PR #136, opened 2025-01-15
mpc contract: remove limits on signature requests & add pytest for updating contract

### LINKED ISSUE #107: Test the upgrade functionality of smart contract
The mpc smart contract allows node operators to propose a update to the smart contract https://github.com/near/mpc/blob/79ec50759146221e7ad8bb04520f13333b75ca07/chain-signatures/contract/src/lib.rs#L524 and vote on it. We should test to make sure it works. The testing should include:
* Unit test
* Testing on localnet
* Testing on testnet

The goal is to verify that it works without any problems, which would allow us to lock the contract.

### LINKED ISSUE #117: Increase number of requests allowed by chain signatures contract
In the contract:
* There is a hard limit on the number of concurrent pending requests [here](https://github.com/near/mpc/blob/79ec50759146221e7ad8bb04520f13333b75ca07/chain-signatures/contract/src/lib.rs#L163)
* A soft limit imposed by scaling costs [here](https://github.com/near/mpc/blob/79ec50759146221e7ad8bb04520f13333b75ca07/chain-signatures/contract/src/lib.rs#L228-L242)

We should change both of these now that we can support more requests.

### LINKED ISSUE #140: Remove charging for signing requests in the smart contract
We should remove charging users in `sign` https://github.com/near/mpc/blob/79ec50759146221e7ad8bb04520f13333b75ca07/chain-signatures/contract/src/lib.rs#L127. Gas cost alone is sufficient to deter spamming.

kevindeforth:
resolves #117:
- introduces a new contract state without the `request_counter` and superfluous config variables
- removes soft and hard limit on the number of concurrent requests

resolves #107:
- introduces a pytest which tests updating the contract from V0 to V1

resolves #140:
- signatures require a deposit of one yoctonear (smallest possible unit)

additionally:
- updates the k256 crate to the [latest audited release](https://crates.io/crates/k256/0.13.4) (includes a security fix)
- adds helper structures for the pytest

**Please note:** 
I was quite conservative changing the contract state, as I am unsure about why certain state was included. 
1. Could we remove `pending_requests` altogether?
2. Currently, the threshold gets defined by calling the function [`init_running()`](https://github.com/Near-One/mpc/blob/fdfca033836d65cabb7d824934ee918dc321b89d/libs/chain-signatures/contract/src/lib.rs#L663). It might make sense to add a `Config` struct in V1, which holds the threshold value and allow for changing the threshold by changing the config. 

edit:
the neard binary from the S3 bucket seems to be an issue for ci right now.
The pytest can be run locally by executing the shell script `pytest/exec_pytest.sh`

edit 2:
The problem was that the logic for extracting the branch name of  the nearcore submodule stopped working.
Right now, the branch is hardcoded to "master" and we are relying on the commit hash to fetch the appropriate binary. Would be more than happy if someone has a more robust solution.


bowenwang1996 (commented):
@kevindeforth we need to be able to upgrade the contract, so I think it makes sense to implement a new struct and put it under `VersionedMpcContract::V1`. Specifically, since you remove `request_counter`, the new contract code will fail to deserialize existing state, which would lead to the contract being completely broken. There should also be a test that test upgrading to the new version

kevindeforth:
> @kevindeforth we need to be able to upgrade the contract, so I think it makes sense to implement a new struct and put it under `VersionedMpcContract::V1`. Specifically, since you remove `request_counter`, the new contract code will fail to deserialize existing state, which would lead to the contract being completely broken. There should also be a test that test upgrading to the new version

thanks for the feedback and the help debugging. I added the struct and pytest.
Please not my updated comment above, specifically about the V1 struct.

pytest/index_signature_request.py
```diff
@@ -1,298 +0,0 @@
...
-        # mpc-node produces way too much output if we run with debug logs
-        nodes[i].run_cmd(cmd=cmd,
-                         extra_env={
-                             'RUST_LOG': 'INFO',
-                             'MPC_SECRET_STORE_KEY': secret_key_hex(i),
-                             'MPC_P2P_PRIVATE_KEY': p2p_private_key(i),
-                             'MPC_ACCOUNT_SK': near_secret_key(i),
-                         })
-
-    return nodes
```
kevindeforth, line 198: this was moved to `common_lib/shared.py` and renamend: 
https://github.com/Near-One/mpc/blob/5bd9f412c571e44751e8c11c387ccb752cd1464a/pytest/common_lib/shared.py#L248

pytest/index_signature_request.py
```diff
@@ -1,298 +0,0 @@
...
-            signature = json.loads(signature)
-            print("SUCCESS! Signature:", signature)
-        except Exception as e:
-            print("Failed to get signature:", e)
-            print("Response:", res)
-            assert False
-
-    res2 = metrics2.get_int_metric_value('mpc_num_sign_responses_timed_out')
-    res3 = metrics2.get_int_metric_value('mpc_num_sign_responses_timed_out')
-    print("Number of nonce conflicts which occurred:", res2, res3)
```
kevindeforth, line 284: this part was moved to `common_lib/shared.py` and is now the following method of the `MpcCluster` helper class.

https://github.com/Near-One/mpc/blob/5bd9f412c571e44751e8c11c387ccb752cd1464a/pytest/common_lib/shared.py#L105

bowenwang1996 (approved):
Looks like there is some code duplication between v0 and v1 that can be refactored. We can do that in a subsequent pull request.

bowenwang1996:
> the neard binary from the S3 bucket seems to be an issue for ci right now.

@andrei-near could you take a look?

pytest/tests/test_signature_request.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python3
+"""
+Starts 2 near validators and 2 mpc nodes.
+Deploys mpc contract and sends a signature request.
+Verifies that the mpc nodes index the signature request.
```
saketh-are, line 5: This test covers more than just indexing now; we could update this comment as well as the function name below.

pytest/tests/test_contract_update.py
```diff
@@ -0,0 +1,39 @@
...
+    contract_v0 = contracts.load_mpc_contract_v0()
+    cluster = shared.start_cluster_with_mpc(2, 2, 1, contract_v0)
+    # assert correct contract is deployed
+    cluster.assert_is_deployed(contract_v0)
+    # do some requests
+    cluster.send_and_await_signature_requests(2)
+    # propose v1
+    contract_v1 = contracts.load_mpc_contract_v1()
+    cluster.propose_update(contract_v1)
+    cluster.vote_update(2, 0)
```
saketh-are, line 29: What happens if signature requests arrive while voting for an update is ongoing? Mainly want to be sure it doesn't break the upgrade process.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -222,6 +257,7 @@ impl VersionedMpcContract {
         const CHEAP_REQUESTS: u32 = 3;
         let pending_requests = match self {
             Self::V0(mpc_contract) => mpc_contract.request_counter,
+            Self::V1(_) => 0,
```
saketh-are, line 260: This is a bit convoluted, maybe just `return U128::from(1)` immediately here instead of faking `pending_requests` to get the behavior we want out of the cost function below.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -287,6 +323,19 @@ impl VersionedMpcContract {
                         Err(InvalidParameters::RequestNotFound.into())
                     }
                 }
+                Self::V1(mpc_contract) => {
```
saketh-are, line 326: It seems all of this logic is identical to V0 above. Maybe you could add a `VersionedMpcContract::get_pending_request(...) -> Option<YieldIndex>` ([like this pattern used in nearcore](https://github.com/near/nearcore/blob/8097202bfa56dba789a4f7fcdb680c2cdaf258b2/core/primitives/src/transaction.rs#L81-L87)) and avoid duplicating the rest of the code.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -654,19 +699,33 @@ impl VersionedMpcContract {
...
             Self::V0(mpc_contract) => &mpc_contract.protocol_state,
+            Self::V1(mpc_contract) => &mpc_contract.protocol_state,
         }
     }
 
     pub fn config(&self) -> &Config {
         match self {
             Self::V0(mpc_contract) => &mpc_contract.config,
+            Self::V1(_) => {
+                panic!("no config")
```
saketh-are, line 727: Perhaps for clarity "V1 contract does not contain a config"

saketh-are (approved):
Thanks @kevindeforth!

pytest/tests/test_contract_update.py
bowenwang1996: That is fine. It will still be processed normally. The upgrade only happens when there are enough votes, which immediately sends a promise to deploy the contract code

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -809,14 +792,7 @@ impl VersionedMpcContract {
         #[callback_unwrap] signature: SignatureResult<SignatureResponse, SignaturePromiseError>,
     ) -> Result<SignatureResponse, Error> {
         match self {
```
saketh-are, line 794: What is the purpose of the outer `match self` here?

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -733,71 +773,14 @@ impl VersionedMpcContract {
...
-                    env::current_account_id(),
-                    "return_signature_on_finish",
-                    &[],
-                    NearToken::from_near(0),
-                    RETURN_SIGNATURE_ON_FINISH_CALL_GAS,
-                );
-                // The return value for this function call will be the value
-                // returned by the `sign_on_finish` callback.
-                env::promise_return(final_yield_promise);
+                handle_signature_state(mpc_contract, contract_signature_request)
```
saketh-are, line 783: I think the behavior here would be clearer if you had something like:
`env::promise_return(update_contract_state_and_construct_signature_promise(mpc_contract, contract_signature_request));`

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -724,7 +764,7 @@ impl VersionedMpcContract {
         match self {
             Self::V0(mpc_contract) => &mpc_contract.config,
             Self::V1(_) => {
-                panic!("no config")
+                panic!("V1 contract des not contain a config")
```
saketh-are, line 767: nit: does

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -62,19 +62,76 @@ impl Default for VersionedMpcContract {
     }
 }
 
+pub trait SignatureStateHandler {
```
saketh-are, line 65: It would be more elegant if `remove_request` were part of this trait alongside add and get.

saketh-are (approved):

libs/chain-signatures/contract/src/lib.rs
kevindeforth: I believe the original author of the contract intended to emphasize that this function is part of the V0 logic and wanted to ensure its behavior does not accidentally get changed.

Because this function is only called via `env::promise_then(..., "return_signature_on_finish", ...)`, searching for references within an IDE might not yield any results and the V0s dependency  on it might get easily overlooked.

Personally, I found this explicit style helpful for familiarizing myself with the contract code, which is why I chose to keep it here.

Issue #135, opened 2025-01-14
Allow mpc nodes to configure multiple access keys for their near account


saketh-are:
Currently we configure a single access key for the near account used to submit signature responses.

Although we can detect nonce conflicts (#113) and resubmit txs as needed, resubmission adds significant delay.

By rotating over multiple access keys when submitting responses we can lessen the incidence of conflicts to begin with. 

Per observations from the load testing tool, a single access key be used to sign around 0.7 transactions per second. If we want to support 300 responses / second across a system of 8 nodes, we should give each node 300 / 8 / .7 ~= 54+ access keys.

PR #139, opened 2025-01-17
use multiple access keys to prevent nonce conflicts

saketh-are:
Implements #135.

Based on observations from the synthetic benchmark tooling, a node should use at least 1.3N keys if it wants to send N tx/s without nonce conflicts. 50 keys each on 8 nodes should be good for 300 sign requests per second. 

Given the large number of keys I chose to use a config file rather than pass them all in via an environment variable as we do for other secrets. The respond account doesn't have any special privileges so risks are lower here. Not sure exactly what would be best practice though, so open to suggestions.

The pytest gains a flag configuring the number of access keys. I confirmed via logs and via near-cli inspection that nodes spread their txs across their access keys as expected. I'm not sure I've actually managed to produce a nonce conflict in localnet even when using only one access key; I think the network setup is too simple and consistent to see reordering.

### DIFF
--- a/node/src/cli.rs
+++ b/node/src/cli.rs
@@ -1,6 +1,6 @@
 use crate::config::{
-    load_config_file, ConfigFile, IndexerConfig, PresignatureConfig, SignatureConfig, SyncMode,
-    TripleConfig, WebUIConfig,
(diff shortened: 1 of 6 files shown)


kevindeforth (approved):
thank you!

PR #148, opened 2025-01-23
Support keyshare storage in GCP and local, natively.

robin-near:
This introduces KeyshareStorage interface, with two implementations, one for local encrypted on disk, one for gcp.

They are written in a way that supports keygen and resharing. In the next PR or two, I'll be making keygen and resharing automatic, and the new key would be written via this interface. For now, keygen is still a separate command and resharing is not implemented.

We use the gcp-cloud client library which seems much better maintained than the google-secretmanager1 library used by the old MPC and has a much better interface.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -528,6 +528,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-compression"
+version = "0.4.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


bowenwang1996 (approved):
Did you test it on gcp?

robin-near:
@bowenwang1996 Testing it. Will make sure it doesn't break.

Issue #155, opened 2025-01-24
Separate the indexer to a separate process


robin-near:

bowenwang1996:
I assume that this is not something we plan to do in the short term?

netrome:
Currently, @pugachAG is investigating how to improve the indexer interface on the near core side. Depending on what the design ends up being, we'll need to re-scope this issue accordingly and clarify what exactly we'd like to achieve with this.

netrome:
@kuksag that's great to know. Can you tell me more about your needs?

netrome:
Oops, yeah I meant https://github.com/near/mpc/issues/592

PR #157, opened 2025-01-25
Preparation for key resharing: completely refactor top level flow.

robin-near:
This is a gigantic PR, but for good reason.

* Rework the top-level flow so that rather than fragmenting the CLI into "start", "keygen" commands, have a single one that reads the contract state to determine what to do. This is critical for key resharing, as resharing is triggered by the contract and not manually.
  * The high-level logic for monitoring the contract state changes and then dispatch tasks to do what we're supposed to do (keygen, signing, resharing) is in `coordinator.rs`. Each time, we create an `MpcJob` that when run, creates a new tokio Runtime. When exiting out of that job due to a contract state change, we kill the Runtime, which effectively stops all tasks running under the job and ensures a clean shutdown.
    * The previous need to wait for a port to be cleaned up during tests is no longer needed; the Runtime shutdown mechanism appears very solid.
  * Key resharing is not yet implemented but a placeholder is clearly carved out for it.
* Because of the new strong requirement for the contract's existence, the indexer is no longer optional. However, this PR abstracts out the indexer using three channels (contract state updates, signature requests, txns to send back to chain). There is a *real indexer* implementation and a *fake indexer* implementation.
  * This also means that the test web endpoints are no longer necessary. Rather, the fake indexer exposes a testing API to submit signature requests and wait for signature responses.
  * This allows us to #155 at some point.
* Because of both the above, testing is now cleaner and more powerful.
  * The Rust integration tests are now much cleaner and more reliable, and the logic they exercise is much closer to production.
  * The Python integration test no longer needs to run the keygen step separately, or need to call the hacky `init_running()`. It now exercises the full end-to-end logic without any invasion.
  * It also means that we no longer need to bootstrap a cluster with the old MPC node, as the new node can start by running key generation automatically, storing secrets, and call vote_pk, and automatically transition to the Running state.

This PR also makes some minor changes to the code which either are required by or make sense together with the larger changes:
* Make the signing code select either the "owner signer" or a "response signer" depending on what transaction it is signing:
  * The owner signer corresponds to the MPC node owner's account. The secret key comes from a high security source. It is necessary to use this to sign vote_pk, vote_reshared, join, etc.
  * The response signer (introduced in #139 ) comes from a pool of many access keys that correspond to arbitrary accounts. These keys come from a low-security source (respond.yaml), but allow more transactions to be sent at the same time without introducing nonce conflicts.
* Fix a network layer bug that a passive computation may be started when we do not yet have all outgoing connections established which triggers a "current connection is not the one expected" error. We now properly handle this case. Also optimize the code for a `watch::Receiver` usage.
* Delete obsolete integration-test docker files and obsolete test_configs directory.


node/src/coordinator.rs
```diff
@@ -0,0 +1,409 @@
...
+                            ),
+                        )?,
+                        stop_fn: Box::new(move |new_state| match new_state {
+                            ContractState::Resharing(new_state) => {
+                                new_state.old_epoch != state.old_epoch
+                                    || new_state.new_participants != state.new_participants
+                            }
+                            _ => true,
+                        }),
+                        timeout_fut: sleep(
```
bowenwang1996, line 170: What happens if there is a fork and the node switches back and force between Resharing and Running? Would that cause some potential inconsistencies? Should we consider using final state here?

bowenwang1996 (approved):

node/src/coordinator.rs
robin-near: Ah, we actually already query the final state. The contract state is proactively queried (once a second) by sending a message to the `ViewClientActor` in `get_mpc_contract_state`, and we specify Final for the finality.

PR #158, opened 2025-01-27
add metrics for passive sign requests from peers

saketh-are:
Adds some metrics giving insight on the passive requests the node receives from mpc peers.

While load testing with a burst of 100 requests, I could observe a few instances of failed lookups. Increasing the channel size fixes the issue for now. In the long term we may rewrite the algorithm there altogether.


robin-near:
What's the submodule change?

saketh-are:
> What's the submodule change?

Committed that by mistake, removed

robin-near (approved):

PR #159, opened 2025-01-28
Mpc terraform config

peter-near:


bowenwang1996:
@peter-near is it still relevant? if not, let's close it

PR #162, opened 2025-01-29
add a test making parallel sign calls using promise::and

### LINKED ISSUE #143: Allow parallel requests that preserve order
Requests from some developers and partners: it would be nice to support parallel requests to `sign` that return in the same order. Let's say that there are 4 calls to `sign` and the results should be an array that preserves the order with which they are submitted. I think the best option is to support `Promise::and` and then the caller can retrieve promise results based on promise indices.

saketh-are:
Closes #143


pytest/tests/test_contracts/parallel/res/contract.wasm
kevindeforth, line 1: It could be beneficial to add a pytest fixture that builds this contract before running the tests.
```python
@pytest.fixture(scope="session", autouse=True)
def compile_parallel_contract():
...
```

pytest/tests/test_parallel_sign_calls.py
```diff
@@ -0,0 +1,54 @@
...
+
+def load_parallel_sign_contract() -> bytearray:
+    """
+    returns test contract for parallel sign
+    """
+    path = constants.mpc_repo_dir / 'pytest/tests/test_contracts/parallel/res/contract.wasm'
+    return load_binary_file(path)
+
+
+def test_parallel_sign_calls():
```
kevindeforth, line 29: This test could be parametrized for additional flexibility

```python
@pytest.mark.parametrize("num_requests", [5])
def test_parallel_sign_calls(num_requests):
```

pytest/tests/test_contracts/parallel/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+        ))
+    }
+
+    #[private]
+    pub fn handle_results(&self) -> u64 {
+        let num_calls = env::promise_results_count();
+        env::log_str(format!("{num_calls} parallel calls completed!").as_str());
+        for i in 0..num_calls {
+            let result = env::promise_result(i);
+            env::log_str(&format!("sign #{i}: {:?}", result));
```
kevindeforth, line 73: This might be a stupid question, but should there be a check if each result corresponds to a success value?

Asking because the pytest asserts if the return value of this function corresponds to the expected number of signatures.

Of course, my comment is superfluous if chaining promises with `and` ensures that the final return value corresponds to success, only if all of the promises succeed (but I am not sure about that, hence asking).

kevindeforth (approved):
Thank you!

pytest/tests/test_contracts/parallel/src/lib.rs
saketh-are: Good question. I added a check in the contract to make sure that each promise result is of type `PromiseResult::Successful`.

Ideally this test would actually parse out the signature results and verify them so that we know the responses are the expected ones and that they came back in the right order. To achieve that I would hope to import `fn check_ec_signature` from the mpc-contract crate and reuse it. However, importing another smart contract marked with `#[near_bindgen]` creates some conflicts:

```
rust-lld: error: duplicate symbol: contract_source_metadata
          >>> defined in /Users/saketh/mpc/pytest/tests/test_contracts/parallel/target/wasm32-unknown-unknown/release/deps/parallel_contract.parallel_contract.d8b97206910cbf48-cgu.1.rcgu.o
          >>> defined in /Users/saketh/mpc/pytest/tests/test_contracts/parallel/target/wasm32-unknown-unknown/release/deps/libmpc_contract.rlib(mpc_contract.mpc_contract.6ff6e7068261a859-cgu.02.rcgu.o)
```

I think what we need to do is move some of the primitives in `mpc-contract` into a separate crate so that we can depend on them here as well. That would also eliminate the need to redefine the `struct SignRequest`; we could just import it. I'll file an issue for this.

Issue #151, opened 2025-01-24
Ensure that participants can eventually successfully keygen or reshare when there are connectivity issues


robin-near:

robin-near:
copied from slack: https://nearone.slack.com/archives/C07UW93JVQ8/p1738180453522019

When we do key resharing, we start the p2p mesh network, wait for all participants to be connected, then initiate the resharing computation. But the "wait for all participants to be connected" is actually very difficult because being "connected" doesn't mean the node is ready to perform the computation.
A concrete example is that 5 nodes are currently running in normal state. Then a key resharing is started and each node realizes it needs to restart the mesh network (due to a different set of nodes to connect to). But this is not coordinated between nodes, so node A can restart and try to connect to other nodes (for a key resharing), while node B is just realizing that it needs to do a resharing. Node A then ends up connecting to node B's original network listener. Let's say C, D, E are all like node A. Then, what we have is:
 Node A, C, D, E: in resharing state, new network mesh, connected to each other and B
 Node B: in running state, old network mesh, accepted a connection from A, C, D, E but are just about to shut down the network
Now nodes A, C, D, E have "all participants connected" and so proceed to do a resharing
but right then, node B shuts down the network to enter the resharing state, causing the connections to break and thus resharing fails.
When resharing fails, each node restarts to try resharing again. But before that happens, B managed to connect to all other nodes and starts a resharing. A, C, D, E all restart (due to failed resharing), but that causes the newly started resharing to fail, and B continues to restart.
This pattern continues. Each time, all nodes are connected but some nodes are thinking "Great, time to start resharing" and some other nodes are thinking "oops, resharing fails, time to retry". And then resharing never actually completes.

robin-near:
Will be solved by #195

PR #164, opened 2025-01-29
Implement key resharing.

robin-near:
This PR implement key resharing. It is expected to work out of the box as long as we manually kick the contract into a Resharing state (by sending join and vote_join; leaving is fine too). Some of the code was adapted from #87 .

Rust-side integration tests are included to test resharing behavior including signature buffering during resharing. **Python integration test as well as real cluster test are TODO**.

Along with that are a few more related or necessary changes:
* Make the coordinator always check if the current node is not a participant, and halt in this case (until contract state change) instead of failing (and then looping).
* Make the network layer's `wait_for_ready` wait for that many connections to be *bidirectionally* connected, not just outgoing. This solves a subtle problem where nodes may repeatedly fail to carry out resharding and run into a subtle retry loop. It is also good to have - can possibly avoid hiding network issues where one side cannot accept incoming connections. Further research may be needed to improve the reliability here (#151 ).
  * Related, also make keygen and resharing loop until receiving the expected keygen/resharing task from the network; this can probably avoid unnecessary failures where other nodes have not transitioned into resharing state yet.
* Some test code refactoring to better support tests where the participant set changes.
* Add `PortSeed` to better avoid port conflicts in tests.

### DIFF
--- a/node/src/coordinator.rs
+++ b/node/src/coordinator.rs
@@ -4,9 +4,12 @@ use crate::indexer::handler::ChainSignatureRequest;
 use crate::indexer::participants::{
     ContractInitializingState, ContractResharingState, ContractRunningState, ContractState,
 };
-use crate::indexer::response::{ChainSendTransactionRequest, ChainVotePkArgs};
+use crate::indexer::response::{
+    ChainSendTransactionRequest, ChainVotePkArgs, ChainVoteResharedArgs,
+};
 use crate::indexer::IndexerAPI;
 use crate::key_generation::{affine_point_to_public_key, run_key_generation_client};
+use crate::key_resharing::run_key_resharing_client;
 use crate::keyshare::{KeyshareStorage, KeyshareStorageFactory};
 use crate::mpc_client::MpcClient;
 use crate::network::{run_network_client, MeshNetworkTransportSender};
@@ -151,7 +154,6 @@ impl Coordinator {
                             "Resharing",
                             self.config_file.cores,
                             Self::run_key_resharing(
-                                self.clock.clone(),
                                 self.secret_db.clone(),
                                 self.secrets.clone(),
                                 self.config_file.clone(),
@@ -274,10 +276,14 @@ impl Coordinator {
             unreachable!();
         }
 
-        let mpc_config = MpcConfig::from_participants_with_near_account_id(
+        let Some(mpc_config) = MpcConfig::from_participants_with_near_account_id(
             contract_state.participants,
             &config_file.my_near_account_id,
-        )?;
+        ) else {
+            tracing::info!("We are not a participant in the initial candidates list; doing nothing until contract state change");
+            futures::future::pending::<()>().await;
+            unreachable!()
+        };
         let (sender, receiver) =
             new_tls_mesh_network(&mpc_config, &secrets.p2p_private_key).await?;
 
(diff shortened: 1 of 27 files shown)


node/src/p2p.rs
```diff
@@ -582,6 +600,70 @@ fn verify_peer_identity(
     Ok(*peer_id)
 }
 
+/// Struct to track bidirectional connectivity between two nodes.
+#[derive(Clone)]
+struct ConnectivityIndicator {
```
bowenwang1996, line 605: Could you add some more comments to explain how this struct works?

bowenwang1996 (approved):
Great work!

node/src/coordinator.rs
```diff
@@ -385,18 +400,109 @@ impl Coordinator {
...
+            None => {
+                if was_participant_last_epoch {
+                    anyhow::bail!("We were a participant last epoch, but we don't have a keyshare");
+                }
+                None
+            }
+        };
+
+        // Delete all presignatures from the previous epoch; they are no longer usable
+        // once we reshare keys.
```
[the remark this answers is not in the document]
robin-near: I think you said triples don't need to be deleted right?

The triples are not tied to keys, only participant IDs, and the triples carry the list of participant IDs with them, so we can still reuse them.

Well... I guess some triples would be useless so we should clear them out. Yeah. let me send a separate PR for that. Less of a urgent thing because it's only relevant for leaving.

node/src/coordinator.rs
```diff
@@ -385,18 +400,109 @@ impl Coordinator {
...
-        tracing::error!("Key resharing is not implemented yet");
-        futures::future::pending::<()>().await;
-        unreachable!()
+        let Some(mpc_config) = MpcConfig::from_participants_with_near_account_id(
+            contract_state.new_participants.clone(),
+            &config_file.my_near_account_id,
+        ) else {
+            tracing::info!("We are not a participant in the new epoch; doing nothing until contract state change");
+            futures::future::pending::<()>().await;
+            unreachable!()
```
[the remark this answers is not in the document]
robin-near: Yeah I'm curious too. I'm relying on just dropping futures a lot. There's an alternative approach in general of using synchronization primitives to gracefully exit (so we don't drop futures), but that is a lot harder to get right and the code becomes sprinkled with `select!`s everywhere.

Hmm, perhaps for this future, we can just return instead of hanging. It would return a different flag saying that it should hang instead of proceeding to the next iteration. Yeah, that would probably work - I'll give it a try.

Wait, do I sound like DeepThink R1? :joy:

node/src/db.rs
```diff
@@ -134,6 +159,17 @@ impl SecretDBUpdate {
         self.batch.delete_cf(&self.db.cf_handle(col), key);
     }
 
+    pub fn delete_all(&mut self, col: DBCol) -> anyhow::Result<()> {
+        let range = self.db.get_cf_key_range(col)?;
+        if let Some(range) = range {
+            self.batch
+                .delete_range_cf(self.db.cf_handle(col), range.start(), range.end());
+            // delete_range_cf deletes ["begin_key", "end_key"), so need one more delete
```
saketh-are, line 167: It's surprising there isn't a nicer way of doing this. One recommendation I saw is to pass (smallest possible key, largest possible key) but that is not that great either.

saketh-are (approved):

node/src/db.rs
```diff
@@ -115,6 +116,30 @@ impl SecretDB {
...
+    fn get_cf_key_range(
+        &self,
+        col: DBCol,
+    ) -> anyhow::Result<Option<std::ops::RangeInclusive<Box<[u8]>>>> {
+        let range = {
+            let mut iter = self
+                .db
+                .iterator_cf(self.cf_handle(col), IteratorMode::Start);
+            let start = iter.next().transpose()?;
+            iter.set_mode(IteratorMode::End);
```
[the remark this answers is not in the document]
robin-near: Oh, this code is copied from nearcore.

node/src/db.rs
robin-near: Same here, copied from nearcore. I'm surprised too....

robin-near: The problem is the smallest key is b"", but there does not exist a largest possible key.

robin-near:
Lol I think when I deal with the futures thing, I'm uncovering a whole rabbit whole of issues. So this will take a little longer.

robin-near:
ok so there's some subtlety with regards to what to do when a node is not actively participating in the MPC network. If the node is not a participant, it should not be buffering the signatures in its channel. Otherwise, this would blow up the memory and when the node joins and becomes a participant, it would be retroactively trying to process all the signature requests and that is not desirable. Added some code to properly deal with that in each scenario along with addressing the `futures::future::pending` hack.

Also, earlier the test that I was doing to assert that I got a signature response was not sufficient; now I check that the response corresponds to the expected signature request. What happens is that because computation can happen before network is fully connected, when a node receives a signature request and it is the secondary leader, it might start a computation for it because it is not yet connected to the primary leader. So in tests we can often see two responses for the same signature request. I've changed it to properly wait for the correct signature response, and that fixes a flake for the faulty cluster test as well.

PR #168, opened 2025-02-01
Reduce mpc contract latency

### LINKED ISSUE #131: Optimize chain signatures contract for latency
In the chain signatures contract, the initial sign call makes a call into sign_helper:
https://github.com/Near-One/mpc/blob/51d5792aafbc47322cf95b0f8e3d0e9da5cd3eec/libs/chain-signatures/contract/src/lib.rs#L175

And the final `return_signature_on_finish` is chained after another promise `clear_state_on_finish`:
https://github.com/Near-One/mpc/blob/51d5792aafbc47322cf95b0f8e3d0e9da5cd3eec/libs/chain-signatures/contract/src/lib.rs#L698-L707

In both cases chaining function calls adds one block of latency. If we can change this we can cut down from the current 5-6 blocks of latency between a signature request and the final response.

kevindeforth:
**Resolves #131 :**
- add contract binaries to compiled-contracts
- add pytest fixtures for contract compilation
- improve pytests for contract update
- add a pytest for state cleanup
- changes the gas requirements for sign call from 50 teragas to 15

**Benchmarks:**
For 40 signatures:
| Contract  | avg. receipts | avg. gas [Tgas] |
| ------------- | ------------- | ------------- |
| V0  | 8  |11.30479597562405|
| V1  | 4  |6.131075775468398 |

**Contract Logic changes:**
Each row is a promise, ignore empty rows.

| V0  | V1 |
| ------------- | ------------- |
 do some checks | do some checks, add request to state, schedule yield promise
add request to state, schedule yield promise | 
 remove signature request from state | clean state & return signature or error 
 return signature or error | 

**Additional Contract State:**
Because of the new logic, failed signature requests need to be periodically removed from the state.
This requires additional contract state:
```Rust
    request_by_timestamp: Vector<(u64, SignatureRequest)>,
    config: ConfigV1,
```
`request_by_timestamp` is iterable and stores the `env::block_height()` during which the request was added to the state.
`config` contains is the timeout after which signature requests can be removed from the contract, it also defines the maximum number of requests that can be removed during a call to `sign()`, `respond()` or `remove_timed_out_requests()`.

**Note:**
The signature of the `sign` function changes with this PR, specifically, the return value is no longer `Result<near_sdk::Promise, Error>`, but all Errors result in `panic`.

edit:
**CI time:**
The new pytest takes a long time and increases the ci time to 40 minutes, which seems quite long. This is expected to increase as we add more tests. If reviewers agree, I would like to restrict the ci to `pytest -m "not slow"`, skipping all tests that are marked as slow via `@pytest.mark.slow`
Alternatively, we could add a marker `ci` to annotate tests that should be included in ci, ignoring all tests that do not have that marker.


libs/chain-signatures/contract/src/lib.rs
```diff
@@ -364,14 +437,21 @@ impl VersionedMpcContract {
...
+            if let Some(YieldIndex { data_id }) = self.get_pending_request(&request) {
+                // Then resolve the promise. This will have no effect if the request already timed
+                // out by the blockchain.
+                env::promise_yield_resume(&data_id, &serde_json::to_vec(&response).unwrap());
+                // Only then clean up the state.
+                // This order of execution ensures that the state is cleaned of the current
+                // response, even if it belongs to an already timed out signature request.
+                match self {
+                    Self::V0(_) | Self::V1(_) => {}
+                    Self::V2(mpc_contract) => mpc_contract.remove_timed_out_requests(),
```
saketh-are, line 450: I think we may need to be careful about gas cost predictability here. If enough timeout requests accumulate, any respond call made afterwards would attempt to clear all of them and run out of gas. If that happens the transaction as a whole fails and none of the requests are cleared, causing the system to get stuck. At the same time we don't want to over-attach gas to respond calls since it limits throughput.

One idea to prevent this would be to limit the number of timed out requests removed per call to some reasonable number.

saketh-are: Also maybe it would be nicer to trigger the cleanup from "sign" calls rather than "respond" calls? Then we don't have the possibility that state accumulates without limit if the signer network happens to be down. Of course ideally that won't ever happen...

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -62,70 +70,82 @@ impl Default for VersionedMpcContract {
...
-    response: &SignatureResponse,
-) -> Result<(), Error> {
-    if let Some(YieldIndex { data_id }) = mpc_contract.get_pending_request(request) {
-        env::promise_yield_resume(&data_id, &serde_json::to_vec(&response).unwrap());
-        Ok(())
-    } else {
-        Err(InvalidParameters::RequestNotFound.into())
+impl MpcContractV2 {
+    fn remove_timed_out_requests(&mut self) {
+        let deadline = env::block_timestamp_ms() - self.request_timeout;
```
saketh-are, line 84: The yielded promise created by the sign call [times out after 200 blocks](https://github.com/near/nearcore/blob/b1bb76ce2e4dd53159c04830ef69e86d4663281e/core/parameters/res/runtime_configs/parameters.snap#L197). You could handle the timeouts accordingly using `env::block_height` instead of `env::block_timestamp_ms`.

bowenwang1996 (commented):
Meta comment: given that V1 is not live yet, let's not introduce V2 unnecessarily and collapse changes in V2 into V1

libs/chain-signatures/contract/src/lib.rs
kevindeforth: cleanup is now triggered in `sign()` and `respond()`. I also added a function that can be called to trigger the cleanup separately.

libs/chain-signatures/contract/src/config/impls.rs
```diff
@@ -12,6 +13,35 @@ const MAX_EXPECTED_PARTICIPANTS: u32 = 32;
 /// that should be in the network.
 const NETWORK_MULTIPLIER: u32 = 128;
 
+// Default delay of five minutes, after that, request is removed from the contract state
+const DEFAULT_REQUEST_TIMEOUT_BLOCKS: u64 = 200;
+
+// The maximum number of requests to remove during a call
+const MAX_NUM_REQUESTS_TO_REMOVE: u32 = 100;
```
saketh-are, line 20: Under the `GAS_FOR_SIGN_CALL` limit we need to account for the worst-case cost of removing the maximum number of requests. It would be helpful to have some estimate documented of how much that costs, especially if other parts of the contract change and `GAS_FOR_SIGN_CALL` needs to be readjusted in the future.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -62,93 +66,66 @@ impl Default for VersionedMpcContract {
...
+    request_by_timestamp: Vector<(u64, SignatureRequest)>,
     proposed_updates: ProposedUpdates,
+    config: ConfigV1,
 }
 
-impl SignatureStateHandler for MpcContractV1 {
+impl MpcContractV1 {
+    fn remove_timed_out_requests(&mut self) -> u32 {
+        let max_to_remove = self.config.max_num_requests_to_remove;
+        let deadline = cmp::max(env::block_height(), self.config.request_timeout_blocks)
```
saketh-are, line 81: `deadline` is fairly ambiguous, maybe something like `minimum_pending_request_height`?

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -202,71 +177,145 @@ impl MpcContract {
...
     #[handle_result]
     #[payable]
-    pub fn sign(&mut self, request: SignRequest) -> Result<near_sdk::Promise, Error> {
+    pub fn sign(&mut self, request: SignRequest) {
         let SignRequest {
             payload,
             path,
             key_version,
         } = request;
+        // First, clear the state and check if there is space in the queue:
```
saketh-are, line 206: Possibly outdated comment, doesn't seem like we check for space in this part

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -364,14 +413,22 @@ impl VersionedMpcContract {
...
-                    handle_signature_response(mpc_contract, &request, &response)
+            // First get the yield promise of the (potentially timed out) request.
+            if let Some(YieldIndex { data_id }) = self.get_pending_request(&request) {
+                // Only then clean up the state.
+                // This order of execution ensures that the state is cleaned of the current
+                // response, even if it belongs to an already timed out signature request.
+                match self {
+                    Self::V0(_) => {}
+                    Self::V1(mpc_contract) => {
+                        mpc_contract.remove_timed_out_requests();
```
saketh-are, line 424: Should this be removed now that `sign` takes the responsibility for cleanup?

libs/chain-signatures/contract/src/config/impls.rs
kevindeforth: Good point! Calling the `remove_timed_out_requests` endpoint and removing 100 signatures from the state requires approx. `37 Tgas`.

Calling `sign` and removing 100 signatures costs approximately `40 Tgas`. I left the the previous gas requirement of 50 TGAS in place to have additional margin and added tests that ensure a signature call does not exceed that limit.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -17,33 +18,36 @@ use k256::Scalar;
...
 
 use crate::config::Config;
 use crate::errors::Error;
 use crate::update::{ProposeUpdateArgs, ProposedUpdates, UpdateId};
-
 pub use state::{
     InitializingContractState, ProtocolContractState, ResharingContractState, RunningContractState,
 };
-
 const GAS_FOR_SIGN_CALL: Gas = Gas::from_tgas(50);
```
bowenwang1996, line 39: Why does this need to be 50Tgas? Let's reduce it for V1

libs/chain-signatures/contract/src/config/impls.rs
bowenwang1996: @kevindeforth no we should not require 50Tgas for each sign call because it means that when `sign` is called from other smart contract, they need to attach at least 50Tgas to the promise and that is too high and limits what the caller contract can do. We should make it as low as possible. One way to do it is to remove 1 request from the state state when `sign` is called.

kevindeforth: Understood. The following is the behavior as of the latest commit:
- Gas requirement for sign call is 10 Tgas.
- Default for `config.max_num_requests_to_remove` is set to 1.
- In the pytests, we now attach only 10 TGas to the sign transactions (as opposed to the previous 150), to reflect how users are expected to interact with it.
- The `remove_timed_out_requests` contract endpoint takes an additional argument `max_num_to_remove: Option<u32>`. The function serves as a back-up in case we need to manually clear the state.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -62,93 +66,66 @@ impl Default for VersionedMpcContract {
...
+    fn remove_timed_out_requests(&mut self, max_num_to_remove: u32) -> u32 {
+        let min_pending_request_height =
+            cmp::max(env::block_height(), self.config.request_timeout_blocks)
+                - self.config.request_timeout_blocks;
+        let mut i = 0;
+        for x in self.request_by_block_height.iter() {
+            if (min_pending_request_height <= x.0) || (i > max_num_to_remove) {
+                break;
+            }
+            let _ = self.pending_requests.remove(&x.1);
```
bowenwang1996, line 88:
```suggestion
            self.pending_requests.remove(&x.1);
```

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -62,93 +66,66 @@ impl Default for VersionedMpcContract {
...
+            i += 1;
+        }
+        let _ = self.request_by_block_height.drain(..i);
```
bowenwang1996, line 91:
```suggestion
        self.request_by_block_height.drain(..i);
```

bowenwang1996 (approved):

pytest/common_lib/constants.py
```diff
@@ -1,9 +1,14 @@
...
+MPC_BINARY_PATH = os.path.join(MPC_REPO_DIR / 'target' / 'release', 'mpc-node')
+CONFIG_PATH = os.path.join(MPC_REPO_DIR / 'pytest' / 'config.json')
 
 TIMEOUT = 60
 NEAR_BASE = 10**24
 TGAS = 10**12
+# Tgas required by the contract for a sign call.
+GAS_FOR_SIGN_CALL = 10
+# maximum number of requests that can be removed during a sign / request call.
+MAX_NUM_REQUSTS_TO_REMOVE = 100
```
saketh-are, line 14: Should we lower this to match the expected usage? Also nit: REQUESTS

pytest/tests/test_contract_update.py
```diff
@@ -7,30 +7,119 @@
...
+    else:
+        print("ensuring config is updated")
+        expected_config = update_args.dump_json()
+        deployed_config = cluster.get_config()
+        assert deployed_config == expected_config
+    print("update completed")
+    cluster.send_and_await_signature_requests(2, add_gas=150 * TGAS)
+
+
+# In case a nonce conflict occurs during a vote_update call, rerun the test once.
```
saketh-are, line 64: In my previous testing I was not able to produce nonce conflicts in localnet even under significant load (10s or 100s of txs). I was running the other pytest test_signature_request with high num_requests and didn't see any requests fail to be included. I believe this is because the local setup is too simple for transaction reordering to happen. So I wonder if there is some other underlying cause for flakiness observed here.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -823,56 +876,88 @@ impl VersionedMpcContract {
...
     }
 
+    /// Upon success, removes the signature from state and returns it.
+    /// Returns an Error if the signature timed out.
+    /// Note that timed out signatures will need to be cleaned up from the state by a different function.
+    #[private]
+    #[handle_result]
+    pub fn return_signature_and_clean_state_on_success(
+        &mut self,
+        contract_signature_request: ContractSignatureRequest,
```
saketh-are, line 886: Wait so why is it that we cannot do the cleanup here in the timeout case? This argument is the only one needed for the cleanup.

pytest/tests/test_contract_update.py
kevindeforth: In the timespan between the rpc request to fetch the current nonce value and until the update transaction is sent, the mpc participant already submitted other transactions and the nonce increased.

This improved significantly after using release version of near core, but I still get the error every now an then.

Still learning about keys and accounts. Does each access key have its own nonce and would it help to use a different access key for the update calls?

libs/chain-signatures/contract/src/lib.rs
kevindeforth: If my assumption is correct, we could still do the cleanup here, but would need to return `Ok(<SignatuseResult, Error>)` and then have the caller handle it. But I think this would break backwards-compatibility.

saketh-are: Ohh, I see, thanks!

saketh-are (approved):
Thanks @kevindeforth!

PR #170, opened 2025-02-03
Key Derivation Function for Frost

kuksag:


SimonRastikian:
I am generally not a big fan of mixing functions created for ECDSA implementation and functions for EdDSA in the same file without clear separation for *both* sets of functions

kuksag:
Will be merged with #194

PR #172, opened 2025-02-03
Frost Protocol Wrapper

### LINKED ISSUE #119: Support frost signatures
@robin-near @bowenwang1996

Here's the plan:

kuksag:


node/src/frost_sign.rs
```diff
@@ -0,0 +1,576 @@
+use std::collections::btree_map::Iter;
+use crate::primitives::ParticipantId;
+use aes_gcm::aead::rand_core::{CryptoRng, RngCore};
+use anyhow::Context;
+use borsh::{BorshDeserialize, BorshSerialize};
+use cait_sith::protocol::{Action, MessageData, Participant, Protocol, ProtocolError};
+use std::collections::BTreeMap;
+
+/// This implementation is valid in the context where we can choose a Leader (Coordinator)
```
robin-near, line 9: @bowenwang1996 you said something about this the last time I mentioned that the cait-sith signature scheme can be optimized to have one party generate a signature rather than all. I forgot what the problem was - do you remember?

node/src/frost_sign.rs
```diff
@@ -0,0 +1,576 @@
...
+            my_commitment,
+            coordinator: None,
+            signing_package: None,
+        }
+    }
+}
+
+impl FrostSignProtocolParticipant {
+    fn handle_ready_to_send_commitment_state(&mut self) -> anyhow::Result<Action<FrostOutput>> {
+        let coordinator = if let Some(coordinator) = self.coordinator {
```
robin-near, line 334: this and below: use the syntax
```
let Some(coordinator) = self.coordinator else {
  return Ok(Action::Wait);
};
```

node/src/frost_sign.rs
```diff
@@ -0,0 +1,576 @@
...
+        let non_active_participants = 0;
+        let coordinators = 1;
+
+        let protocols = build_protocols(
+            max_signers,
+            min_signers,
+            non_active_participants,
+            coordinators,
+        );
+        run_protocol(protocols).unwrap();
```
robin-near, line 530: Is it possible to sanity check the returned results? (One is a coordinator return result, and the others are participants or something like that?)

node/src/frost_sign.rs
```diff
@@ -0,0 +1,576 @@
...
+/// TODO: Can we precompute the first round in advance (just like presigns in ECDSA)?
+///     The answer depends on whether we can do KDF on SigningPackage/SigningShare.
+
+#[derive(Debug)]
+pub enum FrostOutput {
+    Coordinator(frost_ed25519::Signature),
+    Participant,
+}
+
+pub fn frost_sign<RNG: CryptoRng + RngCore + 'static>(
```
robin-near, line 24: Here and below: could you sprinkle some documentation so the reader of this code can more easily follow it? Mostly I'm looking for documentation for important methods whose behavior is not immediately clear, structures fields whose meaning is not immediately clear, etc.

node/src/frost_sign.rs
```diff
@@ -0,0 +1,576 @@
...
+                    self.signing_package = Some(signing_package);
+                } else {
+                    self.state = ParticipantState::Finished
+                }
+            }
+        }
+    }
+}
+
+impl Protocol for FrostSignProtocolParticipant {
```
robin-near, line 411: Maybe add a high level summary of what we're doing here, i.e. adapting the frost protocol to the API of the cait-sith protocol.

node/src/frost_sign.rs
```diff
@@ -0,0 +1,576 @@
...
+                    {
+                        self.signature_shares.add(from, data);
+                    }
+                }
+            }
+        }
+    }
+}
+
+impl FrostSignProtocolCoordinator {
```
robin-near, line 200: merge this impl block with the above?

node/src/frost_sign.rs
```diff
@@ -0,0 +1,576 @@
...
+            CoordinatorMessage::FirstRoundInit => {
+                if self.coordinator.is_none() {
+                    self.coordinator = Some(from)
+                }
+            }
+            CoordinatorMessage::SigningPackage(signing_package) => {
+                if Some(from) == self.coordinator {
+                    self.signing_package = Some(signing_package);
+                } else {
+                    self.state = ParticipantState::Finished
```
robin-near, line 404: I'm having trouble following the logic in the else block. In what case would we receive a coordinator message that is either not from the coordinator or before FirstRoundInit? Is it if we somehow participated in this protocol late and we missed the init message? Either way I think this is worth a comment.

node/src/frost_sign.rs
bowenwang1996: The leader may not be online. We need a way to handle that

kuksag:
Closed in favor of #194

PR #177, opened 2025-02-04
Fix indexer catch-up time calculation

andrei-near:
The indexer was showing incorrect catch-up time estimates, sometimes millions of years.
This was happening because of integer underflow when calculating blocks behind: when the current block height was lower than the last processed block height, the unsigned subtraction would wrap around to a very large number (close to u64::MAX). Because of async fetch_latest_block, block_height was always smaller than last_processed_block_height:

`INFO indexer: Time to catch up calculation: blocks_behind=18446744073709551610, block_processing_speed=0.100, estimated_seconds=18446744073709551615, last_processed_block_height=36, block_height=30 `

Added a safety check to verify this condition and only log catch-up time if indexer is behind


saketh-are (approved):
Thanks @andrei-near

PR #179, opened 2025-02-05
Run Near Core in release

ashamis:
We see occasional timeouts during contract creation in the GitHub tests and I was seeing timeouts approximately in 1 in 5 runs on my mac. In exec_pytest.sh we are building the MPC code in release and linking it to a debug build of near core.

This change makes sure that we build and use the release version of near core when running the python tests.

After this change I have not observed timeouts in contract creation.


kevindeforth (approved):
nice, thank you!

PR #185, opened 2025-02-07
Add timing metrics for triple, pre-signature, and signature generation

ashamis:
Add timing metrics for triple, pre-signature, and signature generation

### DIFF
--- a/node/src/sign.rs
+++ b/node/src/sign.rs
@@ -14,7 +14,7 @@ use cait_sith::{FullSignature, KeygenOutput, PresignArguments, PresignOutput};
 use k256::{AffinePoint, Scalar, Secp256k1};
 use std::sync::atomic::{AtomicBool, AtomicUsize};
 use std::sync::{Arc, Mutex};
-use std::time::{Duration, SystemTime};
+use std::time::Duration;
 use tokio::time::timeout;
 
 /// Performs an MPC presignature operation. This is shared for the initiator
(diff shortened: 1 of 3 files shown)


node/src/triple.rs
```diff
@@ -36,7 +36,13 @@ pub async fn run_many_triple_generation<const N: usize>(
         me.into(),
         threshold,
     )?;
+
+    let start = SystemTime::now();
+
     let triples = run_protocol("many triple gen", channel, me, protocol).await?;
+
+    let duration = start.elapsed()?.as_millis();
+    metrics::MPC_NUM_TRIPLES_GENERATION_TIME_MS.set(duration as i64);
```
kevindeforth, line 45: I would propose to rename the variable to `MPC_LAST_NUM_TRIPLES_GENERATION_TIME_MS`, to better reflect what it tracks.

Given that below we track the total number of triples generated, it could be beneficial to additionally track the total time spent generating triples

kevindeforth (approved):
Thanks, this will be useful!

nit: was a bit cumbersome to checkout your branch locally, because it lives in a forked repository.

node/src/sign.rs
```diff
@@ -45,7 +45,17 @@ pub async fn pre_sign(
             threshold,
         },
     )?;
+    let start = SystemTime::now();
+
     let presignature = run_protocol("presign", channel, me, protocol).await?;
+
+    let duration = start.elapsed()?.as_millis();
+    metrics::MPC_PRE_SIGNATURE_TIME_MS
```
robin-near, line 53: I would suggest to convert these metrics to buckets instead:
 * A bucket metric would automatically keep track of the count and sum, so you don't need to have instance or cumulative.
 * A bucket metric would store a histogram so we can see what are the percentile stats.
 * The calling code doesn't need to do it twice, and doesn't need to incur the overhead of looking up by string keys.

node/src/sign.rs
```diff
@@ -115,7 +125,17 @@ pub async fn sign(
         presign_out,
         msg_hash,
     )?;
+    let start = SystemTime::now();
+
     let signature = run_protocol("sign", channel, me, protocol).await?;
+
+    let duration = start.elapsed()?.as_millis();
```
robin-near, line 132: There is a RAII guard provided by HistogramTimer (`.start_timer()` on a Histogram, see nearcore repo for examples) that will do this for you.

It will also ensure that in case run_protocol returns an error, the elapsed time will still be counted.

btw, for timing purposes, use Instant instead of SystemTime; Instant is guaranteed to be monotonic.

node/src/sign.rs
robin-near: Sorry I meant Histogram, not bucket.

robin-near:
> nit: was a bit cumbersome to checkout your branch locally, because it lives in a forked repository.

Ah we do this as a standard practice for nearcore actually (and probably for a lot of github projects). My workflow is something like
```
git remote add ashamis https://blahblahblah/...
git fetch ashamis
git checkout ashamis/more_metrics
```
(I always work with detached HEAD though, I never bother with local branches)

PR #187, opened 2025-02-07
retry "respond" if contract state still has pending request

### LINKED ISSUE #153: Re-implement indexing of respond tx
In #139 we changed the way respond txs are signed. We introduced a separate account which can be configured with arbitrarily many access keys.

Due to these changes, the logic for indexing the node's own respond txs became outdated and was removed.

Because use of multiple keys mitigates the issue of nonce conflicts, it is far less likely now that a respond tx will be lost after submission. Hence indexing the successful responses and resubmitting as necessary is lower-priority now, but we may still wish to re-implement it to handle the rare loss during transmission.

saketh-are:
In the past we attempted to index `respond` transactions to decide whether they need to be resubmitted. That approach does not work reliably because the account making the respond call may be on a different shard than the signer contract. The node's indexer is configured to track only the shard containing the signer contract.

In this PR, we instead make a view call on `get_pending_request`. It gets executed locally by the node's indexer on the signer contract's state and lets us know whether the respond call has succeeded.

Closes #153.

### DIFF
--- a/node/src/indexer/handler.rs
+++ b/node/src/indexer/handler.rs
@@ -1,9 +1,7 @@
 use crate::hkdf::ScalarExt;
 use crate::indexer::stats::IndexerStats;
-use crate::indexer::IndexerState;
 use crate::metrics;
 use k256::Scalar;
-use near_crypto::PublicKey;
 use near_indexer_primitives::types::AccountId;
 use near_indexer_primitives::views::{
     ActionView, ExecutionOutcomeWithIdView, ExecutionStatusView, ReceiptEnumView, ReceiptView,
@@ -49,9 +47,7 @@ pub(crate) async fn listen_blocks(
     concurrency: std::num::NonZeroU16,
     stats: Arc<Mutex<IndexerStats>>,
     mpc_contract_id: AccountId,
-    account_public_key: Option<PublicKey>,
     sign_request_sender: mpsc::UnboundedSender<ChainSignatureRequest>,
-    indexer_state: Arc<IndexerState>,
 ) {
     let mut handle_messages = tokio_stream::wrappers::ReceiverStream::new(stream)
         .map(|streamer_message| {
@@ -59,9 +55,7 @@ pub(crate) async fn listen_blocks(
                 streamer_message,
                 Arc::clone(&stats),
                 &mpc_contract_id,
-                &account_public_key,
                 sign_request_sender.clone(),
-                indexer_state.clone(),
             )
         })
         .buffer_unordered(usize::from(concurrency.get()));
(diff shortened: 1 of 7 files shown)


node/src/indexer/types.rs
```diff
@@ -128,6 +136,14 @@ impl ChainSendTransactionRequest {
             ChainSendTransactionRequest::VoteReshared(_) => "vote_reshared",
         }
     }
+
+    pub fn gas_required(&self) -> Gas {
```
saketh-are, line 140: Here we introduce the ability to set the attached gas based on the kind of transaction. The gas for `respond` should definitely be lowered. However, I ran into issues in the localnet tests where respond is burning ~238 Tgas for some reason. I am not sure if this is a regression after the recent contract changes or an issue in my local testing. As the gas cost is not the main focus of this PR I propose adjusting it separately after understanding the localnet issue.

node/src/indexer/tx_sender.rs
```diff
@@ -62,24 +68,71 @@ async fn submit_tx(
     match result {
```
robin-near, line 68: Ah! we can actually remove the TODO on the line above now.

node/src/indexer/tx_sender.rs
```diff
@@ -20,31 +24,33 @@ async fn submit_tx(
     indexer_state: Arc<IndexerState>,
     method: String,
     params_ser: String,
-) -> Option<Nonce> {
+    gas: Gas,
+) -> bool {
```
robin-near, line 28: hmm, what do you think of turning this into an `anyhow::Result<()>`? Benefits are (1) it's clear what true/false each means; (2) the caller cannot inadvertently ignore the result without triggering a compiler warning; (3) you can tracing::error once at the call site.

node/src/indexer/tx_sender.rs
```diff
@@ -90,32 +143,53 @@ async fn submit_tx(
...
+                metrics::MPC_OUTGOING_TRANSACTION_OUTCOMES
+                    .with_label_values(&[request.method(), "timed_out"])
+                    .inc();
+                continue;
+            }
+            Err(err) => {
+                metrics::MPC_OUTGOING_TRANSACTION_OUTCOMES
+                    .with_label_values(&[request.method(), "unknown"])
+                    .inc();
+                tracing::warn!(target:"mpc", %err, "encountered error trying to confirm result of transaction {:?}", request);
```
robin-near, line 189: Would this unconditionally be printed if the transaction is VotePk, VoteReshared, or Join?

How about changing `confirm_tx_result` to `should_tx_be_resubmitted` and return `Ok(false)` if it's VotePk, VoteReshared, or Join?

node/src/indexer/handler.rs
```diff
@@ -75,15 +75,14 @@ async fn handle_message(
     mpc_contract_id: &AccountId,
     _account_public_key: &Option<PublicKey>,
     sign_request_sender: mpsc::UnboundedSender<ChainSignatureRequest>,
-    indexer_state: Arc<IndexerState>,
+    _indexer_state: Arc<IndexerState>,
```
robin-near, line 78: should we just remove the two unused fields? Do you see them being used again in the near future?

node/src/indexer/tx_sender.rs
saketh-are: Yeah that sounds good, it doesn't make sense to worry the node operator over this error. In the long run does it make sense to implement this logic for the other transaction types? Or would the coordinator loop resubmit them anyway.

saketh-are: Went with an enum since `Ok(false)` makes "not implemented" indistinguishable from "don't resubmit, observed success". The metrics will be clearer this way.

robin-near (approved):

PR #194, opened 2025-02-10
Sign and DKG protocol wrappers for Frost via `Protocol` trait

### LINKED ISSUE #119: Support frost signatures

kuksag:
This PR aims to simplify development and maintenance of Frost wrappers.
First attempt to describe an explicit state machine was made here https://github.com/Near-One/mpc/pull/172.
After coding/reviewing process, a little more time was spent to investigate how `cait-sith` model of writing protocols can be applied. 

Thus, instead of implementing `Protocol` trait directly, we write an `async` function, and convert it to the trait via `run_protocol`. This is where we need to expose some primitives from `cait-sith` lib, which is done here https://github.com/Near-One/cait-sith/pull/5.

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -10,7 +10,7 @@ anyhow = "1.0.92"
 async-trait = "0.1.83"
 axum = "0.7.9"
 borsh = { version = "1.5.1", features = ["derive"] }
-cait-sith = { git = "https://github.com/Near-One/cait-sith.git", features = ["k256"], rev = "1bb67e2a04f01cc7835ce4d8a611064229acc10c" }
+cait-sith = { git = "https://github.com/kuksag/cait-sith", rev = "59d032582196f3bf7aee402b924b36094bdc4c51", features = ["k256", "internals"] }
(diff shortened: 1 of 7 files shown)


node/Cargo.toml
```diff
@@ -10,7 +10,7 @@ anyhow = "1.0.92"
 async-trait = "0.1.83"
 axum = "0.7.9"
 borsh = { version = "1.5.1", features = ["derive"] }
-cait-sith = { git = "https://github.com/Near-One/cait-sith.git", features = ["k256"], rev = "1bb67e2a04f01cc7835ce4d8a611064229acc10c" }
+cait-sith = { git = "https://github.com/kuksag/cait-sith", rev = "59d032582196f3bf7aee402b924b36094bdc4c51", features = ["k256", "internals"] }
```
kuksag, line 13: This PR can not be merge until https://github.com/Near-One/cait-sith/pull/5

SimonRastikian: Please reverse this change. Near-One/cait-sith#5 is merged

Issue #195, opened 2025-02-10
Implement contract-coordinated key generation and resharing


robin-near:
This solves two problems at once:
* Reliability issues with key resharing, see #151 and resulting (legitimately) flaky test #188 .
* Consensus on the result of key generation and resharing, as described in https://docs.google.com/document/d/1A1yBVwb9PaZ1ebgNrEKYKgZ6EZNSXyO9pGHC6Hk4xrE/edit?tab=t.0

We should use the implementation proposed in the doc at the end, namely:
* When initiating a key generation or resharing, the leader makes a tx to the contract signalling the start of that computation.
* As each party completes the computation they vote for it in the contract.
* If all votes are received on the contract, every party succeeds; if K number of blocks pass before all votes are received on the contract, every party fail.
* Each computation carries a version number (or some other unique ID), so that multiple keygen/resharing attempts do not conflict with each other (especially since technically, it is possible for two computations to be proposed at the same time.)

This requires implementation in both the contract and the MPC node.

Issue #169, opened 2025-02-02
Have a proper story about how signature requests during a node's downtime are handled


robin-near:
When a node was down, how do we handle the signature requests that were submitted while the node was down?

* If we just ignore them (have the indexer only give us the signatures after syncing to head), then even when other nodes ask us to participate in the signing for the request, we may cause them to fail because we haven't indexed the signature request and so won't honor the signing computation. (We would've better been offline.)
* If we don't ignore them (have the indexer give us everything since the last run), then suddenly the node would pick up a bunch of signature requests that were outdated, and the node would suddenly start acting on them, sending a bunch of out-of-date/duplicated response requests.

It seems that we need to think harder about how to handle these.

robin-near:
Superceded by the broader design #226

PR #196, opened 2025-02-10
Fix/workaround test flakes and some minor improvements.

robin-near:
Temporary fix for the following flakes; I have checked that these no longer fail after many runs:
* `test_faulty_cluster` may fail due to:
  * When bringing down a node and then requesting a signature right after, the signature request may be ignored if the primary leader is the downed node and the secondary leader has not yet noticed that the primary leader is offline. The proper fix is #169 ; the temporary fix is to sleep 2 seconds after bringing down a node.
  * When bringing down a node, unowned presignatures that have already been successfully computed by the owner may not have had a chance to be persisted to the DB on the downed node. When this node is back online, this can lead to signing requests failing due to missing unowned presignatures on this node. The proper fix is #169 once again (since we cannot ensure that every presignature is consistently persisted across all participants); the temporary fix is to generate enough presignatures in the test before bringing any node down so that the FIFO will ensure that the presignatures that are being used were generated during the everybody-is-up period.
  * (Fixed in #198 ) ~~Suppose we have nodes A, B, C and C is down. When bringing C back up, A may have connected to B and C first, and start a computation, while B is still only connected to A. Due to receiving the computation from A, B creates its `NetworkTaskChannel` for the computation, which includes querying the current connection version for each participant. The problem is that the connection version for C corresponds to a previously downed connection (the watch channel contains, e.g. (4, empty Weak ptr). This will cause the computation to fail as soon as we try to send a new message, because we're expecting to have connection version 4. Previously, this issue was fixed by handling the case where no connection were ever established, but that fix was incomplete. Instead, we fix the issue correctly here: if by the time we create the `NetworkTaskChannel` we see (x, empty Weak ptr), we pick `x+1` as the expected connection version; indeed, we wish to use the next connection, whenever that is ready. This also (as is usually the case with correct logic) simplifies the `send_message` code.~~
* Resharing tests may fail due to the scenario described in #151 . The permanent fix is #195 ; the temporary fix is to wait 2 seconds before doing any resharing. This makes the network much less available during fail-and-retry loops.

We also make a few minor changes:
* For ease of debugging, when a task spawned with `spawn_checked` fails, we now print an entire "task trace"; it's like a stack trace that gives more context.
  * Related, include the account ID as well as participant ID in the task tracking descriptions during an integration test, so that we can know which node generated task failures.
* Avoid doing a blocking join on the indexer thread in `cli.rs`. This way if either the indexer or the mpc task fails, the program can immediately exit. Previously it was swallowing any fatal initialization errors in the mpc task (such as cannot open DB).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,9 +66,9 @@ jobs:
 
           url="https://s3.us-west-1.amazonaws.com/build.nearprotocol.com/nearcore/${os_and_arch}/${branch_name}/${commit_hash}/neard"
 
-          mkdir -p target/release
-          if ! curl -o target/release/neard "${url}"; then
-            echo "curl failed with URL: ${url}"
+          status_code=$(curl -v -o target/release/neard -w "%{http_code}" "${url}")
+          if [ "$status_code" -ne 200 ]; then
+            echo "curl failed with URL: ${url}, Status Code: ${status_code}"
             exit 1
           fi
           chmod +x target/release/neard
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3701,13 +3701,13 @@ dependencies = [
  "lru 0.12.5",
  "mpc-contract",
  "near-client",
- "near-config-utils 0.0.0",
- "near-crypto 0.0.0",
+ "near-config-utils 2.5.0-rc.1",
+ "near-crypto 2.5.0-rc.1",
  "near-indexer",
  "near-indexer-primitives",
  "near-o11y",
  "near-sdk",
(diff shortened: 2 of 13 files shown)


bowenwang1996:
@robin-near looks like the test `test_faulty_cluster` failed on CI

robin-near:
> @robin-near looks like the test `test_faulty_cluster` failed on CI

god damn it :)

robin-near:
OK it's finally passing.

ashamis (approved):

saketh-are (approved):
Thanks @robin-near !

PR #200, opened 2025-02-11
minor fixes to pytest 

kevindeforth:
This PR:
- fixes the pytest fixture logic for contract compilation (one pytest fixture loaded the contract before the other one compiled it)
- changes the timeout for nonce conflicts in the update pytest and increases the number of signature requests (@ashamis had a test fail, this might help)

### DIFF
--- a/pytest/common_lib/contracts.py
+++ b/pytest/common_lib/contracts.py
@@ -51,17 +51,24 @@ class UpdateArgsV1:
     """
 
     def __init__(self, code_path=None, config=None):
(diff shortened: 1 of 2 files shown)


pbeza:
I'm not familiar enough with the code to confidently review it yet, but this PR fixed the issue I ran into when trying to run `pytest` with `./exec_pytest.sh --verbose --reset-submodules`. 👍🏼

robin-near (approved):

PR #203, opened 2025-02-11
add logging to propose_update in contract

ashamis:
Propose_update is the only function in the MPC start contract that does not log its state. Add this logging.


kevindeforth (approved):
Ok.
Please also add the binary to `libs/chain-signatures/contract/compiled-contracts/v1.wasm`.

PR #211, opened 2025-02-13
Add docker container update script

andrei-near:
This will be used to update mpc container instead of automatic watchtower

### DIFF
--- a/scripts/update-mpc-node.sh
+++ b/scripts/update-mpc-node.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+
+# Exit immediately if a command fails, print commands, and ensure pipelines fail properly
+set -exo pipefail
+
+CONTAINER_NAME="mpc-node"
+IMAGE_NAME="nearone/mpc-node-gcp:mainnet-standalone"
+ENV_FILE=".env"
+VOLUME_PATH="/home/mpc:/data"
+
+# Get currently running container image ID (if exists)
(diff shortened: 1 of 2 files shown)


scripts/update-mpc-node.sh
```diff
@@ -0,0 +1,45 @@
...
+echo "🛑 Stopping existing container..."
+docker stop "$CONTAINER_NAME"
+
+echo "🗑 Removing existing container..."
+docker rm "$CONTAINER_NAME"
+
+
+# Run the new container
+echo "🚀 Starting new container..."
+docker run -d --name "$CONTAINER_NAME" -p 8080:8080 -p 80:80 -p 3000:3030 --restart always -v "$VOLUME_PATH" --env-file "$ENV_FILE" "$IMAGE_NAME"
```
robin-near, line 36: On ports:
 - Are we using "port_override: 80" on mainnet or not? If not, there are some URLs that are https and the port would default to 443 instead of 80.
 - neard port should not be exposed. It's OK to expose here but just wanna make sure the production nodes do not expose that to the public.

robin-near (approved):

scripts/update-mpc-node.sh
andrei-near: port_override - yes, we are using port 80 on mainnet too.
neard port is exposed mainly to be used in health-checks. Once mpc http server starting outside mpc thread implemented, we can remove this port binding. Also, from my checks, most operators have the metrics port locked down.

PR #215, opened 2025-02-14
refactor(audit): shorten CLI's function bodies

pbeza:
This commit addresses finding B1 from the Trail of Bits audit titled _'Functions with too many lines.'_


node/src/cli.rs
```diff
@@ -96,17 +102,60 @@ impl StartCmd {
...
+            }
+            res = indexer_handle => {
+                res??;
+            }
+        }
+        Ok(())
+    }
+
+    fn create_root_future(
+        &self,
```
robin-near, line 137: do you actually need self?

If not, is it possible to just make this function async fn?

node/src/cli.rs
```diff
@@ -182,60 +205,95 @@ impl Cli {
...
-                    serde_json::to_string(&participants_config)?,
-                )?;
-                Ok(())
+                self.run_generate_test_configs(output_dir, participants, threshold)
+                    .await
             }
         }
     }
+
+    async fn run_generate_test_configs(
```
robin-near, line 218: I kinda think this one is a bit over-refactored. I think extracting `create_file_config` reduces a lot of clutter but write_config_files and write_participants_config should just be inlined. wdyt?

node/src/cli.rs
```diff
@@ -1,22 +1,27 @@
-use crate::config::{
-    load_config_file, BlockArgs, KeygenConfig, PresignatureConfig, SecretsConfig, SignatureConfig,
-    SyncMode, TripleConfig, WebUIConfig,
+use crate::{
```
robin-near, line 1: I've been following nearcore's style guide that we don't group imports beyond one level.
It makes sense because grouping it so deeply means that find/replace is a lot harder, and import changes trigger more lines to be changed.

node/src/cli.rs
pbeza: Good point. Fixed. ✅

https://github.com/Near-One/mpc/blob/0142629941a252fff27fe8048a221fce71da5913/node/src/cli.rs#L136-L175

node/src/cli.rs
pbeza: I agree. Fixed. ✅

https://github.com/Near-One/mpc/blob/446cef75ddf6be8737f049f78804e82a6b07f1ac/node/src/cli.rs#L209-L233

node/src/cli.rs
pbeza: OK, I reverted it. ✅

https://github.com/Near-One/mpc/blob/446cef75ddf6be8737f049f78804e82a6b07f1ac/node/src/cli.rs#L1-L19

pbeza:
@robin-near, could you review it again, please? BTW, CI checks seem to be flaky – I had to manually re-run the tests to get them to pass.

node/src/mpc_client.rs
```diff
@@ -312,14 +191,136 @@ impl MpcClient {
...
 
         Ok(())
     }
 
+    async fn monitor_passive_channels_inner<T>(
+        mut channel_receiver: mpsc::Receiver<NetworkTaskChannel>,
+        mpc_client: MpcClient,
+        auto_abort: tracking::AutoAbortTask<T>,
+    ) -> anyhow::Result<(), JoinError> {
+        let _auto_abort_metrics = auto_abort;
```
robin-near, line 207: wait I'm not sure why _auto_abort_metrics needs to be in here - we can just have the run() function keep that as a local variable. Don't need to move it in here.

node/src/mpc_client.rs
```diff
@@ -312,14 +191,136 @@ impl MpcClient {
...
         generate_presignatures.await??;
 
         Ok(())
     }
 
+    async fn monitor_passive_channels_inner<T>(
+        mut channel_receiver: mpsc::Receiver<NetworkTaskChannel>,
+        mpc_client: MpcClient,
+        auto_abort: tracking::AutoAbortTask<T>,
+    ) -> anyhow::Result<(), JoinError> {
```
robin-near, line 206: Can this just return ()? I don't see anything in here that returns an error.

robin-near (approved):

node/src/mpc_client.rs
pbeza: `ping` @robin-near.

robin-near (approved):

PR #216, opened 2025-02-17
[MPC-Contract] compiler optimizations and binary file integrity check

kevindeforth:
This PR:
- adds compilatin optimizations `lto = "fat"`, `codegen-units = 1`, `strip = true`, `opt-level = "s"` and `panic = "abort"`. This reduces the binary file size by 28%.
- bumps the contract version from `1.0.0-rc.5` to `1.0.0-rc.6`;
- adds a pytest that checks if the hash of the compiled contract matches the hash of the current contract version in the `compiled-contracts` folder (excluded from ci, because compilation is platform specific);
- rename `CURRENT_CONTRACT_PATH` to  `COMPILED_CONTRACT_PATH` to better reflect which binary the pytest is using;


bowenwang1996 (approved):
let's use 1.0.0 for the contract version. Thanks!

PR #217, opened 2025-02-17
Chores: keygen default value

andrei-near:
Handle the case when keygen value is missing from config.yaml.
Bonus: simplifying secrets fetching from GCP and avoid additional permissions.

Tested by building a [docker image](https://github.com/Near-One/mpc/actions/runs/13378354585/job/37362209104) and deploying on mpc node [2](http://34.169.143.244/ui/jobs/new-mpc-node-2@default) and [3](http://34.169.143.244/ui/jobs/new-mpc-node-3@default/definition?view=job-spec) .


libs/chain-signatures/Cargo.toml
```diff
@@ -4,6 +4,14 @@ members = [
...
 resolver = "2"
 
+[profile.release]
+lto = "fat"
+codegen-units = 1
+strip = true
+panic = "abort"
+opt-level="s"
+
+
```
kevindeforth, line 14: might have been a rebase issue, these changes are already in main since [PR 216](https://github.com/Near-One/mpc/pull/216/files)

node/src/keyshare/gcp.rs
```diff
@@ -36,31 +35,15 @@ impl GcpKeyshareStorage {
...
-                parent: secret_name,
-                ..Default::default()
-            })
-            .await
-            .context("Failed to list secret versions")?
-            .into_inner();
+        let secret_name = format!(
+            "projects/{}/secrets/{}/versions/latest",
+            self.project_id, self.secret_id
+        );
```
kevindeforth, line 41: I am unfamiliar with this part of the code, but I wonder if we need to update the `store()` function on line 58 of the same file.

libs/chain-signatures/Cargo.toml
andrei-near: yes, this is a Github issue: it shows changes from PR 216 too. I had to rebase after PR216 was merged and this is the result.

node/src/keyshare/gcp.rs
andrei-near: No changes need in in store(). Latest is a default secrets tag. 
Please node that [this entire dev cluster](http://34.169.143.244/ui) is running using this code

robin-near: The reason why I didn't like directly accessing latest is that I'm not sure if it's possible to distinguish from the case of no key vs the case of access denied. Specifically, if we just do this, I don't think it would work for the initial keygen phase, because there's no latest version.

PR #219, opened 2025-02-18
use the main account if respond.yaml is missing

saketh-are:
Tested manually.

### DIFF
--- a/node/src/indexer/real.rs
+++ b/node/src/indexer/real.rs
@@ -3,7 +3,7 @@ use super::participants::{monitor_chain_state, ContractState};
 use super::stats::{indexer_logger, IndexerStats};
 use super::tx_sender::handle_txn_requests;
 use super::{IndexerAPI, IndexerState};
-use crate::config::{load_respond_config_file, IndexerConfig};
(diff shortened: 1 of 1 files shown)


bowenwang1996 (approved):

node/src/indexer/real.rs
```diff
@@ -29,7 +29,13 @@ pub fn spawn_real_indexer(
             let indexer =
                 near_indexer::Indexer::new(indexer_config.to_near_indexer_config(home_dir.clone()))
                     .expect("Failed to initialize the Indexer");
-            let respond_config = load_respond_config_file(&home_dir).unwrap();
+            let respond_config = load_respond_config_file(&home_dir).unwrap_or_else(|err| {
```
robin-near, line 32: I'd prefer if load_respond_config_file returned a Result<Option<...>> where if the file doesn't exist it is explicitly returned as an Option::None. The reason is that it can be easy to make a mistake in the yaml syntax and cause the config to fail to deserialize, and yet it would almost silently fall back to single account.

But we can leave that for later.

robin-near (approved):

PR #229, opened 2025-02-21
Reintroduce deposit and refund for signature requests

### LINKED ISSUE #228: `refund_on_fail` should not send back deposited NEAR which is only 1 yocto NEAR in V1 contract
According to the latest implementation in V1 contract, the required deposit of `sign` function is 1 yocto NEAR: https://github.com/Near-One/mpc/blob/312a51139c9e4b4d48cdb9e2281f09846ca92539/libs/chain-signatures/contract/src/lib.rs#L360-L363

In this case, it makes no sense to refund the failed request which is only 1 yocto NEAR (much more gas will be paid than 1 yN). We may skip the `refund_on_fail` step if the attached NEAR is 1 yN. 

https://github.com/Near-One/mpc/blob/312a51139c9e4b4d48cdb9e2281f09846ca92539/libs/chain-signatures/contract/src/lib.rs#L866-L871

However, we may still need to refund the extra paid NEAR beyond 1 yN if there's any. One way to prevent that is the function doesn't allow deposit more than 1yN for `sign` function (apply `assert_one_yocto()`), but this is not backward compatible for existing projects that are already built with the V0 contract interface.

kevindeforth:
resolves #228 
Contract V1.0.1 with following patches:

- re-introduces a minimum deposit of 1 yoctonear for a signature request, in order to avoid silent function calls from malicious frontends. Thanks to @think-in-universe for pointing out this oversight on our part!
- the contract refunds deposits over 1 yoctoear attached to sign requests. This is because we have to expect that developers might continue to attach deposits to their signature requests, in the expectation that the refund mechanism works as for V0.
-  tests updating the contract from V1 to V1.0.1
- refactor `exeperimental_signature_deposit` so the code is clearer in its intention (c.f. https://github.com/Near-One/mpc/issues/228)


bowenwang1996 (approved):

think-in-universe:
Thanks for the fix! 

Please let me know after the new version is deployed on testnet and we'll test it.

Issue #226, opened 2025-02-19
Implement robust signature generation strategy


robin-near:
https://docs.google.com/document/d/1qWkbMthVVyD89zJOvAUZrrTkc_SqCYwTu-fCb06-csg/edit?tab=t.0#heading=h.nbblafxf4qiy

Copied here:

### Signature Generation

The performance and reliability goals of signature generation are:

* Minimize latency from when a signature request is submitted on chain to when we generate and submit a signature back to the chain.  
* Throughput needs to be high, on the order of 150 signatures per second.  
* On a best effort basis, signature requests should not be dropped.  
* Do not waste assets unnecessarily; ideally have a single leader per signature generation.

The proposed strategy is:

* Periodically (e.g. once per 2s) exchange a **ChainHeight** message from every node to every other node, so that each node maintains a local view of the chain height of every node.  
* When a node indexes a signature request, it places it in the **Pending Signature Request** set.  
* Each node continuously performs the following for each signature request R in the Pending Signature Request set:  
  * If the block height R was in is older than highest peer height minus **SIGNATURE\_HORIZON** (the yield-resume timeout), discard R. (i.e. the signature request has already timed out on chain.)  
  * If a successful response to the signature has been indexed, discard R.  
  * Determine the leader for R’s computation:  
    * Start with the set of block heights every node is at. Remove nodes whose heights are less than the highest peer height minus **NODE\_TOO\_BEHIND\_THRESHOLD** (e.g. 10).  
    * Remove nodes that are offline from the view of this node.  
    * The leader is the node with the lowest Hash(R || node ID) among the remaining nodes.  
  * If the node is the leader, and the last time that the node started a signature request for R is older than **SIGNATURE\_RETRY\_DELAY** (e.g. 10 sec), start a signature computation. The signature computation itself should have a timeout \<= that delay.  
  * When a signature computation succeeds, send the response transaction to the chain.

robin-near:
Implementation complete; need to test this on a cluster.

Also further improvement is desirable in #236

robin-near:
Tested in a cluster of 8 nodes (located in the same region) with 50 QPS and bringing down two nodes and bringing them back up. There were no apparent missed signatures (though the accounting of how many signatures were missed was based on eyeballing of debug pages and not a definitive metric), and debug pages show leader selection fallback working.

Issue #225, opened 2025-02-19
Implement more robust asset generation strategy


robin-near:
https://docs.google.com/document/d/1qWkbMthVVyD89zJOvAUZrrTkc_SqCYwTu-fCb06-csg/edit?tab=t.0#heading=h.8g6o7t9bku9v 

Copied here:

### Asset Generation

Assets include triples and presignatures. The performance and reliability goals for asset generation are:

* Maximize throughput, i.e. number of assets that can be generated per unit of time  
* We do not care about failures; if a single asset fails to generate, we just give up. (But for the purpose of better throughput, we should fail fast.)

The strategy proposed:

* The leader of each asset generation computation determines the participants based on its local view of the active (online) participant set. (This implies that the leader already has an active connection to each of these participants.)  
* The leader sends an explicit **Start** message to each participant; this lets each participant know who the leader is, as well as communicating the input parameters of the computation.  
  * If a participant receives a message before Start, it buffers the message until receiving Start. This can happen if node B receives the Start message, starts the computation by sending something to node C, but C receives from B before receiving its Start message.  
* The leader aborts the computation as soon as any of the following happens:  
  * Any connection to a participant breaks.  
  * The leader receives an Abort message from any participant.  
  * The computation times out.  
* A passive participant (anyone not the leader) has different behavior:  
  * A passive participant, if not initially connected to another participant, waits indefinitely for that participant to be connected (up to the computation-level timeout).  
    * However, once the connection is alive, if it breaks, the passive participant aborts the computation.  
  * A passive participant also aborts the computation if it times out, or encounters another fatal failure such as an input asset not existing on disk.  
  * When a passive participant aborts, it sends an **Abort** message to the leader.  
  * When a passive participant completes the computation, it sends a **Success** message to the leader.  
* If the leader aborts, it sends a **LeaderAbort** message to all other participants. This is a best-effort attempt to clean up resources, but is not necessary.  
* The leader must wait for all Success messages before considering the computation successful.

PR #230, opened 2025-02-22
Implement leader-centric computation and indexer height network APIs.

### LINKED ISSUE #13: Properly understand how to pick channel sizes to avoid deadlocks
Every mpsc channel needs to have a channel size, or be deliberately chosen to be unbounded. Unbounded channels can lead to arbitrarily large blockage somewhere in the system, but bounded channels can lead to hard-to-understand deadlocks. Currently, deadlocks can happen if there are too many protocols running at the same time. We should understand where they happen so we can avoid them.

### LINKED ISSUE #37: The owner of a computation should complete after other participants persist their shares
e.g. participant 1 initiates a triple computation. The triple computation should only be considered complete and written to disk after receiving a confirmation from each other participant that they have written their shares of the triple to disk.

This is to prevent a case where a passive participant's node (say participant 2) is shut down while still in the middle of the computation. The computation is nearly complete and participant 1 already stores its triple share to disk, but participant 2 shut down before completing and persisting its share. Now when participant 1 uses the triple for a presignature computation, participant 2 is unable to fulfill its part of the computation.

This also resolves #8 and simplifies the DistributedAssetStorage code by removing the need for `prepare_unowned`.

The implementation of this would require an additional round of communication outside the cait-sith protocol, as cait-sith protocols are unaware of "ownership", so we would need additional custom network messages for this purpose.

robin-near:
Major changes:

* There are different kinds of network messages now:
  * MpcMessage: can be Start, Computation(data), Abort(msg), Success. See #225 for definitions. Abort is used for both Abort and LeaderAbort in the definitions.
  * IndexerHeightMessage: for syncing indexer heights, for #226 . This is exchanged and handled, but not used yet; that'll be left for a future PR focused on #226 .
* Implement leader-centric computations; make all MPC computations leader-centric:
  * If leader fails, send Abort to everyone else.
  * If follower fails, send Abort to leader.
  * For triple and presignature generation, leader wait for followers to send Success before declaring itself success.
    * TODO with next PR: Resolve #37 which is now very easy to do.
* Computations are now started explicitly with `MpcMessageKind::Start`; any messages received before receiving `Start` are buffered.
  * This supercedes the `deleted_channels` buffer before; any channels that don't receive Start will basically just be ignored.
  * `Start` carries the participant list, so now we don't have to send that along every single message.

Minor changes:
* Network messages are borsh serialized and deserialized now. No more manual serde.
* Change all network channels to unbounded channels. This makes `send` no longer async and removes any possibility of a deadlock due to channel capacities. Resolves #13 .
* Simplifies protocol interfaces so we no longer need to pass in the `participants` list and `me` into triple, presig, sig, etc. computations.
* Computations now wait for every participant to be connected, rather than doing that when `send`ing a message.

~~Minor deviation from specs from #225 :~~ (fixed)
* ~~If a computation *times out*, it does not send Abort messages to others. This is because it's beneficial for timeout to be the last layer of defense (so if it times out, everything just gets dropped forcefully), and it should be OK, because if one party times out, the other parties should also timeout at around the same time.~~

### DIFF
--- a/node/src/network.rs
+++ b/node/src/network.rs
@@ -1,14 +1,21 @@
 pub mod conn;
+pub mod constants;
 pub mod handshake;
+pub mod indexer_heights;
 
-use crate::primitives::{BatchedMessages, MpcMessage, MpcPeerMessage, MpcTaskId, ParticipantId};
+use crate::metrics;
+use crate::primitives::{
+    IndexerHeightMessage, MpcMessage, MpcMessageKind, MpcPeerMessage, MpcStartMessage, MpcTaskId,
+    ParticipantId, PeerMessage,
+};
 use crate::tracking::{self, AutoAbortTask};
-use conn::ConnectionVersion;
-use futures_util::future::BoxFuture;
-use futures_util::FutureExt;
+use conn::{ConnectionVersion, NodeConnectivityInterface};
+use futures::future::BoxFuture;
+use indexer_heights::IndexerHeightTracker;
 use lru::LruCache;
 use std::collections::hash_map::Entry;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
+use std::ops::Deref;
 use std::option::Option;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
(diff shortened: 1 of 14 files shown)


node/src/network.rs
```diff
@@ -124,60 +163,108 @@ impl MeshNetworkClient {
...
-                    sender,
-                    channel: NetworkTaskChannel {
+                let incomplete_channel = IncompleteNetworkTaskChannel { receiver };
+                if let Some((k, _)) = channels
+                    .channels_waiting_for_start
+                    .push(task_id, incomplete_channel)
+                {
+                    if k != task_id {
+                        // Keep channels_waiting_for_start a subset of senders.
+                        channels.senders.remove(&k);
```
bowenwang1996, line 188: Could you add some comment on what this logic means? It is unclear to me what `k != task_id` is supposed to mean

node/src/network.rs
```diff
@@ -250,24 +339,48 @@ pub fn run_network_client(
...
+    participants: Vec<ParticipantId>,
+    /// The version of the connection we have with each other participant.
+    /// This is used to detect dropped connections. See ConnectionVersion for details.
+    connection_versions: HashMap<ParticipantId, ConnectionVersion>,
+    /// The underlying transport layer.
+    transport_sender: Arc<dyn MeshNetworkTransportSender>,
+}
+
+/// A wrapper around NetworkTaskChannel,
+pub struct NetworkTaskChannelWrapper(NetworkTaskChannel);
```
bowenwang1996, line 369: Why is this new struct necessary?

node/src/network.rs
```diff
@@ -293,88 +417,304 @@ impl NetworkTaskChannel {
...
+    }
+
+    /// This is called at the end of a leader-centric computation to communicate the result to
+    /// other parties. For leaders, only failures are communicated to followers; for followers,
+    /// failures are communicated to the leader, and if leader waits for success, also
+    /// communicate successes to the leader.
+    fn communicate_result<T>(
+        &self,
+        result: &anyhow::Result<T>,
+        leader_waits_for_success: bool,
```
bowenwang1996, line 467: Under what scenarios does a leader wait for success?

bowenwang1996 (commented):
You also need to update protocol version right?

node/src/network.rs
robin-near: I'm using a different way to do it; currently it's too ugly.

node/src/network.rs
robin-near: In my new code I documented it.

robin-near:
> You also need to update protocol version right?

I did, it's protocol version 2 now :)

node/src/network.rs
robin-near: Added plenty of extra comments; indeed this was not properly documented.

node/src/network/computation.rs
```diff
@@ -0,0 +1,85 @@
...
+    fn leader_waits_for_success(&self) -> bool;
+}
+
+pub trait MpcLeaderCentricComputationExt<T> {
+    fn perform_leader_centric_computation(
+        self,
+        channel: NetworkTaskChannel,
+        timeout: std::time::Duration,
+    ) -> impl Future<Output = anyhow::Result<T>> + 'static;
+}
```
bowenwang1996, line 31: Why is this a separate trait instead of being part of `MpcLeaderCentricComputation`?

bowenwang1996 (approved):

node/src/network/computation.rs
robin-near: I always feel that inserting a default implementation in a trait just to provide auxiliary utility is a bit weird because it doesn't quite belong to the trait and becomes override-able. The extension syntax is verbose on the other hand. I don't have a strong preference really.

bowenwang1996: I see. I prefer having a default implementation in a trait. It is simpler and easier to read

Issue #236, opened 2025-02-28
Ensure that indexer blocks are not missed upon restart/crash


robin-near:
Currently, even if the indexer is started with SyncMode::FromInterruption, it's possible that block updates are sent by the indexer but missed by the MPC node. Upon restart, the indexer is going to yield from the interruption *it* is aware of, but the missed blocks means that some signature requests would never be indexed. These signature requests would be ignored: if the node is a leader, then it never completes the signature. If the node is a follower but is chosen as a participant, the request would timeout waiting for the indexed signature request.

I propose the following behavior:
* Store a "last indexed final block height" into the DB; commit it at the same time as committing signature requests in a block (also requires committing signature requests of a block in a single commit).
* Upon startup, read the last indexed final block height, and ask the indexer to start yielding blocks from there.
* Signature computation should include the SignatureId as well as the block height from which the signature request was indexed.
* When participating in a signature computation, if the signature is not found while the "last indexed final block height" already passed the specified block height, the computation is aborted rather than waiting forever.

netrome:
We've not integrated the ChainGateway right now. Not sure how the guarantees for restarts/crashes are right now. Let's keep this open and revisit after the integration is completed.

PR #238, opened 2025-02-28
Implement robust signature leader selection and retry

robin-near:
This is an enormous PR, but the bulk of the PR is in self-contained files in src/signing/ and there are no major refactorings. Because the new files are self-contained it is possible to split into more PRs but then the context would also be lost. Still I'm happy to try splitting if needed.

## Overview
See #226 for the motivation and design. The design is mostly kept the same except the retry timeout described below.

The bulk of this PR is to implement the `PendingSignatureRequests`. This is a queue that keeps track of the signature requests that are not yet fulfilled by any node, performs the leader selection logic, and ultimately makes the decision of which signatures to start generating.

As its helper, the `RecentBlocksTracker` consumes optimistic blocks from the indexer, and converts it to a stream of final blocks, as well as providing a query of "should I still pay attention to signature requests in block X".

The indexer code is changed to yield data block-by-block (instead of signature-by-signature), with each block containing not just the signature requests but the responses as well. The request/response pairs are matched via the resuming receipt ID.

Finally, the signature processing flow is changed to periodically query the `PendingSignatureRequests` for any signatures that need generating, and kicking off the requested generations. Failing generations are automatically retried, up till the yield-resume timeout.
image

## Minor Breaking Change
There is a minor breaking change in this PR that it changes the ID that signatures are stored by. Instead of using the receipt ID that generates the signature request, we use the receipt ID that resumes with the signature response. This makes it much easier to find the matching request given the response receipt.

The effect of this breaking change is that all signatures stored in the asset DB are useless. However, this makes no real impact during an upgrade because only recent signatures matter.

## The `PendingSignatureRequests` Queue

This queue maintains a set of signatures for which a matching response has not yet been indexed from the finalized chain. The key function it provides is `get_signatures_to_attempt`, which does the following:
* For each signature in the queue (checking only once a second max), remove it if
  * we attempted it 10 times already;
  * it is older than 200 blocks (yield-resume timeout); or
  * it is in a fork that cannot possibly be on the canonical chain.
* If we're the current leader of the signature:
  * If the signature is *currently being attempted* (the attempt object is still alive), or if the signature was just submitted to the chain less than 10 seconds ago, let it sit for now.
  * Otherwise, create a new attempt for this signature and return it.

How does it select the leader?
* First, remove all nodes that are offline (not connected to the node).
* Then, among the remaining nodes, calculate the highest height they announced.
* Then, remove all nodes (including self) that are more than 10 blocks behind.
* The remaining nodes are eligible.
* For each signature, there is a precomputed leader selection preference order. The current leader is the first one in that order that is eligible.

## The `RecentBlocksTracker`
This indexes the recent window of optimistic blocks from the indexer, handling forks and canonical chains properly. It tolerates incomplete start of the data (missing parents, etc.), and provides a reasonably good classification for each block:

```
pub enum CheckBlockResult {
    /// The block is older than the recent window of blocks we keep.
    OlderThanRecentWindow,
    /// The block is within the recent window, and also finalized by the blockchain
    /// (it is an ancestor (including self) of the latest final block).
    RecentAndFinal,
    /// The block is recent enough, but belongs to a different fork so has no chance
    /// of being included in any canonical chain from this point on.
    NotIncluded,
    /// The block is optimistically included in the chain, and it is on the canonical chain,
    /// but it is not yet part of the final chain. It is also recent enough.
    OptimisticAndCanonical,
    /// The block is optimistically included in the chain, but it is not on the canonical chain.
    /// It is also recent enough.
    OptimisticButNotCanonical,
    /// We have not seen the block yet, but it appears recent, judging from the height.
    Unknown,
}
```

The idea is we only want to work on a signature request if it's either final, optimistically canonical, or unknown. We don't want to deal with signatures that are too old, because when we start up the node we may get some really old signature requests from catching up the chain. We also don't want to process requests that we know are on a discarded fork (this is rare but it's good to handle robustly anyway).

The second functionality of `RecentBlocksTracker` is to provide a stream of finalized blocks, and that's for determining when a signature request is fulfilled by a response on chain *for sure*.

## Indexer Changes
The indexer now yields `ChainBlockUpdate` instead of `ChainSignatureRequest`; the latter is part of the former as a list, and renamed to `SignatureRequestFromChain` to differentiate betwen another `ChainSignatureRequest` of the same name.

```
#[derive(Clone)]
pub struct ChainBlockUpdate {
    pub block: BlockViewLite,
    pub signature_requests: Vec<SignatureRequestFromChain>,
    pub completed_signatures: Vec<SignatureId>,
}
```

The *fake* indexer has also been updated to mimic a real chain by actually creating logical blocks and including signature responses into these blocks.

There's also another subtle change in the coordinator that when we're in an invalid state, we no longer drop new signatures (block updates now). Earlier, we did this because we had no mechanism to ignore old signatures, but now we do and this is unnecessary and causes test flakiness. #236 would be the logical next step to have block updates not only survive state changes, but also survive restarts.

## MpcClient changes
The old flow is to watch for signature requests from the indexer then respond to it if we're the leader or secondary leader. That flow is removed, and replaced with a loop that starts with a tokio::select:
 1. When we see a new block update, give it to the `PendingSignatureRequests` queue.
 2. When we see a debug request from the web component, serve the debug request.
 3. Or every 1 second.

After that, we query the queue for any signatures to attempt (new or retries), and directly go and kick off signature computations.

If a signature computation is retried, the previous progress is saved; specifically, if we had already generated the signature and we just failed to put it on chain, we'll begin the next attempt by just submitting the transaction on chain, without needing to compute again.

## Debugging and Monitoring
A large part of the PR is devoted to debugging and monitoring:
* Two new endpoints are added, `/debug/blocks` and `/debug/signatures`
* Metrics for queue size, number of blocks, final blocks, requests and responses indexed, are added.

`/debug/blocks` provides a drawing of the recent window of blocks (including forks, and marking of final and canonical blocks), for example:
image

`/debug/signatures` provides a list of recent signature requests, both pending and (up to 100) completed:
image
image


node/src/signing/queue.rs
```diff
@@ -0,0 +1,819 @@
...
+    }
+
+    /// Selects the leader given the current state of the network.
+    fn current_leader(&self, eligible_leaders: &HashSet<ParticipantId>) -> Option<ParticipantId> {
+        for candidate_leader in &self.leader_selection_order {
+            if eligible_leaders.contains(candidate_leader) {
+                return Some(*candidate_leader);
+            }
+        }
+        None
```
bowenwang1996, line 232: What happens when the current leader is `None`? Does the request just fail?

node/src/signing/recent_blocks_tracker.rs
```diff
@@ -0,0 +1,943 @@
...
+                }
+                if let Some(final_head) = &self.final_head {
+                    if block_height <= final_head.height {
+                        return CheckBlockResult::NotIncluded;
+                    }
+                }
+                if node.canonical.load(Ordering::Relaxed) {
+                    return CheckBlockResult::OptimisticAndCanonical;
+                }
+                CheckBlockResult::OptimisticButNotCanonical
```
bowenwang1996, line 425: For a block that is not final, why do we need to distinguish between it being on the canonical chain and it not being on the canonical chain?

node/src/indexer/handler.rs
```diff
@@ -161,15 +183,46 @@ fn maybe_get_sign_args(
...
+    if actions.len() != 1 {
+        return None;
+    }
+    let ActionView::FunctionCall {
+        ref method_name, ..
+    } = actions[0]
+    else {
+        return None;
+    };
+    if method_name != "return_signature_and_clean_state_on_success" {
```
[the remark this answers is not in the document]
robin-near: The integration tests checks the metrics; it makes sure that the correct number of responses are indexed, and that the signature requests are removed from the queue. If we changed the method name, that test would fail.

node/src/signing/queue.rs
robin-near: If the current leader is None, the request is ignored for now, but remains in the queue.

node/src/signing/recent_blocks_tracker.rs
robin-near: Signature requests on the canonical chain are attempted, but those not on the canonical chain are not. This is to handle the case where the blockchain may be struggling and having forks, and we don't want to doubly or triply or quadruply etc. attempting the same signature requests because they exist on multiple forks.

node/src/signing/queue.rs
```diff
@@ -0,0 +1,819 @@
...
+/// The inputs to this queue are:
+///  - Every block that comes from the indexer. For each block, we need the list of signature
+///    requests as well as the list of completed signature requests (i.e. responses).
+///  - The set of alive participants (nodes that we're currently connected to).
+///  - The height of the indexer of each participant.
+///
+/// What this queue then provides, via `get_signatures_to_attempt`, is a list of signature requests
+/// that we should attempt to generate a signature for. The list will be generated based on the
+/// following goals:
+///  - Assuming the network state is stable and nodes have consistent views of the connectivity and
```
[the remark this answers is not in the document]
robin-near: No, but they keep exchanging height info so this is more like an eventual consistency model. So although we're going to sometimes have inconsistent views and therefore disagreeing calculations of who the leader is, the hope is that this inconsistency does not cause actual problems. This is a hand-wavy argument. For example, if node A keeps hovering between being 9 and 10 blocks behind, then the other nodes could keep switching between thinking that it's stale vs thinking that it's not. That can keep changing the leader selection. Suppose a signature request has leader preference [A, B, C, ...] then the effect of that flickering is probably just going to be that A and B both end up trying the signature request (because as long as they see themselves as leader in one iteration, they will go ahead and attempt it (regardless of whether someone else is already leading a computation)). And that's okay. What isn't OK is if somehow, for long periods of time, both A and B do not think they are leaders, but C thinks either A or B is a leader, and as a result there be no leaders ever trying the signature. But this seems contrived. For A to think it is ineligible it would need to see a max height more than 10 + its local height, but for B to think A is eligible means it thinks A's height + 10 is >= max height, but its view of A's height should be <= A's local height, so somehow A must be getting consistently higher heights announced than B, and that doesn't seem very likely. Anyway, it's handwavy, but the goal really is to make sure we always do have some leader as long as the view is eventually consistent.

node/src/signing/recent_blocks_tracker.rs
bowenwang1996: I see. Makes sense

bowenwang1996 (approved):

node/src/signing/queue.rs
```diff
@@ -0,0 +1,819 @@
...
+                "{} signs completed: {}",
+                self.completed_requests.len(),
+                ellipsified_shortened_hash_list(&self.completed_requests)
+            )?;
+        }
+        Ok(())
+    }
+}
+
+/// Thin API that need from the network.
```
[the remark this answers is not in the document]
robin-near: The grammar you mean?

PR #239, opened 2025-03-01
[frost] Keyshare protocol

### LINKED ISSUE #119: Support frost signatures

kuksag:
* Add wrappers for `refresh` and `repair` protocols
* Build `reshare` protocol on top of them 

Each file can be reviewed on its own. Separating might lead to the context loss.

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -44,9 +44,8 @@ tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 url = "2"
 x509-parser = "0.16.0"
 
-frost-ed25519 = "2.1.0"
-frost-core = "2.1.0"
-curve25519-dalek = "4.1.3"
+frost-ed25519 = { git = "https://github.com/kuksag/frost", rev = "5807812176b7b7b220286bc4ec7c080da72c04b8" }
+frost-core = { git = "https://github.com/kuksag/frost", rev = "5807812176b7b7b220286bc4ec7c080da72c04b8" }
 
 near-indexer = { git = "https://github.com/near/nearcore", rev = "39882d8d34281fae2e5551c000eeeede2e75f8da" }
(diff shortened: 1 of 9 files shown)


node/Cargo.toml
```diff
@@ -44,6 +44,9 @@ tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 url = "2"
 x509-parser = "0.16.0"
 
+frost-ed25519 = { git = "https://github.com/kuksag/frost", rev = "5807812176b7b7b220286bc4ec7c080da72c04b8" }
+frost-core = { git = "https://github.com/kuksag/frost", rev = "5807812176b7b7b220286bc4ec7c080da72c04b8" }
```
kuksag, line 48: To be discussed

node/Cargo.toml
```diff
@@ -10,7 +10,7 @@ anyhow = "1.0.92"
 async-trait = "0.1.83"
 axum = "0.7.9"
 borsh = { version = "1.5.1", features = ["derive"] }
-cait-sith = { git = "https://github.com/Near-One/cait-sith.git", features = ["k256"], rev = "1bb67e2a04f01cc7835ce4d8a611064229acc10c" }
+cait-sith = { git = "https://github.com/Near-One/cait-sith.git", rev = "d0144f8dd94e6e78f85a526fb45b5c891bf8bf7e", features = ["k256", "internals"] }
```
kuksag, line 13: Blocked on https://github.com/Near-One/cait-sith/pull/8

robin-near:
@kuksag could you please add documentation to every function that isn't immediately obvious, even private ones? I think that'll help with my review as well as helping future readers. Thanks!

kuksag:
This doesn't support threshold changing. Thus it's useless. Closed.

PR #240, opened 2025-03-03
Docker secret vars

andrei-near:
Adjusting the gcp-start script to allow direct use of secrets from env and remove hardcoded gcp dependencies.
In case GCP keyshare secret details are no provided (gcp project and secret name), MPC will use encrypted local store for keyshare storage.
This was tested in a dev cluster with contract deployed on this account:
https://testnet.nearblocks.io/address/signer-d506e3b6-d37f-4d51-a9a0-787ceaf4f1e8.testnet


deployment/gcp-start.sh
```diff
@@ -73,43 +73,30 @@ else
...
+fi
+
 
 echo "Starting mpc node..."
-GCP_PROJECT_ID="${GCP_PROJECT_ID}" \
-GCP_KEYSHARE_SECRET_ID="${GCP_KEYSHARE_SECRET_ID}" \
-MPC_SECRET_STORE_KEY=${LOCAL_ENCRYPTION_KEY} \
-MPC_P2P_PRIVATE_KEY=${P2P_PRIVATE_KEY} \
-MPC_ACCOUNT_SK=${ACCOUNT_SK} \
 /app/mpc-node start
```
robin-near, line 102: bash noob question: the environment variable GCP_PROJECT_ID implicitly transfers into this command right?

robin-near (approved):

robin-near:
I'm gonna merge this for you as I want to stop seeing nodes.tfvars.json conflicts :)

deployment/gcp-start.sh
andrei-near: Yes, ENV vars will be passed down, similar to how MPC_HOME_DIR is passed to mpc-node running env without being explicitly exported in this script

PR #246, opened 2025-03-06
feat: add migration tool migrate keyshares from GCP to local encrypted key file

### LINKED ISSUE #245: Create a CLI command to migrate keyshare from GCP to local encrypted key file

near-bookrock:
This PR adds an option to the main CLI program to migrate key shares from GCP to local encrypted storage. The PR resolves 
#245 .

### DIFF
--- a/node/src/cli.rs
+++ b/node/src/cli.rs
@@ -1,22 +1,25 @@
-use crate::config::{
-    load_config_file, BlockArgs, KeygenConfig, PresignatureConfig, SecretsConfig, SignatureConfig,
-    SyncMode, TripleConfig, WebUIConfig,
+use crate::{
+    config::{
+        load_config_file, AesEncryptionKey, BlockArgs, ConfigFile, IndexerConfig, KeygenConfig,
+        PresignatureConfig, SecretsConfig, SignatureConfig, SyncMode, TripleConfig, WebUIConfig,
+    },
+    coordinator::Coordinator,
+    db::SecretDB,
+    indexer::{real::spawn_real_indexer, IndexerAPI},
+    keyshare::KeyshareStorageFactory,
+    p2p::testing::{generate_test_p2p_configs, PortSeed},
+    tracking::{self, start_root_task},
+    web::start_web_server,
 };
-use crate::config::{ConfigFile, IndexerConfig};
-use crate::coordinator::Coordinator;
-use crate::db::SecretDB;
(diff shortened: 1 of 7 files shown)


node/src/config.rs
```diff
@@ -159,24 +159,51 @@ pub struct ParticipantInfo {
...
+#[derive(Clone, Copy, Debug)]
+pub struct AesEncryptionKey([u8; 16]);
+
+impl From<[u8; 16]> for AesEncryptionKey {
+    fn from(value: [u8; 16]) -> Self {
+        Self(value)
+    }
+}
+
+impl TryFrom<&str> for AesEncryptionKey {
```
robin-near, line 175: There's actually a FromStr trait, and the associated .parse::<AesEncryptionKey>() extension function. It's more semantically clear than TryFrom<&str>.

node/src/cli.rs
```diff
@@ -37,6 +40,48 @@ pub enum Cli {
...
+            secret_id: self.gcp_keyshare_secret_id,
+        }
+        .create()
+        .await?;
+
+        let encryption_key = AesEncryptionKey::try_from(self.secret_store_key_hex.as_str())?;
+
+        let local_key_share_storage = KeyshareStorageFactory::Local {
+            home_dir: self.home_dir.into(),
+            encryption_key: encryption_key,
```
robin-near: tip: Run "cargo clippy" to catch style suggestions like these (encryption_key alone is enough)

node/src/config.rs
near-bookrock: Good point.

andrei-near:
Wouldn't it be better to avoid GCP interactions all together and have mpc cli write given keyshare to local store using provided local encryption key? Since is a 1 time migration, users would get their current keyshare and local encryption key from GCP manually and import to local store, something like this:
$ CURRENT_KEYSHARE='{"epoch":1,"private_share":"1111111111111111111111111111111111111111111111111111111111111111","public_key":"1111111111111111111111111111111111111111111111111111111111111111"}'
$ CURRENT_LOCAL_ENCRYPTION_KEY="AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
$ mpc-node import-keyshare $MY_CURRENT_KEYSHARE $CURRENT_LOCAL_ENCRYPTION_KEY

PR #252, opened 2025-03-12
Abstract away Signature Schema

kuksag:
This PR aims to create abstraction over signature schema.
Most of the changes are refactoring:
* Move existing implementations of `cait-sith`-related code into its `SignatureProvider` (namely, `triple.rs`, `presign.rs`, part of `hkdf.rs` and so on). 

The new logic introduced:
* `MpcClient` made responsible for distributing `MpcTaskId` to each `SignatureProvider`
* `KeyshareStorage` made compatible with any type of keyshare
* `KeyGeneration` and `KeyResharing` semantics does not do keyshare persisting now. This was lifted up into `Coordinator.rs` logic. The reason is because at the moment it's not clear how to handle `epoch` and introduction of new keys. More on that below[1].

What this PR should do, but doesn't (will be addressed in the future PRs):
* Different Database Schemas for each signature provider (and thus flushing policy)
* GCP implementation or removal of such (at the moment I didn't bother with adding changes for it)
* Further step back from `cait-sith` structs and types, such as `Scalar`, which for example used to denote `tweak`, which essentially needs to be just a `[u8; 32]`.

[1]
* What should we do when a set of node wants to generate a keyshare of a new signature provider? 
* Should we use an `epoch = 0` for a new share or use the existing one?
* How do we guarantee atomicity when performing key-resharing? Do we even need this property? What if some node failed to write its share, while others succeeded?

### DIFF
--- a/node/src/config.rs
+++ b/node/src/config.rs
@@ -64,6 +64,20 @@ impl MpcConfig {
             participants,
         })
     }
+
+    /// When performing a key generation or key resharing protocol, someone has to create a channel.
+    /// Don't confuse with Leader Centric Computations.
+    pub fn is_leader_for_keygen(&self) -> bool {
+        let my_participant_id = self.my_participant_id;
+        let participant_with_lowest_id = self
+            .participants
+            .participants
+            .iter()
+            .map(|p| p.id)
+            .min()
+            .expect("Participants list should not be empty");
+        my_participant_id == participant_with_lowest_id
+    }
 }
 
 /// Config for the web UI, which is mostly for debugging and metrics.
--- a/node/src/key_generation.rs
+++ b/node/src/key_generation.rs
@@ -2,7 +2,7 @@ use crate::config::MpcConfig;
 use crate::keyshare::{KeyshareStorage, RootKeyshareData};
 use crate::network::computation::MpcLeaderCentricComputation;
 use crate::network::{MeshNetworkClient, NetworkTaskChannel};
-use crate::primitives::MpcTaskId;
+use crate::primitives::{EcdsaTaskId, MpcTaskId};
 use crate::protocol::run_protocol;
 use anyhow::Context;
 use cait_sith::protocol::Participant;
@@ -53,22 +53,13 @@ pub async fn run_key_generation_client(
     if keyshare_storage.load().await?.is_some() {
         anyhow::bail!("Keyshare already exists, refusing to run key generation");
     }
-    let my_participant_id = client.my_participant_id();
-    let is_leader = my_participant_id
-        == config
-            .participants
-            .participants
-            .iter()
-            .map(|p| p.id)
-            .min()
-            .unwrap();
 
-    let channel = if is_leader {
-        client.new_channel_for_task(MpcTaskId::KeyGeneration, client.all_participant_ids())?
+    let channel = if config.is_leader_for_keygen() {
+        client.new_channel_for_task(EcdsaTaskId::KeyGeneration, client.all_participant_ids())?
     } else {
         loop {
             let channel = channel_receiver.recv().await.unwrap();
-            if channel.task_id() != MpcTaskId::KeyGeneration {
+            if channel.task_id() != MpcTaskId::EcdsaTaskId(EcdsaTaskId::KeyGeneration) {
                 tracing::info!(
                     "Received task ID is not key generation: {:?}; ignoring.",
                     channel.task_id()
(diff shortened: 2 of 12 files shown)


node/src/config.rs
```diff
@@ -64,6 +64,20 @@ impl MpcConfig {
             participants,
         })
     }
+
+    /// A leader – participant with the lowest id.
+    /// Used to calculate who's going to create a channel for a task.
+    pub fn is_leader(&self) -> bool {
```
robin-near, line 70: nit: maybe name this "is_leader_for_keygen" or something like that. The problem is that the leader has a different definition in the case of triple/presig/sig generations.

node/src/signature_provider/ecdsa_secp256k1/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+}
+
+impl RegisteredMpcTask for EcdsaSecp256k1TaskId {
+    fn wrap(self) -> MpcTaskId {
+        MpcTaskId::EcdsaSecp256k1TaskId(self)
+    }
+}
+
+impl KeyshareId for KeygenOutput<Secp256k1> {
+    fn id() -> &'static str {
```
robin-near, line 113: I'd suggest still naming this keyshare_id() because on the KeygenOutput object we would be able to directly call .id() after importing this trait, and that would be confusing.

bowenwang1996 (commented):
Overall looks good. For epochs, it should be global, i.e, not specific to a key. I think we do need GCPStorage to be implemented to avoid breaking compatibility

node/src/keyshare/mod.rs
```diff
@@ -58,22 +52,17 @@ pub enum KeyshareStorageFactory {
 }
 
 impl KeyshareStorageFactory {
-    pub async fn create(&self) -> anyhow::Result<Box<dyn KeyshareStorage>> {
+    pub async fn create(&self) -> anyhow::Result<impl KeyshareStorage> {
```
robin-near, line 55: How is this going to be implementable if Gcp is also going to be supported (and as Bowen said we need to support it)? impl KeyshareStorage is a concrete type; there's no way you can return two different KeyshareStorage types.

I think one idea is to have KeyshareStorage still be a dyn trait (use dyn everywhere), but have load and store take an ID - they will return Vec<u8>. Then, have another wrapper on top which takes a keyshare type parameter and returns the deserialized type of the keyshare. That way we can abstract away the storage of the keyshare with a dyn trait, but still have type safety.

node/src/keyshare/mod.rs
```diff
@@ -1,52 +1,46 @@
...
-use serde::{Deserialize, Serialize};
+use crate::signature_provider::KeyshareId;
+use serde::{de, Deserialize, Serialize};
 
 /// The root keyshare data along with an epoch. The epoch is incremented
 /// for each key resharing. This is the format stored in the old MPC
 /// implementation, and we're keeping it the same to ease migration.
 #[derive(Clone, Serialize, Deserialize)]
-pub struct RootKeyshareData {
+pub struct RootKeyshareData<T: Clone + KeyshareId> {
```
robin-near, line 11: I think we need compatibility with existing keyshares stored on disk or on gcp:
 - The file path is different in the new implementation
 - The serde format is different - we used to store it as flat 3 members, but now it's nested.

I would suggest doing an implicit migration; i.e. on startup, if we discover that the new format doesn't exist but the old does, then create the new one.

For gcp, I don't know how that's going to work. We don't get access to create a separate secret for gcp, so I think we're going to have to be creative here and invent another encoded format that can store multiple keys.

Unfortunately, we do not have any migrate-ability (or backwards compatibility) tests.

node/src/signature_provider/ecdsa_secp256k1/kdf.rs
```diff
@@ -0,0 +1,195 @@
+use crate::hkdf::ScalarExt;
```
robin-near, line 1: :grimacing: This is a ton of moves and breaks history. Is it a lot of work for you to move these files back to the src directory and then follow up with another PR to move them?

Also nit on the package name - I'd suggest shortening, like signature_provider --> providers; ecdsa_secp256k1 -> secp256k1 or just ecdsa. These names do appear in tracing logs, and if it's too long it can make reading them less pleasant.

kuksag: > This is a ton of moves and breaks history. Is it a lot of work for you to move these files back to the src directory and then follow up with another PR to move them?

Will do

> Also nit on the package name

The reasoning behind was that a signature scheme is uniquely defined by the algorithm + its curve. But since we only have 1.5 providers, we can indeed shorten it without loss of distinguishability

kuksag:
Now this PR does following:
* introduces new `trait SignatureProvider`, implementation will follow in the next PR
* changes layout of `MpcTaskId` into `MpcTaskId::EcdsaTaskId` to ease further refactoring

Moreover, changes related to the `KeyshareStorage` will also come into its own different PR

bowenwang1996 (approved):

PR #265, opened 2025-03-18
test: reduce flakiness by reducing the amount of assets buffered in tests

bowenwang1996:
I noticed that we generate a lot of assets (triples and presignatures) in integration tests and they consume a lot of resources. Reducing the amount of assets generated helps reduce flakiness in some integration tests


robin-near:
haha Epic!

robin-near (approved):

PR #269, opened 2025-03-18
Replace `message_hash`, `tweak` and other types with the ones in contract code

kuksag:
Follow-up refactoring after #252 
This branch is stacked on top of #283 

* Reuse `PayloadHash` and `Epsilon` types from contract code
* Reuse `ScalarExt` trait from `crypto-shared`, remove duplicate


node/src/indexer/fake.rs
```diff
@@ -22,7 +21,7 @@ pub struct FakeMpcContractState {
     pub state: ContractState,
     // Not a real MPC contract; here we only index by the payload.
     // We don't test signatures with the same payload anyway.
-    pub pending_signatures: BTreeMap<Scalar, SignatureId>,
+    pub pending_signatures: BTreeMap<[u8; 32], SignatureId>,
```
kuksag, line 24: Compatibility is preserved. Checked with 
```
let mut bytes = [0; 32];
    OsRng.fill_bytes(bytes.as_mut());
    let scalar = Scalar::from_bytes(bytes).unwrap();
    let actual = scalar.to_bytes().to_vec();
    assert_eq!(bytes.as_slice(), actual);
```

bowenwang1996:
@kuksag I debugged this and it is indeed a serialization issue but in a different place. The problem is that in the [`respond` function](https://github.com/Near-One/mpc/blob/17ead40670bf6cb736b0f0975a6baeecc2d06516/libs/chain-signatures/contract/src/lib.rs#L310) inside the contract, it takes `SignatureRequest` as an argument. However, when you submit a transaction to call respond, the argument was formed [here](https://github.com/Near-One/mpc/blob/17ead40670bf6cb736b0f0975a6baeecc2d06516/node/src/indexer/tx_sender.rs#L147) which uses the modified version of `ChainSignatureRequest` that uses `[u8;32]` for payload instead of `Scalar`. As a result, transactions submitted by mpc nodes to call `respond` fail due to deserialization error. To properly fix this, I think you need to change the interface on the contract side as well. cc @near-bookrock

kuksag:
thank you @bowenwang1996. It seems contract interface and types inside node have to change at the same time. Since @near-bookrock has done deeper changes, I will close this PR

libs/chain-signatures/contract/src/crypto_shared/kdf.rs
```diff
@@ -1,18 +1,21 @@
...
+    Scalar, Secp256k1,
 };
 use near_account_id::AccountId;
 use sha3::{Digest, Sha3_256};
 
-// Constant prefix that ensures epsilon derivation values are used specifically for
+// Constant prefix that ensures tweak derivation values are used specifically for
 // near-mpc-recovery with key derivation protocol vX.Y.Z.
-const EPSILON_DERIVATION_PREFIX: &str = "near-mpc-recovery v0.1.0 epsilon derivation:";
+const EPSILON_DERIVATION_PREFIX: &str = "near-mpc-recovery v0.1.0 tweak derivation:";
```
robin-near, line 16: I don't think we can change this? It would break all the derived keys right?

libs/chain-signatures/contract/src/legacy_contract_state/mod.rs
```diff
@@ -306,14 +306,14 @@ impl ScalarExt for Scalar {
...
-        Scalar::from_bytes(hash).expect("Derived epsilon value falls outside of the field")
+        Scalar::from_bytes(hash).expect("Derived tweak value falls outside of the field")
     }
 }
 
 #[derive(BorshDeserialize, BorshSerialize, Serialize, Deserialize, Debug, Clone)]
 #[borsh(crate = "near_sdk::borsh")]
 pub struct SignatureRequest {
-    pub epsilon: SerializableScalar,
+    pub tweak: SerializableScalar,
```
robin-near, line 316: There's no point in changing the legacy contract state - I'd suggest just leaving the original naming in there just in case (as it *is* serialized afterall)

libs/chain-signatures/contract/src/primitives/signature.rs
```diff
@@ -1,9 +1,36 @@
 use crate::crypto_shared;
 
-use crypto_shared::{derive_epsilon, SerializableScalar};
-use k256::Scalar;
+use crypto_shared::derive_tweak;
 use near_sdk::{near, AccountId, CryptoHash};
 
+#[derive(Debug, Clone, Eq, Ord, PartialEq, PartialOrd)]
+#[near(serializers=[borsh, json])]
+pub struct Tweak([u8; 32]);
```
robin-near, line 8: Since we're doing this, should we also change the json encoding to a hex string instead of a long list of integers? @bowenwang1996 I think introducing "hex" as a dependency should be fine - it's very little code, and the derived json code is likely not that much better.

libs/chain-signatures/contract/src/crypto_shared/kdf.rs
kuksag: Oh, wow. Really nice catch. @near-bookrock please fix in #283, I will rebase afterwards.
@near-bookrock Also, please have a look at the other comments, as they are more suitable into your PR

libs/chain-signatures/contract/src/primitives/signature.rs
bowenwang1996: How does changing this struct help? Tweak is computed internally based on account id. It is not something given as an argument to a function

libs/chain-signatures/contract/src/crypto_shared/kdf.rs
near-bookrock: Good catch indeed. Reverted the change **https://github.com/Near-One/mpc/pull/283/commits/91b391faba25972b1874ea0e7e90b22ecc2a78f6**

libs/chain-signatures/contract/src/legacy_contract_state/mod.rs
near-bookrock: Good point. Reverted https://github.com/Near-One/mpc/pull/283/commits/c390eaac93149210956f004eda576c9104f3b13e

PR #270, opened 2025-03-18
feat: Add support for EdDSA/Frost to the MPC smart contract

near-bookrock:


libs/chain-signatures/rust-toolchain.toml
```diff
@@ -1,3 +1,16 @@
 [toolchain]
-channel = "1.81.0" # seems to be required by near runtime (for now)
+channel = "1.84.0"
```
kevindeforth, line 2: I once accidentally updated the cargo version and the contract [broke](https://nearone.slack.com/archives/C07UW93JVQ8/p1736355885343319?thread_ts=1735908645.069139&cid=C07UW93JVQ8), not sure if this has been fixed in the meantime.

PR #278, opened 2025-03-19
Expand Keyshare Storage functionality to support different signature providers

kuksag:
* Introduce "partial keyshare storage" - where we optionally load each keyshare.
This allows adding new signature providers (and thus new data to keyshare to serialize) without breaking existing schema
* Add back-compatibility test 
* Adjust coordinator's code, while preserving the same logic
* Remove `KeyshareId` trait trait as unnecessary

### DIFF
--- a/node/src/keyshare/local.rs
+++ b/node/src/keyshare/local.rs
@@ -1,4 +1,4 @@
-use super::{KeyshareStorage, RootKeyshareData};
+use super::{KeyshareStorage, PartialRootKeyshareData, RootKeyshareData};
 use crate::db;
 use aes_gcm::{Aes128Gcm, KeyInit};
 use anyhow::Context;
@@ -22,7 +22,7 @@ impl LocalKeyshareStorage {
 
 #[async_trait::async_trait]
 impl KeyshareStorage for LocalKeyshareStorage {
-    async fn load(&self) -> anyhow::Result<Option<RootKeyshareData>> {
+    async fn load(&self) -> anyhow::Result<Option<PartialRootKeyshareData>> {
         let cipher = Aes128Gcm::new(GenericArray::from_slice(&self.encryption_key));
         let keyfile = self.home_dir.join("key");
         if !keyfile.exists() {
@@ -32,7 +32,7 @@ impl KeyshareStorage for LocalKeyshareStorage {
             .await
             .context("Failed to read keygen file")?;
         let decrypted = db::decrypt(&cipher, &data).context("Failed to decrypt keygen")?;
-        let keyshare: RootKeyshareData =
+        let keyshare: PartialRootKeyshareData =
             serde_json::from_slice(&decrypted).context("Failed to parse keygen")?;
         Ok(Some(keyshare))
     }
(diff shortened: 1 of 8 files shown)


node/src/keyshare/mod.rs
```diff
@@ -1,34 +1,76 @@
...
+            // We are still generating a key share for some signature provider -> we have to make sure,
+            //  that we do not overwrite existing key shares
+            None => {
+                if let Some(ecdsa) = &existing.ecdsa {
+                    if ecdsa.private_share != to_save.ecdsa.private_share {
+                        return Err(anyhow::anyhow!(
+                            "Refusing to overwrite existing ecdsa keyshare"
+                        ));
+                    }
+                }
```
kuksag, line 70: I couldn't think of a way how to enforce invariants here, i.e. how would a user adding a new keyshare know, that this code needs to be adjusted?

node/src/keyshare/local.rs
```diff
@@ -22,7 +23,7 @@ impl LocalKeyshareStorage {
 
 #[async_trait::async_trait]
 impl KeyshareStorage for LocalKeyshareStorage {
-    async fn load(&self) -> anyhow::Result<Option<RootKeyshareData>> {
+    async fn load(&self) -> anyhow::Result<Option<PartialRootKeyshareData>> {
```
bowenwang1996, line 26: unrelated to this PR, but I don't see why this function needs to be async

node/src/keyshare/mod.rs
```diff
@@ -1,34 +1,76 @@
...
-    pub public_key: AffinePoint,
+    pub ecdsa: <EcdsaSignatureProvider as SignatureProvider>::KeygenOutput,
+}
+
+/// A helper structure that allows loading each share separately.
+/// It's useful when we want to introduce a new signature provider without breaking back-compatibility.
+#[derive(Clone, Serialize, Deserialize)]
+pub struct PartialRootKeyshareData {
+    pub epoch: u64,
+    pub ecdsa: Option<<EcdsaSignatureProvider as SignatureProvider>::KeygenOutput>,
```
bowenwang1996, line 21: Do you need the `as SignatureProvider` part here?

bowenwang1996 (commented):
We need a review from @robin-near as he is working on the multi-domain change

robin-near:
I'm sorry, I don't think I quite follow the strategy you're trying to take. And I apologize for not giving you this important design that I wrote last week - that was my oversight. https://docs.google.com/document/d/17y1lyyYwB5_iq8abv7A10XqAFjYtz8Ak83Wb3eH8j70/edit?tab=t.0 Could you take a look at this - particularly the "Key Storage" section?

Also https://github.com/Near-One/mpc/pull/266 is the PR I have for the contract-side multi-domain changes, if it's relevant.

Let me know your thoughts!

node/src/keyshare/local.rs
kuksag: oh, that's why there was this commit: https://github.com/Near-One/mpc/pull/267/commits/14477ea299011a62f3362c9d8d2e7bb8a4641efd that I've reverted...
Now it makes sense

PR #281, opened 2025-03-20
chore: remove self dependency to `legacy_contract`

near-bookrock:


libs/chain-signatures/contract/src/lib.rs
```diff
@@ -576,9 +577,7 @@ impl VersionedMpcContract {
     #[init(ignore_state)]
     #[handle_result]
     pub fn migrate() -> Result<Self, Error> {
-        if let Some(legacy_contract::VersionedMpcContract::V1(state)) =
-            env::state_read::<legacy_contract::VersionedMpcContract>()
-        {
+        if let Some(state) = env::state_read::<legacy_contract_state::MpcContractV1>() {
```
kevindeforth, line 580: I think we need to deserialize the previous [`VersionedMpcContract`](https://github.com/Near-One/mpc/blob/53f549cf5a58b9072b2f5a13d2e96c78e70ebc7d/libs/chain-signatures/contract/src/lib.rs#L58).

near-bookrock:
`pytest tests/test_contract_update.py::test_contract_update -s -v` is passing

robin-near:
Could you do cargo clippy as well as cargo fmt just to be sure - I feel some code might be dead code, and the CI doesn't run for the contract dir yet.

robin-near (approved):

robin-near:
Actually let me merge for now coz you have another stacked on top.

PR #282, opened 2025-03-20
chore: move `crypto-shared` to a module in contract

near-bookrock:
Stacked on top of https://github.com/Near-One/mpc/pull/281


bowenwang1996 (commented):
Do you actually need all the implementations from legacy contract, i.e, what is included in `legacy_contract_state/mod.rs`? I thought only the structs are needed

near-bookrock:
> Do you actually need all the implementations from legacy contract, i.e, what is included in `legacy_contract_state/mod.rs`? I thought only the structs are needed

I guess you are referencing the changes that were introduced in https://github.com/Near-One/mpc/pull/281? Yes, most of them were needed. I tried first with just the structs, and added the impl blocks as needed. 

It could be that some few methods in the big impl blocks are dead code.

bowenwang1996 (approved):
Please check dependencies added to the smart contract

near-bookrock:
> Please check dependencies added to the smart contract

Good spot.

All the added dependencies are migrated from the [`crypto-shared/cargo.toml`](https://github.com/Near-One/mpc/pull/282/files#diff-9c1b658aa9e295e08244a9ae428fe5fbbbf1a2408bf087815ac58afcb5edd870) that is being deleted as part of moving it into the contract crate.

`subtle` is the only unused dependency, so I deleted it now.

PR #283, opened 2025-03-20
refactor: use `[u8; 32]` instead of Scalar type from `k256` crate in contract

near-bookrock:
Stacked on top of https://github.com/Near-One/mpc/pull/282


kuksag:
I'm not sure whether the pytests will pass, because serialization format between node and contract has changed, as stated here https://github.com/Near-One/mpc/pull/269#issuecomment-2742128636
In that case you should change the node types as well like here: https://github.com/Near-One/mpc/pull/269/commits/9ed08bf6d4c27107e45738031b3f13f50825d1d5

libs/chain-signatures/contract/Cargo.toml
```diff
@@ -19,12 +24,17 @@ k256 = { version = "0.13.4", features = [
     "arithmetic",
     "expose-field",
 ] }
-crypto-shared = { git = "https://github.com/near-one/mpc", rev = "1d4954dff28e8eb988fb7762eff414a602a2b124", package = "crypto-shared" }
 near-gas = { version = "0.2.5", features = ["serde", "borsh", "schemars"] }
+near-account-id = "1"
 thiserror = "1"
+sha3 = "0.10.8"
```
bowenwang1996, line 30: What is this used for?

libs/chain-signatures/contract/src/primitives/signature.rs
```diff
@@ -1,7 +1,36 @@
-use crypto_shared::{derive_epsilon, SerializableScalar};
-use k256::Scalar;
+use crate::crypto_shared;
+
+use crypto_shared::derive_epsilon;
 use near_sdk::{near, AccountId, CryptoHash};
 
+#[derive(Debug, Clone, Eq, Ord, PartialEq, PartialOrd)]
+#[near(serializers=[borsh, json])]
+pub struct Epsilon([u8; 32]);
```
bowenwang1996, line 8: I think we should name it `Tweak` to be consistent with naming inside mpc node code

libs/chain-signatures/contract/Cargo.toml
```diff
@@ -19,12 +24,17 @@ k256 = { version = "0.13.4", features = [
...
+subtle = "2.6.1"
```
bowenwang1996, line 31: same question for this dependency

bowenwang1996 (commented):
In general looks good. Thank you for using new types instead of directly using `[u8; 32]`! One thing I'd like to ask is to double check the dependencies you are adding. Unless they are absolutely necessary, we should not add more dependencies to the contract, as it would increase security risks and also the size of the contract, which leads to higher gas costs for every function call.

libs/chain-signatures/contract/src/primitives/signature.rs
near-bookrock: Good point

near-bookrock: Also replaced all occurrences of epsilon to tweak

libs/chain-signatures/contract/Cargo.toml
near-bookrock: Here https://github.com/Near-One/mpc/blob/27af67ea1f773805cad6c81e573efa21a8dc2974/libs/chain-signatures/contract/src/crypto_shared/kdf.rs#L12

libs/chain-signatures/contract/Cargo.toml
near-bookrock: Subtle is not used. Removed it.

bowenwang1996 (approved):

libs/chain-signatures/contract/src/crypto_shared/kdf.rs
```diff
@@ -1,18 +1,21 @@
...
+    Scalar, Secp256k1,
 };
 use near_account_id::AccountId;
 use sha3::{Digest, Sha3_256};
 
-// Constant prefix that ensures epsilon derivation values are used specifically for
+// Constant prefix that ensures tweak derivation values are used specifically for
 // near-mpc-recovery with key derivation protocol vX.Y.Z.
-const EPSILON_DERIVATION_PREFIX: &str = "near-mpc-recovery v0.1.0 epsilon derivation:";
+const EPSILON_DERIVATION_PREFIX: &str = "near-mpc-recovery v0.1.0 tweak derivation:";
```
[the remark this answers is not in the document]
near-bookrock: You can rebase now.

PR #285, opened 2025-03-21
Implement multi-domain contract-coordinated key storage

robin-near:
This largely adheres to the specification for key storage in this design: https://docs.google.com/document/d/17y1lyyYwB5_iq8abv7A10XqAFjYtz8Ak83Wb3eH8j70/edit?tab=t.0

The strategy I've used here is to actually keep the implementation compatible with the rest of the code, so we can individually merge this PR while keeping the tests passing. However, **this is NOT a good version to actually release to anyone**, **NOR is it a proof or even a suggestion that we can implement this in a backwards-compatible way**, because:
* This will migrate the key storage format, so if anyone runs it in production, it will permanently alter their key storage. We might end up changing things before V2 is actually released, so this would not be good.
* For key import, the imported key *should* still be compatible with old node, but... I would not recommend the risk, so @andrei-near I would recommend that we use a binary older than this PR to import keys.
* Key generation and resharing implementation is no longer resistant to failure. So we should not run these in production either. The reason they are no longer resistant to failure is because keeping that the case with the new key storage would be cumbersome, and we're replacing that code in a matter of days so it isn't worth the trouble.

The addition to the specification is to perform stringent sanity checks before doing keygen and resharing. That's to ensure that when the contract transitions to the Running state, the local node will be able to follow suit. It's not necessary as long as nobody messes with the local storage, but if someone deletes some files in local storage then this would be able to catch issues early.

### DIFF
--- a/node/src/keyshare/mod.rs
+++ b/node/src/keyshare/mod.rs
@@ -3,25 +3,82 @@ pub mod local;
 
 use cait_sith::KeygenOutput;
 use k256::{AffinePoint, Scalar, Secp256k1};
+use mpc_contract::primitives::{
+    domain::DomainId,
+    key_state::{AttemptId, EpochId, KeyEventId},
+};
 use serde::{Deserialize, Serialize};
 
-/// The root keyshare data along with an epoch. The epoch is incremented
-/// for each key resharing. This is the format stored in the old MPC
-/// implementation, and we're keeping it the same to ease migration.
+use crate::indexer::participants::ContractKeyset;
 #[derive(Clone, Serialize, Deserialize)]
-pub struct RootKeyshareData {
-    pub epoch: u64,
+pub struct Secp256k1Data {
     pub private_share: Scalar,
     pub public_key: AffinePoint,
 }
+#[derive(Clone, Serialize, Deserialize)]
+pub enum KeyShareData {
+    Secp256k1(Secp256k1Data),
+}
 
-impl RootKeyshareData {
+// The keystore is a mess right now, because changes will depend on PR #278.
+#[derive(Clone, Serialize, Deserialize)]
+pub struct KeyShare {
+    pub key_id: KeyEventId,
+    pub data: KeyShareData,
+}
+impl KeyShare {
+    //pub fn domain_id(&self) -> DomainId {
+    //    self.key_id.domain_id
+    //}
+    pub fn epoch_id(&self) -> EpochId {
+        self.key_id.epoch_id
+    }
+    pub fn attempt_id(&self) -> AttemptId {
+        self.key_id.attempt_id
+    }
     pub fn keygen_output(&self) -> KeygenOutput<Secp256k1> {
         KeygenOutput {
             private_share: self.private_share,
             public_key: self.public_key,
         }
     }
+    pub fn new(key_id: KeyEventId, keygen_output: KeygenOutput<Secp256k1>) -> Self {
+        Self {
+            key_id,
+            private_share: keygen_output.private_share,
+            public_key: keygen_output.public_key,
+        }
+    }
+}
+pub fn migrate_root_key_share_data(root_keyshare: RootKeyshareData) -> KeyShare {
+    KeyShare {
+        key_id: KeyEventId::new(
+            EpochId::new(root_keyshare.epoch),
+            DomainId::legacy_ecdsa_id(),
+            AttemptId::legacy_attempt_id(),
+        ),
+        private_share: root_keyshare.private_share,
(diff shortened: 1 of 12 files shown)


node/src/keyshare/permanent.rs
```diff
@@ -0,0 +1,222 @@
...
+
+/// Backend for storing the permanent keyshare data.
+#[async_trait::async_trait]
+pub trait PermanentKeyStorageBackend: Send + Sync {
+    /// Loads the latest stored data; or None if no data was ever stored.
+    async fn load(&self) -> anyhow::Result<Option<Vec<u8>>>;
+    /// Stores the data, making it the latest. The identifier is used for the local backend to write
+    /// the data to a separate file and then linking it to the main file (acting as a backup
+    /// mechanism).
+    async fn store(&self, data: &[u8], identifier: &str) -> anyhow::Result<()>;
```
kevindeforth, line 33: We fail in local if the given identifier already exists, whereas in gcp we do not.
Because we check the existence of the file on the `PermamentKeyStore` level, it should not be an issue in practice, but maybe it would be good specify this in the comment.
Or maybe change the behavior of the local storage as well and rename this function `store_or_override` or sth similar.

kevindeforth (approved):
Thank you!
For the compatibility issue, how about before pushing to main we increment the version of the node and make it alpha or something?

node/src/keyshare/mod.rs
```diff
@@ -1,80 +1,396 @@
...
 }
 
-impl RootKeyshareData {
-    pub fn keygen_output(&self) -> KeygenOutput<Secp256k1> {
-        KeygenOutput {
-            private_share: self.private_share,
-            public_key: self.public_key,
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub enum KeyshareData {
+    Secp256k1(Secp256k1KeyshareData),
```
kuksag, line 26: Wouldn't it be better to link it to the existing type of a `SignatureProvider::KeygenOutput`? To avoid duplications and further type-conversions

kuksag (approved):

node/src/keyshare/mod.rs
```diff
@@ -1,80 +1,396 @@
...
+    }
+
+    /// Before generating a key, we must call `ensure_can_generate_key` to check that we are able
+    /// to generate that key and use it afterwards. This requires:
+    ///  - The already generated keys exist either in permanent or temporary storage.
+    ///  - The current permanent key storage is either
+    ///    - In the same epoch as the key generation attempt, and whose keys are a prefix of the
+    ///      already generated keys.
+    ///    - In an older epoch. This can happen if we missed the previous transition from Resharing
+    ///      to Running before it transitions again into Initializing. This is fine.
```
kevindeforth, line 111: I have some trouble understanding this. Shouldn't we require that each node has a valid keyshare for each generated key in the contract? Not just a subset of those?

If the contract is in generating state and expects the nodes to have a key with a certain id, then a node not having that key seems problematic.

The epoch id mismatch due to missing a running state seems like it also should not happen: when we reshare a key, we index the new keyshare under the next epoch id, not the current epoch id. And if we transition from running to initializing, then the epoch id does not change.

node/src/keyshare/mod.rs
```diff
@@ -1,80 +1,396 @@
...
+        self.temporary
+            .delete_keyshares_prior_to_epoch_id(keyset.epoch_id)
+            .await?;
+        Ok(new_keyshares)
+    }
+
+    /// Helper function to verify that the keyshares we have from permanent storage is a prefix
+    /// of the expected keyset, i.e. there are no extra keys, and each key is consistent with the
+    /// expected keyset.
+    async fn verify_existing_keyshares_are_prefix_of_keyset(
```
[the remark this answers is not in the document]
robin-near: Applied; except that the reason why I don't pass in a Keyset is because in the keygen checking code path there is no Keyset. We'd have to artificially construct one (which is fine but like, it isn't really the semantics of Keyset).

node/src/keyshare/permanent.rs
robin-near: Good catch! I wanted to tag it with a name in GCP too but GCP doesn't let me tag a secret version with anything at all. I don't really want to override it though, coz once it has the ability to override anything, there's the possibility of corrupting something useful. So I think the issue is really that on the filesystem we don't have a "append" mechanism, whereas on GCP we do. But since the collision on disk is expected to never happen (unless there's a bug), I think that code can be considered unreachable and thus the discrepancy in the behavior can be ignored.

PR #291, opened 2025-03-22
[DOES NOT COMPILE; for Kevin's change] Draft on integration test refactoring for contract update.

robin-near:


kevindeforth:
Thanks a lot, I used some of the logic in #295 

Side note: pythons `random.choice` may choose the same element multiple times. `random.sample` was probably what you intended to use.

PR #294, opened 2025-03-24
feat(EdDSA): Add support for EdDSA signature requests on the smart contract

ghost:
This PR adds support for submitting sign requests to MPC smart contract. It introduces an optional `domain_id` parameter to the `sign` method.
```rust
pub struct SignRequest {
    pub payload: PayloadHash,
    pub path: String,
    pub key_version: u32,
    pub domain_id: Option<DomainId>,
}
```

The return body object `SignatureResponse` is also changed from a struct to an enum to support Ed25519 signatures. Ed25519 signatures do not contain a `recovery_id` field, whereas `Secp256k1` do.

The new enum `SignatureResponse` contains two variants:

```rust
pub enum SignatureResponse {
    Secp256k1(k256_types::SignatureResponse),
    Edd25519(edd25519_types::SignatureResponse),
}
```
The `Secp256k1` variant contains the same fields as before, meaning backwards compatibility is preserved. The Edd25519 variant is just a tuple struct containing a 64 byte array representing the raw signature.

### DIFF
--- a/libs/chain-signatures/contract/Cargo.toml
+++ b/libs/chain-signatures/contract/Cargo.toml
@@ -11,7 +11,7 @@ anyhow = "1"
 curve25519-dalek = { version = "4.1.3", features = [
     "group",
     "serde",
-], default-features = true }
+], default-features = false }
 borsh = "1.5.0"
 near-sdk = { version = "5.11.0", features = [
     "legacy",
--- a/libs/chain-signatures/Cargo.lock
+++ b/libs/chain-signatures/Cargo.lock
@@ -107,15 +107,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atomic-polyfill"
-version = "1.0.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8cf2bce30dfe09ef0bfaef228b9d414faaf7e563035494d7fe092dba54b300f4"
-dependencies = [
- "critical-section",
-]
-
-[[package]]
 name = "atomic-waker"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -471,12 +462,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f46ad14479a25103f283c0f10005961cf086d8dc42205bb44c46ac563475dca6"
 
 [[package]]
-name = "cobs"
-version = "0.2.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "67ba02a97a2bd10f4b59b25c7973101c79642302776489e030cd13cdab09ed15"
-
-[[package]]
 name = "codespan-reporting"
 version = "0.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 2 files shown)


libs/chain-signatures/contract/Cargo.toml
```diff
@@ -8,13 +8,21 @@ crate-type = ["cdylib", "lib"]
 
 [dependencies]
 anyhow = "1"
+curve25519-dalek = { version = "4.1.3", features = [
+    "group",
+    "serde",
+], default-features = true }
+frost-ed25519 = { version = "2.1.0", default-features = true, features = [
+    "serialization",
+] }
```
bowenwang1996, line 17: What do we need this dependency for?

libs/chain-signatures/contract/src/crypto_shared/types.rs
```diff
@@ -1,141 +1,254 @@
...
+    }
 
-// Is there a better way to force a borsh serialization?
-#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, Clone, Copy, Ord, PartialOrd)]
-pub struct SerializableScalar {
-    pub scalar: Scalar,
-}
+    impl BorshSerialize for SerializableScalar {
+        fn serialize<W: std::io::prelude::Write>(&self, writer: &mut W) -> std::io::Result<()> {
+            let to_ser: [u8; 32] = self.scalar.to_bytes().into();
```
bowenwang1996, line 69: you don't need the call to `.into` here as Borsh can work on `&[u8]`

near-bookrock: [`to_bytes()`](https://docs.rs/k256/latest/k256/struct.Scalar.html#method.to_bytes) returns [`FieldBytes`](https://docs.rs/k256/latest/k256/type.FieldBytes.html), not a byte array.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -282,19 +315,36 @@ impl VersionedMpcContract {
...
+
+        let derived_public_key = match curve_type {
+            CurveType::SECP256K1 => {
+                let derived_public_key =
+                    derive_key_secp256k1(&near_public_key_to_affine_point(public_key), &tweak);
+                let encoded_point = derived_public_key.to_encoded_point(false);
+                let slice: &[u8] = &encoded_point.as_bytes()[1..65];
+                let mut data: Vec<u8> = vec![near_sdk::CurveType::SECP256K1 as u8];
+                data.extend(slice.to_vec());
+                PublicKey::try_from(data)
```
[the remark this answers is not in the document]
near-bookrock: Good point. I kept the implementation the same as before, which used `PublicKey::try_from`. I'll update it to use `from_parts`.

https://github.com/Near-One/mpc/pull/294/files/398b7d0d6c64d54093a8794a49d8c11bbbc54602#diff-f31f89ad11da217fc15c41b4822ee9f4798bbcbb643341e49157c7747f87669bL297

bowenwang1996 (approved):
Approve to unblock. I think @kevindeforth is working on adding some tests

libs/chain-signatures/contract/Cargo.toml
near-bookrock: I pushed a commit for now that removes the dependency, and uses [ED25519_BASEPOINT_POINT](https://docs.rs/curve25519-dalek/latest/curve25519_dalek/constants/constant.ED25519_BASEPOINT_POINT.html).

Issue #241, opened 2025-03-03
[mpc-contract] Improve signature state


kevindeforth:
The current signature state has a few issues:

- successful signature requests are immediately removed from the `pending_request` map, but stay in the `requests_by_blockheight` vector for at least another 200 blocks. This is an issue during load testing, as discovered by @robin-near. Ideally, succcesful signatures are removed from the state immediately.
- It is possible that a signature request times out and is not removed from the state for a while, due to how the cleaning logic works. This is problematic because the contract [rejects signature requests](https://github.com/Near-One/mpc/blob/39925623e859861883d6c3ed58e39599a5efd51e/libs/chain-signatures/contract/src/lib.rs#L260) that are already in the state.
- the gas cost for removing timed out signature requests can be drastically improved by using a different data structure than a vector.

The new solution should ensure that we do not run out of gas during migration and utilize inheritance from storage keys where appropriate.

(also take not of [this](https://github.com/Near-One/mpc/pull/253#discussion_r1998307837) discussion)

PR #295, opened 2025-03-24
Contract coordinated key resharing

kevindeforth:
This draft implements the contract coordinated key resharing mechanism as outlined in #195 and integrates the new v2 contract API with the node and for the pytests.

This Code DOES NOT actually support multiple domains, due to the discussion about triplet stores. It is imperative that we either add that support in a later PR, or disable more than one domain for the same signature scheme in the contract before the releasing.

The tests all pass, but the code inside `coordinator.rs` and `indexer/participants.rs` is likely to change over the next 24 hours, because:
- its robustness & performance might improve by using async rust in the follower loop (which I did not yet manage to implement correctly)
- its structure should be improved to ease integration of multiple domains & signature types.

Follow-ups in later PRs:
- [ ] migration tests for the contract
- [ ] add a test for the `remove_timed_out_request` endpoint after #241


node/src/coordinator.rs
```diff
@@ -317,29 +342,155 @@ impl Coordinator {
...
+                    }
+                    tracing::info!("Key generation complete; Leader calls vote_pk.");
+                    chain_txn_sender
+                        .send(ChainSendTransactionRequest::VotePk(ChainVotePkArgs {
+                            key_event_id: contract_event.id,
+                            public_key: my_public_key,
+                        }))
+                        .await?;
+                } else {
+                    // note: if leader does not exit here, we just hang. Not sure why.
```
robin-near, line 489: I think if we're leader and we see that there is already an event started, we should send vote_abort (lol oops that doesn't even exist on the contract!) and then just sleep and loop. Eventually it will be aborted or time out and then we can start the next attempt.

node/src/coordinator.rs
```diff
@@ -317,29 +342,155 @@ impl Coordinator {
...
-                public_key: my_public_key,
-            }))
-            .await?;
-
-        // Just halt and wait for the running state.
-        Ok(MpcJobResult::HaltUntilInterrupted)
+        let is_leader = mpc_config.is_leader_for_keygen();
+        let keyshare_storage = Arc::new(keyshare_storage);
+        if !is_leader {
+            'follower: loop {
```
robin-near, line 348: I think this needs to be extracted to a separate function, so that when any of the `?`s errors out, we can loop again instead of exiting out of the whole thing.

(btw this should live in a separate file I think; it's not really the coordinator's job)

node/src/coordinator.rs
```diff
@@ -317,29 +342,155 @@ impl Coordinator {
...
+                                    "Received Keygeneration task id: {:?}; spawning task, waiting on contract change. Our id; {:?}" ,
+                                    task_id,
+                mpc_config.my_participant_id,
+                                );
+                let max_timeout = 120;
+                let key_event_receiver_cloned = key_event_receiver.clone();
+                let mpc_config_cloned = mpc_config.clone();
+                let chain_txn_sender_cloned = chain_txn_sender.clone();
+                let keyshare_storage_cloned = keyshare_storage.clone();
+                //tokio::spawn(async move {
```
robin-near, line 371: It isn't necessarily a problematic thing to not spawn. Handling one at a time is okay. Conceptually the follower does not have to do it serially, but in practice, handling one computation at a time isn't going to introduce any problems.

node/src/coordinator.rs
```diff
@@ -317,29 +342,155 @@ impl Coordinator {
...
+                // Wait for the contract to confirm this key event
+                let mut contract_event = key_event_receiver_cloned.borrow().clone();
+                let mut n = 0; // one minute max
+                while contract_event.id != task_key_event_id || !contract_event.started {
```
robin-near, line 375: I think this should be:

- While contract event not started and `state.generating_key.next_attempt_id == task_key_event_id.attempt_id`: wait for a change on `key_event_receiver_cloned`.
- Once we exit that loop, if we haven't started or if `contract_event.id != task_key_event_id`, `bail!("Not the active key ID")`

That way, we do not actually need to even have a timeout here. Why? Because the contract will naturally progress towards higher attempt IDs and this will exit whenever the appropriate timeout happens.

node/src/network/computation.rs
```diff
@@ -33,9 +33,7 @@ pub trait MpcLeaderCentricComputation<T>: Sized + 'static {
 
         // We'll wrap the following future in a timeout below.
         let fut = async move {
-            if !sender.is_leader() {
-                sender.wait_for_all_participants_connected().await?;
-            }
+            sender.wait_for_all_participants_connected().await?;
```
robin-near, line 36: FYI this isn't necessary for the leader since the way it chooses the participants already means the participants are connected. But it's also fine to make this change.

pytest/common_lib/shared.py
```diff
@@ -104,14 +102,39 @@ def sign_tx(self,
...
+        if self.status == MpcNode.NodeStatus.IDLE:
+            return f"⚫\033[90mI: {self.account_id}\033[0m"
+        if self.status == MpcNode.NodeStatus.PARTICIPANT:
+            return f"🟢\033[92mP: {self.account_id}\033[0m"
+        if self.status == MpcNode.NodeStatus.OLD_PARTICIPANT:
+            return f"🟢\033[92mO: {self.account_id}\033[0m"
+        if self.status == MpcNode.NodeStatus.NEW_PARTICIPANT:
+            return f"🟡\033[93mN: {self.account_id}\033[0m"
+        else:
+            return f"❓\033[0m?: {self.account_id}\033[0m"
```
robin-near, line 136: emoji printing is the best :+1: :+1: :+1:

robin-near (commented):
This looks really good! It's exciting that tests pass!

node/src/coordinator.rs
```diff
@@ -317,29 +342,155 @@ impl Coordinator {
...
+                    tracing::info!("leader concluded computation, storing keyshare.");
+                    let keyshare = Keyshare {
+                        key_id: contract_event.id.clone(),
+                        data: KeyshareData::Secp256k1(res.clone()),
+                    };
+                    keyshare_storage.store_key(keyshare).await?;
+                    tracing::info!("leader stored keyshare.");
+                    let my_public_key = affine_point_to_public_key(res.public_key)?;
+                    let key_event_id = contract_event.id;
+                    while contract_event.completed.len() != n_participants - 1 {
```
robin-near, line 471: There's no need to wait for follower votes.

When leader finishes generation it can just go ahead and vote, and just enter the next iteration. Then it naturally waits for the contract to not be started again (or this state gets killed because we're in Running now) and it'll start another computation. We implemented all the great logic at the contract side so we don't need this extra coordination in the node :)

node/src/coordinator.rs
kevindeforth: #297

node/src/key_events.rs
```diff
@@ -0,0 +1,277 @@
...
+        Ok::<_, anyhow::Error>(contract_event)
+    })
+    .await??;
+    if contract_event.id != key_event_id {
+        anyhow::bail!("key event changed");
+    }
+    Ok(contract_event)
+}
+
+pub fn convert(public_key: &near_sdk::PublicKey) -> anyhow::Result<AffinePoint> {
```
robin-near, line 202: Can some of these functions be private?

robin-near (approved):

PR #296, opened 2025-03-25
Simplify handling of pending signature requests

### LINKED ISSUE #241: [mpc-contract] Improve signature state

bowenwang1996:
**Resolves #241**

Remove `requests_by_block_height` and associated logic in favor of a simpler approach where a pending request is removed when a timeout happens. In that case, it takes one more block due to another scheduled promise, but under normal circumstances latency remains the same. Also updated near-sdk-rs version to support yield/resume API in unit tests. Unfortunately due to the lack of compatibility between the rustc version required by near-sdk-rs 5.11 and the current version of near-workspaces, it is imperative that we use wasm-opt to strip the contract of unnecessary features. Otherwise the contract cannot be loaded into near-workspaces runtime and tests fail with `FunctionCallError(CompilationError(PrepareError(Deserialization)))`. It also has the benefit of reducing the size of compiled contract by 15%


libs/chain-signatures/contract/src/lib.rs
```diff
@@ -750,23 +709,32 @@ impl VersionedMpcContract {
...
+                ))
             }
-            Err(_) => Err(SignError::Timeout.into()),
         }
     }
 
+    #[private]
+    pub fn fail_on_timeout(&self) {
+        // To stay consistent with the old version of the timeout error
+        panic!("{}", SignError::Timeout.to_string());
```
robin-near, line 735: What's the diff between `env::panic_str` and `panic!`?

libs/chain-signatures/rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.81.0" # seems to be required by near runtime (for now)
+channel = "1.85.1" # seems to be required by near runtime (for now)
```
robin-near, line 2: Remove the comment?

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -732,7 +691,7 @@ impl VersionedMpcContract {
         env!("CARGO_PKG_VERSION").to_string()
     }
 
-    /// Upon success, removes the signature from state and returns it.
+    /// Upon success, removes the gature from state and returns it.
```
robin-near, line 694: typo

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -750,23 +709,32 @@ impl VersionedMpcContract {
     /// 3. `return_signature_and_clean_state_on_success` fails with `ContractSignatureRequest`
     ///    argument.
```
robin-near, line 710: I think we should update the comment here now.

robin-near (approved):
Wow, this is epic!

libs/chain-signatures/contract/readme.md
```diff
@@ -87,4 +87,8 @@ These functions require the caller to be a participant or candidate.
 Run tests with `cargo nextest run`. This includes the integration tests, for which it is necessary to compile the contract first:
 ```
 cargo build --release --target=wasm32-unknown-unknown
+wasm-opt -Oz -o mpc-contract.wasm target/wasm32-unknown-unknown/release/mpc_contract.wasm
+mv mpc-contract.wasm target/wasm32-unknown-unknown/release/mpc_contract.wasm
```
kevindeforth, line 91: Can you add this to the pytest fixture?

https://github.com/Near-One/mpc/blob/61c4d3ce6ae03ccd3b16f76e083f332f5dc3f3c4/pytest/tests/conftest.py#L42

kevindeforth (approved):
Very nice, I did not know this was possible! 
This removes the need for `test_contract_state_cleanup.py`, so the file can be deleted.
You can use the changes from #295 to run the test suite against the new contract

libs/chain-signatures/contract/src/lib.rs
bowenwang1996: panic_str is slightly better because it doesn't bring in `String` dependencies, but we already use `format!` in other places, so realistically it is not different

PR #299, opened 2025-03-25
Abort endpoint for contract & race condition fix

### LINKED ISSUE #297: leader should vote abort in case the key event was started without them
Can happen if they e.g. reset or crash. 
Requires a vote_abort_key_event() endpoint in the contract.

note: the added contract endpoint will also simplify testing of  the attempt logic.

kevindeforth:
This PR:
- exposes the `abort_key_event` endpoint on the contract and adjust the mpc node for the new API from #293;
- proposes a potential fix for an issue that arose during a pytest: https://github.com/Near-One/mpc/pull/299#discussion_r2014002157;
- removes the unused pytest which was accidentally back in #295  (merge conflicts);
- fixes a race condition due to indexer lag that could result in attempting a successful or partially successful key event instance twice.

The race condition is fixed by making their temporary keystore the source of truth:
- Before starting or following a key event instance, participants check if they are in possession of a matching keyshare by querying their temporary keystore.
- If so, they re-submit a vote to the contract in favor of concluding the current attempt and do not participate in the protocol computation.

Due to the full-threshold requirement in the underlying mpc  protocol, this ensures that if any participant is in possession of a keyshare with Key Id `{epoch, domain, attempt}`, then we will never start another attempt with the same key id.
Note that this does allow participants to compute keyshares for the same key id multiple times, as long as no participant ever successfully stored any of their computed keyshares to disk.

Additional sanity check:
At the end of a key event instance, leader and followers now query their temporary keystore for the existence of the keyshare they just stored in it. They vote success if successful and abort if unsuccessful. This is a sanity check.

resolves #297


node/src/key_events.rs
```diff
@@ -245,7 +314,7 @@ pub struct ResharingArgs {
 
 /// Handles the follower side of a key resharing task.
 /// - Waits for a new network task channel from `channel_receiver`.
-/// - Ignores non-resharing tasks until a resharing task is received.
+/// - Drops the channel and bails if task id other than resharing is received.
```
kevindeforth, line 317: @robin-near : Can you check if this is correct?

In a pytest, I observed this in one of the follower nodes:
```
[2m2025-03-26T11:17:19.900879Z[0m [32m INFO[0m [2mmpc_node::key_events[0m[2m:[0m Expected Key resharing task id, received: EcdsaTaskId(ManyTriples { start: UniqueId(0, 1742987670, 0), count: 64 }); ignoring.
```
There were 80 millions lines like this in stdout before the pytest timed out.

robin-near (approved):

PR #300, opened 2025-03-26
Devnet: vote leave cmd

andrei-near:
implementing vote_leave command to allow removing nodes from the cluster


robin-near:
Oh, this is outdated now; the new V2 contract uses different API for joining and leaving. The devnet scripts would need some changes.

robin-near (approved):

andrei-near:
@robin-near  we can review these later. Merging to have them in the main for now

PR #310, opened 2025-03-28
Eddsa signature provider

kuksag:
Should be merged after https://github.com/Near-One/cait-sith/pull/12.

Add signature provider implementation for Eddsa signature scheme:
* Signing
* DKG
* Resharing 

Follow-up PR should containt:
* Further integration into the logic
* Proper e2e testing

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -10,7 +10,7 @@ anyhow = "1.0.92"
 async-trait = "0.1.83"
 axum = "0.7.9"
 borsh = { version = "1.5.1", features = ["derive"] }
-cait-sith = { git = "https://github.com/kuksag/cait-sith", rev = "84626e50b3bf9135275f9ae610d05674bcba661f" }
+cait-sith = { git = "https://github.com/kuksag/cait-sith", rev = "e89e3d96900c52f4e1097d4031dd8ee3b1eb60df" }
 clap = { version = "4.5.20", features = ["derive", "env"] }
 flume = "0.11.1"
 futures = "0.3.31"
--- a/node/src/mpc_client.rs
+++ b/node/src/mpc_client.rs
@@ -272,6 +272,13 @@ impl MpcClient {
                     .process_channel(channel)
                     .await?
             }
+            MpcTaskId::EddsaTaskId(_) => {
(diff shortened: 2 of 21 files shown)


node/Cargo.toml
```diff
@@ -10,7 +10,7 @@ anyhow = "1.0.92"
 async-trait = "0.1.83"
 axum = "0.7.9"
 borsh = { version = "1.5.1", features = ["derive"] }
-cait-sith = { git = "https://github.com/kuksag/cait-sith", rev = "84626e50b3bf9135275f9ae610d05674bcba661f" }
+cait-sith = { git = "https://github.com/kuksag/cait-sith", rev = "e89e3d96900c52f4e1097d4031dd8ee3b1eb60df" }
```
kuksag, line 13: Blocked on https://github.com/Near-One/cait-sith/pull/12

kuksag: JFYI: Original line also should've been "github.com/**Near-One**/cait-sith", instead of "github.com/**kuksag**/cait-sith".
It was brought by mistake in https://github.com/Near-One/mpc/pull/303, sorry

node/src/mpc_client.rs
```diff
@@ -272,6 +272,13 @@ impl MpcClient {
                     .process_channel(channel)
                     .await?
             }
+            MpcTaskId::EddsaTaskId(_) => {
+                unreachable!()
+                // self.eddsa_signature_provider
+                //     .clone()
+                //     .process_channel(channel)
+                //     .await?
```
bowenwang1996, line 280: Why is this not implemented?

robin-near (approved):

node/src/mpc_client.rs
robin-near: I played with this a bit. We're lacking support for storing keyshares for eddsa and hooking up the keygen and resharing with the keyshare (this needs the eddsa KeygenOutput to be serializable). On the keygen/resharing side, we're missing a DomainRegistry parameter, otherwise it doesn't know what kind of key to generate (it has the DomainId but not the DomainConfig which contains the SignatureScheme). So yeah there's still some hooking up to do.

PR #313, opened 2025-03-28
Python test for contract migration

### LINKED ISSUE #280: Test contract V2 migration logic
It's likely incorrect right now.

kevindeforth:
Test migrating from the new contract to another binary contract and tests migrating from the legacy contract to the new contract.

Resolves #280

1. deploy V2 and generate a new keyset on account_id "test0"
2. migrate from V2 to a dummy contract that does not handle any signature requests.*
3. kill the mpc nodes & deploy V1 contract to account_id "test1"
4. use the public key and participants from step 1 to initialize the contract under "test1" with `init_running`.
5. change the config entry for the mpc nodes, s.t. their mpc_contract now points to "test1" and no longer "test0". Bring the nodes back up.
6. vote to migrate the contract from V1 to V2.
7. do succesful signature requests.

(*step 2 is a fail-safe, in case the nodes were still pointing ant "test0"  instead of "test1")

It's not a perfect test, but for now, good enough. We will need to test the migration of the entire cluster (incl. mpc node code), which will allow us to test anything not covered in this one (such as trailing signature requests).


pytest/tests/test_contract_update.py
```diff
@@ -11,146 +11,125 @@
...
+        "next_id": 2,
+        "participants": participants_map,
+        "account_to_participant_id": account_to_participant_id,
+    }
+
+
+def test_contract_update():
+    update_args = UpdateArgsV1(COMPILED_CONTRACT_PATH)
+    cluster, mpc_nodes, public_key = deploy_and_init_v2()
+    #public_key = cluster.contract_state().keyset().keyset[0].key
```
bowenwang1996, line 74: Delete this line?

pytest/tests/test_contract_update.py
```diff
@@ -11,146 +11,125 @@
...
+    cluster.contract_node.send_txn_and_check_success(tx, 20)
+    tx = cluster.contract_node.sign_tx(cluster.mpc_contract_account(),
+                                       'version',
+                                       json.dumps({}).encode('utf-8'), 1,
+                                       150 * TGAS)
+    cluster.contract_node.send_txn_and_check_success(tx, 20)
+
+    cluster.define_candidate_set(mpc_nodes[:2])
+    cluster.update_participant_status(assert_contract=False)
+    #time.sleep(5)
```
bowenwang1996, line 124: Remove this line?

bowenwang1996 (commented):
What is the purpose of the migration contract?

pytest/tests/test_contract_update.py
```diff
@@ -11,146 +11,125 @@
...
+    for i, p in enumerate(participants_config['participants']):
+        near_account = p['near_account_id']
+        my_pk = p['p2p_public_key']
+        my_addr = p['address']
+        my_port = p['port']
+
+        participants_map[near_account] = {
+            "account_id":
+            near_account,
+            "cipher_pk": [
```
robin-near, line 53: always funny to see a list of zeros in our integration tests :joy:

nit: [0] * 32

robin-near (approved):

robin-near (changes requested):
Ah yeah, Bowen's comment makes sense - we should be doing in-place upgrade instead of moving accounts.

kevindeforth:
> > change the config entry for the mpc nodes, s.t. their mpc_contract now points to "test1" and no longer "test0". Bring the nodes back up.
> 
> I don't think we should test this. It is not how we upgrade. Upgrades are done in place and we don't intend to switch to a different contract because it doesn't really help with anything

The upgrade is done in place:
- Account "test0": new V2 SC is used for coordinating the keygeneration and is then upgraded to a new contract (the one in the migrated folder)
- Account "test1": starts off with the legacy SC, which is then updated to the new V2 SC.

There is no way to do a migration from V2 to V1, because the migrate function in V1 won't support that. So we need two accounts.

robin-near (approved):

pytest/tests/test_contract_update.py
```diff
@@ -11,146 +11,116 @@
...
-    results = cluster.await_txs_responses(tx_hashes)
-    verif_helper = VerificationHelper()
-    shared.verify_txs(results, verif_helper.verify)
-    verif_helper.final_verification()
+    cluster.assert_is_deployed(update_args.code())
+    cluster.contract_state().print()
+    cluster.send_and_await_signature_requests(1)
+
+
+# todo: comprehensive test with entire cluster & node + config updates
```
robin-near, line 126: Let's just file an issue for this coz this isn't a code change right?

PR #314, opened 2025-03-28
eddsa integration 

kuksag:
* Integrate Eddsa signature provider logic
* Make each signature provider store related keys. It's a questionable change. The other approach is to supply share when needed: when creating a signature, when setting-up background task, etc.. But for now it seems ok 

tests for eddsa will come in the follow-up PRs

### DIFF
--- a/node/src/coordinator.rs
+++ b/node/src/coordinator.rs
@@ -16,14 +16,19 @@ use crate::metrics;
 use crate::mpc_client::MpcClient;
 use crate::network::{run_network_client, MeshNetworkTransportSender};
 use crate::p2p::new_tls_mesh_network;
+use crate::providers::eddsa::EddsaSignatureProvider;
 use crate::providers::EcdsaSignatureProvider;
 use crate::runtime::AsyncDroppableRuntime;
 use crate::sign_request::SignRequestStorage;
 use crate::tracking::{self};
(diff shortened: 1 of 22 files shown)


node/src/coordinator.rs
```diff
@@ -377,22 +378,38 @@ impl Coordinator {
...
 
         let sign_request_store = Arc::new(SignRequestStorage::new(secret_db.clone())?);
-        let KeyshareData::Secp256k1(keyshare) = &keyshares.first().unwrap().data;
+
+        let KeyshareData::Secp256k1(ecdsa_keyshare) = &keyshares.get(0).unwrap().data else {
+            anyhow::bail!("first share is expected to be an ecdsa");
+        };
+        let KeyshareData::Ed25519(eddsa_keyshare) = &keyshares.get(1).unwrap().data else {
+            anyhow::bail!("second share is expected to be an eddsa");
+        };
```
kuksag, line 387: will move from hardcoded values, to a `[Box<dyn SignatureProvider>]`, I guess

robin-near (approved):
Really nice! I especially like how you make each signature provider take multiple domains. I think that's the way to go, because now we just need to make the ecdsa presignature generation and retrieval support multiple keys.

PR #317, opened 2025-03-28
Clear out signature requests instead of migrating them.

robin-near:
Also
* Add contract tests to CI.
* Remove `max_num_requests_to_remove`
* Rename `max_event_idle_blocks` to `key_event_timeout_blocks` and default it to 30 instead of 50 (50 is too long... even 30 feels too long).


libs/chain-signatures/contract/src/lib.rs
```diff
@@ -725,46 +725,21 @@ impl VersionedMpcContract {
         if let Some(legacy_contract_state::VersionedMpcContract::V1(state)) =
             env::state_read::<legacy_contract_state::VersionedMpcContract>()
         {
-            // migrate config
             let mut config = Config::default();
             config.request_timeout_blocks = state.config.request_timeout_blocks;
             config.max_num_requests_to_remove = state.config.max_num_requests_to_remove;
```
kevindeforth, line 730: I don't think we need these anymore since #296 got merged.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -707,7 +707,7 @@ impl VersionedMpcContract {
             protocol_state: ProtocolContractState::Running(RunningContractState::new(
                 domains, keyset, parameters,
             )),
-            pending_requests: LookupMap::new(StorageKey::PendingRequests),
+            pending_requests: LookupMap::new(StorageKey::PendingRequestsV2),
             proposed_updates: ProposedUpdates::default(),
```
kevindeforth, line 711: What happens to the data under the old storage keys? Will it be persisted forever? Will it be deleted?
If we have any old updates, what will happen to them?

kevindeforth (approved):

.github/workflows/ci.yml
```diff
@@ -41,25 +41,43 @@ jobs:
           CARGO_TARGET_DIR="target/clippy" \
           RUSTFLAGS="-D warnings" \
           cargo clippy --all-features --all-targets --locked
+          cd ..
+
+          cd libs/chain-signatures
+          CARGO_TARGET_DIR="target/clippy" \
+          RUSTFLAGS="-D warnings" \
+          cargo clippy --all-features --all-targets --locked
+          cd ../..
```
kevindeforth, line 50: the `../..` at the end of the steps are not necessary.

robin-near: It's just to complete the pattern - I tripped myself by not `cd ..`ing at the end of the previous step :)

libs/chain-signatures/contract/src/lib.rs
robin-near: Persisted forever. But we already had that happening in the past anyway. We can just run manual clearing before the upgrade, it's not a big deal - nobody cares.

robin-near: I don't know about old updates, that I think needs a more careful look at.

libs/chain-signatures/contract/src/lib.rs
robin-near: Oh yeah, let me remove those too.

libs/chain-signatures/contract/src/lib.rs
robin-near: OK, I filed #318 for that; it indeed is a problem.

PR #319, opened 2025-03-28
Use real contract's logic in fake indexer.

robin-near:
This will facilitate testing of multi-domain, and strengthen testing of keygen and resharing, as well as making eddsa testing work with the fake automatically.


kuksag (approved):
very nice, thank you

PR #322, opened 2025-03-29
Separate channels for ecdsa/eddsa dkg

kuksag:


node/src/key_events.rs
```diff
@@ -366,13 +367,23 @@ pub async fn keygen_leader(
             }
         }
 
+        let channel = match domain.scheme {
```
robin-near, line 370: oh, interesting. I wonder if this warrants revising the design around the task IDs. Currently, the node works all fine even though the channel ID is always EcdsaTaskId::KeyGeneration/Resharing, because the keygen/resharing code doesn't care about the type of TaskId; it only cares about the key event ID. So I wonder if we should have something like

KeyGeneration
KeyResharing
EcdsaTask(Triple)
EcdsaTask(Presignature)
Signature

(signature is also self-sufficient in describing its domain)

kuksag: > signature is also self-sufficient in describing its domain

The signature task might contain domain-specific data, like `presignature_id` for ecdsa, which we need to be able to communicate to followers. This could have been avoided if there were a deterministic way to choose it

bowenwang1996 (approved):

node/src/key_events.rs
bowenwang1996: Let's merge it for now and discuss refactoring later

PR #328, opened 2025-04-01
fix: Add pre-computed edwards_point of EdDSA keys to contract state

near-bookrock:
The contract is currently storing EdDSA keys in a compressed format. The contract needs to decompress them in order to calculate the derived key.

The decompression step is expensive, and leads sign requests to consume more gas than the maximum allowed limit.

This PR adds the decompressed representation, `EdwardsPoint`, of each `EdDSA` public key to the contract state such that the contract will not need to compute it at runtime for every request.

### DIFF
--- a/libs/chain-signatures/contract/src/crypto_shared/types.rs
+++ b/libs/chain-signatures/contract/src/crypto_shared/types.rs
@@ -1,8 +1,10 @@
 use borsh::{BorshDeserialize, BorshSerialize};
+use curve25519_dalek::EdwardsPoint;
 use k256::{
     elliptic_curve::{bigint::ArrayEncoding, CurveArithmetic, PrimeField},
     AffinePoint, Secp256k1, U256,
 };
+use near_sdk::{near, PublicKey};
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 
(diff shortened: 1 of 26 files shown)


libs/chain-signatures/contract/src/crypto_shared/types.rs
```diff
@@ -12,6 +16,170 @@ pub enum SignatureResponse {
...
 
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+pub enum PublicKeyExtended {
+    Secp256k1 {
+        near_public_key: near_sdk::PublicKey,
+    },
+    Ed25519 {
+        near_public_key: near_sdk::PublicKey,
+        edwards_point: EdwardsPoint,
+    },
```
bowenwang1996, line 27: Add a comment on what these are

bowenwang1996 (commented):
@near-bookrock it seems that you still changed `vote_pk` API. Is it too expensive decompress within `vote_pk`?

near-bookrock:
> @near-bookrock it seems that you still changed `vote_pk` API. Is it too expensive decompress within `vote_pk`?

Did I? I kept the public API `vote_pk` the same, but changed the internal methods that work on the state that are also named `vote_pk`. 

https://github.com/Near-One/mpc/pull/328/files/c062bed679b9070ef8fe7fe9cad0aba361667c69#diff-f31f89ad11da217fc15c41b4822ee9f4798bbcbb643341e49157c7747f87669bR137-R144

bowenwang1996:
Sorry I misread. Thanks for the clarification

bowenwang1996 (approved):

Issue #329, opened 2025-04-02
[Contract] update logic follow-up


kevindeforth:
Follow up to #330 to fully resolve #318 

The contract update logic should probably be simplified:
- allow at most one vote and proposal per participant.
- allow an update to affect either the config or the contract code, but not both simultaneously (it makes it much more complicated to test and is simply not necessary).

Issue #318, opened 2025-03-28
Proposed updates (to code and/or config) persists across updates


robin-near:
Imagine this scenario:
 - Threshold is 5
 - Someone proposes an update A
 - It gets 4 votes.
 - Someone proposes an update B
 - B gets 5 votes and the update happens.
 - Someone else makes another vote to A
 - The update A triggers, which may no longer even be valid at all.

We should instead clear out all the updates on every update.

PR #330, opened 2025-04-02
improve test for contract update, fix contract migration during code upgrade

### LINKED ISSUE #316: Verify (via cluster test, e.g.) that pending signatures during V1 -> V2 upgrade will correctly resume with a timeout error despite being cleared from the state due to migration

kevindeforth:
Resolves #316 and partially addresses #318.

This PR: 
- changes the migration logic. Contract updates are now stored under a new StorageKey, to avoid the behavior described in #318, in the sense that we no longer inherit existing proposals.
- introduces more state to the legacy contract before updating.
- cleans up some fields that are no longer required since #296 

A follow up is required (c.f. #329)


libs/chain-signatures/contract/src/update.rs
```diff
@@ -57,6 +57,9 @@ pub struct ProposeUpdateArgs {
 #[near(serializers=[borsh ])]
 #[derive(Debug)]
 struct UpdateEntry {
+    // todo:
```
bowenwang1996, line 60: create an issue for this?

bowenwang1996 (approved):

PR #333, opened 2025-04-04
Add more tests for `ThresholdParameters::validate_incoming_proposal`

### LINKED ISSUE #289: Add more tests for ThresholdParameters::validate_incoming_proposal
* Check that proposing with non-contiguous new IDs fail
  * Also check case with deleting and adding IDs at the same time
* Check deleting participants only

kuksag:
resolves #289 

* Add tests to cover more code paths


bowenwang1996 (approved):
Thanks!

PR #334, opened 2025-04-04
chore: add docs to EdDSA fields in `PublicKeyExtended`.

near-bookrock:
Follow up PR for https://github.com/Near-One/mpc/pull/328 to address the review comments there. The PR already got merged, hence this follow up PR.


libs/chain-signatures/contract/src/crypto_shared/types.rs
```diff
@@ -85,79 +89,74 @@ impl TryFrom<near_sdk::PublicKey> for PublicKeyExtended {
...
+    pub fn near_public_key_ref(&self) -> &near_sdk::PublicKey {
+        match self {
+            PublicKeyExtended::Secp256k1 { near_public_key } => near_public_key,
+            PublicKeyExtended::Ed25519 {
+                near_public_key_compressed,
+                ..
+            } => near_public_key_compressed,
+        }
+    }
+}
```
bowenwang1996, line 102: Why not implement `AsRef<near_sdk::PublicKey>`?

libs/chain-signatures/contract/src/crypto_shared/types.rs
```diff
@@ -85,79 +89,74 @@ impl TryFrom<near_sdk::PublicKey> for PublicKeyExtended {
...
         }
     }
 
     impl BorshDeserialize for PublicKeyExtended {
         fn deserialize_reader<R: std::io::prelude::Read>(reader: &mut R) -> std::io::Result<Self> {
             let from_ser: Vec<u8> = BorshDeserialize::deserialize_reader(reader)?;
-            let public_key_extended: PublicKeyExtendedHelper = serde_json::from_slice(&from_ser)?;
-            Ok(public_key_extended.into())
+            let deserializable_helper_representation: PublicKeyExtendedHelper =
+                serde_json::from_slice(&from_ser)?;
```
bowenwang1996, line 144: I am a bit confused. Why do we use serde_json here inside implementation for Borsh serialization?

libs/chain-signatures/contract/src/crypto_shared/types.rs
near-bookrock: Implemented.

libs/chain-signatures/contract/src/crypto_shared/types.rs
near-bookrock: Removed serde_json and serialize directly with bytes. Also added  unit test for both key types to make sure serialize/deserialize works.

bowenwang1996 (commented):
@near-bookrock looks like serde_json is still used in BorshDeserialize implementation

near-bookrock:
> @near-bookrock looks like serde_json is still used in `BorshDeserialize` implementation

I kept it for `Affinepoint` as it's already in production. But you're right, given the upgrade is non-backwards compatible anyway, we can break the serialization implementation.

libs/chain-signatures/contract/src/crypto_shared/types.rs
```diff
@@ -225,40 +228,17 @@ pub mod k256_types {
...
-            Ok(SerializableAffinePoint { affine_point })
-        }
-    }
-    #[derive(
-        BorshDeserialize, BorshSerialize, Serialize, Deserialize, Debug, Clone, PartialEq, Eq,
-    )]
+    #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
     pub struct SignatureResponse {
-        pub big_r: SerializableAffinePoint,
+        pub big_r: AffinePoint,
```
bowenwang1996, line 233: Doesn't this change the return value of `sign`? It seems to be that it would break the client side code that handles this response.

near-bookrock: Ah, you are right. Reverted the change, and added regression tests such that we don't break this in the future by mistake.

https://github.com/Near-One/mpc/pull/334/commits/55d6f86f5a75e7cea9223e36b6f4c96d0804b175

bowenwang1996 (approved):

PR #339, opened 2025-04-08
test: test public key derivation in contract

near-bookrock:


libs/chain-signatures/contract/Cargo.toml
```diff
@@ -37,6 +37,7 @@ thiserror = "1"
 sha3 = "0.10.8"
 hex = "0.4.3"
 
+
```
bowenwang1996, line 40: This new line seems unnecessary

bowenwang1996 (approved):

libs/chain-signatures/contract/Cargo.toml
near-bookrock:
```suggestion
```

PR #349, opened 2025-04-09
Support contract V2 for devnet tool

### LINKED ISSUE #343: Node should print error if contract state cannot be serialized

robin-near:
Also fix #343 .

## Testing procedure

This assumes familiarity with whatever is already in the README.md.

First, create a mpc cluster (called `and-upg` here ("andrei upgrade")), as well as a loadtest setup (called `robin8` here). Deploy the infra as usual. All of these are described in README.md.

Then, first deploy V1 contract (this is still the default for the devnet tool)
```sh
target/debug/devnet mpc and-upg deploy-contract --init-participants 3
```

then deploy nodes with old docker image (assuming deploy-infra already performed):
```sh
target/debug/devnet mpc and-upg deploy-nomad --docker-image nearone/mpc-node-gcp:mainnet-release
```

then wait for nodes to be up

then compile contract V2:
```sh
# in libs/chain-signatures
cargo build --release --target=wasm32-unknown-unknown
wasm-opt -Oz -o target/wasm32-unknown-unknown/release/mpc_contract.wasm target/wasm32-unknown-unknown/release/mpc_contract.wasm
```

Upgrade to V2:
```
target/debug/devnet mpc and-upg propose-update-contract --path ../libs/chain-signatures/target/wasm32-unknown-unknown/release/mpc_contract.wasm
```

which prints out a command, execute that to vote for the update:
```
target/debug/devnet mpc and-upg vote-update --update-id=0
```

now restart nodes with V2 build (I was using a build with the node changes in this PR, but that doesn't matter):
```
target/debug/devnet mpc and-upg deploy-nomad --docker-image nearone/mpc-node-gcp:robin-debug1-406eff1
```

Add domains (we add an Eddsa and an Ecdsa key):
```
target/debug/devnet mpc and-upg vote-add-domains --signature-schemes Ed25519,Secp256k1
```

Go to the node's /debug/tasks page and wait until key generation is complete.

Finally, test for signatures in each domain:
```
target/debug/devnet loadtest robin8 run --mpc-network and-upg --qps 1 --domain-id 2
```
(change --domain-id between 0, 1, 2)

I also tested resharing:
```
target/debug/devnet mpc and-upg vote-new-parameters --remove 2
# wait until resharing completes
target/debug/devnet mpc and-upg vote-new-parameters --add 2
# wait again
target/debug/devnet mpc and-upg vote-new-parameters --remove 0
# don't wait for it to complete this time
# this command now reproposes to change the participants to [0, 2] instead
target/debug/devnet mpc and-upg vote-new-parameters --remove 1
# wait again for success
# and then test signatures again
```


devnet/src/mpc.rs
```diff
@@ -387,7 +397,22 @@ struct JoinArgs {
     sign_pk: near_sdk::PublicKey,
 }
 
-impl MpcVoteJoinCmd {
+/// Gets a list of voters who would send the vote txn, based on the cmdline flag (empty list means
+/// all participants; otherwise it's the precise list of participant indices).
+fn get_voter_account_ids<'a>(
+    mpc_setup: &'a MpcNetworkSetup,
+    voters: &[usize],
+) -> Vec<&'a AccountId> {
```
kevindeforth, line 405: Maybe just `get_network_account_ids()` and `node_ids: &[usize]` instead of `voters`?

The term "voter' implies that the corresponding AccountId is a registered participant in the contract, but we don't have that guarantee here. Afai understood, `MpcNetworSetup` just contains all nodes that belong to our cluster and they might or might not all be participants.

kevindeforth (approved):
Nice!

Edit: maybe add the PR description to the devnet readme.

devnet/src/mpc.rs
robin-near: Well.... without the specific context then the special case of "node_ids" being empty would be weird :)

robin-near:
For README I'll make a separate PR.

devnet/src/mpc.rs
kevindeforth: I think I missed something. Isn't it that mpc_setup contains all mpc nodes of our cluster, but not all of them are participants? E.g. we could deploy a cluster of 5 nodes and then kick 2 of them out during a resharing.

robin-near: Yes, but the vector that is being passed in is supposed to come from the CLI. The behavior is:
 if you do something like `./devnet vote-new-parameters`, this will vote from all accounts
but if you do something like `./devnet vote-new-parameters --voters 0`, it will only vote from the 0th account.
So there's the discontinuity when the voters vector is empty, which is what the default value will be.

PR #353, opened 2025-04-09
chore: Use cargo workspaces

near-bookrock:
The crates `libs/chain-signatures/contract`, `/node` and `/devnet` are currently standalone projects. This is annoying to work with since `rust-analyzer` will not analyze the contract crate properly, for example for test code, if opened in the root directory. To work around this we currently have to open two IDE instances, one for node and one for contract, which is not a obvious solution for some developers and  adds unnecessary friction.

This PR unifies the projects in one cargo workspace. The workspace allows us to:
-  Run commands like `cargo clippy`, `cargo test`,  `cargo fmt`, etc. once and run it for all crates in this project. Currently these commands have to be run for every crate.
- Better IDE / `rust-analyzer` support.
- Unification of dependencies dependencies, and `clippy.toml`. Note: I have left the contract dependencies untouched in this PR.
- Our test utils for integration tests can easily be moved out of the contract source code, for example https://github.com/Near-One/mpc/blob/563152319f8a24aebf8b27db0300df481f579a55/libs/chain-signatures/contract/src/state/key_event.rs#L288
As a best practice, we should follow up on this PR by creating a new crate in the workspace with test utilities that is a test target only, which the integration tests can depend on.


.github/workflows/ci.yml
```diff
@@ -37,29 +37,9 @@ jobs:
...
         run: |
           cargo fmt -- --check
-
-          cd devnet
-          cargo fmt -- --check
-          cd ..
-
-          cd libs/chain-signatures
-          cargo fmt -- --check
-          cd ../..
```
near-bookrock, line 62: Removed as they are now all covered by the single ```cargo clippy --all-features --all-targets --locked``` above.

near-bookrock:
> @bowenwang1996 I think we earlier had some concerns about not wanting to touch the dependencies of the contract and that using a workspace would make it harder to manage its dependencies. Are the concerns still valid?

near-bookrock:
> @bowenwang1996 I think we earlier had some concerns about not wanting to touch the dependencies of the contract and that using a workspace would make it harder to manage its dependencies. Are the concerns still valid?

I have compared the final WASM binary size of this PR to current main. Both produce an identical binary size of 564KB. 

I'm not sure I follow why it would be harder to manage dependencies for the contract? We don't need to inherit dependencies from the workspace for the contract.

PR #356, opened 2025-04-09
Upgrade nearcore to 2.6.0-rc.1.

robin-near:
Also fix:
* Coordinator busy loops forever if indexer crashes. (We found it!)
* pytest used incorrect serialization of ed25519 secret key. (How did it work before?)

### DIFF
--- a/node/src/coordinator.rs
+++ b/node/src/coordinator.rs
@@ -228,7 +228,10 @@ impl Coordinator {
                             }
                         }
                     }
-                    _ = self.indexer.contract_state_receiver.changed() => {
+                    res = self.indexer.contract_state_receiver.changed() => {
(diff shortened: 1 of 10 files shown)


node/src/coordinator.rs
```diff
@@ -228,7 +228,10 @@ impl Coordinator {
                             }
                         }
                     }
-                    _ = self.indexer.contract_state_receiver.changed() => {
+                    res = self.indexer.contract_state_receiver.changed() => {
+                        if let Err(_) = res {
```
bowenwang1996, line 232:
```suggestion
                        if res.is_err() {
```

bowenwang1996 (approved):
huh I am really curious how the key serialization worked before. cc @kevindeforth

PR #360, opened 2025-04-10
cli tool to extract derived public key from signature response

### LINKED ISSUE #327: Add a test to make sure derivation path is preserved
We need to make sure that signatures for a specific user is not accidentally altered. Roughly the test does the following:
* Run a cluster with contract and mpc code using the latest commit on main
* Randomly generate an account. Send a signature request for both ecdsa and eddsa
* Stop the cluster. Upgrade contract and mpc code to the new version in some branch that is going to be merged to main
* Make sure that the same request gets the same outcome

kevindeforth:
Resolves #327 

Adds a helper util to derive the public key from a signature response (useful for checking if the derived public key is preserved across contract updates).

Tested with [a cluster on testnet](https://testnet.nearblocks.io/address/mpc-contract-kevin-deriv-655e8c86b972.2a71fea00087.testnet).


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [workspace]
 members = ["node"]
 resolver = "2"
-exclude = ["libs", "pytest"]
+exclude = ["libs", "pytest", "helpers"]
```
near-bookrock, line 4: Why not include it as part of the workspace?

utils/recover_public_key/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+        eprintln!(
+            "Usage: {} <path-to-json>\n\n\
+        The input JSON file should look like:\n\
+        {{\n  \"old_res\": {{ ... }},\n  \"new_res\": {{ \"Secp256k1\": {{ ... }}}},\n  \"msg_hash\": \"...\"\n}}",
+            args.first()
+                .map(|s| s.as_str())
+                .unwrap_or("sig_recovery_check")
+        );
+        std::process::exit(1);
+    }
```
pbeza, line 73: Any specific reason we're not using [`clap`](https://docs.rs/clap/latest/clap/) to parse CLI args? 🤔 It’s kinda the standard for this stuff.

netrome:
Closing this due to inactivity. Looks useful though so feel free to brush it up and reopen.

PR #363, opened 2025-04-11
Smart contract "proposal updates" logic improvements

### LINKED ISSUE #318: Proposed updates (to code and/or config) persists across updates

### LINKED ISSUE #329: [Contract] update logic follow-up

kuksag:
Resolves #329
Resolves #318

* Accept update proposals which are only modifies code, or config, but not both.
* A participant's vote can be counted only once for some proposal. If participant votes again, transfer vote from one proposal to the other
* Remove all proposals after update
* Add tests for introduced code paths

### DIFF
--- a/libs/chain-signatures/contract/tests/updates.rs
+++ b/libs/chain-signatures/contract/tests/updates.rs
@@ -4,7 +4,6 @@ use mpc_contract::config::Config;
 use mpc_contract::state::ProtocolContractState;
 use mpc_contract::update::{ProposeUpdateArgs, UpdateId};
 use near_workspaces::types::NearToken;
-use crate::common::accounts;
 
 pub fn dummy_contract() -> ProposeUpdateArgs {
     ProposeUpdateArgs {
@@ -256,7 +255,6 @@ async fn test_propose_update_contract_many() {
     dbg!(state);
 }
 
-
 #[tokio::test]
 async fn test_propose_incorrect_updates() {
     let (_, contract, accounts, _) = init_env_secp256k1(1).await;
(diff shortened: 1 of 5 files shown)


libs/chain-signatures/contract/tests/updates.rs
```diff
@@ -208,13 +202,13 @@ async fn test_invalid_contract_deploy() {
     let state: ProtocolContractState = execution.json().unwrap();
     dbg!(state);
 }
-// todo: fix this test
```
kuksag, line 211: I believe that was addressed

libs/chain-signatures/contract/tests/updates.rs
```diff
@@ -208,13 +202,13 @@ async fn test_invalid_contract_deploy() {
...
+
 #[tokio::test]
 async fn test_propose_update_contract_many() {
     let (_, contract, accounts, _) = init_env_secp256k1(1).await;
     dbg!(contract.id());
 
-    const PROPOSAL_COUNT: usize = 10;
+    const PROPOSAL_COUNT: usize = 5;
```
kuksag, line 211: I randomly got "Size of the recorded trie storage proof has exceeded ... yada yada", so I decided to drop the number by a little bit

robin-near (approved):
Excellent! Thank you for the thorough testing as well!

libs/chain-signatures/contract/tests/updates.rs
robin-near: Ah I see that's probably an underlying blockchain/test framework limitation.

PR #364, opened 2025-04-11
Remove assets from db with `maybe_discard_owned`

### LINKED ISSUE #320: `maybe_discard_owned` doesn't discard assets on disk?

kuksag:
resolves #320 

* Control what ids were removed from the cold queue and remove them from db too
* Add test scenario that covers the bug


robin-near (approved):
Thank you!

Issue #255, opened 2025-03-14
Export metrics for signature response latency


robin-near:
* Export a histogram prometheus metric for the number of blocks between when a signature request is seen and when the corresponding response is seen.
* Fix the bug that the height used for the response (in the rendering of /debug/signatures) is incorrect (off by 2 blocks)

robin-near:
#374

PR #365, opened 2025-04-14
chore: Add metrics for latency of signature request responses.

near-bookrock:
Part of #255 

This PR adds a new histogram metric `mpc_signature_request_response_latency_blocks` to track the number of blocks between a signature request is observed and its corresponding response is observed on the smart contract.


near-bookrock:
closed in favor of https://github.com/Near-One/mpc/pull/365

PR #366, opened 2025-04-14
chore: Add metrics for latency of signature request responses

near-bookrock:
Part of https://github.com/Near-One/mpc/issues/255

This PR adds a new histogram metric mpc_signature_request_response_latency_blocks to track the number of blocks between a signature request is observed and its corresponding response is observed in a block.


bowenwang1996 (approved):
Can we also add latency in time?

near-bookrock:
> Can we also add latency in time?

I'll add it in a follow up PR.

PR #368, opened 2025-04-15
Add web endpoint /debug/contract

kevindeforth:
This PR exposes a new web endpoint `debug/contract`, that displays the current contract view of an mpc node..

It uses the logic introduced in https://github.com/Near-One/mpc/pull/351 to stringify the current ProtocolState instance and moves this logic to the contract folder (behind a `dev-utils` feature flag).

follow-ups:
- [ ] additionally, display `ThresholdParameterVotes`, `AddDomainVotes` and proposed updates.


robin-near (approved):
I love it!

PR #371, opened 2025-04-16
chore: remove spawn_blocking call wrapping the indexer thread

near-bookrock:
We have a `spawn_blocking` that wraps the join_handle in `spawn_real_indexer`. This is presumably done to await on the `std::thread::JoinHandle` returned from the spawned indexer.

The `spawn_blocking` call lead to the issue we saw the other day with the binary not exiting, because the runtime is unable to drop this task since it never terminates. We can remove this `spawn_blocking call`, and use a [oneshot channel](https://docs.rs/tokio/latest/tokio/sync/oneshot/fn.channel.html) instead that the indexer thread can respond on if it exits.

In general  oneshot channels are a good practice to use when communicating from a synchronous threads to an async task.

### DIFF
--- a/node/src/cli.rs
+++ b/node/src/cli.rs
@@ -141,14 +141,14 @@ impl StartCmd {
         let (web_contract_sender, web_contract_receiver) =
             tokio::sync::watch::channel(ProtocolContractState::NotInitialized);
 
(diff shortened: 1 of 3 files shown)


robin-near (approved):
Ah I see, it's because the runtime can't shutdown when the indexer is still running. Gotcha, thanks!

PR #372, opened 2025-04-16
Requires full threshold votes for keygen and reshare

kevindeforth:
The contract requires:
- all current participants to vote for the new domain in order to enter keygen;
- all prospective participants of the prospective epoch to vote for the proposed parameters in order to enter reshare.

Note that the latter implicitly requires `threshold` number of votes from current participants, because the prospective participant set must have at least as many current participants.

### DIFF
--- a/libs/chain-signatures/contract/src/state/resharing.rs
+++ b/libs/chain-signatures/contract/src/state/resharing.rs
@@ -46,9 +46,9 @@ impl ResharingContractState {
         self.resharing_key.epoch_id()
     }
 
-    /// Casts a vote for a re-proposal. Requires the signer to be an original participant from the
-    /// previous running state. If this exceeds the threshold (from the previous running state),
-    /// returns a new ResharingContractState that we should transition into.
+    /// Casts a vote for a re-proposal. Requires the signer to be a participant of the
+    /// prospective epoch.
+    /// Returns a new ResharingContractState if all proposed participants voted for it.
     pub fn vote_new_parameters(
         &mut self,
         prospective_epoch_id: EpochId,
@@ -171,11 +171,8 @@ mod tests {
         let mut running = gen_running_state(num_domains);
         let proposal = gen_valid_params_proposal(&running.parameters);
         let mut resharing_state = None;
-        let voting_participants = running.parameters.participants().participants()
-            [0..running.parameters.threshold().value() as usize]
-            .to_vec();
-        for (account, _, _) in voting_participants {
-            env.set_signer(&account);
+        for (account, _, _) in proposal.participants().participants() {
+            env.set_signer(account);
             assert!(resharing_state.is_none());
             resharing_state = running
                 .vote_new_parameters(running.keyset.epoch_id.next(), &proposal)
@@ -359,7 +356,6 @@ mod tests {
             .parameters
             .participants()
             .clone();
-        let old_threshold = state.previous_running_state.parameters.threshold().value() as usize;
         {
             let new_participants = state
                 .resharing_key
@@ -383,7 +379,7 @@ mod tests {
         // should be rejected, since all re-proposals must be valid against the original.
         let mut new_participants_1 = old_participants.clone();
         let new_threshold = Threshold::new(old_participants.len() as u64);
-        new_participants_1.add_random_participants_till_n(old_participants.len() * 3 / 2);
+        new_participants_1.add_random_participants_till_n((old_participants.len() * 3).div_ceil(2));
         let new_participants_2 = new_participants_1
             .subset(new_participants_1.len() - old_participants.len()..new_participants_1.len());
         let new_params_1 =
@@ -416,7 +412,7 @@ mod tests {
 
         // Repropose with new_params_1.
         let mut new_state = None;
-        for (account, _, _) in &old_participants.participants()[0..old_threshold] {
+        for (account, _, _) in new_params_1.participants().participants() {
             env.set_signer(account);
             assert!(new_state.is_none());
             new_state = state
--- a/libs/chain-signatures/contract/src/state/running.rs
+++ b/libs/chain-signatures/contract/src/state/running.rs
@@ -114,21 +114,21 @@ impl RunningContractState {
     }
 
     /// Casts a vote for `proposal`, removing any previous votes by `env::signer_account_id()`.
-    /// Fails if the proposal is invalid or the signer is not a participant.
-    /// Returns true if the proposal reached `threshold` number of votes.
+    /// Fails if the proposal is invalid or the signer is not a proposed participant.
+    /// Returns true if all participants of the proposed parameters voted for it.
     pub(super) fn process_new_parameters_proposal(
         &mut self,
         proposal: &ThresholdParameters,
     ) -> Result<bool, Error> {
-        // ensure the signer is a participant
-        let participant = AuthenticatedParticipantId::new(self.parameters.participants())?;
+        // ensure the signer is a proposed participant
+        let participant = AuthenticatedParticipantId::new(proposal.participants())?;
 
         // ensure the proposal is valid against the current parameters
         self.parameters.validate_incoming_proposal(proposal)?;
 
         // finally, vote. Propagate any errors
         let n_votes = self.parameters_votes.vote(proposal, &participant);
-        Ok(self.parameters.threshold().value() <= n_votes)
+        Ok(proposal.participants().len() as u64 == n_votes)
     }
 
     /// Casts a vote for the signer participant to add new domains, replacing any previous vote.
@@ -144,7 +144,7 @@ impl RunningContractState {
         }
         let participant = AuthenticatedParticipantId::new(self.parameters.participants())?;
         let n_votes = self.add_domains_votes.vote(domains.clone(), &participant);
-        if self.parameters.threshold().value() <= n_votes {
+        if self.parameters.participants().len() as u64 == n_votes {
             let new_domains = self.domains.add_domains(domains.clone())?;
             Ok(Some(InitializingContractState {
                 generated_keys: self.keyset.domains.clone(),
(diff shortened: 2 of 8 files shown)


pytest/tests/test_key_event.py
```diff
@@ -37,39 +37,36 @@ def test_single_domain():
...
+    )
 
     cluster.update_participant_status()
     cluster.send_and_await_signature_requests(1)
-    cluster.do_resharing(new_participants=mpc_nodes[0:4],
-                         new_threshold=3,
-                         prospective_epoch_id=3,
-                         wait_for_running=False)
-
-    mpc_nodes[0].run()
```
kevindeforth, line 61: ci is flaky if we keep the killing and restarting of the node, because the node is not back up in time for voting on the reshare.

libs/chain-signatures/contract/src/state/resharing.rs
```diff
@@ -46,9 +46,9 @@ impl ResharingContractState {
         self.resharing_key.epoch_id()
     }
 
-    /// Casts a vote for a re-proposal. Requires the signer to be an original participant from the
-    /// previous running state. If this exceeds the threshold (from the previous running state),
-    /// returns a new ResharingContractState that we should transition into.
+    /// Casts a vote for a re-proposal. Requires the signer to be a participant of the
+    /// prospective epoch.
+    /// Returns a new ResharingContractState if all proposed participants voted for it.
```
near-bookrock, line 51: Just a tip. You can create links in your rust docs, which `rust-analyzer` will also discover and let you go to it's definition from the docs in the editor.

```suggestion
    /// Returns a new [`ResharingContractState`] if all proposed participants voted for it.
```

near-bookrock, line 51:
```suggestion
    /// Returns a new ResharingContractState if all proposed participants voted for a re-proposal.
```

near-bookrock: Is "it" meaning a re-proposal here?

libs/chain-signatures/contract/src/state/running.rs
```diff
@@ -271,12 +271,19 @@ pub mod running_tests {
...
+        for i in 0..participants.participants().len() + 1 {
             loop {
                 let proposal = gen_valid_params_proposal(&state.parameters);
                 if proposals.contains(&proposal) {
                     continue;
                 }
+                if i < participants.participants().len()
+                    && !proposal
+                        .participants()
+                        .is_participant(&participants.participants()[i].0)
```
near-bookrock, line 283: Can't this direct indexing can cause a panic when `i` = `participants.participants().len()`?

libs/chain-signatures/contract/src/state/running.rs
```diff
@@ -271,12 +271,19 @@ pub mod running_tests {
...
+                {
+                    continue;
+                }
```
near-bookrock, line 286: Try to avoid direct indexing, as it can panic if you go out of bounds. It would be more idiomatic to use a `get` here.
```rust
if participants.participants()
    .get(i)
    .is_some_and(|participant| proposal.participants().is_participant(participant.0))
{
    continue;
}
```

libs/chain-signatures/contract/src/state/running.rs
```diff
@@ -294,17 +301,12 @@ pub mod running_tests {
...
-            .enumerate()
-            .take(state.parameters.threshold().value() as usize)
-        {
+        for (i, (account_id, _, _)) in proposal.participants().participants().iter().enumerate() {
             env.set_signer(account_id);
             let res = state
                 .vote_new_parameters(state.keyset.epoch_id.next(), &proposal)
                 .unwrap();
-            if i + 1 < state.parameters.threshold().value() as usize || num_domains == 0 {
+            if i + 1 < proposal.participants().len() || num_domains == 0 {
```
near-bookrock, line 309: Ditto on binding bools to descriptive variable names.

libs/chain-signatures/contract/src/state/running.rs
```diff
@@ -144,7 +144,7 @@ impl RunningContractState {
         }
         let participant = AuthenticatedParticipantId::new(self.parameters.participants())?;
         let n_votes = self.add_domains_votes.vote(domains.clone(), &participant);
-        if self.parameters.threshold().value() <= n_votes {
+        if self.parameters.participants().len() as u64 == n_votes {
```
near-bookrock, line 147: Could you bind verbose or complex expressions to a descriptive variable name so it's easier to follow as a reader?

For example:
```suggestion
        let threshold_number_of_votes_casted = self.parameters.participants().len() as u64 == n_votes;
        if threshold_number_of_votes_casted {
```

near-bookrock (approved):

libs/chain-signatures/contract/src/state/resharing.rs
robin-near: +1; we should be extra clear here because there are three sets of participants involved here: the original, the currently proposed, and the reproposal. It should be clear that it is the prospective participants of the *reproposal* that are required to vote.

I think we should also include a sentence saying that this implies that a threshold number from the original will have endorsed the reproposal, since we do verify that the reproposal is valid. (Otherwise, the documentation raises eyebrows that we're letting anyone else take over the network.)

libs/chain-signatures/contract/src/state/running.rs
robin-near: In this case I would argue that extracting it out is harder to read. I think the code as it is is very clear already. It spells "the number of participants equals the number of votes".

libs/chain-signatures/contract/src/state/running.rs
robin-near: It's a test though and I think in this particular case, the logic is clearer with the explicit i < len comparison because the pattern that kevin's trying to do here is to loop an extra time to have a different behavior. Whether the [i] would go out of bounds is irrelevant since we only do that for the non-extra-iteration case.

libs/chain-signatures/contract/src/state/running.rs
robin-near: I would suggest just commenting.
```
// If we don't have all votes, or if we have no domains, then we shouldn't be transitioning state.
```
(What would the variable be named otherwise?)

libs/chain-signatures/contract/src/state/running.rs
```diff
@@ -114,21 +114,21 @@ impl RunningContractState {
...
+    /// Fails if the proposal is invalid or the signer is not a proposed participant.
+    /// Returns true if all participants of the proposed parameters voted for it.
     pub(super) fn process_new_parameters_proposal(
         &mut self,
         proposal: &ThresholdParameters,
     ) -> Result<bool, Error> {
-        // ensure the signer is a participant
-        let participant = AuthenticatedParticipantId::new(self.parameters.participants())?;
+        // ensure the signer is a proposed participant
+        let participant = AuthenticatedParticipantId::new(proposal.participants())?;
```
robin-near, line 124: I have a concern with this that I don't yet have a good solution for.

Previously, someone can only propose this if they are a current participant. But now, any random person can vote. Although that doesn't threaten correctness, it does open some doors for DoS.

Specifically, someone unrelated can create a bunch of accounts and start proposing (with valid proposals that include themselves). They keep doing this, adding more proposals to the `parameter_votes`, flooding it with useless entries until they cannot even do this because the map is so large. This shouldn't be very hard, because each proposal can be arbitrarily big (just invent some random other participants) so it only takes a handful of transactions.

Now, the contract is in a bad state because not only does every signature request need to spend a ton of gas to decode the state (likely exceeding 10Tgas, making it unable to process signatures), but also we cannot recover from this because to recover, the only way is to trigger a `vote_add_domains` or `vote_new_parameters` but either of these requires adding even *more* data to the state, and we already know that doing so would exceed 300Tgas due to the attack.

libs/chain-signatures/contract/src/state/running.rs
near-bookrock: I see now that this is test code, but in general if you have descriptive variable names you can avoid comments that can eventually become stale, and still keep it easy to follow.

```suggestion
            let missing_threshold_number_of_votes =  i + 1 < proposal.participants().len();
            let has_no_domains = num_domains == 0;
            if missing_threshold_number_of_votes || has_no_domains {
```

libs/chain-signatures/contract/src/state/running.rs
kevindeforth: > Try to avoid direct indexing, as it can panic if you go out of bounds. 

Wait, this can go out of bounds?

```Rust
fn fun(x: Vec<u64>, i: usize) {
    if i < x.len() && do_sth(x[i]) {
        // ...
    }
}

```

near-bookrock: It's test code, so you can ignore my comment if you want.

No, in this specific case it won't since you have the check. But it's better to have a guarantee at compile time, which you get with `get(i)` and then work on the option it returns.

libs/chain-signatures/contract/src/state/running.rs
near-bookrock: My confusion comes from the naming of `n_votes`. `n_votes` is not descriptive enough to tell me what it means, so I don't understand what the if check does.

Now that you tell me it means number of casted votes, it's also obvious to me what the if check does.

Naming our variables something more descriptive from the get go, for example `number_of_casted_votes`, then it would be easy to understand without needing a comment/explanation or a variable.

```rust
 if self.parameters.participants().len() as u64 == casted_votes {
```

kevindeforth: > My confusion comes from the naming of n_votes. n_votes is not descriptive enough to tell me what it means, so I don't understand what the if check does.

The `n` is often used as shorthand for `num` or `number`.

I think we’re landing in the realm of personal preference here. I usually lean toward shorter variable names like `n_votes` to keep the code compact and easy to scan. While `number_of_casted_votes` is more explicit, it also gives the impression of precision that isn’t entirely accurate—we're not counting all casted votes, but rather the votes for this specific proposal. So to be fully accurate, it would have to be something like `number_of_casted_votes_for_proposal`, which feels overly verbose to me.

libs/chain-signatures/contract/src/state/running.rs
kevindeforth: Good call!

I changed the logic s.t. we now require `threshold` old participants to vote for a proposal before new candidates can cast their votes. This should prevent this endpoint from becoming an attack vector for DOS.

This solution required a change to the contract state and thus changes to the migration logic. Python tests were added for that purpose.

robin-near (approved):

PR #374, opened 2025-04-17
Correctly export signature latency metrics.

robin-near:
Followup to #366 and #373, the block height as well as indexed time used was incorrect - we should use the block that contains the signature response, which was being held until it becomes final, instead of the block that triggers the finalization of such a block.

Fixing this required some slight modification to the data structures.

Also, use the clock instead of `Instant::now()` for getting instants. Also add a test that the metrics are correctly computed.

### DIFF
--- a/node/src/signing/debug.rs
+++ b/node/src/signing/debug.rs
@@ -69,25 +69,22 @@ impl Debug for CompletedSignatureRequest {
             "  [completed] blk {:>10} -> {:<24} id: {} rx: {:<44} tries: {:<2}",
             self.indexed_block_height,
             self.completion_delay
-                .map(|(delay_blocks, delay_time)| format!(
-                    "{:>10} (+{}, {})",
-                    self.indexed_block_height + delay_blocks,
-                    delay_blocks,
-                    humantime::format_duration(
-                        round_duration_to_milliseconds(delay_time).unsigned_abs()
-                    ),
-                ))
+                .map(|(delay_blocks, delay_time)| {
+                    let duration_rounded_to_ms =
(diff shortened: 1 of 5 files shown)


node/src/signing/queue.rs
```diff
@@ -150,8 +151,8 @@ pub(super) struct QueuedSignatureRequest {
     /// completed.
     pub active_attempt: Weak<SignatureGenerationAttempt>,
 
-    /// The [`Instant`] the signature was indexed.
-    pub indexing_instant: Instant,
+    /// The time that the signature was indexed.
+    pub time_indexed: near_time::Instant,
```
near-bookrock, line 155: Curious. Why use `near_time::Instant` when it's just a re-export of `std::time::Instant`? https://docs.rs/near-time/latest/near_time/type.Instant.html

node/src/signing/debug.rs
```diff
@@ -79,6 +85,11 @@ impl Debug for CompletedSignatureRequest {
     }
 }
 
+fn round_duration_to_milliseconds(duration: near_time::Duration) -> near_time::Duration {
+    let milliseconds = duration.whole_milliseconds();
+    near_time::Duration::milliseconds(milliseconds as i64)
+}
```
near-bookrock, line 91: Doesn't this seem like an over abstraction when it's only used once?

near-bookrock (approved):
Thanks for the fix!

kevindeforth (approved):

node/src/signing/debug.rs
robin-near: I guess so. I'll just inline it.

PR #376, opened 2025-04-17
Integrate TEE into MPC network

pbeza:


libs/chain-signatures/contract/src/lib.rs
```diff
@@ -73,17 +73,25 @@ pub struct MpcContract {
...
     fn from(value: v0_state::MpcContractV0) -> Self {
-        MpcContract {
+        Self {
             protocol_state: value.protocol_state.into(),
             pending_requests: value.pending_requests,
             proposed_updates: ProposedUpdates::default(),
             config: value.config,
+            allowed_code_hashes: vec![],
+            historical_code_hashes: vec![],
+            upgrade_deadline: Some(TEE_UPGRADE_PERIOD),
```
bowenwang1996, line 90: I think this is not correct. `upgrade_deadline` should be either set to `None`, meaning that there is no pending upgrade, or to `env::block_index() + TEE_UPGRADE_PERIOD`, meaning that there is a pending upgrade with a specific deadline

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -193,6 +204,37 @@ impl MpcContract {
...
+            self.allowed_code_hashes = self
+                .allowed_code_hashes
+                .last()
+                .map_or_else(Vec::new, |last| vec![last.clone()]);
+        }
+        self.allowed_code_hashes.clone()
+    }
+
+    pub fn latest_code_hash(&self) -> Option<CodeHash> {
+        self.allowed_code_hashes.last().cloned()
```
bowenwang1996, line 236: I think we should return `CodeHash`. The invariant is that there should always be a latest code hash, i.e, it is not possible for it to be `None`.

libs/chain-signatures/contract/src/primitives/code_hash.rs
```diff
@@ -0,0 +1,47 @@
...
+#[near(serializers=[borsh, json])]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub struct CodeHash([u8; 32]);
+
+/// Tracks votes to add whitelisted TEE code hashes. Each participant can at any given time vote for
+/// a code hash to add.
+#[near(serializers=[borsh, json])]
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct CodeHashesVotes {
+    pub proposal_by_account: BTreeMap<AuthenticatedParticipantId, CodeHash>,
```
bowenwang1996, line 16: what's the reason for doing it this way vs. `HashMap<CodeHash, HashSet<AuthenticatedParticipantId>>`? I guess there is not much real difference since we are talking about very small number of participants here but just curious

libs/chain-signatures/contract/src/state/mod.rs
```diff
@@ -208,10 +236,7 @@ impl ProtocolContractState {
         }
     }
     pub fn is_running(&self) -> bool {
-        if let ProtocolContractState::Running(_) = self {
-            return true;
-        }
-        false
+        matches!(self, ProtocolContractState::Running(_))
```
bowenwang1996, line 239: nice!

libs/chain-signatures/contract/src/state/running.rs
```diff
@@ -184,22 +194,27 @@ impl RunningContractState {
             Ok(None)
         }
     }
+
+    pub fn vote_code_hash(&mut self, code_hash: CodeHash) -> Result<Option<Self>, Error> {
```
bowenwang1996, line 198: I am a bit confused about the return type of this function. It seems that it either returns an error or returns `Ok(None)`. In this case, why do we make the return type `Result<Option<Self>, Error>` instead of something like `Result<(), Error>`? It also makes the code that calls it look confusing

libs/chain-signatures/contract/src/lib.rs
pbeza: Fixed in https://github.com/Near-One/mpc/pull/376/commits/90db97bc661b0ce7ccc8f08ad9e8687f44b0178b.

libs/chain-signatures/contract/src/primitives/code_hash.rs
pbeza: Funny enough, I asked myself that too!

After digging into the code, I saw there are already a few places where voting is done this way. Probably means we could abstract it out and reuse it across different parts of the code.

https://github.com/Near-One/mpc/blob/a35bc500a2e29592c9aef818b5a31143e5f58715/libs/chain-signatures/contract/src/primitives/domain.rs#L167-L173

https://github.com/Near-One/mpc/blob/a35bc500a2e29592c9aef818b5a31143e5f58715/libs/chain-signatures/contract/src/v0_state/mod.rs#L25-L30

Probably @kevindeforth or @robin-near can help with a bit more context on this.

libs/chain-signatures/contract/src/lib.rs
bowenwang1996: yes it is safe

bowenwang1996 (approved):
Looks good to me. @kevindeforth could you review this change as well?

pbeza:
@kevindeforth +1 on the [code review request](https://github.com/Near-One/mpc/pull/376#pullrequestreview-2799928433). Any idea why the `test_update_v2_running` test is failing? (see the [logs](https://github.com/Near-One/mpc/actions/runs/14707263338/job/41270490353?pr=376)).

libs/chain-signatures/contract/src/state/running.rs
```diff
@@ -36,6 +41,8 @@ pub struct RunningContractState {
     pub parameters_votes: ThresholdParametersVotes,
     /// Votes for proposals to add new domains.
     pub add_domains_votes: AddDomainsVotes,
+    // Votes for proposals to add new TEE code hashes.
+    pub proposed_code_hashes: CodeHashesVotes,
```
kevindeforth, line 45: What was the reason for adding this field to the `RunningContractState` and the `ProtocolContractState` in general?

I understand that there is a grace period of multiple days, during which participants can update their node. So it is fine to keep this completely separate from the ProtocolState and even allow voting on the hash when the network is in Initializing or Resharing state.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -193,6 +204,40 @@ impl MpcContract {
         }
         Ok(())
     }
+
+    pub fn vote_code_hash(&mut self, code_hash: CodeHash) -> Result<(), Error> {
+        let code_hash_votes = self
+            .protocol_state
+            .proposed_code_hashes_count_votes(code_hash.clone())?;
+
+        if code_hash_votes + 1 >= self.threshold()?.value() {
```
kevindeforth, line 213: I think there is  a problem here that leads to double-counting votes:

Assume {Alice, Bob}  are our participants and threshold is 2.
Bob votes for code hash `h`. His vote gets counted.
Bob then submits another vote for `h`.
`code_hash_votes` will return 1 and this condition will evaluate to true, but Alice never voted for the hash.

It is safer to just call `vote_code_hash` and compare its return value to the threshold, since that function ensures that each participant has at most one vote.

libs/chain-signatures/contract/src/primitives/participants.rs
```diff
@@ -16,6 +16,7 @@ pub struct ParticipantInfo {
     pub url: String,
     /// The public key used for verifying messages.
     pub sign_pk: PublicKey,
+    // pub ra_quote: RAQuote,
```
kevindeforth, line 19: can this be removed?

libs/chain-signatures/contract/src/state/mod.rs
```diff
@@ -149,6 +165,15 @@ impl ProtocolContractState {
         .map(|x| x.map(ProtocolContractState::Initializing))
     }
 
+    pub fn vote_code_hash(&mut self, code_hash: CodeHash) -> Result<(), Error> {
+        // TODO verify TEE quote here
```
kevindeforth, line 169: Maybe open an issue for this.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -113,15 +121,18 @@ impl MpcContract {
...
                 DomainRegistry::default(),
                 Keyset::new(EpochId::new(0), Vec::new()),
                 parameters,
             )),
             pending_requests: LookupMap::new(StorageKey::PendingRequestsV2),
             proposed_updates: ProposedUpdates::default(),
+            config: Config::from(init_config),
+            allowed_code_hashes: vec![],
+            historical_code_hashes: vec![],
+            upgrade_deadline: None,
```
kevindeforth, line 135: If we have multiple allowed code hashes, which one does the upgrade deadline refer to?

kevindeforth (changes requested):
The failing test due to a breaking change in the API.

The mpc nodes de-serialize the `ProtocolContractState` from [this endpoint](https://github.com/Near-One/mpc/blob/a35bc500a2e29592c9aef818b5a31143e5f58715/libs/chain-signatures/contract/src/lib.rs#L805).

**Fix**:
- Lets keep the current `ProtocolContractState` and put the TEE stuff in a separate struct. It can be a field in the `MpcContract` struct (c.f. comments).
- modify the [dependency](https://github.com/Near-One/mpc/blob/a35bc500a2e29592c9aef818b5a31143e5f58715/node/Cargo.toml#L64) in the mpc nodes. It is a dangerous that we depend on this contract folder, as we might miss breaking API changes like these. The tests only catched it, because the nodes were trying to de-serialize the contract state of version 2.0.0. struct from chain. It was discovery by accident.

libs/chain-signatures/contract/src/state/mod.rs
pbeza: Good point! https://github.com/Near-One/mpc/issues/378

libs/chain-signatures/contract/src/primitives/participants.rs
pbeza: Yeah, it can be removed for now. We will need it for https://github.com/Near-One/mpc/pull/376 (it's related to [this](https://github.com/Near-One/mpc/pull/376#discussion_r2065690816) comment).

libs/chain-signatures/contract/src/state/running.rs
pbeza: @bowenwang1996 [said](https://docs.nearone.org/doc/integrate-tee-into-mpc-network-OWBOeCofNK?commentId=229f37dd-d80e-46a8-a9d6-13ef20820c13&resolved=) it makes more sense to only allow voting on the hash when we're in the `Running` state, so I just went with that. Not sure why we'd want voting in `Initializing` or `Resharing`. @bowenwang1996, what do you think?

libs/chain-signatures/contract/readme.md
```diff
@@ -37,6 +37,7 @@ stateDiagram-v2
     NotInitialized --> Running : init
     Running --> Initializing : vote_add_domains
     Running --> Resharing : vote_new_parameters
+    Running --> Running : vote_code_hash
```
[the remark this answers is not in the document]
pbeza: I added it here because we have another identity state transition (I’m guessing it’s special and we want to keep it):
https://github.com/Near-One/mpc/blob/a35bc500a2e29592c9aef818b5a31143e5f58715/libs/chain-signatures/contract/readme.md?plain=1#L43

Anyway, fixed here: https://github.com/Near-One/mpc/pull/376/commits/55591017c9f061462eb4eb4e2b795c0e225e859b.✅

libs/chain-signatures/contract/src/lib.rs
pbeza: Good catch! You’re absolutely right. Hopefully fixed it now: https://github.com/Near-One/mpc/pull/376/commits/301b857d5af0c72d61a174cf26183178a918648b ✅

libs/chain-signatures/contract/src/lib.rs
pbeza: I think any of them. AFAIR, we never really talked about it before, but my guess is we let a participant run `vote_new_parameters` only if:
1. They're running a Docker image with a code hash that's in `allowed_code_hashes`.
2. Their `sign_pk` is correct — meaning it's included in the RA quote (TBD in https://github.com/Near-One/mpc/issues/378).

@bowenwang1996 do you agree?

libs/chain-signatures/contract/src/state/running.rs
bowenwang1996: The reason is that it doesn't make sense to vote when we are in initializing or resharing state, since voting requires a stable quorum of participants, which is not the case in initializing or resharing state. This is also to avoid unnecessarily complicating the contract logic.

libs/chain-signatures/contract/src/lib.rs
bowenwang1996: No, `upgrade_deadline` should always be for the most recent allowed hash. The reason is that when a new allowed hash is voted in, it means that an upgrade should happen and at this point there are multiple allowed code hashes (the old version is still allowed), but the after the deadline passes there should only be one allowed hash, which is the new version. Between the successful voting of a new hash and its deadline to upgrade, new code hashes can be voted in, which becomes the latest version and extends the deadline.

kevindeforth: Thanks for the explanations!
I don't see any logic that would remove entries from the `allowed_code_hashes` vector. So with the current implementation, we will allow all previous code hashes whenever we have a pending update.

Idk if this is good or bad or neutral, just wanted to point it out.

libs/chain-signatures/contract/src/state/running.rs
kevindeforth: I see. The issue with tracking the votes here is that it is [a breaking API change](https://github.com/Near-One/mpc/pull/376#pullrequestreview-2802234130).

We would need to change the node s.t. they can handle both contract versions simultaneously. We would need to update the nodes before updating the contract.

Alternatively, we could store the votes in `TeeState` and just require that the contract is in running state whenever a vote is submitted.

libs/chain-signatures/contract/src/lib.rs
bowenwang1996: The call to `allowed_code_hashes` does that. Indeed we should figure out how this function gets called.

kevindeforth: Ah, I missed that. There is still a potential issue though: if a new hash is accepted while the deadline of the previously added hash has not yet expired, the deadline of all hashes in that vector are extended:

t0: h0 is valid
t1: h1 gets added, deadline is set to t1 + d;
t2: h2 gets added, deadline is set to t2 + d;

If t2 < t1 + d, then h0 will be valid until t2+d, which is after t1 + d.

We could avoid this by storing time stamp info with the hash:
```rust
pub struct AllowedCodeHash {
    code_hash: CodeHash,
    added: BlockHeight,
}

pub struct AllowedCodeHashes {
    allowed_code_hashes: Vec<AllowedCodeHash> // ordered by `start`
}

impl AllowedCodeHashes {
    fn clean() -> usize; // removes all expired hashes
    fn insert(code_hash: CodeHash) --> bool; // calls clean(), inserts new hash
    fn get() -> Vec<AllowedCodeHash>; // calls clean(), returns vector
}
```

bowenwang1996: > if a new hash is accepted while the deadline of the previously added hash has not yet expired, the deadline of all hashes in that vector are extended

That is the desired behavior. What problems do you see?

kevindeforth: > That is the desired behavior. What problems do you see?

Consider the previous example:

> t0: h0 is valid
> t1: h1 gets added, deadline is set to t1 + d;
> t2: h2 gets added, deadline is set to t2 + d;
>
> If t2 < t1 + d, then h0 will be valid until t2+d, which is after t1 + d.

Assume the deadline d is 7 days. Then after 8 days, we can have one of the following:

- Scenario 1: h0 is no longer valid. This happens if the time between voting for h1 and h2 was longer than 7 days. In this case, h0 will have been removed from the allowed hashes during the voting of h2.
- Scenario 2: h0 is still a valid hash. This happens if the time between voting for h1 and h2 was shorter than 7 days. In this case, h0 will not have been removed from the allowed hashes.

This doesn't feel right. I would expect the deadline to be hard, i.e. for h0 to expire after 7 days, no matter what.
 But it all depends on the intended use of these hashes.

libs/chain-signatures/contract/src/state/running.rs
```diff
@@ -184,22 +194,27 @@ impl RunningContractState {
             Ok(None)
         }
     }
+
+    pub fn vote_code_hash(&mut self, code_hash: CodeHash) -> Result<u64, Error> {
```
kevindeforth, line 198: can be removed

libs/chain-signatures/contract/src/state/mod.rs
```diff
@@ -72,6 +71,15 @@ impl ProtocolContractState {
             }
         }
     }
+    pub fn clear_code_hashes_votes(&mut self) -> Result<(), Error> {
```
kevindeforth, line 74: can be removed

libs/chain-signatures/contract/src/state/mod.rs
```diff
@@ -149,6 +157,15 @@ impl ProtocolContractState {
         .map(|x| x.map(ProtocolContractState::Initializing))
     }
 
+    pub fn vote_code_hash(&mut self, code_hash: CodeHash) -> Result<u64, Error> {
```
kevindeforth, line 160: idem

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -193,6 +279,45 @@ impl MpcContract {
...
+                    .historical_code_hashes
+                    .push(code_hash.clone());
+                self.tee_state.allowed_code_hashes.insert(AllowedCodeHash {
+                    code_hash,
+                    added: env::block_height(),
+                });
+                self.tee_state.upgrade_deadline = Some(env::block_height() + TEE_UPGRADE_PERIOD);
+            }
+        }
+
```
kevindeforth, line 302: Lets return an error in case the contract is not in running state, since the vote is invalid in such a case.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -66,24 +67,103 @@ impl Default for VersionedMpcContract {
...
+        {
+            return false;
+        }
+
+        // Find the correct position to maintain sorted order by `added`
+        let insert_index = self
+            .allowed_code_hashes
+            .iter()
+            .position(|entry| new_code_hash.added <= entry.added)
+            .unwrap_or(self.allowed_code_hashes.len());
```
kevindeforth, line 119: nit: if we pass only the `CodeHash` as argument to `insert()` and populate the `added` field inside this function (setting to `env::block_height()`), then this step can be skipped and we save some gas.

kevindeforth (approved):
Thank you, that looks great!

Only minor cleanup left:
- The `CodeHashVotes` struct and its associated logic need to be removed from running.rs. CI should pass after that is done.
- Would be nice if the readme contains a short paragraph on the role of the TEE variables, but that can also be done later (in that case, opening an issue would suffice).

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -66,24 +67,103 @@ impl Default for VersionedMpcContract {
...
+        self.allowed_code_hashes.clone()
+    }
+}
+
+#[near(serializers=[borsh])]
+#[derive(Debug)]
+pub struct TeeState {
+    allowed_code_hashes: AllowedCodeHashes,
+    historical_code_hashes: Vec<CodeHash>,
+    upgrade_deadline: Option<BlockHeight>,
```
kevindeforth, line 136: ah, btw, I don't think we need this variable anymore, given that we have the `added` timestamp and `TEE_UPGRADE_PERIOD` is a constant.

libs/chain-signatures/contract/src/lib.rs
kevindeforth: Resolving, this has been addressed with https://github.com/Near-One/mpc/pull/376/commits/fc90940975c796d8297e3afdd9a79165d74ba2f3

libs/chain-signatures/contract/src/lib.rs
pbeza: Fixed in https://github.com/Near-One/mpc/pull/376/commits/e187ae1ac786589c1e5a2f46f445cf117be95206. ✅

libs/chain-signatures/contract/src/lib.rs
pbeza: Fixed in https://github.com/Near-One/mpc/pull/376/commits/cf0b0efeb3c17b9ed6e070456b40de88d77c2eae. ✅

libs/chain-signatures/contract/src/state/running.rs
pbeza: Fixed in https://github.com/Near-One/mpc/pull/376/commits/b179c24f7c97a347d30662e164f29b5192ea6193. ✅

libs/chain-signatures/contract/src/state/mod.rs
pbeza: Fixed in https://github.com/Near-One/mpc/pull/376/commits/b179c24f7c97a347d30662e164f29b5192ea6193. ✅

libs/chain-signatures/contract/src/lib.rs
pbeza: Fixed in https://github.com/Near-One/mpc/pull/376/commits/b179c24f7c97a347d30662e164f29b5192ea6193. ✅

pbeza:
@kevindeforth, thanks for the careful code review and all the suggestions! They were extremely helpful for me to resolve all the comments you added!

There is still [one test](https://github.com/Near-One/mpc/actions/runs/14820378420/job/41606505099?pr=376) that's failing though, even though I'm pretty sure I followed all your suggestions. Any ideas on how to fix it?

> Would be nice if the readme contains a short paragraph on the role of the TEE variables, but that can also be done later (in that case, opening an issue would suffice).

I just opened an issue: https://github.com/Near-One/mpc/issues/393. ✅

kevindeforth:
> There is still [one test](https://github.com/Near-One/mpc/actions/runs/14820378420/job/41606505099?pr=376) that's failing, even though I'm pretty sure I followed all your suggestions. Did I miss something? Any ideas on how to fix it?

I suspect it is because of the `proposed_code_hashes` field inside the running state (c.f. changes to file `running.rs`.

> BTW, do you think it makes sense to disable compiling the TEE-related code I added using some kind of compilation flag? @bowenwang1996 suggested this, but I'm not sure if it's actually necessary. Maybe we can just merge as-is?

Good question. Now that all tee-related information is in `TeeState`, which is a single entry in the contract struct, I think merging as-is is it ok.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -193,6 +284,45 @@ impl MpcContract {
         }
         Ok(())
     }
+
+    pub fn vote_code_hash(&mut self, code_hash: CodeHash) -> Result<(), Error> {
```
barakeinav1, line 288: can you please documentation of the function?

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -193,6 +284,45 @@ impl MpcContract {
...
+        // Ensure the protocol is in the Running state
+        let ProtocolContractState::Running(state) = &self.protocol_state else {
+            return Err(InvalidState::ProtocolStateNotRunning.into());
+        };
+
```
barakeinav1, line 293: I don't think we need TEE quote verification in this function

pbeza: This PR was merged ~3 weeks ago. Let's move the discussion to https://github.com/Near-One/mpc/pull/410.

PR #380, opened 2025-04-30
[Fix] Adjust gas attachment

kevindeforth:
Fix to [this discussion](https://nearone.slack.com/archives/C07UW93JVQ8/p1746013523358199?thread_ts=1746005094.777519&cid=C07UW93JVQ8):

- adjust the gas attachment so we don't run out of gas on failure
- move the cleaning mechanism into the `respond` call. This should allow us to remove the hanging request on testnet, but might have its own set of problems.

### DIFF
--- a/libs/chain-signatures/contract/src/lib.rs
+++ b/libs/chain-signatures/contract/src/lib.rs
@@ -455,14 +455,14 @@ impl VersionedMpcContract {
         if !signature_is_valid {
             return Err(RespondError::InvalidSignature.into());
         }
+
+        let Self::V1(mpc_contract) = self else {
+            env::panic_str("expected V1")
+        };
         // First get the yield promise of the (potentially timed out) request.
-        if let Some(YieldIndex { data_id }) = self.get_pending_request(&request) {
+        if let Some(YieldIndex { data_id }) = mpc_contract.pending_requests.remove(&request) {
             // Finally, resolve the promise. This will have no effect if the request already timed.
             env::promise_yield_resume(&data_id, &serde_json::to_vec(&response).unwrap());
-            let Self::V1(mpc_contract) = self else {
-                env::panic_str("expected V1")
-            };
-            mpc_contract.pending_requests.remove(&request);
             Ok(())
         } else {
             Err(InvalidParameters::RequestNotFound.into())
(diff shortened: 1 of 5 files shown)


libs/chain-signatures/contract/src/lib.rs
```diff
@@ -459,6 +459,10 @@ impl VersionedMpcContract {
         if let Some(YieldIndex { data_id }) = self.get_pending_request(&request) {
             // Finally, resolve the promise. This will have no effect if the request already timed.
             env::promise_yield_resume(&data_id, &serde_json::to_vec(&response).unwrap());
+            let Self::V1(mpc_contract) = self else {
+                env::panic_str("expected V1")
+            };
+            mpc_contract.pending_requests.remove(&request);
```
near-bookrock, line 465: Can we remove the entry before calling `env::promise_yield_resume`? If so, to save some gas you can call `remove` directly in the if let Some, instead of `get` + `remove` to avoid hashing &request twice.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -839,10 +843,10 @@ impl VersionedMpcContract {
         let Self::V1(mpc_contract) = self else {
             env::panic_str("expected V1")
         };
-        mpc_contract.pending_requests.remove(&request);
         match signature {
             Ok(signature) => PromiseOrValue::Value(signature),
             Err(_) => {
+                mpc_contract.pending_requests.remove(&request);
```
near-bookrock, line 849: Why do we need remove it in the error case if it is removed in `respond`. Will `respond` function also panic in the error case here when `return_signature_and_clean_state_on_success` panics?

libs/chain-signatures/contract/src/lib.rs
kevindeforth: nice!

libs/chain-signatures/contract/src/lib.rs
kevindeforth: > Why do we need remove it in the error case if it is removed in respond.

Because `respond` might not necessarily be called.

bowenwang1996 (approved):

near-bookrock (approved):

PR #387, opened 2025-05-01
Introduce channel id

kuksag:


bowenwang1996 (commented):
This change is not backward compatible right?

kuksag:
> This change is not backward compatible right?

Oh, yes you are right. Bumped the version

node/src/protocol_version.rs
```diff
@@ -14,4 +14,4 @@
 /// an incompatible protocol change, we are effectively creating a new network
 /// that is separate from the old network, thus requiring only threshold number
 /// of nodes to upgrade.
-pub const MPC_PROTOCOL_VERSION: u32 = 4;
+pub const MPC_PROTOCOL_VERSION: u32 = 5;
```
kevindeforth, line 17: good, can you also bump the node version in Cargo.toml?

kevindeforth (approved):
Thank you!
Nit: maybe move the definition of `UniqueId` from `assets.rs`  to `helper.rs` or something like that?

kuksag:
> maybe move the definition of UniqueId from assets.rs to helper.rs or something like that?

good point

node/src/protocol_version.rs
kuksag: I feel like those changes need to be either:
* Independent of each other, i.e. change Cargo.toml version only before publishing a release

or

* Be the same thing. i.e. `MPC_PROTOCOL_VERSION` should be the major component of the version
```
pub const MPC_PROTOCOL_VERSION = env!("CARGO_PKG_VERSION_MAJOR");
```
And then every PR should adjust the version (either minor/medium/major), but that will lead to the conflicts.

What do you think?

kevindeforth: You bring up an excellent point, we don't have an agreed-upon versioning process yet (at least, not that is know of).

I would have suggested we take a page from [SemVer](https://semver.org/):
- Major version change --> backwards-incompatible API changes or significant new features / functionality.
- Minor version change --> backwards-compatible API change, new features or functionality
- Patch version change --> backwards-compatible bug-fixes

In the context of MPC, a change in the communication protocol is a breaking change, but a breaking change might not necessarily be the result of a protocol change (e.g. changing a web-endpoint is technically a breaking API change).

To complicate things, in the context of MPC, even minor changes (like new feature support), can behave as breaking - in the sense that added functionality (e.g. eddsa support) does not break existing API, but only works if at least `threshold` nodes support it. Further, attempting to start a computation that is not supported by nodes of older minor versions may result in undefined behavior.

For that reason, I would have proposed that the MPC network requires major and minor version to match between nodes, but should be ok with different patch versions.

That would allow us to slightly deviate from the classic SemVer definition and treat communication protocol changes as minor version changes.

But I would be happy to have a more detailed discussion about this on a different channel, preferably also with @near-bookrock and @robin-near involved.

PR #390, opened 2025-05-02
accept re-submitting requests

kevindeforth:
Unfortunately, the cleanup function that was supposed to clear the state from timed out requests ran out of gas (c.f. #380 ).
This means that any signature requests that timed out since the latest release will remain in the state indefinitely, preventing the user from submitting the same request again.

This PR allows the user to re-submit signature requests. This guarantees to remove the request from the state, because:
 - either the MPC network will submit a signature response and remove the request from the state,
 - or the request will time out and be removed from the state in the cleanup function.

**But beware!** Re-submitting the same request before the time-out duration of 200 blocks has passed, or a response has been submitted, risks to result in another timed-out request. That is because the previous return promise would remove the request from the state and thus any attempt at responding to it will fail.

Timeline of edge case:
t0: submit request, cleanup-promise p0 [is generated](https://github.com/Near-One/mpc/blob/83e0970a3e7428011d3ab1fb7bbd916fdc0e299f/libs/chain-signatures/contract/src/lib.rs#L296)
t1: re-submit request, cleanup promise p1 is generated
t2: cleanup-promise p0 is triggered, [removing the request from the state](https://github.com/Near-One/mpc/blob/83e0970a3e7428011d3ab1fb7bbd916fdc0e299f/libs/chain-signatures/contract/src/lib.rs#L849)
t3: MPC network submits signature response to request from t1. Fails, because it is no longer in the state
t4: cleanup promise p1 is triggered.

### DIFF
--- a/libs/chain-signatures/contract/src/lib.rs
+++ b/libs/chain-signatures/contract/src/lib.rs
@@ -96,9 +96,11 @@ impl MpcContract {
         self.protocol_state.threshold()
     }
 
-    fn add_request(&mut self, request: &SignatureRequest, data_id: CryptoHash) {
+    // returns true if the request was already pending
+    fn add_request(&mut self, request: &SignatureRequest, data_id: CryptoHash) -> bool {
         self.pending_requests
-            .insert(request.clone(), YieldIndex { data_id });
+            .insert(request.clone(), YieldIndex { data_id })
+            .is_some()
     }
 
     fn get_pending_request(&self, request: &SignatureRequest) -> Option<YieldIndex> {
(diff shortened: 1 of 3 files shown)


libs/chain-signatures/contract/src/lib.rs
```diff
@@ -96,9 +96,11 @@ impl MpcContract {
         self.protocol_state.threshold()
     }
 
-    fn add_request(&mut self, request: &SignatureRequest, data_id: CryptoHash) {
+    // returns true if the request was already pending
```
near-bookrock, line 99:
```suggestion
    /// returns true if the request was already pending
```

near-bookrock (approved):
lgtm. I'd wait for @bowenwang1996 or @robin-near to also have a look at this change.

bowenwang1996 (approved):

PR #401, opened 2025-05-09
Devnet - fetch contract from testnet & v1 cleanup

### LINKED ISSUE #400: Devnet mpc-setup improvements
the devnet cli should by default fetch the contract binary from testnet or mainnet and deploy that one.

kevindeforth:
Changes:
- remove V1 logic from devnet (it is no longer required, now that we deployed V2 on both test- and devnet).
- if no path to the contract is indicated, the contract under `v1.signer-prod.testnet` will be copied and re-deployed (resolves #400 )


near-bookrock (approved):
Thanks for the update!

PR #406, opened 2025-05-12
chore: remove unwrap in `monitor_passive_channels_inner`

near-bookrock:


node/src/mpc_client.rs
```diff
@@ -299,14 +299,18 @@ impl MpcClient {
...
             tasks.spawn_checked(
                 &format!("passive task {:?}", channel.task_id()),
                 async move { mpc_clone.process_channel_task(channel).await },
             );
         }
+
+        tracing::info!(
+            "Network task channel receiver is closed. Exiting monitor_passive_channels_inner."
+        );
+        Ok(())
```
kevindeforth, line 313: shouldn't we return an error here?

near-bookrock: Good point. 

I don't have a strong opinion about the return type. If we also return an Error here, we might as well return `anyhow::Error` instead of a result.

Most important is that we don't have these unsafe unwraps.

kevindeforth (approved):
thank you!

PR #408, opened 2025-05-13
Add backward compatibility test

kuksag:
### Overview
An experiment to bring contract state back-compatibility testing.

It requires storing a single precompiled version of the contract. This version corresponds to the latest breaking change in the contract state. 

### Pipeline 
1. Load "latest breaking change" contract. Initialize it.
2. Deploy current version of the contract
3. Call any view function (we want to check that state loads in correctly).
4. If it hasn't succeeded, call `migrate()`. If it's succeeded, prompt user to update "latest breaking change contract"
5. Otherwise, user is asked to add logic into `migrate()`

### Storing an artefact in the repo
Compiled + optimized contract weights ~546kb, which increases the repo size from 3.32 mb -> to 3.86 mb. I find this tradeoff acceptable.

The other way around would be to store a commit hash of the latest incompatible change. That would require some trickery around in-place/on-flight atomic git checkout + contract compilation (so that we do not accidentally break user's working tree)

note: this is not a vibe-coding, I genuinely placed those emojis around 😁


near-bookrock:
Thanks a lot for this!

near-bookrock (approved):

PR #410, opened 2025-05-13
Draft TEE quote verification implementation

pbeza:


libs/chain-signatures/contract/Cargo.toml
```diff
@@ -30,6 +30,7 @@ k256 = { version = "0.13.4", features = [
     "arithmetic",
     "expose-field",
 ] }
+dcap-qvl = { git = "https://github.com/mattlockyer/dcap-qvl" }
```
pbeza, line 33: Some sketchy code changes there that might impact contract verification security. :/ Needs a closer look.

libs/chain-signatures/contract/src/primitives/participants.rs
```diff
@@ -1,18 +1,49 @@
...
+    /// TEE Remote Attestation Quote that proves the participant's identity.
+    pub tee_quote: Vec<u8>,
+    /// Supplemental data for the TEE quote, including Intel certificates to verify it came from
+    /// genuine Intel hardware, along with details about the Trusted Computing Base (TCB)
+    /// versioning, status, and other relevant info.
+    pub quote_collateral: String,
+}
+
+/// We do not use #[derive(Debug)] for ParticipantInfo because TEE quote and collateral fields are
+/// long and can cause HostError(TotalLogLengthExceeded { length: 120461, limit: 16384 }) errors.
```
kevindeforth, line 29: `Participants` is de-serialized from the contract state for pretty much every contract call. 

How large are `tee_quote` and `quote_collateral`? If they are causing this issue with the logging, we might want to store them in such a way that they are de-serialized from state only when needed.

pbeza: > How large are `tee_quote` and `quote_collateral`?

Around `4kB` to a few kB.

> (...) we might want to store them in such a way that they are de-serialized from state only when needed.
How do we that?

pbeza: Thanks! I was hoping it'd be as easy as just swapping `std::Vec<u8>` with `near_sdk::store::Vector<u8>`, but nope. I'm hitting this error now:
```
error[E0369]: binary operation `==` cannot be applied to type `near_sdk::store::Vector<u8>`
   --> libs/chain-signatures/contract/src/primitives/participants.rs:21:5
    |
15  | #[derive(PartialEq, Eq, PartialOrd, Ord, Clone)]
    |          --------- in this derive macro expansion
...
21  |     pub tee_quote: Vector<u8>,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^
    |
note: the foreign item type `near_sdk::store::Vector<u8>` doesn't implement `PartialEq`
   --> /home/patryk/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/near-sdk-5.11.0/src/store/vec/mod.rs:116:1
    |
116 | pub struct Vector<T>
    | ^^^^^^^^^^^^^^^^^^^^ not implement `PartialEq`
    = note: this error originates in the derive macro `PartialEq` (in Nightly builds, run with -Z macro-backtrace for more info)
```
Any ideas? 🤔

libs/chain-signatures/contract/Cargo.toml
barakeinav1, line 33: are we ok with using a git repository that is not part of near one?
should we copy to near-one github?

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -28,31 +30,36 @@ use near_sdk::{
...
 // Register used to receive data id from `promise_await_data`.
 const DATA_ID_REGISTER: u64 = 0;
+
 // Prepaid gas for a `return_signature_and_clean_state_on_success` call
 const RETURN_SIGNATURE_AND_CLEAN_STATE_ON_SUCCESS_CALL_GAS: Gas = Gas::from_tgas(5);
+
 // Prepaid gas for a `update_config` call
 const UPDATE_CONFIG_GAS: Gas = Gas::from_tgas(5);
-// Maximum time after which TEE MPC nodes must be upgraded to the latest version
-const TEE_UPGRADE_PERIOD: BlockHeight = 604800; // ~7 days
```
barakeinav1, line 55: why was this removed? where is the logic to kick out a node in case it was not upgraded?

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -826,15 +828,29 @@ impl VersionedMpcContract {
     }
 
     #[handle_result]
-    pub fn vote_code_hash(&mut self, code_hash: CodeHash) -> Result<(), Error> {
+    pub fn vote_code_hash(
```
barakeinav1, line 831: according to the flow https://docs.nearone.org/doc/integrate-tee-into-mpc-network-OWBOeCofNK#h-update-mpc-docker-image-hash-in-the-contract
-there is no need to perform TEE quote verification since the request is coming from the operator and not the MPC node

libs/chain-signatures/contract/src/lib.rs
pbeza: It wasn't removed — I just moved it here since I kinda felt it fits better there:

https://github.com/Near-One/mpc/blob/e150d95528565a7c3fafe41b6c072b9d1b810244/libs/chain-signatures/contract/src/primitives/tee/proposal.rs#L6-L7

libs/chain-signatures/contract/src/primitives/tee/proposal.rs
```diff
@@ -0,0 +1,99 @@
...
+/// run.
+#[near(serializers=[borsh])]
+#[derive(Debug, Default)]
+pub struct AllowedTeeProposals {
+    /// Whitelisted code hashes, sorted by when they were added (oldest first). Expired entries are
+    /// lazily cleaned up during insertions and lookups.
+    allowed_tee_proposals: Vec<AllowedTeeProposal>,
+}
+
+impl AllowedTeeProposals {
```
barakeinav1, line 36: I don't understand the purpose of this method, what is it for?

barakeinav1 (dismissed):

barakeinav1:
I did see where are you verifying the a new participant quote when calling add_new_paramaters.
can you point me to it?

libs/chain-signatures/contract/src/primitives/tee/proposal.rs
pbeza: Not sure which method you mean. The `AllowedTeeProposal` `struct` is just a proposal to add a new TEE code hash to the whitelist, along with when it was added and the proposer’s TEE quote.

The `pub code_hash: CodeHash` part is probably self-explanatory — it’s the hash of the Docker image that participants vote to whitelist. MPC nodes can only run Docker images that are already whitelisted here. When we add a new code hash to the whitelist, we effectively start a 'timer' for all the previous hashes. After `TEE_UPGRADE_PERIOD`, those older hashes get removed. The `impl AllowedTeeProposals` just implements this idea of a grace period for previously whitelisted hashes.

You can also check out the great review @kevindeforth left on it here: https://github.com/Near-One/mpc/pull/376#discussion_r2065694609. The idea behind this was also discussed with @bowenwang1996 here: https://github.com/Near-One/mpc/pull/376#discussion_r2071923433.

libs/chain-signatures/contract/src/primitives/participants.rs
```diff
@@ -1,18 +1,47 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("ParticipantInfo")
+            .field("url", &self.url)
+            .field("sign_pk", &self.sign_pk)
+            .finish()
+    }
+}
+
+impl ParticipantInfo {
+    pub fn verify_quote(&self, timestamp: u64) -> Result<VerifiedReport, Error> {
```
barakeinav1, line 40: we need to add to verify_quote all checks  that verify the RTMRs
see https://docs.nearone.org/doc/integrate-tee-into-mpc-network-OWBOeCofNK#h-attestation-verification-in-the-contract

libs/chain-signatures/contract/src/primitives/tee/proposal.rs
pbeza: [`clean`](https://github.com/Near-One/mpc/blob/e150d95528565a7c3fafe41b6c072b9d1b810244/libs/chain-signatures/contract/src/primitives/tee/proposal.rs#L38) is a private function, so you can’t call it directly — but I really appreciate your vigilance here! For a moment, I actually thought there might be a bug :)

https://github.com/Near-One/mpc/blob/e150d95528565a7c3fafe41b6c072b9d1b810244/libs/chain-signatures/contract/src/primitives/tee/proposal.rs#L38

pbeza: Now that I think about it, this might actually be a bug — if we call `get()` after the `TEE_UPGRADE_PERIOD` has passed, `clean()` will get triggered and all the whitelisted entries will be wiped... 🤦🏼 @kevindeforth am I missing something here? 🤔 I should’ve added a unit test for this. I’ll fix it. @barakeinav1, thanks again for the curiosity — pretty sure it helped catch a real bug here 🙏🏼

kevindeforth: I thought `clean` only removes expired hashes. https://github.com/Near-One/mpc/blob/e150d95528565a7c3fafe41b6c072b9d1b810244/libs/chain-signatures/contract/src/primitives/tee/proposal.rs#L37

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -259,6 +268,10 @@ impl MpcContract {
         prospective_epoch_id: EpochId,
         proposal: &ThresholdParameters,
     ) -> Result<(), Error> {
+        let tee_verification = self.tee_state.verify_participants();
```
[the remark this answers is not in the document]
pbeza: > have a compiler flag that allows to skip the tee verification step. When we just want to test the node, this will be annoying.

@kuksag AFAIK, is working on it (https://github.com/Near-One/mpc/issues/442#issuecomment-2916549957).

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -72,80 +80,85 @@ impl Default for VersionedMpcContract {
...
 
-impl AllowedCodeHashes {
-    /// Removes all expired code hashes and returns the number of removed entries.
-    fn clean(&mut self, current_block_height: BlockHeight) -> usize {
-        // Find the first non-expired entry
-        let expired_count = self
-            .allowed_code_hashes
+impl TeeState {
+    /// Verifies all TEE participant quotes. Used for periodic verification of TEE participants.
+    pub fn verify_participants(&self) -> bool {
```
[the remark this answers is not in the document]
pbeza: Done here: https://github.com/Near-One/mpc/pull/410/commits/2e6cdaa41c341e71282fc59bdd997cf3e8b8a502. ✅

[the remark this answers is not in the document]
kevindeforth: This looks good, thank you, I can work with that!

libs/chain-signatures/contract/src/primitives/tee/proposal.rs
```diff
@@ -0,0 +1,165 @@
...
+// Maximum time after which TEE MPC nodes must be upgraded to the latest version
+const TEE_UPGRADE_PERIOD: BlockHeight = 7 * 24 * 60 * 100; // ~7 days @ block time of 600 ms, e.g. 100 blocks every 60 seconds
+
+/// Proposal for a new TEE code hash to be added to the whitelist, along with the TEE quote that
+/// includes the RTMR3 measurement among others.
+#[near(serializers=[borsh, json])]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub struct TeeProposal {
+    pub code_hash: CodeHash,
+    pub tee_quote: TeeQuote,
```
kevindeforth, line 15: So this might be another stupid question, but why do we need the `tee_quote` in the Proposal?

I thought the `tee_quote` was the signed structure generated by the TEE platform and that it already includes the `code_hash` (docker image hash).

To my untrained eye, it looks like we are extracting the `code_hash` when verifying the `tee_quote` and then ensuring that the participant is running on a valid image.

https://github.com/Near-One/mpc/blob/5f86401ac055be3b8dae74365cf87749833d38cb/libs/chain-signatures/contract/src/primitives/tee/quote.rs#L93


To be more explicit, I would have expected something like this:
- contract keeps track of the hashes of allowed docker images. This is common to all participants.
- participant submits their `tee_quote`. During verification. This quote is unique to each participant and the used docker image hash is extracted and compared to the list of allowed hashes. Verification fails if the image is not in that list.

Apologies if this is all complete nonsense, I am still learning about TEEs.

libs/chain-signatures/contract/src/primitives/tee/proposal.rs
```diff
@@ -0,0 +1,165 @@
...
+        expired_count
+    }
+
+    /// Inserts a new code hash into the list after cleaning expired entries. Maintains the sorted
+    /// order by `added` (ascending). Returns `true` if the insertion was successful, `false` if the
+    /// code hash already exists.
+    pub fn insert(
+        &mut self,
+        code_hash: CodeHash,
+        tee_quote: TeeQuote,
```
kevindeforth, line 61: maybe just pass `TeeProposal`?

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -72,80 +81,82 @@ impl Default for VersionedMpcContract {
...
+    pub fn tee_status(&self, participants: Vec<AccountId>) -> BTreeMap<AccountId, TeeQuoteStatus> {
+        let now_sec = env::block_timestamp_ms() / 1_000;
+        participants
+            .into_iter()
+            .map(|account_id| {
+                let status = self
+                    .tee_participant_info
+                    .get(&account_id)
+                    .map(|tee_participant_info| {
+                        TeeQuoteStatus::from(tee_participant_info.verify_quote(now_sec))
```
kevindeforth, line 113: Don't we also need to ensure that the participant is running a docker image that is is whitelisted?

libs/chain-signatures/contract/src/primitives/participants.rs
```diff
@@ -15,6 +19,55 @@ pub struct ParticipantInfo {
     pub sign_pk: PublicKey,
 }
 
+#[near(serializers=[borsh, json])]
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Default)]
+pub struct TeeParticipantInfo {
```
kevindeforth, line 24: Lets move this to the `tee` folder.

libs/chain-signatures/contract/src/primitives/key_state.rs
```diff
@@ -239,6 +243,8 @@ pub mod tests {
 
     #[test]
     fn test_authenticated_participant_id() {
+        set_test_env_for_tee_quote_verification();
```
kevindeforth, line 246: these should not be required here.

libs/chain-signatures/contract/src/primitives/key_state.rs
```diff
@@ -252,8 +258,11 @@ pub mod tests {
             assert!(AuthenticatedParticipantId::new(proposed_parameters.participants()).is_err());
         }
     }
+
     #[test]
     fn test_authenticated_account_id() {
+        set_test_env_for_tee_quote_verification();
```
kevindeforth, line 264: idem

libs/chain-signatures/contract/src/primitives/participants.rs
```diff
@@ -179,13 +234,22 @@ impl Participants {
...
+    use std::time::Duration;
 
     #[test]
     fn test_participants() {
+        let now_sec = 1_747_785_600u64; // 2025-05-21 00:00:00 UTC for TEE quote verification
+        let now_ns = Duration::from_secs(now_sec).as_nanos() as u64; // nanoseconds since epoch
+        let mut ctx = VMContextBuilder::new();
+        ctx.block_timestamp(now_ns);
+        testing_env!(ctx.build());
+
```
kevindeforth, line 252: these should not be required in these unit tests.

libs/chain-signatures/contract/src/primitives/mod.rs
```diff
@@ -1,8 +1,8 @@
-pub mod code_hash;
 pub mod domain;
 pub mod key_state;
 pub mod participants;
 pub mod signature;
+pub mod tee;
```
kevindeforth, line 5: I would move the `tee` folder outside of the `primitives` folder, since there is no dependency.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -275,35 +255,57 @@ impl MpcContract {
...
+        raw_tcb_info: String,
+    ) -> Result<(), Error> {
         // Ensure the protocol is in the Running state
         let ProtocolContractState::Running(state) = &self.protocol_state else {
             return Err(InvalidState::ProtocolStateNotRunning.into());
         };
 
-        // Authenticate the participant and cast a vote
-        // TODO: Verify TEE quote here. See GitHub issue #378: https://github.com/Near-One/mpc/issues/378
+        // Check participant identity, verify their TEE quote, and cast a vote
```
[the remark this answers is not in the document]
kevindeforth: agreed with Barak, I don't think we need verification here.

libs/chain-signatures/contract/src/primitives/participants.rs
```diff
@@ -15,6 +19,55 @@ pub struct ParticipantInfo {
...
+                "quote_collateral",
+                &str_preview(&self.quote_collateral, 256),
+            )
+            .finish()
+    }
+}
+
+impl TeeParticipantInfo {
+    pub fn verify_quote(&self, timestamp: u64) -> Result<VerifiedReport, Error> {
+        let tee_collateral = get_collateral(self.quote_collateral.clone());
```
kevindeforth, line 65: so all `get_collateral` seems to be doing, is to de-compose the string.
https://github.com/Near-One/mpc/blob/5f86401ac055be3b8dae74365cf87749833d38cb/libs/chain-signatures/contract/src/primitives/tee/quote.rs#L45

Why don't we just have the nodes send a `QuoteCollateral` struct and use serde json serialization / deserialization?

And then store that struct in the contract instead of the `quote_collateral` string?

libs/chain-signatures/contract/src/primitives/thresholds.rs
```diff
@@ -32,16 +35,22 @@ impl Threshold {
 pub struct ThresholdParameters {
     participants: Participants,
     threshold: Threshold,
+    tee_participant_info: TeeParticipantInfo,
```
kevindeforth, line 38: (just to inform any other reviewers that were not part of the discussion yesterday, this will be removed from here and verification will be done on the `MpcContract` level)

libs/chain-signatures/contract/src/primitives/tee/proposal.rs
barakeinav1: agree with Kevin, code hash is common property while TeeQuote is per participant

libs/chain-signatures/contract/src/primitives/participants.rs
pbeza: Done: https://github.com/Near-One/mpc/pull/410/commits/f6e44c80b1302d1a8c817496b6d6c5549750b0b2. ✅

libs/chain-signatures/contract/src/primitives/key_state.rs
pbeza: Fixed here: https://github.com/Near-One/mpc/pull/410/commits/c6e8945cb85d0854fd3544f1a81dccff38cb1399. ✅

libs/chain-signatures/contract/src/primitives/mod.rs
pbeza: Fixed: https://github.com/Near-One/mpc/pull/410/commits/f9f2bcb95058f5fd9a90c5ac3a279d3c717f5dbe. ✅

libs/chain-signatures/contract/src/primitives/thresholds.rs
```diff
@@ -141,18 +155,32 @@ impl ThresholdParameters {
...
...56544d466b77457759484b6f5a497a6a3043415159494b6f5a497a6a3044415163445167414543366e45774d4449595a4f6a2f69505773437a61454b69370a314f694f534c52466857476a626e42564a66566e6b59347533496a6b4459594c304d784f346d717379596a6c42616c54565978465032734a424b357a6c4b4f420a757a43427544416642674e5648534d4547444157674251695a517a575770303069664f44744a5653763141624f5363477244425342674e5648523845537a424a0a4d45656752614244686b466f64485277637a6f764c324e6c636e52705a6d6c6a5958526c63793530636e567a6447566b63325679646d6c6a5a584d75615735300a5a577775593239744c306c756447567355306459556d397664454e424c6d526c636a416442674e564851344546675155496d554d316c71644e496e7a673753560a55723951477a6b6e4271777744675944565230504151482f42415144416745474d42494741315564457745422f7751494d4159424166384341514577436759490a4b6f5a497a6a3045417749445351417752674968414f572f35516b522b533943695344634e6f6f774c7550524c735747662f59693747535839344267775477670a41694541344a306c72486f4d732b586f356f2f7358364f39515778485241765a55474f6452513763767152586171493d0a2d2d2d2d2d454e442043455254494649434154452d2d2d2d2d0a0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000";
```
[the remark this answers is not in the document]
near-bookrock: We can move these strings to a file and include them at compile time with [include_str](https://doc.rust-lang.org/std/macro.include_str.html).

libs/chain-signatures/contract/src/primitives/thresholds.rs
pbeza: I already removed it. ✅

https://github.com/near/mpc/blob/bc64afbb57f6a854a4a5453709a831fc4203c898/libs/chain-signatures/contract/src/tee/tee_participant.rs#L9-L19

pbeza:
> You should include a git revision when depending on repositories. I.e. include the rev atrribute. See docs if unsure here https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#choice-of-commit

Fixed here: https://github.com/near/mpc/pull/410/commits/bde78767599f84c924f2bbe39f4908ef3d65d1de. ✅

node/src/indexer/fake.rs
```diff
@@ -40,7 +39,8 @@ impl FakeMpcContractState {
         let config = Config {
             key_event_timeout_blocks: 10,
         };
-        let env = Environment::new(None, None, None);
+        let block_timestamp = 1757785600_u64 * 1_000_000_u64; // 2025-05-21 00:00:00 UTC to ensure TEE quote verification succeeds
+        let env = Environment::new(None, None, None, Some(block_timestamp));
```
kevindeforth, line 43: Here too, this should not be required by default.
The entire set-up should still be working outside of TEE for the time being.

kuksag:
PSA: if the building for `wasm32-unknown-unknown` on MacOS breaks for you after [1bd1764](https://github.com/near/mpc/pull/410/commits/1bd1764d2b7fab855be201341a343bf8ab741bc5) refer to https://github.com/briansmith/ring/issues/1824#issuecomment-2059955073

kuksag:
Does anybody have to use wasm-opt with `--enable-bulk-memory` option enabled (otherwise wasm-opt fails)? I tried to use first buildable commit in this branch which is fb4dab58 and the contract can not be deployed in near workspaces `FunctionCallError(CompilationError(PrepareError(Deserialization)))` regardless of using wasm-opt 😞.
Anyways, I couldn't investigate further what package requires bulk memory operations all of a sudden (`dcap-qvl` perhaps), but more importantly I can not run the tests at the moment. Is it MacOS problem?

libs/chain-signatures/contract/src/primitives/participants.rs
```diff
@@ -91,6 +91,7 @@ impl Participants {
     ///  - All participant IDs are unique.
     ///  - All account IDs are unique.
     ///  - The next_id is greater than all participant IDs.
+    ///  - All participant TEE quotes are valid.
```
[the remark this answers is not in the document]
pbeza: Good catch! Thanks! Fixed here: https://github.com/near/mpc/pull/410/commits/5f744a7d24ab541cd0ee84c4cdeca0e4ac4795c0.

libs/chain-signatures/contract/src/primitives/test_utils.rs
```diff
@@ -95,3 +96,11 @@ pub fn gen_threshold_params(max_n: usize) -> ThresholdParameters {
     let k = rand::thread_rng().gen_range(k_min..n + 1);
     ThresholdParameters::new(gen_participants(n), Threshold::new(k as u64)).unwrap()
 }
+
+pub fn set_test_env_for_tee_quote_verification() {
```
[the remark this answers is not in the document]
pbeza: I removed it [here](https://github.com/near/mpc/pull/410/commits/c3476118e5cb2a96fcc0887d89da2dc2a8a943ea#diff-a807c484ed52173d1752b797aae5adec2683c640bad05627c86ea52928790d09) since it's not needed anymore.

libs/chain-signatures/contract/src/primitives/thresholds.rs
```diff
@@ -72,11 +72,13 @@ impl ThresholdParameters {
 
     pub fn validate(&self) -> Result<(), Error> {
         Self::validate_threshold(self.participants.len() as u64, self.threshold())?;
-        self.participants.validate()
+        self.participants.validate()?;
+        Ok(())
```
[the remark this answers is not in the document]
pbeza: I'd expect `clippy` to catch that... :/ [Fixed](https://github.com/near/mpc/pull/410/commits/f7d2dd4ddf9ce41e971cddd1fc0944082d629dca). ✅

node/src/indexer/fake.rs
pbeza: Sorry, these are leftovers from one of my tests. [Fixed](https://github.com/near/mpc/pull/410/commits/c3476118e5cb2a96fcc0887d89da2dc2a8a943ea#diff-a7a28b10ed6f4ebdcf1220c768f10987fdbae5585685c2ad51fc1cf0f09868f5). ✅

libs/chain-signatures/contract/src/state/running.rs
```diff
@@ -229,7 +229,8 @@ pub mod running_tests {
             state.parameters.participants().len(),
             state.parameters.threshold().value()
         );
-        let mut env = Environment::new(None, None, None);
+        let block_timestamp = 1757785600_u64 * 1_000_000_u64; // 2025-05-21 00:00:00 UTC to ensure TEE quote verification succeeds
+        let mut env = Environment::new(None, None, None, Some(block_timestamp));
```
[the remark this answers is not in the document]
pbeza: Sorry, these are leftovers from one of my tests. [Fixed](https://github.com/near/mpc/pull/410/commits/c3476118e5cb2a96fcc0887d89da2dc2a8a943ea#diff-a7a28b10ed6f4ebdcf1220c768f10987fdbae5585685c2ad51fc1cf0f09868f5). ✅

libs/chain-signatures/contract/src/state/resharing.rs
```diff
@@ -178,6 +183,8 @@ mod tests {
     }
 
     fn test_resharing_contract_state_for(num_domains: usize) {
+        set_test_env_for_tee_quote_verification();
```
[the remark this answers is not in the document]
pbeza: [Fixed](https://github.com/near/mpc/pull/410/commits/c3476118e5cb2a96fcc0887d89da2dc2a8a943ea#diff-31c129aa9cb31d291120fad1c2fff6715af7c04d8622106f19b8b8a9f4d5fa90). ✅ I [completely removed](https://github.com/near/mpc/pull/410/commits/c3476118e5cb2a96fcc0887d89da2dc2a8a943ea#diff-a807c484ed52173d1752b797aae5adec2683c640bad05627c86ea52928790d09) `set_test_env_for_tee_quote_verification`.

libs/chain-signatures/contract/tests/sign.rs
```diff
@@ -354,6 +354,7 @@ async fn test_contract_initialization() -> anyhow::Result<()> {
         }))
         .transact()
         .await?;
+    eprintln!("result: {result:?}"); // TODO
```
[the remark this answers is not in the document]
pbeza: Yup. Sorry, overlooked it. [Fixed](https://github.com/near/mpc/pull/410/commits/fb0124a854adb150f9eecfc668921f24365420ac#diff-204489480325207a99607d54c7b225554cb76d81c348dc6f183c1a4719c6384f). ✅

libs/chain-signatures/contract/src/primitives/thresholds.rs
```diff
@@ -180,16 +187,19 @@ mod tests {
...
+        set_test_env_for_tee_quote_verification();
+
         let n: usize = rand::thread_rng().gen_range(2..600);
         let min_threshold = ((n as f64) * 0.6).ceil() as usize;
 
         let participants = gen_participants(n);
         for k in 1..min_threshold {
             let invalid_threshold = Threshold::new(k as u64);
-            assert!(ThresholdParameters::new(participants.clone(), invalid_threshold).is_err());
+            assert!(ThresholdParameters::new(participants.clone(), invalid_threshold,).is_err());
```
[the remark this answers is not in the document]
pbeza: Fixed [here](https://github.com/near/mpc/pull/410/files#diff-8fe5f6792f6ae4aa5b49c0c6415c6fd4acf1c142cf6da2c8245dcc15cb12aa80). ✅

libs/chain-signatures/contract/src/primitives/thresholds.rs
```diff
@@ -180,16 +187,19 @@ mod tests {
...
         }
         assert!(
-            ThresholdParameters::new(participants.clone(), Threshold::new((n + 1) as u64)).is_err()
+            ThresholdParameters::new(participants.clone(), Threshold::new((n + 1) as u64),)
```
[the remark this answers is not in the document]
pbeza: Fixed [here](https://github.com/near/mpc/pull/410/files#diff-8fe5f6792f6ae4aa5b49c0c6415c6fd4acf1c142cf6da2c8245dcc15cb12aa80). ✅

libs/chain-signatures/contract/src/tee/proposal.rs
```diff
@@ -0,0 +1,203 @@
...
+use crate::primitives::key_state::AuthenticatedParticipantId;
+
+// Maximum time after which TEE MPC nodes must be upgraded to the latest version
+const TEE_UPGRADE_PERIOD: BlockHeight = 7 * 24 * 60 * 100; // ~7 days @ block time of 600 ms, e.g. 100 blocks every 60 seconds
+
+/// Proposal for a new TEE code hash to be added to the whitelist, along with the TEE quote that
+/// includes the RTMR3 measurement among others.
+#[near(serializers=[borsh, json])]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub struct TeeProposal {
```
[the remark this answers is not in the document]
pbeza: Fixed [here](https://github.com/near/mpc/pull/410/commits/fb0124a854adb150f9eecfc668921f24365420ac#diff-abefba90d0fc116fd05d9207a90db8daef4d064a38708bc35db1df212236bbe4R14). ✅

libs/chain-signatures/contract/src/state/key_event.rs
```diff
@@ -330,8 +330,8 @@ pub mod tests {
             ctx.block_height(self.block_height);
             ctx.random_seed(self.seed);
             ctx.signer_account_id(self.signer.clone());
-            let unix_time = 1_747_785_600u64; // 2025-05-21 00:00:00 UTC for TEE quote verification
-            let now_ns = Duration::from_secs(unix_time).as_nanos() as u64; // nanoseconds since epoch
+            let now_sec = 1_747_785_600u64; // 2025-05-21 00:00:00 UTC for TEE quote verification
```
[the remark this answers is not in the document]
pbeza: After moving the TEE verification logic, there's no need to mock the block timestamp anymore, so this code is no longer necessary.

libs/chain-signatures/contract/src/state/initializing.rs
```diff
@@ -162,7 +162,8 @@ mod tests {
         num_domains: usize,
         num_generated: usize,
     ) -> (Environment, InitializingContractState) {
-        let mut env = Environment::new(None, None, None);
+        let block_timestamp = 1757785600_u64 * 1_000_000_u64; // 2025-05-21 00:00:00 UTC to ensure TEE quote verification succeeds
+        let mut env = Environment::new(None, None, None, Some(block_timestamp));
```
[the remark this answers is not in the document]
pbeza: Sorry, these are leftovers from one of my tests. [Fixed](https://github.com/near/mpc/pull/410/commits/c3476118e5cb2a96fcc0887d89da2dc2a8a943ea#diff-1a536516b116c3343e8e5d6ad59c691c8dd0d5dedac4faf112d6894d80cb77c5). ✅

libs/chain-signatures/contract/Cargo.toml
pbeza: Will be addressed by @near-bookrock in a separate PR (see: https://github.com/orgs/near/projects/175?pane=issue&itemId=113958602&issue=near%7Cmpc%7C451).

pbeza: > You should include a git revision when depending on repositories. I.e. include the rev atrribute. See docs if unsure here https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#choice-of-commit

Done [here](https://github.com/near/mpc/pull/410/files#diff-adb834fd214ef7b0f6cf6859677ae86fea9e374fe05af3a2ecd28df0d6099b2aR33). ✅

> are we ok with using a git repository that is not part of near one?
> should we copy to near-one github?

Will be addressed by @near-bookrock in a separate PR. Ref: https://github.com/near/mpc/issues/451.

pbeza:
> (...) more importantly I can not run the tests at the moment

WDYM? @kuksag How can I reproduce?

libs/chain-signatures/contract/src/tee/proposal.rs
```diff
@@ -0,0 +1,199 @@
...
+pub struct AllowedTeeProposals {
+    /// Whitelisted code hashes, sorted by when they were added (oldest first). Expired entries are
+    /// lazily cleaned up during insertions and lookups.
+    allowed_tee_proposals: Vec<AllowedDockerImageHash>,
+}
+
+impl AllowedTeeProposals {
+    /// Removes all expired code hashes and returns the number of removed entries.
+    /// Ensures that at least one (the latest) proposal always remains in the whitelist.
+    fn clean(&mut self, current_block_height: BlockHeight) -> usize {
```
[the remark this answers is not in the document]
pbeza: [Fixed](https://github.com/near/mpc/pull/410/commits/d27db9dc8220ab38300c64ea1b27dfdc1443c907). ✅

libs/chain-signatures/contract/src/tee/proposal.rs
```diff
@@ -0,0 +1,199 @@
...
+#[derive(Debug, Clone)]
+pub struct AllowedDockerImageHash {
+    pub image_hash: DockerImageHash,
+    pub added: BlockHeight,
+}
+/// Collection of whitelisted Docker code hashes that are the only ones MPC nodes are allowed to
+/// run.
+#[near(serializers=[borsh])]
+#[derive(Debug, Default)]
+pub struct AllowedTeeProposals {
```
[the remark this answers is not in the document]
pbeza: [Fixed](https://github.com/near/mpc/pull/410/commits/ca891c14bf3c9dfc4bd4d436717c2620bd987f35). ✅

libs/chain-signatures/contract/src/state/key_event.rs
```diff
@@ -308,6 +310,7 @@ pub mod tests {
             let block_height = block_height.unwrap_or(rand::thread_rng().gen());
             ctx.block_height(block_height);
             ctx.random_seed(seed);
+            ctx.block_timestamp(0);
```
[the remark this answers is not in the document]
pbeza: No, I just overlooked it. Thanks for noticing! [Fixed](https://github.com/near/mpc/pull/410/commits/fcc994c04692ffdbdb7df624a02addf0cbd6d254). ✅

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -259,6 +262,19 @@ impl MpcContract {
...
+        let participant_ids = self
+            .tee_state
+            .tee_participant_info
+            .keys()
+            .cloned()
+            .collect();
+        let tee_verification = self.tee_state.tee_status(participant_ids);
+        if tee_verification
+            .values()
+            .any(|status| *status != TeeQuoteStatus::Valid && *status != TeeQuoteStatus::None)
```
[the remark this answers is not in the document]
pbeza: Good point. @kevindeforth, PTAL and let me know if this fix works for you, or if you’d suggest handling it differently: https://github.com/near/mpc/pull/410/commits/f32adde5e5a4d086b287d30fca52d66e8cc1fdf8.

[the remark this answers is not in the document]
pbeza: [Fixed](https://github.com/near/mpc/pull/410/commits/4d610b887e6921c4f1bffae6a1718db4dd05acbe). ✅

libs/chain-signatures/contract/Cargo.toml
```diff
@@ -31,6 +31,7 @@ k256 = { version = "0.13.4", features = [
     "arithmetic",
     "expose-field",
 ] }
+dcap-qvl = { git = "https://github.com/mattlockyer/dcap-qvl", rev = "3caedd8fd0f365841abaf1430ff2922fcb09d9db" }
```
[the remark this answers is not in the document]
pbeza: > Nit: Can we add a comment as to why we're using this particular commit?

It was simply the latest commit available when the dependency was added.

> Also why are we using a fork rather than the upstream from Phala network? Do we have this written down somewhere?

We’re using that fork because the original version from Phala isn’t compatible with NEAR contracts. @mattlockyer forked it and made some [workaround-ish](https://github.com/near/mpc/pull/410#discussion_r2086324618) changes to make it work with NEAR smart contracts. See: https://github.com/near/mpc/issues/451.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -27,42 +31,48 @@ use near_sdk::{
...
-const TEE_UPGRADE_PERIOD: BlockHeight = 604800; // ~7 days
 
 /// Store two version of the MPC contract for migration and backward compatibility purposes.
 /// Note: Probably, you don't need to change this struct.
 #[near_bindgen]
 #[derive(BorshDeserialize, BorshSerialize, Debug)]
 pub enum VersionedMpcContract {
-    /// Previous breaking changes version. We call `migration()` on it to transit into v1
+    /// This is no longer deployed
     V0(MpcContractV0),
```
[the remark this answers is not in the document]
pbeza: Good point. [Fixed](https://github.com/near/mpc/pull/410/commits/7e4d4bf05760ca13dc117b59cae8f08a479dd30d). ✅

libs/chain-signatures/contract/src/tee/quote.rs
```diff
@@ -0,0 +1,377 @@
...
+    // compose hash match expected
+    require!(replayed_compose_hash == expected_compose_hash);
+    // event with compose hash matches report rtmr3
+    require!(replayed_rtmr3 == rtmr3);
+
+    let (_, right) = app_compose.split_once("\\n        image:").unwrap();
+    let (left, _) = right.split_once("\\n").unwrap();
+    let (_, codehash) = left.split_once("@sha256:").unwrap();
+
+    codehash.to_owned()
```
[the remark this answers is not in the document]
pbeza: This code was [copied](https://github.com/search?q=serde_json%3A%3Afrom_str%28%26raw_tcb_info%29.expect%28%22TCB+Info+should+be+valid+JSON%22%29%3B&type=code) from the [Proximity](https://www.proximity.dev/) implementation (which, AFAIK, was originally taken from Phala). We took it as-is. The whole idea behind implementing the TEE stuff quickly was to reuse as much as possible from Proximity. We haven’t had time to fix their implementation yet, but we’ll likely iterate on it once @barakeinav1 reviews the verification logic.

libs/chain-signatures/contract/src/tee/quote.rs
```diff
@@ -0,0 +1,377 @@
...
+    encode(digest)
+}
+
+#[test]
+fn test() {
+    use dcap_qvl::verify;
+    use hex::decode;
+    use serde_json::json;
+
+    let tcb_info = json!(
```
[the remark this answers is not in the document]
pbeza: As explained [here](https://github.com/near/mpc/pull/410#discussion_r2142325849), it was [copied](https://github.com/search?q=%22rootfs_hash%22%3A+%22738ae348dbf674b3399300c0b9416c203e9b645c6ffee233035d09003cccad12f71becc805ad8d97575bc790c6819216%22%2C&type=code) from the Proximity implementation.

libs/chain-signatures/contract/Cargo.toml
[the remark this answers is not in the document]
pbeza: Nit: you should've used `#` instead of `//`.

libs/chain-signatures/contract/src/tee/quote.rs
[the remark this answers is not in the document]
netrome: Not really following the copied-as-is link here. Where's the source code we copied?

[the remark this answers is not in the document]
pbeza: https://github.com/NearDeFi/shade-agent-template/blob/6e6ff7abd577c06b35fab1a51864acd610a080ab/contract/src/collateral.rs#L109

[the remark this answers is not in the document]
pbeza: BTW, this bit of code is supposed to be abstracted and shared between the Proximity and N1 implementations. That was actually @bowenwang1996’s idea, which is why we ended up creating the shared [tee-verification](https://nearone.slack.com/archives/C08V34N1VLZ) Slack channel with them.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -275,39 +293,40 @@ impl MpcContract {
...
-                .push(code_hash.clone());
-            self.tee_state.allowed_code_hashes.insert(code_hash);
+        // let expected_rtmr3 = verification_result.report.as_td10().unwrap().rt_mr3;
+
+        // If the vote threshold is met and the new Docker hash is allowed by the TEE's RTMR3,
+        // update the state
+        if votes >= self.threshold()?.value()
+        // && self
+        //     .tee_state
+        //     .is_code_hash_allowed(code_hash, &expected_rtmr3, raw_tcb_info)
```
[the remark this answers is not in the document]
near-bookrock: Either way of a follow up. Can we delete the commented out code before merging?

libs/chain-signatures/contract/src/tee/quote.rs
```diff
@@ -0,0 +1,377 @@
...
+    fn get_str(v: &Value, key: &str) -> String {
+        v.get(key)
+            .and_then(Value::as_str)
+            .map(str::to_owned)
+            .unwrap_or_else(|| panic!("Missing or invalid field: {}", key))
+    }
+
+    fn get_hex(v: &Value, key: &str) -> Vec<u8> {
+        hex::decode(get_str(v, key)).unwrap_or_else(|_| panic!("Failed to decode hex: {}", key))
+    }
```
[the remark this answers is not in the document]
pbeza: [Copy-paste](https://github.com/near/mpc/pull/410#discussion_r2142325849) from Phala. :)

devnet/Cargo.lock
kevindeforth, line 1: This is probably a merge error, this file is no longer required.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -72,80 +81,79 @@ impl Default for VersionedMpcContract {
...
 
 #[near(serializers=[borsh])]
-#[derive(Debug, Clone)]
-pub struct AllowedCodeHash {
-    code_hash: CodeHash,
-    added: BlockHeight,
+#[derive(Debug)]
+pub struct TeeState {
+    allowed_tee_proposals: AllowedDockerImageHashes,
+    historical_tee_proposals: Vec<DockerImageHash>,
```
kevindeforth, line 87: nit: rename to `allowed_docker_image_hashes` and `historical_docker_image_hashes`

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -72,80 +81,79 @@ impl Default for VersionedMpcContract {
...
-            .position(|entry| entry.added + TEE_UPGRADE_PERIOD >= current_block_height)
-            .unwrap_or(self.allowed_code_hashes.len());
-
-        // Remove all expired entries
-        self.allowed_code_hashes.drain(0..expired_count);
-
-        // Return the number of removed entries
-        expired_count
+impl TeeState {
+    pub fn tee_status(&self, participants: Vec<AccountId>) -> BTreeMap<AccountId, TeeQuoteStatus> {
```
kevindeforth, line 104: nit: comment would be good
```suggestion
    /// maps every element in `participants` to its `TeeQuoteStatus`. If an element of `participants` does not have any TEE information associated to it, then it is mapped to `TeeQuoteStatus::None`.
    pub fn tee_status(&self, participants: Vec<AccountId>) -> BTreeMap<AccountId, TeeQuoteStatus> {
```

libs/chain-signatures/contract/src/tee/tee_participant.rs
```diff
@@ -0,0 +1,65 @@
...
+                "quote_collateral",
+                &str_preview(&self.quote_collateral, 128),
+            )
+            .field("raw_tcb_info", &str_preview(&self.raw_tcb_info, 128))
+            .finish()
+    }
+}
+
+impl TeeParticipantInfo {
+    pub fn verify_quote(&self, timestamp: u64) -> Result<VerifiedReport, Error> {
```
kevindeforth, line 54: can we add a comment about what is verified here? Also, please add a suffix to `timestamp` to indicate the unit (s for seconds, ms for milliseconds, ns for nanoseconds and so on)

libs/chain-signatures/contract/src/lib.rs
kevindeforth: just to bump this comment for visibility. I guess here, we are essentially checking if one of the certificates expired?

Shouldn't we additionally make sure that the docker image hash the participant is running is still valid and in the list of allowed hashes?

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -259,6 +263,19 @@ impl MpcContract {
...
+            .keys()
+            .cloned()
+            .collect();
+        let tee_verification = self.tee_state.tee_status(participant_ids);
+        log!(
+            "vote_new_parameters: prospective_epoch_id={}, proposal={:?}, tee_verification={:?}",
+            prospective_epoch_id,
+            proposal,
+            tee_verification,
+        );
```
kevindeforth, line 278: I don't think we need to log the vale of `tee_status` here.

libs/chain-signatures/contract/src/tee/tee_participant.rs
```diff
@@ -0,0 +1,65 @@
...
+        let tee_collateral = get_collateral(self.quote_collateral.clone())
+            .map_err(|_| Into::<Error>::into(InvalidCandidateSet::InvalidParticipantsTeeQuote))?;
+        let verification_result = verify::verify(&self.tee_quote, &tee_collateral, timestamp);
+        verification_result.map_err(|_| InvalidCandidateSet::InvalidParticipantsTeeQuote.into())
+    }
+
+    pub fn required_deposit(&self) -> NearToken {
+        let bytes_used = std::mem::size_of::<Self>() as u128; // TODO is it a correct estimate?
+        env::storage_byte_cost().saturating_mul(bytes_used)
+    }
```
kevindeforth, line 64: hmm, so probably this is not exact, since on the near blockchain, we store the borsh serialized value, not the raw bytes.

Instead, we could probably use `env::storage_used()` before and after storing the participant info, e.g. something like:
```Rust
    let initial_storage = env::storage_usage();
    // store participant info
    let storage_used = env::storage_usage() - initial_storage;
    let cost = env::storage_byte_cost() * storage_used as u128;
```
https://docs.near.org/smart-contracts/anatomy/environment#storage-used

kevindeforth (commented):
This looks pretty good!
Just a question about quote verification and the storage deposit.

libs/chain-signatures/contract/src/lib.rs
pbeza: [Fixed](https://github.com/near/mpc/pull/410/commits/e1f2582e2396c210b466de8e05e61518ceca105d). ✅

libs/chain-signatures/contract/src/lib.rs
pbeza: [Fixed](https://github.com/near/mpc/pull/410/commits/ffd7c9377221da92b021aebe39e9c195c2ba9ed3). ✅

libs/chain-signatures/contract/src/lib.rs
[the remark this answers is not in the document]
pbeza: [Fixed](https://github.com/near/mpc/pull/410/commits/821a0f3b36ceed73ea38b6071be4a7ca7eb1d272). ✅

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -72,80 +82,79 @@ impl Default for VersionedMpcContract {
...
+        //     .iter()
+        //     .chain(
+        //         self.allowed_tee_proposals
+        //             .get(env::block_height())
+        //             .iter()
+        //             .map(|entry| &entry.proposal),
+        //     )
+        //     .any(|proposal| proposal.tee_quote.get_rtmr3().unwrap() == *expected_rtmr3)
+        // TODO TODO TODO should be:
+        // .any(|proposal| hash(proposal.tee_quote.get_rtmr3().unwrap() || code_hash) == *expected_rtmr3)
```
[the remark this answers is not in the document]
pbeza: @barakeinav1 is reviewing it since that logic probably needs to be reimplemented. Removing it for now — [fixed](https://github.com/near/mpc/pull/410/commits/821a0f3b36ceed73ea38b6071be4a7ca7eb1d272). ✅

libs/chain-signatures/contract/src/tee/tee_participant.rs
pbeza: I only [store](https://github.com/near/mpc/blob/e1ef7be7ce6a13c26d054805648f87cd000a87db/libs/chain-signatures/contract/src/lib.rs#L675) participant info in the contract storage after [calculating](https://github.com/near/mpc/blob/e1ef7be7ce6a13c26d054805648f87cd000a87db/libs/chain-signatures/contract/src/lib.rs#L625) `required_deposit()`, so how can I compute it in that case? 🤔

libs/chain-signatures/contract/src/tee/proposal.rs
```diff
@@ -0,0 +1,200 @@
...
+        };
+
+        // Find the correct position to maintain sorted order by `added`
+        let insert_index = self
+            .allowed_tee_proposals
+            .iter()
+            .position(|entry| new_entry.added <= entry.added)
+            .unwrap_or(self.allowed_tee_proposals.len());
+
+        // Insert at the correct position
```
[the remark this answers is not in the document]
pbeza: [Fixed](https://github.com/near/mpc/pull/410/commits/0737d3416a3f4835d95cad14edaca1eaf13f691d). ✅

libs/chain-signatures/contract/src/tee/proposal.rs
```diff
@@ -0,0 +1,200 @@
...
+        // Remove the old entry if it exists
+        if let Some(pos) = self
+            .allowed_tee_proposals
+            .iter()
+            .position(|entry| entry.image_hash == code_hash)
+        {
+            self.allowed_tee_proposals.remove(pos);
+        }
+
+        // Create the new entry
```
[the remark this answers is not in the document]
pbeza: [Fixed](https://github.com/near/mpc/pull/410/commits/0737d3416a3f4835d95cad14edaca1eaf13f691d). ✅

libs/chain-signatures/contract/src/tee/proposal.rs
```diff
@@ -0,0 +1,200 @@
...
+        self.allowed_tee_proposals.drain(0..expired_count);
+
+        expired_count
+    }
+
+    /// Inserts a new code hash into the list after cleaning expired entries. Maintains the sorted
+    /// order by `added` (ascending). Returns `true` if the insertion was successful, `false` if the
+    /// code hash already exists.
+    pub fn insert(&mut self, code_hash: DockerImageHash, current_block_height: u64) -> bool {
+        // Clean expired entries
```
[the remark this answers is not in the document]
pbeza: [Fixed](https://github.com/near/mpc/pull/410/commits/0737d3416a3f4835d95cad14edaca1eaf13f691d). ✅

libs/chain-signatures/contract/src/lib.rs
pbeza: Probably an outdated comment—not sure which lines you're referring to, but I think it's already fixed [here](https://github.com/near/mpc/pull/410/commits/95e823d7bc6414e02ed84189adf7e264d4ac76e6) (?). Feel free to reopen if not.

libs/chain-signatures/contract/src/primitives/participants.rs
pbeza: That’s something we can address after merging this already huge PR.

libs/chain-signatures/contract/src/primitives/tee/proposal.rs
pbeza: Let’s move the RTMR/verification discussion to a separate PR after we merge this one. Reviewing the RTMR/verification logic is our next big task anyway.

libs/chain-signatures/contract/src/lib.rs
netrome: @pbeza here's the thread.

netrome:
> @kevindeforth, not sure how you added this code review comment — I can't follow up in the same thread, so I'm copy-pasting your comment here:

Yeah this happens when a separate review comments on an existing thread. In this PR that's pretty difficult to navigate. Pinged you in the thread. Otherwise you can manually navigate to the comment in the diff view.

image

<rant>Github UX as it's finest...🫠</rant>

libs/chain-signatures/contract/src/tee/tee_participant.rs
kevindeforth: > I only [store](https://github.com/near/mpc/blob/e1ef7be7ce6a13c26d054805648f87cd000a87db/libs/chain-signatures/contract/src/lib.rs#L675) participant info in the contract storage after [calculating](https://github.com/near/mpc/blob/e1ef7be7ce6a13c26d054805648f87cd000a87db/libs/chain-signatures/contract/src/lib.rs#L625) required_deposit(), so how can I compute it in that case? 🤔

So if we store the info and then realize the deposit does not suffice, we can just return an error and all state changes should get rolled back.

But I am not sure how big the actual difference between borsh serialized state size and rust struct size is, so for me, it is also ok to merge as is (but we should track this in a separate issue).

PR #411, opened 2025-05-13
workflow for contract release

kevindeforth:


near-bookrock (commented):
Could be good to enforce tests to pass with this binary, as a validation criteria in the workflow for the release to pass.

.github/workflows/contract_release.yml
```diff
@@ -0,0 +1,89 @@
...
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git tag "${tag}"
+          git push origin "${tag}"
+
+      - name: Create GitHub Release
+        uses: softprops/action-gh-release@v2
+        with:
+          tag_name: ${{ steps.compute_tag.outputs.tag }}
+          name: Contract Release v${{ steps.compute_tag.outputs.tag }}
```
near-bookrock, line 86: What do you think about auto generating release notes based on the commit history since the last release?

```suggestion
          name: Contract Release v${{ steps.compute_tag.outputs.tag }}
          generate_release_notes: true
```

andrei-near, line 86: wouldn't this overwrite latest mpc code contract? Given there are 2 different products in this repo, maybe it would make sense to move contract code into another repo?

andrei-near: another option would be to just set tags without the release

near-bookrock (approved):

.github/workflows/contract_release.yml
kevindeforth: I like the idea, but doing this correctly might require [some patience](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes).
I will open an issue.

kevindeforth: Yeah, we should probably separate this into its own repo. A prefix will just be confusing.

kevindeforth: Is this not just the output of some LLM, or some commits chained in a row? We have no idea of the quality of those notes, so I'd rather have none of them now and test to see if it is any good.

PR #414, opened 2025-05-13
Add migration logic 

kuksag:
stacked on: https://github.com/Near-One/mpc/pull/408

Current testnet logic runs on this binary https://github.com/Near-One/mpc/releases/tag/2.0.0-rc.1, which runs on 7c277d73b9f69679c8f2b0a58bed51617a333e3c
It can not migrate to the current version, because migration logic has to be updated. This PR addresses this problem. 

Migration from 7c277d73b9f69679c8f2b0a58bed51617a333e3c to the latest commit was succeffully tested.

### Overview:
There are new semantics around V0 and V1 and. 
TLDR: Store only migration for the last incompatible change. 

### Rationale: 
Other approach would be to store each state there was for each breaking change + corresponding migration logic. 
That results in lots of dead code. I propose storing only migration from the last breaking change. If for whatever reason we have to migrate from an older version, we checkout some specific commit in history for "older" -> "recent breaking change" migration, and do "recent" -> "current breaking change"  migration


kuksag:
@kevindeforth `test_update_v2_running` is failing now, because of my new semantics. Please let me know how do you find my proposal.

If we agree on that, then this test seems obsolete due to new back compatibility unit test.
Otherwise, they rightfully co-exists with each other.

libs/chain-signatures/contract/src/v0_state/mod.rs
```diff
@@ -62,3 +53,19 @@ pub struct MpcContractV0 {
     pub proposed_updates: ProposedUpdates,
     pub config: Config,
 }
+
+impl From<MpcContractV0> for MpcContract {
+    fn from(value: MpcContractV0) -> Self {
+        Self {
+            protocol_state: value.protocol_state,
```
kevindeforth, line 60: I don't think this will work. We changed the struct `ThresholdParametersVotes` in PR #372.

previously (what is released on testnet):
```Rust
pub struct ThresholdParametersVotes {
    proposal_by_account: BTreeMap<AuthenticatedParticipantId, ThresholdParameters>,
}
```

now:
```Rust
pub struct ThresholdParametersVotes {
    proposal_by_account: BTreeMap<AuthenticatedAccountId, ThresholdParameters>,
}
```

kevindeforth (commented):
Thank you for fixing the migration of the `Update` struct. 
I would expect the failing test to pass after migration of `ThresholdParametersVotes` has been re-enabled.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -154,22 +158,6 @@ pub struct MpcContract {
     tee_state: TeeState,
 }
 
-impl From<v0_state::MpcContractV0> for MpcContract {
```
kevindeforth, line 157: (this was the logic migrating the `ThresholdParametersVotes` struct)

kuksag:
Let me supply the "proper" v2.0.0, I want to check how the tests are going to behave, to verify a hypothesis of mine (I'm too lazy to compile neard and other stuff locally for pytests). Will revisit later.

kuksag:
> I would expect the failing test to pass after migration of ThresholdParametersVotes has been re-enabled

Yes, you are totally right Kevin 👍 . I was confused why the unit test that I wrote was succeeding... TLDR some non-trivial state has to be supplied. I loaded off some thought in https://github.com/Near-One/mpc/issues/415

kevindeforth (approved):
Thank You very much!

PR #421, opened 2025-05-19
rust compiles contract for testing

kevindeforth:


near-bookrock (approved):

libs/chain-signatures/contract/tests/common.rs
```diff
@@ -77,8 +78,31 @@ pub async fn gen_accounts(worker: &Worker<Sandbox>, amount: usize) -> (Vec<Accou
     let candidates = candidates(Some(accounts.iter().map(|a| a.id().clone()).collect()));
     (accounts, candidates)
 }
-
+pub static INIT: Lazy<()> = Lazy::new(build_contract);
+fn build_contract() {
```
near-bookrock, line 82: Shouldn't this be called to build the contract?

pbeza (approved):

PR #423, opened 2025-05-19
test: remove unused mutations

near-bookrock:
The removed mutations are clones of the participant set in the integration tests. Mutating the clone does nothing for the test.

### DIFF
--- a/node/src/tests/resharing.rs
+++ b/node/src/tests/resharing.rs
@@ -109,12 +109,6 @@ async fn test_key_resharing_multistage() {
         PortSeed::KEY_RESHARING_MULTISTAGE_TEST,
     );
 
-    // Initialize the contract with two fewer participants.
-    let mut participants_1 = setup.participants.clone();
-    participants_1.participants.pop();
-    participants_1.participants.pop();
-    participants_1.threshold = 3;
-
     let domain = DomainConfig {
         id: DomainId(0),
         scheme: SignatureScheme::Secp256k1,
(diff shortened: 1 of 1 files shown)


node/src/tests/resharing.rs
```diff
@@ -109,12 +109,6 @@ async fn test_key_resharing_multistage() {
         PortSeed::KEY_RESHARING_MULTISTAGE_TEST,
     );
 
-    // Initialize the contract with two fewer participants.
```
kevindeforth, line 112: Good catch! Could you adjust the comments of the test as well? There is a discrepancy between what they say and what the test actually does ("`// Test two nodes joining and two old nodes leaving.`")

kevindeforth (approved):

node/src/tests/resharing.rs
near-bookrock: Does it now look better to you?

node/src/tests/resharing.rs
```diff
@@ -89,7 +89,14 @@ async fn test_key_resharing_simple() {
     .is_some());
 }
 
-// Test two nodes joining and two old nodes leaving.
+// Tests that key resharing works, when new nodes join, and the nodes
+// from the original set leave the network.
+//
+// Test scenario:
+// 1. Setup network with 4 nodes.
```
kevindeforth, line 96: Don't we start with 6 nodes?

node/src/tests/resharing.rs
```diff
@@ -145,7 +146,6 @@ async fn test_key_resharing_multistage() {
     // Have the fifth node join.
```
kevindeforth, line 146: actually, a node is leaving here, no?

node/src/tests/resharing.rs
kevindeforth: I might be confused, but it looks to me like it still doesn't match. I commented on some of them, but there are more.

PR #454, opened 2025-05-29
List Mpc and Loadtest setups in devnet

kevindeforth:
Helper command to list MPC and Loadtest setups.
Separating to keep PRs small.

### DIFF
--- a/devnet/src/types.rs
+++ b/devnet/src/types.rs
@@ -3,6 +3,7 @@ use near_crypto::SecretKey;
 use near_sdk::AccountId;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
+use std::fmt;
 use std::path::PathBuf;
 use std::sync::Arc;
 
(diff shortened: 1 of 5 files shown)


devnet/src/types.rs
```diff
@@ -60,6 +61,44 @@ pub struct MpcNetworkSetup {
...
+            "  Number of Responding Access Keys: {}",
+            self.num_responding_access_keys
+        )?;
+        writeln!(
+            f,
+            "  Desired Balance per Responding Account: {}",
+            self.desired_balance_per_responding_account
+        )?;
+
+        //writeln!(f, " Using SSD: {}", self.ssd)?;
```
near-bookrock, line 94: Is this a leftover?
```suggestion
```

devnet/src/types.rs
```diff
@@ -60,6 +61,44 @@ pub struct MpcNetworkSetup {
...
+        writeln!(f, "  Participants:")?;
+        for (i, participant) in self.participants.iter().enumerate() {
+            writeln!(f, "    {}: {}", i + 1, participant)?;
+        }
+
+        if let Some(contract) = &self.contract {
+            writeln!(f, "  Contract: {}", contract)?;
+        } else {
+            writeln!(f, "  Contract: None")?;
+        }
```
near-bookrock, line 76: Nit:
You can rewrite this as

```rust
let contract_formatted = self.contract.map(ToString::to_string).unwrap_or_else(|| "None".to_string());
writeln!(f, "  Contract: {}", contract_formatted)?;
```

devnet/src/types.rs
```diff
@@ -60,6 +61,44 @@ pub struct MpcNetworkSetup {
     pub nomad_server_url: Option<String>,
 }
 
+impl fmt::Display for MpcNetworkSetup {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
near-bookrock, line 65: For consistency since you have `use std::fmt`.
```suggestion
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> fmt::Result {
```

devnet/src/types.rs
```diff
@@ -95,7 +160,7 @@ pub struct ParsedConfig {
 
 pub async fn load_config() -> ParsedConfig {
     const CONFIG_FILE: &str = "config.yaml";
-    let config = std::fs::read_to_string(CONFIG_FILE).unwrap();
+    let config = std::fs::read_to_string(CONFIG_FILE).expect("Did not find 'config.yaml'");
```
near-bookrock, line 163: Thanks a lot for adding `expect` and an explanation to it.

Nit:
expect statements should contain the reason you _expect_ the Result to be Ok, not why it became an error.
I.e. something like 
```suggestion
    let config = std::fs::read_to_string(CONFIG_FILE).expect("A `config.yaml` should exist in the working directory.");
```

Here's a nice link from the std docs for `expect()`:
https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style

near-bookrock (commented):
I haven't run your changes, so I don't know how the output looks like. I might be wrong here, but is the formatted output you are creating not very similar to a pretty print?

See example here 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=2e97b0471c4d16ea84a8a3d6ab0ae4b4

devnet/src/types.rs
kevindeforth: This will actually be needed in a subsequent PR.

near-bookrock (approved):

PR #457, opened 2025-06-03
update comments

kevindeforth:
follow up to #438

### DIFF
--- a/pytest/tests/test_signature_request_during_resharing.py
+++ b/pytest/tests/test_signature_request_during_resharing.py
@@ -12,15 +12,17 @@ def test_single_domain():
     Tests that signature requests are still processed while performing key resharing.
 
     Test scenario:
-        1. Start with 2 nodes and 1 domain.
+        1. Start with 2 nodes and 1 domain for Ecdsa as well as Eddsa.
         2. Add 2 more nodes and start a key resharing.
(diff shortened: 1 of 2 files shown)


pytest/tests/test_signature_request_during_resharing.py
```diff
@@ -12,15 +12,17 @@ def test_single_domain():
...
     """
     # Have the nodes disabled
-    cluster, mpc_nodes = shared.start_cluster_with_mpc(
-        2, 4, 1, load_mpc_contract(), start_mpc_nodes=True
-    )
+    cluster, mpc_nodes = shared.start_cluster_with_mpc(2,
+                                                       4,
+                                                       1,
+                                                       load_mpc_contract(),
+                                                       start_mpc_nodes=True)
```
near-bookrock, line 25: What formatter are you using for python?
I have Black installed, which is the format being removed here.

My formatter is also going haywire with its formatting in other since the style we have is not consistent with it. We should allign the formatter we use in the project, such that we don't re-format every time we work touch a new file.

near-bookrock: Related to this https://nearone.slack.com/archives/D08FW44AN59/p1748870096099459

PR #460, opened 2025-06-03
[DRAFT] Launcher and reproducible build.

thomasknauth:
Launcher script and artifacts to enable reproducible builds for the MPC binary and Docker image.

### DIFF
--- a/launcher.py
+++ b/launcher.py
@@ -28,7 +28,7 @@ def parse_env_file(path):
     return env
 
 def main():
-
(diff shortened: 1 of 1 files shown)


launcher.py
netrome, line 1: One thing I'm not seeing in this file is any interaction with the signer contract. From the system diagram it seems like it should get a whitelist from an onchain registry. Am I misinterpreting or missing anything here?

image

PR #468, opened 2025-06-04
feat(Tee): Automatic kickout mechanism for invalid TEE status

### LINKED ISSUE #424: Enforce timely node upgrade after voting for a new image digest

kevindeforth:
Resolves #424:

This PR introduces a new unprotected endpoint: `verify_tee`.
The purpose of this endpoint is to assess the current TEE status of all participants and take corrective action if necessary:
- If `threshold` or more participants have a `Valid` or `None` TEE status:
    - A re-sharing is triggered, excluding participants with an `Invalid` status, preserving the current threshold.
- If fewer than `threshold` participants have a `Valid` or `None` TEE status:
    - The contract stops accepting new signature requests until enough participants update their status and `verify_tee` is called again.

Note that this endpoint can only be called while the protocol is in running state.

(not sure about where to place the background task in the node @near-bookrock any suggestions?)

Requires additional testing https://github.com/near/mpc/issues/514

### DIFF
--- a/libs/chain-signatures/contract/src/lib.rs
+++ b/libs/chain-signatures/contract/src/lib.rs
@@ -24,7 +24,7 @@ use crypto_shared::{
 };
 use dcap_qvl::verify;
 use errors::{
-    DomainError, InvalidParameters, InvalidState, PublicKeyError, RespondError, SignError,
+    DomainError, InvalidParameters, InvalidState, PublicKeyError, RespondError, SignError, TeeError,
 };
 use k256::elliptic_curve::{sec1::ToEncodedPoint, PrimeField};
 use near_sdk::{
@@ -35,6 +35,7 @@ use near_sdk::{
     AccountId, CryptoHash, CurveType, Gas, GasWeight, NearToken, Promise, PromiseError,
     PromiseOrValue, PublicKey,
 };
+use primitives::participants::Participants;
 use primitives::{
     domain::{DomainConfig, DomainId, DomainRegistry, SignatureScheme},
     key_state::{AuthenticatedParticipantId, EpochId, KeyEventId, Keyset},
@@ -42,7 +43,6 @@ use primitives::{
     thresholds::{Threshold, ThresholdParameters},
 };
 use state::{running::RunningContractState, ProtocolContractState};
-use std::collections::BTreeMap;
 use storage_keys::StorageKey;
 use tee::{
     proposal::{CodeHashesVotes, DockerImageHash},
@@ -56,7 +56,6 @@ const GAS_FOR_SIGN_CALL: Gas = Gas::from_tgas(10);
 // Register used to receive data id from `promise_await_data`.
 const DATA_ID_REGISTER: u64 = 0;
 
-// Prepaid gas for a `return_signature_and_clean_state_on_success` call
 const RETURN_SIGNATURE_AND_CLEAN_STATE_ON_SUCCESS_CALL_GAS: Gas = Gas::from_tgas(6);
 // Prepaid gas for a `update_config` call
 const UPDATE_CONFIG_GAS: Gas = Gas::from_tgas(5);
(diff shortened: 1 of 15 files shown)


libs/chain-signatures/contract/src/lib.rs
```diff
@@ -918,6 +957,68 @@ impl VersionedMpcContract {
...
+                        //let n_participants_new = new_participants.len();
+                        //let new_threshold = (3 * n_participants_new + 4) / 5; // minimum 60%
+                        //let new_threshold = new_threshold.max(2); // but also minimum 2
+                        let new_threshold = threshold;
+
+                        let tp = ThresholdParameters::new(
+                            new_participants,
+                            Threshold::new(new_threshold as u64),
+                        )
+                        .expect("error");
```
near-bookrock, line 1008: What's the expectation here?

node/src/mpc_client.rs
```diff
@@ -92,6 +93,21 @@ impl MpcClient {
             )
         };
 
+        let tee_verification = {
+            let chain_txn_sender = chain_txn_sender.clone();
+            tracking::spawn("tee_verification", async move {
```
near-bookrock, line 98: Can you use tokio directly here? I want us to deprecate the `tracking` module.

node/src/mpc_client.rs
```diff
@@ -92,6 +93,21 @@ impl MpcClient {
             )
         };
 
+        let tee_verification = {
+            let chain_txn_sender = chain_txn_sender.clone();
```
near-bookrock, line 97: Nice shadowing

node/src/mpc_client.rs
```diff
@@ -92,6 +93,21 @@ impl MpcClient {
             )
         };
 
+        let tee_verification = {
```
near-bookrock, line 96:
```suggestion
        let tee_verification_handle = {
```

node/src/mpc_client.rs
```diff
@@ -111,6 +127,7 @@ impl MpcClient {
         monitor_chain.await?;
         let _ = ecdsa_background_tasks.await?;
         let _ = eddsa_background_tasks.await?;
+        tee_verification.await?;
```
near-bookrock, line 130: Awaiting on this join_handle will never complete. The spawned task has a loop with no exits or cancellations.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -918,6 +957,68 @@ impl VersionedMpcContract {
...
+                        contract.accept_signature_requests = true;
+                        log!("All participants have an accepted Tee status");
+                        Ok(true)
+                    }
+                    TeeValidationResult::Partial(new_participants) => {
+                        let threshold = current_params.threshold().value() as usize;
+                        let remaining = new_participants.len();
+                        if threshold > remaining {
+                            log!("Less than `threshold` participants are left with a valid TEE status. This requires manual intervention. We will not accept new signature requests as a safety precaution.");
+                            contract.accept_signature_requests = false;
```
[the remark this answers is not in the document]
kevindeforth: No, why would it be false?

node/src/mpc_client.rs
kevindeforth: In this case, I prefer to stay consistent with the existing code. using tokio directly would look out of place in this function.

node/src/mpc_client.rs
kevindeforth: Good point. I don't think that is good. Shot you a dm.

node/src/mpc_client.rs
near-bookrock: It's not just about looking in place to stay consistent unfortunately.

Any downstream calls you do or introduce that also create depencies to the `tracking` module will make it impossible to remove `tracking` for `tokio` as the `tracking` call can only be done within a `tracking` spawned task.

Given that we have allignment to remove this module, let's create more code that has to be migrated in the close future.

node/src/metrics.rs
```diff
@@ -205,3 +205,12 @@ lazy_static! {
         )
         .unwrap();
 }
+
+lazy_static! {
```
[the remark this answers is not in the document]
kevindeforth: > Let me know if for any reason we need lazy_static in particular.
  
 I wouldn't know about any, but this is very Rust-specific knowledge, so maybe @near-bookrock knows more.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -100,25 +100,51 @@ impl Default for TeeState {
...
+
 impl TeeState {
-    /// maps every element in `participants` to its `TeeQuoteStatus`. If an element of
-    /// `participants` does not have any TEE information associated to it, then it is mapped to
+    /// Performs TEE validation on the given participants.
+    ///
+    /// Returns `TeeValidationResult::Full` if all participants are valid,
+    /// or `TeeValidationResult::Partial` with the subset of valid participants otherwise.
+    ///
+    /// Participants with `TeeQuoteStatus::Valid` or `TeeQuoteStatus::None` are considered valid.
```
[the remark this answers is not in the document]
kevindeforth: yes

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -918,6 +956,65 @@ impl VersionedMpcContract {
...
+                }
+
+                // here, we set it to true, because at this point, we have at least `threshold`
+                // number of participants with an accepted Tee status.
+                contract.accept_signature_requests = true;
+
+                // do we want to adjust the threshold?
+                //let n_participants_new = new_participants.len();
+                //let new_threshold = (3 * n_participants_new + 4) / 5; // minimum 60%
+                //let new_threshold = new_threshold.max(2); // but also minimum 2
```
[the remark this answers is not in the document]
kevindeforth: I asked this [in slack](https://nearone.slack.com/archives/C07UW93JVQ8/p1750140170506589), we do not want to adjust the threshold.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -918,6 +956,65 @@ impl VersionedMpcContract {
...
+                let new_threshold = threshold;
+
+                let threshold_parameters = ThresholdParameters::new(
+                    new_participants,
+                    Threshold::new(new_threshold as u64),
+                )
+                .expect("Require valid threshold parameters"); // this should never happen.
+                current_params.validate_incoming_proposal(&threshold_parameters)?;
+                let res = running_state.transition_to_resharing_no_checks(&threshold_parameters);
```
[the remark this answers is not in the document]
kevindeforth: yes

PR #481, opened 2025-06-10
test: test that threshold from previous running state is used when serving sign requests in resharing

near-bookrock:
This PR adds a regression test to the issue that was solved in: https://github.com/near/mpc/pull/480

The test fails on `main` previous to the merged fix.


kevindeforth (commented):
These pytests take quite a long time to run.
I would prefer to merge `test_single_domain` with `test_threshold_from_previous_running_state_is_maintained`, since the latter also covers the former.

If we want to test this behavior in isolation, we might be advised to implement a Rust integration test instead, since they are significantly faster.

kevindeforth (approved):

PR #490, opened 2025-06-11
chore: bump versions for new release candidate

ghost:


pbeza (approved):

libs/chain-signatures/contract/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mpc-contract"
-version = "2.1.2-rc.1"
+version = "2.2.0-rc.1"
```
netrome, line 3: Given that this code passed release testing, do we need the `rc.1` suffix?

netrome:
> Sorry missed this comment while it was in the merge queue. I don't know what the best convention is, but I think it's fine we call it RC when we release to testnet as it will be running there for a while before we go ahead and release to mainnet.

No worries, I think it's more to align on conventions for the future.

I'm a bit confused about naming it RC for testnet. Do you mean to follow up with a subsequent release to testnet before making the non-RC release to mainnet? Or having a different release on testnet and mainnet?

I don't see any problem with having different releases on testnet and mainnet. On my previous project we typically had a two week window between releasing a release on testnet and on mainnet - which I think is a sensible process. I don't see what value we gain from the RC tag in this context.

PR #493, opened 2025-06-12
chore: Update devnet docs

netrome:
This PR contains three small updates to improve the experience using our devnet CLI:

1. I've added installation instructions for Terraform and Gcloud, which are needed to run the devnet CLI.
2. I included the devnet crate in the root workspace and renames the binary to `mpc-devnet`.
3. I added installation instructions for the CLI and changed the examples to use the `mpc-devnet` command directly instead of the verbose `target/debug/devnet` command.


near-bookrock:
Thanks for adding it to the workspace. Can you delete the old lock file as it is no longer used? https://github.com/near/mpc/blob/chore/devnet_experiment/devnet/Cargo.lock

near-bookrock (approved):

netrome:
> Thanks for adding it to the workspace. Can you delete the old lock file as it is no longer used? https://github.com/near/mpc/blob/chore/devnet_experiment/devnet/Cargo.lock

Good point, will do

PR #495, opened 2025-06-12
chore(indexer): remove duplicate watcher in `monitor_contract_state`

ghost:
We have two "duplicate" watcher senders in `monitor_contract_state` that both send the latest state of the contract over the watcher. The minor difference between them is that one sends the type used in the contract code, `ProtocolContractState`, which is (used in our debug endpoint), whereas the other sends the parsed state used in the node code, `ContractState`.

As agreed in a code walk through a few months ago we concluded that this duplicate watcher could be deleted, and have the debug endpoint use a cloned receiver of the other channel.

### DIFF
--- a/node/src/coordinator.rs
+++ b/node/src/coordinator.rs
@@ -22,6 +22,7 @@ use crate::runtime::AsyncDroppableRuntime;
 use crate::sign_request::SignRequestStorage;
 use crate::tracking::{self};
 use crate::web::SignatureDebugRequest;
+use anyhow::Context;
 use cait_sith::{ecdsa, eddsa};
 use futures::future::BoxFuture;
 use futures::FutureExt;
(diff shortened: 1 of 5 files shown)


kevindeforth (commented):
> As agreed in a code walk through a few months ago we concluded that this duplicate watcher could be deleted, and have the debug endpoint use a cloned receiver of the other channel.

So if my memory serves me right, I thought we agreed to keep the channel that sends the contract state as it is stored on chain and get rid of the channel that sends the internal representation.

That is for two reasons:
1. the debug endpoint is more useful if it displays the current contract state as received by the node from the indexer (because if it does not match expectations, we can't tell if there is an error in the parsing or conversion, or if the indexer is lagging).
2. It should not be the indexers job to convert the contract to the internal represntation.

netrome:
Closing this as stale.

Issue #439, opened 2025-05-26
Signature requests in flight are lost when state transitions away from `Running`


ghost:
## Problem
The MPC network is currently losing signature requests if the signature requests are being processed while the node is transitioning from `Running` to any other state.

## Root Cause
The indexer sends [`ChainBlockUpdate`](https://github.com/Near-One/mpc/blob/9d41a03566c3583b97b88d76fb9e0c8ba144cac0/node/src/indexer/handler.rs#L41-L46)s over an unbounded channel to the mpc node. Incoming `sign` requests to the contract are contained in the `ChainBlockUpdate`s that the indexer sends.

In [`monitor_block_updates`](https://github.com/Near-One/mpc/blob/946a39c3de16fef004d7082f1fdf4cf3230def6c/node/src/mpc_client.rs#L118) we consume these block updates in a task that runs as long as we are in the running state. The message is consumed in this select loop;
https://github.com/Near-One/mpc/blob/946a39c3de16fef004d7082f1fdf4cf3230def6c/node/src/mpc_client.rs#L139

Once the message is consumed from the channel, we don't persist the signature request. That means, if the task running `monitor_block_updates` or the spawned tasks that handle the individual signature requests are dropped, we lose the signature request.

The problem is that when the contract state transitions, we tear down the entire tokio runtime in `coordinator.rs`, and create a new one each time we go into a new state. 

https://github.com/Near-One/mpc/blob/ef3f1e7f862d447de60e91d32dadf68696eb6a58/node/src/coordinator.rs#L248

Dropping the runtime drops all spawned tasks, meaning we lose `sign` requests that were in progress.

netrome:
Still relevant, not immediate priority

netrome:
Still relevant, still annoying, still a good duck.

Issue #409, opened 2025-05-13
`test_basic_multidomain` fails on macOS.


near-bookrock:
When re-entering running state in `test_basic_multidomain` all connections are dropped and attempted to be re-established. The re-establishing of these connections fail when running the integration test on macOS.

netrome:
Tests pass for me now, still take wayyyyy longer than they should and require running in release mode. Let's close this issue for now though.

PR #498, opened 2025-06-13
fix(node): add timeout to avoid race condition in integration test

kevindeforth:
In #394 , we introduced the capability to handle signature requests while being in resharing state.
Unfortunately, this lead to a race condition, resulting in at least one flaky test (c.f. https://github.com/near/mpc/issues/439).

This PR adds a timeout to avoid the race condition.
In addition, it removes a log statement that was a bit excessive.

See these logs as an example of a failed run:

https://github.com/near/mpc/actions/runs/15635164527/job/44048349639

[log file](https://github.com/user-attachments/files/20727510/logs.log)

### DIFF
--- a/node/src/network/conn.rs
+++ b/node/src/network/conn.rs
@@ -172,7 +172,6 @@ impl<I: Send + Sync + 'static, O: Send + Sync + 'static> NodeConnectivity<I, O> 
             anyhow::bail!("Connection was dropped");
         };
 
-        info!("Got a connection handle.");
         Ok(conn)
     }
 }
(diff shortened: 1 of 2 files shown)


netrome:
`test_basic_multidomain` still fails for me so I assume this is orthogonal to https://github.com/near/mpc/issues/409.

Also relying on `tokio::time::sleep` to synchronize between threads is naturally quite brittle and inherently flaky, so while it's an okay band-aid fix for now, I'd like us to be cautious not to rely on this going forward

near-bookrock:
> `test_basic_multidomain` still fails for me so I assume this is orthogonal to #409.
> 
> Also relying on `tokio::time::sleep` to synchronize between threads is naturally quite brittle and inherently flaky, so while it's an okay band-aid fix for now, I'd like us to be cautious not to rely on this going forward

It's orthogonal to this issue. It was failing on `main` for me as well previous to this feature being merged.

near-bookrock (approved):

PR #504, opened 2025-06-16
chore: Add chain-signatures crates to root workspace

netrome:
As I'm exploring the code, I took the liberty to add these to the root workspace as I found that this could be trivially done.


kevindeforth:
Does it make sense to do this now, given that we decided to move the contract (and probably the entire `chain-signatures` folder) into its own repository?

Issue #479, opened 2025-06-08
Generate respond keys inside node


kuksag:
#397 follow-up
should be done after after #476 so we can actually test it

### Overview 
MPC node stores two keys: 
(a) NEAR signer keys – perform protocol actions
(b) NEAR respond keys – respond with generated signatures 
for more details have a look into a knowledge base 😉 

We should generate X number of responding keys and share the public part. Later node operator will authorize those keys for responding accounts

Issue #397, opened 2025-05-08
Generate `sign_pk` and "account key" inside MPC node


thomasknauth:
Today, `sign_pk` is generated outside the MPC node and passed in through env variables. When moving MPC node into TEE, the `sign_pk` key needs to be generated inside the MPC node at first "boot" and saved on disk.

thomasknauth:
Also, the public key should be printed to stdout/logging such that the node operator can figure out what the key is.

thomasknauth:
And, I guess, we want to remove the possibility to set via env variable.

barakeinav1:
we also need to generate the near account key the is responsible for responding with the final generated signature

PR #508, opened 2025-06-16
Pytest refactoring 

kuksag:
Before moving further with https://github.com/near/mpc/issues/479 and closing https://github.com/near/mpc/issues/397 with proper testing, some logic has to be added to the pytests. 

As I had to come up with understanding what's happening in the tests, I found it easier to split some things up. Otherwise it's too complicated for me to grasp.  

After this PR, there will be follow-ups with near signer/responder account creation


kevindeforth (approved):
This looks like an improvement to me. It is a pure refactor, right?

PR #509, opened 2025-06-16
feat(contract): verification of TEE RTMRs 0-2 and MRTD

pbeza:


libs/chain-signatures/contract/src/tee/tee_participant.rs
```diff
@@ -5,6 +5,11 @@ use crate::{
 use dcap_qvl::verify::{self, VerifiedReport};
 use near_sdk::near;
 
+const RTMR0: [u8; 48] = [0u8; 48];
+const RTMR1: [u8; 48] = [0u8; 48];
+const RTMR2: [u8; 48] = [0u8; 48];
+const MRTD: [u8; 48] = [0u8; 48];
```
pbeza, line 11: @barakeinav1, you mentioned we can hardcode these values — what exactly should I put here?

barakeinav1: let me build dstack 0.52 on 2 separate machines to confirm.

libs/chain-signatures/contract/src/tee/tee_state.rs
```diff
@@ -0,0 +1,105 @@
+use crate::{
```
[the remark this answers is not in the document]
pbeza: Yup. Not sure why GitHub isn’t smart enough to show it as moved code.

kevindeforth (approved):
Thank You.

libs/chain-signatures/contract/src/tee/tee_participant.rs
netrome: I guess this must be added before we can merge right? Or will you do it as a follow-up?

libs/chain-signatures/contract/src/tee/tee_state.rs
```diff
@@ -0,0 +1,105 @@
...
+        self.historical_docker_image_hashes
+            .iter()
+            .chain(
+                self.allowed_docker_image_hashes
+                    .get(env::block_height())
+                    .iter()
+                    .map(|entry| &entry.image_hash),
+            )
+            .any(|proposal| proposal.as_hex() == code_hash)
+    }
```
netrome, line 96: Not really confident about the code, but this smells like a bug to me.

Why don't we use the `_code_hash` in this function? Shouldn't we verify the provided docker image hash?

I'm also confused why the `verify_codehash` function returns a code hash - if we're verifying stuff shouldn't it just return a bool or result? I suppose the historical docker image hashes should be compared with the provided docker image hash.

netrome (changes requested):
Mostly looks good to me. Two blockers from my side however:

1. We need to add the hardcoded values or create a follow up for this.
2. The `is_code_hash_allowed` function seems incorrect to me. Please let me know if I'm missing something here.

Finally, this PR has non-trivial conflicts with main since we merged https://github.com/near/mpc/pull/468 - please rebase or merge to resolve the conflict.

If my assumptions are incorrect, feel free to dismiss this review and go ahead and merge.

libs/chain-signatures/contract/src/tee/tee_state.rs
netrome: Ah I see, thanks for the pointer.

pbeza: > Why don't we use the _code_hash in this function? Shouldn't we verify the provided docker image hash?
>
> I'm also confused why the verify_codehash function returns a code hash - if we're verifying stuff shouldn't it just return a bool or result? I suppose the historical docker image hashes should be compared with the provided docker image hash.

@kevindeforth @netrome This isn’t our code — it’s from Phala or Proximity (not sure who the original author is). We just copied it from [here](https://github.com/NearDeFi/shade-agent-template/blob/6e6ff7abd577c06b35fab1a51864acd610a080ab/contract/src/collateral.rs#L34-L62) and haven’t had time to refactor it (though I did slightly refactor some of it in #516).

As a rule of thumb: whenever you see an `unwrap()` or some fishy code, it might have been copy-pasted from another project due to tight deadlines.

PR #511, opened 2025-06-16
feat(contract): add method to contract to get allowed image hashes

### LINKED ISSUE #510: [Task] smart contract must have a method that returns allowed image hashes
For TEE we have a set of image hashes that are allowed by the contract. The contract must expose an endpoint that returns these hashes so the node can fetch them.

near-bookrock:
Resolves #510


netrome (approved):
Approving despite the lack of test coverage due to our aggressive timeline. Can you create a follow-up issue to implement tests for this method so we keep track of this?

Issue #156, opened 2025-01-24
Use tokio for the indexer thread instead of std::thread


robin-near:
The top-level join for the std::thread blocks any errors that may be raised from the mpc task. So if the mpc task fails, the binary doesn't end up exiting. This is very awkward to deal with and it would be better if the indexer is created on a tokio runtime instead.

robin-near:
The immediate issue of joining was resolved in #196 . The remaining item would be to use tokio instead of actix for the indexer tasks.

PR #513, opened 2025-06-16
feat(indexer): Periodically fetch allowed image hashes from mpc contract

near-bookrock:
Resolves https://github.com/near/mpc/issues/459?issue=near%7Cmpc%7C512

Also adds a feature, `tee`, to feature gate tee-related code on the node.

### DIFF
--- a/libs/chain-signatures/contract/src/tee/proposal.rs
+++ b/libs/chain-signatures/contract/src/tee/proposal.rs
@@ -64,15 +64,15 @@ impl CodeHashesVotes {
 
 /// A proposal for a new TEE code hash to be added to the whitelist, along with the time it was
 /// added.
-#[near(serializers=[borsh])]
-#[derive(Debug, Clone)]
+#[near(serializers=[borsh, json])]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct AllowedDockerImageHash {
     pub image_hash: DockerImageHash,
     pub added: BlockHeight,
 }
 /// Collection of whitelisted Docker code hashes that are the only ones MPC nodes are allowed to
 /// run.
-#[near(serializers=[borsh])]
+#[near(serializers=[borsh, json])]
 #[derive(Debug, Default)]
 pub struct AllowedDockerImageHashes {
     /// Whitelisted code hashes, sorted by when they were added (oldest first). Expired entries are
--- a/node/src/indexer/lib.rs
+++ b/node/src/indexer/lib.rs
@@ -80,7 +80,7 @@ pub(crate) async fn get_mpc_contract_state(
 }
 
 #[cfg(feature = "tee")]
-pub(crate) async fn get_mpc_tee_state(
+pub(crate) async fn get_mpc_allowed_image_hashes(
     mpc_contract_id: AccountId,
     client: &actix::Addr<near_client::ViewClientActor>,
 ) -> anyhow::Result<(u64, AllowedDockerImageHashes)> {
--- a/node/src/indexer/lib.rs
+++ b/node/src/indexer/lib.rs
@@ -8,10 +8,17 @@ use near_indexer_primitives::types::AccountId;
 use near_indexer_primitives::views::QueryRequest;
 use near_indexer_primitives::views::QueryResponseKind::CallResult;
 use near_o11y::WithSpanContextExt;
+use serde::Deserialize;
 use std::time::Duration;
 use tokio::time;
 
+#[cfg(feature = "tee")]
+use mpc_contract::tee::proposal::AllowedDockerImageHashes;
+
 const INTERVAL: Duration = Duration::from_millis(500);
+#[cfg(feature = "tee")]
+const ALLOWED_IMAGE_HASHES_ENDPOINT: &str = "allowed_docker_image_hashes";
+const CONTRACT_STATE_ENDPOINT: &str = "state";
 
 pub(crate) async fn wait_for_full_sync(client: &Addr<ClientActor>) {
     loop {
@@ -36,13 +43,17 @@ pub(crate) async fn wait_for_full_sync(client: &Addr<ClientActor>) {
     }
 }
 
-pub(crate) async fn get_mpc_contract_state(
+pub(crate) async fn get_mpc_state<State>(
     mpc_contract_id: AccountId,
     client: &actix::Addr<near_client::ViewClientActor>,
-) -> anyhow::Result<(u64, ProtocolContractState)> {
+    endpoint: &str,
+) -> anyhow::Result<(u64, State)>
+where
+    State: for<'de> Deserialize<'de>,
+{
     let request = QueryRequest::CallFunction {
         account_id: mpc_contract_id,
-        method_name: "state".to_string(),
+        method_name: endpoint.to_string(),
         args: vec![].into(),
     };
     let query = near_client::Query {
@@ -59,4 +70,19 @@ pub(crate) async fn get_mpc_contract_state(
             bail!("got unexpected response querying mpc contract state")
         }
     }
+}
+
+pub(crate) async fn get_mpc_contract_state(
+    mpc_contract_id: AccountId,
+    client: &actix::Addr<near_client::ViewClientActor>,
+) -> anyhow::Result<(u64, ProtocolContractState)> {
+    get_mpc_state(mpc_contract_id, client, CONTRACT_STATE_ENDPOINT).await
+}
+
+#[cfg(feature = "tee")]
+pub(crate) async fn get_mpc_tee_state(
+    mpc_contract_id: AccountId,
+    client: &actix::Addr<near_client::ViewClientActor>,
+) -> anyhow::Result<(u64, AllowedDockerImageHashes)> {
+    get_mpc_state(mpc_contract_id, client, ALLOWED_IMAGE_HASHES_ENDPOINT).await
 }
--- a/node/src/indexer/mod.rs
+++ b/node/src/indexer/mod.rs
@@ -1,6 +1,13 @@
+use handler::ChainBlockUpdate;
+#[cfg(feature = "tee")]
+use mpc_contract::tee::proposal::AllowedDockerImageHash;
+use near_indexer_primitives::types::AccountId;
+use participants::ContractState;
+use std::sync::Arc;
+use tokio::sync::{mpsc, watch};
+use types::ChainSendTransactionRequest;
+
 pub mod configs;
-#[cfg(test)]
-pub mod fake;
 pub mod handler;
 pub mod lib;
 pub mod participants;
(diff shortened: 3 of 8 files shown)


node/src/indexer/tee.rs
```diff
@@ -0,0 +1,63 @@
...
+use tokio::sync::watch;
+
+use crate::indexer::{
+    lib::{get_mpc_tee_state, wait_for_full_sync},
+    IndexerState,
+};
+
+/// This future waits for the indexer to fully sync, and returns
+/// a [`watch::Receiver`] that will be continuously updated with the latest
+/// state [`BlockHeight`] and [`AllowedDockerImageHashes`] when a change is detected
```
near-bookrock, line 13:
```suggestion
/// allowed [`AllowedDockerImageHash`]es when a change is detected
```

node/src/indexer/tee.rs
```diff
@@ -0,0 +1,63 @@
...
+            .await
+            {
+                Ok((block_height, mut allowed_images_on_contract)) => {
+                    let allowed_images: Vec<AllowedDockerImageHash> =
+                        allowed_images_on_contract.get(block_height);
+                    break allowed_images;
+                }
+                Err(e) => {
+                    tracing::error!(target: "mpc", "error reading tee state from chain: {:?}", e);
+                    continue;
```
kevindeforth, line 38: Maybe add a progressive timeout here? 
Repeatedly failing here means there is something wrong and will eventually lead to the node getting kicked out.

Hence, it could be valuable to somehow get the attention of the node operators.

Not sure if/how we are monitoring error messages, but maybe one way to do it is to expose a metric, measuring the number of failed calls to this endpoint. The metric can then easily be integrated into a dashboard.

kevindeforth (commented):
This is good.

I would like to see the docker image hash exposed through a debug endpoint, but maybe @kuksag can handle that, since he is already working on the tee debug endpoint?

kevindeforth (approved):
Approving, since my comments can be addressed in follow-ups.

node/src/indexer/mod.rs
```diff
@@ -63,4 +69,7 @@ pub struct IndexerAPI {
     /// Sender to request transactions be signed (by a TransactionSigner that
     /// the indexer is initialized with) and sent to the chain.
     pub txn_sender: mpsc::Sender<ChainSendTransactionRequest>,
+    #[cfg(feature = "tee")]
+    /// Watcher that keeps track of allowed [`AllowedDockerImageHash`]es on the contract.
+    pub _allowed_docker_images_receiver: watch::Receiver<Vec<AllowedDockerImageHash>>,
```
netrome, line 74: We need to be cautious about feature flagging struct fields. While this is fine, we're limiting ourselves to never making this type a part of the crates public interface - or we'd have a feature toggling a SemVer incompatible change. Which if you didn't know already is [discouraged](https://doc.rust-lang.org/cargo/reference/features.html#semver-compatibility).

Just stressing this here as I've had to work through a lot of pain in the past when having SemVer incompatible changes behind feature flags.

node/src/indexer/tee.rs
```diff
@@ -0,0 +1,63 @@
...
+};
+
+/// This future waits for the indexer to fully sync, and returns
+/// a [`watch::Receiver`] that will be continuously updated with the latest
+/// allowed [`AllowedDockerImageHash`]es when a change is detected
+/// on the MPC smart contract.
+pub async fn monitor_allowed_docker_images(
+    indexer_state: Arc<IndexerState>,
+) -> watch::Receiver<Vec<AllowedDockerImageHash>> {
+    const CONTRACT_STATE_REFRESH_INTERVAL: std::time::Duration = std::time::Duration::from_secs(1);
```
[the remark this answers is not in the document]
netrome: Agreed. Polling every second sounds very aggressive.

node/src/indexer/tee.rs
netrome: Perhaps worth considering an exponential backoff here https://crates.io/crates/backoff.

node/src/indexer/lib.rs
```diff
@@ -60,3 +71,18 @@ pub(crate) async fn get_mpc_contract_state(
...
+
+pub(crate) async fn get_mpc_contract_state(
+    mpc_contract_id: AccountId,
+    client: &actix::Addr<near_client::ViewClientActor>,
+) -> anyhow::Result<(u64, ProtocolContractState)> {
+    get_mpc_state(mpc_contract_id, client, CONTRACT_STATE_ENDPOINT).await
+}
+
+#[cfg(feature = "tee")]
+pub(crate) async fn get_mpc_tee_state(
```
netrome, line 83: Seems like this method doesn't get the full TEE state. Should we rename it?

```suggestion
pub(crate) async fn get_mpc_allowed_image_hashes(
```

node/src/indexer/tee.rs
```diff
@@ -0,0 +1,63 @@
...
+                    continue;
+                }
+            };
+        }
+    };
+
+    let initial_state = fetch_allowed_image_hashes().await;
+    let (sender, receiver) = watch::channel(initial_state);
+
+    actix::spawn(async move {
```
netrome, line 47: Why do we need to use actix here instead of just `tokio::spawn`?

netrome (approved):
Mostly looks good to me. Some questions. I'd like to stress being cautious with how we use feature flags, but I haven't seen anything incorrect right now so that's promising at least.

node/src/indexer/tee.rs
```diff
@@ -0,0 +1,63 @@
...
+        loop {
+            tokio::time::sleep(CONTRACT_STATE_REFRESH_INTERVAL).await;
+            let new_tee_state = fetch_allowed_image_hashes().await;
```
barakeinav1, line 50: what do you do in case there is more than one allow hash.
during an hash upgrade process, I'd expect there to be 2 hashes for some grace period.
and in some corner cases even 3 (in case there a few fast consecutive hash updates) - @pbeza correct me if I'm wrong.

You should take the lasted (how do we check this?, block height) and write only that one to the "IMAGE_DIGEST_FILE"

kevindeforth: I would expect this sort of logic to happen on the receiver side.

node/src/indexer/tee.rs
kevindeforth: c.f. https://github.com/near/mpc/issues/156

node/src/indexer/tee.rs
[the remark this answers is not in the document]
near-bookrock: I don't mind reducing the frequency, but there is no disadvantage to doing it every second either. It's a read-only "query" method on the contract that is only run locally on the indexer, so there is no gas uage associated it. https://github.com/Near-One/mpc/blob/8fc44a0bb5f5076fd3eec95aab6bb10a966fef2c/node/src/indexer/lib.rs#L59-L62

Like Barak is pointing out, it's the exact same mechanism for polling the contract state which is done at the same frequency, but involves much more state.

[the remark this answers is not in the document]
near-bookrock: >Note: Tokio can [auto-advance time](https://docs.rs/tokio/latest/tokio/time/fn.pause.html#auto-advance) so this should not be a problem for unit tests.

This is great for unit and integration tests, but I think Barak is asking for when we do system tests and spin up the whole binary.

node/src/indexer/mod.rs
near-bookrock: Thanks for bringing it up, no I was not aware of this. For some context this module is part of the main binary crate, which we don't expose as a package.

In such cases is it still discouraged? What would a better alternative be?

netrome: Within the same crate it's fine. The problem is if it's part of the public interface and you use it in another crate.

This may sound a bit constructed but is pretty easy to fall into:

Say our crate `shared_library` exposes a public struct `Stuff` with a TEE-specific field `tee_stuff`:

```rust
pub struct Stuff {
  normal_stuff: NormalStuff,
  #[cfg(feature = "tee")]
  tee_stuff: TeeStuff,
}
```

Assume we use this in two utility crates `shared_utils` and `tee_utils`, where the `shared_utils` do not depend on the `tee` feature but the `tee_utils` do depend on it.

Because `shared_utils` do not depend on TEE it should not have to activate this feature flag. Therefore it may contain code like:

```
let stuff = Stuff { normal_stuff: ... }
```

However, the problem is that if we write code that depends on both `shared_utils` and `tee_utils` - cargo will compile the `shared_library` with the TEE feature (for both crates) - which would lead to a compilation error when building `shared_utils`.

I.e. compilation of `shared_utils` would break because we included `tee_utils` in the build. The reason is that cargo assumes that it's always safe to add extra feature flags when compiling shared dependencies. This can pop up in other scenarios too and typically requires ugly workarounds - for example ensuring every downstream dependency of `shared_utils` also expose a `tee` flag.

### TL;DR
Cargo assumes feature flags don't introduce SemVer incompatible changes. Stuff can break in non-intuitive ways if we break this, so let's not.

node/src/indexer/tee.rs
near-bookrock: Yes, it is coming in a follow up PR. Will tag you Barak to review.

libs/chain-signatures/contract/src/tee/proposal.rs
```diff
@@ -64,15 +64,15 @@ impl CodeHashesVotes {
 
 /// A proposal for a new TEE code hash to be added to the whitelist, along with the time it was
 /// added.
-#[near(serializers=[borsh])]
-#[derive(Debug, Clone)]
+#[near(serializers=[borsh, json])]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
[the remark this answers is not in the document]
near-bookrock: Yeah, I know. I'd rather not add a feature flag.`wasm-opt` will remove this code anyway for the deserializer, and hopefully we move away to proper interface types so we don't have this dependency in the future https://github.com/near/mpc/issues/515

Let me know If you feel strongly about it and want to introduce feature flags.

@netrome What do you think about this?

node/src/indexer/mod.rs
near-bookrock: Thanks a lot for the thorough explanation. Yeah, now that you explain it I recall to also have run into some nasty issues with feature unification of transitive dependencies.

So if I understand correctly, it is fine in this scenario since we are doing it in a binary crate?

kevindeforth (approved):
Thank you!

node/src/indexer/mod.rs
netrome: > So if I understand correctly, it is fine in this scenario since we are doing it in a binary crate?

Oh I see I wasn't super clear on this. This scenario is fine because it's a private struct within the crate. If we make this struct public, we have a problem. Although public structs within binary crates doesn't make sense. But we have too much stuff in the binary crate and should break things out to shared libraries, and in that case this may pose a problem depending on how we'll expose/use it.

netrome: I.e.

Bad:
```rust
pub struct Stuff {
  normal_stuff: NormalStuff,
  #[cfg(feature = "tee")]
  tee_stuff: TeeStuff,
}
```

Okay:
```
pub(crate) struct Stuff {
  normal_stuff: NormalStuff,
  #[cfg(feature = "tee")]
  tee_stuff: TeeStuff,
}
```

PR #517, opened 2025-06-17
[DRAFT] test(TEE): test automated kickout mechansim

### LINKED ISSUE #514: Test contracts `verify_tee` with valid and invalid TEE state
need to write tests for the following:

- if participants have valid Tee statuses, they do not get kicked out
- if participants have invalid Tee statuses, they do get kicked out
- if less than threshold participants remain, the contract stops accepting signature requests or responses.

also, add a pytest for checking if the endpoint is called and succeeds.

kevindeforth:
DRAFT! This is incomplete right now!

Follow-up to https://github.com/near/mpc/pull/468.
Meant to resolve #514


kevindeforth:
closing due to change of scope  #514

PR #526, opened 2025-06-19
chore: increase bytes allocated to version in report data from u8 to u16

ghost:


node/src/tee/mod.rs
```diff
@@ -55,17 +74,20 @@ pub async fn create_remote_attestation_info(
     let client_info_response = client.info().await?;
     let tcb_info = client_info_response.tcb_info;
 
-    let report_data = {
+    let report_data: [u8; REPORT_DATA_SIZE] = {
+        let mut report_data = [0u8; REPORT_DATA_SIZE];
+
+        // Copy binary version
+        let byte_representation = binary_version.0.to_be_bytes();
```
barakeinav1, line 81: I didn't see above where you initialized this value. if you didn't let set it to 1.

barakeinav1 (changes requested):

PR #537, opened 2025-06-20
feat(contract): verify `report_data` field of the TEE quote

pbeza:


kevindeforth (approved):
~Pending that one comment~(edit: resolved), this seems as correct as it can get without testing :sunglasses: 
Lets open a issue for testing though (I don't suppose we have a valid attestation at hand, that we could include in a test before EOD?)

pbeza:
> Lets open a issue for testing though

https://github.com/near/mpc/issues/539 ✅

libs/chain-signatures/contract/src/tee/tee_participant.rs
```diff
@@ -94,6 +94,29 @@ impl TeeParticipantInfo {
...
+
+    fn verify_report_data(&self, quote: &Quote, expected_public_key: PublicKey) -> bool {
+        let report_data = match quote.report.as_td10() {
+            Some(r) => r.report_data,
+            None => return false,
+        };
+        let binary_version = report_data[0];
+        let mut public_keys_hash = GenericArray::<u8, U48>::default();
+        public_keys_hash.copy_from_slice(&report_data[4..52]); // 48 bytes for SHA3-384
+        expected_public_key.into_bytes() == public_keys_hash.as_slice()
```
barakeinav1, line 116: you should check for:
reportdata  ==  Version || SHA384(node's public signing key || account key)

byte Version=2 bytes, big Indian;
please see Daniel's PR if not clear.

Issue #465, opened 2025-06-04
Submit remote attestation on startup if contract has changed


near-bookrock:
The node needs to send remote attestation to the contract on startup if it has deteceted a contract update. Linked to https://github.com/Near-One/mpc/issues/464

PR #543, opened 2025-06-23
feat: submit remote attestation on startup

ghost:
Resolves https://github.com/near/mpc/issues/465

### DIFF
--- a/node/src/cli.rs
+++ b/node/src/cli.rs
@@ -1,5 +1,4 @@
 use crate::config::{PersistentSecrets, RespondConfig};
-use crate::tee::remote_attestation::submit_remote_attestation;
 use crate::{
     config::{
         load_config_file, BlockArgs, ConfigFile, IndexerConfig, KeygenConfig, PresignatureConfig,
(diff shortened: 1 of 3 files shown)


node/src/tee/remote_attestation.rs
```diff
@@ -59,19 +66,47 @@ struct UploadResponse {
...
+        match operation().await {
+            Err(err) => {
+                let duration = backoff.next().unwrap_or(MAX_BACKOFF_DURATION);
+                error!(?err, "{description} failed. retrying in: {:?}", duration);
+                continue;
+            }
+            Ok(response) => break response,
+        }
+    }
+}
```
pbeza, line 95: That looks like a pretty generic pattern that should probably be implemented somewhere in the standard libraries, no? Isn’t it equivalent to something like this?

```
use tokio_retry::strategy::ExponentialBackoff;
use tokio_retry::Retry;

let retry_strategy = ExponentialBackoff::from_millis(10).take(10);

let result = Retry::spawn(retry_strategy, || async {
    // your operation
});
```

near-bookrock: Yeah, the example you show is a quite standard pattern and it also exists in the `backon` crate we have for exponential backoffs.

However, I don't want the retry mechanism to return a Result, I want it to retry forever until the Ok case is hit. `Retry`'s future implementation is returning a `Result`, not the value we are trying to produce.

Just note that I also don't find it ideal to loop forever as a backoff mechanism. But we don't have time to implement proper error handling for the audit.

pbeza: Sure. I didn’t mean to be too nitpicky here — I hate when people nitpick about general Rust style, ignoring the core logic of the project where the real difficulty of the implementation lies. Tests are failing, but approving since you probably just need to fix the import with `use`. ✅

PR #524, opened 2025-06-18
feat: Initial launcher script

barakeinav1:

PR #549, opened 2025-06-23
chore(tee): launcher cleanup

kevindeforth:
Follow-up to https://github.com/near/mpc/pull/524, resolving the most pressing clean-up issues raised during the review (https://github.com/near/mpc/pull/524#pullrequestreview-2945983472).

- RPC variables (timeout for a request and timeout between succsessive requests) can now be passed through environment variables.
- RPC variables default behavior is changed (before: waiting indefinitely and doing only one request, now, waiting for 10 seconds max and attempting up to 20 requests).
- additional type checks for user env variables (tags, registry and docker_image name)
- print `stderr` of failed processes instead of just the error code.

There are still a ton of panics in this code and splitting some of these bigger functions up would improve readability. But in the interest of simplifying review and concentrating on the most pressing issues, this is delayed to later PRs.


tee-launcher/launcher.py
```diff
@@ -141,62 +296,113 @@ def main():
...
+                case 'application/vnd.docker.distribution.manifest.v2+json' | \
+                     'application/vnd.oci.image.manifest.v1+json':
+                    config_digest = manifest['config']['digest']
+                    if config_digest == docker_image.digest:
+                        return manifest_resp.headers['Docker-Content-Digest']
+        except:
+            print(
+                "[Warning] Exceeded number of maximum RPC requests for any given attempt. Will continue in the hopes of finding the matching image hash among remaining tags"
+            )
+            # Q: Do we expect all requests to succeed?
```
kevindeforth, line 397: do we?

tee-launcher/launcher.py
```diff
@@ -1,21 +1,107 @@
...
+# only considered if `IMAGE_DIGEST_FILE` does not exist.
+ENV_VAR_DEFAULT_IMAGE_DIGEST = "DEFAULT_IMAGE_DIGEST"
+# the time to wait between rpc requests, in milliseconds. Defaults to 500 milliseconds.
+OS_ENV_VAR_RPC_REQUEST_INTERVAL_MS = 'RPC_REQUST_INTERVAL_MS'
+# the maximum time to wait for an rpc response. Defaults to 10 seconds.
+OS_ENV_VAR_RPC_REQUST_TIMEOUT_SECS = 'RPC_REQUST_TIMEOUT_SECS'
+# the maximum number of attempts for rpc requests until we raise an exception
+OS_ENV_VAR_RPC_MAX_ATTEMPTS = 'RPC_MAX_RETRIES'
+# MUST be set to 1.
+OS_ENV_DOCKER_CONTENT_TRUST = 'DOCKER_CONTENT_TRUST'
```
netrome, line 28: Why do we need it if it must be set to 1?

tee-launcher/launcher.py
```diff
@@ -1,21 +1,107 @@
...
+
+# Dstack user configuration flags
+DSTACK_USER_CONFIG_FILE = '/tapp/user_config'
+
+# Dstack user config. Read from `DSTACK_USER_CONFIG_FILE`
+USER_ENV_VAR_LAUNCHER_IMAGE_TAGS = 'LAUNCHER_IMAGE_TAGS'
+USER_ENV_VAR_LAUNCHER_IMAGE_NAME = 'LAUNCHER_IMAGE_NAME'
+USER_ENV_VAR_LAUNCHER_IMAGE_REGISTRY = 'LAUNCHER_REGISTRY'
```
netrome, line 36: I'm confused. The comment says we should read these from a file but we still have env variables for it? Why not just take a file path?

tee-launcher/launcher.py
```diff
@@ -1,21 +1,107 @@
...
+
+# Default values for dstack user config file.
+DEFAULT_LAUNCHER_IMAGE_NAME = 'nearone/mpc-node-gcp'
+DEFAULT_REGISTRY = 'registry.hub.docker.com'
+DEFAULT_LAUNCHER_IMAGE_TAG = 'latest'
```
netrome, line 41: We should use a specific tag here, right?

tee-launcher/launcher.py
```diff
@@ -1,21 +1,107 @@
...
+
+def is_non_empty_and_cleaned(val: str) -> bool:
+    if not isinstance(val, str):
+        return False
+    if not val.strip():
+        return False
+    return val.strip() == val
+
+
+@dataclass(frozen=True)
```
netrome, line 55: Nice!

tee-launcher/launcher.py
barakeinav1: don't know

netrome: Feels weird to keep looping if we have failed the retry loops. If we hit rate limits, wouldn't it be better to use a pretty long exponential backoff? And if we still fail I think we should raise an exception here and let the launcher fail explicitly instead of hanging and continue sending failing requests.

tee-launcher/launcher.py
```diff
@@ -141,62 +296,113 @@ def main():
...
+    Returns:
+        Response: The successful HTTP response object with status code 200.
+
+    Notes:
+        - Retries indefinitely until the request succeeds.
+        - Prints a warning with the response content on each failure.
+    """
+    for attempt in range(1, rpc_max_attempts + 1):
+        # we sleep at the beginning, to ensure that we respect the timeout. Performance is not a priority in this case.
+        time.sleep(rpc_request_interval_secs)
```
netrome, line 322: We could consider increasing this interval on each loop by a factor of e.g. 1.5 until we hit a max interval of 1 minute (or any value that seems sensible).

netrome (approved):
Nice stuff. Some thoughts and questions from me.

near-bookrock (approved):

barakeinav1 (approved):

tee-launcher/launcher.py
kevindeforth: not sure, this was done by @barakeinav1

tee-launcher/launcher.py
kevindeforth: Good point. I renamed to `DSTACK_USER_CONFIG_*`

tee-launcher/launcher.py
kevindeforth: @barakeinav1 will know this better than me, I preserved his logic

tee-launcher/launcher.py
kevindeforth: > And if we still fail I think we should raise an exception here and let the launcher fail explicitly instead of hanging and continue sending failing requests.

IIUC, then we don't necessarily require these requests to succeed. Seems to me like we are searching for a matching hash https://github.com/near/mpc/blob/3610a24b1434a91dfc46f642cd40e3372d2a8c20/tee-launcher/launcher.py#L393

Before this PR, we ignored any tag for which we got a response different from 200.
This meant that if we hit a rate limit at the hash of interest (or any other error that was recoverable), the launcher would not retry and eventually fail, since we would skip the tag of interest and probably not find it anymore.

With this PR, we try to give us the best chances of success. This might lead to wasting some resources, but I don't understand this endpoint enough to judge that. (I would guess @barakeinav1 has more insights here)

For example, if we request a tag that does not exist, will we get a `400` or a `404`? In such a case, we should probably just skip the tag. But given my limited understanding of this endpoint and how it is expected to behave, I thought it would be better to err on the side of caution and just retry up to the allotted limit.

If it turns out to be bad performance wise, we can always increase the timeout and decrease the maximum number of attempts through the environment variables.

tee-launcher/launcher.py
barakeinav1: no latest will pull the latest, and it should be the latest voted node.

barakeinav1 (approved):

tee-launcher/launcher.py
pbeza: Probably worth adding that to the code comment then.

```suggestion
# MUST be set to 1 to enforce Docker to validate image signatures when launching containers 
# See: https://docs.docker.com/engine/security/trust/#client-enforcement-with-docker-content-trust
OS_ENV_DOCKER_CONTENT_TRUST = 'DOCKER_CONTENT_TRUST'
```

tee-launcher/launcher.py
barakeinav1: from security perspective - we don't have a problem we always validated against approved hash. 
and user can specify a tag if he wants to change it.
using will allow the user to use the launcher without adding config change more often

PR #555, opened 2025-06-23
test(pytest): resolve pytest nonce conflicts

### LINKED ISSUE #523: [Flaky test] Pytest `test_multi_domain` is flaky
Example failure: https://github.com/near/mpc/actions/runs/15729139469/job/44325895124

kevindeforth:
resolves #523 

Before this PR, pytest used the same `signer_key` that was used by the MPC nodes. This resulted in occasional nonce conflicts among the MPC nodes (c.f. #523 ).

In this PR, we generate additional access keys for the mpc node accounts. These keys are used exclusively by the pytest framework (for initiating resharing, initializations, voting on updates etc.).

Note that we do not need to generate additional access keys for the contract accounts, as those are already exclusive to the pytest framework. Any conflicts occurring on that side must be handled through a more elaborate nonce tracking mechanism (not necessary for now).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@ venv
 **/target/
 /pytest/output.log
 /pytest/venv/
-/pytest/*/__pycache__/
+/pytest/**/__pycache__/
 /pytest/nearcore_pytest/nearcore_pytest.egg-info/
 /pytest/tests/test_contracts/parallel/target/
 
--- a/pytest/common_lib/shared/__init__.py
+++ b/pytest/common_lib/shared/__init__.py
@@ -3,7 +3,8 @@ import os
 import pathlib
 import subprocess
 import sys
-from typing import List, Iterable, Optional, Tuple
+from typing import List, Optional, Tuple, cast
+import typing
 
 import base58
 import yaml
@@ -36,16 +37,17 @@ def serialize_key(key: Key):
 
 def deserialize_key(account_id: str, key: str) -> Key:
     assert key.startswith('ed25519:')
-    signing_key = SigningKey(base58.b58decode(key.split(':')[1].encode('ascii'))[:32])
+    signing_key = SigningKey(
+        base58.b58decode(key.split(':')[1].encode('ascii'))[:32])
     return Key.from_keypair(account_id, signing_key)
 
 
 def sign_create_account_with_multiple_access_keys_tx(
-        creator_key: Key,
-        new_account_id,
-        keys: List[Key],
-        nonce,
-        block_hash
+    creator_key: Key,
+    new_account_id,
+    keys: List[Key],
+    nonce,
+    block_hash,
 ) -> bytes:
     create_account_action = create_create_account_action()
     payment_action = create_payment_action(100 * NEAR_BASE)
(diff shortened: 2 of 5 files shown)


pytest/common_lib/shared/__init__.py
```diff
@@ -219,38 +249,66 @@ def start_cluster_with_mpc(
...
-            cluster.contract_node.last_block_hash()
+            nonce,
+            cluster.contract_node.last_block_hash(),
         )
         cluster.contract_node.send_txn_and_check_success(tx)
 
+        candidate_account_id = candidate.signer_key.account_id
+
+        pytest_signer_keys = []
+        for i in range(0, 5):
```
near-bookrock, line 282: Question. Why specifically 5 signer keys?

near-bookrock (approved):

.gitignore
```diff
@@ -6,7 +6,7 @@ venv
 **/target/
 /pytest/output.log
 /pytest/venv/
-/pytest/*/__pycache__/
+/pytest/**/__pycache__/
```
netrome, line 9: Nice. Ran the pytests today for the first time and got some unwanted `__pycache__` folders in my git worktree :sweat_smile:.

pytest/common_lib/shared/__init__.py
```diff
@@ -60,14 +62,14 @@ def sign_create_account_with_multiple_access_keys_tx(
...
+        creator_key.decoded_sk(),
     )
     return serialize_transaction(signed_tx)
 
 
 def start_neard_cluster_with_cleanup(
-        num_validators: int,
-        num_mpc_nodes: int,
+    num_validators: int,
+    num_mpc_nodes: int,
```
netrome, line 72: I take it we don't have any python formatting in our CI? We should probably add a step to verify we follow some standard https://github.com/psf/black format.

pytest/common_lib/shared/__init__.py
```diff
@@ -86,14 +88,18 @@ def start_neard_cluster_with_cleanup(
...
-    )
+    # the config is set to local, so we expect local nodes.
+    nodes: typing.List[LocalNode] = cast(
+        List[LocalNode],
+        start_cluster(num_validators,
+                      num_mpc_nodes,
+                      1,
+                      None, [("epoch_length", 1000),
+                             ("block_producer_kickout_threshold", 80)],
+                      client_config_changes=client_config_changes))
```
netrome, line 102: Do we do explicit type checking somewhere? Why do we need to `cast` here? Afaik that's a noop when running the code.

Ref: https://docs.python.org/3/library/typing.html#typing.cast

netrome (commented):
Some post-merge questions from my side. Otherwise lgtm.

.gitignore
pbeza: I ran into the same issue with `**/target`, which [used to be](https://github.com/near/mpc/commit/79e4ac24ab7fb91056982710a69850c71ce60e78#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947R6) `*/target`. I presume there may be more non-optimal stuff in `.gitignore`.

pytest/common_lib/shared/__init__.py
pbeza: Lets add https://pypi.org/project/pylint/ (or some alternative) too!

pytest/common_lib/shared/__init__.py
kevindeforth: Good question — you're right, cast doesn't do any runtime type checking. I use it here to help the type checker  understand the expected type.

In this case, `start_cluster(...)` returns a `List[BaseNode]` that the type checker can't confidently infer as `List[LocalNode]`.
The cast makes the intended type explicit and improves editor support and readability, even though it doesn't affect execution.

But I am quite a python noob. Most likely, there is a better way of handling this.

pytest/common_lib/shared/__init__.py
netrome: Happy to use yapf as well. That's what I used in past projects, but my understanding is that Black is the more popular one. No strong opinion on either.

Created an issue to do this: https://github.com/near/mpc/issues/593

pytest/common_lib/shared/__init__.py
netrome: Yeah I haven't worked too much with strongly typed Python so don't know what's best practice here.

PR #556, opened 2025-06-24
fix: ignore peers who are behind in a computation

bowenwang1996:
Ignore peers who are behind in a computation to prevent them from ruining a computation because they are stuck

TODO: add tests

### DIFF
--- a/node/src/network.rs
+++ b/node/src/network.rs
@@ -91,6 +91,10 @@ impl NetworkTaskChannelManager {
 const LRU_CAPACITY: usize = 10000;
 
 impl MeshNetworkClient {
+    /// The maximum height difference that we are willing to accept.
+    /// This is used to filter out participants that are too far behind in the indexer height.
+    const MAX_HEIGHT_DIFF: u64 = 100;
+
     /// Primary functionality for the MeshNetworkClient: returns a channel for the given
     /// new MPC task. It is expected that the caller is the leader of this MPC task, and that the
     /// way the MPC task IDs are assigned ensures that no two participants would initiate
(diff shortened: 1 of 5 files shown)


node/src/network.rs
```diff
@@ -142,14 +146,20 @@ impl MeshNetworkClient {
     /// This is a subset of all_participant_ids, and includes our own participant ID.
     pub fn all_alive_participant_ids(&self) -> Vec<ParticipantId> {
         let mut result = Vec::new();
+        let my_height = *self
+            .get_indexer_heights()
```
robin-near, line 150: should grab self.get_indexer_heights() into a variable first to avoid building it repeatedly

libs/nearcore
netrome, line 1: Why do we change the submodule? It's not even used afaik. (i.e. I can delete the submodule and the code still compiles).

kevindeforth: We compile the nearcore client (indexer) from this submodule for the pytests and production. It is not used in Rust integration tests, because we use the fake indexer there.

But I am not sure if this was deliberate. The pytests fail when I use this commit (just like they fail in the CI).
The pytests pass when I revert to the same commit used in `mainnet-release`.

netrome: Thank you for the clarification :pray:

node/src/tests/faulty.rs
```diff
@@ -119,4 +120,101 @@ async fn test_faulty_cluster() {
...
+            .participants
+            .participants
+            .iter()
+            .map(|p| p.id)
+            .collect::<Vec<_>>()
+    );
+
+    tokio::time::sleep(std::time::Duration::from_secs(30)).await;
+
+    assert!(request_signature_and_await_response(
```
robin-near, line 212: should we do this a few times - coz there's a 1/4 chance of succeeding even without a correct implementation right?

robin-near (approved):

node/src/tests/faulty.rs
netrome: Created https://github.com/near/mpc/issues/561

PR #558, opened 2025-06-24
feat(devnet): enable ssd support

kevindeforth:
Requires https://github.com/Near-One/infra-ops/pull/476

### DIFF
--- a/devnet/src/cli.rs
+++ b/devnet/src/cli.rs
@@ -187,6 +187,9 @@ pub struct NewMpcNetworkCmd {
     /// higher amounts here.
     #[clap(long, default_value = "1")]
     pub near_per_responding_account: u128,
+    /// Indicates if the machines should be using SSD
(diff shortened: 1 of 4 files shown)


netrome:
> > > Why should the MPC node keep track of whether or not it's running on a machine with an SSD?
> > > I don't see any other use here than logging this information, and the node has no way of verifying if this is actually true or a misconfiguration.
> > 
> > 
> > Maybe I am missing something. But I don't see any code changes to the MPC? There's only changes to the devnet setup?
> 
> Oh you're right my bad - I missed that this is the devnet setup. Then my question changes, where do we actually set the SSDs? Because I only see this field being used for logging.

Nvm I missed the terraform config. Now I see how this is used. Thank you for calling me out on this @near-bookrock :pray: .

netrome (approved):

near-bookrock (approved):
Thank you!

PR #560, opened 2025-06-24
feat: enable network hardship simulation

kevindeforth:
### This PR introduces:

- a pytest for https://github.com/near/mpc/pull/595
- the ability to simulate network hardship for the indexer.

Additionally, resolves https://github.com/near/mpc/issues/561

### Hardship Testing:
The binary can now be compiled with the feature `--network-hardship-simulation`. If done so, then the indexer will read the contents of a file `listen_blocks.flag`, located in the home directory of the node.

If the the file exists, then it must contain a boolean value. If that value is `false`, then the indexer will stop ingesting blocks, simulating an error similar to what we encountered on mainnet.

This feature can be used to test the fix introduced in #595 in pytests, testnet or mainnet clusters

If we need to put https://github.com/near/mpc/issues/592 on hold or not implement it at all, due to constants of our TEE set-up, then this would be an alternative for simulating network hardship.

### DIFF
--- a/node/src/indexer/real.rs
+++ b/node/src/indexer/real.rs
@@ -3,6 +3,8 @@ use super::participants::{monitor_contract_state, ContractState};
 use super::stats::{indexer_logger, IndexerStats};
 use super::tx_sender::handle_txn_requests;
 use super::{IndexerAPI, IndexerState};
+#[cfg(feature = "network-hardship-simulation")]
+use crate::config::load_listening_blocks_file;
 use crate::config::{IndexerConfig, RespondConfig};
 #[cfg(feature = "tee")]
 use crate::indexer::tee::monitor_allowed_docker_images;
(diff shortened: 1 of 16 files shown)


node/src/indexer/real.rs
```diff
@@ -3,16 +3,40 @@ use super::participants::{monitor_contract_state, ContractState};
...
 use std::path::PathBuf;
 use std::sync::Arc;
+#[cfg(feature = "network-hardship-simulation")]
+use std::time::Duration;
 use tokio::sync::{mpsc, oneshot, watch, Mutex};
 
-#[cfg(feature = "tee")]
-use crate::indexer::tee::monitor_allowed_docker_images;
+#[cfg(feature = "network-hardship-simulation")]
+pub async fn check_block_processing(process_blocks_sender: watch::Sender<bool>, home_dir: PathBuf) {
```
netrome, line 21: It feels weird to have to poll a file for this. Have you considered adding an HTTP endpoint to toggle this? With that we could easily toggle this behavior in devnet by just making a post/put (not sure which is best) request.

The endpoint would also make synchronization easier because it wouldn't return until we've toggled the flag at which point we'd know the block processing has stopped. In this implementation, there could be up to a 2 second lag until the block processing stops, with no way for the process changing this file to await this.

netrome (approved):
Approving but am curious about if you looked into using an HTTP endpoint instead of a file. Seems like it would bring a few benefits if there's no big problem with that approach.

node/src/indexer/real.rs
kevindeforth: > Have you considered adding an HTTP endpoint to toggle this? With that we could easily toggle this behavior in devnet by just making a post/put (not sure which is best) request.

This is an interesting proposition. I have not considered it. For devnet, I thought we would continue the terraform / nomad route to trigger any such behaviors through some config flag or by triggering a script.

I think your proposing a promising idea, but I am hesitant to turn the node --> web-server one way street into a node <--> web-server two way street, if it is just for testing. This might impose a few architectural constraints on the code which I don't see the need for, as long as we can work with config file settings (which are very easy to add or remove from the code).

>  The endpoint would also make synchronization easier because it wouldn't return until we've toggled the flag at which point we'd know the block processing has stopped. In this implementation, there could be up to a 2 second lag until the block processing stops, with no way for the process changing this file to await this.

I don't see benefit in instant feedback, because whatever process is triggering this behavior will most likely still need to monitor the block height of the node it was triggered on (just like in the pytest). That is because our implementation _must_ be forgiving for reasonable delays of the indexer, and thus we will probably always have to wait for a specific delta to pass.

netrome: Fair enough. For https://github.com/near/mpc/issues/602 I think we'll have to consider adding write capabilities to the web server (mutating file system & restart indexer most likely), but it makes sense to think that architecture through.

> I don't see benefit in instant feedback, because whatever process is triggering this behavior will most likely still need to monitor the block height of the node it was triggered on (just like in the pytest). That is because our implementation must be forgiving for reasonable delays of the indexer, and thus we will probably always have to wait for a specific delta to pass.

Not really sure if I follow what you refer to here. For the current tests sure, but I'd imagine that there could be scenarios where we want to stop this blocks processing and assert some property directly afterwards. Though I need to understand the system better before I can come up with a specific example. Perhaps you are right that there are no conceivable cases where we'd need this, but as a rule of thumb I like to be able to await any effects to allow for better testability.

PR #576, opened 2025-06-26
feat(pytest): interactive pytest

kevindeforth:
This is helpful if we need more elaborate testing, but don't want to deploy a cluster with devnet.

### DIFF
--- a/node/src/tests/multidomain.rs
+++ b/node/src/tests/multidomain.rs
@@ -96,16 +96,20 @@ async fn test_basic_multidomain() {
     }
 
     // give nodes time to enter resharing so we don't lose the request.
-    tokio::time::sleep(Duration::from_millis(500)).await;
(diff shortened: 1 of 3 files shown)


kevindeforth:
> Any reason for using a test instead of a stand-alone script?

I still want to run all the pytest fixtures. They are useful, as they compile the contract, ensure we are setting appropriate environment variables and cleanup after the test.

Right now, there is no use-case for which we would prefer to do these steps manually.

netrome:
> > Any reason for using a test instead of a stand-alone script?
> 
> I still want to run all the pytest fixtures. They are useful, as they compile the contract, ensure we are setting appropriate environment variables and cleanup after the test.
> 
> Right now, there is no use-case for which we would prefer to do these steps manually.

Fair enough. I need to learn more about how our pytests are set up.

Issue #579, opened 2025-06-27
[Task] Add validation for `.user-config` env file in TEE launcher


netrome:
### Background

Currenty, our launcher runs the MPC node with env vars from an unmeasured file.

```python
    # Build the docker command we use to start the app, i.e., mpc node
    docker_cmd = ['docker', 'run']

    env_file = '/tapp/.host-shared/.user-config'
    if os.path.isfile(env_file):
        docker_cmd += ['--env-file', env_file]
```

This could be exploited to allow for arbitrary code execution within the CVM.

### User Story

As an operator, I should not be able to run arbitrary code on a TEE enabled MPC node.

### Acceptance Criteria

The provided env file is measured and validated so that it only allows relevant variables that can't be exploited.

netrome:
Note: We probably shouldn't use dotenv format at all, but rather use a more conservative format of passing additional options.

PR #588, opened 2025-06-29
fix(launcher): enforce env var allow-list and custom host/port parsing in user-config

barakeinav1:
fixes trailOfBit 2 security issues:
1. Env file parsing can lead to incorrect tagging
4. Env-File Could Lead to Arbitrary Code Execution

also solves:
[Task] Add validation for .user-config env file in TEE launcher
[#579](https://github.com/near/mpc/issues/579)


tee-launcher/launcher.py
```diff
@@ -94,23 +129,19 @@ def tags(self) -> list[str]:
     def registry(self) -> str:
         return self.spec.registry
 
-
+#Parses a .env-style file into a dictionary of key-value pairs.
```
pbeza, line 132: Is there any good reason we're manually parsing the config file? We could probably replace it with a one-liner using an existing parsing library.

Also, using regex to manually parse almost anything is a red flag for me.

tee-launcher/launcher.py
```diff
@@ -43,6 +45,39 @@
...
+# RUST_BACKTRACE=1
+# RUST_LOG=info
+# MPC_RESPONDER_ID=responder-xyz
+# IMAGE_HASH=sha256:abc123...
+# LATEST_ALLOWED_HASH_FILE=/mnt/shared/image-digest
+# EXTRA_HOSTS=host1:192.168.0.1,host2:192.168.0.2
+# PORTS=11780:11780,2200:2200
+
+# Define an allow-list of permitted environment variables:
+ALLOWED_ENV_VARS = {
```
pbeza, line 66: Why didn’t you add `EXTRA_HOSTS` and `PORTS` to `ALLOWED_ENV_VARS`? Did you overlook it?

tee-launcher/launcher.py
netrome: Created an issue to add black or yapf to our CI.
https://github.com/near/mpc/issues/593

netrome: Note that the report advocates for not using .env at all. The current code doesn't handle variable and command substitution for example, which means we don't resolve the first minor security issue.

Since dotenv comes with some features we don't need, I think we should do as recommended and not use the format at all. Instead we could just take a CSV, JSON or TOML file with the relevant configuration and use a parsing library for that.

tee-launcher/launcher.py
```diff
@@ -94,23 +129,19 @@ def tags(self) -> list[str]:
...
+            # Remove inline comments (unquoted)
+            line = re.split(r'\s+#', line, 1)[0].strip()
             if not line or line.startswith('#'):
                 continue
             key, _, value = line.partition("=")
-            env[key.strip()] = value.strip()
+            key = key.strip()
+            value = value.strip().strip('"').strip("'")  # remove quotes if any
+            env[key] = value
     return env
```
pbeza, line 145: This still isn't correct `.env` file parsing. I strongly [recommend](https://github.com/near/mpc/pull/588/files#r2173889571) reusing an existing library that handles it correctly out of the box. I asked Copilot to find cases where this parsing would break, and it did — I’ve verified some of them.

image

tee-launcher/launcher.py
```diff
@@ -263,14 +294,33 @@ def main():
     # Build the docker command we use to start the app, i.e., mpc node
     docker_cmd = ['docker', 'run']
 
-    env_file = '/tapp/.host-shared/.user-config'
-    if os.path.isfile(env_file):
-        docker_cmd += ['--env-file', env_file]
+    #load environment variables from the user config file
```
pbeza, line 297:
```suggestion
    # Load environment variables from the user config file
```

tee-launcher/launcher.py
pbeza: > Since dotenv comes with some features we don't need, I think we should do as recommended and not use the format at all. (...)

Good point. In that case, @barakeinav1, feel free to ignore or resolve my code review comments related specifically to the .env file.

barakeinav1: we are not using an .env file anymore. I removed this from [here](https://github.com/near/mpc/pull/588/files#:~:text=env_file%20%3D%20%27/tapp/.host%2Dshared/.user%2Dconfig%27).

this was the most dangerous part

tee-launcher/launcher.py
```diff
@@ -263,14 +294,33 @@ def main():
...
+        user_env = parse_env_file(user_env_file)
+        for key, value in user_env.items():
+            if key in ALLOWED_ENV_VARS:
+                docker_cmd += ['--env', f'{key}={value}']
+            elif key == "EXTRA_HOSTS":
+                # Allow optional host entries in the same file, comma-separated
+                for host_entry in value.split(','):
+                    clean_host = host_entry.strip()
+                    if clean_host:
+                        docker_cmd += ['--add-host', clean_host]
```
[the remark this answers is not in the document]
barakeinav1: hmm. this is something we should check. 
this kind of attack (if possible) is very dangerous for us.

tee-launcher/launcher.py
barakeinav1: @pbeza @netrome, chatgpt suggested using TOML instead. what do you think? I'm not familiar with it, but seems to have a more strict parser, and more rigid structure. also it has build in parser in python as part of the standard library.
do you have any other suggestion?

netrome: I like TOML. Afaik it should be a good choice for our use case. Not sure if there are any advanced features that complicates this similar to dotenv. Could be worth researching, but the format should be pretty simple.

tee-launcher/launcher.py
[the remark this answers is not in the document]
barakeinav1: I've added validation for EXTRA_HOSTS and PORTS to prevent CLI injection:
https://github.com/barakwei/mpc/commit/19ef455

tee-launcher/launcher.py
barakeinav1: I parsing the file as type=value. and not loading it a .env

tee-launcher/launcher.py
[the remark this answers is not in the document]
netrome: Nice catch!

tee-launcher/launcher.py
netrome: Discussing in other thread: https://github.com/near/mpc/pull/588#discussion_r2174627021

netrome (approved):

tee-launcher/launcher.py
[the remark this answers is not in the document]
pbeza: This code still doesn't solve the problem since I can just pass [`-e LD_PRELOAD=/path/to/my/malloc.so`](https://stackoverflow.com/questions/426230/what-is-the-ld-preload-trick) as one of the `EXTRA_HOSTS` entries:

https://github.com/near/mpc/blob/a1ae563d1056f5b85cf21895978ee1db6aa64c5d/tee_launcher/launcher.py#L482-L486

netrome (changes requested):
As @pbeza pointed out, the current regexes still allow injection of malicious args.

tee-launcher/launcher.py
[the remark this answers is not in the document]
barakeinav1: @pbeza I add spesific validaiton that there is no "-" at the start of a host name. and also added test for such input in the pytests.
@netrome I took a look at the URLlib, I don't think is is very helpful. the problem is that domain name can start with -. so I added explicit check for it.

pbeza:
Did you add the tests you wrote to the CI workflow? I can't find it.

tee_launcher/launcher.py
```diff
@@ -0,0 +1,545 @@
...
+
+    raise Exception("Image hash not found among tags.")
+
+
+def build_docker_cmd(user_env: dict[str, str], image_digest: str) -> list[str]:
+    docker_cmd = ["docker", "run"]
+
+    for key, value in user_env.items():
+        if key in ALLOWED_ENV_VARS:
+            docker_cmd += ["--env", f"{key}={value}"]
```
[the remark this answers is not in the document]
netrome: Resolving this and we can continue the discussion in the other thread.

tee_launcher/test_launcher_config.py
```diff
@@ -0,0 +1,304 @@
...
+    cmd = build_docker_cmd(env, "sha256:abc123")
+
+    assert "2200:2200" in cmd
+    assert not any("/:/mnt" in arg for arg in cmd)
+
+
+def test_invalid_env_key_is_ignored():
+    env = {
+        "BAD_KEY": "should_not_be_used",
+        "MPC_ACCOUNT_ID": "safe",
```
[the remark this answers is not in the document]
netrome: > ChatGPT [claims](https://chatgpt.com/share/686bba4b-b628-800e-93a7-5e2c3f219cfc) we're safe.

From what I can see we're not safe. Since we do 
```python
            docker_cmd += ["--env", f"{key}={value}"]
```

if we don't validate the "value" (which I don't see anywhere) you should be able to pass in arbitrary extra env variables here by just embedding them in the value.

netrome (changes requested):
Can you add instructions for how to run the tests? I'm running `pytest` from the `tee_launcher` directory and it doesn't work. However `PYTHONPATH=. pytest` works but it feels hacky to have to manually modify the python path to have the tests working.

Also the root Makefile seems to have been copied into this folder in https://github.com/near/mpc/pull/524 - please remove it.

netrome:
> Added a test case that fails, please update the code so that we don't have this vulnerability (unless I'm missing why this isn't a problem). Also, I don't see these tests being executed in our CI, we should add them.

I see @pbeza already commented this https://github.com/near/mpc/pull/588#issuecomment-3044308171

tee_launcher/test_launcher_config.py
```diff
@@ -0,0 +1,304 @@
...
+    assert "--env" in cmd
+    assert "MPC_ACCOUNT_ID=mpc-user-123" in cmd
+    assert "-p" in cmd
+    assert "11780:11780" in cmd
+    assert "--add-host" in cmd
+    assert "node:192.168.1.1" in cmd
+
+    # Make sure injection strings were filtered
+    assert not any("BAD=1" in arg for arg in cmd)
+    assert not any("/:/mnt" in arg for arg in cmd)
```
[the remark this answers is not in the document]
barakeinav1: according to chat GPT this in not correct.

here is some code to prove he is right:

cmd = [
    "--env", 
    "MPC_ACCOUNT_ID=test",
    "--env", 
    "PORTS=11780:11780,--env BAD=1"
]

print("List of arguments:")
for arg in cmd:
    print("  ", arg)

print("\n1. Using 'BAD=1' not in cmd:")
if "BAD=1" not in cmd:
    print("✅ PASS: 'BAD=1' is NOT an exact element in cmd")
else:
    print("❌ FAIL: 'BAD=1' IS an exact element in cmd")

print("\n2. Using any('BAD=1' in arg for arg in cmd):")
if not any("BAD=1" in arg for arg in cmd):
    print("✅ PASS: 'BAD=1' is NOT found as a substring in any element")
else:
    print("❌ FAIL: 'BAD=1' IS found as a substring in at least one element")

print("\n3. Using 'BAD=1' not in ' '.join(cmd):")
joined_cmd = " ".join(cmd)
print("  Joined command string:\n  ", joined_cmd)
if "BAD=1" not in joined_cmd:
    print("✅ PASS: 'BAD=1' is NOT a substring in the full command string")
else:
    print("❌ FAIL: 'BAD=1' IS a substring in the full command string")

tee_launcher/launcher.py
```diff
@@ -0,0 +1,583 @@
...
+
+
+def is_safe_env_value(value: str) -> bool:
+    """
+    Ensures that an environment variable value does not contain dangerous substrings
+    like LD_PRELOAD which may be used for injection.
+    """
+    if not isinstance(value, str):
+        return False
+    return "LD_PRELOAD" not in value
```
netrome, line 97: Okay this feels oddly specific. I'm sure there is more to this validation. Aren't there other things we could do with the value that would be problematic as well?

However, this PR has been stuck for a long time so let's merge what we have now and do any subsequent security enhancements as follow-ups.

netrome (approved):
Still skeptical that just banning `LD_PRELOAD` is sufficient for the env var validation, but given how long this PR has been stuck for - let's get it merged as-is and make additional vulnerability spotting/fixing as follow-ups.

tee_launcher/launcher.py
barakeinav1: I agree, but I can't think of anything else right now.

pbeza:
I generally agree with @netrome’s [comment](https://github.com/near/mpc/pull/588#pullrequestreview-3016334185) that simply banning `LD_PRELOAD` probably isn’t sufficient for environment variable validation. But given how long this PR has been stuck and that we need to test it, let’s merge it as-is and handle any additional fixes in follow-up PRs. ✅ It would probably make sense to reimplement it in Rust at some point anyway.

pbeza (approved):
I generally agree with @netrome’s [comment](https://github.com/near/mpc/pull/588#pullrequestreview-3016334185) that simply banning `LD_PRELOAD` probably isn’t sufficient for environment variable validation. But given how long this PR has been stuck and that we need to test it, let’s merge it as-is and handle any additional fixes in follow-up PRs. :+1: It would probably make sense to reimplement it in Rust at some point anyway.

Issue #565, opened 2025-06-25
[Task] TEE local testability investigation


netrome:
### Background

As part of our audit push, we implemented a lot of code we weren't able to test in our current setup. To be able to continue develop in the code base, we should retain the ability to test as much of our code as possible.

Long term, the idea is to only support TEE enabled MPC nodes. To be able to develop and maintain these nodes, we must have the ability to simulate a local TEE environment.

However, currently we're lacking a plan for how to make our system testable. We need to answer the following questions.

* Which pieces of the system that are currently inherently reliant on the TEE environment?
* How does the MPC node code and the TEE environment interact?
* How can we create a controlled/simulated TEE environment to run the MPC node code on a normal laptop?

### Example test cases we should support
We should be able to write unit and integration tests that asserts that the system expects as expected under the following scenarios:

* An MPC node sumbits an invalid quote to the smart contract.
* An MPC node fails to send a remote attestation to the smart contract.
* An MPC node sends an invalid remote attestation to the smart contract.

### User Story

As a developer, I want to be able to test my code in a non-production environment to be able to identify and catch bugs before we ship them to mainnet.

### Acceptance Criteria

We have identified and documented the main challenges in testing the TEE enabled node code, allowing us to form an actionable plan forward for how to make this testable.

### Resources & Additional Notes

[Phala local development guide](https://docs.phala.network/dstack/local-development)

netrome:
Taking over this as @kevindeforth is on vacation. We urgently need a way to improve our TEE testability to uncover potential untested bugs such as [this one](https://github.com/near/mpc/pull/612).

PR #612, opened 2025-07-07
fix: create tokio enter guard for spawning TEE related tasks

ghost:
We have tokio spawn calls that are TEE feature gated in this function that panic as they are not called from a tokio context. 
This PR adds an [EnterGuard](https://docs.rs/tokio/latest/tokio/runtime/struct.EnterGuard.html) such that `tokio::spawn` can be called.

### DIFF
--- a/node/src/cli.rs
+++ b/node/src/cli.rs
@@ -211,6 +211,8 @@ impl StartCmd {
             .worker_threads(1)
             .build()?;
 
+        let _tokio_enter_guard = root_runtime.enter();
+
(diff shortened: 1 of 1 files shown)


netrome (changes requested):
Before merging I would like to have this code path tested so that the same bug cannot reoccur. If you don't have any ideas I'll present a proposal in the context of https://github.com/near/mpc/issues/565 within the hour and we can sync about this.

If there are strong reasons for merging this sooner, please create a follow-up for the test case and I can approve this PR.

netrome (approved):

PR #621, opened 2025-07-09
feat(tee): custom attestation module

### LINKED ISSUE #613: [Task] Scaffold custom attestation module
### Background

Following the investigation in https://github.com/near/mpc/issues/565#issuecomment-3048435180 we have decided to break out attestation verification logic to a separate module which supports both live TEE based attestation verification and generating and validating local non-TEE based attestations - useful for local tests.

As a first step, we should scaffold that module to ensure we have the right structure with a basic attestation type supporting both local and TEE-based verification. Additionally, we have unit tests covering both cases - with a hard-coded fixture for the TEE based case.

### User Story

As a MPC developer I want to be able to test as much of the code base as possible.

### Acceptance Criteria

There is a dedicated crate in our root workspace with the relevant structures we need for representing attestations.

_No response_

pbeza:
Fixes #613

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -111,7 +111,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn 2.0.103",
+ "syn 2.0.104",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


netrome:
Brain dump from our sync today:
image

Summary:
- The first PR should focus only on the attestation module. We'll follow up with the integration later.
- Most important is to get the core types right. From what we discussed we need at least:
  - An Attestation type containing quote, quote collateral and TCB info.
  - A validation function that takes:
    - Input: Attestation, allowed docker hashes, docker compose template.
    - Output: Whether or not the attestation is valid, a date time for how long it is valid and possibly additional information to allow it to be revoked in the future (could be added as a follow-up. Should only be added if we use it in the contract).

PR #632, opened 2025-07-11
feat(tee): custom attestation module

### LINKED ISSUE #613: [Task] Scaffold custom attestation module

pbeza:
Fixes #613


attestation/src/tcbinfo.rs
```diff
@@ -0,0 +1,4 @@
+/// Dstack event log, a.k.a. the TCB Info.
+pub(crate) struct TcbInfo {
+    tcb_info: String,
+}
```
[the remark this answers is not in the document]
pbeza: Good idea — it’s always better to have a dedicated type rather than using a generic one like `String`. 👍🏼

Fixed [here](https://github.com/near/mpc/pull/632/commits/7a60641c15e8cb92850b373b244506d01a17bafc). ✅

attestation/src/quote.rs
```diff
@@ -0,0 +1,14 @@
+use dcap_qvl::quote::Quote as DcapQuote;
+
+/// TEE Remote Attestation Quote that proves the participant's identity.
+pub(crate) struct Quote(pub(crate) DcapQuote);
```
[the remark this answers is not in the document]
pbeza: Fixed [here](https://github.com/near/mpc/pull/632/commits/002e26751a733de6c6c28f0e31f4c9e4ab1181ad) and [here](https://github.com/near/mpc/pull/632/commits/571a778e5d77a8f18bf704d1fc988a5eb5765b30). ✅

attestation/src/tee_authority.rs
```diff
@@ -0,0 +1,31 @@
...
+enum TeeAuthority {
+    Local(LocalTeeAuthorityConfig),
+    Dstack(DstackTeeAuthorityConfig),
+}
+
+impl TeeAuthority {
+    async fn generate_attestation(
+        &self,
+        _tls_public_key: &PublicKey,
+        _account_public_key: &PublicKey,
```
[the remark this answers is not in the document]
pbeza: Actually, the TEE Report is defined [here](https://github.com/Phala-Network/dcap-qvl/blob/a1b725dbeacc78125c7024e63e4d9a9293570a46/src/quote.rs#L338-L343), and `report_data` [here](https://github.com/Phala-Network/dcap-qvl/blob/a1b725dbeacc78125c7024e63e4d9a9293570a46/src/quote.rs#L217-L218). In our case, `report_data` essentially consists of:
1. the report data version
1. the TLS public key
1. the account public key

So it might make sense to define it as you suggested. 👍🏼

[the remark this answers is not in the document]
pbeza: Fixed [here](https://github.com/near/mpc/pull/632/commits/9da13483977bb6c89a3843675a9cdaa8e929b266). ✅

Note that some of the code is copy-pasted from Daniel’s implementation in [`node/src/tee/remote_attestation.rs`](https://github.com/near/mpc/blob/main/node/src/tee/remote_attestation.rs). I just extracted it for now, with the intention of integrating it properly into the existing codebase in a follow-up PR.

attestation/src/hash.rs
```diff
@@ -0,0 +1,35 @@
+use std::marker::PhantomData;
+
+pub struct Hash32<T> {
```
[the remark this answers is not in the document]
pbeza: Note that some of the code is copy-pasted from [`libs/chain-signatures/contract/src/tee/proposal.rs`](https://github.com/near/mpc/blob/main/libs/chain-signatures/contract/src/tee/proposal.rs). I’ve just extracted it for now, with the intention of properly integrating it into the existing codebase in a follow-up PR.

The use of `std::marker::PhantomData` [was suggested](https://github.com/near/mpc/pull/516#discussion_r2157595936) by @near-bookrock, so all credit goes to him. 👏🏼

attestation/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+// #![no_std]
```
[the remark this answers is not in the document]
pbeza: [Fixed](https://github.com/near/mpc/pull/632/commits/6ec3b98e90b7755cae9e902ec9ce5b1edd65d51d). ✅

pbeza:
@netrome @near-bookrock PTAL and let me know if you have any suggestions for improving the design of the core TEE primitives and interfaces in this PR. These will be integrated with the existing codebase in a follow-up PR.

For now, it's a standalone crate that isn’t used anywhere yet — so we’re allowing dead code for the time being:
https://github.com/near/mpc/blob/6ec3b98e90b7755cae9e902ec9ce5b1edd65d51d/attestation/src/lib.rs#L2

attestation/src/quote.rs
```diff
@@ -0,0 +1,14 @@
...
+
+impl Quote {
+    pub fn new(quote: DcapQuote) -> Self {
+        Self(quote)
+    }
+
+    pub fn get(&self) -> &DcapQuote {
+        &self.0
+    }
+}
```
near-bookrock, line 14: Nit: we could implement `From` and `Deref` instead with the [derive_more](https://docs.rs/derive_more/latest/derive_more/index.html) crate.

attestation/src/report_data.rs
```diff
@@ -0,0 +1,161 @@
...
+        let mut hasher = Sha3_384::new();
+        hasher.update(self.tls_public_key.key_data());
+        hasher.update(self.account_public_key.key_data());
+        hasher.finalize().into()
+    }
+}
+
+impl From<ReportData> for [u8; REPORT_DATA_SIZE] {
+    fn from(report_data: ReportData) -> Self {
+        report_data.to_bytes()
```
near-bookrock, line 80: Nit: reuse implementation below.
```suggestion
        (&report_data).into()
```

attestation/src/attestation.rs
```diff
@@ -0,0 +1,84 @@
...
+        _measurements: Measurements,
+        _public_key: PublicKey,
+    ) -> bool {
+        match context {
+            ValidationContext::Tee => {
+                todo!("Implement Docker image validation logic")
+            }
+            ValidationContext::Local => {
+                todo!("Implement local Docker image validation logic")
+            }
```
near-bookrock, line 81: I guess these will be part of a follow up PR right?

attestation/src/attestation.rs
```diff
@@ -0,0 +1,84 @@
...
+        } else {
+            // Default/zero measurements if TD10 report is not available
+            Self {
+                rt_mr0: [0; 48],
+                rt_mr1: [0; 48],
+                rt_mr2: [0; 48],
+                rt_td: [0; 48],
+            }
+        }
```
near-bookrock, line 46: I have a knoweldge gap with TEE. What does it mean if TD10 report is not available? Should this be a TryFrom insted that fails if it is not available?

attestation/src/report_data.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use near_crypto::{KeyType, SecretKey};
+
+    fn create_test_keys() -> (PublicKey, PublicKey) {
+        let tls_key = SecretKey::from_random(KeyType::ED25519).public_key();
+        let account_key = SecretKey::from_random(KeyType::ED25519).public_key();
```
near-bookrock, line 97: Could we have these not be random, or at least have a seed?

near-bookrock (approved):

attestation/src/attestation.rs
netrome: Was going to ask the same thing. I assume this is something we'd expect on our TDX machines.

attestation/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+#![no_std]
+#![allow(dead_code)]
```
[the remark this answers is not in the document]
netrome: Which warnings? I'd be interested to know since those are probably legitimate warnings in that case.

For example `Attestation::new` is neither public nor used anywhere. I'd expect this module to provide a convenient way to construct these as part of the external interface.

attestation/src/quote.rs
netrome: Agreed. `drive_more` could also help us with the hash newtype.

attestation/src/report_data.rs
```diff
@@ -0,0 +1,161 @@
...
+            tls_public_key,
+            account_public_key,
+        }
+    }
+
+    /// Generates the binary representation of the report data.
+    ///
+    /// Format:
+    /// [version(2 bytes big endian) || sha384(TLS pub key || account pub key) || zero padding]
+    pub fn to_bytes(&self) -> [u8; REPORT_DATA_SIZE] {
```
netrome, line 52: Nit: Perhaps this should be called `to_binary_representation`.

attestation/src/report_data.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+}
+
+impl From<ReportData> for [u8; REPORT_DATA_SIZE] {
+    fn from(report_data: ReportData) -> Self {
+        report_data.to_bytes()
+    }
+}
+
+impl From<&ReportData> for [u8; REPORT_DATA_SIZE] {
```
netrome, line 84: Thinking of the semantics here, this is not a plain type conversion as we're hashing data from the `ReportData` struct. So in this case I'd opt for not having a `From<...>` implementation. Unless this is important for anything in particular @pbeza?

attestation/src/tee_authority.rs
```diff
@@ -0,0 +1,25 @@
...
+impl TeeAuthority {
+    async fn generate_attestation(&self, _report_data: ReportData) -> Attestation {
+        match self {
+            TeeAuthority::Local(_config) => {
+                // Generate attestation using local TEE authority
+                todo!("Implement local TEE attestation generation")
+            }
+            TeeAuthority::Dstack(_config) => {
+                // Generate attestation using Dstack TEE authority
+                todo!("Implement Dstack TEE attestation generation")
```
netrome, line 21: Right here's the attestation generation logic. @pbeza can you create follow-up tasks for these?

attestation/src/attestation.rs
```diff
@@ -0,0 +1,84 @@
...
+
+impl Attestation {
+    fn verify_quote(&self, context: &ValidationContext) -> bool {
+        match context {
+            ValidationContext::Tee => {
+                todo!("Implement TEE validation logic")
+            }
+            ValidationContext::Local => {
+                todo!("Implement local attestation validation logic")
+            }
```
netrome, line 63: And let's create follow-up tasks for these as well.

attestation/src/attestation.rs
netrome: And these.

netrome (changes requested):
I think this is a good start. The only blockers from my side is:

1. For the missing attestation generation logic, and validation logic - let's create issues and reference these in TODO comments.

2. The dead code lint. If there's anything specific that's dead code due to e.g. todos, can we make it more specific? E.g. the fields in the `Attestation` struct are dead code since they are intended for validation. We can add the `#[allow(dead_code)]` on this struct and reference the issue when these will be used in this case.

On the other hand, `TeeAuthority::generate_attestation` counts as dead code but should be part of the public interface of this crate. Removing this lint from the root of the crate allows us to distinguish between what's left to do and what we should update now.

attestation/src/attestation.rs
pbeza: > I guess these will be part of a follow up PR right?
Yes: https://github.com/near/mpc/issues/643

attestation/src/report_data.rs
pbeza: Removed [here](https://github.com/near/mpc/pull/632/commits/e7bb4ef4379bfe5519b8c19a322b43c8c0372c6c). ✅

There’s already a `to_bytes` implementation — the `From<...>` is just a wrapper around it, so I’m fine with removing it.

https://github.com/near/mpc/blob/e7bb4ef4379bfe5519b8c19a322b43c8c0372c6c/attestation/src/report_data.rs#L48-L67

attestation/src/report_data.rs
pbeza: Good point but I [decided to remove it altogether](https://github.com/near/mpc/pull/632/commits/e7bb4ef4379bfe5519b8c19a322b43c8c0372c6c) as @netrome suggested [here](https://github.com/near/mpc/pull/632#discussion_r2206682074).

pbeza:
@netrome please re-review. I'm particularly interested in whether I addressed [this](https://github.com/near/mpc/pull/632#discussion_r2207684523) the way you expected.

attestation/src/attestation.rs
```diff
@@ -48,13 +51,15 @@ impl TryFrom<VerifiedReport> for Measurements {
     }
 }
 
+#[allow(dead_code)]
 enum ValidationContext {
```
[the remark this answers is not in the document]
pbeza: [Fixed](https://github.com/near/mpc/pull/632/commits/7f79689be93584fa6ad75a37d2e40a262d9711d1). ✅

attestation/src/tee_authority.rs
```diff
@@ -4,13 +4,15 @@ pub struct LocalTeeAuthorityConfig;
...
+#[allow(dead_code)]
 enum TeeAuthority {
     Local(LocalTeeAuthorityConfig),
     Dstack(DstackTeeAuthorityConfig),
 }
 
 impl TeeAuthority {
     // TODO(#644): Implement the attestation quote generation logic in the attestation module
+    #[allow(dead_code)]
     async fn generate_attestation(&self, _report_data: ReportData) -> Attestation {
```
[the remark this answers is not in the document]
pbeza: [Fixed](https://github.com/near/mpc/pull/632/commits/7f79689be93584fa6ad75a37d2e40a262d9711d1). ✅

Issue #638, opened 2025-07-14
add P2P key to the http api endpoint that exposes public key.


barakeinav1:
PR https://github.com/near/mpc/pull/445/files#diff-6eab3169730c0d85b9d5933dd86f2ea2953aeb449648295f0a2d22e3a097f845

generated and exposes http api endpoint with public counter parts of the generated keys.
put the P2P key was not added to the API endpoit.

PR #639, opened 2025-07-14
feat(tee): Add p2p public key to StaticWebData and http endpoint

barakeinav1:
Fixes https://github.com/near/mpc/issues/638

### DIFF
--- a/node/src/web.rs
+++ b/node/src/web.rs
@@ -175,10 +175,10 @@ pub async fn start_web_server(
 ) -> anyhow::Result<BoxFuture<'static, anyhow::Result<()>>> {
     use futures::FutureExt;
 
-    tracing::debug!(
-        "Attempting to bind web server to host: {}, port: {}",
-        config.host,
-        config.port
+    tracing::info!(
+        host = %config.host,
+        port = %config.port,
+        "Attempting to bind web server to host",
     );
 
     let router = axum::Router::new()
--- a/node/src/web.rs
+++ b/node/src/web.rs
@@ -121,11 +121,7 @@ pub struct StaticWebData {
 
 fn get_public_keys(
     secrets_config: &SecretsConfig,
-) -> (
-    near_crypto::PublicKey,
-    near_crypto::PublicKey,
-    Vec<near_crypto::PublicKey>,
-) {
+) -> PublicKeys {
     let near_signer_public_key = secrets_config
         .persistent_secrets
         .near_signer_key
--- a/node/src/web.rs
+++ b/node/src/web.rs
@@ -114,16 +114,21 @@ async fn third_party_licenses() -> Html<&'static str> {
 #[derive(Clone, Serialize)]
 pub struct StaticWebData {
     pub near_signer_public_key: near_crypto::PublicKey,
+    pub near_p2p_public_key: near_crypto::PublicKey,
     pub near_responder_public_keys: Vec<near_crypto::PublicKey>,
     pub tee_participant_info: Option<TeeParticipantInfo>,
 }
 
 fn get_public_keys(
     secrets_config: &SecretsConfig,
-) -> (near_crypto::PublicKey, Vec<near_crypto::PublicKey>) {
+) -> (near_crypto::PublicKey,near_crypto::PublicKey, Vec<near_crypto::PublicKey>) {
     let near_signer_public_key = secrets_config
         .persistent_secrets
         .near_signer_key
+        .public_key();
+    let near_p2p_public_key = secrets_config
+        .persistent_secrets
+        .p2p_private_key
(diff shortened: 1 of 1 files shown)


node/src/web.rs
```diff
@@ -114,31 +114,46 @@ async fn third_party_licenses() -> Html<&'static str> {
...
 }
 
 fn get_public_keys(
     secrets_config: &SecretsConfig,
-) -> (near_crypto::PublicKey, Vec<near_crypto::PublicKey>) {
+) -> (
+    near_crypto::PublicKey,
+    near_crypto::PublicKey,
+    Vec<near_crypto::PublicKey>,
+) {
```
near-bookrock, line 128:
```suggestion
) -> PublicKeys {
```

I think a struct would be better to return here over a tuple. It will make it less likely we mix up which key is what.
```rust
PublicKeys {
    near_signer_public_key: near_crypto::PublicKey,
    near_p2p_public_key: near_crypto::PublicKey,
    near_responder_public_keys: Vec<near_crypto::PublicKey>,
}
```

node/src/web.rs
```diff
@@ -180,9 +202,17 @@ pub async fn start_web_server(
             static_web_data,
         });
 
-    let tcp_listener = TcpListener::bind(&format!("{}:{}", config.host, config.port)).await?;
+    // Binding to the address
+    let bind_address = format!("{}:{}", config.host, config.port);
```
near-bookrock, line 206: Thanks for this refactor. I also prefer creating variables with descriptive names for the values we work with.

node/src/web.rs
```diff
@@ -158,12 +173,19 @@ async fn get_public_data(state: State<WebServerState>) -> Json<StaticWebData> {
...
     contract_state_receiver: watch::Receiver<ProtocolContractState>,
 ) -> anyhow::Result<BoxFuture<'static, anyhow::Result<()>>> {
     use futures::FutureExt;
 
+    // Print out the config parameters before attempting to bind
+    tracing::debug!(
+        "Attempting to bind web server to host: {}, port: {}",
+        config.host,
+        config.port
+    );
```
near-bookrock, line 187: Consider using structured logging here instead of string interpolation. Structured fields make logs more searchable and filterable in log aggregation systems. You'll be able to query for specific hosts or ports without parsing the message string.

https://docs.rs/tracing/latest/tracing/#recording-fields

```suggestion
    tracing::info!(
        host = %config.host,
        port = %config.port,
        "Attempting to bind web server to host",
    );
```

node/src/web.rs
```diff
@@ -180,9 +200,19 @@ pub async fn start_web_server(
...
 
-    let tcp_listener = TcpListener::bind(&format!("{}:{}", config.host, config.port)).await?;
+    // Binding to the address
+    let bind_address = format!("{}:{}", config.host, config.port);
+
+    tracing::info!(address = %bind_address, "Binding to address");
+
+    let tcp_listener = TcpListener::bind(&bind_address).await?;
+
+    tracing::info!("Successfully bound to address: {}", bind_address);
```
near-bookrock, line 210: Can you make the bind_address a structured field like the log above with port and host?

node/src/web.rs
```diff
@@ -180,9 +200,19 @@ pub async fn start_web_server(
             static_web_data,
         });
 
-    let tcp_listener = TcpListener::bind(&format!("{}:{}", config.host, config.port)).await?;
+    // Binding to the address
```
near-bookrock, line 203: Rmove comment.

node/src/web.rs
```diff
@@ -158,12 +172,18 @@ async fn get_public_data(state: State<WebServerState>) -> Json<StaticWebData> {
 pub async fn start_web_server(
     root_task_handle: Arc<crate::tracking::TaskHandle>,
     signature_debug_request_sender: broadcast::Sender<SignatureDebugRequest>,
-    config: WebUIConfig,
+    config: WebUIConfig, // assuming this contains host and port
```
near-bookrock, line 175: Remove comment. Is this from ChatGPT? It's fine if it is, but let's try to clean up the comments it outputs as they are often obvious.

near-bookrock (approved):

node/src/web.rs
near-bookrock: https://github.com/near/mpc/pull/639#discussion_r2206976442

node/src/web.rs
barakeinav1: yes- missed this one.

node/src/web.rs
barakeinav1: no sure what do you mean.
feel free to use git suggestion. or paste something here

PR #680, opened 2025-07-20
fix: Adjust invariant checks when selecting participants

kuksag:
## Context
The function: 
```
pub fn select_random_active_participants_including_me(
        &self,
        total: usize,
        peers_to_consider: &[ParticipantId],
    ) -> anyhow::Result<Vec<ParticipantId>>
```
Should return `total` number of online participants amongst `peers_to_consider`, including the caller participant. 

This function has two usages:
1. In generation of triples
2. In generation eddsa signatures

## Bug
This function behaves incorrectly, when the size of `peers_to_consider` is insufficient, to make up for `total` participants.
Extrapolated example: if we pass empty `peers_to_consider`. In that case `[me]` will be returned, because `me` is getting appended regardless. The error should be returned instead.

`IteratorRandom::choose_multiple(_, amount)` silently returns less than `amount` elements, if there are not enough.

## Solution

Make invariant check on return value

## Effect 

This has no practical effect at the moment, because `all_participants_from_config` is being passed into `select_random_active_participants_including_me`

### DIFF
--- a/node/src/network.rs
+++ b/node/src/network.rs
@@ -1095,7 +1095,7 @@ mod tests {
     }
 
     #[test]
-    fn foo() {
+    fn test_random_participant_selection() {
         let num_participants = 4;
         let participant_ids =
             TestGenerators::new(num_participants, num_participants).participant_ids();
--- a/node/src/network.rs
+++ b/node/src/network.rs
@@ -1095,7 +1095,7 @@ mod tests {
     }
 
     #[test]
-    fn test_random_participant_selection() {
+    fn select_random_active_participants_including_me_should_return_error_when_peers_to_consider_is_empty() {
         let num_participants = 4;
         let participant_ids =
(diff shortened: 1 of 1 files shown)


node/src/network.rs
```diff
@@ -1090,6 +1093,30 @@ mod tests {
     struct TestTripleMessage {
         data: u64,
     }
+
+    #[test]
+    fn foo() {
```
near-bookrock, line 1098: Maybe a leftover, but could you put a descriptive name for the test and a doc comment on what is being tested?

near-bookrock (changes requested):

node/src/network.rs
kuksag: oops, yeah that's a leftover

near-bookrock (approved):

netrome:
I assume this failed due to main being broken for a brief time. Taking the liberty to add it to the merge queue again.

PR #682, opened 2025-07-21
test: create pytest to test cancellation of key resharing

### LINKED ISSUE #664: [Task] Create pytest to test cancellation of key resharing
### Background

Follow up from #500 that adds the ability cancel key resharing we need a system test, pytest, that tests key resharing cancellation works as expected.

### User Story

.

### Acceptance Criteria

We have a pytest that tests cancellation.

What needs to be tested:
1. Cancellation of resharing works.
2. State transitions to running.
3. MPC network can serve sign requests as expected.
4. Attempt a new resharing and assert that the old epoch id is not reused.
3. MPC network can serve sign requests as expected with the reshared key shares.

_No response_

near-bookrock:
closes #664

### DIFF
--- a/pytest/common_lib/shared/__init__.py
+++ b/pytest/common_lib/shared/__init__.py
@@ -132,6 +132,7 @@ def generate_mpc_configs(
     num_mpc_nodes: int,
     num_respond_aks: int,
     presignatures_to_buffer: Optional[int],
+    threshold: int,
 ) -> List[Candidate]:
     """
     Generate MPC configs for each participant.
@@ -154,7 +155,7 @@ def generate_mpc_configs(
         "--responders",
         ",".join(responders),
         "--threshold",
-        str(num_mpc_nodes),
+        str(threshold),
         "--desired-responder-keys-per-participant",
         str(num_respond_aks),
     )
@@ -239,16 +240,18 @@ def start_cluster_with_mpc(
     contract,
     presignatures_to_buffer=None,
     start_mpc_nodes=True,
+    threshold=None,
 ):
     validators, observers = start_neard_cluster_with_cleanup(
         num_validators,
         num_mpc_nodes,
     )
 
+    if threshold is None:
+        threshold = num_mpc_nodes
+
     candidates = generate_mpc_configs(
-        num_mpc_nodes,
-        num_respond_aks,
-        presignatures_to_buffer,
+        num_mpc_nodes, num_respond_aks, presignatures_to_buffer, threshold=threshold
     )
 
     move_mpc_configs(observers)
--- a/pytest/common_lib/shared/mpc_cluster.py
+++ b/pytest/common_lib/shared/mpc_cluster.py
@@ -322,22 +322,6 @@ class MpcCluster:
             assert self.wait_for_state("Running"), "failed to conclude resharing"
             self.update_participant_status()
 
-    def do_cancellation(self, new_participants: List[MpcNode]):
-        print(f"\033[91mVoting on cancellation of resharing\033[0m")
-        state = self.contract_state()
-        assert state.is_state(
-            "Resharing"
-        ), "Require resharing state to do cancellation of key resharing."
-
-        args = {}
-        for node in self.get_voters():
-            tx = node.sign_tx(
-                self.mpc_contract_account(), "vote_cancel_resharing", args
-            )
-            node.send_txn_and_check_success(tx)
-
-        assert self.wait_for_state("Running"), "failed to start resharing"
-
     def get_contract_state(self):
         cn = self.contract_node
         txn = cn.sign_tx(self.mpc_contract_account(), "state", {})
--- a/pytest/common_lib/shared/mpc_cluster.py
+++ b/pytest/common_lib/shared/mpc_cluster.py
@@ -322,6 +322,22 @@ class MpcCluster:
             assert self.wait_for_state("Running"), "failed to conclude resharing"
             self.update_participant_status()
 
+    def do_cancellation(self, new_participants: List[MpcNode]):
+        print(f"\033[91mVoting on cancellation of resharing\033[0m")
+        state = self.contract_state()
+        assert state.is_state(
+            "Resharing"
+        ), "Require resharing state to do cancellation of key resharing."
+
+        args = {}
+        for node in self.get_voters():
+            tx = node.sign_tx(
+                self.mpc_contract_account(), "vote_cancel_resharing", args
+            )
+            node.send_txn_and_check_success(tx)
+
+        assert self.wait_for_state("Running"), "failed to start resharing"
+
     def get_contract_state(self):
         cn = self.contract_node
         txn = cn.sign_tx(self.mpc_contract_account(), "state", {})
--- a/pytest/common_lib/shared/near_account.py
+++ b/pytest/common_lib/shared/near_account.py
@@ -21,6 +21,52 @@ def assert_txn_success(res):
     )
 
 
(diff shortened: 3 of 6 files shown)


pytest/common_lib/shared/mpc_cluster.py
```diff
@@ -322,6 +322,22 @@ def do_resharing(
...
+        ), "Require resharing state to do cancellation of key resharing."
+
+        args = {}
+        for node in self.get_voters():
+            tx = node.sign_tx(
+                self.mpc_contract_account(), "vote_cancel_resharing", args
+            )
+            node.send_txn_and_check_success(tx)
+
+        assert self.wait_for_state("Running"), "failed to start resharing"
```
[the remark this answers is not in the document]
netrome: Hmm we should probably add lints to our CI to capture unused arguments. Perhaps [ruff](https://github.com/astral-sh/ruff) would have caught this. I just noticed it could replace black as we just introduced for formatting as well. :sweat_smile:

pytest/tests/test_cancellation_of_key_resharing.py
```diff
@@ -0,0 +1,89 @@
...
+    # Two new nodes join, increase threshold to 3
+    cluster.do_resharing(
+        all_participants,
+        new_threshold=3,
+        prospective_epoch_id=initial_prospective_epoch_id,
+        wait_for_running=False,
+    )
+
+    # Kill one of the new nodes to make resharing stall
+    mpc_nodes[3].kill()
```
netrome, line 42: I'm sad we don't have a better way to control the system to stay in the resharing state. Additionally I assume this is an inherent race condition if resharing somehow happened to complete before we kill this node.

Anyway, I don't see any easy way to resolve this.

netrome (approved):
There's a key error in CI that needs to be addressed. Seems to be related to the new field you added to the `RunningProtocolState`. Beyond that I agree with Kevin's comments, but none are hard blockers. Otherwise looks good to me :ok_hand:

pytest/tests/test_cancellation_of_key_resharing.py
```diff
@@ -0,0 +1,89 @@
...
+
+sys.path.append(str(pathlib.Path(__file__).resolve().parents[1]))
+from common_lib import shared
+from common_lib.contracts import load_mpc_contract
+from common_lib.contract_state import (
+    RunningProtocolState,
+)
+
+
+def test_cancellation_of_key_resharing():
```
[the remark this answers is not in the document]
kevindeforth: @netrome  I did not spend too much time, as I thought it would be generally difficult to do so correctly. Depending on how we do it, it might end up to the equivalent of having one giant test.

The advantage of the current set-up is that with some tweaks, we could run theses tests in parallel. For a beefy ci, this shouldn't be an issue.

pytest/common_lib/shared/mpc_cluster.py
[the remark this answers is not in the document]
near-bookrock: Pushed some changes, and inlined the voting for cancellation to adress the concerns. We now check
- Votes by new participants to cancel are rejected.
- Only threshold votes are required to cancel a resharing.

https://github.com/near/mpc/blob/e535fdc8aa832a3fb1da3168cc1677ffb14aea0a/pytest/tests/test_cancellation_of_key_resharing.py#L65-L85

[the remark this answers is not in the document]
near-bookrock: > Hmm we should probably add lints to our CI to capture unused arguments. Perhaps [ruff](https://github.com/astral-sh/ruff) would have caught this. I just noticed it could replace black as we just introduced for formatting as well. 😅

Very good point.

Should be an easy migration from the good docs they have. I obviously won't prioritise working on that now, but created an issue so we remember at some point  https://github.com/near/mpc/issues/689

pytest/common_lib/shared/__init__.py
```diff
@@ -154,7 +155,7 @@ def generate_mpc_configs(
         "--responders",
         ",".join(responders),
         "--threshold",
-        str(num_mpc_nodes),
+        str(threshold),
```
kevindeforth, line 158: I was trying to wrap my head around this parameter. I don't think this _should_ have any effect on pytests, because the nodes are supposed to fetch that information from the contract, not from their locally stored config.

This seems like an artifact from a previous iteration of these tests. If I am not mistaken, then the `participants.json` file is only relevant for our rust tests.

pytest/common_lib/shared/__init__.py
```diff
@@ -239,16 +240,18 @@ def start_cluster_with_mpc(
     contract,
     presignatures_to_buffer=None,
     start_mpc_nodes=True,
+    threshold=None,
```
kevindeforth, line 243: So, did the tests not pass before this change?

pytest/tests/test_cancellation_of_key_resharing.py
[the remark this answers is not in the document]
kevindeforth: I got the following timings:
```
total time elapsed: 134.13s 
1. time for setup (until line 49): 60.38s
2. time for testing (critical section between 50 and 110): 30.26s
3. time for final resharing (111 until end):  42.77s
```
integrating this with the other existing test would save 1 + 3, so roughly 100s, as we could sandwich it between existing resharing tests.

pytest/common_lib/shared/__init__.py
near-bookrock: I'm not sure I understand.

IIUC the config generated in this function is what's used when intializing the contract in `start_cluster_with_mpc`, which is why I updated it to include a threshold parameter to be able to customize the threshold.

pytest/common_lib/shared/near_account.py
```diff
@@ -21,6 +21,52 @@ def assert_txn_success(res):
     )
 
 
+def assert_txn_execution_error(res, expected_error_msg=None):
```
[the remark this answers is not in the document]
near-bookrock: Good point. Done.

pytest/tests/test_cancellation_of_key_resharing.py
near-bookrock: That's interesting.

Do we have anything in place already to communicate to the node to dynamically adjust it's behaviour during runtime? On top of my head I can imagine doing it through the debug endpoint to submit requests to alter it's behavior,.

near-bookrock: I'm asking because at some point we revive the node such that it can complete resharing.

pytest/tests/test_cancellation_of_key_resharing.py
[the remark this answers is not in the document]
near-bookrock: I created a tracking issue https://github.com/near/mpc/issues/691

pytest/common_lib/shared/__init__.py
kevindeforth: That takes place in `init_cluster`, see also this comment: https://github.com/near/mpc/pull/682#discussion_r2222356401

pytest/tests/test_cancellation_of_key_resharing.py
kevindeforth: > Do we have anything in place already to communicate to the node to dynamically adjust it's behaviour during runtime?

We actually support that! https://github.com/near/mpc/blob/9dd470376c3c3405690e8db1ca3e450ddcd6e762/pytest/common_lib/shared/mpc_node.py#L147

PR #685, opened 2025-07-21
feat: Add TEE Launcher image building step to CI

kuksag:


kuksag:
To be honest I don't know how to test before landing onto main.
I tried to do it in the separate PR with trigger on every PR: https://github.com/near/mpc/pull/687

"Login to Docker Hub" step failed for me, which is kinda make sense – I don't and I should not have an access to the github secrets while making a PR from outside. But I'm not sure if that's going to work on `main` then. 
@andrei-near any thoughts?

netrome:
> To be honest I don't know how to test before landing onto main.

Yeah afaik there's still no good way to run Github Actions locally, which bothers me every time I see a PR like this. But I think we'll have to do the yolo approach of just seeing if this works on main and patch it if it doesn't.

.github/workflows/tee_launcher.yml
```diff
@@ -0,0 +1,60 @@
...
+on:
+  push:
+    branches:
+      - main
+
+  workflow_dispatch:
+    inputs:
+      build-ref:
+        default: 'main'
+        description: "The branch, tag or SHA to build MPC Docker image from. Default to latest commit on main branch."
```
netrome, line 16:
```suggestion
        description: "The branch, tag or SHA to build the launcher Docker image from. Default to latest commit on main branch."
```

netrome (approved):
Just a typo

.github/workflows/tee_launcher.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push Tee Launcher image to Docker Hub
+        uses: Warpbuilds/build-push-action@v6
+        with:
+          context: tee_launcher/
```
andrei-near, line 56: you need to either use context or full Dockerfile path in file argument. I think this will error when both context and file contain tee_launcher. 
No way to test without merging first

kuksag: That should be fine, I tested this behaviour locally

PR #688, opened 2025-07-21
chore: Add `tee_launcher` tests in CI

### LINKED ISSUE #636: Add the Launcher pytests to CI
there is currently several pytests used to test possible and native test for the launcher. that were run manually.
need to add them to CI. 

relevant test are found under:

tee_launcher/test_launcher_config.py

kuksag:
fixes #636


.github/workflows/ci.yml
```diff
@@ -145,3 +145,33 @@ jobs:
...
+
+      - name: Setup python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+
+      - name: Setup virtualenv
+        run: |
+          python3 -m venv pytest/venv
+          source pytest/venv/bin/activate
```
kevindeforth, line 168: just to be a bit more explicit, I would prefer if we didn't store this venv in the pytest folder for now (even though it shouldn't make a difference here).
```suggestion
          python3 -m venv tee_launcher/venv
          source tee_launcher/venv/bin/activate
```

.github/workflows/ci.yml
```diff
@@ -145,3 +145,33 @@ jobs:
...
+          cd tee_launcher 
+          pip install -r requirements.txt
+
+      - name: Run pytest
+        run: |
+          python3 -m venv pytest/venv
+          source pytest/venv/bin/activate
```
kevindeforth, line 175:
```suggestion
          python3 -m venv tee_launcher/venv
          source tee_launcher/venv/bin/activate
```

.github/workflows/ci.yml
```diff
@@ -145,3 +145,33 @@ jobs:
...
+          cd tee_launcher 
+          pytest test_launcher_config.py -v -s -x
```
kevindeforth, line 177: what about `test_launcher.py`?

netrome: Also, just running `pytest` should be sufficient since it then discovers all test files.

When I run this locally I need to set `PYTHONPATH=.` for `test_launcher` to succeed. 

```suggestion
          PYTHONPATH=. pytest -vsx
```

netrome (changes requested):
Mostly looks good to me, but we should also include the `test_launcher` tests as Kevin suggested. I added a suggestion that I think works but feel free to do it any way that makes sense to you.

.github/workflows/ci.yml
kuksag: @netrome do you run under with env activated? (`source venv/bin/activate`), because after that step and `pip install requirements.txt` it should be available in path

PR #708, opened 2025-07-24
chore: rename web endpoint /get_public_data to /public_data

kevindeforth:
resolves https://github.com/near/mpc/issues/709
This should not yet be a breaking change, so it should be pretty safe to do.


netrome (approved):
Thanks you!

Issue #641, opened 2025-07-15
[pytest] Investigate test_key_event.py::test_single_domain flakiness


gilcu3:
### Background

Running test `test_key_event.py::test_single_domain` on CI, this test's result sometimes changes fails without reason.

### User Story

Tests need to have deterministic results to make them trustworthy.

### Acceptance Criteria

Make sure `test_key_event.py::test_single_domain` passes deterministically.

### Resources & Additional Notes

Example failures:
- https://github.com/near/mpc/actions/runs/16286595604/job/45986546229
- https://github.com/near/mpc/actions/runs/16934426117/job/47987506250
- https://github.com/near/mpc/actions/runs/16962633037/job/48078864110
- https://github.com/near/mpc/actions/runs/17074077165/job/48410630014
- https://github.com/near/mpc/actions/runs/17120621617/job/48560393701?pr=885
- https://github.com/near/mpc/actions/runs/17208673174/job/48814906590
- https://github.com/near/mpc/actions/runs/20004461760/job/57367984507
- https://github.com/near/mpc/actions/runs/20021792975/job/57410066923
- https://github.com/near/mpc/actions/runs/21664559303/job/62456669013
```
        assert self.wait_for_state(ProtocolState.RESHARING), "failed to start resharing"
        if wait_for_running:
>           assert self.wait_for_state(ProtocolState.RUNNING), (
                "failed to conclude resharing"
            )
E           AssertionError: failed to conclude resharing

common_lib/shared/mpc_cluster.py:352: AssertionError
=========================== short test summary info ============================
FAILED tests/test_key_event.py::test_single_domain - AssertionError: failed to conclude resharing
```

netrome:
FYI @anodar - let's make sure to make it more robust when we port it to Rust.

anodar:
Closing this as duplicate in favor of https://github.com/near/mpc/issues/2898. Stress testing all tests there.

PR #713, opened 2025-07-24
chore: added rerun for flaky test

gilcu3:
This mitigates #641 . We were installing `pytest-rerunfailures` already in the virtualenv but were not using it.

The reason I am opening this now is that this test failed again for me, this time locally.


kevindeforth (changes requested):
I would prefer if we focused on resolving the underlying cause, instead of adding automatic reruns.

gilcu3:
Makes sense, thanks for the explanation

Issue #712, opened 2025-07-24
Display `responder_account_id` and `mpc_account_id` on public data endpoint


kevindeforth:
### Background

The public_data endpoint displays the responder keys and p2p public key generated by the mpc node.

### User Story

As a developer working on devnet, I want to verify that I am adding the keys to the correct account id.

### Acceptance Criteria

The node should also display the `responder_account_id` and `mpc_account_id` on this endpoint. The information is parsed and used by devnet, instead of relying on the indices.

netrome:
Note that this account is still visible on normal debug endpoints - so this is primarily for the devnet CLI.

netrome:
We already serve these on the node debug/node_config page. Isn't that useful for the devnet CLI? Closing this for now and we can re-open if we have a use-case for it.

Issue #710, opened 2025-07-24
Devnet cleanup after 3.0.0 release.


kevindeforth:
### Background

Account keys used to be generated by devnet, now (version 3.0.0 and newer) they are generated inside by node.
This changes the workflow. Devnet supports both options since <insert PR number>. We can stop supporting the previous behavior after 3.0.0 is released.

### User Story

As a developer, I want a clean code base.

### Acceptance Criteria

Remove all artifacts related to this issue in devnet codebase (c.f. https://github.com/near/mpc/pull/715), as well as the `legacy` artifacts introduced in https://github.com/Near-One/infra-ops/pull/503

PR #715, opened 2025-07-24
fix(devnet): compatibility with upcoming mpc node version 3.0.0

### LINKED ISSUE #521: [Task] Update devnet to work with new key setup
### Background

In https://github.com/near/mpc/pull/445 we changed the p2p and near signer keys to be generated within the MPC nodes rather than provided through args/CLI arguments.

However, this breaks the current devnet configuration. This task is a follow-up to update so our devnet can be set up by requesting the relevant public keys from the signer HTTP API.

### User Story

As a developer I want to be able to set up a local devnet.

### Acceptance Criteria

The devnet setup works.

Meeting notes from discussion about this: [link](https://docs.google.com/document/d/1oQWoHvi5p1FvnNfCOXubvxi2beKL6zN8pOj-SO22gGI/)

kevindeforth:
resolves #521 
requires https://github.com/Near-One/infra-ops/pull/503
related follow-ups:
- https://github.com/near/mpc/issues/710
- https://github.com/near/mpc/issues/712

changes:
- `deploy-contract` command is now split into two commands `deploy-contract` (deploy the code) and `init-contract` (initialize the contract).
- `add-keys` method, adding the public keys generated by the nodes to the mpc and responder accounts (signer and responder keys).
- unused `--max-requests-to-remove` argument has been removed from the contract commands
- devnet no longer panics upon calling `describe` in case the contract is not initialized:
```
thread 'main' panicked at devnet/src/mpc.rs:812:10:
Expected rpc node to respond: ServerError(HandlerError(ContractExecutionError { vm_error: "wasm execution failed with error: HostError(GuestPanic { panic_msg: \"Calling default not allowed.\" })", block_height: 206548722, block_hash: 2LPrNt4BEv2MU6GVRRCZLYXvD1dCUYFRymr5mXtdYWZC }))
```

### DIFF
--- a/devnet/src/mpc.rs
+++ b/devnet/src/mpc.rs
@@ -1,7 +1,7 @@
 #![allow(clippy::expect_fun_call)] // to reduce verbosity of expect calls
 use crate::account::{OperatingAccount, OperatingAccounts};
 use crate::cli::{
-    ListMpcCmd, MpcDeployContractCmd, MpcDescribeCmd, MpcInitContractCmd,
+    ListMpcCmd, MpcAddKeysCmd, MpcDeployContractCmd, MpcDescribeCmd, MpcInitContractCmd,
     MpcProposeUpdateContractCmd, MpcViewContractCmd, MpcVoteAddDomainsCmd, MpcVoteNewParametersCmd,
     MpcVoteUpdateCmd, NewMpcNetworkCmd, RemoveContractCmd, UpdateMpcNetworkCmd,
 };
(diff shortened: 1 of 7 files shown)


devnet/src/mpc.rs
```diff
@@ -199,6 +208,70 @@ impl UpdateMpcNetworkCmd {
     }
 }
 
+#[derive(Clone, Deserialize)]
+pub struct StaticWebData {
+    pub near_signer_public_key: near_crypto::PublicKey,
+    pub near_p2p_public_key: near_crypto::PublicKey,
+    pub near_responder_public_keys: Vec<near_crypto::PublicKey>,
+    pub _tee_participant_info: Option<TeeParticipantInfo>,
+}
```
netrome, line 217: Hmm, I wonder if we should consider reusing the existing `StaticWebData` struct from `node/src/web.rs` to prevent this from going stale. What do you think?

netrome (approved):
Thanks for fixing devnet :pray:

devnet/src/mpc.rs
netrome: Issue: https://github.com/near/mpc/issues/719
and PR: https://github.com/near/mpc/pull/720

PR #722, opened 2025-07-25
feat: added TEE enabled dockerfile + github workflow

### LINKED ISSUE #723: Add workflow to build and publish TEE enabled docker images
### Background

To test and use MPC node images published in dockerhub we need to have the option of using TEE enabled images.

### User Story

.

### Acceptance Criteria

A workflow to generate TEE enabled docker images is added and tested

_No response_

gilcu3:
Closes #723 

This is a PR to test publishing a TEE enabled image to dockerhub. If it works I will transform it in a proper PR.

### DIFF
--- a/.github/workflows/docker_release_tee.yml
+++ b/.github/workflows/docker_release_tee.yml
@@ -11,7 +11,6 @@ on:
         default: 'main'
         description: "The branch, tag or SHA to build MPC TEE enabled Docker image from. Default to latest commit on main branch."
         type: string
-  pull_request:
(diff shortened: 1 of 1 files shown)


gilcu3:
The workflow was successfully tested [here](https://github.com/near/mpc/actions/runs/16523369257). I couldn't test publishing all 3 images because for some reason sanitized_ref was not picked correctly. At the moment I think the new workflow will not be executed by default. Let me know if that's not the correct behavior.

PR #733, opened 2025-07-28
doc: added doc on how to deploy an MPC node within a TEE

### LINKED ISSUE #721: Document how to deploy an MPC node within a TEE
### Background

We've spent a lot of time on manual tests for MPC nodes within a TEE, but most of the team is still not capable of reproducing this setup. While we've had a recorded informal knowledge sharing session, and based on that wrote notes in https://github.com/near/mpc/issues/654 during continued testing - this is still hard to discover and more of an information dump rather than a proper guide.

We will need to put together a proper guide for how to run MPC nodes within a TEE. The guide should be easy to follow and contain sufficient information so that anyone with access to a TDX-enabled server can:

1. Spin up the required dstack components.
2. Start an MPC node and configure it to connect to testnet, mainnet or a local network.

### User Story

As a node operator, I'd like reproducible instructions for how to run an MPC node on a TDX server.

### Acceptance Criteria

We have a guide (e.g. `/docs/running_an_mpc_node_in_tdx.md` or something similar) that any team member can easily read and spin up a proper MPC node on a TDX server.

For the initial guide, we don't need the node to be able to connect to an external network.

Initial draft for external guide: https://docs.google.com/document/d/1SEMIo3YmZ_vgVpqG8N0NqNi52MAqUDqNnVSHWrM4cRk/

gilcu3:
Closes #721

### DIFF
--- a/docs/running_an_mpc_node_in_tdx.md
+++ b/docs/running_an_mpc_node_in_tdx.md
@@ -1,0 +1,228 @@
+# Deploy an MPC node within a TEE
+
+This document describes the steps to deploy an MPC node within a TEE. Its
+expected audience is Near One engineers because it uses internal infrastructure.
+
+## Steps
+
+1. Provision a TDX server.
+2. Install dstack components and start core dstack services.
+3. Deploy an MPC node within dstack.
+4. Observe the public data for the MPC node
+
+### 1. Provision a TDX server.
+
+TODO: Add some sentence here about providers.
+
+1. SSH to NEAR's tdx dev machine. In this case we will use `TDX_SERVER=91.134.92.20`
+
+```bash
+ssh USER_NAME@TDX_SERVER
+```
+
+2. Add your user to the `kvm` group:
+
+```bash
+sudo usermod -a -G kvm USER_NAME
+```
+
+### 2. Install dstack and start core services
+
+1. Go to the data directory (`/mnt/data/` assuming you are on one of our tdx
+   servers)
+2. Follow the commands below
+
+Create a directory with a unique name for installation of dstack. For example
+your username
+
+```bash
+mkdir /mnt/data/WORKING_DIRECTORY_NAME
+```
+
+Give your own user ownership of the directory
+
+```bash
+sudo chown USER_NAME:USER_NAME WORKING_DIRECTORY_NAME/
+```
+
+Enter directory
+
+```bash
+cd WORKING_DIRECTORY_NAME/
+```
+
+3. Follow the instructions in
+   [Dstack's Getting Started section in their README](https://github.com/Dstack-TEE/dstack?tab=readme-ov-file#-getting-started)
+   until before
+   [step 2](https://github.com/Dstack-TEE/dstack?tab=readme-ov-file#2-download-or-build-guest-image).
+   Download the development version of the `0.5.2` by running:
+
+```bash
+# This will download the guest image from the release page.
+../build.sh dl -dev 0.5.2
+```
+
+4. Once it's installed you need to run `dstack-vmm` and `dstack-kms`. It's
+   important these keep running persistently. You can use `tmux` for this
+   purpose:
+
+Run Dstack key management service:
+
+```bash
+# Start a new tmux session
+tmux new -s dstack
+```
+
+```bash
+# Run dstack-kms
+sudo ./dstack-kms -c kms.toml
+```
+
+Open new tmux panel with `Ctrl+B` then `%`
+
+Run Dstack Virtual Machine Manager:
+
+```bash
+sudo ./dstack-vmm -c vmm.toml
+```
+
+
+Look for the log where it shows the localhost + port the VMM is running on. In
+this case `VMM_PORT=16040`. It should look something like:
+
+```log
+2025-07-22T20:04:51.818701Z  INFO rocket::rkt: endpoint=http://127.0.0.1:16040
+```
+
+Detach with `Ctrl+B` then `D`. To reattach later you can run:
+
+```bash
+tmux attach -t dstack
+```
+
+Port forward the port VMM is running on from the TDX host to your local machine.
+
+```bash
+ssh -NL 8080:localhost:VMM_PORT USER_NAME@TDX_SERVER
+```
+
+### 3. Deploy an MPC node within dstack
+
+
+1. Go to the url the `dstack-vmm` is running on `http://localhost:VMM_PORT`
+2. Click `Deploy a new instance` button on the top left.
+3. Fill in the form
+
+- name: `mpc_tee_test`  (this parameter can be selected at will)
+- image: `dstack-dev-0.5.2`
+- 8 CPUs
+- 64GiB Memory
+- 200GiB Storage
+
+4. Find a docker image of the MPC node
+   [here](https://hub.docker.com/r/nearone/mpc-node-gcp/tags). Images with
+   suffix `-tee` are TEE enabled. For the selected image we need the `IMAGE_TAG`
+   and `IMAGE_DIGEST`. For the latest image search `IMAGE_TAG=latest-tee`.
+
+5. Docker Compose File
+
+Copy the docker compose file below, and paste it into the vm creator. Notice you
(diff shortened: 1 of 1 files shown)


docs/running_an_mpc_node_in_tdx.md
```diff
@@ -0,0 +1,247 @@
...
+   # Start a new tmux session
+   tmux new -s dstack
+   ```
+
+   ```bash
+   # Run dstack-kms
+   ./dstack-kms -c kms.toml
+   ```
+
+   Open new tmux panel with `Ctrl+B` then `%`
```
barakeinav1, line 82: I used nohup "./dstack-kms -c kms.toml" > output.log 2>&1 &

but it is probably the same result

docs/running_an_mpc_node_in_tdx.md
```diff
@@ -0,0 +1,247 @@
...
+
+      volumes:
+         - mpc-data:/data
+         - shared-volume:/mnt/shared
+         - /:/host/
+         - /var/run/tappd.sock:/var/run/tappd.sock
+         - /var/run/dstack.sock:/var/run/dstack.sock
+      environment:
+         - MPC_HOME_DIR=/data
+         - MPC_ACCOUNT_ID=$MPC_ACCOUNT_ID
```
barakeinav1, line 141: using env variable here will force you to use KMS.
for production (and some testing) we will use SGX.

so I think it is better to avoid using the encrypted .env file. and put the value directly in the compose file. ( or at least mention this limitation)

docs/running_an_mpc_node_in_tdx.md
```diff
@@ -0,0 +1,247 @@
...
+   Once inside the VM it is possible to call the dstack curl [API endpoints](https://github.com/Dstack-TEE/dstack/blob/c41c17d465b6f0a213e091d677699181b9ba75ca/sdk/curl/api.md#endpoints), for example:
+
+   ```bash
+   curl --unix-socket /var/run/dstack.sock -X POST -H "Content-Type: application/json" -d '{}' http://localhost/Info
+   ```
+
+6. Activate features
+   Only activate the following features:
+
+   - KMS
```
barakeinav1, line 200: KMS for testing. SGX for production

docs/running_an_mpc_node_in_tdx.md
```diff
@@ -0,0 +1,247 @@
...
+   MPC_RESPONDER_ID=mpc-responder-2-barak-launch1-cdd0fd949a48.5035bf56abb0.testnet
+   MPC_NODE_0=mpc-node-0.service.mpc.consul=35.185.233.54
+   MPC_NODE_1=mpc-node-1.service.mpc.consul=34.168.117.59
+   ```
+
+   Several values in the previous file can be modified:
+
+   - IMAGE_TAG and IMAGE_DIGEST: Find a docker image of the MPC node in
+   [mpc-node-gcp/tags](https://hub.docker.com/r/nearone/mpc-node-gcp/tags). Images with
+   suffix `-tee` are TEE enabled. For the selected image we need the tag `IMAGE_TAG`
```
barakeinav1, line 231: is "suffix `-tee" something we support in our build system? and tied to the compilation TEE flag?

docs/running_an_mpc_node_in_tdx.md
```diff
@@ -0,0 +1,247 @@
...
+   and the corresponding digest `IMAGE_DIGEST`. The latest image should be `IMAGE_TAG=latest-tee`.
+   - TODO: comment other parameters
+
+8. Click on deploy
+
+### 4. Observe the public data for the MPC node
+
+Now data of the MPC node deployed can be accessed on your local machine:
+
+1. Go to the VM list in <http://localhost:8090>.  Click on the "View logs button"
```
barakeinav1, line 241: shouldn't be `17190`:`8090` and `17180`  due to the port forwarding you wrote above

docs/running_an_mpc_node_in_tdx.md
gilcu3: Makes sense, I didnt know the encrypted file had limitations, so I just move the env vars there so that everything that needed to be selected by the user was in the env file. Which parameters would cause this limitation then?

docs/running_an_mpc_node_in_tdx.md
gilcu3: I am not sure I am following you here. This is supposed to be opened in the user machine.

docs/running_an_mpc_node_in_tdx.md
barakeinav1: sure - whatever works

docs/running_an_mpc_node_in_tdx.md
barakeinav1: any env parameter.
for some reason phala does not support a "not-encrypted" env file. (they do have a .user-config , but I don't think it automatically interprets to the .env file.

for out use case, however, (since we don't want to use KMS) we don't pass any secrets to the VM.

a possible solution is to move all the env files into the docker-compose file.

in any cause - you are running the MPC alone and not via launcher. so I don't think it matter too much.

we will need to update the doc with the launcher flow as well

docs/running_an_mpc_node_in_tdx.md
barakeinav1: ok, but let just add this comment about KMS limitation.

docs/running_an_mpc_node_in_tdx.md
gilcu3: thanks for the explanation. So what I can do to be more general then, is eliminate the need of the env file here, and put everything in the compose file. Would that make sense for this guide, making it more usable latter?

docs/running_an_mpc_node_in_tdx.md
```diff
@@ -0,0 +1,247 @@
...
+
+6. Activate features
+   Only activate the following features:
+
+   - KMS
+   - Public Logs
+   - Public Sysinfo
+   - Public TCB Info
+
+7. Encrypted Environment Variables
```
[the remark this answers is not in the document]
gilcu3: could you elaborate on what exactly the notice would look like?

docs/running_an_mpc_node_in_tdx.md
```diff
@@ -0,0 +1,268 @@
...
+## Steps
+
+1. Provision a TDX server.
+2. Install dstack components and start core dstack services.
+3. Deploy an MPC node within dstack.
+4. Observe the public data for the MPC node
+
+### 1. Provision a TDX server
+
+TODO: Add some sentence here about providers.
```
netrome, line 15: Not sure I understand this todo. I suggest removing it for now.

```suggestion
```

Alternatively write something like this:
```suggestion
To run the MPC node within a TEE, you need a TDX-enabled server. There are several providers to choose from, such as pheonixNap, OVHcloud, openmetal.io and Hivelocity to name a few.
```

docs/running_an_mpc_node_in_tdx.md
```diff
@@ -0,0 +1,268 @@
...
+      docker inspect nearone/mpc-node-gcp:IMAGE_TAG | jq .[0].Id
+      ```
+
+      - `NEAR_BOOT_NODES`: can be obtained executing
+
+      ```bash
+      curl -X POST https://rpc.testnet.near.org -H "Content-Type: application/json" -d '{"jsonrpc": "2.0", "method": "network_info","params": [], "id": "dontcare"}'| jq -r '.result.active_peers[] as $active_peer | "\($active_peer.id)@\($active_peer.addr)"' | paste -sd',' -
+      ```
+
+      - TODO: comment other parameters
```
netrome, line 159: Where? Let's clear this for now.

netrome:
```suggestion
```

netrome (approved):
I suggest clearing some TODOs because I generally find such comments more distracting than helpful unless they reference actual issues.

Otherwise, nice stuff. Would be good to get this merged and we can tackle any other improvements as follow-up issues.

PR #740, opened 2025-07-30
fix: display error message when participant list is empty in mpc-node generate-test-configs

### LINKED ISSUE #716: [Task] `generate-test-configs` is panicking when participants set is empty
### Background

`generate-test-configs` CLI tool panics with an out of bounds access when the participant set is empty. 
```bash
mpc-node generate-test-configs --output-dir ~/.near --threshold 1
```
```
thread 'main' panicked at node/src/cli.rs:452:42:
index out of bounds: the len is 0 but the index is 0
```

### User Story

.

### Acceptance Criteria

A better error message than this panic should be emitted, or the CLI parser should reject empty participants list.

_No response_

gilcu3:
Closes #716

### DIFF
--- a/node/src/cli.rs
+++ b/node/src/cli.rs
@@ -73,7 +73,7 @@ pub enum CliCommand {
     GenerateTestConfigs {
         #[arg(long)]
         output_dir: String,
-        #[arg(long, value_delimiter = ',')]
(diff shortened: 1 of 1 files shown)


near-bookrock:
Thank you!

PR #742, opened 2025-07-31
chore(devnet): Add support for voting new approved code hash on devnet

### LINKED ISSUE #701: [Task] Add support for voting new approved code hash on devnet
### Background

For E2E testing of TEE we require the contract to have a list of approved docker image hashes. We need to call the `vote_code_hash` method on the contract to add an MPC image hash.

https://github.com/near/mpc/blob/1597faf1e2e1ae744b041cdcf6d2974346381219/libs/chain-signatures/contract/src/lib.rs#L854-L855

### User Story

As a developer I'd like to do E2E testing of a TEE enabled networ.

### Acceptance Criteria

Add a CLI option to the `devnet` CLI that can put a new MPC image hash on the contract.

_No response_

near-bookrock:
Resolves #701 

Tested manually that voting a new image hash works:
https://testnet.nearblocks.io/txns/4A2NTyTQPWu67Bmmf4yuUaMVdAXZsmxxxqYtYydafk7j


libs/chain-signatures/contract/src/primitives/thresholds.rs
```diff
@@ -13,7 +14,7 @@ const MIN_THRESHOLD_ABSOLUTE: u64 = 2;
 /// assert!(dt.value() == 8);
 /// ```
 #[near(serializers=[borsh, json])]
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+#[derive(Debug, Deref, PartialEq, Eq, PartialOrd, Ord, Clone)]
```
kevindeforth, line 17: If we import this, then lets get rid of the `value()` method.

kevindeforth (commented):
I would approve, if it wasn't for the small contract code change. I think we should have either the deref trait, or a getter, but not both at the same time.

near-bookrock:
> I would approve, if it wasn't for the small contract code change. I think we should have either the deref trait, or a getter, but not both at the same time.

Fair. It requires more extensive refactoring to remove it, so I created a tracking issue. https://github.com/near/mpc/issues/744

I'll undo the change.

kevindeforth (approved):

Issue #689, opened 2025-07-22
[Task] replace `black` in favor of `ruff`


ghost:
### Background

In https://github.com/near/mpc/issues/675 we introduced `black` as formatter of choice for the MPC prject, and also enforced its formatting on CI.

We would also like to add lints to our CI to for example catch unused argument in our python code, as was detected in [this PR review](https://github.com/near/mpc/pull/682#discussion_r2220331267). Ruff has better support for lints, and has drop in parity support for `black`.

Ruff also has good official integrations, if not more extensive:
- https://github.com/astral-sh/ruff-action
- https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff
- https://docs.astral.sh/ruff/editors/setup/#helix

### User Story

As a developer I want to make sure we have both consistent formatting, and linting of our python code.

### Acceptance Criteria

Replace `black` with `ruff`.

kuksag:
Let the fun begin 
```
Found 441 errors.
[*] 225 fixable with the `--fix` option (25 hidden fixes can be enabled with the `--unsafe-fixes` option).
```

PR #750, opened 2025-07-31
chore: Add python linter

kuksag:
resolves:  #689


pytest/common_lib/shared/mpc_node.py
```diff
@@ -93,7 +93,7 @@ def reset_mpc_data(self):
         assert not self.is_running
         patterns = [
             "CURRENT",
-            "IDENTITY" "LOCK",
+            "IDENTITYLOCK",
```
kuksag, line 96: that looks like a typo tho, it probably should be:
```
"IDENTITY",
"LOCK"
```

near-bookrock (approved):
Thanks @kuksag 💯

Issue #777, opened 2025-08-05
bug: call order in spawn_real_indexer


kevindeforth:
### Background

`monitor_docker_images` is an async task spawned in the background here
https://github.com/near/mpc/blob/dbcefc2e2389157eff76ce9159d55fa192bf9629/node/src/indexer/real.rs#L125, after a call to `listen_blocks`. But `listen_blocks` runs indefinitely and only returns in case of an error.

### User Story

As a developer and user, I want an MPC node that works.

### Acceptance Criteria

`listen_blocks` should be the last function call inside `spawn_real-indexer` before returning the exit signal.

PR #778, opened 2025-08-05
fix: spawn mointor docker images task before listening to blocks indefinitely

kevindeforth:
resolves https://github.com/near/mpc/issues/777
stacked on top of #776


near-bookrock (approved):
Thanks, and great job spotting it!

Issue #735, opened 2025-07-29
bug: Verification missing RTMR3 type check for event logs


barakeinav1:
### Background

We discovered that some of the TEE event validations (e.g., local-sgx, compose-hash) do not explicitly verify that the event is of type imr3, which is necessary for it to be measured into RTMR3. Without this check, an attacker could inject a fake event of the same name into the event log, causing the smart contract to trust unmeasured or attacker-controlled data.

### User Story

As a smart contract verifier, I need to ensure that only events actually measured into RTMR3 are trusted, so that untrusted or forged events cannot be used to bypass security checks.

### Acceptance Criteria

Acceptance Criteria
 Add a check that ensures any event (e.g., local-sgx, compose-hash, etc.) relied on for verification is of type imr3.

 Update all verification functions (check_local_sgx, check_mpc_hash, etc.) to validate the type field of events.

 Add test coverage with both valid and invalid event types to ensure rejection of non-imr3 data.

 Confirm that events not used to compute RTMR3 are never assumed to be trustworthy unless backed by a separate hash that is verified and measured.

### Resources & Additional Notes

https://nearone.slack.com/archives/C07UW93JVQ8/p1753795609219309?thread_ts=1753709819.171519&cid=C07UW93JVQ8

PR #780, opened 2025-08-05
fix(node): enforce RTMR3 validation on event checks

barakeinav1:
fixes https://github.com/near/mpc/issues/735

### DIFF
--- a/pytest/common_lib/shared/mpc_cluster.py
+++ b/pytest/common_lib/shared/mpc_cluster.py
@@ -381,7 +381,7 @@ class MpcCluster:
             ContractMethod.PROPOSE_UPDATE,
             args,
             # TODO: #771 https://github.com/near/mpc/issues/771
-            deposit=10610730000000000000000000,
+            deposit=10612020000000000000000000,
         )
         res = participant.send_txn_and_check_success(tx, timeout=30)
         return int(


libs/chain-signatures/contract/src/tee/tee_participant.rs
```diff
@@ -284,7 +291,10 @@ impl TeeParticipantInfo {
     ) -> bool {
         let mpc_node_image_digest = event_log
             .iter()
-            .find(|e| e["event"].as_str() == Some("mpc-image-digest"))
+            .find(|e| {
+                e["event"].as_str() == Some("mpc-image-digest")
+                    && e["imr"].as_u64() == Some(RTMR3_INDEX)
+            })
```
near-bookrock, line 297: Ditto

libs/chain-signatures/contract/src/tee/tee_participant.rs
```diff
@@ -183,14 +185,17 @@ impl TeeParticipantInfo {
             Some(r) => hex::encode(r.rt_mr3),
             None => return false,
         };
-        let replayed_rtmr3 = replay_rtmr(event_log.to_owned(), 3);
+        let replayed_rtmr3 = replay_rtmr(event_log.to_owned(), RTMR3_INDEX.try_into().unwrap());
```
near-bookrock, line 188: What type are we converting to here in the `try_into()` call? We should avoid unwraps to not cause panics.
 
```suggestion
        let replayed_rtmr3 = replay_rtmr(event_log.to_owned(), RTMR3_INDEX.try_into().unwrap());
```

near-bookrock: `replay_rtmr` is expecting u8.

Why not have the const `RTMR3_INDEX` be u8 instead of u64 instead?

near-bookrock: I see it will be more casting if you cast up from `u8` to `u64` but it will always give us a strong guarantee at compile time that the conversion never fails, and code won't panic.

It's also obvious to the reader of the code that `RTMR3_INDEX` is supposed to be a small value by setting it to a `u8`.

If you cast up to `u64` from `u8` it will be quite simple, just add `as u64` as a suffix.
```rust
 e["imr"].as_u64() == Some(RTMR3_INDEX as u64)
```

PR #786, opened 2025-08-06
chore: unify rand by downgrading

### LINKED ISSUE #658: Unify `rand` dependency version in workspace and `devnet` crate.
### Background

There is currently a mismatch between the `node` crate and `devnet` crate for their dependenct to the `rand` crate.

https://github.com/near/mpc/blob/354db38367abfdbab04a567bac613cbf41994ed1/devnet/Cargo.toml#L16
https://github.com/near/mpc/blob/c4da39c9afea0954317a0bd83c54a5a0402eb74d/Cargo.toml#L62

We should unify the version by upgrading the rand version to latest version in the workspace Cargo.toml file. It will require some changes in the node crate as there are breaking changes in the API of the rand crate.

### User Story

.

### Acceptance Criteria

The `rand` dependency version in th workspace is bumped to the latest version, and the `devnet` crate uses `{ workspace = true }` for the `rand` crate.

_No response_

gilcu3:
Closes #658

The move from rand 0.8 to rand 0.9 breaks other deps in node, and rand was just minimally used in devnet. Therefore, I opted to downgrade rand in devnet to use the same version  as in the workspace. We can leave the update of rand for the future when our crypto deps also update.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4849,7 +4849,7 @@ dependencies = [
  "near-jsonrpc-primitives 0.28.0",
  "near-primitives 0.28.0",
  "near-sdk",
- "rand 0.9.2",
+ "rand 0.8.5",
  "reqwest 0.12.22",
  "serde",
  "serde_json",
--- a/devnet/Cargo.toml
+++ b/devnet/Cargo.toml
@@ -13,8 +13,7 @@ futures = { workspace = true }
 hex = { workspace = true }
 legacy-mpc-contract = { workspace = true }
 mpc-contract = { workspace = true, features = ["test-utils", "dev-utils"] }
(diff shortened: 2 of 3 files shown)


near-bookrock (changes requested):
Cross posting from slack for visibility:
https://nearone.slack.com/archives/C0912BTG51T/p1754518277359149?thread_ts=1754494109.241189&cid=C0912BTG51T

>It's not solving the issue though per the acceptance criteria which was to upgrade the workspace version and then unify.
I would just put the issue as blocked until k256 is released any day now.
>
>I'm just hesitant on downgrading dependencies when it's redundant changes that will need to be reverted anyway. It also opens a window for more breaking code in devnet crate the meantime.

Issue #696, opened 2025-07-22
[Task] Separate CI tasks into several jobs


kuksag:
### Background

Current state: we have all tasks formatting, linting, tests under the same CI job. Reference: https://github.com/near/mpc/blob/9c6be0fa917ebfdaeed8deb93536df58a3e68243/.github/workflows/ci.yml#L34

As opposed, we can have each task in its own job with the following hierarchy `fmt -> clippy -> [contract_tests, unit_tests, pytests]` with the last step running in parallel. As a side bonus, we get to see 5 different statuses instead of 1 on a github PR page

### User Story

As a developer, I want to run things in parallel when it's possible

### Acceptance Criteria

`contract_tests, unit_tests, pytests` are being run in parallel 

### Resources & Additional Notes

One can use `depends: [job]` option when configuring a github workflow

PR #787, opened 2025-08-06
chore(ci): split ci-tests into multiple jobs

kevindeforth:
Resolves https://github.com/near/mpc/issues/696

Reduces CI time by approx. 20 minutes.

Additionally, we now also unittest the build without the tee flag.

@andrei-near I think we need to change the branch protection rules to allow this PR to pass (c.f. [this thread](https://github.com/orgs/community/discussions/26698#discussioncomment-3252954))

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,7 @@ jobs:
         run: |
           CARGO_TARGET_DIR="target/clippy" \
           RUSTFLAGS="-D warnings" \
-          cargo clippy --locked
+          cargo clippy --all-targets --locked
 
       - name: Run Cargo fmt
         run: |
@@ -85,7 +85,7 @@ jobs:
         run: |
           CARGO_TARGET_DIR="target/clippy" \
           RUSTFLAGS="-D warnings" \
-          cargo clippy --features "tee" --locked
+          cargo clippy --all-targets --features "tee" --locked
 
       - name: Run Cargo fmt
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,8 +31,8 @@ jobs:
           profile-name: "mpc-image-builder"
           file: deployment/Dockerfile-gcp
 
-  ci-tests:
-    name: "Run tests"
+  clippy-mpc-node:
+    name: "MPC Node: clippy and format"
     runs-on: warp-ubuntu-2204-x64-8x
     timeout-minutes: 60
     permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -180,3 +345,21 @@ jobs:
...
+      - clippy-mpc-devnet
+      - clippy-mpc-node-tee
+      - clippy-mpc-contract
+      - mpc-tee-unittests
+      - mpc-unittests
+      - mpc-contract-unittests
+      - mpc-pytests
+      - tee-launcher-tests
+    steps:
+      - run: echo "All tests passed."
```
kevindeforth, line 365: Added this to satisfy the branch protection rule.

Alternatively, we could also change the branch protection rule (c.f. [this thread](https://github.com/orgs/community/discussions/26698#discussioncomment-3252954))

image

.github/workflows/ci.yml
```diff
@@ -55,31 +55,125 @@ jobs:
         run: |
           CARGO_TARGET_DIR="target/clippy" \
           RUSTFLAGS="-D warnings" \
-          cargo clippy --all-features --all-targets --locked
+          cargo clippy --locked
```
gilcu3, line 58: By removing the --all-features flag, aren't we losing some checks? For example at the moment we have a feature called network-hardship-simulation in node.

andrei-near (approved):

.github/workflows/ci.yml
near-bookrock: For clippy we should also check with all features? Why would we not have linting for feature gated code?

.github/workflows/ci.yml
```diff
@@ -89,20 +183,85 @@ jobs:
...
+          cache-all-crates: true
+
+      - name: Install cargo-nextest
+        run: cargo install cargo-nextest
 
       - name: Run cargo-nextest
-        run: cargo nextest run --release --locked --all-features
+        run: cargo nextest run --release --locked
+
+  mpc-contract-unittests:
```
near-bookrock, line 215: Nit:
- Spacing `unittest` -> `unit test`.
Tip:
- This runs **both unit tests and integration tests.** The tests in `/tests` folder are integration tests, so `unit test` could be a bit misleading here with rust conventons https://doc.rust-lang.org/rust-by-example/testing.html

.github/workflows/ci.yml
kevindeforth: I might be confused, but I thought `cargo clippy --all-features` works the same as `cargo --all-features`. I.e. it checks the code resulting from activating all features. In our case, that is equivalent to checking `features = "network-hardship, tee"`. However, we do not run that code in production. Also, we wouldn't be checking the code without the tee flag.

But please correct me, if I got this wrong.

near-bookrock: `all-targets` should at least be added back if we want linting for our tests.

kevindeforth: made an issue: https://github.com/near/mpc/issues/794

PR #788, opened 2025-08-06
fix: use download_config parameter in mpc-node cli

### LINKED ISSUE #714: [Task] mpc-node `init` CLI tool always downloads a configuration file
### Background
Observed when working on setting up a local network in https://github.com/near/mpc/issues/645.

When initializing an mpc node with the following command:

```
mpc-node init --dir ~/.near/mpc-alice --chain-id mpc-localnet --genesis ~/.near/mpc-localnet/genesis.json --boot-nodes $NODE_PUBKEY@localhost:3030
```

The following error is returned:
```
Error: Failed to download the config file from https://s3-us-west-1.amazonaws.com/build.nearprotocol.com/nearcore-deploy/mpc-localnet/rpc/config.json
```

This despite the `--download-config` flag being unset.

The reason is that the `download_config` flag is not used in the `mpc-node` `init` CLI:
https://github.com/near/mpc/blob/b8ee25b6fdf90f38d96fa9d4b11a99cba45a3c93/node/src/cli.rs#L103-L105

This is problematic, because we want to be able to generate a configuration file without downloading 

### User Story
As a developer, I would like to be able to spin up a fresh local MPC network when working from an airplane without WIFI.

### Acceptance Criteria
The MPC node can generate a configuration file without fetching anything over the internet.

_No response_

gilcu3:
Closes #714

This PR sets two config options to None when download_config is false, which would stop the config file being downloaded, as can be seen in the `init_configs` implementation:

https://github.com/near/nearcore/blob/4b2d7219cef9a2a213ecb73b996710fb9903c13e/nearcore/src/config.rs#L949-L958


DSharifi (approved):
Thanks @gilcu3

PR #799, opened 2025-08-08
fix: replay all RTMR3 required events

### LINKED ISSUE #795: replay all RTMR3 events to verify that the digest is valid
### Background

We currently don't check that the `digest` field for each RTMR3 event is valid. Currently we only check it for the  `mpc-image-digest` event, but there are other events we also care about.

Example:
```json
    {
      "imr": 3,
      "event_type": 134217729,
      "digest": "bcb7a884a6ba0970997335c548b61e2486be6065860219c5f436291d57ce62a8feb2405a63dfbbfca63be8ab6cd9e72a",
      "event": "mpc-image-digest",
      "event_payload": "a87f7eb6882446dd714e6d47d9d1b9331cb333f36d3905f172c68adbd06e461f"
    }
```

### User Story

_No response_

### Acceptance Criteria

As part of the `Attestation::verify` in the `attestation` crate we need to:

Fields we care about:
- `mpc-image-digest`
- `key-provider`
- `compose-hash`

For each RTMR3 event listed above we have to check:
- Each event appears only once.
- Each event has a valid digest

_No response_

gilcu3:
Closes #795


attestation/src/attestation.rs
```diff
@@ -12,8 +12,12 @@ use near_sdk::env::sha256;
 /// Expected TCB status for a successfully verified TEE quote.
 const EXPECTED_QUOTE_STATUS: &str = "UpToDate";
 
-const COMPOSE_HASH_EVENT_TYPE: u32 = 134217729;
+// Defined https://github.com/Dstack-TEE/dstack/blob/cfa4cc4e8a4f525d537883b1a0ba5d9fbfd87f1e/tdx-attest/src/lib.rs#L28
+// It is the same for all events
```
near-bookrock, line 16: Nit:
```suggestion
/// The constant values are all defined in https://github.com/Dstack-TEE/dstack/blob/cfa4cc4e8a4f525d537883b1a0ba5d9fbfd87f1e/tdx-attest/src/lib.rs#L28
```

attestation/src/attestation.rs
```diff
@@ -145,17 +149,87 @@ impl Attestation {
...
+        let mpc_image_digest = match hex::decode(mpc_image_digest_payload) {
+            Ok(bytes) => bytes,
+            Err(e) => {
+                tracing::error!(
+                    "Failed to decode hex string for key-provider event: {:?}",
+                    e
+                );
+                return false;
+            }
+        };
```
near-bookrock, line 229: Instead of having a function for each digest check, can you generalize the function? Would want to avoid duplicate check for each event.

near-bookrock (changes requested):

PR #835, opened 2025-08-14
chore: add docker TEE build as a CI check

ghost:
closes https://github.com/near/mpc/issues/836


libs/nearcore
kevindeforth, line 1: why the submodule change?

barakeinav1 (approved):

Issue #842, opened 2025-08-15
fix: API mismatch on node


kevindeforth:
### Background

The node currently calls an endpoint called `submit_remote_attestation`, but no such endpoint exists in the contract.
https://github.com/near/mpc/blob/198f5e88417d42ae067e8d9d47687d54193a638e/node/src/indexer/types.rs#L157

Further, on the contract side, the function name is confusing:
https://github.com/near/mpc/blob/ffe1004e49988eb7331e3487f33a0f5b45128e74/libs/chain-signatures/contract/src/lib.rs#L525

The name implies that the function is only called by prospective participants, not by existing ones. But current participants are supposed to call the same endpoint for submitting their tee information at start-up and periodically.

On the node side, the argument for `SubmitRemoteAttestation` is called `ProposeJoinArgs`
https://github.com/near/mpc/blob/198f5e88417d42ae067e8d9d47687d54193a638e/node/src/indexer/types.rs#L124

Neither one of these names are suitable:
- `ProposeJoinArgs` is ill-fitting for the aforementioned reasons
- `SubmitRemoteAttestation` is not necessarily a great choice either, as he node submits not only the remote attestation, but also ephemeral values.
- The mismatch between the `ChainSendTransactionRequest` method name and argument is not consistent with the existing code base.

### User Story

As a developer, I want a consistent code base.

### Acceptance Criteria

- on the contract, rename the endpoint `propose_join` to `submit_participant_info`.
- on the node:
    - call the correct endpoint
    - rename `ProposeJoinArgs` to `SubmitParticipantInfoArgs`
    - rename `SubmitRemoteAttestation` to `SubmitParticipantInfo`

Issue #846, opened 2025-08-15
unit test for `submit_tee_participant_info`


kevindeforth:
### Background

We don't have one right now

### User Story

As a dev, I want to know that our code works.

### Acceptance Criteria

We have a unit test

pbeza:
This task was addressed in https://github.com/near/mpc/pull/828. ✅

netrome:
Awesome!

Issue #823, opened 2025-08-13
Remove contract code that allows MPC node running without TEE to join the MPC cluster


barakeinav1:
### Background

In the first transition phase to TEE, we will allow both TEE and non TEE node to participate.
we only allow new node with TEE, but we won't kickout existing node.

after the transition phase is finished, we will need to remove the contract code that allow non TEE nodes to participate

code is found under:
tee_status and validate_tee methods in tee_state.rs 

### User Story

as user of the MPC network, I'd like to insure that only nodes running inside TEE are allow to join / participate in the network.

### Acceptance Criteria

contract code for only allowing TEE node to join or participate is updated and tested

barakeinav1:
@DSharifi I don't understand why is this issue closed.
This issue should be close as part of moving to full TEE enforcement.

DSharifi:
Created #1710 and #1711 as sub issues with @barakeinav1  to cover this task

netrome:
Keeping this around for the hard launch

SimonRastikian:
The changes in https://github.com/near/mpc/pull/2954/changes let me think that the issue should soon not be needed -- Except for the related ones

PR #851, opened 2025-08-15
fix: propose join and rename to submit_participant_info

### LINKED ISSUE #841: Fix `propose_join`
### Background

In `propose_join`, we add the proposed participant to the tee state
https://github.com/near/mpc/blob/ffe1004e49988eb7331e3487f33a0f5b45128e74/libs/chain-signatures/contract/src/lib.rs#L559 

Unfortunately, we do not verify the tee information submitted by the caller. Instead, we check whether or not the caller is already a known participant and reject them only in case they have an invalid state. This makes no sense, as an existing participant with an invalid state should be allowed to update their state. And further, we would probably not want to store invalid states in the tee state.

https://github.com/near/mpc/blob/ffe1004e49988eb7331e3487f33a0f5b45128e74/libs/chain-signatures/contract/src/lib.rs#L549

### User Story

As a developer, I want working code.

### Acceptance Criteria

`propose_join` should respect the following behavior:

1. it should ensure the submitted tee state is valid, aborting the transaction if it is not.
2. It should store the valid tee information to the tee state of the contract:
    a. replacing any existing information if the caller already has a tee state
    b. adding new information if the caller does not yet have a tee state
3. In case of 2b), a deposit must be charged.

https://nearone.slack.com/archives/C07UW93JVQ8/p1755181284788119

kevindeforth:
Resolves #841 and contract part of #842

### DIFF
--- a/libs/chain-signatures/contract/src/lib.rs
+++ b/libs/chain-signatures/contract/src/lib.rs
@@ -520,6 +520,8 @@ impl VersionedMpcContract {
         }
     }
 
+    /// (Prospective) Participants can submit their tee participant information through this
+    /// endpoint.
     #[payable]
     #[handle_result]
     pub fn submit_participant_info(
@@ -551,7 +553,7 @@ impl VersionedMpcContract {
                 .message("TeeQuoteStatus is invalid".to_string()));
         }
 
-        // Add a new proposed participant to the contract state
+        // Add the participant information to the contract state
         mpc_contract
             .tee_state
             .add_participant(account_id.clone(), proposed_tee_participant.clone());
--- a/libs/chain-signatures/contract/src/lib.rs
+++ b/libs/chain-signatures/contract/src/lib.rs
@@ -522,14 +522,14 @@ impl VersionedMpcContract {
 
     #[payable]
     #[handle_result]
-    pub fn propose_join(
+    pub fn submit_participant_info(
         &mut self,
         #[serializer(borsh)] proposed_tee_participant: TeeParticipantInfo,
         #[serializer(borsh)] sign_pk: PublicKey,
     ) -> Result<(), Error> {
         let account_id = env::signer_account_id();
         log!(
-            "propose_join: signer={}, proposed_tee_participant={:?}",
+            "submit_participant_info: signer={}, proposed_tee_participant={:?}",
             account_id,
             proposed_tee_participant,
         );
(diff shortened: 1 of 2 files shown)


libs/chain-signatures/contract/src/tee/tee_state.rs
```diff
@@ -72,6 +68,22 @@ impl TeeState {
...
+    pub(crate) fn verify_tee_participant(
+        &mut self,
+        account_id: &AccountId,
+        sign_pk: &PublicKey,
+    ) -> Result<TeeQuoteStatus, Error> {
+        let tee_participant_info = self.tee_participant_info.get(account_id);
+
+        if let Some(tee_participant_info) = tee_participant_info {
+            let tee_participant_info: TeeParticipantInfo = tee_participant_info.clone();
+            self.verify_tee_participant_info(&tee_participant_info, sign_pk)
```
netrome, line 81: Why do we clone the participant info if we later send a reference to the function?

netrome (approved):
Good stuff! Love these small reviews and clear issues. Makes it so much easier to squeeze in a small PR review before lunch :star_struck: 

Would love to see tests for this, but I see you created a follow-up for that already since it requires the attestation module to be integrated. https://github.com/near/mpc/issues/846

Otherwise, just one question about a clone.

libs/chain-signatures/contract/src/tee/tee_state.rs
```diff
@@ -36,28 +36,24 @@ impl Default for TeeState {
         }
     }
 }
+
 impl TeeState {
-    /// Verifies the TEE quote and Docker image
-    pub(crate) fn verify_tee_participant(
+    /// Returns either `Valid` or `Invalid`.
```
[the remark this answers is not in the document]
kevindeforth: yes, this will be solved with https://github.com/near/mpc/issues/823.
I removed the comment.

netrome:
@near-bookrock - why did you remove this PR from the merge queue? 
image

netrome (approved):
Looks better now, thank you! :

netrome:
> I wanted to have time to also review before it got merged. https://nearone.slack.com/archives/C0912BTG51T/p1755252073726839?thread_ts=1755250728.252599&cid=C0912BTG51T.

I see, fair enough.

> It was opened at 11:38 GMT+2 and moved to merge queue at 11:54 GMT+2, so there wasn't much time to review if I also wanted to take a look.

Yep, this is the gold standard imo. Small PRs and quick review turnaround. Nothing wrong with post-merge reviews and follow-ups either. But in this case when you're reviewing just as it got added to the merge queue I agree it can be reasonable to hold the merge for a short bit.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -542,18 +544,16 @@ impl VersionedMpcContract {
         };
 
         // Verify the TEE quote and Docker image for the proposed participant
```
barakeinav1, line 546: the comment says // Verify the TEE quote and Docker image for the proposed participant 
        but now verify_tee_participant_info only verify the image and not the quote  - which one is correct?

PR #869, opened 2025-08-18
chore(tee): remove account key from report data

### LINKED ISSUE #866: Remove account_key from the report_data
### Background

This is a follow-up task for #855. Together with @near-bookrock and @barakeinav1, we’re considering removing the `AccountKey` from the `ReportData`, and only keeping the TLS public key in the first iteration.

Currently, we don’t store the `AccountKey` in the contract, but it’s needed for periodic re-verification of all peers when a node calls the contract to re-create the report data. We already store the TLS public key for each node, which will be used in the report data.

From what we understand, the only security benefit of storing the `AccountKey` is that it lets us gate certain contract methods so that they can only be called by the node itself.

### User Story

_No response_

### Acceptance Criteria

`ReportData` consisting only of the version and the hash of the node’s TLS key.

https://nearone.slack.com/archives/C07UW93JVQ8/p1755510828968719?thread_ts=1755499467.319929&cid=C07UW93JVQ8

pbeza:
Fixes #866

### DIFF
--- a/attestation/tests/app_compose.rs
+++ b/attestation/tests/app_compose.rs
@@ -37,8 +37,6 @@ fn test_app_compose_deserialization() {
 fn test_app_compose_from_tcb_info() {
     let dstack_tcb_info: DstackTcbInfo = serde_json::from_str(TEST_TCB_INFO_STRING).unwrap();
     let app_compose = dstack_tcb_info.app_compose;
-    println!("APP_COMPOSE:\n{}", app_compose);
-    println!("TEST_APP_COMPOSE_STRING:\n{}", TEST_APP_COMPOSE_STRING);
     assert_eq!(app_compose, TEST_APP_COMPOSE_STRING);
 }
 


attestation/src/report_data.rs
```diff
@@ -125,41 +123,28 @@ impl ReportData {
...
 
         let td_report = quote.report.as_td10().expect("Should be a TD 1.0 report");
 
-        let expected_rtmr3 = [
-            0x90, 0x5f, 0x81, 0x71, 0x7d, 0x9, 0x5f, 0x14, 0xf5, 0x94, 0x47, 0xed, 0xff, 0xe3,
-            0x3e, 0x1e, 0x3c, 0xe4, 0x85, 0xda, 0xaa, 0x95, 0x3a, 0xc6, 0xfe, 0xb8, 0x21, 0x7f,
-            0xc6, 0x5a, 0x3d, 0x50, 0xb2, 0x96, 0x2e, 0x41, 0x15, 0xe, 0xb8, 0x53, 0x3b, 0xda,
-            0xbe, 0xb2, 0x35, 0x49, 0x31, 0x34,
-        ];
-        assert_eq!(td_report.rt_mr3, expected_rtmr3,);
```
DSharifi, line 143: I removed this sanity check, as it was effectively testing the dcap crate, and not our code.

attestation/tests/assets/create_assets.sh
```diff
@@ -0,0 +1,53 @@
...
+# Extract collateral
+jq -r '.tee_participant_info.quote_collateral | fromjson' "$INPUT_FILE" > "$OUTPUT_DIR/collateral.json"
+
+# Extract quote
+jq -c '.tee_participant_info.tee_quote' "$INPUT_FILE" > "$OUTPUT_DIR/quote.json"
+
+# Extract tcb_info
+jq -r '.tee_participant_info.raw_tcb_info | fromjson' "$INPUT_FILE" > "$OUTPUT_DIR/tcb_info.json"
+
+# Extract launcher_image_compose.yaml. It is whitespace sensitive, and we need it's exact hash
```
DSharifi, line 45:
```suggestion
# Extract launcher_image_compose.yaml. It is whitespace sensitive, and we need its exact hash
```

DSharifi (approved):

barakeinav1 (approved):

attestation/tests/app_compose.rs
```diff
@@ -37,6 +37,8 @@ fn test_app_compose_deserialization() {
 fn test_app_compose_from_tcb_info() {
     let dstack_tcb_info: DstackTcbInfo = serde_json::from_str(TEST_TCB_INFO_STRING).unwrap();
     let app_compose = dstack_tcb_info.app_compose;
+    println!("APP_COMPOSE:\n{}", app_compose);
+    println!("TEST_APP_COMPOSE_STRING:\n{}", TEST_APP_COMPOSE_STRING);
```
pbeza, line 41: Do we want to keep those debug prints?

PR #873, opened 2025-08-18
chore: implement borsh schema for attestation crate types

### LINKED ISSUE #859: implement borsh schema for `attestation` crate types
### Background

The attestation crate contains public types that the contract uses in it's API signature for the public methods. These types must implement `borshschema` trait for the contract to compile for the  `attestation` crate to be integrated in #770.

### User Story

_No response_

### Acceptance Criteria

Implement `BorshSchema` trait for all public types in `attestation` crate, similar to what is already done in `primitives` crate.

_No response_

ghost:
closes #859

### DIFF
--- a/libs/chain-signatures/Cargo.lock
+++ b/libs/chain-signatures/Cargo.lock
@@ -292,7 +292,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "borsh",
- "dcap-qvl",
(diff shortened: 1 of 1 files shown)


kuksag:
> Just for the record, can you remind me what not(target_arch = "wasm32") in the cfg_attrs you added is for, and whether it's really needed?

Because ABI is built for a default target (`x86-linux-*` whatever), thus `not(target_arch = "wasm32")`. The `abi` is on by default when using `cargo near`, because in `cargo near` you can't separately control compilation flags for ABI and Contract building. When using `cargo build --target=wasm32-unknown` you can omit the flag

PR #874, opened 2025-08-19
chore(test): more detailed error checking in contract test

kevindeforth:
Follow-up to #871 
To ensure the call fails due to caller not being a participant

### DIFF
--- a/libs/chain-signatures/contract/tests/tee.rs
+++ b/libs/chain-signatures/contract/tests/tee.rs
@@ -174,7 +174,10 @@ async fn test_vote_code_hash_doesnt_accept_account_id_not_in_participant_list() 
     };
     let expected = format!("{:?}", InvalidState::NotParticipant);
     let err_str = format!("{:?}", err);
-    assert!(err_str.contains(&expected));
+    assert!(
+        err_str.contains(&expected),
+        "expected failure due to voter not being a participant"
+    );
     Ok(())
 }
 


libs/chain-signatures/contract/tests/tee.rs
```diff
@@ -164,11 +164,17 @@ async fn test_vote_code_hash_doesnt_accept_account_id_not_in_participant_list()
         0x90, 0xab, 0xcd, 0xef, 0x12, 0x34, 0x56, 0x78, 0x90, 0xab, 0xcd, 0xef, 0x12, 0x34, 0x56,
         0x78, 0x90,
     ]);
-    assert!(random_account
-            .call(contract.id(), "vote_code_hash")
-            .args_json(serde_json::json!({"code_hash": hash}))
-            .transact()
-            .await?.is_failure(), "vote_code_hash should not accept votes from a randomly generated account id that is not in the participant list");
```
DSharifi, line 171: Could you keep the error message, or some message to make it very obvious what failed if we have a regression?

DSharifi (approved):
Thanks for this follow up!

libs/chain-signatures/contract/tests/tee.rs
kevindeforth: good point, Idk why I removed that!

libs/chain-signatures/contract/tests/tee.rs
```diff
@@ -164,11 +164,17 @@ async fn test_vote_code_hash_doesnt_accept_account_id_not_in_participant_list()
...
+        .call(contract.id(), "vote_code_hash")
+        .args_json(serde_json::json!({"code_hash": hash}))
+        .transact()
+        .await?;
+    let Err(err) = res.into_result() else {
+        panic!("vote_code_hash should not accept votes from a randomly generated account id that is not in the participant list");
+    };
+    let expected = format!("{:?}", InvalidState::NotParticipant);
+    let err_str = format!("{:?}", err);
+    assert!(err_str.contains(&expected));
```
DSharifi, line 177: Nit:
We could have a message here as well. But feel free to merge as is

Issue #875, opened 2025-08-19
Research: Security & Implementation Feasibility of Including AccountKey in TEE report_data and restricting methods to accept only that key


barakeinav1:
### Background

TL/DL
Storing AccountKey in contract state and report data adds some security but adds scope to the MVP.

Summary
in the original TEE design we added AccountKey to the report data, but we didn't follow up on the security benefits (e.g prevent DOS, or something worse)
Due to implementation complexity we plan to remove the AccountKey hash from the TEE report_data and proceed with only the TLS public key. This issue tracks a follow-up security + feasibility study to decide whether we should add AccountKey back later, and if so, how.

see discussion :
https://nearone.slack.com/archives/C07UW93JVQ8/p1755499467319929

Context / Background

Current PR under review: #855 (includes pieces from #839).
Related cleanup: #866 (removal of public key from contract, per discussion).

Discussion highlights:
Including AccountKey could let the contract restrict sensitive calls to the node running inside a TEE (mitigate operator-initiated DoS, e.g., repeated vote_abort_key_event_instance).

Near-SDK currently does not support enumerating all keys for an account  complicating verification if we don’t store AccountKey.

Alternative considered: have submit_tee_information (or submit_participant_info) optionally accept the node’s public key; trade-offs include no on-chain verification that the provided key belongs to the AccountId.

### User Story

As a product owner I'd like to: get recommendation on whether to:
Keep AccountKey omitted permanently,
Store AccountKey per participant in contract state and include it in report_data.

Questions to Answer

Security benefit:
What concrete DoS or misuse cases are prevented by gating these methods by AccountKey inside TEE?

Candidates:
start_keygen_instance(key_event_id)
start_resharing_instance(key_event_id)
vote_pk(key_event_id, public_key)
vote_reshared(key_event_id)
vote_abort_key_event_instance

Can an operator still achieve a comparable DoS via other contract interactions even with AccountKey gating?

Feasibility & cost:
What contract/state changes are required to store AccountKey per participant?
how will it affect the operator ability to call submit_participant_info, and how can we make sure the operator really owns the public key he submits.

### Acceptance Criteria

A short design doc  that includes:

- Threat model deltas (with/without AccountKey).
- DoS scenarios matrix and likelihood/impact.
- Implementation sketch(es) for each option with state layout, API changes.

- Recommendation for the option chosen.

### Resources & Additional Notes

see discussion :
https://nearone.slack.com/archives/C07UW93JVQ8/p1755499467319929

[...]

Issue #857, opened 2025-08-18
High level design document for MPC node TEE integration


netrome:
### Background

While we've been working for some time on the TEE integration for MPC nodes, we still haven't consolidated our current knowledge into a coherent and simple high level design outline. We're now approaching the point where we want to provide good resources for external operators with resources to help them run nodes within TEEs. While working on an external operator guide, we've found that this setup is pretty complex and our operators would benefit from having a clear mental model of the MPC nodes and the major TEE components.

Beyond the node operators, this would also help us in the team maintain our knowledge of the system and high level architecture.

### User Story

As a node operator, I'd like to have a design document available to help me build a high level mental model over the MPC system when running in TEEs.

### Acceptance Criteria

We have a design document that concisely and accurately outlines the main components of the MPC system when running within TEEs, their responsibilities and how they interact to ensure the integrity of the MPC system after the TEE integration.

### Resources & Additional Notes

TEE integration research resources in outline: https://docs.nearone.org/doc/tee-research-6AAZfSqYlk
Internal node setup guide with dstack: https://docs.google.com/document/d/1OO2hGEx0WuC_H2R56fQ6mnK-HO4xpwy5qpYZMesAZrc/
External guide for MPC node deployment in TDX: https://docs.google.com/document/d/1SEMIo3YmZ_vgVpqG8N0NqNi52MAqUDqNnVSHWrM4cRk

PR #877, opened 2025-08-19
docs: document disaster recovery plan

### LINKED ISSUE #814: Disaster Recovery Documentation
### Background

Formally type up and review the disaster recovery plan discussed in https://nearone.slack.com/archives/C07UW93JVQ8/p1753830474083739

### User Story

_No response_

### Acceptance Criteria

A clear proposal on disaster recovery, as well as follow-up issues for the implementation.

_No response_

kevindeforth:
Resolves #814


TEE.md
```diff
@@ -0,0 +1,79 @@
...
+On a high-level, the back-up mechanism works like this:
+
+1. The node running inside the TEE and the back-up recipient must agree on a key used to encrypt the secrets.
+2. The node encrypts the sensitive secret shares with the derived key.
+3. The node exports the encrypted secret shares.
+
+#### Agreeing on an encryption key
+A pair-wise key establishement scheme can be leveraged to establish a symmetric key. A suitable implementation can be found in [libsodium](https://doc.libsodium.org/) and its rust wrapper [libsodium-rs](https://docs.rs/crate/libsodium-rs/latest). Specifically, the **key exchange protocol based on X25519** would be a good fit for the given set-up (c.f. [libsodium](https://doc.libsodium.org/key_exchange) and [libsodium-rs](https://docs.rs/libsodium-rs/latest/libsodium_rs/crypto_kx/index.html)).
+
+_Note: This key exchange protocol is on Curve25519, which is different from the Edwards curve used by NEAR._
```
netrome, line 37: Why do we use a different curve? Any Diffie-Hellman based key exchange should work for most curves anyway.

That said, Curve25519 is a nice curve and I'm happy to use it - but if we already have keys on another curve it makes more sense to reuse them.

TEE.md
```diff
@@ -0,0 +1,79 @@
...
+
+The key establishment scheme requires that the node as well as the back-up recipient have *mutually authenticated Curve25519 public keys*. The NEAR blockchain can be leveraged for this, more specifically:
+- the node operator whitelists a NEAR-account associated to the back-up recipient in the MPC smart contract.
+- the back-up recipient submits a public key (Curve25519) to the MPC smart contract. The smart contract only accepts it if the account has been whitelisted by the node operator.
+- the node submits a public key (Curve25519) to the MPC smart contract. The smart contract only accepts it if the public key submitting this transaction is an access key of the node operators account.
+
+_Note: Curve25519 and Edwards curve are [birationally equivalent](https://crypto.stackexchange.com/questions/43013/what-does-birational-equivalence-mean-in-a-cryptographic-context), so one could theoretically use NEAR account keys. But it seems cleaner if we just have the node and the back-up recipient generate dedicated keys for this._
```
netrome, line 44: Isn't Edwards curves a family of curves[^1]? Which one do we refer to here? Afaik you can define Edwards curve's over many different fields.

[^1]: https://en.wikipedia.org/wiki/Edwards_curve

TEE.md
```diff
@@ -0,0 +1,79 @@
...
+
+#### Encrypting the secret
+The node running inside the TEE monitors the contract and can derive an encryption key using its secret key and the public key submitted by the back-up recipient.
+
+#### Exporting the encrypted secret
+The back-up service triggers the export of the back-up through a call to the smart contract or a dedicated web-endpoint.
+The node returns the encrypted key either:
+- as a response to a request to its web-endpoint;
+- by printing its hex-value to the logs;
+- or through some other mechanism (@Barak, are there alternatives?)
```
netrome, line 54: I'd prefer to just default to an HTTP endpoint in the initial design. If there are any problems with this, we can always make a follow-up to update the design - but I don't see any big risks with just plain HTTP. We could require some auth header if we want to protect the endpoint, but the key share is encrypted in either case so even a public endpoint wouldn't be a problem afaik.

TEE.md
```diff
@@ -0,0 +1,79 @@
...
+
+The back-up mechanism can then decrypt the received ciphertext and store the secret somewhere safe.
+
+_Todo: improve this section with more details. It is vague._
```
netrome, line 58: I don't think we need to elaborate more here. Conventional databases has protected sensitive data for several decades. Though I'd recommend keeping any data-at-rest encrypted and only decrypt when the data is in-flight.

```suggestion
The back-up mechanism can then decrypt the received ciphertext to verify the received shares, before storing the encrypted shares in a secure database.
```

TEE.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Restoration Mechanism:
+#### Encryption Key Agreement & Encrypting the secret 
+Same as above, just reversed roles
+
+#### Importing the encrypted secret into the node
+Similar to above, but reversed roles. The node makes a web request to the back-up service. (@Barak is this possible?)
+_Todo: improve this section with more details. It is vague._
```
netrome, line 66: I don't see why this shouldn't be possible.

However, it would be nice if the backup service didn't expose any HTTP endpoint but instead the backup service posts the data to an endpoint on the node. This way, the node doesn't have to be aware of the backup service, making for a more loosely coupled design.

This also allows for an initial design where the backup service is just a simple CLI saving the shares to disk, which perhaps could be a good-enough MVP for the initial soft-launch.

TEE.md
```diff
@@ -0,0 +1,79 @@
...
+
+
+### Notes:
+- keep the web-endpoint private. Should we use a new webserver for that, on a different port?
```
netrome, line 70: I don't think any of the endpoints needs to be private. But it could be wise to add authentication to the recovery endpoints just to be safe.

TEE.md
```diff
@@ -0,0 +1,79 @@
...
+- instead of relying on web-endpoints, could we use TLS keys & connect via network, or connect via unix sockets?
+- for the hard-launch, the back-up service will need to run inside a TEE and must pass attestation verification on the contract.
```
netrome, line 72: In the same contract I take it. I guess we'll need to elaborate more here, but we don't have to do this now as we're still lacking design docs for the original attestation flow (TBA in https://github.com/near/mpc/issues/857).

netrome (commented):
I like it! Some nits and comments so far from me. The most significant high-level feedback is that I think we should start with an initial design revolving around adding HTTP endpoints on the MPC nodes. If there's any reason to later upgrade the transport layer, we can do that as a follow-up, but the way I see it:

1. We'll likely not have any problems with using HTTP for this.
2. Keeping the initial design lean and updating it as we uncover things is likely faster than keeping more options open for a longer time.

TEE.md
```diff
@@ -0,0 +1,79 @@
...
+## Overview
+A trusted execution environment (TEE) is an environment isolated from the operating system. It provides security guarantees about confidentiality and integrity of the code and memory executed inside.
+
+For the MPC network, the security guarantees provided by TEEs are attractive mainly because they allow the use of more relaxed threat models, unlocking performant MPC protocols.
+
+TEEs provide their security guarantees by restricting how anything outside of the TEE can interact with the code running inside the TEE. This is great to protect against malicious actors, but it also restricts the honest actors. It has to be expected that debugging and handling of emergencies will become much more difficult compared to running an MPC node outside of a TEE.
+
+This is a good moment to remind the reader that threshold crypotgrapy requires `threshold` operational nodes. - anything less than that, and the protocol finds itself incapacitated, not able to move any funds.
+
+Therefore, Near one will roll-out their TEE implementation in two phases:
```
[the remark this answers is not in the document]
kevindeforth: Thanks! I actually prefer to not write in first-person for documents like this.

TEE.md
```diff
@@ -0,0 +1,79 @@
...
+- Soft Launch: node operators are invited to try out and run their node inside a TEE. This will happen in a coordinated manner.
```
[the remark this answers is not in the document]
kevindeforth: Oh, thanks, completely missed that!
I guess this makes sense.

TEE.md
kevindeforth: For now, it probably has to be the same contract. In the future, we could add a different contract for that.
The main challenge for this would most likely be on the node. Idk how easy it is with the current code to monitor multiple contracts, but I would expect it to require quite a bit of effort.

TEE.md
kevindeforth: >  it would be nice if the backup service didn't expose any HTTP endpoint but instead the backup service posts the data to an endpoint on the node. 

Oh yeah, that would be way nicer!

TEE.md
```diff
@@ -0,0 +1,79 @@
...
+Same as above, just reversed roles
+
+#### Importing the encrypted secret into the node
+Similar to above, but reversed roles. The node makes a web request to the back-up service. (@Barak is this possible?)
+_Todo: improve this section with more details. It is vague._
+
+
+### Notes:
+- keep the web-endpoint private. Should we use a new webserver for that, on a different port?
+- instead of relying on web-endpoints, could we use TLS keys & connect via network, or connect via unix sockets?
```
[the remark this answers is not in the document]
kevindeforth: > Do you mean using the P2P network?

I meant the node and the back-up service could establish a TCP connection. The back-up service could connect to the node via their IP and a specified port.

This gives the node operator more control over when to allow exports. They could block the port most of the time and only allow connections on it when they actually want to make a back-up. 

It wold also simplify some of the logic related to the encryption. With this, the back-up service could simply share its TLS key on the contract.

> An unix socket would force any user to have direct access to the same machine the node runs on. That feels like an unnecessary constraint.

Yeah, I thought that was a feature, more than a bug. 

>  I see no reason not to just use an HTTP endpoint for this.
This service is, despite all the precautions, quite sensitive. My main concern is for it to open a DoS attack vector or worse, get exploited in such a way that unauthorized third parties receive a key-share (encrypted or plaintext, I would like to avoid either).

The web endpoint is assumed to not require too much protection. At some point, we also discussed if we should ask operators to expose it, s.t. we can collect health metrics.

Those are the reasons why I would like this service to be somewhat separate from our existing web-endpoints.

TEE.md
kevindeforth: good point! https://github.com/near/mpc/pull/877/commits/731460722abfeb726d5c1ae58aa5f46c104c4e88

TEE.md
```diff
@@ -0,0 +1,77 @@
...
+
+_Note: This key exchange protocol is on Curve25519, which is different from the Edwards curve used by NEAR._
+
+The key establishment scheme requires that the node as well as the back-up recipient have *mutually authenticated Curve25519 public keys*. The NEAR blockchain can be leveraged for this, more specifically:
+- the node operator whitelists a NEAR-account associated to the back-up recipient in the MPC smart contract.
+- the back-up recipient submits a public key (Curve25519) to the MPC smart contract. The smart contract only accepts it if the account has been whitelisted by the node operator.
+- the node submits a public key (Curve25519) to the MPC smart contract. The smart contract only accepts it if the public key submitting this transaction is an access key of the node operators account.
```
barakeinav1, line 44: thinking out loud here: 
is it important the access key be owned (e.g private key available ) explicitly by the node and not the operator?

TEE.md
barakeinav1: we can make the endpoint accessible on the local machine.
but from security perspective, we really anyway on cryptography so it doesn't really matter.

TEE.md
barakeinav1: for the soft launch it doesn't really matter.
for the hard launch  it will.
so now we have a concrete use case for putting the account key in the report data :-)

TEE.md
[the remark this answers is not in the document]
barakeinav1: we already asked operator to make the web endpoint public.
peter and andre specifically asked for this in the operators guide as well.

I also see the sense of having a different endpoint /communication channel for the recovery mechanism in order to reduce the attack surface.

TEE.md
netrome: Well, you know more than I. I thought Curve25519 was the twisted Edwards curve used in the Ed25519 scheme - but TIL they are different curves.

TEE.md
netrome: Perfect! I commented in the research issue: https://github.com/near/mpc/issues/875#issuecomment-3204816474

TEE.md
```diff
@@ -0,0 +1,77 @@
...
+- Hard Launch: All mainnet nodes are running within TEEs. Their key shares are backed-up outside of inside a different TEE.
+
+In order to protect the network from the worst case scenario (complete loss of funds due to loss of key shares or an incapacitated network), a disaster recovery plan is prepared.
+
+
+## Disaster Recovery
+Distaster recovery is a plan intended to prevent a permanent loss of the signing quorum.
+As long as the secret shares of the MPC nodes are securely backed-up outside of the TEE environment in which the node is running, it highly likely that the network will be able to recover from otherwise catastrophic events.
+
+Therefore, the disaster recovery plan aims at establishing a secure mechanism allowing to back-up the secret shares held by the nodes. It encompasses two steps:
```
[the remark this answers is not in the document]
kevindeforth: I was purposefully vague here. Wanted to avoid insinuating the operators would gain access to plaintext shares.
Once we have everything set-up, it would be also possible to make periodic back-up, in which case the operators would be more or less out of the loop.

TEE.md
```diff
@@ -0,0 +1,77 @@
...
+- the node operator whitelists a NEAR-account associated to the back-up recipient in the MPC smart contract.
+- the back-up recipient submits a public key (Curve25519) to the MPC smart contract. The smart contract only accepts it if the account has been whitelisted by the node operator.
+- the node submits a public key (Curve25519) to the MPC smart contract. The smart contract only accepts it if the public key submitting this transaction is an access key of the node operators account.
+
+_Note: Curve25519 and edwards25519 are [birationally equivalent](https://crypto.stackexchange.com/questions/43013/what-does-birational-equivalence-mean-in-a-cryptographic-context), so one could theoretically use NEAR account keys. But it seems cleaner if we just have the node and the back-up recipient generate dedicated keys for this._
+
+#### Encrypting the secret
+The node running inside the TEE monitors the contract and can derive an encryption key using its secret key and the public key submitted by the back-up recipient.
```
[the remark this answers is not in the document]
kevindeforth: Nice one, I love these :D 

So I guess it is:
- the backup
- the backup recipient
- to back up

But is it
- _the back up service_ or _the backup service_?
- _the back up mechanism_ or _the backup mechanism_?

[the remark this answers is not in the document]
DSharifi: I would go with "backup" for the last two.

Checking with my friend Claude as well:
```
The key test is: if you can substitute "backup" with another adjective like "emergency" or "secondary," then it's one word. If it's describing an action (what someone does), then it's two words.
Examples:

"The backup system failed" → "The emergency system failed" ✓
"We need to back up the data" → "We need to emergency the data" ✗ (this is the action)

So in your cases, since "service" and "mechanism" are nouns being described by "backup," it's definitely one word!
```

TEE.md
```diff
@@ -0,0 +1,77 @@
+# TEE Integration
+
+## Overview
+A trusted execution environment (TEE) is an environment isolated from the operating system. It provides security guarantees about confidentiality and integrity of the code and memory executed inside.
```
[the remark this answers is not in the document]
DSharifi: You're right about the TEE + environment is awkward, let's not do that.


I'm fine with both `within` and `inside`. The nit is that the sentence ending is a bit awkward since we don't specify "within" what ...?

```suggestion
A trusted execution environment (TEE) is an environment isolated from the operating system. It provides security guarantees about confidentiality and integrity of the code and memory executed within the TEE.
```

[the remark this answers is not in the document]
DSharifi: I think this reads even better
```suggestion
A trusted execution environment (TEE) is an environment isolated from the operating system. TEEs provide security guarantees about confidentiality and integrity of the code and memory executed inside it.
```

TEE.md
```diff
@@ -0,0 +1,77 @@
...
+3. The node exports the encrypted secret shares.
+
+#### Agreeing on an encryption key
+A pair-wise key establishment scheme can be leveraged to establish a symmetric key. A suitable implementation can be found in [libsodium](https://doc.libsodium.org/) and its rust wrapper [libsodium-rs](https://docs.rs/crate/libsodium-rs/latest). Specifically, the **key exchange protocol based on X25519** would be a good fit for the given set-up (c.f. [libsodium](https://doc.libsodium.org/key_exchange) and [libsodium-rs](https://docs.rs/libsodium-rs/latest/libsodium_rs/crypto_kx/index.html)).
+
+_Note: This key exchange protocol is on Curve25519, which is different from the Edwards curve used by NEAR._
+
+The key establishment scheme requires that the node as well as the backup recipient have *mutually authenticated Curve25519 public keys*. The NEAR blockchain can be leveraged for this, more specifically:
+- the node operator whitelists a NEAR-account associated to the backup recipient in the MPC smart contract.
+- the backup recipient submits a public key (Curve25519) to the MPC smart contract. The smart contract only accepts it if the account has been whitelisted by the node operator.
```
pbeza, line 43: How do we make sure the public key on the contract is always up to date? Is it possible to end up in a situation where the backup provider or recipient generates a new public key (e.g., after a restart), so the backup is saved but can no longer be decrypted because it was encrypted using keys that are no longer accessible?

TEE.md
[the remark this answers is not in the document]
netrome: > Yeah, I thought that was a feature, more than a bug.

It's also a risk. The backup service should not run on the same physical or virtual machine as the node.

TEE.md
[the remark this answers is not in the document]
kevindeforth: a single verb remained. https://github.com/near/mpc/pull/877/commits/101820398a5ca411e3a0e738657e3933387d51f8

TEE.md
```diff
@@ -0,0 +1,77 @@
...
+2. The node encrypts the sensitive secret shares with the derived key.
+3. The node exports the encrypted secret shares.
+
+#### Agreeing on an encryption key
+A pair-wise key establishment scheme can be leveraged to establish a symmetric key. A suitable implementation can be found in [libsodium](https://doc.libsodium.org/) and its rust wrapper [libsodium-rs](https://docs.rs/crate/libsodium-rs/latest). Specifically, the **key exchange protocol based on X25519** would be a good fit for the given set-up (c.f. [libsodium](https://doc.libsodium.org/key_exchange) and [libsodium-rs](https://docs.rs/libsodium-rs/latest/libsodium_rs/crypto_kx/index.html)).
+
+_Note: This key exchange protocol is on Curve25519, which is different from the Edwards curve used by NEAR._
+
+The key establishment scheme requires that the node as well as the backup recipient have *mutually authenticated Curve25519 public keys*. The NEAR blockchain can be leveraged for this, more specifically:
+- the node operator whitelists a NEAR-account associated to the backup recipient in the MPC smart contract.
```
pbeza, line 42: > - the node operator whitelists a NEAR-account associated to the backup recipient in the MPC smart contract.

Do I understand correctly that node operators will effectively need to manually select — by whitelisting them in the contract — which other nodes will act as backup recipients for their node? Also, can they select multiple backup recipients?

kevindeforth: I will make this clearer in the next commit.
The back-up service is independent from the nodes. They are completely separate binaries and may run on different machines.
There is a 1-1 relationship between nodes and back-up services.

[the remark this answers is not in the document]
kevindeforth: Hmm, I don't quite understand this part:
>  since every node is expected to eventually act as a backup recipient, couldn’t all of them simply submit their Curve25519 public keys unconditionally when joining the MPC network?

The back-up service is separate from the node.

TEE.md
kevindeforth: The key used for encrypting the secret at rest is different from the key used for encrypting the secret in transit.

TEE.md
```diff
@@ -0,0 +1,211 @@
...
+
+For the MPC network, the security guarantees provided by TEEs are attractive for two reasons:
+1. They relax the threat models (e.g. honest-but-curious instead of malicious adversaries). This allows the adoption of significantly more efficient MPC protocols.
+2. They help enforce backward secrecy. Since TEEs can guarantee that former nodes never gain lasting possession of plaintext secret-shares, collusion attacks after departure become infeasible.
+
+TEEs provide their security guarantees by restricting how anything outside of the TEE can interact with the code running inside the TEE. This is great to protect against malicious actors, but it also restricts the honest actors. It has to be expected that debugging and handling of emergencies will become much more difficult compared to running an MPC node outside of a TEE.
+
+This is a good moment to remind the reader that threshold cryptography requires threshold operational nodes. Anything less than that, and the protocol stalls, in which case, no funds can be moved.
```
gilcu3, line 12:
```suggestion
This is a good moment to remind the reader that threshold cryptography requires n > threshold operational nodes. Anything less than that, and the protocol stalls, in which case, no funds can be moved.
```
or something similar to make it slightly more readable

TEE.md
netrome: > may run on different machines.

**Should** run on different machines.

netrome (approved):
Great stuff. I'm also getting more warmed up to the idea of the TLS solution, as long as we can implement it cleanly. Though I still think the HTTP based variant might be easier, but would love to be proven wrong here.

TEE.md
```diff
@@ -0,0 +1,203 @@
...
+If the contract is in `Recovery` state, then the backup service follows the protocol outlined under [Recovery](#recovery)
+
+##### Backup 
+1. The backup service looks up the details of its MPC node in the set of participants of the `Running` protocol state.
+2. The backup service establishes a p2p connection with the node with mutual TLS authentication.
+3. The backup service waits for the node to send the key shares, stores them securely and then closes the connection.
+
+##### Recovery
+1. The backup service looks up the details of its MPC node in the set of prospective participants of the `Recovery` protocol state.
+2. the backup service establishes a p2p connection with the node with mutual TLS authentication.
```
barakeinav1, line 85: after moving to recovery, the node may (or may not) need update the TLS on the contract. Is there some signal so that the backup service will try to establish a session with the node?

Maybe it should be the other way around, that the node established the p2P connection with the backup service.

TEE.md
```diff
@@ -0,0 +1,203 @@
...
+        - Submit their information to `submit_participant_info`.  
+        - Wait to be included in the protocol via a `resharing`.  
+
+    - **If** the account ID of their node operator **is** in the set of current participants **and** the TLS key of the participant info matches theirs:  
+        - Submit their information to `submit_participant_info`.  
+        - Engage normally with the protocol.  
+        
+    - **If** the account ID of their node operator **is** in the set of current participants, **but** the TLS key of the participant info does **not** match theirs:  
+        - Submit their information to `submit_participant_recovery_info`. 
+        - This puts the protocol in `Recovery` mode (c.f. also the [contract](#contract) section).
```
barakeinav1, line 104: I wandering if now one operator can cause denlie of service more easily than before?
this can be avoided if we enforce that only the node key can trigger moving to recovery.

TEE.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Backup service
+The backup service has two responsibilities:
+1. _Backup:_ Securely fetch the secret shares of the current key set from the MPC node it must back up and store the secrets in a secure manner.
+2. _Recovery:_ Securely provide the backup shares to the MPC node if required.
+
+For the hard-launch, the backup service must run inside a TEE.
+
+The backup service needs a current view of the MPC smart contract on the NEAR blockchin. For that reason, it will be operating a NEAR light-client node.
+
```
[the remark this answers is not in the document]
kevindeforth: nice one! https://github.com/near/mpc/pull/877/commits/b3ae57f30b2f1632cb94ab59450d1e1c1bbac395

TEE.md
```diff
@@ -0,0 +1,203 @@
...
+#### Node
+The **MPC node** will expose a web endpoint over which the backup service can submit requests. These endpoints require some sort of authentication using the published public keys _(todo: yet to be specified)_.
+The exposed endpoints are:
+- GET /shares_backup - with an authentication header
+    - Returns the encrypted shares, if a valid backup service is registered.
+- POST /shares_recover
+    - Posts encrypted shares to a new node.
+
+
+## Todo
```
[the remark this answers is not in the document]
kevindeforth: done: 8bdba01d

TEE.md
kevindeforth: > after moving to recovery, the node may (or may not) need update the TLS on the contract.

at that point, that information is already in the contract. The node already submitted it.

TEE.md
kevindeforth: That's a good point.
This should not be a big issue though for two reasons:
1. we allow to enter a `Resharing` state from a `Recovery`, precisely for that reason. If an operator mis-behaves, they will just get kicked out.
2. in the medium term, we will be able to handle signature requests while being in `Recovery` mode, as long as the remaining participants can form a signing quorum.

TEE.md
```diff
@@ -0,0 +1,203 @@
...
+    end
+```
+##### Recovery Protocol
+The node follows this protocol when the protocol state is in `Recovery`:
+
+- **If** this nodes account ID matches the account ID of the recovering node:
+    - **If** this node is the recovering node (matches the TLS key):
+        - They wait for the backup service to provide the keyshares.
+        - Once in possession of key shares, the node calls `conclude_recovery()`, which resumes the protocol in the `Running` state.
+    - **Else** they shut down (_note: This case means that the node is about to be de-commissioned. The node must shut down now, because otherwise, it will itself request a "Recovery" state once the contract resumes `Running`_).
```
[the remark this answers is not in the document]
kevindeforth: Yeah, this can be hard to parse. I made a graph for this, did you get a chance to look at it? https://github.com/near/mpc/blob/8bdba01da17e37bc9ef7b572877d71c8c70d05de/TEE.md#onboarding-1

If still unclear, please follow-up.

TEE.md
```diff
@@ -0,0 +1,203 @@
...
+In order to protect the network from the worst case scenario (permanent loss of the signing quorum), a disaster recovery system is to be deployed.
+
+## Disaster Recovery on a high-level
+The purpose of the disaster recovery system is to prevent a permanent loss of the signing quorum.
+As long as the secret shares of the MPC nodes are securely backed-up outside of the TEE environment in which the node is running, it is highly likely that the network will be able to recover from otherwise catastrophic failures. Therefore, each operator of an MPC node will be required to operate their own **backup service**. This service is separate from the MPC node and should run on a different machine.
+
+### Backup service
+The backup service has two responsibilities:
+1. _Backup:_ Securely fetch the secret shares of the current key set from the MPC node it must back up and store the secrets in a secure manner.
+2. _Recovery:_ Securely provide the backup shares to the MPC node if required.
```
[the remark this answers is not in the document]
kevindeforth: Appreciate the attention to detail, but I prefer to be explicit here.

TEE.md
```diff
@@ -0,0 +1,203 @@
...
+- `Initializing` for generating new key shares
+- `Running` when handling signature request
+- `Resharing` when the network is resharing their secret keys for a change of the participant set, or for a change of the cryptography threshold
+
+Until now, when a node operator wanted to switch their machine, they needed to do so through a `Resharing` - they had to leave the network and then re-join.
+Unfortunately, it may not be possible to use the same mechanism for recovering from disaster, as by definition, a disaster implies that the network lost it signing quorum and thus, its ability to execute a resharing.
+To account for this, a fourth protocol state must be introduced: `Recovery`.
+
+The purpose of this state is to:
+- allow participants to change their participant information (e.g. TLS keys, IP address, and anything other than their account id);
```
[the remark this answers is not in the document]
kevindeforth: > Why do we allow nodes to change IP address and other information in case of recovery of shares?

Because this way, they can use the back-up service to switch their node from one machine to another.

> Could you add why the account id is the only thing the participants can not change?

there is a paragraph on this here already. The account id is a unique identifier for the node operator: 

https://github.com/near/mpc/blob/8bdba01da17e37bc9ef7b572877d71c8c70d05de/TEE.md?plain=1#L92

PR #878, opened 2025-08-20
chore: integrate the attestation module into the contract code

### LINKED ISSUE #770: Integrate the `attestation` module into the contract code
### Background

Now that the following issues are resolved:
- https://github.com/near/mpc/issues/642
- https://github.com/near/mpc/issues/643
- https://github.com/near/mpc/issues/644

and the [`attestation`](https://github.com/near/mpc/tree/main/attestation) module is complete, we need to replace the contract’s existing TEE-related logic with the one from the `attestation` module.

### User Story

As a developer, I’d like to avoid duplicating code between the `attestation` module and the contract. Instead, I want to rely solely on the `attestation` module, which is better tested and more robust than the old implementation in the contract code.

### Acceptance Criteria

No duplicated TEE-related code between the `attestation` module and the contract. The contract should use the `attestation` module for all TEE-related logic.

https://github.com/near/mpc/pull/758 needs to be merged first, as it’s a prerequisite for this issue.

DSharifi:
closes #770


libs/chain-signatures/contract/readme.md
```diff
@@ -196,7 +196,7 @@ These functions require the caller to be a participant or candidate.
...
 | `propose_update(args: ProposeUpdateArgs)`                                           | Proposes an update to the contract, requiring an attached deposit.                                                                                                                                                                      | `Result<UpdateId, Error>` | TBD             | TBD                |
 | `vote_update(id: UpdateId)`                                                         | Votes on a proposed update. If the threshold is met, the update is executed.                                                                                                                                                            | `Result<bool, Error>`     | TBD             | TBD                |
-| `propose_join(proposed_tee_participant: TeeParticipantInfo, signer_pk: PublicKey)` | Submits the tee participant info for a potential candidate. c.f. TEE section | `Result<(), Error>` | TBD | TBD |
+| `propose_join(proposed_tee_participant: DstackAttestation, signer_pk: PublicKey)`   | Submits the tee participant info for a potential candidate. c.f. TEE section | `Result<(), Error>` | TBD | TBD |
```
DSharifi, line 199: The type is actually the enum 
https://github.com/near/mpc/blob/f15568d06ce397ed0375dcbcf8cd03fb913db841/attestation/src/attestation.rs#L36-L40
`DstackAttestation` type is technically true in production, as it will be the only accepted enum variant and we won't allow the `Local` variant.

I'm not sure if we should keep this as is, or reflect the source code.

DSharifi: Isn't this the purpose of the ABI generation we now support?

node/src/indexer/types.rs
```diff
@@ -136,7 +136,7 @@ pub(crate) enum ChainSendTransactionRequest {
     VoteAbortKeyEventInstance(ChainVoteAbortKeyEventInstanceArgs),
     VerifyTee(),
     #[cfg(feature = "tee")]
-    SubmitParticipantInfo(SubmitParticipantInfoArgs),
+    SubmitParticipantInfo(Box<SubmitParticipantInfoArgs>),
```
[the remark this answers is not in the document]
netrome: Fair enough. I'd consider skipping the box here too but I don't think it matters that much either way. This enum shouldn't be in any hot path anyway afaik.

PR #885, opened 2025-08-20
feat: add CKD support to the MPC contract

### LINKED ISSUE #860: CKD smart contract methods
### Background

As part of the [Confidential Key Derivation](https://github.com/near/threshold-signatures/blob/main/docs/confidential_key_derivation.md) functionality (CKD), the MPC contract needs to have methods to provide Near Accounts with (encrypted) deterministic secrets.

### User Story

_No response_

### Acceptance Criteria

The MPC contract contains methods to support the CKD functionality:
```rust
pub request_app_private_key(request: CKDrequest);
pub respond_app_private_key(request: CKDrequest, response: CKDresponse);
```

https://docs.nearone.org/doc/low-level-system-design-GcD8YKiczY

gilcu3:
Closes #860 

- A few structures of signatures were unified with the ckd ones.

### DIFF
--- a/libs/chain-signatures/contract/src/storage_keys.rs
+++ b/libs/chain-signatures/contract/src/storage_keys.rs
@@ -10,7 +10,6 @@ pub enum StorageKey {
     /// Proposed updates to the contract code and config.
     _DeprecatedProposedUpdatesEntries,
     _DeprecatedRequestsByTimestamp,
-    /// Pending requests.
     PendingSignatureRequestsV2,
     ProposedUpdatesEntriesV2,
     ProposedUpdatesVotesV2,


libs/chain-signatures/contract/src/lib.rs
```diff
@@ -428,6 +450,93 @@ impl VersionedMpcContract {
...
+                    log!("refund excess deposit {diff} to {predecessor}");
+                    Promise::new(predecessor.clone()).transfer(diff);
+                }
+            }
+        }
+
+        // TODO: do we need this here?
+        let Self::V2(mpc_contract) = self else {
+            env::panic_str("expected V2")
+        };
```
gilcu3, line 510: This is a check for V2 contract version. Afaiu it should not be needed here. If that's the case and can be removed, let me know.

libs/chain-signatures/contract/src/primitives/ckd.rs
```diff
@@ -0,0 +1,25 @@
+use near_sdk::{near, AccountId};
+
+// TODO: do we need the Default trait here?
+#[derive(Clone, Debug)]
+#[near(serializers=[json])]
+pub struct CKDRequestArgs {
+    pub app_public_key: near_sdk::PublicKey,
```
gilcu3, line 7: This could be a `SerializableAffinePoint` instead, making the ckd API more uniform. Let me know which one should be preferable. The good thing about this is that it is a much simpler type, the bad is that it includes edwards curves as well. Using k256_types::PublicKey is also an option, but would require adding all the serialization code for that type as well.

libs/chain-signatures/contract/src/lib.rs
kevindeforth: I think it's needed to extract the `mpc_contract` and check the `accept_requests` flag below.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -1080,10 +1225,36 @@ impl VersionedMpcContract {
...
+        match ck {
+            Ok(ck) => PromiseOrValue::Value(ck),
+            Err(_) => {
+                mpc_contract.pending_ckd_requests.remove(&request);
+                let promise = Promise::new(env::current_account_id()).function_call(
+                    "fail_on_timeout".to_string(),
+                    vec![],
+                    NearToken::from_near(0),
+                    Gas::from_tgas(2), // TODO: shouldn't this be a constant somewhere?
```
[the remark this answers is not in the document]
kevindeforth: yeah, lets make it a constant.

libs/chain-signatures/contract/src/primitives/ckd.rs
kevindeforth: what would be your preference?

libs/chain-signatures/contract/src/storage_keys.rs
```diff
@@ -10,7 +10,7 @@ pub enum StorageKey {
     /// Proposed updates to the contract code and config.
     _DeprecatedProposedUpdatesEntries,
     _DeprecatedRequestsByTimestamp,
-    /// Pending signature requests.
+    /// Pending requests.
```
kevindeforth, line 13: need to add a storage key for the ckd requests (at the end of the enum).
https://docs.near.org/smart-contracts/anatomy/collections#instantiation

libs/chain-signatures/contract/tests/common.rs
```diff
@@ -512,6 +522,95 @@ pub async fn sign_and_validate(
...
+fn hash2curve(app_id: &[u8]) -> ProjectivePoint {
+    const DOMAIN: &[u8] = b"NEAR CURVE_XOF:SHAKE-256_SSWU_RO_";
+    <Secp256k1 as GroupDigest>::hash_from_bytes::<ExpandMsgXof<sha3::Shake256>>(
+        &[app_id],
+        &[DOMAIN],
+    )
+    .unwrap()
+}
+
+/// Derives a confidential key following https://github.com/near/threshold-signatures/blob/main/docs/confidential_key_derivation.md
```
kevindeforth, line 539: nice! I like that you chose to not rely on the threshold-signatures implementation. This should catch regressions!

libs/chain-signatures/contract/tests/ckd.rs
```diff
@@ -0,0 +1,268 @@
...
+    let (_, sk) = worker.generate_dev_account_credentials();
+    worker.create_root_account_subaccount(account_id, sk).await
+}
+
+#[tokio::test]
+async fn test_contract_ckd_request() -> anyhow::Result<()> {
+    let (worker, contract, _, sks) = init_env_secp256k1(1).await;
+    let sk = match &sks[0] {
+        common::SharedSecretKey::Secp256k1(sk) => sk,
+        common::SharedSecretKey::Ed25519(_) => unreachable!(),
```
kevindeforth, line 22: Noob question: do we support this for ed25519?

kevindeforth (commented):
Good work!
The storage key is the only issue I see. But will review again to make sure.

libs/chain-signatures/contract/src/primitives/ckd.rs
gilcu3: I think I prefer SerializableAffinePoint, will change the code to reflect that if you have nothing against it

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -1038,9 +1178,16 @@ impl VersionedMpcContract {
         }
     }
 
-    pub fn get_pending_request(&self, request: &SignatureRequest) -> Option<YieldIndex> {
+    pub fn get_pending_signature_request(&self, request: &SignatureRequest) -> Option<YieldIndex> {
```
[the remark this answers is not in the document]
gilcu3: Alright, I think the name will be a bit misleading, but I will rename it back, no need to do this on this PR.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -996,10 +1135,11 @@ impl VersionedMpcContract {
             protocol_state: ProtocolContractState::Running(RunningContractState::new(
                 domains, keyset, parameters,
             )),
-            pending_requests: LookupMap::new(StorageKey::PendingRequestsV2),
+            pending_signature_requests: LookupMap::new(StorageKey::PendingRequestsV2),
+            pending_ckd_requests: LookupMap::new(StorageKey::PendingRequestsV2),
```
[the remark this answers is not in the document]
gilcu3: I was pondering about that but couldn't figure why. Could you explain a bit?

libs/chain-signatures/contract/src/lib.rs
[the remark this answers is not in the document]
gilcu3: Given that CI did not fail when I changed the name, and that this function is used by the indexer:

https://github.com/near/mpc/blob/b89d1084bcbd2fdc777140a4dda38de616b810ef/node/src/indexer/tx_sender.rs#L102

I think we are missing a test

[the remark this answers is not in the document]
gilcu3: I just searched in the Node pytests [here](https://github.com/near/mpc/actions/runs/17108677640/job/48524202345) and couldn't find that warning. But maybe it is just not visible, I have found several places where tracing logs are not visible.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -425,6 +447,95 @@ impl VersionedMpcContract {
...
+        // TODO: do we need this here?
+        let Self::V2(mpc_contract) = self else {
+            env::panic_str("expected V2")
+        };
+
+        if !mpc_contract.accept_requests {
+            env::panic_str(&TeeError::TeeValidationFailed.to_string())
+        }
+
+        env::log_str(&serde_json::to_string(&near_sdk::env::random_seed_array()).unwrap());
```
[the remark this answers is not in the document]
gilcu3: open issue https://github.com/near/mpc/issues/918 as nobody in the team seems to know

PR #887, opened 2025-08-20
fix(tee): ensure participants run valid TEE or none

### LINKED ISSUE #843: Ensure participants run valid tee or none before starting a resharing
### Background

In`vote_new_parameters`, we currently do not check the tee status of prospective participants.
https://github.com/near/mpc/blob/ffe1004e49988eb7331e3487f33a0f5b45128e74/libs/chain-signatures/contract/src/lib.rs#L595

We should allow only participants with a valid Tee status or `None`

### User Story

As a node user, I want to know that participants  running inside a tee only receive a valid keyshare if they pass the attestation verification.

### Acceptance Criteria

Verify Tee attestations inside `propose_new_parameters`. Only accept votes if the verification passes.

https://github.com/near/mpc/issues/841

pbeza:
Fixes #843

### DIFF
--- a/libs/chain-signatures/contract/src/lib.rs
+++ b/libs/chain-signatures/contract/src/lib.rs
@@ -605,9 +605,27 @@ impl VersionedMpcContract {
             env::signer_account_id(),
             proposal,
         );
+
+        // Validate TEE status of all prospective participants
+        // Only allow participants with Valid or None TEE status
         match self {
             Self::V2(mpc_contract) => {
-                mpc_contract.vote_new_parameters(prospective_epoch_id, &proposal)
+                let validation_result =
+                    mpc_contract.tee_state.validate_tee(proposal.participants());
+                let final_proposal = match validation_result {
+                    TeeValidationResult::Full => proposal,
+                    TeeValidationResult::Partial(valid_participants) => {
+                        if valid_participants.len() == 0 {
+                            return Err(InvalidParameters::InvalidTeeRemoteAttestation.message(
+                                "No participants with valid TEE status found".to_string(),
+                            ));
+                        }
+                        ThresholdParameters::new(valid_participants, proposal.threshold().clone())
+                            .map_err(|e| InvalidParameters::MalformedPayload.message(e.to_string()))?
+                    }
+                };
+
+                mpc_contract.vote_new_parameters(prospective_epoch_id, &final_proposal)
             }
             _ => env::panic_str("expected V2"),
         }
--- a/libs/chain-signatures/contract/src/lib.rs
+++ b/libs/chain-signatures/contract/src/lib.rs
@@ -616,11 +616,12 @@ impl VersionedMpcContract {
                     TeeValidationResult::Full => {
                         mpc_contract.vote_new_parameters(prospective_epoch_id, &proposal)
                     }
-                    TeeValidationResult::Partial(_) => {
-                        Err(InvalidParameters::InvalidTeeRemoteAttestation.message(
-                            "One or more participants have invalid TEE status".to_string(),
-                        ))
-                    }
+                    TeeValidationResult::Partial(invalid_participants) => Err(
+                        InvalidParameters::InvalidTeeRemoteAttestation.message(format!(
+                            "The following participants have invalid TEE status: {:?}",
+                            invalid_participants
+                        )),
+                    ),
                 }
             }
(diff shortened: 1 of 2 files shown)


libs/chain-signatures/contract/src/lib.rs
```diff
@@ -605,9 +605,25 @@ impl VersionedMpcContract {
...
+                let final_proposal = match validation_result {
+                    TeeValidationResult::Full => proposal,
+                    TeeValidationResult::Partial(valid_participants) => {
+                        if valid_participants.len() == 0 {
+                            return Err(InvalidParameters::InvalidTeeRemoteAttestation.message(
+                                "No participants with valid TEE status found".to_string(),
+                            ));
+                        }
+                        ThresholdParameters::new(valid_participants, proposal.threshold().clone())
+                            .map_err(|e| InvalidParameters::MalformedPayload.message(e.to_string()))?
```
kevindeforth, line 622: I would say we need to reject in case the result is anything other than `Full`. 
Otherwise, the contract is not doing what the caller would expect.
As a operator, I vote for a specific set of participants. If that is not a valid set, then my vote should simply not count.

kevindeforth (changes requested):
Will use this button for this one.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -414,11 +414,12 @@ impl VersionedMpcContract {
...
-    /// Older key versions will always work but newer key versions were never held by older signers
-    /// Newer key versions may also add new security features, like only existing within a secure enclave.
-    /// The signature_scheme parameter specifies which signature scheme we're querying the latest version
-    /// for. The default is Secp256k1. The default is **NOT** to query across all signature schemes.
+    /// Key versions refer new versions of the root key that we may choose to generate on cohort
+    /// changes Older key versions will always work but newer key versions were never held by
+    /// older signers Newer key versions may also add new security features, like only existing
+    /// within a secure enclave. The signature_scheme parameter specifies which signature scheme
+    /// we're querying the latest version for. The default is Secp256k1. The default is **NOT**
+    /// to query across all signature schemes.
```
pbeza, line 422: Sorry for reformatting some unrelated comments — my VSCode is set to wrap at 100 chars. We should probably enforce a comment wrap width at some point. I can revert if you prefer.

barakeinav1 (approved):

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -605,9 +607,21 @@ impl VersionedMpcContract {
             env::signer_account_id(),
             proposal,
         );
+
         match self {
             Self::V2(mpc_contract) => {
-                mpc_contract.vote_new_parameters(prospective_epoch_id, &proposal)
+                let validation_result =
```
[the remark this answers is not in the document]
pbeza: IIUC, we usually don’t extract small functions unless they’re reusable in multiple places.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -605,9 +607,21 @@ impl VersionedMpcContract {
...
+                    mpc_contract.tee_state.validate_tee(proposal.participants());
+                match validation_result {
+                    TeeValidationResult::Full => {
+                        mpc_contract.vote_new_parameters(prospective_epoch_id, &proposal)
+                    }
+                    TeeValidationResult::Partial(_) => {
+                        Err(InvalidParameters::InvalidTeeRemoteAttestation.message(
+                            "One or more participants have invalid TEE status".to_string(),
```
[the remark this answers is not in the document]
pbeza: Fixed: https://github.com/near/mpc/pull/887/commits/f98e460096b7d3f7a96920c43ceecb09af49a3f0.

kevindeforth (approved):
Thank You!

libs/chain-signatures/contract/src/lib.rs
DSharifi: Good spot. We have this invariant check both here now, and also in `running.rs` to process the actual vote count:
https://github.com/near/mpc/blob/72eb2bc9b9be71955fa6ec60cdb70ba402990cf0/libs/chain-signatures/contract/src/state/running.rs#L143-L145

Wonder if we can make this invariant check in a single place so they're always unified.

PR #890, opened 2025-08-20
refactor: Consistent usage of SerializableEdwardsPoint in crypto_shared/types.rs

### LINKED ISSUE #757: EdwardPoint vs SerializableEdwardsPoint
### Background

In libs/chain-signatures/contract/src/crypto_shared/types.rs we use two types for edward points:
SerializableEdwardsPoint and EdwardsPoint. That leads to some code repetition when defining schemas in https://github.com/near/mpc/pull/751#discussion_r2247628502. I think we should use SerializableEdwardsPoint as a proper wrapper of EdwardsPoint everywhere, therefore avoiding code duplication.

### User Story

_No response_

### Acceptance Criteria

A single type is used uniformly

_No response_

netrome:
Closes #757 

This PR simplifies the serialization trait implementations for `PublicKeyExtended` by using `SerializableEdwardsPoint` consistently throughout the module. Additionally, I've broken out the `serialize` submodule to a separate file and implemented more useful traits on the `SerializableEdwardsPoint` type to make using it more ergonomic.

Note: I considered replacing other usages of `EdwardsPoint` with `SerializableEdwardsPoint` in the rest of the code base but I don't see the (Edwards) point of doing that, so I let it be for now.

### DIFF
--- a/libs/chain-signatures/contract/src/crypto_shared/types.rs
+++ b/libs/chain-signatures/contract/src/crypto_shared/types.rs
@@ -48,14 +48,6 @@ pub enum PublicKeyExtended {
         /// Serialized compressed Edwards-y point.
         near_public_key_compressed: near_sdk::PublicKey,
         /// Decompressed Edwards point used for curve arithmetic operations.
-        #[cfg_attr(
-            all(feature = "abi", not(target_arch = "wasm32")),
-            schemars(with = "[u8; 32]"),
-            borsh(schema(with_funcs(
-                declaration = "<[u8; 32] as ::borsh::BorshSchema>::declaration",
-                definitions = "<[u8; 32] as ::borsh::BorshSchema>::add_definitions_recursively"
-            ),))
-        )]
         edwards_point: SerializableEdwardsPoint,
     },
 }
--- a/libs/chain-signatures/contract/src/crypto_shared/types.rs
+++ b/libs/chain-signatures/contract/src/crypto_shared/types.rs
@@ -1,14 +1,15 @@
+pub mod serializable;
+
 use std::fmt::Display;
 
 use borsh::{BorshDeserialize, BorshSerialize};
-use curve25519_dalek::EdwardsPoint;
 use k256::{
     elliptic_curve::{group::GroupEncoding, CurveArithmetic, PrimeField},
     AffinePoint, Secp256k1,
 };
-use near_sdk::near;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
+use serializable::SerializableEdwardsPoint;
 
 #[cfg_attr(
     all(feature = "abi", not(target_arch = "wasm32")),
@@ -37,7 +38,7 @@ pub struct CKDResponse {
     derive(::near_sdk::schemars::JsonSchema),
     derive(::borsh::BorshSchema)
 )]
-#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
 pub enum PublicKeyExtended {
     Secp256k1 {
         near_public_key: near_sdk::PublicKey,
@@ -55,7 +56,7 @@ pub enum PublicKeyExtended {
                 definitions = "<[u8; 32] as ::borsh::BorshSchema>::add_definitions_recursively"
             ),))
         )]
-        edwards_point: EdwardsPoint,
+        edwards_point: SerializableEdwardsPoint,
     },
 }
 
@@ -102,7 +103,7 @@ impl TryFrom<near_sdk::PublicKey> for PublicKeyExtended {
                     .ok_or(PublicKeyExtendedConversionError::PublicKeyLengthMalformed)?
                     .map_err(|_| PublicKeyExtendedConversionError::PublicKeyLengthMalformed)?;
 
-                let edwards_point = curve25519_dalek::EdwardsPoint::from_bytes(public_key_bytes)
+                let edwards_point = SerializableEdwardsPoint::from_bytes(public_key_bytes)
                     .into_option()
                     .ok_or(PublicKeyExtendedConversionError::FailedDecompressingToEdwardsPoint)?;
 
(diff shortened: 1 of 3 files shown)


libs/chain-signatures/contract/src/crypto_shared/types/serializable.rs
```diff
@@ -0,0 +1,82 @@
...
+    Debug,
+    PartialEq,
+    Serialize,
+    Deserialize,
+    Eq,
+    Clone,
+    Copy,
+    derive_more::From,
+    derive_more::AsRef,
+    derive_more::Deref,
```
DSharifi, line 26: Curious, was the qualification vs import a consicous decision?

libs/chain-signatures/contract/src/crypto_shared/types/serializable.rs
```diff
@@ -0,0 +1,82 @@
...
+            .map(SerializableEdwardsPoint)
+    }
+}
+
+#[cfg(any(test, feature = "test-utils"))]
+impl SerializableEdwardsPoint {
+    pub fn random(rng: impl rand::RngCore) -> Self {
+        Self(EdwardsPoint::random(rng))
+    }
+}
```
DSharifi, line 82: We can move this to the `test-utils` module?

libs/chain-signatures/contract/src/crypto_shared/types.rs
```diff
@@ -55,7 +56,7 @@ pub enum PublicKeyExtended {
                 definitions = "<[u8; 32] as ::borsh::BorshSchema>::add_definitions_recursively"
             ),))
         )]
```
gilcu3, line 58: I think these lines should not be needed anymore

libs/chain-signatures/contract/src/crypto_shared/types/serializable.rs
```diff
@@ -0,0 +1,82 @@
...
+        let bytes: [u8; 32] = BorshDeserialize::deserialize_reader(reader)?;
+
+        EdwardsPoint::from_bytes(&bytes)
+            .into_option()
+            .ok_or(std::io::Error::new(
+                std::io::ErrorKind::InvalidData,
+                "The provided bytes is not a valid edwards point.",
+            ))
+            .map(SerializableEdwardsPoint)
+    }
```
DSharifi, line 74: If we implement the GroupEncoding trait, why not:
```suggestion
        Self::from_bytes(&bytes)
            .into_option()
            .ok_or(std::io::Error::new(
                std::io::ErrorKind::InvalidData,
                "The provided bytes is not a valid edwards point.",
            ))
    }
```

gilcu3 (approved):
Thank you! Only one change that might be needed

libs/chain-signatures/contract/src/crypto_shared/types/serializable.rs
```diff
@@ -0,0 +1,82 @@
...
+        schemars(with = "[u8; 32]"),
+        borsh(schema(with_funcs(
+            declaration = "<[u8; 32] as ::borsh::BorshSchema>::declaration",
+            definitions = "<[u8; 32] as ::borsh::BorshSchema>::add_definitions_recursively"
+        ),))
+    )]
+    EdwardsPoint,
+);
+
+impl GroupEncoding for SerializableEdwardsPoint {
```
DSharifi, line 40: Why do we need this trait, instead of using the derive_mode traits for operations?

DSharifi (commented):
Question around the whole existence of SerializableEdwardsPoint. If we use the cfg_attr defined above for borsh, and [#[serde(with = "module")]](https://serde.rs/field-attrs.html#with) for serde, do we need the this wrapper struct at all?

libs/chain-signatures/contract/src/crypto_shared/types.rs
DSharifi: Yes exactly. Which tirggered this question from me, if we can remove our wrapper `SerializableEdwardsPoint` if we we can use these field attributes to make de/serialization work.

libs/chain-signatures/contract/src/crypto_shared/types/serializable.rs
netrome: Good point.

libs/chain-signatures/contract/src/crypto_shared/types/serializable.rs
netrome: I'd prefer to be consistent with the wrapped struct. I.e. `EdwardsPoint` implements `GroupEncoding` so the wrapper should as well, since we're using one of the methods from it.

Although I agree, and find it weird to have `from_bytes` as an explicit method in the `GroupEncoding` trait rather than just have a trait bound on `From<_>`.

netrome:
> Question around the whole existence of SerializableEdwardsPoint. If we use the cfg_attr defined above for borsh, and [#[serde(with = "module")]](https://serde.rs/field-attrs.html#with) for serde, do we need the this wrapper struct at all?

The cfg_attr is only for BorshSchema, not the actual serializers. I made a quick experiment naively just using `EdwardsPoint` but got the expected compile error about missing `BorshSerialize`:

```
   Compiling mpc-contract v2.2.0-rc.1 (/home/marten/near/mpc/libs/chain-signatures/contract)
error[E0277]: the trait bound `EdwardsPoint: borsh::BorshSerialize` is not satisfied
  --> contract/src/crypto_shared/types.rs:57:9
   |
39 | #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
   |                                                               -------------- required by a bound introduced by this call
...
57 |         edwards_point: EdwardsPoint,
   |         ^^^^^^^^^^^^^ the trait `borsh::BorshSerialize` is not implemented for `EdwardsPoint`
```

While there are probably more clean solutions to this code, but this PR is at least an improvement over the status quo so let's do this first and then anyone can revisit this for further refactors later.

libs/chain-signatures/contract/src/crypto_shared/types.rs
netrome: Done in https://github.com/near/mpc/pull/890/commits/d44ee04cd5830319520e675e06a42ed55db7ff48

netrome:
> In case anyone takes a look for further refactor. We need to use the field level attribute [#[borsh(serialize_with = ...)]](https://docs.rs/borsh/latest/borsh/derive.BorshSerialize.html#5-borshserialize_with---field-level-attribute). It works very similarly to serde's with attribute.

Makes sense. I think this would likely be a very nice simplification of the code. Let's not open an issue for it though unless anyone is ready to do it.

PR #914, opened 2025-08-21
chore!: remove near_crypto from node code

### LINKED ISSUE #880: remove near_crypto::PublicKey in favor of near_sdk::PublicKey
### Background

We'd like to remove the `near_crypto::PublicKey` type from our codebase as we already depend on `near_sdk::PublicKey` and want to avoid awkward conversions between the two types.

### Acceptance Criteria

- Remove `near_crypto` dependency and replace all `near_crypto::PublicKey` usage with `near_sdk::PublicKey`.

_No response_

DSharifi:
closes #880

Unfortunately we can't keep get completely rid of `near_crypto` for now as a dependency as it's needed for creating Transactions for the indexer in:
https://github.com/near/mpc/blob/9c801c98a02e3610049e19a397988260b66b2cd4/node/src/indexer/tx_signer.rs#L54-L66

This PR is a breaking change as the serialization/deserialization of keys with the `ed25519_dalek` crate uses binary representation of the keys, whereas `ner_crypto` uses bs58 encoding. This is fine as we also have another breaking change that is unreleased which to generate the private keys inside the node.

Devnet will be "broken" as seriialization/deserialization of the keys still use `near_crypto` implementation. This will easily be fixed in a follow up in https://github.com/near/mpc/issues/915


netrome:
This is also needed for https://github.com/near/mpc/issues/943

devnet/src/types.rs
```diff
@@ -1,21 +1,94 @@
...
+    }
+}
+
 /// Locally stored Near account information.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct NearAccount {
     pub account_id: AccountId,
-    pub access_keys: Vec<SecretKey>,
+    #[serde(with = "serialize_access_keys")]
+    pub access_keys: Vec<SigningKey>,
```
netrome, line 86: Why do we need custom serialization for these keys?

devnet/src/types.rs
```diff
@@ -30,10 +103,11 @@ pub enum NearAccountKind {
 /// Locally stored MPC participant keys and other info.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct MpcParticipantSetup {
-    pub p2p_private_key: SecretKey, // todo: this can eventually be removed [(#710)](https://github.com/near/mpc/issues/710)
+    pub p2p_private_key: SigningKey, // todo: this can eventually be removed [(#710)](https://github.com/near/mpc/issues/710)
     /// The account this participant uses to respond to signature requests.
     pub responding_account_id: AccountId,
-    pub p2p_public_key: Option<PublicKey>, // todo: this can eventually be made non-optional [(#710)](https://github.com/near/mpc/issues/710)
+    #[serde(with = "serialize_p2p_public_key")]
```
netrome, line 109: Same here.

node/src/indexer/participants.rs
```diff
@@ -257,14 +257,11 @@ pub fn convert_participant_infos(
...
-        // impl TryFrom<near_sdk::PublicKey> for near_crypto::PublicKey but it's test-only.
-        // For lack of better option we use this to-string from-string conversion instead.
-        let Ok(p2p_public_key) = near_crypto::PublicKey::from_str(&String::from(&info.sign_pk))
-        else {
-            anyhow::bail!("invalid participant public key {:?}", info.sign_pk);
-        };
+
+        let p2p_public_key =
+            ed25519_dalek::VerifyingKey::from_near_sdk_public_key(&info.sign_pk)
+                .with_context(|| format!("Invalid public key length for peer: {:?}", info.url))?;
```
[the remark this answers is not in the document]
netrome: Indeed.

netrome (changes requested):
This looks good to me, but I don't see why we need the custom serialization methods in devnet. If this is to enforce them to be serialized as hex strings, I'd prefer if this was explicit from naming or docs.

So we should either:
1. Remove the custom serialization modules.
or
2. Rename them and/or document why they are needed.

netrome:
> Note: Since this PR is needed for both #943 and #963, and because @DSharifi is off today and likely also tomorrow, I'll take some time tonight (unless anyone beats me to it) to remove the serialization logic and rebase this PR so we can get it merged.

Looking more at the code, I found it pretty clear that we want to serialize these as hex strings. So I kept it but renamed the serialization modules and moved them to the bottom of the file[^1].

[^1]: The first thing I want to see when reading a module named "types" is not serialization logic.

netrome (approved):
LGTM now

Thanks for the refactor @DSharifi :pray: 

And thanks for rebasing and updating @netrome *self five
self_five

DSharifi:
MPC binary fails to start when deploying on nomad. Check error log: http://34.53.93.19/ui/jobs/mpc-node-0@default?activeTask=abb174b6-b51a-e591-fd35-78503349797a-mpc-node
```log
error: unexpected argument '--boot-nodes' found

Usage: mpc-node init --dir <DIR> --chain-id <CHAIN_ID> --download-genesis --download-config

...
```

kevindeforth:
Investigation outcome with @DSharifi:

- devnet defaults to `testnet-release` which should be [testnet-release-7a304c1](https://hub.docker.com/layers/nearone/mpc-node-gcp/testnet-release-7a304c1/images/sha256-66bc3eb732d2d74321b869a629be070ffe36b4c4b0890c18ffad4b59a0f875aa), which is at commit 7a304c14b962f60a46a25366c0f2e5dcac796b29 from the [testnet-release branch](https://github.com/near/mpc/commits/testnet-release/)
- that commit does not include changes from https://github.com/near/mpc/pull/667 (c.f. cli.rs from that commit: https://github.com/near/mpc/blob/7a304c14b962f60a46a25366c0f2e5dcac796b29/node/src/cli.rs#L62 )

netrome:
> MPC binary fails to start when deploying on nomad. Check error log: http://34.53.93.19/ui/jobs/mpc-node-0@default?activeTask=abb174b6-b51a-e591-fd35-78503349797a-mpc-node
> 
> ```
> error: unexpected argument '--boot-nodes' found
> 
> Usage: mpc-node init --dir <DIR> --chain-id <CHAIN_ID> --download-genesis --download-config
> 
> ...
> ```
> 
> Edit: This issue is not due to this PR. Devnet is also broken on `main` branch, see http://34.145.55.28/ui/jobs/mpc-node-0@default?activeTask=493f7f91-9c24-824a-5152-1d1dba7e09ad-mpc-node.
> 
> The issue is that we added a new CLI flag, `boot-nodes`, in #667 which we use to start the node. Since the mpc binary version deployed by default when using Devnet is older than the commit at #667, `clap` will reject the CLI call as `boot-nodes` is not a recognized argument. Devnet has probably been broken since #667 was merged.

Nice observation. I just created a follow-up to fix Devnet https://github.com/near/mpc/issues/970

netrome:
> devnet defaults to testnet-release which should be [testnet-release-7a304c1](https://hub.docker.com/layers/nearone/mpc-node-gcp/testnet-release-7a304c1/images/sha256-66bc3eb732d2d74321b869a629be070ffe36b4c4b0890c18ffad4b59a0f875aa), which is at commit https://github.com/near/mpc/commit/7a304c14b962f60a46a25366c0f2e5dcac796b29 from the [testnet-release branch](https://github.com/near/mpc/commits/testnet-release/)

Why does devnet default to `testnet-release`? Shouldn't it default to whatever version we run the code on? I.e. if I check out testnet-release and run the devnet-cli, it should be using the `testnet-release` branch. If I'm on latest master, it should use the latest master.

Issue #923, opened 2025-08-22
Bump `GAS_FOR_SIGN_CALL` and `RETURN_SIGNATURE_AND_CLEAN_STATE_ON_SUCCESS_CALL_GAS`


netrome:
### Background

We have started experiencing some testnet issues. See thread: https://nearone.slack.com/archives/C07UW93JVQ8/p1755843932247499

### Acceptance Criteria

We've bumped the parameters to:
* `GAS_FOR_SIGN_CALL`: 15
* `RETURN_SIGNATURE_AND_CLEAN_STATE_ON_SUCCESS_CALL_GAS`: 7

- [ ] in testnet : https://github.com/near/mpc/pull/924
- [ ] in mainnet
- [ ] on main branch: https://github.com/near/mpc/pull/925

PR #925, opened 2025-08-22
fix: increase gas limit for sign calls

kevindeforth:
resolves (partially) #923


netrome (approved):

gilcu3:
Should it also be increased in the pytests?

gilcu3 (approved):

PR #928, opened 2025-08-22
fix: CI get docker manifest timeouts

### LINKED ISSUE #927: Dockerhub registry timeout
### Background

In CI a few times we have encountered timeout issues when interacting with the dockerhub registry in TEE Launcher pytests.

### User Story

CI should not fail due to small network timeouts

### Acceptance Criteria

The current timeout is increased, and the connection is retried a few times before giving up.

https://nearone.slack.com/archives/C0912BTG51T/p1755847439396639?thread_ts=1755843168.976579&cid=C0912BTG51T

https://github.com/near/mpc/actions/runs/17148851147/job/48650368214

gilcu3:
Closes #927 

This was actually a bug. The function doing the retries was doing a single retry, throwing an exception on the first failure.

### DIFF
--- a/tee_launcher/launcher.py
+++ b/tee_launcher/launcher.py
@@ -420,7 +420,6 @@ def request_until_success(
             continue
         else:
             return manifest_resp
-        
 
     raise RuntimeError(
         f"Failed to get succesful response from {url} after {rpc_max_attempts} attempts."


netrome (approved):
Nice, thanks for fixing :pray:

Issue #861, opened 2025-08-18
Indexer picks up CKD smart contract calls


gilcu3:
### Background

As part of the [Confidential Key Derivation](https://github.com/near/threshold-signatures/blob/main/docs/confidential_key_derivation.md) functionality (CKD), the MPC Node Indexer needs to pick up CKD requests accepted by the MPC contract #860 

To achieve that, the following changes are needed:
### node/src/indexer/handler.rs

* modify fn **handle_message**
* add fn **maybe_get_ckd_args**( receipt: &ReceiptView, execution_outcome: &ExecutionOutcomeWithIdView, mpc_contract_id: &AccountId, )
* add fn **maybe_get_ckd_completion**( receipt: &ReceiptView, mpc_contract_id: &AccountId,)
* add struct **CKDRequestFromChain**
* modify struct **ChainBlockUpdate**

### node/src/mpc_client.rs

* modify fn **MpcClient::monitor_block_updates** 
* modify fn **MpcClient::process_channel_task**

### node/src/primitives.rs

* modify struct **MpcTaskId**

### node/src/metrics.rs

* add metrics such that MPC_NUM_CKD_RESPONSES_INDEXED, MPC_NUM_CKD_REQUESTS_INDEXED

### Acceptance Criteria

The Indexer in `node/src/indexer/` is able to pick up CKD contract requests. All functionalities that currently exists for edDSA signatures are also implemented for CKD requests.

### Resources & Additional Notes

https://docs.nearone.org/doc/low-level-system-design-GcD8YKiczY

gilcu3:
Following advice from https://github.com/near/mpc/pull/920#pullrequestreview-3154037436, this issue will be divided in mainly 4 parts:

- [ ] indexer (here, in mpc_client.rs, we just skip if we receive CKDRequest)
- [ ] store + web (db logic + handle requests in mpc_client.rs + web.rs)
- [ ] queue refactor
- [ ] provider + integration with threshold-signatures

netrome:
> I was keeping it open until the sub-issue is done :| As that's the only way I can actually test that the current implementation is not missing something. Hope to finish it ASAP

Okay sure. Personally I'd just close this issue and track the integration test implementation separately - as you already have a separate (sub) issue for it. Any discovered bug or problem could be opened as either a separate issue or done directly alongside the integration test if it's just a small fix.

But I'm okay with keeping this open for the time being then as well.

PR #968, opened 2025-08-27
feat: CKD indexer - queue refactor

gilcu3:
Solves part 3 of https://github.com/near/mpc/issues/861#issuecomment-3223531579

- it looks bigger than it really is, the problem is github does not recognize that the `queue.rs` module was just moved and modified


node/src/requests/queue.rs
kevindeforth, line 1: this is just moved + abstracted, right?

kevindeforth (approved):
Thank you!

> looks bigger than it really is, the problem is github does not recognize that the queue.rs module was just moved and modified

Yep, that's classic. I did not compare the `queue.rs` file line by line with the old one - that would be a bit tedious.

If you think an in-depth review is required, then I suggest to split this into 3 separate PRs: 1) add abstraction 2) move file, 3) add ckd support.

PR #973, opened 2025-08-28
chore: upgrade rust version to 1.89

### LINKED ISSUE #972: upgrade rust compiler version to 1.89
### Background

upgrade rust compiler version to 1.89

### User Story

As a developer I want our compiler to be up to date.

### Acceptance Criteria

upgrade rust compiler version to 1.89

_No response_

DSharifi:
closes #972 

All changes added are to satisfy the new clippy lints.


netrome (approved):
Nice, thank you!

pbeza (approved):

netrome (changes requested):
Right forgot. You should update the dockerfiles as well, and we risk hitting [this problem](https://github.com/near/mpc/pull/964#discussion_r2306802070) - so please try that first. Otherwise we might need to settle for an older version.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -77,17 +77,30 @@ const CLEAN_TEE_STATUS_GAS: Gas = Gas::from_tgas(3);
...
     V1(MpcContractV1),
     /// Current actual version
     V2(MpcContract),
+    // IMPORTANT: When adding a new variant, update the compile-time assertion below.
 }
+// Compile-time assertion to ensure the production contract variant is the largest.
+// This prevents wasted storage space from enum padding in production.
+const _: () = {
+    assert!(
+        size_of::<MpcContract>() >= size_of::<MpcContractV1>(),
```
netrome, line 100: Shouldn't this be:

```suggestion
        size_of::<VersionedMpcContract>() >= size_of::<MpcContract>(),
```

DSharifi: I went with this initially. But there's a flaw in checking for equality. There will also be some storage used to tag the enum variant.

Equality check will pass right due to alignement padding and word sizes from the compiler. But we don't have a guarantee for that.

See this example in the rust playground https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=172f744e1f4b6060fce44e310d5e570a

DSharifi: >Nvm, an enum is typically slightly larger than it's largest variant. A better variant would be to do something like

It's actually the same size right now.

PR #986, opened 2025-09-01
test: test that new participant without valid TEE attestation can not join participants

### LINKED ISSUE #848: modify unit test for `submit_new_parameters` after attestation module works
### Background

In #843 we change the behavior of the endpoint. We should test the new behavior once we have working attestation crate.

### User Story

As a dev, I want test coverage

### Acceptance Criteria

Test the newly implemented behavior.

_No response_

DSharifi:
closes #848

### DIFF
--- a/libs/chain-signatures/contract/src/lib.rs
+++ b/libs/chain-signatures/contract/src/lib.rs
@@ -12,11 +12,12 @@ pub mod utils;
 pub mod v0_state;
 
 use crate::crypto_shared::types::CKDResponse;
+use crate::tee::quote::TeeQuoteStatus;
 use crate::{
     errors::{Error, RequestError},
     primitives::ckd::{CKDRequest, CKDRequestArgs},
     storage_keys::StorageKey,
-    tee::{proposal::AllowedDockerImageHashes, quote::TeeQuoteStatus, tee_state::TeeState},
+    tee::{proposal::AllowedDockerImageHashes, tee_state::TeeState},
     update::{ProposeUpdateArgs, ProposedUpdates, Update, UpdateId},
     v0_state::MpcContractV1,
 };
--- a/libs/chain-signatures/contract/tests/common.rs
+++ b/libs/chain-signatures/contract/tests/common.rs
@@ -32,6 +32,7 @@ use mpc_contract::{
     crypto_shared::k256_types::SerializableAffinePoint,
     primitives::signature::{Payload, SignRequestArgs},
 };
+use mpc_primitives::hash::MpcDockerImageHash;
 use near_sdk::{log, CurveType, PublicKey};
 use near_workspaces::{
     network::Sandbox,
@@ -666,4 +667,19 @@ pub async fn submit_participant_info(
         .transact()
         .await?;
     Ok(result.is_success())
+}
+
+pub async fn vote_for_hash(
+    account: &Account,
+    contract: &Contract,
+    hash: &MpcDockerImageHash,
+) -> anyhow::Result<()> {
+    check_call_success(
+        account
+            .call(contract.id(), "vote_code_hash")
+            .args_json(serde_json::json!({"code_hash": hash}))
+            .transact()
+            .await?,
+    );
+    Ok(())
 }
--- a/libs/chain-signatures/contract/tests/tee.rs
+++ b/libs/chain-signatures/contract/tests/tee.rs
@@ -1,13 +1,10 @@
 pub mod common;
 
-use crate::common::gen_accounts;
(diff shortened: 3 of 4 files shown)


crates/contract/tests/sandbox/tee.rs
```diff
@@ -143,7 +143,9 @@ async fn test_vote_code_hash_doesnt_accept_account_id_not_in_participant_list()
         .transact()
         .await?;
     let Err(err) = res.into_result() else {
-        panic!("vote_code_hash should not accept votes from a randomly generated account id that is not in the participant list");
+        panic!(
+            "vote_code_hash should not accept votes from a randomly generated account id that is not in the participant list"
+        );
     };
```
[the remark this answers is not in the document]
pbeza: With formatting? I’m curious what the root cause is. Let me know once you figure it out, since I still have an issue with formatting when saving files in VSCode.

crates/contract/tests/sandbox/vote.rs
```diff
@@ -853,3 +855,63 @@ async fn test_successful_resharing_after_cancellation_clears_cancelled_epoch_id(
...
+            }))
+            .transact()
+            .await
+            .unwrap()
+            .into_result()
+            .expect_err("calling `vote_new_parameters` must fail when one participant has invalid TEE status.");
+
+        let error_message = call_result.to_string();
+        let expected_error_message = InvalidParameters::InvalidTeeRemoteAttestation.to_string();
+        assert!(error_message.contains(&expected_error_message));
```
pbeza, line 907: Do we actually need to stringify before comparing? Comparing error strings feels fragile (though at least we’re not hardcoding the error strings here, which is good).

crates/contract/tests/sandbox/vote.rs
```diff
@@ -853,3 +855,63 @@ async fn test_successful_resharing_after_cancellation_clears_cancelled_epoch_id(
...
+        let new_account = new_accounts.pop().unwrap();
+        (new_account, new_account_id, new_participant_info)
+    };
+
+    new_participants
+        .insert(new_account_id.clone(), new_participant_info)
+        .unwrap();
+
+    let threshold_parameters =
+        ThresholdParameters::new(new_participants, Threshold::new(3)).unwrap();
```
pbeza, line 886: I always feel a bit lost when I see a new threshold set in a test and can’t remember what the default was in `init_env_secp256k1`. Anyway, just bikeshedding about the testing infra—`init_env_secp256k1` implicitly sets the threshold to `2`. It’s not very intuitive unless you’ve spent a lot of time with the codebase (which we did).

crates/contract/tests/sandbox/vote.rs
```diff
@@ -853,3 +855,63 @@ async fn test_successful_resharing_after_cancellation_clears_cancelled_epoch_id(
...
+    let existing_params = initial_running_state.parameters.clone();
+    let mut new_participants = existing_params.participants().clone();
+
+    // Add a new participant
+    let (new_account, new_account_id, new_participant_info) = {
+        let (mut new_accounts, participants) = gen_accounts(&worker, 1).await;
+        let (new_account_id, _, new_participant_info) =
+            participants.participants().first().unwrap().clone();
+        let new_account = new_accounts.pop().unwrap();
+        (new_account, new_account_id, new_participant_info)
```
pbeza, line 878: [bikeshedding] I wish `gen_accounts` were more user-friendly so we didn’t have to do all that conversion dance.

crates/contract/tests/sandbox/vote.rs
```diff
@@ -853,3 +855,63 @@ async fn test_successful_resharing_after_cancellation_clears_cancelled_epoch_id(
...
+    let (worker, contract, mut current_participant_accounts, _) = init_env_secp256k1(1).await;
+
+    let state: ProtocolContractState = contract.view("state").await.unwrap().json().unwrap();
+    let ProtocolContractState::Running(initial_running_state) = state else {
+        panic!("State is not running: {:#?}", state)
+    };
+
+    let initial_epoch_id = initial_running_state.keyset.epoch_id;
+    let existing_params = initial_running_state.parameters.clone();
+    let mut new_participants = existing_params.participants().clone();
```
pbeza, line 870: Isn’t `new_participants` a bit confusing in this context? It’s actually the old participants plus one new participant, right?

crates/contract/tests/sandbox/vote.rs
DSharifi: Yeah, fully agree.

PR #995, opened 2025-09-02
chore: use configuration value instead of constant for `TEE_UPGRADE_PERIOD`

### LINKED ISSUE #994: Use configuration value in place of hard coded constants for `TEE_UPGRADE_PERIOD`
### Background

In [proposal.rs](https://github.com/near/mpc/blob/main/libs/chain-signatures/contract/src/tee/proposal.rs#L7) we have a hard-coded constant for the duration an old MPC image hash is considered valid during the grace period where it superseded by a new image hash.

https://github.com/near/mpc/blob/e5ecc161bf1cffced6875a5e6aa0dd7af6ad225b/libs/chain-signatures/contract/src/tee/proposal.rs#L7

We'd like to use a configuration value instead of a hard-coded constant for two reasons:
- We want to be able to tune the deadline for a graceful upgrade period in production without having to upgrade the whole contract.
- For test purposes we'd like to be able to tune this graceful upgrade period, such that we can perform system and integration tests of this behavior.

### User Story

_No response_

### Acceptance Criteria

The const `TEE_UPGRADE_PERIOD` is removed and replaced with a runtime configurable value instead.

_No response_

DSharifi:
closes #994 

I'm sad that once again we have limitations on what we can put in the contract state structs because of the serde/borsh bounds. We have these bounds because we don't have any interface types and use the contract types directly as return types on the contract method, and the node depends on them.

This means we can't for example use structures such as `Rc` + `RefCell` to store the configuration value, instead of passing down values for every call.

See revert commit:
https://github.com/near/mpc/commit/a98c6d940af9c83db93203c67156afc347257e8f


libs/chain-signatures/contract/src/config.rs
netrome, line 1: Did you refactor the `mod.rs` -> `config.rs` now even before https://github.com/near/mpc/pull/998 landed? :sweat_smile: I was so confused by the diff, because `InitConfig` seemed unused and I thought it was a new struct.

netrome (approved):
Looks good to me. I'm sad that we have to have dynamic configuration for these deadlines/timeouts to be able to test them. Although we'll want to support configuring support for local attestations anyway.

libs/chain-signatures/contract/src/tee/proposal.rs
```diff
@@ -63,7 +60,7 @@ pub struct AllowedDockerImageHash {
 /// Collection of whitelisted Docker code hashes that are the only ones MPC nodes are allowed to
 /// run.
 #[near(serializers=[borsh, json])]
-#[derive(Clone, Debug, Default, PartialEq, Eq)]
+#[derive(Clone, Default, Debug, PartialEq, Eq)]
```
gilcu3, line 63: Curiosity: is this an extra clippy rule you have enabled?

PR #1031, opened 2025-09-04
fix: Use version controlled contracts in compatibility tests

### LINKED ISSUE #1026: Backwards compatibility test is broken
### Background

See slack thread https://nearone.slack.com/archives/C07UW93JVQ8/p1756984840167529

We're currently getting these errors:

```
cargo nextest run mpc-contract::back_compatibility test_back_compatiblity_mainnet
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache
   Compiling dstack-sdk-types v0.1.0 (https://github.com/Dstack-TEE/dstack.git?rev=04de4e422bb06f075b4215b2cfc410f5d7ac7aed#04de4e42)
   Compiling mpc-primitives v0.0.0 (/home/marten/near/mpc/crates/primitives)
   Compiling attestation v0.1.0 (/home/marten/near/mpc/crates/attestation)
   Compiling test_utils v0.1.0 (/home/marten/near/mpc/crates/test_utils)
   Compiling mpc-contract v2.2.0-rc.1 (/home/marten/near/mpc/libs/chain-signatures/contract)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 17.79s
────────────
 Nextest run ID c0e342c1-24fb-4c56-b2f3-23d44258e4a4 with nextest profile: default
    Starting 1 test across 10 binaries (105 tests skipped)
        SLOW [> 60.000s] mpc-contract::back_compatibility test_back_compatiblity_mainnet
        FAIL [  62.237s] mpc-contract::back_compatibility test_back_compatiblity_mainnet
  stdout ───

    running 1 test
    test test_back_compatiblity_mainnet has been running for over 60 seconds
    test test_back_compatiblity_mainnet ... FAILED

    failures:

    failures:
        test_back_compatiblity_mainnet

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1 filtered out; finished in 62.22s

  stderr ───
    2025-09-04T11:19:06.428325Z ERROR neard::cli: ERROR: net.core.rmem_max is set to 212992, expected 8388608. Please run `scripts/set_kernel_params.sh`.
    2025-09-04T11:19:06.428348Z ERROR neard::cli: ERROR: net.core.wmem_max is set to 212992, expected 8388608. Please run `scripts/set_kernel_params.sh`.
    2025-09-04T11:19:06.428356Z ERROR neard::cli: ERROR: net.ipv4.tcp_rmem is set to 4096 131072 6291456, expected 4096 87380 8388608. Please run `scripts/set_kernel_params.sh`.
    2025-09-04T11:19:06.428361Z ERROR neard::cli: ERROR: net.ipv4.tcp_wmem is set to 4096 16384 4194304, expected 4096 16384 8388608. Please run `scripts/set_kernel_params.sh`.
    2025-09-04T11:19:06.428369Z ERROR neard::cli: ERROR: net.ipv4.tcp_slow_start_after_idle is set to 1, expected 0. Please run `scripts/set_kernel_params.sh`.

    thread 'test_back_compatiblity_mainnet' panicked at contract/tests/back_compatibility.rs:145:48:
    called `Result::unwrap()` on an `Err` value: error decoding response body: expected value at line 1 column 1

    Caused by:
        expected value at line 1 column 1

  Cancelling due to test failure
────────────
     Summary [  62.239s] 1 test run: 0 passed, 1 failed, 105 skipped
        FAIL [  62.237s] mpc-contract::back_compatibility test_back_compatiblity_mainnet
error: test run failed
```

They appeared out of nowhere, and are reproducible on old commits on main which suggests a time bomb.

### User Story

_No response_

### Acceptance Criteria

The test passes.

Ideally it's also fixed so that there are no new time bombs, and the test should work without a network connection.

_No response_

netrome:
Closes #1026 

In this PR I've introduced a new small `contract_history` crate aiming to store historic bytecodes of deployed contracts for version compatibility tests. This crate primarily defines the **current**[^1] mainnet and testnet contracts.

I've replaced the contract compatibility tests to use this module instead. For the pytests, I added a similar module and an export binary that writes the current tests to the python module.

To ensure this module is up to date, I've added tests that can verify the contracts against testnet and mainnet. These tests are not run on PRs since they may break for reasons out of our control. However, I added a workflow that we run on `main` and release tags to ensure we'll quickly be able to notice if anything is off.

[^1]: Current means the contracts deployed to the networks at the time of the checked out commit.


crates/contract_history/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "contract_history"
+version = "0.1.0"
+edition = "2024"
```
gilcu3, line 4: question: should we keep editions in sync?

crates/contract_history/README.md
```diff
@@ -0,0 +1,5 @@
+# Historic MPC contracts
+This crate stores binaries of historic contracts, which we can use for compatibility tests.
+
+Additionally, to support our pytests we have a binary in `bin/export_contracts.rs` that
+takes the latest mainnet and testnet contracts and writes them to the appropriate pytest folder.
```
gilcu3, line 5:
```suggestion
# Historic MPC contracts

This crate stores binaries of historic contracts, which we can use for compatibility tests.

Additionally, to support our pytests we have a binary in `bin/export_contracts.rs` that
takes the latest mainnet and testnet contracts and writes them to the appropriate pytest folder.

```
just to make the linter happy

crates/contract_history/src/bin/export_contracts.rs
```diff
@@ -0,0 +1,18 @@
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let target_dir = std::path::Path::new("pytest/common_lib/current_contracts");
```
gilcu3, line 2: does this mean this must be run from the root dir?

crates/contract_history/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+    version_2_2_0()
+}
+
+pub const fn current_testnet() -> &'static [u8] {
+    version_2_2_0()
+}
+
+pub const fn version_2_2_0() -> &'static [u8; 566653] {
+    include_bytes!("../archive/signer-2_2_0.wasm")
+}
```
gilcu3, line 11: why do we need to fix the output size here?

crates/contract_history/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+
+        let response = reqwest::Client::new()
+            .post(rpc_endpoint)
+            .json(&body)
+            .send()
+            .await
+            .expect("contract request should succeed")
+            .json::<RpcResponse>()
+            .await
+            .expect("contract response should be valid JSON");
```
gilcu3, line 56: I think it is sane to have a small timeout here anyway, 10 seconds solved the issue from yesterday for example

libs/chain-signatures/contract/Cargo.toml
```diff
@@ -78,9 +78,10 @@ assert_matches = "1.5.0"
 rand = "0.8"
 tokio = { version = "1", features = ["full"] }
 sha2 = "0.10.8"
-reqwest = { version = "0.11", features = ["json"] }
-base64 = "0.21"
+reqwest = { version = "0.11", features = ["json"] } # TODO: Remove
+base64 = "0.21" # TODO: Remove
```
gilcu3, line 82: by remove you mean to use the workspace version?

.github/workflows/contract-verification.yml
gilcu3, line 1: afaik this would only run in the merge queue, or manually, right?

gilcu3 (approved):
Thanks! Looks great! Left mostly questions :)

crates/contract_history/Cargo.toml
netrome: Created follow-up issue https://github.com/near/mpc/issues/1035

crates/contract_history/src/lib.rs
netrome: We don't, but given that we know the size here I found it nice to be explicit. This will never change, and this way we have the size of the contracts visible in the code. So I think this is strictly better than just converting this to an arbitrary slice pointer.

crates/contract_history/src/lib.rs
netrome: Good point. I'll add one.

libs/chain-signatures/contract/Cargo.toml
netrome: Oh! Forgot about this. This dependency should be unused. Thanks for spotting.

netrome: Done in https://github.com/near/mpc/pull/1031/commits/35fafcabce91932be9a433e0e18d686056a925a2

crates/contract_history/src/bin/export_contracts.rs
gilcu3: I am not sure what is the expected workflow, maybe adding that to the readme would help

crates/contract_history/src/lib.rs
gilcu3: You mean it will never change until we update the contract in testnet/mainnet?

.github/workflows/contract-verification.yml
netrome: Nope, to have this run in the merge queue we'd have to use the `merge_group` event. [Reference](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#configuring-continuous-integration-ci-workflows-for-merge-queues).

So this should work as intended.

libs/chain-signatures/contract/Cargo.toml
netrome: Ahhh I'm blind :sweat_smile: Let's also forbid unused crate dependencies.

netrome: Fixed in https://github.com/near/mpc/pull/1031/commits/06362c969e1995ea31e737e45e85348ed2b51868. Set to auto-merge again. I'll forbid unused dependencies in a follow-up since it's bound to reveal additional crates that needs to be pruned.

pytest/common_lib/contracts.py
```diff
@@ -33,33 +37,24 @@ def build_view_code_request(account_id: str) -> dict:
     }
 
 
-class NetworkRpc(Enum):
-    MAINNET = "https://rpc.mainnet.near.org"
-    TESTNET = "https://rpc.testnet.near.org"
-
+def fetch_testnet_contract() -> bytearray:
```
kevindeforth, line 40: Why not just delete those functions and replace their calls with `current_*_contract()`?

pytest/common_lib/contracts.py
```diff
@@ -16,8 +16,12 @@
     / "migration"
     / "migration_contract.wasm"
 )
-TESTNET_ACCOUNT_ID = "v1.signer-prod.testnet"
-MAINNET_ACCOUNT_ID = "v1.signer"
+CURRENT_MAINNET_CONTRACT_PATH = (
+    MPC_REPO_DIR / "pytest" / "common_lib" / "current_contracts" / "signer_mainnet.wasm"
```
kevindeforth, line 20: Why don't we rely on the `contract_history` crate? 

Having the `.wasm` binaries in the pytest folder means yet another two binaries to maintain (which, I thought should be handled by the `contract_history` crate).

crates/contract_history/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+    use super::*;
+
+    use base64::{Engine as _, prelude::BASE64_STANDARD};
+
+    #[tokio::test]
+    async fn mainnet_contract_should_be_up_to_date() {
+        let on_chain_right_now =
+            fetch_contract_code("https://rpc.mainnet.near.org", "v1.signer").await;
+
+        assert_eq!(current_mainnet(), on_chain_right_now);
```
kevindeforth, line 25: I think comparing the hashes should be sufficient.

crates/contract_history/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    #[tokio::test]
+    async fn testnet_contract_should_be_up_to_date() {
+        let on_chain_right_now =
+            fetch_contract_code("https://rpc.testnet.near.org", "v1.signer-prod.testnet").await;
+
+        assert_eq!(current_testnet(), on_chain_right_now);
```
kevindeforth, line 33: idem

pytest/common_lib/current_contracts/signer_mainnet.wasm
kevindeforth, line 1: lets remove these and rely on the `contract_history` crate (we could add a cli that returns the current_mainnet and current_testnet versions).

kevindeforth (changes requested):
Pretty nice! 
Just, I would prefer not to have binaries in the pytest folder.

.github/workflows/contract-verification.yml
kevindeforth: Is it possible to have this run on every CI, but not require that test to pass?

crates/contract_history/src/lib.rs
gilcu3: is there a way to get the hash directly from the rpc? In that case it would make sense :) Else I don't see the advantage

netrome:
> Just, I would prefer not to have binaries in the pytest folder.

Sure, I agree it's better to copy them on the fly as part of pytest fixtures as you suggested.

crates/contract_history/src/bin/export_contracts.rs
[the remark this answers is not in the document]
netrome: Done in https://github.com/near/mpc/pull/1031/commits/6a5f763a68e5197e947891817039ffb670751f47 if CI is happy this should be ready for re-review.

pytest/common_lib/contracts.py
netrome: Yeah that's why I added the explicit export script. Wasn't too happy about this solution, so the pytest fixture variant you proposed is much better.

Updated in https://github.com/near/mpc/pull/1031/commits/6a5f763a68e5197e947891817039ffb670751f47

pytest/tests/test_contract_update.py
```diff
@@ -38,11 +35,8 @@ def test_update_from_current():
     cluster.assert_is_deployed(new_contract.code())
 
 
-@pytest.mark.parametrize(
-    "fetch_contract", [fetch_mainnet_contract, fetch_testnet_contract]
-)
-def test_update_to_current(fetch_contract):
-    current = fetch_contract()
+def test_update_to_current(current_contract):
+    current = current_contract
```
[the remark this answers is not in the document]
netrome: Yeah that's a bit more explicit. I find the whole fixture system in pytests a tad too implicit for my taste in general. It's not obvious that just running a single test causes the contract to be compiled etc.

pytest/tests/conftest.py
```diff
@@ -51,16 +52,36 @@ def pytest_addoption(parser):
     )
 
 
+@pytest.fixture(scope="session")
+def contract_binaries(request):
```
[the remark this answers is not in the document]
netrome: Oh... will remove. Thanks for spotting!

crates/contract_history/src/lib.rs
gilcu3: just tested, if that hash is the hash of the code, then we can also get the hash only using  https://docs.near.org/api/rpc/contracts#view-account

.github/workflows/contract-verification.yml
```diff
@@ -0,0 +1,37 @@
+name: Contract Verification
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  push:
+    branches: [main]
+    tags: ["v*"]
```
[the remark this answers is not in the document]
netrome: Main broke due to https://github.com/near/mpc/issues/1039 and I got an email about it now. I think this will be sufficient :)

Issue #948, opened 2025-08-25
Attestation Service


kevindeforth:
### Background

Contract changing state from Running --> Recovery --> Running might be missed by some nodes. Need a way to update participant set while in Running state. In that case, nodes will need to re-establish p2p connections and purge their assets to avoid timing out requests.

### User Story

As a node operator, I want my node to realize when the credentials of other nodes have changed and establish a new connection when required.

### Acceptance Criteria

A pytest with two nodes:
- submit (in the same block, if possible) the transaction initiating the recovery mode and the one resolving it.
- observe that both nodes remove their pre-computed assets.

PR #1032, opened 2025-09-04
feat: Improve asset cleanup behavior when entering running or resharing

### LINKED ISSUE #1020: Improve Asset Cleanup Behavior when entering running state
### Background

We call `delete_stale_triples_and_presignatures` when entering a running or a resharing state:

https://github.com/near/mpc/blob/b0cfd13f13fa5ed6411109f4a505925f5c0e3bf8/node/src/coordinator.rs#L386

That function deletes all assets (triples & pre-signatures) if the current epoch id differes from the epoch id stored in the database.

https://github.com/near/mpc/blob/b0cfd13f13fa5ed6411109f4a505925f5c0e3bf8/node/src/coordinator.rs#L695

With the changes for #948, it is possible that one of the participants executed a recovery and is not in possession of any pre-computed assets.

The node must handle this case. Even if the `epoch_id` is the same, the node must now check if one of the participants executed a recovery or not. This can be done by the node storing the latest known participant config to the database and comparing the current config with the previously known config. If any of the p2p keys changed, then the node must delete their owned assets involving those peers.

### User Story

As a user, I want to avoid downtime due to lost assets after a recovery.

### Acceptance Criteria

- Nodes store the latest seen participants config in the database.
- in case any of their peers changed their TLS key, the node must remove all owned assets involving any such node.

_No response_

kevindeforth:
Resolves #1020 

Would have liked to do a minor refactor of the database organization, but that would have significantly increased the scope of this PR.

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -13,7 +13,6 @@ attestation = { workspace = true }
 axum = { workspace = true }
 backon = { workspace = true }
 borsh = { workspace = true }
-bincode = { workspace = true }
 threshold-signatures = { workspace = true }
 clap = { workspace = true }
 curve25519-dalek = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1517,15 +1517,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "55248b47b0caf0546f7988906588779981c43bb1bc9d0c44087278f80cdb44ba"
 
 [[package]]
-name = "bincode"
-version = "1.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b1f45e9417d87227c7a56d22e471c6206462cba514c7590c09aff4cf6d1ddcad"
-dependencies = [
- "serde",
-]
-
-[[package]]
 name = "bindgen"
 version = "0.65.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4891,7 +4882,6 @@ dependencies = [
  "attestation",
  "axum 0.7.9",
  "backon",
- "bincode",
  "borsh",
  "built",
  "clap",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,6 @@ async-trait = "0.1.83"
 axum = "0.7.9"
 backon = { version = "1.5.1", features = ["tokio-sleep"] }
 borsh = { version = "1.5.1", features = ["derive", "unstable__schema"] }
-bincode = { version = "1.3" }
 bs58 = { version = "0.5.1" }
 clap = { version = "4.5.20", features = ["derive", "env"] }
 curve25519-dalek = "4.1.3"
--- a/node/src/assets/cleanup.rs
+++ b/node/src/assets/cleanup.rs
@@ -70,8 +70,8 @@ pub fn delete_stale_triples_and_presignatures(
 
     let mut update_writer = db.update();
     tracing::info!("Updating epoch data: {:?}.", current_epoch_data);
-    let bytes = bincode::serialize(&current_epoch_data)?;
-    update_writer.put(DBCol::EpochData, EPOCH_ID_KEY, &bytes);
+    let current_epoch_data_ser = serde_json::to_vec(&current_epoch_data)?;
+    update_writer.put(DBCol::EpochData, EPOCH_ID_KEY, &current_epoch_data_ser);
     tracing::info!("Updated epoch id entry");
     update_writer.commit()?;
     Ok(())
@@ -87,7 +87,7 @@ fn get_epoch_data(db: &Arc<SecretDB>) -> anyhow::Result<Option<EpochDataWrapper>
     let Some(db_res): Option<Vec<u8>> = db.get(DBCol::EpochData, EPOCH_ID_KEY)? else {
         return Ok(None);
     };
-    if let Ok(epoch_data) = bincode::deserialize::<EpochData>(&db_res) {
+    if let Ok(epoch_data) = serde_json::from_slice::<EpochData>(&db_res) {
         return Ok(Some(EpochDataWrapper::Current(epoch_data)));
     }
 


TEE.md
```diff
@@ -66,61 +66,58 @@ The MPC contract is used for authenticating the backup service and the MPC node.
...
 
-Unlike the `Resharing` state, entering this state does not require `threshold` votes, but rather, a single vote is sufficient. However, the `AccountId` of all participants must be preserved. Only secondary participant details may be changed.
+A recovery process may occur only while the protocol is in a `Running` state. This is to simplify the problem space. Having a recovery process in parallel to a `Resharing` or `Initializing` protcol state requires to account for many edge cases and race conditions, which we would prefer to avoid.
 
-In the first iteration of the implementation, it is okay if signature requests are not accepted while the protocol is in the `Recovery` state.
+Unlike changes in the `ProtocolState`, changes in `Recovery` process do not require a signing quorum. Instead, each participant can initiate and conclude a recovery process at their own discretion.
+To avoid making the `Recovery` process a DOS attack vector, protocol state changes must have priority over ongoing `Recovery` processes. If the protocol state changes into a `Resharing` or `Initializing` state, any ongoing `Recovery` processes will simply be cancelled.
```
netrome, line 86: Nice updates!

libs/chain-signatures/contract/src/primitives/key_state.rs
```diff
@@ -135,6 +135,11 @@ impl Keyset {
             .key
             .clone())
     }
+
+    #[cfg(feature = "dev-utils")]
```
netrome, line 139: I see we have both the `dev-utils` and `test-utils` feature flags. What is the distinction between the two? I'd advocate for just having a single `test-utils`[^1] feature flag for any logic that's only relevant for testing.

[^1]: I'm more used to the `test-utils` name than `dev-utils` for this purpose, but have no strong opinion on which one we go with.

netrome: Note: This should be done as a follow-up.

node/src/asset_cleanup.rs
```diff
@@ -0,0 +1,416 @@
...
+    Current(EpochData),
+}
+
+fn get_epoch_data(db: &Arc<SecretDB>) -> anyhow::Result<Option<EpochDataWrapper>> {
+    let Some(db_res): Option<Vec<u8>> = db.get(DBCol::EpochData, EPOCH_ID_KEY)? else {
+        return Ok(None);
+    };
+    if let Ok(epoch_data) = bincode::deserialize::<EpochData>(&db_res) {
+        return Ok(Some(EpochDataWrapper::Current(epoch_data)));
+    }
```
netrome, line 30: Unrelated: I'm surprised we don't have a wrapper on the database level which also does the serialization. Also I notice we sometimes use JSON and sometimes bincode for serializing our data. Why do we use different serialization formats for different columns?

node/src/asset_cleanup.rs
```diff
@@ -0,0 +1,416 @@
...
+    tracing::info!("Updating epoch data: {:?}.", current_epoch_data);
+    let bytes = bincode::serialize(&current_epoch_data)?;
+    update_writer.put(DBCol::EpochData, EPOCH_ID_KEY, &bytes);
+    tracing::info!("Updated epoch id entry");
+    update_writer.commit()?;
+    Ok(())
+}
+
+#[cfg(test)]
+mod test_utils {
```
netrome, line 164: This is quite a lot of logic. I'd prefer to move it out to a separate file.

netrome (commented):
Partial review. Will continue in ~30 minutes.

libs/chain-signatures/contract/src/primitives/key_state.rs
kevindeforth: `test-utils` is for tests, whereas `dev-utils` is for production dependencies in the MPC node.

node/src/asset_cleanup.rs
kevindeforth: Yeah, that's an excellent observation. Would love to fix this in a refactor.

node/src/asset_cleanup.rs
kevindeforth: Yeah, I was wondering about the right place for the logic in this file. I see the following optionss.
- make a `db` folder and move the logic from `db.rs` to that folder, together with the one in this file
- make an `assets` folder and move this logic there
- keep this in the `src` folder and just make yet another file behind the `test` compilation flag.

node/Cargo.toml
```diff
@@ -13,6 +13,7 @@ attestation = { workspace = true }
 axum = { workspace = true }
 backon = { workspace = true }
 borsh = { workspace = true }
+bincode = { workspace = true }
```
netrome, line 16: Oh this PR introduces bincode serialization? As much as I love bincode, it's good if we're consistent in which format we use to serialize the data in our database. Is there any reason not to go with JSON serialization?

Secondly, if we want a separate binary format here I'd also consider borsh over bincode because we already use it for our smart contracts and it has a strict specification.

node/src/asset_cleanup.rs
```diff
@@ -0,0 +1,416 @@
...
+    anyhow::bail!("Can't deserialize EPOCH_ID entry: {:?}", db_res);
+}
+
+enum AssetCleanup {
+    Keep,
+    DeleteAll,
+    KeepOnly(Vec<primitives::ParticipantId>),
+}
+
+fn get_cleanup_behavior(
```
netrome, line 50: Nit: I find the `get_` prefix a bit confusing here since we're not getting anything from a database or external service.

```suggestion
fn cleanup_behavior(
```

node/src/asset_cleanup.rs
```diff
@@ -0,0 +1,416 @@
...
+    let tls_key_changed: bool =
+        persitent_participants.len() != current_epoch_data.participants.participants.len();
+    if tls_key_changed {
+        AssetCleanup::KeepOnly(persitent_participants)
+    } else {
+        AssetCleanup::Keep
+    }
+}
+
+fn get_cleanup_behavior_during_update(
```
netrome, line 85: Same here

```suggestion
fn cleanup_behavior_during_update(
```

node/src/asset_cleanup.rs
```diff
@@ -0,0 +1,416 @@
...
+    } else {
+        AssetCleanup::Keep
+    }
+}
+
+// This function compares the `current_epoch_data` with the epoch data stored in the database.
+// If the epoch id changed, then all assets (triples & presignatures) are deleted.
+// If the epoch id did not change, then the node will discard all owned assets (triples &
+// presignatures) involving any nodes that changed their TLS key.
+pub fn delete_stale_triples_and_presignatures(
```
netrome, line 105: Code organization nit: This feels like the most significant function in this module so I'd find it more readable to have it defined first.

node/src/asset_cleanup.rs
netrome: From my shallow knowledge, I'd start with just replacing this module with an `asset_cleanup/test_utils.rs` module because that's what we have right now.

However, pondering some more it boils down to the purpose of the module. Is this general test utilities for database interactions? -> `db/test_utils.rs`, or is this relating only to asset generation then `assets/test_utils.rs` makes more sense.

netrome: Pondering some more about this I'd propose to make `assets_cleanup` as submodule of `assets` as well. What would you think about this structure for the modules?

```
assets.rs
assets/cleanup.rs
assets/test_utils.rs
```

libs/chain-signatures/contract/src/primitives/key_state.rs
netrome: Okay fair enough. Feels like that code should belong in the MPC node and not the contract then, but okay I can live with this.

node/src/providers/ecdsa/triple.rs
```diff
@@ -19,6 +20,8 @@ use threshold_signatures::protocol::Participant;
 
 pub struct TripleStorage(DistributedAssetStorage<PairedTriple>);
 
+// IMPORTANT: if we add domain_ids as an idetifier to triples, ensure to also update the asset cleanup mechanism on startup (the other place where this constant is used).
+pub const TRIPLE_STORE_DOMAIN_ID: Option<DomainId> = None;
```
netrome, line 24: It feels weird to have a const "None" value. Couldn't we just remove the option and argument to the `TripleStorage`?

netrome (commented):
Some more comments. I'm especially curious about the introduction of bincode for this particular column.

Will get back to this PR after lunch and understand the actual cleanup logic better.

node/src/assets.rs
```diff
@@ -347,6 +348,32 @@ where
     last_id: Mutex<Option<UniqueId>>,
 }
 
+/// Iterates over a key range in column `db_col`, determined by  [`DistributedAssetStorage::<T>::make_prefix_range(my_participant_id, domain_id)`],
+/// Deletes all entries in `db_col` that evaluate `false` for `is_subset_of_active_participants(persistent_participants)`
+pub fn clean_db<T>(
+    db: &Arc<SecretDB>,
+    db_col: DBCol,
+    persistent_participants: &Vec<ParticipantId>,
```
netrome, line 356: It's idiomatic to take references to slices instead of vecs

```suggestion
    persistent_participants: &[ParticipantId],
```

node/src/assets.rs
```diff
@@ -347,6 +348,32 @@ where
     last_id: Mutex<Option<UniqueId>>,
 }
 
+/// Iterates over a key range in column `db_col`, determined by  [`DistributedAssetStorage::<T>::make_prefix_range(my_participant_id, domain_id)`],
+/// Deletes all entries in `db_col` that evaluate `false` for `is_subset_of_active_participants(persistent_participants)`
+pub fn clean_db<T>(
+    db: &Arc<SecretDB>,
```
netrome, line 354: References to pointers is also a bit weird. You should be able to just do

```suggestion
    db: &SecretDB,
```

node/src/assets.rs
```diff
@@ -347,6 +348,32 @@ where
     last_id: Mutex<Option<UniqueId>>,
 }
 
+/// Iterates over a key range in column `db_col`, determined by  [`DistributedAssetStorage::<T>::make_prefix_range(my_participant_id, domain_id)`],
+/// Deletes all entries in `db_col` that evaluate `false` for `is_subset_of_active_participants(persistent_participants)`
+pub fn clean_db<T>(
```
netrome, line 353: It feels like this could/should be a method on the `DistributedAssetStorage` type. Or is there anything preventing this?

node/src/assets.rs
netrome: Nvm I see the `db.update()` method requires this. We should refactor this separately, but don't need to do it in this PR.

node/src/coordinator.rs
```diff
@@ -382,9 +382,24 @@ impl Coordinator {
...
+                epoch_id: current_epoch_id,
+                participants: current_participants_config,
+            };
+            delete_stale_triples_and_presignatures(
+                &secret_db,
+                current_epoch_data,
+                my_participant_id,
+                all_domains,
+            )?;
+        }
```
netrome, line 402: Does this mean we only run this when we enter the running state? What if the nodes are running for a long time and one participant updates their state?

netrome (changes requested):
Mostly looks good. Some nits and questions from my side. My main concerns that I'd like to see updated or get more clarification on are:

1. What happens if a participant updates their parameters while the contract is running (maybe I'm misunderstanding something here).
2. Can we be consistent with the database serialization format?

Beyond that it would be great to see an integration test where we start a few nodes, generate triplets and presignatures, have one node update their config and assert that the old assets have been wiped. I guess part of this is blocked by [#948](https://github.com/near/mpc/issues/948) so we'd have to make a follow up for it. What do you think?

node/src/coordinator.rs
netrome: Got clarification on slack. The purpose of https://github.com/near/mpc/pull/1032 is to handle that.

node/src/asset_cleanup.rs
kevindeforth: done, nice!

node/src/providers/ecdsa/triple.rs
kevindeforth: There already is no such option for the `TripleStorage` constructor: https://github.com/near/mpc/blob/9250afd3f6055d2e726908310c7a064f112a0e35/node/src/providers/ecdsa/triple.rs#L26

node/src/assets.rs
netrome: Ah right, because the storage is an in-memory copy of the on-disk database. Makes sense, thanks for clarifying. Although it feels weird to explicitly load all data in-memory in the first place. Afaik RocksDB will keep most relevant data in-memory already.

node/Cargo.toml
netrome: > Not sure why I used bincode, probably skill issue.

Wouldn't call it a skill issue. In isolation it's a good choice. Bincode is probably (most certainly) a better choice than serde_json overall. It's just a matter of consistency.

PR #1036, opened 2025-09-05
chore: test for new docker images

andrei-near:
Adding a CI test for resulting Docker image.
This will try to start a generic docker mpc container and report back if it fails.
Example failure run: https://github.com/near/mpc/actions/runs/17494915394/job/49693182812?pr=1036

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,9 +63,8 @@ jobs:
           # Check if container is actually running
           sleep 60
           if [ -z "$(docker ps --filter "id=$CONTAINER_ID" --format "{{.ID}}")" ]; then
-            echo "❌ Container failed configuration validation or startup. Checking logs..."
             docker logs "$CONTAINER_ID" 2>&1 | head -50
-            echo "❌ Container cannot start properly - this indicates a real problem"
+            echo "❌ Container cannot initialize/start properly"
             exit 1
           fi
           
--- a/deployment/Dockerfile-mpc
+++ b/deployment/Dockerfile-mpc
@@ -1,4 +1,4 @@
-FROM rust:latest AS builder
+FROM rust:1.85 AS builder
 RUN apt-get update -y && apt-get install -y --no-install-recommends clang  # needed for rocksdb
 
 WORKDIR /app


.github/workflows/ci.yml
```diff
@@ -30,6 +30,37 @@ jobs:
...
+          # Test container startup - fail if container can't start
+          timeout -k 10 120s docker run --rm \
+          -e MPC_HOME_DIR="/tmp" \
+          -e MPC_ACCOUNT_ID=test_image.near \
+          -e RUST_BACKTRACE="full" \
+          -e RUST_LOG="mpc=debug,info" \
+          -e MPC_SECRET_STORE_KEY=BD399143F5B3126098B0EAA023A0E730 \
+          -e MPC_P2P_PRIVATE_KEY=ed25519:2WBi5gRyJntYA7dCyddiwNV2yNTdr5uZhrP4WX5GNBL5DSVCSR7ESvcXF2DBfY5oPYhzBmHnguPVSXjr6UCi8h1g \
+          -e MPC_ACCOUNT_SK=ed25519:6X3Bnghdf89WjHQfHDFYjW2UeNaNqCQ1AUGxX7zgvbJT4KSoeXzuHEdux6A2jsphTZTmS4SUGQRyYqC2ik3UrMP \
+          -e NEAR_BOOT_NODES=ed25519:ERguu7jQuYk8pxNsRC6FdezvNsegBPva1GRGqjmtD7i2@10.10.10.10:24567 \
```
netrome, line 48: What nodes are these? Could we set dummy values here instead? I assume the purpose of this test is not to have this node connect to another node anyway,

.github/workflows/ci.yml
```diff
@@ -30,6 +30,37 @@ jobs:
...
+          -e MPC_CONTRACT_ID=v1.signer \
+          -e MPC_IMAGE_HASH=5ba283860c0efa3d4c3e08a76a2b77fab4725baad4f48504eac858e04af7fd64 \
+          -e MPC_LATEST_ALLOWED_HASH_FILE=/tmp/image-digest.bin \
+          -e MPC_ENV=mainnet test_image_tag_ci
```
netrome, line 52: It would be more resilient to use localnet here imo. This way we don't accidentally try to connect to mainnet nodes.

```suggestion
          -e MPC_ENV=localnet test_image_tag_ci
```

netrome (changes requested):
Mostly looks good to me. However, it would be good to not try to connect to mainnet nodes from our CI. Also there's a CI error due to a 137 exit code. Not sure what that means.

PR #1040, opened 2025-09-05
fix: Add missing bs58 dependency in contract_history crate

### LINKED ISSUE #1039: Add bs58 as dev dependency for `contract_history` crate
### Background

I have no idea how this happened, but for some reason `bs58` was not included as a dev-dependency in https://github.com/near/mpc/pull/1031/commits/1a4dce76b1ef1acb1413f600f0c032d8e28b21ff - which breaks CI on our main.

This is very strange, since I tested this on my laptop before pushing - which should not be possible if I had not added the dependency. I also strictly remember adding this - although memories can be false. I have a few theories of what happened:

1. I forgot to add `Cargo.toml` while committing.
2. Someone has hacked my laptop and decided to prank me.
3. Cosmic background radiation.

From my shell history we can rule out 1, since I always do `git add -A` followed by `git commit -a`, so it's most likely 3.

### User Story

As a dev I want our CI to be green.

### Acceptance Criteria

We've added the missing bs58 dependency.

_No response_

netrome:
Closes #1039

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2079,7 +2079,7 @@ dependencies = [
 name = "contract_history"
 version = "0.1.0"
 dependencies = [
- "base64 0.22.1",
+ "bs58 0.5.1",
  "clap",
  "reqwest 0.12.23",
  "serde",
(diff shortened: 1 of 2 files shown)


gilcu3 (approved):
It was a good trick to test the CI from the previous PR 🫣

PR #1061, opened 2025-09-09
feat: Change of participant set leads to exit of running state.

### LINKED ISSUE #948: Attestation Service

kevindeforth:
Resolves #948

### DIFF
--- a/libs/chain-signatures/contract/src/primitives/participants.rs
+++ b/libs/chain-signatures/contract/src/primitives/participants.rs
@@ -192,9 +192,9 @@ impl Participants {
         account_id: AccountId,
         new_info: ParticipantInfo,
     ) -> Result<(), Error> {
-        for (a_id, _, info) in self.participants.iter_mut() {
-            if *a_id == account_id {
-                *info = new_info.clone();
+        for (participant_id, _, participant_info) in self.participants.iter_mut() {
+            if *participant_id == account_id {
+                *participant_info = new_info.clone();
                 return Ok(());
             }
         }
--- a/libs/chain-signatures/contract/src/primitives/participants.rs
+++ b/libs/chain-signatures/contract/src/primitives/participants.rs
@@ -52,10 +52,12 @@ impl Participants {
             participants: Vec::new(),
         }
     }
+
     #[allow(clippy::len_without_is_empty)]
     pub fn len(&self) -> usize {
         self.participants.len()
     }
+
     pub fn insert_with_id(
         &mut self,
         account_id: AccountId,
@@ -77,12 +79,15 @@ impl Participants {
         self.next_id.0 = id.0 + 1;
         Ok(())
     }
+
     pub fn insert(&mut self, account_id: AccountId, info: ParticipantInfo) -> Result<(), Error> {
         self.insert_with_id(account_id, info, self.next_id.clone())
     }
+
     pub fn participants(&self) -> &Vec<(AccountId, ParticipantId, ParticipantInfo)> {
         &self.participants
     }
+
     pub fn next_id(&self) -> ParticipantId {
         self.next_id.clone()
     }
@@ -109,6 +114,7 @@ impl Participants {
         }
         Ok(())
     }
+
     pub fn init(
         next_id: ParticipantId,
         participants: Vec<(AccountId, ParticipantId, ParticipantInfo)>,
@@ -129,12 +135,14 @@ impl Participants {
             .map(|(_, p_id, _)| p_id.clone())
             .ok_or_else(|| crate::errors::InvalidState::NotParticipant.into())
     }
+
     pub fn info(&self, account_id: &AccountId) -> Option<&ParticipantInfo> {
         self.participants
             .iter()
             .find(|(a_id, _, _)| a_id == account_id)
             .map(|(_, _, info)| info)
     }
+
     pub fn account_id(&self, id: &ParticipantId) -> Result<AccountId, Error> {
         self.participants
             .iter()
@@ -142,11 +150,13 @@ impl Participants {
             .map(|(a_id, _, _)| a_id.clone())
             .ok_or_else(|| crate::errors::InvalidState::ParticipantIndexOutOfRange.into())
     }
+
     pub fn is_participant(&self, account_id: &AccountId) -> bool {
         self.participants
             .iter()
             .any(|(a_id, _, _)| a_id == account_id)
     }
+
     /// Returns a subset of the participants according to the given range of indices.
     pub fn subset(&self, range: std::ops::Range<usize>) -> Participants {
         let participants = self.participants[range]
@@ -157,6 +167,7 @@ impl Participants {
             participants: participants.collect(),
         }
     }
+
     pub fn add_random_participants_till_n(&mut self, n: usize) {
         let mut rng = rand::thread_rng();
         while self.len() < n {
@@ -165,6 +176,7 @@ impl Participants {
             self.insert(account, pinfo).unwrap();
         }
     }
+
     pub fn remove(&mut self, account: &AccountId) {
         if let Some(pos) = self
             .participants
@@ -173,6 +185,20 @@ impl Participants {
         {
             self.participants.remove(pos);
         }
+    }
+
+    pub fn change_info(
+        &mut self,
+        account_id: AccountId,
+        new_info: ParticipantInfo,
+    ) -> Result<(), Error> {
+        for (a_id, _, info) in self.participants.iter_mut() {
+            if *a_id == account_id {
+                *info = new_info.clone();
+                return Ok(());
+            }
+        }
+        Err(crate::errors::InvalidState::NotParticipant.into())
     }
 }
 
--- a/libs/chain-signatures/contract/src/primitives/participants.rs
+++ b/libs/chain-signatures/contract/src/primitives/participants.rs
@@ -187,7 +187,7 @@ impl Participants {
         }
     }
 
-    pub fn change_info(
+    pub fn update_info(
         &mut self,
         account_id: AccountId,
         new_info: ParticipantInfo,
--- a/node/src/config.rs
+++ b/node/src/config.rs
@@ -61,11 +61,14 @@ impl MpcConfig {
     pub fn from_participants_with_near_account_id(
         participants: ParticipantsConfig,
         my_near_account_id: &AccountId,
+        my_p2p_public_key: &ed25519_dalek::VerifyingKey,
     ) -> Option<Self> {
         let my_participant_id = participants
             .participants
             .iter()
-            .find(|p| &p.near_account_id == my_near_account_id)?
+            .find(|p| {
+                &p.near_account_id == my_near_account_id && &p.p2p_public_key == my_p2p_public_key
+            })?
             .id;
         Some(Self {
             my_participant_id,
--- a/node/src/coordinator.rs
+++ b/node/src/coordinator.rs
@@ -684,7 +684,7 @@ impl Drop for ReportCurrentJobGuard {
     }
 }
 
-/// returns true if one if one of the following occurs:
+/// returns true if one of the following occurs:
 /// - the epoch id changes
 /// - a resharing starts
 /// - the participant set changes
--- a/node/src/coordinator.rs
+++ b/node/src/coordinator.rs
@@ -2,7 +2,9 @@ use crate::assets::cleanup::{delete_stale_triples_and_presignatures, EpochData};
 use crate::config::{ConfigFile, MpcConfig, ParticipantsConfig, SecretsConfig};
 use crate::db::SecretDB;
 use crate::indexer::handler::ChainBlockUpdate;
-use crate::indexer::participants::{ContractKeyEventInstance, ContractRunningState, ContractState};
+use crate::indexer::participants::{
+    ContractKeyEventInstance, ContractResharingState, ContractRunningState, ContractState,
+};
 use crate::indexer::types::ChainSendTransactionRequest;
 use crate::indexer::IndexerAPI;
 use crate::key_events::{
@@ -28,6 +30,7 @@ use crate::web::DebugRequest;
 use futures::future::BoxFuture;
 use futures::FutureExt;
 use mpc_contract::primitives::domain::{DomainId, SignatureScheme};
+use mpc_contract::primitives::key_state::EpochId;
 use near_time::Clock;
 use std::collections::HashMap;
 use std::future::Future;
@@ -116,8 +119,7 @@ impl Coordinator {
                     // For initialization state, we generate keys and vote for the public key.
                     // We give it a timeout, so that if somehow the keygen and voting fail to
                     // progress, we can retry.
-                    let (key_event_sender, key_event_receiver) =
-                        watch::channel(state.key_event.clone());
+                    let (key_event_receiver, stop_fn) = make_initializing_stop_fn(state.key_event);
                     MpcJob {
                         name: "Initializing",
                         fut: Self::create_runtime_and_run(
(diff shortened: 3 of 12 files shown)


libs/chain-signatures/contract/src/primitives/participants.rs
```diff
@@ -174,6 +186,20 @@ impl Participants {
             self.participants.remove(pos);
         }
     }
+
+    pub fn change_info(
```
netrome, line 190: Nit(name suggestion): I'd call this method `update_info`. It feels more conventional but I don't have a motivation for why at the top of my head.

```suggestion
    pub fn update_info(
```

libs/chain-signatures/contract/src/primitives/participants.rs
```diff
@@ -174,6 +186,20 @@ impl Participants {
...
+        &mut self,
+        account_id: AccountId,
+        new_info: ParticipantInfo,
+    ) -> Result<(), Error> {
+        for (a_id, _, info) in self.participants.iter_mut() {
+            if *a_id == account_id {
```
netrome, line 196: Nit: This is slightly easier to read imo.

```suggestion
        for (participant_id, _, info) in self.participants.iter_mut() {
            if *participant_id == account_id {
```

node/src/config.rs
```diff
@@ -61,11 +61,14 @@ impl MpcConfig {
...
         my_near_account_id: &AccountId,
+        my_p2p_public_key: &ed25519_dalek::VerifyingKey,
     ) -> Option<Self> {
         let my_participant_id = participants
             .participants
             .iter()
-            .find(|p| &p.near_account_id == my_near_account_id)?
+            .find(|p| {
+                &p.near_account_id == my_near_account_id && &p.p2p_public_key == my_p2p_public_key
+            })?
```
netrome, line 71: So the purpose of this is that if our node configuration is not present in the participant config, the node will halt?

Shouldn't we also check the address and port here? Or is that just a practical limitation since the MPC node is unaware of this information?

node/src/coordinator.rs
```diff
@@ -736,3 +683,131 @@ impl Drop for ReportCurrentJobGuard {
...
+) -> (watch::Receiver<ContractKeyEventInstance>, StopFn) {
+    let (key_event_sender, key_event_receiver) = watch::channel(key_event.clone());
+    let key_event_sender = key_event_sender.clone();
+    (
+        key_event_receiver,
+        Box::new(move |new_state| {
+            stop_initializing(new_state, key_event.id.epoch_id, &key_event_sender)
+        }),
+    )
+}
```
netrome, line 813: Just so that I'm not missing anything important, there's no logic change here right? Except for the extra log line if the key event receiver is closed. Beyond that, this seems like a pure (and nice) refactor - correct?

node/src/coordinator.rs
```diff
@@ -736,3 +683,131 @@ impl Drop for ReportCurrentJobGuard {
...
+                return true;
+            }
+            if new_state.resharing_state.is_some() {
+                tracing::info!("A resharing started.");
+                return true;
+            }
+            if new_state.participants != current_participant_set {
+                tracing::info!("Participant details changed.");
+                return true;
+            }
```
netrome, line 709: And this is the new condition. Looks good!

node/src/coordinator.rs
```diff
@@ -736,3 +683,131 @@ impl Drop for ReportCurrentJobGuard {
...
+            current_running_epoch_id,
+            current_participant_set.clone(),
+        )
+    })
+}
+
+/// returns true if one of the following occurs:
+///     - epoch id changed
+///     - resharing concludes
+///     - key event receiver closes the channel.
```
netrome, line 735: Right and naturally we don't allow participant updates during resharing. Makes sense.

node/src/indexer/fake.rs
```diff
@@ -215,6 +216,46 @@ impl FakeMpcContractState {
             }
         }
     }
+
+    pub fn change_participant_info(
```
netrome, line 220: Nit: Not sure why I react to this. Just pure intuition/habit. Feel free to disagree/ignore.

```suggestion
    pub fn update_participant_info(
```

node/src/indexer/fake.rs
```diff
@@ -479,11 +526,13 @@ pub struct FakeIndexerManager {
     ckd_request_sender: mpsc::UnboundedSender<CKDRequestFromChain>,
 
     /// Allows nodes to be disabled during tests. See `disable()`.
-    node_disabler: HashMap<AccountId, NodeDisabler>,
+    node_disabler: BTreeMap<Uid, NodeDisabler>,
     /// Allows nodes' indexers to be paused during tests.
-    indexer_pauser: HashMap<AccountId, IndexerPauser>,
+    indexer_pauser: BTreeMap<Uid, IndexerPauser>,
```
netrome, line 531: Does ordering matter here now?

node/src/indexer/fake.rs
```diff
@@ -542,14 +591,15 @@ impl Drop for PausedIndexer {
...
     }
 }
-
+#[derive(Clone, Copy, PartialEq, PartialOrd, Eq, Ord, From)]
+pub struct Uid(usize);
 /// Runs the fake indexer logic for one node.
 struct FakeIndexerOneNode {
     /// Account under which transactions by this node are originated.
-    account_id: AccountId,
+    uid: Uid,
```
netrome, line 599: The docstring is confusing here now. What are these Uids anyway? Node IDs? 

```suggestion
/// Internal ID used to organize nodes for tests
uid: Uid,
```

node/src/indexer/fake.rs
```diff
@@ -542,14 +591,15 @@ impl Drop for PausedIndexer {
         self.indexer_suspended.send(false).unwrap();
     }
 }
-
+#[derive(Clone, Copy, PartialEq, PartialOrd, Eq, Ord, From)]
+pub struct Uid(usize);
```
netrome, line 595: Just guessing from the context. This is not a user ID right? Or what does the `U` stand for?

```suggestion
pub struct NodeId(usize);
```

node/src/indexer/fake.rs
```diff
@@ -479,11 +526,13 @@ pub struct FakeIndexerManager {
...
     /// Allows nodes to be disabled during tests. See `disable()`.
-    node_disabler: HashMap<AccountId, NodeDisabler>,
+    node_disabler: BTreeMap<Uid, NodeDisabler>,
     /// Allows nodes' indexers to be paused during tests.
-    indexer_pauser: HashMap<AccountId, IndexerPauser>,
+    indexer_pauser: BTreeMap<Uid, IndexerPauser>,
     /// Allows modification of the contract.
     contract: Arc<tokio::sync::Mutex<FakeMpcContractState>>,
+
+    account_id_by_uid: Arc<Mutex<BTreeMap<Uid, AccountId>>>,
```
netrome, line 535: This does not seem to be held over any await boundaries, so I'd recommend using the `std::sync::Mutex` over the tokio one.

node/src/indexer/fake.rs
```diff
@@ -678,8 +731,9 @@ impl FakeIndexerManager {
 
     /// Adds a new node to the fake indexer. Returns the API for the node, a task that
     /// runs the node's logic, and the running job name to passed to the coordinator.
-    pub fn add_indexer_node(
+    pub async fn add_indexer_node(
```
netrome, line 734: From what I can see the only reason to make this async now is due to the use of a `tokio::sync::Mutex` over the `std` one. I'd recommend to go with the `std` mutex here instead.

netrome: Also this is a good read for more context: https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use

node/src/tests.rs
```diff
@@ -264,10 +283,10 @@ pub struct IntegrationTestSetup {
 impl IntegrationTestSetup {
     /// Generates test node configs and a fake indexer; each config can then be used
     /// to start running the node.
-    pub fn new(
+    pub async fn new(
```
netrome, line 286: Shouldn't need to be async if we change the Mutex

node/src/tests.rs
```diff
@@ -321,17 +340,18 @@ impl IntegrationTestSetup {
                     near_signer_key: ed25519_dalek::SigningKey::generate(&mut OsRng),
                     near_responder_keys: vec![ed25519_dalek::SigningKey::generate(&mut OsRng)],
                 },
-                local_storage_aes_key: rand::random(),
+                local_storage_aes_key: rand::random(), // change this after the fact
```
netrome, line 343: Not following this comment, can you elaborate?

node/src/tests.rs
```diff
@@ -264,10 +283,10 @@ pub struct IntegrationTestSetup {
 impl IntegrationTestSetup {
     /// Generates test node configs and a fake indexer; each config can then be used
     /// to start running the node.
-    pub fn new(
+    pub async fn new(
         clock: Clock,
         temp_dir: &Path,
-        participant_accounts: Vec<AccountId>,
+        participant_accounts: Vec<AccountId>, // we create per account. lol.
```
netrome, line 289: Hate to be boring here, but I don't get it, should this be changed?

node/src/tests/changing_participant_details.rs
```diff
@@ -0,0 +1,144 @@
...
+use near_o11y::testonly::init_integration_logger;
+use near_time::Clock;
+
+/// Runs a cluster of 3 nodes, but with only 2 participants.
+/// Two nodes of the cluster are assigned the same account id.
+/// Each node has its own home directory and tls key.
+/// After the conclusion of the key initialization and passing of a sanity check, the participant
+/// set will be forcefully changed.
+#[tokio::test]
+async fn test_changing_participant_set() {
```
netrome, line 21: Nice test!

netrome (changes requested):
Nice stuff. While I've added a bunch of nits and smaller suggestions, overall I think this looks very good. Only two small hard blockers that I'd like to see updated before approving:

1. The `Uid` type in the fake indexer invalidated an existing doc string that must be updated. I'd also find it helpful with a more informative name to clarify the purpose of this type.
2. The `account_id_by_uid` field uses a `tokio::sync::Mutex` forcing a bunch of previously sync functions to be async. This should not be necessary, and I'd like to use a normal `std::sync::Mutex` instead.


Other than that I'm very eager to get this through!

node/src/indexer/fake.rs
kevindeforth: Uid stands for [unique identifier](https://en.wikipedia.org/wiki/Unique_identifier).
Until now, we used the `AccountId` as a unique identifier in the FakeIndexer. This won't work with the proposed disaster recovery plan.

node/src/tests.rs
kevindeforth: oh yeah, that was a note to myself :D

node/src/tests.rs
kevindeforth: Not relevant anymore. Removed.

node/src/config.rs
kevindeforth: > Shouldn't we also check the address and port here? 

I don't think that is necessary. When all nodes run inside a TEE, then the TLS key effectively acts as a unique identifier for the nodes.

node/src/coordinator.rs
kevindeforth: Exactly, I didn't intend to change any logic here, just a refactor.

node/src/coordinator.rs
kevindeforth: Yeah, I was thinking of adding a sanity check here. If we ever encounter the case where the participant set changes during a resharing, then this is a serious bug and means the contract has an issue.

I didn't add the check, because it's unclear how we would want to recover from that error.

node/src/keyshare.rs
```diff
@@ -269,6 +269,19 @@ impl KeyshareStorage {
     }
 }
 
+#[cfg(test)]
+impl KeyshareStorage {
+    // Do NOT use this for production
```
DSharifi, line 274: Make it a doc instead of comment.
```suggestion
    /// Do NOT use this for production
```

DSharifi (changes requested):
Looks good to me as well. Only hard blocker is the `Uid` name as well.

node/src/keyshare.rs
```diff
@@ -269,6 +269,19 @@ impl KeyshareStorage {
...
+    pub async fn put_keyshares(&self, keyshares: Vec<Keyshare>) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
DSharifi: (It's fine to keep the `Impl`, but I'm cautious about the current fn name). We should at least change the current fn name and prefix it with something like `dangerous_`

node/src/coordinator.rs
```diff
@@ -736,3 +683,131 @@ impl Drop for ReportCurrentJobGuard {
...
+    current_participant_set: ParticipantsConfig,
+) -> StopFn {
+    Box::new(move |new_state| {
+        stop_running(
+            new_state,
+            current_running_epoch_id,
+            current_participant_set.clone(),
+        )
+    })
+}
```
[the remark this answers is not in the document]
DSharifi: Is it really more readable? There's a cognitive overhead of having to jump to the code definition at a single use call site vs. having all the logic visible. Especially when it's just a wrapper with single digit lines of logic. The function is also several hundred lines down so I have to jump to definition to see the logic.

If we change the variable name from `stop_fn` to `running_stop_fn` it's also more readable and should give you the same result without the cognitive load of jumping to the wrapper.

```rust
let stop_fn = make_running_stop_fn(
    running_state.keyset.epoch_id,
    running_state.participants.clone(),
);


// NOTE THERE `make_running_stop_fn` IS NOT VISIBLE WHEN READING THE CODE ABOVE 


fn make_running_stop_fn(
    current_running_epoch_id: EpochId,
    current_participant_set: ParticipantsConfig,
) -> StopFn {
    Box::new(move |new_state| {
        stop_running(
            new_state,
            current_running_epoch_id,
            current_participant_set.clone(),
        )
    })
}
```
vs

```rust
let running_stop_fn = Box::new(move |new_state| {
        stop_running(
            new_state,
            current_running_epoch_id,
            current_participant_set.clone(),
        )
    });
```

node/src/config.rs
netrome: > When all nodes run inside a TEE, then the TLS key effectively acts as a unique identifier for the nodes.

Right because it's generated inside the TEE so you can't reuse it across instances. Fair point.

node/src/coordinator.rs
netrome: Right yeah but since resharing can be cancelled it makes sense for the resharing operation to fail first and recovery to happen after.

node/src/coordinator.rs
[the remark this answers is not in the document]
netrome: Personally I like the wrapper more in this case, but this is definitely a matter of taste. However there are a few things that I think are significant improvements with the wrapper:

1. It does make the calling function leaner. In most cases I'm not chasing down every single code path, and having the calling function cleaner makes it easier to grasp what it's doing.
2. Testability. In this case, `make_running_stop_fn` can be tested in isolation. If you suspect there's a bug in this logic, it's quicker to verify when hunting it down.
3. Scope reduction. Calling `make_running_stop_function` only captures the arguments passed to it. A closure may catch anything in the scope, and I have to read all lines of it to know what variables it interacts with. This way, a function call is inherently simpler than a closure.

That said, I do see that for very small functions the extra overhead of declaring a new function and the cost of jumping in the code is sometimes not worth it.

DSharifi (approved):

node/src/tests.rs
netrome: Nice, seems like you had fun coding this :blush:

PR #1066, opened 2025-09-10
test(contract): system test for `submit_tee_participant_info`

### LINKED ISSUE #849: System test for `submit_tee_participant_info`
### Background

We need a pytest ensuring the nodes succesfully call the endpoint `submit_tee_participant_info`, to avoid API mismatches like we had in #842 

### User Story

As a dev, I want test coverage.

### Acceptance Criteria

We have a pytest ensuring the nodes succesfully call the endpoint `submit_tee_participant_info`

_No response_

pbeza:
Fixes #849


crates/contract/src/lib.rs
```diff
@@ -1041,7 +1041,9 @@ impl MpcContract {
                 let threshold = current_params.threshold().value() as usize;
                 let remaining = participants_with_valid_attestation.len();
                 if threshold > remaining {
-                    log!("Less than `threshold` participants are left with a valid TEE status. This requires manual intervention. We will not accept new signature requests as a safety precaution.");
+                    log!(
+                        "Less than `threshold` participants are left with a valid TEE status. This requires manual intervention. We will not accept new signature requests as a safety precaution."
+                    );
```
DSharifi, line 1046: Why the change here?

pytest/tests/test_submit_participant_info.py
```diff
@@ -0,0 +1,89 @@
...
+                    "   ❌ No TEE accounts registered - submit_participant_info calls failed"
+                )
+                success = False
+        else:
+            print("   ❌ No SuccessValue found in contract response")
+            success = False
+    else:
+        print("   ❌ Invalid contract response format")
+        success = False
+
```
DSharifi, line 85: Instead of these nested `if`s why not just assert?

crates/contract/src/lib.rs
DSharifi: Check if you have formatting settings in your `.vscode` folder in the root directory of the repo.

pytest/tests/test_submit_participant_info.py
pbeza: Good point! Fixed: https://github.com/near/mpc/pull/1066/commits/b645a5b45f05c53194412c564fe744fba0a07346. ✅

pbeza:
@DSharifi, I just addressed all the review comments. Feel free to re-review. Thanks!

pytest/tests/test_submit_participant_info.py
```diff
@@ -0,0 +1,94 @@
...
+    initial_participants = 2
+    total_nodes = 4
+    cluster, mpc_nodes = shared.start_cluster_with_mpc(
+        2, total_nodes, 1, load_mpc_contract(), start_mpc_nodes=False
+    )
+
+    # Start only the initial participant nodes first
+    for i, node in enumerate(mpc_nodes[:initial_participants]):
+        print(f"   Starting initial participant node {i + 1}: {node.p2p_public_key}")
+        node.run()
```
DSharifi, line 35: Why set `start_mpc_nodes` to False if we are starting the node immediately afterwards?

DSharifi (approved):
Thanks for the good work @pbeza 

I think the test has much more logic than needed. Doesn't this give us all the coverage we need?
```python3
def test_submit_participant_info_endpoint():
    initial_participants = 2
    total_nodes = 4
    cluster, mpc_nodes = shared.start_cluster_with_mpc(
        2, total_nodes, 1, load_mpc_contract(), start_mpc_nodes=False
    )
    cluster.init_cluster(mpc_nodes[:initial_participants], 2)

    attestations_submitted = cluster.wait_for_nodes_to_have_attestation(
        mpc_nodes
    )

    assert attestations_submitted, (
        f"Timeout: Not all {len(additional_nodes)} additional nodes submitted attestations within the timeout period. "
        f"Check the debug output above for which nodes failed to submit."
    )
```

Issue #527, opened 2025-06-19
Pytest follow-up. Use restricted access keys


kevindeforth:
In the pytests, we currently make the signer_pk and respond keys full access keys . It would be better to only grant them access for calling the contract methods that they actually require.

old branch:
https://github.com/near/mpc/blob/4af607d092eb1ecfe22f44a490b4e98bc1fd7d8f/pytest/common_lib/shared/__init__.py#L45

netrome:
Unblocked, we just needed to kill some processes using shared ports. Unfortunately the error messages were very verbose making this take more time than it should.

PR #1070, opened 2025-09-11
feat(pytest): restrict signer keys to MPC contract method (clean)

barakeinav1:
fixes https://github.com/near/mpc/issues/527


kevindeforth (approved):
Thanks a lot!

PR #1067, opened 2025-09-10
feat: enforce valid TEE attestations for all nodes

DSharifi:
closes #1052

PR #1094, opened 2025-09-12
fix: retry sending of attestation submission transaction

### LINKED ISSUE #1069: MPC node should retry to submit participant info in case it failed
### Background

During first boot flow, submit_tee_participant_info is called by the MPC node in order to submit its attestation information and P2P key.

This method must be called ONLY after the operator has added the node’s key to their account. otherwise it will fail.
There are several options to handle such a failure.
see 
https://nearone.slack.com/archives/C07UW93JVQ8/p1755244298678639?thread_ts=1755181284.788119&cid=C07UW93JVQ8

Option choose:  The MPC node will check if the submission was successful and retry to submit in a loop.

check state:  get_tee_accounts and compare to the node's account.
https://github.com/near/mpc/blob/123dde9454e267e2f17363b7f1f120ab124806c8/libs/chain-signatures/contract/src/lib.rs#L1097C12-L1097C27

 

### Acceptance Criteria

Code is implemented and tested.

slack discussion 

https://nearone.slack.com/archives/C07UW93JVQ8/p1755244298678639?thread_ts=1755181284.788119&cid=C07UW93JVQ8

DSharifi:
closes #1069 

The changes here are tested in pytests in #1067

### DIFF
--- a/node/src/indexer/tx_sender.rs
+++ b/node/src/indexer/tx_sender.rs
@@ -20,6 +20,8 @@ use std::time::Duration;
 use tokio::sync::mpsc;
 use tokio::time;
 
+const GET_TEE_ACCOUNTS_METHOD_NAME: &str = "get_tee_accounts";
+
 /// Creates, signs, and submits a function call with the given method and serialized arguments.
 async fn submit_tx(
     tx_signer: Arc<TransactionSigner>,
@@ -180,7 +182,7 @@ async fn observe_tx_result(
                         block_reference: BlockReference::Finality(Finality::Final),
                         request: QueryRequest::CallFunction {
                             account_id: indexer_state.mpc_contract_id.clone(),
-                            method_name: "get_tee_accounts".to_string(),
+                            method_name: GET_TEE_ACCOUNTS_METHOD_NAME.to_string(),
                             args: get_pending_request_args.into(),
                         },
                     }


barakeinav1:
looks good.
I don't have a particular comments in addition to what was siad before.
 
BTW, we call submit_remote_attestationonly in case the MPC node's hash is matches the valid hash on the contract,  correct? otherwise, 
the node will just call submit_remote_attestation without any possibility of success.

pbeza (approved):

netrome:
> If so, I think submitting the new (real) attestation should just succeed, so GET_TEE_ACCOUNTS_METHOD_NAME won’t be called—since there’s no need to retry the attestation submission. Or am I missing something?

I think the point is that we use the current state to determine whether or not to retry, irrespective of the status of the transaction. Right now the flow is roughly:

1. Submit transaction.
2. Observe if we have a valid attestation or not.
3. If invalid, retry.

Imagine now that we submit an updated attestation, but the transaction fails or isn't picked up for any reason. We should retry in this case, but since the contract has a valid attestation already we won't.

So yeah that's a bug. Great observation @barakeinav1!

DSharifi:
Closed in favor of #1174 which uses the new `TransactionSender` API

PR #1102, opened 2025-09-13
chore: Dsharifi/contract in workspace pls

DSharifi:


netrome (approved):

deployment/Dockerfile-mpc-tee
```diff
@@ -15,6 +15,7 @@ WORKDIR /app
 COPY .git .git
 COPY node node
 COPY devnet devnet
+COPY pytest/test_contracts pytest/test_contracts
```
netrome, line 18: Why do we add the test contracts to the dockerfile?

I know this is just part of the build image so it doesn't affect the size of the final image but still.

netrome: Right thanks for clarifying :pray:

PR #1111, opened 2025-09-15
chore: remove enum `VersionedMpcContract` contract state in favor of a single version struct

### LINKED ISSUE #1110: `MpcContractVersion` struct does not need to be an enum
### Background

The contract implementation is annoying to work with. It's represented as an enum, where each variant of the enum is a versioned variant.

For every single public method, we currently try to extract `V2` variant of the enum, otherwise we panic:
https://github.com/near/mpc/blob/5f62f1c11517eee8d9883fe0fed798e5e6a75a66/libs/chain-signatures/contract/src/lib.rs#L1077-L1080

We currently have 37 occurrences of such matches or "if let else".

I can only assume the motivation of the enum representation was a misinformed decision to be able to do upgrades. That assumption is a misinformed decision and is pure tech debt. We don't need to represent this state as an enum for upgrades to work.

An upgrade allows us to use a generic parameters for the struct definition of the previous contract state, meaning we can use new structs to represent the new contract state while serialization will still work.

https://github.com/near/mpc/blob/5f62f1c11517eee8d9883fe0fed798e5e6a75a66/libs/chain-signatures/contract/src/lib.rs#L1273-L1284

### User Story

_No response_

### Acceptance Criteria

Remove `MpcContractVersion` from the contract implementation.

_No response_

DSharifi:
closes #1110

### DIFF
--- a/crates/contract/tests/expired_attestation.rs
+++ b/crates/contract/tests/expired_attestation.rs
@@ -9,7 +9,7 @@ use mpc_contract::{
     },
     state::ProtocolContractState,
     tee::tee_state::NodeId,
-    VersionedMpcContract,
+    MpcContract,
 };
 
 use assert_matches::assert_matches;
@@ -21,7 +21,7 @@ use near_sdk::{
 use std::time::Duration;
 
 struct TestSetup {
-    contract: VersionedMpcContract,
+    contract: MpcContract,
     participants_list: Vec<(AccountId, ParticipantId, ParticipantInfo)>,
 }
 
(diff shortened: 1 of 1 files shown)


kevindeforth (changes requested):
https://nearone.slack.com/archives/C0912BTG51T/p1757928286968949?thread_ts=1757923837.366669&cid=C0912BTG51T

> Are we sure this is the motivation? One benefit the VersionedMpcContract provides is that the distinction between exposed endpoint and non-exposed endpoint is very easy to make.

By exposed endpoint, do you mean methods that will be callable as smart contract methods? That hasn't changed. We still need to explicitly expose them with the #[near_bindgen] procedural macro.
See the difference between these two Impls.
Impl that exposes contract methods: https://github.com/near/mpc/blob/ddd68966fb6bce484a5af2d51083a731b5b34279/libs/chain-signatures/contract/src/lib.rs#L124-L126
Impl that does not expose methods: https://github.com/near/mpc/blob/ddd68966fb6bce484a5af2d51083a731b5b34279/libs/chain-signatures/contract/src/lib.rs#L96-L97

kevindeforth (approved):
I came to appreciate. Thanks :+1:

PR #1113, opened 2025-09-15
chore: update ci to use reproducible builds

### LINKED ISSUE #1090: Update CI to use reproducible builds
### Background

In https://github.com/near/mpc/pull/1064 we achieved reproducible builds. Now we need to update our CI so that these builds are published to dockerhub.

### User Story

_No response_

### Acceptance Criteria

The images published to dockerhub are reproducible.

_No response_

gilcu3:
Closes #1090

### DIFF
--- a/.github/workflows/docker_release.yml
+++ b/.github/workflows/docker_release.yml
@@ -34,7 +34,11 @@ jobs:
       - name: Get SOURCE_DATE_EPOCH
         shell: bash
         run: |
-          echo "SOURCE_DATE_EPOCH=$(git log -1 --pretty=%ct)" >> "$GITHUB_ENV"
+
+      - name: Fix file timestamps
+        shell: bash
+        run: |
+          find . \( -type f -o -type d \) -exec touch -d @"${{ env.SOURCE_DATE_EPOCH }}" {} +
 
       - name: Sanitize build-ref for Docker tags
         shell: bash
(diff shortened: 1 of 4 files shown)


deployment/Dockerfile-mpc-tee
```diff
@@ -41,7 +14,7 @@ RUN \
 ENV SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt
 
 WORKDIR /app
-COPY --from=builder /app/target/x86_64-unknown-linux-gnu/release/mpc-node mpc-node
+COPY target/reproducible/mpc-node mpc-node
```
netrome, line 17: This looks strange to me. Can you explain why we can copy the `mpc-node` binary from outside the docker environment? Shouldn't we build it when we build the container?

netrome (changes requested):
I find it strange that we're delegating the building of the MPC node to the host environment. Is this intentional and if so, why?

(Nothing necessarily needs to be changed, but I really want to understand this before I'm comfortable merging this.)

deployment/Dockerfile-mpc-tee
gilcu3: because `repro-env` reuses the outside folder, so its result is copied there

netrome: Alright makes sense. Thanks for the explanation.

netrome (approved):
Thanks for the explanation.

Issue #1119, opened 2025-09-16
Fix contract integration test TLS key


kevindeforth:
### Background

We currently use a constant hard-coded tls key in our [integration tests](https://github.com/near/mpc/blob/5f3d0deecf8135a2d3c4102de923b599fdf9e6b8/libs/chain-signatures/contract/tests/common.rs#L83).

With #1109 in mind, we would prefer to have unique TLS keys for each participant.

### User Story

As a developer, I want broad test coverage.

### Acceptance Criteria

Generate a random TLS key for each participant.

DSharifi:
Closed by https://github.com/near/mpc/pull/1120

Issue #1118, opened 2025-09-16
investigate `clean_tee_status`


kevindeforth:
### Background

The contract [method `clean_tee_status`](https://github.com/near/mpc/blob/5bac07e013f3a1e6a373bc706f313aecbfce81d9/libs/chain-signatures/contract/src/lib.rs#L1189)  is [called](https://github.com/near/mpc/blob/5bac07e013f3a1e6a373bc706f313aecbfce81d9/libs/chain-signatures/contract/src/lib.rs#L930) at the end of a resharing. We currently [attach 3 Tgas](https://github.com/near/mpc/blob/5bac07e013f3a1e6a373bc706f313aecbfce81d9/libs/chain-signatures/contract/src/lib.rs#L75) to the call.
If there are more than 3 attestations to remove from the state, the cleanup call has failed in the integration test.
The`clean_tee_status` endpoint is private, in order to avoid an attack vector, where a new participant could be stopped from joining by repeatedly calling this endpoint.

One possible solution could be to allow existing participants to call `clean_tee_status`. If a participant mis-behaves and maliciously calls it, they could be removed from the network through a resharing (by kicking them out).

### User Story

Balance the need for lean contract state with opening a potential DOS attack vector.

### Acceptance Criteria

Implement a fix for the described problem, or reason why it is not relevant.

gilcu3:
related to #1461

Issue #1109, opened 2025-09-15
mock attestations should include TLS key (and also node account key)


kevindeforth:
### Background

It would be helpful for testing if constrained mock attestations would include the TLS key
https://github.com/near/mpc/blob/5f62f1c11517eee8d9883fe0fed798e5e6a75a66/crates/attestation/src/attestation.rs#L91

### User Story

As a developer, I want test capailities

### Acceptance Criteria

Mock attestation with constraints must include a TLS key.

netrome:
Sure, makes sense. Keeping this around but marking it as low priority as it's not clear what the testing gap is (i.e. which particular tests would we like to add?)

netrome:
Setting as P4

PR #1120, opened 2025-09-16
chore: index nodes by account id & TLS key in TEE State

### LINKED ISSUE #1084: Migration Service:  TEE Attestations must be indexed by p2p keys in the contract
### Background

For the migration service, the contract must allow the same node operator (`AccountId`) to store multiple TEE attestations. Instead of using `AccountId` as a unique identifier in [`TEEState`](https://github.com/near/mpc/blob/2d0a7f62835da102bc21db0adb9bbf74c6b88afd/libs/chain-signatures/contract/src/tee/tee_state.rs#L29), Attestations shall be stored with respect to the TLS key.

### User Story

As a node operator, I would like to migrate my node to a new machine without interrupting the network.

### Acceptance Criteria

An `AccountId` can have multiple TEE attestations, as long as they are for different TLS keys.

_No response_

kevindeforth:
Resolves #1084, #1119 and #1145

Introduces `NodeUid` as a unique identifier for participants nodes.
```rust
pub struct NodeUid {
    /// Operator account
    pub account_id: AccountId,
    /// TLS public key
    pub tls_public_key: PublicKey,
}
```

Reasonable assumptions about the network state are:

1. Each node operator is uniquely identified by their `AccountId` (existing assumption).
2. A node operator may only have one participating node at any point in time (enforced by the `Participants` struct).
3. A node operator may have multiple nodes submitting an attestation to the contract (enabled by this PR);
4. Each node is uniquely identified by their `tls_public_key` (not really enforceable on the contract).

Note that the following give probabilistic guarantees for point 4, if nodes run inside TEEs:
- a new node must generate a new TLS key upon startup;
- the TLS key is measured and included in the attestation.

Uncovered issues:
- https://github.com/near/mpc/issues/1118
- https://github.com/near/mpc/issues/1109
- https://github.com/near/mpc/issues/1144

### DIFF
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -50,7 +50,7 @@ use primitives::{
 use state::{running::RunningContractState, ProtocolContractState};
 use tee::{
     proposal::MpcDockerImageHash,
-    tee_state::{NodeUid, TeeValidationResult},
+    tee_state::{NodeId, TeeValidationResult},
 };
 
 /// Gas required for a sign request
@@ -739,7 +739,7 @@ impl VersionedMpcContract {
 
         // Add the participant information to the contract state
         let is_new_attestation = mpc_contract.tee_state.add_participant(
-            NodeUid {
+            NodeId {
                 account_id: account_id.clone(),
                 tls_public_key,
             },
@@ -1116,7 +1116,7 @@ impl VersionedMpcContract {
     /// Returns all accounts that have TEE attestations stored in the contract.
     /// Note: This includes both current protocol participants and accounts that may have
     /// submitted TEE information but are not currently part of the active participant set.
-    pub fn get_tee_accounts(&self) -> Vec<NodeUid> {
+    pub fn get_tee_accounts(&self) -> Vec<NodeId> {
         log!("get_tee_accounts: signer={}", env::signer_account_id());
         match self {
             Self::V2(contract) => contract.tee_state.get_tee_accounts(),
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -738,7 +738,7 @@ impl VersionedMpcContract {
         }
 
         // Add the participant information to the contract state
-        mpc_contract.tee_state.add_participant(
+        let is_new_attestation = mpc_contract.tee_state.add_participant(
             NodeUid {
                 account_id: account_id.clone(),
                 tls_public_key,
@@ -748,7 +748,7 @@ impl VersionedMpcContract {
 
         // Both participants and non-participants can propose. Non-participants must pay for the
         // storage they use; participants do not.
-        if self.voter_account().is_err() {
+        if self.voter_account().is_err() || is_new_attestation {
             let storage_used = env::storage_usage() - initial_storage;
             let cost = env::storage_byte_cost().saturating_mul(storage_used as u128);
             let attached = env::attached_deposit();
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -48,7 +48,10 @@ use primitives::{
     thresholds::{Threshold, ThresholdParameters},
 };
 use state::{running::RunningContractState, ProtocolContractState};
-use tee::{proposal::MpcDockerImageHash, tee_state::TeeValidationResult};
+use tee::{
+    proposal::MpcDockerImageHash,
+    tee_state::{NodeUid, TeeValidationResult},
+};
 
 /// Gas required for a sign request
 const GAS_FOR_SIGN_CALL: Gas = Gas::from_tgas(15);
@@ -725,7 +728,7 @@ impl VersionedMpcContract {
             .tee_state
             .verify_proposed_participant_attestation(
                 &proposed_participant_attestation,
-                tls_public_key,
+                tls_public_key.clone(),
                 mpc_contract.config.tee_upgrade_deadline_duration_blocks,
             );
 
@@ -735,9 +738,13 @@ impl VersionedMpcContract {
         }
 
         // Add the participant information to the contract state
-        mpc_contract
-            .tee_state
-            .add_participant(account_id.clone(), proposed_participant_attestation);
+        mpc_contract.tee_state.add_participant(
+            NodeUid {
+                account_id: account_id.clone(),
+                tls_public_key,
+            },
+            proposed_participant_attestation,
+        );
 
         // Both participants and non-participants can propose. Non-participants must pay for the
         // storage they use; participants do not.
@@ -1109,7 +1116,7 @@ impl VersionedMpcContract {
     /// Returns all accounts that have TEE attestations stored in the contract.
     /// Note: This includes both current protocol participants and accounts that may have
     /// submitted TEE information but are not currently part of the active participant set.
-    pub fn get_tee_accounts(&self) -> Vec<AccountId> {
+    pub fn get_tee_accounts(&self) -> Vec<NodeUid> {
         log!("get_tee_accounts: signer={}", env::signer_account_id());
         match self {
             Self::V2(contract) => contract.tee_state.get_tee_accounts(),
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -1184,7 +1184,7 @@ impl VersionedMpcContract {
                     contract.protocol_state = ProtocolContractState::Resharing(resharing);
                 }
 
-                Ok(false)
+                Ok(true)
             }
         }
     }
--- a/crates/contract/src/primitives/participants.rs
+++ b/crates/contract/src/primitives/participants.rs
@@ -1,6 +1,10 @@
 use crate::errors::{Error, InvalidCandidateSet, InvalidParameters};
+
 use near_sdk::{near, AccountId, PublicKey};
 use std::{collections::BTreeSet, fmt::Display};
+
+#[cfg(any(test, feature = "test-utils"))]
+use crate::tee::tee_state::NodeUid;
 
 pub mod hpke {
     pub type PublicKey = [u8; 32];
@@ -198,6 +202,16 @@ impl Participants {
             }
         }
         Err(crate::errors::InvalidState::NotParticipant.into())
+    }
+
+    pub fn get_node_uids(&self) -> BTreeSet<NodeUid> {
+        self.participants()
+            .iter()
+            .map(|(account_id, _, p_info)| NodeUid {
+                account_id: account_id.clone(),
+                tls_public_key: p_info.sign_pk.clone(),
+            })
+            .collect()
     }
 }
 
--- a/crates/contract/src/primitives/participants.rs
+++ b/crates/contract/src/primitives/participants.rs
@@ -4,7 +4,7 @@ use near_sdk::{near, AccountId, PublicKey};
 use std::{collections::BTreeSet, fmt::Display};
 
 #[cfg(any(test, feature = "test-utils"))]
-use crate::tee::tee_state::NodeUid;
+use crate::tee::tee_state::NodeId;
 
 pub mod hpke {
     pub type PublicKey = [u8; 32];
@@ -204,10 +204,10 @@ impl Participants {
         Err(crate::errors::InvalidState::NotParticipant.into())
     }
 
-    pub fn get_node_uids(&self) -> BTreeSet<NodeUid> {
+    pub fn get_node_ids(&self) -> BTreeSet<NodeId> {
         self.participants()
             .iter()
-            .map(|(account_id, _, p_info)| NodeUid {
+            .map(|(account_id, _, p_info)| NodeId {
                 account_id: account_id.clone(),
                 tls_public_key: p_info.sign_pk.clone(),
             })
--- a/crates/contract/src/primitives/participants.rs
+++ b/crates/contract/src/primitives/participants.rs
@@ -130,6 +130,12 @@ impl Participants {
             participants,
         }
     }
+    pub fn info(&self, account_id: &AccountId) -> Option<&ParticipantInfo> {
+        self.participants
+            .iter()
+            .find(|(a_id, _, _)| a_id == account_id)
+            .map(|(_, _, info)| info)
+    }
 }
 
 #[cfg(any(test, feature = "test-utils"))]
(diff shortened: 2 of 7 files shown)


libs/chain-signatures/contract/src/primitives/participants.rs
```diff
@@ -200,6 +203,16 @@ impl Participants {
...
+
+    pub fn get_node_uids(&self) -> BTreeSet<NodeUid> {
+        self.participants()
+            .iter()
+            .map(|(account_id, _, p_info)| NodeUid {
+                account_id: account_id.clone(),
+                tls_public_key: p_info.sign_pk.clone(),
+            })
+            .collect()
+    }
```
DSharifi, line 215: We don't "need" to have this test utility in the source code as part of a `test-utils` feature. Nothing here relies on private attributes or methods.

So we can define this impl extension or as a function in our test utilties, for example in `common.rs`.

libs/chain-signatures/contract/src/tee/tee_state.rs
```diff
@@ -14,6 +14,15 @@ use mpc_primitives::hash::LauncherDockerComposeHash;
 use near_sdk::{env, near, store::IterableMap, AccountId, PublicKey};
 use std::collections::HashSet;
 
+#[near(serializers=[borsh, json])]
+#[derive(Debug, Eq, Ord, PartialEq, PartialOrd, Clone, Hash)]
+pub struct NodeUid {
```
DSharifi, line 19: Struct makes sense. But I do wonder if it's confusing to call this `NodeUid` if it's only an unique identifier only in the context of TEE attestation.

libs/chain-signatures/contract/src/tee/tee_state.rs
```diff
@@ -73,19 +82,19 @@ impl TeeState {
     /// Verifies the TEE quote and Docker image
     pub(crate) fn verify_tee_participant(
         &mut self,
-        account_id: &AccountId,
-        tls_public_key: PublicKey,
```
DSharifi, line 77: nice

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -735,9 +738,13 @@ impl VersionedMpcContract {
...
-        mpc_contract
-            .tee_state
-            .add_participant(account_id.clone(), proposed_participant_attestation);
+        mpc_contract.tee_state.add_participant(
+            NodeUid {
+                account_id: account_id.clone(),
+                tls_public_key,
+            },
+            proposed_participant_attestation,
+        );
```
DSharifi, line 747: Doesn't this open an attack vector where nodes that are participants can flood the contract with large attestations?

Previously the node would only accept one attestation.

libs/chain-signatures/contract/tests/common.rs
```diff
@@ -674,3 +685,34 @@ pub async fn submit_participant_info(
...
+    };
+    Ok(running_state.parameters.participants().clone())
+}
+
+pub async fn submit_tee_attestations(
+    contract: &Contract,
+    env_accounts: &mut [Account],
+    node_uids: &BTreeSet<NodeUid>,
+) -> anyhow::Result<()> {
+    env_accounts.sort_by(|a, b| a.id().cmp(b.id()));
```
DSharifi, line 708: I'd appreciate avoiding variable names like `a` and `b`

libs/chain-signatures/contract/tests/tee.rs
```diff
@@ -333,37 +332,31 @@ async fn test_clean_tee_status_denies_external_account_access() -> Result<()> {
...
+    let additional_uids = additional_participants.get_node_uids();
+    submit_tee_attestations(&contract, &mut additional_accounts, &additional_uids).await?;
 
     // Verify we have TEE data for all accounts before cleanup
     let tee_participants_before = get_tee_accounts(&contract).await?;
     assert_eq!(
-        tee_participants_before.len(),
-        accounts.len() + additional_accounts.len()
+        tee_participants_before,
+        &additional_uids | &participant_uids
```
DSharifi, line 359: Is this a union of the two sets?

libs/chain-signatures/contract/tests/integration_tee_cleanup_after_resharing.rs
```diff
@@ -111,84 +156,55 @@ async fn test_tee_cleanup_after_full_resharing_flow() -> Result<()> {
...
-            .max_gas()
-            .transact()
-            .await?,
-    );
-
-    // Wait for threshold participants to vote for resharing (2 out of 3)
-    // The transition should happen after 2 votes when threshold is reached
-    for account in initial_accounts.iter().take(2) {
+    for domain_id in domain_ids {
+        // Use hardcoded key event ID for test simplicity
```
DSharifi, line 164: stale comment?
```suggestion
        // Use hardcoded key event ID for test simplicity
```

libs/chain-signatures/contract/tests/integration_tee_cleanup_after_resharing.rs
```diff
@@ -29,73 +34,113 @@ use test_utils::attestation::p2p_tls_key;
...
+    let final_participants = assert_running_return_participants(&contract)
+        .await
+        .expect("Expected contract to be in Running state after resharing.");
+
+    // Get current participants to compare
+    let final_participants_node_uids = final_participants.get_node_uids();
+    // Verify only the new participants remain
+    assert_eq!(final_participants_node_uids, expected_tee_post_resharing);
+    // Verify TEE participants are properly cleaned up
+    time::sleep(Duration::from_secs(1)).await;
```
DSharifi, line 126: Relying on a sleep/time for synchronization will cause flakiness.

I assume the test must have also previously already needed it? But I don't see any sleeps from before?

libs/chain-signatures/contract/src/tee/tee_state.rs
netrome: This should be a relatively universal identifier for the node afaik. But I'd also question how important the `U` is in this context as opposed to just calling it `NodeId`.

libs/chain-signatures/contract/tests/common.rs
netrome, line 708: Naming aside, this can be simplified:

```suggestion
    env_accounts.sort_by_key(Account::id)
```

netrome (approved):
Looks good to me. Nice stuff!

libs/chain-signatures/contract/tests/expired_attestation.rs
```diff
@@ -81,168 +21,107 @@ impl TestSetup {
...
-    );
-    // Before clean_tee_status() cleanup, we still have old TEE accounts
-    assert_eq!(setup.contract.get_tee_accounts().len(), PARTICIPANT_COUNT);
-}
-
-/// **Unit test for TEE cleanup of non-participants** - Tests that `clean_tee_status()` removes
-/// TEE accounts for accounts that are no longer in the participant list.
-/// This simulates cleanup after participant removal (e.g., post-resharing).
-#[test]
-fn test_clean_tee_status_removes_non_participants() {
```
kevindeforth, line 182: Can be removed, this is almost a duplicate of the one in `tee.rs`
https://github.com/near/mpc/blob/95649d0982a98942d6249e4d7efaf0758d0155b4/libs/chain-signatures/contract/tests/tee.rs#L335

DSharifi (changes requested):
The tests in `expired_attestation.rs` need to be kept. They are not the same as the one in `tee.rs`

The one in `tee.rs`  tests that the `[private]` attribute is set, I.E. only the contract can call the methods.

The tests in `expired_attestation.rs` were just added today. They test the flow of cleanups after resharing work as a result of resharing through expired attestations.

CC: @Patrick1904  as he has more context as author of these tests

pbeza:
> CC: @Patrick1904 as he has more context as author of these tests

@DSharifi my GitHub username is @pbeza. :)

> The tests in `expired_attestation.rs` need to be kept. They are not the same as the one in `tee.rs`
>
> The one in `tee.rs` tests that the `[private]` attribute is set, I.E. only the contract can call the methods.
>
> The tests in `expired_attestation.rs` were just added today. They test the flow of cleanups after resharing work as a result of resharing through expired attestations.

I agree with @DSharifi. From my perspective, these are two different tests covering TEE cleanup from different angles, as Daniel described, so I’d keep both. I don’t have the full context on why we want to drop one of them—just because they look similar?

kevindeforth:
> The tests in `expired_attestation.rs` need to be kept. They are not the same as the one in `tee.rs`

We keep this one:
```rust
/// **Integration test for participant kickout after expiration** - Tests expired attestation removal. This test demonstrates the complete kickout mechanism using direct contract calls:
/// 1. Initialize contract with 3 secp256k1 participants in Running state at time T=1s
/// 2. Submit valid attestations for first 2 participants at time T=1s
/// 3. Submit expiring attestation for 3rd participant with expiry at time T+10s
/// 4. Fast-forward blockchain time to T+20s using VMContextBuilder
/// 5. Call verify_tee() which detects expired attestation and returns false
/// 6. Contract automatically transitions from Running to Resharing state
/// 7. Start resharing instance and have valid participants vote for completion
/// 8. Contract transitions back to Running state with filtered participant set (2 participants)
/// 9. Verify final state: 2 participants in Running state but 3 TEE accounts remain (cleanup tested separately)
``` 
 https://github.com/near/mpc/blob/95649d0982a98942d6249e4d7efaf0758d0155b4/libs/chain-signatures/contract/tests/expired_attestation.rs#L85
 
 But we remove 
 ```rust
/// **Unit test for TEE cleanup of non-participants** - Tests that `clean_tee_status()` removes
/// TEE accounts for accounts that are no longer in the participant list.
/// This simulates cleanup after participant removal (e.g., post-resharing).
```
https://github.com/near/mpc/blob/95649d0982a98942d6249e4d7efaf0758d0155b4/libs/chain-signatures/contract/tests/expired_attestation.rs#L182

which tests the same thing as 
 ```rust
/// **TEE cleanup functionality** - Tests that the clean_tee_status contract method works correctly when called by the contract itself.
/// Unlike the access control test above, this demonstrates the positive case: the contract can successfully clean up
/// TEE data for accounts that are no longer participants. Uses the test method to populate initial TEE state.
```
https://github.com/near/mpc/blob/95649d0982a98942d6249e4d7efaf0758d0155b4/libs/chain-signatures/contract/tests/tee.rs#L335



> 
> The one in `tee.rs` tests that the `[private]` attribute is set, I.E. only the contract can call the methods.

yes, but that is a different test to the one we are discussing and irrelevant for this PR.
```rust
/// **Access control validation** - Tests that external accounts cannot call the private clean_tee_status contract method.
/// This verifies the security boundary: only the contract itself should be able to perform internal cleanup operations.
```
https://github.com/near/mpc/blob/95649d0982a98942d6249e4d7efaf0758d0155b4/libs/chain-signatures/contract/tests/tee.rs#L303


> The tests in `expired_attestation.rs` were just added today. They test the flow of cleanups after resharing work as a result of resharing through expired attestations.

Not exactly. The tests in `expired_attesation.rs` test that the contract enters a resharing state if `verify_tee` is called while an existing participant has an expired attestation. The mechanism you are describing is tested in `integration_tee_cleanup_after_resharing.rs`

pbeza:
OK, I re-read the implementation of both:

- https://github.com/near/mpc/blob/95649d0982a98942d6249e4d7efaf0758d0155b4/libs/chain-signatures/contract/tests/expired_attestation.rs#L182
- https://github.com/near/mpc/blob/95649d0982a98942d6249e4d7efaf0758d0155b4/libs/chain-signatures/contract/tests/tee.rs#L335

These tests are quite similar conceptually (with some minor differences, like the number of participants and checking that [only the contract can call the `clean_tee_status` method](https://github.com/near/mpc/blob/95649d0982a98942d6249e4d7efaf0758d0155b4/libs/chain-signatures/contract/tests/tee.rs#L369-L375)), but they use different test primitives (`VMContext` vs sandbox), which is already enough reason for me to keep both.  

I’m not sure I follow why there’s so much insistence on removing one of them (sorry, I haven’t spent too much time looking into the context of this PR yet), but from my perspective `test_clean_tee_status_removes_non_participants` is probably less important than `test_participant_kickout_after_expiration`. So if you want to drop that one, I won’t oppose too strongly.

BTW: I think this discussion might never have started if I hadn’t followed [this suggestion](https://github.com/near/mpc/pull/1099#discussion_r2354662273), where I refactored [a single test](https://github.com/near/mpc/blob/68a67ccbb6ba00a356bbab891ff4eb1ea06cb42a/libs/chain-signatures/contract/tests/expired_attestation.rs#L22-L147) into two separate ones—one of which now looks very similar to the one in `tee.rs`.  :P

libs/chain-signatures/contract/tests/expired_attestation.rs
```diff
@@ -81,168 +21,107 @@ impl TestSetup {
...
-    testing_env!(VMContextBuilder::new()
-        .block_timestamp(INITIAL_TIMESTAMP_NANOS)
-        .attached_deposit(NearToken::from_near(1))
-        .build());
-    let domain_id = DomainId::default();
 
-    let mut setup = TestSetup::new(PARTICIPANT_COUNT, THRESHOLD);
-
-    assert_eq!(setup.contract.get_tee_accounts().len(), 0);
+    let (worker, contract, env_accounts, _) = init_env_secp256k1(1).await;
```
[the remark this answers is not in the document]
kevindeforth: Closing this in favor of keeping the `VmContext` framework.
@netrome and I agree that it would be better to either move the existing tests closer to the code they test and there is no reason to have them as integration tests, this is beyond the scope of this PR.
In the interest of moving things along, we "fixed" the unit tests for this PR and opened https://github.com/near/mpc/issues/1144 as a follow-up.

libs/chain-signatures/contract/src/lib.rs
```diff
@@ -1184,7 +1184,7 @@ impl VersionedMpcContract {
                     contract.protocol_state = ProtocolContractState::Resharing(resharing);
                 }
 
-                Ok(false)
+                Ok(true)
```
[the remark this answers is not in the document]
DSharifi: I mean that the return value will never be observed by the node who is responsible for calling this method. So the return value doesn't serve any purpose.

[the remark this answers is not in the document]
kevindeforth: > I mean that the return value will never be observed by the node who is responsible for calling this method. So the return value doesn't serve any purpose.

Hmm, IMO, the fact that the node doesn't track the transaction outcome does not invalidate that our methods return values. They can still be inspected in our testing frameworks and are observable through manual inspection on Nearblocks 

But I do think that it would be much desired for our nodes to track return values as well.

DSharifi (approved):

libs/chain-signatures/contract/tests/expired_attestation.rs
pbeza: Discussed offline: https://nearone.slack.com/archives/C0912BTG51T/p1758033418475589.

Issue #1126, opened 2025-09-16
Transitive dependency `alloy` will not compile


DSharifi:
### Background

Alloy won't compile:
https://github.com/alloy-rs/alloy/issues/2876

A fix has been added to the `dstack` repo in https://github.com/Dstack-TEE/dstack/pull/335, but until th fix is merged and released we need to pin serde.

For more context see:
https://nearone.slack.com/archives/C07UW93JVQ8/p1758059102473569

### Acceptance Criteria

Pin serde to version `1.0.219` so `main` can be compiled.

PR #1129, opened 2025-09-17
chore(deps): bump `dstack-sdk` and remove serde pin

DSharifi:
Follow up to #1126 

`dstack-sdk` has a new version released that bumps their alloy version.


gilcu3 (approved):
Thanks!

Issue #515, opened 2025-06-17
Decouple node and contract through shared interface types


ghost:
### Background
We should look into creating a crate for defining the interface types shared between the node and contract. Right now we return the raw internal representation from the contract, meaning we can't do changes to the state representation without breaking its API's. That way we can keep it fluid on either side without leading to breaking changes on the other side.

This change should also mean that the node code should no longer have a direct dependency to the contract and its types.

### User Story
As an MPC developer I'd like to remove the direct dependency between the node and contract types so that I can modify internal state representations without breaking the API and maintain flexibility in evolving both components independently.

### Acceptance Criteria
The node can no longer depend on the contract, and the contract can no longer return types that are defined in the contract.

Suggested solution:
- Look into using `cargo-deny` on CI to enforce no workspace members depend on `mpc-contract`.
```toml
# deny.toml
[bans]
deny = [
    { name = "mpc-contract" },
]
```
- Move the MPC primitives in `libs/contract/src/primitives` to `mpc/crates/primitives`. 
- Create interface types in a module or new crate in `mpc/crates/` that defines the arguments and return values for all the public methods of the contract.

DSharifi:
Another argument for getting this done, as it's constraining the implementation of the contract:
https://github.com/near/mpc/pull/995#issue-3377704594

PR #1151, opened 2025-09-18
chore: create DTO types for the attestation submission method

DSharifi:


Cargo.toml
```diff
@@ -40,6 +28,7 @@ overflow-checks = true
 [workspace.dependencies]
 #workspace members
 attestation = { path = "crates/attestation" }
+data-transfer-objects = { path = "crates/data_transfer_objects" }
```
netrome, line 31: This crate is dedicated for the contract interface afaik, so I'd like the name of the crate to reflect that. Perhaps `contract-dtos` or `contract-interface-types`. This is basically a first step towards https://github.com/near/mpc/issues/515 as well.

crates/contract/src/data_transfer_object_mapping.rs
```diff
@@ -0,0 +1,157 @@
...
+    attestation::{Attestation, DstackAttestation, MockAttestation},
+    collateral::{Collateral, QuoteCollateralV3},
+    EventLog, TcbInfo,
+};
+use data_transfer_objects::dto_attestation::{
+    DtoAttestation, DtoCollateral, DtoDstackAttestation, DtoEventLog, DtoMockAttestation,
+    DtoTcbInfo,
+};
+
+pub(crate) trait ConvertDtoToContractType {
```
netrome, line 11: Two things:

1. This should be a public trait right?
2. Nit: Naming suggestion: I think `IntoContractType` is more concise and conveys the same information. Open to other suggestions, but having `Dto` as part of the name here sounds weird.

```suggestion
pub trait IntoContractType {
```

crates/data_transfer_objects/src/dto_attestation.rs
```diff
@@ -0,0 +1,187 @@
...
+
+type Sha256Digest = [u8; 32];
+
+#[allow(clippy::large_enum_variant)]
+#[derive(Clone, Debug, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
+#[cfg_attr(
+    all(feature = "abi", not(target_arch = "wasm32")),
+    derive(schemars::JsonSchema)
+)]
+pub enum DtoAttestation {
```
netrome, line 15: Nit: I prefer to skip the `Dto` prefix to the type here. The fact that this type lives in the DTO crate should be sufficient.

```suggestion
pub enum Attestation {
```

crates/data_transfer_objects/src/dto_attestation.rs
```diff
@@ -0,0 +1,187 @@
...
+
+        f.debug_struct("DstackAttestation")
+            .field("quote", &truncate_debug(&self.quote, MAX_BYTES))
+            .field("collateral", &truncate_debug(&self.collateral, MAX_BYTES))
+            .field("tcb_info", &truncate_debug(&self.tcb_info, MAX_BYTES))
+            .finish()
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
```
netrome, line 93: Why don't we derive `Hash`, `PartialEq` etc. here?

netrome (commented):
From an early look this looks good to me. I have some naming suggestions and nits, but the general pattern looks very nice :+1: 

I was a bit torn on the use of associated types over generics in the conversion traits, but for our use case I cannot imagine when the same DTO would map to different types so this does makes things simpler.

crates/contract/src/data_transfer_object_mapping.rs
DSharifi: 1. No, only the contract should have access to these mappings. At some point the contract should be a standalone binary, and not a library we depend from other crates. So everything in that crate should not need to be more public than `pub(crate)`.

2. Good point, that's indeed a better name. https://github.com/near/mpc/pull/1151/commits/cfd7e87e267113fe840fd7ea6eceac00afbd876d

crates/dtos_contract/src/attestation.rs
```diff
@@ -0,0 +1,187 @@
...
+
+        f.debug_struct("DstackAttestation")
+            .field("quote", &truncate_debug(&self.quote, MAX_BYTES))
+            .field("collateral", &truncate_debug(&self.collateral, MAX_BYTES))
+            .field("tcb_info", &truncate_debug(&self.tcb_info, MAX_BYTES))
+            .finish()
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
```
DSharifi: Thought we could add them on a need by basis since it's a workspace member. Would you prefer to add them preemptively?

DSharifi: Also, given that it's a DTO and it's use case is only for the wire, does it make sense to implement these traits?

netrome: I wouldn't rule out the possibility of someone wanting to compare these types or some type containing these. I think C-COMMON-TRAITS still apply to these https://rust-lang.github.io/api-guidelines/interoperability.html?highlight=eager#types-eagerly-implement-common-traits-c-common-traits

Cargo.toml
```diff
@@ -1,19 +1,7 @@
...
-    "crates/tee_authority", 
-    "crates/test_utils", 
-    "crates/tls",
-    "crates/contract_history",
-    "crates/test_migration_contract",
-    "crates/test_parallel_contract",
-    "crates/devnet",
-    "crates/node",
-]
+members = ["crates/*"]
```
netrome, line 4: I prefer having these explicitly spelled out. Is this inconvenient in any way?

When using `cargo new` it typically auto-adds these anyway.

netrome (approved):
Nice stuff!

crates/dtos_contract/src/attestation.rs
DSharifi: Good point. Added these common traits except for `Display`, `Copy`, and `Default`.

Issue #1105, opened 2025-09-14
dstack reproducible build


barakeinav1:
### Background

We need to create or (use existing) a reproducible dsack OS image. since this is measured inside the contract RTMRs0-2.

### User Story

I'd like to be be able to link between the binary's running in the CVM to source code.

### Acceptance Criteria

a recipe for deterministically reproducible build is create
exact commits hash are define for use.
RTMR measurements are  retrieved.

barakeinav1:
using stack 0.54

MRTD: f06dfda6dce1cf904d4e2bab1dc370634cf95cefa2ceb2de2eee127c9382698090d7a4a13e14c536ec6c9c3c8fa87077
RTMR0: 3744b154069500a466f514253b49858299b2e1bdc44e3d557337d81e828bedf6a0410f27d3a18c932e5e49e1c4215737
RTMR1: a7b523278d4f914ee8df0ec80cd1c3d498cbf1152b0c5eaf65bad9425072874a3fcf891e8b01713d3d9937e3e0d26c15
RTMR2: 24847f5c5a2360d030bc4f7b8577ce32e87c4d051452c937e91220cab69542daef83433947c492b9c201182fc9769bbe

netrome:
(Or at least link to it, as Phala will also host these images)

barakeinav1:
Machine measurements:
MRTD: f06dfda6dce1cf904d4e2bab1dc370634cf95cefa2ceb2de2eee127c9382698090d7a4a13e14c536ec6c9c3c8fa87077
RTMR0: e673be2f70beefb70b48a6109eed4715d7270d4683b3bf356fa25fafbf1aa76e39e9127e6e688ccda98bdab1d4d47f46
RTMR1: a7b523278d4f914ee8df0ec80cd1c3d498cbf1152b0c5eaf65bad9425072874a3fcf891e8b01713d3d9937e3e0d26c15
RTMR2: 24847f5c5a2360d030bc4f7b8577ce32e87c4d051452c937e91220cab69542daef83433947c492b9c201182fc9769bbe

PR #1154, opened 2025-09-21
docs: Update RTMR generation instructions for dstack 0.5.4

barakeinav1:
fixes [1105](https://github.com/near/mpc/issues/1105)

### DIFF
--- a/docs/running_an_mpc_node_in_tdx_external_guide.md
+++ b/docs/running_an_mpc_node_in_tdx_external_guide.md
@@ -79,29 +79,29 @@ Make sure to complete:
 This ensures that your TDX setup is correctly configured.  
 
 ---
-
 #### 2. Dstack Setup and Configuration
-
-##### Dstack Setup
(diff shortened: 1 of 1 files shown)


docs/running_an_mpc_node_in_tdx_external_guide.md
```diff
@@ -199,24 +194,72 @@ echo "  Hash offset:   $HASH_OFFSET"
...
+docker build . -t dstack-mr
+```
+
+Run:
+```bash
+docker run --rm \
+  -v "$(pwd)":/dstack-0.5.4 \
+  dstack-mr \
+  measure -c 8 -m 64G /dstack-0.5.4/metadata.json
+```
```
netrome, line 252: So this is just to build the `dstack-mr` tool which we use to verify measurements? Is there any reason why we need to build this as opposed to having phala publish some official binary that could be verified and installed?

PR #1166, opened 2025-09-22
chore: use time durations instead of block numbers for tee grace period

### LINKED ISSUE #1160: Use time durations instead of block numbers for TEE grace period
### Background

Currently, the grace period for TEE upgrades in the MPC contract is defined in terms of block heights (e.g., number of blocks approximating 7 days). This approach assumes a stable block production rate, but block times can vary due to network incidents, protocol changes, or intentional tuning (e.g., reducing block time). As a result, block height is not a reliable proxy for elapsed time.

Additionally, upgrade deadlines are intended for node operators, who think in terms of human-readable timeframes (e.g., "7 days" or "September 9, 2025, 15:55 UTC"), not block counts (e.g., "1,000,000 blocks"). Communicating deadlines in block numbers adds unnecessary friction and requires manual or error-prone conversion to time.

Since each finalized block already includes a monotonically increasing timestamp (with at most 2 minutes deviation), using timestamps and durations directly would allow us to enforce deadlines more accurately and provide a better user experience.

### User Story

As an MPC node operator, I want upgrade deadlines to be defined and enforced in terms of human-readable time durations so that I can clearly understand when upgrades must be completed by.

### Acceptance Criteria

Change the TEE ugprade period to use durations over blocks in the contract.

https://nearone.slack.com/archives/C07UW93JVQ8/p1756735190432329

DSharifi:
closes #1160 

The changes to the node code come as a side effect of the node (wrongly) depending on the `AllowedMpcDockerImage` struct. This assumption is wrong for two reasons:
1. `AllowedMpcDockerImage` is not returned on the contract method which queries for allowed image hashses.
2. `AllowedMpcDockerImage`is an internal structure of the contract.

Other bug fixes:
- https://github.com/near/mpc/blob/main/crates/contract/src/tee/proposal.rs#L71-L78. The grace period should be defined as 1 week from the time when the new hash is voted in, but the implementation defines it as 1 week from the time the old hash itself was added.

- (Minor, should never be possible to reproduce in prod); If two hashes are voted in at the same time stamp, I'd expect the last hash to take precedence as the latest allowed image hash. Currently the first one added takes precedence if both have the same time stamps/block number.


crates/contract/src/tee/proposal.rs
```diff
@@ -50,70 +50,60 @@ impl CodeHashesVotes {
...
             })
-            .unwrap_or(self.allowed_tee_proposals.len());
-
-        // Never remove all proposals; always keep at least one (the latest)
-        let expired_count = expired_count.min(self.allowed_tee_proposals.len().saturating_sub(1));
-
-        self.allowed_tee_proposals.drain(0..expired_count);
+            .map(|(_i, entry)| entry)
+            .cloned()
+            .collect()
```
netrome, line 89: This looks incorrect. Why do we return the current entry if the successor is valid?

If the logic is "skip the latest entry, and check the grace period of the rest" it should instead be something like:

```suggestion
            .rev()
            .skip(1)
            .filter(|entry| {
                let grace_period_deadline = entry.added + tee_upgrade_deadline_duration;

                current_time <= grace_period_deadline
            })
            .cloned()
            .collect()
```

Note that I'm returning in reverse order here. I don't think order matters, and it seems more sensible to return the most recent one first anyway.

Let me know if I'm missing something here. Otherwise it would also be good to add test coverage for this.

DSharifi: This is precisely the bug that is being removed. Here you are defining the expiration deadline as the `tee_upgrade_deadline_duration` plus the entry time of the potentially stale hash itself, not the entry time of its successor which is what should be starting the ticking timer to invalidate it.

DSharifi: > This looks incorrect. Why do we return the current entry if the successor is valid?

Isn't that the whole point of having a grace period?

pbeza (approved):

PR #1180, opened 2025-09-24
chore: leader info on debug page

andrei-near:


node/src/signing/debug.rs
```diff
@@ -105,7 +107,14 @@ impl Debug for CompletedSignatureRequest {
             &format!("{:?}", self.request.id)[0..6],
             format!("{:?}", self.request.receipt_id),
             self.progress.lock().unwrap().attempts,
-        )
+        )?;
+
+        // Add leader election information - show only the elected leader
+        if let Some(elected_leader) = self.leader_selection_order.last() {
```
gilcu3, line 113: I am not sure if the last element of the vector is the selected leader. If I remember correctly, it was the first, and then in case of failure the second and so on...

andrei-near: it's the last one based on my observations

Issue #415, opened 2025-05-14
Improved backwards compatibility testing


kevindeforth:
With #408 migration testing without depending on the mpc-nodes was enabled.

These tests can be enhanced by:
- [ ] fetching the latest breaking change from git, once #412 has been addressed
- [ ] adding state to the test (i.e. add some update proposals, parameter proposals, signature requests etc.).

kuksag:
@kevindeforth 👍 

> adding state to the test

During https://github.com/Near-One/mpc/pull/414 I realized that populating a contract with some state is an important task. If some more complex state is provided, migration might fail. 
I think [spooning](https://docs.near.org/smart-contracts/testing/integration-test#spooning-contracts) the contract is might be the right way.

> fetching the latest breaking change from git

...Which led to the idea why even care about when the last breaking change occurred, when in fact we care about specific contracts: (a) contract on testnet (b) contract on mainnet. Why not fetch them directly from the explorer or via the other way? the pitfall might be that it will increase time by a bit (loading 500kb from the web), but I think it might be optimized.

kevindeforth:
> ...Which led to the idea why even care about when the last breaking change occurred, when in fact we care about specific contracts: (a) contract on testnet (b) contract on mainnet. Why not fetch them directly from the explorer or via the other way? the pitfall might be that it will increase time by a bit (loading 500kb from the web), but I think it might be optimized.

That is a good point. On devnet, the default behavior since #401 is to fetch the contract from testnet.

We could probably do something similar for pytests (and thus ci), by querying an RPC node.

PR #1186, opened 2025-09-25
test: rewrite upgrade test as a sandbox integration tests

### LINKED ISSUE #1185: Rewrite `test_update_to_current` as a sandbox integration tests
### Background

The pytest `test_update_to_current` can be removed and written as a rust integration test for the contract crate instead.

https://github.com/near/mpc/blob/3a5a7f3a3dd3593ffb14a75edc243c09d674c326/pytest/tests/test_contract_update.py#L41

This test is currently blocking #1069, as it's failign on CI for #1174. The reason is that this test assumes that new version of the nodes are backwards compatible with the old contract binary, which is no the case.

The test expectation is wrong, but we would lose coverage by just deleting it ([slack message](https://nearone.slack.com/archives/C0912BTG51T/p1758803719162309?thread_ts=1758653924.725649&cid=C0912BTG51T)):
>But it also tests the migration logic on the contract (e2e test for proposing and voting for a contract upgrade).
Some of that is handled on the integration tests in the contract folder, but the pytests cover, afaik, more behavior (at least, they add more state to the contract than what is currently done in the integration tests).

We can delete the test and add the coverage it has as a sandbox test.

### User Story

_No response_

### Acceptance Criteria

- `test_update_to_current` pytest is deleted
- Test coverage from `test_update_to_current` is added as a sandbox integration test.

https://nearone.slack.com/archives/C0912BTG51T/p1758802334671489?thread_ts=1758653924.725649&cid=C0912BTG51T

DSharifi:
closes #1185


crates/contract/tests/sandbox/common.rs
```diff
@@ -725,3 +726,95 @@ pub async fn submit_tee_attestations(
...
+        let state: ProtocolContractState = contract.view("state").await.unwrap().json().unwrap();
+        match state {
+            ProtocolContractState::Initializing(state) => {
+                assert_eq!(state.domains.domains().len(), domain_attempt + 1);
+            }
+            _ => panic!("should be in initializing state"),
+        };
+
+        check_call_success(
+            accounts[0]
```
kevindeforth, line 759: For this to succeed, we need `accounts[0]` to be the participant with the lowest `ParticipantId`.
There should be a function comment stating this.

crates/contract/tests/sandbox/contract_upgrade.rs
```diff
@@ -106,12 +121,59 @@ async fn back_compatibility(network: Network) -> anyhow::Result<()> {
...
-    back_compatibility(Network::Mainnet).await.unwrap();
-}
+async fn upgrade_keeps_participants_and_domains_intact(
+    #[values(Network::Mainnet, Network::Testnet)] network: Network,
+) {
+    const EPOCH_ID: u64 = 0;
+
+    let worker = near_workspaces::sandbox().await.unwrap();
+    let contract = deploy_old(&worker, network).await.unwrap();
+    let (accounts, _participants) = init_contract(worker, &contract).await.unwrap();
```
kevindeforth, line 141: As an alternative, we could simplify the logic and use `init_running`.

crates/contract/tests/sandbox/contract_upgrade.rs
```diff
@@ -106,12 +121,59 @@ async fn back_compatibility(network: Network) -> anyhow::Result<()> {
...
+
+    // Add the domains above to the contract so we have additional state on the contract
+    // that should be persisted after the upgrade
+    call_contract_key_generation(domains_to_add, &accounts, &contract, EPOCH_ID).await;
+
+    let state_pre_upgrade: ProtocolContractState =
+        contract.view("state").await.unwrap().json().unwrap();
+
+    assert!(healthcheck(&contract).await.unwrap());
+
```
kevindeforth, line 166: I think it would be good to add some more state. In the pytest, we added parameter proposals:
(https://github.com/near/mpc/blob/3d21df4e24146b2b28afbb729d5535151e7f74b7/pytest/tests/test_contract_update.py#L51-L57) 

It wouldn't be too bad to additionally add a bunch of signature requests and one or two other contract upgrades

kevindeforth (commented):
Thank you.
I would like to some more state added before the migration call (at least what we have in the pytest we are about to delete, but even better if we were to resolve https://github.com/near/mpc/issues/415 with this PR)

crates/contract/tests/sandbox/contract_upgrade.rs
```diff
@@ -106,12 +121,59 @@ async fn back_compatibility(network: Network) -> anyhow::Result<()> {
...
+    let contract = upgrade_to_new(contract).await.unwrap();
+    migrate(&contract)
+        .await
+        .expect("❌ Back compatibility check failed: migration() failed");
```
kevindeforth, line 170: here, can you add a check comparing the actual vs expected hash of the contract?

kevindeforth (changes requested):
I think there are still a few details to work out s.t. we don't lose test coverage compared to the pytests.

crates/contract/tests/sandbox/contract_upgrade.rs
DSharifi: I'm not sure I understand. By hash do you mean the hash of the compiled binary? And what is the expected hash?

crates/contract/tests/sandbox/common.rs
```diff
@@ -725,3 +726,95 @@ pub async fn submit_tee_attestations(
...
+                .transact()
+                .await
+                .unwrap(),
+        );
+
+        println!("start_keygen_instance completed");
+
+        let public_key = match domain.scheme {
+            SignatureScheme::Secp256k1 => {
+                let dummy_bytes = [*domain.id as u8; 64];
```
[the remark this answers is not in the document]
DSharifi: Changed to use a proper random key

crates/contract/tests/sandbox/contract_upgrade.rs
kevindeforth: Exactly :+1:
We can compute the expected hash, since have the code we want to deploy.

DSharifi: Ah I see. So you want a sanity check that the returned hash is what we upgraded to?

DSharifi: Added

1e4766d2b1ffdd7629697e8aa7c5d8c987887bce

crates/contract/tests/sandbox/contract_upgrade.rs
DSharifi: Also added proposal update:
1d4b03b17e906c3f8ed9fe9f52b2ac8d507f056c

crates/contract/tests/sandbox/contract_upgrade.rs
```diff
@@ -106,12 +121,59 @@ async fn back_compatibility(network: Network) -> anyhow::Result<()> {
...
+    // Add the domains above to the contract so we have additional state on the contract
+    // that should be persisted after the upgrade
+    call_contract_key_generation(domains_to_add, &accounts, &contract, EPOCH_ID).await;
+
+    let state_pre_upgrade: ProtocolContractState =
+        contract.view("state").await.unwrap().json().unwrap();
+
+    assert!(healthcheck(&contract).await.unwrap());
+
+    let contract = upgrade_to_new(contract).await.unwrap();
```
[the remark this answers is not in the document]
DSharifi: Fixed in https://github.com/near/mpc/pull/1186/commits/898d39df004f84fd83f12cede3abe1c86348df64

crates/contract/tests/sandbox/updates.rs
```diff
@@ -142,32 +137,7 @@ async fn test_propose_update_config() {
 #[tokio::test]
 async fn test_propose_update_contract() {
```
kevindeforth, line 138: so, this test is not exactly a drop-in replacement for the deleted pytest
https://github.com/near/mpc/blob/e224461bcd1259fd213f370fbfac1b009558c19f/pytest/tests/test_contract_update.py#L29-L37

Compared to the deleted test:
- we don't have any state when we do the upgrade
- we don't change the contract code to which we want to upgrade. That means that if for some reason, the update fails throwing an error message, the state method call and the code hash check in `propose_and_vote_contract_to_current_binary` will pass even if the migration failed.

That is why, in the pytest, we deploy the [test_migration_contract](https://github.com/near/mpc/blob/e7b631f59acbb1f8834b4ac15df6a676ddd2327c/crates/test_migration_contract/src/lib.rs)

crates/contract/tests/sandbox/sign.rs
```diff
@@ -30,8 +31,14 @@ async fn test_contract_sign_request() -> anyhow::Result<()> {
 
     for msg in messages {
         println!("submitting: {msg}");
-        let (payload, respond_req, respond_resp) =
-            create_response(predecessor_id, msg, path, &sks[0]).await;
+        let (payload, respond_req, respond_resp) = create_message_payload_and_response(
+            DomainId::legacy_ecdsa_id(),
```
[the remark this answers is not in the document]
DSharifi: going with first option as that is what we had from before. Tracking issue to resolve https://github.com/near/mpc/issues/1194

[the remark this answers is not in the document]
DSharifi: b23d8e8c3184112d0faf9b5d499b30e6fd1c5a9a

crates/contract/tests/sandbox/common.rs
```diff
@@ -499,20 +515,40 @@ pub async fn sign_and_validate(
         .await?;
     dbg!(&status);
 
+    Ok(status)
+}
+
+pub async fn respond_to_sign_request(
```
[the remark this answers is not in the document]
DSharifi: d060f765e1aaa2c3c94d665b4b32c349312f02c0

crates/contract/tests/sandbox/common.rs
```diff
@@ -75,6 +77,13 @@ pub const PARTICIPANT_LEN: usize = 3;
 /// unexpectedly in the future.
 pub const GAS_FOR_VOTE_RESHARED: Gas = Gas::from_tgas(15);
 
+/// This is the current deposit required for a contract deploy. This is subject to change but make
+/// sure that it's not larger than 2mb. We can go up to 4mb technically but our contract should
+/// not be getting that big.
+///
+/// TODO(#771): Reduce this to the minimal value possible after #770 is resolved
+pub const CURRENT_CONTRACT_DEPLOY_DEPOSIT: NearToken = NearToken::from_millinear(13333);
```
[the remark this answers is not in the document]
DSharifi: Changed to pytest value. Although I don't see why we care about this value as long as it's reasonable:

d08f0e82a5a824c3d3f8e4d322861e768366d7a6

crates/contract/tests/sandbox/contract_upgrade.rs
DSharifi: I don't see it immediately how to simplify it.

Could you elaborate on how we can use `init_running` to simplify the logic?

crates/contract/tests/sandbox/common.rs
```diff
@@ -621,6 +657,61 @@ pub async fn derive_confidential_key_and_validate(
...
+        .await
+        .expect("state request succeeds");
+
+    let _state: ProtocolContractState = state_request_execution
+        .json()
+        .expect("state is deserializable.");
+
+    vote_update_till_completion(contract, accounts, &proposal_id).await;
+
+    let code_hash_post_upgrade = contract.view_code().await.unwrap();
```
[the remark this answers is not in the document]
DSharifi: You're right. Good catch

1c494e954ab8641e35adf7f90417ba33389b006a

crates/contract/tests/sandbox/common.rs
```diff
@@ -725,3 +816,106 @@ pub async fn submit_tee_attestations(
...
+                    .await
+                    .unwrap(),
+            );
+        }
+
+        let state: ProtocolContractState = contract.view("state").await.unwrap().json().unwrap();
+        match state {
+            ProtocolContractState::Running(state) => {
+                assert_eq!(state.keyset.epoch_id.get(), expected_epoch_id);
+                assert_eq!(state.domains.domains().len(), domain_attempt + 1);
```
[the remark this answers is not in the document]
DSharifi: To not slow down all tests that will depend on `propose_and_vote_contract_update_to_current_binary`, how about we have an isolated test which tests this in isolation?

Feel free to create a tracking issue.

crates/contract/tests/sandbox/upgrade_to_current_contract.rs
```diff
@@ -0,0 +1,303 @@
...
+    }
+}
+
+struct PendingSignRequest {
+    transaction: TransactionStatus,
+    signature_request: SignatureRequest,
+    signature_response: SignatureResponse,
+}
+
+async fn init_contract(
```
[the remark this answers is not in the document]
DSharifi: 01b8f25ad826838192535b72bb6f15e5a803a04f

crates/contract/tests/sandbox/common.rs
```diff
@@ -725,3 +816,106 @@ pub async fn submit_tee_attestations(
...
+            SignatureScheme::Ed25519 => {
+                let (public_key, private_key) = new_ed25519();
+                let shared_secret_key = SharedSecretKey::Ed25519(private_key);
+
+                (shared_secret_key, public_key)
+            }
+            SignatureScheme::CkdSecp256k1 => {
+                unimplemented!("Old contract does not support CKD.")
+            }
+        };
```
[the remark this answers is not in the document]
DSharifi: Thanks, that's a good spot.

4f5c7140d5c3e85ae8bd5562398272d44e933aea

crates/contract/tests/sandbox/upgrade_to_current_contract.rs
```diff
@@ -0,0 +1,303 @@
...
+
+/// Checks the contract in the following order:
+/// 1. Are there any state-breaking changes?
+/// 2. If so, does `migrate()` still work correctly?
+///
+/// These checks use the previous contract version (the one that introduced breaking changes)
+/// as a baseline. If step 2 fails, you will be prompted to update the baseline contract.
+#[rstest]
+#[tokio::test]
+async fn back_compatibility_without_state(
```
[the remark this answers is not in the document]
DSharifi: Let's keep it. It's a much more isolated test, and a good signal to have if it fails.

crates/contract/tests/sandbox/upgrade_to_current_contract.rs
```diff
@@ -0,0 +1,303 @@
...
+        .max_gas()
+        .transact()
+        .await
+        .unwrap()
+        .unwrap();
+
+    // Add the domains above to the contract so we have additional state on the contract
+    // that should be persisted after the upgrade
+    let shared_secret_keys =
+        call_contract_key_generation(&domains_to_add, &accounts, &contract, EPOCH_ID).await;
```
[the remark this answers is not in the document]
DSharifi: Let's keep it as is. `init_running` is currently non backwards compatible as well since we changed the `inig_config` parameter.

Feel free to create a follow up issue if you want to separate functions for the old and current version of the contract.

crates/contract/tests/sandbox/upgrade_from_current_contract.rs
```diff
@@ -142,32 +137,7 @@ async fn test_propose_update_config() {
 #[tokio::test]
 async fn test_propose_update_contract() {
```
DSharifi: Good point.

92099175b4625b79fec94c36cc4dcf63914ac1da

crates/contract/tests/sandbox/contract_upgrade.rs
kevindeforth: `init_running` allows to initialize the contract with a keyset, so we don't need to go through the key generation procedure.

This allows us to narrow the scope of tests. A failure in the key generation logic only means _"running the key generation workflow with the old contract failed"_ - that's not helpful in asserting whether or not we can successfully upgrade the contract code.

But it's not the end of the world if we do keep the current workflow.

PR #1192, opened 2025-09-26
metrics tracking participants in failed comps

kevindeforth:


netrome:
Context: https://github.com/near/mpc-private/issues/28

This PR creates more visibility on which participants took part of failed computations.

node/src/providers/ecdsa/sign.rs
```diff
@@ -45,12 +46,25 @@ impl EcdsaSignatureProvider {
...
+            Ok(res) => res,
+            e => {
+                participants.iter().for_each(|id| {
+                    metrics::PARTICIPANT_TOTAL_TIMES_SEEN_IN_FAILED_COMPUTATION
+                        .with_label_values(&[&id.raw().to_string()])
+                        .inc();
+                });
+                return e;
+            }
+        };
```
netrome, line 67: Any reason we only track this for ECDSA signatures when we are the leader? I'd assume it would be nice to track computations we're participating in as follower as well, and also EDDSA signatures. Or was this intentional for any reason?

andrei-near (approved):
This can be merged. we'd also need this in main branch

node/src/providers/ecdsa/sign.rs
netrome: Thanks!

PR #1193, opened 2025-09-26
fix: restore exact workflow before #1113 not including MPC launcher

gilcu3:
Follow up to https://github.com/near/mpc/pull/1157

### DIFF
--- a/.github/workflows/docker_release_legacy.yml
+++ b/.github/workflows/docker_release_legacy.yml
@@ -33,6 +33,13 @@ jobs:
         run: |
           echo "sha_short=$(git rev-parse --short HEAD)" >> "$GITHUB_ENV"
   
+      - name: Sanitize build-ref for Docker tags
+        shell: bash
+        run: |
+          # Replace forward slashes with hyphens to create valid Docker tags
+          sanitized_ref=$(echo "${{ github.event.inputs.build-ref }}" | sed 's/\//-/g')
+          echo "sanitized_ref=$sanitized_ref" >> "$GITHUB_ENV"
+
       - name: Initialize submodules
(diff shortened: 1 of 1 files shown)


.github/workflows/docker_release_legacy.yml
```diff
@@ -41,12 +48,12 @@ jobs:
...
+      - name: Build and push MPC Node TEE Docker image to Docker Hub
         uses: Warpbuilds/build-push-action@v6
         with:
           context: .
           profile-name: "mpc-image-builder"
           push: true
-          file: deployment/Dockerfile-gcp
-          tags: nearone/mpc-node-gcp:latest,nearone/mpc-node-gcp:${{ github.event.inputs.build-ref }},nearone/mpc-node-gcp:${{ github.event.inputs.build-ref }}-${{ env.sha_short }}
\ No newline at end of file
+          file: deployment/Dockerfile-mpc-tee
```
andrei-near, line 58: legacy docker images have no tee. this should not be changed

.github/workflows/docker_release_legacy.yml
```diff
@@ -41,12 +48,12 @@ jobs:
...
+          tags: nearone/mpc-node-tee:latest,nearone/mpc-node-tee:${{ env.sanitized_ref }},nearone/mpc-node-tee:${{ env.sanitized_ref }}-${{ env.sha_short }}
```
andrei-near, line 59: ditto

andrei-near (changes requested):

gilcu3:
> Can you you test it to make sure it works?

Sure, do you know exactly what file was the last one working?

andrei-near (approved):

PR #1201, opened 2025-09-28
chore(node): use one key pair for both TLS AND near access key

DSharifi:
Proof of concept PR to
- merge the two key pairs generated on the ndoe, TLS and Account key, into one key: https://nearone.slack.com/archives/C07UW93JVQ8/p1759043700318139
- remove`NodeId`struct and its usage as a unique identifier for attestations, and only rely on the nodes' public keys as identifiers as they are assumed to be unique: 

Remaining work:
- Pytests and integration tests will also have to be updated and simplified to not rely on `NodeId` struct.
- Pytests must be simplified to only fetch a single key from the web endpoint.

### DIFF
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -23,7 +23,10 @@ use crate::{
     errors::{Error, RequestError},
     primitives::ckd::{CKDRequest, CKDRequestArgs},
     storage_keys::StorageKey,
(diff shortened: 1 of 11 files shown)


gilcu3:
Could we close this (given decided not to use it a few weeks ago)? (just to clean-up the PR list page)

PR #1202, opened 2025-09-29
fix: Resubmit participant info more frequently

pbeza:


DSharifi (changes requested):
Looks great overall!

One hard blocker: the attestation submission logic shouldn’t live in the `coordinator` + `mpc_client` since those modules should stay focused on the MPC protocol.

Could you move the periodic re-submission into `tee/remote_attestation.rs` instead? You can implement it as a looping future there, and then explicitly spawn it in `cli.rs` right after the initial attestation is submitted.

crates/node/src/indexer/fake.rs
```diff
@@ -487,9 +487,7 @@ impl FakeIndexerCore {
                     }
                     ChainSendTransactionRequest::VerifyTee() => {}
                     ChainSendTransactionRequest::SubmitParticipantInfo(_participant_info) => {
-                        unimplemented!(
-                            "Submitting participant info is not implemented for tests yet."
-                        )
+                        // TODO(#1203): Submitting participant info is not implemented for tests yet.
```
DSharifi, line 490: Thanks for creating the issue.

I think the reason you saw this panic first now is that the code path in `cli.rs` is not executed in some of our integration tests, whereas the current placement for re-submissions in `coordinator.rs` is run.

crates/node/src/mpc_client.rs
```diff
@@ -130,6 +138,41 @@ impl MpcClient {
...
+                                chain_txn_sender.clone(),
+                                fresh_attestation,
+                                tls_public_key,
+                            )
+                            .await
+                            {
+                                tracing::error!(
+                                    "Failed to submit fresh remote attestation: {:?}",
+                                    e
+                                );
```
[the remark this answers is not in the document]
pbeza: Fixed ✅
- https://github.com/near/mpc/pull/1202/commits/fa7a568e3e7666e4f55da1907187cb664ef1044a
- https://github.com/near/mpc/pull/1202/commits/4a6f90b77203eb1980b583fbcc3efe026a66e454

crates/node/src/tee/remote_attestation.rs
```diff
@@ -80,3 +81,44 @@ pub async fn submit_remote_attestation(
         .await
         .context("failed to submit attestation after multiple retry attempts")?
 }
+
+/// Periodically generates and submits fresh attestations at regular intervals.
+///
+/// This function runs indefinitely, generating a fresh attestation every 10 minutes
```
DSharifi, line 87:
```suggestion
/// This future runs indefinitely, generating a fresh attestation every 10 minutes
```

crates/node/src/tee/remote_attestation.rs
```diff
@@ -80,3 +81,44 @@ pub async fn submit_remote_attestation(
...
+/// and submitting it to the blockchain. It's designed to be spawned as a background task.
```
DSharifi, line 88: Nit:
```suggestion
/// and submitting it to the blockchain.
```

crates/node/src/tee/remote_attestation.rs
```diff
@@ -80,3 +81,44 @@ pub async fn submit_remote_attestation(
...
+pub async fn periodic_attestation_submission<T: TransactionSender + Clone>(
+    tee_authority: TeeAuthority,
+    tx_sender: T,
+    tls_sdk_public_key: near_sdk::PublicKey,
+    tls_public_key: VerifyingKey,
```
DSharifi, line 93:
```suggestion
    tls_public_key: VerifyingKey,
```

You can create the SDK key version from the `VerifyingKey` type. No need to request both arguments from the caller, when the underlying key is the same with differing formats.

crates/node/src/tee/remote_attestation.rs
```diff
@@ -80,3 +81,44 @@ pub async fn submit_remote_attestation(
...
+) -> anyhow::Result<()> {
+    const ATTESTATION_RESUBMISSION_INTERVAL: Duration = Duration::from_secs(10 * 60);
```
DSharifi, line 95: Could you place this constant to the top of the file alongside the existing constants?

crates/node/src/tee/remote_attestation.rs
```diff
@@ -80,3 +81,44 @@ pub async fn submit_remote_attestation(
...
+
+    loop {
+        interval.tick().await;
+
+        let report_data = ReportData::new(tls_sdk_public_key.clone());
+        let fresh_attestation = match tee_authority.generate_attestation(report_data).await {
+            Ok(attestation) => attestation,
+            Err(e) => {
+                tracing::error!(
+                    error = ?e,
```
DSharifi, line 106: tip:
```suggestion
            Err(error) => {
                tracing::error!(
                    ?error,
```

netrome (changes requested):
I don't see any tests for this. While the logic is fairly simple and looks correct, I think we should strive to test any new functionality we add.

Unless I'm missing something, it should be possible to add a unit test for the `periodic_attestation_submission` without changing too much code. Tokio time can already be controlled in tests, you just need to make sure the test starts with frozen time and explicitly advance the time to avoid it being flaky (also beware of Tokio [auto-advancing time](https://docs.rs/tokio/latest/tokio/time/fn.advance.html#auto-advance)).

With that said, I typically prefer to rely on dependency injection for time-related functionality (e.g. creating a trait for the interval and implementing that over the tokio::time::Interval type) which I think would be nice here as well.

The tx_sender already provides an interface that could be used to catch and verify the submitted attestation.

Beyond this, it would be nice to test this on a system level but given our challenges managing time on that level we might have to let that pass for now.

What do you think?

crates/node/src/cli.rs
```diff
@@ -383,6 +384,26 @@ impl StartCmd {
         submit_remote_attestation(indexer_api.txn_sender.clone(), attestation, tls_public_key)
             .await?;
 
+        // Spawn periodic attestation submission task
+        let tls_sdk_public_key = tls_public_key.to_near_sdk_public_key()?;
+        let tee_authority_clone = TeeAuthority::try_from(self.tee_authority.clone())?;
```
DSharifi, line 389: Instead of this fallible `try_from` logic, have already constructed a `tee_authority` on line 242 above in `run()`. Let's reuse that tee_authority instead of reconstructing it which is fallible.

DSharifi (approved):

crates/node/src/tee/remote_attestation.rs
pbeza: Good point, not sure why I didn’t catch that myself. Anyway, fixed here: https://github.com/near/mpc/pull/1202/commits/f4497f34067290b5793d2169f19dce072bab9827. ✅

crates/node/src/tee/remote_attestation.rs
pbeza: Nice! Fixed here: https://github.com/near/mpc/pull/1202/commits/f4497f34067290b5793d2169f19dce072bab9827. ✅

crates/node/src/tee/remote_attestation.rs
```diff
@@ -80,3 +81,44 @@ pub async fn submit_remote_attestation(
...
+/// Periodically generates and submits fresh attestations at regular intervals.
+///
+/// This function runs indefinitely, generating a fresh attestation every 10 minutes
+/// and submitting it to the blockchain. It's designed to be spawned as a background task.
+pub async fn periodic_attestation_submission<T: TransactionSender + Clone>(
+    tee_authority: TeeAuthority,
+    tx_sender: T,
+    tls_sdk_public_key: near_sdk::PublicKey,
+    tls_public_key: VerifyingKey,
+) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
pbeza: Yup, good point — it used to be fallible, but [now](https://github.com/near/mpc/commit/f4497f34067290b5793d2169f19dce072bab9827#diff-454ca9f9a7e8ff74a84ffe709988d6f6ac448b7646b0c236b42729929a6c78c8R101), after addressing [this comment](https://github.com/near/mpc/pull/1202#discussion_r2398421085), it has become infallible.

crates/node/src/tee/remote_attestation.rs
```diff
@@ -1,19 +1,46 @@
...
+        }
+    }
+}
+
+impl IntervalTicker for TokioIntervalTicker {
+    async fn tick(&mut self) {
+        self.interval.tick().await;
+    }
+}
+
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/pull/1202/commits/7ab95065d83179256672286086ac67644b5d3492. 🫡

PR #1216, opened 2025-10-01
feat: import keyshares into non empty keyshare storage

### LINKED ISSUE #1214: feat: import keyshares into non-empy keyshare storage
### Background

For the TEE hard launch, we want a Backup Service that securely stores key-shares of an MPC node (c.f. #949 and [the docs](https://github.com/near/mpc/blob/main/docs/migration-service.md#backup-service)).

The backup service will need to:
- securely store key-shares
- ensure it is in possession of the current key-shares, according to its view of the MPC smart-contract.
- ensure validity of the imported key-shares

The node handles its keyshare storage through the [`KeyshareStorage`](https://github.com/near/mpc/blob/a28272d33bfdccd48fca7fea5c7544332fec7fdc/crates/node/src/keyshare.rs#L86) struct. The struct already helps to ensure some of the properties listed above.

It would make sense for the Backup Service to use the same struct for storing keyshares.

With #1213 we introduce the ability to import key-shares into an empty keyshare storage. For the backup-service, we would need to extend the capability to import key-shares into a non-empty keyshare storage, as it will repeatedly receive current key-shares from the node.

### User Story

As a node operator, I want a backup-service that can securely store keys from different epochs.

### Acceptance Criteria

Be able to import keyshares into a non-empty `KeyshareStorage`. Be mindful of edge cases. Specifically:
- the importing shares must be an exact match for the current contract keyset
- if the permanent `KeyshareStorage` contains keyshares of the same epoch, then they must be a prefix of the importing shares.
- (probably optional): if the temporary `KeyshareStorage` contains any keyshares of matching [`KeyEventId`], then they must be an exact match for the keyshares provided in the backup.

_No response_

kevindeforth:
Resolves #1214

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1499,6 +1499,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "backup-cli"
+version = "0.1.0"
+dependencies = [
+ "clap",
+ "tokio",
+]
+
+[[package]]
 name = "base16ct"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ members = [
     "crates/test_parallel_contract",
     "crates/test_utils",
     "crates/tls",
+    "crates/backup-cli",
 ]
 
 [profile.dev-release]
--- a/crates/backup-cli/Cargo.toml
+++ b/crates/backup-cli/Cargo.toml
@@ -1,0 +1,8 @@
+[package]
+name = "backup-cli"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+clap = { workspace = true }
+tokio = { workspace = true }
--- a/crates/backup-cli/src/adapters.rs
+++ b/crates/backup-cli/src/adapters.rs
@@ -1,0 +1,59 @@
+use crate::{
+    ports::{ContractInterface, KeyShareRepository, P2PClient, SecretsRepository},
+    types,
+};
+
+pub struct DummySecretsStorage {}
+
+impl SecretsRepository for DummySecretsStorage {
+    type Error = String;
+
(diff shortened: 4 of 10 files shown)


crates/node/src/keyshare.rs
```diff
@@ -289,16 +288,30 @@ impl KeyshareStorage {
...
+    /// - A required keyshare is missing in both temporary storage and backup,
+    /// - The reconstructed keyshares differ from the backup (indicating corruption),
+    /// - Or storing to permanent storage fails.
     ///
     /// # Returns
     /// * `Ok(())` if the backup was successfully imported and stored permanently.
     /// * `Err(anyhow::Error)` if any validation or storage step fails.
-    #[allow(dead_code)] // todo: remove after integration with onboarding function
     pub async fn import_backup(
+        // while technically not required to be mut, we must not call read functions in parallel
```
netrome: Isn't this a write function?
```suggestion
        // while technically not required to be mut, we must not call write functions in parallel
```

Otherwise it's nice to use the borrow checker here to guard against concurrent writes.

crates/node/src/keyshare.rs
```diff
@@ -313,15 +326,47 @@ impl KeyshareStorage {
...
+            let key_id =
+                KeyEventId::new(contract_keyset.epoch_id, domain.domain_id, domain.attempt);
+
+            // if we don't have the keyshare in temporary storage, we load it from the backup
+            let keyshare: Keyshare = self
+                .temporary
+                .load_keyshare(key_id)
+                .await?
+                .or_else(|| backup.iter().find(|share| share.key_id == key_id).cloned())
+                .ok_or_else(|| anyhow::anyhow!("Missing keyshare {:?}", key_id))?;
```
netrome: Nit:

> Error messages are typically concise lowercase sentences without trailing punctuation

From https://doc.rust-lang.org/std/error/trait.Error.html

```suggestion
                .ok_or_else(|| anyhow::anyhow!("missing keyshare {:?}", key_id))?;
```

netrome (approved):

PR #1221, opened 2025-10-03
chore: remove redundant `allow` attribute

pbeza:


netrome (approved):

pbeza:
> Thanks! So, it does not have too many arguments anymore?

Correct. I introduced that attribute while addressing code review comments in https://github.com/near/mpc/pull/1202, but in the second round of review it turned out to be redundant, and I forgot to remove it.

PR #1223, opened 2025-10-03
feat: re-submit attestation if node detects it has no attestation on chain

### LINKED ISSUE #1198: Node should re-submit attestation if it detects it has no attestation on chain
### Background

In addition to implementing periodic attestation submissions on a timer, #1051,  we should implement a trigger for nodes to re-submit attestations if the node detects its attestation was removed for any reason on the contract.

A node's attestation can be removed if:
- The attestation becomes invalid due to a timeout. In practice the periodic re-submissions implemented in #1051 should frequent enough to cover attestations on chain expiring.
- A resharing is proposed where the node is removed. Once the resharing completes, the contract removes the attestation for all nodes that are no longer participants. This behavior is fine in practice for production, but problematic for our pytests where we kick nodes out and make them rejoin the network:
https://github.com/near/mpc/blob/808d08854be7b318f9d364aa1b1d365a3dce9f09/pytest/tests/test_key_event.py#L62-L67

### User Story

_No response_

### Acceptance Criteria

The node automatically submits a new attestation to the contract if it detects it has no active attestations stored on chain.

_No response_

pbeza:
Fixes #1198

### DIFF
--- a/crates/node/src/cli.rs
+++ b/crates/node/src/cli.rs
@@ -227,6 +227,13 @@ pub struct ExportKeyshareCmd {
 
 impl StartCmd {
     async fn run(self) -> anyhow::Result<()> {
+        let root_runtime = tokio::runtime::Builder::new_multi_thread()
+            .enable_all()
+            .worker_threads(1)
+            .build()?;
+
+        let _tokio_enter_guard = root_runtime.enter();
+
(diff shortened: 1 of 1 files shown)


netrome (approved):
Nice. Thanks for the updates

Issue #1227, opened 2025-10-05
Use the same type to configure the node in integration tests and when running live


netrome:
### Background

Currently there's a whole bunch of logic in the `StartCmd::run` function which is inherently untestable. While we touch some of it in pytests, we're still hitting bugs like https://github.com/near/mpc/issues/1224 which should be simple to catch with basic unit tests.

While we do have integration tests in place already, they run completely different setup logic using `OneNodeTestConfig::run`. It feels unnatural to have completely different ways of setting up the node for tests and for production. This inherently creates a split which reduces test coverage and creates many code paths which are only touched in production.

A better way to ensure closer parity between the tests and production would be to use a single node configuration, call it `NodeConfig`. This config could be set up to use the appropriate mocks/fakes for test setups, while still being configurable to run with a real indexer, time etc. for production. The only logic residing in `cli.rs` should be pure conversions between CLI args to the appropriate types needed to construct a `NodeConfig` without any side effects.

### User Story

As a developer I want to be able to test our code.

### Acceptance Criteria

* We run the same node logic in our integration tests as we do in production, only differing in which configuration we pass to it.

* The logic in `cli.rs` only does type conversions. No runtimes are created. No tasks are spawned. No requests are being made etc.

### Resources & Additional Notes

Note: This task is exploration heavy and requires investigating the code and having a close dialogue with the team to solicit early feedback and ensure we arrive at a better architecture that better aligns our tests with reality.

PR #1229, opened 2025-10-05
fix: Run attestation requests within a tokio runtime

### LINKED ISSUE #1224: MPC node panics on startup in TEE
### Background
After I reconfigured one of our TDX servers in https://github.com/near/mpc-private/issues/27 I deployed the latest MPC node with dstack and noticed the following error:

```
2025-10-03T13:29:16.401476357Z Starting mpc node...
2025-10-03T13:29:16.409440888Z 
2025-10-03T13:29:16.409452142Z thread 'main' panicked at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.47.1/src/net/unix/stream.rs:894:18:
2025-10-03T13:29:16.409457347Z there is no reactor running, must be called from the context of a Tokio 1.x runtime
2025-10-03T13:29:16.409461653Z stack backtrace:
2025-10-03T13:29:16.409476073Z 2025-10-03T13:29:16.407018Z DEBUG mpc_node::config: p2p and near account secret key not found. Generating...
2025-10-03T13:29:16.409480856Z 2025-10-03T13:29:16.408162Z DEBUG mpc_node::config: p2p and near account key generated in /data/secrets.json
2025-10-03T13:29:16.414723793Z    0:     0x55f7ebceb010 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h6d42cc84fc840290
2025-10-03T13:29:16.414743454Z    1:     0x55f7ebd15c43 - core::fmt::write::h5af61a909e3ec64d
2025-10-03T13:29:16.414748276Z    2:     0x55f7ebce64b3 - std::io::Write::write_fmt::h5a7b54aa6e4a315d
2025-10-03T13:29:16.414759191Z    3:     0x55f7ebceae62 - std::sys::backtrace::BacktraceLock::print::h555579e7396c26ac
2025-10-03T13:29:16.414763525Z    4:     0x55f7ebcebd6a - std::panicking::default_hook::{{closure}}::h9128866118196224
2025-10-03T13:29:16.414767596Z    5:     0x55f7ebcebbd5 - std::panicking::default_hook::h52e9e7314e0255f6
2025-10-03T13:29:16.414771426Z    6:     0x55f7ebcec702 - std::panicking::rust_panic_with_hook::h541791bcc774ef34
2025-10-03T13:29:16.414778355Z    7:     0x55f7ebcec4aa - std::panicking::begin_panic_handler::{{closure}}::h6479a2f0137c7d19
2025-10-03T13:29:16.414782284Z    8:     0x55f7ebceb529 - std::sys::backtrace::__rust_end_short_backtrace::ha04e7c0fc61ded91
2025-10-03T13:29:16.414786436Z    9:     0x55f7ebcec13d - rust_begin_unwind
2025-10-03T13:29:16.414792364Z   10:     0x55f7e933ca70 - core::panicking::panic_fmt::h5764ee7030b7a73d
2025-10-03T13:29:16.414796104Z   11:     0x55f7e932e3a1 - tokio::runtime::scheduler::Handle::current::panic_cold_display::hdd0abbb7943e8dd1
2025-10-03T13:29:16.416426156Z /app/start.sh: line 119:    13 Aborted                 (core dumped) /app/mpc-node start $tee_authority
```

I believe this is because we're calling `tee_authority.generate_attestation(report_data).await?;` outside of a Tokio runtime,
which works for local attestations but when creating the live attestations it's not unreasonable to assume this function
assumes a running Tokio runtime.

Naturally we should fix so that this function runs within a Tokio runtime. Also it would be good to think how we can reduce the risk of similar problems reoccurring, although we've greatly reduced the surface of these potential bugs since we removed the `tee` feature flag.

### User Story

As a node operator I want functioning software.

### Acceptance Criteria

The MPC node doesn't panic on startup in a TEE.

netrome:
Closes #1224 

Forgive me, team, for I have sinned. I have opened this PR without adding test coverage of the change. While the change is trivial and should be easy to review, I did not find any easy way to test this. And since this bug currently breaks the node, it would be nice to get the fix through before refactoring the code to make this testable. 

I have created a follow-up (#1227) to ensure this part of the code is easier to test, but I fear that will be a significant effort to refactor and we might need to push it until after the soft launch. Unless we're hitting many more bugs during E2E testing at which point we can reconsider prioritizing this refactor.

Also this bug is a deja-vu. We fixed a similar issue in https://github.com/near/mpc/pull/612 but for some reason, the runtime guard introduced there is not around anymore. This also shows why it's important to have tests to guard against regressions. @DSharifi - was there any reason to kill the runtime guard that you recall?

### DIFF
--- a/crates/node/src/cli.rs
+++ b/crates/node/src/cli.rs
@@ -225,6 +225,13 @@ pub struct ExportKeyshareCmd {
 
 impl StartCmd {
     async fn run(self) -> anyhow::Result<()> {
+        let root_runtime = tokio::runtime::Builder::new_multi_thread()
+            .enable_all()
+            .worker_threads(1)
+            .build()?;
+
+        let _tokio_enter_guard = root_runtime.enter();
+
         // Load configuration and initialize persistent secrets
         let home_dir = PathBuf::from(self.home_dir.clone());
         let config = load_config_file(&home_dir)?;
@@ -247,11 +254,6 @@ impl StartCmd {
         // Create communication channels and runtime
         let (debug_request_sender, _) = tokio::sync::broadcast::channel(10);
         let root_task_handle = Arc::new(OnceLock::new());
-
-        let root_runtime = tokio::runtime::Builder::new_multi_thread()
-            .enable_all()
-            .worker_threads(1)
-            .build()?;
 
         let (protocol_state_sender, protocol_state_receiver) =
             watch::channel(ProtocolContractState::NotInitialized);


netrome:
> LGTM, but yeah — it’d be nice to find a proper way to test it so we don’t run into the same kind of bug for the third time.

Amen to that. As much as I'd love to do https://github.com/near/mpc/issues/1227, if you have any ideas of lower-hanging fruits or quick solutions in the meantime I'd love to hear them.

PR #1230, opened 2025-10-05
chore: use NEAR style base58 encoding + `ed25519` prefix on node HTTP server.

### LINKED ISSUE #1226: NEAR key endpoint should return  base58 encoded public keys
### Background

In #880 we removed `near_crypto::PublicKey` in favor of `ed25519::VerifyingKey`. As part of that migration the `/public_data` endpoint now serializes public keys generated on the node as a vector of 32 bytes instead of base58 encoding with an `ed25519:` prefix.

Although serializing the public key to 32 bytes is the default serialization format of `ed25519_dalek::VerifyingKey`, downstream users (node operators) of the `/public_data` endpoint will need to use these public keys with the NEAR CLI to create access keys.

The NEAR cli expects `base58` encoding of the key with a `ed25519:` prefix, so we should change the node endpoint to also return `base58` encoding of the keys such that node operators don't have to do this tedious conversion themselves.

### User Story

As a node operator that delegates access keys to the nodes, I'd like the public keys of the node to be formatted the same way as the NEAR cli 

### Acceptance Criteria

- `/public_data` endpoint on the node returns public keys serialized with `base58` encoding pluse `ed25519:` prefix. 

_No response_

DSharifi:
closes #1226


netrome (approved):
Would be nice with some extra test coverage, but the current set is sufficient. I'll take this over (@DSharifi is off), extend the test coverage and add some more tests before merging.

netrome:
Updated and set to auto-merge.

crates/node-types/src/http_server.rs
```diff
@@ -0,0 +1,144 @@
...
+    use rand_chacha::{ChaCha20Rng, rand_core::SeedableRng};
+
+    fn gen_verifying_key(seed: u64) -> VerifyingKey {
+        let mut rng = ChaCha20Rng::seed_from_u64(seed);
+        let sk = SigningKey::generate(&mut rng);
+        sk.verifying_key()
+    }
+
+    #[test]
+    fn verifying_key_bs58__should_serialize_to_exact_snapshot() {
```
pbeza, line 35: Is the double underscore intentional? Why not replace `__` with a single `_`? (The same question applies to several other places in the code.)

Issue #1184, opened 2025-09-25
Indexer should not be mapping `ProtocolContractState` to `ContractState`


DSharifi:
### Background

Our indexer code is currently updating two watchers for the MPC contract state. See https://github.com/near/mpc/blob/60e4a4cbd22715f78c4a4a2e8ed1a873568ee661/crates/node/src/indexer/participants.rs#L201-L202

The `contract_state_sender` contains the node's representation of the MPC state, and is what is used in the implementation of the node coordinator: https://github.com/near/mpc/blob/60e4a4cbd22715f78c4a4a2e8ed1a873568ee661/crates/node/src/coordinator.rs#L103

The other `protocol_state_sender` is the raw representation of the contract state, and currently used in the web server endpoint. 

Note that both these senders are sending the same underlying data, but the `contract_state_sender` applies a transformation first on the raw state of the contract before sending it.

We can remove the `contract_state_sender`, and instead only have the `protocol_state_sender` which the node coordinator can also rely on. The transformation from the raw state to `ContractState  is specific to the MPC node and coordinator, and should be done by the coordinator. That means we only need one watcher in the indexer for the raw state.

Same applies to the migration state receivers introduced in https://github.com/near/mpc/pull/1250

### Acceptance Criteria

- The second watcher in `participants.rs`, `contract_state_sender`, is removed.
- The indexer does not perform transformations from `ProtocolContractState` into `ContractState`.
- The mpc coordinator does the mapping from `ProtocolContractState` into `ContractState` instead of receiving the mapped version over a channel.

### Resources & Additional Notes

https://github.com/near/mpc/issues/1182#issuecomment-3332730632

kevindeforth:
Thanks for opening this!

kevindeforth:
Some issues observed while working with the indexer:
- We mix actix and tokio. This is probably fine, because actix is based on tokio, but it makes the code slightly harder to follow and maintain. We have an [open issue](https://github.com/near/mpc/issues/156) for moving away from actix, but it's a bit complicated due to our reliance on [nearcore intenals](https://github.com/near/mpc/blob/279131a8ba8481d6f82529a72dcabf52dc0d2ca3/crates/node/src/indexer/real.rs#L75).
https://github.com/near/nearcore/blob/b2e47a2f22303cbfa77368b19e003bf74bc23e3a/chain/indexer/src/lib.rs#L154-L163, which expose the type `actix::Addr`.
- Extending indexer functionality is cumbersome
[...]

Issue #1249, opened 2025-10-07
Migration Service: write onboard logic


kevindeforth:
### Background

Write the onboarding logic for MPC nodes.

### Acceptance Criteria

Either test onboarding logic in an integration test (might be tricky, due to https://github.com/near/mpc/issues/1227), or test it in a pytest (which requires web-endpoints to be added).

Issue #1173, opened 2025-09-23
Create DTO types for types used in migration service


kevindeforth:
### Background

https://github.com/near/mpc/pull/1162#pullrequestreview-3257527673

### Acceptance Criteria

All exposed types related to the migration service are specified in the dto crate.

### Resources & Additional Notes
https://github.com/near/mpc/pull/1250#discussion_r2413643177
_No response_

netrome:
Oops, commented in the wrong issue. This one isn't as critical. Setting as P3

PR #1250, opened 2025-10-07
feat: indexer fetches migration state from contract and displays it on a web-endpoint

kevindeforth:
This PR:
- integrates the contract migration logic with the indexer
- adds debug web-endpoints to display the current view the node has on the migration state in the contract
- tests the aforementioned logic in a comprehensive pytest, verifying that debug endpoint and indexer work as expected.

 We also hook-up the fake indexer with the migration state logic, which will be useful for when we work on #1249 .

Note that part of the indexer integration follows the existing, concerning pattern  of spawning channels without a true separation of concerns. While working on this, I also, initially, resolved https://github.com/near/mpc/issues/1184, but found the PR would become quite a bit too large and lose focus, which led to a reversion of the refactor and a minimal diff seen herein.

Some observations about working with the indexer and a proposal for abstraction have been added [here](https://github.com/near/mpc/issues/1184#issuecomment-3378017936)

### DIFF
--- a/crates/node/src/indexer/migrations.rs
+++ b/crates/node/src/indexer/migrations.rs
@@ -1,0 +1,98 @@
+use std::{collections::BTreeMap, sync::Arc};
+
+use ed25519_dalek::VerifyingKey;
+use mpc_contract::node_migrations::{BackupServiceInfo, DestinationNodeInfo};
+use near_sdk::AccountId;
+use tokio::sync::watch;
+
+use crate::{
+    indexer::{
+        lib::{get_mpc_migration_info, wait_for_full_sync},
+        IndexerState,
+    },
+    migration_service::types::MigrationInfo,
+};
+
+pub type ContractMigrationInfo =
+    BTreeMap<AccountId, (Option<BackupServiceInfo>, Option<DestinationNodeInfo>)>;
+
+const MIGRATION_INFO_REFRESH_INTERVAL: std::time::Duration = std::time::Duration::from_secs(1);
+
+/// Blocks until the indexer has a current view of the blockchain.
+/// Spawns a monitoring task to fetch the migration info from the contract.
+/// Returns a tokio watch channel on which the latest migration state can be watched.
+///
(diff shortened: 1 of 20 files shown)


crates/node/src/migration_service/types.rs
```diff
@@ -0,0 +1,55 @@
...
+#[derive(PartialEq, Debug, Clone)]
+pub struct MigrationInfo {
+    pub backup_service_info: Option<BackupServiceInfo>,
+    pub active_migration: bool,
+}
+
+impl MigrationInfo {
+    pub fn from_contract_state(
+        my_account_id: &AccountId,
+        my_p2p_tls_key: &VerifyingKey,
```
gilcu3, line 16: If this type is not actually used for some crypto operation here, then I think it is better to use the dtos_contract type: `Ed25519PublicKey`

kevindeforth: Good observation!
This is the type currently exposed by the contract API. We have an issue to change it, but that is out of scope for this PR: https://github.com/near/mpc/issues/1173.

gilcu3: Interesing :) I thought I had changed that, but maybe the change is not in main yet.

gilcu3: Indeed it was changed https://github.com/near/mpc/pull/1225

crates/node/src/indexer/migrations.rs
```diff
@@ -0,0 +1,98 @@
...
+use crate::{
+    indexer::{
+        lib::{get_mpc_migration_info, wait_for_full_sync},
+        IndexerState,
+    },
+    migration_service::types::MigrationInfo,
+};
+
+pub type ContractMigrationInfo =
+    BTreeMap<AccountId, (Option<BackupServiceInfo>, Option<DestinationNodeInfo>)>;
```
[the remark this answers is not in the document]
kevindeforth: This will be handled with #1173

PR #1251, opened 2025-10-08
chore: remove unneeded deps in attestation crate

### LINKED ISSUE #1235: Remove redundant dependencies of attestation crate
### Background

Currently the attestation crate depends on `k256`, `near-sdk`, while these deps are not really needed.

To remove `near-sdk` the type:
```rust
pub struct ReportDataV1 {
    tls_public_key: near_sdk::PublicKey,
}
```

should be modified so that it enforces only the conditions necessary for attestation, for example that is 64 bytes of data.

### User Story

_No response_

### Acceptance Criteria

`ReportDataV1` is redefined and unnecessary deps are removed

https://github.com/near/mpc/pull/1225#discussion_r2405746850

gilcu3:
Closes #1235 

- The change from `near_sdk::PublicKey` to `[u8; 32]` is open to discussion. I actually believe the crate could be made independent of what goes into the report data, in which care report data would simply be a type containing bytes. The change implemented is in that direction, but creating much less disruptive changes.

### DIFF
--- a/crates/attestation/src/report_data.rs
+++ b/crates/attestation/src/report_data.rs
@@ -54,7 +54,13 @@ pub struct ReportDataV1 {
     BorshDeserialize,
     BorshSerialize,
 )]
-struct Ed25519PublicKey([u8; 32]);
+pub struct Ed25519PublicKey([u8; 32]);
+
+impl core::borrow::Borrow<[u8]> for Ed25519PublicKey {
+    fn borrow(&self) -> &[u8] {
+        &self.0
+    }
+}
 
 /// report_data_v1: [u8; 64] =
 ///   [version(2 bytes big endian) || sha384(TLS pub key) || zero padding]
@@ -63,7 +69,7 @@ impl ReportDataV1 {
     const PUBLIC_KEYS_OFFSET: usize = BINARY_VERSION_OFFSET + BINARY_VERSION_SIZE;
     const PUBLIC_KEYS_HASH_SIZE: usize = 48;
 
-    pub fn new(tls_public_key: [u8; 32]) -> Self {
+    pub fn new(tls_public_key: impl Into<Ed25519PublicKey>) -> Self {
         Self {
             tls_public_key: tls_public_key.into(),
         }
@@ -116,7 +122,7 @@ pub enum ReportData {
 }
 
 impl ReportData {
-    pub fn new(tls_public_key: [u8; 32]) -> Self {
+    pub fn new(tls_public_key: impl Into<Ed25519PublicKey>) -> Self {
         ReportData::V1(ReportDataV1::new(tls_public_key))
     }
 


crates/attestation/src/report_data.rs
```diff
@@ -34,7 +34,8 @@ impl ReportDataVersion {
 
 #[derive(Debug, Clone, Constructor)]
 pub struct ReportDataV1 {
-    tls_public_key: near_sdk::PublicKey,
+    // This corresponds to an Ed25519 public key
+    tls_public_key: [u8; 32],
```
pbeza, line 38: I’d prefer to have a self-explanatory strong type here rather than a raw byte array with a comment explaining it, tbh.

crates/attestation/src/report_data.rs
```diff
@@ -85,11 +86,7 @@ impl ReportDataV1 {
 
     /// Generates SHA3-384 hash of TLS public key only.
     fn public_keys_hash(&self) -> [u8; Self::PUBLIC_KEYS_HASH_SIZE] {
-        let mut hasher = Sha3_384::new();
-        // Skip first byte as it is used for identifier for the curve type.
-        let key_data = &self.tls_public_key.as_bytes()[1..];
-        hasher.update(key_data);
-        hasher.finalize().into()
+        Sha3_384::digest(self.tls_public_key).into()
```
pbeza, line 89: I trust you that this is effectively the same as the original implementation you modified. But in that case, this wrapper function doesn’t make much sense to me anymore — I think we could just inline the calls to it and remove it entirely.

crates/attestation/src/report_data.rs
```diff
@@ -99,7 +96,7 @@ pub enum ReportData {
 }
 
 impl ReportData {
-    pub fn new(tls_public_key: near_sdk::PublicKey) -> Self {
+    pub fn new(tls_public_key: [u8; 32]) -> Self {
```
pbeza, line 99: I’d prefer to have a self-explanatory strong type here rather than a raw byte array. TLS public keys are used so widely across our codebase that we should have a single type representing them, easily convertible to all the necessary formats.

crates/attestation/src/report_data.rs
netrome: One option is to have a named bytes wrapper defined in the attestation crate. It would convey that this crate only cares about the raw bytes while still being a self-explanatory strong type that doesn't need a comment. What do you think of that?

Something like:
```
struct Ed25519PubKey([u8; 32]);
```

and deriving all appropriate traits for it (i.e. Clone, Copy, Hash, Deref, AsRef, Borrow etc.)

gilcu3: ok, fixed in 04157008ae78bf844d6442caf3e4f71996d92cd1

netrome: > by Borrow you mean std::borrow::Borrow, it seems I cannot use std here?

Yes but this one should be in `core` as well as `core::borrow::Borrow`.

gilcu3: > It also prevents us from accidentally passing another [u8; 32] value here (like a hash output or whatever).

How does it prevent that? The current solution allows the input of [u8; 32], else we would need to make this new type public and use conversion every time it is used? Was that the point?

crates/attestation/src/report_data.rs
```diff
@@ -32,18 +32,43 @@ impl ReportDataVersion {
...
+struct Ed25519PublicKey([u8; 32]);
+
 /// report_data_v1: [u8; 64] =
 ///   [version(2 bytes big endian) || sha384(TLS pub key) || zero padding]
 impl ReportDataV1 {
     /// V1-specific format constants
     const PUBLIC_KEYS_OFFSET: usize = BINARY_VERSION_OFFSET + BINARY_VERSION_SIZE;
     const PUBLIC_KEYS_HASH_SIZE: usize = 48;
 
+    pub fn new(tls_public_key: [u8; 32]) -> Self {
```
netrome, line 66:
```suggestion
    pub fn new(tls_public_key: impl Into<Ed25519PublicKey>) -> Self {
```

crates/attestation/src/report_data.rs
netrome: Right yeah that's correct. Although putting the newtype in the interface makes the expected type clearer, and then the caller could choose whether or not they want to convert on their end.

netrome (approved):
Nice stuff!

PR #1270, opened 2025-10-12
feat: Add public key enforcement feature

### LINKED ISSUE #1205: Restrict MPC Contract APIs to MPC Node Account Keys
### Background

Currently, MPC contract APIs can be called by any NEAR account, leaving the system open to potential DoS attacks and misuse. While protocol-level fixes allow TEEs to validate encrypted keys, the contracts themselves cannot enforce restrictions.
This task implements the design outlined in [Issue #875](https://github.com/near/mpc/issues/875)
, ensuring that only MPC nodes with valid attestations (and in some cases, operators) can call critical APIs.

### User Story

As a developer/operator of MPC services,
I want contract APIs to only accept calls from verified MPC nodes (with valid TEE attestations) or authorized operators,
so that the system is resilient against DoS attacks and unauthorized acces

### Acceptance Criteria

 Node attestation integration:

MPC node public key hash included in TEE report_data.

Contract verifies that submitted public key matches the attested key.

 API restrictions applied:

Node-only APIs restricted to verified MPC nodes.

Operator-only APIs restricted to operator accounts with minimal permission keys.

View-only APIs remain open.

 Contract upgrade path:

Phase 1: Add Option<account public key> field to Node_ID (backward compatibility).

Phase 2: Make the account key mandatory.

 Documentation updated:

Operator guide updated with explicit API access lists.

 Tests:

Add integration tests to ensure restricted APIs reject invalid or unauthorized callers.

Migration/backup service workflows verified against new restrictions.

Design doc: [Issue #875](https://github.com/near/mpc/issues/875)

Related discussions:

Restricting API: Slack thread

Node migration service: [PR #877](https://github.com/near/mpc/pull/877/files#r2287333508)

Consider corner cases such as conclude_node_migration, which can be called by a non-participant node but must still validate signer key against attested data.

barakeinav1:
fixes #1205

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5065,6 +5065,7 @@ dependencies = [
  "gcloud-sdk",
  "hex",
  "humantime",
+ "hyper 0.14.32",
  "itertools 0.12.1",
  "k256",
  "lru 0.12.5",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,6 +86,7 @@ ed25519-dalek = { version = "2.2.0", features = ["serde", "digest", "pkcs8", "ra
 elliptic-curve = "0.13.8"
 flume = "0.11.1"
 fs2 = "0.4"
+hyper = { version = "0.14.32", features = ["server", "http1", "tcp", "stream"]} # due to axum
 futures = "0.3.31"
 gcloud-sdk = { version = "0.26.2", default-features = false, features = [
     "google-cloud-secretmanager-v1",
--- a/crates/node/Cargo.toml
+++ b/crates/node/Cargo.toml
@@ -56,6 +56,7 @@ node-types = { workspace = true }
 mpc-contract = { workspace = true }
 mpc-tls = { workspace = true }
 itertools = { workspace = true }
+hyper = { workspace = true }
 
 [dev-dependencies]
 assert_matches = { workspace = true }
--- a/crates/node/src/migration_service.rs
+++ b/crates/node/src/migration_service.rs
@@ -1,2 +1,3 @@
 pub mod onboarding;
 pub mod types;
+pub mod web;
--- a/crates/node/src/migration_service/types.rs
+++ b/crates/node/src/migration_service/types.rs
@@ -4,18 +4,49 @@ use mpc_contract::{
     primitives::key_state::Keyset,
 };
 use near_sdk::AccountId;
+use serde::Serialize;
 use tokio_util::sync::CancellationToken;
 
 use crate::{
     config::{NodeStatus, ParticipantStatus},
     indexer::{migrations::ContractMigrationInfo, participants::ContractState},
     providers::PublicKeyConversion,
+    trait_extensions::convert_to_contract_dto::TryIntoNodeType,
 };
 
-#[derive(PartialEq, Debug, Clone)]
+pub struct NodeBackupServiceInfo {
+    pub p2p_key: VerifyingKey,
+}
+
+impl NodeBackupServiceInfo {
+    pub fn from_contract(info: BackupServiceInfo) -> anyhow::Result<Self> {
+        let p2p_key = match info.public_key.try_into_node_type() {
+            Ok(res) => res,
+            Err(err) => {
+                anyhow::bail!("can't convert key: {}", err);
+            }
+        };
+        Ok(Self { p2p_key })
+    }
+}
+
+#[derive(PartialEq, Debug, Clone, Serialize)]
 pub struct MigrationInfo {
     pub backup_service_info: Option<BackupServiceInfo>,
     pub active_migration: bool,
+}
+
+impl MigrationInfo {
+    pub fn get_pk_backup_service(self) -> Option<VerifyingKey> {
+        self.backup_service_info
+            .and_then(|info| match NodeBackupServiceInfo::from_contract(info) {
+                Ok(service) => Some(service.p2p_key),
+                Err(err) => {
+                    tracing::warn!("could not convert backup service info: {}", err);
+                    None
+                }
+            })
+    }
 }
 
 #[derive(Clone)]
@@ -132,9 +163,28 @@ pub mod tests {
         config,
         indexer::{migrations::ContractMigrationInfo, participants::ContractState},
         providers::PublicKeyConversion,
+        trait_extensions::convert_to_contract_dto::TryIntoNodeType,
     };
 
     use super::{MigrationInfo, OnboardingJob};
+
+    #[test]
+    fn test_migration_get_pk_backup_service() {
+        let empty = MigrationInfo {
+            backup_service_info: None,
+            active_migration: true,
+        };
+        assert!(empty.get_pk_backup_service().is_none());
+
+        let public_key = bogus_ed25519_public_key();
+        let pk_converted = public_key.clone().try_into_node_type().unwrap();
+        let backup_service_info = Some(BackupServiceInfo { public_key });
+        let populated = MigrationInfo {
+            backup_service_info,
+            active_migration: true,
+        };
+        assert_eq!(populated.get_pk_backup_service(), Some(pk_converted))
+    }
 
     #[test]
     fn test_migration_status_constructor_empty() {
--- a/crates/node/src/migration_service/web.rs
+++ b/crates/node/src/migration_service/web.rs
@@ -1,0 +1,203 @@
+#![allow(dead_code)]
+pub mod authentication;
+pub mod client;
+pub mod server;
+pub mod types;
+
+#[cfg(test)]
+mod tests {
+    use std::sync::Arc;
+
+    use ed25519_dalek::SigningKey;
+    use mpc_contract::node_migrations::BackupServiceInfo;
+    use rand::rngs::OsRng;
+    use tokio::sync::watch;
+    use tokio_util::sync::CancellationToken;
+
+    use crate::keyshare::{test_utils::KeysetBuilder, Keyshare};
+    use crate::migration_service::web::client::{
+        connect_to_web_server, make_hello_request, make_keyshare_get_request,
+        make_set_keyshares_request,
+    };
+    use crate::migration_service::web::{server::start_web_server, types::WebServerState};
+    use crate::{
+        config::WebUIConfig, migration_service::types::MigrationInfo, p2p::testing::PortSeed,
+    };
+
+    const LOCALHOST_IP: &str = "127.0.0.1";
+
+    struct TestSetup {
+        client_key: SigningKey,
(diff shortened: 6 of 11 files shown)


crates/contract/src/tee/tee_state.rs
```diff
@@ -14,15 +14,35 @@ use borsh::{BorshDeserialize, BorshSerialize};
...
 
 #[near(serializers=[borsh, json])]
-#[derive(Debug, Eq, Ord, PartialEq, PartialOrd, Clone, Hash)]
+#[derive(Debug, Ord, PartialOrd, Clone)] //#[derive(Debug, Ord, PartialOrd, Clone)]
 pub struct NodeId {
     /// Operator account
     pub account_id: AccountId,
     /// TLS public key, MUST BE of type Ed25519
     pub tls_public_key: near_sdk::PublicKey,
+    pub account_public_key: Option<near_sdk::PublicKey>,
```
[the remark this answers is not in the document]
barakeinav1: add [#1284](https://github.com/near/mpc/issues/1284) for follow up task

crates/attestation/src/report_data.rs
```diff
@@ -92,14 +96,16 @@ impl ReportDataV1 {
...
             .copy_from_slice(&version_bytes);
 
-        // Generate and copy hash of public keys
-        let public_keys_hash: [u8; Self::PUBLIC_KEYS_HASH_SIZE] =
-            Sha3_384::digest(self.tls_public_key.as_ref()).into();
+        // Hash both TLS and account public keys
+        let mut hasher = Sha3_384::new();
+        hasher.update(self.tls_public_key.as_ref());
+        hasher.update(self.account_public_key.as_ref());
+        let public_keys_hash: [u8; Self::PUBLIC_KEYS_HASH_SIZE] = hasher.finalize().into();
```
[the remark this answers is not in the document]
barakeinav1: got some compilation and reference problems for this change, prefer to keep as is.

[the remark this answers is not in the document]
barakeinav1: but I did add a new [compute_public_keys_hash ](https://github.com/near/mpc/pull/1270/commits/436be64ece5a71da35b3208ff0e6207e64a839b8#diff-cfcc3ce4d911b95bf27fec97349742a09d1bb7a6108be87f40c4785855fa6407R94
)

netrome (changes requested):
Just one TODO comment that should be addressed.

crates/contract/src/tee/tee_state.rs
```diff
@@ -246,36 +301,70 @@ impl TeeState {
...
+    /// Find a NodeId by its TLS public key.
+    pub fn find_node_id_by_tls_key(&self, tls_public_key: &near_sdk::PublicKey) -> Option<NodeId> {
+        self.participants_attestations
+            .get(tls_public_key)
+            .map(|(node_id, _)| node_id.clone())
+    }
+    pub fn is_caller_an_attested_node(&self) -> bool {
+        let signer_pk = env::signer_account_pk();
+        let signer_id = env::signer_account_id();
+        //let predecessor = env::predecessor_account_id(); //TODO check what is better predecessor or signer
```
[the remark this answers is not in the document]
barakeinav1: it probably should be predecessor (but I'm not 100% sure).
so I want to check it a bit further. before changing.
I moving this to [#1288](https://github.com/near/mpc/issues/1288) as well.


see this https://docs.near.org/smart-contracts/anatomy/environment#predecessor-and-signer.

see some background here:
https://near.zulipchat.com/#narrow/channel/469556-community-support/topic/Does.20an.20internal.20call.20change.20the.20predecessor_account_id/with/543161605

netrome (dismissed):
Just a dangling todo, otherwise should be ready to merge. Please fix and I'll be happy to approve.

crates/node/src/config.rs
```diff
@@ -399,10 +399,17 @@ pub struct RespondConfig {
...
+        // using main account for responding.
         Self {
+            account_id: config.my_near_account_id.clone(),
+            access_keys: vec![secrets.near_signer_key.clone()],
+        }
+        /*Self {
             account_id: config.near_responder_account_id.clone(),
             access_keys: secrets.near_responder_keys.clone(),
-        }
+        }*/
```
netrome: Also we shouldn't keep commented out sections of code in production code.

```suggestion
```

crates/contract/src/primitives/participants.rs
```diff
@@ -204,12 +204,16 @@ impl Participants {
         }
     }
 
+    /// Returns the set of NodeIds corresponding to the participants.
```
pbeza:
```suggestion
    /// Returns the set of [`NodeId`]s corresponding to the participants.
```

crates/contract/src/primitives/participants.rs
```diff
@@ -204,12 +204,16 @@ impl Participants {
...
+    /// Note that the account_public_key field in NodeId is None.
+    /// This is because NodeId is used in contexts where account_public_key is not needed (only TLS key is needed).
```
pbeza:
```suggestion
    /// Note that the `account_public_key` field in [`NodeId`] is `None`.
    /// This is because [`NodeId`] is used in contexts where `account_public_key` is not needed (only TLS key is needed).  
```

crates/contract/src/tee/tee_state.rs
```diff
@@ -14,15 +14,35 @@ use borsh::{BorshDeserialize, BorshSerialize};
 use contract_interface::types::Ed25519PublicKey;
 use mpc_primitives::hash::LauncherDockerComposeHash;
 use near_sdk::{env, near, store::IterableMap, AccountId};
+use std::hash::Hash;
+use std::hash::Hasher;
```
pbeza:
```suggestion
use std::hash::{Hash, Hasher};
```

crates/node/src/config.rs
barakeinav1: sure, I thought to leave this so it would be clearer for whoever takes up issue [#1296]
but I'll remove it.

crates/attestation/src/report_data.rs
```diff
@@ -120,10 +139,19 @@ impl ReportDataV1 {
...
     V1(ReportDataV1),
 }
-
 impl ReportData {
-    pub fn new(tls_public_key: impl Into<Ed25519PublicKey>) -> Self {
-        ReportData::V1(ReportDataV1::new(tls_public_key))
+    /// Creates a new ReportData instance.
+    ///
+    /// * `tls_public_key`: The TLS key of the MPC node
+    /// * `account_public_key`: The NEAR account signing key
```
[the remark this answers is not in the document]
kevindeforth: I thought that was just markdown and that it didn't matter. 
This is something that can be automated and enforced though.
So if we want to be consistent across the code base, I would suggest we include a format check and add it to the CI.
Otherwise, this will just be a back-and-forth of personal preferences

crates/contract/src/lib.rs
```diff
@@ -434,8 +435,11 @@ impl MpcContract {
         response: SignatureResponse,
     ) -> Result<(), Error> {
         let signer = env::signer_account_id();
+
         log!("respond: signer={}, request={:?}", &signer, &request);
 
+        self.tee_state.assert_caller_is_attested_node();
```
kevindeforth: Doesn't this only assert that the node has a valid attestation? I don't think it enforces the node to be a participant in the current epoch.

crates/contract/src/lib.rs
```diff
@@ -516,6 +520,8 @@ impl MpcContract {
         let signer = env::signer_account_id();
         log!("respond_ckd: signer={}, request={:?}", &signer, &request);
 
+        self.tee_state.assert_caller_is_attested_node();
```
kevindeforth: idem

kevindeforth (dismissed):
It seems to me this is not resolving the API restriction listed as an acceptance criteria in #1205. In case this was planned as a follow-up, please dismiss my review, but since the PR description doesn't indicate such things, I will request changes until we know this is deliberate.

IIUC, a node can have an attestation in `participant_attestatins` (which, in that case, would be a confusing name), without being a participant

https://github.com/near/mpc/blob/bc14426d1ae2254faa3563f9d3df97a241c09ffb/crates/contract/src/tee/tee_state.rs#L200-L208


https://github.com/near/mpc/blob/bc14426d1ae2254faa3563f9d3df97a241c09ffb/crates/contract/src/lib.rs#L541-L611

crates/contract/src/tee/tee_state.rs
```diff
@@ -246,36 +299,70 @@ impl TeeState {
...
+    /// Find a NodeId by its TLS public key.
+    pub fn find_node_id_by_tls_key(&self, tls_public_key: &near_sdk::PublicKey) -> Option<NodeId> {
+        self.participants_attestations
+            .get(tls_public_key)
+            .map(|(node_id, _)| node_id.clone())
+    }
+    pub fn is_caller_an_attested_node(&self) -> bool {
+        let signer_pk = env::signer_account_pk();
+        let signer_id = env::signer_account_id();
+        //let predecessor = env::predecessor_account_id(); //TODO (#1288)  check what is better predecessor or signer
```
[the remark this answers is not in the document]
barakeinav1: yes, this will be added in [#1205 follow-up: public enforcement part 2](https://github.com/near/mpc/issues/1288)
#1288

netrome:
Looking back at this now, I think we missed a part of the original issue:

> This task implements the design outlined in https://github.com/near/mpc/issues/875
, ensuring that **only MPC nodes** with valid attestations (and in some cases, operators) can call critical APIs.

Right now the restriction is that the caller has to be the account associated with an MPC node. However, this is an account created outside of the TEEs, entirely under the operator's control. Naturally this doesn't break any security assumptions during the soft launch, but down the line we'll need stronger guarantees that the node operators cannot forge respond calls to CKD requests for example.

barakeinav1:
> Right now the restriction is that the caller has to be the account associated with an attested MPC node. However, this is an >account created outside of the TEEs, entirely under the operator's control. Naturally this doesn't break any security assumptions >during the soft launch, but down the line we'll need stronger guarantees that the node operators cannot forge respond calls to >CKD requests for example.

@netrome the public key we verify does come from the node, and not from the operator

netrome:
> @netrome The account is created outside of the TEE, but the private key associated with node is created by the node and never leave. The corresponding public key is what is added an enforced on the contract to restrict APIs. The node send a hash of the key it as part of the attestation report_data, and we verify the key matches the hash, before storing it on the contract.

Thanks for clarifying, I see I missed the public key check in the `is_caller_an_attested_participant` function.

PR #1274, opened 2025-10-13
chore: add support for CKD BLS in contract sandbox tests

### LINKED ISSUE #1243: contract sandbox tests for CKD
### Background

After https://github.com/near/mpc/pull/1239 the contract sandbox tests for CKD were left as a follow-up

### User Story

_No response_

### Acceptance Criteria

We have contract sandbox tests for CKD

_No response_

gilcu3:
Closes #1243

### DIFF
--- a/crates/contract/tests/sandbox/common.rs
+++ b/crates/contract/tests/sandbox/common.rs
@@ -956,9 +956,8 @@ pub async fn call_contract_key_generation<const N: usize>(
     let existing_domains = {
         let state: ProtocolContractState = contract.view("state").await.unwrap().json().unwrap();
         match state {
-            ProtocolContractState::Initializing(state) => state.domains.domains().len(),
             ProtocolContractState::Running(state) => state.domains.domains().len(),
-            _ => 0,
+            _ => panic!("ProtocolContractState must be Running"),
         }
     };
 


crates/contract/tests/sandbox/common.rs
```diff
@@ -418,16 +394,17 @@ pub fn new_bls12381() -> (dtos::PublicKey, ckd::KeygenOutput) {
     (pk, keygen_output)
 }
 
-pub async fn init_env_ed25519(
+pub async fn init_env(
     num_domains: usize,
+    scheme: SignatureScheme,
```
kevindeforth: Can we pass a vector of schemes and create domains accordingly? Or will that break to many tests / blow-up the refactor?

crates/contract/tests/sandbox/upgrade_to_current_contract.rs
```diff
@@ -261,3 +281,120 @@ async fn all_participants_get_valid_mock_attestation_for_soft_launch_upgrade() -
...
+/// 6. Upgrades the contract to the new version and runs `migrate()`.
+/// 7. Asserts that the state (participants, domains, proposals, signature requests, etc.)
+///    is identical post-upgrade.
+/// 10. Adds new domains, including CKD
+/// 11. Submits new signature and ckd requests
+/// 12. Confirms that pending signature and ckd requests created before and after the upgrade
+///    can still be responded to.
+#[rstest]
+#[tokio::test]
+async fn upgrade_allows_new_request_types(
```
kevindeforth: Is seems like this is an extension of `upgrade_preserves_state_and_requests`. It's nice to have a comprehensive test like this, but I would suggest to test the second part in isolation. There should be a `test_contract_sign_request_ckd` in the `sign.rs` file.

It might be nice to extract logic used by different tests into separate functions.

kevindeforth (commented):
Very nice improvements!
A few nits & questions in the interest of coherence across the code.

crates/contract/tests/sandbox/upgrade_to_current_contract.rs
gilcu3: Here what I was testing was exactly that the new contract with the old state can handle new request types. The second part was already tested elsewhere. If you think this test is not necessary we could simply remove it.

PR #1281, opened 2025-10-14
fix: use more efficient DomainRegistry

### LINKED ISSUE #1255: Use of public_key method in contract
### Background

In the context of contract methods `sign` and `request_app_private_key`, if we are only interested in the domain id, we should look that one up in our `DomainRegistry` instead of calling `self.public_key`.
The latter makes a bunch of conversions which increase gas usage and are not required here.

### User Story

Less gas is better

### Acceptance Criteria

We use the  `DomainRegistry` instead of `self.public_key`

Original idea https://github.com/near/mpc/pull/1239#discussion_r2415873490

gilcu3:
Closes #1255


kevindeforth (approved):
Excellent!

PR #1313, opened 2025-10-19
feat: add enforcement that contract call are by attested participants

### LINKED ISSUE #1288: #1205 follow-up: node must be attested and a participant.
### Background

This issue tracks follow-up work postponed from the **Public Key Enforcement** PR.

 node must be attested and a participant.
Add check to assert_caller_is_attested_node to check that the MPC node is also a participant (in addition to being attested) 

https://nearone.slack.com/archives/C0912BTG51T/p1760349135203859?thread_ts=1760289141.266809&cid=C0912BTG51T

barakeinav1:
fix #1288

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1503,6 +1503,12 @@ name = "backup-cli"
 version = "0.1.0"
 dependencies = [
  "clap",
+ "ed25519-dalek",
+ "rand 0.8.5",
+ "rand_core 0.6.4",
+ "serde",
+ "serde_json",
+ "thiserror 2.0.16",
  "tokio",
 ]
 
--- a/crates/backup-cli/Cargo.toml
+++ b/crates/backup-cli/Cargo.toml
@@ -5,4 +5,10 @@ edition = "2024"
 
 [dependencies]
 clap = { workspace = true }
+ed25519-dalek = { workspace = true }
+rand = { workspace = true }
+rand_core = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+thiserror = { workspace = true }
 tokio = { workspace = true }
--- a/crates/backup-cli/src/adapters.rs
+++ b/crates/backup-cli/src/adapters.rs
@@ -1,21 +1,11 @@
+use ed25519_dalek::VerifyingKey;
+
 use crate::{
-    ports::{ContractInterface, KeyShareRepository, P2PClient, SecretsRepository},
+    ports::{ContractInterface, KeyShareRepository, P2PClient},
     types,
 };
 
-pub struct DummySecretsStorage {}
-
-impl SecretsRepository for DummySecretsStorage {
-    type Error = String;
-
-    async fn store_private_key(&self, _private_key: &types::PrivateKey) -> Result<(), Self::Error> {
-        Ok(())
-    }
-
-    async fn load_private_key(&self) -> Result<types::PrivateKey, Self::Error> {
-        Ok(types::PrivateKey {})
-    }
-}
+pub mod secrets_storage;
 
 pub struct DummyKeyshareStorage {}
 
@@ -50,10 +40,7 @@ pub struct DummyContractInterface {}
 impl ContractInterface for DummyContractInterface {
     type Error = String;
 
-    async fn register_backup_data(
-        &self,
-        _public_key: &types::PublicKey,
-    ) -> Result<(), Self::Error> {
+    async fn register_backup_data(&self, _public_key: &VerifyingKey) -> Result<(), Self::Error> {
         Ok(())
     }
 }
--- a/crates/backup-cli/src/adapters/secrets_storage.rs
+++ b/crates/backup-cli/src/adapters/secrets_storage.rs
@@ -1,0 +1,169 @@
+use std::{path::Path, sync::Arc};
+
+use tokio::{
+    fs::File,
+    io::{AsyncRead, AsyncReadExt, AsyncSeek, AsyncSeekExt, AsyncWrite, AsyncWriteExt},
+    sync::Mutex,
+};
+
+use crate::{ports::SecretsRepository, types};
+
+const SECRETS_FILE_NAME: &str = "secrets.json";
+
+pub struct SharedJsonSecretsStorage<D>(Arc<Mutex<JsonSecretsStorage<D>>>);
+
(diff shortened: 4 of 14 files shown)


crates/contract/src/lib.rs
```diff
@@ -1322,6 +1325,40 @@ impl MpcContract {
...
+    /// Active phases:
+    /// - `Initializing` → uses proposed participants from generating_key
+    /// - `Running` → uses current active participants
+    /// - `Resharing` → uses new participants from resharing proposal
+    ///
+    /// Panics if:
+    /// - The protocol is not active (e.g., NotInitialized)
+    /// - The caller is not attested or not in the relevant participants set
+    pub fn assert_caller_is_attested_participant_and_protocol_active(&self) {
+        // Select participants based on the current protocol state
```
netrome: Nit: redundant comment

```suggestion
```

crates/contract/src/tee/tee_state.rs
```diff
@@ -338,30 +338,51 @@ impl TeeState {
...
-        // Check if caller matches an attested node
-        let result = self.participants_attestations.values().any(|(node_id, _)| {
-            match &node_id.account_public_key {
-                Some(pk) => pk == &signer_pk,
-                None => {
-                    // TODO (#823) Legacy fallback for mock nodes (no account_public_key) -remove when TEE is enforced
-                    node_id.account_id == signer_id
+
+        // find all participant entries that belong to the caller
+        for (account_id, _pid, participant_info) in participants.participants().iter() {
```
netrome: It feels like it would be sufficient to use `.iter().find(...)` here to simplify the code a bit and avoid some nesting.

crates/contract/src/tee/tee_state.rs
```diff
@@ -338,30 +338,51 @@ impl TeeState {
...
+            if account_id != &signer_id {
+                continue;
+            }
+
+            // Step 2: lookup attestation by TLS key (sign_pk)
```
netrome: Also a redundant comment

```suggestion
```

netrome (approved):
Given that this revealed a bug that we didn't test for, would it be possible to add a test case that covers this case (i.e. a non-attested participant attempts to send a CKD response)?

I'm not considering this a hard blocker right now since the code fixes a bug, but as a general principle I'd like us to have a more test-driven mindset.

crates/contract/src/tee/tee_state.rs
```diff
@@ -338,30 +338,51 @@ impl TeeState {
             .get(tls_public_key)
             .map(|(node_id, _)| node_id.clone())
     }
-    pub fn is_caller_an_attested_node(&self) -> bool {
+    /// Returns true if the caller has at least one participant entry
```
pbeza:
```suggestion
}

    /// Returns true if the `participants` have at least one participant entry
```

crates/contract/src/tee/tee_state.rs
```diff
@@ -338,30 +338,51 @@ impl TeeState {
...
-                Some(pk) => pk == &signer_pk,
-                None => {
-                    // TODO (#823) Legacy fallback for mock nodes (no account_public_key) -remove when TEE is enforced
-                    node_id.account_id == signer_id
+
+        // find all participant entries that belong to the caller
+        for (account_id, _pid, participant_info) in participants.participants().iter() {
+            if account_id != &signer_id {
+                continue;
+            }
```
pbeza: It’s probably [not possible](https://nearone.slack.com/archives/C0912BTG51T/p1760941606850299?thread_ts=1760289141.266809&cid=C0912BTG51T) to have two participants with the same `AccountID`, so we don’t need a loop here — just use `find()`.

crates/contract/src/lib.rs
```diff
@@ -1322,6 +1325,40 @@ impl MpcContract {
...
+                state.generating_key.proposed_parameters().participants()
+            }
+            ProtocolContractState::Running(state) => state.parameters.participants(),
+            ProtocolContractState::Resharing(state) => {
+                state.resharing_key.proposed_parameters().participants()
+            }
+            ProtocolContractState::NotInitialized => {
+                panic!("Protocol must be Initializing, Running, or Resharing to perform this operation");
+            }
+        };
```
kevindeforth: nit: might be nice to make add this to `state.rs` and make it a method of `ProtocolContracteState`.
https://github.com/near/mpc/blob/3d83f1f4867d2aa43eb274df2d809419e690a004/crates/contract/src/state.rs#L179

crates/contract/src/tee/tee_state.rs
```diff
@@ -338,30 +338,51 @@ impl TeeState {
...
     }
-    pub fn is_caller_an_attested_node(&self) -> bool {
+    /// Returns true if the caller has at least one participant entry
+    /// whose TLS key matches an attested node belonging to the caller account.
+    ///
+    /// Handles multiple participants per account and supports legacy mock nodes.
+    pub fn is_caller_an_attested_participant(&self, participants: &Participants) -> bool {
         let signer_pk = env::signer_account_pk();
         let signer_id = env::signer_account_id();
-        //let predecessor = env::predecessor_account_id(); //TODO (#1288)  check what is better predecessor or signer
```
kevindeforth: _note to any reviewers: will be handled in https://github.com/near/mpc/pull/1314_

kevindeforth (dismissed):
~I think we need to add the access key public key to the `ParticipantInfo` data. Otherwise, the best we can do is:~
- ~check if the caller has a valid attestation~
- ~check if there is a participant of same account id~

Actually, this is sufficient, because of the following:
a) we look up the participant info `participant_info` through the `predecessor_id` (signer_id, until #1314 is resolved)
b) we look up the attestation associated to the tls key stored in `participant_info`. If the attestation is found & valid, then the tls key is a unique identifier for the node.
c) we extract the account access key from the attestation and compare it with the `env::signer_pk()`.

barakeinav1:
> I think we need to add the access key public key to the `ParticipantInfo` data. Otherwise, the best we can do is:
> 
> * check if the caller has a valid attestation
> * check if there is a participant of same account id
> 
> return true if both are true.
> 
> But that doesn't mean that the caller node is also the participating node.

We also check TLS key:  

 if let Some((node_id, _attestation)) = self
                .participants_attestations
                .get(&participant_info.sign_pk).

PR #1317, opened 2025-10-20
feat: backup-cli http over mtls support

### LINKED ISSUE #1291: Backup Service: cli for HTTP requests with the node
### Background

With #1283, we added the required web server and client logic for the migration service. We should integrate that logic and hook it up with cli commands

### User Story

_No response_

### Acceptance Criteria

I’m able to fetch and store secrets from/to an MPC node using something like:

```
cargo run -p backup-cli -- get-keyshares
cargo run -p backup-cli -- put-keyshares
```

(modulo subparams or config that might be needed to connect to the MPC node)

In practice, one need to replace this dummy version of the P2P client:

https://github.com/near/mpc/blob/1bf863cdc8f2556dbbaee04b2a790c1822a8828e/crates/backup-cli/src/adapters.rs#L34-L46

- https://docs.google.com/document/d/12EawnD4273RSrT8Mzw1i-lxf3oln3vg9q0ojYoUJ3G8/edit?usp=sharing

gilcu3:
Closes #1291

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,8 +26,8 @@ contract_history = { path = "crates/contract_history" }
 mpc-contract = { path = "crates/contract", features = [
     "dev-utils",
 ] }
-mpc-primitives = { path = "crates/primitives", features = ["abi"] }
 mpc-node = { path = "crates/node" }
+mpc-primitives = { path = "crates/primitives", features = ["abi"] }
 mpc-tls = { path = "crates/tls" }
 node-types = { path = "crates/node-types" }
 tee_authority = { path = "crates/tee_authority" }
--- a/crates/backup-cli/Cargo.toml
+++ b/crates/backup-cli/Cargo.toml
@@ -5,8 +5,8 @@ edition = "2024"
 
 [dependencies]
 anyhow = { workspace = true }
-contract-interface = { workspace = true }
 clap = { workspace = true }
+contract-interface = { workspace = true }
 ed25519-dalek = { workspace = true }
 mpc-contract = { workspace = true, features = ["dev-utils"] }
 mpc-node = { workspace = true }


crates/backup-cli/src/types.rs
```diff
@@ -1,7 +1,9 @@
 use ed25519_dalek::SigningKey;
+use mpc_node::keyshare::Keyshare;
 use rand_core::CryptoRngCore;
 use serde::{Deserialize, Serialize};
-pub struct KeyShares {}
+
+pub struct KeyShares(pub Vec<Keyshare>);
```
gilcu3: Adding this instead of using simply `Vec<Keyshare>` made the code slightly worse in a few places, let me know if it is ok not to use this type

crates/backup-cli/src/adapters/contract_interface.rs
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+}
+
+impl ContractInterface for SimpleContractInterface {
+    type Error = Error;
+
+    async fn register_backup_data(&self, _public_key: &VerifyingKey) -> Result<(), Self::Error> {
+        Ok(())
+    }
```
netrome:
```suggestion
    async fn register_backup_data(&self, _public_key: &VerifyingKey) -> Result<(), Self::Error> {
        // TODO(https://github.com/near/mpc/issues/1290)
        unimplemented!()
    }
```

crates/backup-cli/src/adapters/contract_interface.rs
```diff
@@ -0,0 +1,81 @@
...
+    #[error("could not read from file: {0}")]
+    Read(tokio::io::Error),
+
+    #[error("failed to deserialize secrets")]
+    JsonDeserialization(serde_json::Error),
+}
+
+pub struct SimpleContractInterface {
+    contract_state_path: PathBuf,
+}
```
netrome: When do we need this type? I guess this is just a mock for tests, and not intended to be the "real" implementation right? In this case I'd expect this to be implemented in a test module or test helper module.

crates/backup-cli/src/adapters/p2p_client.rs
```diff
@@ -0,0 +1,142 @@
...
+    type Error = Error;
+
+    async fn get_keyshares(&self, keyset: &Keyset) -> Result<types::KeyShares, Self::Error> {
+        let mut send_request = client::connect_to_web_server(
+            &self.p2p_private_key,
+            &self.mpc_node_url,
+            &self.mpc_node_p2p_key,
+        )
+        .await
+        .map_err(Error::ServerConnection)?;
```
netrome: Nit: To be consistent with the previous review, we could technically improve our isolation of I/O here, but it would require refactoring `connect_to_web_server` and would be pretty complex, so I'd leave that as a follow-up exercise and not indulge in this exercise this time.

The cost is that our tests here actually will need to use I/O which means they can technically interfere or fail based on the host environment. However, that's not a big issue in practice and I would not expect any flakiness here.

crates/node/src/test_utils.rs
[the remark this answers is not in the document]
gilcu3: done https://github.com/near/mpc/issues/1318

crates/backup-cli/src/types.rs
gilcu3: It is quite simple indeed, but still would make it different from the rest of the implementation, and does not bring any gain at least right now. Anyhow, it should be very easy to put back in the future if we actually need something. The definition of this type will probably be defined only once in the node-types crate. At that moment it might be worth it to define a proper type.

crates/backup-cli/src/backup.rs
```diff
@@ -69,28 +107,52 @@ pub async fn register_backup_service(
...
+fn verifying_key_from_str(mpc_node_p2p_key: &str) -> VerifyingKey {
+    let mpc_node_p2p_key = contract_types::Ed25519PublicKey::from_str(mpc_node_p2p_key)
+        .expect("Invalid mpc_node_p2p_key value");
+    VerifyingKey::from_bytes(mpc_node_p2p_key.as_bytes()).expect("Invalid mpc_node_p2p_key value")
+}
+
+fn get_keyset_from_contract_state(contract_state: &ProtocolContractState) -> Keyset {
+    match contract_state {
+        ProtocolContractState::NotInitialized | ProtocolContractState::Resharing(_) => {
+            panic!("keyset not available in current contract state")
```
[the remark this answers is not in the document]
netrome: That's hard to know of course, but even if it's not I think the result makes reading it easier.

crates/backup-cli/src/types.rs
netrome: Fair enough.

crates/node/src/p2p.rs
```diff
@@ -557,6 +557,8 @@ pub mod testing {
         pub const MIGRATION_WEBSERVER_SUCCESS_TEST_GET_KEYSHARES: Self = Self(14);
         pub const MIGRATION_WEBSERVER_SUCCESS_TEST_SET_KEYSHARES: Self = Self(15);
         pub const MIGRATION_WEBSERVER_CHANGE_MIGRATION_INFO: Self = Self(16);
+        pub const BACKUP_CLI_WEBSERVER_GET_KEYSHARES: Self = Self(17);
+        pub const BACKUP_CLI_WEBSERVER_PUT_KEYSHARES: Self = Self(18);
```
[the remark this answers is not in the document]
netrome: Good point. Although this is another argument for avoiding system interactions in tests :sweat_smile:

crates/backup-cli/src/adapters/contract_interface.rs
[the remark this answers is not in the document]
netrome: Ah okay so you assume we manually take the state and write it to a file using the CLI (or just the explorer). Right yeah I recall us discussing that.

netrome (approved):
Nice stuff - thanks for the updates!

PR #1319, opened 2025-10-21
feat: mpc node spawns recovery web endpoint (1295)

### LINKED ISSUE #1295: Migration service: wire web-endpoints with node
### Background

With #1283 , we added the web-endpoints. Due to the size of the PR and to unblock ongoing works on the backup-service, we omitted wiring the migration-web-server logic with the MPC node.
We should do so. We might only be able to test in integration tests for the time being, but that should suffice until we can tackle #1294 

### User Story

_No response_

### Acceptance Criteria

Wire the web-endpoints with the node and write an integration test for importing and exporting secrets

_No response_

kevindeforth:
Resolves #1295 and concludes the onboarding logic on the MPC node side.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5083,13 +5083,13 @@ dependencies = [
  "mpc-contract",
  "mpc-tls",
  "near-client",
- "near-config-utils 2.8.0",
- "near-crypto 2.8.0",
+ "near-config-utils 2.9.0",
+ "near-crypto 2.9.0",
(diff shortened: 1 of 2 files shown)


netrome (changes requested):
~~Nit: I realized we need to update the start script to not break the TEE launcher for new nodes. Also we need to remember to update existing configs before upgrading.~~

Edit: I see you already did this.

PR #1339, opened 2025-10-29
feat: update to current ts version, more changes than I expected, some tricky deps changes as well

### LINKED ISSUE #1331: Modify API to match https://github.com/near/threshold-signatures/pull/180
### Background

Following the auditor's recommendation https://github.com/near/threshold-signatures/pull/180 now inputs the tweak in the HKDF function. This implied a change in the syntax of the rerandomization function.
We should be compatible with the new api

### User Story

_No response_

### Acceptance Criteria

Modify API to add the tweak when extracting the randomness during rerandomization

https://github.com/near/threshold-signatures/pull/180

gilcu3:
Closes #1331 

- BTW the huge change in Cargo.lock was triggered by some collision of serde's deps. It might be good to think about making the ts repo just another crate here, as conflicts like this, and API breaks, could happen at any point

### DIFF
--- a/crates/node/src/test_utils.rs
+++ b/crates/node/src/test_utils.rs
@@ -9,9 +9,11 @@ use threshold_signatures::ecdsa::ot_based_ecdsa::triples::TripleGenerationOutput
 use threshold_signatures::ecdsa::ot_based_ecdsa::PresignOutput;
 use threshold_signatures::ecdsa::ot_based_ecdsa::{PresignArguments, RerandomizedPresignOutput};
 use threshold_signatures::ecdsa::{RerandomizationArguments, Signature};
+use threshold_signatures::errors::ProtocolError;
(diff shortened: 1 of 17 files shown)


crates/node/src/test_utils.rs
```diff
@@ -23,6 +25,54 @@ pub struct TestGenerators {
...
 
+// +++++++++++++++++ Any Protocol +++++++++++++++++ //
+/// Run a protocol to completion, synchronously.
+///
+/// This works by executing each participant in order.
+///
+/// The reason this function exists is as a convenient testing utility.
+/// In practice each protocol participant is likely running on a different machine,
+/// and so orchestrating the protocol would happen differently.
+pub fn run_protocol<T>(
```
gilcu3: Adding this was also required for an unintended API break :| We thought the function was not being used

netrome (approved):
Thanks!

PR #1350, opened 2025-10-30
fix: Use port 24566 instead of 24567 for boot node arg in localnet guide

### LINKED ISSUE #1349: Use correct port in localnet guide
### Background

Currently the localnet guide still points to port 24567 for boot nodes, while the system has been updated to use 24566 during normal operations. We should update the guide to reflect this.

### User Story

As a developer I want accurate docs.

### Acceptance Criteria

The localnet guide uses the right port for the boot nodes.

_No response_

netrome:
closes #1349


gilcu3 (approved):
Thanks!

PR #1353, opened 2025-10-30
feat: ensure docker compose file is up to date in get docker compose hash

### LINKED ISSUE #1348: Ensure docker-compose file is up-to-date in `get_docker_compose_hash`
### Background

Currently, the docker compose file in `get_docker_compose_hash` looks like the following:

```rust
    pub fn get_docker_compose_hash(
        mpc_docker_image_hash: MpcDockerImageHash,
    ) -> LauncherDockerComposeHash {
        let filled_yaml = format!("version: '3.8'\n\nservices:\n  launcher:\n    image: barakeinavnear/launcher@sha256:1ea7571baf18bd052359abd2a1f269e7836f9bad2270eb55fc9475aa327f8d96\n\n # isuse #531: TODO (security): Replace with a specific image digest\n    container_name: launcher\n\n    environment:\n      - DOCKER_CONTENT_TRUST=1\n      - DEFAULT_IMAGE_DIGEST=sha256:{}\n\n    volumes:\n      - /var/run/docker.sock:/var/run/docker.sock\n      - /var/run/dstack.sock:/var/run/dstack.sock\n      - /tapp:/tapp:ro\n      - shared-volume:/mnt/shared:ro\n\n    security_opt:\n      - no-new-privileges:true\n\n    read_only: true\n\n    tmpfs:\n      - /tmp\n\nvolumes:\n  shared-volume:\n    name: shared-volume",
            mpc_docker_image_hash.as_hex()
        );
        let hash = sha256(filled_yaml.as_bytes());
        assert!(
            hash.len() == 32,
            "Docker compose hash must be 32 bytes long"
        );

        let mut hash_arr = [0u8; 32];
        hash_arr.copy_from_slice(&hash);

        LauncherDockerComposeHash::from(hash_arr)
    }
}
```

This should closely match the `/tee_launcher/launcher_docker_compose.yaml` file. However these have diverged right now as the launcher image differs between the two. Moreover, it's hard to read the current code and verify that those files are an exact match otherwise.

It would be better to include a template file in the repo which we include directly in the code (e.g. using `include_str!`, and use that template as a source of truth for the launcher docker compose file.

Secondly, since we have reproducible builds it should be possible to ensure this template always matches the current version of the code by including a verification step in CI that checks the hash.

### User Story

As a developer I'd like to ensure I can deploy the MPC contract with a sensible docker compose template that matches my expectations.

### Acceptance Criteria

We have a CI step that ensures that the `get_docker_compose_hash` function uses a compose file that corresponds to the current[^1] launcher image.

[^1]: Current as in, the launcher image you get when you build the launcher from the current commit.

_No response_

gilcu3:
Closes #1348

- CI worked on the [first try](https://github.com/near/mpc/actions/runs/18943715822/job/54089055394?pr=1353)! 
- CI with repro manifest worked in [20th try](https://github.com/near/mpc/actions/runs/18951849019/job/54118315202?pr=1353) :O 
- Aaand the [image published](https://hub.docker.com/layers/nearone/mpc-launcher/1348-ensure-docker-compose-file-is-up-to-date-in-get_docker_compose_hash-8c26295/images/sha256-bab4577e61bebcbcbed9fff22dd5fa741ded51465671638873af8a43e8f7373b) with the hardcoded hash

### DIFF
--- a/deployment/build-images.sh
+++ b/deployment/build-images.sh
@@ -134,7 +134,10 @@ if $USE_PUSH; then
     if $USE_LAUNCHER; then
         temp_dir=$(mktemp -d)
         echo "using $temp_dir"
+        # This compresses the built image to a local directory, which implicitly computes the manifest
+        # digest in $temp_dir/manifest.json
         skopeo copy --all --dest-compress docker-daemon:$LAUNCHER_IMAGE_NAME:latest dir:$temp_dir
+        # Then we publish the image from the directory, making sure the manifest digest does not change
         skopeo copy --preserve-digests dir:$temp_dir docker://docker.io/nearone/$LAUNCHER_IMAGE_NAME:$branch_name-$short_hash
     fi
 
--- a/scripts/build-and-verify-launcher-docker-image.sh
+++ b/scripts/build-and-verify-launcher-docker-image.sh
@@ -1,10 +1,12 @@
 #! /usr/bin/env bash
 
 ./deployment/build-images.sh --launcher
-# notice that any change to crates/contract/assets/launcher_docker_compose.yaml.template needs to be reflected here
+# Notice that any change to the format of crates/contract/assets/launcher_docker_compose.yaml.template needs to be reflected here
 published_manifest_hash=$(sed -n '5p' crates/contract/assets/launcher_docker_compose.yaml.template | grep -o '@sha256:.*' | cut -c 9-)
 temp_dir=$(mktemp -d)
 echo "using $temp_dir"
+# This compresses the built image to a local directory, which implicitly computes the manifest
+# digest in $temp_dir/manifest.json
 skopeo copy --all --dest-compress docker-daemon:mpc-launcher:latest dir:$temp_dir
 expected_manifest_hash=$(sha256sum $temp_dir/manifest.json | cut -d' ' -f1)
 if [ "${published_manifest_hash}" == "${expected_manifest_hash}" ]; then


.github/workflows/ci.yml
```diff
@@ -70,6 +70,34 @@ jobs:
...
+          expected_image_hash=$(docker inspect $published_image | jq -r .[0].Id)
+          if [ "${image_hash}" == "${expected_image_hash}" ]; then
+            echo "Launcher docker image hash verified"
+          else
+            echo "Launcher docker image hash verification failed"
+            echo "Got: ${image_hash}"
+            echo "Expected: ${expected_image_hash}"
+            exit 1
+          fi
+
```
gilcu3: When doing things like this I always wonder where is the border between github CI scripts, and bash scripts in the repo. Atm we have both kinds.

crates/contract/assets/launcher_docker_compose.yaml.template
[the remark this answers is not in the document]
netrome: I think this is a good place.

crates/contract/assets/launcher_docker_compose.yaml.template
```diff
@@ -0,0 +1,29 @@
+version: '3.8'
+
+services:
+  launcher:
+    image: nearone/mpc-launcher@sha256:45aa2d7bab715d62520be3fc560a0198bbe167be80211b22367bd1dd28919e8f
```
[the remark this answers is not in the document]
gilcu3: I am not 100% sure how docker and dockerhub handle this. Like if two identical images are published, I am not sure the manifest files will coincide. My guess is that they wouldn't, else making the manifest hash reproducible would have been possible, and I couldn't when I tried

.github/workflows/ci.yml
gilcu3: Good, at some point I might open issues + PR doing exactly that :)

.github/workflows/docker_launcher_release.yml
```diff
@@ -31,6 +31,11 @@ jobs:
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
+      
+      - name: Install skopeo
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y skopeo
```
netrome: Should we pin the version here somewhere to avoid time bombs?

gilcu3: as we discussed, version pinning for ubuntu packages might create a time bomb by itself, so leaving it as is

PR #1369, opened 2025-11-03
feat: More detailed error messages when attestation validation fails

### LINKED ISSUE #1356: More detailed error messages when attestation validation fails.
### Background

Currently when submitting an invalid remote attestation, we're getting this error in the transaction:

```
status: Failure(ActionError(ActionError { index: Some(0), kind: FunctionCallError(ExecutionError("Smart contract panicked: Invalid TEE Remote Attestation.: TeeQuoteStatus is invalid")) })),
```

This could mean anything from a signature validation error or the docker compose file mismatching the expected value, or the MPC docker image hash not existing.

This makes it very hard to troubleshoot the setup as a developer.
To improve the situation, we should ensure more informative error messages are returned when attestation verification fails.

### User Story

As a developer I'd like to be able to troubleshoot when my system is misconfigured.

### Acceptance Criteria

We provide different error messages when attestation verification fails for different reasons.

_No response_

netrome:
Closes #1356 

This was a bit more work than I anticipated, but now we should have much more clarity in why attestation submissions fail when they do.

Unfortunately I have not added tests for the new error scenarios, which ideally we should have. However, since there are no functional changes in this PR and due to our limited time frame for the soft launch I've chosen to de-prioritize this.

### DIFF
--- a/crates/attestation/src/attestation.rs
+++ b/crates/attestation/src/attestation.rs
@@ -442,10 +442,7 @@ impl Attestation {
             &tcb_info.compose_hash,
         )?;
 
-        Self::validate_app_compose_payload(
-            &app_compose_event.event_payload,
-            &tcb_info.app_compose,
-        )
+        Self::validate_app_compose_payload(&app_compose_event.event_payload, &tcb_info.app_compose)
     }
 
     /// Validates app compose configuration against expected security requirements.


crates/attestation/src/attestation.rs
```diff
@@ -39,6 +45,55 @@ pub struct DstackAttestation {
...
+    #[error("invalid event type {0}")]
+    InvalidEventType(u32),
+    #[error("failed to decode event digest `{0}`")]
+    EventDecoding(String),
+    #[error("failed to parse app compose JSON: {0}")]
+    AppComposeParsing(String),
+    #[error("no {0} event in event log")]
+    MissingEvent(&'static str),
+    #[error("duplicate {0} events in event log")]
+    DuplicateEvent(&'static str),
```
gilcu3: question: why choosing `&'static str` here instead of `String` like in many other cases?

crates/attestation/src/attestation.rs
```diff
@@ -287,51 +366,86 @@ impl Attestation {
...
+        let app_compose: AppCompose = serde_json::from_str(&tcb_info.app_compose)
+            .map_err(|e| VerificationError::AppComposeParsing(e.to_string()))?;
 
-        let mut events = tcb_info
-            .event_log
-            .iter()
-            .filter(|event| event.event == COMPOSE_HASH_EVENT && event.imr == RTMR3_INDEX);
-
-        let payload_is_correct = events.next().is_some_and(|event| {
-            event.event_payload == tcb_info.compose_hash
```
gilcu3: I think this line is no longer verified in the new implementation, at least not explicitly as it was before. We compare the hash in `validate_app_compose_payload`, but `event_payload` is not used as far I can see

crates/attestation/src/attestation.rs
```diff
@@ -287,51 +366,86 @@ impl Attestation {
...
+            &tcb_info.rtmr2,
+            &hex::encode(expected_measurements.rtmrs.rtmr2),
+        )?;
+        compare_hex_hashes(
+            "mtrd_tcb_info",
+            &tcb_info.mrtd,
+            &hex::encode(expected_measurements.rtmrs.mrtd),
+        )?;
+
+        Ok(())
```
gilcu3: same nit
```suggestion
        compare_hex_hashes(
            "mtrd_tcb_info",
            &tcb_info.mrtd,
            &hex::encode(expected_measurements.rtmrs.mrtd),
        )
```

crates/attestation/src/attestation.rs
```diff
@@ -147,57 +216,52 @@ impl Attestation {
...
+        self.verify_rtmr3(report_data, &attestation.tcb_info)?;
+        self.verify_app_compose(&attestation.tcb_info)?;
+        self.verify_local_sgx_digest(&attestation.tcb_info, &expected_measurements)?;
+        self.verify_mpc_hash(&attestation.tcb_info, allowed_mpc_docker_image_hashes)?;
+        self.verify_launcher_compose_hash(
+            &attestation.tcb_info,
+            allowed_launcher_docker_compose_hashes,
+        )?;
+
+        Ok(())
```
gilcu3:
```suggestion
        self.verify_launcher_compose_hash(
            &attestation.tcb_info,
            allowed_launcher_docker_compose_hashes,
        )
```

gilcu3 (approved):
This looks much better now, thank you!

Left mostly nits

crates/attestation/src/attestation.rs
netrome: Because these events are statically known and expected to be set when calling, whereas I've used `Strings` for nested error messages in other places. And because &'static str is more conservative than String, it's better to start with that imo.

PR #1406, opened 2025-11-07
test: improve unit tests

kevindeforth:
follow-up to https://github.com/near/mpc/pull/1405, https://github.com/near/mpc/pull/1403 and https://github.com/near/mpc/pull/1166

changes to some unit tests of TEE related logic:
- additional checks after voting for a code hash
- removal of cases for `test_latest_allowed_image_hash_is_written`. The cases don't add any value to the test since we removed the timestamp in https://github.com/near/mpc/pull/1166

context:
https://nearone.slack.com/archives/C07UW93JVQ8/p1762500941029699

### DIFF
--- a/crates/contract/tests/inprocess/expired_attestation.rs
+++ b/crates/contract/tests/inprocess/expired_attestation.rs
@@ -322,6 +322,19 @@ fn test_clean_tee_status_removes_non_participants() {
     assert_eq!(final_participant_count, PARTICIPANT_COUNT);
 }
 
+macro_rules! assert_allowed_code_hashes {
+    ($test_setup:expr, $blocktime_ns:expr, $expected_value:expr $(,)?) => {{
+        set_system_time($blocktime_ns);
+        let res: Vec<[u8; 32]> = $test_setup
+            .contract
+            .allowed_code_hashes()
+            .iter()
+            .map(|hash| *hash.clone())
+            .collect();
+        assert_eq!(res, $expected_value);
+    }};
+}
+
 /// **Test for grace-period expiry of older code hashes**
 ///
 /// Verifies that when participants vote for a new image hash, the older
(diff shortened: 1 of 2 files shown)


crates/contract/tests/inprocess/expired_attestation.rs
```diff
@@ -322,6 +322,19 @@ fn test_clean_tee_status_removes_non_participants() {
...
+        let res: Vec<[u8; 32]> = $test_setup
+            .contract
+            .allowed_code_hashes()
+            .iter()
+            .map(|hash| *hash.clone())
+            .collect();
+        assert_eq!(res, $expected_value);
+    }};
+}
+
```
netrome: What's the point of making this a macro as opposed to just having it be a normal function?

```suggestion
fn assert_allowed_code_hashes(
    test_setup: &TestSetup,
    blocktime_ns: u64,
    expected_value: &[[u8; 32]],
) {
    set_system_time(blocktime_ns);
    let res: Vec<[u8; 32]> = test_setup
        .contract
        .allowed_code_hashes()
        .iter()
        .map(|hash| *hash.clone())
        .collect();
    assert_eq!(res, expected_value);
}
```

netrome: Hmm interesting point. Fair enough :+1:

crates/node/src/tee/allowed_image_hashes_watcher.rs
```diff
@@ -212,53 +212,49 @@ mod tests {
...
-    ) {
-        let latest_allowed_image = allowed_images.last().unwrap().clone();
-        let current_image = image_hash_1();
-
-        let cancellation_token = CancellationToken::new();
-        let (sender, receiver) = watch::channel(allowed_images);
-        let (sender_shutdown, mut receiver_shutdown) = mpsc::channel(1);
-
-        let write_is_called = Arc::new(Notify::new());
+    async fn test_latest_allowed_image_hash_is_written() {
```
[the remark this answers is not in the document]
netrome: Yeah but functionally it's basically the same test right? Just that we had the explicit cases outlined before, and now we use a loop instead?

netrome:
> > I'm not convinced test_latest_allowed_image_hash_is_written is in a better state now than before.
> 
> oh, really? What is the benefit of keeping the test cases?

Yeah as I said in the discussion, it seems like a readability tradeoff of using the macro with the separate test cases or an internal loop. I think both have their minor advantages and drawbacks and are kind of equivalent at the end of the day.

Mainly just wanted to make sure there wasn't anything I missed here.

PR #1412, opened 2025-11-07
feat: validate attestation before submission

### LINKED ISSUE #1357: Validate attestation before submitting it to the MPC contract.
### Background

Currently our nodes does not log any detailed informations on transaction failures. Therefore, it is very painful to debug when something goes wrong. One risky operation is the submission of remote attestations. This transaction will fail if the attestation isn't valid, which becomes very tricky to debug if you don't have more information.

Therefore, the node should validate the attestation before submitting it. This way, if the attestation is invalid we'll be able to log more detailed information in the node directly to help any operator fix the problem.

### User Story

As a node operator I'd like to get detailed error logs if my node tries to submit invalid attestations.

### Acceptance Criteria

MPC nodes validate attestations before submitting them, and provide detailed error logs if this operation fails.

_No response_

gilcu3:
Closes #1357 

- ~~I still need to add tests~~
- The decision to add a new contract method is based on this [discussion](https://nearone.slack.com/archives/C07UW93JVQ8/p1762434622861229)
- I chose not to handle the unification of `periodic_attestation_submission` and `monitor_attestation_removal` to avoid making this too complex. I can handle that as a follow up

### DIFF
--- a/crates/contract/tests/sandbox/mod.rs
+++ b/crates/contract/tests/sandbox/mod.rs
@@ -4,6 +4,7 @@ pub mod sign;
 pub mod tee;
 pub mod upgrade_from_current_contract;
 pub mod upgrade_to_current_contract;
+pub mod user_views;
 pub mod vote;
 
 pub mod common;
--- a/crates/contract/tests/sandbox/user_views.rs
+++ b/crates/contract/tests/sandbox/user_views.rs
@@ -1,12 +1,13 @@
-pub mod common;
-use common::init_env_secp256k1;
+use mpc_contract::primitives::domain::SignatureScheme;
 use near_sdk::{CurveType, PublicKey};
 use serde_json::json;
 use std::str::FromStr;
 
+use crate::sandbox::common::init_env;
+
 #[tokio::test]
 async fn test_key_version() -> anyhow::Result<()> {
-    let (_, contract, _, _) = init_env_secp256k1(1).await;
+    let (_, contract, _, _) = init_env(&[SignatureScheme::Secp256k1]).await;
 
     let version: u32 = contract
         .view("latest_key_version")
@@ -21,7 +22,7 @@ async fn test_key_version() -> anyhow::Result<()> {
 
 #[tokio::test]
 async fn test_public_key() -> anyhow::Result<()> {
-    let (_, contract, _, _) = init_env_secp256k1(1).await;
+    let (_, contract, _, _) = init_env(&[SignatureScheme::Secp256k1]).await;
 
     let key: String = contract
         .view("public_key")
@@ -38,7 +39,7 @@ async fn test_public_key() -> anyhow::Result<()> {
 
 #[tokio::test]
 async fn test_derived_public_key() -> anyhow::Result<()> {
-    let (_, contract, _, _) = init_env_secp256k1(1).await;
+    let (_, contract, _, _) = init_env(&[SignatureScheme::Secp256k1]).await;
 
     let key: String = contract
         .view("derived_public_key")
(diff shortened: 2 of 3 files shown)


crates/node/src/tee/allowed_hashes_watcher.rs
```diff
@@ -182,10 +186,92 @@ where
...
+            self.receiver_from_contract.borrow_and_update().clone();
+
+        *self.allowed_hashes_in_contract.write().await = allowed_launcher_compose_hashes;
+
+        Ok(())
+    }
+}
+
+/// Send shutdown signal when watcher exits if
+/// it was not due to cancellation token being canceled
```
DSharifi:
```suggestion
/// it was not due to cancellation token being cancelled
```

crates/node/src/indexer/tee.rs
```diff
@@ -74,6 +77,64 @@ pub async fn monitor_allowed_docker_images(
     }
 }
 
+/// This future waits for the indexer to fully sync, and returns
+/// a [`watch::Receiver`] that will be continuously updated with the
+/// allowed [`LauncherDockerComposeHash`]es when a change is detected
+/// on the MPC smart contract.
+pub async fn monitor_allowed_launcher_compose_hashes(
```
DSharifi: There's quite some duplication with the `monitor_allowed_docker_images` which is defined right above. Could you refactor the loginc into a re-usable function which both methods can depend on?

crates/node/src/tee/allowed_hashes_watcher.rs
```diff
@@ -182,10 +186,92 @@ where
             tracing::error!("Currently running node image not in set of allowed image hashes.");
         }
 
+        *self.allowed_hashes_in_contract.write().await = allowed_image_hashes;
```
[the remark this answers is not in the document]
gilcu3: Can we use the watch receiver to receive values in two distinct places? Like now we need to receive them to write them on disk, but also to validate attestations.

crates/node/src/tee/remote_attestation.rs
```diff
@@ -85,25 +91,80 @@ pub async fn submit_remote_attestation(
...
+    )
+}
+
+pub async fn validate_and_submit_remote_attestation(
+    tx_sender: impl TransactionSender,
+    attestation: Attestation,
+    tls_public_key: Ed25519PublicKey,
+    account_public_key: Ed25519PublicKey,
+    allowed_image_hashes_in_contract: watch::Receiver<Vec<MpcDockerImageHash>>,
+    allowed_launcher_compose_hashes_in_contract: watch::Receiver<Vec<LauncherDockerComposeHash>>,
```
netrome: Nit: I don't see why this function needs to take a `warch::Receiver` as opposed to just receiving a reference to the latest values

gilcu3: I needed to select the proper place to convert from `watch::Receiver` to the reference to the latest values. The place I chose was basically `validate_and_submit_remote_attestation` function. What place do you suggest for doing that?

crates/node/src/tee/remote_attestation.rs
```diff
@@ -85,25 +91,80 @@ pub async fn submit_remote_attestation(
...
+    let allowed_launcher_compose_hashes_in_contract =
+        allowed_launcher_compose_hashes_in_contract.borrow().clone();
+
+    validate_remote_attestation(
+        &attestation,
+        tls_public_key.clone(),
+        account_public_key,
+        &allowed_image_hashes_in_contract,
+        &allowed_launcher_compose_hashes_in_contract,
+    )
```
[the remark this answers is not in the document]
gilcu3: I wonder where exactly would the program fail. It should be in the borrow call above, but afaik it would block, waiting for a value to be received, which is not optimal anyway. I will handle this and return a special result for that case.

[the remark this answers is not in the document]
gilcu3: I handled this in https://github.com/near/mpc/pull/1412/commits/db2ca7d1ac9e57efd44fa57f3da2bf0bf40ae22c when the validation fails because the lists are empty, we will see that in the logs

Issue #1408, opened 2025-11-07
Migration monitor sends updates on each block


netrome:
### Background

Currently we use the following loop to monitor and send migration state updates:

```rust
            loop {
                interval.tick().await;
                let response = fetch_migrations_once(indexer_state.clone()).await;
                tracing::debug!(target: "indexer", "fetched mpc migration state {:?}", response);
                migration_state_sender.send_if_modified(|watched_state| {
                    if *watched_state != response {
                        tracing::info!("contract migration state changed: {:?}", response);
                        *watched_state = response.clone();
                        true
                    } else {
                        false
                    }
                });
```

However, since the `response` contains the block height, this loop perceives it as the migration state changes on each new block. This is incorrect since we only want to monitor if the actual migration state has changed.

### User Story

As a developer I want correct code.

### Acceptance Criteria

The `monitor_migrations` function only sends updated states when the migration state has changed.

PR #1419, opened 2025-11-10
fix: compare only migration info in `monitor_migrations` function

pbeza:
Fixes https://github.com/near/mpc/issues/1408

### DIFF
--- a/crates/node/src/indexer/migrations.rs
+++ b/crates/node/src/indexer/migrations.rs
@@ -253,7 +253,6 @@ mod tests {
         let (my_account, my_key) = test_account_and_key();
 
         // When: Processing a response where my account is added
-        let new_block_height: u64 = 200u64;
         let expected_backup_service = BackupServiceInfo {
             public_key: my_key.into_contract_interface_type(),
         };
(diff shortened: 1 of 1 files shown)


crates/node/src/indexer/migrations.rs
```diff
@@ -99,3 +117,203 @@ async fn fetch_migrations_once(indexer_state: Arc<IndexerState>) -> (u64, Contra
...
+            contract_migration_sender,
+            mut contract_migration_receiver,
+            my_migration_sender,
+            mut my_migration_receiver,
+        ) = create_test_state(empty_migration_info);
+        let (my_account, my_key) = test_account_and_key();
+
+        // When: Processing a response where my account is added
+        let expected_backup_service = BackupServiceInfo {
+            public_key: Ed25519PublicKey(my_key.to_bytes()),
```
gilcu3: better to use `my_key.into_contract_interface_type()` for the conversion here

pbeza:
IIUC, from this:

> (...) Changing the log to "debug" and re-phrasing it sounds like an acceptable solution. Alternatively / additionally, we can also keep printing the current log, but only so in case the migration state has actually changed.

the main takeaway is that the channel should update on every block, but info logs should appear only when the migration state actually changes. If that’s correct, please take a look at the PR where I implemented this idea: https://github.com/near/mpc/pull/1441.

PR #1441, opened 2025-11-12
fix: reduce log noise in migration monitor task

### LINKED ISSUE #1408: Migration monitor sends updates on each block

pbeza:
Only log at `info` level when migration state actually changes, not on every block height update. Block height changes are still logged at `debug` level for monitoring purposes.

Fixes #1408

### DIFF
--- a/crates/node/src/indexer/migrations.rs
+++ b/crates/node/src/indexer/migrations.rs
@@ -45,10 +45,14 @@ pub async fn monitor_migrations(
             loop {
                 interval.tick().await;
                 let response = fetch_migrations_once(indexer_state.clone()).await;
-                tracing::debug!(target: "indexer", "fetched mpc migration state {:?}", response);
+                tracing::debug!(target: "indexer", "fetched mpc migration state at block {}: {:?}", response.0, response.1);
+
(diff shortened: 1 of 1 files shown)


pbeza:
Cross-linking the relevant thread for visibility: https://github.com/near/mpc/pull/1419#issuecomment-3521132117.

Issue #1447, opened 2025-11-12
Node fails to sync after delayed start (possible state sync / retention issue)


barakeinav1:
### Background

While testing the MPC node startup sequence, it was observed that if the node is not started immediately after the localnet setup, it fails to synchronize. The issue appears even though the latest fix from [PR #1414](https://github.com/near/mpc/pull/1414)
 is already included.
Logs indicate that the node syncs block headers but not actual blocks, resulting in errors such as:

`Failed to get balance. Waiting for sync? account=signer err=Account sam.test.near does not exist while viewing at block #0
error reading TEE accounts from chain: Account mpc-contract.test.near does not exist while viewing at block #0
INFO indexer: # 0 | Blocks processing: 0 | Blocks done: 0. Bps 0.00 b/s`

Discussion suggests this might be related to the default neard configuration, which retains only a limited number (3–5 epochs) of historical blocks, preventing older or delayed nodes from syncing.

possible fix solutions:

Increasing gc_num_epochs_to_keep in neard configuration.  (prefered)
Using a bucket/state-sync approach if full history is not retained.

### User Story

As an MPC node operator,
I want the node to be able to fully sync even if it starts after the network has progressed several epochs,
so that the system remains reliable and resilient to restarts or deployment delays.

### Acceptance Criteria

- [ ] A node started after some time (e.g., after multiple epochs) can still sync fully.
- [ ] 
- [ ]  The node successfully retrieves and processes all missing blocks, not just headers.
- [ ] 
- [ ]  Relevant configuration parameters (e.g., gc_num_epochs_to_keep, state_sync_enabled) are clearly documented and set correctly for MPC localnet setups.
- [ ] 
- [ ]  No “account does not exist at block #0” or similar errors appear in the logs during sync.
- [ ] 
- [ ]  Verified that enabling enable_debug_rpc or increasing history retention resolves or isolates the issue.
- [ ] 
- [ ]  Root cause (e.g., localnet history pruning, missing headers, port/config issues) identified and documented.

### Resources & Additional Notes

https://nearone.slack.com/archives/C07UW93JVQ8/p1762938765939719

PR #1448, opened 2025-11-12
fix(localnet): ensure MPC node can sync after delay by updating neard retention policy

barakeinav1:
fixes [#1447](https://github.com/near/mpc/issues/1447)
this will give us retention policy of 10000*5(minutes) =~34 days.

also enabling enable_debug_rpc - for better debug metrices


netrome (approved):

barakeinav1:
tested, and node are able to catch up even after 7 epochs

PR #1451, opened 2025-11-13
chore: update testnet contract

gilcu3:
Closes https://github.com/near/mpc/issues/1452


crates/contract/src/lib.rs
```diff
@@ -1601,7 +1603,7 @@ impl MpcContract {
 }
 
 fn try_state_read<T: borsh::BorshDeserialize>() -> Result<Option<T>, std::io::Error> {
-    env::storage_read(b"state")
+    env::storage_read(b"STATE")
```
gilcu3: ups yeah I had seen it as well, so shame on me

netrome: Shhhh :hushed:

PR #1478, opened 2025-11-17
fix: fetch MPC_SECRET_STORE_KEY

andrei-near:


netrome:
Done in https://github.com/near/mpc/pull/1480

PR #1480, opened 2025-11-17
fix: fetch mpc secret store key and add gcp image

### LINKED ISSUE #1479: Set SECRET_STORE_KEY_HEX if not already present in `start.sh`
### Background

From https://nearone.slack.com/archives/C090K7MC8CR/p1763392628828609

The testnet upgrade is failing because we're not setting the `SECRET_STORE_KEY_HEX` environment variable in some deployments.

From Andrei:

> we discussed this when dev cluster was updated so that SECRET_STORE_KEY_HEX is assigned the value of  GCP_LOCAL_ENCRYPTION_KEY_SECRET_ID or MPC_SECRET_STORE_KEY

### User Story

_No response_

### Acceptance Criteria

We set the appropriate value for the SECRET_STORE_KEY_HEX env variable if it's unset in `start.sh`

_No response_

gilcu3:
closes #1479

### DIFF
--- a/deployment/build-images.sh
+++ b/deployment/build-images.sh
@@ -143,10 +143,10 @@ if $USE_PUSH; then
     if [ -z "$branch_name" ]; then
         branch_name="detached"
     fi
-    sanitizded_branch_name="${branch_name//\//-}"
+    sanitized_branch_name="${branch_name//\//-}"
 
     short_hash=$(git rev-parse --short HEAD)
-    image_tag="$sanitizded_branch_name-$short_hash"
+    image_tag="$sanitized_branch_name-$short_hash"
     echo "Using branch-hash tag: $image_tag"
 
     if $USE_LAUNCHER; then


deployment/Dockerfile-node
```diff
@@ -1,19 +1,18 @@
-FROM debian:bookworm-slim@sha256:acd98e6cfc42813a4db9ca54ed79b6f702830bfc2fa43a2c2e87517371d82edb
+# FROM debian:bookworm-slim@sha256:acd98e6cfc42813a4db9ca54ed79b6f702830bfc2fa43a2c2e87517371d82edb
+FROM google/cloud-sdk:debian_component_based
```
netrome: Note for posterity: We need this image to support the `gcloud` command in `start.sh`

deployment/build-images.sh
```diff
@@ -72,6 +77,9 @@ fi
 DOCKERFILE_NODE=deployment/Dockerfile-node
 : "${NODE_IMAGE_NAME:=mpc-node}"
 
+DOCKERFILE_NODE_GCP=deployment/Dockerfile-node-gcp
+: "${NODE_GCP_IMAGE_NAME:=mpc-node-gcp}"
```
netrome: Hmm I'm a bit torn on whether we should have a separate image tag for this or not. But sure, if we're maintaining different images for now it does make sense. Let's keep this for now and get rid of it in a subsequent release.

netrome (approved):
Thanks, let's get this through and make a new patch release.

PR #1488, opened 2025-11-18
fix(contract):  fix ProposeUpdate vote method and add unit test

### LINKED ISSUE #1485: chore: add unit tests for contract update logic
### Background

The logic in [crates/contract/src/update.rs](https://github.com/near/mpc/blob/main/crates/contract/src/update.rs) is currently not unit-tested. The logic is only covered by high-level tests, asserting that we can upgrade the contract code. This means that it is easy to miss small inconsistencies in code.

For example, the method `vote` may lead to an inconsistent state in the `ProposedUpdate` struct: if a participant submits a vote for a non-existent `UpdateId`, then the pair `(AccountId, UpdateId)` will be tracked in the `vote_by_participant` map, but the vote will never be counted.
To keep a consistent state, `ProposedUpdate` currently relies on the caller reverting any changes in case `vote` returns `None`

https://github.com/near/mpc/blob/9c34614e87280932da9d6cdc3cee8537b6443b7a/crates/contract/src/update.rs#L137-L147

Writing unit tests can help to discover small inconsistencies like the above. Good unit tests often require that one can reason well about the behavior of a small unit of code.

### User Story

As a dev, I want test coverage.
As a dev, I want code that is easy to reason about.

### Acceptance Criteria

- Make the `vote` method above easy to reason about
- add unit test coverage for the `ProposedUpdates` struct. 

_No response_

kevindeforth:
resolves #1485

### DIFF
--- a/crates/contract/src/update.rs
+++ b/crates/contract/src/update.rs
@@ -246,7 +246,7 @@ mod tests {
         assert_eq!(found, expected);
     }
 
-    /// Ensures that `ProposedUpdates::propose()`:
+    /// Ensures that [`ProposedUpdates::propose`]:
     /// - returns the [`UpdateId`] of the proposed update
     /// - inserst the [`UpdateEntry`] into the entries map
     /// - does **not** record a vote for the newly proposed update
@@ -465,7 +465,7 @@ mod tests {
     impl TryFrom<&ProposedUpdates> for TestUpdateVotes {
         type Error = anyhow::Error;
 
-        /// converts `ProposedUpdates` to `TestUpdateVotes`. Fails if `ProposedUpdates` has a n
+        /// converts `ProposedUpdates` to `TestUpdateVotes`. Fails if `ProposedUpdates` has an
         /// inconsistent state.
         fn try_from(value: &ProposedUpdates) -> anyhow::Result<Self> {
             let id = value.id.0;


crates/contract/src/update.rs
```diff
@@ -175,6 +179,17 @@ impl ProposedUpdates {
         }
         Some(promise)
     }
+
+    /// Removes the vote for [`AccountId`]
+    pub fn remove_vote(&mut self, voter: &AccountId) {
+        if let Some(previous_id) = self.vote_by_participant.remove(voter) {
+            if let Some(entry) = self.entries.get_mut(&previous_id) {
+                entry.votes.remove(voter);
```
pbeza: Shouldn't we remove `entry` from `self.entries` if no votes remain after this removal? 🤔

crates/contract/src/update.rs
```diff
@@ -199,3 +214,289 @@ fn bytes_used(update: &Update) -> u128 {
...
+        let proposed_updates = ProposedUpdates::default();
+        let expected = TestUpdateVotes {
+            id: 0,
+            entries: BTreeMap::new(),
+        };
+        let found: TestUpdateVotes = (&proposed_updates).try_into().unwrap();
+        assert_eq!(found, expected);
+    }
+
+    /// Ensures that `ProposedUpdates::propose()`:
```
pbeza:
```suggestion
    /// Ensures that [`ProposedUpdates::propose`]:
```

crates/contract/src/update.rs
```diff
@@ -199,3 +214,289 @@ fn bytes_used(update: &Update) -> u128 {
...
+
+    /// Ensure that the default struct is empty
+    #[test]
+    fn test_proposed_updates_starts_empty() {
+        let proposed_updates = ProposedUpdates::default();
+        let expected = TestUpdateVotes {
+            id: 0,
+            entries: BTreeMap::new(),
+        };
+        let found: TestUpdateVotes = (&proposed_updates).try_into().unwrap();
```
pbeza: Nit: Initially, I was a bit confused about how `ProposedUpdates` gets transformed into `TestUpdateVotes` — I missed that you put `TryFrom` at the very end of the file. It also adds a bit of mental overhead to understand why that intermediate representation is needed, but I’m okay with it. Just sharing my thoughts.

crates/contract/src/update.rs
```diff
@@ -199,3 +214,289 @@ fn bytes_used(update: &Update) -> u128 {
...
+                    },
+                ),
+            ]),
+        };
+
+        let found: TestUpdateVotes = (&proposed_updates).try_into().unwrap();
+        assert_eq!(found, expected);
+    }
+
+    /// asserts that a vote for a non-existing update id by [`AccountId`] removes any previous vote by [`AccountId`]
```
pbeza: Should we document this behavior here?

https://github.com/near/mpc/blob/be7acd28c36a3c3d44ebad69c44d72614b578c5e/crates/contract/README.md?plain=1#L264

crates/contract/src/update.rs
```diff
@@ -175,6 +179,17 @@ impl ProposedUpdates {
...
         Some(promise)
     }
+
+    /// Removes the vote for [`AccountId`]
+    pub fn remove_vote(&mut self, voter: &AccountId) {
+        if let Some(previous_id) = self.vote_by_participant.remove(voter) {
+            if let Some(entry) = self.entries.get_mut(&previous_id) {
+                entry.votes.remove(voter);
+            } else {
+                env::log_str("inconsistent voting set");
```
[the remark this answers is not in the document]
kevindeforth: > Does the current implementation allow us to end up in this inconsistent state?

The contract endpoints prevent this for the existing code on main (c.f. also https://github.com/near/mpc/pull/1488#discussion_r2538616075)

With the implementation of `ProposedUpdates::vote` in this PR, we should never be in an inconsistent state.

PR #1490, opened 2025-11-18
fix: remove balance checks

### LINKED ISSUE #1489: Balance check fails on testnet
### Background

Several operators are seeing these errors:

```
2025-11-18T07:46:24.702976Z DEBUG mpc_node::indexer::lib: fetching account balance for <account>
2025-11-18T07:46:24.703139Z  INFO mpc_node::indexer::balances: Failed to get balance. Waiting for sync? account=signer err=The node does not track the shard ID 4
```

This is because our indexer does not track the shard where the MPC account lives in the typical scenario. Therefore we should remove this balance check.

### User Story

As a node operator I don't want a broken balance check.

### Acceptance Criteria

We've removed the balance check and associated metric since they are not working in production.

Note: The balance check was introduced in https://github.com/near/mpc/pull/776

gilcu3:
Closes #1489


gilcu3:
> Thanks! Feels unfortunate that we implemented and removed this, so it would be nice to figure out long term how to better cover the sharded setup in our system tests. But let's save that for a bit.

ups thanks for the review, but I thought this had been merged already :O the queue had defeated it :(

Issue #1269, opened 2025-10-12
re-enabled and Update tests that need data from a real MPC node


barakeinav1:
### Background

the following test need to be Update with real data from the MPC node running in TDX, and then  re-enabled.
test list:
test_from_str_valid

edit: The test update below moved  another task:
test_verify_method_signature (will be fix in another task
[#1493](https://github.com/near/mpc/issues/1493))

### Acceptance Criteria

the test in the list above are enabled in CI and are passing.

PR #1492, opened 2025-11-18
fix(test): enable and update test_from_str_valid

barakeinav1:
fixes [#1269](https://github.com/near/mpc/issues/1269)


gilcu3:
> It's also mentioned in the [acceptance criteria](https://github.com/near/mpc/issues/1269) of the task.

true, @barakeinav1 did you mean this as a partial solve? If that's the case, maybe you can reflect that, and avoid closing the issue

gilcu3 (approved):
Approving again!

PR #1506, opened 2025-11-20
feat: contract allows querying update proposals

### LINKED ISSUE #1415: Allow querying update proposals on the MPC contract
### Background

Currently when we upgrade the MPC contract it happens in two steps:

1. Someone proposes an update using the `propose_update` method. In doing so, they upload the bytecode of the new contract and receive an update ID.
2. The node operators uses the `vote_update` method to vote for the update. Once the update gets enough votes, the contract is upgraded.

However, there is an obvious problem here. `vote_update` only takes a single update ID as a parameter but there is currently no convenient way to

1. Query the set of existing proposals to spot that an update is proposed.
2. Validate the proposal before voting for it.

### User Story

As an operator I'd like to know what proposal I'm voting for.

### Acceptance Criteria

The contract exposes a method to query data about update proposals.

Note: The endpoint should not return the full bytecode for practical reasons. Just a sha256 digest would be enough.

kevindeforth:
Resolves #1415

### DIFF
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -2869,7 +2869,7 @@ mod tests {
         expected_votes
     }
 
-    fn test_proposed_updates_case(protocol_contract_state: ProtocolContractState) {
+    fn test_proposed_updates_case_given_state(protocol_contract_state: ProtocolContractState) {
         let mut contract = MpcContract::new_from_protocol_sate(protocol_contract_state);
 
         assert_eq!(contract.proposed_updates(), dtos::ProposedUpdates(vec![]));
@@ -2908,10 +2908,12 @@ mod tests {
             }
         };
         let mut expected = vec![code_update, config_update];
+        // sorting to have consistent order
         expected.sort();
 
         let mut res = contract.proposed_updates();
         res.0.iter_mut().for_each(|update| update.votes.sort());
+        // sorting to have consistent order
         res.0.sort();
 
         assert_eq!(dtos::ProposedUpdates(expected), res);
@@ -2920,19 +2922,19 @@ mod tests {
     #[test]
     pub fn test_proposed_updates_interface_running() {
         let protocol_contract_state = ProtocolContractState::Running(gen_running_state(2));
-        test_proposed_updates_case(protocol_contract_state);
+        test_proposed_updates_case_given_state(protocol_contract_state);
     }
 
     #[test]
     pub fn test_proposed_updates_interface_resharing() {
         let protocol_contract_state = ProtocolContractState::Resharing(gen_resharing_state(2).1);
-        test_proposed_updates_case(protocol_contract_state);
+        test_proposed_updates_case_given_state(protocol_contract_state);
     }
 
     #[test]
     pub fn test_proposed_updates_interface_initialzing() {
         let protocol_contract_state =
             ProtocolContractState::Initializing(gen_initializing_state(2, 1).1);
-        test_proposed_updates_case(protocol_contract_state);
+        test_proposed_updates_case_given_state(protocol_contract_state);
     }
 }


crates/contract-interface/src/types/updates.rs
```diff
@@ -0,0 +1,72 @@
...
+)]
+#[cfg_attr(
+    all(feature = "abi", not(target_arch = "wasm32")),
+    derive(schemars::JsonSchema)
+)]
+pub struct Update {
+    pub update_id: u64,
+    pub update_hash: UpdateHash,
+    pub votes: Vec<AccountId>,
+}
```
gilcu3: this is the dto type for one of the entries of  `entries: IterableMap<UpdateId, UpdateEntry>` where

```
struct UpdateEntry {
    update: Update,
    votes: HashSet<AccountId>,
    bytes_used: u128,
}
```

I guess the new names are better, but I am worried that the reason for the dto type to be different from the original gets lost.

crates/contract/src/lib.rs
```diff
@@ -2836,4 +2842,98 @@ mod tests {
...
+        let mut contract = MpcContract::new_from_protocol_sate(protocol_contract_state);
+
+        assert_eq!(contract.proposed_updates(), dtos::ProposedUpdates(vec![]));
+
+        let code_update = {
+            // propose update
+            let code = [0u8; 1000];
+            let hash = Sha256::digest(code);
+            let update = Update::Contract(code.into());
+            let expected_update_hash = dtos::UpdateHash::Code(hash.into());
```
gilcu3: orthogonal to this PR, but it would probably be much nicer if `Update::Contract` had a digest method instead of this, so that the code needed would be:
```suggestion
            let code = [0u8; 1000];
            let update = Update::Contract(code.into());
            let expected_update_hash = update.digest();
```

gilcu3 (dismissed):
Thank you, specially for the tests added!

I left some comments/questions, maybe helpful for future readers

crates/contract/src/lib.rs
kevindeforth: nice idea!

crates/contract/src/dto_mapping.rs
```diff
@@ -399,3 +403,38 @@ impl IntoInterfaceType<dtos::PublicKey> for &near_sdk::PublicKey {
...
+                update_hash: update.into_dto_type(),
+                votes: votes
+                    .iter()
+                    .map(|account_id| account_id.into_dto_type())
+                    .collect(),
+            })
+            .collect();
+        dtos::ProposedUpdates(updates)
+    }
+}
```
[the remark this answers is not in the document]
gilcu3: thanks for the thorough explanation!

crates/contract/src/lib.rs
```diff
@@ -2836,4 +2842,98 @@ mod tests {
...
+                update_hash: expected_update_hash,
+                votes: expected_votes,
+            }
+        };
+        let mut expected = vec![code_update, config_update];
+        expected.sort();
+
+        let mut res = contract.proposed_updates();
+        res.0.iter_mut().for_each(|update| update.votes.sort());
+        res.0.sort();
```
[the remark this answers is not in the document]
gilcu3: I think this is answered already in you previous answer, so marking as resolved

crates/contract/src/update.rs
```diff
@@ -567,4 +574,63 @@ mod tests {
...
+            expected_votes
+        };
+
+        let votes_1 = {
+            let account_id = gen_account_id();
+            let expected_votes = HashSet::from([account_id.clone()]);
+            let votes_1 = proposed_updates
+                .vote(&update_id_1, account_id.clone())
+                .unwrap();
+            assert_eq!(&HashSet::from([account_id.clone()]), votes_1);
```
pbeza:
```suggestion
            assert_eq!(&expected_votes, votes_1);
```

crates/contract/src/lib.rs
```diff
@@ -2836,4 +2842,98 @@ mod tests {
...
+    fn propose_and_vote(
+        contract: &mut MpcContract,
+        update: Update,
+        expected_update_id: u64,
+    ) -> Vec<dtos::AccountId> {
+        let update_id = contract.proposed_updates.propose(update.clone());
+        assert_eq!(update_id.0, expected_update_id);
+        // generate two accounts for voting
+        let account_id_0 = gen_account_id();
+        let account_id_1 = gen_account_id();
```
pbeza: Given the [discussion](https://nearone.slack.com/archives/C07UW93JVQ8/p1763644116745819?thread_ts=1763628576.510629&cid=C07UW93JVQ8) brought up by @gilcu3, it might be worth making the tests easily parameterized over the number of accounts, if possible. It would also reduce the copy-paste. Not pushing for it though, since we already have plenty of tests with a fixed number of participants/accounts/whatever.

pbeza (dismissed):
LGTM. Left a few nits.

Issue #1461, opened 2025-11-14
test_tee_cleanup_after_full_resharing_flow is flaky


gilcu3:
### Background

Example:

- https://github.com/near/mpc/actions/runs/19357024149/job/55380201686?pr=1459

this is the important part of the log:
```
ExecutionOutcome {
                transaction_hash: HfR9YHYdcssYd65jH8RK7E3tVSsovqZZqjdUUoshoRTZ,
                block_hash: 6oZ39eFCRN15bmbddt7iDkguzkzY18hkfPXZhcAtAVZk,
                logs: [
                    "clean_tee_status: signer=dev-20251114070655-47226657978817.test.near",
                ],
                receipt_ids: [
                    7sFBM9ZRWcqV2iMmPgVnjUZeu5Gn4gzihiPfie4ZLqZi,
                ],
                gas_burnt: NearGas {
                    inner: 3888095274944,
                },
                tokens_burnt: NearToken {
                    inner: 388809527494400000000,
                },
                executor_id: AccountId(
                    "dev-20251114070652-75601919128858.test.near",
                ),
                status: Failure(ActionError(ActionError { index: Some(0), kind: FunctionCallError(ExecutionError("Exceeded the prepaid gas.")) })),
            }
```

### User Story

No flaky tests please

### Acceptance Criteria

The test is no longer flaky.

PR #1510, opened 2025-11-20
feat: sandbox tests support for any number of participants

### LINKED ISSUE #1508: sandbox tests support for any number of participants
### Background

Currently many of the sandbox tests in the contract have hardcoded values which do not allow their correct usage after `PARTICIPANT_LEN` value is changed

### User Story

test should be able to scale to any number of participants

### Acceptance Criteria

The tests are adapted so that testing with any number of participant is possible.

[long context](https://nearone.slack.com/archives/C07UW93JVQ8/p1763628576510629)

gilcu3:
Closes #1508 

- For some tests, I could not make it work with `PARTICIPANT_LEN = 10` locally, there are to follow-up issues: #1518 and #1461
- For some did not make sense to scale them, they were thought to test special behaviors.
- Currently, all tests using the global const  `PARTICIPANT_LEN` did scale properly, the rest did not. I did not change the variable anyway, to avoid CI becoming too slow.

For posterity, in [40da56a](https://github.com/near/mpc/pull/1510/commits/40da56ae89075281ef0eafeb2f082fe1cd285a55) I left the `PARTICIPANT_LEN = 10`, the result is in [this run](https://github.com/near/mpc/actions/runs/19567239865/job/56032133163) (it was only 7 minutes, so a bit slower only, I think we could make it the new default!)


crates/contract/tests/sandbox/common.rs
```diff
@@ -919,6 +915,17 @@ pub async fn assert_running_return_participants(
...
+pub async fn assert_running_return_threshold(contract: &Contract) -> anyhow::Result<Threshold> {
+    let final_state: ProtocolContractState = contract.view("state").await?.json()?;
+    let ProtocolContractState::Running(running_state) = final_state else {
+        panic!(
+            "Expected contract to be in Running state: {:?}",
+            final_state
+        );
+    };
+    Ok(running_state.parameters.threshold())
+}
```
kevindeforth: Why have an error path for this return value? Is there any place where we would not want to panic if `view("state")` fails?

pbeza (approved):

PR #1525, opened 2025-11-22
fix: Use patched near core supporting reproducible builds

### LINKED ISSUE #1524: Node images doesn't build with nearcore 2.10.0-rc.3
### Background

See failing workflow here: https://github.com/near/mpc/actions/runs/19594360580/job/56117201009

This also fails locally for me with the same error.

After our recent changes, it seems like we cannot build the node images anymore as they now seem to try to use `curl` which naturally is something we don't have in our reproducible environment.

We need to look into how to fix so we can build our node images again.

### User Story

As an MPC developer I want our builds to succeed.

### Acceptance Criteria

We can build our node image again.

_No response_

netrome:
fixes #1524 (I hope :crossed_fingers:)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4937,7 +4937,7 @@ dependencies = [
 [[package]]
 name = "near-async"
 version = "2.10.0-rc.3"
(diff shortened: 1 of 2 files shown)


netrome:
This worked locally. Building image with github action as well to verify before merging: https://github.com/near/mpc/actions/runs/19595942036

Issue #765, opened 2025-08-04
bug: MPC writes to image-digest.bin is not atomic


barakeinav1:
### Background

In case the MPC node writes to image-digest.bin and there is a power lost. the  image-digest.bin will contain a value that will not be accepted by the launcher on next boot.
currently there is no recovery for the node from this case.

possible solutions
-create our own backup file
-use OS provided atomic writes

### Acceptance Criteria
We have decided if want accept this risk or find a solution. In the latter case, the solution is either implemented or deferred to a follow-up issue.

PR #1527, opened 2025-11-23
feat(launcher): add support for multiple MPC hashes with fallback logic

### LINKED ISSUE #1475: MPC network becomes unrecoverable after voting in a bad code hash
### Background

During localnet testing of the MPC node update flow, we discovered a scenario where the MPC network becomes **permanently bricked** if a voted `code_hash` corresponds to a malfunctioning or invalid MPC node image.

Sequence:

1. Start with a working MPC network.
2. Vote for a new `code_hash` that corresponds to a **broken/non‑bootable image**.
3. All existing nodes detect the update and **persist the new hash** to disk.
4. After restart, the launcher attempts to load the MPC image for that hash.
5. The image fails to load → MPC node never comes up.
6. No nodes remain operational, so **no one can vote in another code hash**.
7. The network becomes stuck unless manually repaired on disk.

### User Story

**As an operator**  
I want MPC nodes to validate a newly voted code hash **before committing it**  
so the network cannot be bricked by a malformed or broken image.

### Acceptance Criteria

A design for solving the issue above is made.
a follow up task for execution/code/documentation are made.

[slack](https://nearone.slack.com/archives/C07UW93JVQ8/p1763375192810089)

### LINKED ISSUE #1541: Fix used variables in laucnher.py
### Background

in the laucnher.py code below 
`int(os.environ.get(OS_ENV_VAR_RPC_REQUST_TIMEOUT_SECS, "10"))`   - is not used.

in request_until_success the code passes

`rpc_request_timeout_secs=rpc_request_interval_secs,
`

 maybe some oversight and it should be fixed.

Need to review and fix accordingly.

```

 # RPC timing configuration
    int(os.environ.get(OS_ENV_VAR_RPC_REQUST_TIMEOUT_SECS, "10"))
    rpc_request_interval_ms = int(
        os.environ.get(OS_ENV_VAR_RPC_REQUEST_INTERVAL_MS, "1000")
    )
    rpc_request_interval_secs = rpc_request_interval_ms / 1000.0
    rpc_max_attempts = int(os.environ.get(OS_ENV_VAR_RPC_MAX_ATTEMPTS, "20"))
```

### User Story

as a developer Code should work as intended, and be clear

### Acceptance Criteria

fix launcher code.
test as needed.

_No response_

barakeinav1:
fixes #1475
fixes #1541
fixes [#765](https://github.com/near/mpc/issues/765)
testing WIP, can be found https://github.com/near/mpc-private/pull/79

CI test that is failing, this is due to launcher measurement update that is needed due to the launcher code update.
Since, this requires updates to several files, I'll do it in this PR, after the main review is completed.

### DIFF
--- a/crates/contract/assets/launcher_docker_compose.yaml.template
+++ b/crates/contract/assets/launcher_docker_compose.yaml.template
@@ -2,7 +2,7 @@ version: '3.8'
 
 services:
   launcher:
-    image: nearone/mpc-launcher@sha256:6e3d8048150494e6f96bb603bd5ad9420e10b5cc5ec3e965b28703e974bd03f5
+    image: nearone/mpc-launcher@sha256:d8f5f45aadc54e5ddc0b4109c75de67ab62ac6eb1726c026070452691b9035f9
 
     container_name: launcher
 
--- a/tee_launcher/launcher_docker_compose.yaml
+++ b/tee_launcher/launcher_docker_compose.yaml
@@ -2,7 +2,7 @@ version: '3.8'
 
 services:
   launcher:
-    image: nearone/mpc-launcher@sha256:6e3d8048150494e6f96bb603bd5ad9420e10b5cc5ec3e965b28703e974bd03f5
+    image: nearone/mpc-launcher@sha256:d8f5f45aadc54e5ddc0b4109c75de67ab62ac6eb1726c026070452691b9035f9
 
     container_name: launcher
 


crates/node/src/tee/allowed_image_hashes_watcher.rs
```diff
@@ -16,12 +16,11 @@ use tokio_util::sync::CancellationToken;
 
 #[cfg(test)]
 use mockall::automock;
-
 #[cfg_attr(test, automock)]
 pub trait AllowedImageHashesStorage {
     fn set(
         &mut self,
-        latest_allowed_image_hash: &MpcDockerImageHash,
+        approved_hashes: Vec<MpcDockerImageHash>,
```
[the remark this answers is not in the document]
barakeinav1: I agree that taking a slice (&[MpcDockerImageHash]) would normally be more idiomatic, this was the first approach.
But then I discovered that this method is part of a trait that is mocked in existing unit tests using mockall, and mockall has a known limitation:

I got this error:
`Mockall cannot generate mocks for async trait methods that take borrowed parameters (&[T]).`

So unless, you think it is very important, I'd prefer to keep this instead of refactoring the existing unitests.

crates/node/src/tee/allowed_image_hashes_watcher.rs
```diff
@@ -31,39 +30,37 @@ pub struct AllowedImageHashesFile {
...
+            .collect();
 
-        tracing::info!(
-            ?self.file_path,
-            ?latest_allowed_image_hash,
-            "Writing latest allowed image hash to disk."
-        );
+        let json = serde_json::json!({
+            "approved_hashes": hash_strings
+        });
```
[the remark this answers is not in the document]
barakeinav1: This format should be shared between Launcher and MPC.
I think defining a shared json format is easier and more intuitive in this case.

I added some comments and tests in both MPC/Launcher - to catch any change.

tee_launcher/launcher.py
```diff
@@ -246,13 +256,186 @@ def get_image_spec(dstack_config: dict[str, str]) -> ImageSpec:
...
+            return False
+
+        return True
+
+    except Exception as e:
+        logging.error(f"Validation failed for {image_digest}: {e}")
+        return False
+
+
+def select_valid_hash(
```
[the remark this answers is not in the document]
barakeinav1: done

tee_launcher/launcher.py
```diff
@@ -306,96 +491,37 @@ def main():
...
-    proc = run(docker_cmd)
-
-    if proc.returncode:
-        raise RuntimeError("docker run non-zero exit code %d", proc.returncode)
+    # --------------------------------------------------------
+    # PHASE 2 + 3:
+    # - Extend RTMR3 with validated hash
+    # - Launch MPC container
+    # --------------------------------------------------------
+    extend_rtmr3_and_launch(valid_hash, dstack_config)
```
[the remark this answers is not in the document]
barakeinav1: it is an over the very very long main we had before :-) 
but I fully agree, I'll break it into separate functions

crates/node/src/tee/allowed_image_hashes_watcher.rs
```diff
@@ -30,40 +29,49 @@ pub struct AllowedImageHashesFile {
     file_path: PathBuf,
 }
 
+// important: must stay aligned with the launcher implementation in:
+// mpc/tee_launcher/launcher.py
+const JSON_KEY_APPROVED_HASHES: &str = "approved_hashes";
```
DSharifi: Does it make sense to have this constant in a config file that they both can share? Otherwise, why can't we make it simpler by just having a line separated list of hashes instead of a JSON object?

crates/node/src/tee/allowed_image_hashes_watcher.rs
```diff
@@ -212,12 +219,16 @@ mod tests {
...
-    /// Assert that the image with highest block height is always written to storage.
+    /// Ensures that whenever the allowed image hash list changes,
+    /// the MPC node writes the full list of allowed hashes to storage,
+    /// preserving the ordering received from the contract.
+    ///
+    /// The contract always sends hashes oldest → newest, and the watcher
+    /// must pass the entire vector untouched to the storage backend.
     #[rstest]
     #[tokio::test]
     async fn test_latest_allowed_image_hash_is_written() {
```
DSharifi:
```suggestion
    async fn test_allowed_image_hashes_are_written() {
```

netrome (approved):
Thanks for the updates. I still would prefer to use structs to declare the json schema in Rust. In Python I think you can use dataclasses for this, but the extraction code was so small I don't care as much of it on that side.

However this is not a blocker if you disagree.

DSharifi (dismissed):

crates/node/src/tee/allowed_image_hashes_watcher.rs
barakeinav1: >>Does it make sense to have this constant in a config file that they both can share?

ideally, yes it should be shared. 
but since we have test on both side (python,rust) to catch any change, this is good enough.

>>otherwise, why can't we make it simpler by just having a line separated list of hashes instead of a JSON object?

We can. but then you could argue that there could be parsing errors. so we should use a standard format and parser (like json?  :-) )

tee_launcher/launcher.py
```diff
@@ -246,13 +261,192 @@ def get_image_spec(dstack_config: dict[str, str]) -> ImageSpec:
...
+        raise RuntimeError(f"Failed to parse {IMAGE_DIGEST_FILE}: {e}")
+
+    hashes = data.get(JSON_KEY_APPROVED_HASHES)
+    if not isinstance(hashes, list) or not hashes:
+        raise RuntimeError(
+            f"Invalid JSON in {IMAGE_DIGEST_FILE}: approved_hashes missing or empty"
+        )
+
+    # Contract writes oldest → newest; we need newest → oldest
+    reversed_hashes = list(reversed(hashes))
```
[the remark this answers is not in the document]
DSharifi: When designing an API, you design it for simplicity and for use case for the user of the API. Not for making it convenient for the implementer.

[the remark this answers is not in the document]
barakeinav1: MPC node<->launcher is an internal interface between them.
MPC contract API is public. This may break something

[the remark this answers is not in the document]
barakeinav1: In the contract, I can update those 2 APIs:
I can change the order here easily. 

Bthen, There is a reverse order between the external API and the internal TEE state:
If you are ok with this, I'll go ahead and make the change
changing the TEE state is much more work, and I don't' think we should do it.

Personally, I still think that just changing the MPC node, (saving in reverse order ) is totally fine (and less riskier than changing the contract API)


```
pub fn allowed_code_hashes(&self) -> Vec<MpcDockerImageHash> {
        log!("allowed_code_hashes: signer={}", env::signer_account_id());

        let tee_upgrade_deadline_duration =
            Duration::from_secs(self.config.tee_upgrade_deadline_duration_seconds);

        let mut hashes = self.tee_state
            .get_allowed_mpc_docker_image_hashes(tee_upgrade_deadline_duration)
            
            //reverse code here
            hashes.reverse();    // ← reverse in-place
             hashes
    }

    pub fn latest_code_hash(&mut self) -> MpcDockerImageHash {
        log!("latest_code_hash: signer={}", env::signer_account_id());

        let tee_upgrade_deadline_duration =
            Duration::from_secs(self.config.tee_upgrade_deadline_duration_seconds);

        self.tee_state
            .get_allowed_mpc_docker_image_hashes(tee_upgrade_deadline_duration)
            .first()
            .expect("there must be at least one allowed code hash")
            .clone()
    }
```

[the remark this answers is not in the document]
DSharifi: Thanks for outlining the diff.

> Bthen, There is a reverse order between the external API and the internal TEE state:

Yes, it is totally fine if the order are different in the internal structure of the contract and the API it exposes. **We don't want our APIs to be tightly coupled to their implementation details.**

There is no risk to change the contract API for `allowed_code_hashes`, as this API is not used in production.

DSharifi (dismissed):

tee_launcher/test_launcher_config.py
```diff
@@ -372,4 +373,12 @@ def test_ld_preload_in_ports2():
...
-# Additional tests could go here to check other edge cases
+def test_json_key_matches_node():
+    """
+    Ensure the JSON key used by the launcher to read approved image hashes
+    stays aligned with the MPC node implementation.
+
+    If this test fails, it means the launcher and MPC node are using different
+    JSON field names, which would break MPC hash propagation.
+    """
+    assert launcher.JSON_KEY_APPROVED_HASHES == "approved_hashes"
```
gilcu3: I think having the same comment above where `JSON_KEY_APPROVED_HASHES` is defined would have the same effect, and will avoid any surprising test failure. This does not really guarantee that it is in sync with the MPC node, so a test does not help here.

gilcu3 (commented):
Left some comments, I can review again tomorrow after the changes we agreed upon are done

tee_launcher/test_launcher_config.py
barakeinav1: better safe than sorry.
I added reference to the MPC node code

tee_launcher/launcher.py
```diff
@@ -246,13 +261,190 @@ def get_image_spec(dstack_config: dict[str, str]) -> ImageSpec:
...
+        raise RuntimeError(f"Failed to parse {IMAGE_DIGEST_FILE}: {e}")
+
+    # Extract approved_hashes, order newest → oldest
+    hashes = data.get(JSON_KEY_APPROVED_HASHES)
+    if not isinstance(hashes, list) or not hashes:
+        raise RuntimeError(
+            f"Invalid JSON in {IMAGE_DIGEST_FILE}: approved_hashes missing or empty"
+        )
+
+    SHA256_REGEX = re.compile(r"^sha256:[0-9a-f]{64}$")
```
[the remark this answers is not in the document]
barakeinav1: ok

tee_launcher/launcher.py
```diff
@@ -246,13 +261,190 @@ def get_image_spec(dstack_config: dict[str, str]) -> ImageSpec:
...
-        return open(IMAGE_DIGEST_FILE).readline().strip()
+def load_approved_hashes(dstack_config: dict) -> list[str]:
+    """
+    Load approved MPC image hashes from JSON file.
+
+    Contract + MPC Node write them oldest → newest:
+        ["sha256:h1", "sha256:h2", "sha256:h3"]
+
+    Launcher must try newest → oldest:
+        ["sha256:h3", "sha256:h2", "sha256:h1"]
```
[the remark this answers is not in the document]
barakeinav1: refactored

DSharifi (dismissed):

crates/node/src/tee/allowed_image_hashes_watcher.rs
```diff
@@ -338,31 +362,48 @@ mod tests {
             Err(TryRecvError::Empty),
             "Shutdown signal was sent unexpectedly."
         );
-        let event_loop_is_alive = !receiver_shutdown.is_closed();
-        assert!(event_loop_is_alive, "Event loop should be running.");
+
+        assert!(
+            !receiver_shutdown.is_closed(),
+            "Event loop should be running."
+        );
```
gilcu3: I think the previous state is better, as it is more informative

crates/node/src/tee/allowed_image_hashes_watcher.rs
```diff
@@ -371,15 +412,106 @@ mod tests {
...
+        assert_matches!(
+            receiver_shutdown.try_recv(),
+            Err(TryRecvError::Empty),
+            "Shutdown should NOT be sent when list is empty"
+        );
+    }
+    #[test]
+    fn test_json_key_matches_launcher() {
+        assert_eq!(JSON_KEY_APPROVED_HASHES, "approved_hashes");
+    }
```
gilcu3: same as in pytests, I think this test does not add any value. The comment above the constant already does this. But feel free to disregard this, anyway the test is not expensive :)

tee_launcher/test_launcher_config.py
```diff
@@ -372,4 +386,80 @@ def test_ld_preload_in_ports2():
...
+
+def test_override_present(monkeypatch):
+    override_value = "sha256:" + "a"*64
+    approved = [override_value, "sha256:" + "b"*64]
+
+    fake_json = make_digest_json(approved)
+
+    monkeypatch.setattr("tee_launcher.launcher.os.path.isfile", lambda _: True)
+    monkeypatch.setattr("builtins.open", mock_open(read_data=fake_json))
```
gilcu3: I was thinking to use this the other day, didn't think would see it so soon, nice!

tee_launcher/launcher.py
```diff
@@ -246,13 +262,167 @@ def get_image_spec(dstack_config: dict[str, str]) -> ImageSpec:
...
+    return selected
+
+
+def validate_image_hash(
+    image_digest: str,
+    dstack_config: dict,
+    rpc_request_timeout_secs: float,
+    rpc_request_interval_secs: float,
+    rpc_max_attempts: int,
+) -> bool:
```
gilcu3: can we add a test for this function?

tee_launcher/launcher.py
```diff
@@ -246,13 +262,167 @@ def get_image_spec(dstack_config: dict[str, str]) -> ImageSpec:
...
+        logging.info(f"MPC hash {image_digest} validated successfully.")
+        return True
+
+    except Exception as e:
+        logging.error(f"Validation failed for {image_digest}: {e}")
+        return False
+
+
+def extend_rtmr3(valid_hash: str) -> None:
+    full, bare = split_digest(valid_hash)
```
gilcu3: same, I think this function deserves a test

tee_launcher/launcher.py
```diff
@@ -306,103 +478,42 @@ def main():
...
-
-    if proc.returncode:
-        raise RuntimeError(
-            f"docker image inspect returned non-zero exit code {proc.returncode}:\n{proc.stderr}"
-        )
 
-    pulled_image_digest = proc.stdout.decode("utf-8").strip()
+    # Choose exactly one allowed hash (override or newest)
+    selected_hash = load_and_select_hash(dstack_config)
+    logging.info(f"Selected MPC image hash: {selected_hash}")
```
gilcu3: this log is repeated. The same message appears at the end of function `load_and_select_hash`, therefore please remove one of the two

tee_launcher/launcher.py
```diff
@@ -547,27 +663,24 @@ def build_docker_cmd(user_env: dict[str, str], image_digest: str) -> list[str]:
...
+                if is_safe_port_mapping(clean) and is_valid_port_mapping(clean):
+                    docker_cmd += ["-p", clean]
                 else:
-                    logging.warning(
-                        f"Ignoring invalid or unsafe PORTS entry: {clean_port}"
-                    )
+                    logging.warning(f"Ignoring invalid or unsafe PORTS entry: {clean}")
         else:
-            logging.warning(f"Ignoring unauthorized environment variable: {key}")
+            logging.info(f"Ignoring non-MPC variable: {key}")
```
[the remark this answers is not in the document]
gilcu3: I don't get it then. Is this something that cannot happen then? I traced where it comes from, and it is from a file that can be edited by the user. So certainly looks like a warning is more appropriate, as it would mean user mistakes there

tee_launcher/launcher.py
barakeinav1: correct

tee_launcher/launcher.py
```diff
@@ -547,27 +663,24 @@ def build_docker_cmd(user_env: dict[str, str], image_digest: str) -> list[str]:
...
-                clean_host = host_entry.strip()
-                if is_safe_host_entry(clean_host) and is_valid_host_entry(clean_host):
-                    docker_cmd += ["--add-host", clean_host]
+                clean = host_entry.strip()
+                if is_safe_host_entry(clean) and is_valid_host_entry(clean):
+                    docker_cmd += ["--add-host", clean]
                 else:
                     logging.warning(
-                        f"Ignoring invalid or unsafe EXTRA_HOSTS entry: {clean_host}"
+                        f"Ignoring invalid or unsafe EXTRA_HOSTS entry: {clean}"
```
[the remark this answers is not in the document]
barakeinav1: agree.

tee_launcher/launcher.py
```diff
@@ -515,30 +629,32 @@ def get_manifest_digest(
...
+def split_digest(d: str) -> tuple[str, str]:
+    """
+    Returns (full_digest, bare_digest).
+    full_digest: 'sha256:abcd...'
+    bare_digest: 'abcd...'
+    """
+    if not d.startswith("sha256:"):
+        raise ValueError(f"Invalid digest (missing sha256: prefix): {d}")
+
+    return d, d.split(":", 1)[1]
```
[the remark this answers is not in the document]
gilcu3: yes both are equivalent, but my proposal avoids having two variables with different names and equal value and semantics later

tee_launcher/launcher.py
gilcu3: We agreed that this could be handled in a follow-up, that uses a small test image published in dockerhub that guarantees our logic is consistent without manual testing

gilcu3: the follow up issue https://github.com/near/mpc/issues/1567

barakeinav1: added this task Add pytest for launcher.validate_image_hash[#1567](https://github.com/near/mpc/issues/1567)
need to connect to dockerhub (causes a dependency in our CI)

PR #1537, opened 2025-11-24
feat: allow participants to withdraw their update vote

### LINKED ISSUE #1421: mpc contract: allow participants to remove update votes
### Background

Currently, we do not allow participants to withdraw a vote for an update.

### User Story

as a node operator, I want the ability to remove a vote for an update

### Acceptance Criteria

The smart contract exposes an endpoint that allows the removal of an update vote.

_No response_

kevindeforth:
resolves #1421

### DIFF
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -981,7 +981,7 @@ impl MpcContract {
     }
 
     /// Removes an update vote by the caller
-    /// panics if the account is not in a running state or if the caller is not a participant
+    /// panics if the contract is not in a running state or if the caller is not a participant
     pub fn remove_update_vote(&mut self) {
         log!("remove_update_vote: signer={}", env::signer_account_id(),);
         let ProtocolContractState::Running(_running_state) = &self.protocol_state else {
@@ -1664,6 +1664,7 @@ mod tests {
     use near_sdk::{test_utils::VMContextBuilder, testing_env, NearToken, VMContext};
     use primitives::key_state::{AttemptId, KeyForDomain};
     use rand::seq::SliceRandom;
+    use rand::SeedableRng;
     use rand::{rngs::OsRng, RngCore};
     use rand_core::CryptoRngCore;
     use sha2::{Digest, Sha256};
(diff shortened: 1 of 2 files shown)


crates/contract/src/lib.rs
```diff
@@ -1652,6 +1663,7 @@ mod tests {
     };
     use near_sdk::{test_utils::VMContextBuilder, testing_env, NearToken, VMContext};
     use primitives::key_state::{AttemptId, KeyForDomain};
+    use rand::seq::SliceRandom;
```
gilcu3: nice! never seen that one before

crates/contract/src/lib.rs
```diff
@@ -2950,4 +2962,90 @@ mod tests {
...
+            );
+        }
+    }
+
+    #[test]
+    #[should_panic(expected = "not a voter")]
+    fn test_remove_update_vote_panics_if_non_voter() {
+        let running_state = gen_running_state(2);
+        let protocol_contract_state = ProtocolContractState::Running(running_state);
+        let mut contract = MpcContract::new_from_protocol_sate(protocol_contract_state);
```
gilcu3: not introduced in this PR, but if it does not break API, we should:
```suggestion
        let mut contract = MpcContract::new_from_protocol_state(protocol_contract_state);
```

crates/contract/src/lib.rs
```diff
@@ -2950,4 +2962,90 @@ mod tests {
...
+
+    #[test]
+    pub fn test_remove_update_vote_running() {
+        let running_state = gen_running_state(2);
+        let participants = running_state.parameters.participants().clone();
+        let protocol_contract_state = ProtocolContractState::Running(running_state);
+        let mut contract = MpcContract::new_from_protocol_sate(protocol_contract_state);
+        let expected = propose_and_vote_code(0, &mut contract);
+        for (account_id, _, _) in participants.participants() {
+            contract
```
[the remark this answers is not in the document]
kevindeforth: > does the contract hash direct vote command?
> e.g contract.vote(..) - is yes, better use it directly.

I might be misunderstanding the question - are you asking if the contract has a direct "vote" command, and suggesting I should use that instead?

If so, then I disagree - the contracts `vote` endpoint is a much more complex function and would require additional test logic. This is a unit test supposed to test `remove_vote` only - it's generally preferred to isolate tested code as much as possible (which, this test is already not doing that great of a job at).

PR #1539, opened 2025-11-24
fix: use correct nearcore commit in submodule

gilcu3:
Follow up to #1525


barakeinav1:
changes seem fine. thanks.
one question - when I open the "files changed" tab, I don't see the 2 files open as am used to.
but rather need to press the files in order to open them. 
is this since this is a submodule ?

image

gilcu3:
> changes seem fine. thanks. one question - when I open the "files changed" tab, I don't see the 2 files open as am used to. but rather need to press the files in order to open them. is this since this is a submodule ?
> 

yes, that's the only way with submodules. One thing I learned today is that if you click on the "2 files" link, you can actually see what's the commit the submodule is using now.

image

PR #1547, opened 2025-11-25
fix: patch nearcore version 210

### LINKED ISSUE #1540: Testnet nodes unstable after upgrading to 3.0.5
### Background

The 3.0.5 upgrade has lead to what seems like a memory leak. We're noticing all upgraded nodes have memory usage that grows linearly until the nodes are restarted.

### User Story

_No response_

### Acceptance Criteria

We've identified the cause and fixed the nodes.

Slack thread: https://nearone.slack.com/archives/C07UW93JVQ8/p1764000206122069

gilcu3:
Closes #1540

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -141,14 +141,14 @@ x509-parser = "0.16.0"
 zstd = "0.13.3"
 
 # NEAR CORE DEPENDENCIES:
-near-async = { git = "https://github.com/near/nearcore", branch = "14668-bug-utoipa-swagger-ui-v902-breaks-reproducible-builds" }
-near-client = { git = "https://github.com/near/nearcore", branch = "14668-bug-utoipa-swagger-ui-v902-breaks-reproducible-builds" }
-near-config-utils = { git = "https://github.com/near/nearcore", branch = "14668-bug-utoipa-swagger-ui-v902-breaks-reproducible-builds" }
-near-crypto = { git = "https://github.com/near/nearcore", branch = "14668-bug-utoipa-swagger-ui-v902-breaks-reproducible-builds" }
-near-indexer = { git = "https://github.com/near/nearcore", branch = "14668-bug-utoipa-swagger-ui-v902-breaks-reproducible-builds" }
-near-indexer-primitives = { git = "https://github.com/near/nearcore", branch = "14668-bug-utoipa-swagger-ui-v902-breaks-reproducible-builds" }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -141,14 +141,14 @@ x509-parser = "0.16.0"
...
-near-o11y = { git = "https://github.com/near/nearcore", branch = "14668-bug-utoipa-swagger-ui-v902-breaks-reproducible-builds" }
-near-time = { git = "https://github.com/near/nearcore", branch = "14668-bug-utoipa-swagger-ui-v902-breaks-reproducible-builds" }
+near-async = { git = "https://github.com/near/nearcore", branch = "bug-utoipa-swagger_patch_nearcore_210rc4" }
+near-client = { git = "https://github.com/near/nearcore", branch = "bug-utoipa-swagger_patch_nearcore_210rc4" }
+near-config-utils = { git = "https://github.com/near/nearcore", branch = "bug-utoipa-swagger_patch_nearcore_210rc4" }
+near-crypto = { git = "https://github.com/near/nearcore", branch = "bug-utoipa-swagger_patch_nearcore_210rc4" }
+near-indexer = { git = "https://github.com/near/nearcore", branch = "bug-utoipa-swagger_patch_nearcore_210rc4" }
+near-indexer-primitives = { git = "https://github.com/near/nearcore", branch = "bug-utoipa-swagger_patch_nearcore_210rc4" }
+near-o11y = { git = "https://github.com/near/nearcore", branch = "bug-utoipa-swagger_patch_nearcore_210rc4" }
+near-time = { git = "https://github.com/near/nearcore", branch = "bug-utoipa-swagger_patch_nearcore_210rc4" }
```
DSharifi: Ignore since it's time critical, and we already depended on a specific branch.

Use `rev` instead of `branch` for git dependencies, if you don't have a release tag to point to. Otherwise the build will give different results depending on the state of the branch at build time.

netrome: Yeah Cargo.lock catches that. Though I agree a `rev` is better than a `branch`. I used a `branch` instead of a `rev` for the 2.9.1 fix because I was faced with a bunch of weird version mismatch errors when I used a `rev` (I tried for quite a while to make it work), that I think @DSharifi resolved in the 2.10.0 migration anyway. Then I think I just got used to using the branch.

netrome:
Okay so the `rc.4` version is set in nearcore but the tag hasn't been published yet. Good to know.

Issue #1555, opened 2025-11-26
External accounts can call the `migrate` function directly


DSharifi:
### Background

There is a bug in `near_sdk` where the `#[private]` attribute is silently ignored if combined with `#[init]`. We assume our `migrate` function is private, but it is not.

See: https://github.com/near/near-sdk-rs/pull/1410

### Acceptance Criteria

Add runtime check for methods with #[private] and #[init] that the method is called by the contract itself. Otherwise the call should panic.

PR #1556, opened 2025-11-26
fix: only allow contract itself to call `migrate` function

DSharifi:
closes https://github.com/near/mpc/issues/1555

### DIFF
--- a/crates/contract/tests/inprocess/expired_attestation.rs
+++ b/crates/contract/tests/inprocess/expired_attestation.rs
@@ -17,7 +17,7 @@ use mpc_contract::{
 use assert_matches::assert_matches;
 use near_account_id::AccountId;
 use near_sdk::{test_utils::VMContextBuilder, testing_env, NearToken, VMContext};
-use std::time::Duration;
+use std::{str::FromStr, time::Duration};
 use utilities::AccountIdExtV2;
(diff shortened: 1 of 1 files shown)


crates/contract/src/lib.rs
```diff
@@ -1237,6 +1237,7 @@ impl MpcContract {
     #[init(ignore_state)]
     #[handle_result]
     pub fn migrate() -> Result<Self, Error> {
```
kevindeforth: We can remove the `#[private]` decorator, since it's not doing anything.

kevindeforth (approved):

crates/contract/src/lib.rs
DSharifi: Yeah, I was a bit conflicted as well on it. But if they fix this in a semver compatible release, which they should, then we don't need to add it again.

DSharifi:
> Thanks! Looks good to me, except that I expected the test + check for the other method we were talking about, `init_running`

We can't test it to be private, since it will always panic if called after intialization.

gilcu3 (dismissed):
Thank you!

gilcu3 (approved):

kevindeforth (dismissed):

PR #1559, opened 2025-11-26
feat: clear update votes from non-participants after resharing

### LINKED ISSUE #1482: Bug: Update votes should be cleared post resharing
### Background

Currently, the contract does not remove stale votes from previous participants after a resharing.

### User Story

As a user / dev / node operater, I only want to count votes of current participants

### Acceptance Criteria

contract update votes from non-participants are removed after concluding a resharing

_No response_

pbeza:
Fixes #1482

### DIFF
--- a/crates/contract/tests/sandbox/update_votes_cleanup_after_resharing.rs
+++ b/crates/contract/tests/sandbox/update_votes_cleanup_after_resharing.rs
@@ -39,7 +39,6 @@ async fn update_votes_from_kicked_out_participants_are_cleared_after_resharing()
             config: None,
         })
         .deposit(CURRENT_CONTRACT_DEPLOY_DEPOSIT)
-        .max_gas()
         .transact()
         .await?
         .json()?;


crates/contract/src/lib.rs
```diff
@@ -814,16 +820,17 @@ impl MpcContract {
...
+            self.protocol_state = new_state;
 
-        if resharing_concluded {
+            // Spawn a promise to clear update votes from accounts that are no longer participants
+            Promise::new(env::current_account_id()).function_call(
+                "clean_non_participant_update_votes".to_string(),
+                vec![],
+                NearToken::from_yoctonear(0),
+                CLEAN_NON_PARTICIPANT_UPDATE_VOTES_GAS,
+            );
```
DSharifi: Is there a reason to do this in a promise?

crates/contract/src/lib.rs
```diff
@@ -1117,6 +1121,30 @@ impl MpcContract {
         }
     }
 
+    /// Private endpoint to clean update votes from non-participants after resharing.
+    /// This can only be called by the contract itself via a promise.
+    #[private]
+    #[handle_result]
+    pub fn clean_non_participant_update_votes(&mut self) -> Result<(), Error> {
```
DSharifi: Clean can mean a lot of things. We also have `clear_votes_from_non_participants`. Let's make it more explicit and consistent
```suggestion
    pub fn remove_non_participant_update_votes(&mut self) -> Result<(), Error> {
```

DSharifi (changes requested):
Thanks for the good work

I'd like to see two things to approve:
1. Add integration tests in a sandbox where we test that the votes are cleared. You can use one of the tests in `upgrade_from_curren_contract.rs` as a skeleton, as we have tests already that submit updates. The issue with only having the unit test is that we are not testing that `clean_non_participant_update_votes` is ever called by the contract itself, since we are driving the contract logic from this unit test. `clear_votes_from_non_participants` is being called by the test driver instead of the contract in the current tests.

2. Add a sandbox regression test to verify that `clean_non_participant_update_votes` is private.

crates/contract/src/update.rs
```diff
@@ -191,6 +191,26 @@ impl ProposedUpdates {
         }
     }
 
+    /// Removes votes from accounts that are not in the provided participant list.
+    /// This should be called after resharing concludes to clear votes from non-participants.
+    pub fn clear_votes_from_non_participants<F>(&mut self, is_participant: F)
+    where
```
gilcu3: hmm the way this function is used is quadratic, we need to do something better if we plan to scale this to a hundred of participants. notice that `is_participant` function is linear

gilcu3 (changes requested):
Left some questions, but mostly worried about the quadratic complexity issue.

crates/contract/src/update.rs
```diff
@@ -191,6 +191,26 @@ impl ProposedUpdates {
         }
     }
 
+    /// Removes votes from accounts that are not in the provided participant list.
+    /// This should be called after resharing concludes to clear votes from non-participants.
+    pub fn clear_votes_from_non_participants<F>(&mut self, is_participant: F)
```
kevindeforth: Lets add a unit test for this function.

crates/contract/src/lib.rs
DSharifi: @kevindeforth do you have more context on why we use promises for cleanup convention?

crates/contract/src/lib.rs
```diff
@@ -3045,4 +3077,48 @@ mod tests {
...
+
+        let ProtocolContractState::Running(running_state) = &contract.protocol_state else {
+            panic!("Expected Running state");
+        };
+        for vote in &votes_after {
+            assert!(
+                running_state.is_participant(vote),
+                "Non-participant vote should have been removed"
+            );
+        }
```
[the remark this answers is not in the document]
pbeza: Thanks for catching it! Fixed here: https://github.com/near/mpc/pull/1559/commits/2e1d34bbd63a5fc0b69cbc187c2744aa4a8cb1d0. ✅

pbeza:
> Thanks for the good work
> 
> I'd like to see two things to approve:
> 
> 1. Add integration tests in a sandbox where we test that the votes are cleared. You can use one of the tests in `upgrade_from_curren_contract.rs` as a skeleton, as we have tests already that submit updates. The issue with only having the unit test is that we are not testing that `clean_non_participant_update_votes` is ever called by the contract itself, since we are driving the contract logic from this unit test. `clear_votes_from_non_participants` is being called by the test driver instead of the contract in the current tests.
> 2. Add a sandbox regression test to verify that `clean_non_participant_update_votes` is private.

Good points! I addressed both of your comments ([ref1](https://github.com/near/mpc/pull/1559/commits/60401b7dd783a3cefae1fb67904fdd1793cd30e1), [ref2](https://github.com/near/mpc/pull/1559/commits/03187e48f458c67630e215957280b8e9e80ad8c8)), as well as the other reviewers’ feedback. Feel free to take another look, @DSharifi @gilcu3 @kevindeforth

The only open item is this question for you, @kevindeforth: https://github.com/near/mpc/pull/1559#discussion_r2567950816 — PTAL when you get a chance. Thanks!

crates/contract/src/update.rs
kevindeforth: We can make it non-quadratic without a huge refactor:
1. put existing voting accounts into a set --> O(nlog(n))
2. put current participants into another set --> O(mlog(m))
3. take set intersection --> O(min(m, n)log(max(n,m))) if we iterate over the smaller & binary search over the larger.

But there probably is an even more efficient approach.

crates/contract/tests/sandbox/integration_update_votes_cleanup_after_resharing.rs
```diff
@@ -0,0 +1,191 @@
...
+        participants::Participants,
+        thresholds::ThresholdParameters,
+    },
+    state::ProtocolContractState,
+    update::{ProposeUpdateArgs, UpdateId},
+};
+
+/// Tests that update votes from non-participants are cleared after resharing.
+#[tokio::test]
+async fn test_update_votes_cleared_after_resharing() -> Result<()> {
```
DSharifi: Nit:
```suggestion
async fn update_votes_from_kicked_out_participants_are_cleared_after_resharing() -> Result<()> {
```

crates/contract/src/update.rs
DSharifi: Let's keep the PR focused, and not block on Big O. it's not obvious that the current implementation is worse than optimizing for Big O after thinking more about it later.

In practice, switching to `HashSet`s will likely lead to worse performance, due to the overhead of hashing and memory allocation, compared to simply looping over co-located data in a `Vec`.

Even with `n~=m=100`, resulting in 10k checks, the participant struct (currently 24 bytes) are small enough to fit entirely within the CPU cache, meaning the linear scan has cache locality, whereas building a `Set` or `HashSet` leads to pointer chasing and expensive hashing operations.

crates/contract/tests/sandbox/common.rs
```diff
@@ -75,7 +75,7 @@ const DUMMY_MIGRATION_CONTRACT_PACKAGE_NAME: &str = "test-migration-contract";
 /// nodes usually attach the maximum available gas.  
 /// For testing, we use this constant to attach a fixed amount to each call and detect if gas usage increases  
 /// unexpectedly in the future.
-pub const GAS_FOR_VOTE_RESHARED: Gas = Gas::from_tgas(22);
+pub const GAS_FOR_VOTE_RESHARED: Gas = Gas::from_tgas(24);
```
gilcu3: does this mean that the gas was tight? Could you let me know which test failed?

crates/contract/tests/sandbox/integration_update_votes_cleanup_after_resharing.rs
gilcu3: name of this file should be:

`integration_update_votes_cleanup_after_resharing`

and maybe it is a good moment to rename `integration_tee_cleanup_after_resharing` as well. All other files in the folder follow the convention of not using the `integration_` prefix

crates/contract/tests/sandbox/integration_update_votes_cleanup_after_resharing.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    for account in &env_accounts[0..2] {
+        account
+            .call(contract.id(), "vote_update")
+            .args_json(json!({"id": update_id}))
+            .max_gas()
+            .transact()
+            .await?
+            .into_result()?;
+    }
```
gilcu3: for this we have a helper function now, which is probably better to use here: `execute_async_transactions`

crates/contract/tests/sandbox/integration_update_votes_cleanup_after_resharing.rs
```diff
@@ -0,0 +1,191 @@
...
+            .call(contract.id(), "vote_new_parameters")
+            .args_json(json!({
+                "prospective_epoch_id": prospective_epoch_id,
+                "proposal": new_threshold_parameters,
+            }))
+            .max_gas()
+            .transact()
+            .await?
+            .into_result()?;
+    }
```
gilcu3: now I realized that using `execute_async_transactions` here will make the test much faster! maybe we should use that for every contract call that we make in a cycle

crates/contract/tests/sandbox/integration_update_votes_cleanup_after_resharing.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    // Propose update and have first 2 participants vote on it
+    let update_id: UpdateId = env_accounts[0]
+        .call(contract.id(), "propose_update")
+        .args_borsh(ProposeUpdateArgs {
+            code: Some(vec![1u8; 1000]),
+            config: None,
+        })
+        .deposit(near_workspaces::types::NearToken::from_near(1))
+        .max_gas()
```
gilcu3: nit: now that https://github.com/near/mpc/pull/1566 is almost done, we could use the constants defined there for our tests

gilcu3 (changes requested):
Left some nits/questions and a small blocker. There is a function that should be made more efficient, unless I am missing something

crates/contract/tests/sandbox/update_votes_cleanup_after_resharing.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    for account in &env_accounts[0..2] {
+        account
+            .call(contract.id(), "vote_update")
+            .args_json(json!({"id": update_id}))
+            .max_gas()
+            .transact()
+            .await?
+            .into_result()?;
+    }
```
pbeza: https://github.com/near/mpc/pull/1559/commits/b95755431435d14aae517856976f945c7e6b84dd

crates/contract/tests/sandbox/update_votes_cleanup_after_resharing.rs
```diff
@@ -0,0 +1,191 @@
...
+        .iter()
+        .min_by_key(|a| {
+            resharing_state
+                .resharing_key
+                .proposed_parameters()
+                .participants()
+                .id(&a.id().as_v2_account_id())
+                .unwrap()
+        })
+        .unwrap()
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/pull/1559/commits/6804962f37fb17c9349c505cc94321944a341027

crates/contract/src/lib.rs
```diff
@@ -1122,6 +1129,37 @@ impl MpcContract {
...
+
+    pub(crate) fn remove_votes_from_non_participants(&mut self, participants: &Participants) {
+        let non_participants: Vec<AccountId> = self
+            .proposed_updates
+            .voters()
+            .into_iter()
+            .filter(|voter| !participants.is_participant(voter))
+            .collect();
+        self.proposed_updates.remove_votes(&non_participants);
+    }
```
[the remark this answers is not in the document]
pbeza: This function is used in two places—the function above and one of the tests—so it can’t be inlined. I’ve already addressed this here: https://github.com/near/mpc/pull/1559#discussion_r2571653091.

Feel free to reopen the comment if you think it still needs changes.

crates/contract/src/update.rs
```diff
@@ -191,6 +192,30 @@ impl ProposedUpdates {
...
+    pub fn remove_non_participant_votes(&mut self, participants: &Participants) {
+        let non_participants: Vec<AccountId> = self
+            .vote_by_participant
+            .keys()
+            .filter(|voter| !participants.is_participant(voter))
+            .cloned()
+            .collect();
+
+        self.remove_votes(&non_participants);
+    }
```
kevindeforth: nit: I would inline this in in `lib.rs` - it's one less method to test and maintain.

crates/contract/src/lib.rs
```diff
@@ -3071,4 +3112,46 @@ mod tests {
             .build());
         contract.remove_update_vote();
     }
+
+    #[test]
+    pub fn test_update_votes_cleared_after_resharing() {
```
kevindeforth: The name of this function is mis-leading: it does not test whether votes are cleared after a resharing, but if `PrposodeUpdates::remove_non_participant_votes()` works.

This test doesn't need to be in this file, as it does not test a contract method. It can be massively simplified and moved to `votes.rs`

crates/contract/src/lib.rs
```diff
@@ -81,6 +82,13 @@ const MINIMUM_SIGN_REQUEST_DEPOSIT: NearToken = NearToken::from_yoctonear(1);
 /// Minimum deposit required for CKD requests
 const MINIMUM_CKD_REQUEST_DEPOSIT: NearToken = NearToken::from_yoctonear(1);
 
+/// Prepaid gas for a `cleanup_orphaned_node_migrations` call
+/// todo: benchmark [#1164](https://github.com/near/mpc/issues/1164)
+const CLEAN_NODE_MIGRATIONS: Gas = Gas::from_tgas(3);
```
gilcu3: nit: this should have the _GAS suffix as the other constants

crates/contract/src/update.rs
```diff
@@ -223,6 +224,30 @@ impl ProposedUpdates {
...
+
+    /// Returns all account IDs that have voted.
+    pub fn voters(&self) -> Vec<AccountId> {
```
gilcu3: this method is currently only used in tests. So we either remove it or
```suggestion
    /// Returns all account IDs that have voted.
    #[cfg(test)]
    pub fn voters(&self) -> Vec<AccountId> {
```

crates/contract/src/update.rs
```diff
@@ -223,6 +224,30 @@ impl ProposedUpdates {
...
+    }
+
+    /// Removes votes from accounts that are not participants.
+    pub fn remove_non_participant_votes(&mut self, participants: &Participants) {
+        let non_participants: Vec<AccountId> = self
+            .vote_by_participant
+            .keys()
+            .filter(|voter| !participants.is_participant(voter))
+            .cloned()
+            .collect();
```
gilcu3: can we add a comment that this step is quadratic and we might need to improve it in the future?

crates/contract/src/lib.rs
DSharifi: Please move this gas value to the config struct in `config.rs`, where all other gas values live.

crates/contract/src/lib.rs
```diff
@@ -961,19 +967,32 @@ impl MpcContract {
...
+        let current_participants: HashSet<AccountId> = running_state
+            .parameters
+            .participants()
+            .participants()
+            .iter()
+            .map(|(account_id, _, _)| account_id.clone())
+            .collect();
+
+        // Only count votes from current participants
+        let valid_votes_count = all_votes.intersection(&current_participants).count();
```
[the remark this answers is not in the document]
pbeza: > but if the gas measurements are not affected by this, I am fine with going for this, as it is more readable

Actually, I like your original point here. We should optimize for gas usage, and introducing a `HashSet` likely impacts that (though I haven’t compared the gas usage with and without it).

Hopefully this is effectively the same as what you suggested: https://github.com/near/mpc/pull/1559/commits/1ff787e6bb1d2d5c387af80b1735dc56a308e408.

https://github.com/near/mpc/blob/1ff787e6bb1d2d5c387af80b1735dc56a308e408/crates/contract/src/lib.rs#L973-L981

Thanks for bringing an algorithmic eye to the review! 🙂

crates/contract/tests/sandbox/common.rs
```diff
@@ -74,7 +74,7 @@ const DUMMY_MIGRATION_CONTRACT_PACKAGE_NAME: &str = "test-migration-contract";
 /// nodes usually attach the maximum available gas.  
 /// For testing, we use this constant to attach a fixed amount to each call and detect if gas usage increases  
 /// unexpectedly in the future.
-pub const GAS_FOR_VOTE_RESHARED: Gas = Gas::from_tgas(22);
+pub const GAS_FOR_VOTE_RESHARED: Gas = Gas::from_tgas(34);
```
gilcu3: was this needed as part of the gas experiments only? Because now it looks like a big bump, please revert if not needed

crates/contract/src/lib.rs
```diff
@@ -3059,4 +3089,98 @@ mod tests {
...
+    /// - Adding a 2nd participant vote should meet threshold (returns true)
+    #[test]
+    pub fn test_vote_update_filters_non_participant_votes() {
+        // given: a running state with 3 participants and threshold of 2
+        let running_state = gen_running_state(1);
+        let running_state = RunningContractState::new(
+            running_state.domains.clone(),
+            running_state.keyset.clone(),
+            ThresholdParameters::new(gen_participants(3), Threshold::new(2)).unwrap(),
+        );
```
kevindeforth: I would find it cleaner to do one of the following:
- use the output of `gen_running_state(1)` directly - we just need to loop over the participants for voting.
- not use `gen_running_state()`, as we don't really require domains or keysets

crates/contract/src/lib.rs
```diff
@@ -3059,4 +3089,98 @@ mod tests {
...
+        let update_id = UpdateId(0);
+
+        // Add votes from 2 current participants
+        let participants = participants.participants();
+        let (p1, p2) = (participants[0].0.clone(), participants[1].0.clone());
+        contract.proposed_updates.vote(&update_id, p1.clone());
+        contract.proposed_updates.vote(&update_id, p2.clone());
+
+        // Sanity check: 4 total votes (2 non-participants + 2 participants)
+        assert_eq!(contract.proposed_updates.voters().len(), 4);
```
kevindeforth: **nit:**
It's always nice to compare against the full expected items, not just the expected number of items (like you do on line 3184).

Additionally to `ProposedUpdates::voters()`, I would advise to also compare against the output of `ProposedUpdates::all_updates()`

crates/contract/tests/sandbox/update_votes_cleanup_after_resharing.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    // Propose update and have first 2 participants vote on it
+    let update_id: UpdateId = env_accounts[0]
+        .call(contract.id(), "propose_update")
+        .args_borsh(ProposeUpdateArgs {
+            code: Some(vec![1u8; 1000]),
+            config: None,
+        })
+        .deposit(near_workspaces::types::NearToken::from_near(1))
+        .max_gas()
```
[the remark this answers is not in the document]
pbeza: From what I see in the existing code, [we use](https://github.com/search?q=repo%3Anear%2Fmpc%20propose_update&type=code) `CURRENT_CONTRACT_DEPLOY_DEPOSIT` for `propose_update`, which is a bit strange — I’d expect that constant to be used for deploying a new contract, not proposing an update. I’m probably missing something here. Anyway, to keep things consistent with the current codebase, I just reused that constant here: https://github.com/near/mpc/pull/1559/commits/4de6cf7ec4512b047199acca6a10f8e777ce8889.

crates/contract/src/update.rs
```diff
@@ -663,4 +691,73 @@ mod tests {
...
+        let (acc0, acc1, acc2) = (gen_account_id(), gen_account_id(), gen_account_id());
+        proposed_updates.vote(&update_id, acc0.clone());
+        proposed_updates.vote(&update_id, acc1.clone());
+        proposed_updates.vote(&update_id, acc2.clone());
+
+        proposed_updates.remove_votes(&[acc0, acc2]);
+
+        let result: TestUpdateVotes = (&proposed_updates).try_into().unwrap();
+        assert_eq!(result.entries.len(), 1);
+        assert_eq!(result.entries[&update_id.0].votes, HashSet::from([acc1]));
```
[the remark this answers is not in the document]
pbeza: Makes sense! Fixed: https://github.com/near/mpc/pull/1559/commits/a7d5455a1563206f6a32c0c31289422869497adb. ✅

kevindeforth (commented):
Nice, thanks for addressing the comments!

PR #1563, opened 2025-11-26
feat: embed abi in contract

### LINKED ISSUE #1562: embed abi in the contract
### Background

Embedding the ABI in the contract will improve our developer friendliness. The problem is that it might increase its size too much.

### User Story

Having embedded ABI makes it easier for developers to use the contract

### Acceptance Criteria

The contract is compiled with embedded ABI by default

_No response_

gilcu3:
Closes #1562 

Example [MPC contract](https://testnet.nearblocks.io/address/gilcu3-contract.testnet?tab=contract) with embedded abi deployed in testnet

Contract size change: before -> 1.2M, after -> 1.2M, so it seems compression worked pretty well.

Note: I am not removing the abi feature from the contract because IIRC it would break clippy/tests somehow

### DIFF
--- a/crates/contract/Cargo.toml
+++ b/crates/contract/Cargo.toml
@@ -31,7 +31,6 @@ container_build_command = [
     "--profile=release-contract",
     "--features",
     "abi",
-    "--no-embed-abi",
 ]
 
 [package.metadata.cargo-shear]


DSharifi:
> 1.2M, after -> 1.2M

What about in order of kibibytes?

DSharifi (approved):

PR #1584, opened 2025-11-30
chore: Update dockerhub configuration parameter and add integration test for validate_image_hash using Docker Hub image

### LINKED ISSUE #1567: Add pytest for  launcher.validate_image_hash
### Background

As a follow up for https://github.com/near/mpc/pull/1527

We want to add a pytest for launcher.py->validate_image_hash
note - this test will need to fetch an image from dockerhub (this creates a dependency in CI the on docker hub)  and validate it.

This is why this it is  separate task.

planed solution:
1. upload a small docker image to dockerhub.
2. call validate_image_hash that will
   a. pull the image from dockerhub
   b. inspect the image, and compare the image ID against an approved hardcoded hash

### User Story

As a developer, I'd like to be sure that my code is covered by ULTs

### Acceptance Criteria

Test is added and running in CI

_No response_

barakeinav1:
fixes #1567
fixes [#1585](https://github.com/near/mpc/issues/1585)


tee_launcher/test_launcher_config.py
```diff
@@ -449,3 +449,76 @@ def test_missing_file_fallback(monkeypatch):
...
+ATTEMPT_VALUES = [3, 5]
+
+
+@pytest.mark.ci_excluded
+def test_validate_image_hash_stress():
+    """
+    Stress-test validate_image_hash() with reasonable permutations of RPC timing params.
+    This test is intentionally slow and WILL hit the real Docker registry.
+    Run manually:  pytest -m stress
+    """
```
gilcu3: I don't think we need this test. I believe our purpose is not to be able to use dockerhub with any set of parameters, at most we need a lower bound in some parameters, so a single test should be enough

tee_launcher/pytest.ini
```diff
@@ -0,0 +1,6 @@
+[pytest]
+markers =
+    slow: marks tests as slow (deselect with '-m "not slow"')
+    ci_excluded: marks tests that should be excluded from ci
+    no_atexit_cleanup: Skip the autouse atexit cleanup fixture for this test
```
gilcu3: from this list we are only using ci_excluded. There is no need to add the others

tee_launcher/test_launcher_config.py
gilcu3: > > I don't think we need this test
> 
> why not? - it tests that different value are acceptable. Anyway, this test will not run in CI by default.

if it does not run in CI, it looks even less suitable to me. What is the test objective? Surely if the timeouts are too small, or the network is bad, the test will fail, but that's totally outside of our control, so the test IMO does not achieve any purpose

> > believe our purpose is not to be able to use dockerhub with any set of parameters,
> 
> I don't understand this comment.

related to the above, IMO this test is not testing our system, it is testing if dockerhub works with certain timeouts, which does not depend on us at all. As long as we have a set of minimal values that is tested to work (done in the other test), then we are fine

tee_launcher/launcher.md
```diff
@@ -1,89 +1,173 @@
...
+The launcher ensures integrity, validates the MPC image before execution, and interacts with dstack’s attestation agent.
 
-This is a secure launcher script for initializing and attesting a Docker-based MPC node.  
-It is designed to run inside a TEE-enabled environment (e.g., Intel TDX) to add and ensures the integrity and trustworthiness of the image before launching it.
+The launcher is a single Python script: **launcher.py**.
 
+---
 
-## 🔐 Features
+# 🔐 Features
```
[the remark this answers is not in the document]
gilcu3: in markdown there should be a single `#` header in the file, that's why it was changed at some point.

tee_launcher/test_launcher_config.py
barakeinav1: We support different a configuration values in case there is a problem with dockerhub.
The test checks that different values are accepted by our code, and at least for now work with dockerhub.

tee_launcher/test_launcher_config.py
```diff
@@ -449,3 +449,76 @@ def test_missing_file_fallback(monkeypatch):
...
+INTERVAL_VALUES = [0.1, 0.5, 1.0]
+ATTEMPT_VALUES = [3, 5]
+
+
+@pytest.mark.ci_excluded
+def test_validate_image_hash_stress():
+    """
+    Stress-test validate_image_hash() with reasonable permutations of RPC timing params.
+    This test is intentionally slow and WILL hit the real Docker registry.
+    Run manually:  pytest -m stress
```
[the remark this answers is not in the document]
barakeinav1: changed name

tee_launcher/pytest.ini
netrome:
```suggestion
    ci_excluded: marks tests that should be excluded from ci
```

As Reynaldo said, the `slow` and `no_atexit_cleanup` flags are unused. Let's remove them.

tee_launcher/test_launcher_config.py
netrome: I agree, this test feels redundant as it does not trigger any new code paths on our end.

tee_launcher/pytest.ini
gilcu3: if test `test_validate_image_hash_stress` is removed as suggested, even `ci_excluded` will not be used, so the file itself can be removed

gilcu3 (dismissed):
Thank you for taking the time!

Left only nits.

gilcu3 (dismissed):

netrome (changes requested):
Right, we still need to add the workflow to build the new docker image so we don't rely on your personal registry.

PR #1588, opened 2025-12-01
chore: make attestation crate independent of the mpc

### LINKED ISSUE #1570: Make attestation crate independent of the MPC
### Background

Currently, the attestation crate contains many values/functions that are mpc dependent. It also contains hard-coded values such as `EXPECTED_LOCAL_SGX_EVENT_DIGEST`. To be able to use the crate for the backup-service and the example CKD TEE application, this crate needs to be more general.

Currently the attestation crate is used by several other crates in the workspace:
- contract
- node
- node-types
- tee_authority
- test_utils

For this we can:
- create a new crate `mpc-attestation`, that can be used by the current workspace crates exactly as the attestation crate is currently used
- generalize the `attestation` crate so that it is fully independent of the mpc crates

### User Story

As a user we get a new attestation crate, which is nice!

### Acceptance Criteria

The attestation crate becomes independent of the mpc details

_No response_

gilcu3:
Closes #1570 

- Each commit can be pretty much reviewed independently
- I did not change any API on purpose. This leads to a TODO regarding a clone, that we might want to remove in a follow-up
- No test was removed, although several of them were moved.
- Before we had a single `local_sgx_event_digest`. Now with each of the possible measurements of the other registers, we also get the digest, as was the case before #1428. I would go a bit further and change the API, in a follow-up. I believe the verify function should take a single set of measurements, and the caller should simply implement custom logic if they need it (like we do with the prod/dev dstack images).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -886,7 +886,7 @@ checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
 name = "attestation"
-version = "3.0.6"
+version = "3.1.0"
 dependencies = [
  "borsh",
  "dcap-qvl",
@@ -1113,7 +1113,7 @@ dependencies = [
 
 [[package]]
 name = "backup-cli"
-version = "3.0.6"
+version = "3.1.0"
 dependencies = [
  "anyhow",
  "clap",
@@ -1626,7 +1626,7 @@ dependencies = [
 
 [[package]]
 name = "ckd-example-cli"
-version = "3.0.6"
+version = "3.1.0"
 dependencies = [
  "anyhow",
  "blstrs",
@@ -1800,7 +1800,7 @@ checksum = "245097e9a4535ee1e3e3931fcfcd55a796a44c643e8596ff6566d68f09b87bbc"
 
 [[package]]
 name = "contract-interface"
-version = "3.0.6"
+version = "3.1.0"
 dependencies = [
  "borsh",
  "bs58 0.5.1",
@@ -1814,7 +1814,7 @@ dependencies = [
 
 [[package]]
 name = "contract_history"
-version = "3.0.6"
+version = "3.1.0"
 dependencies = [
  "bs58 0.5.1",
  "clap",
@@ -4649,7 +4649,7 @@ checksum = "7843ec2de400bcbc6a6328c958dc38e5359da6e93e72e37bc5246bf1ae776389"
 
 [[package]]
 name = "mpc-attestation"
-version = "3.0.6"
+version = "3.1.0"
 dependencies = [
  "attestation",
  "borsh",
@@ -4667,7 +4667,7 @@ dependencies = [
 
 [[package]]
 name = "mpc-contract"
-version = "3.0.6"
+version = "3.1.0"
 dependencies = [
  "anyhow",
  "assert_matches",
@@ -4712,7 +4712,7 @@ dependencies = [
 
 [[package]]
 name = "mpc-devnet"
-version = "3.0.6"
+version = "3.1.0"
 dependencies = [
  "anyhow",
  "borsh",
@@ -4743,7 +4743,7 @@ dependencies = [
 
 [[package]]
 name = "mpc-node"
-version = "3.0.6"
+version = "3.1.0"
 dependencies = [
  "aes-gcm",
  "anyhow",
@@ -4811,7 +4811,7 @@ dependencies = [
 
 [[package]]
 name = "mpc-primitives"
-version = "3.0.6"
+version = "3.1.0"
 dependencies = [
  "borsh",
  "derive_more 2.0.1",
@@ -4825,7 +4825,7 @@ dependencies = [
 
 [[package]]
 name = "mpc-tls"
-version = "3.0.6"
+version = "3.1.0"
 dependencies = [
  "anyhow",
  "ed25519-dalek",
@@ -6609,7 +6609,7 @@ dependencies = [
 
 [[package]]
 name = "node-types"
-version = "3.0.6"
+version = "3.1.0"
 dependencies = [
  "anyhow",
  "bs58 0.5.1",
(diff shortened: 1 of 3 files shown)


crates/attestation/src/report_data.rs
gilcu3: this file was just moved with minor modifications, shame that github does not see it like that

crates/attestation/src/attestation.rs
```diff
@@ -93,10 +87,8 @@ pub enum VerificationError {
     },
     #[error("the mock attestation is invalid per definition")]
     InvalidMockAttestation,
-    #[error("the allowed mpc image hashes list is empty")]
-    EmptyAllowedMpcImageHashesList,
-    #[error("the allowed mpc laucher compose hashes list is empty")]
-    EmptyAllowedMpcLauncherComposeHashesList,
+    #[error("custom error: `{0}`")]
+    Custom(String),
```
gilcu3: this is so that down-stream users can provide their own error messages. The first down-stream user is the mpc-crate, so errors that previously had names, now become this Custom variant

crates/attestation/src/measurements.rs
```diff
@@ -5,22 +5,9 @@ use borsh::{BorshDeserialize, BorshSerialize};
...
-///
-/// Digest format:
-///   digest = SHA384( event_type + ":" + "key-provider" + ":"+payload) )
-///
-/// Payload format: sha256 {"name":"local-sgx", "id": "<mr_enclave of the provider>"}
-pub const EXPECTED_LOCAL_SGX_EVENT_DIGEST: [u8; 48] = [
-    0x74, 0xca, 0x93, 0x9b, 0x8c, 0x3c, 0x74, 0xaa, 0xb3, 0xc3, 0x09, 0x66, 0xa7, 0x88, 0xf7, 0x74,
-    0x39, 0x51, 0xd5, 0x4a, 0x93, 0x6a, 0x71, 0x1d, 0xd0, 0x14, 0x22, 0xf0, 0x03, 0xff, 0x9d, 0xf6,
-    0x66, 0x6f, 0x3c, 0xc5, 0x49, 0x75, 0xd2, 0xe4, 0xf3, 0x5c, 0x82, 0x98, 0x65, 0x58, 0x3f, 0x0f,
-];
```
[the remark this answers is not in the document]
gilcu3: yeah, the removal happened after that commit, so you will need to keep reviewing to understand it :) the TLDR is that that values was already part of the tcb_info.json files we have in assets. So I am taking the constant from there instead.

crates/attestation/src/measurements.rs
```diff
@@ -85,10 +80,7 @@ impl ExpectedMeasurements {
                 mrtd: decode_rtmr("mrtd", &tcb_info.mrtd)?,
             };
 
-            Ok(ExpectedMeasurements {
-                rtmrs,
```
barakeinav1: The purpose of version is to have the ability to change the report data format in the future.
don't we check that the version is 1 now?

crates/attestation/src/measurements.rs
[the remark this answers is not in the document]
barakeinav1: I also don't want to loss the documentation on how we calculated this value.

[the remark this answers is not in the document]
gilcu3: >I also don't want to loss the documentation on how we calculated this value.

I think there is no need to lose the documentation, I can add it back to the point in which we take it from the json file.

> I don't think we should take EXPECTED_LOCAL_SGX_EVENT_DIGEST from the tcb_info.json.
It has a long list of even't and we don't use them.

the point is that we should take it from the same place we take the RTMRs and we should make it part of the ExpectedMeasurements struct. How that is done, if taken from the original `tcb_info.json` file, or from a reduced one with only the values we care about, is fine for me. Here I just did not want to modify the part of how we store the values.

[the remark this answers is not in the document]
gilcu3: > Not sure about this. Since the key-provider and the RTMR are orthogonal.

they are both expected measurements, and they come from the same source, at least in our workflow

crates/attestation/src/measurements.rs
gilcu3: I thought we would, but we didn't. Happy to add it back (and fix the bug of where should we actually check it). But it feels to me it does not belong to the ExpectedMeasurements struct anyway.

crates/attestation/src/attestation.rs
```diff
@@ -143,41 +133,13 @@ pub enum MockAttestation {
 pub(crate) fn verify_mock_attestation(
     mock_attestation: &MockAttestation,
     timestamp_seconds: u64,
-    allowed_mpc_docker_image_hashes: &[MpcDockerImageHash],
-    allowed_launcher_docker_compose_hashes: &[LauncherDockerComposeHash],
 ) -> Result<(), VerificationError> {
```
kevindeforth: can we add `expected_measurements_list: &[ExpectedMeasurements]` here?

crates/mpc-attestation/src/attestation.rs
```diff
@@ -0,0 +1,207 @@
...
+    Invalid,
+    /// Pass validation depending on the set constraints
+    WithConstraints {
+        mpc_docker_image_hash: Option<MpcDockerImageHash>,
+        launcher_docker_compose_hash: Option<LauncherDockerComposeHash>,
+
+        /// Unix time stamp for when this attestation expires.  
+        expiry_time_stamp_seconds: Option<u64>,
+    },
+}
```
kevindeforth: How will this `MockAttestation` struct play with the `MockAttestation` inside the attestation crate?

https://github.com/near/mpc/blob/2f09f9197b46b96083d63e161444c7f86045114b/crates/attestation/src/attestation.rs#L120-L131

gilcu3: atm they dont play. As you noticed the MockAttestation from the attestation crate is not used. Lets continue this in the other comment or in the internal thread

barakeinav1 (approved):

PR #1594, opened 2025-12-02
fix: wrong tag name in gcp image creation

gilcu3:
Follow up to #1480

- Until we stop depending on gcp, all operators should use mpc-node-gcp image (for simplicity, in reality the non-gcp operators should use the mpc-node tag). So this does not imply any change in the release process, except the guide from https://github.com/near/mpc/pull/1551 will need a little fix


netrome:
Small follow-up issue: https://github.com/near/mpc/issues/1595

PR #1614, opened 2025-12-05
fix: use gas value in the new config for the update promise

### LINKED ISSUE #1615: Aattach the gas value from the new config for config updates
### Background

When a vote for a config update is executed, we should attach the gas value from the new config for the promise that udpates the config. The contract is now using the current config value, which is problematic in scenarios this attached gas parameter is incorrectly set.

The issue with using the current config value is that in a scenario where `vote_update` consumes too much gas and in combination with the current `config.contract_upgrade_deposit_tera_gas` exceeds max gas, 300tgas, then the whole transaction will fail.

https://github.com/near/mpc/blob/37af62bffbed58f4bda9e348ce756a0fec42f4ce/crates/contract/src/update.rs#L209-L216

### User Story

_No response_

### Acceptance Criteria

Use the gas value `contract_upgrade_deposit_tera_gas` from the **new config** when spawning promises for config updates.

_No response_

DSharifi:
closes #1615

### DIFF
--- a/crates/contract/src/update.rs
+++ b/crates/contract/src/update.rs
@@ -208,7 +208,8 @@ impl ProposedUpdates {
             }
             Update::Config(config) => {
                 // If we vote for a new config, we should use
-                // attach the gas value
+                // the value `contract_upgrade_deposit_tera_gas` from the config
+                // as the new gas value
                 let new_config_gas_value = Gas::from_tgas(config.contract_upgrade_deposit_tera_gas);
                 promise = promise.function_call(
                     "update_config",


crates/contract/src/update.rs
```diff
@@ -207,11 +207,14 @@ impl ProposedUpdates {
                 );
             }
             Update::Config(config) => {
+                // If we vote for a new config, we should use
+                // attach the gas value
```
DSharifi:
```suggestion
                // If we vote for a new config, we should use
                // the value `contract_upgrade_deposit_tera_gas` from the config
                // as the new gas value
```

Issue #1598, opened 2025-12-02
Expect a specific version of the attestation report data in the contract


pbeza:
### Background

@gilcu3 [noticed](https://nearone.slack.com/archives/C07UW93JVQ8/p1764593972363819) that we currently don’t use the `report_data_version` field in the `ExpectedMeasurements` struct for anything. The report-data version should probably be hardcoded in the contract so that only a specific version of the report-data format is accepted.

### Acceptance Criteria

The contract accepts only a specific version of the report data attestation. Right now, we have only one version that should be allowed.

### Resources & Additional Notes

- https://nearone.slack.com/archives/C07UW93JVQ8/p1764593972363819

gilcu3:
Cross-linking the thread related to the previous link: [thread](https://nearone.slack.com/archives/C0912BTG51T/p1765211008933019?thread_ts=1765208944.174099&cid=C0912BTG51T)

Citing @DSharifi:
>So we will need to store the version tag. Otherwise if we introduce a new version variant, how will the contract know which variant to use when re-constructing it?

Currently the contract expects a single version of the report data, and constructs the object before verifying attestation. If in the future the nodes support a different report-data version, the contract will need to be aware of it, and to do so the only way is to get that information somehow, for example in the request. Of course that is not something that needs to be done now, as it will only be needed in the future when a new report data version exists. The problem is that it will imply a small breaking change.

Issue #1617, opened 2025-12-05
Remove `votes` from `UpdateEntry`


netrome:
### Background

We've observed very high gas usage for the `vote_update` contract method. This originates from the fact that we store update votes in the `UpdateEntry` struct:

```rust
struct UpdateEntry {
    update: Update,
    votes: HashSet<AccountId>,
    bytes_used: u128,
}
```

This struct should only represent the update and not be concerned with bookkeeping of votes, as that is the purpose of the `ProposedUpdates` struct.

```rust
pub struct ProposedUpdates {
    vote_by_participant: IterableMap<AccountId, UpdateId>,
    entries: IterableMap<UpdateId, UpdateEntry>,
    id: UpdateId,
}
```

The problem is that when voting for contract proposals we need to load the whole `UpdateEntry` containing the entire contract bytecode to register the vote.

This is how the method currently looks:
```rust
    /// Removes any existing vote by `voter`.
    /// Sets `voter`s vote for the update with the given id.
    ///
    /// Returns `Some(votes)` if the given [`UpdateId`] exists, otherwise `None`.
    pub fn vote(&mut self, id: &UpdateId, voter: AccountId) -> Option<&HashSet<AccountId>> {
        // If participant has voted before, remove their vote
        self.remove_vote(&voter);
        // ensure that the update the participant is voting for exists
        let Some(update_entry) = self.entries.get_mut(id) else {
            env::log_str(&format!("no update with id {:?} exists", id));
            return None;
        };
        // record the vote
        self.vote_by_participant.insert(voter.clone(), *id);
        update_entry.votes.insert(voter);
        Some(&update_entry.votes)
    }
```

Instead if we only tracked votes in `ProposedUpdates` we would not have to load the whole proposed contract on every vote, which would significantly reduce the gas costs of voting.

### User Story

As a node operator I don't want to have to pay for loading the whole contract on every vote.

### Acceptance Criteria

It's possible to vote without loading the proposed contract bytecode.

PR #1626, opened 2025-12-08
refactor: remove `ReportData::new()`

pbeza:
The issue #1598 seems to be invalid, since this line already implicitly checks the expected version of the report data:

https://github.com/near/mpc/blob/b4f7acb5424459ae1c75a0fa46f1003c07ab92a9/crates/attestation/src/attestation.rs#L254

but I took the opportunity to refactor it a bit, as @barakeinav1 suggested in the issue thread. `ReportData::new()` should abstract over any specific version of `ReportData` and shouldn’t assume any particular set of fields stored in it.

### DIFF
--- a/crates/mpc-attestation/tests/test_attestation_verification.rs
+++ b/crates/mpc-attestation/tests/test_attestation_verification.rs
@@ -38,7 +38,7 @@ fn test_verify_method_signature() {
     let account_key = account_key();
 
     let report_data: ReportData = ReportDataV1::new(tls_key, account_key).into();
-    let timestamp_s = 1763626832_u64; // Thursday, 20 November 2025 08:20:32
+    let timestamp_s = 1763626832_u64; //Thursday, 20 November 2025 08:20:32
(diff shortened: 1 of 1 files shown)


gilcu3:
> > Thank you! Certainly an improvement, although I am not sure if we should close the linked issue. As mentioned @DSharifi in [thread](https://nearone.slack.com/archives/C0912BTG51T/p1765211008933019?thread_ts=1765208944.174099&cid=C0912BTG51T) we might need to do something more to enable upgrading the ReportData version.
> 
> Fair enough. In that case, I’ll merge it as-is but without the `Fixes #1598` in the commit message, so it doesn’t close the original issue. I’m still not 100% sure what the intended direction for that issue is — if you have any suggestions, feel free to add them to #1598. I’ll rethink it as well and pick it up again after I finish #1617, which has higher priority right now.

Sure, added https://github.com/near/mpc/issues/1598#issuecomment-3631261119

PR #1627, opened 2025-12-08
feat: add derivation path support for ckd

### LINKED ISSUE #1622: extend CKD with derivation path
### Background

Our customers expect to have a derivation path parameter to the CKD request so that a single account can manage many TEE apps. We can compute the app_id in a similar way than we currently compute the tweak for signatures in:

https://github.com/near/mpc/blob/c15b0a63a8942945b60bf92e769dadc4ba51b1a7/crates/contract/src/crypto_shared/kdf.rs#L16-L30

### User Story

It is simpler to have a single account managing several TEE apps

### Acceptance Criteria

A path parameter is added so that the same account can obtain many confidential keys

[Discussion](https://nearone.slack.com/archives/C07UW93JVQ8/p1764958230601659)

gilcu3:
Closes #1622 

There are a couple of TODOs that will be removed once we update the reference to the ts repo. Did not want to do here to avoid cluttering the PR.

### DIFF
--- a/crates/contract/tests/sandbox/common.rs
+++ b/crates/contract/tests/sandbox/common.rs
@@ -57,7 +57,7 @@ use std::{
 };
 use threshold_signatures::{
     blstrs,
-    confidential_key_derivation::{self as ckd, ciphersuite::hash_to_curve, ElementG1, BLS12381SHA256},
+    confidential_key_derivation::{self as ckd, ciphersuite::hash_to_curve, BLS12381SHA256},
     ecdsa as ts_ecdsa, eddsa,
     frost_ed25519::{self, keys::SigningShare, Ed25519Group, Group as _, VerifyingKey},
     frost_secp256k1::{self, Secp256K1Group},
@@ -717,7 +717,6 @@ pub fn create_response_ckd(
     let mut hash_input = [0u8; 128];
     hash_input[..96].copy_from_slice(&pk);
     hash_input[96..].copy_from_slice(app_id.as_ref());
-
 
     let big_s = hash_to_curve(&hash_input) * msk;
     let y = ckd::Scalar::random(OsRng);


crates/ckd-example-cli/src/cli.rs
```diff
@@ -11,6 +11,9 @@ pub struct Args {
     #[arg(long, env)]
     pub domain_id: DomainId,
 
+    #[arg(long, env)]
+    pub path: String,
```
DSharifi: I think this path flag could be more descriptive. Path to what?

If you want to keep the CLI flag concise and keep `flag` name. could you add a doc which will also be displayed to the user when running `--help`?

crates/ckd-example-cli/README.md
```diff
@@ -30,14 +30,16 @@ domain is `Bls12381`.
 
 - `mpc-ckd-public-key`: the public key associated to the domain id above.
 
+- `path`: derivation path for the confidential key, which allows a single account to request several keys
```
DSharifi: Would it not be better to have these flag docs in the Struct definition such that the user can see it when running `--help`?

pytest/common_lib/ckd.py
```diff
@@ -70,12 +71,21 @@ def verify_bls_signature(
     pk = G2Element.from_bytes(bytes(public_key.to_compressed_bytes()))
     sig = G1Element.from_bytes(bytes(signature.to_compressed_bytes()))
     # hash_to_curve
+    # TODO: this will need to be updated once ts repo with https://github.com/near/threshold-signatures/pull/246 is merged
```
DSharifi: Can you link to a tracking issue as well?

crates/contract/tests/sandbox/ckd.rs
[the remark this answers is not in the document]
gilcu3: added in 86b038d0cfadf3b9ff52abd456d0ef0fbbba82f0

crates/ckd-example-cli/src/ckd.rs
```diff
@@ -130,3 +137,13 @@ fn derive_strong_key(
     hk.expand(info, &mut okm).map_err(|err| anyhow!("{err}"))?;
     Ok(okm)
 }
+
+const APP_ID_DERIVATION_PREFIX: &str = "near-mpc v0.1.0 app_id derivation:";
```
SimonRastikian: Nicely Salted XD

crates/contract/tests/sandbox/common.rs
```diff
@@ -697,14 +698,26 @@ pub fn create_response_ckd(
...
-    let app_id = account_id.as_bytes();
+    let app_id = derive_app_id(account_id, derivation_path);
     let app_pk: ckd::ElementG1 = app_public_key.into_contract_type();
-    let msk = signing_key;
-    let big_s = hash_to_curve(app_id) * msk;
+    let msk = key_package.private_share.to_scalar();
+    let pk = key_package.public_key.to_element().to_compressed();
+
+    let mut hash_input = [0u8; 128];
+    hash_input[..96].copy_from_slice(&pk);
```
SimonRastikian: Please use the constant provided by ElementG1 instead of 96 (I know both are similar currently).
Additionally, please create hash_input as vec with capacity the sum of the lengths.

pytest/common_lib/ckd.py
```diff
@@ -70,12 +71,21 @@ def verify_bls_signature(
...
+    # TODO(#1628): this will need to be updated once ts repo with https://github.com/near/threshold-signatures/pull/246 is merged
     Hm = G1Element.from_message(message, NEAR_CKD_DOMAIN)
     return sig.pair(G2Element.generator()) == Hm.pair(pk)
 
 
+APP_ID_DERIVATION_PREFIX = "near-mpc v0.1.0 app_id derivation:"
+
+
+def derive_app_id(account_id: str, path: str) -> bytes:
+    return sha3_256(f"{APP_ID_DERIVATION_PREFIX}{account_id},{path}".encode()).digest()
```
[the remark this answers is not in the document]
gilcu3: this is python code, so cannot be done unfortunately :(

crates/contract/tests/sandbox/common.rs
```diff
@@ -697,14 +698,26 @@ pub fn create_response_ckd(
...
+        derivation_path,
+    );
 
-    let app_id = account_id.as_bytes();
+    let app_id = derive_app_id(account_id, derivation_path);
     let app_pk: ckd::ElementG1 = app_public_key.into_contract_type();
-    let msk = signing_key;
-    let big_s = hash_to_curve(app_id) * msk;
+    let msk = key_package.private_share.to_scalar();
+    let pk = key_package.public_key.to_element().to_compressed();
```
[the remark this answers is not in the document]
gilcu3: true, let me see if I can call from there

SimonRastikian (approved):

PR #1633, opened 2025-12-09
chore: update reference to ts repo

### LINKED ISSUE #1628: Update reference to ts repo
### Background

After https://github.com/near/threshold-signatures/pull/246 and https://github.com/near/mpc/pull/1627 are merged, we need to update the reference to the ts repo so that the transition to CKD with derivation paths is complete.

### User Story

_No response_

### Acceptance Criteria

The reference to the ts repo is updated, and the tests that will fail are fixed.

_No response_

gilcu3:
Closes #1628

### DIFF
--- a/crates/node/Cargo.toml
+++ b/crates/node/Cargo.toml
@@ -46,7 +46,6 @@ near-time = { workspace = true }
 node-types = { workspace = true }
 prometheus = { workspace = true }
 rand = { workspace = true }
-# rand_core = { workspace = true }
 rocksdb = { workspace = true }
 rustls = { workspace = true }
 serde = { workspace = true }


Cargo.toml
```diff
@@ -123,8 +123,8 @@ sha3 = "0.10.8"
 signature = "2.2.0"
 tempfile = "3.22.0"
 thiserror = "2.0.12"
-# Last commit on main branch as of 30-10-2025
-threshold-signatures = { git = "https://github.com/near/threshold-signatures", rev = "53f928c9e7d21f5858fd928b7524c8ce8bf8b854" }
+# Last commit on main branch as of 09-12-2025
+threshold-signatures = { git = "https://github.com/near/threshold-signatures", rev = "1982f1b4de944d6d3011ee3f4ebbecfc9dd31cd9" }
```
DSharifi: We really should start creating releases over there as well 😅

DSharifi (approved):

crates/node/Cargo.toml
```diff
@@ -46,6 +46,7 @@ near-time = { workspace = true }
 node-types = { workspace = true }
 prometheus = { workspace = true }
 rand = { workspace = true }
+# rand_core = { workspace = true }
```
kevindeforth: is there a reason to keep this comment?

gilcu3: leftover that needs to be removed, at the end I managed to achieve what I needed without it

crates/node/src/providers/ckd/sign.rs
```diff
@@ -139,8 +139,8 @@ impl MpcLeaderCentricComputation<Option<(ElementG1, ElementG1)>> for CKDComputat
             cs_participants.as_slice(),
             channel.sender().get_leader().into(),
             channel.my_participant_id().into(),
-            self.keygen_output.private_share,
-            self.app_id.as_ref(),
+            self.keygen_output,
+            AppId::try_new(self.app_id.as_ref())?,
```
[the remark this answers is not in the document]
gilcu3: still, I will move the conversion earlier to the indexer as you mention, which makes more sense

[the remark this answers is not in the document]
kevindeforth: Thanks!
> Although now I notice that it cannot fail, as the new failurr case is only an upper bound we put in the value size, and the contract now enforces that size to be 32 bytes.

Yeah, I figured this is one of these "should never happen" failures, but it felt somehow wrong to construct the computation, put it in the queue and attempt it, just to realize it would have never needed to go in the queue to begin with.

kevindeforth (approved):

PR #1637, opened 2025-12-09
test: migration system test

### LINKED ISSUE #1294: Migration Service: system test
### Background

Write a system test for e2e migration service.

### User Story

_No response_

### Acceptance Criteria

Migrate a python test node via backup-service and web-endpoints.

_No response_

kevindeforth:
resolves #1294


crates/backup-cli/src/cli.rs
```diff
@@ -44,25 +44,26 @@ pub struct RegisterArgs {
 
 #[derive(clap::Args, Debug)]
 pub struct GetKeysharesArgs {
-    /// URL of the MPC node to retrieve keyshares from.
+    /// host address of the MPC node to retrieve keyshares from (`host:port`, not a url).
     #[arg(long, env)]
-    pub mpc_node_url: String,
+    pub mpc_node_address: String,
```
DSharifi: Same here about using https://doc.rust-lang.org/stable/std/net/enum.SocketAddr.html

pytest/common_lib/shared/__init__.py
```diff
@@ -222,27 +223,34 @@ def start_neard_cluster_with_cleanup(
...
+
+Loader.add_constructor("!Block", block_constructor)
 
 
 def generate_mpc_configs(
     num_mpc_nodes: int,
     num_respond_aks: int,
-    presignatures_to_buffer: Optional[int],
-) -> List[Candidate]:
+    presignatures_to_buffer: int | None,
```
DSharifi: Super nit:

```suggestion
    presignatures_to_buffer: Optional[int],
```

pytest/common_lib/shared/__init__.py
```diff
@@ -376,53 +415,68 @@ def start_cluster_with_mpc(
     (key, nonce) = cluster.contract_node.get_key_and_nonce()
     create_txs = []
     access_txs = []
-    mpc_nodes: List[MpcNode] = []
-    pytest_keys_per_node = []
-    secondary_near_account: Optional[NearAccount] = None
+    mpc_nodes: list[MpcNode] = []
+    pytest_keys_per_node = [[] for _ in range(len(configs))]
```
DSharifi: Nit:
This feels more "pythonic"
```suggestion
    pytest_keys_per_node = [[] for _ in configs]
```

crates/backup-cli/src/backup.rs
```diff
@@ -40,13 +40,15 @@ pub async fn run_command(args: cli::Args) {
...
+
         cli::Command::GetKeyshares(subcommand_args) => {
             let home_dir = PathBuf::from(args.home_dir);
             let secrets_storage =
                 adapters::secrets_storage::SharedJsonSecretsStorage::<File>::open_read(
                     home_dir.as_path(),
                 )
                 .await
+                .inspect_err(|e| eprintln!("Failed to create storage: {e}"))
                 .expect("failed to create secrets storage");
```
DSharifi: This message is already output from the panic that will arise from `expect`?

crates/backup-cli/src/backup.rs
```diff
@@ -164,6 +166,7 @@ pub async fn get_keyshares(
     let keyshare = mpc_p2p_client
         .get_keyshares(&keyset)
         .await
+        .inspect_err(|err| eprintln!("{err:?}"))
         .expect("fail to get key shares");
```
DSharifi: Same here. The expect will panic with your error message.

pytest/common_lib/shared/__init__.py
```diff
@@ -1,18 +1,18 @@
+from dataclasses import dataclass
 import json
 import os
 import pathlib
 import subprocess
 import sys
-from typing import List, Optional, Tuple, cast
-import typing
+from typing import List, cast
```
gilcu3: do we still need `List`? it seemed we were switching to `list`

.github/workflows/ci.yml
```diff
@@ -268,6 +268,9 @@ jobs:
       - name: Build mpc node
         run: cargo build -p mpc-node --release --features=network-hardship-simulation
 
+      - name: Build backup-cli
+        run: cargo build -p backup-cli --release
```
DSharifi:
```suggestion
        run: cargo build -p backup-cli --release --locked
```

pytest/common_lib/shared/__init__.py
```diff
@@ -268,14 +276,24 @@ def generate_mpc_configs(
...
     if presignatures_to_buffer:
         cmd = cmd + (
             "--desired-presignatures-to-buffer",
             str(presignatures_to_buffer),
         )
     subprocess.run(cmd)
 
-    candidates = []
+    for i in migrating_nodes:
+        responders.append(responders[i])
```
gilcu3: is this correct? looks quite suspicious 😶‍🌫️

crates/node/src/cli.rs
```diff
@@ -580,6 +585,45 @@ impl Cli {
...
+            threshold,
+            desired_triples_to_buffer,
+            desired_presignatures_to_buffer,
+            desired_responder_keys_per_participant,
+        )
+        .await
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    async fn _run_generate_test_configs(
```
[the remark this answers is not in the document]
kevindeforth: I thought it was cleaner this way, but can inline if you prefer.

pytest/common_lib/constants.py
```diff
@@ -3,6 +3,9 @@
 
 MPC_REPO_DIR = pathlib.Path(__file__).resolve().parents[2]
 MPC_BINARY_PATH = os.path.join(MPC_REPO_DIR / "target" / "release", "mpc-node")
+BACKUP_SERVICE_BINARY_PATH = os.path.join(
+    MPC_REPO_DIR / "target" / "release", "backup-cli"
+)
 CONFIG_PATH = os.path.join(MPC_REPO_DIR / "pytest" / "config.json")
```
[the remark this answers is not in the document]
kevindeforth: ooh, nice!

pytest/exec_pytest.sh
```diff
@@ -120,6 +120,12 @@ if ! log_output bash -c "cd '$GIT_ROOT' && cargo build --quiet --color=always -p
     exit 1
 fi
 
+printf "\nBuilding backup cli"
+if ! log_output bash -c "cd '$GIT_ROOT' && cargo build --quiet --color=always -p backup-cli --release"; then
```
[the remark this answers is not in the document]
kevindeforth: I don't think this would be beneficial here - this script is intended to simplify our developers life and not meant for use in production. If the lock file needs to be updated, then that's not a defect.

pytest/common_lib/shared/__init__.py
kevindeforth: This is very much intended. Let me see if I can circle back to this comment later for an explanation.

crates/node/src/cli.rs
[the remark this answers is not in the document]
kevindeforth: > Why have _run_generate_test_configs as a separate function when it's just a continuation of run_generate_test_configs?

ok, Reynaldo also commented on that, will change. https://github.com/near/mpc/pull/1637#discussion_r2605997089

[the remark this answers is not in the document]
netrome: I'd find it a bit cleaner to inline and break out the logic for extending the participants and responders to a separate helper function instead.

Something like:
```rust
fn run_generate_test_configs(...) {
    let participants = duplicate_migrating_accounts(&migrating_nodes, participants);
    let responders = duplicate_migrating_accounts(&migrating_nodes, responders);
    
    ... same logic as before
}
```

pytest/tests/test_migration_service.py
```diff
@@ -0,0 +1,139 @@
...
+        while n_try < 50:
+            if running_state_matches_participant_key(
+                cluster, migrating_node.account_id(), target_node.p2p_public_key
+            ):
+                print("successfully migrated node")
+                break
+            else:
+                print("waiting for migration to conclude")
+                time.sleep(1)
+                n_try += 1
```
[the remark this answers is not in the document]
kevindeforth: oh, nice catch! That is no longer needed and was from before I put that loop there. Must have overlooked removing it.

I was wondering why it always succeeded first try xD

pytest/tests/test_migration_service.py
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env python3
+"""
+Tests node migrations.
+Starts a cluster with 2 participating nodes and two target nodes.
+Migrates nodes #0 to #2 and node #1 to #3
+"""
```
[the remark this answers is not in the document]
kevindeforth: I am not sure what degree of detail you would like here, as I think the string is quite descriptive. Do you have a suggestion?

pytest/common_lib/shared/__init__.py
```diff
@@ -222,27 +223,34 @@ def start_neard_cluster_with_cleanup(
...
+
+class Loader(yaml.SafeLoader):
+    pass
+
+
+def block_constructor(loader, node):
+    return loader.construct_mapping(node)
+
+
+Loader.add_constructor("!Block", block_constructor)
```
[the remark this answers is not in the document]
netrome: +1 I tried looking at the source but I don't get much wiser.

```python
    @classmethod
    def add_constructor(cls, tag, constructor):
        if not 'yaml_constructors' in cls.__dict__:
            cls.yaml_constructors = cls.yaml_constructors.copy()
        cls.yaml_constructors[tag] = constructor
```

So this one seems to register this constructor function somewhere. But our constructor is just a thin wrapper over `loader.construct_mapping` which looks a bit opaque to me:

```python
    def construct_mapping(self, node, deep=False):
        if not isinstance(node, MappingNode):
            raise ConstructorError(None, None,
                    "expected a mapping node, but found %s" % node.id,
                    node.start_mark)
        mapping = {}
        for key_node, value_node in node.value:
            key = self.construct_object(key_node, deep=deep)
            if not isinstance(key, collections.abc.Hashable):
                raise ConstructorError("while constructing a mapping", node.start_mark,
                        "found unhashable key", key_node.start_mark)
            value = self.construct_object(value_node, deep=deep)
            mapping[key] = value
        return mapping
```

netrome (commented):
Some more comments + questions.

pytest/common_lib/shared/__init__.py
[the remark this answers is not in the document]
kevindeforth: I removed it in https://github.com/near/mpc/pull/1637/commits/8304e6cbdd0f50bb11b39e99c8f801cc95478200, we already have that logic in `yaml_safeloader.py`.
We have the following custom tag in our config.yaml:
```yaml
  sync_mode: !Block
```
This tells pyyaml to ignore this and not attempt to construct it. It will just set `sync_mode: None`.

pytest/common_lib/shared/backup_service.py
```diff
@@ -0,0 +1,115 @@
...
+        pk_bytes = sk.verify_key.encode()
+        pk_b58 = base58.b58encode(pk_bytes).decode()
+        return f"ed25519:{pk_b58}"
+
+
+@dataclass
+class BackupService:
+    home_dir: str = field(init=False)
+
+    def __post_init__(self):
```
netrome: Let's use a normal constructor here:

```suggestion
    def __init__(self):
```

pytest/common_lib/shared/__init__.py
```diff
@@ -121,7 +123,7 @@ def deserialize_key(account_id: str, key: str) -> Key:
 def sign_create_account_with_multiple_access_keys_tx(
     creator_key: Key,
     new_account_id: str,
-    keys: List[Key],
+    keys: list[Key],
```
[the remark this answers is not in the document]
kevindeforth: Yeah, I was also confused by that. But my linter and ChatGPT tell me this is how the young kids to it nowadays. Seems like `typing.List[]` is deprecated now: https://docs.python.org/3/library/typing.html#deprecated-aliases

pytest/common_lib/shared/backup_service.py
```diff
@@ -0,0 +1,115 @@
...
+    def near_public_key(self) -> str:
+        sk = SigningKey(self.p2p_private_key)
+        pk_bytes = sk.verify_key.encode()
+        pk_b58 = base58.b58encode(pk_bytes).decode()
+        return f"ed25519:{pk_b58}"
+
+
+@dataclass
+class BackupService:
+    home_dir: str = field(init=False)
```
netrome:
```suggestion
```

pytest/common_lib/shared/backup_service.py
```diff
@@ -0,0 +1,115 @@
...
+        )
+
+    def near_public_key(self) -> str:
+        sk = SigningKey(self.p2p_private_key)
+        pk_bytes = sk.verify_key.encode()
+        pk_b58 = base58.b58encode(pk_bytes).decode()
+        return f"ed25519:{pk_b58}"
+
+
+@dataclass
```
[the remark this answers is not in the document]
kevindeforth: Good point, a plain class makes more sense. Changed in https://github.com/near/mpc/pull/1637/commits/31daa5ad20d4ec5f93b30c335592d59f1ddd87ce

pytest/exec_pytest.sh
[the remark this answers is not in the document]
gilcu3: I think the simple path to avoid dev headaches is to assume that `--locked` is passed everywhere by default, cannot see anyone wanting to use latest deps (instead of the ones in the lock file) within a test script. But if you disagree I am fine with leaving as is (the status quo). We could raise that later, it is really orthogonal to the PR

pytest/tests/test_migration_service.py
```diff
@@ -0,0 +1,142 @@
...
+        while n_try < 50:
+            if running_state_matches_participant_key(
+                cluster, migrating_node.account_id(), target_node.p2p_public_key
+            ):
+                print("successfully migrated node")
+                break
+            else:
+                print("waiting for migration to conclude")
+                time.sleep(1)
+                n_try += 1
```
netrome: Nit: It would be nice to break this retry logic out to some helper function. It's unfortunate that we don't have a way to await this state change without having to poll.

netrome (approved):
Thanks for the updates. Looks good to me. Some minor comments on the actual test case.

pytest/tests/test_migration_service.py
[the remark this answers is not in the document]
DSharifi: You could go in more detail on what's being checked during the migration, for example that liveness is checked by sending signature and CKD requests.

Some of our other pytests also go in detail like this:
- https://github.com/near/mpc/blob/b4f7acb5424459ae1c75a0fa46f1003c07ab92a9/pytest/tests/test_cancellation_of_key_resharing.py#L18-L29
- https://github.com/near/mpc/blob/b4f7acb5424459ae1c75a0fa46f1003c07ab92a9/pytest/tests/test_key_event.py#L22-L31

crates/node/src/cli.rs
```diff
@@ -559,36 +562,56 @@ impl Cli {
...
+
     #[allow(clippy::too_many_arguments)]
-    async fn run_generate_test_configs(
+    fn run_generate_test_configs(
         &self,
         output_dir: &str,
-        participants: &[AccountId],
-        responders: &[AccountId],
+        participants: Vec<AccountId>,
+        responders: Vec<AccountId>,
```
pbeza: Why this change? I assumed we generally prefer passing `&[...]`. It’s more flexible — callers can pass a `Vec<T>`, arrays, slices, or anything that dereferences to a slice.

kevindeforth: we pass by value, not by reference.

netrome (approved):
Still looks good to me :)

PR #1645, opened 2025-12-10
feat: support for robust ecdsa signature in the mpc-node

### LINKED ISSUE #1640: MPC nodes processes Robust ECDSA signature requests.
### Background

Once the contract has been extended to support the new Robust ECDSA signature scheme variant, we should upgrade the MPC node indexer to observe these requests and run the Robust ECDSA protocol to handle them.

Updated detailed plan:

- [x] Create a single small PR with the first commit introducing the SignatureSheme variant. #1670 
- [x] Break out the pytest reorganization (not including the Robust ECDSA test) and CI beefiness bump to a single PR. #1671 
- [x] Small refactor in node integration tests, to make adding ECDSA tests easier later. #1672 
- [x] A PR with the singature provider implementation, including more unit tests if possible. Ideally we'd also break out presignature generation to a separate PR as that's quite a big function that should also be implemented and tested in isolation. However these are quite tightly coupled so if it's hard to break them out I can live with reviewing them together. #1679
- [ ] Integrating the provider in the node and adding integration tests in the node, and system tests with pytest. #1682

### User Story

As a node operator I want to be able to support the new Robust ECDSA scheme.

### Acceptance Criteria

The MPC nodes can process Robust ECDSA signature requests.

[Thread](https://nearone.slack.com/archives/C0912BTG51T/p1765836138470169?thread_ts=1765374628.931179&cid=C0912BTG51T)

gilcu3:
Closes #1640

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ on:
 jobs:
   docker-tee-build:
     name: "Build MPC Node TEE Docker image"
-    runs-on: warp-ubuntu-2204-x64-8x
+    runs-on: warp-ubuntu-2204-x64-16x
     timeout-minutes: 60
     permissions:
       contents: read
--- a/.github/workflows/docker_build_node.yml
+++ b/.github/workflows/docker_build_node.yml
@@ -13,7 +13,7 @@ on:
 jobs:
   build-and-push-images:
     name: "Build and push Docker node image with commit hash"
-    runs-on: warp-ubuntu-2204-x64-8x
+    runs-on: warp-ubuntu-2204-x64-16x
     permissions:
       contents: read
 
--- a/.github/workflows/docker_build_node_gcp.yml
+++ b/.github/workflows/docker_build_node_gcp.yml
@@ -13,7 +13,7 @@ on:
 jobs:
   build-and-push-images:
     name: "Build and push Docker node gcp image with commit hash"
-    runs-on: warp-ubuntu-2204-x64-8x
+    runs-on: warp-ubuntu-2204-x64-16x
     permissions:
       contents: read
 


crates/node/src/providers/robust_ecdsa/presign.rs
```diff
@@ -0,0 +1,296 @@
...
+    /// Continuously generates presignatures, trying to maintain the desired number of
+    /// presignatures available, using the desired number of concurrent computations as
+    /// specified in the config. Most of the time, this process would be waiting for
+    /// more triples to be generated, as presignature generation is significantly faster
+    /// than triple generation.
+    ///
+    /// Generated triples will be written to `presignature_store` as owned triples. Note
+    /// that this function does not take care of the passive side of presignature
+    /// generation (i.e. other participants of the computations this function initiates),
+    /// so that needs to be separately handled.
```
gilcu3: need to update, now this does not have triples

crates/node/src/providers/robust_ecdsa/presign.rs
```diff
@@ -0,0 +1,293 @@
...
+                    &running_participants,
+                ) {
+                    Ok(participants) => participants,
+                    Err(e) => {
+                        tracing::warn!(
+                            "Can't choose active participants for a triple: {}. Sleeping.",
+                            e
+                        );
+                        // that should not happen often, so sleeping here is okay
+                        tokio::time::sleep(Duration::from_millis(100)).await;
```
gilcu3: we should certainly have an issue for removing these sleep ops, not doing so just in case something actually breaks

netrome (commented):
Shallow skimming. Some nits, but so far so good. Happy to go deeper once the PR is ready.

crates/node/src/providers/robust_ecdsa/sign.rs
```diff
@@ -0,0 +1,216 @@
...
+use threshold_signatures::ecdsa::robust_ecdsa::{PresignOutput, RerandomizedPresignOutput};
+use threshold_signatures::ecdsa::{RerandomizationArguments, Signature, SignatureOption};
+use threshold_signatures::frost_secp256k1::VerifyingKey;
+use threshold_signatures::participants::Participant;
+use threshold_signatures::ParticipantList;
+use tokio::time::timeout;
+
+impl RobustEcdsaSignatureProvider {
+    pub(super) async fn make_signature_leader(
+        self: Arc<Self>,
```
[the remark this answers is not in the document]
gilcu3: I will go ahead and fix for all signatures, seems weird why was it added in the first place, maybe just tech-debt

crates/node/src/p2p.rs
```diff
@@ -520,51 +521,82 @@ pub mod testing {
...
+        pub const BASE_PORT: u16 = 10000;
+        // This constant must be equal to the total number of ports defined below
+        pub const TOTAL_DEFINED_PORTS: u16 = 19;
+        // Maximum number of nodes that can be handled without port collisions
+        pub const MAX_NODES: u16 = 10;
+        // Maximum number of cases that can be handled without port collisions
+        pub const MAX_CASES: u16 = 4;
+        // Each function below corresponds to a port per node. Each defines an offset,
+        // and all offsets must be different
+        pub const TOTAL_PORTS_PER_NODE: u16 = 3;
```
gilcu3: changes in this file were needed to be able to use rstest with  different domains and still avoid port collisions

crates/node/src/tests.rs
```diff
@@ -47,8 +47,8 @@ mod research;
 mod resharing;
 
 const DEFAULT_BLOCK_TIME: std::time::Duration = std::time::Duration::from_millis(300);
-const DEFAULT_MAX_PROTOCOL_WAIT_TIME: std::time::Duration = std::time::Duration::from_secs(30);
-const DEFAULT_MAX_SIGNATURE_WAIT_TIME: std::time::Duration = std::time::Duration::from_secs(30);
+const DEFAULT_MAX_PROTOCOL_WAIT_TIME: std::time::Duration = std::time::Duration::from_secs(60);
+const DEFAULT_MAX_SIGNATURE_WAIT_TIME: std::time::Duration = std::time::Duration::from_secs(60);
```
gilcu3: unfortunately the old timeout is now too small

crates/node/src/providers/robust_ecdsa.rs
```diff
@@ -0,0 +1,251 @@
...
+/// This function translates the current threshold from the contract
+/// to the threshold expected by the robust-ecdsa scheme, which
+/// is semantically different.
+/// The function should be no longer needed when these issues are solved:
+/// https://github.com/near/threshold-signatures/issues/255
+/// https://github.com/near/mpc/issues/1649
+pub(super) fn translate_threshold(threshold: usize, number_of_participants: usize) -> usize {
+    let number_of_signers = get_number_of_signers(threshold, number_of_participants);
+    (number_of_signers - 1) / 2
+}
```
netrome: We should add some unit tests for this function. This should be trivial but it would illustrate very explicitly how the values translate.

crates/node/src/providers/robust_ecdsa.rs
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+pub(super) fn get_number_of_signers(threshold: usize, _number_of_participants: usize) -> usize {
+    // TODO: this is the case for the other schemes, might not be our best choice for this one
+    threshold
+}
```
netrome: This looks incorrect. Shouldn't it be 2t + 1?

crates/node/src/providers/robust_ecdsa/presign.rs
```diff
@@ -0,0 +1,302 @@
...
+use threshold_signatures::participants::Participant;
+
+pub struct PresignatureStorage(DistributedAssetStorage<PresignOutputWithParticipants>);
+
+impl PresignatureStorage {
+    pub fn new(
+        clock: Clock,
+        db: Arc<SecretDB>,
+        my_participant_id: ParticipantId,
+        alive_participant_ids_query: Arc<dyn Fn() -> Vec<ParticipantId> + Send + Sync>,
```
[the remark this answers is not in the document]
netrome: Looking more on this implementation, it feels weird to pass a closure containing the network client over just passing a reference or clone of the client itself. Smells like over-engineering.

crates/node/src/providers/robust_ecdsa/presign.rs
netrome: Ouch yeah this is bad. Sleeping here is not okay, it makes anything touching these code paths untestable.

crates/node/src/providers/robust_ecdsa/presign.rs
```diff
@@ -0,0 +1,302 @@
...
+}
+
+/// Performs an MPC presignature operation. This is shared for the initiator
+/// and for passive participants.
+pub struct PresignComputation {
+    threshold: usize,
+    keygen_out: KeygenOutput,
+}
+
+#[async_trait::async_trait]
```
netrome: Why do we use `async_trait`? This has been redundant since 1.75 - clearly stated in the crate readme that any human would read before introducing the dependency. This pre-dates the MPC node code base, which makes this smell very vibe-coded, as LLMs often still add `async_trait` whenever they get to roam too freely.

Anyway, that's what we're here to fix. Welcome to the AI-slop cleanup crew :sweat_smile:

pytest/common_lib/shared/__init__.py
```diff
@@ -553,3 +571,211 @@ def new_signer_key(account_id: str) -> Key:
...
+    return result if result is not None else 0
+
+
+def get_node_metrics_all_nodes(cluster: MpcCluster):
+    number_nodes = len(cluster.mpc_nodes)
+
+    network_metrics = [NodeMetrics(0, 0, 0, 0, 0) for _ in range(number_nodes)]
+    for i in range(len(cluster.mpc_nodes)):
+        network_metrics[i].queue_size = get_metric_value_for_node(
+            cluster, "mpc_pending_signatures_queue_size", i
```
netrome: These helpers are just moved around right? It would be nice to have them in a separate helper module to not pollute `__init__.py`. Also this refactor could be broken out to a separate PR as you said in the meeting today.

pytest/tests/robust_ecdsa/conftest.py
```diff
@@ -0,0 +1,37 @@
...
+PRESIGNATURES_TO_BUFFER = 6
+
+
+@pytest.fixture(scope="package")
+def shared_cluster():
+    """
+    Spins up a cluster with three nodes per test, initializes the contract and adds
+    domains. Returns the cluster in a running state.
+    """
+    # TODO: this does not work when using 7 nodes
```
netrome: Dangling TODO - we should create an issue for this.

netrome (changes requested):
Initial pass. Mostly looks good to me. Some dangling todos and other minor things. However, this is pretty huge so just for the sake of organizing our conversations and maintaining a decent commit history I'd like to split this out to separate PRs. I'm thinking:

1. Create a single small PR with the first commit introducing the `SignatureSheme` variant.
2. Break out the pytest reorganization (not including the Robust ECDSA test) and CI beefiness bump to a single PR.
3. A PR with the singature provider implementation, including more unit tests if possible. Ideally we'd also break out presignature generation to a separate PR as that's quite a big function that should also be implemented and tested in isolation. However these are quite tightly coupled so if it's hard to break them out I can live with reviewing them together.
4. Integrating the provider in the node and adding system tests.

Let me know what you think, and if you see other ways we could slice this up.

crates/node/src/providers/robust_ecdsa.rs
gilcu3: the inputs refer to the threshold as stored in the contract. As we agreed a few days ago, and as it happens for other schemes, the number of signers equals the threshold is our invariant. The other function that computes the robust ecdsa threshold is the one using the equation you mention

crates/node/src/providers/robust_ecdsa.rs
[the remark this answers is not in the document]
gilcu3: I am happy to make providers testable, but not sure how easy that is, considering the current architecture. If anything becomes too complicated, I think would be better to add this without too many differences from the others, then as a follow-up finally fix all of them together. In the current state it should actually be easy to avoid replicating code for key generation and resharing, so that should go in its own PR.

crates/node/src/providers/robust_ecdsa/presign.rs
gilcu3: :cry: alright, will clean :smile:

netrome: Let's not do it in this PR though since the trait definition is from the existing code.

Issue #1634, opened 2025-12-09
Avoid parsing tcb_info.json inside contract to reduce TEE verification gas costs


barakeinav1:
### Background

Our TEE verification flow currently parses the full [tcb_info.json ](https://github.com/near/mpc/blob/406ce734acc1899cc83fce842a7ee4e65df8956c/crates/mpc-attestation/assets/tcb_info.json) (and also tcb_info_dev.json)  inside the contract for every attestation that is verified. This significantly increases gas usage and contributes to verify_tee costing ~270 TGas per call (for 2 users). The relevant values from tcb_info.json.

improvements:
1. Only 5 fields of the json file are really needed.
2. all the values are known in compile time, and should not require parsing a json that could fail.
3. Move them to a more efficient (storage/gas) storage location /stract 

open:
How do we handle leftover data from the existing json files?

### User Story

As MPC user, I'd like my contract to be efficient as possible. allowing many MPC nodes participants. 

### Acceptance Criteria

New solution implemented.
1. Need to test upgrade flow.
2. test regular (existing) attestation verification follow
3. if possible  - measure gas costs

PR #1646, opened 2025-12-10
refactor(attestation): move TCB measurement extraction to build time

barakeinav1:
thanks all for the review, but I'm closing this in favor of[#1659](https://github.com/near/mpc/pull/1659)

related  #1634

**This PR removes all runtime JSON parsing from the contract at runtime.**  
Instead, TCB measurements (MRTD, RTMR0–2, key-provider digest) are extracted **at build time** from readable JSON files and compiled into static Rust structures.

This makes attestation verification:

- faster  
- cheaper (less gas)  
- safer (no JSON decode failures)  
- easier to maintain

### DIFF
--- a/crates/mpc-attestation/build.rs
+++ b/crates/mpc-attestation/build.rs
@@ -1,179 +1,25 @@
 use std::env;
-use std::fs::{self, File};
-use std::io::Write;
 use std::path::PathBuf;
+
+include!("build_support/lib.rs");
+
 
 const ASSETS_DIR_NAME: &str = "assets";
 
 fn main() {
-    // Find assets directory
-    let manifest_dir =
-        env::var("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR must be set by Cargo");
-
-    let assets_dir = PathBuf::from(&manifest_dir).join(ASSETS_DIR_NAME);
-
+    // Rerun if directory changes
     println!("cargo:rerun-if-changed={}", ASSETS_DIR_NAME);
 
-    // First pass: ensure directory exists + register rerun triggers
-    let entries = fs::read_dir(&assets_dir).unwrap_or_else(|e| {
-        panic!(
-            "Failed to read assets directory '{}': {}.\n\
-             This directory must exist and contain tcb_info*.json files.",
-            assets_dir.display(),
-            e
-        )
-    });
+    let manifest_dir = env::var("CARGO_MANIFEST_DIR")
+        .expect("CARGO_MANIFEST_DIR must be set by Cargo");
 
-    let mut measurement_files = Vec::new();
+    let in_dir = PathBuf::from(manifest_dir).join(ASSETS_DIR_NAME);
 
-    for entry in entries {
-        let entry = entry.unwrap_or_else(|e| {
-            panic!(
-                "Failed to read an entry inside '{}': {}",
-                assets_dir.display(),
-                e
-            )
-        });
+    let out_dir = PathBuf::from(env::var("OUT_DIR")
+        .expect("OUT_DIR must be set by Cargo"));
 
-        let path = entry.path();
-
-        if path.extension().and_then(|x| x.to_str()) == Some("json") {
-            println!("cargo:rerun-if-changed={}", path.display());
-
-            // Only include tcb_info*.json
-            let filename = path
-                .file_name()
-                .and_then(|x| x.to_str())
-                .unwrap_or_else(|| {
-                    panic!(
-                        "Found a JSON file with invalid UTF-8 filename inside '{}': {:?}",
-                        assets_dir.display(),
-                        path.file_name()
-                    )
-                });
-
-            if filename.starts_with("tcb_info") {
-                measurement_files.push(path);
-            }
-        }
-    }
-
-    if measurement_files.is_empty() {
-        panic!(
-            "No tcb_info*.json files found in directory '{}'. \
-             Add files such as tcb_info.json or tcb_info_dev.json.",
-            assets_dir.display()
-        );
-    }
-
-    // Write generated Rust file
-    let out_dir = PathBuf::from(env::var("OUT_DIR").expect("OUT_DIR must be provided by Cargo"));
     let out_file = out_dir.join("measurements_generated.rs");
 
-    let mut f = File::create(&out_file).unwrap_or_else(|e| {
-        panic!(
-            "Failed to create output file '{}': {}",
-            out_file.display(),
-            e
-        )
-    });
-
-    writeln!(
-        f,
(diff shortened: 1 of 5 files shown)


crates/mpc-attestation/build.rs
```diff
@@ -0,0 +1,96 @@
...
+                     mrtd: {:?}, \
+                     rtmr0: {:?}, \
+                     rtmr1: {:?}, \
+                     rtmr2: {:?} \
+                 }}, \
+                 key_provider_event_digest: {:?}, \
+             }},",
+            mrtd, rtmr0, rtmr1, rtmr2, key_provider_digest
+        )
+        .unwrap();
```
gilcu3: I naively thought we could do this better with some crate

gilcu3 (commented):
Just commenting as some cleanup are still coming, but overall looks good, did not check if it works as expected though

crates/mpc-attestation/src/attestation.rs
```diff
@@ -6,6 +6,16 @@ use attestation::{
...
+include!(concat!(env!("OUT_DIR"), "/measurements_generated.rs"));
+
+pub struct ExpectedMeasurementsSet;
+
+impl ExpectedMeasurementsSet {
+    pub fn all() -> &'static [ExpectedMeasurements] {
+        EXPECTED_MEASUREMENTS
+    }
+}
+
```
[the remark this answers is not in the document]
gilcu3: I am curious now :smile: what would be the issue in that case?

crates/mpc-attestation/build.rs
```diff
@@ -0,0 +1,108 @@
...
+                .unwrap()
+                .starts_with("tcb_info")
+        {
+            measurement_files.push(path);
+        }
+    }
+
+    // Output file
+    let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
+    let out_file = out_dir.join("measurements_generated.rs");
```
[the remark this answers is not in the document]
barakeinav1: done

crates/mpc-attestation/build.rs
barakeinav1: didn't find any official create.

crates/mpc-attestation/measurements_build.md
```diff
@@ -0,0 +1,112 @@
...
+---
+
+## ➕ Adding a New Measurements File
+
+1. Obtain the new TCB info JSON from a node:
+
+```bash
+curl "http://<node-ip>:<port>/public_data" \
+  | jq -r '.tee_participant_info.Dstack.tcb_info' \
+  > crates/mpc-attestation/assets/tcb_info_new.json
```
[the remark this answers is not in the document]
barakeinav1: updated doc a bit

crates/mpc-attestation/measurements_build.md
```diff
@@ -0,0 +1,112 @@
...
+cargo build -p mpc-attestation
+```
+
+The build script will:
+
+- parse the JSON files  
+- extract the required measurement fields  
+- decode them into fixed byte arrays  
+- generate Rust constants  
+- embed them statically into the WASM contract
```
gilcu3: this will not happen as we are not building the contract here, so better remove this line and modify the previous one to say that the constants will be embedded when the contract is built (or something similar)

crates/mpc-attestation/measurements_build.md
```diff
@@ -0,0 +1,112 @@
...
+
+```
+crates/mpc-attestation/assets/
+```
+
+2. Rebuild:
+
+```bash
+cargo clean -p mpc-attestation
+cargo build -p mpc-attestation
```
gilcu3: same, probably the clean step is no longer needed

crates/mpc-attestation/build.rs
```diff
@@ -0,0 +1,187 @@
...
+const ASSETS_DIR_NAME: &str = "assets";
+
+fn main() {
+    // Find assets directory
+    let manifest_dir =
+        env::var("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR must be set by Cargo");
+
+    let assets_dir = PathBuf::from(&manifest_dir).join(ASSETS_DIR_NAME);
+
+    println!("cargo:rerun-if-changed={}", ASSETS_DIR_NAME);
```
gilcu3:
```suggestion
    println!("cargo::rerun-if-changed={}", ASSETS_DIR_NAME);
```

https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed

crates/mpc-attestation/src/attestation.rs
```diff
@@ -6,6 +6,16 @@ use attestation::{
...
+///
+/// This combines prod/dev (or any future) measurement JSON files
+/// into a single slice generated at build time.
+pub fn all_expected_measurements() -> &'static [ExpectedMeasurements] {
+    EXPECTED_MEASUREMENTS
+}
+
 pub use attestation::attestation::{DstackAttestation, VerificationError};
 
 use mpc_primitives::hash::{LauncherDockerComposeHash, MpcDockerImageHash};
```
gilcu3: these should be below all the `use` statements

gilcu3 (changes requested):
Looks good overall, but found a bug in the cargo annotation and noticed we are passing the path parameter to several functions that not need it. Added a comment with a suggestion on how to achieve that.

crates/mpc-attestation/build.rs
```diff
@@ -0,0 +1,187 @@
...
+                "Failed to read an entry inside '{}': {}",
+                assets_dir.display(),
+                e
+            )
+        });
+
+        let path = entry.path();
+
+        if path.extension().and_then(|x| x.to_str()) == Some("json") {
+            println!("cargo:rerun-if-changed={}", path.display());
```
pbeza: I don’t think we need it since we already have it for the parent directory. See: https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed:
> If the path points to a directory, it will scan the entire directory for any modifications.

But please test it to double-check.

pbeza (approved):

crates/mpc-attestation/build_support/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+/// Extract one 48-byte hex measurement field (mrtd, rtmrX)
+fn decode_measurement(tcb: &Value, field: &str) -> Result<[u8; 48], String> {
+    let hex = tcb[field]
+        .as_str()
+        .ok_or_else(|| format!("Missing string field '{}'", field))?;
+
+    decode_hex(hex).map_err(|e| format!("{}: {}", field, e))
+}
+
+/// Extract key-provider digest
```
pbeza: Let’s avoid repeating what’s already clear from the function name.
```suggestion
```

PR #1663, opened 2025-12-15
perf(contract): contract should not store full attestation submission

### LINKED ISSUE #1629: contract should not store full Attestation submission
### Background

@barakeinav1 observed that gas usage is too high when deploying two nodes with TEE, and attempting to vote in a new node. We also observed that verify_tee is using [~270Tgas](https://testnet.nearblocks.io/txns/3u9t5MNgtGuBHXtUQqrenJKiRR9WEJgGforVxP5UKWTp?tab=execution#4o7VGQ3izcc5cDCGZNWAMrPwXAVucN6zLZhgQwGXtbYt) per call, which is only to verify 2-3 nodes with attestation.

We believe our current tests did not uncover this bug earlier, as all our sandbox tests that use a runtime use mocked attestations which are very cheap to verify.

The issue is that resharing votes failed due to max gas (300tgas) being exceeded. This is due to resharing votes currently re-verifying every participant + the proposed one.

### User Story

_No response_

### Acceptance Criteria

- [ ] the contract does not store the full attestation for any participant
- [ ] re-verifications only manually checks for expiry and the allowed hash
- [ ] re-run @barakeinav1 's manual test and verify that gas usage is acceptable

_No response_

DSharifi:
closes #1629

### DIFF
--- a/crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
+++ b/crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
@@ -2808,10 +2808,10 @@ expression: abi
             {
               "type": "object",
               "required": [
-                "Dtack"
+                "Dstack"
               ],
               "properties": {
-                "Dtack": {
+                "Dstack": {
                   "$ref": "#/definitions/VerifiedDstackAttestation"
                 }
               },


crates/contract-interface/src/types/attestation.rs
```diff
@@ -29,6 +29,54 @@ pub enum Attestation {
...
+    all(feature = "abi", not(target_arch = "wasm32")),
+    derive(schemars::JsonSchema)
+)]
+pub struct VerifiedDstackAttestation {
+    /// The digest of the MPC image running.
+    pub mpc_image_hash: [u8; 32],
+    /// The digest of the MPC image running.
+    pub launcher_compose_hash: [u8; 32],
+    /// Unix timestamp for when the attestation was created.
+    pub creation_time_stamp_seonds: u64,
```
barakeinav1: where do we get the time from?

crates/contract-interface/src/types/config.rs
```diff
@@ -108,6 +116,7 @@ mod tests {
             clean_tee_status_tera_gas: Some(10),
             cleanup_orphaned_node_migrations_tera_gas: Some(3),
             remove_non_participant_update_votes_tera_gas: Some(5),
+            attestation_max_validity_duration_seconds: Some(1912312),
```
barakeinav1: this equals to 22 days, why did you choose this number?

crates/contract/src/tee/tee_state.rs
```diff
@@ -197,10 +229,11 @@ impl TeeState {
         }
     }
 
-    pub fn validate_tee(
+    pub fn re_veirfy_participants(
```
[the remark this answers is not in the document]
barakeinav1: also since this method also does cleanup
consider caliing it:

reverify_and_update_participants
or
validate_and_cleanup_participants

crates/contract-interface/src/types/attestation.rs
DSharifi: The comment is s currently misleading. The timestamp is taken from "now" time argument during the initial verification, so it's the timestamp of when the contract first saw the attestation and not the creation date.

In a follow up PR we should refactor this internal implementation to use the timestamp embedded in the attestation instead.

barakeinav1 (approved):
approving with some comments to be address

crates/contract/tests/sandbox/common.rs
```diff
@@ -89,11 +89,11 @@ pub const GAS_FOR_VOTE_NEW_PARAMETERS: Gas = Gas::from_tgas(22);
 /// TODO(#1571): Gas cost for voting on contract updates. Reduced somewhat after
 /// optimization (#1617) by avoiding full contract code deserialization; there’s likely still
 /// room for further optimization.
-pub const GAS_FOR_VOTE_UPDATE: Gas = Gas::from_tgas(232);
+pub const GAS_FOR_VOTE_UPDATE: Gas = Gas::from_tgas(241);
```
gilcu3: do you remember why this bump was needed? We continue approaching the limit, although this cannot be completely solved until #1674 is solved

crates/contract/src/lib.rs
```diff
@@ -1424,12 +1426,15 @@ impl MpcContract {
...
-        {
-            panic!("Caller must be an attested participant");
-        }
+            .is_caller_an_attested_participant(participants);
+
+        assert_matches::assert_matches!(
+            attestation_check,
+            Ok(()),
+            "Caller must be an attested participant"
+        );
```
gilcu3: this is the only place where assert_matches is used in non-testing code. To avoid introducing the new dependency, it is better to simply do 
```
assert!(
    attestation_check.is_ok(),
    "Caller must be an attested participant"
);
```

crates/contract/Cargo.toml
```diff
@@ -50,6 +50,7 @@ __abi-generate = ["near-sdk/__abi-generate"]
 
 [dependencies]
 anyhow = { workspace = true }
+assert_matches = { workspace = true }
```
gilcu3: please remove this dep, I am aware it is very lean, but the least deps we have in the contract prod the better IMO. The point where it is actually used can be easily swapped

crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
```diff
@@ -2798,6 +2798,72 @@ expression: abi
...
+              "items": {
+                "type": "integer",
+                "format": "uint8",
+                "minimum": 0.0
+              },
+              "maxItems": 32,
+              "minItems": 32
+            }
+          }
+        },
```
netrome: Given that we change the interface here, I'm curious about compatibility guarantees. We've discussed this synchronously now and it seems like this would cause nodes to fail to observe attestations in previous contract versions.

We need to figure out how to make this change backwards compatible to ensure nodes can submit and observe attestations in the old contract.

crates/contract/tests/sandbox/common.rs
DSharifi: Was needed due to the change of using BtreeMap isntead of IterableMap IIRC

crates/contract/src/lib.rs
DSharifi: I agree on not needlessly adding more dependencies, but think we can make an exception for crates like `assert_matches` that don't pull any transitive deps nor contain any code except for macros;

In this case the macro just expands to:

```rust
match attestation_check {
    Ok(()) => {}
    ref e => {
        ::std::rt::panic_fmt(
            format_args!(
                "assertion failed: `{0:?}` does not match `{1}`: {2}",
                e,
                "Ok(())",
                format_args!("Caller must be an attested participant"),
            ),
        );
    }
};
```

What's nice with it is that for failures, you'll see what the error case was.

crates/contract/src/v3_0_2_state.rs
```diff
@@ -87,27 +95,47 @@ impl From<ProposedUpdates> for crate::update::ProposedUpdates {
...
+
+        let crate::ProtocolContractState::Running(running_state) = &protocol_state else {
+            env::panic_str("Contract must be in running state when migrating.");
+        };
+
+        // For the soft release we give every participant a mocked attestation.
+        // Since this upgrade has a non-backwards compatible change, instead of manually mapping the attestations
+        // we give everyone a new mock attestation again instead.
+        let threshold_parameters = &running_state.parameters.participants();
+        let tee_state = crate::TeeState::with_mocked_participant_attestations(threshold_parameters);
```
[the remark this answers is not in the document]
DSharifi: That's the idea. And even if they have, it doesn't matter if we overwrite it with a mock one.
- The mock is also valid
- Assuming someone runs a TEE image, they will re-periodically submit a new one and overwrite the mock

DSharifi:
Carved out flaky test fix into https://github.com/near/mpc/pull/1695

crates/mpc-attestation/src/attestation.rs
```diff
@@ -23,126 +21,220 @@ use crate::alloc::string::ToString;
...
 
-                dstack_attestation
+                #[cfg(any(test, debug_assertions))]
+                let accepted_measurements = [
+                    include_measurements!("assets/tcb_info.json"),
+                    include_measurements!("assets/tcb_info_dev.json"),
+                ];
+
+                #[cfg(not(any(test, debug_assertions)))]
+                let accepted_measurements = [include_measurements!("assets/tcb_info.json")];
```
barakeinav1: should we use a feature flag instead?
e.g
#[cfg(feature = "accept-dev-measurements")]

barakeinav1: also current testing setup instruction ((https://github.com/near/mpc/blob/main/docs/localnet/localnet.md)
)  only choose the production measurement 
cargo near build non-reproducible-wasm --features abi --manifest-path crates/contract/Cargo.toml --locked

libs/nearcore
DSharifi: Will revert this

barakeinav1:
I open a bug for the issue I found:
Bug: MPC node crashes if attestation submission happens before the node keys are added to the near account[#1763](https://github.com/near/mpc/issues/1763)

crates/contract-interface/src/types/attestation.rs
```diff
@@ -29,6 +29,54 @@ pub enum Attestation {
...
+    Deserialize,
+    BorshSerialize,
+    BorshDeserialize,
+)]
+#[cfg_attr(
+    all(feature = "abi", not(target_arch = "wasm32")),
+    derive(schemars::JsonSchema)
+)]
+pub enum VerifiedAttestation {
+    Dtack(VerifiedDstackAttestation),
```
barakeinav1:
```suggestion
    Dstack(VerifiedDstackAttestation),
```

crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
```diff
@@ -2798,6 +2803,72 @@ expression: abi
...
         },
+        "VerifiedAttestation": {
+          "oneOf": [
+            {
+              "type": "object",
+              "required": [
+                "Dtack"
+              ],
+              "properties": {
+                "Dtack": {
```
barakeinav1:
```suggestion
                "Dstack": {
```

crates/node/src/indexer/tx_sender.rs
```diff
@@ -207,11 +210,61 @@ async fn observe_tx_result(
...
+                &submit_participant_info_args.proposed_participant_attestation;
+
             let submitted_attestation_is_on_chain =
-                attestation_stored_on_contract.is_some_and(|stored_attestation| {
-                    stored_attestation
-                        == submit_participant_info_args.proposed_participant_attestation
-                });
+                match (stored_attestation, submitted_attestation) {
+                    (
+                        VerifiedAttestation::Dtack(verified_dstack_attestation),
```
barakeinav1:
```suggestion
                        VerifiedAttestation::Dstack(verified_dstack_attestation),
```

PR #1671, opened 2025-12-16
test: refactor pytests, several improvements preparing for robust ecdsa integration

gilcu3:
Solves part 2 of #1640

### DIFF
--- a/pytest/tests/test_cancellation_of_key_resharing.py
+++ b/pytest/tests/test_cancellation_of_key_resharing.py
@@ -8,6 +8,7 @@ sys.path.append(str(pathlib.Path(__file__).resolve().parents[1]))
 from common_lib.shared.transaction_status import assert_txn_execution_error
 from common_lib import shared
 from common_lib.contracts import load_mpc_contract
+from common_lib.constants import TRANSACTION_TIMEOUT
 from common_lib.contract_state import (
     ProtocolState,
     RunningProtocolState,
@@ -73,7 +74,7 @@ def test_cancellation_of_key_resharing():
             continue
 
         tx = node.sign_tx(cluster.mpc_contract_account(), "vote_cancel_resharing", {})
-        response = node.near_node.send_tx_and_wait(tx)
+        response = node.near_node.send_tx_and_wait(tx, timeout=TRANSACTION_TIMEOUT)
         assert_txn_execution_error(response, expected_error_msg="Not a participant")
 
     # Vote with a threshold number of the running nodes


pytest/common_lib/shared/mpc_cluster_metrics.py
gilcu3: this file needs to be separated from metrics.py because there is a cyclic dependency (used in type annotations) between the MpcCluster type and the metrics helper functions defined here.

pytest/common_lib/shared/__init__.py
```diff
@@ -184,6 +185,12 @@ def sign_add_access_keys_tx(
     return serialize_transaction(signed_tx)
 
 
+def kill_observer(observer: LocalNode):
+    observer.kill(gentle=True)
+    observer.reset_data()
+    adjust_indexing_shard(observer)
```
netrome: This is new right? Why do we need this?

netrome (approved):
Nice!

pytest/common_lib/shared/metrics.py
```diff
@@ -14,8 +15,36 @@ class IntMetricName(str, Enum):
...
+            res.requests_indexed = self.requests_indexed - other.requests_indexed
+            res.responses_indexed = self.responses_indexed - other.responses_indexed
+            res.matching_responses_indexed = (
+                self.matching_responses_indexed - other.matching_responses_indexed
+            )
+            res.mpc_cluster_failed_signatures_count = (
+                self.mpc_cluster_failed_signatures_count
+                - other.mpc_cluster_failed_signatures_count
+            )
+            return res
```
kevindeforth: Does this just return `None` if it is not an instance & do we risk silent errors?
What about the following
```suggestion
    def __sub__(self, other):
        if not isinstance(other, NodeMetrics):
            return NotImplemented
            
        res = NodeMetrics(0, 0, 0, 0, 0)
        res.queue_size = self.queue_size - other.queue_size
        res.requests_indexed = self.requests_indexed - other.requests_indexed
        res.responses_indexed = self.responses_indexed - other.responses_indexed
        res.matching_responses_indexed = (
            self.matching_responses_indexed - other.matching_responses_indexed
        )
        res.mpc_cluster_failed_signatures_count = (
            self.mpc_cluster_failed_signatures_count
            - other.mpc_cluster_failed_signatures_count
        )
        return res
```

pytest/tests/shared_cluster_tests/conftest.py
```diff
@@ -7,6 +7,9 @@
 
 from common_lib import shared, contracts, contract_state
 
+TRIPLES_TO_BUFFER = 200
+PRESIGNATURES_TO_BUFFER = 100
```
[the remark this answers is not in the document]
gilcu3: I did not want to err on the low side. Several tests do request a lot of signatures, so these values seemed fair given that they are used by several tests. I did not try to compute an upper bound, but by order of magnitude I don't think we can get away with much less

pytest/common_lib/shared/mpc_cluster_metrics.py
```diff
@@ -0,0 +1,213 @@
...
+        )
+        network_metrics[i].requests_indexed += get_metric_value_for_node(
+            cluster, "mpc_pending_ckds_queue_requests_indexed", i
+        )
+        network_metrics[i].responses_indexed += get_metric_value_for_node(
+            cluster, "mpc_pending_ckds_queue_responses_indexed", i
+        )
+        network_metrics[i].matching_responses_indexed += get_metric_value_for_node(
+            cluster, "mpc_pending_ckds_queue_matching_responses_indexed", i
+        )
```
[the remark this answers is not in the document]
gilcu3: fixed in https://github.com/near/mpc/pull/1671/commits/632c45502358162436771d1969c389fdb3c7c5d0

netrome (approved):

Issue #1306, opened 2025-10-17
don't sleep in integration tests


kevindeforth:
## Background

Even after https://github.com/near/mpc/pull/1678, we still have one sleep in the sign.rs sandbox tests. 

The sleep is required in test where we verify that in case of duplicate requests, only the most recent signature request receives a response and the first one times out (c.f. [slack thread](https://nearone.slack.com/archives/C0912BTG51T/p1765988408324409?thread_ts=1765973579.653919&cid=C0912BTG51T)).

There are a few approaches to solve this:
- change the contract implementation such that it immediately resolves the promise of the previous signature request.
- use a different test framework that allows query transaction outcomes other than pending and final.

netrome:
Still relevant, could potentially be a bit complex

netrome:
Will be resolved in https://github.com/near/mpc/pull/3187

PR #1678, opened 2025-12-16
test: fix sign sandbox tests

### LINKED ISSUE #1666: test: fix sandbox "sign"
### Background

We should fix this function to reduce test time:

https://github.com/near/mpc/blob/3137309f6652586865800cf08dcb9b6dde32511f/crates/contract/tests/sandbox/common.rs#L659-L688

Additionally, we should delete these lines of "test", because in prod, we expect one of the consecutive signature requests to time out. The only reason this test succeeds is because it it is not testing what it claims to be testing.
https://github.com/near/mpc/blob/0faab75ee131bcf81300721b9b0b1a3a046a1b91/crates/contract/tests/sandbox/sign.rs#L69-L100

### User Story

As a dev, I like fast and correct unit tests.

### Acceptance Criteria

- Instead of sleeping, use sandbox `fast_forward` to advance by one block, such that a request can be submitted.
Advance by 200 blocks if a timeout is desired.
- fix the test s.t. it tests actual production behavior

_No response_

kevindeforth:
resolves #1666

main improvements:
- _all except one_ test inside sign.rs no longer rely sleeps _(edit: almost, there is still one sleep we can't get rid of with the current sandbox capabilities, c.f. [slack](https://nearone.slack.com/archives/C0912BTG51T/p1765988408324409?thread_ts=1765973579.653919&cid=C0912BTG51T))_
- tests inside sing.rs are significantly faster, due to the use of "fast forward"

follow-ups:
1. https://github.com/near/mpc/pull/1683
2. https://github.com/near/mpc/issues/1684
3. a re-scoped https://github.com/near/mpc/issues/1306

### DIFF
--- a/crates/contract/tests/sandbox/sign.rs
+++ b/crates/contract/tests/sandbox/sign.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use crate::sandbox::common::{
     candidates, init, init_env, submit_signature_response, ContractSetup, PARTICIPANT_LEN,
 };
@@ -64,6 +66,8 @@ async fn test_contract_sign_request_all_schemes() -> anyhow::Result<()> {
                 .await
                 .unwrap();
             let status_2 = req.sign_ensure_included(&alice, &contract).await?;
+            // unfortunately, we still can't completely get rid of this sleep
+            tokio::time::sleep(Duration::from_secs(3)).await;
             worker
                 .fast_forward(NUM_BLOCKS_BETWEEN_REQUESTS)
                 .await


crates/contract/tests/sandbox/common.rs
```diff
@@ -761,35 +810,38 @@ pub async fn submit_ckd_response(
...
+        if start.elapsed() >= timeout {
+            anyhow::bail!("timed out waiting for request to appear in queue");
+        }
 
-    if let Some(response) = respond {
-        assert_eq!(
-            &returned_resp, &response.response,
-            "Returned signature request does not match"
-        );
+        tokio::time::sleep(Duration::from_millis(200)).await;
```
gilcu3: isn't this too big? Or this is supposed to take an amount of time in the order of seconds? My worry is that if a test uses a chain of many requests then waiting this long would make it much slower. The operation inside the loop does not seem to inefficient, so a lower sleep, maybe 50ms or less seems better IMO

crates/contract/tests/sandbox/sign.rs
```diff
@@ -14,17 +12,22 @@ use mpc_contract::{
 use near_workspaces::types::NearToken;
 use utilities::AccountIdExtV1;
 
-const ECDSA_SIGNATURE_SCHEMES: &[SignatureScheme] =
-    &[SignatureScheme::Secp256k1, SignatureScheme::V2Secp256k1];
+const NON_CKD_SCHEMES: &[SignatureScheme] = &[
+    SignatureScheme::Secp256k1,
+    SignatureScheme::V2Secp256k1,
+    SignatureScheme::Ed25519,
+];
```
gilcu3: I hope that at some point we don't need this distinction anymore. Initially I preferred to have CKD separately because the parameters were quite different, and it was not a signature, but they seem to be converging more and more. 

Still, I think we should keep the system open for the introduction of MPC protocols that are not signatures, so maybe it is fine to have at least 2 separate paths for that.

pbeza (approved):

crates/contract/tests/sandbox/sign.rs
```diff
@@ -40,46 +43,43 @@ async fn test_contract_sign_request() -> anyhow::Result<()> {
...
+                let req = key.create_sign_request(&predecessor_id.as_v2_account_id(), msg, path);
+                req.sign_and_validate(&alice, &contract, attested_account)
+                    .await
+                    .unwrap();
+            }
+        }
+
+        {
+            // check that in case of duplicate request, only the most recent will be signed:
+            let msg = "welp";
```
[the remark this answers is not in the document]
kevindeforth: yeah, it's for readability. We might want to split this into multiple tests and the indentation helps with that.

crates/contract/tests/sandbox/sign.rs
```diff
@@ -40,46 +43,43 @@ async fn test_contract_sign_request() -> anyhow::Result<()> {
 
     for key in &keys {
         for msg in messages {
+            {
+                println!("submitting: {msg}");
```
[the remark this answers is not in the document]
kevindeforth: I see these statements when nextest fails.
edit: also, this is existing code. I tend to agree with you, but would prefer not to do too much bike-shedding in this PR, given that we already have https://github.com/near/mpc/pull/1683 open.

crates/contract/tests/sandbox/sign.rs
```diff
@@ -40,46 +46,51 @@ async fn test_contract_sign_request() -> anyhow::Result<()> {
...
-                attested_account,
-            )
-            .await?;
+            let status_1 = req.sign_ensure_included(&alice, &contract).await?;
+            worker
+                .fast_forward(NUM_BLOCKS_BETWEEN_REQUESTS)
+                .await
+                .unwrap();
+            let status_2 = req.sign_ensure_included(&alice, &contract).await?;
+            // unfortunately, we still can't completely get rid of this sleep
```
netrome: :cry:

netrome (approved):
Quick skimming, thanks for doing this!

PR #1687, opened 2025-12-17
feat: implement echo connection check with ping/pong mechanism

### LINKED ISSUE #1647: Implement echo connection check
### Background

While we haven't concluded the investigation in https://github.com/near/mpc/issues/1206 because it is difficult to replicate this state, it is evident that implementing an echo/ping-pong check would remove the possibility of the network ending up in these states.

What is an echo check? It's a check where one node sends a message (just an incremental ID is sufficient) to another node, and the other node echos this message back to the first node. When the node receives the echo message, it knows the connection is up in both directions and can also measure the network latency/round-trip time as a bonus.

### User Story

As a node operator I'd like my node to be able to notice if loses it's connection with a peer.

### Acceptance Criteria

MPC nodes uses a ping-pong protocol to determine if their peers are up or not.

_No response_

pbeza:
Fixes #1647

### DIFF
--- a/crates/node/src/p2p.rs
+++ b/crates/node/src/p2p.rs
@@ -1494,12 +1494,30 @@ mod tests {
             assert_eq!(seq, expected_seq);
         }
 
+        /// Asserts that no ping has been sent yet.
+        fn expect_no_ping(&mut self) {
+            let pings = self.ping_receiver.as_mut().expect("pings was dropped");
+            match pings.try_recv() {
+                Err(mpsc::error::TryRecvError::Empty) => {
+                    // Expected: no ping in the channel
+                }
+                Err(mpsc::error::TryRecvError::Disconnected) => {
+                    panic!("ping channel was unexpectedly closed");
+                }
+                Ok(packet) => {
+                    panic!("expected no ping, but received: {:?}", packet);
+                }
+            }
+        }
+
         /// Sends a pong with the given sequence number.
         fn send_pong(&self, seq: u64) {
             self.pong_sender.send(PongInfo { seq }).unwrap();
         }
 
-        /// Returns true if the keepalive task detected a pong timeout.
+        /// Returns true if the keepalive task detected a pong timeout. In production, this
+        /// cancellation triggers reconnection via `run_persistent_connection`, which waits
+        /// on `wait_for_close()` and loops to establish a new connection.
         fn timed_out(&self) -> bool {
             self.closed.is_cancelled()
         }
@@ -1511,47 +1529,68 @@ mod tests {
         let mut env = KeepaliveTestEnv::new();
         let handle = env.spawn_keepalive();
 
-        env.advance(Duration::from_secs(1)).await;
         env.expect_ping(KeepaliveTestEnv::FIRST_PING_SEQ).await;
 
-        // when: advance to just before timeout - should NOT trigger
+        // when: advance to just before timeout
         env.advance(env.config.pong_timeout - Duration::from_nanos(1))
             .await;
         assert!(
             !env.timed_out(),
             "connection should not be cancelled before timeout"
         );
+        // no second ping should be sent while waiting for pong
+        env.expect_no_ping();
 
-        // when: advance past timeout - should trigger
+        // when: advance past timeout
         env.advance(Duration::from_nanos(2)).await;
-        let _ = handle.await;
-
-        // then
         assert!(
             env.timed_out(),
             "connection should be cancelled on pong timeout"
         );
+
+        let result = handle.await;
+        assert!(result.is_ok());
     }
 
     #[tokio::test(start_paused = true)]
     async fn test_pong_received_keeps_connection_alive() {
         start_root_task_with_periodic_dump(async move {
-            // given
+            // given: a keepalive task monitoring connection health
             let mut env = KeepaliveTestEnv::new();
             let _handle = env.spawn_keepalive_tracked();
 
-            // when: process multiple ping/pong cycles
+            // when: we respond to each ping with a matching pong
             for expected_seq in 1..=3u64 {
-                env.advance(Duration::from_secs(1)).await;
                 env.expect_ping(expected_seq).await;
+                assert!(
+                    !env.timed_out(),
+                    "connection should be alive while waiting for pong"
+                );
                 env.send_pong(expected_seq);
-                tokio::task::yield_now().await;
+                assert!(
+                    !env.timed_out(),
+                    "connection should remain alive after pong {expected_seq}"
+                );
             }
 
-            // then
+            // then: connection survives even near timeout boundary
+            env.expect_ping(4).await;
+            env.advance(env.config.pong_timeout - Duration::from_nanos(1))
+                .await;
+            assert!(!env.timed_out(), "should survive until just before timeout");
+            env.send_pong(4);
             assert!(
                 !env.timed_out(),
-                "connection should still be alive after successful pongs"
+                "pong at last moment should keep connection alive"
+            );
+
+            // then: connection times out when pongs stop
+            env.expect_ping(5).await;
+            env.advance(env.config.pong_timeout + Duration::from_nanos(1))
+                .await;
+            assert!(
+                env.timed_out(),
+                "connection should timeout when pongs stop arriving"
             );
         })
         .await;
@@ -1563,56 +1602,57 @@ mod tests {
         let mut env = KeepaliveTestEnv::new();
         let handle = env.spawn_keepalive();
 
-        env.advance(Duration::from_secs(1)).await;
         env.expect_ping(KeepaliveTestEnv::FIRST_PING_SEQ).await;
 
         // when: send pong with wrong sequence
         env.send_pong(KeepaliveTestEnv::INITIAL_SEQ);
-        tokio::task::yield_now().await;
+
+        // then: connection should not be cancelled yet, and no new ping sent
+        assert!(!env.timed_out(), "connection should not be cancelled yet");
+        env.expect_no_ping();
+
+        // when: advance to just before timeout
+        env.advance(env.config.pong_timeout - Duration::from_nanos(1))
+            .await;
 
         // then: connection should not be cancelled yet
         assert!(!env.timed_out(), "connection should not be cancelled yet");
+        env.expect_no_ping();
 
-        // when: advance past timeout
-        env.advance(env.config.pong_timeout + Duration::from_secs(1))
-            .await;
-        let _ = handle.await;
+        // when: advance to the timeout
+        env.advance(env.config.pong_timeout).await;
 
-        // then
+        // then: connection should be cancelled
         assert!(
             env.timed_out(),
             "connection should be cancelled after timeout"
         );
+
+        let result = handle.await;
+        assert!(result.is_ok());
     }
 
-    /// Verifies that when the ping channel is closed externally (e.g., connection teardown),
-    /// the keepalive task exits gracefully without triggering a timeout cancellation.
+    /// Verifies that when `ping_sender.send()` fails (receiver dropped), the keepalive task
+    /// exits gracefully without cancelling the `closed` token.
     ///
-    /// This distinguishes between two exit scenarios:
-    /// - **Channel closed**: Another component initiated shutdown (e.g., sender task dropped).
-    ///   The keepalive should simply exit - cancelling `closed` would be redundant since
-    ///   the connection is already being torn down elsewhere.
-    /// - **Pong timeout**: The keepalive detected a dead connection first. It must cancel
-    ///   `closed` to notify other tasks (like `wait_for_close`) that the connection died.
+    /// This tests the error path in `run_keepalive` where sending a ping fails because the
+    /// sender task has already dropped its end of the channel (e.g., due to TLS stream error).
+    /// In this case, the keepalive should simply return - the `closed` token will be cancelled
+    /// by the sender task's `DropToCancel` guard, so cancelling it here would be redundant.
     #[tokio::test(start_paused = true)]
     async fn test_ping_channel_closed_exits_gracefully() {
-        // given
+        // given: ping receiver is dropped before spawning keepalive
         let mut env = KeepaliveTestEnv::new();
         env.drop_ping_receiver();
 
-        // when
+        // when: keepalive task tries to send ping, it fails immediately
         let handle = env.spawn_keepalive();
-        env.advance(Duration::from_secs(1)).await;
-        let result = tokio::time::timeout(Duration::from_millis(100), handle).await;
+        handle.await.expect("task should exit cleanly");
 
-        // then
-        assert!(
-            result.is_ok(),
-            "keepalive should exit when ping channel is closed"
-        );
+        // then: closed token should NOT be cancelled
         assert!(
             !env.timed_out(),
-            "connection should not be cancelled on channel close"
+            "closed token should not be cancelled when ping channel is closed"
         );
     }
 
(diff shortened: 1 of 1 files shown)


crates/node/src/p2p.rs
```diff
@@ -393,8 +578,41 @@ pub async fn new_tls_mesh_network(
...
                     match packet {
-                        Packet::Ping => {
-                            // Do nothing. Pings are just for TCP keepalive.
+                        Packet::Ping(seq) => {
+                            // Send Pong via our outgoing connection to the peer
+                            if let Some(conn) = connections.get(&peer_id) {
+                                if let Some(outgoing_conn) =
+                                    conn.connectivity.any_outgoing_connection()
+                                {
+                                    let _ = outgoing_conn.sender.send(Packet::Pong(seq));
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/pull/1687/commits/bbdcc29c9e366c4958d89005a75840ea07d7b621

pbeza:
> (...) can you try disabling that one only and see if CI passes?

Unfortunately, commenting out that one test didn’t help. There seems to be something going on during resharing when some nodes disconnect and then reconnect. I’ll get back to this early next year (🎄), as I wasn’t able to even quickly monkey-patch it, unfortunately.

crates/node/src/p2p.rs
```diff
@@ -173,23 +297,70 @@ impl TlsConnection {
...
+                    seq += 1;
+                    let ping_sent_at = Instant::now();
+                    if sender_clone.send(Packet::Ping(seq)).is_err() {
                         // The receiver side will be dropped when the sender task is
                         // dropped (i.e. connection is closed).
                         break;
                     }
+
+                    // Wait for either a pong response or timeout
+                    tokio::select! {
```
DSharifi: You don't need to use a select! here if you want a timeout.

Checkout [`tokio::time::timeout`](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html) instead.

crates/node/src/p2p.rs
```diff
@@ -243,49 +414,53 @@ impl PersistentConnection {
...
+                        Ok(conn) => {
+                            let new_conn = Arc::new(conn);
                             tracing::info!(
                                 "Outgoing {} --> {} connected",
                                 my_id,
                                 target_participant_id
                             );
-                            new_conn
+
+                            // Register connection and drain any buffered Pongs
```
DSharifi: This comment is confusing. "drain" implies we'd discard the buffered pongs, but in the code below you are actually consuming the pongs to send them?

crates/node/src/p2p.rs
```diff
@@ -243,49 +414,53 @@ impl PersistentConnection {
...
+                                let _ = new_conn.sender.send(Packet::Pong(seq));
+                            }
+
+                            new_conn.wait_for_close().await;
                         }
                         Err(e) => {
-                            tracing::info!(
-                                "Could not connect to {}, retrying: {}, me {}",
+                            tracing::warn!(
+                                "Could not connect to {}: {:#}, retrying (me: {})",
```
DSharifi: Same comment as above regarding structured logging

pbeza: https://github.com/near/mpc/pull/1687/commits/873e1e0c30b5d311dbaf176bdce58b8ace8ba6dc ✅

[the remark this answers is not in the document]
pbeza: This comment is outdated — I’m now using structured logging.

crates/node/src/p2p.rs
```diff
@@ -406,8 +583,50 @@ pub async fn new_tls_mesh_network(
...
+                        Packet::Ping(seq) => {
+                            // Send Pong via our outgoing connection to the peer, maintaining
+                            // unidirectional I/O design. If outgoing isn't ready, buffer the Pong.
+                            if let Some(conn) = connections.get(&peer_id) {
+                                if let Some(outgoing_conn) =
+                                    conn.connectivity.any_outgoing_connection()
+                                {
+                                    if outgoing_conn.sender.send(Packet::Pong(seq)).is_err() {
+                                        tracing::info!(
+                                            "Outgoing connection to {} is dead, closing incoming connection for clean reconnect",
```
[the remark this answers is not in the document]
pbeza: I added a comment explaining this here: https://github.com/near/mpc/pull/1687/commits/b2c02feacd5c158d1f4e48a4f2fa0ef910dca4d2.

crates/node/src/p2p.rs
```diff
@@ -30,54 +26,170 @@ use tracing::info;
...
+    /// the `closed` token. This ensures pings are sent at exactly 1-second intervals while only
+    /// sending when the previous ping received a response.
     _keepalive_task: AutoAbortTask<()>,
-    /// This is cancelled when the connection is closed. Used to wait for the
-    /// connection to close.
+    /// Watch channel sender for receiving pong notifications from the incoming connection handler.
+    ///
+    /// The incoming handler sends PongInfo when it receives a Pong packet. The keepalive task
+    /// monitors this channel to detect connection health and calculate RTT. This enables clean
+    /// async communication without mutexes.
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/pull/1687/commits/7ec54a52111ec6897e4d110bc6afeeaf315350e6 ✅

pbeza:
Ready for re-review. After merging `origin/main`, I started hitting a failing `cargo deny` check due to https://rustsec.org/advisories/RUSTSEC-2026-0002.html.

I added a workaround in this commit: https://github.com/near/mpc/pull/1687/commits/0ebbc18002c450709e9013bbba95435ccae5ca8a. If you see a cleaner way to upgrade [`lru`](https://crates.io/crates/lru) (which seems to be the culprit here), let me know.

@DSharifi, feel free to take a look.

crates/node/src/p2p.rs
```diff
@@ -173,15 +303,61 @@ impl TlsConnection {
...
+                        }
+                        Err(_) => {
+                            tracing::warn!(
+                                peer = %target_participant_id,
+                                timeout = ?Self::PONG_TIMEOUT,
+                                "no pong received, closing connection"
+                            );
+                            closed_for_keepalive.cancel();
+                            return;
+                        }
```
kevindeforth: What is the reason for spawning this logic in the constructor, as opposed to in the `run_network_client` function, and the `run_receive_messages_loop` function, like we currently do for Indexer Heights and MPC tasks?
https://github.com/near/mpc/blob/31060c32e922e5232648b3278c9fd12c31d219e7/crates/node/src/network.rs#L383

pbeza: The `keepalive_task` was already responsible for sending pings, so extending it to handle pongs as well felt like a natural choice, IMO.

Also, I’m not sure we need to be that careful when refactoring this particular piece of code — it’s already quite complex and does a lot of work in the constructor. AFAIK, a full reimplementation of the networking layer is likely to be one of this year’s priorities (though that’s still to be decided by @netrome).

crates/node/src/p2p.rs
```diff
@@ -30,54 +26,170 @@ use tracing::info;
...
+    /// sending when the previous ping received a response.
     _keepalive_task: AutoAbortTask<()>,
-    /// This is cancelled when the connection is closed. Used to wait for the
-    /// connection to close.
+    /// Watch channel sender for delivering pong notifications to the keepalive task.
+    ///
+    /// The incoming handler sends PongInfo when it receives a Pong packet. The keepalive task
+    /// monitors this channel to detect connection health and calculate RTT. This enables clean
+    /// async communication without mutexes.
+    pong_tx: tokio::sync::watch::Sender<PongInfo>,
```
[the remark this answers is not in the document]
pbeza: Same comment as [here](https://github.com/near/mpc/pull/1687#discussion_r2676064102).

crates/node/src/p2p.rs
```diff
@@ -243,49 +420,66 @@ impl PersistentConnection {
...
+                                    peer = %target_participant_id,
+                                    "failed to send pending pong, reconnecting"
+                                );
+                                continue;
+                            }
+
+                            new_conn.wait_for_close().await;
+                            tracing::info!(
+                                peer = %target_participant_id,
+                                "connection closed, reconnecting"
```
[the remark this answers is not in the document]
pbeza: No strong reason — it was kind of an arbitrary decision to put it where it seemed like it could work.

Honestly, just wrapping my head around the current implementation and getting something workable was hard enough, so I didn’t really try to redesign it properly — especially since the existing design is already pretty messy.

If you feel strongly that this should live somewhere else, I’m happy to move it — just please elaborate a bit on how you’re imagining it should be implemented.

crates/node/src/p2p.rs
pbeza: Got it, I see your point — I’m just not convinced a small refactor of already messy code will help much, especially if we’re going to reimplement it soon anyway. Let’s sync with @netrome and @DSharifi on Monday.

crates/node/src/p2p.rs
```diff
@@ -30,54 +26,170 @@ use tracing::info;
...
+        handshake::p2p_handshake,
+        MeshNetworkTransportReceiver, MeshNetworkTransportSender,
+    },
+    primitives::{
+        IndexerHeightMessage, MpcMessage, MpcPeerMessage, ParticipantId, PeerIndexerHeightMessage,
+        PeerMessage,
+    },
+    tracking::{self, AutoAbortTask, AutoAbortTaskCollection},
+};
+
```
netrome: Was there any thought behind putting these imports after the `TCP_NODELAY` constant? Otherwise I'd like to preserve the normal "imports > constants > types/functions > tests" order in this file.

netrome (commented):
Partial review so far. Just a nit about import ordering and chiming in on the keepalive task refactor conversation.

Will continue reviewing and understanding this better ~~after lunch~~ tonight.

netrome:
> Will continue reviewing and understanding this better ~~after lunch~~ tonight.

Had to prioritize other work for the evening. Will circle back to this tomorrow instead.

crates/node/src/p2p.rs
```diff
@@ -173,15 +303,61 @@ impl TlsConnection {
...
+                    let ping_sent_at = Instant::now();
+                    if sender_clone.send(Packet::Ping(seq)).is_err() {
                         // The receiver side will be dropped when the sender task is
                         // dropped (i.e. connection is closed).
-                        break;
+                        return;
+                    }
+                    crate::metrics::MPC_P2P_PING_SEQUENCE_SENT
+                        .with_label_values(&[&peer_id_str])
+                        .set(seq as i64);
```
netrome: Tiny blocker: We should not use `as` conversions, especially this one can wrap for large values. While this isn't likely to happen I'd like to keep our code base clean and make a habit of not using these conversions. I think it's safe to return an error and close the task if we encounter a value larger than i64::MAX here.

```suggestion
                    let Ok(seq_i64) = i64::try_from(seq) else {
                        tracing::warn!(seq, "TLS sequence number exceeded i64::MAX - closing connection");
                        return;
                    };
                    crate::metrics::MPC_P2P_PING_SEQUENCE_SENT
                        .with_label_values(&[&peer_id_str])
                        .set(seq_i64);
```

deny.toml
```diff
@@ -32,7 +32,14 @@ ignore = [
...
 [bans]
 multiple-versions = "deny"
-skip = [{ name = "prost-types" }]
+skip = [
+  { name = "prost-types" },
+  # Allow multiple versions pulled transitively by near-sdk/nearcore
+  { name = "lru" },
+  { name = "near-fmt" },
+  { name = "near-primitives" },
+  { name = "near-time" },
```
netrome: Why do we need this here now? This PR doesn't change any dependencies.

netrome (changes requested):
Thanks for doing this, I know this part of the node is quite complex to dive into and I don't fully understand it myself.

However, I think this should be possible to test so I'd like to see at least some basic tests covering the changed behavior. This is the only big blocker I see.

Beyond this there's a few tiny blockers as well:

1. Unconventional import order (from previous review)
2. A silent lossy as conversion.
3. Implicit overflowing arithmitic.

crates/node/src/p2p.rs
```diff
@@ -173,15 +303,61 @@ impl TlsConnection {
...
+                let peer_id_str = target_participant_id.to_string();
+                let mut seq: u64 = 0;
+                let mut interval = tokio::time::interval(Self::PING_INTERVAL);
+                interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+
                 loop {
-                    tokio::time::sleep(std::time::Duration::from_secs(1)).await;
-                    if sender_clone.send(Packet::Ping).is_err() {
+                    interval.tick().await;
+                    seq += 1;
```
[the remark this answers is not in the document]
pbeza: You’re quite optimistic about the MPC network uptime. 😄 `u64::MAX = 18_446_744_073_709_551_615` (seconds). Converted to years, that’s **~584 billion years**. Anyway, I fixed it here to be on the safe side: https://github.com/near/mpc/pull/1687/commits/eed1da866a606b070a272a81337e357fd1cc076d. ✅

crates/node/src/p2p.rs
pbeza: https://github.com/near/mpc/pull/1687/commits/0164312676c69491afdde90daa7a35d728d98e6f ✅

pbeza:
Feel free to do an early re-review, @kevindeforth @netrome @DSharifi. There’s still a bit of a mystery around one test I’d like to dig into tomorrow—specifically this one (it passes even if you replace `+` with `-` [here](https://github.com/near/mpc/blob/0fe1bcbeb11609e9852f670381050bbf12934d77/crates/node/src/p2p.rs#L1413)):
https://github.com/near/mpc/blob/0fe1bcbeb11609e9852f670381050bbf12934d77/crates/node/src/p2p.rs#L1403-L1422

Overall, the code shouldn’t change much unless you ask for more changes.

I did refactor it a bit to make it easier to test and tried to follow Kevin’s suggestions. This part of the code is pretty hard to reason about, so I might’ve missed something — would appreciate an extra careful review 🙏

netrome:
> I did refactor it a bit to make it easier to test and tried to follow Kevin’s suggestions. This part of the code is pretty hard to reason about, so I might’ve missed something — would appreciate an extra careful review 🙏

Sounds good. I'll do a quick pass of the PR now and look a bit deeper again after lunch.

crates/node/src/p2p.rs
```diff
@@ -381,78 +694,121 @@ pub async fn new_tls_mesh_network(
...
-                            let mut buf = vec![0; len as usize];
-                            tokio::time::timeout(
-                                std::time::Duration::from_secs(MESSAGE_READ_TIMEOUT_SECS),
-                                stream.read_exact(&mut buf),
-                            )
-                            .await??;
-                            received_bytes += 4 + len as u64;
-
-                            let packet = Packet::try_from_slice(&buf)
-                                .context("Failed to deserialize packet")?;
```
netrome: Hmm, the fact that we return here on errors sounds like we'll crash if old nodes encounter Pong messages. If that's the case, we'll need to deploy this in two steps. First, make a small PR to make the packet deserialization robust towards unfamiliar packets which we have to release first, and then rebase this PR on top of that and make this in a subsequent release.

netrome: Hmm... reflecting more on this, simply adding tolerance for unrecognized packets won't solve this as the ping-pong mechanism would still kill the connection with any node who doesn't recognize it. Instead, we have to break up this change into two steps. Step 1 is to create a version of the node that recognizes these packets, returns Pong messages if received but still sends the old Ping (without any value) messages.

Once that's released we can follow up with a PR that sends the new ping messages and drops the connection if the pong messages aren't received.

@pbeza - what do you think of this?

PR #1704, opened 2025-12-19
feat: example ckd contract implementation

### LINKED ISSUE #1605: TEE example application - smart contract
### Background

As part of the CKD project, we need to build an example TEE application, as explained in the [CKD docs](https://github.com/near/threshold-signatures/blob/main/docs/confidential_key_derivation/confidential_key_derivation.md). To begin with, we should have a crate in the workspace implementing the smart-contract part.

### User Story

With an example application users will have a much easier path to using the feature

### Acceptance Criteria

We’ve implemented a simple smart contract for the TEE based application that uses the CKD system that can act as a reference for future implementations. The smart-contract needs to validate attestations using our attestation crate.

_No response_

gilcu3:
Closes #1605 

This will probably stay as a draft until I am back in January, as there are a few scope decisions to be made.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1672,6 +1672,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "ckd-example-contract"
+version = "3.2.0"
+dependencies = [
+ "attestation",
+ "contract-interface",
+ "include-measurements",
+ "near-sdk",
+ "serde_json",
+]
+
+[[package]]
 name = "clang-sys"
 version = "1.8.1"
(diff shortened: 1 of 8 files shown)


gilcu3:
> This will probably stay as a draft until I am back in January, as there are a few scope decisions to be made.

@netrome about this, how far do we want to go with this? To have a full example I would need to also add some TEE app, which would probably include a launcher and a lot of the infra that we already have for the MPC node. I feel this might be an overkill. WDYT? The issue with this PR is that for attestation to minimally work, we need some measurements, which depend on the TEE app to be attested.

Issue #1716, opened 2026-01-03
Remove `WebUIConfig` custom type in favor of `std::net::SocketAddr`


DSharifi:
### Background

The type `WebUIConfig` is effectively encoding a `SocketAddr` but has not validated that the host string is valid. For all use cases we want to use `WebUIConfig`, we could use `SocketAddr` instead.

https://github.com/near/mpc/blob/c900991863654298e0075366e35db41132d484ea/crates/node/src/config.rs#L96-L98

### Acceptance Criteria

Remove the `WebUIConfig` and replace it with `SocketAddr`

PR #1717, opened 2026-01-03
chore: use `SocketAddr` instead of custom struct for addresses in configs

DSharifi:
closes https://github.com/near/mpc/issues/1716


kevindeforth (approved):
Thanks!
Before merging, can you confirm that devnet still works?

docs/localnet/localnet.md
```diff
@@ -193,12 +193,8 @@ cat > ~/.near/mpc-frodo/config.yaml << 'EOF'
 my_near_account_id: frodo.test.near
 near_responder_account_id: frodo.test.near
 number_of_responder_keys: 1
-web_ui:
-  host: localhost
-  port: 8081
-migration_web_ui:
-  host: localhost
-  port: 8079
```
netrome: This broke backwards compatibility, we need to revert or fix.

PR #1723, opened 2026-01-06
feat(node): add web endpoint to collect CPU profiles with `pprof-rs`

### LINKED ISSUE #1630: Add support for pprof/flamegraph to the MPC binary
### Background

To help the investigation in https://github.com/near/mpc/issues/1175 we'd like to enable us to debug and inspect where we are spending CPU time.

### User Story

As a developer I'd like flamegraphs for being able to debug and inspect where we are spending CPU time so that I can investigate our performance bottlenecks and degradations.

### Acceptance Criteria

We have a way to generate flamegraphs to profile the MPC node.

_No response_

DSharifi:
closes #1630

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3527,6 +3527,8 @@ version = "0.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5419bdc4f6a9207fbeba6d11b604d481addf78ecd10c11ad51e76c2f6482748d"
 dependencies = [
+ "allocator-api2",
+ "equivalent",
  "foldhash 0.2.0",
  "serde",
 ]
@@ -4512,6 +4514,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "lru"
+version = "0.16.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a1dc47f592c06f33f8e3aea9591776ec7c9f9e4124778ff8a3c3b87159f7e593"
+dependencies = [
+ "hashbrown 0.16.0",
+]
+
+[[package]]
 name = "lru-slab"
 version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


crates/node/src/profiler/web_server.rs
```diff
@@ -0,0 +1,56 @@
...
+
+            (
+                StatusCode::OK,
+                [(header::CONTENT_TYPE, CONTENT_TYPE_SVG)],
+                svg_buffer,
+            )
+                .into_response()
+        }
+        Err(e) => (StatusCode::INTERNAL_SERVER_ERROR, format!("Error: {:?}", e)).into_response(),
+    }
```
pbeza: Let's not use `unwrap()` if possible.

```suggestion
    let flamegraph_svg = pprof_report.and_then(|report| {
        let mut svg_buffer = Vec::new();
        report.flamegraph(&mut svg_buffer)?;
        Ok(svg_buffer)
    });

    match flamegraph_svg {
        Ok(svg_buffer) => (
            StatusCode::OK,
            [(header::CONTENT_TYPE, CONTENT_TYPE_SVG)],
            svg_buffer,
        )
            .into_response(),
        Err(e) => (StatusCode::INTERNAL_SERVER_ERROR, format!("Error: {e:?}")).into_response(),
    }
```

crates/node/src/profiler/pprof.rs
```diff
@@ -0,0 +1,150 @@
...
+        "libsystem_c",
+        "CoreFoundation",
+        "Foundation"
+    ]
+};
+
+/// Collects a pprof profile for the specified duration and frequency (Hz).
+pub(super) async fn collect_pprof(
+    duration: Duration,
+    frequency_hz: i32,
```
pbeza: I think it’s an unsigned `int` under the hood, so we should use `uint` (`u32`?).

deployment/start.sh
```diff
@@ -92,6 +92,7 @@ near_responder_account_id: $responder_id
 number_of_responder_keys: 50
 web_ui: 0.0.0.0:8080
 migration_web_ui: 0.0.0.0:8079
+pprof_bind_address: 0.0.0.0:34001
```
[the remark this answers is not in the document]
DSharifi: It must be bound to 0.0.0.0 since it's running a docker container, if we want to reach it from the docker host

pbeza (dismissed):

Issue #1590, opened 2025-12-01
Support running MPC nodes with the launcher in non-TDX environments


netrome:
### Background

We've rolled out the 3.0 release to both mainnet and testnet now, making the network support running nodes on TDX servers. However, on these servers we're using a very different approach to configure and run the MPC nodes, using docker compose and our dedicated launcher process.

To reduce the surface of potential TEE-only bugs we should use the same docker compose + launcher setup to manage our non-TEE deployments. This should only require small modifications to our launcher code, but would bring several benefits:

* We'd be able to spot most launcher bugs outside of TEEs.
* We can upgrade non-TEE nodes and TEE nodes by just voting for the appropriate code hash.
* It would be easier for developers to mimic the production setup, making it easier to test upgrades and proactively catch bugs.

### User Story

As a node operator I'd like to be able to run the same setup outside of TEEs as I run inside of TEEs.

### Acceptance Criteria

The launcher can be used outside of TEEs.

PR #1735, opened 2026-01-11
feat(launcher): Add ability to use the launcher also for non tee setups.

barakeinav1:
fixes [#1590](https://github.com/near/mpc/issues/1590)

This PR allows the MPC launcher to run both inside TEEs and outside TEEs using the same orchestration flow, without weakening any existing TEE security guarantees.

A new measured PLATFORM mode (TEE / NONTEE) explicitly controls whether dstack/attestation is required. TEE behavior remains fail-closed and fully attested, while non-TEE deployments can exercise the launcher logic without dstack.

High-level changes

- Introduce measured PLATFORM={TEE|NONTEE} environment variable
- Enforce strict, fail-closed behavior in PLATFORM=TEE
- Skip dstack / RTMR extension in PLATFORM=NONTEE
- Gate dstack socket mount and env vars on platform
- Explicitly signal MPC node via MPC_TEE_MODE

tests:
added a few python tests.
I also run a manually test an started an MPC node using the launcher

### DIFF
--- a/crates/node/src/coordinator.rs
+++ b/crates/node/src/coordinator.rs
@@ -425,7 +425,7 @@ where
             let network_client = network_client.clone();
             let mpc_config = mpc_config.clone();
 
-            tracking::spawn_checked("key_resharing", async move {
+            tracking::spawn_checked("key resharing", async move {
                 Self::run_key_resharing(
                     &config_file,
                     keyshare_storage.clone(),
--- a/crates/node/src/key_events.rs
+++ b/crates/node/src/key_events.rs
@@ -511,7 +511,7 @@ pub async fn keygen_follower(
         };
 
         tasks.spawn_checked(
-            &format!("Key generation follower for {:?}", key_event_id),
+            &format!("key generation follower for {:?}", key_event_id),
             keygen_computation(
                 key_event_receiver.clone(),
                 channel,
@@ -641,7 +641,7 @@ pub async fn resharing_follower(
         };
 
         tasks.spawn_checked(
-            &format!("Key resharing follower for {:?}", key_event_id),
+            &format!("key resharing follower for {:?}", key_event_id),
             resharing_computation(
                 key_event_receiver.clone(),
                 channel,
--- a/crates/node/src/mpc_client.rs
+++ b/crates/node/src/mpc_client.rs
@@ -511,7 +511,7 @@ impl MpcClient {
         while let Some(channel) = channel_receiver.recv().await {
             let mpc_clone = mpc_client.clone();
             tasks.spawn_checked(
-                &format!("passive task {:?}", channel.task_id()),
+                &format!("passive task; task_id: {:?}", channel.task_id()),
                 async move { mpc_clone.process_channel_task(channel).await },
             );
         }
--- a/crates/node/src/p2p.rs
+++ b/crates/node/src/p2p.rs
@@ -366,75 +366,80 @@ pub async fn new_tls_mesh_network(
     let connectivities_clone = connectivities.clone();
     let my_id = config.my_participant_id;
     info!("Spawning incoming connections handler.");
-    let incoming_connections_task = tracking::spawn("Handle incoming connections", async move {
-        let mut tasks = AutoAbortTaskCollection::new();
-        while let Ok((tcp_stream, _)) = tcp_listener.accept().await {
-            let message_sender = message_sender.clone();
-            let participant_identities = participant_identities.clone();
-            let tls_acceptor = tls_acceptor.clone();
-            let connectivities = connectivities_clone.clone();
-            tasks.spawn_checked::<_, ()>("Handle connection", async move {
-                let tcp_stream = configure_tcp_stream(tcp_stream)?;
-                let mut stream = tls_acceptor.accept(tcp_stream).await?;
+    let incoming_connections_task =
+        tracking::spawn("incoming connection handler tls mesh network", async move {
+            let mut tasks = AutoAbortTaskCollection::new();
+            while let Ok((tcp_stream, _)) = tcp_listener.accept().await {
+                let message_sender = message_sender.clone();
+                let participant_identities = participant_identities.clone();
+                let tls_acceptor = tls_acceptor.clone();
+                let connectivities = connectivities_clone.clone();
+                tasks.spawn_checked::<_, ()>(
+                    "new connection handler tls mesh network",
+                    async move {
+                        let tcp_stream = configure_tcp_stream(tcp_stream)?;
+                        let mut stream = tls_acceptor.accept(tcp_stream).await?;
 
-                let peer_id = verify_peer_identity(stream.get_ref().1, &participant_identities)?;
-                tracking::set_progress(&format!("Authenticated as {}", peer_id));
-                p2p_handshake(&mut stream, TlsConnection::HANDSHAKE_TIMEOUT)
-                    .await
-                    .context("p2p handshake")?;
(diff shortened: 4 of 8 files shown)


tee_launcher/test_launcher_config.py
```diff
@@ -496,3 +499,271 @@ def test_validate_image_hash():
...
+    monkeypatch.setenv(launcher.ENV_VAR_DEFAULT_IMAGE_DIGEST, default_digest)
+
+    # Provide a temp user config file so main() passes env into build_docker_cmd()
+    user_config = tmp_path / "user_config"
+    user_config.write_text(
+        "\n".join(
+            [
+                "MPC_ACCOUNT_ID=mpc-user-123",
+                "PORTS=11780:11780,--env BAD=1",  # BAD should be ignored
+                "EXTRA_HOSTS=node:192.168.1.1,--volume /:/mnt",  # injection should be ignored
```
DSharifi: Can we move these magic values into constants?

DSharifi (dismissed):

tee_launcher/launcher_docker_compose_nontee.yaml
```diff
@@ -0,0 +1,30 @@
...
+    container_name: launcher
+
+    environment:
+      - PLATFORM=NONTEE
+      - DOCKER_CONTENT_TRUST=1
+      - DEFAULT_IMAGE_DIGEST=sha256:00006c1059cc0219005b21956a4df8238b0cc33ad559a578a63169de4e28c81e
+
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+      - ./user-config.conf:/tapp/user_config:ro
```
gilcu3: what's the reason this line is not in the TEE yaml but exists here?

tee_launcher/launcher_docker_compose_nontee.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+    read_only: true
+    tmpfs:
+      - /tmp
+
+volumes:
+  shared-volume:
+    name: shared-volume
+  mpc-data:
+    name: mpc-data
```
gilcu3: why do we need this? Can't we also have the mpc-data as an internal volume to the docker image? That would make it easier to lose it, but at the same time would make it much closer to the actual TEE deployment

tee_launcher/test_launcher_config.py
```diff
@@ -496,3 +499,271 @@ def test_validate_image_hash():
...
+    assert "BAD=1" not in cmd_str
+    assert "/:/mnt" not in cmd_str
+
+    # Required mounts / flags from build_docker_cmd
+    assert "--security-opt" in cmd_str
+    assert "no-new-privileges:true" in cmd_str
+    assert "/tapp:/tapp:ro" in cmd_str
+    assert "shared-volume:/mnt/shared" in cmd_str
+    assert "mpc-data:/data" in cmd_str
+    assert f"--name {launcher.MPC_CONTAINER_NAME}" in cmd_str
```
gilcu3: given that the command in this test will not change, we could actually match the whole `cmd_str` instead (or in addition). That could have caught the bug I found above.

tee_launcher/using_the_launcher_in_nontee_setup.md
gilcu3: Thank you very much for this doc!

tee_launcher/launcher.py
```diff
@@ -685,15 +739,22 @@ def get_bare_digest(full_digest: str) -> str:
...
 
     docker_cmd = ["docker", "run"]
 
     # Required environment variables
     docker_cmd += ["--env", f"MPC_IMAGE_HASH={bare_digest}"]
     docker_cmd += ["--env", f"MPC_LATEST_ALLOWED_HASH_FILE={IMAGE_DIGEST_FILE}"]
-    docker_cmd += ["--env", f"DSTACK_ENDPOINT={DSTACK_UNIX_SOCKET}"]
+
+    if platform == PLATFORM_TEE:
+        docker_cmd += ["--env", "MPC_TEE_MODE=1"]
```
gilcu3: do we plan to use the `MPC_TEE_MODE` env variable in the node? Is this to future proof something?

tee_launcher/launcher.py
```diff
@@ -748,6 +807,10 @@ def build_docker_cmd(user_env: dict[str, str], image_digest: str) -> list[str]:
         image_digest,  # IMPORTANT: Docker must get the FULL digest
     ]
 
+    # Mount dstack socket ONLY in TEE mode.
+    if platform == PLATFORM_TEE:
+        docker_cmd += ["-v", f"{DSTACK_UNIX_SOCKET}:{DSTACK_UNIX_SOCKET}"]
+
```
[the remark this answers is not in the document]
barakeinav1: BTW, I have tested only the nontee flow.
I'll test the TEE as well after the review (since it is harder to test)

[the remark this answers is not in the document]
gilcu3: >there is no bug here.
github shows this as removing the third -v, but it could have showed this as the second -v instead. (it doesn't matter).

alright will check again, sad that we cannot trust github to show the proper diff

>BTW, I have tested only the nontee flow.
I'll test the TEE as well after the review (since it is harder to test)

indeed. using some `monkeypatch` magic, shouldn't we be able to test the TEE part, at least to some extent?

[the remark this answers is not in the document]
gilcu3: just checked, yeah my bad it is not a bug :) will modify the description

tee_launcher/launcher_docker_compose_nontee.yaml
barakeinav1: In TEE deployments, /tapp/user_config is provided by dstack inside the CVM (it copies the operator’s config into the guest filesystem at that path), so we don’t need a compose bind-mount for it.

In non-TEE deployments there is no dstack guest agent to populate /tapp/user_config, so we bind-mount ./user-config.conf to /tapp/user_config:ro to provide the same configuration interface to the launcher.

gilcu3: thanks, yeah it makes total sense then to make the non-tee env as close as possible to the tee one

tee_launcher/test_launcher_config.py
```diff
@@ -124,7 +127,7 @@ def test_build_docker_cmd_sanitizes_ports_and_hosts():
         "EXTRA_HOSTS": "node:192.168.1.1,--volume /:/mnt",
         "MPC_ACCOUNT_ID": "mpc-user-123",
     }
-    cmd = build_docker_cmd(env, "sha256:abc123")
+    cmd = build_docker_cmd(TEE, env, "sha256:abc123")
```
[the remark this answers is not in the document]
barakeinav1: ok

tee_launcher/launcher.py
```diff
@@ -36,6 +37,13 @@ class RpcTimingConfig(NamedTuple):
     rpc_max_attempts: int
 
 
+#
+# Platform mode (MUST come from measured docker-compose env in TEE deployments)
+#
+ENV_VAR_PLATFORM = "PLATFORM"  # expected values: "TEE" or "NONTEE"
+PLATFORM_TEE = "TEE"
+PLATFORM_NONTEE = "NONTEE"
+
```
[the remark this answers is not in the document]
barakeinav1: ok

DSharifi (approved):

gilcu3 (approved):

PR #1738, opened 2026-01-11
build(rust): add support for Nix build environment

### LINKED ISSUE #1753: Create dev environment for rust
### Background

child task of #1737 

### User Story

_No response_

### Acceptance Criteria

`nix develop` has all the tooling needed to build our rust codebase

_No response_

DSharifi:
closes #1753

Edit:
Note to reviewers, the python environment is explicitly disabled in this PR, as I want to follow up with a proper setup that has the full pytest environment included.


nix/python-env.nix
```diff
@@ -0,0 +1,217 @@
...
+
+        doCheck = false;
+      };
+
+      # NB! downgrade to python 3.12 if blspy is needed
+      #
+      # blspy = pyfinal.buildPythonPackage rec {
+      #   pname = "blspy";
+      #   version = "2.0.2";
+      #   pyproject = true;
```
gilcu3: I understand these comments will make sense when the pytest part is enabled? Could we remove them in the meantime?

nix/python-env.nix
```diff
@@ -0,0 +1,217 @@
...
+    scikit-learn
+    scipy
+    semver
+    toml
+    tqdm
+    urllib3
+    jmespath
+    borsh-construct
+    nearup
+    # blspy
```
gilcu3: what was the problem with that package? is this nix file enforcing python3.11 which is the one we need and use in CI?

flake.nix
```diff
@@ -0,0 +1,211 @@
...
+      self,
+      nixpkgs,
+      rust-overlay,
+    }:
+    let
+      lib = nixpkgs.lib;
+
+      systems = [
+        "x86_64-linux"
+        "aarch64-linux"
```
gilcu3: do we have anyone in this arch?

gilcu3 (approved):
Thank you, tested and worked on my env. Left some questions

EDIT: cargo-shear failed because probably you are using a newer version. Please sync versions with the ones used in CI, we can update them later in sync

README.md
```diff
@@ -7,6 +7,21 @@ This repository contains the code for the Near MPC node. It is a rewrite of [Nea
 - **Nearcore Node**: This repository depends on the nearcore node, included as a submodule in the `/libs` directory.
 - **Other Dependencies**: All other dependencies are handled by Cargo.
 
+## Development Environment (Nix)
+
+This repository includes a [flake.nix](flake.nix) file that provides a reproducible development environment using Nix. This environment pre-configures the Rust toolchain, LLVM/Clang tools, NEAR CLI tools, and all necessary system dependencies.
+Prerequisites
+  - **Nix**: Install Nix with [Flakes enabled](https://nixos.wiki/wiki/Flakes)
```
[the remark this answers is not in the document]
gilcu3: nix recommends the script installer, and arch recommends their package
```
extra/nix 2.33.0-1 [installed]
```
and I prefer pacman to handle stuff, so I use the distro package, which requires the nix-daemon running.

flake.nix
```diff
@@ -0,0 +1,200 @@
...
+          nearTools = with pkgs; [
+            python3Packages.keyring
+            near-cli-rs
+            cargo-near
+          ];
+
+          miscTools = with pkgs; [
+            git
+            zizmor
+            binaryen
```
gilcu3: we need to pin versions of these tools to follow ci, I tried cargo make and it failed because of that. new zizmor versions are less permissive. Same happens with these tools:

```
cargo-binstall
cargo-deny
cargo-insta
cargo-make
cargo-nextest
cargo-shear
cargo-sort
```

pbeza: > I'm adding it now. But it would be the wrong way to go about it to keep enforcing the 1;1 mapping.

Why’s that? 🤔

> If we want to maintain that, then CI might as well use nix

Why not? 🤔

PR #1761, opened 2026-01-14
chore: remove dead Python code

pbeza:
Fixes https://github.com/near/mpc/issues/1762

Inspired by https://github.com/near/mpc/pull/1725#discussion_r2687318118.


netrome (approved):
Nice, thanks for spotting and removing the dead code! :pray:

PR #1769, opened 2026-01-15
test: handle transaction nonces locally

### LINKED ISSUE #1606: flaky pytests since nearcore 2.10
### Background

Recently we have had a few pytests fail with:

```

AssertionError: {
 "jsonrpc": "2.0",
 "error": {
  "name": "HANDLER_ERROR",
  "cause": {
   "info": {},
   "name": "INVALID_TRANSACTION"
  },
  "code": -32000,
  "message": "Server error",
  "data": {
   "TxExecutionError": {
    "InvalidTxError": {
     "InvalidNonce": {
      "ak_nonce": 30,
      "tx_nonce": 30
     }
    }
   }
  }
 },
 "id": "dontcare"
}
```

Examples:
- https://github.com/near/mpc/actions/runs/19893409371/job/57017619575?pr=1603
- https://github.com/near/mpc/actions/runs/19927453444/job/57130706678
- https://github.com/near/mpc/actions/runs/20141149427/job/57808757473
- https://github.com/near/mpc/actions/runs/20163080061/job/57879868229?pr=1645
- https://github.com/near/mpc/actions/runs/20177192755/job/57928213036?pr=1645
- https://github.com/near/mpc/actions/runs/20276497537/job/58226661136?pr=1676

My best bet is that it is related to the underlying changes in pytests that happened in that release.

### User Story

_No response_

### Acceptance Criteria

The source of the flakiness is found and fixed.

_No response_

gilcu3:
Closes #1606 (again)

### DIFF
--- a/pytest/common_lib/shared/near_account.py
+++ b/pytest/common_lib/shared/near_account.py
@@ -50,7 +50,7 @@ class NearAccount:
             )
             if nonce is None:
                 nonce = 0
-            nonces += [Nonce(nonce)]
+            nonces.append(Nonce(nonce))
         self._pytest_signer_nonces = nonces
 
     def account_id(self) -> str:


pytest/common_lib/shared/__init__.py
gilcu3: this is the only remaining place where we do not manage the nonce. The reason is a bit complicated to explain, but basically is related to the dance we do here to create accounts and access_keys. Given that we are anyway moving away from pytests, I did not want to spend too much time refactoring that. The bug we were hitting should not re-appear here, so it should be fine.

pytest/common_lib/shared/near_account.py
```diff
@@ -37,6 +37,26 @@ def __init__(
...
+                    nonce := running_node.get_nonce_for_pk(
+                        acc=key.account_id,
+                        pk=key.pk,
+                    )
+                )
+                is not None
+                else 0
+            )
+            for key in self._pytest_signer_keys
+        ]
```
DSharifi: As much as I like list comprehensions, I think a simple for loop is much easier to undestand here when it spans 13 lines plus a subtle walrus operator `:=`.

DSharifi (changes requested):
Thanks for cleaning this up.

Blocker:
- Can you just change the complex list comprehension to a simple for loop?

pytest/common_lib/shared/near_account.py
```diff
@@ -126,7 +145,30 @@ def sign_tx(
...
+            last_block_hash,
+            key.account_id,
+            key.decoded_pk(),
+            key.decoded_sk(),
+        )
+        return signed_tx
+
+
+class Nonce:
+    def __init__(self, nonce=0):
```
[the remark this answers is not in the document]
gilcu3: true, it was used in some stage, will remove

PR #1775, opened 2026-01-15
fix(network): add small timeout between establishing new persistent connections

### LINKED ISSUE #1829: investigation: check if adding timeout in Persistent connection still breaks after improving handshake
### Background

check if [this PR](https://github.com/near/mpc/pull/1775) still breaks CI after resolving #1783

### User Story

as a dev, I like non-ddos code

### Acceptance Criteria

merge or close #1775

_No response_

kevindeforth:
resolves https://github.com/near/mpc/issues/1829

----
before #1831:
A timeout of 100 ms makes many tests fail.
A timeout of 10 ms seems to pass at least locally.
That's slightly concerning, as we would like a timeout, so we don't accidentally DDOS the receiving node.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4904,6 +4904,7 @@ dependencies = [
  "near-sdk",
  "near-time 2.10.5",
  "node-types",
+ "num_enum",
  "pprof",
  "prometheus",
  "rand 0.8.5",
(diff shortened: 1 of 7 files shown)


kevindeforth:
closing this, it seems to be a persistent issue.

PR #1778, opened 2026-01-16
chore: nodes accept peer with same or higher protocol version

### LINKED ISSUE #1777: prepare network prototocol for updates
### Background

We currently use a u32 as the network protocol version number.
The nodes currently require that their peer is running the exact same protocol.
This is too restrictive and makes updating the protocol network (without breaking changes) impossible.

In a first PR, nodes should accept all peers that are running the same or a higher version of the network protocol.

### User Story

As a dev, I want to update the network protocol without breaking the network.

### Acceptance Criteria

Nodes accept peers that are running the same or newer protocol versions than themselves. 

_No response_

kevindeforth:
Resolves #1777, in preparation of further network updates.

----
for context: https://github.com/near/mpc/pull/1778#discussion_r2697598183
> For future reference, we plan to keep track of the agreed protocol version in a future release so newer nodes will know which version of the protocol they need to "speak". However, in this case the old nodes are simply "trusting" newer nodes to decide if they want to accept the connection and not, and the responsibility is on the newer nodes to communicate in a backwards compatible way.

### DIFF
--- a/crates/node/src/network/handshake.rs
+++ b/crates/node/src/network/handshake.rs
@@ -63,10 +63,11 @@ mod tests {
     #[tokio::test]
     async fn test_p2p_handshake_accept_higher_version() {
         let (a, mut b) = tokio::io::duplex(1024);
+        let new_version = MPC_PROTOCOL_VERSION + 1;
         let a_result = do_handshake(a);
         let mut buf = [0u8; 5];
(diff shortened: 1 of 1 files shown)


SimonRastikian (dismissed):
Things look good to me  aside from some clippy issues

PR #1780, opened 2026-01-16
ci: refactor CI tests to group fast tests in a single run

### LINKED ISSUE #1779: consolidate CI actions
### Background

Currently we have many actions that run quite fast. It would be nice to consolidate them so that not so many runners are spawned for cheap checks.

### User Story

Less runner usage optimized CI

### Acceptance Criteria

All the cheap CI actions are consolidated in a single action, re-using cargo-make to avoid having so much logic in CI

_No response_

gilcu3:
Closes #1779


.github/workflows/ci-fast.yml
```diff
@@ -0,0 +1,65 @@
+name: Fast CI checks
```
netrome: Nit, using the name "fast" feels weird here. Perhaps this should just be `CI checks` and we can break out the slow ones (i.e. nextest for now) instead. Longer term, it would be nice to have unit tests be part of the normal "fast checks" as well, but since our unit and integration tests are a bit mixed and some are quite slow that has to wait a bit.

netrome (approved):
Looks good to me, but I think it would be even more clear if we had the default "fast" checks in `ci.yml` and instead only broke out the remaining slow checks (`nextest`, building docker image etc.) instead. Wdyt?

.github/workflows/ci-fast.yml
gilcu3: yeah separating integration tests outside unit tests would make this faster. Yeah I can add the fast checks to the other file, but seemed already a bit cluttered.

Makefile.toml
```diff
@@ -12,6 +12,7 @@ dependencies = [
   "zizmor",
   "check",
   "clippy",
+  "clippy-all-features",
```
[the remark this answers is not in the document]
netrome: True. Though we're still talking minutes to compile the project.

.github/workflows/ci.yml
```diff
@@ -336,89 +277,53 @@ jobs:
...
 
-      - name: cargo-deny check
-        run: |
-          cargo deny --all-features check
+      - name: Install cargo-make
+        uses: taiki-e/install-action@81ee1d48d9194cdcab880cbdc7d36e87d39874cb # v2.62.45
+        with:
+          tool: cargo-make@0.37.24
+
+      - name: Run cargo-make fast checks
```
netrome: I'm a bit torn on the name "fast checks" but I don't have any good better suggestions. Perhaps "hygiene checks"? Or "sanity checks"? Eh "fast checks" still works.

Makefile.toml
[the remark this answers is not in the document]
gilcu3: yes, the idea is to order the jobs with speed in mind. If something easy fails it will fail fast.

PR #1801, opened 2026-01-19
fix: TCP user keepalive

andrei-near:
Another layer of protectiong agains half-open TCP connections during active writes. With this, we have 3 layers in place:
Write timeouts (10s): Application-level detection
TCP_USER_TIMEOUT (30s): OS-level detection for unacknowledged data
TCP keepalive (existing): Detects idle dead connections

This resolves the issue where one side thinks the connection is closed but the other side continues writing.

### DIFF
--- a/crates/node/src/config.rs
+++ b/crates/node/src/config.rs
@@ -8,7 +8,12 @@ use serde::{Deserialize, Serialize};
 #[cfg(feature = "network-hardship-simulation")]
 use std::fs;
 
-use std::{net::SocketAddr, path::Path};
+use std::{
+    net::{Ipv4Addr, SocketAddr},
+    path::Path,
+};
+
+const DEFAULT_PPROF_PORT: u16 = 34001;
 
 pub type AesKey256 = [u8; 32];
 pub type AesKey128 = [u8; 16];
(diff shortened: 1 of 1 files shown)


crates/node/src/p2p.rs
```diff
@@ -491,6 +524,33 @@ fn configure_tcp_stream(tcp_stream: TcpStream) -> anyhow::Result<TcpStream> {
...
+                std::mem::size_of::<u32>() as libc::socklen_t,
+            );
+            if result != 0 {
+                return Err(anyhow::anyhow!(
+                    "Failed to set TCP_USER_TIMEOUT: {}",
+                    std::io::Error::last_os_error()
+                ));
+            }
+        }
+    }
```
DSharifi: We can avoid using unsafe libc calls and use the `socket2` crate from the rustlang team which we already depend on to set the `TCP_USER_TIMEOUT` with 
https://docs.rs/socket2/latest/socket2/struct.Socket.html#method.set_tcp_user_timeout

There's code just above using this crate for setting socket options, so it should be fairly straight forward to change

andrei-near: Nice! checked socket2 docs but did not found at the time this option.
Reviewing the code

DSharifi (dismissed):

DSharifi (approved):

netrome (approved):
Thanks for doing this! :pray:

PR #1816, opened 2026-01-20
chore: make pprof web server endpoint queriable

### LINKED ISSUE #1815: make pprof web server endpoint queriable
### Background

In #1630 we added functionality to create pprof flamegraphs with default duration of 30s and 1000hz sampling frequency.

In production we often see that these values of the sampling differ quite a lot in where spend CPU time, thus we'd like to be able to configure the durations and frequencies to get better reports for average times over long runs.

To do this the axum web point that accepts pprof requests should accept query parameters that can configure the duration and frequency.

### User Story

_No response_

### Acceptance Criteria

The pprof web server endpoint can accept query parameters that defines the pprof sampling duration and frequency.

_No response_

DSharifi:
closes #1815

### DIFF
--- a/pytest/tests/shared_cluster_tests/test_web_endpoints.py
+++ b/pytest/tests/shared_cluster_tests/test_web_endpoints.py
@@ -130,9 +130,13 @@ def assert_pprof_endpoint(pprof_address: str):
     )
 
     assert response.status_code == 200
-    assert response.headers["Content-Type"].startswith("image/svg+xml")
 
-    # Basic SVG sanity checks
-    body = response.text
-    assert body.startswith("<svg")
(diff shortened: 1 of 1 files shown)


netrome (approved):
Nice!

gilcu3 (approved):
Thank you!

Just to make sure I understand the changes, the pytest part is just adding a missing test there? I don't really see the relation with the change in the node

DSharifi:
> Just to make sure I understand the changes, the pytest part is just adding a missing test there? I don't really see the relation with the change in the node

Correct, it's testing an existing endpoint. But it's also sanity checking that the query parameters work. This test would have failed before the change to the web server.

PR #1837, opened 2026-01-23
chore(metrics): track tokio runtime and task metrics and export it to prometheus

### LINKED ISSUE #1799: metrics: include tokio runtime metrics to node
### Background

We should add  [tokio-metrics](https://docs.rs/tokio-metrics/latest/tokio_metrics/struct.RuntimeMetrics.html) to our metrics, so we can correlate CPU usage and degradations with interesting metrics such as mean poll duration poll durations and queue depths.

I also have suspicion that we are severly blocking the runtime with long poll durations, and need these metrics to confirm that.

### User Story

_No response_

### Acceptance Criteria

include tokio runtime metrics on the node

_No response_

DSharifi:
closes #1799


gilcu3 (changes requested):
Looks good to me, although I may have found a problem.
- ~~Why are we calling `run_monitor_loop` twice?~~ EDIT: not twice, they were different functions, approving as the rest was non-blocking

On a side note, did you consider using [metrics_exporter_prometheus](https://docs.rs/metrics-exporter-prometheus/latest/metrics_exporter_prometheus/) as recommended in the [tokio_metrics docs](https://docs.rs/tokio-metrics/latest/tokio_metrics/#monitoring-and-publishing-metrics)?

gilcu3 (dismissed):
Thank you, please check out my previous comment:

> On a side note, did you consider using [metrics_exporter_prometheus](https://docs.rs/metrics-exporter-prometheus/latest/metrics_exporter_prometheus/) as recommended in the [tokio_metrics docs](https://docs.rs/tokio-metrics/latest/tokio_metrics/#monitoring-and-publishing-metrics)?

netrome (dismissed):
No hard blockers from my side, but it seems Copilot found one we should address.

Though I feel strongly about the introduction of dynamic dispatch and I think we

1. Don't need polymorphism here and can simplify the code.
2. If you still want polymorphism, I'd prefer enums over trait objects in general since they preserve more type information, doesn't force you to care about "dyn compatibility" in async code and are typically more performant.

netrome (approved):

gilcu3 (approved):

PR #1861, opened 2026-01-26
feat: change `Participants` from a `Vec`-based structure to a `BTreeMap` for improved gas usage

### LINKED ISSUE #1572: Replace `Vec` with `Set`/`Map`-based data structure in `Participants` for improved performance
## Description

The `Participants` struct currently uses a `Vec<(AccountId, ParticipantId, ParticipantInfo)>` to store participants:
https://github.com/near/mpc/blob/a39b60a4246903cd06772ecf3a35d35a9166ecf7/crates/contract/src/primitives/participants.rs#L40-L45

This has several drawbacks:

1. **O(n) lookup complexity**
    Methods like `is_participant()`, `info()`, and `id()` require linear search through the vector
1. **Manual uniqueness validation**
    The `validate()` method needs to explicitly check for duplicate IDs and accounts by constructing temporary `BTreeSet`s
1. **Insertion complexity**
    `insert_with_id()` must scan the entire vector to check for duplicates
1. **Inefficient operations**

Several operations that should be `O(1)` or `O(log n)` are `O(n)`. Replacing the `Vec` with a `Set` or `HashMap` would:
- Provide `O(log n)` or `O(1)` lookups
- Enforce uniqueness by construction
- Simplify the code by removing manual duplicate checks
- Improve performance, especially as the number of participants grows

**Context**: This improvement was identified [during](https://github.com/near/mpc/pull/1559/files/0b4947e1d6baa4e5e13623a84e48c85c4695c57e#r2567608506) review of PR #1559 where we added `remove_votes_from_non_participants` functionality that filters voters by checking `is_participant()` on each one.

### User Story

As a developer, I want to use optimal data structures to improve the performance of operations.

### Acceptance Criteria

- [ ] Replace `participants: Vec<(AccountId, ParticipantId, ParticipantInfo)>` with an appropriate `Set`/`Map`
- [ ] Update all methods in `Participants` to work with the new data structure
- [ ] Remove or simplify the `validate()` method since uniqueness will be enforced by construction
- [ ] Performance of `is_participant()`, `info()`, and `id()` methods is improved from `O(n)` to `O(log n)` or better
- [ ] The `participants()` method continues to return an iterator or collection that maintains the same API contract for external callers
- [ ] Bonus points for replacing the `(AccountId, ParticipantId, ParticipantInfo)` tuple with a proper struct.

https://github.com/near/mpc/pull/1559/files/0b4947e1d6baa4e5e13623a84e48c85c4695c57e#r2567608506

pbeza:
Fixes #1572


crates/backup-cli/assets/contract_state.json
```diff
@@ -165,4 +165,4 @@
       "proposal_by_account": {}
     }
   }
-}
\ No newline at end of file
+}
```
gilcu3: nit: tools should not modify file endings without user intervention

kevindeforth (commented):
preliminary review

crates/contract/tests/sandbox/common.rs
```diff
@@ -394,9 +394,12 @@ pub async fn submit_attestations(
...
 ) {
-    let futures: Vec<_> = participants
-        .participants()
-        .iter()
+    // Sort participants by ParticipantId to match the order of accounts
+    let mut sorted_participants: Vec<_> = participants.participants().collect();
+    sorted_participants.sort_by_key(|(_, participant_id, _)| participant_id.get());
+
+    let futures: Vec<_> = sorted_participants
+        .into_iter()
```
gilcu3: this looks like a trick. I would rather refactor the tests so that they do not assume a given order of participants. If that's too much, we can add TODOs to handle as a follow up

gilcu3 (changes requested):
I think it looks good overall. Requesting changes mostly related to the future of the code:

- I think `next_id` in Participants should be removed, as it was meant to guarantee an invariant that is no longer needed. Correct me if I am wrong and this was also used for something else
- There are several new structs that are used during migration only. I think we should separate them from the implementation in such a way that they can be easily removed later. I think the current code mixes them.
- I think tests that require participants order should be fixed to avoid requiring that, but that can be a follow-up

crates/contract/src/primitives/participants.rs
```diff
@@ -37,13 +38,100 @@ impl Display for ParticipantId {
...
 }
 
+/// A participant entry with all fields. Only available in tests.
+#[cfg(any(test, feature = "test-utils"))]
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct ParticipantEntry {
+    pub account_id: AccountId,
+    pub id: ParticipantId,
+    pub info: ParticipantInfo,
+}
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/pull/1861#discussion_r2770311370 ✅

crates/backup-cli/assets/contract_state.json
pbeza: https://github.com/near/mpc/pull/1861/commits/8c7d99046043a4b43e6d7b6519de0917ec04a857

crates/contract/src/primitives/thresholds.rs
```diff
@@ -311,15 +311,17 @@ mod tests {
     fn test_proposal_non_unique_ids() {
         let params = gen_threshold_params(10);
 
-        // Add duplicate participants
+        // With BTreeMap-based Participants, duplicate AccountIds are automatically deduplicated.
+        // We can still test that the validate_incoming_proposal catches proposals with
+        // incorrect participant counts.
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/pull/1861/commits/bd6fdff728c91976e96053db5c5d7ca5bd4562c8

crates/contract/src/state/key_event.rs
```diff
@@ -207,6 +206,24 @@ impl KeyEvent {
...
     }
+
+    /// Constructs a KeyEvent from its parts during migration.
+    /// This bypasses the normal initialization logic.
+    pub fn from_migration(
+        epoch_id: EpochId,
+        domain: DomainConfig,
+        parameters: ThresholdParameters,
+        instance: Option<KeyEventInstance>,
+        next_attempt_id: AttemptId,
```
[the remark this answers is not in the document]
pbeza: Yes — would you suggest moving it to the migration file? If so, `KeyEvent`’s fields are private, so `from_migration` would need to live in the same module to construct it via a struct literal. Alternatively, we could make the fields `pub(crate)`. WDYT?

crates/contract/tests/sandbox/tee_cleanup_after_resharing.rs
```diff
@@ -83,17 +83,20 @@ async fn test_tee_cleanup_after_full_resharing_flow() -> Result<()> {
     assert_eq!(initial_and_non_participants, expected_node_ids);
 
     // Now, we do a resharing. We only retain `threshold` of the initial participants
+    // Sort by ParticipantId to ensure deterministic selection regardless of iteration order
+    let mut sorted_participants: Vec<_> = initial_participants.participants().collect();
+    sorted_participants.sort_by_key(|(_, participant_id, _)| participant_id.get());
+
```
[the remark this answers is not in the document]
pbeza: Fixed here: https://github.com/near/mpc/pull/1861/commits/9d60d3d88b769e4039f8c4fef258a5a037811920. ✅

crates/contract/src/primitives/participants.rs
```diff
@@ -37,13 +38,91 @@ impl Display for ParticipantId {
...
+#[cfg_attr(
+    all(feature = "abi", not(target_arch = "wasm32")),
+    derive(schemars::JsonSchema)
+)]
+pub struct Participants {
+    /// The next [`ParticipantId`] to assign when inserting a new participant.
+    /// Always greater than all existing participant IDs.
+    next_id: ParticipantId,
+    /// Primary storage mapping [`AccountId`] to [`ParticipantData`].
+    participants: BTreeMap<AccountId, ParticipantData>,
```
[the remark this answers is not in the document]
pbeza: Outdated comment? Let me ask Copilot to take another look.

crates/contract/src/primitives/participants.rs
```diff
@@ -69,121 +87,142 @@ impl Participants {
...
             .iter()
-            .find(|(_, p_id, _)| p_id == id)
-            .map(|(a_id, _, _)| a_id.clone())
+            .find(|(_, data)| data.id == *id)
+            .map(|(a_id, _)| a_id.clone())
             .ok_or_else(|| crate::errors::InvalidState::ParticipantIndexOutOfRange.into())
     }
 
     /// Returns a subset of the participants according to the given range of indices.
     pub fn subset(&self, range: std::ops::Range<usize>) -> Participants {
```
gilcu3: I did not know we had this function!

crates/contract/src/v3_4_1_state.rs
```diff
@@ -6,26 +6,214 @@
...
+            OldProtocolContractState::Resharing(state) => {
+                crate::state::ProtocolContractState::Resharing(state.into())
+            }
+        }
+    }
+}
+
+/// StaleData in v3.4.1 is empty (participant_attestations was cleaned up in 3.4.0 → 3.4.1).
+#[derive(Debug, Default, BorshSerialize, BorshDeserialize)]
+struct StaleData {}
```
gilcu3: why do we need this struct if that's the case?

crates/contract/tests/sandbox/upgrade_to_current_contract.rs
```diff
@@ -41,6 +41,52 @@ fn contract_code(network: Network) -> &'static [u8] {
...
+            "participants": participants_vec
+        },
+        "threshold": params.threshold()
+    })
+}
+
+/// Wrapper around [`execute_key_generation_and_add_random_state_with_proposal`]
+/// that serialises the threshold proposal in the legacy Vec-of-tuples format
+/// expected by old contract binaries.
+async fn add_random_state_to_old_contract(
```
gilcu3: more accurate, and seems to align better with existing code

```suggestion
async fn execute_key_generation_and_add_random_state_to_old_contract(
```

crates/contract/src/tee/tee_state.rs
```diff
@@ -254,14 +252,16 @@ impl TeeState {
...
             .collect();
 
-        if participants_with_valid_attestation.len() != participants.len() {
-            let participants_with_valid_attestation =
-                Participants::init(participants.next_id(), participants_with_valid_attestation);
+        if !invalid_accounts.is_empty() {
+            let mut participants_with_valid_attestation = participants.clone();
+            for account in &invalid_accounts {
+                participants_with_valid_attestation.remove(account);
+            }
```
gilcu3: this change added the requirement of the "remove" function to the prod contruct that was previously test only. I think the previous code was slightly more efficient, wdyt?

gilcu3 (approved):
This was a big PR, I had several comments, but none of them blocking IMO. There are also the comments above by claude. I think many of them, both mine and claude's are good to have.

kevindeforth (changes requested):
This PR currently does two things:
- it changes the contract-internal Participants representation
- it propagates those changes to the contract-interface crate.

It is unclear to me, why we need to add the new internal representation to the contract-interface crate. It makes reviewing this PR much harder.
To me, it looks like there is a breaking change in the abi.

I would prefer we split this up into two separate PR's, with the first one only doing the contract-internal changes. If we stay backwards-compatible, then this shouldn't require changes to any of our other crates.

Requesting changes for now - but we can discuss this on slack or over a call.

gilcu3 (changes requested):
Unfortunately we figured that the current PR introduces breaking changes wrt the 3.4.1 contract, checked with localnet automation:
```
mpc on  1572-replace-vec-with-setmap-based-data-structure-in-participants-for-improved-performance via 🦀 v1.86.0 took 59s
❯ ./scripts/launch-localnet.sh
Using mpc-contract binary from ./target/near/mpc_contract/mpc_contract.wasm
Creating network with 2 mpc nodes and threshold 2
Logs will be stored in /tmp/mpc-localnet.9QpDvx
Cleaning ~/.near folder
Started: neard PID: 261214
Waiting 60 seconds for neard to start properly
Creating mpc-contract account
Deploying mpc-contract
Creating mpc-node accounts
Creating mpc nodes configuration
Starting mpc nodes
Started: mpc-node-1.test.near PID: 266942
Started: mpc-node-2.test.near PID: 266944
Waiting 20 seconds for mpc nodes to start properly
Adding account keys for the nodes
Initializing contract
Adding domains to contract
Waiting 20 seconds for key generation to happen
Command near contract call-function as-read-only mpc-contract.test.near state json-args {} network-config mpc-localnet now 2>&1 | grep Running failed. Retrying in 2s...
Command near contract call-function as-read-only mpc-contract.test.near state json-args {} network-config mpc-localnet now 2>&1 | grep Running failed. Retrying in 2s...
Command near contract call-function as-read-only mpc-contract.test.near state json-args {} network-config mpc-localnet now 2>&1 | grep Running failed. Retrying in 2s...
Command near contract call-function as-read-only mpc-contract.test.near state json-args {} network-config mpc-localnet now 2>&1 | grep Running failed. Retrying in 2s...
```

For some reason the old contract does not transition to Running state

pbeza:
Thanks for all the reviews!

I’m [putting](https://nearone.slack.com/archives/C0912BTG51T/p1771498006179029?thread_ts=1771163703.865359&cid=C0912BTG51T) this on hold for now, since addressing all the concerns would take too much time and we need to prioritize the HOT Wallet migration work.

netrome:
Feels bad to have to close this one, but I don't see us getting back to this anytime soon. Feel free to open if you feel like picking this up again.

PR #1878, opened 2026-01-27
chore: enable ignored tests

### LINKED ISSUE #1150: Fix and re-enable broken contract tee tests
### Background

In https://github.com/near/mpc/issues/1148 we disabled three test cases that were broken due to a time bomb. Essentially all of these tests submit expired attestations, because the sandbox environment currently uses the system time.

### User Story

As a developer I want to have test coverage without time bombs.

### Acceptance Criteria

The ignored tests are re-enabled and execute well without time bombs.

_No response_

gilcu3:
Closes #1150

### DIFF
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -1756,6 +1756,7 @@ mod tests {
     use sha2::{Digest, Sha256};
     use test_utils::attestation::{
         image_digest, mock_dto_dstack_attestation, near_account_key, p2p_tls_key,
+        VALID_ATTESTATION_TIMESTAMP,
     };
     use test_utils::contract_types::dummy_config;
     use threshold_signatures::confidential_key_derivation as ckd;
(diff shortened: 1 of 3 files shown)


crates/contract/src/lib.rs
```diff
@@ -3608,4 +3612,168 @@ mod tests {
...
+            contract,
+            participant_account_ids,
+            attestation,
+            tls_key,
+            mpc_hash,
+            near_public_key,
+        )
+    }
+
+    /// Sets up a contract with an approved MPC hash by having the first two participants vote for it.
```
barakeinav1:
```suggestion
    /// Sets up a contract with an approved MPC hash by having the participants vote for it.
```

barakeinav1 (dismissed):
looks good. thanks.
one tiny comment

crates/contract/tests/sandbox/tee.rs
```diff
@@ -588,7 +526,8 @@ async fn get_attestation_overwrites_when_same_tls_key_is_reused() {
     let validation_success =
         submit_participant_info(participant_account, &contract, &first_attestation, &tls_key)
             .await
-            .unwrap();
+            .unwrap()
+            .is_success();
```
DSharifi: Now that you improved `submit_participant_info` to return the result instead of a bool, can you avoid casting it back into a bool to assert on that?

Ideally you do another unwrap/expect or assert that it matches ok. Otherwise when we have a regression we don't get a meaningful error message for what the error was.

barakeinav1 (approved):

PR #1885, opened 2026-01-28
test: add CI test for mpc-node through non-tee launcher

### LINKED ISSUE #1758: Add CI job to validate NON-TEE launcher-based MPC deployment
### Background

follow up from https://github.com/near/mpc/pull/1735

With the introduction of NON-TEE launcher support, we now have a new supported deployment mode for MPC nodes that mirrors the TEE setup (docker-compose + launcher), but runs outside of a CVM.

While we already have CI coverage for:

building the MPC node image, and

validating that the MPC node container can initialize and start (TEE-oriented flow),

we currently do not have CI coverage that exercises the launcher itself in NON-TEE mode.

This means:

regressions in the launcher logic (platform handling, env propagation, sanitization, container startup) may go unnoticed,

the newly added NON-TEE path is not validated end-to-end in CI.

Proposed Change

Add a new CI job (separate from the existing TEE build job) that:

Builds (or pulls) an MPC node Docker image.

Runs the launcher Docker image in NONTEE mode using a docker-compose file.

Provides a test user_config file via bind-mount (as in non-TEE deployments).

Verifies that:

the launcher starts successfully,

the MPC container is launched by the launcher,

the MPC container remains running for a fixed period (e.g. 30–60 seconds).

Fails the job if:

the launcher exits unexpectedly,

the MPC container fails to start,

or the MPC container crashes shortly after startup.

This should be conceptually similar to the existing CI logic used to validate MPC node startup (e.g. docker run … sleep 60 … docker ps), but routed through the launcher instead of invoking the MPC image directly.

### User Story

add an MPC developer I'd like to have CI check for the launcher /mpc builds

### Acceptance Criteria

CI job is deployed

_No response_

gilcu3:
Closes #1758

### DIFF
--- a/scripts/check-mpc-node-docker-starts.sh
+++ b/scripts/check-mpc-node-docker-starts.sh
@@ -25,6 +25,12 @@ if $USE_LAUNCHER; then
   export LAUNCHER_IMAGE_NAME
   docker compose -f launcher_docker_compose_nontee.yaml up -d
   sleep 10
+  launcher_logs=$(docker logs --tail 10 "$LAUNCHER_IMAGE_NAME" 2>&1)
+  if ! echo "$launcher_logs" | grep "MPC launched successfully."; then
+    echo "MPC launcher image did not start properly"
+    echo "$launcher_logs"
+    exit 1
+  fi
   CONTAINER_ID=$(docker ps -aqf "name=^mpc-node$")
 else
   touch /tmp/image-digest.bin


tee_launcher/launcher_docker_compose_nontee.yaml
```diff
@@ -1,9 +1,7 @@
-version: '3.8'
-
```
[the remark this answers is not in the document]
barakeinav1: sure, this is not measured so we can safely removed it.

we do have this also in launcher_docker_compose.yaml - should we remove it there as well (separate PR)?

scripts/check-mpc-node-docker-starts.sh
```diff
@@ -0,0 +1,73 @@
...
+    -e MPC_ACCOUNT_ID=test_image.near \
+    -e RUST_BACKTRACE="full" \
+    -e RUST_LOG="mpc=debug,info" \
+    -e MPC_SECRET_STORE_KEY=BD399143F5B3126098B0EAA023A0E730 \
+    -e MPC_P2P_PRIVATE_KEY=ed25519:2WBi5gRyJntYA7dCyddiwNV2yNTdr5uZhrP4WX5GNBL5DSVCSR7ESvcXF2DBfY5oPYhzBmHnguPVSXjr6UCi8h1g \
+    -e MPC_ACCOUNT_SK=ed25519:6X3Bnghdf89WjHQfHDFYjW2UeNaNqCQ1AUGxX7zgvbJT4KSoeXzuHEdux6A2jsphTZTmS4SUGQRyYqC2ik3UrMP \
+    -e NEAR_BOOT_NODES=ed25519:ERguu7jQuYk8pxNsRC6FdezvNsegBPva1GRGqjmtD7i2@10.10.10.10:24567 \
+    -e MPC_CONTRACT_ID=v1.signer_test \
+    -e MPC_IMAGE_HASH=5ba283860c0efa3d4c3e08a76a2b77fab4725baad4f48504eac858e04af7fd64 \
+    -e MPC_LATEST_ALLOWED_HASH_FILE=/tmp/image-digest.bin \
```
barakeinav1:
```suggestion
    -e MPC_LATEST_ALLOWED_HASH_FILE=/data/image-digest.bin \
```

scripts/check-mpc-node-docker-starts.sh
```diff
@@ -0,0 +1,73 @@
...
+done
+
+: "${NODE_IMAGE_NAME:=mpc-node}"
+: "${LAUNCHER_IMAGE_NAME:=mpc-launcher-nontee}"
+
+if $USE_LAUNCHER; then
+  cd tee_launcher
+  export LAUNCHER_IMAGE_NAME
+  docker compose -f launcher_docker_compose_nontee.yaml up -d
+  CONTAINER_ID=$(docker compose -f launcher_docker_compose_nontee.yaml ps -q launcher)
```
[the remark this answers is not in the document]
gilcu3: alright, added some commits, including the check you proposed, so this should be fixed

scripts/check-mpc-node-docker-starts.sh
```diff
@@ -0,0 +1,80 @@
...
+
+if [ -z "$CONTAINER_ID" ]; then
+  echo "❌ Failed to start container"
+  exit 1
+fi
+
+echo "Container started: $CONTAINER_ID"
+
+# Check if container is actually running
+sleep 60
```
SimonRastikian: How did we come up to this number?

SimonRastikian: can we not just add a while loop for 1 min. If things work under this time you continue otherwise you raise an error?

gilcu3: there is not much loss in CI here waiting for 1 min, as other steps take much longer anyway. Proper way would be to check if it is up every second for example, but that is something to optimize in a future PR :)

Issue #657, opened 2025-07-16
[Task] Upgrade near core dependencies in `devnet` crate


ghost:
### Background

These near-core deps below are all stale, and need to be upgraded. We already have these dependencies defined in the workspace, and we should use the workspace version. The problem is that there are breaking changes between the devnet version and the latest which the workspace and node depends on.

https://github.com/near/mpc/blob/c4da39c9afea0954317a0bd83c54a5a0402eb74d/devnet/Cargo.toml#L14-L18

### User Story

.

### Acceptance Criteria

All the near-core dependencies in `devnet` crate depend on the workspace version. This will require some code changes in the devnet code to account for the breaking API changes.

netrome:
We'll get to this when we revamp devnet in the scalability work.

PR #1894, opened 2026-01-29
chore: bump near crates, remove AccountId conversions

### LINKED ISSUE #1893: unify account_id usage
### Background

In #1511 we had to patch the `AccountId` type from `near-sdk` and the `near-account-id` crate, as we needed both but they were different. Later `near-sdk` was updated with a compatible version. We should bump this version and remove the workarounds introduced

### User Story

_No response_

### Acceptance Criteria

A single `AccountId` type is used in all our crates

[context](https://nearone.slack.com/archives/C07UW93JVQ8/p1765880647921989)

gilcu3:
Closes #1893

- Only in devnet some `AccountId` conversions are still required. These can only be removed when #657 is solved


crates/contract/src/primitives/key_state.rs
```diff
@@ -5,7 +5,6 @@ use crate::errors::{DomainError, Error, InvalidState};
 use near_account_id::AccountId;
 use near_sdk::{env, near};
 use std::fmt::Display;
-use utilities::AccountIdExtV1;
```
netrome: Why haven't we removed this crate entirely? It seems like these are still used in devnet, but couldn't we bump the crates there as well?

netrome (approved):
Nice stuff. Just one question.

crates/contract/src/primitives/key_state.rs
netrome: Reynaldo already mentioned this on slack (prior to me asking x) ) - devnet needs to be bumped separately due to breaking changes, see https://github.com/near/mpc/issues/657

PR #1905, opened 2026-01-29
docs: proposal for indexer API

### LINKED ISSUE #1904: Propose Indexer API
### Background

Propose indexer API in the context of #1892 

### User Story

as a dev, I like to get early feedback on design decisions

### Acceptance Criteria

propose and agree on an indexer API

_No response_

kevindeforth:
resolves #1904


docs/migration-service-tee-integration.md
```diff
@@ -0,0 +1,310 @@
...
+- block stream 
+- transaction sender
+
+
+#### State View
+
+The indexer should offer a convenient method for viewing and subscribing to MPC contract state. While it would be sufficient to only expose a simple view function like the following:
+
+```rust
+#[async_trait::async_trait]
```
pbeza: IIUC, since Rust 1.75 (December 2023), `async` functions in traits are stable, so you can drop this attribute.


```suggestion
```

docs/migration-service-tee-integration.md
```diff
@@ -0,0 +1,310 @@
...
+Alternatively, if we are concerned about having `tokio::watch` and `oneshot` channels in our API, we could return a struct with the following traits:
+
+```rust
+#[async_trait::async_trait]
+trait ContractStateStream<T> {
+    /// is synchronous, contains the last seen value
+    fn latest(&self) -> (BlockHeight, &T);
+    /// must be cancellation safe
+    /// returned BlockHeight is monotonically increasing
+    async fn next(&mut self) -> (BlockHeight, &T);
```
pbeza: Maybe `BlockHeight` should just be part of `T` (which I assume represents the contract state)?

docs/migration-service-tee-integration.md
```diff
@@ -0,0 +1,310 @@
...
+
+Alternatively, if we are concerned about having `tokio::watch` and `oneshot` channels in our API, we could return a struct with the following traits:
+
+```rust
+#[async_trait::async_trait]
+trait ContractStateStream<T> {
+    /// is synchronous, contains the last seen value
+    fn latest(&self) -> (BlockHeight, &T);
+    /// must be cancellation safe
+    /// returned BlockHeight is monotonically increasing
```
pbeza: I don’t think guaranteeing a monotonically increasing `BlockHeight` is enough. We probably want a stronger guarantee here — namely that we receive all blocks, with no gaps, no? (see [related discussion](https://nearone.slack.com/archives/C07UW93JVQ8/p1765239374976249?thread_ts=1765238400.243289&cid=C07UW93JVQ8))

docs/migration-service-tee-integration.md
```diff
@@ -0,0 +1,310 @@
...
+}
+
+/// Defines the timeout between consecutive attempts
+pub enum BackoffStrategy {
+    Fixed(Duration),
+    Exponential {
+        initial: Duration,
+        max: Duration,
+    },
+}
```
pbeza: Do we already have something like this in the current indexer? (sorry, I haven’t dug into the indexer code deeply enough yet to check myself). This sounds like a pretty natural thing to configure for an indexer, so I’d expect it to be defined somewhere already — I’d be surprised if it isn’t.

docs/migration-service-tee-integration.md
```diff
@@ -0,0 +1,310 @@
...
+```
+
+#### BlockStream
+
+For the next few months, the only expected user of block streams is the MPC node. This might change in the future if we use the indexer crate for monitoring and testing our MPC code, but for now, it seems safe to just port the existing design to the indexer until we have a better idea of what changes we would want from it.
+
+```Rust
+// c.f. https://github.com/near/mpc/issues/236 for start_block_height
+trait MpcEventSubscriber {
+    async fn subscribe(interval: Duration, channel_size: usize, start_block_height) -> anyhow::Result<mpsc::Receiver<ChainBlockUpdate>>;
```
pbeza: Would it make sense to reuse `SubscriberPolicy` here instead of passing `interval: Duration`?

docs/migration-service-tee-integration.md
```diff
@@ -0,0 +1,310 @@
...
+}
+
+pub struct ChainBlockUpdate {
+    pub block: BlockViewLite,
+    pub signature_requests: Vec<SignatureRequestFromChain>,
+    pub completed_signatures: Vec<SignatureId>,
+    pub ckd_requests: Vec<CKDRequestFromChain>,
+    pub completed_ckds: Vec<CKDId>,
+}
```
pbeza: I’d suggest linking directly to the code instead — otherwise it’s not clear whether the struct was introduced by you or copied from existing code. Since the link points to a specific commit, it won’t get outdated anyway (we don’t allow force-pushes to `main` 🙂).

```suggestion
https://github.com/near/mpc/blob/e10da17ec59fee4c511f245e2a37d243fc952774/crates/node/src/indexer/handler.rs#L69-L75
```

docs/migration-service-tee-integration.md
```diff
@@ -0,0 +1,310 @@
...
+The node is already running an indexer. It makes sense to re-use the existing logic for the following reasons:
+- Generally, it is preferred to re-use code where applicable;
+- The indexer exposes some Near internals, which have experienced breaking changes in the past. It would thus be good to only have to maintain one place where we depend on these internals.
+- There exist larger plans for the indexer. It is quite useful to have a standalone binary that is capable of monitoring the MPC contract and recognizing transactions related to the MPC network. We could use it as a tool for monitoring our production deployments.
+
+On a high-level, we expect to be doing the following:
+
+1. Clean up the indexer API in the MPC node, such that it will be easy to separate it from the node. We **should not** nearcore internals - we would like to have a single point of failure in case nearcore decides to change the indexer API.
+2. Move the indexer code into its own crate.
+3. Add the indexer crate as a dependency to the migration and backup service.
```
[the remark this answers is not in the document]
netrome: Yes, I'd assume moving the indexer code into its own crate in 2. entails having the MPC node depend on this crate.

docs/migration-service-tee-integration.md
```diff
@@ -0,0 +1,310 @@
...
+#### State View
+
+The indexer should offer a convenient method for viewing and subscribing to MPC contract state. While it would be sufficient to only expose a simple view function like the following:
+
+```rust
+#[async_trait::async_trait]
+trait MpcContractStateView {
+    async fn get<T: DeserializeOwned>(
+        &self,
+        endpoint: MpcContractEndpoint,
```
netrome: What's an MpcContractEndpoint and why is it needed for the state view? I'd spontaneously expect this getter to not take any parameters as input.

docs/migration-service-tee-integration.md
```diff
@@ -0,0 +1,310 @@
...
+            TEE_ACCOUNTS_ENDPOINT,
+            policy.clone(),
+        )?;
+
+    let my_migration_info_receiver =
+        subscribe_as_watch::<MigrationInfo>(
+            &indexer.state,
+            MIGRATION_INFO_ENDPOINT,
+            policy,
+        )?;
```
pbeza: Does it actually make sense to have this copy-pasted setup with separate tokio tasks polling each of these independently? Why not poll them all together in one go somehow? (Sorry if that’s a dumb idea — I might definitely be missing something.)

docs/migration-service-tee-integration.md
```diff
@@ -0,0 +1,310 @@
...
+    async fn next(&mut self) -> (BlockHeight, &T);
+}
+
+#[async_trait::async_trait]
+trait MpcContractStateSubscriber {
+    async fn subscribe<T: DeserializeOwned + PartialEq + Send + 'static>(
+        &self,
+        endpoint: MpcContractEndpoint,
+        subscriber_policy: SubscriberPolicy,
+    ) -> anyhow::Result<Box<dyn ContractStateStream<T> + Send>>;
```
netrome: I'd like to avoid trait objects unless absolutely necessary. Any reason `ContractStateStream` couldn't be a concrete type (potentially with the generic parameter, although I'm also not convinced why we need generics here)?

docs/migration-service-tee-integration.md
```diff
@@ -0,0 +1,310 @@
...
+}
+```
+
+where `SubscriberPolicy` details how the indexer should query new contract state (e.g. exponential backoff, interval, etc), the behavior on error etc.
```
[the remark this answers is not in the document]
kevindeforth: We are not imposing any changes with respect to poll times.
The whole idea is that the user of the library decides how often they want to query. If they want 250ms, they get 250ms. If they want something else, they get something else.

docs/migration-service-tee-integration.md
```diff
@@ -0,0 +1,310 @@
...
+    pub signature_requests: Vec<SignatureRequestFromChain>,
+    pub completed_signatures: Vec<SignatureId>,
+    pub ckd_requests: Vec<CKDRequestFromChain>,
+    pub completed_ckds: Vec<CKDId>,
+}
+```
+
+##### Transaction Sender
+
+Similarly, it's probably fine for the first iteration to keep the existing `TransactionSender` trait:
```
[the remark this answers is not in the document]
kevindeforth: +1.

docs/migration-service-tee-integration.md
kevindeforth: Mabye. This is related to this comment here: https://github.com/near/mpc/pull/1905#discussion_r2764051645

and what I wrote here:
https://github.com/near/mpc/blob/3f8baef5b31d47a198d8abe69c25121338780931/docs/migration-service-tee-integration.md?plain=1#L174

tldr: I prefer not to touch this one for this PR, as it is not necessary for the TEE integration and I don't see any downside in leaving such a refactor for later.

docs/migration-service-tee-integration.md
```diff
@@ -0,0 +1,310 @@
...
+
+For the next few months, the only expected user of block streams is the MPC node. This might change in the future if we use the indexer crate for monitoring and testing our MPC code, but for now, it seems safe to just port the existing design to the indexer until we have a better idea of what changes we would want from it.
+
+```Rust
+// c.f. https://github.com/near/mpc/issues/236 for start_block_height
+trait MpcEventSubscriber {
+    async fn subscribe(interval: Duration, channel_size: usize, start_block_height) -> anyhow::Result<mpsc::Receiver<ChainBlockUpdate>>;
+}
+
+pub struct ChainBlockUpdate {
```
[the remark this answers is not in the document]
netrome: Right. In that case it sounds like a subscription model where you provide a declaration of which event types you want to listen to might be a good idea.

Issue #1897, opened 2026-01-29
Remove post_upgrade_cleanup method


netrome:
### Background

The smart contract has a method `post_upgrade_cleanup` that we needed for the 3.3 upgrade to clean some old attestations. This is not needed for subsequent upgrades, and we should remove the method.

### User Story

As an MPC user I don't want unused methods lying around

### Acceptance Criteria

We've removed the `post_upgrade_cleanup` contract method.

### Resources & Additional Notes

Note: We also need to remember to call this method after the mainnet upgrade.

PR #1914, opened 2026-01-30
chore: update 3.3.2 mainnet contract history, clean-up previous contract migrations

### LINKED ISSUE #1912: 3.3.2 in mainnet contract-history
### Background

The contract has been updated to 3.3.2 in mainnet. We need now to update the contract history to fix CI in main

### User Story

_No response_

### Acceptance Criteria

contract-history is updated for mainnet

_No response_

gilcu3:
Closes #1912 and #1897

### DIFF
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -49,13 +49,12 @@ use errors::{
 };
 use k256::elliptic_curve::PrimeField;
 
-use mpc_attestation::attestation::Attestation;
 use mpc_primitives::hash::LauncherDockerComposeHash;
 use near_sdk::{
     env::{self, ed25519_verify},
     log, near_bindgen,
     state::ContractState,
-    store::{IterableMap, LookupMap},
+    store::LookupMap,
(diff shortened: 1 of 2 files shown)


crates/contract/src/lib.rs
```diff
@@ -18,8 +18,7 @@ pub mod tee;
 pub mod update;
 #[cfg(feature = "dev-utils")]
 pub mod utils;
-pub mod v3_0_2_state;
-pub mod v3_2_0_state;
+pub mod v3_3_2_state;
```
gilcu3: I am not 100% if we were leaving the old states on purpose, but taking into account that our contract size is already near dangerous sizes, I opted for only leaving the necessary code there

DSharifi: We had a window where 3_0_2 was on mainnet and 3_2_0 was on testnet.

We can and should delete both the previous state modules in this PR now that they are unused.

DSharifi (approved):

netrome (approved):
Thanks!

Issue #1918, opened 2026-01-30
Foreign chain transactions high level design


netrome:
### Background

In https://github.com/near/mpc/pull/1851 a new feature has been proposed to extend the MPC network to allow MPC nodes to verify foreign chain transactions.

Since this is a big feature, it would be very helpful to compile a design proposal to facilitate effective design conversations and ensure we can make effective progress on getting this merged.

### User Story

As a developer I'd like to have key design decisions documented to ensure we're aligned and allowing us to proceed and focus on implementation details.

### Acceptance Criteria

We have a design doc for the foreign transaction validation feature. The design doc should contain the following:

1. **Motivation.** Why is this feature important? What are the use-cases we want to support?
2. **High level component design.** What are the major components we're implementing, and how are they interacting? How does the flow look end to end when using this feature? Charts following the c4 model would be helpful here.
3. **Risks**. What are the major risks if we implement this feature? Can we migrate pieces of it, or will this cause a big maintenance burden going forward.

### Resources & Additional Notes

Prototype implementation PR: https://github.com/near/mpc/pull/1851

Meeting notes from a discussion on this:
- We'll start small with only supporting foreign transaction status verification.
  - This is sufficient for the bridge use cases.
  - This will not help us migrate the Hot wallet use case.
  - Hot bridge should be able to work with this, but it would require significant refactors on their end.
- Supported RPC providers should be configured in the MPC contract.
  - We'll require a threshold number of votes to add a new RPC provider.
  - Nodes, not operators, will vote for the RPC providers as soon as they see a proposal they have configured API keys for.
- Each MPC node will call a single RPC provider, determined using consistent hashing similar to how we do leader election.

**Use case: Omnibridge**
See this quote from Bowen - this feature is key to allow using the MPC network to move assets from other chains to near.

> Chain Signatures is used in Omnibridge starting from Day 1. Near → Foreign Chain always uses chain signatures, whether the destination chain is Bitcoin, Zcash, Solana, Ethereum, etc. The other direction (foreign chain to Near) uses a variety of proving mechanisms including light clients and wormhole. However, we are also working on migrating that entirely to chain signatures.

PR #1922, opened 2026-01-30
docs: Draft Proposal: Separating Oracle and MPC Networks

kevindeforth:
A proposal for #1918 on how to keep oracle and MPC network separate. Alternative to #1920 and #1917

A key concern with  the previously discussed designs is that the MPC network is homogeneous by definition, while oracle networks are inherently heterogeneous.
Oracle functionality varies significantly across chains and may evolve independently over time. Attempting to handle both responsibilities within a single binary is likely to introduce complexity and operational risk. We would like the MPC network to remain functional, even if some oracle functionality is broken due to unforeseen events.
Separating the functionalities into different binaries and contracts _will_ give us those guarantees.

Proposal:
MPC nodes may still participate in oracle verification, but oracle and MPC functionality should be implemented as separate binaries with independent governance and release cycles.
This separation allows the oracle and MPC networks to evolve independently, paying tribute to their inherently different natures.

Benefits:
- **Clear separation of concerns:**
 We keep the core responsibility of the MPC network (signing signatures) separate from oracle functionality. Changes or failures in RPC providers or chain-specific verification logic would not require urgent MPC releases and would instead be confined to the oracle layer. This ensures signing functionality will not be unnecessarily blocked due to issues in the oracle layer.
- While the initial oracle implementation may rely on RPC queries, the architecture should not constrain future improvements. In particular, it should allow for the adoption of light clients running inside TEEs to verify foreign-chain transactions directly.
- Different chains require different trust models. For some chains, a single authenticated RPC response may be sufficient, while others may require cross-checking multiple providers or clients. A dedicated oracle network enables these strategies to be defined and evolved independently on a per-chain basis.

### DIFF
--- a/docs/foreign_chain_transactions.md
+++ b/docs/foreign_chain_transactions.md
@@ -1,0 +1,59 @@
+```mermaid
+---
+title: Attestation-Gated Signing Architecture - System Context
+---
+flowchart TD
+
+    subgraph OffChain["Off Chain"]
+        RPC["**RPC Providers / Light Clients**"]
+        FC["**Foreign Chain**"]
+        RPC -->|"4. read state"| FC
+    end
+
+    %% Oracle layer
+    subgraph OracleLayer["Oracle Network"]
+        OC["**Oracle Contract**
+            _Attest foreign chain state.
+             Handles requirements for Oracle Nodes._"]
+        
+        ON["**Oracle Nodes**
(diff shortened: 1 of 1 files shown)


netrome:
Still an interesting idea but closing this for now and we can reopen when we want to revisit this discussion.

PR #1925, opened 2026-01-31
docs: Post-discussion updates for foreign chain transaction design doc

### LINKED ISSUE #1924: Initial round of updates to foreign chain transaction design doc
### Background

In https://github.com/near/mpc/pull/1920 we merged an initial draft of the foreign chain transaction design doc. This got some good feedback and ideas (some on slack, some in the PR), that we should incorporate as a next update.

1. Make DomainId required.
2. Enforce strict domain separation between foreign chain transactions and signature generation.
3. (optional) add an extra tweak to foreign chain transaction key derivation.
4. Better naming (`ForeignTransactionId` and `ForeignBlockId`).
5. RPC base URLs should be visible in the contract.
6. Other small fixes.

### User Story

As a developer I want to maintain an up-to-date design doc.

### Acceptance Criteria

We've made the above updates to the foreign chain transaction design doc.

Note: We need to redesign this feature a bit to accomodate the bridge use case (just pure tx payloads isn't sufficient). I'll do that as a separate PR.

netrome:
Closes #1924 

This PR incorporates review feedback and conclusions from various discussions into the design doc keeping it up-to-date with the latest design ideas.

Note: We have a pretty big design change ahead of us as we found that just verifying transaction existence isn't sufficient for the bridge use case. I added a note about this, I'll follow up with a design proposal to handle this and we have a meeting on Monday to discuss it.

### DIFF
--- a/docs/foreign_chain_transactions.md
+++ b/docs/foreign_chain_transactions.md
@@ -69,16 +69,13 @@ respond_verify_foreign_tx({ request, response }) // Respond method for signers
 ```rust
 // Contract DTOs
 pub struct VerifyForeignTxRequestArgs {
-    pub chain: ForeignChain,
-    pub tx_id: ForeignTransactionId, // TxID is the payload we're signing
+    pub request: ForeignChainRpcRequest,
     pub path: String, // Key derivation path
     pub domain_id: DomainId,
 }
 
 pub struct VerifyForeignTxRequest {
-    // Constructed from the args
-    pub chain: ForeignChainRpcRequest,
-    pub tx_id: ForeignTransactionId,
+    pub request: ForeignChainRpcRequest,
     pub tweak: Tweak,
     pub domain_id: DomainId,
 }
@@ -95,12 +92,12 @@ pub enum ForeignChainRpcRequest {
 }
 
 pub struct SolanaRpcRequest {
-    pub tx_id: SolanaTxId,
+    pub tx_id: SolanaTxId, // This is the payload we're signing
     pub finality: Finality, // Optimistic or Final
 }
 
 pub struct BitcoinRpcRequest {
-    pub tx_id: BitcoinTxId,
+    pub tx_id: BitcoinTxId, // This is the payload we're signing
     pub confirmations: usize, // required confirmations before considering final
 }
 


docs/foreign_chain_transactions.md
```diff
@@ -70,21 +70,21 @@ respond_verify_foreign_tx({ request, response }) // Respond method for signers
 // Contract DTOs
 pub struct VerifyForeignTxRequestArgs {
     pub chain: ForeignChain,
-    pub tx_id: TransactionId, // TxID is the payload we're signing
+    pub tx_id: ForeignTransactionId, // TxID is the payload we're signing
     pub path: String, // Key derivation path
-    pub domain_id: Option<DomainId>, // Defaults to 0 (legacy ECDSA)
+    pub domain_id: DomainId,
```
DSharifi: Can't we remove this `domain_id` field in the API args? The contract can use the default/latest `DomainId`, and return the public key/domain id used in the signature?

I think it will make the API simpler.

netrome: Wouldn't that risk breaking downstream applications? If we switch domain, we'd switch the public key used for verification and other contract may have hard-coded this key already.

DSharifi: We can return the public key used as part of the response

DSharifi: https://github.com/near/mpc/blob/9bc04d466df3b4ea4b1cc3059c77826e1bb4f817/crates/contract-interface/src/types/foreign_chain.rs#L30-L38

docs/foreign_chain_transactions.md
```diff
@@ -110,6 +110,32 @@ pub enum Finality{
...
+Requests are only accepted for domains matching the purpose:
+- `sign()` may only target domains with purpose `Sign`.
+- `verify_foreign_transaction()` may only target domains with purpose `ForeignTx`.
+
+```rust
+pub enum DomainPurpose {
+    Sign,
+    ForeignTx,
+    CKD,
+}
```
gilcu3: Yeah I think this is good long-term, so that we are no longer tied one curve <-> one purpose/application as we are now

gilcu3 (dismissed):
LGTM

docs/foreign_chain_transactions.md
gilcu3: I thought this was about including the `domain_id` in the request, wasn't it?

PR #1983, opened 2026-02-05
fix(network): TCP Listener task must not die

### LINKED ISSUE #1979: TCP Listener task can silently die
### Background

We recently observed a mainnet incident where nodes were unable to establish an outgoing connection with two nodes.

Compare the tasks debug page from the two faulty nodes with a healthy node:
faulty node 1:

Image

faulty node 2:

Image

healthy node:

Image

noteworthy is the absence of `incoming connection handler tls mesh network` task in the faulty nodes.

The likely culprit is this line here in the listener task:
https://github.com/near/mpc/blob/3abb324b67813dd5caf5aecf3ae7d96ab5c5ff94/crates/node/src/p2p.rs#L379

if `tcp_listener.accept()` returns an error, we silently fail.

### User Story

As a dev, I like a good network implementation

### Acceptance Criteria

if `tcp_listener.accept()` returns an error, we should not silently fail.
We should log the error and re-establish the outgoing network.

https://nearone.slack.com/archives/C07UW93JVQ8/p1770272891921099

kevindeforth:
resolves #1979 

follow-up issue for test: https://github.com/near/mpc/issues/1984

### DIFF
--- a/nix/cargo-tools.nix
+++ b/nix/cargo-tools.nix
@@ -58,6 +58,13 @@ in
     cargoHash = "sha256-oO5Kp5A2v1w6EUwgcHhyagZDIK7a/2d9uTiCoXHuHhY=";
   })
 
+  (buildTool {
+    pname = "git-cliff";
+    version = "2.12.0";
+    hash = "sha256-V/dLd4yzMLOhMyoV5j/jKSGi0ZSYRFcZuUwlyz+Q3fk=";
+    cargoHash = "sha256-8OhCb2b22S7/UaUAeaRUg0+haRIQ9+1m8eEiItMVTR4=";
+  })
+
   # --- STANDARD NIXPKGS VERSIONS ---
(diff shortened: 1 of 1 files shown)


crates/node/src/p2p.rs
```diff
@@ -496,141 +498,30 @@ pub async fn new_tls_mesh_network(
...
-                                received_bytes, peer_id
-                            ));
-                        }
-                    },
-                );
+                        );
+                    }
+                    Err(err) => {
+                        tracing::error!("error accepting tcp stream: {}", err);
+                        tokio::time::sleep(TLS_ACCEPT_TIMEOUT_ON_ERROR).await;
```
gilcu3: I am not sure we need a timeout here. The line 
```
match tcp_listener.accept().await
```
should already avoid the hot loop no?

netrome (dismissed):

PR #1990, opened 2026-02-05
feat(dtos): add Participants JSON serialization types to contract-interface

### LINKED ISSUE #2001: Move contract interface types (ThresholdParameters, protocol states) to contract-interface crate for Participants BTreeMap migration
### Background

PR [#1861](https://github.com/near/mpc/pull/1861) changes `Participants` from `Vec<(AccountId, ParticipantId, ParticipantInfo)>` to `BTreeMap<AccountId, ParticipantData>` for O(log n) lookups. To maintain JSON backward compatibility, we need to introduce DTO serialization types (`ParticipantId`, `ParticipantInfo`, `ParticipantData`, `ParticipantsJson`, `ParticipantsField`, `ParticipantsJsonDeserialize`) in the `contract-interface` crate.

As identified in [Kevin's review](https://github.com/near/mpc/pull/1990#pullrequestreview-3762663339), we need to:

1. **Use the DTO types in the actual contract interface** — the contract should no longer serialize `Participants` directly; instead, it should convert to the DTO type via an `.into_dto_type()` (or similar) method.
2. **Move additional types to `contract-interface`** — `Participants` is embedded in `ThresholdParameters`, which appears in the protocol contract states (`Running`, `Initializing`, `Resharing`), which are returned by the `state()` view function. These types must also be represented in the `contract-interface` crate to ensure backward compatibility is enforced at the DTO boundary.

This is significant but mostly mechanical/boilerplate work — defining DTO mirrors of the internal types and wiring up conversions.

### User Story

As a contract developer, I need all types that appear in contract method inputs/outputs to be defined in the `contract-interface` crate, so that internal type changes (like the `Participants` Vec→BTreeMap migration in #1861) cannot accidentally break the JSON API.

### Acceptance Criteria

- [ ] The `Participants` DTO types from #1990 are used in the contract's public interface (not the internal `Participants` struct)
- [ ] An `.into_dto_type()` (or `From`/`Into` impl) conversion exists from internal `Participants` to the DTO type
- [ ] `ThresholdParameters` (or its DTO equivalent) is defined in `contract-interface`
- [ ] Protocol state types (`Running`, `Initializing`, `Resharing`) that are returned by the `state()` view function have DTO representations in `contract-interface`
- [ ] The `state()` view function serializes using the DTO types, not the internal types
- [ ] Existing JSON output format is preserved (backward compatible) — validated by unit tests

- Parent PR: [#1861 — feat: change Participants from Vec to BTreeMap](https://github.com/near/mpc/pull/1861)
- DTO types PR: [#1990 — feat(dtos): add Participants JSON serialization types to contract-interface](https://github.com/near/mpc/pull/1990)
- Kevin's review comment with full context on the `contract-interface` crate purpose and required work: [#1990 review](https://github.com/near/mpc/pull/1990#pullrequestreview-2934855244)
- The `contract-interface` crate's purpose: ensure backward compatibility by having all contract I/O types defined there, decoupled from internal types
- This is largely boilerplate work and a good candidate for LLM-assisted development (per Kevin's note)

pbeza:
## Summary

Fixes #2001.

This PR addresses the code review feedback on https://github.com/near/mpc/pull/1861#discussion_r2769051413 by moving the JSON serialization helpers from the primitives module to the `contract-interface` (DTOs) crate.

## Context

PR #1861 changes `Participants` from `Vec<(AccountId, ParticipantId, ParticipantInfo)>` to `BTreeMap<AccountId, ParticipantData>` for O(log n) lookups. To maintain JSON backward compatibility, it added serialization helpers directly in the primitives module. We’re now moving them to the DTOs crate.

## Next Steps

After this PR merges:
1. Rebase #1861 on `main`
2. Update `participants.rs` to import these types from `contract_interface::types`
3. Remove the local helper definitions

### DIFF
--- a/crates/contract/src/dto_mapping.rs
+++ b/crates/contract/src/dto_mapping.rs
@@ -746,7 +746,7 @@ impl IntoInterfaceType<dtos::ThresholdParametersVotes> for &ThresholdParametersV
     fn into_dto_type(self) -> dtos::ThresholdParametersVotes {
         dtos::ThresholdParametersVotes {
             proposal_by_account: self
-                .proposal_by_account()
+                .proposal_by_account
                 .iter()
                 .map(|(account, params)| (account.into_dto_type(), params.into_dto_type()))
                 .collect(),
@@ -758,7 +758,7 @@ impl IntoInterfaceType<dtos::AddDomainsVotes> for &AddDomainsVotes {
     fn into_dto_type(self) -> dtos::AddDomainsVotes {
         dtos::AddDomainsVotes {
             proposal_by_account: self
-                .proposal_by_account()
+                .proposal_by_account
                 .iter()
                 .map(|(participant, domains)| {
                     (
--- a/crates/contract/src/primitives/domain.rs
+++ b/crates/contract/src/primitives/domain.rs
@@ -168,15 +168,10 @@ impl DomainRegistry {
 #[near(serializers=[borsh, json])]
 #[derive(Debug, Clone, Default, PartialEq, Eq)]
 pub struct AddDomainsVotes {
-    proposal_by_account: BTreeMap<AuthenticatedParticipantId, Vec<DomainConfig>>,
+    pub(crate) proposal_by_account: BTreeMap<AuthenticatedParticipantId, Vec<DomainConfig>>,
 }
 
 impl AddDomainsVotes {
-    /// Returns a reference to the underlying map of votes.
-    pub fn proposal_by_account(&self) -> &BTreeMap<AuthenticatedParticipantId, Vec<DomainConfig>> {
-        &self.proposal_by_account
-    }
-
     /// Votes for the proposal, returning the total number of voters so far who
     /// have proposed the exact same domains to add.
     /// If the participant had voted already, this replaces the existing vote.
--- a/crates/contract/src/primitives/votes.rs
+++ b/crates/contract/src/primitives/votes.rs
@@ -8,15 +8,10 @@ use std::collections::BTreeMap;
 #[near(serializers=[borsh, json])]
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
 pub struct ThresholdParametersVotes {
-    proposal_by_account: BTreeMap<AuthenticatedAccountId, ThresholdParameters>,
+    pub(crate) proposal_by_account: BTreeMap<AuthenticatedAccountId, ThresholdParameters>,
 }
 
 impl ThresholdParametersVotes {
-    /// Returns a reference to the underlying map of votes.
-    pub fn proposal_by_account(&self) -> &BTreeMap<AuthenticatedAccountId, ThresholdParameters> {
-        &self.proposal_by_account
-    }
-
     /// return the number of votes for `proposal` casted by members of `participants`
     pub fn n_votes(&self, proposal: &ThresholdParameters, participants: &Participants) -> u64 {
         self.proposal_by_account


kevindeforth (changes requested):
What I really like about this PR is that we have a unit test for backwards compatibility!

Blockers as of now:
- we should be using the type defined in `contract_interface` in the interface (this is the cumbersome part)
- we should not be introducing anything more than what is strictly needed _right now_.

Some context on the `contract_interface` crate:
The purpose of that crate is to ensure backwards-compatibility (or, at least, make it much easier to spot backwards-breaking changes).
The goal is to have all types that occur as input or output of contract methods defined in that crate.
This allows us to modify node- or contract-internal types, without having to worry if it will break the interface.

In this PR, we want to:
- [x] define a struct in the `contract_interface` crate that will serialize to and deserialize from the same json format as the existing `Participants` struct. This task is mostly achieved, as evidenced by the unit tests. But we should not introduce new types in this  PR (one of the blockers).
- [ ] ensure that we no longer serialize the `Participants` struct in the contract. Instead, we need an `.into_dto_type()` method for the type that is used internally in the contract. It is that method, that will then later be modified in #1861.

The second task is a bit cumbersome in this case, because the `Participants` struct is used in `ThresholdParameters`, which occurs in our protocol contract states, which are returned in the `state` function (which must remain backwards compatible) - IIRC, none of these are yet in the contract_interface crate and might also need to be added.

As mentioned in the other PR, this is significant work, but we were bound to run into this issue sooner or later, so it's definitely worth doing!

Note: I do suspect this is a good task for an LLM. It's mostly boilerplate and the underlying patterns are simple.

crates/contract-interface/src/types/participants.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+/// Helper enum for deserializing both legacy Vec and new Map formats.
+#[derive(Clone, Debug, Eq, PartialEq, Deserialize)]
+#[serde(untagged)]
+pub enum ParticipantsField {
+    /// Legacy format: array of [AccountId, ParticipantId, ParticipantInfo] tuples
+    Vec(Vec<(AccountId, ParticipantId, ParticipantInfo)>),
+    /// New format: map of AccountId -> ParticipantData
+    Map(BTreeMap<AccountId, ParticipantData>),
```
kevindeforth: we don't want to add anything new in this PR, please remove this.

crates/contract/src/lib.rs
```diff
@@ -1595,8 +1595,9 @@ impl MpcContract {
         }
     }
 
-    pub fn state(&self) -> &ProtocolContractState {
-        &self.protocol_state
+    pub fn state(&self) -> contract_interface::types::ProtocolContractState {
```
kevindeforth: nice!

crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
```diff
@@ -2337,6 +2331,26 @@ expression: abi
             }
           }
         },
+        "KeyForDomain2": {
```
kevindeforth: Is this a particularity of abi? Why do we get these suffix-2 fields?

kevindeforth (commented):
Shallow review.
There are still some details that we need to get right, but it looks like a solid improvement.

crates/contract-interface/src/types/state.rs
```diff
@@ -0,0 +1,348 @@
...
+pub struct RunningContractState {
+    pub domains: DomainRegistry,
+    pub keyset: Keyset,
+    pub parameters: ThresholdParameters,
+    pub parameters_votes: ThresholdParametersVotes,
+    pub add_domains_votes: AddDomainsVotes,
+    pub previously_cancelled_resharing_epoch_id: Option<EpochId>,
+}
+
+impl RunningContractState {
```
kevindeforth: idem

crates/contract-interface/src/types/state.rs
```diff
@@ -0,0 +1,348 @@
...
+    derive(schemars::JsonSchema)
+)]
+pub struct ResharingContractState {
+    pub previous_running_state: RunningContractState,
+    pub reshared_keys: Vec<KeyForDomain>,
+    pub resharing_key: KeyEvent,
+    pub cancellation_requests: HashSet<AuthenticatedAccountId>,
+}
+
+impl ResharingContractState {
```
kevindeforth: idem

crates/contract/tests/sandbox/utils/interface.rs
```diff
@@ -60,3 +64,63 @@ impl IntoContractType<near_sdk::PublicKey> for &dtos::PublicKey {
         }
     }
 }
+
+impl IntoContractType<near_sdk::PublicKey> for &dtos::PublicKeyExtended {
```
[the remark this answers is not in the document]
pbeza: I created an issue for it: https://github.com/near/mpc/issues/2060. ✅

crates/contract/tests/sandbox/vote.rs
```diff
@@ -149,19 +167,22 @@ async fn test_cancel_keygen() -> anyhow::Result<()> {
...
+        };
+        assert!(
+            running
+                .keyset
+                .domains
+                .iter()
+                .all(|k| k.domain_id.0 != next_domain_id),
+            "No key should be registered for the cancelled domain"
         );
+        assert_eq!(running.domains.domains.len(), ALL_SIGNATURE_SCHEMES.len());
```
kevindeforth: nit (follow-up): we can add a check for the `public_key` method of the contract.

kevindeforth (approved):
Good work, thank you!

I only have nits, that can be addressed in a follow-up. If subsequent reviewers don't have anything to add, then I would opt for merging this ASAP, as it's a really important and nice improvement!

---

_The following remark is out of scope, but I would like to get the discussion started._

Code ergonomics can be improved in the sandbox tests. This was already an existing issue, but this PR aggravates the despair of the situation.

Specifically, we often see this sort of code pattern:

```rust
let state = get_state(&contract).await;
        let ProtocolContractState::Initializing(ref init) = state else {
            panic!("expected initializing state");
        };
        assert_eq!(init.domains.next_domain_id, next_domain_id + 1);
        let found = init
            .domains
            .domains
            .iter()
            .find(|d| d.id.0 == next_domain_id)
            .unwrap();
        assert_eq!(next_domain_id, found.id.0);
        assert_eq!((*scheme).into_interface_type(), found.scheme);
```

We will probably want a nicer interface for these repeated checks. Maybe something like:
```Rust
let expected_state = ExpectedStateBuilder::new().state(Running).epoch(epoch_id).includes_domain(next_domain_id, scheme).build();
expected_state.assert(&contract).await.unwrap();
```
This would make it significantly easier to review.

Again, I would do this in a follow-up PR, because the exact interface might be subject to debate. Lets merge this.

crates/contract-interface/src/types/participants.rs
```diff
@@ -0,0 +1,81 @@
...
+    Eq,
+    PartialEq,
+    Ord,
+    PartialOrd,
+    Hash,
+    Serialize,
+    Deserialize,
+    BorshSerialize,
+    BorshDeserialize,
+)]
```
gilcu3: nit: can we add Into and From conversions here?

crates/contract-interface/src/types/participants.rs
```diff
@@ -0,0 +1,81 @@
...
+
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[cfg_attr(
+    all(feature = "abi", not(target_arch = "wasm32")),
+    derive(schemars::JsonSchema)
+)]
+pub struct ParticipantsJson {
+    pub next_id: ParticipantId,
+    pub participants: Vec<(AccountId, ParticipantId, ParticipantInfo)>,
+}
```
gilcu3: Is this needed only during the migration? If yes we need to add a comment to make that explicit

crates/contract-interface/src/types/state.rs
```diff
@@ -0,0 +1,309 @@
+//! DTO types for protocol contract state.
+//!
+//! These types mirror the internal contract state types and are used for JSON serialization
+//! in the contract's public API (e.g., the `state()` view function).
```
gilcu3: aren't this types also going to be used in the contract storage? shouldn't they implement Borsh serialization?

crates/contract/src/lib.rs
```diff
@@ -1595,8 +1595,9 @@ impl MpcContract {
...
+        use crate::dto_mapping::IntoInterfaceType;
```
gilcu3: any special reason to use the inline `use` statement?

I think our codebase in general prefers the global:
```
use crate::dto_mapping::IntoInterfaceType as _;
```

crates/contract/tests/sandbox/utils/mpc_contract.rs
```diff
@@ -1,26 +1,24 @@
...
+use mpc_contract::tee::tee_state::NodeId;
 use mpc_primitives::hash::MpcDockerImageHash;
 use near_workspaces::{result::ExecutionFinalResult, Account, Contract};
 
 pub async fn get_state(contract: &Contract) -> ProtocolContractState {
     contract.view("state").await.unwrap().json().unwrap()
 }
 
-pub async fn get_participants(contract: &Contract) -> anyhow::Result<Participants> {
+pub async fn get_participants(contract: &Contract) -> anyhow::Result<ParticipantsJson> {
```
gilcu3: are we planning to work with the `ParticipantsJson` type from now on? I find this a bit weird, I would have assumed that the function would now convert from `ParticipantsJson` to `Participants` and then the callers would remain unchanged.

crates/contract/tests/sandbox/utils/resharing_utils.rs
```diff
@@ -130,12 +147,12 @@ pub async fn vote_reshared(
 pub async fn do_resharing(
     remaining_accounts: &[Account],
     contract: &Contract,
-    new_threshold_parameters: ThresholdParameters,
+    new_threshold_parameters: impl Serialize,
```
gilcu3: same question here

crates/contract-interface/src/types/participants.rs
kevindeforth: Good catch. I was expecting this to be resolved together with  https://github.com/near/mpc/pull/1990#discussion_r2785863282. I will be more careful reviewing next time.

crates/contract/tests/sandbox/vote.rs
```diff
@@ -127,14 +141,18 @@ async fn test_cancel_keygen() -> anyhow::Result<()> {
...
-        assert_eq!(expected_domain_config, found);
+        assert_eq!(init.domains.next_domain_id, next_domain_id + 1);
+        let found = init
+            .domains
+            .domains
+            .iter()
+            .find(|d| d.id.0 == next_domain_id)
+            .unwrap();
+        assert_eq!(next_domain_id, found.id.0);
+        assert_eq!((*scheme).into_interface_type(), found.scheme);
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/pull/1990/commits/39a37e79e698c644db7884492dc115bf5b523bb1

crates/contract/tests/sandbox/vote.rs
pbeza: https://github.com/near/mpc/pull/1990/commits/d4abc006d217fecacc29bc6c91066b418d08bb06

crates/contract-interface/src/types/participants.rs
pbeza: https://github.com/near/mpc/pull/1990/commits/871730704fe0233ee6d26b5c523bdd4dc78b69a4

crates/contract-interface/src/types/participants.rs
```diff
@@ -0,0 +1,81 @@
...
+                },
+            )],
+        };
+
+        let json = serde_json::to_string(&participants_json).unwrap();
+        assert_eq!(
+            json,
+            r#"{"next_id":1,"participants":[["alice.near",0,{"url":"https://alice.com","sign_pk":"ed25519:abc"}]]}"#
+        );
+    }
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/pull/1990/commits/142b662c68a94f232bbcc19ae77bb230ab530c25

crates/contract-interface/src/types/state.rs
gilcu3: that's true, but for pragmatic reasons and to avoid duplicate structs, we are also using them for contract storage. See for example how many types in this crate already implement the Borsh traits

crates/contract-interface/src/types/participants.rs
gilcu3: I am a bit lost about your https://github.com/near/mpc/pull/1990#discussion_r2785863282 @kevindeforth does it mean this struct should be removed from this PR as well?

crates/contract-interface/src/types/state.rs
```diff
@@ -0,0 +1,309 @@
...
+// Re-export DomainId from primitives to avoid duplicate JsonSchema definitions
+// that cause non-deterministic ABI snapshots.
+pub use super::primitives::DomainId;
+
+// =============================================================================
+// Simple Wrapper Types (newtypes)
+// =============================================================================
+
+/// Epoch identifier for key generation/resharing cycles.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/pull/1990/commits/69ec4cdd4ab827aeb5b63c55efd2db0d70913910

crates/contract/src/lib.rs
pbeza: No particular reason — I just overlooked it. Fixed here: https://github.com/near/mpc/pull/1990/commits/3cc0371a283ff56b613ce666abff32fcf5375c7c. ✅

crates/contract/tests/sandbox/tee_cleanup_after_resharing.rs
```diff
@@ -85,27 +85,33 @@ async fn test_tee_cleanup_after_full_resharing_flow() -> Result<()> {
...
-    let new_threshold_parameters =
-        ThresholdParameters::new(new_participants, threshold.clone()).unwrap();
+    let new_threshold_parameters = ThresholdParameters::new(
+        new_participants,
+        mpc_contract::primitives::thresholds::Threshold::new(threshold.0),
+    )
+    .unwrap();
 
-    let prospective_epoch_id = EpochId::new(6);
+    let prospective_epoch_id = contract_interface::types::EpochId(6);
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/pull/1990/commits/c8fb0f527d9249eb3995dddf07552d6fd76d7ea2

crates/contract/src/dto_mapping.rs
```diff
@@ -565,6 +579,297 @@ impl From<contract_interface::types::Config> for Config {
...
+                edwards_point,
+            } => dtos::PublicKeyExtended::Ed25519 {
+                near_public_key_compressed: String::from(near_public_key_compressed),
+                edwards_point: edwards_point.to_bytes(),
+            },
+            PublicKeyExtended::Bls12381 { public_key } => dtos::PublicKeyExtended::Bls12381 {
+                public_key: public_key.clone(),
+            },
+        }
+    }
```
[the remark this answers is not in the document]
pbeza: Not sure I follow — what would you suggest doing here instead?

crates/contract-interface/src/types/participants.rs
pbeza: > I am a bit lost about your https://github.com/near/mpc/pull/1990#discussion_r2785863282 @kevindeforth does it mean this struct should be removed from this PR as well?

Not sure why you’d want to remove it. 👀 Isn’t the whole point of introducing the `Vec`-based `Participants` as a DTO to enable a later migration to the new `BTreeMap`-based version of `Participants`? (cc @kevindeforth)

crates/contract/src/primitives/domain.rs
```diff
@@ -172,6 +172,11 @@ pub struct AddDomainsVotes {
 }
 
 impl AddDomainsVotes {
+    /// Returns a reference to the underlying map of votes.
+    pub fn proposal_by_account(&self) -> &BTreeMap<AuthenticatedParticipantId, Vec<DomainConfig>> {
+        &self.proposal_by_account
+    }
```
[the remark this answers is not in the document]
pbeza: > what about if the field is pub(crate)? (...)

It offers less encapsulation, but I actually prefer it — less boilerplate getters. Fixed here: https://github.com/near/mpc/pull/1990/commits/36e7b0bc2d030fe4f4375dc740b223418adace5c. ✅ Thanks!

PR #2008, opened 2026-02-06
feat: implement verify foreign key logic in the contract

### LINKED ISSUE #1965: implement verify foreign tx function in the contract
### Background

Currently we have function stubs in the contract for the verify foreign tx feature. We should implemented them before the node can be tested.

### User Story

_No response_

### Acceptance Criteria

The functions are implemented and tested in the contract. This includes the implementation of the tweak derivation as explained in the design doc

_No response_

gilcu3:
Closes #1965

### DIFF
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -2299,13 +2299,13 @@ mod tests {
     }
 
     #[test]
-    fn test_signature_simple() {
+    fn respond__should_succeed_when_response_is_valid_and_request_exists() {
         test_signature_common(true, false);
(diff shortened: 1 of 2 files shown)


netrome (approved):
Nice stuff. Just a testing opinion, pushing for the new convention. I'd find it much easier to read in this case, because the current test name doesn't really say anything and readers are therefore forced to read the code to understand what the test is supposed to test. I know this mirrors existing code so not a blocker ofc.

kevindeforth (approved):
Thank You!

PR #2014, opened 2026-02-09
fix: broken reproducibility

### LINKED ISSUE #1932: Investigate broken reproducibility
### Background

It seems like we cannot reproduce the most recent releases locally. For example this is my output when trying to reproduce the 3.4.0 gcp node image:

```
#12 exporting to docker image format
#12 sending tarball 5.6s done
#12 DONE 12.7s
commit hash: f2504d4a2eb6de21798e8aa3abaeba19c4fd9167
SOURCE_DATE_EPOCH used: 0
node binary hash: 9d0735498b5615dbda377237784fce558fe6c6033b4d3d0c4c9a8098fe69f03e
node gcp docker image hash: sha256:c6c327b733aae6d8f7fdd53ea0bad6f5a467a242214d0fe1a3f7328edb151284
```

### User Story

As an MPC operator I'd like to be able to verify the binaries I'm running

### Acceptance Criteria

We've identified why our builds aren't reproducible anymore.

_No response_

gilcu3:
Closes #1932 

The cluprit seems to have been a mix of `jemalloc` and the `built` crate. The jemalloc part was fixed in https://github.com/near/mpc/pull/1935. Would be nice to add it back

I got repro with this branch, locally:
```
commit hash: d56d8f112796d582aaf39923bce55f184b0cca70
SOURCE_DATE_EPOCH used: 0
node binary hash: 7bcec996bac31018387d166397815a364d07416a1dfbe703e39a667b6d2adc56
node docker image hash: sha256:24f18b462ac770a3701f84a34eff8fc4470abf15c4bb2291ed1537d9e470fd5e
```
and in CI https://github.com/near/mpc/actions/runs/21820809884/job/62953357578

added also a nit, as we are using/testing claude locally we should ignore the local config folder

This is the TL;DR of the fix:

The built crate generates ~20 constants via include!(), and we only use 4 of them. The unused ones still end up in the binary (they were pub const in a pub mod), and some would differ between local and CI (why this was not happening before is a mystery, but hopefully now we are more robust):

  - GIT_HEAD_REF: Locally Some("refs/heads/main"), in CI (detached HEAD from actions/checkout) it's None. Different string lengths → different .rodata layout.

  These unused but still-present constants apparently shifted the positions of stuff in .rodata., obtained with:
```bash
readelf --wide --decompress --hex-dump=.rodata target/reproducible/mpc-node
```


DSharifi (approved):

barakeinav1 (approved):
thank you!

PR #2021, opened 2026-02-09
docs: design hot migration

### LINKED ISSUE #2018: Design HOT migration
### Background

HOT want to migrate their keyshares to NEAR's MPC network.
We need to figure out how to do that securely.

### User Story

As a dev, I like a good plan.

### Acceptance Criteria

Write a markdown file outlining how to migrate the HOT keyshare to the MPC network

https://docs.nearone.org/doc/hot-migration-EzdSbjWjiW

kevindeforth:
resolves #2018

### DIFF
--- a/docs/hot_migration.md
+++ b/docs/hot_migration.md
@@ -29,7 +29,24 @@ _- Lets schedule an audit to assess the safety of this and ensure we don't accid
 
 ### Key derivation
 
-At this point, it is unclear what exact key derivation mechanism HOT uses. We need to account for the possibility that it might be different to ours and that signature payloads might be different.
+HOT seems to be using the same key derivation method as NEAR, so we don't expect any difficulties on that end (with the exception that their `tweak` corresponds to a uid relatedto the `walled_id` discussed above).
+
+HOT:
+```rust
+/// Derives a public key from a tweak and a master public key by computing PK + [tweak] G
+pub fn derive_public_key(public_key: AffinePoint, tweak: Scalar) -> AffinePoint {
+    (AffinePoint::GENERATOR * tweak + public_key).to_affine()
+}
+```
+
+us:
+```rust
+/// Derives the verifying key as X + tweak . G
+pub fn derive_verifying_key(&self, public_key: &VerifyingKey<C>) -> VerifyingKey<C> {
+    let derived_share = public_key.to_element() + C::Group::generator() * self.value();
+    VerifyingKey::new(derived_share)
+}
+```
 
 ### Keyshare Import
 
@@ -54,7 +71,7 @@ We don't expect to be migrating any other keyshares. We aim for a design that ca
 
 ### Contract Implementation
 
-#### State
+#### MpcContract changes
 
 The contract will need extra state.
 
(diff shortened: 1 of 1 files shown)


docs/hot_migration.md
```diff
@@ -0,0 +1,148 @@
...
+
+
+Option 1:
+In the coordinator, we add a new contract state `ConfirmKeyMigration`, during which the nodes resolve a simple challenge to check whether they can compute a result.
+
+Option 2:
+Node continues to monitor the `HotDomainsInfo` in a separate thread. Once it sees that all partiicpants voted, they resolve a challenge individually.
+
+Option 3:
+We immediately entere a resharing state and do a key refresh
```
netrome: What's the problem with not having a challenge and just using the sign function to verify that it works after the migration?

netrome (commented):
Some early feedback from my side.

pbeza (commented):
What do we want to do with this one-time code after successfully importing the HOT key shares? Do we just remove it, or could we reuse parts of it in the test framework—if there’s a testing scenario where manually importing secret shares would be useful? Not sure if that makes sense though; maybe this is already easy to do, or simply not needed for any test scenarios.

docs/hot_migration.md
kevindeforth: Because by then, those keyshares will be part of the keyset and have been loaded into permanent keyshare storage.
We don't support removing or modifying individual keyshares in the permanent keyshare storage.

netrome (dismissed):
Thanks for updating 🙏

docs/hot_migration.md
```diff
@@ -0,0 +1,296 @@
...
+        self = HotDomainMigrationState::PrepareImport(...)
+    }
+}
+
+```
+
+
+### Node Implementation
+
+We can add a `MigratingHotState` in the `ContractRunningState` of the nodes [indexer](https://github.com/near/mpc/blob/f1b0d197ea29767acc6ed4631f35e2cc4901b61e/crates/node/src/indexer/participants.rs#L106C12-L106C32). We spawn a separate task for monitoring the filesysem, importing keyshares and engaging with the system.
```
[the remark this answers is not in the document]
kevindeforth: Where would you spawn this task?

docs/hot_migration.md
kevindeforth: >  If something is wrong, it should be possible to write a script to do some surgery to change these as needed.

Can you elaborate on this?

docs/hot_migration.md
```diff
@@ -0,0 +1,296 @@
...
+We will need to import keyshares. We will assume that HOT can provide one valid keyshare for each of our production nodes and that the threshold matches our production threshold. Import will be coordinated by the contract and it will consist of thee stages:
+
+1. Import preparation:
+    - Nodes monitor the file system for the existence of keyshares from HOT in a specific directory.
+    - Nodes query the contract to get the current `KeyEventId` for this import.
+    - Nodes validate and write the HOT keyshares to their temporary keyshare storage (under the `keyEventId` handle).
+    - Nodes confirm to the contract that they prepared the key import.
+2. Keyshare validation:
+    - Nodes test if they can generate a valid signature request for the newly imported keyshares.
+    - Nodes test if they can run a key refresh with the newly imported keyshares.
```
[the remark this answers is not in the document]
netrome: > but if we import the keyshares and then the resharing fails, we end up in a very hard-to-debug state

How would this be hard to debug? First, we should try a signature generation before resharing. If the nodes can generate valid signatures, they have valid shares and therefore resharing should succeed. I'd be very curious to learn if it's even theoretically possible to have a state in which this isn't true.

If signature generation fails, we know the keyshares are wrong and need to check that each operator loaded the right shares.

PR #2023, opened 2026-02-09
test: System test for foreign chain policy voting

### LINKED ISSUE #2022: Automatic foreign chain policy voting system test
### Background

In https://github.com/near/mpc/pull/1997 we've implemented the automatic foreign chain policy voting feature. While the PR itself is well-tested with unit and integration tests, there's still a small risk of mismatch between what transaction the node sends (e.g. the method name is a hard-coded string literal) and what the contract expects.

Therefore, we should complement these with a system test (pytest) that spins up the whole system with the expected configuration and verifies these votes are passed.

### User Story

As a developer I want to ensure our features are tested end to end.

### Acceptance Criteria

We have a pytest that verifies that nodes can be configured with foreign chain policies, and votes for these successfully.

_No response_

netrome:
Closes #2022

### DIFF
--- a/pytest/tests/test_foreign_chain_policy_auto_voting.py
+++ b/pytest/tests/test_foreign_chain_policy_auto_voting.py
@@ -48,7 +48,7 @@ def _normalize_policy(policy: dict[str, Any]) -> list[tuple[str, tuple[str, ...]
     normalized = []
     for chain_cfg in chains:
         chain_name = chain_cfg["chain"]
-        providers = tuple(sorted(p["rpc_url"] for p in chain_cfg.get("providers", [])))
+        providers = tuple(sorted(p["rpc_url"] for p in chain_cfg["providers"]))
         normalized.append((chain_name, providers))
     return sorted(normalized)
 


pytest/common_lib/shared/mpc_cluster.py
```diff
@@ -354,6 +354,24 @@ def do_resharing(
...
+        if args is None:
+            args = {}
+        encoded_args = base64.b64encode(json.dumps(args).encode("utf-8")).decode(
+            "utf-8"
+        )
+        res = self.contract_node.near_node.call_function(
+            self.mpc_contract_account(),
+            function_name,
+            encoded_args,
+            timeout=10,
```
SimonRastikian: How is this value decided?

netrome: Updated in https://github.com/near/mpc/pull/2023/commits/593bccd6b54a85f4257b742f45ad8e29908cd82d

pytest/common_lib/shared/mpc_cluster.py
```diff
@@ -354,6 +354,24 @@ def do_resharing(
             )
             self.update_participant_status()
 
+    def view_contract_function(
+        self, function_name: str, args: dict[str, Any] | None = None
+    ) -> Any:
+        if args is None:
```
gilcu3: isnt this more or less what we do with the function(s) below?

pytest/tests/test_foreign_chain_policy_auto_voting.py
```diff
@@ -0,0 +1,172 @@
...
+    last_error = None
+    while time.monotonic() < deadline:
+        try:
+            if predicate():
+                return
+        except Exception as err:
+            last_error = err
+        time.sleep(poll_interval_sec)
+
+    raise AssertionError(f"timed out waiting for {description}") from last_error
```
gilcu3: this seems very general. Can we put it somewhere accessible for all tests?

pytest/common_lib/shared/mpc_cluster.py
netrome: This is a view-only call, we should be able to migrate some of the other calls to use this one. I was surprised to find we didn't use view/read-only calls already. I think this will make our tests a bit faster but should be done as a follow-up imo.

PR #2027, opened 2026-02-10
chore: remove usage of `near_o11` for metrics and test logger

### LINKED ISSUE #1908: Remove `near-o11y` nearcore metric dep wrapper
### Background

We depend on nearcore's prometheus wrapper. It's only used in a few metrics and should be removed such that we use the prometheus crate directly as in our all other metrics.

https://github.com/near/mpc/blob/59bb8669f6bd94933d4adb8858fbbe1c0053810f/crates/node/src/metrics.rs#L17-L48

### User Story

_No response_

### Acceptance Criteria

remove `near-o11y` from the workspace and use prometheus directly for the metrics that rely on it.

_No response_

DSharifi:
part of #1908 

This PR doesn't fully close #1908 since `near_o11y`' since the `.send_async()` implementation of the nearcore inner client expects a SpanWrap<Status> type, which is defined in `near_o11y`, meaning we can't fully remove this dependency yet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2442,9 +2442,9 @@ dependencies = [
 
 [[package]]
 name = "deranged"
-version = "0.5.5"
+version = "0.5.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ececcb659e7ba858fb4f10388c250a7252eb0a27373f1a72b8748afdd248e587"
+checksum = "cc3dc5ad92c2e2d1c193bbbbdf2ea477cb81331de4f3103f267ca18368b988c4"
 dependencies = [
  "powerfmt",
  "serde_core",
@@ -10334,7 +10334,7 @@ checksum = "be35209fd0781c5401458ab66e4f98accf63553e8fae7425503e92fdd319783b"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.108",
+ "syn 2.0.114",
 ]
 
 [[package]]


crates/node/src/network.rs
```diff
@@ -1147,15 +1147,14 @@ mod fault_handling_tests {
...
     use crate::tests::into_participant_ids;
     use crate::tracking::testing::start_root_task_with_periodic_dump;
-    use near_o11y::testonly::init_integration_logger;
     use std::sync::Arc;
     use threshold_signatures::test_utils::TestGenerators;
     use tokio::sync::mpsc;
     use tokio_util::sync::CancellationToken;
 
     #[tokio::test]
+    #[test_log::test]
```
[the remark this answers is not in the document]
DSharifi: From their docs, that variant you're suggesting is a fix if it's chained with test macros that don't play nicely with test_log.

pbeza (approved):

SimonRastikian (approved):

crates/node/src/indexer.rs
```diff
@@ -429,8 +428,12 @@ impl IndexerClient {
                 is_health_check: false,
                 detailed: false,
             };
-
-            let status_response = self.client.send_async(status_request.span_wrap()).await;
+            let status_response = self
+                .client
+                .send_async(
+                    near_o11y::span_wrapped_msg::SpanWrappedMessageExt::span_wrap(status_request),
+                )
```
gilcu3: nit: if this is the only remaining place we use near_o11y, probably a todo list is appropriate

gilcu3 (approved):
Thank you!

crates/node/src/indexer.rs
DSharifi: I thougt about it. But it's irremovable unless nearcore removes it from the API of the indexer.

And if they do remove it, then `near_o11y` should be detected as unused by cargo shear.

PR #2041, opened 2026-02-10
ci: remove `cargo-about` third-party licenses check from CI

### LINKED ISSUE #2040: remove `cargo-about` from CI check
### Background

`cargo-about` update is quite noisy in PRs every time a change to the lock file occurs. We only care about the license notices for release builds, so we are fine with just having it generated at the release process instead of for every PR.

### User Story

_No response_

### Acceptance Criteria

remove `cargo-about` check in CI

_No response_

DSharifi:
closes #2040

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -33,9 +33,7 @@ dependencies = [
 
 [tasks.check-extra]
 description = "Run extra checks"
-dependencies = [
-  "cargo-deny",
-]
+dependencies = ["cargo-deny"]
 
 [tasks.check]
 command = "cargo"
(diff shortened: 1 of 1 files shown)


RELEASES.md
```diff
@@ -47,9 +47,10 @@ In practice when making a release, you need to do the following things:
 
 1. Update the changelog.
 2. Bump the crate versions.
-3. Create the release tag.
-4. Create release artifacts.
-5. Create the release on GitHub.
+3. Update license versions
```
gilcu3: could you add a sentence somewhere to let people know how to do that

PR #2055, opened 2026-02-11
feat: integrate foreign chain tx feature in the node

### LINKED ISSUE #1960: integrate extractors and indexer changes in the node
### Background

Once #1941 #1943 #1942 are done (possibly a few more as well), the changes need to be integrated and tested in the node.

### User Story

_No response_

### Acceptance Criteria

Verify foreign transaction changes are integrated and tested in the node.

_No response_

gilcu3:
Closes #1960

### DIFF
--- a/crates/node/src/storage.rs
+++ b/crates/node/src/storage.rs
@@ -38,7 +38,7 @@ impl SignRequestStorage {
         true
     }
 
-    /// Blocks until a signature request with given id is present, then returns it.
+    /// Returns when a signature request with given id is present, then returns it.
     /// This behavior is necessary because a peer might initiate computation for a signature
     /// request before our indexer has caught up to the request. We need proof of the request
     /// from on-chain in order to participate in the computation.
@@ -104,7 +104,7 @@ impl CKDRequestStorage {
         true
     }
 
-    /// Blocks until a ckd request with given id is present, then returns it.
+    /// Returns when a ckd request with given id is present, then returns it.
     /// This behavior is necessary because a peer might initiate computation for a ckd
     /// request before our indexer has caught up to the request. We need proof of the request
     /// from on-chain in order to participate in the computation.
@@ -170,7 +170,7 @@ impl VerifyForeignTransactionRequestStorage {
         true
     }
 
-    /// Blocks until a verify foreign tx request with given id is present, then returns it.
+    /// Returns when a verify foreign tx request with given id is present, then returns it.
     /// This behavior is necessary because a peer might initiate computation for a verify foreign tx
     /// request before our indexer has caught up to the request. We need proof of the request
     /// from on-chain in order to participate in the computation.


crates/node/src/db.rs
```diff
@@ -31,16 +32,18 @@ impl DBCol {
             DBCol::Presignature => "presignature",
             DBCol::SignRequest => "sign_request",
             DBCol::CKDRequest => "ckd_request",
+            DBCol::VerifyForeignTxRequest => "verify_foreign_tx_request",
```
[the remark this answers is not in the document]
gilcu3: We can give it a try in a follow-up, to see if it is possible to do so while remaining backward compatible.

crates/node/src/requests/metrics.rs
```diff
@@ -26,6 +26,73 @@ pub static MPC_PENDING_VERIFY_FOREIGN_TX_QUEUE_ATTEMPTS_GENERATED: LazyLock<
...
+                 indexed by the pending verify foreign tx queue"
+    )
+    .unwrap()
+});
+pub static VERIFY_FOREIGN_TXS_REQUEST_RESPONSE_LATENCY_BLOCKS: LazyLock<prometheus::Histogram> =
+    LazyLock::new(|| {
+        // High resolution for 1-9 blocks
+        let mut buckets = linear_buckets(1.0, 1.0, 9).unwrap();
+        // 10, 15, 22.5, 33.75, 50, ..., 256, 384
+        let exponential_buckets = exponential_buckets(10.0, 1.5, 91).unwrap();
```
DSharifi: 91 buckets is  overkill. Did you mean 9 buckets here?

crates/node/src/requests/metrics.rs
```diff
@@ -8,6 +8,91 @@ pub static MPC_PENDING_CKDS_QUEUE_SIZE: LazyLock<prometheus::IntGauge> = LazyLoc
...
+pub static MPC_PENDING_VERIFY_FOREIGN_TXS_QUEUE_MATCHING_RESPONSES_INDEXED: LazyLock<
+    prometheus::IntCounter,
+> = LazyLock::new(|| {
+    prometheus::register_int_counter!(
+        "mpc_pending_verify_foreign_txs_queue_matching_responses_indexed",
+        "Number of verify foreign tx responses that match previously indexed verify foreign tx requests,
+                 indexed by the pending verify foreign tx queue"
+    )
+    .unwrap()
+});
```
DSharifi: As a best practice, all the counters should have `_total` suffix.

Example for this one:
```suggestion
pub static MPC_PENDING_VERIFY_FOREIGN_TXS_QUEUE_MATCHING_RESPONSES_INDEXED_TOTAL: LazyLock<
    prometheus::IntCounter,
> = LazyLock::new(|| {
    prometheus::register_int_counter!(
        "mpc_pending_verify_foreign_txs_queue_matching_responses_indexed_total",
        "Number of verify foreign tx responses that match previously indexed verify foreign tx requests,
                 indexed by the pending verify foreign tx queue"
    )
    .unwrap()
});
```

crates/node/src/requests/metrics.rs
```diff
@@ -8,6 +8,91 @@ pub static MPC_PENDING_CKDS_QUEUE_SIZE: LazyLock<prometheus::IntGauge> = LazyLoc
...
+pub static VERIFY_FOREIGN_TXS_REQUEST_RESPONSE_LATENCY_BLOCKS: LazyLock<prometheus::Histogram> =
+    LazyLock::new(|| {
+        // High resolution for 1-9 blocks
+        let mut buckets = linear_buckets(1.0, 1.0, 9).unwrap();
+        // 10, 15, 22.5, 33.75, 50, ..., 256, 384
+        let exponential_buckets = exponential_buckets(10.0, 1.5, 91).unwrap();
+        buckets.extend(exponential_buckets);
+
+        prometheus::register_histogram!(
+            "mpc_verify_foreign_tx_request_response_latency_blocks",
```
DSharifi:
```suggestion
            "mpc_verify_foreign_txs_request_response_latency_blocks",
```

crates/node/src/storage.rs
```diff
@@ -137,6 +137,79 @@ impl CKDRequestStorage {
...
+    }
+
+    /// If given request is already in the database, returns false.
+    /// Otherwise, inserts the request and returns true.
+    pub fn add(&self, request: &VerifyForeignTxRequest) -> bool {
+        let key = borsh::to_vec(&request.id).unwrap();
+        if self
+            .db
+            .get(DBCol::VerifyForeignTxRequest, &key)
+            .expect("Unrecoverable error reading from database")
```
DSharifi: two nits:
1. Why not make this method fallible, instead of panicking?
2. Expect message should contain _why_ we expect this to not fail, not what failed.

crates/node/src/mpc_client.rs
```diff
@@ -500,12 +553,103 @@ impl MpcClient {
...
+                                    .inc();
+
+                                let response = match this
+                                    .domain_to_scheme
+                                    .get(&verify_foreign_tx_attempt.request.domain_id)
+                                {
+                                    Some(SignatureScheme::Secp256k1) => {
+                                        let response = timeout(
+                                            Duration::from_secs(this.config.signature.timeout_sec),
+                                            this.verify_foreign_tx_provider.clone().make_signature(
```
DSharifi: In a follow up, or after the MVP we should detach the `make_signature` call in a new task, to avoid sequentially doing one RPC request after another before processing pending requests.

DSharifi (changes requested):
Looks good!

Some Blockers :
- Please take a look at my comments, especially for the metrics file
- Some leftover comments
- Use constants for metric labels

crates/node/src/db.rs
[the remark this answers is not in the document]
netrome: Yeah I don't think this is trivial, so definitely worth looking into as a follow-up

crates/node/src/providers/verify_foreign_tx.rs
```diff
@@ -0,0 +1,117 @@
...
+    async fn run_key_resharing_client(
+        _new_threshold: usize,
+        _key_share: Option<SigningShare>,
+        _public_key: VerifyingKey,
+        _old_participants: &ParticipantsConfig,
+        _channel: NetworkTaskChannel,
+    ) -> anyhow::Result<Self::KeygenOutput> {
+        // this method is never called, as we are re-using the ecdsa signature provider
+        unimplemented!()
+    }
```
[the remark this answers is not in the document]
gilcu3: we are reusing the existing at the moment, so we cannot call it. We could simply return an error instead, this method is only implemented because of the trait restrictions, but that might be removable later.

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -0,0 +1,168 @@
...
+            Duration::from_secs(self.config.signature.timeout_sec),
+            self.verify_foreign_tx_request_store.get(id),
+        )
+        .await??;
+        metrics::MPC_NUM_PASSIVE_SIGN_REQUESTS_LOOKUP_SUCCEEDED.inc();
+
+        let response_payload = self
+            .execute_foreign_chain_request(&foreign_tx_request.request)
+            .await?;
+
```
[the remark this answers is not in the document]
netrome: Definitely. We need to make some bigger refactors here anyway to support consistent hashing as well.

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -0,0 +1,168 @@
...
+                let Some(bitcoin_provider_config) = bitcoin_provider_config else {
+                    anyhow::bail!("found empty list of providers for bitcoin")
+                };
+
+                let public_node_url = bitcoin_provider_config.rpc_url.clone();
+
+                let http_client = foreign_chain_inspector::build_http_client(
+                    public_node_url,
+                    bitcoin_provider_config.auth.clone().try_into()?,
+                )?;
```
netrome: Also important for a follow-up is that we don't build a live http client here but instead take a client constructor as input or something similar to allow running this with mock clients for tests. Right now this is not possible to unit test.

netrome (dismissed):
While I think there's plenty of room for improvement, I don't have any strict blockers from my side (beyond what Daniel already observed).

I think it will be most effective if we merge this relatively soon as-is and we can tackle most improvements in parallel as follow-ups. My main concerns are:

1. The broken forwards compatibility for database columns (kind of unrelated but still).
2. The lack of provider unit tests, partially due to the http client not being mockable.

DSharifi (dismissed):

PR #2075, opened 2026-02-12
feat: add extractor for evm `Log`s 

### LINKED ISSUE #2066: Event extractor for the Abstract blockchain
### Background

As discussed in https://nearone.slack.com/archives/C0AC9JP4Y3E/p1770857443070669

We want to add an extractor for Abstract that finds an event from a log index and returns a hash of it.

The event structure should look like
```rust
pub struct EventPayload {
    pub emitter_address: [u8; 20],
    pub topics: Vec<[u8; 32]>,
    pub data: Vec<u8>,
}
```

The extractor should take tx_hash and log_index as input, extract the appropriate event payload and return a hash of this payload.

### User Story

As a bridge developer I'd like to be able to verify events for Abstract.

### Acceptance Criteria

We have an EventHash extractor for Abstract.

_No response_

DSharifi:
closes #2066

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3277,7 +3277,6 @@ name = "foreign-chain-inspector"
 version = "3.4.1"
 dependencies = [
  "assert_matches",
- "borsh",
  "derive_more 2.1.1",
  "ethereum-types",
  "foreign-chain-rpc-interfaces",
@@ -3296,7 +3295,6 @@ dependencies = [
 name = "foreign-chain-rpc-interfaces"
 version = "3.4.1"
 dependencies = [
- "borsh",
  "derive_more 2.1.1",
  "ethereum-types",
  "jsonrpsee",
--- a/crates/foreign-chain-inspector/Cargo.toml
+++ b/crates/foreign-chain-inspector/Cargo.toml
@@ -5,7 +5,6 @@ edition.workspace = true
 license.workspace = true
 
 [dependencies]
-borsh = { workspace = true }
 derive_more = { workspace = true }
 ethereum-types = { workspace = true }
 foreign-chain-rpc-interfaces = { workspace = true }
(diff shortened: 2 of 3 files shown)


crates/foreign-chain-rpc-interfaces/src/evm.rs
DSharifi: This file is also just a move from `ethereum.rs` with addition of the new `Log` struct

crates/foreign-chain-inspector/src/abstract_chain/inspector.rs
```diff
@@ -1,11 +1,12 @@
...
+use sha2::Digest;
 
 use crate::{
     EthereumFinality, ForeignChainInspectionError, ForeignChainInspector,
-    abstract_chain::{AbstractBlockHash, AbstractTransactionHash},
+    abstract_chain::{AbstractBlockHash, AbstractTransactionHash, LogHash},
 };
 
-use crate::rpc_schema::ethereum::{
+use foreign_chain_rpc_interfaces::evm::{
```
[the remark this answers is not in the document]
gilcu3: > It's mainly for having access to the transport types to create responses with the mock client. See that the duplicate types are now deleted in the integration tests.

I thought we could define the mock client in the same crate, therefore reducing the complexity of this. As this is anyway an internal crate, wasn't that a possibility?

> I don't think it makes sense to make the transport types public part of the inspector crate either, so I think this is a good middle ground.

This is one I don't fully understand. Why these types being public don't make sense?

crates/foreign-chain-rpc-interfaces/src/evm.rs
```diff
@@ -0,0 +1,105 @@
...
+        let topics_len = self.topics.len() as u32;
+        BorshSerialize::serialize(&topics_len, writer)?;
+
+        for topic in &self.topics {
+            BorshSerialize::serialize(&topic.0, writer)?;
+        }
+
+        Ok(())
+    }
+}
```
netrome: Why do we need this to be explicitly serialized?

netrome (changes requested):
Overall looks good to me but the as conversion in the custom borsh serialization method is a blocker imo.

PR #2079, opened 2026-02-13
chore: make API types specific per chain

### LINKED ISSUE #2074: revisit the response format of foreign RPC requests
### Background

In the contract interface type we have defined the response type as a signature, and a payload type which is used to calculate the 32 bytes of payload to sign https://github.com/near/mpc/blob/d84680c06dd5745096beb0efc0d03a6fe380899e/crates/contract-interface/src/types/foreign_chain.rs#L45-L53

The payload type is defined as:
https://github.com/near/mpc/blob/d84680c06dd5745096beb0efc0d03a6fe380899e/crates/contract-interface/src/types/foreign_chain.rs#L568-L572

This means that we return back the original request and the extracted values back to the caller, such that they can derive the payload themselves to verify the signature. However in the use-case we are currently building for, bridge team, they only require the signature, and nothing else.

The reason is that the bridge team don't need any of the extracted values. They already have all these extracted values, and use the MPC network as an oracle to verify that the extracted values they already have are valid. I.e. they can calculate the payload that needs to be signed before they even call the sign function of the MPC contract.

### User Story

_No response_

### Acceptance Criteria

Solidify the design of the external signing feature on whether extracted values need to be part of the response of the sign requests

_No response_

DSharifi:
closes #2074

### DIFF
--- a/crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
+++ b/crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
@@ -1407,6 +1407,22 @@ expression: abi
             }
           }
         },
+        "BitcoinExtractedValue": {
+          "oneOf": [
+            {
+              "type": "object",
+              "required": [
+                "BlockHash"
+              ],
+              "properties": {
+                "BlockHash": {
+                  "$ref": "#/definitions/Hash256"
+                }
+              },
+              "additionalProperties": false
+            }
+          ]
+        },
         "BitcoinExtractor": {
           "type": "string",
           "enum": [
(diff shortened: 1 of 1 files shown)


crates/contract-interface/src/types/foreign_chain.rs
```diff
@@ -257,8 +284,50 @@ pub enum BitcoinExtractor {
...
+    Deserialize,
+    BorshSerialize,
+    BorshDeserialize,
+)]
+#[cfg_attr(
+    all(feature = "abi", not(target_arch = "wasm32")),
+    derive(schemars::JsonSchema)
+)]
+pub enum BitcoinExtractedValue {
+    BlockNumber(u64),
```
netrome: I don't think this one is used yet.

netrome (changes requested):

netrome (changes requested):
We need to update the response types in the design doc as well.

netrome (dismissed):
Thanks!

crates/contract-interface/src/types/foreign_chain.rs
gilcu3: removed

PR #2104, opened 2026-02-16
chore: add `From` and `TryFrom` conversions between dto and foreign chain inspector types

### LINKED ISSUE #2085: create From/TryFrom conversions between contract-interface types and `foreign-inspector-rpc` types
### Background

Currently we need to do painful conversions between contract-interface types for extractors and extracted values, and the types of the API of the foreign inspector crate. This is currently done manually, and tedious to do at the call sites.

We should create convenient conversions between these two types.

See the code just needed for the simple bitcoin inspector:
https://github.com/near/mpc/blob/ab324c12f3bb3b22afc088b0c99b5a12915a90f2/crates/node/src/providers/verify_foreign_tx/sign.rs#L100-L149

### User Story

_No response_

### Acceptance Criteria

implement `From`/`TryFrom` conversions between all foreign_chain types that the inspector crate has "duplicate" types of.

_No response_

DSharifi:
closes #2085

### DIFF
--- a/crates/foreign-chain-inspector/src/contract_interface_conversions.rs
+++ b/crates/foreign-chain-inspector/src/contract_interface_conversions.rs
@@ -6,6 +6,8 @@ use crate::EthereumFinality;
 use crate::abstract_chain::inspector::{AbstractExtractedValue, AbstractExtractor};
 use crate::bitcoin::BitcoinExtractedValue;
 use crate::bitcoin::inspector::BitcoinExtractor;
+use crate::starknet::StarknetExtractedValue;
+use crate::starknet::inspector::{StarknetExtractor, StarknetFinality};
 
 #[derive(Debug, thiserror::Error)]
 pub enum ConversionError {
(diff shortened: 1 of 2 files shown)


crates/foreign-chain-inspector/src/contract_interface_conversions.rs
```diff
@@ -0,0 +1,353 @@
...
+
+#[derive(Debug, thiserror::Error)]
+pub enum ConversionError {
+    #[error("unsupported variant for conversion: {context}")]
+    UnsupportedVariant { context: &'static str },
+    #[error("integer overflow during conversion: {context}")]
+    IntegerOverflow { context: &'static str },
+}
+
+// --- BlockConfirmations ---
```
netrome: Not a fan of these section comments, can we remove them? Alternatively if you want to separate these we could make them separate modules (and possibly also separate files)

netrome (approved):
Lgtm, a few nits

PR #2114, opened 2026-02-16
fix: include topics in signature responses for EVM requests by using borsh over json

### LINKED ISSUE #2089: Topics occupy too much data, breaking NEAR limits
### Background

In https://github.com/near/mpc/pull/2087 we learned that the EvmLog extractor returns too much data, causing respond calls to fail to the contract. We need to figure out a better way to approach this.

We've discussed an alternative variant where full payloads are never passed to or from the contract, but only the expected/returned payload hash. We probably want to do something like this, but other alternatives may also be interesting to consider.

### User Story

_No response_

### Acceptance Criteria

We can support full EvmLog extractors without breaking contract call limits.

_No response_

DSharifi:
closes #2089

### DIFF
--- a/crates/contract-interface/src/types/foreign_chain.rs
+++ b/crates/contract-interface/src/types/foreign_chain.rs
@@ -209,7 +209,7 @@ pub struct StarknetRpcRequest {
 )]
 #[cfg_attr(
     all(feature = "abi", not(target_arch = "wasm32")),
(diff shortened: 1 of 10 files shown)


DSharifi:
Ignore the @Claude review. This feature is not released, so it's not breaking a rolling release.

PR #2124, opened 2026-02-17
fix: make run_receive_messages_loop infallible, log error on internal failures

### LINKED ISSUE #2101: `run_receive_messages_loop` can fail causing permanent disconnection for a running node
### Background

Last Friday we observed a node being disconnected from all other nodes. This node was lacking some tasks from the `debug/tasks` page, for example the "Network receive message loop" task wasn't running and neither was the incoming connection handler.

The problem is that there are several fallible calls in the `run_receive_messages_loop` that can fail under certain circumstances. And if this function returns an error, the network receive messages loop will be down forever. Additionally, any errors from this function are silently ignored in the node.

Instead of tearing down the network on errors, we should be able to log any error that can occur during normal operations and not kill the receive message loop task.

### User Story

As a node operator I don't want permanent connection failures.

### Acceptance Criteria

The run receive messages loop is rewritten to be infallible and handles any encountered errors.

See this slack thread for more context: https://nearone.slack.com/archives/C07UW93JVQ8/p1770965404288609

gilcu3:
Closes #2101

### DIFF
--- a/crates/node/src/network.rs
+++ b/crates/node/src/network.rs
@@ -377,6 +377,36 @@ enum SenderOrNewChannel {
     NewChannel(NetworkTaskChannel),
 }
 
+async fn run_receive_message(
+    client: Arc<MeshNetworkClient>,
+    receiver: &mut Box<dyn MeshNetworkTransportReceiver>,
+    new_channel_sender: &mpsc::UnboundedSender<NetworkTaskChannel>,
+    indexer_heights: Arc<IndexerHeightTracker>,
+) -> anyhow::Result<()> {
+    let message = receiver.receive().await?;
+    match message {
+        PeerMessage::Mpc(message) => {
+            let channel_id = message.message.channel_id;
+            let start_msg = match &message.message.kind {
+                MpcMessageKind::Start(start_msg) => Some(start_msg),
+                _ => None,
+            };
+            match client.sender_for(channel_id, start_msg, message.from) {
+                SenderOrNewChannel::Sender(sender) => {
+                    sender.send(message)?;
+                }
+                SenderOrNewChannel::NewChannel(channel) => {
+                    new_channel_sender.send(channel)?;
+                }
+            }
+        }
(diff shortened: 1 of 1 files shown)


gilcu3:
@claude can you explain this comment:
> Why this is critical: Issue https://github.com/near/mpc/issues/2101 describes permanent disconnection as the problem. This PR prevents the loop from crashing, but replaces it with a silent failure mode where messages are dropped without proper visibility. In production, operators won't know the node is degraded until signature requests start timing out.

it seems to me that it can only happen when the coordinator has died, which means that the node would have died anyway

EDIT: the job to reply https://github.com/near/mpc/actions/runs/22094386928 failed, don't know why

crates/node/src/network.rs
```diff
@@ -377,6 +377,36 @@ enum SenderOrNewChannel {
     NewChannel(NetworkTaskChannel),
 }
 
+async fn run_receive_message(
```
SimonRastikian: I understood what the function does, but it may deserve a couple of comments or a description about when it might fail

SimonRastikian (approved):

crates/node/src/network.rs
```diff
@@ -385,28 +415,17 @@ async fn run_receive_messages_loop(
...
-            }
+        if let Err(err) = run_receive_message(
+            client.clone(),
+            &mut receiver,
+            &new_channel_sender,
+            indexer_heights.clone(),
+        )
+        .await
+        {
+            tracing::warn!("run_receive_message failed with error: {err}");
```
SimonRastikian: Is it possible to DOS by filling the log with errors like this?

gilcu3: there is a line that should wait to receive a message, so a hot loop should not happen here unless the error is not transient, which is my main worry with claude's review

gilcu3:
@claude review

Please take into account my comment above

kevindeforth (commented):
Came too late, but looks good!

PR #2131, opened 2026-02-17
chore: validate local RPC provider config with on chain config

### LINKED ISSUE #2086: Check that the node's config matches the foreign chain policy in the contract when responding to foreign chain requests
### Background

Currently our foreign chain provider implementation doesn't make any check that it's configuration matches the contract when processing requests. We should add this.

### User Story

_No response_

### Acceptance Criteria

Nodes doesn't process foreign transaction verification requests unless their provider config matches the policy in the contract.

_No response_

DSharifi:
closes #2086

### DIFF
--- a/crates/node/src/mpc_client.rs
+++ b/crates/node/src/mpc_client.rs
@@ -8,6 +8,7 @@ use crate::indexer::types::{
     ChainCKDRespondArgs, ChainSendTransactionRequest, ChainSignatureRespondArgs,
     ChainVerifyForeignTransactionRespondArgs,
 };
+use crate::indexer::ReadForeignChainPolicy;
 use crate::metrics;
 use crate::network::{MeshNetworkClient, NetworkTaskChannel};
 use crate::primitives::MpcTaskId;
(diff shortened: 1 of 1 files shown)


crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -232,3 +244,176 @@ impl<ForeignChainPolicyReader: Send + Sync> VerifyForeignTxProvider<ForeignChain
...
+        on_chain: dtos::ForeignChainPolicy,
+    },
+    #[error("requested chain {requested:?} is not present in the on-chain foreign chain policy")]
+    ChainNotInPolicy { requested: dtos::ForeignChain },
+}
+
+async fn validate_foreign_chain_policy(
+    foreign_chains_config: &ForeignChainsConfig,
+    policy_reader: &impl ReadForeignChainPolicy,
+    request: &dtos::ForeignChainRpcRequest,
```
gilcu3: nit: I believe this code should be located in some other module

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -232,3 +244,176 @@ impl<ForeignChainPolicyReader: Send + Sync> VerifyForeignTxProvider<ForeignChain
...
+        .get_foreign_chain_policy()
+        .await
+        .map_err(ValidateForeignChainPolicyError::FetchOnChainPolicy)?;
+
+    if on_chain_policy != local_policy {
+        return Err(ValidateForeignChainPolicyError::PolicyMismatch {
+            local: local_policy,
+            on_chain: on_chain_policy,
+        });
+    }
```
netrome: We discussed whether we should use exact matching or just subset matching here, but I think this is a good and conservative start. We can relax this later if we want.

PR #2135, opened 2026-02-17
feat!: Domain separation

### LINKED ISSUE #2076: Domain separation support
### Background

For foreign transaction validation we want to be able to provide strong guarantees that we don't use the same domain for these requests and for arbitrary sign requests.

From the design doc:

## Domain Separation

To prevent callers from using plain `sign()` requests that could be mistaken for validated foreign-chain
transactions, we enforce domain separation by extending `DomainConfig` with a `DomainPurpose` enum.
Requests are only accepted for domains matching the purpose:

* `sign()` may only target domains with purpose `Sign`.
* `verify_foreign_transaction()` may only target domains with purpose `ForeignTx`.

```rust
pub enum DomainPurpose {
    Sign,
    ForeignTx,
    CKD,
}

pub struct DomainConfig {
    pub id: DomainId,
    pub scheme: SignatureScheme,
    pub purpose: DomainPurpose,
}
```

Compatibility note: legacy contract state does not include `DomainPurpose`. New nodes reading old state
must infer the purpose (e.g., treat existing Secp256k1/Ed25519/V2Secp256k1 domains as `Sign` and
Bls12381 domains as `CKD`) until a migration writes explicit purposes.

### User Story

As a bridge I don't want to risk validate payloads that could have been forged.

### Acceptance Criteria

We've implemented domain separation as per the foreign chain transaction design doc.

_No response_

netrome:
closes #2076 

This PR introduces domain purposes as a `BTreeMap<DomainId, DomainPurpose>` in the contract. See the change in [docs/foreign_chain_transactions.md](https://github.com/near/mpc/pull/2135/changes#diff-efe2abd54085303e9cae3132932ce2342119a05b449233938c85d08720936236) for a high-level overview of the design change.

### DIFF
--- a/crates/contract/src/state/running.rs
+++ b/crates/contract/src/state/running.rs
@@ -160,10 +160,11 @@ impl RunningContractState {
         let participant = AuthenticatedParticipantId::new(self.parameters.participants())?;
         let n_votes = self.add_domains_votes.vote(domains.clone(), &participant);
         if self.parameters.participants().len() as u64 == n_votes {
-            let domain_configs: Vec<DomainConfig> =
-                domains.iter().map(|(d, _)| d.clone()).collect();
-            let purposes: Vec<(DomainId, DomainPurpose)> =
-                domains.iter().map(|(d, p)| (d.id, *p)).collect();
+            let (domain_configs, purposes): (Vec<DomainConfig>, Vec<(DomainId, DomainPurpose)>) =
+                domains
+                    .iter()
+                    .map(|(domain, purpose)| (domain.clone(), (domain.id, *purpose)))
+                    .unzip();
             let new_domains = self.domains.add_domains(domain_configs.clone())?;
             Ok(Some(AddDomainsOutcome {
                 new_state: InitializingContractState {
--- a/crates/contract/src/state/running.rs
+++ b/crates/contract/src/state/running.rs
@@ -150,11 +150,10 @@ impl RunningContractState {
     /// If the number of votes for the same set of new domains reaches the number of participants,
     /// returns the InitializingContractState we should transition into to generate keys for these
     /// new domains, along with the purposes for each new domain.
-    #[allow(clippy::type_complexity)]
     pub fn vote_add_domains(
         &mut self,
         domains: Vec<(DomainConfig, DomainPurpose)>,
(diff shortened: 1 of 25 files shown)


crates/contract/src/state/running.rs
```diff
@@ -180,13 +191,49 @@ impl RunningContractState {
     pub fn is_participant(&self, account_id: &AccountId) -> bool {
         self.parameters.participants().is_participant(account_id)
     }
+
+    fn validate_domain_purpose_compatibility(
+        domains: &[(DomainConfig, DomainPurpose)],
+    ) -> Result<(), Error> {
+        for (domain, purpose) in domains {
```
gilcu3: shouldn't this go higher in the file?

docs/foreign_chain_transactions.md
```diff
@@ -232,17 +232,51 @@ pub enum DomainPurpose {
...
+### Backwards compatibility
+
+* **Old nodes + new contract**: Old nodes never query the purpose map and continue to work without
+  purpose enforcement.
+* **New nodes + old contract**: The map is empty. New nodes infer purpose from the signature scheme
+  (Secp256k1/Ed25519/V2Secp256k1 → `Sign`, Bls12381 → `CKD`). Since no `ForeignTx` domains exist
+  on old contracts, inference is always correct.
+* **Mixed nodes during rolling upgrade**: Old nodes won't enforce purpose checks, but `ForeignTx`
+  domains won't exist until all nodes upgrade and vote to add them.
+* **Domain lookup**: When resolving purpose for a domain, check the map first; if absent, fall back
```
gilcu3: hmm I see this is the trick. Was it better than allowing the new node to read both the old contract and the new contract state, and for the old one infer purposes?

gilcu3 (approved):
Looks good to me, although as you can see in several comments, I would have preferred to keep the implementation as it was in the doc.

crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
```diff
@@ -1,5 +1,6 @@
 ---
 source: crates/contract/tests/abi.rs
+assertion_line: 49
 expression: abi
```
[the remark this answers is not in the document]
gilcu3: for some reason it was not getting autogenerated in the past. Did you add any version or changed anything? Removing the line is enough, but we should try to figure what actually causes it. I had seen it before in the ts repo only

PR #2163, opened 2026-02-18
feat: Domain separation

### LINKED ISSUE #2076: Domain separation support

netrome:
Alternative approach for domain separation. Supersedes #2135 

Closes #2076

### DIFF
--- a/crates/contract/src/dto_mapping.rs
+++ b/crates/contract/src/dto_mapping.rs
@@ -639,7 +639,7 @@ impl IntoInterfaceType<dtos::DomainConfig> for &DomainConfig {
         dtos::DomainConfig {
             id: self.id.into_dto_type(),
             scheme: self.scheme.into_dto_type(),
-            purpose: self.purpose,
+            purpose: Some(self.purpose),
         }
     }
 }
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -533,9 +533,7 @@ impl MpcContract {
                     expected: DomainPurpose::ForeignTx,
                     actual: domain_config.purpose,
                 }
-                .message(
-                    "verify_foreign_transaction() requires a domain with purpose ForeignTx",
-                )
+                .message("verify_foreign_transaction() requires a domain with purpose ForeignTx")
                 .to_string(),
             );
         }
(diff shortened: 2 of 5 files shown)


crates/contract/src/primitives/domain.rs
```diff
@@ -321,12 +365,47 @@ pub mod tests {
...
+    #[rstest]
+    #[case(SignatureScheme::Secp256k1, DomainPurpose::Sign)]
+    #[case(SignatureScheme::Ed25519, DomainPurpose::Sign)]
+    #[case(SignatureScheme::V2Secp256k1, DomainPurpose::Sign)]
+    #[case(SignatureScheme::Bls12381, DomainPurpose::CKD)]
+    fn test_infer_purpose_from_scheme(
+        #[case] scheme: SignatureScheme,
+        #[case] expected: DomainPurpose,
+    ) {
+        assert_eq!(infer_purpose_from_scheme(scheme), expected);
```
netrome: Obviously not using given-when-then here. Feels fine for these cases for now.

netrome:
> (I don't think we need it, but it is much better than copilot. I wonder why it is not running by default anymore)

Yeah I thought it was a bug.

gilcu3:
> > (I don't think we need it, but it is much better than copilot. I wonder why it is not running by default anymore)
> 
> Yeah I thought it was a bug.

it was, fixed in #2168

DSharifi (approved):

PR #2179, opened 2026-02-19
feat: add on chain metrics for sign request payload version

### LINKED ISSUE #2178: Add on-chain metrics for sign request payload version usage
## Problem

We currently have no visibility into how many sign requests use the deprecated v1 payload (`deprecated_payload`) versus the new v2 payload (`payload_v2`). Without this data, we cannot make an informed decision about when it is safe to deprecate and remove support for the old payload version.

## Proposed Solution

Add on-chain counters to the signer contract that track:
- `sign_with_v1_payload_count`: number of sign requests using the deprecated `deprecated_payload` field
- `sign_with_v2_payload_count`: number of sign requests using the `payload_v2` field

Expose these counters via a `metrics()` view method so we can monitor usage over time and determine when v1 payload usage has dropped to zero (or near-zero), at which point we can proceed with deprecation.

DSharifi:
## Summary

Closes #2178

### DIFF
--- a/crates/contract-interface/src/types/metrics.rs
+++ b/crates/contract-interface/src/types/metrics.rs
@@ -2,6 +2,7 @@ use borsh::{BorshDeserialize, BorshSerialize};
 use serde::{Deserialize, Serialize};
 
 #[derive(
+    Default,
     Debug,
     Clone,
     Eq,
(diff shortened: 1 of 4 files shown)


netrome (dismissed):
Edit: Discussed on slack. Since these are public methods we want to deprecate it's nice to have these metrics visible.

gilcu3 (dismissed):
Thank you!

PR #2190, opened 2026-02-20
fix: add_domain_votes are preserved after resharing

### LINKED ISSUE #1601: Bug: Completing a resharing wipes add domain votes.
### Background

We observed on testnet that CKD votes was missing when we initiated a CKD vote. During this vote we made a participant set upgrade. Now looking at the code I note that once we conclude a participant upgrade in `vote_reshared` we'll initiate a new running state with:

```rust
                return Ok(Some(RunningContractState::new(
                    self.previous_running_state.domains.clone(),
                    Keyset::new(self.prospective_epoch_id(), self.reshared_keys.clone()),
                    self.resharing_key.proposed_parameters().clone(),
                )));
```

The new function will initiate the state without any domain votes:

```rust
    pub fn new(domains: DomainRegistry, keyset: Keyset, parameters: ThresholdParameters) -> Self {
        RunningContractState {
            domains,
            keyset,
            parameters,
            parameters_votes: ThresholdParametersVotes::default(),
            add_domains_votes: AddDomainsVotes::default(),
            previously_cancelled_resharing_epoch_id: None,
        }
    }
```

Thus effectively clearing any existing votes.

### User Story

As a node operator I don't want my votes to be wiped if the participant set is upgraded.

### Acceptance Criteria

We don't remove add domain votes after concluding a resharing, except if the voter is no longer a part of the participant set.

_No response_

gilcu3:
Closes #1601

- I tried not going beyond the specific issue at hand, to avoid changing the semantics of contract state changes regarding previous domain votes. It might be that we want to preserve votes for other transitions, but I refrained from doing so here.

### DIFF
--- a/crates/contract/src/primitives/domain.rs
+++ b/crates/contract/src/primitives/domain.rs
@@ -224,12 +224,14 @@ impl AddDomainsVotes {
 
     /// Filters out existing votes no longer in the participant set
     pub fn get_remaining_votes(&self, participants: &Participants) -> Self {
-        let mut remaining_votes = BTreeMap::new();
-        for (participant_id, vote) in &self.proposal_by_account {
-            if participants.is_participant_given_participant_id(&participant_id.get()) {
-                remaining_votes.insert(participant_id.clone(), vote.clone());
-            }
-        }
+        let remaining_votes = self
+            .proposal_by_account
+            .iter()
+            .filter(|&(participant_id, _vote)| {
+                participants.is_participant_given_participant_id(&participant_id.get())
+            })
+            .map(|(participant_id, vote)| (participant_id.clone(), vote.clone()))
+            .collect();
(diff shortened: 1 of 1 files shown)


crates/contract/src/state/running.rs
```diff
@@ -64,6 +71,7 @@ impl RunningContractState {
                     self.domains.clone(),
                     self.keyset.clone(),
                     self.parameters.clone(),
+                    self.add_domains_votes.clone(),
```
netrome: Took me a while to fin this line, but I guess this is where the magic happens now - because we add `AddDomainsVotes::default()` at all other call sites.

netrome (approved):
Nice stuff - thank you!

DSharifi (approved):

PR #2195, opened 2026-02-20
build: make sure nix includes all needed tools

### LINKED ISSUE #2194: verify missing tools in nix env
### Background

Once in a while we should verify that nix contains all the needed tools. It can be done for example by doing:
```
# remove cache
rm -rf target/
# run all cargo make steps without using the system binaries
nix develop --ignore-environment --command bash -c 'cargo make check-all'
```

Doing that I already discovered we don't have `perl` while we need it.

### User Story

_No response_

### Acceptance Criteria

Add missing deps to the nix shell. Document how to detect them in the future.

_No response_

gilcu3:
Closes #2194 

- The advertised command in the doc does not check absolutely all tools are in, but does cover the majority of it. Therefore, I don't think we should be more verbose about that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,18 @@ Add the following to your project settings in `.vscode/settings.json`:
 }
 ```
 
+### Verifying the Nix shell is complete
+
+Because `nix develop` inherits the system `$PATH`, missing packages can go unnoticed
+if the host already has them installed. To verify the shell provides everything needed,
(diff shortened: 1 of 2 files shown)


DSharifi (approved):

SimonRastikian (approved):
Cool! Thank you!

Issue #2186, opened 2026-02-19
Use contract interface sign request args instead of internal type in contract


DSharifi:
### Background

We are in the process of creating an SDK for the MPC contract. Currently for sign requests, we don't have proper DTO types, instead a local type defined in the contract `SignRequestArgs`.

The `SignRequestArgs` has fields for the payload, contain a special collection type `Bytes` which provides a bounded vector type over bytes.

We should move this bounded bytes type to the same collections crate where we have `NonEmptyBtreeSet` and `NoEmptyBtreeMap`, such that we can reuse the bounded bytes type in other crates where we serialize/deserialize signature request args without having to depend on the contract crate.

There is a quite sound implementation, with much better ergonomics and liberal license that we can copy https://github.com/ergoplatform/bounded-vec/blob/develop/src/bounded_vec.rs 

### Acceptance Criteria

The Bytes type in the contract is removed, and is replaced in a separate crate.

netrome:
Okay turns out this is not completed yet. We still have the `Bytes` type in the internal sign request args type. We should migrate to use the contract interface types throughout.

PR #2165, opened 2026-02-18
chore: use contract interface sign request args instead of internal type in contract

DSharifi:
closes #2186

PR #2199, opened 2026-02-20
feat: add new collection type, `BoundedVec`, to the bounded collections crate

DSharifi:
part of #2186 

This PR is broken out of #2165, by only creating the bounded vec type, and not integrating it into the contract yet. See slack thread for motivation https://nearone.slack.com/archives/C0912BTG51T/p1771583540732579?thread_ts=1771583231.086569&cid=C0912BTG51T

### DIFF
--- a/crates/bounded-collections/src/bounded_vec.rs
+++ b/crates/bounded-collections/src/bounded_vec.rs
@@ -292,6 +292,24 @@ impl<T, const L: usize, const U: usize> BoundedVec<T, L, U, witnesses::NonEmpty<
         self.inner.last().unwrap()
     }
 
+    /// Returns the last and all the rest of the elements
+    pub fn split_last(&self) -> (&T, &[T]) {
+        self.inner.split_last().unwrap()
+    }
+
+    /// Return a new BoundedVec with indices included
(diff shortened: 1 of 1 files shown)


crates/bounded-collections/src/bounded_vec.rs
DSharifi: Most of the implementation is taken from https://github.com/ergoplatform/bounded-vec/blob/develop/src/bounded_vec.rs (license allows this).

Some small changes have been made, such as not implementing AsMut since that can break the invariants.

Borsh and scehma implementations have also been added for ABI generation.

crates/bounded-collections/src/bounded_vec.rs
```diff
@@ -0,0 +1,1346 @@
...
+        self.inner.iter()
+    }
+
+    /// Returns an iterator that allows to modify each value
+    pub fn iter_mut(&mut self) -> IterMut<T> {
+        self.inner.iter_mut()
+    }
+}
+
+impl<T, const L: usize, const U: usize> BoundedVec<T, L, U, witnesses::NonEmpty<L, U>> {
```
gilcu3: in this type we implement mapped, mapped_ref, try_mapped, try_mapped_ref. But we don't do the same for the PossiblyEmpty variant. Was that a conscious decision?

PR #2202, opened 2026-02-20
feat: create `near-mpc-sdk` crate with types for sign requests

### LINKED ISSUE #2159: create SDK crate meant developers building on top of MPC contract
### Background

We currently don't have any SDKs in place for external developers to use when building on top of the MPC signer contract. 

We should create an sdk that contains API types, and contract method names needed to interact with the MPC smart contract

### User Story

_No response_

### Acceptance Criteria

Create SDK crate to interact with MPC contract as a user

_No response_

DSharifi:
stacked on top of #2199

closes #2159

Creates a new crate, `near-mpc-sdk`, which is meant for users that want to call the mpc contract.

This PR adds types for building sign request payloads, and deserializing responses.

#### Testing
- Unit test added to the sign module in the new crate
- The sandbox integration tests in for the contract have been changed to use the SDK for building payloads, and responses.


gilcu3 (dismissed):
LGTM.

Left a questions about the pattern used

crates/contract-interface/src/types/sign.rs
```diff
@@ -0,0 +1,28 @@
...
+
+const ECDSA_PAYLOAD_SIZE_BYTES: usize = 32;
+const EDDSA_PAYLOAD_SIZE_LOWER_BOUND_BYTES: usize = 32;
+const EDDSA_PAYLOAD_SIZE_UPPER_BOUND_BYTES: usize = 1232;
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+pub struct SignRequestArgs {
+    pub path: String,
+    #[serde(rename = "payload_v2")]
+    pub payload: Payload,
```
netrome: opinion: How about we just rename the field to be explicit?

```suggestion
    pub payload_v2: Payload,
```

netrome (changes requested):
Nice stuff. I think the EDDSA bounds need a better explanation but that's my only blocker.

crates/contract-interface/src/types/sign.rs
DSharifi: It will make make it more awkward for the sdk users seeing a `_v2` field

crates/contract/tests/sandbox/utils/sign_utils.rs
```diff
@@ -490,11 +490,16 @@ fn create_response_secp256k1(
         panic!("unable to use recovery id of 0 or 1");
     };
 
-    let respond_resp = SignatureResponse::Secp256k1(k256_types::Signature {
-        big_r: SerializableAffinePoint {
-            affine_point: big_r,
+    use k256::elliptic_curve::sec1::ToEncodedPoint;
```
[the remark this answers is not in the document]
netrome: Yeah from the perspective of local reasonability, this is strictly better imo. Even better would be to create a dedicated scope just for the import and calling `.to_encoded_point` so the import doesn't even leak to the rest of the function.

But anyway not important.

PR #2203, opened 2026-02-20
ci: gh action release automation

### LINKED ISSUE #1971: Create an automated bash script for the Release
### Background

The release process is a sequence of repetitive steps. There is no reason IMO not to have a bash script that completes these steps for us avoiding any mistakes.

Even better if we can integrate this bash script into the CI. One would just click a button that allows everything to happen. Release would be equivalent to one button.

### User Story

_No response_

### Acceptance Criteria

The script should be given a version number and should run all the commands in https://github.com/near/mpc/blob/main/RELEASES.md.
Then the user after running this script does the release buttons in the GH repo

_No response_

SimonRastikian:
Closes #1971 .

This should be combined with PR github.com/near/mpc/pull/2200 for a full automation with little manual setup.

This PR triggers a GH action on main:
* If the merged PR comes from a branch named release/vX.X.X (with proper semver) it launches the task
* The task first creates a tag with the proper version
* It then builds the three necessary docker images
* It collects the ID and Digest from each
* It builds the contract and renames it 
* It collects the hash value of the contract
* It makes a tar file of it
* It creates a draft release with the changelog pulled from CHANGELOG.md
* It adds all the ID and digest and hash values in the draft.

Only real problem is that I do not see a proper way of testing this without merging into main and without merging a dummy release/v0.0.0 PR.
Any suggestions?


SimonRastikian:
@claude  review

SimonRastikian:
@claude review this PR

netrome (changes requested):
Explicitly requested the change we discussed. Will review this deeper once that's updated.

netrome (changes requested):
The tag filter should be updated

gilcu3 (changes requested):
We need to install cargo-near

PR #2204, opened 2026-02-21
docs: Archive Signer design for legacy HOT key support

### LINKED ISSUE #2062: TEE application for long-term support of legacy keys
### Background

One of the biggest challenges surrounding MPC keys is legacy support.
It has to be assumed that any point in time, there are user funds locked in the keys managed by the MPC network. This comes at a cost: every key that is generated by the MPC network and exposed to users must be supported for eternity.

HOT labs is encountering this issue. They would like to retire their MPC network. We initially explored the option of porting their keyshares to our existing MPC network in #2018 and the associated #2021. While technically feasible, it's not something that NEAR has appetite for.

The main issue with importing foreign private keys to our network is the potential lack of security guarantees. NEAR  would be held responsible if anything happened _after_ importing the keyshares, but in the context of Threshold Keys, this is essentially the same thing as being held responsible for everything that happened _before_ taking ownership of them.

It doesn't matter if NEAR network can provide strong security guarantees - if the key was compromised in the past, those guarantees are irrelevant, because the old keyshares still map to the same secret key.

From a PR perspective, the above could potentially be disastrous. While a compromised imported foreign key does not indicate underlying security concerns with the NEAR MPC network, it will be a hard task to communicate this to the broader public. It is highly likely such an exploit would have negative consequences for the NEAR ecosystem.

#### Task

It was determined that an acceptable compromise is to migrate existing users to the keys currently managed by NEAR's MPC network and to move the legacy keyshares into an application running inside a trusted execution environment.
This ensures the following:
- NEAR does not assume liability for foreign key shares.
- HOT will already be compatible with the MPC network and once TEE integration is concluded, migration to another (more secure, due to guarantees of TEE's) key, should be easier.
- HOT users experience long-term support with acceptable security.

This issue is about designing the TEE application for long term support. In particular, the following **requirements** must be met:
- The application must run inside a TEE.
- There must be on-chain proof that the application is running inside a TEE.
- Updates to the code must be possible.
- The application must be able to generate signature responses compatible with what the NEAR MPC network does today.

### User Story

- As a stakeholder in the NEAR ecosystem, I like to avoid unnecessary risk exposure.
- As a HOT wallet user, I need long term support for my wallet.
- As a developer, I like to know the rough design before starting to code.

### Acceptance Criteria

Compose a design doc outlining the different components of a TEE application capable of long-term support of foreign key in line with above requirements. In particular, the following questions should be answered:
- How will the application submit proof it's running inside a TEE to the blockchain?
- How will the application receive the key-shares from the current HOT MPC nodes?
- How can we upgrade the application in the future?

#2018, #2021 
https://nearone.slack.com/archives/C0ADS52AZV3/p1770742473871499

pbeza:
Closes #2062

### DIFF
--- a/docs/securing-mpc-with-tee-design-doc.md
+++ b/docs/securing-mpc-with-tee-design-doc.md
@@ -595,6 +595,184 @@ _Note_ - Each MPC node will periodically call the `verify_tee`every 7 days.
 If any participant fails this check (since it did not submit a remote attestation with the new MPC docker image hash), then this participant will be automatically kicked out from the network, and a key re-sharing between the remaining participants will start.  
 Note - In case the number of remaining participants is less that the threshold. The node will not be kicked out, instead the contract will stop to accept signing requests until this is solved.
 
+# CVM Upgrades
+
+In the previous sections, we described how MPC node upgrades are
+performed.\
+The current design allows upgrading the MPC Docker image through
+on-chain voting, while keeping the same sealing key derived by the [Gramine Key Provider](#gramine-key-provider) across upgrades.
+
+This works because:
+
+-   The MPC Docker image hash is measured into **RTMR3**.
+-   The sealing key is derived from:
+    -   The Launcher's docker-compose measurements.
+    -   The OS measurements.
+-   The MPC node hash itself is **not** part of the sealing key
+    derivation material.
+
+As a result, MPC node upgrades do **not** change the sealing key, and
+existing encrypted key shares remain accessible after upgrade.
+
+However, there may be a need to upgrade:
+
+-   The **Launcher image**.
+-   The **Operating System (CVM base image)**.
+-   The **Dstack runtime version**.
+
+Unlike MPC node upgrades, these components are part of the measurement
+chain used to derive the sealing key.\
+Changing them will result in a **different sealing key**, meaning
+previously sealed key shares cannot be decrypted.
+
+Therefore, CVM upgrades require a controlled migration process.
+
+## Upgrade Impact Summary
+
+
+| Component        | Expected Frequency | Affects Sealing Key |
+|------------------|-------------------|---------------------|
+| MPC Node Image  | High              | No                  |
+| Launcher Image  | Low               | Yes                 |
+| OS Measurements | Very Low          | Yes                 |
+
+## Governance Changes
+
+To support CVM upgrades, the contract must support governance over:
+
+-   Approved **Launcher image hashes**.
+-   Approved **OS measurement identities** (MRTD, RTMR0--2).
+
+### Launcher Upgrade APIs
+
+Add the following contract methods:
+
+``` rust
+// requires a vote by a threshold of participants
+pub fn vote_add_new_launcher_image(
+    &mut self,
+    launcher_hash: Sha256Hash
+) -> Result<(), Error>
+// requires a vote by ALL of participants
+pub fn vote_remove_launcher_image(
+    &mut self,
+    launcher_hash: Sha256Hash
+) -> Result<(), Error>
+```
+
+Once the voting threshold is reached:
+
+-   The new Launcher hash is added to the list of approved Launcher
+    images.
+-   The contract accepts attestation quotes containing either the old or
+    new Launcher measurement.
+-   Multiple Launcher versions may temporarily coexist during migration.
+
+### OS Measurement Upgrade
+
+Two approaches are possible:
+
+1.  Add explicit voting APIs to add/remove approved OS measurements
+    (similar to Launcher upgrades).
+2.  Introduce new OS measurements as part of a contract upgrade.
+
+decision: Use explicit voting APIs, as it keeps OS upgrades decoupled from contract deployment.
+
+``` rust
+// requires a vote by a threshold of participants
+pub fn vote_add_os_measurement(
+    &mut self,
+    os_measurement: OsMeasurement
+) -> Result<(), Error>
+
+// requires a vote by ALL of participants
+pub fn vote_remove_os_measurement(
+    &mut self,
+    os_measurement: OsMeasurement
+) -> Result<(), Error>
+
+pub struct OsMeasurement {
+    pub mrtd: HexDigest,
+    pub rtmr0: HexDigest,
+    pub rtmr1: HexDigest,
+    pub rtmr2: HexDigest,
+}
+```
+
+
+## CVM Upgrade Flow
+
+Upgrading the Launcher or OS requires deploying a new CVM instance and
+migrating the key shares, since the sealing key will change.
+
+```mermaid
+sequenceDiagram
+    title CVM Upgrade Flow
+
+    participant Participants
+    participant Contract
+    participant Operator
+    participant OldNode
+    participant NewNode
+    participant Migration
+
+    Participants ->> Contract: Vote to approve new Launcher hash and/or OS measurements
+    alt Not enough votes
+        Contract ->> Contract: Increment vote counter
+    else Threshold reached
+        Contract ->> Contract: Add new Launcher/OS measurements to approved list
+        Note over Contract: Both old and new measurements are temporarily accepted
+    end
+
+    Operator ->> NewNode: Deploy new CVM with new Launcher/OS
+    NewNode ->> Contract: Submit remote attestation
+    Contract ->> Contract: Verify attestation matches approved measurements
+    Contract -->> NewNode: Mark node as valid
+
+    Note over Operator: Follow migration flow (see node-migration-guide.md)
+    Operator ->> Migration: Initiate migration from OldNode to NewNode
+    Migration ->> OldNode: Pull encrypted keyshares
+    Migration ->> NewNode: Restore keyshares
+    NewNode ->> Contract: Migration complete
+
+    Note over OldNode: Old node is deactivated
+
+    Participants ->> Contract: Vote to remove old measurements
+    alt Not enough votes
+        Contract ->> Contract: Increment vote counter
+    else All participants votes reached
+        Contract ->> Contract: Remove old measurements
+        Note over Contract: Only new measurements accepted
+    end
+```
+
+1.  **Governance Approval**\
+    Participants vote to approve the new Launcher image hash and/or new
+    OS measurements.\
+    After the voting threshold is reached, the contract temporarily
+    supports both the old and new measurements.
+
+2.  **Deploy New CVM**\
+    The operator deploys a new node using the approved Launcher/OS
+    version and submits remote attestation.\
+    The contract validates the new measurement set.
+
+3.  **Key Share Migration**\
+    The operator follows the migration flow to migrate the key share
+    from the old node to the new node via the migration service.
+
+    For full details see:
+
+    -   Node migration guide: `./docs/node-migration-guide.md`
+    -   Migration service design: `./docs/migration-service.md`
+
+4.  **Remove Old Measurements**\
+    After all nodes have been migrated, participants vote to remove the
+    old Launcher and/or OS measurements from the approved list.
+
+This completes the CVM upgrade cycle.
+
+
 # Threat Model
 
 This section outlines the threat model and key security considerations behind the solution design.


docs/hot-tee-signing-design.md
```diff
@@ -0,0 +1,580 @@
...
+1. Governors vote for a new HOT TEE App Docker image hash on-chain via `vote_code_hash()`.
+2. When `vote_threshold` is reached, the new hash is added to the allowed list.
+3. The running app's Contract State Subscriber detects the new allowed hash.
+4. The app stores the new hash to an encrypted file on disk.
+5. The HOT operator restarts the CVM.
+6. On restart, the Launcher pulls the new image, verifies the hash, extends RTMR3, starts the container.
+7. The new app submits a fresh attestation.
+8. After the upgrade deadline (configurable, default 7 days), old image hashes expire.
+
+## Redundancy and Recovery
```
barakeinav1: you treated Redundancy, but didn't mention recovery.
e.g the assumption is that we spin up 2 (identical) nodes during the initial key migration from the MPC network.

but what about down the line, I assume we want to support key backup or key migration
maybe use CKD or migration service?

docs/hot-tee-signing-design.md
```diff
@@ -0,0 +1,580 @@
...
+[tee-proposal]: https://github.com/near/mpc/blob/main/crates/contract/src/tee/proposal.rs
+[launcher-template]: https://github.com/near/mpc/blob/main/crates/contract/assets/launcher_docker_compose.yaml.template
+
+1. A YAML template ([`launcher_docker_compose.yaml.template`][launcher-template]) contains a `{{DEFAULT_IMAGE_DIGEST_HASH}}` placeholder.
+2. The placeholder is replaced with the approved Docker image hash.
+3. The filled YAML is SHA256-hashed to produce the `LauncherDockerComposeHash`.
+
+During attestation verification, the contract replays the TDX event log to reconstruct RTMR3 and checks that both the Docker image hash and launcher compose hash match the allowed lists. This ensures the attesting CVM is running an approved image via an approved launcher configuration.
+
+The HOT TEE governance contract will need its own launcher compose template, since the HOT TEE app has a different Docker Compose configuration than the MPC node.
```
kevindeforth: This is pretty useful, also for the back-up service.
I would suggest to move this to a stand-alone doc, explaining how the TEE attestation and updates are handled.
This can then be used by the backup-service and allows this doc to focus on the more HOT-migration specific stuff.

kevindeforth (commented):
Thanks a lot!
I very much appreciate the detailed Phase 2 (just, maybe give it a different name).
Left a few comments, mostly related to structure.

netrome (commented):
Some more comments. Nothing blocking from my side but there are still TODOs in the doc so I'll do a full review once this is no longer a draft.

docs/hot-tee-signing-design.md
```diff
@@ -0,0 +1,567 @@
+# Archive Signer
+
+This document outlines the design of the Archive Signer — a TEE application for long-term support of legacy HOT Wallet keys.
+
+## Background
+
+### Problem
```
kevindeforth: nit: "Background" is already sufficient

```suggestion
```

docs/hot-tee-signing-design.md
```diff
@@ -0,0 +1,567 @@
...
+
+HOT Labs operates a threshold MPC network for ECDSA (secp256k1) and EdDSA (ed25519) signing on behalf of HOT Wallet users. HOT Labs wants to retire this network.
+
+To support this, we'll provide a standalone application running inside a Trusted Execution Environment (TEE) that holds the reconstructed full private keys and handles signing requests. HOT will then be able to operate an instance of this application for their keyshares.
+
+c.f. [#2062][issue-2062], [#2018][issue-2018], [#2021][pr-2021]
+
+[issue-2062]: https://github.com/near/mpc/issues/2062
+[issue-2018]: https://github.com/near/mpc/issues/2018
+[pr-2021]: https://github.com/near/mpc/pull/2021
```
kevindeforth: If it's just a loose collection of links and they are not embedded in the text, then I prefer to add them at the end in the "Related Issues" section.

If they really belong here, help the reader understand why they are relevantr. For example:
_To support this, we'll provide a standalone application running inside a Trusted Execution Environment (TEE) that holds the reconstructed full private keys and handles signing request (c.f. [#2062](https://github.com/near/mpc/issues/2062))._

docs/hot-tee-signing-design.md
```diff
@@ -0,0 +1,567 @@
...
+#### Phase 2a: CVM Boot and Encryption Key Publication
+
+1. The CVM boots and the Archive Signer starts.
+2. On first boot (no keys on encrypted disk), the app generates an **ephemeral encryption key pair** inside the TEE.
+3. The app exposes the ephemeral public key via a local HTTP endpoint (`GET /import/public_key`). This endpoint is only available during the initial import phase and is disabled after key reconstruction succeeds.
+4. The app generates a TDX attestation quote binding the ephemeral public key in `report_data`, so operators can verify the public key belongs to a genuine TEE running the approved image.
+
+#### Phase 2b: Operator Encrypts and Delivers Keyshares
+
+1. Each operator fetches the CVM's ephemeral public key from `GET /import/public_key` and verifies the accompanying TDX attestation to confirm it originates from a genuine TEE.
```
kevindeforth: I guess we will need to provide a CLI for this step.

docs/hot-tee-signing-design.md
```diff
@@ -0,0 +1,567 @@
...
+- **Running the existing MPC stack with a single node** — [`threshold-signatures`][threshold-sigs] does not support fewer than 2 shares, and carries unnecessary complexity (P2P networking, resharing, triple/presignature generation, block event indexing).
+- **Running two MPC nodes inside one CVM** — higher maintenance burden and susceptible to the same bugs seen in mainnet/testnet deployments.
+
+[threshold-sigs]: https://github.com/near/threshold-signatures
+
+A custom binary was chosen because it is lightweight and easy to reason about, reuses modular TEE attestation and chain indexer components already being developed ([indexer design][indexer-design]), is future-proof (can be reused as a "graveyard" for NEAR's own deprecated domains), and aligns with other ongoing priorities (backup service, [#1891][issue-1891]).
+
+See also: [#2062][issue-2062], [#2018][issue-2018], [#2021][pr-2021].
+
+[issue-1891]: https://github.com/near/mpc/issues/1891
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/pull/2204/commits/fcd6df993c25837262482094fa37004be312c1d2

docs/hot-tee-signing-design.md
```diff
@@ -0,0 +1,535 @@
...
+
+CLIENT["Authenticated Accounts"]
+HOT_CONTRACT["HOT TEE Governance<br/>Contract (NEAR)"]
+RPC["Foreign Chain RPCs"]
+
+CLIENT -->|"POST /sign"| SIGNING
+SIGNING -->|"cross-chain auth"| RPC
+SIGNING ---|"persist / load on boot"| DISK
+TEE_GOV -->|"view: allowed hashes, foreign chain policy<br/>call: submit_participant_info"| HOT_CONTRACT
+
```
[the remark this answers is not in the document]
pbeza: Yes, there's a runtime dependency: the TEE Context polls `ForeignChainPolicy` from the governance contract (trusted RPC providers per chain), and the validation SDK needs these endpoints for cross-chain authorization calls. Splitting into two binaries is probably feasible if the governance side exposes the policy via a local endpoint or shared config. Thanks for the suggestion!

docs/hot-tee-signing-design.md
```diff
@@ -0,0 +1,535 @@
...
+    governors: BTreeSet<AccountId>,
+    /// Number of governor votes required to approve an action.
+    vote_threshold: u32,
+    /// TEE state: allowed image hashes, attestations.
+    /// Reuses the existing TeeState structure from mpc-contract.
+    tee_state: TeeState,
+    /// Trusted RPC providers per chain for cross-chain authorization.
+    /// Reuses the existing ForeignChainPolicy from contract-interface.
+    foreign_chain_policy: ForeignChainPolicy,
+    foreign_chain_policy_votes: ForeignChainPolicyVotes,
```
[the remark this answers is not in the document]
pbeza: Yeah, I think we should be able to reuse the foreign chain policy from the contract. And if not, we can probably generalize it a bit to make it reusable.

docs/hot-tee-signing-design.md
```diff
@@ -0,0 +1,535 @@
...
+class HOT_CONTRACT ext;
+class CLIENT ext;
+class RPC ext;
+```
+
+### Shared Components
+
+The Archive Signer is built on three reusable layers from this repository (bottom-up):
+
+- **[Embedded Indexer Node](#embedded-indexer-node)** — runs a full `near-indexer` (including `neard`) inside the CVM for trustless chain access.
```
kevindeforth:
```suggestion
- **[Embedded Indexer Node](#embedded-indexer-node)** — runs a full `near-indexer` (including `neard`) inside the CVM for trustless NEAR chain access.
```

docs/hot-tee-signing-design.md
```diff
@@ -0,0 +1,535 @@
...
+
+TEE_CTX --> CHAIN
+
+classDef new stroke:#d97706,stroke-width:3px;
+class HOT_APP new;
+```
+
+### Embedded Indexer Node
+
+The Archive Signer embeds a full `near-indexer` (which includes a `neard` node), the same as the MPC node, rather than using a lightweight RPC client. The embedded neard is used exclusively for **TEE governance operations**: monitoring the HOT governance contract for allowed Docker image hashes, launcher compose hashes, and foreign chain policy, and submitting TEE attestation transactions. Running `neard` inside the CVM eliminates external RPC trust assumptions — the app verifies chain state directly, with no external trust assumptions beyond the NEAR network itself.
```
kevindeforth: nit: unnecessary filler word and repetition.

```suggestion
The Archive Signer embeds a full `near-indexer` (which includes a `neard` node), the same as the MPC node, rather than using a lightweight RPC client. The embedded neard is used for **TEE governance operations**: monitoring the HOT governance contract for allowed Docker image hashes, launcher compose hashes, and foreign chain policy, and submitting TEE attestation transactions.
```

docs/tee-lifecycle.md
```diff
@@ -0,0 +1,280 @@
...
+        APP ->> SC: submit_participant_info(attestation, tls_pk)
+    end
+```
+
+Individual services may add steps between "Start application container" and the image hash check — for example, the Archive Signer performs [key import][key-import] on first boot.
+
+[launcher-pattern]: securing-mpc-with-tee-design-doc.md#launcher-pattern
+[key-import]: hot-tee-signing-design.md#key-import-process
+
+## TEE Context
```
kevindeforth: This section needs some more work, I propose to label this doc as WIP and continue on it after merging this PR to keep the discussion focused.

kevindeforth (approved):

docs/tee-lifecycle.md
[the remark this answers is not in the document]
pbeza: OK, I realized that if a node is running outdated firmware that appears in the attestation collateral advisories, the periodic re-attestation task (`periodic_attestation_submission`) will keep trying to submit an outdated TDX attestation.

The `monitor_attestation_removal` task won’t help either, since it only triggers when an attestation is not present for the node. Removing an attestation can only be done via `verify_tee`, which is currently called every 2 days for the MPC node — and we don’t have that mechanism for the Archive Signer.

I suggest adding the `verify_tee` call to the TEE Context, since it will likely be useful for other modules that rely on it as well. I propose tracking this as a follow-up in issue #2282.

Good catch, thanks!

docs/hot-tee-signing-design.md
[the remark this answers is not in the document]
netrome: Okay I wonder if it's worth adding an arrow in the chart to highlight this dependency then?

Whether it's worth separating the binaries or not, I'm not sure myself but I think it's nice to know if we have two completely decoupled concepts with a minimal boundary irrespective of if we bundle them in a single binary or split the binaries.

PR #2214, opened 2026-02-23
docs: added asset generation doc

### LINKED ISSUE #2207: document triple and presignature asset generation
### Background

Currently the triple and presignature generation is not well understood. We have observed in production cases in which:
- The presignature generation task is not reported as running
- The node is able to compute signatures as leader
- The number of available presignatures does not decrease

Also, it would be good to know how exactly when/how assets are marked as available, online or offline. This is a premise to check if our current solution is scalable.

### User Story

A better system understanding will allow us to improve it

### Acceptance Criteria

The assets generation process, and how its results appear in metrics, is explained in a doc.

_No response_

gilcu3:
Closes #2207

### DIFF
--- a/.github/prompts/pr-review.prompt.md
+++ b/.github/prompts/pr-review.prompt.md
@@ -33,10 +33,16 @@ PRIORITY CHECKS (report only if found):
    - Incorrect ownership patterns or lifetime issues
    - Concurrency issues (Arc/Mutex misuse, data races)
 
-6. Code Quality
+6. Security
+   - Injection vulnerabilities (e.g., command injection, path traversal, prompt injection)
+   - Hardcoded secrets or credentials in source code
+   - Secret values (tokens, keys, credentials) leaking through any output channel: serialization, debug formatting, logs, error messages, or API responses
+   - New config fields containing secrets must be protected from accidental exposure
+   - Sensitive data lingering in memory without zeroization where cryptographic material is involved
+
+7. Code Quality
    - Poor modularity (functions >100 lines, god objects)
    - Violated Single Responsibility Principle
-   - Security vulnerabilities (injection, hardcoded secrets)
 
 REVIEW STYLE:
 - List only issues that should block the merge
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -987,6 +987,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "attestation-cli"
+version = "3.6.0"
+dependencies = [
+ "anyhow",
+ "attestation",
+ "bs58 0.5.1",
+ "clap",
+ "mpc-attestation",
+ "mpc-crypto-types",
+ "mpc-primitives",
+ "node-types",
+ "reqwest 0.12.28",
+ "serde_json",
+ "sha2",
+ "tempfile",
+ "test-utils",
+ "time",
+ "tokio",
+ "url",
+]
+
+[[package]]
 name = "attohttpc"
 version = "0.19.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1640,7 +1662,7 @@ dependencies = [
  "serde-untagged",
  "serde-value",
  "thiserror 2.0.18",
- "toml",
+ "toml 0.8.23",
  "unicode-xid",
  "url",
 ]
@@ -5554,6 +5576,7 @@ dependencies = [
  "tokio-rustls",
  "tokio-stream",
  "tokio-util",
+ "toml 1.0.6+spec-1.1.0",
  "tower",
  "tracing",
  "tracing-subscriber",
@@ -9949,6 +9972,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "serde_spanned"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f8bbf91e5a4d6315eee45e704372590b30e260ee83af6639d64557f51b067776"
+dependencies = [
+ "serde_core",
+]
+
+[[package]]
 name = "serde_urlencoded"
 version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -10961,9 +10993,24 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dc1beb996b9d83529a9e75c17a1686767d148d70663143c7854d8b4a09ced362"
 dependencies = [
  "serde",
- "serde_spanned",
+ "serde_spanned 0.6.9",
  "toml_datetime 0.6.11",
  "toml_edit 0.22.27",
+]
+
+[[package]]
+name = "toml"
+version = "1.0.6+spec-1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "399b1124a3c9e16766831c6bba21e50192572cdd98706ea114f9502509686ffc"
+dependencies = [
+ "indexmap 2.13.0",
+ "serde_core",
+ "serde_spanned 1.0.4",
+ "toml_datetime 1.0.0+spec-1.1.0",
+ "toml_parser",
+ "toml_writer",
+ "winnow",
 ]
 
 [[package]]
@@ -10985,6 +11032,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "toml_datetime"
+version = "1.0.0+spec-1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "32c2555c699578a4f59f0cc68e5116c8d7cabbd45e1409b989d4be085b53f13e"
+dependencies = [
+ "serde_core",
+]
+
+[[package]]
 name = "toml_edit"
 version = "0.22.27"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -10992,7 +11048,7 @@ checksum = "41fe8c660ae4257887cf66394862d21dbca4a6ddd26f04a3560410406a2f819a"
 dependencies = [
  "indexmap 2.13.0",
  "serde",
- "serde_spanned",
+ "serde_spanned 0.6.9",
  "toml_datetime 0.6.11",
  "toml_write",
  "winnow",
@@ -11012,9 +11068,9 @@ dependencies = [
 
 [[package]]
 name = "toml_parser"
-version = "1.0.7+spec-1.1.0"
+version = "1.0.9+spec-1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "247eaa3197818b831697600aadf81514e577e0cba5eab10f7e064e78ae154df1"
+checksum = "702d4415e08923e7e1ef96cd5727c0dfed80b4d2fa25db9647fe5eb6f7c5a4c4"
 dependencies = [
  "winnow",
 ]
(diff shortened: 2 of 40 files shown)


docs/asset-generation.md
```diff
@@ -0,0 +1,351 @@
...
+during **signature generation**. It applies only to **ECDSA** at this point.
+
+## Pipeline overview
+
+```
+Triples  ──►  Presignatures  ──►  Signatures
+(slow)         (fast)              (1 round)
+```
+
+Each MPC signature requires one **presignature**, and each presignature
```
netrome: I think this is a bit confusing, since this doesn't apply to Robust ECDSA. Et least not triples.

docs/asset-generation.md
```diff
@@ -0,0 +1,352 @@
...
+Presignature generation is significantly faster, and signature generation
+is a single round.
+
+Triples are not domain-specific; a single shared triple store feeds
+presignature generation for all ECDSA domains. Presignatures are
+per-domain because they incorporate the domain's key share.
+
+**Capacity rule of thumb:** With `desired_triples_to_buffer = N`, the
+node can produce at most `N / 2` presignatures before new triples must
+be generated (since each presignature consumes a pair of triples).
```
kevindeforth: this seems like a repetition? What's the motivation of stating this here? If it needs to be mentioned, lets add it to line 16.

docs/asset-generation.md
```diff
@@ -0,0 +1,352 @@
...
+
+## Owned vs unowned assets
```
kevindeforth: how about naming this "Asset Storage"

docs/asset-generation.md
```diff
@@ -0,0 +1,352 @@
...
+
+Every asset has exactly one **owner**, the participant who initiated its
+generation. Only the owner may pick which asset to consume next for a new
+computation. All other participants in the same asset's group hold an
```
kevindeforth: _Only the owner may pick which asset to consume next for a new computation_

This sounds weird. The intention is to say that one can only pick owned assets for computations? If so, then maybe it would be good to add more context of what it means to be a leader of a computation for e.g. pre-signatures or signatures.

docs/asset-generation.md
```diff
@@ -0,0 +1,352 @@
...
+**unowned** copy.
+
+| Role | Storage | Retrieval |
+|------|---------|-----------|
+| Owner (leader) | In-memory queue + RocksDB | `take_owned()` — pops next usable asset, blocks if none available |
+| Follower | RocksDB only | `take_unowned(id)` — looks up by specific ID chosen by the leader |
+
+When the leader starts a presignature or signature computation it
+broadcasts the asset ID. Followers look the asset up by that ID.
```
kevindeforth: this is the first time we mention that assets have identifiers. Would be good to mention this when we explain what assets are.

docs/asset-generation.md
gilcu3: rewrote, let me know how it reads now

docs/asset-generation.md
gilcu3: explained above

docs/asset-generation.md
```diff
@@ -0,0 +1,352 @@
...
+
+**Follower-side generation:** Followers do not run their own generation
+loops for assets they don't own. Instead, the generic protocol runner
```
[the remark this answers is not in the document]
gilcu3: added a clarification that it is explained in the next section

docs/asset-generation.md
```diff
@@ -0,0 +1,427 @@
...
+  over time.
+
+### Owned vs unowned
+
+Every asset is generated by a set of participants, one of whom is the
+**leader** (owner), while the rest are **followers**. All participants
+hold a different share of the same asset, identified by a common unique
+ID.
+
+- The **owner** (leader) decides when to consume an asset. Owned assets
```
kevindeforth: ni:
```suggestion
- The **owner** (leader) decides when to consume their asset. Owned assets
```

docs/asset-generation.md
```diff
@@ -0,0 +1,427 @@
...
+   This check is at most ~1 second old.
+
+2. **The asset has not been used before.** It is removed from storage
+   atomically on retrieval. Each asset is consumed exactly once.
+
+3. **No real-time liveness guarantee.** A participant can go offline
+   between the liveness check and the start of the protocol. If that
+   happens, the protocol will fail and the asset is lost. There is no
+   reservation or retry mechanism for the asset itself — the node simply
+   uses the next available one.
```
kevindeforth: This is does not list a "property" of assets.

I would suggest to make it a remark after stating the properties.

```suggestion
Remark: **No real-time liveness guarantee.** A participant can go offline
   between the liveness check and the start of the protocol. If that
   happens, the protocol will fail and the asset is lost. There is no
   reservation or retry mechanism for the asset itself — the node simply
   uses the next available one.
```

docs/asset-generation.md
```diff
@@ -0,0 +1,427 @@
...
+
+## Asset storage
+
+Both owned and unowned assets are persisted to RocksDB. Owned assets are
+additionally kept in an in-memory queue (the `DoubleQueue` described
+below) so the leader can quickly select the next usable one. Unowned
+assets live only in RocksDB, indexed by ID, and are retrieved on demand
+when the leader broadcasts which asset to use.
+
+Followers do not run their own generation loops. Instead, the generic
```
kevindeforth: _generation loop_ is undefined. I think this sentence is pointless here because if we spell it out it reads:
_Followers do not initiate generation of unowned assets_

So it would be better to mention this under #Owned vs unowned and explain what it means to be a "follower" and a "leader"

netrome (dismissed):
I find this really helpful, thanks for the writeup!

docs/asset-generation.md
```diff
@@ -0,0 +1,400 @@
...
+```
+Triples  ──►  Presignatures  ──►  Signatures
+(slow)         (fast)              (1 round)
+```
+
+Each MPC signature requires one **presignature**, and each presignature
+requires a pair of **triples** (called a `PairedTriple`). Triple generation
+is the bottleneck: it involves heavy OT-based cryptographic computation.
+Presignature generation is significantly faster, and signature generation
+is a single round.
```
kevindeforth: nit: We don't need to concern the reader with the notion of communication rounds.
```suggestion
is fairly cheap.
```

docs/asset-generation.md
```diff
@@ -0,0 +1,428 @@
...
+   their share, then run the protocol.
+4. The signature computation is a single round. The leader does **not**
+   wait for all followers to confirm success (`leader_waits_for_success`
+   returns `false`).
+
+## Asset cleanup on startup
+
+Source: `crates/node/src/assets/cleanup.rs`
+
+On node startup, `delete_stale_triples_and_presignatures()` compares
```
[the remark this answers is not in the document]
gilcu3: it does execute after resharing as well. Clarified that it is not only node startup, but more generally when entering running state

docs/asset-generation.md
```diff
@@ -0,0 +1,428 @@
...
+
+Followers do not initiate asset generation, as expected. Instead, the
+generic protocol runner (`crates/node/src/protocol.rs`) handles incoming
+generation requests from leaders over the P2P network. When a leader
+spawns a triple or presignature protocol, each follower in the chosen
+participant set receives the protocol messages, runs its side of the MPC
+computation, and stores the result as an unowned asset.
+
+This asymmetry means only the owner's queue management matters for
+throughput. Followers are passive.
```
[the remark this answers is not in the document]
gilcu3: will take the borrower suggestion. That will lead to some rewrites, will try to make it into an isolated commit to make reviewing a bit easier

docs/asset-generation.md
```diff
@@ -0,0 +1,404 @@
...
+is fairly cheap.
+
+The system supports an arbitrary number of secret keys. Each key is
+uniquely identified by a `DomainId`, which maps to a combination of
+signature scheme (ECDSA/EdDSA), protocol (OT-Based ECDSA, Robust ECDSA,
+FROST EdDSA) and application (signatures, CKD, foreign transactions).
+See #1649 for more information on domains.
+Triples are not domain-specific; a single shared triple store feeds
+presignature generation for all OT-based ECDSA domains. Presignatures
+are per-domain because they incorporate the domain's key share.
```
kevindeforth: connect with previous paragraph and make it part of intro:

```suggestion
Our system supports multiple cryptographic protocols and an arbitrary number of secret keys per protocol. Each key is uniquely identified by a `DomainId`, which maps to a combination of
signature scheme (ECDSA/EdDSA), protocol (OT-Based ECDSA, Robust ECDSA,
FROST EdDSA) and application (signatures, CKD, foreign transactions).
See #1649 for more information on domains.

Currently, only our OT-based ECDSA protocol requires pre-computed triples: each signature requires one **presignature**, and each presignature requires a pair of **triples** (called a `PairedTriple`)

Triples  ──►  Presignatures  ──►  Signatures
(slow)         (fast)              (fastest)


Note that presignatures are computed from the domain's key share, which makes them domain-specific, whereas triples are simply protocol-specific (a triple can be used to generate a presignature for any domain of the same protocol).
```

docs/asset-generation.md
```diff
@@ -0,0 +1,404 @@
...
+is the bottleneck: it involves heavy OT-based cryptographic computation.
+Presignature generation is significantly faster, and signature generation
+is fairly cheap.
+
+The system supports an arbitrary number of secret keys. Each key is
+uniquely identified by a `DomainId`, which maps to a combination of
+signature scheme (ECDSA/EdDSA), protocol (OT-Based ECDSA, Robust ECDSA,
+FROST EdDSA) and application (signatures, CKD, foreign transactions).
+See #1649 for more information on domains.
+Triples are not domain-specific; a single shared triple store feeds
```
kevindeforth: we haven't defined what a "triple store" is at this point.

docs/asset-generation.md
```diff
@@ -0,0 +1,404 @@
...
+
+The asset queue manages asset usage with three goals:
+
+- Efficiently provide an asset compatible with the current set of online
+  participants when one is needed.
+- Ensure each asset is consumed at most once.
+- Prevent the store from growing indefinitely by evicting unusable assets
+  over time.
+
+### Owned vs unowned
```
kevindeforth: this paragraph is very nice now! only nit:
place it _before_ `## Queue design`, not after

docs/asset-generation.md
```diff
@@ -0,0 +1,404 @@
...
+### Which assets get discarded?
+
+The queue discards assets to prevent the store from filling up with
+unusable entries:
+
+- **Offline assets** are the only ones subject to discarding. They are
+  evicted from the back of the queue when the store reaches its target
+  buffer size.
+- **Online assets are never discarded.** If an asset is online, it is
+  moved to the front of the queue, even during a discard pass.
```
kevindeforth: This applies to owned assets only, right? How are unowned assets maintained?

docs/asset-generation.md
```diff
@@ -0,0 +1,404 @@
...
+unowned assets, since it delegates to `clean_db()`. The only event that
+clears them is a full asset wipe on epoch change (resharing). In normal
+operation this constitutes a slow disk storage leak.
+
+## Background asset generation loops
+
+Background generation is launched by `spawn_background_tasks()`
+(`crates/node/src/providers/ecdsa.rs`). It spawns:
+
+1. **One** triple generation loop (shared across all domains).
```
kevindeforth: Didn't we [state earlier](https://github.com/near/mpc/pull/2214#discussion_r2896258354) that triples are only shared between OT-based ECDSA domains?

kevindeforth (commented):
Good progress!
Still a few suggestions, that aim to make the document more coherent.

The main blocker for me right now is the `## Background asset generation loops` section, which I think is too low-level and hard to parse. Left a suggestion for that.

docs/asset-generation.md
gilcu3: I don't think we need to define triple store, seems quite obvious to the reader that a triple store is the place where triples are stored.

docs/asset-generation.md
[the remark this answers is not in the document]
gilcu3: I read your suggestion a couple of times, and it does not feel better than the current text, which I like a bit more. So unless you feel strongly that there is something to be fixed here, I am leaving the text as is

kevindeforth (dismissed):

netrome (dismissed):

kevindeforth (approved):
Thank you!

PR #2254, opened 2026-02-26
feat(sdk): chain-specific builder entry points for foreign chain request builder

### LINKED ISSUE #2253: SDK: Add chain-specific builder entry points to ForeignChainRequestBuilder
### Background

The `ForeignChainRequestBuilder` currently uses chain-prefixed methods like `with_abstract_tx_id()`, `with_starknet_tx_id()` etc. directly on the initial builder state. This means chain-specific methods are all mixed at the top level, which is confusing when multiple chains are supported and doesn't guide the user through the correct builder flow.

Instead, the builder should first require the user to select a chain (`.bitcoin()`, `.starknet()`, `.abstract_chain()`), and then present chain-agnostic methods (`.with_tx_id()`, `.with_finality()`, etc.) that are scoped to the selected chain.

### Acceptance Criteria
`ForeignChainRequestBuilder` requires the user to select a chain explicitly.

DSharifi:
closes #2253

### DIFF
--- a/crates/near-mpc-sdk/src/foreign_chain/abstract_chain.rs
+++ b/crates/near-mpc-sdk/src/foreign_chain/abstract_chain.rs
@@ -66,20 +66,17 @@ impl From<BuildableAbstractRequest> for ForeignChainRpcRequestWithExpectations {
     }
 }
 
-impl ForeignChainRequestBuilder<NotSet, NotSet, NotSet> {
-    pub fn abstract_chain(
-        self,
-    ) -> ForeignChainRequestBuilder<AbstractRequest<NotSet, NotSet>, NotSet, NotSet> {
-        ForeignChainRequestBuilder {
+impl ForeignChainRequestBuilder<AbstractRequest<NotSet, NotSet>, NotSet, NotSet> {
+    pub fn new_abstract() -> Self {
+        Self {
             request: AbstractRequest {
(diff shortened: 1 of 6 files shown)


crates/near-mpc-sdk/src/foreign_chain/abstract_chain.rs
```diff
@@ -67,7 +67,25 @@ impl From<BuildableAbstractRequest> for ForeignChainRpcRequestWithExpectations {
 }
 
 impl ForeignChainRequestBuilder<NotSet, NotSet, NotSet> {
-    pub fn with_abstract_tx_id(
+    pub fn abstract_chain(
+        self,
+    ) -> ForeignChainRequestBuilder<AbstractRequest<NotSet, NotSet>, NotSet, NotSet> {
```
netrome: Does this need to take self? It would be nicer to construct it directly with `ForeignChainRequestBuilder::abstract_chain()` and skipping the `new` imo

DSharifi: Good point.

I changed it to `new_bitcoin`, `new_abstract`, etc. so it's clear it's the entry point of the builder, and keeping the `new` convertion for construction.

Similar to `new_current_thread` and `new_multi_thread` for tokio builder.

PR #2257, opened 2026-02-26
fix: bug in from_be_bytes_mod_order when not a multiple of 8 bytes

### LINKED ISSUE #2256: Fix bytes-to-scalar bug in CKD ScalarWrapper::from_be_bytes_mod_order
### Background

`ScalarWrapper::from_be_bytes_mod_order` in `crates/threshold-signatures/src/confidential_key_derivation/scalar_wrapper.rs` has a bug on the remainder branch (line 47):

```rust
res = res * res.shl(count * 8) + blstrs::Scalar::from(remainder);
```

The extra `res *` squares the accumulator instead of just shifting it. The correct code is:

```rust
res = res.shl(count * 8) + blstrs::Scalar::from(remainder);
```

Currently unreachable — the only caller (`FromOkm::from_okm`) always passes 48 bytes (multiple of 8), so `count` is always 0 and this branch never executes. The existing stress test only checks for panics, not correctness, so it doesn't catch this.

### Acceptance Criteria

- Fix the remainder branch to `res = res.shl(count * 8) + blstrs::Scalar::from(remainder)`
- Add a correctness test for non-8-aligned inputs

gilcu3:
Closes #2256

### DIFF
--- a/crates/threshold-signatures/src/confidential_key_derivation/scalar_wrapper.rs
+++ b/crates/threshold-signatures/src/confidential_key_derivation/scalar_wrapper.rs
@@ -67,6 +67,7 @@ mod tests {
     };
     use rand::Rng as _;
     use rand_core::{RngCore, SeedableRng};
+    use rstest::rstest;
 
     #[test]
     // This test only makes sense if `overflow-checks` are enabled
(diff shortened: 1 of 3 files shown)


crates/threshold-signatures/src/confidential_key_derivation/scalar_wrapper.rs
```diff
@@ -83,6 +83,28 @@ mod tests {
...
+        let expected = blstrs::Scalar::from(0x0001_0203_u64);
+        assert_eq!(result.0, expected);
+
+        // 9 bytes: one full 8-byte chunk + 1 remainder byte
+        let bytes = [1, 2, 3, 4, 5, 6, 7, 8, 9];
+        let result = ScalarWrapper::from_be_bytes_mod_order(&bytes);
+        let expected =
+            blstrs::Scalar::from(0x0102_0304_0506_0708_u64).shl(8) + blstrs::Scalar::from(9u64);
+        assert_eq!(result.0, expected);
+    }
```
DSharifi: Nit:
Could be separate test cases, or parameterized

DSharifi (approved):

netrome (approved):
Thanks for fixing this 🙏

PR #2268, opened 2026-02-27
fix: use bounded incoming message buffers for all protocols

### LINKED ISSUE #2247: Remote DoS -- Node OOM crash
### Background

The async inbox keeps a `HashMap<MessageHeader, SubMessageQueue>` and blindly allocates a new unbounded queue for every header it sees. A malicious peer can fabricate arbitrary _channel_/_waitpoint_ identifiers that the honest state machine will never poll. Those spoofed messages are never drained, so the map and the buffered payloads grow without bound; repeated packets trigger memory exhaustion and **take the participant down**.

### Acceptance Criteria

- **Quick patch**: Treat the buffer as a cache—only accept a bounded number of outstanding waitpoints per channel and evict or drop messages once the cap is hit.
- **Robust fix**: Introduce explicit waitpoint registration/capabilities. Require the receiver to announce which waitpoints are valid (or hand out per-waitpoint tokens) before accepting traffic. Messages that don't match an outstanding registration are rejected up front.

### Context

Transferred from https://github.com/near/threshold-signatures/issues/185 (archived repo), originally opened by @0xsecaas.

gilcu3:
Closes #2247

Added per-protocol incoming messages buffer capacity constants throughout threshold-signatures. `Comms::with_buffer_capacity(max)` rejects messages for new headers once the cap
  is reached; messages for existing entries still flow. Honest participants always use the same buffer capacity for each protocol.

Now each protocol now declares its own maximum:
  - Simple protocols (sign, presign, CKD, DKG): small constants (0–7), derived by counting waitpoints
  - Triple generation: $131 \cdot N \cdot (P-1) + 7$, derived from the sub-protocol structure (and empirical tests), when N is the number of triples (used in batch triple generation) and P the number of participants. This is an exact formula for N = 1, but just an upper bound for N > 1. The reason for this is an optimization in one of the subprotocols, which makes computing the total with a constant formula impossible (because it uses hashing to decide some execution branches). 

Majority of the code is the added tests to guarantee that the new buffer capacity bounds do not break existing implementations.

This solves the issue because the buffer is always bounded. A malicious can still stall the protocol, but can no longer cause an OOM, which seems a strict improvement. A better solution might be possible, but would certainly require breaking changes, so this one seems the best we can get for now.

As a byproduct, the computed numbers make it extremely clear how many rounds each protocol has, which is something that we did not know previously (for the triples for example)

### DIFF
--- a/crates/threshold-signatures/src/ecdsa/robust_ecdsa/sign.rs
+++ b/crates/threshold-signatures/src/ecdsa/robust_ecdsa/sign.rs
@@ -497,7 +497,6 @@ mod test {
                     &presignature,
                 );
             let comms = Comms::with_buffer_capacity(usize::MAX);
-            let comms_ref = comms.clone();
             let participant_list =
                 crate::participants::ParticipantList::new(&participants).unwrap();
             let fut = super::fut_wrapper(
@@ -509,8 +508,8 @@ mod test {
                 rerandomized,
                 msg_scalar,
             );
+            comms_refs.push((p, comms.clone()));
             let prot = make_protocol(comms, fut);
-            comms_refs.push((p, comms_ref));
             protocols.push((p, Box::new(prot)));
         }
 
--- a/crates/threshold-signatures/src/confidential_key_derivation/protocol.rs
+++ b/crates/threshold-signatures/src/confidential_key_derivation/protocol.rs
@@ -319,7 +319,6 @@ mod test {
 
         for &p in &participants {
             let comms = Comms::with_buffer_capacity(usize::MAX);
-            let comms_ref = comms.clone();
             let private_share = SigningShare::new(f.eval_at_participant(p).unwrap().0);
             let key_pair = KeygenOutput {
                 public_key: pk,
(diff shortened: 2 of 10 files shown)


crates/threshold-signatures/src/ecdsa/robust_ecdsa/sign.rs
```diff
@@ -18,6 +18,12 @@ use frost_core::serialization::SerializableScalar;
 use subtle::ConditionallySelectable;
 type C = Secp256K1Sha256;
 
+/// Maximum incoming buffer entries for the coordinator in the robust ECDSA sign protocol.
+pub(crate) const ROBUST_ECDSA_SIGN_MAX_INCOMING_COORDINATOR_ENTRIES: usize = 1;
+/// Maximum incoming buffer entries for non-coordinator participants in the robust ECDSA sign protocol.
+#[cfg(test)]
```
kevindeforth: idem

SimonRastikian (changes requested):
I reviewed most of the PR. 
As explained in the meeting, we all know this is not the cleanest solution. I hope you could open an Issue describing a cleaner way to solve it. (This is independent of my approval)

The changes I hope to see is more helper functions that cleanup the code. Otherwise thanks for takling this.

Issue #503, opened 2025-06-16
define launcher upgrade process


barakeinav1:
upgrade of launcher's code  is not possible while keeping any secrets,
need to define a contingency in case there is a critical bug that requires a fix in the launche

netrome:
Current strategy is to successively restart nodes and do a resharing after each restarted node. That should be sufficient for the initial testnet launch. Afterwards, we can work on this in the context of our upcoming disaster recovery milestone as per https://docs.nearone.org/doc/tee-integration-0YgwlmO6Fq.

PR #2270, opened 2026-03-01
docs:  CVM upgrade mechanism with launcher and OS measurement voting

### LINKED ISSUE #737: Define a process for updating the hardcoded RTMR values in the smart contract
### Background

Whenever we change anything in the launcher or the TEE-enabled MPC Docker image, we effectively change the TEE measurements (a.k.a. [RTMRs](https://docs.phala.network/phala-cloud/tees-attestation-and-zero-trust-security/attestation#runtime-measurement-fields)). For simplicity—and given the tight deadlines—we were okay with hardcoding the expected RTMR values in the smart contract. But now it’s becoming a problem, since we don’t have a clear process for keeping those values up to date.

### User Story

As a developer, I’d like to understand how we ensure the smart contract always stays up to date with the latest RTMRs.

### Acceptance Criteria

A clear process is defined (and ideally implemented—though implementation can be handled in a separate PR) to ensure RTMR values in the smart contract are always kept up to date.

- https://nearone.slack.com/archives/C0912BTG51T/p1756478901934479?thread_ts=1756392939.982719&cid=C0912BTG51T
- https://github.com/near/mpc/pull/971#pullrequestreview-3169034406
- https://github.com/near/mpc/pull/971#discussion_r2309428735
- https://github.com/near/mpc/blob/1e1fe0c418a3b7f0079a8f2851885d048d50012f/attestation/src/measurements.rs#L12-L41
- https://github.com/near/mpc/blob/1e1fe0c418a3b7f0079a8f2851885d048d50012f/attestation/src/measurements.rs#L68-L86

barakeinav1:
fixes https://github.com/near/mpc/issues/503

Closes #737

### DIFF
--- a/docs/securing-mpc-with-tee-design-doc.md
+++ b/docs/securing-mpc-with-tee-design-doc.md
@@ -596,7 +596,7 @@ Note - In case the number of remaining participants is less that the threshold. 
 In the previous sections, we described how MPC node upgrades are
 performed.\
 The current design allows upgrading the MPC Docker image through
-on-chain voting, while keeping the same sealing key across upgrades.
+on-chain voting, while keeping the same sealing key derived by the [Gramine Key Provider](#gramine-key-provider) across upgrades.
 
 This works because:
 


docs/securing-mpc-with-tee-design-doc.md
```diff
@@ -591,6 +591,162 @@ _Note_ - Each MPC node will periodically call the `verify_tee`every 7 days.
...
+
+Using explicit voting APIs is recommended, as it keeps OS upgrades
+decoupled from contract deployment.
+
+## CVM Upgrade Flow
+
+Upgrading the Launcher or OS requires deploying a new CVM instance and
+migrating the key shares, since the sealing key will change.
+
+```mermaid
```
kevindeforth: nice graph!

docs/securing-mpc-with-tee-design-doc.md
```diff
@@ -591,6 +591,162 @@ _Note_ - Each MPC node will periodically call the `verify_tee`every 7 days.
...
+### OS Measurement Upgrade
+
+Two approaches are possible:
+
+1.  Add explicit voting APIs to add/remove approved OS measurements
+    (similar to Launcher upgrades).
+2.  Introduce new OS measurements as part of a contract upgrade.
+
+Using explicit voting APIs is recommended, as it keeps OS upgrades
+decoupled from contract deployment.
```
kevindeforth: I agree with this statement.

gilcu3: From previous discussions around #737 I am not totally sure we have consensus on this. I don't have anything against voting, but agree that maybe voting for everything is not the best approach. If some values are expected to change very slowly I am not sure if voting is better than contract deployments (which IIRC was proposed recently as an alternative). We should probably sync about this CC: @netrome

gilcu3 (changes requested):
Reads very good, thank you!

I requested a small change. I think the votes must be unanimous, as all node operators need to do manual actions when this happens.

Also, I think regarding how the RTMRs are updated (voting vs contract upgrade) we should have a quick sync to reach consensus, I believe I have seen people supporting both possibilities in the past. Would love to hear pros and cons of each approach.

docs/securing-mpc-with-tee-design-doc.md
gilcu3: very nice graph indeed :D I think we should do similar ones for the Threat model, to mention which attacks are out of scope for example

PR #2277, opened 2026-03-02
refactor: unify crypto conversions

### LINKED ISSUE #2267: cryptographic conversions
### Background

Recently we have agreed that we need to have a single source of truth for crypto type conversions. Currently we do them in many places, which could lead to problems. We agreed we could add this in the contract-interface crate behind features. The conversions needed are from dto types to:
- friendly representation for near host functions
- k256 secp256k1 types
- ed25519 dalek types
- blstrs types

We could have each of these sets of conversions behind a feature, giving maximum liberty to the user. None of them would be default features.

### User Story

Making easy to use these types would lead to higher security guarantees.

### Acceptance Criteria

The conversions are implemented in the contract-interface crate, and all its usage across the codebase is modified to use them.

_No response_

gilcu3:
Closes #2267 

There are more lines added than removed basically because I added a lot of tests that probably did not exist before

### DIFF
--- a/crates/contract-interface/src/types/conversions/ed25519_dalek.rs
+++ b/crates/contract-interface/src/types/conversions/ed25519_dalek.rs
@@ -3,8 +3,8 @@ use crate::types::crypto::Ed25519PublicKey;
 use crate::types::primitives::Ed25519Signature;
 use curve25519_dalek::edwards::CompressedEdwardsY;
 
-impl From<&CompressedEdwardsY> for Ed25519PublicKey {
-    fn from(point: &CompressedEdwardsY) -> Self {
+impl From<CompressedEdwardsY> for Ed25519PublicKey {
+    fn from(point: CompressedEdwardsY) -> Self {
         Ed25519PublicKey::from(point.to_bytes())
     }
 }
@@ -23,8 +23,8 @@ impl TryFrom<&Ed25519PublicKey> for ed25519_dalek::VerifyingKey {
     }
 }
 
-impl From<&ed25519_dalek::Signature> for Ed25519Signature {
-    fn from(sig: &ed25519_dalek::Signature) -> Self {
+impl From<ed25519_dalek::Signature> for Ed25519Signature {
+    fn from(sig: ed25519_dalek::Signature) -> Self {
         Ed25519Signature::from(sig.to_bytes())
     }
 }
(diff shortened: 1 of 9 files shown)


crates/contract-interface/Cargo.toml
```diff
@@ -6,12 +6,21 @@ edition = { workspace = true }
 
 [features]
 abi = ["borsh/unstable__schema", "schemars", "bounded-collections/abi"]
+blstrs = ["dep:blstrs"]
+ed25519-dalek = ["dep:curve25519-dalek", "dep:ed25519-dalek"]
```
gilcu3: this is a bit misleading, but I did not one to create one more feature

netrome:
> There are more lines added than removed basically because I added a lot of tests that probably did not exist before

There's something slightly unsettling about the "probably" in that sentence :sweat_smile:

crates/contract-interface/Cargo.toml
netrome: Right because we need both for the same conversions. I think that's fine, the feature represent the `ed25519-dalek` conversions whereas the dependencies are the deps needed for those conversions.

crates/contract-interface/src/types/conversions/near.rs
```diff
@@ -0,0 +1,281 @@
...
+    type Error = CryptoConversionError;
+    fn try_from(pk: &PublicKey) -> Result<Self, Self::Error> {
+        near_sdk::PublicKey::try_from(pk.clone())
+    }
+}
+
+impl From<Ed25519PublicKey> for near_sdk::PublicKey {
+    fn from(pk: Ed25519PublicKey) -> Self {
+        near_sdk::PublicKey::from_parts(near_sdk::CurveType::ED25519, pk.0.into())
+            .expect("Ed25519PublicKey always has correct size")
```
DSharifi: These expects can be scary if near_sdk does a change to their representation.

Do they not have any safe constructors?

crates/contract-interface/src/types/conversions/near.rs
```diff
@@ -0,0 +1,281 @@
...
+    }
+}
+
+impl From<&Ed25519PublicKey> for near_sdk::PublicKey {
+    fn from(pk: &Ed25519PublicKey) -> Self {
+        near_sdk::PublicKey::from(pk.clone())
```
DSharifi: Same here about cloning

Issue #2208, opened 2026-02-23
Use strong typing for thresholds sent to the thresholds signatures algorithms


netrome:
### Background

Currently we pass `threshold` parameters as `usize` when interfacing with the signing and DKG algorithms. However, we've introduced stronger types on the crypto side allowing us to be more specific about which parameter we're providing. We should explicitly pass either reconstruction bounds or max malicious parameters as thresholds to the signature schemes.

### User Story

As a developer I want to use strong types to reduce the risk of accidentally passing the wrong threshold to the signing schemes.

### Acceptance Criteria

We're using dedicated threshold types when interacting with the threshold-signature algorithms

PR #2279, opened 2026-03-02
refactor: use strong threshold types in signature provider layer

anodar:
Propagates threshold as ReconstructionLowerBound instead of usize, converts into MaxMalicious for robust_ecdsa.

Closes: https://github.com/near/mpc/issues/2208

### DIFF
--- a/crates/node/src/providers/robust_ecdsa.rs
+++ b/crates/node/src/providers/robust_ecdsa.rs
@@ -295,7 +295,8 @@ mod tests {
         let max_size = 30;
         for threshold in 5..max_size {
             for number_of_participants in threshold..max_size {
-                let number_of_signers = get_number_of_signers(threshold, number_of_participants).unwrap();
+                let number_of_signers =
+                    get_number_of_signers(threshold, number_of_participants).unwrap();
                 let new_threshold = translate_threshold(threshold, number_of_participants)
                     .unwrap()
                     .value();
@@ -312,7 +313,8 @@ mod tests {
         let max_size = 30;
         for threshold in 5..max_size {
             for number_of_participants in threshold..max_size {
-                let number_of_signers = get_number_of_signers(threshold, number_of_participants).unwrap();
+                let number_of_signers =
+                    get_number_of_signers(threshold, number_of_participants).unwrap();
                 assert!(threshold <= number_of_signers && number_of_signers <= number_of_participants, "Failed for threshold={threshold}, number_of_participants={number_of_participants}");
             }
         }


crates/node/src/providers/ckd/key_resharing.rs
```diff
@@ -21,7 +22,7 @@ impl CKDProvider {
         let new_keyshare = KeyResharingComputation {
             threshold: new_threshold,
             old_participants: old_participants.participants.iter().map(|p| p.id).collect(),
-            old_threshold: old_participants.threshold as usize,
+            old_threshold: ReconstructionLowerBound::from(old_participants.threshold as usize),
```
netrome: Unrelated to this PR, but we should probably do something about this `as` conversion. While in practice this is fine, as a general rule of thumb we try to get rid of these conversions in favor of explicit conversions and error handling.

crates/node/src/providers/robust_ecdsa.rs
```diff
@@ -262,10 +268,10 @@ pub(super) fn get_number_of_signers(threshold: usize, _number_of_participants: u
 pub(super) fn translate_threshold(
     threshold: usize,
     number_of_participants: usize,
-) -> anyhow::Result<usize> {
+) -> anyhow::Result<MaxMalicious> {
```
netrome: Unfortunately, I think we'll have to keep this method untyped and return a `usize`, because when we call this in `run_key_generation_client`, `threshold` represents the reconstruction bound. But when we call this method in the robust ECDSA provider, the output is `MaxMalicious`.

Though, looking deeper into this this starts to smell like a potential bug in the existing code. It looks like we're running DKG with `MaxMalicious` and not the `ReconstructionLowerBound` 🤔 

Note: this is an ugly workaround that shouldn't be needed once we support per-domain thresholds, which @SimonRastikian is currently working on.

netrome (changes requested):
Mostly looks good, but we need to sort out the `translate_threshold` function which may be an existing bug.

crates/node/src/providers/ckd/key_resharing.rs
SimonRastikian: There is a function for the structure .value() that should be used

crates/node/src/providers/ckd/sign.rs
```diff
@@ -25,7 +26,8 @@ impl CKDProvider {
     ) -> anyhow::Result<((ElementG1, ElementG1), VerifyingKey)> {
         let ckd_request = self.ckd_request_store.get(id).await?;
 
-        let threshold = self.mpc_config.participants.threshold as usize;
+        let threshold =
+            ReconstructionLowerBound::from(self.mpc_config.participants.threshold as usize);
```
SimonRastikian: same here

crates/node/src/providers/robust_ecdsa.rs
anodar: Nice catch! Should be fixed now.

crates/threshold-signatures/src/thresholds.rs
```diff
@@ -23,3 +23,10 @@ impl ReconstructionLowerBound {
         self.0
     }
 }
+
+/// Lower bound to reconstruct the secret is `MaxMalicious` + 1.
+impl From<MaxMalicious> for ReconstructionLowerBound {
+    fn from(m: MaxMalicious) -> Self {
+        Self(m.0.saturating_add(1))
```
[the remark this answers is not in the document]
anodar: I initially had + but then changed it to follow https://github.com/near/mpc/blob/main/CONTRIBUTING.md#use-safe-arithmetic-methods

Although does seem unnecessary as bound should never be that high anyway, happy to revert back to +.

crates/node/src/providers/ecdsa.rs
```diff
@@ -248,7 +249,9 @@ impl SignatureProvider for EcdsaSignatureProvider {
                     &format!("generate presignatures for domain {}", domain_id.0),
                     Self::run_background_presignature_generation(
                         self.client.clone(),
-                        self.mpc_config.participants.threshold as usize,
+                        ReconstructionLowerBound::from(
+                            self.mpc_config.participants.threshold as usize,
```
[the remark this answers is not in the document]
anodar: Same as above, it's u64 type.

netrome (dismissed):
Thanks for updating. One nit, otherwise lgtm.

netrome (dismissed):
Thanks for the update, just one nit about code organization 😅

gilcu3 (dismissed):

crates/node/src/providers/robust_ecdsa.rs
```diff
@@ -141,38 +143,42 @@ impl SignatureProvider for RobustEcdsaSignatureProvider {
...
-        threshold: usize,
+        threshold: ReconstructionLowerBound,
         channel: NetworkTaskChannel,
     ) -> anyhow::Result<Self::KeygenOutput> {
         let number_of_participants = channel.participants().len();
-        let robust_ecdsa_threshold = translate_threshold(threshold, number_of_participants)?;
-        EcdsaSignatureProvider::run_key_generation_client_internal(robust_ecdsa_threshold, channel)
-            .await
+        let robust_ecdsa_threshold =
+            translate_threshold(threshold.value(), number_of_participants)?;
```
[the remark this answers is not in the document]
SimonRastikian: Discussion moved to https://nearone.slack.com/archives/C0912BTG51T/p1772544339814719?thread_ts=1772525279.345459&cid=C0912BTG51T

crates/threshold-signatures/src/thresholds.rs
```diff
@@ -23,3 +30,14 @@ impl ReconstructionLowerBound {
...
+/// Lower bound to reconstruct the secret is `MaxMalicious` + 1.
+impl TryFrom<MaxMalicious> for ReconstructionLowerBound {
+    type Error = ThresholdError;
+
+    fn try_from(m: MaxMalicious) -> Result<Self, Self::Error> {
+        m.0.checked_add(1)
+            .map(Self)
+            .ok_or(ThresholdError::IntegerOverflow)
+    }
+}
```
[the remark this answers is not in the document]
anodar: Not at all, makes perfect sense.  Addressed it.

PR #2309, opened 2026-03-04
fix: reject wrong payload versions for foreign chain transaction

### LINKED ISSUE #2296: Validate payload_version in verify_foreign_transaction
### Background

`VerifyForeignTransactionRequestArgs` accepts `payload_version: u8`, but neither the contract nor the node validates or dispatches on it. The node unconditionally constructs `ForeignTxSignPayload::V1` regardless of the requested version. A request with any `payload_version` value is processed identically to version 1.

PR #2254 hardcodes `DEFAULT_PAYLOAD_VERSION` in the SDK, but callers bypassing the SDK can still submit arbitrary values that are silently ignored. When future payload versions are introduced, this creates a risk of silent payload format mismatch.

### Acceptance Criteria

- Contract rejects requests with unsupported `payload_version` values (anything other than 1):
  ```rust
  if request.payload_version != 1 {
      env::panic_str("unsupported payload_version; only version 1 is currently supported");
  }
  ```
- Node rejects unsupported `payload_version` values instead of silently falling back to V1:
  ```rust
  let payload = match foreign_tx_request.payload_version {
      1 => ForeignTxSignPayload::V1(ForeignTxSignPayloadV1 { request: request.clone(), values }),
      v => anyhow::bail!("unsupported payload_version: {v}"),
  };
  ```

### Context

[RG-F1400](https://github.com/near/mpc-private/blob/6398193cc85a9664f911147e5b517ebf2854562e/vulnerability_audit/findings/valid/RG-F1400-payload-version-ignored.md) (Low severity)

gilcu3:
Closes #2296

### DIFF
--- a/crates/contract-interface/src/types/foreign_chain.rs
+++ b/crates/contract-interface/src/types/foreign_chain.rs
@@ -1087,4 +1087,38 @@ mod tests {
         // Then
         assert_ne!(hash_a, hash_b);
     }
+
+    #[rstest]
+    #[case(ForeignTxPayloadVersion::V1, 1)]
+    fn foreign_tx_payload_version__serializes_as_u8(
+        #[case] version: ForeignTxPayloadVersion,
+        #[case] expected: u8,
+    ) {
+        assert_eq!(
+            serde_json::to_value(version).unwrap(),
+            serde_json::json!(expected)
+        );
+        assert_eq!(borsh::to_vec(&version).unwrap(), vec![expected]);
+    }
+
+    #[rstest]
+    #[case(1, ForeignTxPayloadVersion::V1)]
+    fn foreign_tx_payload_version__deserializes_from_u8(
+        #[case] input: u8,
+        #[case] expected: ForeignTxPayloadVersion,
+    ) {
+        let json: ForeignTxPayloadVersion =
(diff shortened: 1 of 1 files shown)


DSharifi (changes requested):
Thanks for addressing the bug.

Like we discussed on Slack, let's enforce this with  numbered enums, so invalid payload versions become inconstructible, instead of runtime checks and avoid the added fallibility/panic checks.

crates/contract-interface/src/types/foreign_chain.rs
```diff
@@ -1043,4 +1087,38 @@ mod tests {
...
+    }
+
+    #[rstest]
+    #[case(0)]
+    #[case(2)]
+    fn foreign_tx_payload_version__rejects_unknown_version(#[case] input: u8) {
+        serde_json::from_value::<ForeignTxPayloadVersion>(serde_json::json!(input)).unwrap_err();
+        borsh::from_slice::<ForeignTxPayloadVersion>(&[input]).unwrap_err();
+    }
 }
```
netrome: Thanks for adding these! 🙏

netrome (approved):
Looks great now, thanks!

PR #2332, opened 2026-03-06
feat: allow configuration files for full config of the mpc node

### LINKED ISSUE #2331: make mpc binary configurable with a signle configuration file
### Background

Currently the MPC binary requires many environment variables and CLI flags to run. This is quite problematic, due to these flags having to be set dynamically by the TEE launcher.

When the launcher sets these flags it has to build a docker command string, but also safeguard against string injection attacks since these env variables are all appeneded to the docker command string.

We should change the CLI args of the mpc binary to accept a single configuration file, which the launcher just forwards to the MPC binary. That way we can:
- still dynamically change config values without launcher hash changing
- avoid runtime checks and manual sanitization of env variables which is brittle
- organize all configuration values in one place, instead of having

### User Story

_No response_

### Acceptance Criteria

~~- The launcher does not forward any dynamically added docker env flags.~~
- The MPC binary accepts a configuration file for all it's configuration values.

_No response_

DSharifi:
closes #2331

### DIFF
--- a/pytest/common_lib/shared/mpc_node.py
+++ b/pytest/common_lib/shared/mpc_node.py
@@ -131,6 +131,15 @@ class MpcNode(NearAccount):
             for file_path in pathlib.Path(self.home_dir).glob(pattern):
                 file_path.unlink()
 
+    @staticmethod
+    def _strip_none(obj):
+        """Recursively remove keys with None values since TOML has no null."""
+        if isinstance(obj, dict):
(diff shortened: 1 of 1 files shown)


DSharifi:
@claude review

pbeza (approved):
That’s quite a bit of config-related heavy lifting. Good job! 💪🏼 I only did a relatively shallow review, but it LGTM. 👍

PR #2348, opened 2026-03-09
fix: threshold signatures sensitive types print custom debug message

### LINKED ISSUE #2347: Debug should make sure not to leak secrets
### Background

Deriving debug might cause the leakage of secret cryptographic material.
It is important to be able to control what we print as Debug. Some structure should not simply derive debug (especially if they could deal with secret material). For others it's fine. 

### User Story

_No response_

### Acceptance Criteria

* Create a tailored debug for the struct Action which could handle secret material 
* KeygenOutput contains secret key and thus should not derive debug as is
* Double check the leftover structures that derive debug and make sure they are safe 

_No response_

SimonRastikian:
Closes #2347


gilcu3:
> @gilcu3 I feel my redaction makes sense, however this is breaking the test: ` test confidential_key_derivation::ciphersuite::tests::test_hash_to_scalar ... FAILED` What do you suggest, do we delete this test or do we take down the redaction from scalarWrapper?

The test is very important to keep. We should find a way to keep it without taking out the redaction, for example using the Debug capability of the inner type.

PR #2386, opened 2026-03-10
fix: do not crash when vectors received are smaller than expected

### LINKED ISSUE #2353: Unchecked vector indexing on network-deserialized data causes panic in triple generation
### Background

In `crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/triples/generation.rs`, the `do_generation_many` function deserializes vectors from network messages via `chan.recv()` and `recv_from_others()`, then indexes into them with `vec[i]` for `i in 0..N` without bounds checking. A malicious participant can send a vector shorter than `N` elements, causing an out-of-bounds panic on all honest nodes simultaneously.

There are 6 affected sites across rounds 0, 2, 3, 4, 5, and 6.

From audit finding [RG-F1006](https://github.com/near/mpc-private/blob/main/vulnerability_audit/findings/valid/RG-F1006-unchecked-vector-indexing-panic.md).

### Acceptance Criteria

- Add length validation immediately after deserialization at all 6 sites in `do_generation_many`, returning a `ProtocolError` on mismatch instead of panicking
- For `PolynomialCommitmentsMessageMany` (round 2), validate all 6 `Vec` fields

gilcu3:
Closes #2353

### DIFF
--- a/crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/triples/generation.rs
+++ b/crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/triples/generation.rs
@@ -119,13 +119,6 @@ struct PolynomialCommitmentsMessageMany<const N: usize> {
     phi_proof1_v: [dlog::Proof<Secp256K1Sha256>; N],
 }
 
-/// Converts a `Vec<T>` to `[T; N]`, returning a `ProtocolError` on length mismatch.
-fn vec_to_array<T, const N: usize>(v: Vec<T>) -> Result<[T; N], ProtocolError> {
-    v.try_into().map_err(|v: Vec<T>| {
-        ProtocolError::Other(format!("expected vector of length {N}, got {}", v.len()))
-    })
-}
-
 #[allow(clippy::too_many_lines)]
 async fn do_generation_many<const N: usize>(
     comms: Comms,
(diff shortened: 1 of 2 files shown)


kevindeforth (dismissed):
Thank you!
It's quite a bit of boilerplate, did you consider introducing a helper? Something like:

```rust
verify_len::<(Vec<_>, Vec<>>)>(a_i_j_v, b_j_i_v)?;
```

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/triples/generation.rs
```diff
@@ -94,15 +95,35 @@ struct ParallelToMultiplicationTaskOutputMany {
...
+    #[serde_as(as = "[_; N]")]
+    phi_proof1_v: [dlog::Proof<Secp256K1Sha256>; N],
+}
+
+/// Converts a `Vec<T>` to `[T; N]`, returning a `ProtocolError` on length mismatch.
+fn vec_to_array<T, const N: usize>(v: Vec<T>) -> Result<[T; N], ProtocolError> {
+    v.try_into().map_err(|v: Vec<T>| {
+        ProtocolError::Other(format!("expected vector of length {N}, got {}", v.len()))
+    })
 }
```
netrome: Three nits:

1. I'd prefer to define this after `do_generation_many` used since it's a small helper function and less significant, which makes it easier to read the code top to bottom.
2. It would be nice with a dedicated error variant here instead of reusing the big `ProtocolError` type.
3. (style opinion) this would look nice as an extension trait `trat ToArray` with a `.to_array()` method.

netrome (approved):

Issue #2341, opened 2026-03-08
Introduce state viewer functionality


kevindeforth:
### Background

Related to https://github.com/near/mpc/issues/2112

### User Story

as a dev, I like isolated PRs

### Acceptance Criteria

Create the chain-gateway crate and expose the `ContractStateSubscriber` trait

PR #2415, opened 2026-03-11
refactor: Use unit type instead of custom `NoArgs` type

netrome:
suggestion for #2341 - using unit type instead of the custom `NoArgs` type.


crates/chain-gateway/src/state_viewer/traits.rs
```diff
@@ -150,7 +150,7 @@ impl<V: ViewRaw + Clone> SubscribeToContractMethod for V {
     where
         T: DeserializeOwned + Send + Clone,
     {
-        ContractMethodSubscription::new(self.clone(), contract, view_method, b"{}".to_vec())
+        ContractMethodSubscription::new(self.clone(), contract, view_method, b"null".to_vec())
```
kevindeforth:
```suggestion
        ContractMethodSubscription::new(self.clone(), contract, view_method, &())
```

kevindeforth (commented):
I agree with Claude\s reservations about this - I am unsure this will reliably work and is different to how we do it currently in the repo.

If you don't mind, I would prefer to keep the current NoArgs {} struct, as it's a drop-in replacement for our current logic.

For json args, this might work, but a contract method can also expect a borsh-serialized value as input, at which point, I am unsure this will reliably work.

PR #2417, opened 2026-03-11
fix: improving security by adding zeroization to the secret cryptography material

### LINKED ISSUE #2416: Missing zeroization in threshold signatures
### Background

It would be safer to introduce zeroization for cryptographic types that could deal with secret material.

### User Story

_No response_

### Acceptance Criteria

Add zeroization for:
Polynomial<C>
dlog::Witness<C>
dlogeq::Witness<C>
Randomness
PresignArguments
Shares
PresignOutput
TranscriptRng

_No response_

SimonRastikian:
Closes #2416

### DIFF
--- a/crates/threshold-signatures/src/frost/eddsa/sign.rs
+++ b/crates/threshold-signatures/src/frost/eddsa/sign.rs
@@ -194,12 +194,12 @@ async fn do_sign_coordinator_v2(
     threshold: ReconstructionLowerBound,
     me: Participant,
     keygen_output: KeygenOutput,
-    mut presignature: PresignOutput,
+    presignature: &PresignOutput,
     message: Vec<u8>,
 ) -> Result<SignatureOption, ProtocolError> {
     // --- Round 1
     let signing_package = frost_ed25519::SigningPackage::new(
-        std::mem::take(&mut presignature.commitments_map),
+        presignature.commitments_map.clone(),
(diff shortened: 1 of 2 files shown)


crates/threshold-signatures/src/crypto/polynomials.rs
```diff
@@ -17,6 +18,15 @@ pub struct Polynomial<C: Ciphersuite> {
...
 
+impl<C: Ciphersuite> Zeroize for Polynomial<C>
+where
+    Scalar<C>: Zeroize,
+{
+    fn zeroize(&mut self) {
+        self.coefficients.zeroize();
+    }
+}
+
```
gilcu3: this will not zeroize unless explicitly called. The more robust solution is to derive `ZeroizeOnDrop`. I tried doing that here and it doesn't work, so it is something we need to figure out here. Also calling zeroize in a vector is only best-effort, so if we don't manage we might have to live with this one

crates/threshold-signatures/src/crypto/proofs/strobe_transcript.rs
```diff
@@ -98,6 +98,7 @@ impl Transcript {
     }
 }
 
+#[derive(Zeroize, ZeroizeOnDrop)]
```
gilcu3: no need to Zeroize

PR #2437, opened 2026-03-13
chore: force usage of constructor for PermanentKeyshareData

### LINKED ISSUE #1217: force usage of constructor for `PermanentKeyshareData`
### Background

The node makes a lot of assumptions about the order of shares in `PermanentKeyshareData` (c.f. https://github.com/near/mpc/pull/1215)

By forcing the use of the constructor, we can reduce the chance of bugs occuring due to code assumptions not being met.

https://github.com/near/mpc/blob/e64dcba9853470f54d1a405286c9125ef830d37b/crates/node/src/keyshare/permanent.rs#L10

### User Story

As a dev, I want to avoid implied behavior that can be easily missed. 

### Acceptance Criteria

Make the fields inside `PermanentKeyshareStorage` private.

_No response_

gilcu3:
Closes #1217

The issue was expanded to cope with ( see the comments below):

> Unfortunately, this PR isn't helping in decoupling the node and the contract behavior.
> Ideally, we want to get rid of the assumption that keyshares stored on the node follow the same order with which they are stored in the contract, and that the resharing of keys happen in the exact same order in which the keys are stored in the KeyshareStorage.


crates/node/src/keyshare/permanent.rs
```diff
@@ -21,7 +21,18 @@ impl PermanentKeyshareData {
...
+        self.epoch_id
+    }
+
+    pub fn keyshares(&self) -> &[Keyshare] {
+        &self.keyshares
+    }
+
+    pub fn into_keyshares(self) -> Vec<Keyshare> {
+        self.keyshares
+    }
```
kevindeforth: I think we need to add some sort of sorting mechanism.
Either in the getters or in the setters or both.

crates/node/src/keyshare.rs
```diff
@@ -388,16 +443,21 @@ impl KeyshareStorage {
                 .ok_or_else(|| anyhow::anyhow!("missing keyshare {:?}", key_id))?;
             new_keyshares.push(keyshare);
         }
+        // Sort by domain_id for PermanentKeyshareData::new() invariant
+        new_keyshares.sort_by_key(|k| k.key_id.domain_id);
```
kevindeforth: same here, we can replace lines 418-447 with:
```rust
let mut domains: Vec<_> = contract_keyset.domains.iter().collect();
domains.sort_by_key(|d| d.domain_id);
let permanent = self._load_matching_from_permanent(contract_keyset).await?;
let mut new_keyshares = Vec::with_capacity(domains.len());
for domain in domains {
    if let Some(p) = &permanent {
        if let Some(keyshare) = p.keyshare_by_domain_id(domain.domain_id) {
            new_keyshares.push(keyshare.clone());
            continue;
        }
    }

    let key_id = KeyEventId::new(
        contract_keyset.epoch_id,
        domain.domain_id,
        domain.attempt,
    );

    let keyshare = self
        .temporary
        .load_keyshare(key_id)
        .await?
        .or_else(|| backup.iter().find(|share| share.key_id == key_id).cloned())
        .ok_or_else(|| anyhow::anyhow!("missing keyshare {:?}", key_id))?;

    new_keyshares.push(keyshare);
}

```

kevindeforth (approved):
Thank you, only nits!

crates/node/src/keyshare.rs
```diff
@@ -282,30 +323,31 @@ impl KeyshareStorage {
...
     }
 
-    /// Helper function to verify that the keyshares we have from permanent storage is a prefix
-    /// of the expected keyset, i.e. there are no extra keyshares, and each keyshare matches the
-    /// keyset entry at the same index.
-    fn verify_existing_keyshares_are_prefix_of_keyset(
+    /// Helper function to verify that each existing keyshare matches a corresponding entry
+    /// in the expected keyset by `domain_id` (order-independent).
+    fn verify_existing_keyshares_match_keyset(
         existing_keyshares: &[Keyshare],
```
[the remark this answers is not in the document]
kevindeforth: yeah, I don't think it is relevant perfornance wise, because we don't hold many keyshares. I would be more interested in the guarantees the type system could provide us.
But in any case, I think it's out of scope.

Issue #2443, opened 2026-03-13
Add design doc in the repo about pytest deprecation


anodar:
### Background

Rough overview of the components we're building (MPC orchestration, node process management etc.), and how we intend to use them. Scaffolding traits or example functions

### Acceptance Criteria

Have a pytest migration design doc in the repo with rough overview.

PR #2446, opened 2026-03-14
docs: Add design doc for Rust E2E test infrastructure (pytest deprecation)

anodar:
Closes: https://github.com/near/mpc/issues/2443

### DIFF
--- a/crates/contract/tests/sandbox/utils/consts.rs
+++ b/crates/contract/tests/sandbox/utils/consts.rs
@@ -39,7 +39,7 @@ pub const GAS_FOR_VOTE_BEFORE_THRESHOLD: Gas = Gas::from_tgas(5);
 /// Maximum gas expected for the threshold vote that triggers the contract update.
 /// This vote is more expensive because it deploys the new contract code and executes
 /// the migration function.
-pub const MAX_GAS_FOR_THRESHOLD_VOTE: Gas = Gas::from_tgas(178);
+pub const MAX_GAS_FOR_THRESHOLD_VOTE: Gas = Gas::from_tgas(180);
 
 /* --- Deposit constants --- */
 /// This is the current deposit required for a contract deploy. This is subject to change but make


docs/pytest-deprecation.md
```diff
@@ -0,0 +1,663 @@
...
+    /// Sends SIGTERM (gentle=true) or SIGKILL (gentle=false) to the process.
+    /// Equivalent of Python MpcNode.kill(gentle).
+    pub fn kill(&mut self, gentle: bool) -> anyhow::Result<()>;
+
+    /// Kill then start. New process, same config and data directory.
+    /// Equivalent of Python MpcNode.restart().
+    pub fn restart(&mut self, gentle: bool) -> anyhow::Result<()>;
+
+    /// Returns true if the child process is still running.
+    pub fn is_running(&mut self) -> bool;
```
netrome: Some design thoughts, I'd spontaneously like to leverage the type system a bit more strongly here. I'd consider making `start` a constructor that returns an `MpcNode` and have `kill` take ownership of self, so that during normal operations one could always assume that `MpcNode` is a handle to a running process. What do you think?

docs/pytest-deprecation.md
```diff
@@ -0,0 +1,663 @@
...
+pub struct NearBlockchain {
+    /// The RPC URL (e.g. "http://localhost:3030" or "https://rpc.testnet.near.org").
+    rpc_url: String,
+    /// RPC client for sending transactions and queries.
+    client: near_api::JsonRpcClient,
+    /// The deployed MPC signer contract account ID.
+    contract_id: Option<AccountId>,
+    /// Signer for the account that deploys and manages the contract.
+    signer: near_api::Signer,
+}
```
netrome: Given that we have a signer account for the MPC contract this isn't really just a `NearBlockchain`. What do we want to represent with this? `ContractAccountHandle`?

docs/pytest-deprecation.md
```diff
@@ -0,0 +1,663 @@
...
+├─────────────────────────────────────────────────┤
+│              NearBlockchain                       │
+│  RPC client for any NEAR network:                │
+│  - Account creation + key management             │
+│  - Contract deployment                           │
+│  - Transaction submission + polling              │
+│  - View method calls (contract state)            │
+│  Same code → localhost:3030 or testnet RPC       │
+└─────────────────────────────────────────────────┘
+```
```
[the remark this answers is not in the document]
anodar: Changed to mermaid chart.

docs/pytest-deprecation.md
```diff
@@ -0,0 +1,695 @@
...
+user journes covered end-to-end as close to production as possible.
+---
+
+## Implementation order
+
+1. **`MpcNode`** — Node process manager. Start/stop/kill a single mpc-node binary,
+   write config TOML, scrape metrics. This is the foundation everything else builds on.
+   *([#2440](https://github.com/near/mpc/issues/2440))*
+
+2. **`NearBlockchain`** — RPC client wrapper using `near-api-rs`. Deploy contract,
```
gilcu3: it seems https://github.com/r-near/near-kit-rs might be another option here

PR #2457, opened 2026-03-16
chore: create interface crate for types shared between rust launcher and node

### LINKED ISSUE #2456: create interface crate for the rust launcher to reuse
### Background

The rust launcher should be able to reuse types and paths in the mpc binary. We should break these interfaces out into a separate crates so the rust launcher and the node binary can both depend on it.

### User Story

_No response_

### Acceptance Criteria

we have a new crate that defines the structure of allowed image hashes

_No response_

DSharifi:
closes #2456

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4985,7 +4985,7 @@ dependencies = [
 
 [[package]]
 name = "launcher-interface"
-version = "3.6.0"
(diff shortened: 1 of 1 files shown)


SimonRastikian (approved):
Thanks!

PR #2507, opened 2026-03-18
ci: more cpu and ram for cargo tests

DSharifi:


gilcu3:
@DSharifi I believe what is happening here is that warpbuilds is not giving a proper runner with 32 cores as they promise. So 16 might be the best we can get. That or their 32x runners are very faulty. Still having CI un under 10 minutes seems good enough :partying_face:

PR #2518, opened 2026-03-19
ci: move reproducible build of contract as separate task

### LINKED ISSUE #2519: CI: don't use reproducible build of contract for pytests
### Background

reproducible builds of the contract is quite slow, as it runs the build in a separate container, with cold cache. This is currently a bottleneck for pytests on CI.

We should move the repro build as a separate job, and use non-repro for pytests to speed up the pyttests CI job.

### User Story

_No response_

### Acceptance Criteria

We should move the repro build as a separate job, and use non-repro for pytests to speed up the pyttests CI job.

_No response_

DSharifi:
closes #2519 

Building the contract is a bottleneck requirement to start the pytests. This PR moves the reproducible build of the contract as a separate job, and uses normal cargo build + wasm-opt for building for the pytests.

The building of the contract for pytests:

Before: ~1m 36s
After: ~19s (with a cold cache)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -292,19 +292,7 @@ jobs:
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           persist-credentials: false
-
-      - name: Cache Rust dependencies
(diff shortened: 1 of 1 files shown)


DSharifi:
@claude review

PR #2522, opened 2026-03-19
docs: designing domain separation

### LINKED ISSUE #1649: Design: Contract may allow individual thresholds per domain
#### Current State
Currently, the purpose of `DomainId` in the contract is to serve as a unique identifier for a private key that is secret shared across parties:
https://github.com/near/mpc/blob/45ef47a72f6f9dc10a3a418ceedecc106b7a5680/crates/contract/src/primitives/domain.rs#L12

The `KeyEventId` struct serves as a unique identifier for a set of secret shares that, when combined, result in the private key identified by `DomainId`:
https://github.com/near/mpc/blob/45ef47a72f6f9dc10a3a418ceedecc106b7a5680/crates/contract/src/primitives/key_state.rs#L76-L80

`EpochId` identifies a specific set of participants, whereas `AttemptId` identifies a specific computation that resulted in the keyshares.

The struct `DomainConfig` specifes the signature scheme for a domain:
https://github.com/near/mpc/blob/45ef47a72f6f9dc10a3a418ceedecc106b7a5680/crates/contract/src/primitives/domain.rs#L53-L56

https://github.com/near/mpc/blob/45ef47a72f6f9dc10a3a418ceedecc106b7a5680/crates/contract/src/primitives/domain.rs#L38-L42

#### Problem Statement
We are encountering the following limitations with this design:
- The current design implicitly assumes that we have a single protocol / algorithm for each `SignatureScheme`.
- The names of the enum values of `SignatureScheme` (which identify curves), imply that we have a single protocol scheme for each curve.
- The contract assumes all domains adhere to the same threshold, which is specified in `ThresholdParameters`.

https://github.com/near/mpc/blob/45ef47a72f6f9dc10a3a418ceedecc106b7a5680/crates/contract/src/primitives/thresholds.rs#L33-L36

Those assumptions are invalidated with the addition of robust ecdsa. With that addition, we have:
- different protocols over the same curve
- protocols may have different cryptographic thresholds
- the governance threshold of the DAO consisting of the participants in the MPC contract may differ from the cryptographic thresholds

#### Proposition (WIP)
If the contract aims to be a reference point for the cryptographic and governance configuration of the network, something like the following would be a more suitable choice (**wip, please leave suggestions on how to improve it in the comments**):

```rust
pub enum Curve {
    Secp256k1,
    Ed25519,
    Bls12381,
}

pub enum Protocol {
     CaitSith,
     Frost,
     ConfidentialKeyDerivation,
     DamgardEtAl, // the new robust ecdsa
}

/// specifies the domain configuration
pub struct KeyConfig {
    scheme: Scheme,
    curve: Curve,
    /// number of shares required to reconstruct the secret key
    reconstruction_threshold: ReconstructionThreshold, 
}

/// maps domains to their protocol configuration
pub struct KeyConfigs {
    config: IterableMap<KeyConfigId, KeyConfig>,
    next_config_id: u64,
}

/// specifies the current participant set and the voting threshold in matters of governance
pub struct  GovernanceBody {
    participants: Participants,
    voting_threshold: VotingThreshold,
}

 pub struct ReconstructionThreshold(pub u64); 
 pub struct KeyConfigId(pub u64); 
 pub struct VotingThreshold(pub u64); 
```

This design would give maximum flexibility at the cost of requiring additional validation logic. 

For example, we currently only support pairing `CaitSith` with `Secp256k1`, `Frost` with `Ecdsa`, `Ckd` with `BLS12381` and `DamgardEtAl` with `Secp256k1`.

Additionally, in case of changing protocol parameters, validation should take place that the desired `reconstruction_bounds` are achievable. For example, it should not be possible to define a `reconstruction_bound` of 10 if the current number of participants is 9.
Protocol-specific validation would need to take place. For example, in the case of `DamgardEtAl`, we require `2*reconstruction_bound-1 <= num_participants.`

suggestion, **(heavy wip)**

```rust

impl KeyConfigs {
    fn validate_resharing(self, new_reconstruction_threshold: usize, num_participants_retained: usize) -> bool {
        // returns true if the resharing is possible
    },
    fn min_max_reconstruction_threshold(self, num_new_participants: usize, num_participants_retained: usize) -> Option<(usize, usize)> {}
    fn min_max_number_of_participants_threshold(self, num_reconstruction_threshold: usize) -> Option<(usize, usize)> {}
}
```

### User Story

- As a user of chain signatures, I would like to be able to determine cryptographic and governance thresholds of the MPC network by reading the contract state.
- As a developer of the MPC network, I would like to add new protocol schemes without requiring major contract refactors.

### Acceptance Criteria

Decide on a sensible design for storing protocol and governance information.

https://nearone.slack.com/archives/C0912BTG51T/p1765374628931179

SimonRastikian:
Closes #1649


docs/design/domain-separation.md
```diff
@@ -0,0 +1,1021 @@
+# Domain Separation: Protocol & Governance Configuration Design
+
+The addition of Robust ECDSA (aka DamgardEtAl) invalidates three assumptions in the current design:
+
+✗ There is one protocol per curve (now: both CaitSith and DamgardEtAl operate over Secp256k1).$\newline$
+✗ All domains share a single cryptographic threshold. The node already has a `translate_threshold()` hack to bridge this gap.$\newline$
```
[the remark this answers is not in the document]
netrome: Was just about to say, this doesn't render well in GitHub

Image

docs/design/domain-separation.md
```diff
@@ -0,0 +1,972 @@
...
+
+/// Returns the minimum number of participants required across all domains.
+pub fn min_participants_required(domains: &DomainRegistry) -> u64 {
+    domains.iter().map(|d| {
+        match d.key_config.protocol {
+            Protocol::DamgardEtAl => 2 * d.key_config.reconstruction_threshold.0 - 1,
+            _ => d.key_config.reconstruction_threshold.0,
+        }
+    }).max().unwrap_or(0)
+}
```
kevindeforth: nit: we might want to implemen thee as a method of the `Protocol` struct, such that we have:
```rust
pub fn min_participants_required(domains: &DomainRegistry) -> u64 {
    domains.iter().map(|d| { 
        d.key_config.protocol.min_participants_required()
    })
    .max()
    .unwrap_or(2)
}
```

docs/design/domain-separation.md
```diff
@@ -0,0 +1,972 @@
...
+    if t < min_relative {
+        return Err(Error::VotingThresholdBelowMinimumRelative);
+    }
+    Ok(())
+}
+```
+
+---
+
+## 4. Backwards-Compatible Migration Strategy
```
kevindeforth: Regarding the implementation plan, I think 9 PR's is a bit much for this change. Have you considered to compress it to 3?

1. remove the mpc node --> mpc contract dependency.
2. Implement the herein described internal changes, but map them to the existing DTO types (we can use the `into_dto_type()` methods for that). If these changs don't touch the `contract-intreface` craet, then due to step 1, we are sure to be backwards compatible.
3. Add the new structs to the `contract_interface` crate, alter the `state` endpoint and make the nodes deserialize the new types.

Issue #2482, opened 2026-03-17
Generalize Hash newtype to support Sha384Digest ([u8; 48])


barakeinav1:
### Background

As a contributor working on TEE measurement types, I want to reuse the existing Hash<T, const N: usize> newtype
  from primitives/hash.rs for Sha384Digest so that we avoid duplicating hex serialization, borsh, and schemars
  boilerplate across the codebase.

  Context

  primitives/hash.rs defines a generic Hash<T, const N: usize> newtype (source) that wraps [u8; N] with hex serde,
  borsh, and display. Currently it's only used with N=32.

  Sha384Digest in contract/src/tee/measurements.rs is a separate newtype wrapping [u8; 48] with the same traits (hex
   serde, borsh, schemars). It could be replaced with Hash<Measurement, 48> (or similar marker type) if the generic
  Hash is extended to support:

  - schemars::JsonSchema (behind abi feature, with schemars(with = "String") for arrays >32)
  - The serde(transparent) pattern used by Sha384Digest

  Scope

  - Extend Hash<T, N> in primitives/hash.rs to support N=48 and schemars
  - Replace Sha384Digest with an alias like pub type Sha384Digest = Hash<Measurement, 48>
  - Verify borsh schema compatibility (no on-chain layout change)

  Related: #2447 (review comment by @pbeza 
https://github.com/near/mpc/pull/2447#discussion_r2947009456

### User Story

As a contributor, I want Sha384Digest to reuse the generic Hash<T, N> newtype from primitives/hash.rs so that hex
  serialization, borsh, and schemars boilerplate isn't duplicated.

### Acceptance Criteria

 Acceptance Criteria:
  - Hash<T, N> supports N=48 with schemars JsonSchema (behind abi feature)                                          
  - Sha384Digest is replaced with a type alias using Hash                                                           
  - Borsh schema unchanged (no on-chain layout change)   
  - All existing tests pass

PR #2564, opened 2026-03-23
refactor: generalize `Hash32<T>` to `Hash<T, N>` and replace `Sha384Digest`

pbeza:
Address #2482

### DIFF
--- a/crates/foreign-chain-inspector/src/hash.rs
+++ b/crates/foreign-chain-inspector/src/hash.rs
@@ -3,16 +3,14 @@ pub(crate) fn parse_hex_hash<const N: usize, T: From<[u8; N]>>(
     s: &str,
 ) -> Result<T, mpc_primitives::hash::HashParseError> {
     let decoded = hex::decode(s)?;
-    let bytes: [u8; N] =
-        decoded
-            .try_into()
-            .map_err(|v: Vec<u8>| mpc_primitives::hash::HashParseError::InvalidLength {
-                expected: N,
-                got: v.len(),
-            })?;
+    let bytes: [u8; N] = decoded.try_into().map_err(|v: Vec<u8>| {
+        mpc_primitives::hash::HashParseError::InvalidLength {
+            expected: N,
+            got: v.len(),
+        }
+    })?;
     Ok(T::from(bytes))
 }
-
 
 /// Generates a 32-byte hash newtype with hex JSON serialization and `FromStr`.
 /// Unlike the primitives crate's `hash_newtype!`, this does not include borsh or schema support.


crates/contract/src/tee/measurements.rs
```diff
@@ -136,21 +135,23 @@ impl AllowedMeasurements {
...
-    pub mrtd: Sha384Digest,
-    pub rtmr0: Sha384Digest,
-    pub rtmr1: Sha384Digest,
-    pub rtmr2: Sha384Digest,
-    pub key_provider_event_digest: Sha384Digest,
+    pub mrtd: MrtdHash,
+    pub rtmr0: Rtmr0Hash,
+    pub rtmr1: Rtmr1Hash,
+    pub rtmr2: Rtmr2Hash,
+    pub key_provider_event_digest: KeyProviderEventDigestHash,
```
gilcu3: I believe the correct term for these is precisely digest. Specially in the last variable it becomes a problem, as the name contains redundant word now. It should be simply `KeyProviderEventDigest`

crates/foreign-chain-inspector/src/hash.rs
```diff
@@ -0,0 +1,76 @@
...
+                let decoded = hex::decode(s)?;
+                let bytes: [u8; 32] =
+                    decoded
+                        .try_into()
+                        .map_err(|v: Vec<u8>| mpc_primitives::hash::HashParseError::InvalidLength {
+                            expected: 32,
+                            got: v.len(),
+                        })?;
+                Ok(Self { bytes })
+            }
```
netrome: Same here, the logic could be broken out from the macro

netrome (dismissed):
Looking really good. Only nit is that I'd consider breaking out any logic from functions in macro declarations to standalone separate functions, to minimize the size of the macro declarations.

crates/foreign-chain-rpc-interfaces/Cargo.toml
```diff
@@ -7,8 +7,8 @@ license.workspace = true
 [dependencies]
 derive_more = { workspace = true }
 ethereum-types = { workspace = true }
+hex = { workspace = true }
 jsonrpsee = { workspace = true }
-mpc-primitives = { workspace = true }
```
gilcu3: does this mean that the type `Hash32` itself becomes deprecated? Where does it make sense to keep it?

crates/contract/src/tee/measurements.rs
```diff
@@ -1,9 +1,59 @@
...
+digest_newtype!(
+    /// SHA-384 digest of the RTMR2 TDX measurement.
+    Rtmr2Hash
+);
+digest_newtype!(
+    /// SHA-384 digest of the key provider event.
+    KeyProviderEventDigest
+);
+
 use crate::primitives::{key_state::AuthenticatedParticipantId, participants::Participants};
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/pull/2564/commits/9980b3cf446a46749253f7bb156c07549cab2bfc

netrome (approved):
Thanks for the updates!

PR #2567, opened 2026-03-23
fix: cleanup stale votes after resharing

### LINKED ISSUE #2381: Clean up stale `CodeHashesVotes` entries after resharing
### Background

During a resharing event, the post-resharing cleanup (`clean_non_participants` called from `CLEAN_TEE_STATUS`) removes stale entries from `stored_attestations` but does **not** touch `CodeHashesVotes.proposal_by_account`. The `count_votes` function performs an unfiltered count over all entries in the map, regardless of whether the voting accounts are still current participants.

This is inconsistent with other vote types: `ThresholdParametersVotes` is reset on each epoch transition, and `AddDomainsVotes` is filtered through `get_remaining_votes`. `CodeHashesVotes` has no equivalent cleanup.

As a result, participants who are about to be removed during resharing can pre-seed votes for a malicious Docker image hash. After resharing, those stale votes persist and count toward the quorum. In a concrete scenario with N=5 and threshold T=3: two outgoing participants pre-seed votes, and after resharing only one fresh vote from a current participant is needed to reach quorum — instead of the expected three.

### User Story

As a **contract maintainer**, I want stale `CodeHashesVotes` entries from removed participants to be cleaned up after resharing, so that only votes from current participants count toward the code-hash approval quorum.

### Acceptance Criteria

- Cleanup promis also returns stale entries from `CodeHashesVotes`
- `count_votes` only counts votes from current participants (probably requires filtering at count time).
- A test demonstrates the scenario: participants vote before resharing, are removed during resharing, and their votes do not count toward quorum after resharing.

https://github.com/near/mpc-private/pull/207

gilcu3:
Closes #2381

### DIFF
--- a/crates/contract/src/tee/measurements.rs
+++ b/crates/contract/src/tee/measurements.rs
@@ -4,7 +4,7 @@ use near_sdk::{log, near};
 use serde_with::serde_as;
 use std::collections::BTreeMap;
 
-use crate::primitives::key_state::AuthenticatedParticipantId;
+use crate::primitives::{key_state::AuthenticatedParticipantId, participants::Participants};
 
 /// Tracks votes for adding or removing OS measurements.
 /// Each participant can have at most one active vote at a time.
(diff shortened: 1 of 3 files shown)


netrome (approved):
Nice, thanks!

PR #2593, opened 2026-03-25
chore: make sure docs explain why we have deposits, handle deposits in a single function

### LINKED ISSUE #1073: Improve README and documentation of deposit checks
### Background

For `sign` and `ckd` transactions we have deposit checks, to assert that the incoming transaction have at least 1 yoctonear desposited in the transaction:
https://github.com/near/mpc/blob/3381c80084ad2caeb92c65bc2c676066a6c38abe/libs/chain-signatures/contract/src/lib.rs#L349-L367

This check is currently very obscure and hard to understand what's going on. 

The reason we have this check is due to: https://nearone.slack.com/archives/C07UW93JVQ8/p1740154744237259?thread_ts=1740117904.553419&cid=C07UW93JVQ8

### User Story

As a developer I'd like to be able to understand what the deposit mechanism does, and why it's needed when reading the code base.

### Acceptance Criteria

- Improve [`contract/README.md`](https://github.com/near/mpc/blob/main/libs/chain-signatures/contract/readme.md) with an explainer on **why** a 1 yoctonear deposit is required. There is a note about the requirement, but with no reasoning already in https://github.com/near/mpc/blob/main/libs/chain-signatures/contract/readme.md#submitting-a-signature-request
- put deposit check behind a function with a clear name and add a doc or comments to explain what the check does and why it is needed. Ideally point to the README.

_No response_

gilcu3:
Closes #1073

### DIFF
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -81,6 +81,41 @@ const MINIMUM_SIGN_REQUEST_DEPOSIT: NearToken = NearToken::from_yoctonear(1);
 /// Minimum deposit required for CKD requests
 const MINIMUM_CKD_REQUEST_DEPOSIT: NearToken = NearToken::from_yoctonear(1);
 
+/// Checks that the caller attached at least `minimum_deposit` and refunds any excess.
+///
+/// A non-zero deposit is required so that the transaction must be signed by a
+/// full-access key (or a function-call access key whose `deposit` allowance is
+/// explicitly set). This prevents a **malicious frontend** from silently
+/// submitting signature requests on behalf of a user via a restricted
+/// function-call access key, because such keys cannot attach deposits by
+/// default. In other words, requiring a deposit ensures the user (or their
+/// full-access key) explicitly authorised the call.
+///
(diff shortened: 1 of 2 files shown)


crates/contract/src/lib.rs
```diff
@@ -81,6 +81,41 @@ const MINIMUM_SIGN_REQUEST_DEPOSIT: NearToken = NearToken::from_yoctonear(1);
...
 
+/// Checks that the caller attached at least `minimum_deposit` and refunds any excess.
+///
+/// A non-zero deposit is required so that the transaction must be signed by a
+/// full-access key (or a function-call access key whose `deposit` allowance is
+/// explicitly set). This prevents a **malicious frontend** from silently
+/// submitting signature requests on behalf of a user via a restricted
+/// function-call access key, because such keys cannot attach deposits by
+/// default. In other words, requiring a deposit ensures the user (or their
+/// full-access key) explicitly authorised the call.
```
gilcu3: I learned something today, as this was very far from what I thought was the reason

Issue #2597, opened 2026-03-25
Unify duplicate `hash_newtype!` macros across `primitives` and `foreign-chain-inspector`


pbeza:
### Background

We currently have two `hash_newtype!`-style macros with the same purpose but different derive sets:

1. **`mpc-primitives`** ([`crates/primitives/src/hash.rs`](https://github.com/near/mpc/blob/main/crates/primitives/src/hash.rs)) — generates hash newtypes with `borsh`, `schemars`, and `serde` support (for on-chain / contract types):
   https://github.com/near/mpc/blob/500bcfbbc91c410feed04b306f56dd586dae78c9/crates/primitives/src/hash.rs#L14-L130
3. **`foreign-chain-inspector`** ([`crates/foreign-chain-inspector/src/hash.rs`](https://github.com/near/mpc/blob/main/crates/foreign-chain-inspector/src/hash.rs)) — lightweight `serde`-only hash newtypes for RPC / off-chain types:
   https://github.com/near/mpc/blob/500bcfbbc91c410feed04b306f56dd586dae78c9/crates/foreign-chain-inspector/src/hash.rs#L15-L72
Having two separate macros with identical names is confusing and duplicates logic.

### User Story

As a developer working on this codebase, I want a single source of truth for hash newtype generation so that I don't have to maintain two nearly-identical macros and remember which one to use.

### Acceptance Criteria

- [ ] A single `hash_newtype!` (or equivalent generic `Hash<T, N>`) macro/type exists, usable from both on-chain (`mpc-contract`) and off-chain (`foreign-chain-inspector`) crates.
- [ ] Crates that don't need `borsh` are not forced to depend on it (e.g., via feature flags or by keeping borsh derives behind a feature gate).
- [ ] The duplicate macro in `foreign-chain-inspector` is removed.
- [ ] All existing tests pass.

### Resources & Additional Notes

- Discussion originated in PR #2564: [comment](https://github.com/near/mpc/pull/2564#issuecomment-4120359213) and [follow-up](https://github.com/near/mpc/pull/2564#issuecomment-4125861752).
- PR #2596 was a spike showing what `#[macro_export]` unification would look like — it requires a `_macro_deps` re-export module (`$crate::_macro_deps::*`) so the macro can reference `borsh`, `serde`, `hex`, etc. from the caller's context. This added non-trivial boilerplate.
- A feature-flag approach (e.g., `features = ["borsh"]` on `mpc-primitives`) may be cleaner than the re-export trick, but needs investigation.

PR #2619, opened 2026-03-25
refactor: unify duplicate hash newtype macros into a single generic `HashDigest<S, N>` type in `primitives`

pbeza:
Replace three separate `hash_newtype!` / `digest_newtype!` macros (in `primitives`, `foreign-chain-inspector`, and `contract/tee/measurements`) with a single generic `HashDigest<S, N>` struct and a `define_hash!` macro in `mpc-primitives`. This eliminates code duplication across crates while supporting arbitrary byte lengths and consistent serde/borsh/schema implementations.

Addresses #2597


netrome (changes requested):
While there's probably a good reason for it, I don't see why the macro needs to generate a type alias using a generic type rather than simply just generating a new wrapper type. This would circumvent the need for any phantom data and marker types afaik as well.

PR #2625, opened 2026-03-26
feat: chain-gateway block event subscriber

### LINKED ISSUE #2587: Chain Gateway: Implement Block event subscriber
### Background

implement the block event subscriber designed in #1904 

### User Story

This is needed, such that we can make the MPC node depend on the indexer crate and reduce code surface

### Acceptance Criteria

Implement the BlockEventSubscriber outlined in the chain-gateway design doc

_No response_

kevindeforth:
resolves #2587 
Relevant files from the current mpc node:
handler: https://github.com/near/mpc/blob/main/crates/node/src/indexer/handler.rs
stats: https://github.com/near/mpc/blob/main/crates/node/src/indexer/stats.rs

Note: during implementation, we discovered that the block event API design from https://github.com/near/mpc/pull/2103 will probably want to be revised slightly. That will be a follow-up, as the current PR is a good basis for that https://github.com/near/mpc/issues/2680

### DIFF
--- a/crates/backup-cli/src/adapters/keyshare_storage.rs
+++ b/crates/backup-cli/src/adapters/keyshare_storage.rs
@@ -43,7 +43,7 @@ impl KeyShareRepository for KeyshareStorageAdapter {
         let data = self.storage.load().await?;
 
         match data {
-            Some(permanent_data) => Ok(permanent_data.keyshares().to_vec()),
+            Some(permanent_data) => Ok(permanent_data.keyshares_vec()),
             None => Ok(vec![]),
         }
     }
--- a/crates/foreign-chain-inspector/src/abstract_chain/inspector.rs
+++ b/crates/foreign-chain-inspector/src/abstract_chain/inspector.rs
@@ -99,7 +99,7 @@ pub enum AbstractExtractedValue {
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum AbstractExtractor {
     BlockHash,
-    Log { log_index: usize },
+    Log { log_index: u64 },
 }
 
 impl AbstractExtractor {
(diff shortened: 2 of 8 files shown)


docs/chain-gateway-design.md
```diff
@@ -502,26 +502,20 @@ If we want this interface to be re-usable in other parts of our code, we can cre
...
 
-/// an identifier for a subscription
-pub struct SubscriptionId(pub u64);
+/// An identifier for a subscription, returned by `subscribe()`.
+pub struct BlockEventId(pub u64);
 
-/// Filter - can be easily extended later
-pub enum SubscriptionFilter {
+/// Filter — can be easily extended later.
+pub enum BlockEventFilter {
```
gilcu3: em-dash?

The comment does not look informative to me, maybe we can be more explicit or remove

crates/chain-gateway-test-contract/src/consts.rs
```diff
@@ -0,0 +1,23 @@
...
+pub const SET_VALUE: &str = "set_value";
+// private method for setting value
+pub const PRIVATE_SET: &str = "private_set";
+// spawn a promise to set the value
+pub const SET_VALUE_IN_PROMISE: &str = "set_value_in_promise";
+// spawns a promise to set the value and spawns a callback promise to set the value
+pub const SPAWN_PROMISE_WITH_CALLBACK: &str = "spawn_promise_with_callback";
+
+/* Gas constants */
+// teragas as u64. We don't use near_sdk::Gas on purpose, such that the near indexer can re-use
```
gilcu3: for that you should use https://crates.io/crates/near-gas

crates/chain-gateway-test-contract/res/chain_gateway_test_contract.wasm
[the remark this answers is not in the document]
gilcu3: My initial point is simply it is bad to have binaries in git repos. For that reason we recompile the contracts in pytests instead of storing the wasm files, as we used to do in the past. These contracts also "never" change. Sharing the logic for building should be very easy, so we can take that in a follow up. IMO that logic is too simple to need sharing, the example above uses abi for example, but that's not needed in general.

In any case, what avoids here using the same pre-step (either contract build, or contract load) for all tests that need it? If building, it would not need any fancy lock file, and the code I shared above would be more than enough.

PR #2646, opened 2026-03-30
fix: ensure asset id origins are validated before use

### LINKED ISSUE #2002: Unique IDs from other participants are not validated
### Background

From audit 2025/02/06:

‭The MPC client does not validate asset IDs from other nodes. This means that nodes can‬
‭maliciously choose asset IDs to overwrite shared values such as Beaver triples and‬
‭pre-signatures owned by other nodes.‬
‭Since asset IDs are not validated, a malicious node is free to choose asset IDs belonging to‬
‭other nodes. If a malicious node chooses an asset ID already used by another node when a‬
‭Beaver triple or pre-signature is generated, this will cause other nodes to replace the‬
‭original asset when the new unowned asset is committed to the database.‬

### User Story

_No response_

### Acceptance Criteria

Short term, validate asset IDs generated by other nodes to ensure that the asset ID‬
‭corresponds to the participant ID and does not correspond to a previously reserved asset‬
‭ID or a previously generated asset.‬

‭Long term, investigate whether it is possible to use existing consensus primitives to have‬
‭the nodes agree on unique IDs. We also recommend the team to simplify the‬
i‭mplementation of the distributed asset storage.‬

_No response_

gilcu3:
Closes #2002


kevindeforth (commented):
This PR introduces an implicit assumption that the leader will always pick a new unique id that was not used beforehand.

With the current implementation, we are likely to respect this assumption, because the unique ids are generated based on seconds since unix epoch.

So in the event of, for example, a node migrating to a new machine and losing its owned assets, it is currently unlikely they were to request a tripe computation with the same unique id.

However, we have no guarantees that this assumption will always be respected. For example, if in the future, we replaced the `UniqeId` with a simple counter, followers would engage in asset computation, but refuse to add it to the store, leading to wasted compute cycles.

This PR goes beyond the acceptance criteria and has borrowers refuse to override assets, even if the owner of the asset dictates this.

Is there a security issue if we were to allow an owner to override owned assets?

Issue #381, opened 2025-04-30
MPC node dependency management


kevindeforth:
Modify the [contract dependency](https://github.com/Near-One/mpc/blob/a35bc500a2e29592c9aef818b5a31143e5f58715/node/Cargo.toml#L64) in the mpc nodes.

It is sub-optimal that we depend directly on the contract folder, as we might miss breaking API changes of the `ProtocolContractState`.

netrome:
Related PR https://github.com/near/mpc/pull/2831

gilcu3:
The node currently depends on the contract for tests, so we can declare this issue done.

PR #2673, opened 2026-03-31
refactor: delete v2secp256k1 and add Protocol enum

### LINKED ISSUE #2442: Add Protocol separation in the Smart Contract
### Background

To properly separate the domains and allow different threshold for different domains, we first need to allow a separation of the protocols that we make use of as in 

```rust
pub enum Curve {
    Secp256k1,
    Curve25519,
    Bls12381,
}

pub enum Protocol {
      CaitSith(Curve),
      Frost(Curve),
      ConfidentialKeyDerivation(Curve),
      DamgardEtAl(Curve),
  }
```

### User Story

A proper domain separation is essential for us to be able to introduce the Robust ECDSA scheme

### Acceptance Criteria
* Update the documentation to meet the new Protocol structure design
* Delete V2Secp256k1 from Curve
* Add the Protocol structure
* Add acceptable mappings between Protocol and Curve as follows 
```rust
impl Protocol {
    pub fn cait_sith() -> Self { Self::CaitSith(Curve::Secp256k1) }
    pub fn frost() -> Self { Self::Frost(Curve::Curve25519) }
    pub fn ckd() -> Self { Self::ConfidentialKeyDerivation(Curve::Secp256k1) }
    pub fn damgard_et_al() -> Self { Self::DamgardEtAl(Curve::Secp256k1) }
}

impl Protocol {
    pub fn curve(&self) -> Curve {
        match self {
            Self::CaitSith(c)
            | Self::Frost(c)
            | Self::ConfidentialKeyDerivation(c)
            | Self::DamgardEtAl(c) => *c,
        }
    }
}
```

_No response_

SimonRastikian:
Closes #2442 
Even though the Borsh has changed, this is a non-breaking change as the scheme V2Secp256k1 was never deployed


crates/contract/src/primitives/domain.rs
```diff
@@ -63,12 +62,58 @@ impl Default for Curve {
...
+/// Each variant is parameterized with a [`Curve`], making it possible
+/// for any protocol to operate over different curves in the future.
+#[near(serializers=[borsh, json])]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Protocol {
+    CaitSith(Curve),
+    Frost(Curve),
+    ConfidentialKeyDerivation(Curve),
+    DamgardEtAl(Curve),
+}
```
gilcu3: I didn't think we would go for each protocol supporting all curves from the beginning

crates/contract/src/primitives/domain.rs
```diff
@@ -63,12 +62,58 @@ impl Default for Curve {
...
 
+/// Identifies the threshold signature protocol.
+/// Each variant is parameterized with a [`Curve`], making it possible
+/// for any protocol to operate over different curves in the future.
+#[near(serializers=[borsh, json])]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Protocol {
+    CaitSith(Curve),
+    Frost(Curve),
+    ConfidentialKeyDerivation(Curve),
```
gilcu3: Lets just use Ckd here, the current name is very verbose
```suggestion
    Ckd(Curve),
```

gilcu3 (changes requested):
This PR is currently deleting the possibility of using `DamgardEtAl` from the contract and the node. We should aim to do this transition without losing that feature, else we are just duplicating the work of integrating it again. Also I see the design doc was changed, we should reach some consensus as to why, given we had already agreed on it.

crates/contract/src/primitives/domain.rs
SimonRastikian: I kind of like the verbosity at this layer as people coming after are not expected to have much knowledge of what CKD means

kevindeforth (commented):
Thanks a lot for updating the `Protocol` enum!

My advise is to fix https://github.com/near/mpc/issues/381 before continuing with this. As it stands right now, we are introducing a regression in the node. Resolving #381 will prevent us from doing that and make the scope of this PR much smaller.

gilcu3:
superseded by #3115 and #3085

PR #2699, opened 2026-04-02
chore: unify testing contract build steps

### LINKED ISSUE #2687: Unify contract build helpers into a common test crate
### Background

Contract WASM build and load logic is duplicated across several Rust test harnesses:

1. **Sandbox tests** (`crates/contract/tests/sandbox/utils/contract_build.rs`) — full `cargo build` + `wasm-opt` pipeline with file-lock caching and TTL-based staleness. Builds `mpc-contract` and `test-migration-contract`.
2. **E2E tests** (`crates/e2e-tests/tests/request_lifecycle.rs:21-38`) — `load_contract_wasm()` reads a pre-built WASM from disk (or `MPC_CONTRACT_WASM` env var), no build step, just documents the build commands in comments.
3. **Chain-gateway test contract** (`crates/chain-gateway-test-contract/src/lib.rs:13-15`) — `compiled_wasm()` embeds a pre-compiled WASM checked into git via `include_bytes!`, consumed by `crates/chain-gateway/tests/common/accounts.rs`.

Each implementation independently resolves the workspace root, constructs artifact paths, and either runs or documents the same `cargo build --target=wasm32-unknown-unknown --profile=release-contract` + `wasm-opt -Oz` steps.

The chain-gateway test contract also has a checked-in WASM binary (`crates/chain-gateway-test-contract/res/chain_gateway_test_contract.wasm`) that must be manually rebuilt when the source changes.

### Acceptance Criteria

- A shared Rust crate (e.g. extend `crates/test-utils` or create `crates/contract-build`) provides reusable contract build and load functions.
- Sandbox tests and e2e tests use the shared crate instead of their own implementations.
- The chain-gateway test contract loading is consolidated into the same pattern.
- The checked-in `chain_gateway_test_contract.wasm` is either removed in favor of build-on-demand, or clearly documented as intentional.
- Existing test workflows (CI, local dev) continue to work.

gilcu3:
Closes #2687

### DIFF
--- a/crates/contract/tests/abi.rs
+++ b/crates/contract/tests/abi.rs
@@ -1,12 +1,13 @@
 fn compile_project() -> (Vec<u8>, serde_json::Value) {
     let project_dir = std::path::Path::new(env!("CARGO_MANIFEST_DIR"));
+    let out_dir = project_dir.join("../../target/near/contract");
     let to_utf8 = |p: std::path::PathBuf| {
         cargo_near_build::camino::Utf8PathBuf::from_path_buf(p).expect("path must be valid UTF-8")
     };
 
     let opts = cargo_near_build::BuildOpts {
         manifest_path: Some(to_utf8(project_dir.join("Cargo.toml"))),
-        out_dir: Some(to_utf8(project_dir.join("../../target/near/contract"))),
+        out_dir: Some(to_utf8(out_dir.clone())),
         features: Some("abi".to_string()),
         profile: Some("release-contract".to_string()),
         ..Default::default()
@@ -15,10 +16,7 @@ fn compile_project() -> (Vec<u8>, serde_json::Value) {
     let contract_path = test_utils::contract_build::build_contract_path(opts);
 
     let wasm = std::fs::read(&contract_path).unwrap();
-    let abi_path = contract_path
-        .parent()
-        .unwrap()
-        .join("mpc_contract_abi.json");
+    let abi_path = out_dir.join("mpc_contract_abi.json");
     let abi_str = std::fs::read_to_string(abi_path).unwrap();
     let abi = serde_json::from_str::<serde_json::Value>(&abi_str).unwrap();
     (wasm, abi)
--- a/crates/test-utils/src/contract_build.rs
+++ b/crates/test-utils/src/contract_build.rs
@@ -1,4 +1,3 @@
-use fs2::FileExt;
 use std::path::{Path, PathBuf};
 
 /// Returns the workspace root directory.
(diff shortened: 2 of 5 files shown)


crates/test-utils/src/contract_build.rs
anodar: nit: I know your PR didn't introduce it but I think we should avoid generic names like `utils` have those utilities grouped into modules with more descriptive name that indicate their use.  I'd put this into different module and name it contract_test or something better (naming is hard). 

I'm not sure what's a best practice in Rust for it, but at least for Go it would be named something like that:  https://google.github.io/styleguide/go/best-practices#util-packages

crates/chain-gateway/tests/common/accounts.rs
```diff
@@ -23,8 +23,11 @@ impl Contract {
     }
 }
 
-pub(super) fn compiled_test_contract_wasm() -> &'static [u8] {
-    chain_gateway_test_contract::compiled_wasm()
+pub(super) fn compiled_test_contract_wasm() -> Vec<u8> {
+    test_utils::contract_build::ContractBuilder::new(
+        "crates/chain-gateway-test-contract/Cargo.toml",
+    )
+    .build()
```
kevindeforth: Will this build the contract for every test?

crates/test-utils/src/contract_build.rs
```diff
@@ -0,0 +1,96 @@
+use std::path::{Path, PathBuf};
+
+/// Returns the workspace root directory.
+fn workspace_root() -> PathBuf {
+    Path::new(env!("CARGO_MANIFEST_DIR")).join("../..")
+}
+
+/// Builds a contract and returns the path to the compiled WASM artifact.
+///
+/// Uses `cargo near build non-reproducible-wasm` under the hood. The caller
```
kevindeforth: nit: could we add the option for reproducible builds?

crates/contract/tests/abi.rs
```diff
@@ -1,19 +1,19 @@
...
+    };
 
-    let cargo_opts = cargo_near_build::BuildOpts {
-        manifest_path: Some(
-            cargo_near_build::camino::Utf8PathBuf::from_path_buf(project_path.join("Cargo.toml"))
-                .unwrap(),
-        ),
+    let opts = cargo_near_build::BuildOpts {
+        manifest_path: Some(to_utf8(project_dir.join("Cargo.toml"))),
+        out_dir: Some(to_utf8(project_dir.join("../../target/near/contract"))),
```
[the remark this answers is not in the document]
gilcu3: because else we get a collision with another contract being built in contract tests

crates/test-utils/src/contract_build.rs
gilcu3: it would change the code a little bit (not the same function in carg-near-build) and require docker, so we better tackle that in a follow-up once we plan to use them

kevindeforth (approved):
Thanks!

crates/contract/tests/sandbox/utils/consts.rs
```diff
@@ -47,6 +47,6 @@ pub const MAX_GAS_FOR_THRESHOLD_VOTE: Gas = Gas::from_tgas(180);
 /// not be getting that big.
 ///
 /// TODO(#771): Reduce this to the minimal value possible after #770 is resolved
-pub const CURRENT_CONTRACT_DEPLOY_DEPOSIT: NearToken = NearToken::from_millinear(15000);
+pub const CURRENT_CONTRACT_DEPLOY_DEPOSIT: NearToken = NearToken::from_millinear(16000);
```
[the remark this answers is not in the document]
gilcu3: the problem is that there are restrictions on contract size as well. For RPC nodes requests it is 1.5MB, for contract size itself is 4MB

PR #2707, opened 2026-04-05
test: debug Phala attestation tcb_info_issuer_chain parsing failure

barakeinav1:
## DO NOT MERGE — debugging only

Temporary debug changes to investigate the `tcb_info_issuer_chain` parsing error when uploading TDX quotes to Phala's attestation service.

### Changes
- Dump raw Phala response before parsing (length, first/last 500 chars)
- Show response context around the error column on parse failure
- Fall back to Mock attestation instead of crashing (so the node stays alive for further debugging)

### Background
All MPC nodes fail Dstack attestation with:
```
Missing or invalid field: tcb_info_issuer_chain at line 1 column 5620
```

See mpc-private#261 for full investigation.

### To test
Build and push the MPC node image from this branch, deploy in a TEE CVM, check logs.


netrome:
Closing as stale, hope the investigation went well.

PR #2714, opened 2026-04-06
feat: add local PCCS proxy to replace Phala dependency

### LINKED ISSUE #467: remove dependency on phala PCS endpoint
Currently MPC nodes are calling Phala's https://proof.t16z.com/api/upload endpoint to get TDX collateral
this may cause a centralization problem.
system: level options:
1. each node operator will be responsible to bring up his own PCS  
2. use public available PCS +backups
3. post collateral on chain.

MPC node:
Have a ability to configure the PCS Path/URL

### LINKED ISSUE #1545: Introduce independent CRL endpoint and remove dependency on Phala’s CRL endpoint
### Background

Right now our contract [relies](https://nearone.slack.com/archives/C07UW93JVQ8/p1749027135551629?thread_ts=1748938185.491209&cid=C07UW93JVQ8) on Phala’s collateral endpoint (https://proof.t16z.com/api/upload) to fetch Intel’s attestation collateral (including CRLs). This introduces an external dependency we don’t control: if Phala lags behind Intel’s CRL updates, serves stale collateral, or changes their API, our on-chain verification may incorrectly accept revoked quotes or fail in unexpected ways.

Given that Intel periodically updates their Certificate Revocation Lists — including recent revocations of fake quotes generated by the TEE.fail bot — we should not depend on a third-party service for this critical security step.

This task is to design and introduce our own independent collateral/CRL endpoint, fully under our control, and update the contract + tooling to rely on it. This endpoint should fetch, cache, verify, and expose all collateral required for SGX/TDX quote validation. We should also add unit tests using known invalid test vectors (e.g., TEE.fail quotes) to ensure the new pipeline correctly handles CRL updates.

### User Story

_No response_

### Acceptance Criteria

- **Independent Collateral Service**
  - Implement a new endpoint (NEAR-hosted) that fetches Intel collateral + CRLs directly from Intel.
  - Include basic caching, refresh logic, and error handling.
  - Update the contract and verification logic to use this endpoint instead of Phala.
  - You can probably reuse the [reference PCCS service][pccs] provided by Intel for that.

- **Security & Correctness**
  - Ensure CRLs are refreshed regularly and exposed in a format compatible with existing verification logic.
  - Define fallback behavior for stale or unavailable collateral.

- **Testing**
  - Add tests using known invalid/fake quotes (e.g., [TEE.fail bot](https://x.com/teedotfail)) to confirm revoked quotes are rejected.
  - Add integration tests covering collateral fetch, CRL updates, and end-to-end verification flow.

- **Documentation**
  - Document the new collateral/CRL pipeline, refresh frequency, and debugging workflow.
  - Describe the deprecation and removal of the Phala dependency.

- **Outcome**
  - The system no longer relies on Phala and only uses it as a fallback.
  - Intel CRL updates propagate reliably and affect attestation validation.

[pccs]: https://cc-enabling.trustedservices.intel.com/intel-tdx-enabling-guide/02/infrastructure_setup/#provisioning-certificate-caching-service-pccs

Reference PCCS service provided by Intel:
https://cc-enabling.trustedservices.intel.com/intel-tdx-enabling-guide/02/infrastructure_setup/#provisioning-certificate-caching-service-pccs

barakeinav1:
## Summary

Reference/POC PR for replacing Phala's collateral endpoint with a local PCCS proxy. This PR contains the combined implementation that was later split into:

- **#2893** — Node `quote_upload_url` config option (closes #1545)
- **#2894** — `pccs-proxy` Rust crate + Python proxy (closes #467)

## What's in this PR

- **Rust PCCS proxy** (`crates/pccs-proxy/`) — drop-in replacement for Phala's endpoint, uses `dcap-qvl` for quote parsing and `axum` for HTTP, fetches 4 Intel PCCS endpoints in parallel
- **Python PCCS proxy** (`scripts/local-pccs-proxy.py`) — zero-dependency fallback
- **Node config: `quote_upload_url`** — optional override in `[mpc_node_config]` to point at any collateral endpoint
- **rstest-parametrized test** covering both Phala and local proxy endpoints
- **Operator docs** (`docs/local-pccs-proxy.md`)

## E2E test results

Full 2-node CVM cluster tested on TDX hardware:
- [x] Both nodes fetched collateral from local PCCS proxy (confirmed in proxy logs)
- [x] Attestation accepted by contract (`get_tee_accounts` returns both nodes)
- [x] Key generation completed across 4 domains (ECDSA, EdDSA, CKD, ForeignTx)
- [x] ECDSA signature returned successfully
- [x] Also tested with Phala endpoint for comparison — both work

## Design doc

https://github.com/near/mpc-private/pull/284

Relates to #1545, #467


docs/local-pccs-proxy.md
```diff
@@ -0,0 +1,263 @@
...
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/python3 -u /opt/mpc/scripts/local-pccs-proxy.py --port 8082
+Restart=on-failure
+RestartSec=5
+
+[Install]
+WantedBy=multi-user.target
+```
```
pbeza: IMHO, this should go in a standalone file instead of being inlined in the doc.

docs/local-pccs-proxy.md
```diff
@@ -0,0 +1,263 @@
...
+  --profile test-release test_upload_quote_for_collateral_with_local_pccs_proxy
+```
+
+Override the URL via environment variable:
+
+```bash
+LOCAL_PCCS_PROXY_URL=http://localhost:9090/api/v1/attestations/verify \
+  cargo test -p tee-authority --features external-services-tests \
+  --profile test-release test_upload_quote_for_collateral_with_local_pccs_proxy
+```
```
pbeza: I’m not sure it’s a good idea to reference unit tests in the docs — they can easily get out of sync if test names change. It feels like docs should stay a few layers of abstraction above the code.

docs/local-pccs-proxy.md
```diff
@@ -0,0 +1,263 @@
...
+
+### Quote parsing fails
+
+The proxy only supports TDX v4 quotes (TEE type 0x81). SGX quotes are not supported. Check the error message in the proxy logs for details.
+
+## Related issues
+
+- [#467](https://github.com/near/mpc/issues/467) -- Remove dependency on Phala PCS endpoint
+- [#1545](https://github.com/near/mpc/issues/1545) -- Introduce independent CRL endpoint and remove dependency on Phala's CRL endpoint
```
pbeza: We probably shouldn’t refer to GitHub issues in the docs.

```suggestion
```

crates/tee-authority/src/tee_authority.rs
```diff
@@ -509,4 +509,60 @@ mod tests {
...
+    /// python3 scripts/local-pccs-proxy.py --port 8082 --pccs-url https://localhost:8081
+    ///
+    /// # Run this test:
+    /// LOCAL_PCCS_PROXY_URL=http://localhost:8082/api/v1/attestations/verify \
+    ///   cargo test -p tee-authority --features external-services-tests \
+    ///     --profile test-release test_upload_quote_for_collateral_with_local_pccs_proxy
+    /// ```
+    #[tokio::test]
+    #[cfg(feature = "external-services-tests")]
+    async fn test_upload_quote_for_collateral_with_local_pccs_proxy() {
```
pbeza: This test is almost a copy-paste of the one above. Could we maybe parametrize it with [`rstest`](https://docs.rs/rstest/latest/rstest/), or use some other way to reduce the duplication?

crates/node-config/src/start.rs
```diff
@@ -24,6 +24,11 @@ pub struct StartConfig {
     /// Node configuration (indexer, protocol parameters, etc.).
     pub node: ConfigFile,
     pub log: LogConfig,
+    /// Optional override for the TDX quote collateral endpoint URL.
+    /// When set, overrides the URL injected by the launcher in the `[tee]` section.
+    /// Example: `"http://51.68.219.1:8082/api/v1/attestations/verify"`
+    #[serde(default)]
+    pub quote_upload_url: Option<String>,
```
pbeza: Is the `[tee]` section configuration measured? If so, it might be a bit confusing that we can override things that were supposed to be measured (and thus fixed and trusted) via runtime config. WDYT?

pbeza: It would be better to use `Url` instead of `String`.

crates/pccs-proxy/src/main.rs
```diff
@@ -0,0 +1,263 @@
...
+    #[arg(long, default_value = "8082")]
+    port: u16,
+
+    /// URL of the local Intel PCCS.
+    #[arg(long, default_value = "https://localhost:8081")]
+    pccs_url: String,
+
+    /// Address to bind to.
+    #[arg(long, default_value = "0.0.0.0")]
+    bind: String,
```
pbeza: You can use more appropriate type here instead of plain `String` (like `SocketAddr`). You can also check how we handle addresses in other places with `clap`.

docs/local-pccs-proxy.md
```diff
@@ -0,0 +1,263 @@
...
+Intel PCCS (host, port 8081)
+    │
+    │  Fetches/caches from Intel PCS
+    ▼
+Intel Provisioning Certification Service (api.trustedservices.intel.com)
+```
+
+## Prerequisites
+
+- **Intel PCCS** installed and running on the host (package `sgx-dcap-pccs`, typically listening on `https://localhost:8081`)
```
pbeza: From my experience, this isn’t trivial. We could point people to a guide on how to install and configure it.

crates/pccs-proxy/src/main.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    let state = Arc::new(AppState {
+        pccs_base_url: args.pccs_url.trim_end_matches('/').to_string(),
+        http,
+    });
+
+    let app = Router::new()
+        .route("/api/v1/attestations/verify", post(verify_attestation))
+        .route("/health", get(health))
+        .layer(DefaultBodyLimit::max(1024 * 1024)) // 1MB
```
pbeza: Could you extract it into a constant?

crates/pccs-proxy/src/main.rs
```diff
@@ -0,0 +1,263 @@
...
+}
+
+#[tokio::main]
+async fn main() {
+    tracing_subscriber::fmt()
+        .with_env_filter(
+            tracing_subscriber::EnvFilter::try_from_default_env()
+                .unwrap_or_else(|_| "pccs_proxy=info".parse().unwrap()),
+        )
+        .init();
```
pbeza, line 241: Why do we need that? Is it used somewhere else too?

crates/pccs-proxy/src/main.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    let args = Args::parse();
+
+    let http = Client::builder()
+        .danger_accept_invalid_certs(true) // PCCS uses self-signed cert
+        .timeout(std::time::Duration::from_secs(30))
```
pbeza: Could you extract it into a constant?

crates/pccs-proxy/src/main.rs
```diff
@@ -0,0 +1,263 @@
...
+    #[arg(long, default_value = "https://localhost:8081")]
+    pccs_url: String,
+
+    /// Address to bind to.
+    #[arg(long, default_value = "0.0.0.0")]
+    bind: String,
+}
+
+struct AppState {
+    pccs_base_url: String,
```
pbeza: Let’s use `Url` here instead of a raw `String`.

crates/pccs-proxy/src/main.rs
```diff
@@ -0,0 +1,263 @@
...
+    /// URL of the local Intel PCCS.
+    #[arg(long, default_value = "https://localhost:8081")]
+    pccs_url: String,
+
+    /// Address to bind to.
+    #[arg(long, default_value = "0.0.0.0")]
+    bind: String,
+}
+
+struct AppState {
```
pbeza: This name feels a bit odd — it feels more like a PCCS client than application state.

crates/pccs-proxy/src/main.rs
```diff
@@ -0,0 +1,263 @@
...
+
+#[derive(Serialize)]
+struct HealthResponse {
+    status: &'static str,
+}
+
+#[derive(Serialize)]
+struct CollateralWrapper {
+    quote_collateral: CollateralResponse,
+}
```
pbeza, line 38: Why do we have that wrapper in the first place? Why not use `CollateralResponse` directly?

crates/pccs-proxy/src/main.rs
```diff
@@ -0,0 +1,263 @@
...
+        error_response(StatusCode::INTERNAL_SERVER_ERROR, e.to_string())
+    })?;
+
+    tracing::info!("Collateral returned successfully");
+    Ok(Json(CollateralWrapper {
+        quote_collateral: collateral,
+    }))
+}
+
+async fn extract_hex_field(multipart: &mut Multipart) -> anyhow::Result<String> {
```
pbeza: It’s not clear from the name that we’re actually looking for the quote’s hex.

crates/pccs-proxy/src/main.rs
```diff
@@ -0,0 +1,263 @@
...
+    while let Some(field) = multipart
+        .next_field()
+        .await
+        .map_err(|e| anyhow::anyhow!("multipart error: {e}"))?
```
pbeza, line 104: Shouldn’t we introduce proper error types to make error handling more robust? This also applies to other places in the code as well.

crates/pccs-proxy/src/main.rs
pbeza: Can we restructure the code so we don’t need `mut`?

crates/pccs-proxy/src/main.rs
```diff
@@ -0,0 +1,263 @@
...
+        tracing::error!("Failed to extract hex field: {e}");
+        error_response(StatusCode::BAD_REQUEST, e.to_string())
+    })?;
+
+    tracing::info!(quote_len = quote_hex.len(), "Received quote");
+
+    let collateral = get_collateral(&state, &quote_hex).await.map_err(|e| {
+        tracing::error!("Failed to get collateral: {e}");
+        error_response(StatusCode::INTERNAL_SERVER_ERROR, e.to_string())
+    })?;
```
pbeza, line 91: Fetching collateral on every verification request feels like a lot of overhead. We should probably cache it and only refresh it when explicitly requested (e.g., via a flag in the request) or on a timer.

scripts/local-pccs-proxy.py
```diff
@@ -0,0 +1,369 @@
...
+      "pck_crl_issuer_chain": "<PEM>",
+      "root_ca_crl": "<hex>",
+      "pck_crl": "<hex>",
+      "pck_certificate_chain": "<PEM>"
+    }
+  }
+
+Usage:
+  python3 scripts/local-pccs-proxy.py [--port 8082] [--pccs-url https://localhost:8081]
+"""
```
pbeza, line 31: Wasn’t this already implemented in [`main.rs`](https://github.com/near/mpc/blob/b193d4afbf156f432e6d0b1aaa464f9987dedcca/crates/pccs-proxy/src/main.rs)? Let’s prefer using Rust over Python for new code.

pbeza (changes requested):
A couple of nits and 2 blockers:
- https://github.com/near/mpc/pull/2714#discussion_r3072513740
- https://github.com/near/mpc/pull/2714#discussion_r3072520605

crates/pccs-proxy/src/main.rs
barakeinav1: The local Intel PCCS already caches collateral from Intel (configurable refresh schedule, currently daily via Refresh Schedule in pccs config). Our proxy is a thin pass-through, and does local call to it — it doesn't fetch from Intel directly.
 
that said, we didn't really define what is the purpose of this proxy. should it serve only nearOne node (in this case we don't have any bottlenecks) or serve other as well.

scripts/local-pccs-proxy.py
barakeinav1: I initially  create a python PCCS proxy (easier to develop), and then added a RUST version (not tested yet)

crates/node-config/src/start.rs
barakeinav1: No, the `[tee]` section is not measured. What is measured is the launcher's docker-compose file, which includes `PLATFORM=TEE` — and the launcher prevents that from being overridden (it's a reserved key).

As for `quote_upload_url` — we don't need to measure it. The collateral data itself is cryptographically signed by Intel (certificates, CRL signatures, TCB info signature). `dcap_qvl::verify::verify()` validates all of these signatures regardless of where the data was fetched from. So it doesn't matter if the collateral came from Phala, a local PCCS, or anywhere else — if the signatures don't verify against Intel's root CA, attestation fails.

crates/pccs-proxy/src/main.rs
barakeinav1: The wrapper is needed for API compatibility. Phala's endpoint returns `{ "quote_collateral": { ... } }` and the MPC node deserializes it via:

```rust
#[derive(Deserialize)]
struct UploadResponse {
    quote_collateral: Collateral,
}
```

(`crates/tee-authority/src/tee_authority.rs:160`). Removing the wrapper would change the JSON shape and break the node's deserialization.

crates/pccs-proxy/src/main.rs
barakeinav1: Axum's \`Multipart::next_field()\` takes \`&mut self\` — this is part of axum's API, not something we can avoid:

\`\`\`rust
pub async fn next_field(&mut self) -> Result<Option<Field<'_>>, MultipartError>
\`\`\`

The \`mut\` is required to iterate through the multipart fields.

scripts/local-pccs-proxy.py
barakeinav1: This is a draft/POC PR. Both implementations are included to demonstrate the approach. The Python version was written first for quick prototyping and used in the initial E2E tests. The Rust version was added afterwards as the production-ready implementation. Happy to remove the Python version if preferred.

crates/pccs-proxy/src/main.rs
barakeinav1: The caller already gets specific error messages in the JSON body (e.g. `"Failed to parse TDX quote: ..."`, `"Missing header: TCB-Info-Issuer-Chain"`, etc.), just not structured error codes.

The MPC node doesn't match on error details — it only checks for HTTP 200 vs non-200 and retries on failure:

```rust
if status != StatusCode::OK {
    bail!("Got unexpected HTTP status code: response from collateral endpoint: {:?}, expected: {:?}",
        status, StatusCode::OK);
}
```

Phala's endpoint behaves the same way (unstructured errors, no typed codes). Since this is a drop-in replacement for Phala, adding structured error types would be an API extension that no consumer uses, and would diverge from the API contract we're matching. Happy to add them if a use case comes up.

scripts/local-pccs-proxy.py
pbeza: Makes sense — let’s remove the Python version then.

crates/node-config/src/start.rs
pbeza: > As for quote_upload_url — we don't need to measure it. The collateral data itself is cryptographically signed by Intel (certificates, CRL signatures, TCB info signature). dcap_qvl::verify::verify() validates all of these signatures regardless of where the data was fetched from. So it doesn't matter if the collateral came from Phala, a local PCCS, or anywhere else — if the signatures don't verify against Intel's root CA, attestation fails.

True. The only issue is that a malicious or misconfigured PCCS can serve stale collateral (e.g., outdated CRLs) or omit updates, which could lead to accepting outdated TCB levels — but we can handle that in a follow-up.

crates/pccs-proxy/src/main.rs
pbeza: Gotcha — since it’s not obvious from just reading the code, could you add a comment explaining that?

crates/pccs-proxy/src/main.rs
pbeza: I mean, technically it’s not really “local” if we configure the PCCS via an address/URL that the proxy uses:

https://github.com/near/mpc/blob/b193d4afbf156f432e6d0b1aaa464f9987dedcca/crates/pccs-proxy/src/main.rs#L18-L20

In practice, we might still run it locally. I’ll leave it to other reviewers, but it could indeed be redundant to have two layers of caching if the connection to PCCS is stable.

crates/tee-authority/src/tee_authority.rs
barakeinav1: It was addressed in commit 46308a3b ("refactor: address PR review feedback")

docs/local-pccs-proxy.md
barakeinav1: Fixed — moved to standalone file. See [46308a3b](https://github.com/near/mpc/pull/2714/commits/46308a3b).

docs/local-pccs-proxy.md
barakeinav1: Fixed in [46308a3b](https://github.com/near/mpc/pull/2714/commits/46308a3b).

docs/local-pccs-proxy.md
barakeinav1: Fixed in [46308a3b](https://github.com/near/mpc/pull/2714/commits/46308a3b).

crates/tee-authority/src/tee_authority.rs
barakeinav1: Fixed — refactored into rstest-parametrized test. See [46308a3b](https://github.com/near/mpc/pull/2714/commits/46308a3b).

crates/node-config/src/start.rs
barakeinav1: Fixed in [46308a3b](https://github.com/near/mpc/pull/2714/commits/46308a3b).

crates/pccs-proxy/src/main.rs
barakeinav1: Fixed — merged --bind/--port into --listen SocketAddr. See [46308a3b](https://github.com/near/mpc/pull/2714/commits/46308a3b).

docs/local-pccs-proxy.md
barakeinav1: Fixed — linked to Intel's setup guide. See [46308a3b](https://github.com/near/mpc/pull/2714/commits/46308a3b).

crates/pccs-proxy/src/main.rs
barakeinav1: Fixed in [46308a3b](https://github.com/near/mpc/pull/2714/commits/46308a3b).

crates/pccs-proxy/src/main.rs
barakeinav1: Removed — using inline JSON now. See [46308a3b](https://github.com/near/mpc/pull/2714/commits/46308a3b).

crates/pccs-proxy/src/main.rs
barakeinav1: Fixed in [46308a3b](https://github.com/near/mpc/pull/2714/commits/46308a3b).

crates/pccs-proxy/src/main.rs
barakeinav1: Fixed in [46308a3b](https://github.com/near/mpc/pull/2714/commits/46308a3b).

crates/pccs-proxy/src/main.rs
barakeinav1: Fixed — renamed to PccsClient. See [46308a3b](https://github.com/near/mpc/pull/2714/commits/46308a3b).

crates/pccs-proxy/src/main.rs
barakeinav1: Fixed — renamed to extract_quote_hex. See [46308a3b](https://github.com/near/mpc/pull/2714/commits/46308a3b).

scripts/local-pccs-proxy.py
barakeinav1: Will remove Python version from the proxy PR (#2894). It stays in #2714 as reference.

crates/node-config/src/start.rs
barakeinav1: Tracked in near/mpc-private#288 for investigation.

crates/pccs-proxy/src/main.rs
barakeinav1: Already added — see the doc comment on `CollateralWrapper` at line 33-34.

crates/pccs-proxy/src/main.rs
barakeinav1: Agreed — the naming is about the typical deployment (proxy colocated with PCCS on the same host), but it works with any reachable PCCS.

barakeinav1:
Superseded by #2944, which takes a cleaner approach — calling `dcap_qvl::collateral::get_collateral()` directly from the node instead of adding a proxy service. No new binary to deploy, no extra network hop.

PR #2720, opened 2026-04-07
ci: fix rust-tee launcher failing with OOM

### LINKED ISSUE #2661: investigation: Rust launcher non-TEE CI check OOM during testnet genesis download
### Background

The new Rust launcher non-TEE CI check (`rust-launcher-nontee-check`) intermittently fails with exit code 137 (OOM) when the MPC node downloads the testnet genesis file. The Python launcher non-TEE check does not have this issue on the same `2x` runner.

### Current workaround

 The runner was also bumped from `2x` to `4x`, but OOM still occurred on `4x`.

Reduced the wait time from 60s to 15s so the check completes before the genesis download exhausts memory. - seems to solve the issue, but further investigation on the root cause should be done

### Observations during testing

- The MPC node image used is `main-9515e18` (supports TOML config via `start-with-config-file`)
- The Python launcher check uses the `start` subcommand with `MPC_ENV=testnet` and doesn't OOM
- The Rust launcher check uses `start-with-config-file` with `[near_init] chain_id = "testnet", download_genesis = true`
- Exit 137 (SIGKILL from kernel OOM killer) happened on both `2x` (~8GB) and `4x` (~16GB) runners
- Exit 139 (SIGSEGV) happened when `[near_init]` was omitted entirely (node crashed accessing uninitialized NEAR home dir)
- With `download_genesis = false` and no genesis file, the node also crashes
- The 15s wait works because the node is still alive during the genesis download phase — it hasn't OOMed yet

### Possible root causes

1. The `start-with-config-file` code path may use more memory during genesis processing than the `start` subcommand
2. The Rust launcher Docker image may be larger, leaving less memory headroom for the node
3. Different timing — the Rust launcher uses `docker compose up -d` (v2) which may have different memory characteristics than `docker run` (used by Python launcher)

### Acceptance criteria

- [ ] Investigate why the Rust launcher path OOMs but the Python launcher path doesn't
- [ ] Find a fix that allows the check to run with 60s wait on a `2x` or `4x` runner
- [ ] Or document why a shorter wait / larger runner is necessary

### References

- PR #2640 discussion: https://github.com/near/mpc/pull/2640
- Workaround commit: reduced wait to 15s, runner bumped to 4x

gilcu3:
Closes #2661 

EDIT:

The OOM was due to testnet requiring genesis download and processing, which requires ~20 GiB of RAM. Switched to mainnet to avoid the issue.

### DIFF
--- a/deployment/cvm-deployment/user-config.toml
+++ b/deployment/cvm-deployment/user-config.toml
@@ -1,5 +1,7 @@
 # Minimal config for non-TEE CI check.
 # Only needs enough for the launcher to start and pull the MPC node image.
+# Uses mainnet chain_id with embedded genesis to avoid downloading the large
+# testnet genesis file, which requires ~20 GiB of RAM.
 
 [launcher_config]
 image_tags = ["main-9515e18"]
@@ -20,12 +22,13 @@ format = "plain"
 filter = "info"
 
 [mpc_node_config.near_init]
-chain_id = "testnet"
+chain_id = "mainnet"
 boot_nodes = "ed25519:ERguu7jQuYk8pxNsRC6FdezvNsegBPva1GRGqjmtD7i2@10.10.10.10:24567"
-download_genesis = true
+download_genesis = false
+download_config = "rpc"
 
 [mpc_node_config.secrets]
-secret_store_key_hex = "BD399143F5B3126098B0EAA023A0E730BD399143F5B3126098B0EAA023A0E730"
+secret_store_key_hex = "BD399143F5B3126098B0EAA023A0E730"
 backup_encryption_key_hex = "0000000000000000000000000000000000000000000000000000000000000000"
 
 [mpc_node_config.node]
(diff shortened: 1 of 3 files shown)


deployment/cvm-deployment/user-config.toml
```diff
@@ -20,12 +22,13 @@ format = "plain"
...
-chain_id = "testnet"
+chain_id = "mainnet"
 boot_nodes = "ed25519:ERguu7jQuYk8pxNsRC6FdezvNsegBPva1GRGqjmtD7i2@10.10.10.10:24567"
-download_genesis = true
+download_genesis = false
+download_config = "rpc"
 
 [mpc_node_config.secrets]
-secret_store_key_hex = "BD399143F5B3126098B0EAA023A0E730BD399143F5B3126098B0EAA023A0E730"
+secret_store_key_hex = "BD399143F5B3126098B0EAA023A0E730"
```
gilcu3: this was a bug, we use aes 128 for this key

PR #2731, opened 2026-04-07
chore: unify test-utils in the node and ts crates

### LINKED ISSUE #2244: Unify test_utils functions
### Background

In https://github.com/near/threshold-signatures/pull/187 we brought to this repo several test functions that were used by the node. This creates a bit of duplication, as we had similar features in this repo already.

### Acceptance Criteria

The test_utils functionality is unified so that it can be reused by the node and the threshold-signatures crate.

https://github.com/near/threshold-signatures/pull/187#pullrequestreview-3398202536

### Context

Moved from https://github.com/near/threshold-signatures/issues/188

gilcu3:
Closes #2244

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,23 +65,6 @@ jobs:
         run: |
           ./scripts/check-mpc-node-docker-starts.sh --launcher
 
-  rust-launcher-nontee-check:
-    name: "MPC Rust Launcher non-tee check"
-    runs-on: warp-ubuntu-2404-x64-2x
-    timeout-minutes: 60
-    permissions:
-      contents: read
-
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
-        with:
-          persist-credentials: false
-
-      - name: Check mpc-node docker image can initialize through the Rust launcher and start
-        run: |
-          ./scripts/check-mpc-node-docker-starts.sh --rust-launcher
-
   docker-launcher-build-and-verify:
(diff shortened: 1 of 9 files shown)


crates/node/src/keyshare/test_utils.rs
```diff
@@ -20,7 +21,10 @@ pub fn generate_dummy_keyshares<R: CryptoRng + RngCore + SeedableRng + Send + 's
     attempt_id: u64,
     rng: &mut R,
 ) -> (Keyshare, Keyshare) {
-    let keyshares = TestGenerators::new(2, 2.into()).make_ecdsa_keygens(rng);
+    let keyshares: std::collections::HashMap<_, _> =
+        run_keygen::<Secp256K1Sha256, _>(&generate_participants_with_random_ids(2, rng), 2, rng)
```
pbeza: Do both `2`s refer to the number of participants? If so, it’s probably worth introducing a constant for that.

gilcu3: first one is number of participants, second must be threshold. Will introduce variables to make that explicit

crates/node/src/requests/queue.rs
```diff
@@ -1274,8 +1266,7 @@ mod tests {
     #[test_log::test]
     fn test_ckd_request_latency_debug() {
         let clock = FakeClock::default();
-        let participants =
-            into_participant_ids(&TestGenerators::new_contiguous_participant_ids(4, 3.into()));
```
DSharifi: Question.
What happened to the second parameter, `3` here?

gilcu3: it was the threshold, unused for the test

PR #2739, opened 2026-04-07
feat(contract): introduce generic voting structs

### LINKED ISSUE #1573: contract: abstract voting struct
### Background

We currently have a few places where we allow votes:
- `ProposedUpdates`
- `CodeHashesVotes` in `TeeState`
- `ThresholdParametersVotes`

for each one of these, we have custom vote logic.
We could abstract the voting logic in a generic type and ensure it is loaded lazily:

```rust
pub struct Votes<VoterId, ProposalId, Proposal> {
    pub votes: IterableMap<VoterId, ProposalId>,
    pub proposals: IterableMap<ProposalId, ProposalEntry<Proposal>>,
    pub next_id: ProposalId,
}
pub struct ProposalEntry<Proposal> {
    pub proposed: Proposal,
    pub num_votes: u64,
}
```

We would need (at least) the following methods:
```rust
impl Votes<VoterId, ProposalId, Proposal> {
    /// returns the `ProposalId`
    pub fn propose(&mut self, proposal: Proposal) -> ProposalId,
    /// returns the number of votes for [`ProposalId`]
    pub fn vote(&mut self, voter_id: VoterId, proposal_id: ProposalId) -> Result<u64>,
    /// removes any vote by `voter_id`
    pub fn remove_vote(&mut self, voter_id: VoterId) -> bool,
    /// removes all votes and proposals
    pub fn clear(&mut self)
    /// removes any vote not belonging to an element in [`BTreeSet<VoterId>`]
    pub fn retain_votes(voter_ids_to_keep: BTreeSet<VoterId>)
}
```

we might also want const iterators to the underlying `votes` and `proposals` maps.

### User Story

As a user of the mpc service, this is related to #1571 and should reduce variance for gas costs to function calls.
As a developer and maintainer of the project, I like if I don't need to maintain too much code

### Acceptance Criteria

Implement an efficient abstraction for tracking votes and proposals

note: do not deserialize UpdateEntry (e.g. contract code) when registering a vote.

kevindeforth:
resolves #1573

### DIFF
--- a/crates/contract/src/primitives/threshold_votes.rs
+++ b/crates/contract/src/primitives/threshold_votes.rs
@@ -1,0 +1,137 @@
+use super::thresholds::ThresholdParameters;
+use super::{key_state::AuthenticatedAccountId, participants::Participants};
+use near_sdk::{log, near};
+use std::collections::BTreeMap;
+
+/// Tracks votes for ThresholdParameters (new participants and threshold).
+/// Each current participant can maintain one vote.
+// TODO: Replace with Votes<AuthenticatedAccountId, ThresholdParameters> from generic_votes.rs
+// once this type is moved out of RunningContractState (which requires Clone + PartialEq + JSON).
+#[near(serializers=[borsh, json])]
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct ThresholdParametersVotes {
+    pub(crate) proposal_by_account: BTreeMap<AuthenticatedAccountId, ThresholdParameters>,
+}
+
+impl ThresholdParametersVotes {
+    /// return the number of votes for `proposal` casted by members of `participants`
+    pub fn n_votes(&self, proposal: &ThresholdParameters, participants: &Participants) -> u64 {
+        self.proposal_by_account
+            .iter()
+            .filter(|&(acc, prop)| {
+                participants
+                    .participants()
+                    .iter()
+                    .any(|(acc_id, _, _)| acc.get() == acc_id)
+                    && prop == proposal
+            })
+            .count() as u64
+    }
+    /// Registers a vote by `participant` for `proposal`.
+    /// Removes any existing votes by `participant`.
+    /// Returns the number of participants who have voted for the same proposal (including the new
+    /// vote).
+    pub fn vote(
+        &mut self,
+        proposal: &ThresholdParameters,
+        participant: AuthenticatedAccountId,
+    ) -> u64 {
+        if self
+            .proposal_by_account
+            .insert(participant, proposal.clone())
+            .is_some()
+        {
+            log!("removed one vote for signer");
+        }
+        self.proposal_by_account
+            .values()
+            .filter(|&prop| prop == proposal)
+            .count() as u64
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::ThresholdParametersVotes;
+    use crate::primitives::{
+        key_state::AuthenticatedAccountId,
+        participants::Participants,
+        test_utils::{gen_participant, gen_threshold_params},
+    };
+    use near_sdk::{test_utils::VMContextBuilder, testing_env};
+
+    #[test]
+    fn test_voting_and_removal() {
+        let mut participants = Participants::default();
+        let p0 = gen_participant(0);
+        participants.insert(p0.0.clone(), p0.1).expect("error");
+        let mut ctx = VMContextBuilder::new();
+        ctx.signer_account_id(p0.0);
+        testing_env!(ctx.build());
+        let participant =
+            AuthenticatedAccountId::new(&participants).expect("expected authentication");
+        let params = gen_threshold_params(30);
+        let mut votes = ThresholdParametersVotes::default();
+        assert_eq!(votes.vote(&params, participant.clone()), 1);
+        assert_eq!(votes.n_votes(&params, &participants), 1);
+        let params2 = gen_threshold_params(30);
+        assert_eq!(votes.vote(&params2, participant), 1);
+        assert_eq!(votes.n_votes(&params2, &participants), 1);
+        assert_eq!(votes.n_votes(&params, &participants), 0);
+
+        // new participant
+        let p1 = gen_participant(1);
+        participants.insert(p1.0.clone(), p1.1).expect("error");
+        ctx.signer_account_id(p1.0);
+        testing_env!(ctx.build());
+        let participant =
+            AuthenticatedAccountId::new(&participants).expect("expected authentication");
+        assert_eq!(votes.vote(&params, participant.clone()), 1);
+        assert_eq!(votes.n_votes(&params2, &participants), 1);
+        assert_eq!(votes.vote(&params2, participant), 2);
+        assert_eq!(votes.n_votes(&params2, &participants), 2);
+        assert_eq!(votes.n_votes(&params2, params2.participants()), 0);
+        assert_eq!(votes.n_votes(&params, &participants), 0);
+    }
+
+    #[test]
+    fn test_non_participant_votes_not_counted() {
+        // given: two participants vote for a proposal
+        let mut old_participants = Participants::default();
+        let (p0, p1) = (gen_participant(0), gen_participant(1));
+        old_participants.insert(p0.0.clone(), p0.1.clone()).unwrap();
+        old_participants.insert(p1.0.clone(), p1.1.clone()).unwrap();
+
+        let mut ctx = VMContextBuilder::new();
+        let auth_p0 = {
+            ctx.signer_account_id(p0.0);
+            testing_env!(ctx.build());
+            AuthenticatedAccountId::new(&old_participants).unwrap()
+        };
+        let auth_p1 = {
+            ctx.signer_account_id(p1.0);
+            testing_env!(ctx.build());
+            AuthenticatedAccountId::new(&old_participants).unwrap()
+        };
+
+        let params = gen_threshold_params(30);
+        let mut votes = ThresholdParametersVotes::default();
+        votes.vote(&params, auth_p0);
+        votes.vote(&params, auth_p1);
+        assert_eq!(votes.n_votes(&params, &old_participants), 2);
+
+        // when: checking votes against a different participant set (simulating post-resharing)
+        let mut new_participants = Participants::default();
+        let p2 = gen_participant(2);
+        new_participants.insert(p2.0.clone(), p2.1).unwrap();
+
+        // then: votes from non-participants are not counted
+        assert_eq!(votes.n_votes(&params, &new_participants), 0);
+
+        ctx.signer_account_id(p2.0);
+        testing_env!(ctx.build());
+        let auth_p2 = AuthenticatedAccountId::new(&new_participants).unwrap();
+        votes.vote(&params, auth_p2);
+        assert_eq!(votes.n_votes(&params, &new_participants), 1);
+    }
+}
--- a/crates/contract/src/primitives/threshold_votes.rs
+++ b/crates/contract/src/primitives/threshold_votes.rs
@@ -1,11 +1,11 @@
-use super::thresholds::ThresholdParameters;
-use super::{key_state::AuthenticatedAccountId, participants::Participants};
+use crate::primitives::thresholds::ThresholdParameters;
+use crate::primitives::{key_state::AuthenticatedAccountId, participants::Participants};
 use near_sdk::{log, near};
 use std::collections::BTreeMap;
 
 /// Tracks votes for ThresholdParameters (new participants and threshold).
 /// Each current participant can maintain one vote.
-// TODO: Replace with Votes<AuthenticatedAccountId, ThresholdParameters> from generic_votes.rs
+// TODO(#2825): Replace with Votes<AuthenticatedAccountId, ThresholdParameters> from generic_votes.rs
 // once this type is moved out of RunningContractState (which requires Clone + PartialEq + JSON).
 #[near(serializers=[borsh, json])]
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
--- a/crates/contract/src/primitives/votes.rs
+++ b/crates/contract/src/primitives/votes.rs
@@ -31,7 +31,7 @@ where
     V: VoterBounds,
     P: ProposalBounds,
 {
-    /// Constructs a enw Votes struct. Prpoosals will be stored in an iterable map under
+    /// Constructs a new Votes struct. Proposals will be stored in an iterable map under
     /// `storage_key`.
     pub fn new(storage_key: impl IntoStorageKey) -> Self {
         Self {
@@ -48,7 +48,7 @@ where
         let proposal_id = self.proposal_registry.register(proposal);
         let votes = self
             .vote_for(voter, proposal_id)
-            .expect("proposal id is expect do exist");
+            .expect("proposal id must exist");
         (proposal_id, votes)
     }
 
--- a/crates/contract/src/primitives/votes.rs
+++ b/crates/contract/src/primitives/votes.rs
@@ -6,7 +6,7 @@ use std::{collections::BTreeMap, collections::BTreeSet};
 
 use near_sdk::{near, IntoStorageKey};
 use proposal_registry::ProposalRegistry;
-use types::{ProposalBounds, ProposalId, VoterBounds, VoterSet};
+use types::{ProposalBounds, ProposalHash, ProposalId, VoterBounds, VoterSet};
 use votes_registry::{VoteRegistry, VoteResult};
 
 use crate::errors::VoteError;
@@ -103,10 +103,10 @@ where
 
     /// Returns a snapshot of the current state, that can be serde deserialized.
     /// Specifically, returns a map from proposal to votes
-    pub fn snapshot(&self) -> BTreeMap<ProposalId, (P, BTreeSet<V>)> {
+    pub fn snapshot(&self) -> BTreeMap<ProposalId, ((ProposalHash, P), BTreeSet<V>)> {
         let all_proposals = self.proposal_registry.all();
         let all_votes = self.vote_registry.all();
-        let merged: BTreeMap<ProposalId, (P, BTreeSet<V>)> = all_proposals
+        let merged: BTreeMap<ProposalId, ((ProposalHash, P), BTreeSet<V>)> = all_proposals
             .into_iter()
             .map(|(pid, proposal)| {
                 let votes = all_votes.get(&pid).cloned().unwrap_or_default();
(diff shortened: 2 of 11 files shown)


crates/contract/src/primitives/votes/types.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+}
+
+/// The set of voters who voted for a particular proposal. Always non-empty when stored
+/// inside `VotesByProposal`.
+#[near(serializers=[borsh])]
+#[derive(Debug)]
+pub struct VoterSet<V>(pub(super) BTreeSet<V>)
+where
+    V: VoterBounds;
```
netrome, line 36: As always, let's not put trait bounds on types. That's almost never helpful, and can lead to unintended limitations, for example when used in enums.

See https://stackoverflow.com/questions/49229332/should-trait-bounds-be-duplicated-in-struct-and-impl for more details


```suggestion
pub struct VoterSet<V>(pub(super) BTreeSet<V>);
```

netrome: I see this fails because the `near` macro doesn't add the `Ord` trait bound on the Borsh implementations. Why do we need the macro in the first place? Can't we just derive the borsh traits explicitly?

crates/contract/src/primitives/votes/votes_registry.rs
```diff
@@ -0,0 +1,418 @@
...
+use super::types::VoterSet;
+use super::types::{ProposalId, VoterBounds};
+
+/// Helper struct to keep track of submitted votes.
+/// Allows efficient look-up of votes by voter and votes by proposal.
+#[near(serializers=[borsh])]
+#[derive(Debug)]
+pub(super) struct VoteRegistry<V>
+where
+    V: VoterBounds,
```
netrome, line 16: Same here. I do like the pattern of breaking out the shared trait bounds though. Looks great.


```suggestion
pub(super) struct VoteRegistry<V>;
```

crates/contract/src/primitives/votes/proposal_registry.rs
```diff
@@ -0,0 +1,192 @@
...
+use super::types::{ProposalBounds, ProposalId};
+
+/// Keeps track of proposals and assigns stable ids.
+/// `id_by_proposal` and `proposals_by_id` are inverse of one another.
+/// `next_id` is the next ID to assign (monotonically increasing).
+#[near(serializers=[borsh])]
+#[derive(Debug)]
+pub(super) struct ProposalRegistry<P>
+where
+    P: ProposalBounds,
```
netrome, line 14: Same here - let's not put trait bounds on structs. It can lead to severe headaches when you least expect it and there should typically be no good reason to do it.

```suggestion
pub(super) struct ProposalRegistry<P>;
```

crates/contract/src/primitives/votes.rs
```diff
@@ -1,135 +1,424 @@
...
+/// Keeps track of proposals and votes.
+/// Invariants:
+/// - Each voter has exactly one vote.
+/// - Each proposal has at least one vote.
+#[near(serializers=[borsh])]
+#[derive(Debug)]
+pub struct Votes<V, P>
+where
+    V: VoterBounds,
+    P: ProposalBounds,
```
netrome, line 23: Idem

```suggestion
pub struct Votes<V, P>;
```

netrome (commented):
Just started looking. Noticed a lot of trait bounds on structs, which should be avoided. I'll continue reviewing tomorrow.

crates/contract/src/primitives/votes/proposal_registry.rs
netrome: As we discussed on slack, this struct may need to retain the trait bounds since it holds an `IterableMap`. I'd recommend collecting those trait bounds as `IterableMapBounds` though.

crates/contract/src/primitives/votes/proposal_registry.rs
```diff
@@ -0,0 +1,192 @@
...
+{
+    id_by_proposal: HashMap<P, ProposalId>,
+    proposals_by_id: IterableMap<ProposalId, P>,
```
netrome: Why do we use a `HashMap` in one direction and an `IterableMap` in the other?

crates/contract/src/primitives/votes/types.rs
```diff
@@ -0,0 +1,152 @@
...
+#[near(serializers=[borsh])]
+#[derive(Debug, PartialEq, PartialOrd, Eq, Ord, Clone, Copy, From, Deref, Into)]
+pub struct ProposalId(pub(crate) u64);
+
+impl ProposalId {
+    pub(super) fn next(&self) -> Self {
+        let (next, overflow) = self.0.overflowing_add(1);
+        if overflow {
+            near_sdk::env::panic_str("overflow in proposal id")
+        }
```
netrome: Nit: Seems like the intention is to do a checked operation here rather than allowing/handling overflows.

```suggestion
        let Some(next) = self.0.checked_add(1) else {
            near_sdk::env::panic_str("overflow in proposal id")
        };
```

crates/contract/src/primitives/votes/votes_registry.rs
```diff
@@ -0,0 +1,418 @@
...
+{
+    votes_by_voter: BTreeMap<V, ProposalId>,
+    votes_by_proposal: BTreeMap<ProposalId, VoterSet<V>>,
```
netrome, line 19: No strong opinion but I'm curious of the choice of `BTreeMap` here, when we use `HashMap` and `IterableMap` in the proposal registry. They all seem very similar to me so I'd expect us to be able to converge to a single map type.

crates/contract/src/primitives/votes/proposal_registry.rs
```diff
@@ -0,0 +1,192 @@
...
+/// Keeps track of proposals and assigns stable ids.
+/// `id_by_proposal` and `proposals_by_id` are inverse of one another.
+/// `next_id` is the next ID to assign (monotonically increasing).
+#[near(serializers=[borsh])]
+#[derive(Debug)]
+pub(super) struct ProposalRegistry<P>
+where
+    P: ProposalBounds,
+{
+    id_by_proposal: HashMap<P, ProposalId>,
```
netrome: If we assume `P` may be a relatively large structure, it feels bad to have to keep this as keys in a map since that would require loading all proposals on every contract call.

From what I can see, the purpose of this map is to deduplicate `P` when multiple people submit the same proposal. We would be able to achieve the same property of `ProposalId` was a hash of the proposals instead of and incremental number.

Did you consider that or is there anything that would be problematic in that case? (Like gas cost overhead of computing the hash function perhaps)

crates/contract/src/primitives/threshold_votes.rs
```diff
@@ -0,0 +1,137 @@
+use super::thresholds::ThresholdParameters;
+use super::{key_state::AuthenticatedAccountId, participants::Participants};
+use near_sdk::{log, near};
+use std::collections::BTreeMap;
+
+/// Tracks votes for ThresholdParameters (new participants and threshold).
+/// Each current participant can maintain one vote.
+// TODO: Replace with Votes<AuthenticatedAccountId, ThresholdParameters> from generic_votes.rs
```
netrome: Is this to be done in this PR?

netrome: Otherwise if it's for a follow-up please reference it here.

crates/contract/src/primitives/threshold_votes.rs
```diff
@@ -0,0 +1,137 @@
+use super::thresholds::ThresholdParameters;
+use super::{key_state::AuthenticatedAccountId, participants::Participants};
```
netrome: Nit: I find fully qualified imports easier to read.
```suggestion
use crate::primitives::thresholds::ThresholdParameters;
use crate::primitives::{key_state::AuthenticatedAccountId, participants::Participants};
```

netrome (changes requested):
Nice stuff overall, only a minor hard blocker about toe TODO in threshold votes - it's unclear to me if this is something intended for this PR that was overlooked or a planned follow-up. If it's the latter we should reference the issue in the TODO.

crates/contract/src/primitives/votes/proposal_registry.rs
kevindeforth: `ProposalId` is just a wrapper around `u64`. Using an `IterableMap` to store that type is likely to add more bloat than anything else.
However, `proposal` could be potentially large, so we want to make sure we deserialize only when neeeded.
edit: ah, I see I had an oversight here. Related comment: https://github.com/near/mpc/pull/2739#discussion_r3062772121

_edit 2: made this a BTreeMap: https://github.com/near/mpc/pull/2739/commits/632b1ea200d408c28035cb180a62605ee6037c99_

crates/contract/src/primitives/threshold_votes.rs
kevindeforth: That code was just moved from the previous votes.rs. Will fix though. https://github.com/near/mpc/pull/2739/commits/1c4ae34ca572611378ed1be6b39f7a22cc446cee

crates/contract/src/primitives/votes/proposal_registry.rs
kevindeforth: Oh, I must have hallucinated writing this, because I somehow assumed that `HashMap` wouldn't store the full proposal. Will fix.

_(fixed in https://github.com/near/mpc/pull/2739/commits/632b1ea200d408c28035cb180a62605ee6037c99)_

crates/contract/src/primitives/votes/votes_registry.rs
kevindeforth: These votes will have at most `num_voters` entries each and contain pretty small data (type `u64` and `AccountId`, both of which are rather small-ish).

This is fine until we exceed 100 participants, which I don't think will happen anytime soon.

crates/contract/src/primitives/threshold_votes.rs
netrome: Ah I see :sweat_smile:

crates/contract/src/primitives/threshold_votes.rs
kevindeforth, line 1: note: this file was `votes.rs` before and was just renamed. This is the diff to the file currently on main:
```
< use super::thresholds::ThresholdParameters;
< use super::{key_state::AuthenticatedAccountId, participants::Participants};
---
> use crate::primitives::thresholds::ThresholdParameters;
> use crate::primitives::{key_state::AuthenticatedAccountId, participants::Participants};
7a8,9
> // TODO(#2825): Replace with Votes<AuthenticatedAccountId, ThresholdParameters> from generic_votes.rs
> // once this type is moved out of RunningContractState (which requires Clone + PartialEq + JSON).
```

crates/contract/src/primitives/votes/votes_registry.rs
kevindeforth: After https://github.com/near/mpc/pull/2739/commits/1c306a85b9801acd427f8a5d97add1ec7948a05d, we no longer use `HashMap`. Instead, we have the following `BTreeMaps`:

- `Voter <-> ProposalId`, containing at most `num_voters` entries
- `ProposalId <-> VoterSet<Voter>`. The map contains at most `num_voters` entries and its values `VoterSets` are disjoint, with their union containing at most `num_voters` elements. 
-  `ProposalHash <-> ProposalId`, containing at most `num_voters` entries, likely much fewer.

And we have one `IterableMap`: `proposal_by_id: IterableMap<ProposalId, (ProposalHash, Proposal)>`.

While `proposal_by_id` also contains less than `num_voters` entries and thus, is still below the threshold for which native collection types are recommended (taking [this guide](https://docs.near.org/smart-contracts/anatomy/collections) as reference), we prefer an `IterableMap` here, because `Proposal` can be potentially quite large.
An `IterableMap` ensures we only deserialize the proposal when truly needed.

crates/contract/src/primitives/votes/proposal_registry.rs
```diff
@@ -0,0 +1,208 @@
...
+/// `id_by_proposal` and `proposals_by_id` are inverse of one another.
+/// `next_id` is the next ID to assign (monotonically increasing).
+#[near(serializers=[borsh])]
+#[derive(Debug)]
+pub(super) struct ProposalRegistry<P>
+where
+    P: ProposalBounds,
+{
+    id_by_proposal: BTreeMap<ProposalHash, ProposalId>,
+    proposals_by_id: IterableMap<ProposalId, (ProposalHash, P)>,
```
DSharifi, line 17: note for myself:
is ProposalHash used here?

crates/contract/src/primitives/votes/proposal_registry.rs
```diff
@@ -0,0 +1,208 @@
...
+where
+    P: ProposalBounds,
+{
+    /// builds a new proposal. The caller is responsibe for ensuring that `storage_key` is empty.
+    pub(super) fn new(storage_key: impl IntoStorageKey) -> Self {
+        Self {
+            id_by_proposal: BTreeMap::new(),
+            proposals_by_id: IterableMap::new(storage_key),
+            // we should probably set this.
+            next_id: ProposalId(0),
```
DSharifi: Let's add a tracking issue, seems like a todo?

crates/contract/src/primitives/votes/proposal_registry.rs
```diff
@@ -0,0 +1,208 @@
...
+        }
+    }
+
+    /// Stores the proposal if new, or looks up the existing matching proposal.
+    /// Returns the proposal id for this proposal.
+    pub(super) fn register(&mut self, proposal: P) -> ProposalId {
+        let encoded = borsh::to_vec(&proposal).expect("borsh serialization failed");
```
DSharifi: nit:
expect message should tell the reader why we expect it to not fail, instead of what actually failed. The panic should already contain that,

crates/contract/src/primitives/votes/proposal_registry.rs
```diff
@@ -0,0 +1,208 @@
...
+        self.id_by_proposal.insert(proposal_hash, proposal_id);
+        self.proposals_by_id
+            .insert(proposal_id, (proposal_hash, proposal));
+        proposal_id
+    }
+
+    /// Removes the proposal under [`ProposalId`]
+    pub(super) fn remove(&mut self, proposal_id: &ProposalId) {
+        if let Some((proposal_hash, _)) = self.proposals_by_id.remove(proposal_id) {
+            self.id_by_proposal.remove(&proposal_hash);
```
DSharifi: could be good to add debug_assert! here that this remove happened

crates/contract/src/primitives/votes/proposal_registry.rs
```diff
@@ -0,0 +1,208 @@
...
+        }
+    }
+
+    /// Returns true if a proposal exists for [`ProposalId`], returns false otherwise.
+    pub(super) fn contains(&self, proposal_id: &ProposalId) -> bool {
+        self.proposals_by_id.contains_key(proposal_id)
+    }
+
+    /// clears the maps
```
DSharifi: The maps are internal implementation detail

```suggestion
    /// clears all registered proposals
```

crates/contract/src/primitives/votes/proposal_registry.rs
```diff
@@ -0,0 +1,208 @@
...
+    pub(super) fn clear(&mut self) {
+        self.id_by_proposal.clear();
+        self.proposals_by_id.clear();
+    }
+
+    /// Returns the registry in a form fit for json serialization.
+    pub(super) fn all(&self) -> BTreeMap<ProposalId, (ProposalHash, P)> {
```
DSharifi, line 72: Note:
This function is currently dead code (only used in tests)

crates/contract/src/primitives/votes/types.rs
```diff
@@ -0,0 +1,154 @@
...
+#[near(serializers=[borsh])]
+#[derive(Debug, PartialEq, PartialOrd, Eq, Ord, Clone, Copy, From, Deref, Into)]
+pub struct ProposalId(pub(crate) u64);
+
+impl ProposalId {
+    pub(super) fn next(&self) -> Self {
+        let Some(next) = self.0.checked_add(1) else {
+            near_sdk::env::panic_str("overflow in proposal id")
+        };
+        ProposalId(next)
```
DSharifi, line 26: Wonder if a mutable increment would be nicer to have, so we don't have to increment and assign.

crates/contract/src/primitives/votes/types.rs
```diff
@@ -0,0 +1,154 @@
...
+impl<V> VoterSet<V>
+where
+    V: VoterBounds,
+{
+    pub(super) fn new() -> Self {
+        Self(BTreeSet::new())
+    }
+
+    // counts all the votes for which `predicate` returns true
+    pub fn count_for(&self, predicate: impl Fn(&V) -> bool) -> usize {
```
DSharifi, line 47: How do you envision this being used?

Currently I only see it used in the unit tests.

DSharifi (commented):
Thanks, for this change. It looks quite good, but I haven't fully understood why some of the APIs are designed as they are, but I assume there's a good reason for it.

I think it would be easier to understand the API if you could migrate one of the existing voting patterns to use these new structs as a PR stacked on top of this one.

crates/contract/src/primitives/votes/proposal_registry.rs
kevindeforth: oh, this can be removed.

crates/contract/src/primitives/votes/types.rs
kevindeforth: Some of our current voting structs are only comparing the number of collected votes to the required threshold for accepting a proposal. The issue with this is that we risk counting votes of former participants, who have since left the network. We rely on votes getting cleaned up after each resharing.

Now, the downside of course is that calling this method is more expensive than simply comparing two integers. But I have not yet seen any indication that this would result in intolerable gas costs.

kevindeforth:
> Thanks, for this change. It looks quite good, but I haven't fully understood why some of the APIs are designed as they are, but I assume there's a good reason for it.
> 
> I think it would be easier to understand the API if you could migrate one of the existing voting patterns to use these new structs as a PR stacked on top of this one.

Thanks for the review! Regarding usage of this API, please take a look at #2930.

kevindeforth:
Putting this on hold for now.

Using the hash allows us to save some code bytes and simplify the structs.
The main idea is to not store the Proposal itself, but only the Hash of the proposal.

Take a look at this comment: https://github.com/near/mpc/pull/2930#issuecomment-4264767358.

I will open another PR with the simplified voting structs such that we can compare.

_c.f. https://github.com/near/mpc/pull/2934_

kevindeforth:
closing this in favor of https://github.com/near/mpc/pull/2934

PR #2740, opened 2026-04-07
refactor: Separation of the node and the smart contract

### LINKED ISSUE #381: MPC node dependency management

SimonRastikian:
Closes #381


SimonRastikian:
Closed as this should be split

PR #2746, opened 2026-04-08
fix: assert in crypto layer replaced by proper error handling

SimonRastikian:
Closes #2745


crates/threshold-signatures/src/crypto/proofs/strobe.rs
```diff
@@ -153,7 +153,7 @@ impl Strobe128 {
     fn begin_op(&mut self, flags: u8, more: bool) {
         // Check if we're continuing an operation
         if more {
-            assert_eq!(
+            debug_assert_eq!(
                 self.cur_flags, flags,
                 "You tried to continue op {:#b} but changed flags to {:#b}",
                 self.cur_flags, flags,
```
gilcu3: this will not trigger in release builds, so we cannot do this change without going deeper

gilcu3 (changes requested):
Minor change requested, as the current code will accepts what was previously rejected. If it is too cumbersome to use Errors there, it is strictly better to leave the asserts, unless we are convinced these were not needed.

crates/threshold-signatures/src/crypto/proofs/strobe.rs
SimonRastikian: done

gilcu3 (dismissed):
Thank you!

DSharifi (approved):

gilcu3 (approved):

Issue #2684, opened 2026-04-01
Migrate all pytests to new Rust E2E framework or as Integration test


anodar:
### Background

We have new E2E framework in Rust. We should get rid of tests in python and either migrate them to Rust E2E test or to integration tests.

### Acceptance Criteria

No pytests no more

### Done

#2839

```  
  ┌─────────────────────────────────────────────────────────────────────┬────────────────────────────────────────────────────────────────────────────────────────────────┐
  │                               pytest                                │                                              Rust                                              │
  ├─────────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ shared_cluster_tests/test_web_endpoints::test_web_endpoints         │ e2e-tests/web_endpoints.rs::test_web_endpoints                                                 │
  ├─────────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ test_submit_participant_info::test_submit_participant_info_endpoint │ e2e-tests/submit_participant_info.rs::test_submit_participant_info                             │
  ├─────────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ robust_ecdsa/test_requests::test_request_lifecycle                  │ e2e-tests/request_lifecycle.rs::mpc_cluster__should_successfully_process_robust_ecdsa_requests │
  ├─────────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ shared_cluster_tests/test_requests::test_request_lifecycle          │ e2e-tests/request_lifecycle.rs::mpc_cluster__should_sign_with_scheme_matching_domain           │
  └─────────────────────────────────────────────────────────────────────┴────────────────────────────────────────────────────────────────────────────────────────────────┘
```

  #2841
```
  ┌───────────────────────────────────────────────────────────────────────────────────────────────────────────┬──────────────────────────────────────────────────────────────────┐
  │                                                  pytest                                                   │                               Rust                               │
  ├───────────────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ test_request_during_resharing::test_threshold_from_previous_running_state_is_maintained                   │ resharing.rs::test_signature_requests_in_resharing_are_processed │
  ├───────────────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
[...]

PR #2759, opened 2026-04-08
test: improve e2e framework, implement tests for web_endpoints, submit_participant_info, robust_ecdsa

### LINKED ISSUE #2839: test: port web_endpoints, submit_participant_info, and robust_ecdsa from pytest to Rust e2e
## Background

The MPC project is migrating its test suite from pytest to Rust-based e2e tests as part of the pytest deprecation effort. This task covers porting three test suites (web_endpoints, submit_participant_info, robust_ecdsa) and improving the e2e testing framework infrastructure.

## User Story

As an MPC developer, I need the web_endpoints, submit_participant_info, and robust_ecdsa pytest tests migrated to Rust e2e tests so that we can deprecate the pytest infrastructure and maintain a unified Rust test suite.

## Acceptance Criteria

We have 
- `web_endpoints`
- `submit_participant_info`
- `robust_ecdsa` 
tests implemented in Rust and infra required for it.

Implemented in #2759. Part of broader effort tracked in #2684.

anodar:
Closes: https://github.com/near/mpc/issues/2839
Work towards: https://github.com/near/mpc/issues/2684

Framework:
- Options pattern for setup_cluster(port_seed, |c| { ... })
- Shared test helpers in tests/common.rs
- send_ckd_request, get_tee_accounts, begin_resharing/start_resharing
- Attestation wait, epoch ID fix, participant ID preservation in resharing
- vote_cancel_resharing_from, reset_and_start_nodes, client_for_voting
- build-contract-wasm cargo make task, nextest e2e test group

Tests ported from pytest:
- web_endpoints: HTTP endpoint health checks on all nodes
- submit_participant_info: TEE attestation auto-submission
- robust_ecdsa: V2Secp256k1 signing (6 nodes, threshold 5)


crates/e2e-tests/src/blockchain.rs
```diff
@@ -3,6 +3,8 @@ use near_kit::FinalExecutionOutcome;
 use near_mpc_contract_interface::types::ProtocolContractState;
 use serde::de::DeserializeOwned;
 
+const DEFAULT_CALL_GAS: near_kit::Gas = near_kit::Gas::from_tgas(300);
```
gilcu3: we normally call this MAX_GAS, although that has recently changed I believe

crates/e2e-tests/src/mpc_node.rs
```diff
@@ -81,6 +85,24 @@ impl MpcNode {
...
+    /// Create a marker file that tells the MPC node a key event attempt was
+    /// already started. The node reads these on startup to determine which
+    /// attempt ID to use next.
+    pub fn reserve_key_event_attempt(
+        &self,
+        epoch_id: u64,
+        domain_id: u64,
+        attempt_id: u64,
+    ) -> anyhow::Result<()> {
+        let dir = self.setup.home_dir.join("temporary_keys");
```
gilcu3: is this a constant that should be exposed by the node (or a common crate)?

crates/e2e-tests/src/mpc_node.rs
```diff
@@ -207,7 +229,29 @@ impl MpcNodeSetup {
...
+    /// Removes keyshares, temporary keys, and RocksDB state, but keeps the
+    /// NEAR chain data (`data/`, `config.json`, `genesis.json`, `node_key.json`)
+    /// so the node can resume syncing quickly after restart.
+    ///
+    /// Re-writes secrets.json and start_config.toml from the saved setup.
+    pub fn reset_mpc_state(&self) -> anyhow::Result<()> {
+        // Nuke and recreate the home directory. SIGKILL can leave NEAR indexer
+        // data in a corrupted state that prevents restart, so we start fully
+        // fresh. The NEAR indexer will re-init from genesis_path/boot_nodes in
+        // start_config.toml and sync from block 0.
```
gilcu3: this seems to contradict the message the doc string of the function

crates/e2e-tests/tests/common.rs
```diff
@@ -0,0 +1,130 @@
+#![allow(dead_code)]
```
gilcu3: could we put this on the actual `dead_code` blocks?

crates/e2e-tests/tests/common.rs
```diff
@@ -0,0 +1,130 @@
...
+/// }).await;
+/// ```
+pub async fn setup_cluster(
+    port_seed: u16,
+    configure: impl FnOnce(&mut MpcClusterConfig),
+) -> (MpcCluster, RunningContractState) {
+    tracing_subscriber::fmt()
+        .with_env_filter(
+            tracing_subscriber::EnvFilter::try_from_default_env()
+                .unwrap_or_else(|_| "info,e2e_tests=debug".parse().unwrap()),
```
gilcu3: do we need debug by default? normally this is better set by whoever is running the tests

crates/e2e-tests/tests/common.rs
```diff
@@ -0,0 +1,130 @@
...
+
+pub fn generate_eddsa_payload() -> serde_json::Value {
+    let bytes: [u8; 32] = rand::random();
+    json!({ "Eddsa": hex::encode(bytes) })
+}
+
+pub fn generate_ckd_app_public_key() -> CKDAppPublicKey {
+    let point = G1Projective::generator() * Scalar::from(42u64);
+    CKDAppPublicKey::AppPublicKey(Bls12381G1PublicKey::from(&point))
+}
```
gilcu3: we are striving to make testing deterministic when possible. For this kind of functions, they should receive the `&mut rng` as parameter. Check how it is usually done for example in the ts crate tests

gilcu3 (commented):
This is a somehow partial review, as there are quite a few things I don't fully understand. Left some questions regarding those, and also asked for some changes (sorry if some of those were existing code that just moved around)

crates/e2e-tests/src/blockchain.rs
```diff
@@ -86,6 +88,32 @@ impl NearBlockchain {
...
 
+    /// Create a client that fetches a fresh nonce from the blockchain,
+    /// bypassing the global nonce cache. Used when MPC node processes have
+    /// been sending transactions from the same account, making the cached
+    /// nonce stale.
+    ///
+    /// Works by appending a unique fragment to the RPC URL — the nonce cache
+    /// is keyed by `(url, account, pubkey)`, so a different URL forces a
+    /// fresh fetch. The `#` fragment is not sent to the server.
+    pub fn client_for_voting(
```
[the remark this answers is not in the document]
anodar: There are 2 paths that same account sends transaction: from test directly (not yet in this change, but in resharing one), and the mpc node itself.
Near-kit caches the nonce and it gets out of sync as mpc binary does its own [nonce book keeping](https://github.com/near/mpc/blob/069ebdaee866791da7a1e00f29f30785cbb7e676/crates/node/src/indexer/tx_signer.rs#L18).

I was getting nonce errors and that's what Claude suggested and seemed reasonable to me. Another option would have been to increase `max_nonce_retries` from 10 to 100 or something like that.

crates/e2e-tests/tests/common.rs
```diff
@@ -0,0 +1,130 @@
...
+        .expect("failed to start cluster");
+
+    let state = cluster
+        .get_contract_state()
+        .await
+        .expect("failed to get contract state");
+    let running = match state {
+        ProtocolContractState::Running(r) => r,
+        other => panic!("expected Running state, got: {other:?}"),
+    };
```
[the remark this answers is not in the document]
anodar: Great catch! Changed to wait for state.

crates/e2e-tests/tests/common.rs
```diff
@@ -0,0 +1,130 @@
...
+                .join("../../target/wasm32-unknown-unknown/release-contract/mpc_contract.wasm")
+        });
+
+    std::fs::read(&wasm_path).unwrap_or_else(|e| {
+        panic!(
+            "Failed to read contract WASM at {}: {e}\n\
+             Build it first:\n  \
+             cargo build -p mpc-contract --target=wasm32-unknown-unknown \
+             --profile=release-contract --locked\n  \
+             wasm-opt -Oz <path> -o <path>",
```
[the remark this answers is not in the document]
anodar: Change to cargo make as we have command there, so that if we make change in how we build it we don't have to do it in two places. 

Changed makefile to use cargo near as well.

crates/e2e-tests/tests/web_endpoints.rs
```diff
@@ -0,0 +1,180 @@
...
+
+        // GET /health
+        let resp = client
+            .get(format!("http://{web_addr}/health"))
+            .send()
+            .await
+            .unwrap();
+        assert_eq!(resp.status(), 200, "node {i}: /health status");
+        let body = resp.text().await.unwrap();
+        assert!(body.contains("OK"), "node {i}: /health body: {body}");
```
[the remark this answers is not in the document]
anodar: Done.

crates/e2e-tests/tests/robust_ecdsa.rs
```diff
@@ -0,0 +1,56 @@
...
+///
+/// Setup: 6 nodes, threshold 5, single V2Secp256k1 domain, no triples
+/// (robust ECDSA doesn't need them), 6 presignatures buffered.
+#[tokio::test]
+async fn test_robust_ecdsa() {
+    let (cluster, running) = common::setup_cluster(common::ROBUST_ECDSA_PORT_SEED, |c| {
+        c.num_nodes = 6;
+        c.initial_participants = 6;
+        c.threshold = 5;
+        c.domains = vec![DomainConfig {
```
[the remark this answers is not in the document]
anodar: Good to know. Merged it with request lifecycle one.

crates/e2e-tests/tests/common.rs
gilcu3: hmm we have similar code in the sandbox, and I think we don't have a dead_code mark there, so there must be a way to achieve it

.config/nextest.toml
```diff
@@ -1,3 +1,12 @@
+[test-groups.e2e]
+# E2E tests each spawn a NEAR sandbox + multiple MPC node processes.
+# Limit parallelism to avoid overwhelming the machine.
+max-threads = 10
```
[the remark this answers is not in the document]
anodar: Sounds good, got rid of it.

crates/e2e-tests/tests/common.rs
anodar: Sandbox puts them inside tests folder and has test.rs next to it so that it compiles as single binary. I prefer to keep it flat rather than having additional `tests` folder inside `e2e-tests` (seems redundant), but I don't have strong arguments against it, besides not having whole file as allow(dead_code) is probably good idea. 

Aligned it with sandbox and dropped `allow(dead_code)`

crates/e2e-tests/tests/common.rs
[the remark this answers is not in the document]
anodar: I'm not sure I follow what we will need to add.

Currently e2e test job depends on `mpc-pytest-build` which builds the contract, and e2e test job sets it in MPC_CONTRACT_WASM env variable that then load_contract_wasm reads from.

crates/e2e-tests/tests/common.rs
```diff
@@ -0,0 +1,150 @@
...
+    CKDAppPublicKey, ProtocolContractState, RunningContractState,
+};
+use near_mpc_crypto_types::Bls12381G1PublicKey;
+use serde_json::json;
+
+pub const POLL_INTERVAL: Duration = Duration::from_millis(500);
+pub const SIGN_REQUEST_PER_SCHEME_PORT_SEED: u16 = 1;
+pub const WEB_ENDPOINTS_PORT_SEED: u16 = 2;
+#[expect(dead_code)]
+pub const KEY_RESHARING_PORT_SEED: u16 = 3;
```
netrome, line 17: Why do we add this constant if it's dead code? Is it a placeholder for upcoming tests?

crates/e2e-tests/tests/common.rs
```diff
@@ -0,0 +1,130 @@
...
+///
+/// ```ignore
+/// // Default 3-node cluster:
+/// setup_cluster(SEED, |_| {}).await;
+///
+/// // Custom 4-node cluster with 2 initial participants:
+/// setup_cluster(SEED, |c| {
+///     c.num_nodes = 4;
+///     c.initial_participants = 2;
+/// }).await;
```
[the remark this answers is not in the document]
netrome: Oh interesting. Definitely unconventional. More common would be to just construct and pass the config explicitly:

```rust
let config = MpcClusterConfig {
    num_nodes: 4,
    initial_participant_indices: vec![0, 1],
    ..MpcClusterConfig::default_for_test(SEED, contract_wasm);
};
setup_cluster(SEED, config)
```

Alternatively for better ergonomics something resembling the builder pattern would be more conventional:

```rust
let config = MpcClusterConfig::default_for_test(SEED, contract_wasm).num_nodes(4).initial_participant_indices([0, 1]);
```

That said, the `configure` closure reads very nicely so I'm happy to keep it. Always a fan of passing functions around :)

.config/nextest.toml
```diff
@@ -1,3 +1,6 @@
+[[profile.default.overrides]]
+filter = 'package(e2e-tests) and not test(/^port_allocator/)'
+
```
netrome: What's the purpose of this? It seems to be a filter without any overrides, which essentially is a no-op. https://nexte.st/docs/configuration/per-test-overrides/ Can we clean it out?

crates/e2e-tests/src/cluster.rs
```diff
@@ -240,6 +284,36 @@ impl MpcCluster {
...
+                    tracing::info!(node = idx, "node is healthy");
+                    return Ok(());
+                }
+            }
+            anyhow::ensure!(
+                tokio::time::Instant::now() < deadline,
+                "node {idx} did not become healthy within 120s"
+            );
+            tokio::time::sleep(POLL_INTERVAL).await;
+        }
```
netrome: Did you consider using `backon` for this and the other retry loops? https://crates.io/crates/backon - we already have it in the workspace.

crates/e2e-tests/tests/request_lifecycle.rs
```diff
@@ -131,3 +55,47 @@ async fn test_request_lifecycle() {
         tracing::info!(domain_id = ?domain.id, "sign request returned valid signature");
     }
 }
+
+#[tokio::test]
+async fn test_robust_ecdsa() {
```
netrome: I would love it if these tests also followed our conventions of `//Given //When //Then` sections, and the double underscore naming convention - renaming it to something like `mpc_cluster__should_successfully_process_robust_ecdsa_requests`

crates/e2e-tests/tests/request_lifecycle.rs
```diff
@@ -131,3 +55,47 @@ async fn test_request_lifecycle() {
...
+        c.threshold = 5;
+        c.domains = vec![DomainConfig {
+            id: DomainId(0),
+            scheme: SignatureScheme::V2Secp256k1,
+            purpose: Some(DomainPurpose::Sign),
+        }];
+        c.triples_to_buffer = 0;
+        c.presignatures_to_buffer = 6;
+    })
+    .await;
```
netrome, line 95: Yeah this reads very nicely indeed! 🚀 🤩

crates/e2e-tests/tests/submit_participant_info.rs
```diff
@@ -0,0 +1,43 @@
...
+            .await
+            .expect("failed to query TEE accounts");
+
+        if tee_accounts.len() >= 4 {
+            tracing::info!(
+                count = tee_accounts.len(),
+                "all nodes submitted attestations"
+            );
+            return;
+        }
```
netrome: I guess we should error here if for any reason we got 5

netrome (dismissed):
Nice stuff! Some questions and thoughts. The nextest.toml config should be cleaned up though.

crates/e2e-tests/tests/common.rs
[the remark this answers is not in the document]
netrome: Ugh, I don't like the idea of having to declare this as a submodule of every test that uses it. Would be nicer to break it out as a separate crate. Couldn't this be a sub module of the existing e2e tests crate?

Also I don't think that part of the book is focused on naming conventions so don't hesitate to deviate from it.

crates/e2e-tests/tests/common.rs
anodar: Yes it's a placeholder for tests in https://github.com/near/mpc/pull/2760 and https://github.com/near/mpc/pull/2761

crates/e2e-tests/tests/request_lifecycle.rs
anodar: Done.

crates/e2e-tests/src/cluster.rs
anodar: Changed to use it.

crates/e2e-tests/tests/submit_participant_info.rs
anodar: Addressed.

.config/nextest.toml
anodar: That's leftover from when I used group and max concurrency. Cleaned it up.

crates/e2e-tests/src/cluster.rs
```diff
@@ -293,20 +294,20 @@ impl MpcCluster {
...
-                    return Ok(());
-                }
-            }
-            anyhow::ensure!(
-                tokio::time::Instant::now() < deadline,
-                "node {idx} did not become healthy within 120s"
-            );
-            tokio::time::sleep(POLL_INTERVAL).await;
-        }
+        (|| async {
```
anodar, line 310: @netrome  I'm not convinced this reads better 👀

deny.toml
```diff
@@ -29,6 +29,18 @@ ignore = [
   "RUSTSEC-2026-0002", # TODO(#1730): Remove once nearcore updates to a fixed version
   "RUSTSEC-2026-0009", # TODO(#1999): Remove once MSVR is bumped to 1.88
   "RUSTSEC-2023-0089", # atomic-polyfill is unmaintained
+  # TODO(#1730): Remove once nearcore updates to a fixed version
```
anodar: These were all added today: https://rustsec.org/advisories/

crates/e2e-tests/src/blockchain.rs
```diff
@@ -52,6 +54,32 @@ impl NearBlockchain {
...
+        primary_key: &SigningKey,
+        extra_key: &SigningKey,
+    ) -> anyhow::Result<()> {
+        self.root_client
+            .transaction(name)
+            .create_account()
+            .transfer(near_kit::NearToken::from_near(balance_near))
+            .add_full_access_key(near_kit::PublicKey::Ed25519(
+                primary_key.verifying_key().to_bytes(),
+            ))
```
gilcu3: oh my, they (`near-kit`) are not using key types from `near-sdk`, so we have more duplicate key types :| Nothing we can do at the moment

crates/e2e-tests/src/mpc_node.rs
```diff
@@ -207,7 +231,30 @@ impl MpcNodeSetup {
...
+        if self.home_dir.exists() {
+            std::fs::remove_dir_all(&self.home_dir)
+                .with_context(|| format!("failed to remove {}", self.home_dir.display()))?;
+        }
+        std::fs::create_dir_all(&self.home_dir)
+            .with_context(|| format!("failed to recreate {}", self.home_dir.display()))?;
+        self.write_secrets_json()?;
+        self.write_start_config()?;
```
gilcu3, line 235: do we need this here because we are deleting them right before?

crates/e2e-tests/tests/common.rs
```diff
@@ -0,0 +1,150 @@
...
+    let initial_participant_indices = config.initial_participant_indices.clone();
+    let presignatures_to_buffer = config.presignatures_to_buffer;
+    let cluster = MpcCluster::start(config)
+        .await
+        .expect("failed to start cluster");
+
+    cluster
+        .wait_for_state(
+            |s| matches!(s, ProtocolContractState::Running(_)),
+            Duration::from_secs(120),
```
gilcu3: nit: lets reuse this constant

crates/e2e-tests/tests/common.rs
```diff
@@ -0,0 +1,150 @@
...
+pub fn generate_eddsa_payload(rng: &mut impl rand::Rng) -> serde_json::Value {
+    let mut bytes = [0u8; 32];
+    rng.fill(&mut bytes);
+    json!({ "Eddsa": hex::encode(bytes) })
+}
+
+pub fn generate_ckd_app_public_key() -> CKDAppPublicKey {
+    let point = G1Projective::generator() * Scalar::from(42u64);
+    CKDAppPublicKey::AppPublicKey(Bls12381G1PublicKey::from(&point))
+}
```
gilcu3: we also can take `&mut rng` in this one

crates/e2e-tests/tests/request_lifecycle.rs
```diff
@@ -81,34 +21,22 @@ async fn test_request_lifecycle() {
...
-    // When: we send a sign request for each Sign domain.
-    // Then: each returns a valid signature matching the requested scheme.
+    let mut rng = rand::rngs::StdRng::seed_from_u64(0);
     for domain in &sign_domains {
         let payload = match domain.scheme {
-            SignatureScheme::Secp256k1 => generate_ecdsa_payload(),
-            SignatureScheme::Ed25519 => generate_eddsa_payload(),
+            SignatureScheme::Secp256k1 => common::generate_ecdsa_payload(&mut rng),
+            SignatureScheme::Ed25519 => common::generate_eddsa_payload(&mut rng),
             _ => continue,
```
gilcu3: we could add robust-ecdsa (and maybe ckd here)

Makefile.toml
```diff
@@ -138,5 +138,13 @@ description = "Check that sandbox Docker image version matches nearcore"
 command = "bash"
 args = ["scripts/check-sandbox-image-version.sh"]
 
+[tasks.build-contract-wasm]
+description = "Build and optimize MPC contract WASM for E2E tests"
+script = '''
+cargo near build non-reproducible-wasm --no-wasmopt --features abi --profile=release-contract --manifest-path crates/contract/Cargo.toml --locked
```
gilcu3: here we should not use the no-wasmopt, so that a single step is needed

crates/e2e-tests/Cargo.toml
```diff
@@ -3,9 +3,11 @@ name = "e2e-tests"
 edition = { workspace = true }
 version = { workspace = true }
 license = { workspace = true }
+autotests = false
```
gilcu3, line 6: is this a temporary flag?

deny.toml
gilcu3: ups are we affected by all of that now? btw notice that we dont need to add this in the PR as it will not break CI

crates/e2e-tests/src/cluster.rs
```diff
@@ -107,6 +113,9 @@ pub struct MpcCluster {
     pub contract: DeployedContract,
     pub nodes: Vec<MpcNodeState>,
     pub node_keys: Vec<SigningKey>,
+    /// Separate access keys used by the test to cast votes on node accounts.
+    /// Disjoint from `node_keys` so the MPC node's own nonce sequence is never disturbed.
+    pub voting_keys: Vec<SigningKey>,
```
gilcu3: better call them operator keys, as thats what they are

crates/e2e-tests/src/cluster.rs
```diff
@@ -240,6 +285,36 @@ impl MpcCluster {
...
+            ConstantBuilder::default()
+                .with_delay(POLL_INTERVAL)
+                .with_max_times(240),
+        )
+        .await
+        .with_context(|| format!("node {idx} did not become healthy within 120s"))
+    }
+
+    /// Query all accounts that have TEE attestations stored in the contract.
+    pub async fn get_tee_accounts(&self) -> anyhow::Result<Vec<serde_json::Value>> {
```
gilcu3, line 327: this return type seems very broad. Not sure if we could do better, but might be worth it

crates/e2e-tests/src/cluster.rs
```diff
@@ -258,43 +333,153 @@ impl MpcCluster {
...
+        &self,
+        new_participants: &[usize],
+        new_threshold: usize,
+    ) -> anyhow::Result<()> {
+        self.start_resharing(new_participants, new_threshold)
+            .await?;
         self.wait_for_state(
             |s| matches!(s, ProtocolContractState::Running(_)),
-            Duration::from_secs(120),
+            Duration::from_secs(180),
```
gilcu3: is this a result of flakiness?

crates/e2e-tests/tests/common.rs
```diff
@@ -0,0 +1,150 @@
...
+/// // Custom 4-node cluster with 2 initial participants:
+/// setup_cluster(SEED, |c| {
+///     c.num_nodes = 4;
+///     c.initial_participant_indices = vec![0, 1];
+/// }).await;
+/// ```
+pub async fn setup_cluster(
+    port_seed: u16,
+    configure: impl FnOnce(&mut MpcClusterConfig),
+) -> (MpcCluster, RunningContractState) {
```
gilcu3, line 43: shouldn't this function decide which domains to add? I guess for simplicity we can add all of them always, but that will have a performance hit

gilcu3 (dismissed):
Thank you for the fixes.

Left some more comments, but nothing blocking IMO

crates/e2e-tests/src/mpc_node.rs
anodar: Yes otherwise node won't have start config and secrets to start up later.

Makefile.toml
kevindeforth: > so that a single step is needed

I think the difference is that cargo-near uses `-Os` and in the line below, we use `-Oz`?

crates/e2e-tests/Cargo.toml
kevindeforth: What do you mean "temporary"? I think this just disables automatic test recognition and allows to define them specifically, as on line 37 in this file. Right now, we run everything in `tests/e2e.rs`

crates/e2e-tests/src/blockchain.rs
```diff
@@ -52,6 +54,32 @@ impl NearBlockchain {
...
+            .add_full_access_key(near_kit::PublicKey::Ed25519(
+                extra_key.verifying_key().to_bytes(),
+            ))
+            .send()
+            .await
+            .map_err(|e| anyhow::anyhow!("failed to create account {name}: {e}"))?;
+        Ok(())
+    }
+
```
kevindeforth: nit: we don't need both `create_account` an `create_account_with_extra_key`. We can simply to have a single method, taking as input a slice:
```rust
pub async fn create_account_with_keys(
    &self,
    name: &str,
    balance_near: u128,
    keys: &[SigningKey],
) -> anyhow::Result<()> {
    let mut tx = self
        .root_client
        .transaction(name)
        .create_account()
        .transfer(near_kit::NearToken::from_near(balance_near));

    for key in keys {
        tx = tx.add_full_access_key(near_kit::PublicKey::Ed25519(
            key.verifying_key().to_bytes(),
        ));
    }

    tx.send()
        .await
        .map_err(|e| anyhow::anyhow!("failed to create account {name}: {e}"))?;

    Ok(())
}
```

.config/nextest.toml
netrome: Thanks :pray:

crates/e2e-tests/src/cluster.rs
```diff
@@ -87,6 +92,7 @@ impl MpcClusterConfig {
             presignatures_to_buffer: DEFAULT_PRESIGNATURES_TO_BUFFER,
             sandbox_version: DEFAULT_SANDBOX_VERSION.to_string(),
             home_base: None,
+            initial_participant_indices: (0..3_usize).collect(),
```
kevindeforth: nit: Since we don't allow initializing a contract with no participants, I would suggest to consider an empty vector here as an instruction to initialize the cluster with all participants.

crates/e2e-tests/src/cluster.rs
```diff
@@ -467,18 +686,28 @@ fn create_test_dir(home_base: &Option<PathBuf>) -> anyhow::Result<tempfile::Temp
     }
 }
 
-fn generate_node_keys(num_nodes: u64) -> (Vec<SigningKey>, Vec<SigningKey>, Vec<SigningKey>) {
+fn generate_node_keys(
```
kevindeforth: also nit: the function name could be improved. `voting_key` isn't really a node key, it's an operator key.

Maybe just `generate_signing_keys`?

crates/e2e-tests/tests/common.rs
anodar: We're not adding domain later, we do pass domain to setup with that configure callback. I think it's nicer that way because we're explicitly configuring domains rather than always having them.

crates/e2e-tests/src/cluster.rs
anodar: Yes I think resharing cancellation test was timing out but that's in followup PR so reverted this change and I'll cross that bridge once I get there.

crates/e2e-tests/src/cluster.rs

crates/e2e-tests/Cargo.toml
anodar: No it's same approach as in chain gateway, to have e2e.rs as single binary.

Makefile.toml
anodar: That produces large wasm even without no-wasmopt and with "--no-abi". I can't find any example of `cargo near build non-reproducible-wasm....` in the codebase (that's then used in tests) that would work here. 

CI also does exactly how it's done here.

crates/e2e-tests/tests/request_lifecycle.rs
anodar: You mean basically merging this and mpc_cluster__should_successfully_process_robust_ecdsa_requests below ?  I opted for them being separate since cluster setup is different and those two tests can run concurrently instead of sequentially.

Added CKD though.

crates/e2e-tests/tests/common.rs
anodar: Done.

crates/e2e-tests/tests/common.rs
anodar: Done.

deny.toml
anodar: Yeah previous commit was failing on that: https://github.com/near/mpc/actions/runs/24216136234/job/70697013260

You're right, it's not required check so I can revert that.

crates/e2e-tests/src/cluster.rs
anodar: Leaving as is for now. We can change to more suitable type once/if we start using the error more meaningfully.

crates/e2e-tests/src/cluster.rs
netrome: Sure, but if you break out the health check to a separate function I do think it does read better:

```rust
    pub async fn wait_for_node_healthy(&self, idx: usize) -> anyhow::Result<()> {
        let node = match &self.nodes[idx] {
            MpcNodeState::Running(n) => n,
            _ => anyhow::bail!("node {idx} is not running"),
        };
        let ctx = HealthCheckCtx {
            client: reqwest::Client::new(),
            url: format!("http://{}/health", node.web_address()),
            node_idx: idx,
        };
        check_if_node_is_healthy
            // 120s deadline: 120_000ms / 500ms = 240 attempts
            .retry(
                ConstantBuilder::default()
                    .with_delay(POLL_INTERVAL)
                    .with_max_times(240),
            )
            .context(ctx)
            .await
            .1
            .with_context(|| format!("node {idx} did not become healthy within 120s"))
    }
```

with these helpers:

```rust
struct HealthCheckCtx {
    client: reqwest::Client,
    url: String,
    node_idx: usize,
}

impl HealthCheckCtx {
    async fn check(&self) -> anyhow::Result<()> {
        let resp = self.client.get(&self.url).send().await?;
        anyhow::ensure!(resp.status() == 200, "not yet healthy");
        tracing::info!(node = self.node_idx, "node is healthy");
        Ok(())
    }
}

fn check_if_node_is_healthy(
    ctx: HealthCheckCtx,
) -> impl Future<Output = (HealthCheckCtx, Result<(), anyhow::Error>)> {
    async move {
        let result = ctx.check().await;
        (ctx, result)
    }
}
```

note that this requires `backon::RetryableWithContext` to be in scope.

netrome: But feel free to do whatever you prefer here. I have no strong opinions either way. A plain loop also works fine, and is arguable simpler than the backon future+closure+trait magic.

netrome (dismissed):
Thanks for the updates

crates/e2e-tests/src/cluster.rs
```diff
@@ -712,19 +968,28 @@ async fn wait_for_contract_state(
     timeout: Duration,
     predicate: impl Fn(&ProtocolContractState) -> bool,
 ) -> anyhow::Result<()> {
-    let deadline = tokio::time::Instant::now() + timeout;
-    loop {
+    // timeout / 500ms attempts
```
kevindeforth: nit: comment can be removed. the 500ms will be inaccurate as soon as we change `POLL_TIMEOUT`.

crates/e2e-tests/src/cluster.rs
```diff
@@ -712,19 +968,28 @@ async fn wait_for_contract_state(
...
+            .with_delay(POLL_INTERVAL)
+            .with_max_times(max_times),
+    )
+    .await
+    .with_context(|| {
+        format!(
+            "contract state predicate not satisfied within {}s",
+            timeout.as_secs()
+        )
+    })
```
kevindeforth, line 1010: remark:

I see this was suggested in a different context by @netrome (https://github.com/near/mpc/pull/2759#discussion_r3059936312).

For the record: I don't think this is better than the simple loop it used to be. Went from 14 lines to 22 lines and requires the reader to know about `ConstantBuilder`.

No need to change antyhing, this relates to form and I think these are personal preferences.

crates/e2e-tests/src/mpc_node.rs
```diff
@@ -207,7 +231,30 @@ impl MpcNodeSetup {
...
+    /// restart, so the entire home directory is removed and recreated. The NEAR
+    /// indexer will re-init from genesis_path/boot_nodes in start_config.toml
+    /// and sync from block 0.
+    ///
+    /// Re-writes secrets.json and start_config.toml from the saved setup.
+    pub fn reset_mpc_state(&self) -> anyhow::Result<()> {
+        // Nuke and recreate the home directory. SIGKILL can leave NEAR indexer
+        // data in a corrupted state that prevents restart, so we start fully
+        // fresh. The NEAR indexer will re-init from genesis_path/boot_nodes in
+        // start_config.toml and sync from block 0.
```
kevindeforth: What's the purpose of this comment? It seems to state the same thing from a few lines above.

crates/e2e-tests/src/mpc_node.rs
```diff
@@ -81,6 +87,24 @@ impl MpcNode {
         Ok(None)
     }
 
+    /// Create a marker file that tells the MPC node a key event attempt was
+    /// already started. The node reads these on startup to determine which
+    /// attempt ID to use next.
+    pub fn reserve_key_event_attempt(
```
kevindeforth: This this method is not used anywhere.

crates/e2e-tests/src/mpc_node.rs
kevindeforth: What's the purpose of this comment? It seems to state the same thing from a few lines above.

crates/e2e-tests/tests/common.rs
```diff
@@ -0,0 +1,150 @@
...
+        .await
+        .expect("cluster did not reach Running state");
+    let running = match cluster
+        .get_contract_state()
+        .await
+        .expect("failed to get contract state")
+    {
+        ProtocolContractState::Running(r) => r,
+        other => panic!("expected Running state, got: {other:?}"),
+    };
```
kevindeforth: We could be a bit more efficient if `wait_for_state` returned the value of the protocol state.

```rust
let protocol_state = cluster
        .wait_for_state(
            |s| matches!(s, ProtocolContractState::Running(_)),
            CLUSTER_WAIT_TIMEOUT,
        )
        .await
        .expect("cluster did not reach Running state");

let ProtocolContractState::Running(running) = protocol_state else {
    panic!(...)
};
```

crates/e2e-tests/src/cluster.rs
```diff
@@ -712,19 +969,28 @@ async fn wait_for_contract_state(
     timeout: Duration,
     predicate: impl Fn(&ProtocolContractState) -> bool,
 ) -> anyhow::Result<()> {
```
kevindeforth: In at least one place, we call this method followed by  `get_contract_state()`. Why don't we have it return the contract state on success and save us an RPC query?

crates/e2e-tests/tests/common.rs
```diff
@@ -0,0 +1,150 @@
...
+pub async fn wait_for_presignatures(
+    cluster: &MpcCluster,
+    participant_indices: &[usize],
+    presignatures_to_buffer: usize,
+) {
+    let expected = i64::try_from(presignatures_to_buffer).expect("presignatures exceeds i64::MAX");
+    let deadline = tokio::time::Instant::now() + CLUSTER_WAIT_TIMEOUT;
+    loop {
+        let values = cluster
+            .get_metric_all_nodes("mpc_owned_num_presignatures_available")
```
kevindeforth: I would prefer if we declared these as const variables in a dedicated file. That way, it's easy to spot which metrics these tests depend on.

crates/e2e-tests/tests/common.rs
```diff
@@ -0,0 +1,150 @@
...
+        .map(PathBuf::from)
+        .unwrap_or_else(|_| {
+            Path::new(env!("CARGO_MANIFEST_DIR"))
+                .join("../../target/wasm32-unknown-unknown/release-contract/mpc_contract.wasm")
+        });
+
+    std::fs::read(&wasm_path).unwrap_or_else(|e| {
+        panic!(
+            "Failed to read contract WASM at {}: {e}\n\
+             Build it first: cargo make build-contract-wasm",
```
kevindeforth: Hmm, I would prefer we don't require cargo make for this and if the tests compiled the contracts before use. 

Are you aware of the contract [builder crate](https://github.com/near/mpc/blob/main/crates/test-utils/src/contract_build.rs)?

I would hope for `cargo nextest run` to be sufficient for running all tests in this workspace and not introduce additional manual steps.

Right now, these tests fail on my machine when I run them with nextest:

```bash
    thread 'web_endpoints::test_web_endpoints' panicked at crates/e2e-tests/tests/common.rs:60:10:
    failed to start cluster: failed to create account and deploy to mpc.sandbox: Network error: HTTP 413 Payload Too Large: Failed to buffer the request body: length limit exceeded
```

crates/e2e-tests/tests/request_lifecycle.rs
```diff
@@ -1,133 +1,122 @@
...
-            "Failed to read contract WASM at {}: {e}\n\
-             Build it first:\n  \
-             cargo build -p mpc-contract --target=wasm32-unknown-unknown --profile=release-contract --locked\n  \
-             wasm-opt -Oz <path> -o <path>",
-            wasm_path.display()
-        )
-    })
-}
+    let mut rng = rand::rngs::StdRng::seed_from_u64(0);
+    for domain in &running.domains.domains {
```
kevindeforth, line 20: Here, it would be diligent to add a check that the domains vector is of a certain minimal length.

crates/e2e-tests/tests/common.rs
kevindeforth: > shouldn't this function decide which domains to add? I guess for simplicity we can add all of them always, but that will have a performance hit


Yeah, I would prefer if this was set by the test itself.

crates/e2e-tests/tests/request_lifecycle.rs
```diff
@@ -1,133 +1,122 @@
-use std::path::{Path, PathBuf};
-use std::time::Duration;
+#![expect(non_snake_case)]
```
kevindeforth: I would prefer if we added this specifically where it's needed and not for the entire file.

crates/e2e-tests/tests/request_lifecycle.rs
```diff
@@ -1,133 +1,122 @@
...
-        .expect("failed to start cluster");
+                match (&domain.scheme, &signature) {
+                    (SignatureScheme::Secp256k1, SignatureResponse::Secp256k1(_)) => {}
+                    (SignatureScheme::Ed25519, SignatureResponse::Ed25519 { .. }) => {}
+                    _ => panic!(
+                        "signature scheme mismatch: requested {:?}, got {:?}",
+                        domain.scheme, signature
+                    ),
+                }
+                tracing::info!(domain_id = ?domain.id, "sign request returned valid signature");
```
kevindeforth, line 56: out of scope: would be nice to add validation logic and verify the signature. The contract already does it, but it wouldn't hurt to add it here as a sanity check.

crates/e2e-tests/tests/submit_participant_info.rs
```diff
@@ -0,0 +1,42 @@
...
+/// Tests that non-participant MPC nodes automatically submit their TEE
+/// attestations to the contract after startup.
+///
+/// Setup: 4 nodes started, but only 2 are initial participants. The other 2
+/// should submit attestations on their own. We verify all 4 appear in the
+/// contract's TEE accounts list.
+#[tokio::test]
+async fn test_submit_participant_info() {
+    let (cluster, _running) =
+        common::setup_cluster(common::SUBMIT_PARTICIPANT_INFO_PORT_SEED, |c| {
```
kevindeforth, line 14: hmm, so spinning up these clusters is quite expensive and takes a bit of time. It would be nice if we had some logic to share clusters between tests, similar to the `shared_cluster` from the pytests. It led to significant reduction in ci time.

kevindeforth (commented):
First pass.
I don't have anything blocking, just a few nits and general observations. You already have two approvals, so feel free to merge.

crates/e2e-tests/src/blockchain.rs
anodar: Done.

crates/e2e-tests/src/cluster.rs

crates/e2e-tests/src/cluster.rs

crates/e2e-tests/src/cluster.rs

crates/e2e-tests/src/mpc_node.rs
anodar: Dropped.

crates/e2e-tests/tests/common.rs
anodar: Done.

crates/e2e-tests/src/cluster.rs

crates/e2e-tests/tests/common.rs
anodar: Done.

crates/e2e-tests/tests/common.rs
anodar: Done.

crates/e2e-tests/tests/request_lifecycle.rs
anodar: Done.

crates/e2e-tests/tests/request_lifecycle.rs
anodar: It's 100% of functions in this file that require it :D  Changed.

crates/e2e-tests/tests/request_lifecycle.rs
anodar: Will follow up in next PR's as it's explicitly marked out of scope.

crates/e2e-tests/tests/submit_participant_info.rs
anodar: Tests are running concurrently, so it's not a bottleneck, but will introduce if it does become as I port more tests.

crates/e2e-tests/tests/common.rs
```diff
@@ -0,0 +1,155 @@
...
+        return std::fs::read(&default_path).unwrap_or_else(|e| {
+            panic!(
+                "Failed to read contract WASM at {}: {e}",
+                default_path.display()
+            )
+        });
+    }
+
+    tracing::info!("MPC_CONTRACT_WASM not set and pre-built WASM not found — building contract");
+    test_utils::contract_build::ContractBuilder::new("crates/contract/Cargo.toml").build()
```
kevindeforth, line 137: might make sense to run this either way, just to ensure that the contract is up to date when running the tests.

Similarly, might make sense to re-compile the node.

In either case, I these things are follow-up, lets merge this chunker.

kevindeforth (approved):
Thanks a lot!

netrome (approved):

PR #2777, opened 2026-04-09
refactor: simplify contract error handling

### LINKED ISSUE #2402: Fix typos in error messages
### Background

Two error messages in `crates/contract/src/errors.rs` contain typos: `"dose"` instead of `"does"` (line 8) and `"requestued"` instead of `"requested"` (line 42).

### User Story

As an API consumer, I need correct error messages so that I can understand failures without confusion.

### Acceptance Criteria

- [ ] Fix `"Node dose not have an ongoing recovery"` → `"Node does not have an ongoing recovery"`
- [ ] Fix `"requestued key's scheme"` → `"requested key's scheme"`

Finding: `KD-F037-typos-in-error-messages.md` — https://github.com/near/mpc-private/pull/214

kevindeforth:
resolves https://github.com/near/mpc/issues/2776 and https://github.com/near/mpc/issues/2402

Removes a bunch of unused errors.
Reduces contract size by approximately 1k bytes.


kevindeforth:
> Minor note (non-blocking): The test in `vote.rs:790-797` constructs an `InvalidTeeRemoteAttestation { reason: String::new() }` to get a prefix string for `contains()` matching — it works but is a bit roundabout. A simple `error_message.contains("Invalid TEE Remote Attestation:")" would be more readable. Not a blocker.

We have a bunch of hard-coded string message comparisons throughout this codebase and they are generally not desired. If the error message changes, the test will also need to be adjusted.
Constructing the error, while cumbersome, makes the test more robust.

crates/contract/src/errors.rs
```diff
@@ -28,22 +30,15 @@ pub enum RequestError {
...
 #[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
 pub enum RespondError {
     #[error("The provided signature is invalid.")]
     InvalidSignature,
-    #[error("The provided signature scheme does not match the requestued key's scheme")]
-    SignatureSchemeMismatch,
+    #[error("The provided signature scheme does not match. MPC response: {mpc_scheme:?}, user request: {user_scheme:?}")]
+    SignatureSchemeMismatch {
+        mpc_scheme: Box<dtos::SignatureResponse>,
+        user_scheme: Box<crate::crypto_shared::types::PublicKeyExtended>,
```
gilcu3, line 40: this looks pretty weird. Why do we need a Box? The current message also seems less clear than the previous one

crates/contract/src/errors.rs
```diff
@@ -72,18 +65,8 @@ pub enum KeyEventError {
...
-    #[error("Voting account is neither a participant, nor a proposed participant.")]
-    VoterNotParticipantNorProposedParticipant,
-    #[error("This participant already registered a vote.")]
-    ParticipantVoteAlreadyRegistered,
     #[error("Voting account is not the leader of the current reshare or keygen instance.")]
     VoterNotLeader,
-    #[error("Inconsistent voting state")]
-    InconsistentVotingState,
-    #[error("Voter already aborted the current key event.")]
-    VoterAlreadyAborted,
```
gilcu3, line 86: were these errors not used? or are they defined somewhere else now?

crates/contract/tests/sandbox/sign.rs
```diff
@@ -266,7 +266,7 @@ async fn test_contract_request_deposits_all_schemes() -> anyhow::Result<()> {
             .into_result()
             .unwrap_err()
             .to_string()
-            .contains(&errors::InvalidParameters::InsufficientDeposit.to_string()));
+            .contains("Attached deposit is lower than required"));
```
gilcu3, line 269: is it not possible to convert the expected error to string anymore?

crates/contract/tests/sandbox/tee.rs
```diff
@@ -154,9 +153,11 @@ async fn test_vote_code_hash_doesnt_accept_account_id_not_in_participant_list()
             "vote_code_hash should not accept votes from a randomly generated account id that is not in the participant list"
         );
     };
-    let expected = format!("{:?}", InvalidState::NotParticipant);
     let err_str = format!("{:?}", err);
-    assert!(err_str.contains(&expected));
+    assert!(
+        err_str.contains("NotParticipant"),
+        "Expected NotParticipant error, got: {err_str}"
+    );
```
gilcu3, line 160: also here, not possible anymore to match against the string representation of the error?

crates/contract/src/lib.rs
```diff
@@ -729,11 +719,7 @@ impl MpcContract {
             public_key: dtos::PublicKey::Bls12381(public_key),
         } = self.public_key_extended(request.domain_id)?
         else {
-            env::panic_str(
-                &InvalidParameters::InvalidDomainId
-                    .message("Selected domain is not compatible with CKD")
-                    .to_string(),
-            );
+            env::panic_str("Domain is not compatible with CKD (expected Bls12381 curve)");
```
gilcu3, line 722: hmm I see, so this is where some error types were just removed.

gilcu3 (approved):
Thank you!

Left some minor comments

crates/contract/tests/sandbox/tee.rs
anodar: I think we should be able to do similarly to how it is below with InvalidTeeRemoteAttestation.

We just need to convert `random_account` (`near_workspaces::types::account type`) to `near_account_id::AccountId;` type. 

Claude says should be possible with `random_account.id().to_string().parse().unwrap()`

crates/contract/tests/sandbox/sign.rs
anodar: This seems tricky because now `InsufficientDeposit` looks like this:

`InsufficientDeposit { attached: u128, required: u128 }`, you can't easily calculate how much was required I assume.

We could do string matching differently though to improve it a bit by doing something like: adding try_from implementation to InvalidParameters and then do string matching there. That way hardcoded error strings matching are contained next to definition and every time you need to this error in some test you won't have to hardcode string. Will be easier to update error string too (2 places instead of (number of tests using it) + 1.

anodar (approved):

crates/contract/src/errors.rs
kevindeforth: they were not used.

crates/contract/src/lib.rs
kevindeforth: no, not multiple. Just this one (It was misleading and applied inconsistently).

crates/contract/src/errors.rs
kevindeforth: Box is diligent, because these types have very different sizes.

crates/contract/tests/sandbox/tee.rs
kevindeforth: The problem with matching on error types is that the contract doesn't output json errors. It outputs strings.
We can change that, but it's orthogonal to this PR.

crates/contract/tests/sandbox/sign.rs
kevindeforth: > We could do string matching differently though to improve it a bit by doing something like: adding try_from implementation to InvalidParameters and then do string matching there. That way hardcoded error strings matching are contained next to definition and every time you need to this error in some test you won't have to hardcode string. Will be easier to update error string too (2 places instead of (number of tests using it) + 1.

That is a valid way to do this.
But I don't think the boilerplate is worth it. If we change the error message, the test will fail and we can fix it, so there isn't a huge risk.

It's not ideal and I would prefer to be matching on the error type, but for that, we need to change the return type, c.f.  https://github.com/near/mpc/pull/2777#discussion_r3063095066

crates/contract/tests/sandbox/tee.rs
anodar: I didn't mean matching error type, but constructing same error (with account id) and then converting to string and comparing like that instead of hardcoding string that error contains.

PR #2780, opened 2026-04-09
build: switch MPC node image push to skopeo with --preserve-digests

barakeinav1:
## Summary

- Switch MPC node and node-gcp image push from `docker push` to `skopeo copy --preserve-digests`, matching the pattern already used for launcher images
- Add `skopeo` to required commands when `--push` is used
- Output manifest digest after push for both node images

Closes #2778

## Test plan

- [x] CI passes
- [x] `./deployment/build-images.sh --node --push` uses skopeo and outputs manifest digest
- [x] check https://github.com/near/mpc/actions/workflows/docker_build_node_gcp.yml and https://github.com/near/mpc/actions/workflows/docker_build_node.yml work

### DIFF
--- a/deployment/build-images.sh
+++ b/deployment/build-images.sh
@@ -181,18 +181,22 @@ if $USE_PUSH; then
 
     if $USE_NODE; then
         temp_dir=$(mktemp -d)
-        skopeo copy --all --dest-compress docker-daemon:$NODE_IMAGE_NAME:latest dir:$temp_dir
-        skopeo copy --preserve-digests dir:$temp_dir docker://docker.io/nearone/$NODE_IMAGE_NAME:$image_tag
-        node_manifest_digest="sha256:$(sha256sum $temp_dir/manifest.json | cut -d' ' -f1)"
-        rm -rf "$temp_dir"
+        trap 'rm -rf -- "$temp_dir"' EXIT
+        skopeo copy --all --dest-compress "docker-daemon:$NODE_IMAGE_NAME:latest" "dir:$temp_dir"
+        skopeo copy --preserve-digests "dir:$temp_dir" "docker://docker.io/nearone/$NODE_IMAGE_NAME:$image_tag"
+        node_manifest_digest="sha256:$(sha256sum "$temp_dir/manifest.json" | cut -d' ' -f1)"
+        rm -rf -- "$temp_dir"
+        trap - EXIT
     fi
 
     if $USE_NODE_GCP; then
         temp_dir=$(mktemp -d)
-        skopeo copy --all --dest-compress docker-daemon:$NODE_GCP_IMAGE_NAME:latest dir:$temp_dir
-        skopeo copy --preserve-digests dir:$temp_dir docker://docker.io/nearone/$NODE_GCP_IMAGE_NAME:$image_tag
-        node_gcp_manifest_digest="sha256:$(sha256sum $temp_dir/manifest.json | cut -d' ' -f1)"
-        rm -rf "$temp_dir"
+        trap 'rm -rf -- "$temp_dir"' EXIT
+        skopeo copy --all --dest-compress "docker-daemon:$NODE_GCP_IMAGE_NAME:latest" "dir:$temp_dir"
+        skopeo copy --preserve-digests "dir:$temp_dir" "docker://docker.io/nearone/$NODE_GCP_IMAGE_NAME:$image_tag"
+        node_gcp_manifest_digest="sha256:$(sha256sum "$temp_dir/manifest.json" | cut -d' ' -f1)"
+        rm -rf -- "$temp_dir"
+        trap - EXIT
     fi
 
     if $USE_RUST_LAUNCHER; then
--- a/deployment/build-images.sh
+++ b/deployment/build-images.sh
@@ -70,6 +70,10 @@ if $USE_NODE || $USE_RUST_LAUNCHER; then
     require_cmds repro-env podman
 fi
 
+if $USE_PUSH; then
+    require_cmds skopeo
+fi
+
 if ! docker buildx &>/dev/null; then
   die "Please install docker-buildx"
 fi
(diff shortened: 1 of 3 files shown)


deployment/build-images.sh
```diff
@@ -176,13 +180,23 @@ if $USE_PUSH; then
...
     if $USE_NODE; then
-        docker tag $NODE_IMAGE_NAME nearone/$NODE_IMAGE_NAME:$image_tag
-        docker push nearone/$NODE_IMAGE_NAME:$image_tag
+        temp_dir=$(mktemp -d)
+        trap 'rm -rf -- "$temp_dir"' EXIT
+        skopeo copy --all --dest-compress "docker-daemon:$NODE_IMAGE_NAME:latest" "dir:$temp_dir"
+        skopeo copy --preserve-digests "dir:$temp_dir" "docker://docker.io/nearone/$NODE_IMAGE_NAME:$image_tag"
+        node_manifest_digest="sha256:$(sha256sum "$temp_dir/manifest.json" | cut -d' ' -f1)"
+        rm -rf -- "$temp_dir"
+        trap - EXIT
```
gilcu3: why are these two needed? the first no needed because this will only run in CI

deployment/build-images.sh
```diff
@@ -176,13 +180,23 @@ if $USE_PUSH; then
...
     fi
 
     if $USE_NODE_GCP; then
-        docker tag $NODE_GCP_IMAGE_NAME nearone/$NODE_GCP_IMAGE_NAME:$image_tag
-        docker push nearone/$NODE_GCP_IMAGE_NAME:$image_tag
+        temp_dir=$(mktemp -d)
+        trap 'rm -rf -- "$temp_dir"' EXIT
+        skopeo copy --all --dest-compress "docker-daemon:$NODE_GCP_IMAGE_NAME:latest" "dir:$temp_dir"
+        skopeo copy --preserve-digests "dir:$temp_dir" "docker://docker.io/nearone/$NODE_GCP_IMAGE_NAME:$image_tag"
+        node_gcp_manifest_digest="sha256:$(sha256sum "$temp_dir/manifest.json" | cut -d' ' -f1)"
```
gilcu3: nit: we could put this in a function

gilcu3 (dismissed):
Thank you!

barakeinav1:
**"nit: we could put this in a function"**: Done — extracted `skopeo_push` helper. All four image types (launcher, node, node-gcp, rust-launcher) now use the same function and consistently output manifest digests.

deployment/build-images.sh
barakeinav1: added

deployment/build-images.sh
barakeinav1: this just a good practice to clean things up,
but you are right, since this only runs in CI, the cleanup isn't strictly needed. 
I can  remove if you prefer keeping it minimal

gilcu3: the `trap - EXIT` in particular seems is only needed to guarantee the line before is executed, and that one (rm -rf) is never needed in CI, so just to keep the simplicity of the script we better remove them both. putting traps could have future unintended effects once the script is modified for example

deployment/build-images.sh
```diff
@@ -169,41 +169,36 @@ if $USE_PUSH; then
...
+        local image_name="$1"
+        local tag="$2"
+        local td
+        td=$(mktemp -d)
+        trap 'rm -rf -- "$td"' EXIT
+        skopeo copy --all --dest-compress "docker-daemon:${image_name}:latest" "dir:$td"
+        skopeo copy --preserve-digests "dir:$td" "docker://docker.io/nearone/${image_name}:${tag}"
+        echo "sha256:$(sha256sum "$td/manifest.json" | cut -d' ' -f1)"
+        rm -rf -- "$td"
+        trap - EXIT
```
gilcu3: please remove these lines, which will never be needed in CI and just make following the logic harder

```
trap 'rm -rf -- "$td"' EXIT
rm -rf -- "$td"
trap - EXIT
```

gilcu3 (dismissed):

deployment/build-images.sh
gilcu3: notice even in dev computers, something you put in temp gets deleted eventually, so most software does not go to great length to clean files there

deployment/build-images.sh
```diff
@@ -165,31 +169,36 @@ if $USE_PUSH; then
...
```
netrome: Do we have any exit handler that needs to be cleared?

netrome (dismissed):
Would be nice to remove the lines Reynaldo suggested but I don't consider it a hard blocker.

barakeinav1:
Removed trap/cleanup from `skopeo_push` in dfd394d4 — agreed, not needed in CI and adds unnecessary complexity.

deployment/build-images.sh
barakeinav1: removed

deployment/build-images.sh
barakeinav1: removed

deployment/build-images.sh
barakeinav1: removed in dfd394d4 — the function no longer sets or clears any traps.

deployment/build-images.sh
```diff
@@ -165,31 +169,33 @@ if $USE_PUSH; then
...
+        echo "sha256:$(sha256sum "$td/manifest.json" | cut -d' ' -f1)"
+    }
+
     if $USE_LAUNCHER; then
-        temp_dir=$(mktemp -d)
-        echo "using $temp_dir"
-        # This compresses the built image to a local directory, which implicitly computes the manifest
-        # digest in $temp_dir/manifest.json
-        skopeo copy --all --dest-compress docker-daemon:$LAUNCHER_IMAGE_NAME:latest dir:$temp_dir
-        # Then we publish the image from the directory, making sure the manifest digest does not change
```
gilcu3, line 174: nit: would be good to keep this comments in the new function, as they kind of explain the hack that achieves repro

gilcu3 (dismissed):

deployment/build-images.sh
barakeinav1: done

gilcu3: I see no commit :(

netrome (dismissed):

deployment/build-images.sh
barakeinav1: Added back in 4449ae78 — restored the comments explaining the two-step skopeo flow.

barakeinav1: The trap removal was in dfd394d4, and the comments were restored in 4449ae78. Both should be visible now.

gilcu3 (approved):

netrome (approved):

PR #2784, opened 2026-04-09
feat(contract): API to submit foreign chain config without RPC url consensus

### LINKED ISSUE #2783: contract allows nodes to submit configuration without requiring consensus on RPC providers
### Background

implement contract part of parent issue

### User Story

_No response_

### Acceptance Criteria

implement contract part of parent issue

_No response_

DSharifi:
closes #2783 

This PR allows nodes to submit their RPC configuration on chain, and exposes a new API to get supported chains which doesn't require all nodes to use same RPC urls. Instead a chain is considered "supported" as long as all nodes have a configuration of it, regardless of their RPC urls.

The previous API to submit "policies" will be deprecated in a follow up PR.

For all the changes that will be done see  https://github.com/near/mpc/pull/2779


DSharifi:
> **One concern (non-blocking):**
> 
> Migration panics on non-Running state** (`v3_8_1_state.rs:46`): The `From<MpcContract>` migration panics if the contract isn't in `Running` state.
>

Migrations always happen in running state. We always use this pattern in the From implementation for previous states.

crates/contract/src/lib.rs
netrome, line 1: Hmm, beware that we're getting dangerously close to contract size limits. We're already at 1512365 bytes on `main` right now. This PR puts us at 1542950 bytes (measured now on my machine).

I guess this will be reduced once we remove the foreign chain policy votes, which we'll do once this is released right?

crates/contract/src/lib.rs
```diff
@@ -187,6 +191,34 @@ impl ForeignChainPolicyVotes {
     }
 }
 
+#[near(serializers=[borsh])]
+#[derive(Debug)]
+struct ForeignChainSupport {
+    votes_per_chain: IterableMap<dtos::AccountId, dtos::ForeignChainConfiguration>,
```
netrome: These aren't votes right? Or I'm not seeing any aggregation or thing that happens once votes reach a quorum. Isn't this just a configuration? The name looks misleading to me.

netrome (changes requested):
Mostly looks good to me but I find the `votes_per_chain` name misleading which made me confused when I initially read this.

crates/contract/src/lib.rs
DSharifi: > I guess this will be reduced once we remove the foreign chain policy votes, which we'll do once this is released right?

Yes

crates/contract/src/lib.rs
DSharifi: Yeah, this is a good point.

How about `foreign_chain_configuration_by_node`?

netrome: Sounds good.

DSharifi: Updated

crates/contract/tests/sandbox/foreign_chain_policy.rs
kevindeforth, line 1: Reading the acceptance criteria of the issue:

> - nodes can use RPC providers configured locally, without requiring voting nor unanimous consensus on it.
> - nodes still publish their RPC configuration for on chain visibility

I would have expected a test where nodes submit different rpc configs for the same chain, and that chain still being supported.

edit: we have this in `lib.rs`, it's called `get_supported_foreign_chains__different_rpc_urls_per_participant_is_fine`

kevindeforth, line 1: There is quite a bit of code duplication in this file.
It would be nice to introduce helper methods for registering and chain configurations and fetching supported chains, such that we reduce boilerplate and make it easier to write or extend tests.

crates/contract/tests/sandbox/foreign_chain_policy.rs
```diff
@@ -184,3 +188,173 @@ async fn vote_foreign_chain_policy_deduplicates_duplicate_chain_keys() {
...
+            .transact()
+            .await
+            .unwrap()
+            .into_result();
+        assert_matches!(result, Ok(_));
+    }
+
+    // Then: only Bitcoin is in the supported set (Starknet is not unanimous)
+    let supported: Vec<String> = contract
+        .view("get_supported_foreign_chains")
```
kevindeforth, line 308: Lets replace this string with `GET_SUPPORTED_FOREIGN_CHAINS` from the contract interface crate.

crates/contract/src/v3_8_1_state.rs
```diff
@@ -43,20 +43,142 @@ pub struct MpcContract {
...
+        let supported: dtos::SupportedForeignChains = policy
+            .chains
+            .keys()
+            .cloned()
+            .collect::<BTreeSet<_>>()
+            .into();
+
+        assert_eq!(supported.len(), all_chains.len());
+        for chain in &all_chains {
+            assert!(supported.contains(chain));
```
kevindeforth, line 181: I must be missing something. We are not testing actual migration logic here, but verifying properties of the output of `make_policy`, which is a function defined in this test module.

If we want to test the actual migration logic, we need to construct an instance of `MpcContract` and test the migration method above, or break the relevant logic out to a function and test that in isolation.

crates/contract/src/lib.rs
```diff
@@ -187,6 +191,37 @@ impl ForeignChainPolicyVotes {
     }
 }
 
+#[near(serializers=[borsh])]
+#[derive(Debug)]
+struct NodeForeignChainConfigurations {
+    foreign_chain_configuration_by_node:
+        IterableMap<dtos::AccountId, dtos::ForeignChainConfiguration>,
```
kevindeforth, line 198: We should clean this map and remove former participants when they leave the network - I don't see that logic anywhere.

_edit: Closing, I later found there was a todo for this._

crates/contract/src/lib.rs
```diff
@@ -1846,6 +1907,54 @@ impl MpcContract {
         self.foreign_chain_policy_votes.to_dto()
     }
 
+    pub fn get_supported_foreign_chains(&self) -> dtos::SupportedForeignChains {
```
kevindeforth, line 1910: Do we have benchmarks for the gas costs of this method?
Did we consider alternative data structures for `foreign_chain_configuration_by_node`?

It's currently `IterableMap<dtos::AccountId, dtos::ForeignChainConfiguration>`, which is a nested map:
`Participant <-> { ForeignChain <-> Set<RpcProvider>> }`, which essentially is: `AccountId <-> {Enum <-> Set<String> }`

It seems to me this is optimizing for insertion, but not for lookup.
Right now, we only use it to check whether a specific foreign chain is supported by all participants, which would be more suitable for `IterableMap<ForeginChain, BTreeSet<AccountId>>`.

kevindeforth (approved):
some comments, but nothing blocking.

crates/contract/src/lib.rs
DSharifi: ~~It's a read method. We don't use gas for reads~~

Edit: It's used in the verify foreign transaction function

netrome (approved):
Thanks for updating!

PR #2785, opened 2026-04-09
refactor: simplify launcher image validation by accepting manifest digest directly

### LINKED ISSUE #2613: [rust-launcher] Add manifest traversal depth limit
## Context

Follow-up from PR #2326 (port node launcher to Rust). Tracked in #2598.

## Problem

`get_manifest_digest()` in `crates/tee-launcher/src/main.rs` (line 302) processes tags from a VecDeque. Multi-platform manifests add child digests back to the deque. In theory, this loop is unbounded if manifests keep referencing other manifests.

## Action

Add a `MAX_MANIFEST_DEPTH` constant (e.g., 10) and track traversal depth. Return an error if exceeded.

Low priority -- unlikely with real Docker registries.

### LINKED ISSUE #2627: [rust-launcher] Simplify image validation by accepting manifest hash directly
### Background

The Rust tee-launcher currently stores and validates MPC images using the **config digest** (image ID from `docker inspect`). To pull an image by digest, the launcher must:

1. Authenticate with Docker Hub (`auth.docker.io/token`)
2. Iterate through image tags, fetching manifests with retry logic
3. Handle multi-arch image indexes (find amd64/linux platform)
4. Match the config digest against the approved hash
5. Extract the manifest digest from the `Docker-Content-Digest` response header
6. Pull by manifest digest

This is ~150 lines of registry API code (`get_manifest_digest` in `main.rs`).

Now that we have reproducible Docker builds, the **manifest digest is deterministic** — it can be known at voting time. If the contract stored manifest digests instead of (or in addition to) config digests, the launcher could skip the registry API entirely and just `docker pull image@sha256:<manifest_hash>`, then verify the config digest locally after pulling.

Reference: [gilcu3's comment on PR #2621](https://github.com/near/mpc/pull/2621#discussion_r2994674420)

### User Story

As a launcher developer, I want to simplify the image validation flow so there is less code to maintain and fewer external API calls that can fail.

### Acceptance Criteria

- [ ] Evaluate whether the contract should store manifest digests, config digests, or both
- [ ] If manifest digests: update voting flow and contract storage
- [ ] Simplify `get_manifest_digest` to direct pull-by-digest
- [ ] Verify the config digest locally after pulling
- [ ] Remove or simplify registry auth/retry code

- Current implementation: `get_manifest_digest()` in `crates/tee-launcher/src/main.rs`
- PR #2621 (Rust launcher crate)
- This is a follow-up optimization, not a blocker for the initial launcher release

barakeinav1:
## Summary

Remove the Docker registry API from the tee-launcher. The approved hashes file now contains manifest digests, so the launcher can `docker pull image@sha256:<digest>` directly without querying the registry.

**Removed (~630 lines):**
- `registry.rs` entirely — Docker registry auth, manifest fetching, retry/backoff, multi-platform resolution, config digest matching
- `oci-client`, `backon`, `rustls`, `rstest` dependencies
- All registry-related error variants
- rustls crypto provider setup in `main.rs`
- `external-services-tests` feature flag

**Simplified:**
- `validation.rs`: 79 → 33 lines (just `docker pull` + verify exit code)
- `error.rs`: 90 → 65 lines
- `lib.rs`: now emits manifest digest to RTMR3 (not config digest)

**Note:** `LauncherConfig` still accepts `rpc_*` fields for backward compat with existing TOML configs. They are unused — removal is a follow-up.

Closes #2627. Plan and research in the [issue comment](https://github.com/near/mpc/issues/2627#issuecomment-4214536344).

also closes #2613

## Requires coordination

- The approved hashes file must contain manifest digests (not config digests)
- The contract/node side needs to store and propagate manifest digests — tracked separately
- Build script must output manifest digests (#2778)

## Test plan

- [x] `cargo check -p tee-launcher` passes
- [x] `cargo test -p tee-launcher` passes (30 tests)


crates/tee-launcher/README.md
```diff
@@ -54,12 +49,7 @@ port_mappings = [
 
 | Field | Required | Description |
 |-------|----------|-------------|
-| `image_tags` | Yes | Docker image tags to search, e.g. `["3.7.0"]` |
-| `image_name` | Yes | Docker image name, e.g. `"nearone/mpc-node"` |
-| `registry` | Yes | Docker registry hostname, e.g. `"registry.hub.docker.com"` |
-| `rpc_request_timeout_secs` | Yes | Per-request timeout for registry API calls (seconds) |
-| `rpc_request_interval_secs` | Yes | Initial retry interval for registry API calls (seconds) |
-| `rpc_max_attempts` | Yes | Maximum registry API retry attempts |
+| `image` | Yes | Full Docker image reference. Include registry prefix for non-Docker Hub registries. Examples: `"nearone/mpc-node"` (Docker Hub), `"ghcr.io/nearone/mpc-node"` |
```
barakeinav1: The registry, image_name, and image_tags fields were replaced with a single image field. These fields existed
  because the old code needed to construct Docker registry API URLs separately                                       
  (https://{registry}/v2/{image_name}/manifests/{tag}). Now that we pull directly with docker pull                   
  image@sha256:<digest>, Docker handles registry resolution natively from the image reference. A single image field  
  (e.g., nearone/mpc-node or ghcr.io/nearone/mpc-node) is all that's needed and matches standard Docker conventions.
  The rpc_* fields (timeout, interval, max_attempts) were also removed as they only controlled the now-deleted
  registry retry logic.

crates/tee-launcher/README.md
```diff
@@ -69,18 +59,16 @@ Arbitrary TOML table passed through to the MPC node. The launcher writes this ve
 
 ## Supported Registries
 
-The launcher uses the [OCI Distribution Specification](https://github.com/opencontainers/distribution-spec) for registry communication. Auth endpoints are discovered automatically via the `WWW-Authenticate` challenge on `/v2/`, so there is no hard-coded auth URL.
+The launcher pulls images using `docker pull <image>@sha256:<digest>`. Any registry that Docker supports works out of the box. Set the `image` field to include the registry prefix:
```
barakeinav1: Note on removing the OCI Distribution Spec client (oci-client):
                                                                                                                     
  The old code used oci-client to interact with registries directly via the OCI Distribution Spec. This gave us: (1) 
  automatic auth challenge discovery via WWW-Authenticate, (2) multi-platform manifest resolution to amd64/linux, and
   (3) config digest verification inside the manifest before pulling.  
                                                                                                                     
  With the new docker pull image@sha256:<manifest_digest> approach, all of this is handled by Docker itself:
  - Auth: Docker uses its credential helpers (~/.docker/config.json), which actually supports private registries —
  the old code only supported anonymous/public images                                                                
  - Multi-platform: Not needed — we pull by exact manifest digest, which already points to a single-platform manifest
   (platform was resolved at build time)                                                                             
  - Digest verification: Docker verifies the pulled content matches the digest — if it doesn't, the pull fails       
                                                                                                              
  Net result: same guarantees, 430 fewer lines, and we gained private registry support.

crates/tee-launcher/README.md
gilcu3: I am not 100% about this, but of course could be certainly the case

barakeinav1:
Addressed review comments in b6f0a846:

- **Unicode curly quotes in `scripts/update-mpc-node.sh`** — Fixed. Replaced with ASCII single quotes.
- **`single-node.sh` missing `MPC_IMAGE`** — Fixed. Added `export MPC_IMAGE="nearone/mpc-node"`.
- **`single-node-readme.md` references `MPC_IMAGE_TAGS` only** — Fixed. Added documentation for `MPC_IMAGE`.
- **`external-services-tests` feature flag** — Already re-added in earlier commit (64c90e55). Feature is present in both `Cargo.toml` and `validation.rs`.
- **Error formatting duplication** — Checked: the error chain is `"Image pull failed: docker pull failed for <ref>: exit code <N>: stderr=..."`. No actual duplication — the outer message says "Image pull failed" and the inner says "docker pull failed for <ref>" with details. These are at different layers and both add context.
- **`registry` field unused** — Already addressed: replaced `image_name`/`registry`/`image_tags` with single `image` field in commit 80201a0f.

crates/tee-launcher/Cargo.toml
```diff
@@ -13,17 +13,11 @@ path = "src/lib.rs"
 name = "tee-launcher"
 path = "src/main.rs"
 
-[features]
-external-services-tests = []
-
 [dependencies]
-backon = { workspace = true }
 clap = { workspace = true }
 dstack-sdk = { workspace = true }
```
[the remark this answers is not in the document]
barakeinav1: Fixed in later commit — `external-services-tests` feature re-added.

crates/tee-launcher/src/lib.rs
```diff
@@ -37,19 +36,21 @@ pub async fn run() -> Result<(), LauncherError> {
...
+    // The approved hashes file now contains manifest digests.
+    // We can pull directly by digest without querying the Docker registry API.
+    let manifest_digest = select_image_hash(
         approved_hashes_on_disk.as_ref(),
         &args.default_image_digest,
         config.launcher_config.mpc_hash_override.as_ref(),
     )?;
 
-    let manifest_digest = validate_image_hash(&config.launcher_config, image_hash.clone()).await?;
+    pull_and_verify(&config.launcher_config.image_name, &manifest_digest)?;
```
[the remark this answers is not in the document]
barakeinav1: Fixed — replaced `image_name`/`registry`/`image_tags` with a single `image` field (e.g., `nearone/mpc-node` or `ghcr.io/nearone/mpc-node`). Docker handles registry resolution from the image reference.

crates/tee-launcher/src/validation.rs
```diff
@@ -1,79 +1,31 @@
...
-            ImageDigestValidationFailed::PulledImageHasMismatchedDigest {
-                pulled_image_id,
-                expected_image_id: image_hash,
-            },
-        );
+    if !pull.status.success() {
+        let stderr = String::from_utf8_lossy(&pull.stderr);
+        return Err(ImageDigestValidationFailed::DockerPullFailed(format!(
+            "docker pull {reference} failed: {stderr}"
+        )));
```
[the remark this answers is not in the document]
barakeinav1: Checked — no actual duplication. The error chain is: `Image pull failed: docker pull failed for <ref>: exit code <N>: stderr=...`. Each layer adds different context.

crates/tee-launcher/README.md
barakeinav1: Confirmed via TEE E2E testing — the single `image` field works correctly with Docker Hub, and the approach supports other registries by including the prefix (e.g., `ghcr.io/nearone/mpc-node`).

scripts/update-mpc-node.sh
```diff
@@ -5,19 +5,27 @@ IMAGE_NAME="nearone/mpc-node-gcp:testnet-release"
...
-# Get currently running container image ID (if exists)
-RUNNING_IMAGE_ID=$(docker inspect --format "{{.Image}}" "$CONTAINER_NAME" 2>/dev/null || echo "")
+# Get the manifest digest of the image used by the running container (if any).
+# RepoDigests contains "image@sha256:<manifest_digest>" after a pull.
+RUNNING_DIGEST=""
+if docker inspect "$CONTAINER_NAME" &>/dev/null; then
+    RUNNING_IMAGE=$(docker inspect --format "{{.Config.Image}}" "$CONTAINER_NAME" 2>/dev/null || echo "")
+    if [ -n "$RUNNING_IMAGE" ]; then
+        RUNNING_DIGEST=$(docker inspect --format ‘{{index .RepoDigests 0}}’ "$RUNNING_IMAGE" 2>/dev/null | grep -oP ‘sha256:\K[0-9a-f]{64}’ || echo "")
+    fi
```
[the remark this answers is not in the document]
barakeinav1: Fixed in b6f0a846 — replaced Unicode curly quotes with ASCII.

scripts/update-mpc-node.sh
```diff
@@ -5,19 +5,27 @@ IMAGE_NAME="nearone/mpc-node-gcp:testnet-release"
...
 
 # Pull latest image
-echo "📥 Pulling latest image: $IMAGE_NAME..."
+echo "Pulling latest image: $IMAGE_NAME..."
 docker pull "$IMAGE_NAME"
 
-# Get the latest image ID
-LATEST_IMAGE_ID=$(docker inspect --format "{{.Id}}" "$IMAGE_NAME")
+# Get the manifest digest of the freshly pulled image
+LATEST_DIGEST=$(docker inspect --format ‘{{index .RepoDigests 0}}’ "$IMAGE_NAME" 2>/dev/null | grep -oP ‘sha256:\K[0-9a-f]{64}’ || echo "")
```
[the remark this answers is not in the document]
barakeinav1: Fixed in b6f0a846.

crates/tee-launcher/src/validation.rs
```diff
@@ -1,79 +1,105 @@
...
+}
+
+#[cfg(all(test, feature = "external-services-tests"))]
+mod integration_tests {
+    use super::*;
+
+    /// Known manifest digest for nearone/testing:83b52da4e2270c688cdd30da04f6b9d3565f25bb
+    /// This is a small test image that should always be available.
+    const DOCKER_HUB_IMAGE: &str = "nearone/testing";
+    const DOCKER_HUB_MANIFEST_DIGEST: &str =
```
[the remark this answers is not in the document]
barakeinav1: The feature was re-added in a later commit. Both `Cargo.toml` and `validation.rs` reference it.

localnet/tee/scripts/rust-launcher/node.conf.localnet.toml.tpl
```diff
@@ -1,10 +1,5 @@
 [launcher_config]
-image_tags = ["${MPC_IMAGE_TAGS}"]
-image_name = "${MPC_IMAGE_NAME}"
-registry = "${MPC_REGISTRY}"
-rpc_request_timeout_secs = 10
-rpc_request_interval_secs = 1
-rpc_max_attempts = 20
+image = "${MPC_IMAGE}"
```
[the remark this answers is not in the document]
barakeinav1: Fixed in b6f0a846 — added `export MPC_IMAGE="nearone/mpc-node"` to `single-node.sh`.

localnet/tee/scripts/rust-launcher/single-node-readme.md
```diff
@@ -27,9 +27,8 @@ export MACHINE_IP=<host-ip>
 
 # MPC node image tag (must support TOML config / start-with-config-file)
 # Verify DEFAULT_IMAGE_DIGEST in deployment/cvm-deployment/launcher_docker_compose.yaml
-# matches this tag:
-#   docker pull nearone/mpc-node:$MPC_IMAGE_TAGS
-#   docker inspect --format='{{.Id}}' nearone/mpc-node:$MPC_IMAGE_TAGS
+# matches this tag's manifest digest. You can look it up on Docker Hub or
+# in the registry's tag listing — no need for `docker inspect`.
 export MPC_IMAGE_TAGS=main-9515e18
 ```
```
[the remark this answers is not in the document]
barakeinav1: Fixed in b6f0a846 — updated readme to document `MPC_IMAGE`.

crates/tee-launcher/src/validation.rs
```diff
@@ -1,79 +1,109 @@
...
-    if docker_inspect_failed {
-        return Err(ImageDigestValidationFailed::DockerInspectFailed(
-            "docker inspect terminated with unsuccessful status".to_string(),
-        ));
+/// Pulls the image by manifest digest with retry logic.
+///
+/// The approved hashes file contains manifest digests, so we can pull directly
+/// without querying the Docker registry API. Docker verifies the content
+/// matches the digest during the pull.
+pub async fn pull_and_verify(
```
barakeinav1: Note on retry logic: The docker pull retry retries on all errors indiscriminately. The old registry API code had explicit error classification (only retry on server errors, not on auth failures or manifest-not-found). For docker pull, distinguishing error types would require parsing stderr strings (e.g., grepping for "manifest unknown", "unauthorized", "connection refused"), which is brittle across Docker versions. So we retry all failures up to pull_max_retries (default 5). Non-transient errors will fail identically each retry but the cap limits waste.

docs/localnet/tee-localnet.md
```diff
@@ -114,13 +114,7 @@ Update Sam/Frodo.toml fields:
...
 ```
 
-The MPC_IMAGE_TAGS should match the MPC node image hash used in the docker-compose file.
-e.g:
-
-```shell
-$Docker inspect nearone/mpc-node:main_3.0.3 | grep "Id"
-"Id": "sha256:abc",
-```
+The MPC_IMAGE_TAGS should match the MPC node image whose manifest digest is set as `DEFAULT_IMAGE_DIGEST` in the docker-compose file. You can look up the manifest digest on Docker Hub or the registry's tag listing.
```
gilcu3: are you sure about this? it appears we are removing `MPC_IMAGE_TAGS` from some places in this PR

localnet/tee/scripts/rust-launcher/how-to-run-localnet-tee-setup-script.md
```diff
@@ -90,7 +90,7 @@ export VMM_RPC=http://127.0.0.1:10000
 # MPC node image (must support TOML config / start-with-config-file)
 export MPC_IMAGE_NAME=nearone/mpc-node
 export MPC_IMAGE_TAGS=main-9515e18
-export MPC_REGISTRY=registry.hub.docker.com
+export MPC_IMAGE=nearone/mpc-node
```
gilcu3, line 91: could be good to add an example on how to pull the image from another registry, given that we still plan to support that. Not sure what the correct place for that would be though

localnet/tee/scripts/rust-launcher/node.conf.localnet.toml.tpl
```diff
@@ -1,10 +1,7 @@
...
+pull_max_retries = 5
+pull_retry_interval_secs = 2
```
gilcu3, line 4: don't we need also `pull_timeout_secs`:

```suggestion
pull_retry_interval_secs = 2
pull_timeout_secs = 10
```

localnet/tee/scripts/rust-launcher/single-node-readme.md
```diff
@@ -27,10 +27,14 @@ export MACHINE_IP=<host-ip>
...
-#   docker pull nearone/mpc-node:$MPC_IMAGE_TAGS
-#   docker inspect --format='{{.Id}}' nearone/mpc-node:$MPC_IMAGE_TAGS
+# matches this tag's manifest digest. You can look it up on Docker Hub or
+# in the registry's tag listing.
 export MPC_IMAGE_TAGS=main-9515e18
+
+# Docker image name (repository). Do not include a tag — the manifest digest
+# determines the actual image version. The tag above is only used by the
+# deploy script for operator reference.
+# export MPC_IMAGE=nearone/mpc-node  # (set automatically by single-node.sh)
```
gilcu3: we should not only use it for operator reference. tags are important for us, that's how we get testnet-release and mainnet-release tags. Please double-check this

localnet/tee/scripts/rust-launcher/test-verify-and-upgrade.sh
```diff
@@ -215,54 +215,49 @@ verify_cluster() {
...
-# Get the Docker image config digest for a given tag from the registry.
-# This is what the rust launcher compares against the approved hashes.
-fetch_image_config_digest() {
-  local image_name="$1" tag="$2" registry="${3:-registry.hub.docker.com}"
+# Get the Docker image manifest digest for a given tag from the registry.
+# The launcher uses manifest digests (not config digests) to identify images.
+# This fetches the digest via the Docker-Content-Digest header, which is the
+# sha256 hash of the manifest blob itself.
+fetch_image_manifest_digest() {
+  local image_name="$1" tag="$2"
```
gilcu3: I know we are keeping this for convenience, but it seems as convenient to pass the actual digest (given that tag name will also not be fixed and digest can be copied)

scripts/update-mpc-node.sh
```diff
@@ -5,19 +5,27 @@ IMAGE_NAME="nearone/mpc-node-gcp:testnet-release"
...
+if docker inspect "$CONTAINER_NAME" &>/dev/null; then
+    RUNNING_IMAGE=$(docker inspect --format "{{.Config.Image}}" "$CONTAINER_NAME" 2>/dev/null || echo "")
+    if [ -n "$RUNNING_IMAGE" ]; then
+        RUNNING_DIGEST=$(docker inspect --format '{{index .RepoDigests 0}}' "$RUNNING_IMAGE" 2>/dev/null | grep -oP 'sha256:\K[0-9a-f]{64}' || echo "")
+    fi
+fi
 
 # Pull latest image
-echo "📥 Pulling latest image: $IMAGE_NAME..."
+echo "Pulling latest image: $IMAGE_NAME..."
```
gilcu3, line 19: I agree with AI (or human) changing this :)

crates/tee-launcher/src/validation.rs
```diff
@@ -1,79 +1,109 @@
...
+
+        Ok(())
+    };
+
+    let max_delay_secs = 60;
+    let backoff = ExponentialBuilder::default()
+        .with_min_delay(Duration::from_secs(retry_interval_secs.min(max_delay_secs)))
+        .with_factor(1.5)
+        .with_max_delay(Duration::from_secs(max_delay_secs))
+        .with_max_times(max_retries);
```
gilcu3: seems weird to have max_retries and retry_interval_secs as function parameters, and then max_delay_secs defined here. IMO all should be parameters

crates/tee-launcher/src/validation.rs
```diff
@@ -1,79 +1,109 @@
...
+                detail: e.to_string(),
+            })?;
+
+        if !pull.status.success() {
+            let stderr = String::from_utf8_lossy(&pull.stderr).to_string();
+            let stdout = String::from_utf8_lossy(&pull.stdout).to_string();
+            return Err(ImageDigestValidationFailed::DockerPullFailed {
+                reference: reference.clone(),
+                detail: format!(
+                    "exit code {}: stderr={stderr}, stdout={stdout}",
```
gilcu3:
```suggestion
                    "exit code {}:\nstderr:\n{stderr}\nstdout:\n{stdout}",
```

crates/tee-launcher/src/validation.rs
```diff
@@ -1,79 +1,109 @@
...
+///
+/// The approved hashes file contains manifest digests, so we can pull directly
+/// without querying the Docker registry API. Docker verifies the content
+/// matches the digest during the pull.
+pub async fn pull_and_verify(
+    image_name: &str,
+    manifest_digest: &DockerSha256Digest,
+    max_retries: usize,
+    retry_interval_secs: u64,
+) -> Result<(), ImageDigestValidationFailed> {
```
gilcu3: I am surprised we did not have this function before. How were we pulling the image?

crates/tee-launcher/src/types.rs
```diff
@@ -50,24 +49,32 @@ pub struct Config {
...
     pub port_mappings: Vec<PortMapping>,
 }
 
+fn default_pull_max_retries() -> usize {
+    5
+}
+
+fn default_pull_retry_interval_secs() -> u64 {
+    2
+}
```
gilcu3, line 85: these should be constants IMO

crates/tee-launcher/src/types.rs
```diff
@@ -50,24 +49,32 @@ pub struct Config {
...
+    /// Include registry prefix for non-Docker Hub registries.
+    /// Examples: `"nearone/mpc-node"`, `"ghcr.io/nearone/mpc-node"`.
+    pub image: String,
+    /// Maximum number of retries for `docker pull`. Defaults to 5.
+    #[serde(default = "default_pull_max_retries")]
+    pub pull_max_retries: usize,
+    /// Initial delay between pull retries in seconds (exponential backoff base).
+    /// Capped at 60 seconds. Defaults to 2.
+    #[serde(default = "default_pull_retry_interval_secs")]
+    pub pull_retry_interval_secs: u64,
```
gilcu3: we are missing the equivalent of `rpc_request_timeout_secs` here

crates/tee-launcher/src/lib.rs
```diff
@@ -37,27 +36,35 @@ pub async fn run() -> Result<(), LauncherError> {
...
     )?;
 
-    let manifest_digest = validate_image_hash(&config.launcher_config, image_hash.clone()).await?;
+    pull_and_verify(
+        &config.launcher_config.image,
+        &manifest_digest,
+        config.launcher_config.pull_max_retries,
+        config.launcher_config.pull_retry_interval_secs,
+    )
+    .await?;
```
gilcu3: what exactly are we verifying here? I think the digest verification happens already in `launch_mpc_container`, docker compose pulls the image and checks it, there is nothing extra we need to do here.

crates/tee-launcher/src/main.rs
```diff
@@ -1,13 +1,5 @@
 #[tokio::main]
 async fn main() {
-    // Install the default rustls crypto provider before any TLS usage.
-    // Required because rustls is configured with default-features=false,
-    // and indirect consumers like hyper-rustls (via reqwest) call
-    // ClientConfig::builder() without an explicit provider.
-    rustls::crypto::ring::default_provider()
-        .install_default()
-        .expect("Failed to install default rustls CryptoProvider");
```
gilcu3, line 9: I am not sure we need this atm, but would rather leave it there just in case, as not having it might cause a runtime crash as we recently saw

crates/tee-launcher/src/error.rs
```diff
@@ -77,29 +50,15 @@ pub enum LauncherError {
...
-    #[error(
-        "pulled image has mismatching image ID. pulled: {pulled_image_id}, expected: {expected_image_id}"
-    )]
-    PulledImageHasMismatchedDigest {
-        expected_image_id: DockerSha256Digest,
-        pulled_image_id: DockerSha256Digest,
-    },
+    #[error("docker pull failed for {reference}: {detail}")]
+    DockerPullFailed { reference: String, detail: String },
 }
```
gilcu3, line 64: this error will disappear, as we don't need to pull the image explicitly anymore

crates/tee-launcher/src/config.rs
```diff
@@ -33,15 +33,22 @@ fn insert_reserved(
...
 
-/// Validate that `image_name` contains only safe characters for Docker image names.
-/// Rejects values that could inject YAML syntax (newlines, colons in unexpected places, etc.)
-/// when substituted into the compose template.
+/// Validate that `image_name` contains only safe characters for Docker image references.
+/// Allows registry/name:tag format. Rejects values that could inject YAML syntax
+/// (newlines, spaces, etc.) when substituted into the compose template.
 pub fn validate_image_name(image_name: &str) -> Result<(), LauncherError> {
-    // Docker image names: [a-zA-Z0-9][a-zA-Z0-9._/-]*
+    // Docker image references: [a-zA-Z0-9][a-zA-Z0-9._/:-]*
```
gilcu3: you mean docker manifest digest?

crates/tee-launcher/src/config.rs
```diff
@@ -33,15 +33,22 @@ fn insert_reserved(
...
-            .all(|b| b.is_ascii_alphanumeric() || b == b'/' || b == b'-' || b == b'.' || b == b'_');
+        && image_name.bytes().all(|b| {
+            b.is_ascii_alphanumeric()
+                || b == b'/'
+                || b == b'-'
+                || b == b'.'
+                || b == b'_'
+                || b == b':'
+        });
     if !is_valid {
```
gilcu3: if we want to check format including registry, we need a much better regex for it

gilcu3 (changes requested):
We should remove `pull_and_verify` and add back the rustls default provider.

Left a bunch of questions/suggestions

scripts/update-mpc-node.sh
barakeinav1: some previous AI probably added this (chatGPT used to love adding icons), and new AI (claude) removed it  :-)

docs/localnet/tee-localnet.md
barakeinav1: `MPC_IMAGE_TAGS` is not removed — it's still required by the deploy scripts (`deploy-tee-localnet.sh` line 45, `set-localnet-env.sh` line 16) and used by the upgrade test to look up manifest digests and log which version is deployed.

What was removed is `image_tags` from the TOML launcher config — the launcher no longer needs tags because it pulls by manifest digest directly. The TOML template changed from `image_tags = ["${MPC_IMAGE_TAGS}"]` to `image = "${MPC_IMAGE}"`, which may look like the tags were dropped, but the env var (script env)  is still  used by the scripts.

localnet/tee/scripts/rust-launcher/how-to-run-localnet-tee-setup-script.md
barakeinav1: Good idea. The right place long-term would be the external operator guide, but we don't have an external repo for that yet. For now I'll add an example here in `how-to-run-localnet-tee-setup-script.md` showing how to set `MPC_IMAGE` for a non-Docker Hub registry (e.g., `ghcr.io/nearone/mpc-node`).

crates/tee-launcher/src/validation.rs
barakeinav1: Fixed in 4f33c71f — `pull_max_delay_secs` is now a configurable `LauncherConfig` field (default 60).

crates/tee-launcher/src/validation.rs
barakeinav1: Fixed in 4f33c71f — using newlines between stderr/stdout.

localnet/tee/scripts/rust-launcher/how-to-run-localnet-tee-setup-script.md
barakeinav1: Added in 4f33c71f — example for `ghcr.io/nearone/mpc-node` in how-to-run doc.

crates/tee-launcher/src/validation.rs
barakeinav1: Before, `registry.rs` queried the Docker registry API to resolve config digest → manifest digest, then `validation.rs` did `docker pull image@manifest_digest` followed by `docker inspect` to verify the config digest. Now we skip the registry lookup and pull directly by manifest digest.

scripts/update-mpc-node.sh
barakeinav1: Thanks!

barakeinav1:
Regarding adding back the rustls default provider:

The dstack SDK has two communication modes: HTTP (reqwest with rustls-tls) and Unix socket. Our launcher uses `DstackClient::new(Some(DSTACK_UNIX_SOCKET))` which sets it to Unix socket mode — it communicates via `/var/run/dstack.sock`, not HTTPS. So reqwest's TLS is never used and the rustls crypto provider is not needed.

Verified this works in TEE E2E testing — deploy, attestation, upgrade, and hash override all passed without the provider installed.

The rustls provider was only needed for the old `oci-client` registry API calls (which used reqwest over HTTPS to Docker Hub).

docs/localnet/tee-localnet.md
gilcu3: this seems confusing. Could we just use the same mechanism (manifest digest) in the test scripts as well?

barakeinav1:
Regarding removing `pull_and_verify` — could you clarify what you have in mind?

Currently it wraps `docker pull image@sha256:<digest>` with retry logic (exponential backoff via backon). If we remove it, where should the pull happen? Options I can think of:

1. Inline the `docker pull` in `lib.rs::run()` without retry
2. Let Docker Compose handle the pull implicitly (it pulls if the image isn't local)
3. Something else?

The retry was added because `docker pull` can fail on transient network issues. But if you think it's unnecessary I'm happy to simplify.

localnet/tee/scripts/rust-launcher/node.conf.localnet.toml.tpl
barakeinav1: We don't have a `pull_timeout_secs` field. The old `rpc_request_timeout_secs` was a per-HTTP-request timeout for the registry API calls (oci-client/reqwest). Since we now use `docker pull`, Docker manages its own connection and transfer timeouts internally — there's no way to pass a timeout to `docker pull` via CLI.

The total time is bounded by `pull_max_retries` * `pull_max_delay_secs`, which caps retry duration. See [`validation.rs`](https://github.com/near/mpc/blob/barak/simplify-image-validation/crates/tee-launcher/src/validation.rs#L14-L19) and [`types.rs`](https://github.com/near/mpc/blob/barak/simplify-image-validation/crates/tee-launcher/src/types.rs#L50-L68) for the config fields.

gilcu3:
> 2\. Let Docker Compose handle the pull implicitly (it pulls if the image isn't local)

this is exactly what we were doing before, what I am saying is that we do not need to change that behavior by doing an extra pull before docker compose runs

> The rustls provider was only needed for the old oci-client registry API calls (which used reqwest over HTTPS to Docker Hub).

true, I just prefer keeping it so that we don't shoot ourselves in the foot in the future. If the crate does not depend on rustls at all anymore, then we can remove it. If it still does, probably we should keep it.

barakeinav1:
> this is exactly what we were doing before, what I am saying is that we do not need to change that behavior by doing an extra pull before docker compose runs

Regarding `pull_and_verify` — you're right that docker compose already pulls the image by digest. Renamed to `pull_with_retry` in 28757d4c to clarify the purpose: it pre-pulls the image with exponential backoff retry so transient network failures are handled before compose starts. Docker compose doesn't retry on pull failures.

The flow:
1. `pull_with_retry` → `docker pull image@sha256:<digest>` with retries → image cached locally
2. `launch_mpc_container` → `docker compose up -d` → uses cached image, no re-pull needed

crates/tee-launcher/src/lib.rs
barakeinav1: changed to pull_and_retry (see https://github.com/near/mpc/pull/2785#issuecomment-4234645503)

localnet/tee/scripts/rust-launcher/single-node-readme.md
barakeinav1: You're right — fixed in ab3d5baa. Removed the "do not include a tag" guidance. Tags like `testnet-release` and `mainnet-release` are important for identifying versions. Updated the doc comment, README, and single-node-readme to say tags can be included and are useful, while the manifest digest determines the actual image pulled.

localnet/tee/scripts/rust-launcher/test-verify-and-upgrade.sh
barakeinav1: Valid point — passing the digest directly would skip the tag-to-digest lookup. Keeping the tag-based flow for now since it matches the deploy script workflow (scripts use tags like `MPC_IMAGE_TAGS` throughout), but agree it could be simplified in the future.

crates/tee-launcher/src/main.rs
barakeinav1: I will retest this again on dstack CVM before I merge, if it passes, do you still want to keep this?

crates/tee-launcher/src/types.rs
barakeinav1: Already replied on the TOML template comment — `docker pull` doesn't accept a timeout flag via CLI. Docker manages its own connection/transfer timeouts internally. The old `rpc_request_timeout_secs` was a per-HTTP-request timeout for the reqwest/oci-client registry calls, which are now gone. Total retry duration is bounded by `pull_max_retries * pull_max_delay_secs`.

crates/tee-launcher/src/types.rs
barakeinav1: These are serde default value functions — `#[serde(default = "...")]` requires a function path in Rust, it doesn't accept a const directly. This is the standard/idiomatic pattern for serde defaults. If there's a cleaner way to do this I'm happy to change it (not a Rust expert), but from what I've seen this is how it's typically done.

crates/tee-launcher/src/config.rs
barakeinav1: No — this function validates the `image` config field (the Docker image reference like `nearone/mpc-node` or `ghcr.io/nearone/mpc-node:tag`), not a digest. The comment says "Docker image names" which is what the `image` field contains. The function name `validate_image_name` matches this. Digests are validated separately by the `DockerSha256Digest` type parser.

crates/tee-launcher/src/config.rs
barakeinav1: Agreed it's not a full Docker reference parser. It's intentionally a simple allowlist to prevent YAML/compose template injection — the `image` field is interpolated into `"{{IMAGE_NAME}}@{{MANIFEST_DIGEST}}"` in the compose template. The dangerous chars (`"`, `\`, newlines, spaces) are all blocked.

No security concern with the current set. No valid Docker registries are blocked either — Docker references use `[a-zA-Z0-9._/:-]` which we allow. A full reference parser (RFC grammar) would be more correct but adds complexity for no practical benefit here.

docs/localnet/tee-localnet.md
barakeinav1: Updated in 0b26939b — added explicit steps to get the manifest digest for a tag and verify it matches `DEFAULT_IMAGE_DIGEST`. Should be clearer now.

barakeinav1: Updated in 2c857316 — replaced `MPC_IMAGE_TAGS` with `MPC_MANIFEST_DIGEST` in all rust-launcher scripts and docs. Scripts now take digests directly.

localnet/tee/scripts/rust-launcher/single-node-readme.md
barakeinav1: Updated in 2c857316 — `single-node-readme.md` now uses `MPC_MANIFEST_DIGEST` instead of `MPC_IMAGE_TAGS`.

localnet/tee/scripts/rust-launcher/test-verify-and-upgrade.sh
barakeinav1: Done in 2c857316 — upgrade test now takes a manifest digest directly (`upgrade sha256:abc...`). Removed the `fetch_image_manifest_digest` function (~40 lines of registry API calls).

docs/localnet/tee-localnet.md
barakeinav1: Done in 2c857316 — all rust-launcher scripts now use `MPC_MANIFEST_DIGEST` directly instead of tags.

barakeinav1: actually, let me remove the tag from the doc use digests directly . (user should be able to convert easily by themselves if needed)

crates/tee-launcher/README.md
barakeinav1: Confirmed via TEE E2E testing — the single `image` field works correctly. Updated terminology to 'Docker image name (repository)' for clarity.

crates/tee-launcher/src/error.rs
barakeinav1: Discussed in the PR thread — renamed to `pull_with_retry` in 28757d4c. The function adds retry with exponential backoff for transient network failures, which docker compose doesn't provide. See the [PR comment](https://github.com/near/mpc/pull/2785#issuecomment-4234892453) for details. Waiting for your response on whether to keep or remove.

crates/tee-launcher/src/main.rs
barakeinav1: tested, there is not issue due to the removal of rustls

barakeinav1:
> > 2. Let Docker Compose handle the pull implicitly (it pulls if the image isn't local)
> 
> this is exactly what we were doing before, what I am saying is that we do not need to change that behavior by doing an extra pull before docker compose runs
> 
> > The rustls provider was only needed for the old oci-client registry API calls (which used reqwest over HTTPS to Docker Hub).
> 
> true, I just prefer keeping it so that we don't shoot ourselves in the foot in the future. If the crate does not depend on rustls at all anymore, then we can remove it. If it still does, probably we should keep it.

tested on dstsack CVM there is not issue due to the removal of `rustls`

crates/tee-launcher/src/config.rs
gilcu3: yeah I did not mean using a full grammar. But something like a regex for matching explicitly what the the domain, the optional port, then the image name and finally tag would be more appropriate

gilcu3:
> tested on dstsack CVM there is not issue due to the removal of `rustls`

that's good, but to be on the safe side what we need to check is if the current launcher crate still depends on rustls or not.

crates/tee-launcher/src/types.rs
```diff
@@ -50,24 +49,40 @@ pub struct Config {
...
-    pub rpc_request_interval_secs: u64,
-    /// Maximum number of registry API retry attempts. Set via `rpc_max_attempts`.
-    pub rpc_max_attempts: u32,
+    /// Docker image name (repository). A tag can be included to identify the
+    /// configured version (e.g., `"nearone/mpc-node:testnet-release"`), but the
+    /// manifest digest from the approved hashes file determines the actual image
+    /// pulled. Include registry prefix for non-Docker Hub registries.
+    /// Examples: `"nearone/mpc-node"`, `"nearone/mpc-node:3.8.1"`,
+    /// `"ghcr.io/nearone/mpc-node"`.
+    pub image: String,
```
gilcu3: nit: should use more descriptive name. Perhaps `image_reference`? I could not come up with a good name, but image is too broad

crates/tee-launcher/src/validation.rs
```diff
@@ -1,79 +1,109 @@
...
+            );
+        })
+        .await?;
+
+    tracing::info!(%reference, "image pulled successfully");
+    Ok(())
+}
+
+#[cfg(all(test, feature = "external-services-tests"))]
+mod integration_tests {
```
gilcu3: nit: we are usually using mod tests for everything. If that's not the case and we use integration_tests elsewhere, please disregard this nit

docs/running-an-mpc-node-in-tdx-external-guide.md
```diff
@@ -1208,35 +1202,33 @@ git checkout 828f816be36aed6f0d2438e0131b3e9d7d0931ad
   running.
 
 ```bash
-$ ./deployment/build-images.sh
+$ ./deployment/build-images.sh --node --push
```
gilcu3: push will certainly fail here for operators. But I understand why it was added. Make a comment about it, or a TODO so that only the actual push requires --push, but the skopeo part is always done so that manifests are always printed

localnet/tee/scripts/rust-launcher/deploy-tee-localnet.sh
```diff
@@ -765,8 +765,7 @@ render_node_files_range() {
     export MPC_ENV
 
     export MPC_IMAGE_NAME="nearone/mpc-node"
```
gilcu3: this one is not needed anymore, right? else we have a duplicate with MPC_IMAGE below

localnet/tee/scripts/rust-launcher/how-to-run-localnet-tee-setup-script.md
```diff
@@ -87,10 +87,15 @@ export MACHINE_IP=<EXTERNAL_SERVER_IP>
 export BASE_PATH=/path/to/meta-dstack/dstack  # must contain vmm/src/vmm-cli.py
 export VMM_RPC=http://127.0.0.1:10000
 
-# MPC node image (must support TOML config / start-with-config-file)
+# MPC node image
 export MPC_IMAGE_NAME=nearone/mpc-node
-export MPC_IMAGE_TAGS=main-9515e18
-export MPC_REGISTRY=registry.hub.docker.com
+export MPC_IMAGE=nearone/mpc-node
```
gilcu3: duplicate var?

localnet/tee/scripts/rust-launcher/node.conf.localnet.toml.tpl
```diff
@@ -1,10 +1,8 @@
...
+pull_max_delay_secs = 60
```
gilcu3, line 5: how is that before we supported multiple tags, and now only a single image? isn't that a regression?

gilcu3 (dismissed):
Thank you!

Worth checking before merging:
- Launcher supported multiple tags before, now a single image manifest. Worth checking if we are fine with that

localnet/tee/scripts/rust-launcher/node.conf.localnet.toml.tpl
barakeinav1: Not a regression — the old image_tags was a search mechanism for the config-digest-based flow. The launcher had
  a config digest (from the contract) but needed a manifest digest to pull. So it iterated through the tags,      
  querying the registry for each one, checking if the config digest matched.                                      
                                                                                                                  
 With manifest digests, the launcher already has the exact digest to pull (from the approved hashes file written
  by the MPC node, or DEFAULT_IMAGE_DIGEST on first boot). No search needed — docker pull image@sha256:<digest>  works directly. Multiple tags served no purpose anymore.

gilcu3: makes sense, thanks for explaining

localnet/tee/scripts/rust-launcher/deploy-tee-localnet.sh
barakeinav1: Removed in 24df06c1 — consolidated to `MPC_IMAGE` everywhere.

localnet/tee/scripts/rust-launcher/how-to-run-localnet-tee-setup-script.md
barakeinav1: Fixed in 24df06c1 — removed `MPC_IMAGE_NAME`, only `MPC_IMAGE` remains.

docs/running-an-mpc-node-in-tdx-external-guide.md
barakeinav1: Good point. Removed `--push` from the operator build example in 74c5d524. Created #2843 to track computing the manifest digest locally without requiring `--push`.

crates/tee-launcher/src/validation.rs
barakeinav1: Fixed in 4940074c — renamed to `mod tests` to match codebase convention (e.g., `contract-history` uses the same pattern with `external-services-tests`).

crates/tee-launcher/src/types.rs
barakeinav1: Renamed to `image_reference` in 02a8b0d1. Updated all TOML configs, templates, and docs.

crates/tee-launcher/src/config.rs
barakeinav1: Improved validation in 02a8b0d1 — now checks: first char must be alphanumeric, rejects trailing `:`/`/`/`.`, rejects `//`, `::`, `:.` sequences. Not a full Docker reference grammar but catches structurally invalid references beyond just safe characters.

crates/tee-launcher/src/config.rs
gilcu3, line 418: sorry for the nit: feel free to just leave it if the 2nd reviewer does not ask for changes.

We kind of avoid these comments, as they seem only useful for AI. test names should express the test intention, if that's too complicated then a comment above the test would do it

gilcu3 (approved):

crates/tee-launcher/src/config.rs
netrome: And.. if it's for code organization it's better to create separate modules generally.

crates/tee-launcher/src/main.rs
```diff
@@ -1,9 +1,9 @@
...
 async fn main() {
-    // Install the default rustls crypto provider before any TLS usage.
-    // Required because rustls is configured with default-features=false,
-    // and indirect consumers like hyper-rustls (via reqwest) call
-    // ClientConfig::builder() without an explicit provider.
+    // Install the default rustls crypto provider. Currently not needed
+    // (dstack SDK communicates via unix socket, not TLS), but kept as a
+    // safety net: if a dependency adds a TLS code path in the future,
+    // missing this call would cause a runtime panic.
     rustls::crypto::ring::default_provider()
```
netrome, line 7: Thanks for keeping this around. Definitely less than ideal that this panic exists.

crates/tee-launcher/src/config.rs
```diff
@@ -33,17 +33,36 @@ fn insert_reserved(
...
+/// Checks structural validity beyond just safe characters:
+/// - Must not be empty
+/// - Must start with an alphanumeric character
+/// - Must not end with `:`, `/`, or `.`
+/// - Must not contain `//`, `::`, or `:.`
+/// - Only allows `[a-zA-Z0-9._/:-]`
+///
+/// This prevents YAML injection when the value is interpolated into the
+/// compose template, while also catching obviously malformed references.
+pub fn validate_image_reference(image_ref: &str) -> Result<(), LauncherError> {
```
netrome, line 47: While orthogonal to this PR, it feels very brittle to have to rely on this manual verification. Why can't we just use serde to construct the yaml template and not have to worry about injections in the first place?

netrome (approved):
The PR looks alright but I'm not a fan of needing manual validation logic to prevent YAML injections.

crates/tee-launcher/src/config.rs
netrome: Created https://github.com/near/mpc/pull/2847 to illustrate what I mean

barakeinav1: let open a dedicate issue for this.

barakeinav1:
Opened #2850 to track switching compose template rendering from string interpolation to serde (per @netrome's suggestion). Draft PR #2847 illustrates the approach.

crates/tee-launcher/src/config.rs
barakeinav1: post merge question for next time:
could you clarify what you had in mind? looking at the our code conventions we almost alway use inline `mod tests` in the same file.
                                                                                
                                                                                                                
are you suggesting that when the test block gets large (we have 30+ tests in config.rs), it should be split into a separate module file? Or do you prefer separate test modules as a general practice?

crates/tee-launcher/src/config.rs
barakeinav1: [2850](https://github.com/near/mpc/issues/2850)

PR #2820, opened 2026-04-10
test: add integration tests for foreign chain requests

### LINKED ISSUE #2821: test: add sandbox integration tests for foreign chain verification requests
### Background

for the foreign chain verification feature, we don't have any integration tests that tests the contract works as expected when receiving requests.

### User Story

_No response_

### Acceptance Criteria

we have sandbox integration tests for foreign tx requests

_No response_

DSharifi:
closes #2821


DSharifi:
@claude review

DSharifi:
@claude review

crates/contract/tests/sandbox/common.rs
```diff
@@ -523,6 +536,140 @@ fn hash(code: &[u8]) -> [u8; 32] {
...
+    }
+}
+
+/// Poll the contract until a pending foreign-tx request appears (or panic after timeout).
+pub async fn await_pending_foreign_tx_request_observed_on_contract(
+    contract: &Contract,
+    request: &near_mpc_contract_interface::types::VerifyForeignTransactionRequest,
+) {
+    let args = json!({ "request": request });
+    for _ in 0..30 {
```
anodar: optional nit: I think it's a bit nicer to instead have timeout, sleep time constants, and loop timeout/sleep times instead.  That way it's clearer what's condition for timing out.

crates/contract/tests/sandbox/common.rs
```diff
@@ -523,6 +536,140 @@ fn hash(code: &[u8]) -> [u8; 32] {
...
+        }
+        tokio::time::sleep(std::time::Duration::from_millis(250)).await;
+    }
+    panic!("Timed out waiting for pending foreign-tx request on-chain");
+}
+
+/// Sign a foreign-tx payload hash with the root secret key and return the
+/// payload and contract-level response DTO.
+pub fn sign_foreign_tx_response(
+    request: &near_mpc_contract_interface::types::ForeignChainRpcRequest,
```
anodar, line 654: What's the convention we follow when to use full paths rather just importing in the beginning of the file ? There's lot of boilerplate here with `near_mpc_contract_interface::types::` prefixes on all argument / return types.

anodar (dismissed):

crates/contract/tests/sandbox/common.rs
DSharifi: We have no conventions. I personally don't care too much in test code helpers.

DSharifi: updated to use `dtos::`

crates/contract/tests/sandbox/common.rs
DSharifi: good point. fixed.

anodar (dismissed):

crates/contract/tests/sandbox/common.rs
```diff
@@ -523,6 +538,148 @@ fn hash(code: &[u8]) -> [u8; 32] {
...
+    pub worker: Worker<Sandbox>,
+    pub contract: Contract,
+    pub accounts: Vec<Account>,
+    pub domain_id: DomainId,
+    pub secret_key: threshold_signatures::ecdsa::KeygenOutput,
+}
+
+/// Initialise a contract in Running state with a single ForeignTx Secp256k1
+/// domain.
+pub async fn setup_foreign_tx_env() -> ForeignTxEnv {
```
gilcu3, line 551: if we have a single module for foreign chain requests, what's the point of defining these structs and functions in the common file?

crates/contract/tests/sandbox/common.rs
```diff
@@ -543,3 +700,61 @@ pub async fn generate_participant_and_submit_attestation(
...
+    )]
+}
+
+pub fn bnb_evm_request() -> ForeignChainRpcRequest {
+    ForeignChainRpcRequest::Bnb(EvmRpcRequest {
+        tx_id: EvmTxId([0xbb; 32]),
+        extractors: vec![EvmExtractor::BlockHash],
+        finality: EvmFinality::Finalized,
+    })
+}
```
gilcu3, line 760: none of these functions should be in this file IMO, they are strictly about foreign chain requests

crates/contract/tests/sandbox/common.rs
```diff
@@ -523,6 +538,148 @@ fn hash(code: &[u8]) -> [u8; 32] {
     hasher.finalize().into()
 }
 
+pub struct ForeignTxEnv {
+    pub worker: Worker<Sandbox>,
+    pub contract: Contract,
+    pub accounts: Vec<Account>,
+    pub domain_id: DomainId,
+    pub secret_key: threshold_signatures::ecdsa::KeygenOutput,
+}
```
gilcu3, line 547: this is a slight specialization of `SandboxTestSetup`. We should reuse that as we had been doing for all other sandbox tests, and also reuse `init_env` instead of creating `setup_foreign_tx_env`. Both are mostly duplicate code

gilcu3: If we want to gain usability, we could use a thin wrapper over `SandboxTestSetup` and `init_env` for foreign chain transactions.

gilcu3 (changes requested):
Thank you!

In the current PR we are adding duplicate code, and not following the current code organization. We should sort these out before merging.

crates/contract/tests/sandbox/common.rs
DSharifi: Since they are helpers to construct the test setup, keeping them in `common.rs` maybe it should be renamed to `helpers.rs`, will allow us to reuse the logic in the future and keep mostly test logic in the new module.

For example for contract upgrades we should also add tests that these requests are persisted and can be responded to afterwards, so these helpers could be used there.

crates/contract/tests/sandbox/common.rs
DSharifi: I think they are fine to have here, since they are just test helpers. Like mentioned above we should use these for contract upgrade tests as well.

crates/contract/tests/sandbox/common.rs
DSharifi: Discussed on slack. Will keep as is, and add a builder in a follow up https://github.com/near/mpc/pull/2844

gilcu3: we decided to remove the code duplication in a follow-up https://github.com/near/mpc/pull/2844

gilcu3 (dismissed):
Thank you!

gilcu3 (approved):

anodar (approved):

PR #2822, opened 2026-04-10
feat: bnb support in foreign chain inspector

### LINKED ISSUE #2789: BNB support in foreign chain inspector
### Background

sub issue of #2788 
foreign chain inspector crate supports BNB chain

### User Story

_No response_

### Acceptance Criteria

foreign chain inspector crate supports BNB chain

_No response_

DSharifi:
closes #2789

Most of the diff is just moving code around. The moved code comes from the fact that most chains that follow the EVM spec will have identical insepctor implementations and integration tests. For that reason this PR makes the inspector generic over EVM chains, and also changes the integration tests into a macro! which allows creating new inspectors and tests with just a couple of lines for future chains.


crates/foreign-chain-inspector/tests/evm_inspector.rs
```diff
@@ -0,0 +1,416 @@
...
+        transaction_hash: H256([3; 32]),
+        block_hash: H256([4; 32]),
+        block_number: U64([5]),
+        address: H160([6; 20]),
+        data: "test_log".to_string(),
+        topics: vec![H256([7; 32]), H256([8; 32])],
+    }
+}
+
+macro_rules! evm_inspector_tests {
```
gilcu3, line 69: nice!

crates/foreign-chain-inspector/tests/evm_inspector.rs
```diff
@@ -0,0 +1,416 @@
...
+                let mock_client = mock_evm_client(block_response, tx_response);
+                let inspector = Inspector::new(mock_client);
+
+                let extracted_values = inspector
+                    .extract(tx_id, finality, vec![extractor])
+                    .await
+                    .unwrap();
+
+                assert_eq!(vec![expected], extracted_values);
+            }
```
gilcu3, line 113: nit: we lost all the given when then comments in these tests, and they existed before

gilcu3 (dismissed):
LGTM

crates/foreign-chain-inspector/tests/evm_inspector.rs
DSharifi: fixed

gilcu3 (approved):

barakeinav1:
@claude review

crates/foreign-chain-inspector/src/evm/inspector.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum EvmExtractedValue<Chain: EvmChain> {
+    BlockHash(Chain::BlockHash),
+    Log(Log),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum EvmExtractor {
```
barakeinav1, line 116: Question: Was it intentional to leave `EvmExtractor` non-generic over `Chain`, while `EvmExtractedValue<Chain>` is parameterized? This means extractors are interchangeable across chains — e.g. nothing prevents passing an extractor constructed for Ethereum into a `BnbInspector`.

barakeinav1 (approved):

crates/foreign-chain-inspector/src/evm/inspector.rs
DSharifi: Very good question, and good to spot. Yes, since we've already defined the block hashes per chain, but I do question now if this should also be changed to remove the generics. The way we call these inspectors, we don't benefit much from the type safety anyway.

Let's keep it as is, and I'll get back to this once I add more EVM based chains to see how it looks if we remove it.

PR #2823, opened 2026-04-10
feat: node handles BNB foreign chain transactions

### LINKED ISSUE #2791: BNB node performs foreign chain signature requests
### Background

The node is able to verify and respond to foreign transaction requests for BNB chain

### User Story

_No response_

### Acceptance Criteria

The node is able to verify and respond to foreign transaction requests for BNB chain

_No response_

DSharifi:
closes #2791

stacked on top of #2822


DSharifi:
@claude review

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -241,6 +242,51 @@ where
 
                 values.into_iter().map(Into::into).collect()
             }
+            dtos::ForeignChainRpcRequest::Bnb(request) => {
```
barakeinav1, line 248: nit: This ~45-line block is a near-identical copy of the Abstract handler above (lines 199-243). The only differences are the config field name, inspector type, and extractor type. The same will be true for every future EVM chain (Polygon, Arbitrum, Avalanche, etc.).

Worth considering a shared helper like `execute_evm_request<I, E>(config, request_id, my_participant_index, request)` that encapsulates the common pattern (select provider → build client → convert types → extract with timeout). PR #2822 already did this at the inspector level with `EvmInspector<Client, Chain>` — this would be the equivalent at the handler level.

Not blocking since the pattern is already established with Abstract.

barakeinav1 (commented):
added some Non-blocking observations,
not logical bugs found, backward compatibility is fine, test coverage is sufficient.

crates/node-config/src/foreign_chains/bnb.rs
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    fn validate(&self, chain_label: &str, provider_name: &str) -> anyhow::Result<()> {
+        auth::validate_auth_config(&self.auth, &self.rpc_url, chain_label, provider_name)
+    }
+}
+
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[serde(rename_all = "kebab-case")]
+pub enum BnbApiVariant {
```
barakeinav1, line 47: Question: Are all these API variants actually relevant for BNB? Alchemy and Infura don't support BNB Chain. This looks copy-pasted from `AbstractApiVariant` without filtering for BNB-specific providers. Unused variants won't cause runtime issues, but they're misleading for operators configuring the node — someone might try to configure `api_variant: "infura"` and wonder why it doesn't work.

pytest/tests/test_foreign_transaction_validation.py
```diff
@@ -522,6 +540,84 @@ def verify_response(res):
...
+    tx = cluster.request_node.sign_tx(
+        cluster.mpc_contract_account(),
+        "verify_foreign_transaction",
+        args,
+        gas=GAS_FOR_VERIFY_FOREIGN_TX_CALL * TGAS,
+        deposit=VERIFY_FOREIGN_TX_DEPOSIT,
+    )
+
+    # Send, await, and verify response
+    def verify_response(res):
```
barakeinav1: nit: `verify_response` is now duplicated 4 times in this file (Bitcoin, Abstract, BNB, Starknet) with nearly identical logic. Consider extracting to a shared helper at the top of the file. Each test could pass the expected signature scheme if needed.

barakeinav1 (dismissed):

pbeza (dismissed):
I didn’t follow too closely how foreign chain txs were implemented before, so I’m by no means an expert here, but overall LGTM. ✅

One thing that stands out is quite a bit of copy-paste across different chains — e.g. what @barakeinav1 [pointed out](https://github.com/near/mpc/pull/2823#pullrequestreview-4105051675), or here:

https://github.com/near/mpc/blob/e67277e06281218485e5129082f66f4177b88550/crates/node-config/src/foreign_chains.rs#L40-L111

Might be worth refactoring to reduce that (if feasible) — for example, by putting the chains into some kind of container and iterating over them — but it’s probably out of scope for this PR.

_Edit: I think Copilot’s comment might be valid here: https://github.com/near/mpc/pull/2823#discussion_r3078922807. If so, this would be a blocker unless it was intentional._

crates/node-config/src/foreign_chains/bnb.rs
DSharifi: It's a good point. The API variant they choose is not something they will arbitrarily set, as it requires the operator sign up with the RPC provider for the specific chain. So you can't really just setup "infura" with credentials, and already have an API key set up for the specific chain.

But it's a good point that these RPC providers are currently just copy pasted across chains, and should be generalized.

Created a tracking issue for this https://github.com/near/mpc/issues/2866

crates/node/src/providers/verify_foreign_tx/sign.rs
DSharifi: Creating a follow up issue to tackle this

DSharifi: https://github.com/near/mpc/issues/2867

pbeza (approved):

barakeinav1 (approved):

pytest/tests/test_foreign_transaction_validation.py
```diff
@@ -482,40 +501,49 @@ def test_verify_foreign_transaction_abstract(
...
-        assert isinstance(payload_hash, str), (
-            f"Expected hex string payload_hash, got: {type(payload_hash)}"
-        )
-        assert len(payload_hash) == 64, (
-            f"Expected 64 hex chars in payload_hash, got: {len(payload_hash)}"
-        )
+    # Find the Secp256k1 domain
+    contract_state = cluster.contract_state()
+    domains = contract_state.get_running_domains()
+    print(f"domains: {domains}, types: {[(type(d), dir(d)) for d in domains]}")
```
barakeinav1, line 522: nit: This looks like a leftover debug print — `dir(d)` dumps all attributes of the object. Should be removed before merge.

```suggestion
    secp_domain = next(d for d in domains if d.curve == "Secp256k1")
```

DSharifi: Let's keep it. I often miss having debug information in the tests when things fail

PR #2824, opened 2026-04-10
feat: support BNB foreign transactions in the near-mpc-sdk crate

### LINKED ISSUE #2812: BNB foreign-chain request support in mpc-SDK
### Background

When supporting foreign chain validation requests for the BNB chain, we should also extend the near-mpc-sdk crate's functionality to also create foreign chain validation requests to the BNB chain.

### User Story

As a user of the near-mpc-sdk I want to be able to construct foreign chain validaiton requests for the BNB chain.

### Acceptance Criteria

The near-mpc-sdk crate is able to create requests and verify BNB foreign chain requests

_No response_

DSharifi:
closes #2812


DSharifi:
@claude review

crates/near-mpc-sdk/tests/abstract_chain.rs
DSharifi, line 1: sanity check to see that the public API is preserved

crates/near-mpc-sdk/src/foreign_chain.rs
```diff
@@ -21,8 +23,8 @@ pub use near_mpc_contract_interface::types::{
 
 #[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, BorshSerialize, BorshDeserialize)]
 pub struct ForeignChainSignatureVerifier {
-    expected_extracted_values: Vec<ExtractedValue>,
-    request: ForeignChainRpcRequest,
+    pub(crate) expected_extracted_values: Vec<ExtractedValue>,
+    pub(crate) request: ForeignChainRpcRequest,
```
anodar: Do these really need to be public? Shouldn't private fields from parent module be accessible in submodules ? 

Same for `ForeignChainRpcRequestWithExpectations` fields.

crates/near-mpc-sdk/src/foreign_chain/abstract_chain.rs
```diff
@@ -1,152 +1,52 @@
...
 impl ForeignChainRequestBuilder<AbstractRequest<NotSet, NotSet>, NotSet> {
     pub fn new_abstract() -> Self {
         Self {
-            request: AbstractRequest {
+            request: EvmRequest {
                 tx_id: NotSet,
                 finality: NotSet,
                 expected_block_hash: None,
                 expected_logs: vec![],
+                _chain: std::marker::PhantomData,
```
anodar, line 34: Could we not drop Abstract struct entirely and just have wrap function in EvmRequest? 

Something like:

```

  pub struct EvmRequest<TxId, Finality> {
      pub(crate) tx_id: TxId,
      pub(crate) finality: Finality,
      pub(crate) expected_block_hash: Option<EvmBlockHash>,
      pub(crate) expected_logs: Vec<ExpectedLog>,
      pub(crate) wrap: fn(EvmRpcRequest) -> ForeignChainRpcRequest,
  }
```

We'd avoid this phantomData here too.

crates/near-mpc-sdk/src/foreign_chain/bnb.rs
```diff
@@ -0,0 +1,259 @@
...
+                expected_logs: vec![],
+                _chain: std::marker::PhantomData,
+            },
+            domain_id: NotSet,
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
```
anodar, line 42: We can probably deduplicate these unit tests across chains in single file (maybe evm.rs), that runs same tests for different chains.

anodar (dismissed):

crates/near-mpc-sdk/src/foreign_chain/bnb.rs
DSharifi: Good point. Moved it all to evm.rs except for the test to check that the right payload is constructed which is specific per chain

anodar (dismissed):
Thank you!

crates/near-mpc-sdk/src/foreign_chain/abstract_chain.rs
netrome: I like this. I guess this is slightly more expensive than the phantom data variant, and considering this may be used on-chain that might actually matter though.

netrome: Though thinking more, removing the marker type could lead to less structs being generated before compilation which could also impact binary sizes positively.

Anyway as with anything performance related it's better to measure and adjust - I'm okay with both solutions.

crates/near-mpc-sdk/src/foreign_chain/evm.rs
```diff
@@ -0,0 +1,327 @@
...
+        assert_eq!(builder.request.finality, EvmFinality::Finalized);
+    }
+
+    #[test]
+    fn with_expected_block_hash_sets_expected_value() {
+        let tx_id = EvmTxId::from([123; 32]);
+        let expected_hash = [9; 32];
+
+        let builder = ForeignChainRequestBuilder::new_abstract()
+            .with_tx_id(tx_id)
```
netrome, line 195: Nit: Would be nice with given, when, then sections here as well.

netrome (dismissed):
Nice stuff. Just chiming in + a nit.

anodar (approved):

netrome (approved):

crates/near-mpc-sdk/src/foreign_chain/abstract_chain.rs
DSharifi: Keeping as is. We can revisit once I add the other evm chains and we can better see how to simplify this.

PR #2827, opened 2026-04-13
fix: block GCP keyshare storage config in TEE mode

### LINKED ISSUE #2485: Check configuration values that the launcher needs to intercept
### Background

There are a couple of configuration values for the node, which the user should not be able to set when running in TEEs.

For example, the user should not set `[mpc_config.tee.authority]`.

We also need to look into the GCP config, if that allows the node to dump secrets to gcp secret storage, even if they are running in TEEs.

```rust
/// GCP keyshare storage settings. Optional — omit if not using GCP.
    pub gcp: Option<GcpStartConfig>,
```

### User Story

_No response_

### Acceptance Criteria

We have gone over all configuration values that the node accepts, and sanity checked that the launcehr intercepts all config values that the user should not be able to configure.

_No response_

barakeinav1:
## Summary
- Block the `[gcp]` config section when the launcher runs in TEE mode
- In TEE mode, an operator should not be able to send keyshares to an external GCP Secret Manager project — this would defeat the TEE protection against key theft
- The `gcp` key is rejected with a dedicated `TeeRestrictedConfigKey` error (distinct from `ReservedConfigKey` used for always-blocked keys like `tee`)
- GCP config remains allowed in non-TEE mode (no change to existing behavior)
- Added a CI guard test in `node-config` that fails if the `gcp` field name is renamed without updating the launcher

## Changes
- `crates/tee-launcher/src/config.rs`: Added `reject_in_tee_mode()` helper, block `gcp` in TEE mode
- `crates/tee-launcher/src/error.rs`: Added `TeeRestrictedConfigKey` error variant
- `crates/tee-launcher/src/lib.rs`: Pass `platform` to `intercept_node_config()`
- `crates/node-config/src/start.rs`: Added test guarding the `gcp` field name coupling
- Added tests for both TEE (rejected) and non-TEE (allowed) cases

Closes #2485

## Test plan
- [x] `cargo test -p tee-launcher` — 48 tests pass
- [x] `cargo test -p mpc-node-config` — 17 tests pass
- [ ] CI


crates/tee-launcher/src/error.rs
```diff
@@ -83,6 +83,9 @@ pub enum LauncherError {
     #[error("User config contains reserved key [{0}] — remove it from mpc_node_config")]
     ReservedConfigKey(String),
 
+    #[error("[{0}] is not allowed in TEE mode")]
```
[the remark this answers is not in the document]
barakeinav1: prefer to keep it simple

barakeinav1:
@claude review

crates/tee-launcher/src/config.rs
```diff
@@ -1,22 +1,41 @@
...
     insert_reserved(
         &mut node_config,
         "tee",
         toml::Value::try_from(tee_config).expect("tee config serializes to TOML"),
     )?;
+
+    if platform == Platform::Tee {
+        // Must match the field name in node-config's StartConfig::gcp
+        reject_in_tee_mode(&node_config, "gcp")?;
+    }
```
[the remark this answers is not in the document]
barakeinav1: update description

crates/tee-launcher/src/config.rs
```diff
@@ -239,11 +258,54 @@ key = "value"
...
+project_id = "my-project"
+"#,
+        )
+        .unwrap();
+
+        // when
+        let result = intercept_node_config(config, &sample_tee_config(), Platform::NonTee);
+
+        // then
+        assert!(result.is_ok());
```
DSharifi: Nit:
I'm surprised clippy didn't fire here.

We've configured it to lint on this pattern to encourage `assert_matches!(...)`

crates/tee-launcher/src/config.rs
```diff
@@ -1,22 +1,41 @@
...
 ) -> Result<toml::Table, LauncherError> {
     insert_reserved(
         &mut node_config,
         "tee",
         toml::Value::try_from(tee_config).expect("tee config serializes to TOML"),
     )?;
+
+    if platform == Platform::Tee {
+        // Must match the field name in node-config's StartConfig::gcp
+        reject_in_tee_mode(&node_config, "gcp")?;
```
DSharifi: Let's create a const for this "gcp" string.

crates/tee-launcher/src/config.rs
```diff
@@ -1,22 +1,41 @@
...
 }
 
+/// Return an error if the user config contains a key that is blocked in TEE mode.
+fn reject_in_tee_mode(table: &toml::Table, key: &str) -> Result<(), LauncherError> {
+    if table.contains_key(key) {
+        return Err(LauncherError::TeeRestrictedConfigKey(key.to_string()));
+    }
+    Ok(())
+}
+
```
DSharifi: Let's either inline this logic (called once), or we use a constant defined blacklisted keys that it checks.

DSharifi (changes requested):
Looks nice, with good test coverage.

One small blocker:
- "gcp" string for the table should be a const.

Added some optional comments as well

DSharifi (dismissed):

crates/node-config/src/start.rs
```diff
@@ -165,3 +165,28 @@ pub enum DownloadConfigType {
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// The tee-launcher blocks the "gcp" key in TEE mode using the hardcoded
+    /// string "gcp" (see crates/tee-launcher/src/config.rs).
+    /// This test ensures the TOML field name for GCP config is still "gcp".
+    /// If this field is renamed, update the launcher's blocked key list too.
+    #[test]
+    fn gcp_toml_field_name_is_gcp() {
```
netrome: Nice! I wonder if we could have this enforced by the type system somehow but

crates/tee-launcher/src/config.rs
```diff
@@ -1,19 +1,37 @@
...
+    platform: Platform,
 ) -> Result<toml::Table, LauncherError> {
     insert_reserved(
         &mut node_config,
         "tee",
         toml::Value::try_from(tee_config).expect("tee config serializes to TOML"),
     )?;
+
+    if platform == Platform::Tee {
+        for key in TEE_BLOCKED_KEYS {
```
netrome, line 28: Using a hard-coded string feels a bit brittle here. Couldn't we try to parse the config and check if the config contains a gcp field in this case?

Additionally I question why we allow the `gcp` field in the first place. I'd advocate for just dropping it completely.

netrome (dismissed):
The change looks correct but I'm not sold on the approach. Additionally, I think we could make things simpler by just dropping support for the `gcp` option altogether. I don't see why we'd need it in the launcher.

netrome (approved):

crates/tee-launcher/src/config.rs
DSharifi: This is what the code is already doing? Notice that `node_config` is a `toml::Table` and we are calling `contains_key` on it.

The issue is that the node operator can pass arbitrary configs to the node, and in our current implementation a GCP config will lead to all the keys being dumped to GCP key storage.

DSharifi (approved):

PR #2830, opened 2026-04-13
ci: make rust-cache key depend on week

### LINKED ISSUE #2828: Weekly Rust cache refresh does not actually refresh
### Background

`.github/workflows/ci.yml` runs a Saturday cron (`0 12 * * 6`) intended to rebuild the Rust cache each week. It relies on `Swatinem/rust-cache` with `lookup-only: \${{ github.event_name == 'schedule' }}` to skip restore on schedule so the build runs clean, then save a fresh cache. The save never happens.

Last Saturday's scheduled run (https://github.com/near/mpc/actions/runs/24282207602, job `mpc-pytest-build`) shows the pattern:

```
Found cache key \"v0-rust-mpc-pytest-build-Linux-x64-1d14cc0c-44e225f4\" full match: true.
Lookup only - skipping download
...
Post job cleanup.
Cache up-to-date.   <- save skipped
```

Root cause is upstream and cannot be worked around with config:

- On a full key match, `Swatinem/rust-cache`'s restore step does not call `saveState()` (see [`restore.ts`](https://github.com/Swatinem/rust-cache/blob/master/src/restore.ts)), so the post step sees \"cache up to date\" and skips save ([`save.ts`](https://github.com/Swatinem/rust-cache/blob/master/src/save.ts)). `lookup-only` / `save-if` do not override this.
- The skip exists because the backend itself rejects re-reserving an existing key. From [Swatinem/rust-cache#277](https://github.com/Swatinem/rust-cache/pull/277):

  > Failed to CreateCacheEntry: Received non-retryable error: Failed request: (409) Conflict: cache entry with the same key, version, and scope already exists

- The rust-cache action has no input to force overwrite (verified against every input in `action.yml` and every merged PR). The #181 reporter's own workaround in eqlabs/pathfinder exhibits the same silent skip in their scheduled run logs.

The only viable approaches are (a) deleting the server-side cache entry before save, or (b) rotating the cache key. (a) is fragile because it requires reproducing Swatinem's internal key and `cacheVersion` formulas exactly.

### Acceptance Criteria

- On scheduled runs (Saturday cron), the Rust cache is genuinely refreshed (a new entry is saved and reused by later runs in the week).
- Subsequent non-scheduled runs within the same week restore that cache.

### Proposed minimal change

Rotate `prefix-key` weekly across the five `Swatinem/rust-cache` invocations in `ci.yml` by prepending a `date -u +%Y-%U` step and appending `-week-\${{ env.WEEK }}` to each `prefix-key`. Remove the `lookup-only: \${{ github.event_name == 'schedule' }}` lines, which no longer serve a purpose once keys rotate. Old caches age out via Warpbuild's default eviction.

gilcu3:
Closes #2828 

Solving this has required a lot of tinkering, I really hope it works this time. See the issue for a full explanation of why achieving this has been hard

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,9 +11,12 @@ on:
   push:
     branches:
       - main
-  # Weekly cron to clear and rebuild Rust cache (https://github.com/Swatinem/rust-cache/issues/181).
+  # Primes the new week's Rust cache on main right after the weekly `prefix-key` rolls
+  # over at UTC Sun 00:00. Without this, the first PR on Sunday pays a full cold-build
+  # cost and (since `save-if` requires main) doesn't save, so every subsequent Sunday
+  # PR keeps cold-building until something merges to main. See near/mpc#2828.
   schedule:
-    - cron: "0 12 * * 6"
+    - cron: "5 0 * * 0"
 
 jobs:
   docker-tee-build:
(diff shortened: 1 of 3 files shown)


anodar (approved):
Nice!

DSharifi (approved):

PR #2831, opened 2026-04-13
refactor: decouple node production code from mpc-contract crate

SimonRastikian:
Follows https://github.com/near/mpc/pull/2740, which was a huge PR
Closes partly https://github.com/near/mpc/issues/381

## Details

Switch all node production and test imports from `mpc_contract::*` to `near_mpc_contract_interface::types::*`. The contract crate remains as a dev-dependency for test utilities that still construct internal contract state objects.

Key changes:
- Expand interface crate with shared types: kdf (derive_tweak, derive_app_id), CKDRequest/CKDResponse, SignRequest, LegacySignRequestArgs, YieldIndex, BackupServiceInfo, DestinationNodeInfo, NodeId
- Rename Curve to SignatureScheme throughout the node (Edwards25519 → Ed25519, domain.curve → domain.scheme)
- Switch method calls to field access on interface DTOs (e.g. key_event.epoch_id() → key_event.epoch_id)
- Rewrite protocol_state_to_string as local format_protocol_state
- Remove mpc-contract from regular dependencies (kept as dev-dep)

## Coming up in a next PR
Removinh mpc-contract from dev-dependencies entirely by rewriting test files that import mpc_contract::*:


SimonRastikian:
@claude  review this PR

crates/contract/src/crypto_shared/kdf.rs
```diff
@@ -9,43 +9,16 @@ use k256::{
...
-    // indicate the end of the account id in derivation path.
-    // Do not reuse this hash function on anything that isn't an account
-    // ID or it'll be vulnerable to Hash Malleability/extension attacks.
-    let derivation_path = format!("{derivation_prefix}{},{}", predecessor_id, path);
-    let mut hasher = Sha3_256::new();
-    hasher.update(derivation_path);
-    let hash: [u8; 32] = hasher.finalize().into();
-    hash
+    dtos::kdf::derive_app_id(predecessor_id, derivation_path)
 }
```
SimonRastikian, line 22: Moved to smart contract interface (kdf.rs) as this logic is repetitive

crates/near-mpc-contract-interface/src/types/sign.rs
```diff
@@ -54,3 +54,85 @@ pub enum Payload {
...
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct LegacySignRequestArgs {
+    pub path: String,
+    pub payload_v2: Option<Payload>,
+    #[serde(rename = "payload")]
+    pub deprecated_payload: Option<[u8; 32]>,
+    pub domain_id: Option<DomainId>,
+    #[serde(rename = "key_version")]
+    pub deprecated_key_version: Option<u32>,
+}
```
SimonRastikian, line 100: The contract and interface have parallel type hierarchies such as Payload, DomainId... that need to be unified before we can eliminate the duplication. That's part of the next PR which will try to get rid of shared_crypto

SimonRastikian: The contract's SignRequestArgs and the interface's LegacySignRequestArgs are the same struct with different names but this helps splitting SC and Node

SimonRastikian: Replaces type called from the smart contract in crates/node/src/indexer/handler.rs

crates/node/src/migration_service/types.rs
```diff
@@ -131,32 +129,59 @@ fn infer_migration_status(
...
-            ed25519_dalek::VerifyingKey::from_near_sdk_public_key(
-                &info.destination_node_info.sign_pk,
-            )
-            .inspect_err(|_| tracing::warn!(target: "Migration Service", "Error parsing public key from chain."))
-            .is_ok_and(|key| key == *my_p2p_public_key)
+            near_sdk::PublicKey::from_str(&info.destination_node_info.sign_pk)
+                .map_err(|e| anyhow::anyhow!("{}", e))
+                .and_then(|pk| ed25519_dalek::VerifyingKey::from_near_sdk_public_key(&pk))
+                .inspect_err(|_| tracing::warn!(target: "Migration Service", "Error parsing public key from chain."))
+                .is_ok_and(|key| key == *my_p2p_public_key)
```
SimonRastikian, line 136: This is because DestinationNodeInfo.sign_pk changed from near_sdk::PublicKey (contract type) to String (interface DTO type).

crates/node/src/migration_service/types.rs
```diff
@@ -131,32 +129,59 @@ fn infer_migration_status(
...
-            test_utils::{
-                bogus_ed25519_near_public_key, bogus_ed25519_public_key, gen_participant,
-            },
-        },
-        state::{
-            test_utils::{gen_initializing_state, gen_resharing_state, gen_running_state},
-            ProtocolContractState,
-        },
+    use mpc_contract::state::test_utils::{
+        gen_initializing_state, gen_resharing_state, gen_running_state,
```
SimonRastikian, line 145: Still depends on contract but will be removed in a followup PR

crates/node/src/tests/faulty.rs
```diff
@@ -36,14 +36,14 @@ async fn test_faulty_cluster() {
 
     let domain = DomainConfig {
         id: DomainId(0),
-        curve: Curve::Secp256k1,
-        purpose: DomainPurpose::Sign,
+        scheme: SignatureScheme::Secp256k1,
+        purpose: Some(DomainPurpose::Sign),
```
SimonRastikian, line 40: One of the many renamings curve to scheme

crates/node/src/web.rs
```diff
@@ -386,7 +386,126 @@ async fn contract_state(state: State<WebServerState>) -> String {
         // Clone to avoid holding a lock
         .clone();
 
-    protocol_state_to_string(&protocol_state)
+    format_protocol_state(&protocol_state)
+}
+
+fn format_protocol_state(contract_state: &ProtocolContractState) -> String {
```
SimonRastikian, line 392: This is purely for human readable state.
I suggest this stays here in this PR. In a next PR I would remove more of the de duplicated logic in utils.rs and in here however to delete the deduplicated logic I would need to do it cleanly on the SC side:

  So the plan is:                                                                                                                                                                                                                                      
  1. Add impl Display for ProtocolContractState in the interface crate (using the node's version which works on DTO types)                                                                                                                             
  2. Node's web.rs: replace format_protocol_state(&state) with state.to_string()                                                                                                                                                                       
  3. Contract's utils.rs: remove protocol_state_to_string, convert internal state to DTO and use Display
  4. devnet/src/mpc.rs: switch to interface type's Display

I did not do it here not to make the current PR larger

crates/node/src/indexer/participants.rs
```diff
@@ -200,11 +199,11 @@ impl ContractState {
             ProtocolContractState::Resharing(state) => {
                 let resharing_state = Some(ContractResharingState {
                     new_participants: convert_participant_infos(
-                        state.resharing_key.proposed_parameters().clone(),
+                        state.resharing_key.parameters.clone(),
                         port_override,
                     )?,
                     reshared_keys: Keyset {
-                        epoch_id: state.prospective_epoch_id(),
+                        epoch_id: state.previous_running_state.keyset.epoch_id.next(),
```
barakeinav1, line 206: Bug: `previous_running_state.keyset.epoch_id.next()` is not equivalent to the original `state.prospective_epoch_id()`.

The contract's `prospective_epoch_id()` returns `self.resharing_key.epoch_id()` (see `crates/contract/src/state/resharing.rs:43-44`). This value **increments on each re-proposal** via `vote_new_parameters()` (resharing.rs:77), so after a re-proposal it can be `previous + 2`, `previous + 3`, etc.

`previous_running_state.keyset.epoch_id` is frozen at the time resharing started and never changes, so `.next()` always returns the initial resharing epoch.

The interface DTO already exposes the correct value:
```rust
epoch_id: state.resharing_key.epoch_id,
```

**Impact**: After a re-proposal during resharing, nodes would construct `Keyset`s with an incorrect epoch ID, causing epoch mismatches that could prevent the resharing protocol from completing.

crates/node/src/web.rs
```diff
@@ -386,7 +386,126 @@ async fn contract_state(state: State<WebServerState>) -> String {
...
+            }
+            output.push_str("  Parameters:\n");
+            params_to_string(&mut output, &state.parameters);
+        }
+        ProtocolContractState::Resharing(state) => {
+            output.push_str("Contract is in Resharing state\n");
+            output.push_str(&format!(
+                "  Epoch transition: original {} --> prospective {}\n",
+                state.previous_running_state.keyset.epoch_id,
+                state.previous_running_state.keyset.epoch_id.next()
```
barakeinav1: Same bug as in `participants.rs`: this should use `state.resharing_key.epoch_id` instead of `state.previous_running_state.keyset.epoch_id.next()`.

```suggestion
                state.resharing_key.epoch_id
```

barakeinav1 (changes requested):
Two instances of an incorrect prospective epoch ID in the Resharing state — see inline comments.

crates/node/src/indexer/fake.rs
```diff
@@ -40,14 +39,24 @@ use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet, VecDeque};
 use std::sync::{atomic::AtomicBool, Arc};
 use tokio::sync::{broadcast, mpsc, watch};
 
+/// Convert an interface type to a contract type via JSON round-trip.
+fn to_contract<T: serde::de::DeserializeOwned>(dto: &impl serde::Serialize) -> T {
+    serde_json::from_value(serde_json::to_value(dto).unwrap()).unwrap()
+}
+
+/// Convert a contract type to an interface type via JSON round-trip.
+fn to_interface<T: serde::de::DeserializeOwned>(contract: &impl serde::Serialize) -> T {
```
barakeinav1, line 48: nit: These generic `to_contract`/`to_interface` JSON round-trip helpers are duplicated in multiple files with slight variations:

- `fake.rs`: generic `to_contract<T>` / `to_interface<T>` (here)
- `tests.rs`: `to_contract_domain_config`, `to_contract_domain_configs`, `to_interface_keyset`
- `assets/test_utils.rs`: `to_interface_params`
- `migration_service/types.rs`: `to_interface_state`
- `backup-cli/contract_state_fixture.rs`: inline `serde_json::from_value(serde_json::to_value(...))`

The generic versions here in `fake.rs` are the cleanest. Consider extracting them into a shared test utility (e.g. in `test-utils` or a `tests/common.rs` module) and reusing everywhere. This will also make the next PR (removing `mpc-contract` from dev-deps) easier since you'll have a single place to update.

crates/node/src/config.rs
```diff
@@ -477,7 +477,15 @@ pub fn auth_config_to_rpc_auth(
 #[expect(non_snake_case)]
 pub mod tests {
     use assert_matches::assert_matches;
-    use mpc_contract::primitives::test_utils::bogus_ed25519_near_public_key;
+    fn bogus_ed25519_near_public_key() -> near_sdk::PublicKey {
```
barakeinav1, line 480: nit: `bogus_ed25519_near_public_key()` is now copy-pasted identically in both `config.rs:480` and `migration_service/types.rs:153`. Previously these came from `mpc_contract::primitives::test_utils`. Consider extracting to a shared test helper to avoid the duplication.

barakeinav1 (commented):
Additional review comments — non-blocking nits based on team conventions.

crates/near-mpc-contract-interface/src/types/tee.rs
```diff
@@ -0,0 +1,14 @@
...
+
+use serde::{Deserialize, Serialize};
+
+/// Identifier for a TEE node.
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
+pub struct NodeId {
+    /// Operator account.
+    pub account_id: near_account_id::AccountId,
+    /// TLS public key, encoded as a string (e.g., "ed25519:...").
+    pub tls_public_key: String,
```
barakeinav1, line 11: nit: Using bare `String` for public keys is a type-safety degradation compared to the contract's `near_sdk::PublicKey`. You lose validated-at-construction guarantees — a malformed string won't be caught until someone tries to parse it downstream (e.g. `near_sdk::PublicKey::from_str` in `convert_participant_infos`).

I understand the motivation — avoiding the `near-sdk` dependency in the interface crate. But consider a lightweight validated newtype (e.g. `NearPublicKeyString`) that validates format on construction, rather than a bare `String`. That would preserve the decoupling while recovering type safety.

Same applies to `account_public_key` on line 13, and `sign_pk: String` in `ParticipantInfo`.

Not blocking for this PR, but worth considering as a follow-up for the interface crate's string-typed key fields.

barakeinav1:
## String vs typed structs in interface DTOs

This PR introduces several fields that use bare `String` where the contract had typed structs (e.g., `near_sdk::PublicKey`). I understand the motivation — avoiding the `near-sdk` dependency in the interface crate — but it's worth considering the trade-off explicitly, since it's a type-safety degradation across a security-sensitive codebase.

**Affected fields:**

| Struct | Field | Contract type | Interface type |
|---|---|---|---|
| `ParticipantInfo` | `sign_pk` | `near_sdk::PublicKey` | `String` |
| `NodeId` | `tls_public_key` | `near_sdk::PublicKey` | `String` |
| `NodeId` | `account_public_key` | `Option<near_sdk::PublicKey>` | `Option<String>` |
| `DestinationNodeInfo` | `signer_account_pk` | `near_sdk::PublicKey` | `String` |
| `AccountId` | inner | `near_account_id::AccountId` | `String` (wrapper) |

Note that `BackupServiceInfo.public_key` kept a typed `Ed25519PublicKey`, so the approach is inconsistent — some fields got proper types, others got bare strings.

**What's lost:**
- **Validated-at-construction** — A `near_sdk::PublicKey` guarantees valid key format. A `String` can hold anything and won't fail until someone parses it downstream (e.g., `near_sdk::PublicKey::from_str` in `convert_participant_infos`).
- **Equality semantics** — Typed keys compare by bytes. Strings compare literally, which is weaker (e.g., different encodings of the same key).
- **Discoverability** — A `String` field doesn't tell the reader what format to expect without reading the doc comment.

**Suggestion:** Consider introducing lightweight validated newtypes (e.g., `NearPublicKeyString`) in the interface crate that validate format on construction. This preserves the `near-sdk` decoupling while recovering type safety. Not blocking for this PR, but worth thinking about — especially since this is a security-critical system where a malformed key string could cause subtle failures.

barakeinav1:
There are some design choices I'd like to get answers on as well:

1. **Should `near-sdk` be a dependency of the interface crate?** If not then we need to duplicate structs, or use strings (see next point)

2. **Should we use strings instead of structs in the interface?** (avoids duplicating structs or importing `near-sdk`), but this has less type safety — e.g. `ParticipantInfo.sign_pk` is now a bare `String` instead of `near_sdk::PublicKey`

3. **The `PublicKeyExtended → PublicKey` conversion used to be infallible, now it's a `TryFrom` that can fail at runtime.** Should we validate earlier (e.g. at deserialization) instead?

4. **The contract↔interface conversion in tests uses `serde_json::from_value(to_value(&x))` — nothing enforces the two types stay in sync.** Fine as a stopgap, or do we want snapshot tests before this spreads?

5. **`derive_tweak` / `derive_app_id` have security-sensitive prefix strings and aren't just DTOs** — should they live in a separate crate like `mpc-primitives` instead of the interface crate?

gilcu3:
> * **Should `near-sdk` be a dependency of the interface crate?** If not then we need to duplicate structs, or use strings (see next point)

`near-sdk` should not be a dependency unless it is really needed. For that case we had it feature gated before IIRC. So we would need to check here if that's the case. Could result in a new (or reuse existing) feature for example, if only the node would use the types.

> * **Should we use strings instead of structs in the interface?** (avoids duplicating structs or importing `near-sdk`), but this has less type safety — e.g. `ParticipantInfo.sign_pk` is now a bare `String` instead of `near_sdk::PublicKey`

We have types for public keys already in the dto, we should reuse those.


> * **The `PublicKeyExtended → PublicKey` conversion used to be infallible, now it's a `TryFrom` that can fail at runtime.** Should we validate earlier (e.g. at deserialization) instead?

Normally yes, better to fail at deserialization, but we need to check in this case if that is possible or not. 

> * **The contract↔interface conversion in tests uses `serde_json::from_value(to_value(&x))` — nothing enforces the two types stay in sync.** Fine as a stopgap, or do we want snapshot tests before this spreads?

Yeah that does not sound like the patterns we should use. Existing code should contain better patterns already.

> * **`derive_tweak` / `derive_app_id` have security-sensitive prefix strings and aren't just DTOs** — should they live in a separate crate like `mpc-primitives` instead of the interface crate?

Where to put them was a previous question I had in mind as well. The problem lies on who is supposed to use them. If we want our users to have access to them, and that probably makes sense, then they could be in the contract interface, or in another crate that is also public (I dont think we have any right now which fulfills this purpose). We could have a small discussion about this on slack.

gilcu3:
We split this up in many smaller PRs, still in progress, will be done once #381 is closed

PR #2832, opened 2026-04-13
refactor: make preconditions uniform for requests in the contract

### LINKED ISSUE #1172: Reduce duplication between the `sign` and `request_app_private_key` smart contract methods
### Background

This is a follow-up issue to the discussion started in https://github.com/near/mpc/pull/1165#issuecomment-3324128290:

> > > We have the exact same deposit mechanism and refund tests for ckd as well. Could you update these as well?
> > 
> > Oh, good point, thanks! BTW, is there any reason we basically copy-pasted `sign()` into `request_app_private_key()` instead of extracting the common parts? They look quite similar. Same goes for the tests.
> 
> Don't think so. Would be nice to factor out the common parts.

The above discussion refers to the following two functions, along with their corresponding tests updated in #1165:

https://github.com/near/mpc/blob/5fcd4788bf881c0e78de5c28c35f51bcca39bdfe/crates/contract/src/lib.rs#L139-L248

https://github.com/near/mpc/blob/5fcd4788bf881c0e78de5c28c35f51bcca39bdfe/crates/contract/src/lib.rs#L314-L419

### User Story

_No response_

### Acceptance Criteria

Refactor the `sign` and `request_app_private_key` functions to reduce code duplication, along with their corresponding tests (see the tests updated in #1165 for reference).

- https://github.com/near/mpc/pull/1165#issuecomment-3324128290
- https://github.com/near/mpc/pull/1165#issuecomment-3324142595

gilcu3:
Closes #1172 

Before:
```
❯ ls -l /home/rey/opt/near/mpc/target/near/mpc_contract/mpc_contract.wasm
-rw------- 1 rey rey 1508960 Apr 13 12:04 /home/rey/opt/near/mpc/target/near/mpc_contract/mpc_contract.wasm
```

After:

```
❯ ls -l /home/rey/opt/near/mpc/target/near/mpc_contract/mpc_contract.wasm
-rw------- 1 rey rey 1508324 Apr 13 11:57 /home/rey/opt/near/mpc/target/near/mpc_contract/mpc_contract.wasm
```

Mere 636 bytes :|

### DIFF
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -221,6 +221,103 @@ impl MpcContract {
             .insert(request.clone(), YieldIndex { data_id })
             .is_some()
     }
+
+    /// Common preconditions enforced on every user-facing request method (`sign`,
+    /// `request_app_private_key`, `verify_foreign_transaction`):
+    ///
+    /// 1. The target domain exists and its purpose matches `expected_purpose`.
+    /// 2. The caller attached enough prepaid gas to perform the yield/resume flow.
+    /// 3. The caller attached at least `minimum_deposit` (excess is refunded).
+    /// 4. The contract is currently accepting user requests.
+    ///
+    /// Returns the validated domain config and the caller's account id.
+    fn check_request_preconditions(
+        &self,
+        domain_id: DomainId,
+        expected_purpose: DomainPurpose,
+        minimum_gas: Gas,
+        minimum_deposit: NearToken,
+    ) -> (DomainConfig, AccountId) {
+        // 1. Look up the domain and check its purpose.
+        let domains = match self.protocol_state.domain_registry() {
+            Ok(domains) => domains,
+            Err(err) => env::panic_str(&err.to_string()),
+        };
+        let Some(domain_config) = domains.get_domain_by_domain_id(domain_id) else {
+            env::panic_str(
+                &InvalidParameters::DomainNotFound {
+                    provided: domain_id,
+                }
+                .to_string(),
+            );
+        };
+        if domain_config.purpose != expected_purpose {
+            env::panic_str(
+                &InvalidParameters::WrongDomainPurpose {
+                    domain_id: domain_config.id,
+                    expected: expected_purpose,
+                    actual: domain_config.purpose,
+                }
+                .to_string(),
+            );
+        }
+        let domain_config = domain_config.clone();
+
+        // 2. Make sure the call will not run out of gas doing yield/resume logic.
+        let prepaid_gas = env::prepaid_gas();
+        if prepaid_gas < minimum_gas {
+            env::panic_str(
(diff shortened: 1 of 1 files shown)


pbeza (approved):

netrome:
> Mere 636 bytes :|

I know I already dropped this one for a similar change but still...

image

crates/contract/src/lib.rs
```diff
@@ -485,24 +496,13 @@ impl MpcContract {
                 .return_ck_and_clean_state_on_success_call_tera_gas,
         );
 
-        let promise_index = env::promise_yield_create(
+        let callback_args = serde_json::to_vec(&(&request,)).unwrap();
```
netrome, line 499: Unrelated to the PR, why do we need the callback args to be serialized as a tuple (or array I guess the format will be) explicitly?

crates/contract/src/lib.rs
```diff
@@ -2650,6 +2600,99 @@ mod tests {
...
+        let (config, predecessor) = contract.check_request_preconditions(
+            DomainId::default(),
+            DomainPurpose::Sign,
+            Gas::from_tgas(1),
+            NearToken::from_yoctonear(1),
+        );
+        assert_eq!(config.id, DomainId::default());
+        assert_eq!(config.curve, Curve::Secp256k1);
+        assert_eq!(config.purpose, DomainPurpose::Sign);
+        assert_eq!(predecessor.as_str(), "contract_account.near");
```
netrome, line 2627: Looks good, but I still think it would look better with

```suggestion
    fn check_request_preconditions__returns_domain_config_and_predecessor_on_valid_call() {
        // Given
        let mut rng = rand::rngs::StdRng::from_seed([42u8; 32]);
        let (_, contract, _) = basic_setup(Curve::Secp256k1, &mut rng);
        
        // When
        let (config, predecessor) = contract.check_request_preconditions(
            DomainId::default(),
            DomainPurpose::Sign,
            Gas::from_tgas(1),
            NearToken::from_yoctonear(1),
        );
        
        // Then
        assert_eq!(config.id, DomainId::default());
        assert_eq!(config.curve, Curve::Secp256k1);
        assert_eq!(config.purpose, DomainPurpose::Sign);
        assert_eq!(predecessor.as_str(), "contract_account.near");
```

netrome (approved):
Looks good, thanks!

crates/contract/src/lib.rs
netrome: (Goes for all tests)

gilcu3: right, forgot here :/

crates/contract/src/lib.rs
gilcu3: no idea, I just followed the existing pattern

PR #2844, opened 2026-04-13
test: update sandbox integration test to use a builder pattern for setup

### LINKED ISSUE #2853: Generalize init/init_env setup for foreign tx integration tests
### Background

See this thread for motivation; [In the ](https://github.com/near/mpc/pull/2820#discussion_r3071167702)

We want to avoid duplication of logic for setups of the integration tests in common.rs. See slack thread for motivation on why to use builders to solve this https://nearone.slack.com/archives/C0912BTG51T/p1776095552496359?thread_ts=1775823886.494019&cid=C0912BTG51T

### User Story

_No response_

### Acceptance Criteria

use a builder for the integration test setup generalized for all test scenarios

_No response_

DSharifi:
closes #2853


crates/contract/tests/sandbox/common.rs
```diff
@@ -151,112 +150,144 @@ pub async fn init_contract_running(
...
+            init_config: None,
+        }
+    }
+
+    /// Returns the first key with `ForeignTx` purpose.
+    pub fn foreign_tx_key(&self) -> &DomainKey {
+        self.keys
+            .iter()
+            .find(|k| k.domain_config.purpose == DomainPurpose::ForeignTx)
+            .expect("No ForeignTx domain in setup. Did you call .foreign_tx() on the builder?")
```
anodar:
```suggestion
            .expect("No ForeignTx domain in setup. Did you call .with_foreign_tx_domain() on the builder?")
```

crates/contract/tests/sandbox/common.rs
```diff
@@ -151,112 +150,144 @@ pub async fn init_contract_running(
...
+
+    pub async fn build(self) -> SandboxTestSetup {
+        let (worker, contract) = init().await;
+        let (accounts, participants) = gen_accounts(&worker, self.number_of_participants).await;
+        let threshold_parameters = make_threshold_params(&participants);
+
+        let mut keys = Vec::new();
+        let mut domain_configs = Vec::new();
+        let mut key_for_domains = Vec::new();
+        let mut domain_id_counter = 0u64;
```
anodar: Do you really need domain_id_counter? couldn't that just be domain_configs.size() ? 

You could still keep `domain_id` itself to make it explicit but counter seems redundant.

crates/contract/tests/sandbox/common.rs
```diff
@@ -151,112 +150,144 @@ pub async fn init_contract_running(
...
+
+        // Sign-purpose domains from curves
+        for curve in &self.curves {
+            let (pk, sk) = make_key_for_domain(*curve);
+            let purpose = infer_purpose_from_curve(*curve);
+            let domain_id = DomainId(domain_id_counter);
+            domain_id_counter += 2;
```
anodar: Why is this being increased by 2? Doesn't each iteration just add single domain?

anodar (dismissed):

crates/contract/tests/sandbox/common.rs
```diff
@@ -151,112 +150,144 @@ pub async fn init_contract_running(
...
+        self.keys
+            .iter()
+            .find(|k| k.domain_config.purpose == DomainPurpose::ForeignTx)
+            .expect("No ForeignTx domain in setup. Did you call .foreign_tx() on the builder?")
+    }
+}
+
+pub struct SandboxTestSetupBuilder {
+    curves: Vec<Curve>,
+    foreign_tx: bool,
```
gilcu3, line 181: I did not expect we would need this bool here, looks very weird. The Setup builder should receive instead all the necessary information, Curve, Purpose, etc

crates/contract/tests/sandbox/common.rs
```diff
@@ -151,112 +150,144 @@ pub async fn init_contract_running(
...
+
+        let mut keys = Vec::new();
+        let mut domain_configs = Vec::new();
+        let mut key_for_domains = Vec::new();
+        let mut domain_id_counter = 0u64;
+
+        // Sign-purpose domains from curves
+        for curve in &self.curves {
+            let (pk, sk) = make_key_for_domain(*curve);
+            let purpose = infer_purpose_from_curve(*curve);
```
gilcu3, line 219: it should take the purpose as input exactly to avoid this, that we don't need anymore. That would make all requests uniform

gilcu3 (commented):
Added some comments from a couple of days ago before they become still. For context, see [thread](https://nearone.slack.com/archives/C0912BTG51T/p1776175259790209?thread_ts=1775823886.494019&cid=C0912BTG51T)

crates/contract/tests/sandbox/common.rs
DSharifi: foreign transaction domains can only be Ecdsa. Why allow a curve parameter?

crates/contract/tests/sandbox/common.rs
DSharifi: `infer_purpose_from_curve` is pre-existing. Look how this `build` method is mostly moved code from `init_with_candidates`. This is not something introduced in this PR.

The scope of the PR is to add a builder, so we have a common setup interface that can be reused for foreign transaction tests.

crates/contract/tests/sandbox/common.rs
gilcu3: it was mainly for making it future proof, but we agreed in sync that this is fine for now, the builder pattern will make it simpler anyway

gilcu3 (approved):
Thank you!

anodar (approved):
Thank you!

PR #2847, opened 2026-04-13
suggestion: Use serde_yaml instead of manual string replacement for docker compose template

netrome:
Suggested follow-up after #2785 

The purpose of this PR is to clarify what I mean in this comment https://github.com/near/mpc/pull/2785#discussion_r3075203357 - this feels like the natural way we should render these docker compose templates and I'm curious to know if there's any good reason not to do it this way.

The PR is purely Claude generated but I've skimmed it and it seems in line with what I picture we should do - although it would probably need to be cleaned up before merging.


netrome:
serde_yaml is unmaintained so skipping this

PR #2851, opened 2026-04-14
fix: remove stale foreign chain votes from non-participants

gilcu3:
Closes #2769 

before
```
❯ ls -l /home/rey/opt/near/mpc/target/near/mpc_contract/mpc_contract.wasm
-rw------- 1 rey rey 1538703 Apr 14 09:02 /home/rey/opt/near/mpc/target/near/mpc_contract/mpc_contract.wasm
```

after
```
❯ ls -l /home/rey/opt/near/mpc/target/near/mpc_contract/mpc_contract.wasm
-rw------- 1 rey rey 1546077 Apr 14 09:00 /home/rey/opt/near/mpc/target/near/mpc_contract/mpc_contract.wasm
```

delta: +8KB, IMO could be fine, we are not yet close to the limit after this PR

### DIFF
--- a/crates/contract/src/config.rs
+++ b/crates/contract/src/config.rs
@@ -28,6 +28,8 @@ const DEFAULT_CLEAN_TEE_STATUS_TERA_GAS: u64 = 10;
 const DEFAULT_CLEANUP_ORPHANED_NODE_MIGRATIONS_TERA_GAS: u64 = 3;
 /// Prepaid gas for a `remove_non_participant_update_votes` call
 const DEFAULT_REMOVE_NON_PARTICIPANT_UPDATE_VOTES_TERA_GAS: u64 = 5;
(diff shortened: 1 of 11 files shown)


DSharifi (approved):

DSharifi (changes requested):
See slack thread: https://nearone.slack.com/archives/C0912BTG51T/p1776157059539019?thread_ts=1776150370.414399&cid=C0912BTG51T


`CLEAN_FOREIGN_CHAIN_DATA` succeeding but the main transaction failing will lead to all foreign chain transactions being rejected, since one node is no longer supporting any of the chains.

DSharifi (approved):
Discussed sync. The cleanup promise is only executed by the near runtime IFF the base transaction succeeds

anodar (approved):

PR #2855, opened 2026-04-14
chore: add ExpectedMeasurements to MockAttestation

### LINKED ISSUE #2490: Add measurements constraint to MockAttestation::WithConstraints
### Background

**As a** developer writing contract tests with mock attestations, **I want** `MockAttestation::WithConstraints` to support an `expected_measurements` field **so that** I can test measurement validation without requiring real Dstack attestation data.

`MockAttestation::WithConstraints` currently has constraints for:
- `mpc_docker_image_hash: Option<MpcDockerImageHash>`
- `launcher_docker_compose_hash: Option<LauncherDockerComposeHash>`
- `expiry_timestamp_seconds: Option<u64>`

But not for OS measurements (MRTD, RTMR0-2, key-provider event digest), which were added in #2447.

`crates/mpc-attestation/src/attestation.rs:53`

Related: #2447, #2448
Raised by @pbeza in https://github.com/near/mpc/pull/2447#discussion_r2952583450

### User Story
as an MPC developer I'd like to be able to test contract/mpc logic using mock attestation

## Acceptance Criteria
- Add `expected_measurements: Option<ExpectedMeasurements>` to `WithConstraints`
- `verify_mock_attestation` checks measurements when the field is `Some`
- `re_verify` for mock `WithConstraints` checks measurements against allowed set
- Existing tests pass (borsh layout change is safe since no real `WithConstraints` attestations on-chain)

gilcu3:
Closes #2490

### DIFF
--- a/crates/contract/tests/sandbox/gas_thresholds.json
+++ b/crates/contract/tests/sandbox/gas_thresholds.json
@@ -39,11 +39,11 @@
       "update_info": 7448
     },
     "400": {
-      "len": 10920,
-      "is_participant": 11078,
-      "info": 11078,
+      "len": 11500,
+      "is_participant": 11500,
+      "info": 11500,
       "validate": 13701,
-      "serialization": 11676,
+      "serialization": 12050,
       "insert": 20449,
       "update_info": 20612
     }
--- a/crates/mpc-attestation/src/attestation.rs
+++ b/crates/mpc-attestation/src/attestation.rs
@@ -28,7 +28,6 @@ pub enum Attestation {
     Mock(MockAttestation),
 }
 
-#[expect(clippy::large_enum_variant)]
 #[derive(Clone, Debug, Serialize, Deserialize, BorshDeserialize, BorshSerialize)]
 #[cfg_attr(
     all(feature = "abi", not(target_arch = "wasm32")),
@@ -39,6 +38,7 @@ pub enum VerifiedAttestation {
     Mock(MockAttestation),
 }
 
+#[expect(clippy::large_enum_variant)]
 #[derive(Debug, Default, Clone, Serialize, Deserialize, BorshDeserialize, BorshSerialize)]
 #[cfg_attr(
     all(feature = "abi", not(target_arch = "wasm32")),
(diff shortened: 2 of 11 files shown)


crates/contract/tests/sandbox/gas_thresholds.json
```diff
@@ -39,11 +39,11 @@
...
-      "is_participant": 11078,
-      "info": 11078,
+      "len": 11400,
+      "is_participant": 11410,
+      "info": 11410,
       "validate": 13701,
-      "serialization": 11676,
+      "serialization": 11930,
       "insert": 20449,
       "update_info": 20612
```
gilcu3, line 48: does not seem intuitive, but got a CI failure here after the contract change, which is minimal but does increase contract state size, which apparently influences these numbers

crates/mpc-attestation/src/attestation.rs
```diff
@@ -488,4 +504,87 @@ mod tests {
...
+                rtmr2: [byte; 48],
+            },
+            key_provider_event_digest: [byte; 48],
+        }
+    }
+
+    #[test]
+    fn measurements_constraint_passes_if_in_allowed_list() {
+        // given
+        let measurements = make_measurements(42);
```
barakeinav1, line 523: will always pass since 42 is the answer for everything :-)

pbeza (approved):

barakeinav1:
Verified backward compatibility by querying all live attestations on both testnet (`v1.signer-prod.testnet`) and mainnet (`v1.signer`). Every attestation is `Mock::Valid` — no `WithConstraints` exists on-chain. So the Borsh schema addition is safe.

Worth noting for future reference: if a `WithConstraints` **had** been serialized on-chain, this change would break deserialization (Borsh reads fields sequentially — old 3-field data can't deserialize into the new 4-field struct). The change is safe only because `WithConstraints` has never been submitted to a deployed contract. There's no code-level guard preventing it — it's just that no node software constructs `WithConstraints` outside of tests.

Not blocking, but something to keep in mind for future Borsh schema changes to mock types.

crates/mpc-attestation/src/attestation.rs
```diff
@@ -488,4 +504,87 @@ mod tests {
...
+                rtmr0: [byte; 48],
+                rtmr1: [byte; 48],
+                rtmr2: [byte; 48],
+            },
+            key_provider_event_digest: [byte; 48],
+        }
+    }
+
+    #[test]
+    fn measurements_constraint_passes_if_in_allowed_list() {
```
barakeinav1, line 521: Nice test coverage

barakeinav1 (approved):
thanks!

Issue #2857, opened 2026-04-14
test: port parallel_sign_calls, ckd_verification, and lost_assets (dead node) from pytest to Rust e2e


anodar:
## Background

The MPC project is migrating its test suite from pytest to Rust-based e2e tests as part of the pytest deprecation effort. This task covers porting three test suites: `parallel_sign_calls` (concurrent sign calls via the test contract), `ckd_verification` (cryptographic correctness of CKD responses), and `lost_assets` (dead node cleanup and recovery).

## User Story

As an MPC developer, I need the `parallel_sign_calls`, `ckd_verification`, and `lost_assets` pytest tests migrated to Rust e2e so that we can deprecate the pytest infrastructure and maintain a unified Rust test suite.

## Acceptance Criteria

- [ ] Port `parallel_sign_calls` (robust ECDSA): 3 parallel V2Secp256k1 calls via the test parallel contract
  - Deploy `test-parallel-contract`, fire 3 concurrent robust ECDSA sign calls in one transaction
  - Assert all 3 calls complete successfully
  - Verify signature queue drains and attempt counter increments by at least N
- [ ] Port `parallel_sign_calls` (mixed): 6 parallel calls (2 ECDSA + 2 EdDSA + 2 CKD) via the test parallel contract
  - Deploy `test-parallel-contract`, fire mixed calls in one transaction
  - Assert all 6 calls complete successfully
  - Verify both signature and CKD queues drain
- [ ] Port `ckd_verification` (AppPublicKey): send CKD request, verify response cryptographically
  - ElGamal-decrypt the response: `secret = big_c - big_y * private_key`
  - Verify BLS pairing: `e(H(pk_mpc || app_id), pk_mpc) == e(secret, G2)`
- [ ] Port `ckd_verification` (AppPublicKeyPV): same verification with publicly-verifiable key pair (G1 + G2)
- [ ] Port `lost_assets` (dead node): kill + wipe node, wait for presignature purge, recover
  - Kill node 0 and wipe its DB; verify surviving nodes detect it offline
  - Wait for `mpc_owned_num_presignatures_online >= N` and `mpc_owned_num_presignatures_with_offline_participant == 0` on survivors
  - Drain presignature buffer via sign requests; restart dead node and verify it rejoins
  - Confirm presignature buffer rebuilds on all 3 nodes; final sign request succeeds

## Resources & Additional Notes

Implemented in #TODO. Part of broader effort tracked in #2684.

PR #2859, opened 2026-04-14
test: migrate cdk_verification, cdk_pv_verification, lost_assets, parallel_sign_calls tests from pytests

anodar:
Work towards: https://github.com/near/mpc/issues/2857

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -9,5 +9,13 @@ max-threads = 16
 filter = 'package(mpc-contract) and test(/sandbox/)'
 test-group = 'contract-sandbox'
 
+[test-groups.e2e]
+max-threads = 8
+
+[profile.ci-e2e]
+[[profile.ci-e2e.overrides]]
+filter = 'package(e2e-tests)'
+test-group = 'e2e'
+
 [profile.ci-other]
 default-filter = 'not package(mpc-node) and not package(mpc-contract) and not package(e2e-tests) and not package(contract-history) and not test(=tee_authority::tests::test_upload_quote_for_collateral_with_phala_endpoint) and not (test(/^registry::integration_tests::/) and package(tee-launcher))'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -622,7 +622,7 @@ jobs:
 
       - name: Run E2E tests
         run: |
-          RUST_LOG=info,e2e_tests=debug cargo nextest run --cargo-profile=test-release -p e2e-tests --all-features --locked 2>&1
+          RUST_LOG=info,e2e_tests=debug cargo nextest run --cargo-profile=test-release --profile ci-e2e -p e2e-tests --all-features --locked 2>&1
         env:
           MPC_CONTRACT_WASM: ${{ github.workspace }}/target/wasm32-unknown-unknown/release-contract/mpc_contract.wasm
           MPC_PARALLEL_CONTRACT_WASM: ${{ github.workspace }}/target/wasm32-unknown-unknown/release-contract/test_parallel_contract.wasm
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -577,7 +577,7 @@ jobs:
   mpc-e2e-tests:
     name: "MPC E2E tests"
     needs: mpc-pytest-build
-    runs-on: warp-ubuntu-2404-x64-16x
+    runs-on: warp-ubuntu-2404-x64-32x
     timeout-minutes: 60
     permissions:
       contents: read
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -625,6 +625,7 @@ jobs:
           RUST_LOG=info,e2e_tests=debug cargo nextest run --cargo-profile=test-release -p e2e-tests --all-features --locked 2>&1
         env:
           MPC_CONTRACT_WASM: ${{ github.workspace }}/target/wasm32-unknown-unknown/release-contract/mpc_contract.wasm
+          MPC_PARALLEL_CONTRACT_WASM: ${{ github.workspace }}/target/wasm32-unknown-unknown/release-contract/test_parallel_contract.wasm
 
   ci-extra:
     name: "Extra CI checks"
--- a/crates/e2e-tests/src/blockchain.rs
+++ b/crates/e2e-tests/src/blockchain.rs
@@ -1,9 +1,17 @@
+use std::time::Duration;
+
 use ed25519_dalek::SigningKey;
 use near_kit::FinalExecutionOutcome;
 use near_mpc_contract_interface::types::ProtocolContractState;
 use serde::de::DeserializeOwned;
 
 const MAX_GAS: near_kit::Gas = near_kit::Gas::from_tgas(300);
+
+/// Retries for `AccessKeyNotFound` errors that occur when a transaction's
+/// `view_access_key` query (using `Finality::Final`) races against account
+/// creation that was only waited for with `ExecutedOptimistic`.
+const ACCESS_KEY_RETRY_ATTEMPTS: u32 = 20;
+const ACCESS_KEY_RETRY_DELAY: Duration = Duration::from_millis(500);
 
 /// RPC client for any NEAR network (sandbox or testnet).
 ///
@@ -118,13 +126,35 @@ impl DeployedContract {
         method: &str,
         args: serde_json::Value,
     ) -> anyhow::Result<FinalExecutionOutcome> {
-        self.client
-            .call(&self.contract_id, method)
-            .args(args)
-            .gas(MAX_GAS)
-            .send()
-            .await
-            .map_err(|e| anyhow::anyhow!("contract call `{method}` failed: {e}"))
+        // near-kit's send() queries view_access_key with Finality::Final, but
+        // create_account_and_deploy only waits for ExecutedOptimistic. The key
+        // may not appear in the finalized state yet — retry until it does.
+        for attempt in 0..ACCESS_KEY_RETRY_ATTEMPTS {
+            match self
+                .client
+                .call(&self.contract_id, method)
+                .args(args.clone())
+                .gas(MAX_GAS)
+                .send()
+                .await
+            {
+                Ok(outcome) => return Ok(outcome),
+                Err(near_kit::Error::Rpc(rpc_err))
+                    if matches!(*rpc_err, near_kit::RpcError::AccessKeyNotFound { .. }) =>
+                {
+                    tracing::debug!(
+                        attempt,
+                        "Access key not found querying finalized state, retrying after {:?}",
+                        ACCESS_KEY_RETRY_DELAY,
+                    );
+                    tokio::time::sleep(ACCESS_KEY_RETRY_DELAY).await;
+                }
+                Err(e) => return Err(anyhow::anyhow!("contract call `{method}` failed: {e}")),
+            }
+        }
+        Err(anyhow::anyhow!(
+            "contract call `{method}` failed: access key not found after {ACCESS_KEY_RETRY_ATTEMPTS} retries"
+        ))
     }
 
     pub async fn call_from(
--- a/crates/e2e-tests/src/blockchain.rs
+++ b/crates/e2e-tests/src/blockchain.rs
@@ -15,7 +15,7 @@ const ACCESS_KEY_RETRY_DELAY: Duration = Duration::from_millis(500);
 
 /// RPC client for any NEAR network (sandbox or testnet).
 ///
-/// Wraps a `near_kit::Near` client signed as the root/funder account.
+/// Wraps `near_kit::Near` client signed as the root/funder account.
 /// Whether the RPC URL points to a local Docker sandbox or NEAR testnet,
 /// the code path is identical.
 pub struct NearBlockchain {
--- a/crates/e2e-tests/src/blockchain.rs
+++ b/crates/e2e-tests/src/blockchain.rs
@@ -1,17 +1,9 @@
-use std::time::Duration;
-
 use ed25519_dalek::SigningKey;
 use near_kit::FinalExecutionOutcome;
 use near_mpc_contract_interface::types::ProtocolContractState;
 use serde::de::DeserializeOwned;
 
 const MAX_GAS: near_kit::Gas = near_kit::Gas::from_tgas(300);
-
-/// Retries for `AccessKeyNotFound` errors that occur when a transaction's
-/// `view_access_key` query (using `Finality::Final`) races against account
-/// creation that was only waited for with `ExecutedOptimistic`.
-const ACCESS_KEY_RETRY_ATTEMPTS: u32 = 20;
-const ACCESS_KEY_RETRY_DELAY: Duration = Duration::from_millis(500);
 
 /// RPC client for any NEAR network (sandbox or testnet).
 ///
@@ -62,7 +54,8 @@ impl NearBlockchain {
                 .add_full_access_key(near_kit::PublicKey::Ed25519(key.verifying_key().to_bytes()));
         }
 
-        tx.send()
+        tx.wait_until(near_kit::TxExecutionStatus::Final)
+            .send()
             .await
             .map_err(|e| anyhow::anyhow!("failed to create account {name}: {e}"))?;
         Ok(())
@@ -81,6 +74,7 @@ impl NearBlockchain {
             .transfer(near_kit::NearToken::from_near(balance_near))
             .add_full_access_key(near_kit::PublicKey::Ed25519(key.verifying_key().to_bytes()))
             .deploy(wasm.to_vec())
+            .wait_until(near_kit::TxExecutionStatus::Final)
             .send()
             .await
             .map_err(|e| anyhow::anyhow!("failed to create account and deploy to {name}: {e}"))?;
@@ -126,35 +120,13 @@ impl DeployedContract {
         method: &str,
         args: serde_json::Value,
     ) -> anyhow::Result<FinalExecutionOutcome> {
-        // near-kit's send() queries view_access_key with Finality::Final, but
-        // create_account_and_deploy only waits for ExecutedOptimistic. The key
-        // may not appear in the finalized state yet — retry until it does.
-        for attempt in 0..ACCESS_KEY_RETRY_ATTEMPTS {
-            match self
-                .client
-                .call(&self.contract_id, method)
-                .args(args.clone())
-                .gas(MAX_GAS)
-                .send()
-                .await
-            {
-                Ok(outcome) => return Ok(outcome),
-                Err(near_kit::Error::Rpc(rpc_err))
-                    if matches!(*rpc_err, near_kit::RpcError::AccessKeyNotFound { .. }) =>
-                {
-                    tracing::debug!(
-                        attempt,
-                        "Access key not found querying finalized state, retrying after {:?}",
-                        ACCESS_KEY_RETRY_DELAY,
-                    );
-                    tokio::time::sleep(ACCESS_KEY_RETRY_DELAY).await;
-                }
-                Err(e) => return Err(anyhow::anyhow!("contract call `{method}` failed: {e}")),
-            }
-        }
-        Err(anyhow::anyhow!(
-            "contract call `{method}` failed: access key not found after {ACCESS_KEY_RETRY_ATTEMPTS} retries"
-        ))
+        self.client
+            .call(&self.contract_id, method)
+            .args(args)
+            .gas(MAX_GAS)
+            .send()
+            .await
+            .map_err(|e| anyhow::anyhow!("contract call `{method}` failed: {e}"))
     }
 
     pub async fn call_from(
--- a/crates/e2e-tests/src/cluster.rs
+++ b/crates/e2e-tests/src/cluster.rs
@@ -29,6 +29,9 @@ pub const DEFAULT_TRIPLES_TO_BUFFER: usize = 20;
 pub const DEFAULT_PRESIGNATURES_TO_BUFFER: usize = 10;
 pub const CLUSTER_WAIT_TIMEOUT: Duration = Duration::from_secs(120);
 const SIGN_GAS: near_kit::Gas = near_kit::Gas::from_tgas(15);
+// AppPublicKeyPV does an on-chain bls12381_pairing_check (2 pairs) before yielding,
+// which costs significantly more than a plain CKD or sign request.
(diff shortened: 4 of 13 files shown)


crates/e2e-tests/tests/lost_assets.rs
```diff
@@ -0,0 +1,117 @@
...
+        .domains
+        .domains
+        .iter()
+        .find(|d| matches!(d.purpose, Some(DomainPurpose::Sign)))
+    {
+        for _ in 0..PRESIGNATURES_TO_BUFFER {
+            let payload = match domain.scheme {
+                SignatureScheme::Secp256k1 | SignatureScheme::V2Secp256k1 => {
+                    common::generate_ecdsa_payload(&mut rng)
+                }
```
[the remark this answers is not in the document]
anodar: self-high-five

crates/e2e-tests/src/blockchain.rs
```diff
@@ -118,13 +126,35 @@ impl DeployedContract {
...
+                    tracing::debug!(
+                        attempt,
+                        "Access key not found querying finalized state, retrying after {:?}",
+                        ACCESS_KEY_RETRY_DELAY,
+                    );
+                    tokio::time::sleep(ACCESS_KEY_RETRY_DELAY).await;
+                }
+                Err(e) => return Err(anyhow::anyhow!("contract call `{method}` failed: {e}")),
+            }
+        }
```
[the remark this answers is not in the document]
anodar: > so this sleep is wasted.

0.5 second wasted in couple of minutes, not a big deal.

crates/e2e-tests/tests/parallel_sign_calls.rs
```diff
@@ -0,0 +1,172 @@
...
+    let sig_attempts_delta = common::sum_metric(&cluster, metrics::SIGNATURES_QUEUE_ATTEMPTS).await
+        - initial_sig_attempts;
+    let ckd_attempts_delta =
+        common::sum_metric(&cluster, metrics::CKDS_QUEUE_ATTEMPTS).await - initial_ckd_attempts;
+    let total_delta = sig_attempts_delta + ckd_attempts_delta;
+
+    assert_eq!(
+        total_delta, N as i64,
+        "expected exactly {N} total leader attempts across all nodes and queues, got {total_delta}"
+    );
```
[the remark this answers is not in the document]
anodar: Addressed.

crates/e2e-tests/tests/lost_assets.rs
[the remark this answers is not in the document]
anodar: Done.

.config/nextest.toml
```diff
@@ -9,5 +9,13 @@ max-threads = 16
 filter = 'package(mpc-contract) and test(/sandbox/)'
 test-group = 'contract-sandbox'
 
+[test-groups.e2e]
+max-threads = 8
```
netrome, line 13: Why do we need this configuration? Does this impact the tests significantly? I.e. do they break with more threads?

.github/workflows/ci.yml
```diff
@@ -577,7 +577,7 @@ jobs:
   mpc-e2e-tests:
     name: "MPC E2E tests"
     needs: mpc-pytest-build
-    runs-on: warp-ubuntu-2404-x64-16x
+    runs-on: warp-ubuntu-2404-x64-32x
```
netrome, line 580: Why do we need a beefier runner?

crates/e2e-tests/src/blockchain.rs
```diff
@@ -1,13 +1,21 @@
...
 use near_mpc_contract_interface::types::ProtocolContractState;
 use serde::de::DeserializeOwned;
 
 const MAX_GAS: near_kit::Gas = near_kit::Gas::from_tgas(300);
 
+/// Retries for `AccessKeyNotFound` errors that occur when a transaction's
+/// `view_access_key` query (using `Finality::Final`) races against account
+/// creation that was only waited for with `ExecutedOptimistic`.
+const ACCESS_KEY_RETRY_ATTEMPTS: u32 = 20;
+const ACCESS_KEY_RETRY_DELAY: Duration = Duration::from_millis(500);
```
netrome: Why do we race? Can't we just await account creation to be completed instead of having a retry loop?

crates/e2e-tests/src/blockchain.rs
```diff
@@ -118,13 +126,35 @@ impl DeployedContract {
...
-            .call(&self.contract_id, method)
-            .args(args)
-            .gas(MAX_GAS)
-            .send()
-            .await
-            .map_err(|e| anyhow::anyhow!("contract call `{method}` failed: {e}"))
+        // near-kit's send() queries view_access_key with Finality::Final, but
+        // create_account_and_deploy only waits for ExecutedOptimistic. The key
+        // may not appear in the finalized state yet — retry until it does.
+        for attempt in 0..ACCESS_KEY_RETRY_ATTEMPTS {
```
netrome: It feels wrong to have the retry logic within the `call` function. The explanation above seems specific for certain calls, shouldn't this retry exist on a higher level? Or not at all?

crates/e2e-tests/src/blockchain.rs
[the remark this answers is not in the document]
netrome: I wasted much more time than that reading Claude's comment 😂

crates/e2e-tests/tests/lost_assets.rs
```diff
@@ -0,0 +1,138 @@
...
+        &cluster,
+        &alive,
+        metrics::OWNED_PRESIGNATURES_ONLINE,
+        |v| v >= PRESIGNATURES_TO_BUFFER as i64,
+        CLUSTER_WAIT_TIMEOUT,
+    )
+    .await;
+
+    // then — surviving nodes can still process sign requests.
+    let mut rng = rand::rngs::StdRng::seed_from_u64(0);
```
netrome: Nit: Typically I'd expect the `rng` to be declared in the given section making it easy to see that the test uses randomness. But this is also okay since it's closer to when it's used.

crates/e2e-tests/tests/parallel_sign_calls.rs
```diff
@@ -0,0 +1,172 @@
...
+
+use e2e_tests::{CLUSTER_WAIT_TIMEOUT, metrics};
+use near_mpc_contract_interface::types::{Curve, DomainConfig, DomainId, DomainPurpose};
+use serde_json::json;
+
+/// 3 parallel robust ECDSA (V2Secp256k1) sign calls via the test parallel contract.
+/// Verifies all calls succeed and the signature queue processes all requests.
+#[tokio::test]
+#[expect(non_snake_case)]
+async fn parallel_sign_calls__robust_ecdsa() {
```
netrome: If we use the double underscore I think we should follow the <sut>__should_<test_assertion> naming convention. It forces us to reflect, what is the system under test? And what is the behavior we're testing here?

Perhaps `mpc_cluster__should_successfully_process_parallel_robust_ecdsa_requests`?

crates/e2e-tests/tests/parallel_sign_calls.rs
```diff
@@ -0,0 +1,172 @@
...
+        attempts_delta >= N as i64,
+        "expected >= {N} total leader attempts across all nodes, got {attempts_delta}"
+    );
+}
+
+/// 6 parallel calls (2 ECDSA + 2 EdDSA + 2 CKD) via the test parallel contract.
+/// Verifies all calls succeed and both the signature and CKD queues process all requests.
+#[tokio::test]
+#[expect(non_snake_case)]
+async fn parallel_sign_calls__mixed() {
```
netrome: Same here. Perhaps something like `mpc_cluster__should_successfully_process_mixed_parallel_requests`

netrome (dismissed):
Some questions, thoughts and considerations but nothing blocking.

crates/e2e-tests/tests/parallel_sign_calls.rs
anodar: Same as above, done.

crates/e2e-tests/tests/parallel_sign_calls.rs
anodar: I initially did like that, but then realized all the e2e tests start with `mpc_cluster__should ....` in all files :D So figured it doesn't add any value if all tests start with that.

Reverted to that though.

crates/e2e-tests/src/blockchain.rs
anodar: Dropped it.

crates/e2e-tests/src/blockchain.rs
anodar: Looks like near_kit provides `wait_until(near_kit::TxExecutionStatus::Final)` so changed to that.

.github/workflows/ci.yml
anodar: Currently pytests are running concurrently on 10 different runners, some of them e.g. Robust ecdsa on `warp-ubuntu-2404-x64-16x`,  key-event-single-domain: `warp-ubuntu-2404-x64-8x`,  and rest on 2x. 

This runner will be running all of the tests concurrently. I think it makes sense to have beefier machine.

.config/nextest.toml
anodar: Yes this does impact tests. Because when we run tests concurrently all at once, it really uses up 100% CPU / RAM and waiting for state times out (e.g. [0]). For reference on my machine M4 MAX 36GB, without this it freezes the system, mouse stutters etc... 


[0] https://github.com/near/mpc/actions/runs/24402333715/job/71276487173

crates/e2e-tests/tests/lost_assets.rs
anodar: Done.

crates/e2e-tests/src/blockchain.rs
netrome: Thanks :pray:

.config/nextest.toml
netrome: > without this it freezes the system, mouse stutters etc...

Well that's how you know it's working... 😂

netrome (approved):
Thanks for updating 🙏

crates/e2e-tests/tests/parallel_sign_calls.rs
netrome: Thanks. I do think it looks nice when running a lot of tests, and when searching across the code base.

.config/nextest.toml
gilcu3, line 13: why 8 and not 16? seems a bit wasteful given we will be using the beefier 32x machine

crates/e2e-tests/src/blockchain.rs
```diff
@@ -54,7 +54,8 @@ impl NearBlockchain {
                 .add_full_access_key(near_kit::PublicKey::Ed25519(key.verifying_key().to_bytes()));
         }
 
-        tx.send()
+        tx.wait_until(near_kit::TxExecutionStatus::Final)
+            .send()
```
gilcu3, line 58: I thought this was the default. What was exactly happening before?

crates/e2e-tests/src/blockchain.rs
```diff
@@ -73,6 +74,7 @@ impl NearBlockchain {
             .transfer(near_kit::NearToken::from_near(balance_near))
             .add_full_access_key(near_kit::PublicKey::Ed25519(key.verifying_key().to_bytes()))
```
gilcu3, line 75: can we make all these conversion functions part of a single module inside the crate, so that we can decide where to put them at a later stage?

crates/e2e-tests/src/cluster.rs
```diff
@@ -29,6 +29,9 @@ pub const DEFAULT_TRIPLES_TO_BUFFER: usize = 20;
 pub const DEFAULT_PRESIGNATURES_TO_BUFFER: usize = 10;
 pub const CLUSTER_WAIT_TIMEOUT: Duration = Duration::from_secs(120);
 const SIGN_GAS: near_kit::Gas = near_kit::Gas::from_tgas(15);
+// AppPublicKeyPV does an on-chain bls12381_pairing_check (2 pairs) before yielding,
+// which costs significantly more than a plain CKD or sign request.
+pub const CKD_PV_GAS: near_kit::Gas = near_kit::Gas::from_tgas(100);
```
gilcu3, line 34: I wonder if we have these constants somewhere else already. I image the sandbox tests would need the same

crates/e2e-tests/src/cluster.rs
```diff
@@ -632,6 +635,19 @@ impl MpcCluster {
         &self,
         domain_id: DomainId,
         app_public_key: CKDAppPublicKey,
+    ) -> anyhow::Result<near_kit::FinalExecutionOutcome> {
+        self.send_ckd_request_with_gas(domain_id, app_public_key, SIGN_GAS)
+            .await
+    }
```
gilcu3, line 641: I see a bit of an asymmetry here. The gas needed can be determined by the CKDAppPublicKey variant, so this function could handle both cases gracefully, and be merged with the helper function below

crates/e2e-tests/tests/common.rs
```diff
@@ -134,6 +183,32 @@ pub fn load_contract_wasm() -> Vec<u8> {
...
+        return std::fs::read(&default_path).unwrap_or_else(|e| {
+            panic!(
+                "Failed to read parallel contract WASM at {}: {e}",
+                default_path.display()
+            )
+        });
+    }
+    tracing::info!("pre-built parallel contract WASM not found — building");
+    test_utils::contract_build::ContractBuilder::new("crates/test-parallel-contract/Cargo.toml")
+        .build()
```
gilcu3, line 209: this is nice, but might not be enough. What if the contract on disk is stale? For pytests we had two options:
- always build the contract (for test-parallel that is kind of very fast anyway)
- add a parameter so that the user can decide to skip the build

both apply also for the mpc-contract and mpc-node binaries. No need to handle them here, but would be good to get aligned on this. Today I just opened #2875 when I noticed that I cannot really run the e2e tests locally for example (I had a stale mpc-node binary)

crates/e2e-tests/tests/lost_assets.rs
```diff
@@ -0,0 +1,138 @@
...
+    .await;
+
+    common::wait_metric_on_nodes(
+        &cluster,
+        &alive,
+        metrics::OWNED_PRESIGNATURES_ONLINE,
+        |v| v >= PRESIGNATURES_TO_BUFFER as i64,
+        CLUSTER_WAIT_TIMEOUT,
+    )
+    .await;
```
gilcu3, line 78: is this aligned with the previous pytest logic? Seems quite hard to grasp, and depends on node internals. I would at least try to simplify the comment explaining what happens, because `cold_ready` might not ring a bell for most of the team

crates/e2e-tests/tests/parallel_sign_calls.rs
gilcu3, line 1: we made these tests into separate functions only because robust-ecdsa was resulting in flakiness. Would be better to simply have a single function to test them all. In case that cannot be done yet, would be good to add a comment as to why.

crates/e2e-tests/tests/ckd_verification.rs
```diff
@@ -0,0 +1,181 @@
...
+    let secret: G1Affine = (big_c - big_y * private_key).into();
+    let mpc_pk_affine: G2Affine = mpc_pk.into();
+
+    let app_id = derive_app_id(account_id, path);
+    let hash_input = [mpc_public_key.as_slice(), &app_id].concat();
+    let hash_point: G1Affine =
+        G1Projective::hash_to_curve(&hash_input, NEAR_CKD_DOMAIN, &[]).into();
+
+    blstrs::pairing(&hash_point, &mpc_pk_affine) == blstrs::pairing(&secret, &G2Affine::generator())
+}
```
gilcu3, line 49: as we had system tests in python all these code was not possible to reuse from the rust implementation. Now that we can, we should certainly do so. If it is not readily available, we can create and issue+TODO to do that as a follow up.

gilcu3 (approved):
LGTM

Left a bunch of comments and suggestions (some might require non-negligible work or follow-ups)

.config/nextest.toml
anodar: max-threads is number of concurrent tests, not actual OS threads. Each test spawns multiple mpc node + near sandbox. We don't even have 16 tests, that would be equivalent of not restricting at all. I won't be surprised if we have to restrict it further or/and fan out into 2 runners.

It's clear we need to restrict it, but I don't have data that would back up the exact number. I've created issue for logging some telemetry to get a better idea on it https://github.com/near/mpc/issues/2883

crates/e2e-tests/src/blockchain.rs
anodar: Default is: `TxExecutionStatus::ExecutedOptimistic` https://github.com/r-near/near-kit-rs/blob/a611d5e1ba6cc5edfe0219908c391ed51c5c8ff3/crates/near-kit/src/types/block_reference.rs#L347

crates/e2e-tests/src/blockchain.rs
anodar: Will follow up.

crates/e2e-tests/src/cluster.rs
anodar: Considering you're suggesting we merge `send_ckd_request` and `send_ckd_request_with_gas` and pass amount that works for both, using `CKD_PV_GAS` constant in signing requests doesn't make sense. I might as well just drop both SIGN and CDK_PV_GAS and use already existing MAX_GAS.

crates/e2e-tests/src/cluster.rs
anodar: I might as well just use MAX_GAS and drop both of those constants.

crates/e2e-tests/tests/ckd_verification.rs
anodar: It is in `ckd-example-cli` https://github.com/near/mpc/blob/afb4a832dec75eeef3199d7780c2659bd381f617/crates/ckd-example-cli/src/ckd.rs#L100, we probably don't want to depend on that rather factor it out from there in separate library crate. Will add TODO for it in followup.

crates/e2e-tests/tests/common.rs
anodar: Thanks for creating an issue for it. I've assigned it to my self and added context about contracts too. Will follow up.

crates/e2e-tests/tests/lost_assets.rs
anodar: I previously had it aligned with the pytest logic where after killing the node we just check that num_offline_presignatures is 0, and presignatures_online is 8.

There's a race condition where this condition will be met immediately after killing the node (even though we assert assert_num_live_connections, it doesn't mean it updated offline presignatures yet).

Here's the feedback I got from Claude review initially:
```
The test races ahead before offline detection runs: both wait conditions (ONLINE >= 8 and OFFLINE == 0) are trivially satisfied immediately after killing node 0, while all 8 original 3-party presignatures are still classified as ONLINE and OFFLINE is still 0. The test then fires 8 sign requests against stale 3-party presignatures that require the dead node, causing spurious test failures. The fix is to first wait for OFFLINE > 0 (confirming detection ran), then wait for OFFLINE == 0 and ONLINE >= 8.
```


I think comment above is overly detailed, all it has to say is that we need to verify that online presignatures do decrease first before it goes back up to 8.  That should make Rust test more robust that pytest counterpart. Will reword it in followup.

crates/e2e-tests/tests/parallel_sign_calls.rs
anodar: It's the same reason why I didn't merge request lifecycle tests and instead just put in the same file: cluster has different setup (although maybe that's not required ?!), and allows concurrent execution of tests.

anodar:
Thanks for the reviews! Will follow up on open comments.

.config/nextest.toml
gilcu3: >We don't even have 16 tests,

good point, but we do have more than 8 no?

anodar: Yes we have 14 (2 of them are unit tests that complete in 9s), so really we have 12.

crates/e2e-tests/src/cluster.rs
gilcu3: better not using max_gas, so that we know if there is an increase. My proposal is that send_ckd_request can determine how much gas it needs, because the type of ckd request (with public verifiability or not) is determined by the parameter type

crates/e2e-tests/src/cluster.rs
gilcu3: see https://github.com/near/mpc/pull/2859#discussion_r3085343046

crates/e2e-tests/tests/parallel_sign_calls.rs
gilcu3: yeah hopefully the different cluster setup is not needed anymore!

crates/e2e-tests/tests/ckd_verification.rs
gilcu3: we have separate creates (the ts crate) which should offer most of the functionality

.config/nextest.toml
anodar: As discussed, this should probably be dropped even lower. Will investigate in https://github.com/near/mpc/issues/2898

crates/e2e-tests/src/blockchain.rs
anodar: Moved to conversations.rs in: https://github.com/near/mpc/pull/2907

crates/e2e-tests/src/cluster.rs
anodar: Done in: https://github.com/near/mpc/pull/2907

crates/e2e-tests/src/cluster.rs
anodar: Done in: https://github.com/near/mpc/pull/2907

crates/e2e-tests/tests/lost_assets.rs
anodar: Done in: https://github.com/near/mpc/pull/2907

crates/e2e-tests/tests/parallel_sign_calls.rs
anodar: Merged those two in: https://github.com/near/mpc/pull/2907

crates/e2e-tests/tests/ckd_verification.rs
anodar: Reworded in: https://github.com/near/mpc/pull/2907

PR #2861, opened 2026-04-14
fix: remove race condition in `take_unowned`

### LINKED ISSUE #2856: Remove potential race condition in `take_unowned`
### Background

Currently, `take_unowned` performs two separate database operations to get and delete presignatures. There's no synchronization which means the same asset could be popped to two concurrent threads.

### User Story

_No response_

### Acceptance Criteria

`take_unowned` guarantees to only read the same asset at most once.

_No response_

netrome:
closes #2856

Unfortunately untested due to the current lack of isolation of DB operations. Will be fixed in https://github.com/near/mpc/issues/2862


crates/node/src/assets.rs
```diff
@@ -1358,4 +1384,58 @@ mod tests {
             }
         }
     }
+
+    #[test]
+    fn take_unowned__should_allow_exactly_one_concurrent_caller_to_succeed() {
```
netrome: Note: I realize this test is meaningless since it would most likely not hit the race condition in practice even without the fix in place.

We should introduce a proper abstraction over the database to make this race condition testable. I'll add that as a follow-up and delete this test.

netrome: Created https://github.com/near/mpc/issues/2862

DSharifi (approved):

pbeza (approved):

PR #2863, opened 2026-04-14
build: compute manifest digest locally without requiring --push

### LINKED ISSUE #2843: [build] Compute manifest digest locally without requiring --push
### Background

Follow-up from PR #2785 review comment by @gilcu3. Since we now use manifest digests for voting (#2785), operators need to verify the manifest digest when building locally. Currently `deployment/build-images.sh` only computes and prints the manifest digest when `--push` is used (because the skopeo step runs during the push).

### User Story

As an MPC operator, I want to build the Docker image locally and see the manifest digest so that I can verify it matches what I'm asked to vote for, without needing Docker Hub credentials.

### Acceptance Criteria

- [ ] `./deployment/build-images.sh --node` (without `--push`) outputs the manifest digest
- [ ] `--push` still works as before (push + digest)
- [ ] `skopeo` is required for all builds (not just `--push`)
- [ ] Update the operator guide example to not require `--push`

- PR #2785 (simplify launcher image validation — where the review comment originated)
- PR #2780 (skopeo for node push — introduced `skopeo_push` helper)
- Current workaround: operators look up the manifest digest on Docker Hub separately
- The `skopeo_push` function in `build-images.sh` already does the compress-to-dir step — it just needs to be split so the local compress always runs

**Suggested approach:** Always run `skopeo copy --dest-compress docker-daemon:... dir:$td` to compute the manifest digest locally. Only run the second `skopeo copy --preserve-digests dir:$td docker://...` when `--push` is set.

barakeinav1:
### Summary

Always compute the manifest digest after building a Docker image, even without `--push`. Operators can now verify the manifest digest locally without Docker Hub credentials.

### Changes

- Add `skopeo_compress()` function that compresses the image to a temp dir and computes the manifest digest
- Call it after every image build (launcher, rust-launcher, node, node-gcp)
- The `--push` step reuses the existing temp dir (no double compress)
- `skopeo` is now always required (was only required with `--push`)
- Manifest digest is always printed in the output
- Updated operator guide to show manifest digest in build output, removed TODO(#2843)

### Before

```bash
$ ./deployment/build-images.sh --node
node docker image hash: sha256:0e48003c...
# No manifest digest without --push
```

### After

```bash
$ ./deployment/build-images.sh --node
node docker image hash: sha256:0e48003c...
node manifest digest: sha256:331cfec9...
```

Closes #2843

### Test plan

- [x] `./deployment/build-images.sh --node` outputs manifest digest
- [x] `./deployment/build-images.sh --node --push` still pushes correctly
- [x] CI docker build jobs pass

### DIFF
--- a/deployment/build-images.sh
+++ b/deployment/build-images.sh
@@ -70,9 +70,7 @@ if $USE_NODE || $USE_RUST_LAUNCHER; then
     require_cmds repro-env podman
 fi
 
-if $USE_PUSH; then
-    require_cmds skopeo
-fi
+require_cmds skopeo
 
 if ! docker buildx &>/dev/null; then
   die "Please install docker-buildx"
@@ -128,9 +126,26 @@ get_image_hash() {
     docker inspect $image_name | jq -r .[0].Id
 }
 
+# Compress a locally built image via skopeo and compute its manifest digest.
+# Sets two global variables: <prefix>_manifest_digest and <prefix>_skopeo_dir
+# Usage: skopeo_compress <image_name> <variable_prefix>
+skopeo_compress() {
+    local image_name="$1"
+    local prefix="$2"
+    local td
+    td=$(mktemp -d)
+    # Compress the built image to a local directory, which implicitly computes
+    # the manifest digest in $td/manifest.json
+    skopeo copy --all --dest-compress "docker-daemon:${image_name}:latest" "dir:$td"
+    local digest="sha256:$(sha256sum "$td/manifest.json" | cut -d' ' -f1)"
+    eval "${prefix}_manifest_digest=\"$digest\""
+    eval "${prefix}_skopeo_dir=\"$td\""
+}
+
 if $USE_LAUNCHER; then
     build_reproducible_image $LAUNCHER_IMAGE_NAME $DOCKERFILE_LAUNCHER
     launcher_image_hash=$(get_image_hash $LAUNCHER_IMAGE_NAME)
+    skopeo_compress "$LAUNCHER_IMAGE_NAME" launcher
 fi
 
 if $USE_RUST_LAUNCHER; then
@@ -139,6 +154,7 @@ if $USE_RUST_LAUNCHER; then
 
     build_reproducible_image $RUST_LAUNCHER_IMAGE_NAME $DOCKERFILE_RUST_LAUNCHER
     rust_launcher_image_hash=$(get_image_hash $RUST_LAUNCHER_IMAGE_NAME)
+    skopeo_compress "$RUST_LAUNCHER_IMAGE_NAME" rust_launcher
 fi
 
 if $USE_NODE || $USE_NODE_GCP; then
@@ -149,11 +165,13 @@ fi
 if $USE_NODE; then
     build_reproducible_image $NODE_IMAGE_NAME $DOCKERFILE_NODE
     node_image_hash=$(get_image_hash $NODE_IMAGE_NAME)
+    skopeo_compress "$NODE_IMAGE_NAME" node
 fi
 
 if $USE_NODE_GCP; then
     build_reproducible_image $NODE_GCP_IMAGE_NAME $DOCKERFILE_NODE_GCP
     node_gcp_image_hash=$(get_image_hash $NODE_GCP_IMAGE_NAME)
+    skopeo_compress "$NODE_GCP_IMAGE_NAME" node_gcp
 fi
 
 if $USE_PUSH; then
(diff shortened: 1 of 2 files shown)


deployment/build-images.sh
```diff
@@ -70,9 +70,7 @@ if $USE_NODE || $USE_RUST_LAUNCHER; then
     require_cmds repro-env podman
 fi
 
-if $USE_PUSH; then
-    require_cmds skopeo
-fi
+require_cmds skopeo
 
 if ! docker buildx &>/dev/null; then
```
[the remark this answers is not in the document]
barakeinav1: Fixed in 821055e5 — updated header to list skopeo and all image flags.

deployment/build-images.sh
```diff
@@ -128,9 +126,26 @@ get_image_hash() {
...
+# Usage: skopeo_compress <image_name> <variable_prefix>
+skopeo_compress() {
+    local image_name="$1"
+    local prefix="$2"
+    local td
+    td=$(mktemp -d)
+    # Compress the built image to a local directory, which implicitly computes
+    # the manifest digest in $td/manifest.json
+    skopeo copy --all --dest-compress "docker-daemon:${image_name}:latest" "dir:$td"
+    local digest="sha256:$(sha256sum "$td/manifest.json" | cut -d' ' -f1)"
```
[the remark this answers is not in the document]
barakeinav1: Intentionally not cleaning up. Per team feedback on PR #2780, temp dir cleanup is unnecessary in CI (ephemeral runners) and /tmp is cleaned periodically on dev machines.

deployment/build-images.sh
```diff
@@ -128,9 +126,26 @@ get_image_hash() {
...
+    eval "${prefix}_manifest_digest=\"$digest\""
+    eval "${prefix}_skopeo_dir=\"$td\""
```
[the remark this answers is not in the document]
barakeinav1: Fixed in 821055e5 — replaced `eval` with `printf -v`.

docs/running-an-mpc-node-in-tdx-external-guide.md
```diff
@@ -1208,10 +1208,10 @@ commit hash: 828f816be36aed6f0d2438e0131b3e9d7d0931ad
 SOURCE_DATE_EPOCH used: 0
 node binary hash: 86c8f7d8913d6fe37a6992bba165d15a3a1d88fbf6cdff605e4827d5183721bc
 node docker image hash: sha256:0e48003c0ac6ec01e79ce47aa094379e7a8fac428512dfeb18d49d558e100a53
+node manifest digest: sha256:331cfec941671ac343c52847e255eb36a280da65535d2a1e4d002c4c64686e19
 ```
 
-<!-- TODO(#2843): build script should output manifest digest without --push -->
-The `node docker image hash` is the config digest of the locally built image. To get the manifest digest (what you vote for), look it up on DockerHub for the same image tag. The launcher pulls the image directly by manifest digest — Docker verifies the content matches during the pull.
+The `node manifest digest` is what you vote for. The launcher pulls the image directly by this digest — Docker verifies the content matches during the pull.
```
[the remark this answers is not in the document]
barakeinav1: Fixed in 821055e5 — added note to strip `sha256:` prefix when voting.

deployment/build-images.sh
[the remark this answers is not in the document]
netrome: Yes agree

deployment/build-images.sh
```diff
@@ -139,6 +155,7 @@ if $USE_RUST_LAUNCHER; then
 
     build_reproducible_image $RUST_LAUNCHER_IMAGE_NAME $DOCKERFILE_RUST_LAUNCHER
     rust_launcher_image_hash=$(get_image_hash $RUST_LAUNCHER_IMAGE_NAME)
+    skopeo_compress "$RUST_LAUNCHER_IMAGE_NAME" rust_launcher
```
netrome: shouldn't we assign the `launcher_manifest_digest` here somewhere? Or where do we do this?

netrome (commented):
Had a quick look, not following where we assign the manifest digest variables - would appreciate some pointers. I'll get back to this in 30 minutes.

deployment/build-images.sh
barakeinav1: The assignment happens inside `skopeo_compress` via `printf -v`. When called with prefix `rust_launcher`, it dynamically creates two global variables:

```bash
printf -v "rust_launcher_manifest_digest" '%s' "$digest"
printf -v "rust_launcher_skopeo_dir" '%s' "$td"
```

So after `skopeo_compress "$RUST_LAUNCHER_IMAGE_NAME" rust_launcher`, the variable `$rust_launcher_manifest_digest` is available. Same pattern for `node`, `node_gcp`, `launcher`. It's bash's indirect variable assignment — avoids repeating the compress logic for each image.

netrome: Ah I see, thanks for highlighting :pray:

netrome (dismissed):
Thanks!

deployment/build-images.sh
```diff
@@ -70,9 +71,7 @@ if $USE_NODE || $USE_RUST_LAUNCHER; then
     require_cmds repro-env podman
 fi
 
-if $USE_PUSH; then
-    require_cmds skopeo
-fi
+require_cmds skopeo
```
gilcu3: we have another line for required_cmds (line 68)

deployment/build-images.sh
```diff
@@ -128,9 +127,26 @@ get_image_hash() {
...
+    local image_name="$1"
+    local prefix="$2"
+    local td
+    td=$(mktemp -d)
+    # Compress the built image to a local directory, which implicitly computes
+    # the manifest digest in $td/manifest.json
+    skopeo copy --all --dest-compress "docker-daemon:${image_name}:latest" "dir:$td"
+    local digest="sha256:$(sha256sum "$td/manifest.json" | cut -d' ' -f1)"
+    printf -v "${prefix}_manifest_digest" '%s' "$digest"
+    printf -v "${prefix}_skopeo_dir" '%s' "$td"
```
gilcu3: to avoid this bash obscure magic, we could simply return the resulting folder, then compute the digest after. Alternative is to create the folder before calling the function and passing it as parameter. Third option, do the optional push inside this function. Anything but bash black magic 😟

gilcu3 (dismissed):
Thank you!

Approved as it is correct, but would really prefer to avoid the bash magic and use easy to read syntax. Proposed a few alternatives for that

deployment/build-images.sh
barakeinav1: Simplified in a3372293 — `skopeo_compress` now just echoes the temp dir path. Caller computes the digest with a `manifest_digest_from_dir` helper. No more `printf -v` indirection.

deployment/build-images.sh
barakeinav1: Fixed in a3372293 — merged `skopeo` into the main `require_cmds` line.

gilcu3 (approved):

netrome (approved):

PR #2869, opened 2026-04-14
chore: reduce contract WASM size and add CI size check

### LINKED ISSUE #2868: Switch contract to `opt-level = "z"` and add WASM size CI guard
### Background

The contract WASM binary is approaching NEAR's `max_transaction_size` limit of 1,572,864 bytes ([reduced from 4MB in protocol v69](https://github.com/near/nearcore/blob/master/core/parameters/res/runtime_configs/69.yaml)). As of the latest main, the binary is 1,550,594 bytes — only 22KB of headroom. Adding new contract methods (e.g. backup service TEE attestation in #2730) pushes the binary over this limit, causing `413 Payload Too Large` errors in sandbox tests and preventing contract deployment.

### Acceptance Criteria

- [x] Switch `opt-level` from `"s"` to `"z"` in the `release-contract` profile — saves ~69KB (4.5%)
- [x] Add CI check for contract WASM size with ratcheting baseline (`.github/contract-size-baseline.toml`)
- [x] Update gas regression thresholds and bump `cleanup_orphaned_node_migrations` gas from 3 to 4 TGas
- [ ] Investigate further optimizations if needed (e.g. stripping panic location strings from dependencies)

- NEAR `max_transaction_size`: 1,572,864 bytes since [protocol v69](https://github.com/near/nearcore/blob/master/core/parameters/res/runtime_configs/69.yaml)
- Reproducible build with `opt-level = "s"`: 1,550,594 bytes (22KB headroom)
- Reproducible build with `opt-level = "z"`: 1,481,277 bytes (91KB headroom)
- The [Rust docs](https://doc.rust-lang.org/rustc/codegen-options/index.html#opt-level) describe `"s"` as "optimize for binary size" and `"z"` as "optimize for binary size, but more aggressively. Often results in larger binaries than `s`". In practice this is not always true — in our case `"z"` produces a 4.5% smaller binary
- Biggest remaining bloat: panic location strings from cryptographic dependencies (dcap-qvl-webpki, ring) — ~25KB of local file paths embedded in the data section
- Unblocks: #2730 (backup service TEE attestation)
- PR: #2869

pbeza:
## Summary

Three changes to address the contract binary approaching NEAR's `max_transaction_size` limit (1,572,864 bytes since [protocol v69](https://github.com/near/nearcore/blob/master/core/parameters/res/runtime_configs/69.yaml)):

1. **Switch `opt-level` from `"s"` to `"z"`** in the `release-contract` profile

   | opt-level | WASM size | Headroom |
   |---|---|---|
   | `"s"` (before) | 1,550,594 | 22,270 |
   | `"z"` (after) | 1,481,277 | 91,587 |
   | **Savings** | **69,317 (4.5%)** | |

   The [Rust docs](https://doc.rust-lang.org/rustc/codegen-options/index.html#opt-level) describe `"s"` as "optimize for binary size" and `"z"` as "optimize for binary size, but more aggressively. Often results in larger binaries than `s`". In practice this is not always true — in our case `"z"` produces a 4.5% smaller binary.

2. **Add CI size check** after the reproducible contract build
   - Reads the hard limit and expected size from `.github/contract-size-baseline.toml`
   - Fails if the binary size differs from the baseline (grew or shrank)
   - When the contract grows, the developer must explicitly update the baseline
   - When the contract shrinks, the developer updates the baseline to lock in the improvement

3. **Adjust gas thresholds and config** — `opt-level "z"` produces smaller but slightly slower code
   - Update gas regression test thresholds
   - Bump `DEFAULT_CLEANUP_ORPHANED_NODE_MIGRATIONS_TERA_GAS` from 3 to 4 TGas (promise exceeded prepaid gas during resharing)

Closes #2868

## Test plan
- [x] Verified WASM sizes back-to-back with same toolchain
- [x] Script tested locally
- [x] Baseline updated to match reproducible build output
- [x] Gas regression tests pass (3 consecutive runs)
- [x] Resharing tests pass (3 consecutive runs)
- [x] CI passes

### DIFF
--- a/.github/scripts/check-contract-wasm-size.sh
+++ b/.github/scripts/check-contract-wasm-size.sh
@@ -1,0 +1,76 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Checks the contract WASM binary size against:
+# 1. NEAR's max_transaction_size hard limit
+# 2. A committed baseline that ratchets down over time
+#
+# Usage: bash .github/scripts/check-contract-wasm-size.sh <path-to-wasm>
+
+WASM_PATH="${1:?Usage: $0 <path-to-wasm>}"
+BASELINE_FILE=".github/contract-size-baseline.toml"
+
+read_toml_value() {
+    grep "^$1 " "$BASELINE_FILE" | sed 's/.*= *//'
+}
+
+if [[ ! -f "$WASM_PATH" ]]; then
+    echo "❌ WASM file not found: $WASM_PATH"
+    exit 1
+fi
+
+if [[ ! -f "$BASELINE_FILE" ]]; then
+    echo "❌ Baseline file not found: $BASELINE_FILE"
+    exit 1
+fi
+
+MAX_TX_SIZE=$(read_toml_value "max_transaction_size")
+BASELINE=$(read_toml_value "expected_size")
+
+if [[ -z "$MAX_TX_SIZE" || -z "$BASELINE" ]]; then
+    echo "❌ Failed to read max_transaction_size or expected_size from $BASELINE_FILE"
+    exit 1
+fi
+
+SIZE=$(wc -c < "$WASM_PATH" | tr -d ' ')
+HEADROOM=$(( MAX_TX_SIZE - SIZE ))
+DELTA=$(( SIZE - BASELINE ))
+
+echo "Contract WASM size report"
+echo "========================="
+echo "  Binary:     $WASM_PATH"
+echo "  Size:       $SIZE bytes"
+echo "  Hard limit: $MAX_TX_SIZE bytes (NEAR max_transaction_size)"
+echo "  Headroom:   $HEADROOM bytes"
+echo "  Baseline:   $BASELINE bytes"
+if [[ "$DELTA" -gt 0 ]]; then
+    echo "  Delta:      +$DELTA bytes"
+elif [[ "$DELTA" -lt 0 ]]; then
+    echo "  Delta:      $DELTA bytes"
+fi
+echo ""
+
+# Hard limit check
+if [[ "$SIZE" -gt "$MAX_TX_SIZE" ]]; then
+    OVER=$(( SIZE - MAX_TX_SIZE ))
+    echo "❌ EXCEEDS hard limit by $OVER bytes"
+    echo "   The contract binary is too large to deploy in a single transaction"
+    echo "   Reduce size by removing methods, shrinking dependencies,"
+    echo "   or optimizing the build profile"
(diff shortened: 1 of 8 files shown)


scripts/check-contract-wasm-size.sh
```diff
@@ -0,0 +1,76 @@
...
+set -euo pipefail
+
+# Checks the contract WASM binary size against:
+# 1. NEAR's max_transaction_size hard limit
+# 2. A committed baseline that ratchets down over time
+#
+# Usage: bash .github/scripts/check-contract-wasm-size.sh <path-to-wasm>
+
+WASM_PATH="${1:?Usage: $0 <path-to-wasm>}"
+BASELINE_FILE=".github/contract-size-baseline.toml"
```
netrome, line 11: Nit: Code organization - this file (and the current script) doesn't really have anything to do with GitHub. Feels like they could live in our normal `scripts` folder (or a subfolder).

netrome (dismissed):
Nice stuff - just an organization nit

.github/scripts/check-contract-wasm-size.sh
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Checks the contract WASM binary size against:
+# 1. NEAR's max_transaction_size hard limit
+# 2. A committed baseline that ratchets down over time
+#
+# Usage: bash .github/scripts/check-contract-wasm-size.sh <path-to-wasm>
+
+WASM_PATH="${1:?Usage: $0 <path-to-wasm>}"
```
gilcu3: could we make the default be target/.... (the default wasm location of the contract)

gilcu3 (dismissed):
Thank you!

Would be nice to enable running this locally with cargo-make. How I picture it is having a single command in the Makefile.toml to build the contract and run this script. Then would make it part of the normal checks (not the fast checks, although this might make sense, or not)

.github/scripts/check-contract-wasm-size.sh
pbeza: https://github.com/near/mpc/pull/2869/changes/d0a1e57232d3e54334b2fae38c03a43f343f139e

scripts/check-contract-wasm-size.sh
pbeza: https://github.com/near/mpc/pull/2869/changes/d0a1e57232d3e54334b2fae38c03a43f343f139e

gilcu3 (approved):

netrome (approved):

PR #2870, opened 2026-04-14
refactor: participants should depend on contract interface not contract

### LINKED ISSUE #2167: participants.rs should depend on contract-interface, not contract
### Background

We currently depend on contract-internal types in the node:
https://github.com/near/mpc/blob/534caf1e5ebb82cfec193d59dcdd3b52e5c66416/crates/node/src/indexer/participants.rs#L7-L11

We should not do that. We should only rely on the contract-interface

### User Story

As a dev, I want to avoid accidentally breaking API's.

### Acceptance Criteria

The node's participants module does no longer depend on the mpc-contract.

Re scope:
in #2870 we reduced the usage of mpc-contract types in participants, but could not remove completely, as that required also removing types related to keyshares. That part of this issue should be solved by #2881 

_No response_

gilcu3:
Closes #2167 

The remaining non-test mpc-contract usage in the node:

  Participants boundary (this PR's scope)

  - indexer/participants.rs — DomainConfig, KeyEventId, KeyForDomain, Keyset (converted at boundary from interface types)

  Key state / keyshares

  - keyshare.rs — Keyset, EpochId, KeyEventId, KeyForDomain
  - keyshare/temporary.rs — EpochId, KeyEventId
  - keyshare/permanent.rs — DomainId, EpochId
  - keyshare/compat.rs — DomainId, AttemptId, EpochId, KeyEventId

  Coordinator / key events

  - coordinator.rs — Curve, DomainId, EpochId
  - key_events.rs — Curve, DomainConfig, KeyEventId, KeyForDomain, Keyset
  - assets/cleanup.rs — EpochId

  Signature / request types

  - types.rs — DomainId, Payload, Tweak
  - indexer/types.rs — CKDResponse, Tweak, Payload
  - indexer/handler.rs — CKDRequest, Payload, SignRequest, SignRequestArgs
  - mpc_client.rs — derive_tweak, CKDResponse, Curve, DomainId
  - providers/*/sign.rs (4 files) — Tweak, Payload, Bytes

  Migration

  - migration_service/types.rs — BackupServiceInfo, DestinationNodeInfo, Keyset
  - migration_service/onboarding.rs — Keyset
  - migration_service/web/server.rs — Keyset
  - indexer/migrations.rs — BackupServiceInfo, DestinationNodeInfo

  TEE

  - tee/remote_attestation.rs — TEE types
  - tee/allowed_image_hashes_watcher.rs — NodeImageHash
  - indexer/tee.rs — LauncherDockerComposeHash, NodeImageHash, NodeId

  Indexer

  - indexer.rs — YieldIndex, TEE types

It is a lot, but hopefully can be done by parts


crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -203,9 +205,7 @@ pub struct DomainConfig {
     // Accepts "scheme" for compat with pre-3.9.0 contracts. Remove after 3.9.0 deployment.
     #[serde(alias = "scheme")]
     pub curve: Curve,
-    /// `None` when reading state from an old contract that predates domain purposes.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub purpose: Option<DomainPurpose>,
+    pub purpose: DomainPurpose,
```
gilcu3, line 208: this was leftover from a previous migration which became annoying when doing changes here

crates/near-mpc-crypto-types/src/conversions/near.rs
```diff
@@ -72,6 +72,47 @@ impl TryFrom<&near_sdk::PublicKey> for Secp256k1PublicKey {
     }
 }
 
+impl TryFrom<PublicKeyExtended> for PublicKey {
```
gilcu3: IMO only one of these types is needed in the interface, but tried not to fix that to keep the PR "lean"

crates/node/src/web.rs
```diff
@@ -386,7 +385,8 @@ async fn contract_state(state: State<WebServerState>) -> String {
         // Clone to avoid holding a lock
         .clone();
 
-    protocol_state_to_string(&protocol_state)
+    // TODO: check if the debug output looks nice, otherwise port protocol_state_to_string
+    format!("{:#?}", protocol_state)
```
gilcu3, line 389: I can also remove the TODO. This is about the page /debug/contract, for which the format is not too important

crates/node/src/indexer/participants.rs
```diff
@@ -1,80 +1,86 @@
...
 pub struct ContractKeyEventInstance {
-    pub id: KeyEventId,
-    pub domain: DomainConfig,
+    pub id: ContractKeyEventId,
+    pub domain: ContractDomainConfig,
     pub started: bool,
     pub completed: BTreeSet<ParticipantId>,
-    pub completed_domains: Vec<KeyForDomain>,
+    pub completed_domains: Vec<ContractKeyForDomain>,
 }
```
gilcu3, line 26: this must remain using contract types, else the cascade of changes in the keyshares (that have some logic, not only data) becomes tricky. We need to solve this in a follow-up

netrome:
> It is a lot, but hopefully can be done by parts

Thanks for breaking this down :pray:

crates/near-mpc-contract-interface/src/types/state.rs
netrome: Thanks :pray:

crates/near-mpc-crypto-types/src/conversions/near.rs
netrome: Yeah I agree.

crates/node/src/web.rs
netrome: YEah please remove it if you don't plan to address it here or create a follow-up issue for it.

crates/near-mpc-crypto-types/src/conversions/near.rs
netrome: Yeah why do we have both? It's not crystal clear and I don't recall right now, is it for backwards compatibility somewhere?

netrome (dismissed):
Thanks for doing this!

crates/node/src/web.rs
gilcu3: I would love to address here, but the needed function is also used in other crate. So to avoid duplicate code would need to decide that :| Will add an issue

netrome: Yeah makes sense. Thanks

crates/near-mpc-crypto-types/src/conversions/near.rs
gilcu3: I need to check if any of them is currently used or not, I had a huge amount of back and forth to make all conversions work, so might have slipped.

gilcu3: turns out this was not needed, removed

crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -248,7 +248,16 @@ pub struct Keyset {
 
 /// Identifier for a key event (generation or resharing attempt).
 #[derive(
-    Clone, Copy, Debug, Eq, PartialEq, Serialize, Deserialize, BorshSerialize, BorshDeserialize,
+    Clone,
+    Copy,
+    Debug,
+    Eq,
+    PartialEq,
+    Hash,
```
kevindeforth, line 256: This `Hash` is used by the node, not the contract. In the interest of keeping the contract size small, have you considered feature-gating this?

crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -33,6 +33,7 @@ use super::primitives::DomainId;
     derive_more::From,
     derive_more::Into,
     derive_more::AsRef,
+    derive_more::Display,
```
kevindeforth, line 36: nit: These things add to the contract size over time, why do we need to add this?

crates/node/src/indexer/participants.rs
barakeinav1: can you add a TODO here, or at least paste here the follow issue?

barakeinav1 (dismissed):
looks good, thanks!

crates/node/src/indexer/participants.rs
gilcu3: the todo is not really needed because it will break as soon as the dep is removed. The issue we agreed this morning was https://github.com/near/mpc/issues/2881 but it is not 100% it will remove this, the parent issue #381 for sure will do it

kevindeforth (dismissed):
Thank you!
nit: This doesn't resolve https://github.com/near/mpc/issues/2167 fully, so would be good to re-scope the issue.

crates/near-mpc-contract-interface/src/types/state.rs
gilcu3: Yeah probably many of them are not needed, will open an issue for that

crates/near-mpc-contract-interface/src/types/state.rs
gilcu3: I think this one was needed for the debug endpoint, will handle in the same issue I opened for the comment above

crates/near-mpc-contract-interface/src/types/state.rs
gilcu3: opened https://github.com/near/mpc/issues/2899

gilcu3:
> Thank you! nit: This doesn't resolve #2167 fully, so would be good to re-scope the issue.

Re scoped to mention what was done and what is left

barakeinav1 (approved):

kevindeforth (approved):

PR #2893, opened 2026-04-15
feat: make quote_upload_url configurable in MPC node config

### LINKED ISSUE #1545: Introduce independent CRL endpoint and remove dependency on Phala’s CRL endpoint

barakeinav1:
## Summary
- Add optional `quote_upload_url` field to `StartConfig` (`[mpc_node_config]` section)
- When set, overrides the default Phala endpoint injected by the launcher
- Lets operators point at a local PCCS proxy or any alternative collateral endpoint
- Collateral is cryptographically signed by Intel regardless of source — no security impact

## Changes
- `crates/node-config/` — new `quote_upload_url: Option<url::Url>` field with serde default
- `crates/node/src/config/start.rs` — `into_tee_authority` accepts URL override
- `crates/node/src/run.rs` — passes override through
- `crates/tee-authority/` — refactored tests with rstest, updated error message
- `.config/nextest.toml` — updated filter for renamed tests
- `localnet/tee/scripts/` — commented-out config example

## Test plan
- [x] `cargo check` passes for mpc-node, mpc-node-config, tee-launcher
- [x] Existing tests pass
- [x] E2E tested with local PCCS proxy on TDX CVM (see near/mpc#2714)

Relates to #467, #1545

Closes #1545

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -18,4 +18,4 @@ filter = 'package(e2e-tests)'
 test-group = 'e2e'
 
 [profile.ci-other]
-default-filter = 'not package(mpc-node) and not package(mpc-contract) and not package(e2e-tests) and not package(contract-history) and not test(=tee_authority::tests::test_upload_quote_for_collateral_with_phala_endpoint) and not (test(/^registry::integration_tests::/) and package(tee-launcher))'
+default-filter = 'not package(mpc-node) and not package(mpc-contract) and not package(e2e-tests) and not package(contract-history) and not test(/^tee_authority::tests::test_upload_quote_for_collateral/) and not (test(/^registry::integration_tests::/) and package(tee-launcher))'
--- a/.github/workflows/external-services-tests.yml
+++ b/.github/workflows/external-services-tests.yml
@@ -49,6 +49,6 @@ jobs:
         if: always()
         run: cargo nextest run -p tee-launcher --features external-services-tests --cargo-profile test-release --locked
 
-      - name: Run tee-authority tests
+      - name: Run tee-authority Phala collateral tests
         if: always()
-        run: cargo nextest run -p tee-authority --features external-services-tests --cargo-profile test-release --locked -E 'test(test_upload_quote_for_collateral_with_phala_endpoint)'
+        run: cargo nextest run -p tee-authority --features external-services-tests --cargo-profile test-release --locked -E 'test(/test_upload_quote_for_collateral::case_[0-9]+_phala/)'
--- a/.github/workflows/external-services-tests.yml
+++ b/.github/workflows/external-services-tests.yml
@@ -51,4 +51,4 @@ jobs:
 
       - name: Run tee-authority Phala collateral tests
         if: always()
-        run: cargo nextest run -p tee-authority --features external-services-tests --cargo-profile test-release --locked -E 'test(/test_upload_quote_for_collateral::case_[0-9]+_phala/)'
+        run: cargo nextest run -p tee-authority --features external-services-tests --cargo-profile test-release --locked -E 'test(test_upload_quote_for_collateral::case_1_phala_endpoint) | test(test_upload_quote_for_collateral::case_2_phala_via_override)'
--- a/.github/workflows/external-services-tests.yml
+++ b/.github/workflows/external-services-tests.yml
@@ -51,4 +51,4 @@ jobs:
 
       - name: Run tee-authority Phala collateral tests
         if: always()
-        run: cargo nextest run -p tee-authority --features external-services-tests --cargo-profile test-release --locked -E 'test(test_upload_quote_for_collateral::case_1_phala_endpoint) | test(test_upload_quote_for_collateral::case_2_phala_via_override)'
+        run: cargo nextest run -p tee-authority --features external-services-tests --cargo-profile test-release --locked -E 'test(/test_upload_quote_for_collateral::case_[0-9]+_phala/)'
--- a/.github/workflows/external-services-tests.yml
+++ b/.github/workflows/external-services-tests.yml
@@ -51,4 +51,4 @@ jobs:
 
       - name: Run tee-authority Phala collateral tests
         if: always()
-        run: cargo nextest run -p tee-authority --features external-services-tests --cargo-profile test-release --locked -E 'test(test_upload_quote_for_collateral::case_1_phala_endpoint) | test(test_upload_quote_for_collateral::case_2_phala_via_override)'
+        run: cargo nextest run -p tee-authority --features external-services-tests --cargo-profile test-release --locked -E 'test(/test_upload_quote_for_collateral::case_[0-9]+_phala_/)'
--- a/crates/node-config/src/start.rs
+++ b/crates/node-config/src/start.rs
@@ -24,15 +24,10 @@ pub struct StartConfig {
     /// Node configuration (indexer, protocol parameters, etc.).
     pub node: ConfigFile,
     pub log: LogConfig,
-    /// TDX quote collateral endpoint URL. Defaults to Phala's public endpoint
-    #[serde(default = "default_quote_upload_url")]
-    pub quote_upload_url: url::Url,
-}
-
-pub fn default_quote_upload_url() -> url::Url {
-    launcher_interface::DEFAULT_PHALA_TDX_QUOTE_UPLOAD_URL
-        .parse()
-        .expect("default quote upload URL is valid")
+    /// Optional override for the TDX quote collateral endpoint URL.
+    /// When set, overrides the URL injected by the launcher in the `[tee]` section.
+    #[serde(default)]
+    pub quote_upload_url: Option<url::Url>,
 }
 
 impl StartConfig {
--- a/crates/node-config/src/start.rs
+++ b/crates/node-config/src/start.rs
@@ -26,9 +26,6 @@ pub struct StartConfig {
     pub log: LogConfig,
     /// Optional override for the TDX quote collateral endpoint URL.
     /// When set, overrides the URL injected by the launcher in the `[tee]` section.
-    /// This does not affect attestation security — the collateral data is
-    /// cryptographically signed by Intel regardless of where it is fetched from.
-    /// Example: `"http://51.68.219.1:8082/api/v1/attestations/verify"`
     #[serde(default)]
     pub quote_upload_url: Option<url::Url>,
 }
--- a/crates/node-config/src/start.rs
+++ b/crates/node-config/src/start.rs
@@ -24,10 +24,16 @@ pub struct StartConfig {
     /// Node configuration (indexer, protocol parameters, etc.).
     pub node: ConfigFile,
     pub log: LogConfig,
-    /// Optional override for the TDX quote collateral endpoint URL.
-    /// When set, overrides the URL injected by the launcher in the `[tee]` section.
-    #[serde(default)]
-    pub quote_upload_url: Option<url::Url>,
+    /// TDX quote collateral endpoint URL.
+    /// Defaults to Phala's public endpoint if not set in config.
+    #[serde(default = "default_quote_upload_url")]
+    pub quote_upload_url: url::Url,
+}
+
+pub fn default_quote_upload_url() -> url::Url {
+    "https://cloud-api.phala.network/api/v1/attestations/verify"
+        .parse()
+        .expect("default quote upload URL is valid")
 }
 
 impl StartConfig {
--- a/crates/node-config/src/start.rs
+++ b/crates/node-config/src/start.rs
@@ -24,6 +24,13 @@ pub struct StartConfig {
     /// Node configuration (indexer, protocol parameters, etc.).
     pub node: ConfigFile,
     pub log: LogConfig,
+    /// Optional override for the TDX quote collateral endpoint URL.
+    /// When set, overrides the URL injected by the launcher in the `[tee]` section.
+    /// This does not affect attestation security — the collateral data is
+    /// cryptographically signed by Intel regardless of where it is fetched from.
+    /// Example: `"http://51.68.219.1:8082/api/v1/attestations/verify"`
+    #[serde(default)]
+    pub quote_upload_url: Option<url::Url>,
 }
 
 impl StartConfig {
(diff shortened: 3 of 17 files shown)


crates/tee-authority/src/tee_authority.rs
```diff
@@ -477,22 +477,39 @@ mod tests {
...
+    /// Parameterized test for collateral upload endpoints.
+    ///
+    /// - `phala_endpoint`: Tests against Phala using the default config path.
+    /// - `phala_via_override`: Tests against Phala using the URL override path.
+    /// - `local_pccs_proxy`: Tests against the local PCCS proxy.
+    ///   Requires the proxy to be running. Override URL via `LOCAL_PCCS_PROXY_URL`.
+    #[rstest::rstest]
+    #[case::phala_endpoint(None)]
+    #[case::phala_via_override(Some("https://cloud-api.phala.network/api/v1/attestations/verify"))]
+    #[case::local_pccs_proxy(Some("http://localhost:8082/api/v1/attestations/verify"))]
```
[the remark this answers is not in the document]
barakeinav1: The `local_pccs_proxy` case is excluded from CI via the nextest `ci-other` filter (`not test(/^tee_authority::tests::test_upload_quote_for_collateral/)`). The `external-services-tests` workflow now uses a regex that only matches Phala cases (`test(/test_upload_quote_for_collateral::case_[0-9]+_phala/)`), so the local proxy case won't run there either.

.config/nextest.toml
```diff
@@ -18,4 +18,4 @@ filter = 'package(e2e-tests)'
 test-group = 'e2e'
 
 [profile.ci-other]
-default-filter = 'not package(mpc-node) and not package(mpc-contract) and not package(e2e-tests) and not package(contract-history) and not test(=tee_authority::tests::test_upload_quote_for_collateral_with_phala_endpoint) and not (test(/^registry::integration_tests::/) and package(tee-launcher))'
+default-filter = 'not package(mpc-node) and not package(mpc-contract) and not package(e2e-tests) and not package(contract-history) and not test(/^tee_authority::tests::test_upload_quote_for_collateral/) and not (test(/^registry::integration_tests::/) and package(tee-launcher))'
```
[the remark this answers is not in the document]
barakeinav1: Fixed — updated the `-E` filter to `test(/test_upload_quote_for_collateral::case_[0-9]+_phala/)` which matches the rstest-generated names for Phala cases only.

crates/node/src/run.rs
```diff
@@ -58,6 +58,11 @@ pub async fn run_mpc_node(config: StartConfig) -> anyhow::Result<()> {
             launcher_interface::types::TeeAuthorityConfig::Local => "local",
         },
         image_hash = %config.tee.image_hash,
+        quote_upload_url = %match (&config.quote_upload_url, &config.tee.authority) {
+            (Some(url), _) => url.as_str(),
+            (None, launcher_interface::types::TeeAuthorityConfig::Dstack { quote_upload_url, .. }) => quote_upload_url.as_str(),
+            (None, launcher_interface::types::TeeAuthorityConfig::Local) => "n/a",
```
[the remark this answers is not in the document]
barakeinav1: Fixed — Local mode now always logs `quote_upload_url=n/a` regardless of whether the field is set. The override is only shown when `tee_authority=dstack`.

.config/nextest.toml
barakeinav1, line 21: Regular CI (ci-other profile, runs on every PR): All 3 test cases are excluded by the regex.?
Same behavior as before — previously excluded by exact name match on the single test.                                                                                                                               
                                                                                                                                             
External services CI (external-services-tests.yml, runs on every push to main + weekly, non-blocking): Runs phala_endpoint             
  phala_via_override (both hit Phala's live endpoint). Excludes local_pccs_proxy (needs a running proxy). 
 Before this PR, it ran the single                                           
  test_upload_quote_for_collateral_with_phala_endpoint test.

note - I don't think anyone monitors the failed tests if they happen. but maybe I'm wrong.  FYI @netrome

crates/tee-authority/src/tee_authority.rs
```diff
@@ -504,9 +521,12 @@ mod tests {
...
                 assert!(!collateral.qe_identity.is_empty());
                 assert!(!collateral.qe_identity_signature.is_empty());
+                assert!(!collateral.pck_crl_issuer_chain.is_empty());
+                assert!(!collateral.root_ca_crl.is_empty());
+                assert!(!collateral.pck_crl.is_empty());
             }
-            Ok(Err(e)) => panic!("Test failed: {e:?}"),
-            Err(e) => panic!("Test timed out: {e:?}"),
+            Ok(Err(e)) => panic!("Collateral upload failed ({url}): {e:?}"),
+            Err(e) => panic!("Collateral upload timed out ({url}): {e:?}"),
```
pbeza:
```suggestion
            Ok(Err(e)) => panic!("quote upload failed ({url}): {e:?}"),
            Err(e) => panic!("quote upload timed out ({url}): {e:?}"),
```

.github/workflows/external-services-tests.yml
```diff
@@ -49,6 +49,6 @@ jobs:
         if: always()
         run: cargo nextest run -p tee-launcher --features external-services-tests --cargo-profile test-release --locked
 
-      - name: Run tee-authority tests
+      - name: Run tee-authority Phala collateral tests
         if: always()
-        run: cargo nextest run -p tee-authority --features external-services-tests --cargo-profile test-release --locked -E 'test(test_upload_quote_for_collateral_with_phala_endpoint)'
+        run: cargo nextest run -p tee-authority --features external-services-tests --cargo-profile test-release --locked -E 'test(/test_upload_quote_for_collateral::case_[0-9]+_phala/)'
```
pbeza: Having a regex here makes it quite hard to reason about which tests are actually being run (and why). It’s also brittle if test names change. Let’s make it more explicit:

```suggestion
        run: cargo nextest run -p tee-authority --features external-services-tests --cargo-profile test-release --locked -E 'test(test_upload_quote_for_collateral::case_1_phala_endpoint) | test(test_upload_quote_for_collateral::case_2_phala_via_override)'
```

crates/node-config/src/start.rs
```diff
@@ -24,6 +24,10 @@ pub struct StartConfig {
     /// Node configuration (indexer, protocol parameters, etc.).
     pub node: ConfigFile,
     pub log: LogConfig,
+    /// Optional override for the TDX quote collateral endpoint URL.
+    /// When set, overrides the URL injected by the launcher in the `[tee]` section.
+    #[serde(default)]
+    pub quote_upload_url: Option<url::Url>,
```
pbeza: Optional nit: Reasoning about `Option`s adds some mental overhead when interpreting `None`. Conceptually, `quote_upload_url` should always be present, so it might make sense to get rid of the `Option`. Something along these lines: https://github.com/near/mpc/pull/2921.

localnet/tee/scripts/rust-launcher/node.conf.localnet.toml.tpl
```diff
@@ -8,6 +8,9 @@ ${PORTS_TOML}]
 
 [mpc_node_config]
 home_dir = "/data"
+# Override the collateral endpoint URL. Defaults to Phala's endpoint if omitted.
+# To use the local PCCS proxy: "http://${MACHINE_IP}:8082/api/v1/attestations/verify"
```
pbeza: Copy-paste of the below line.
```suggestion
```

pbeza (changes requested):

localnet/tee/scripts/rust-launcher/node.conf.localnet.toml.tpl
barakeinav1: Fixed.

.github/workflows/external-services-tests.yml
barakeinav1: Fixed — using explicit test names now.

crates/tee-authority/src/tee_authority.rs
barakeinav1: Fixed.

crates/node-config/src/start.rs
barakeinav1: Agreed — adopting your approach from #2921 (required field with serde default). Also cleaning up the launcher: removing `quote_upload_url` from `TeeAuthorityConfig::Dstack` and `launcher-interface` so the node config is the single source of truth.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -477,22 +477,39 @@ mod tests {
...
 
+    /// Parameterized test for collateral upload endpoints.
+    ///
+    /// - `phala_endpoint`: Tests against Phala using the default config path.
+    /// - `phala_via_override`: Tests against Phala using the URL override path.
+    /// - `local_pccs_proxy`: Tests against the local PCCS proxy.
+    ///   Requires the proxy to be running. Override URL via `LOCAL_PCCS_PROXY_URL`.
+    #[rstest::rstest]
+    #[case::phala_endpoint(None)]
+    #[case::phala_via_override(Some("https://cloud-api.phala.network/api/v1/attestations/verify"))]
```
pbeza, line 488:
```suggestion
    #[case::phala_via_override(Some(DEFAULT_PHALA_TDX_QUOTE_UPLOAD_URL))]
```

barakeinav1: Keeping the hardcoded URL — we moved `DEFAULT_PHALA_TDX_QUOTE_UPLOAD_URL` out of `launcher-interface` into `node-config` as `default_quote_upload_url()` (returns `Url`). The test takes `Option<&str>` so a string literal is simpler here. The `phala_via_override` case validates that an explicit URL works end-to-end without needing a local proxy.

pbeza (dismissed):

.github/workflows/external-services-tests.yml
```diff
@@ -49,6 +49,6 @@ jobs:
         if: always()
         run: cargo nextest run -p tee-launcher --features external-services-tests --cargo-profile test-release --locked
 
-      - name: Run tee-authority tests
+      - name: Run tee-authority Phala collateral tests
         if: always()
-        run: cargo nextest run -p tee-authority --features external-services-tests --cargo-profile test-release --locked -E 'test(test_upload_quote_for_collateral_with_phala_endpoint)'
+        run: cargo nextest run -p tee-authority --features external-services-tests --cargo-profile test-release --locked -E 'test(test_upload_quote_for_collateral::case_1_phala_endpoint) | test(test_upload_quote_for_collateral::case_2_phala_via_override)'
```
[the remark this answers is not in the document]
pbeza: Oh, I didn’t realize that — that might go against what [I suggested](https://github.com/near/mpc/pull/2893#discussion_r3091931214). Sorry. :(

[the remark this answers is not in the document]
barakeinav1: I'll revert back to regex.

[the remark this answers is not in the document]
barakeinav1: Fixed — switched to regex `test(/test_upload_quote_for_collateral::case_[0-9]+_phala_/)` for stability against case reordering.

pbeza:
Sorry, reverting my approval — I think I [found](https://github.com/near/mpc/pull/2894#issuecomment-4266769461) a way to get rid of the PCCS proxy altogether, so this PR might not be needed anymore. Will update shortly.

pbeza:
> Sorry, reverting my approval — I think I [found](https://github.com/near/mpc/pull/2894#issuecomment-4266769461) a way to get rid of the PCCS proxy altogether, so this PR might not be needed anymore. Will update shortly.

Confirmed. See: https://github.com/near/mpc/pull/2944. I'd suggest closing this PR.

barakeinav1:
Superseded by #2944, which takes a cleaner approach — calling `dcap_qvl::collateral::get_collateral()` directly from the node instead of adding a proxy service. No new binary to deploy, no extra network hop.

Issue #2883, opened 2026-04-15
Add basic telemetry on CPU usage in E2E tests to refine max-threads value


anodar:
### Background

Currently we have max-threads=8 to restrict concurrency in tests, otherwise we get timeouts.

We should collect data about CPU / Memory usage during test execution and adjust if 32x runner has enough resource for it.

### Acceptance Criteria

We have data backing up the max-threads value.

anodar:
We already have those metrics in https://app.warpbuild.com/ci/observability

PR #2907, opened 2026-04-15
refactor: merge parallel tests, refactor code, reword comments

### LINKED ISSUE #2857: test: port parallel_sign_calls, ckd_verification, and lost_assets (dead node) from pytest to Rust e2e

anodar:
Follow-up to #2859

Closes: https://github.com/near/mpc/issues/2857


crates/e2e-tests/tests/parallel_sign_calls.rs
```diff
@@ -4,26 +4,44 @@ use e2e_tests::{CLUSTER_WAIT_TIMEOUT, metrics};
...
 use serde_json::json;
 
-/// 3 parallel robust ECDSA (V2Secp256k1) sign calls via the test parallel contract.
-/// Verifies all calls succeed and the signature queue processes all requests.
-// TODO(#2898): re-enable once the flakiness is fixed.
-#[ignore]
+/// 9 parallel calls (3 robust ECDSA + 2 ECDSA + 2 EdDSA + 2 CKD) via the test parallel
+/// contract, against a 6-node / threshold-5 cluster that carries all four signing-scheme
+/// domains. Verifies all calls succeed and both the signature and CKD queues drain.
 #[tokio::test]
```
anodar, line 10: Re-enabling this test even if it turns out flaky at least I can test that it passes on this PR since it merges two tests. I've reduced max-threads to 4 and referenced issue there to debug flakiness.

Since this isn't blocking shouldn't create any friction besides a bit of a noise (if it turns out to still be flaky).

.config/nextest.toml
```diff
@@ -10,7 +10,8 @@ filter = 'package(mpc-contract) and test(/sandbox/)'
 test-group = 'contract-sandbox'
 
 [test-groups.e2e]
-max-threads = 8
+# TODO(#2898): tune this if parallel_sign_calls test is still flaky.
+max-threads = 4
```
anodar, line 14: That is, if flakiness isn't because of too much concurrency, we'll bump this back up and address actual root cause. Or drop this TODO as I get to that issue and investigate / confirm this was the root cause.

DSharifi (approved):
Thank you

.config/nextest.toml
netrome: I'm a bit bothered by this flakyness. Why don't we run these tests on a single thread? And additionally, is this resource contention something we should be worried about in production as well?

crates/e2e-tests/src/conversions.rs
```diff
@@ -0,0 +1,16 @@
...
+
+/// Extract the public half of `key` as a [`near_kit::PublicKey`].
+pub fn signing_key_to_near_public_key(key: &SigningKey) -> near_kit::PublicKey {
+    near_kit::PublicKey::Ed25519(key.verifying_key().to_bytes())
+}
+
+/// Convert `key` into a [`near_kit::SecretKey`].
+pub fn signing_key_to_near_secret_key(key: &SigningKey) -> near_kit::SecretKey {
+    near_kit::SecretKey::Ed25519(key.to_bytes())
+}
```
netrome, line 16: Nit: For conversions I find it nice to add helper traits, since it makes the calling code nicer (purely subjective).

I.e. you could introduce a trait like `ToNearKey` that contains two methods: `to_near_public_key()` and `to_near_secret_key()` which would allow you to do `key.to_near_secret_key()` in the code using this.

crates/e2e-tests/tests/ckd_verification.rs
```diff
@@ -1,32 +1,24 @@
 use crate::common;
 
-use blstrs::{G1Affine, G1Projective, G2Affine, G2Projective, Scalar};
-use e2e_tests::CKD_PV_GAS;
+use blstrs::{G1Projective, G2Projective, Scalar};
 use group::Group as _;
 use group::ff::Field as _;
-use group::prime::PrimeCurveAffine as _;
+use mpc_contract::crypto_shared::kdf::derive_app_id;
```
netrome, line 6: Feels like this one should be broken out to a shared primitives module or something. Don't like the idea of depending on the contract here.

Wanna do this as a follow-up?

netrome (approved):
Some nits, otherwise looks good from a quick look.

.config/nextest.toml
gilcu3: I dont think it affects production, the contention comes from running several tests at the same time. 

About this change:
I think 4 is a very low number here (the runner has 32 cores), I presume you are using it to try to fix the general flakiness problem, but it might be more nuanced. Also yesterday the runners seemed much slower than usual, so maybe it was a very bad day to measure these things.

crates/e2e-tests/tests/ckd_verification.rs
gilcu3: I think we need to do this either way to break the mpc-contract dependency in the node, so this will surely be handled then. Good to keep in mind though.

crates/e2e-tests/tests/ckd_verification.rs
```diff
@@ -166,10 +138,9 @@ async fn ckd_pv_response__passes_cryptographic_verification() {
     let big_c: Bls12381G1PublicKey =
         serde_json::from_value(response["big_c"].clone()).expect("failed to parse big_c");
 
-    let account_id = cluster.default_user_account().as_str();
     assert!(
         verify_ckd(
-            account_id,
+            cluster.default_user_account(),
```
gilcu3, line 143: this decoupling is a bit odd. For CKD we need the account_id as part of the request, which basically means that send_ckd_request should get account_id as parameter, else when verifying we need to "guess" which account made the request

crates/e2e-tests/tests/parallel_sign_calls.rs
gilcu3: lets hope it does not become flaky :)

crates/e2e-tests/tests/parallel_sign_calls.rs
```diff
@@ -4,26 +4,44 @@ use e2e_tests::{CLUSTER_WAIT_TIMEOUT, metrics};
...
-#[expect(non_snake_case)]
-async fn mpc_cluster__should_successfully_process_parallel_robust_ecdsa_requests() {
-    const N: u64 = 3;
+async fn mpc_cluster_should_successfully_process_robust_ecdsa_and_mixed_parallel_requests() {
```
gilcu3, line 11: name is odd now:
```suggestion
async fn mpc_cluster_should_successfully_process_parallel_requests() {
```

crates/e2e-tests/tests/parallel_sign_calls.rs
```diff
@@ -123,29 +71,44 @@ async fn mpc_cluster__should_successfully_process_mixed_parallel_requests() {
...
             json!({
                 "target_contract": cluster.contract.contract_id(),
-                "ecdsa_calls_by_domain": { "0": PER_TYPE },
-                "eddsa_calls_by_domain": { "1": PER_TYPE },
-                "ckd_calls_by_domain": { "2": PER_TYPE },
+                "ecdsa_calls_by_domain": { "1": ECDSA_CALLS },
+                "eddsa_calls_by_domain": { "2": EDDSA_CALLS },
+                "ckd_calls_by_domain": { "3": CKD_CALLS },
                 "seed": 43u64,
             }),
```
gilcu3, line 95: why this hack instead of just using one parallel call?

crates/e2e-tests/Cargo.toml
```diff
@@ -32,8 +32,9 @@ tracing = { workspace = true }
 [dev-dependencies]
 blstrs = { workspace = true }
 group = { workspace = true }
-sha3 = { workspace = true }
+mpc-contract = { workspace = true }
```
gilcu3, line 35: hopefully we can remove this soon

gilcu3 (commented):
Post merge review

Thanks for the fixes!

I am mostly worried about two things:
- 4 threads seems low, but thinking about it again it does depend on how many nodes our tests run, as each node should take ~1 core to possibly avoid contention.
- There is a "hack" in the test-parallel-requests tests that probably should be simpler to do, unless I missed something

.config/nextest.toml
anodar: > I dont think it affects production, the contention comes from running several tests at the same time.

That's correct.

> I think 4 is a very low number here (the runner has 32 cores), I presume you are using it to try to fix the general flakiness problem,

This is not a fix, coming up with right number based on some data and deflaking tests is tracked separately in  https://github.com/near/mpc/issues/2883 and https://github.com/near/mpc/issues/2883. 

This just allows me to re-enable disabled test to observe all tests running without spamming CI failures too much. This test being disabled may contribute in some other test passing that would otherwise failed so having this enabled gives more accurate picture.

It still has TODO on it as a comment.

crates/e2e-tests/tests/parallel_sign_calls.rs
anodar: That's how I had it initially but transaction was running out of gas (even with MAX_GAS). Comment above tries to explain that:

```
// when — fire 3 robust ECDSA + (2 ECDSA + 2 EdDSA + 2 CKD) across two concurrent
// transactions. All 9 in one tx would exhaust the 300 TGas per-transaction cap
// (9 × 30 TGas per sub-call + 10 TGas callback leaves no gas for the entrypoint).
  
```

gilcu3: I think the limit is no longer 300 TGas, we can already go to 1 PGas (= 1000TGas), might be useful here

crates/e2e-tests/tests/ckd_verification.rs
anodar: @gilcu3 is there an issue already for that or should I create one for tracking it ?

crates/e2e-tests/src/conversions.rs
anodar: Addressed in: https://github.com/near/mpc/pull/2922

crates/e2e-tests/tests/ckd_verification.rs
anodar: Makes sense, done in: https://github.com/near/mpc/pull/2922/changes

crates/e2e-tests/tests/parallel_sign_calls.rs
anodar: Done in: https://github.com/near/mpc/pull/2922/changes

crates/e2e-tests/tests/ckd_verification.rs
gilcu3: there are a lot of issues for it, no worries it wont be forgotten

Issue #2898, opened 2026-04-15
[Rust E2E] Fix flaky E2E tests


anodar:
### Background

- mpc_cluster__should_successfully_process_parallel_robust_ecdsa_requests is failing sometimes
e.g:

    - https://github.com/near/mpc/actions/runs/24448726670/job/71432180643?pr=2896 [logs](https://gist.github.com/anodar/19b6a337329c41a992900858815cfb11)
    - https://github.com/near/mpc/actions/runs/24446942608/job/71426098878?pr=2714 [logs](https://gist.github.com/anodar/963ad7f616a300a94d4c7e98269e40e6) 
     - https://github.com/near/mpc/actions/runs/24415859367/job/71325077493?pr=2870 [logs](https://gist.github.com/anodar/d51256f1c8e76e8f5f682453726ca415)

- ~web_endpoints~ (Should now be addressed in https://github.com/near/mpc/pull/2908)

    - This didn't happen in CI but I got this test failing locally once.  Reason is we don't differentiate between no samples (legitimate case) and failure.

- test_key_resharing

    - https://github.com/near/mpc/actions/runs/24498256339/job/71598574859 [logs](https://gist.github.com/anodar/1a4bcd0239ce0d542e827b7df1b1bc18)
    - https://github.com/near/mpc/actions/runs/24652881721/job/72079806071?pr=2954 [logs](https://gist.github.com/anodar/63820cf3c7a225969c6290e948487008)

- ~migration_service::migration_service__should_migrate_nodes_via_backup_cli~ (should now be addressed in https://github.com/near/mpc/pull/3056)
https://github.com/near/mpc/actions/runs/24887698860/job/72872072139 [logs](https://gist.github.com/anodar/5468404529f5831dc7480c531bf16819)

### Acceptance Criteria

No flaky tests

### Related pytest flakiness

We have related flakiness issue in pytests as well: https://github.com/near/mpc/issues/2898

rsmihir389:
Hey — I've been seeing this exact pattern a lot. Flaky E2E tests like this (especially parallel distributed system tests with timing sensitivity) usually cost teams $30-40 per rerun in CI time + developer context-switching.

I built a tool that detects flaky tests and shows the dollar cost: [kleore.com](https://www.kleore.com). Here's what a report looks like on a real repo: [sample report](https://www.kleore.com/report/cmn0knika00x7dc29p23yyha6).

60-second install as a GitHub App, free tier shows the cost. Might help here.

[...]

PR #2908, opened 2026-04-15
fix: web_server web_endpoints

anodar:
The e2e test was flaky — on a quiet/idle node during the 1-second sampling window, no CPU samples would be captured, producing an empty SVG and failing the assertion. The fix makes the server distinguish "no samples" from "real flamegraph" via status code, and teaches the test that "no samples" is a legitimate outcome.


DSharifi (approved):
Thank you

netrome (approved):
Thank you!

PR #2910, opened 2026-04-15
refactor: reduce trait derives on `contract-interface` types

### LINKED ISSUE #2899: Reduce trait derives on contract-interface types
### Background

The `near-mpc-contract-interface` crate is used by both the contract (compiled to WASM, where size matters) and the node (where additional traits like `Hash`, `Ord`, etc. may be needed for `BTreeMap`/`HashMap` keys).

Some types currently derive traits that are only needed by the node, increasing contract WASM size. Example: `KeyEventId` derives `Hash` (added in #2870) which is needed by the node but not the contract.

### Acceptance Criteria

- Audit trait derives on `near-mpc-contract-interface` types and identify those only needed by node consumers.
- Consider feature-gating node-only derives behind a feature flag (e.g. `node-extras`) that the node enables but the contract does not.

pbeza:
Remove unnecessary `Ord`, `PartialOrd`, and `Hash` derives from ~40 types in `near-mpc-contract-interface` that were never used as collection keys, reducing contract WASM size.

Add a `node-extras` feature flag that gates `Hash` on `DomainId` and `ParticipantId` — types only needed as `HashMap`/`HashSet` keys in the node.

Propagate the removal to downstream consumers: node task ID types, `foreign-chain-inspector`, and `near-mpc-sdk`.

Closes: #2899


pbeza:
Closing it based on what I explained in https://github.com/near/mpc/issues/2899#issuecomment-4253307647.

PR #2912, opened 2026-04-15
docs: design solution for allowing per node foreign tx rpc provider configurations

### LINKED ISSUE #2901: Design solution for allowing per-node foreign tx RPC provider configurations
### Background

We already started working on this in the context of https://github.com/near/mpc/issues/2648 - but to create a better venue for understanding and aligning on the new changes, let's write down the intended design and a migration plan and ensure we have full alignment on the new model before finishing the implementation.

### User Story

As an MPC stakeholder I'd like to have significant protocol changes documented.

### Acceptance Criteria

We have documentation of the intended design update to support updating foreign tx RPC providers without requiring full consensus of nodes.

_No response_

DSharifi:
closes #2901


docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
```diff
@@ -0,0 +1,101 @@
...
+    name: String,
+}
+```
+
+
+### Nodes checking if local configuration is valid
+On startup, the node will check its [local foreign chain configuration ](https://github.com/near/mpc/blob/4b2e758ee468738579af298f482aa13f9b5d269f/crates/node-config/src/foreign_chains.rs#L23-L37) to assert that all RPC providers are whitelisted on chain.
+
+If an RPC provider with a base url that is not whitelisted is detected, then that RPC entry should be dropped and an error log emitted.
+> We can alternatively hard crash the node, so it's obvious when a misconfiguration is done.
```
netrome: An error log should be sufficient. We should also add alerts on errors, and assume they don't go by unnoticed which they may today due to our logs still being noisy.

```suggestion
```

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
```diff
@@ -0,0 +1,101 @@
...
+
+Since the nodes are running in a Trusted Execution Environment (TEE), we use this functionality to have the node guard against node operators
+that might use malicious RPC urls.
+
+
+### Individual Node quorum of RPC providers for verification requests
+When a foreign TX verification request is processed by a set of nodes, every node will each individually query all their respectively configured
+RPC providers. A node will consider the foreign tx verified iff at least a quorum number of RPC providers could verify the transaction.
```
netrome: I'm considering if we should drop the "all" from this formulation leaving it explicitly undefined since it shouldn't be a problem if a node chooses a random subset if it has a lot of providers configured.

```suggestion
When a foreign TX verification request is processed by a set of nodes, every node will each individually query their respectively configured
RPC providers. A node will consider the foreign tx verified iff at least a quorum number of RPC providers could verify the transaction.
```

netrome (dismissed):
Very nice stuff - thanks for putting this together!

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
DSharifi: Perhaps a warning is better here to notify the configured RPC provider will be ignored. I don't think we should cause an alert if someone has a misconfiguration that can be detected on startup.

netrome (dismissed):

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
```diff
@@ -0,0 +1,101 @@
...
+## What
+As part of [#2648](https://github.com/near/mpc/issues/2648), we would like to be able to configure nodes with RPC providers without
+requiring all other nodes to have the exact same configuration.
+
+### Requirements
+1. RPC providers can be whitelisted by being voted into the contract by node operators submitting votes. This differs from the current solution where the node is the one to submit votes for RPC providers it has in its configuration.
+
+2. Nodes do not need to have local configurations with all RPC providers that are whitelisted, a quorum number of RPC providers is fine.
+
+3. Every node partaking in a foreign signature verification request will query all its locally configured RPC providers independently of other nodes. A quorum of the RPC providers must all agree on the verification.
```
anodar: I think there's 2 constants here that are referred with single one

1. Minimum number of RPC providers that each node should have
2. Minimum number of RPC providers that must agree within single node on the response

Those 2 constants don't need to be the same, e.g. we could require node should have at least 5 providers, but at least 4 of them should return the same response to be considered valid.

Otherwise if nodes are free to have them equal, we still have that risk that single invalid RPC provider will invalidate verification.

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
```diff
@@ -0,0 +1,101 @@
...
+
+struct RpcPolicies {
+    foreign_chains: BTreeMap<ForeignChain, RpcPolicy>,
+}
+
+struct RpcPolicy {
+    whitelisted_rpc_providers: BTreeSet<RpcProvider>,
+    // the minimum number of RPC providers that must verify the foreign
+    // tx request for a node to consider the tx as verified.
+    quorum_threshold: u8,
```
anodar, line 64: Expanding on above comment, it would me to also have another field e.g. `min_rpc_providers` which in practice would be at least quorum + 1.

anodar (changes requested):
I think you accidentally bumped nearcore.

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
DSharifi: It's a quite good suggestion to have this min requirement. But if we enforce this it's not going to help much in terms of availability if a node operator doesn't meet this minimum requirement, since we always need all nodes to meet the criteria, and this would lead to the chain not being supported alltogether.

I created a new task https://github.com/near/mpc/issues/2936 as a venue for discussing this idea further as it can be added as a follow up if we go for it.

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
DSharifi: See comment above

netrome (approved):

anodar (approved):
Thank you!

PR #2915, opened 2026-04-15
feat: foreign chain inspector supports base

### LINKED ISSUE #2797: Base support in foreign chain inspector
### Background

sub issue of #2792
foreign chain inspector crate supports Base chain

### User Story

_No response_

### Acceptance Criteria

foreign chain inspector crate supports Base chain

_No response_

DSharifi:
closes #2797

### DIFF
--- a/crates/foreign-chain-inspector/tests/base_rpc_manual.rs
+++ b/crates/foreign-chain-inspector/tests/base_rpc_manual.rs
@@ -15,7 +15,7 @@ async fn inspector_extracts_block_hash_against_live_rpc_provider() {
 
     let threshold = EthereumFinality::Finalized;
 
-    // Example transaction on Base mainnet (block 33554432) with 16 logs
+    // Example transaction on Base mainnet (block 33554432) with 16 logs;
     // https://basescan.org/tx/0xa11eaa1236e80f26ddc7aca164f2ba4c6c2726405cb12b1aa8f52c520bad99e1
     let transaction_id: BaseTransactionHash =
         "a11eaa1236e80f26ddc7aca164f2ba4c6c2726405cb12b1aa8f52c520bad99e1"
--- a/crates/foreign-chain-inspector/tests/base_rpc_manual.rs
+++ b/crates/foreign-chain-inspector/tests/base_rpc_manual.rs
@@ -1,0 +1,60 @@
+use assert_matches::assert_matches;
+use foreign_chain_inspector::{
+    EthereumFinality, ForeignChainInspector, RpcAuthentication,
+    base::{
+        BaseBlockHash, BaseTransactionHash,
+        inspector::{BaseExtractedValue, BaseExtractor, BaseInspector},
+    },
+};
+
+#[tokio::test]
+#[ignore = "manual test to sanity check against live Base RPC provider"]
+async fn inspector_extracts_block_hash_against_live_rpc_provider() {
+    // given
+    const BASE_RPC_URL: &str = "https://mainnet.base.org";
+
+    let threshold = EthereumFinality::Finalized;
+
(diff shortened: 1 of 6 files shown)


DSharifi:
@claude review

crates/foreign-chain-inspector/tests/base_rpc_manual.rs
```diff
@@ -0,0 +1,60 @@
...
+            threshold,
+            vec![
+                BaseExtractor::BlockHash,
+                BaseExtractor::Log { log_index: 0 },
+                BaseExtractor::Log { log_index: 1 },
+                BaseExtractor::Log { log_index: 2 },
+            ],
+        )
+        .await
+        .expect("extract should succeed");
```
gilcu3, line 49: closely related to this. I just ran tests with:
```
❯ cargo nextest run --cargo-profile=test-release --locked -p foreign-chain-inspector --run-ignored only
   Compiling httpmock v0.8.2
   Compiling foreign-chain-inspector v3.8.1 (/home/rey/opt/near/mpc/crates/foreign-chain-inspector)
    Finished `test-release` profile [optimized + debuginfo] target(s) in 11.51s
────────────
 Nextest run ID e01cfc70-ab1d-4992-9cfb-26375ee9120f with nextest profile: default
    Starting 5 tests across 10 binaries (68 tests skipped)
        PASS [   0.109s] foreign-chain-inspector::bnb_rpc_manual inspector_extracts_block_hash_against_live_rpc_provider
        PASS [   0.188s] foreign-chain-inspector::bitcoin_rpc_manual inspector_extracts_block_hash_against_live_rpc_provider
        PASS [   0.353s] foreign-chain-inspector::base_rpc_manual inspector_extracts_block_hash_against_live_rpc_provider
        PASS [   0.358s] foreign-chain-inspector::abstract_rpc_manual inspector_extracts_block_hash_against_live_rpc_provider
        FAIL [   0.407s] foreign-chain-inspector::starknet_rpc_manual inspector_extracts_block_hash_against_live_rpc_provider
```
failure is not this one, but it does show a couple of problems:
```
stderr ───
    thread 'inspector_extracts_block_hash_against_live_rpc_provider' panicked at crates/foreign-chain-inspector/tests/starknet_rpc_manual.rs:45:10:
    extract should succeed: TransactionFailed
```
so basically the error does not show exactly what happens, which is a pattern that should be fixed here, and also these tests never run, so if they stop working we wont notice. We already have optional tests that depends on external services, so we could add these manual tests to that check (optional).

That the error does not show what happens is a small blocker for me, and would be good to take this opportunity to fix it for the other manual tests as well (should be trivial)

gilcu3 (changes requested):
Thanks!

Small blocker on the error not showing the reason when the manual test fails, see the comment below.

Please also fix it for the other manual tests, as probably the same things happens there. An optional suggestion is to add these tests to the external services optional tests we already have, so that we are aware if they become stale.

gilcu3 (approved):
Approving after discussing sync

crates/foreign-chain-inspector/tests/base_rpc_manual.rs
DSharifi: Discussed sync. This is caused by a test not added in this PR.

Furthermore it's caused by poor error display implementation which is not introduced by this PR. https://github.com/near/mpc/blob/fe4e8d4cac5ab6e6878ba7f1248d86c44e543f05/crates/foreign-chain-inspector/src/lib.rs#L50-L71

kevindeforth (approved):

gilcu3:
@DSharifi removed from queue until I finish the release, else changelog will not be accurate

PR #2920, opened 2026-04-16
chore: Prepare for 3.9.0 release 

### LINKED ISSUE #2713: Make 3.9.0 release
### Background

Create 3.9.0 release.

### User Story

_No response_

### Acceptance Criteria

We've created 3.9.0 release

_No response_

gilcu3:
Closes #2713

### DIFF
--- a/.github/contract-size-baseline.toml
+++ b/.github/contract-size-baseline.toml
@@ -3,4 +3,4 @@
 max_transaction_size = 1572864
 
 # Expected contract WASM size in bytes (reproducible builds must match exactly)
-expected_size = 1481556
+expected_size = 1481703
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,164 @@ All notable changes to this project will be documented in this file.
 
 This changelog is maintained using [git-cliff](https://git-cliff.org/) and [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).
 
+## [3.9.0] - 2026-04-16
+
+### 🚀 Features
+
+- [#2727](https://github.com/near/mpc/pull/2727)(@barakeinav1): Add metrics for TEE attestation generation attempts (#2727)
+
+- [#2818](https://github.com/near/mpc/pull/2818)(@DSharifi): Support BNB chain in the contract (#2818)
+
+- [#2784](https://github.com/near/mpc/pull/2784)(@DSharifi): *(contract)* API to submit foreign chain config without RPC url consensus (#2784)
+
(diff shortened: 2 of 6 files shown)


scripts/check-contract-wasm-size.sh
```diff
@@ -62,10 +62,8 @@ fi
 
 # Baseline check
 if [[ "$DELTA" -gt 0 ]]; then
-    echo "❌ Contract grew by $DELTA bytes"
-    echo "   If this growth is intentional, update the baseline in $BASELINE_FILE:"
+    echo "❌ Contract grew by $DELTA bytes — consider updating the baseline in $BASELINE_FILE:"
     echo "   expected_size = $SIZE"
-    exit 1
 elif [[ "$DELTA" -lt 0 ]]; then
```
gilcu3, line 69: on the tangent, we should not need to update the value for tiny growth (here we should changed version). We already have a hard blocker in the script

gilcu3: sorry for making the fix in this PR, I am being pragmatic :)

DSharifi: CC: @pbeza  since you introduced this hard check.

I'm personally fine with this.

DSharifi (dismissed):
thanks

scripts/check-contract-wasm-size.sh
gilcu3: When I reviewed the PR introducing the script I missed this would break CI, I thought basically that only the hard upper bound was enforced

pbeza: Personally, I’m fine with updating the contract size whenever we change the contract — it’s just as tedious as manually updating a `cargo insta` snapshot. But I get that the contract size might change more often than the snapshot, so I’m fine with your change too. ✅

pbeza (dismissed):

scripts/check-contract-wasm-size.sh
gilcu3: > But I get that the contract size might change more often than the snapshot, so I’m fine with your change too. ✅

thanks! Even here that we only change the version number, it already grew by 60 bytes or something :cry:

netrome: Hmm I kind of like requiring an explicit change even for small increments - especially now when we're so close to the limit.

I'd prefer to discuss this separately at least. It doesn't make sense to have this check in the script in the first place if we don't fail on it, so I think we should tidy this up more if/when we remove it.

Also, why did the version bump impact contract size? 🤔

netrome (changes requested):
Let's do the wasm size check separately, or we should remove the "exact" check completely.

netrome (approved):
Thanks!

DSharifi (approved):

PR #2921, opened 2026-04-16
refactor: make StartConfig::quote_upload_url required

pbeza:
## Summary
- Add `quote_upload_url` field to `StartConfig` as a required `Url` with a serde default (Phala's public endpoint)
- Simplify `into_tee_authority` — the node config is now the single source of truth for the collateral endpoint URL, removing the override/fallback dance with the launcher's `TeeAuthorityConfig::Dstack`
- Enable `serde` feature on workspace `url` dependency

## Test plan
- [x] `cargo clippy -p mpc-node-config -p mpc-node -p e2e-tests --locked -- -D warnings` passes
- [ ] CI passes
- [ ] Existing TOML configs that omit `quote_upload_url` still work (serde default kicks in)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -215,7 +215,7 @@ tracing-subscriber = { version = "0.3.23", features = [
     "json",
 ] }
 tracing-test = "0.2.6"
-url = "2"
+url = { version = "2", features = ["serde"] }
 x509-parser = "0.18.1"
 zeroize = { version = "1.8.2", features = ["zeroize_derive"] }
 zstd = "0.13.3"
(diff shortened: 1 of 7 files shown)


barakeinav1:
Thanks for this — we're adopting your approach (required field with serde default) in #2893, along with additional cleanup: removing `quote_upload_url` from `TeeAuthorityConfig::Dstack` and `launcher-interface` so the launcher no longer writes a URL nobody reads. This is a good window to do it since the launcher isn't in production yet. Will close #2921 once #2893 is updated.

netrome:
@pbeza - this seems like a simple one, anything blocking us from pushing it through?

netrome:
Otherwise let's close it as stale.

pbeza:
> Otherwise let's close it as stale.

Closing this as stale. `main` has diverged quite a bit since I opened it, and it no longer seems relevant.

PR #2924, opened 2026-04-16
chore: leave some amount of contract increase without breaking CI

### LINKED ISSUE #2925: Contract WASM size check fails CI on any size increase
### Background

The contract WASM size check (`scripts/check-contract-wasm-size.sh`) currently fails CI when the contract grows by even a single byte over the baseline. This makes unrelated PRs fail when a legitimate feature increases the contract size, requiring each such PR to also bump the baseline.

PR #2924 proposes adding a 500-byte tolerance so small, intentional increases don't block CI, while still flagging them in the output.

### Acceptance Criteria

- The size check allows small contract increases (up to a defined threshold) without failing CI
- Increases within the threshold are still reported in the output so they're visible
- Increases beyond the threshold still fail CI with instructions to update the baseline

gilcu3:
Closes #2925

### DIFF
--- a/.github/contract-size-baseline.toml
+++ b/.github/contract-size-baseline.toml
@@ -1,6 +1,0 @@
-# NEAR max_transaction_size since protocol v69
-# https://github.com/near/nearcore/blob/master/core/parameters/res/runtime_configs/69.yaml
-max_transaction_size = 1572864
-
-# Expected contract WASM size in bytes (reproducible builds must match exactly)
-expected_size = 1481703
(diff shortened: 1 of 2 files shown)


pbeza (dismissed):

DSharifi (dismissed):

.github/contract-size-baseline.toml
DSharifi, line 1: Nit:
we could have kept this file or something similar for configuring the hard limit

gilcu3: It felt a bit pointless to keep it for a single variable (the second one was not used and became a comment)

DSharifi (approved):

pbeza (approved):

PR #2929, opened 2026-04-16
refactor: pre-build http clients to avoid duplication

### LINKED ISSUE #2867: node: `execute_foreign_chain_request` contains a lot of duplication for verifying foreign TXs
### Background

See discussion https://github.com/near/mpc/pull/2823#discussion_r3078656405

There is a lot of duplication when verifying foreign transactions inside `sign.rs` module on the node. We should look into generalizing the code to avoid so much duplication for every new chain we add

### User Story

_No response_

### Acceptance Criteria

reduced duplication of code on the match arms for verifying foreign TXs inside `sign.r`

_No response_

DSharifi:
closes #2867

### DIFF
--- a/crates/node/src/providers/verify_foreign_tx.rs
+++ b/crates/node/src/providers/verify_foreign_tx.rs
@@ -7,7 +7,11 @@ use crate::providers::{EcdsaSignatureProvider, SignatureProvider};
 use crate::storage::VerifyForeignTransactionRequestStorage;
 use crate::types::VerifyForeignTxId;
 use borsh::{BorshDeserialize, BorshSerialize};
+use foreign_chain_inspector::abstract_chain::inspector::AbstractInspector;
+use foreign_chain_inspector::bitcoin::inspector::BitcoinInspector;
+use foreign_chain_inspector::bnb::inspector::BnbInspector;
 use foreign_chain_inspector::http_client::HttpClient;
+use foreign_chain_inspector::starknet::inspector::StarknetInspector;
 use mpc_node_config::{ConfigFile, ForeignChainsConfig};
 use near_mpc_contract_interface::types as dtos;
 use std::sync::Arc;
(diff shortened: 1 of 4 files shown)


crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -399,6 +305,19 @@ async fn validate_foreign_chain_policy(
...
+    inspectors: &'a [T],
+    request_id: &CryptoHash,
+    my_participant_index: usize,
+    chain_name: &str,
+) -> anyhow::Result<&'a T> {
+    let index = select_provider(inspectors.len(), request_id, my_participant_index)
+        .with_context(|| format!("{chain_name} provider config is missing"))?;
+    Ok(&inspectors[index])
+}
+
```
DSharifi: This will be removed in a follow up, as nodes will no longer just query 1 RPC provider, but all of them in the new design.

crates/node/src/providers/verify_foreign_tx.rs
```diff
@@ -1,22 +1,68 @@
...
 
+/// Pre-built HTTP clients for each foreign chain, keyed in provider config order.
+///
+/// Built once at startup so that request handling only needs to select an index
+/// instead of re-parsing config and constructing clients on every call.
+pub(crate) struct ForeignChainInspectors<Client> {
+    pub bitcoin: Vec<BitcoinInspector<Client>>,
+    pub abstract_chain: Vec<AbstractInspector<Client>>,
+    pub bnb: Vec<BnbInspector<Client>>,
+    pub starknet: Vec<StarknetInspector<Client>>,
```
DSharifi, line 31: This will be changed in a follow up to contain only one inspector instead of a Vec of multiple one. The inspector crate will be extended to contain the logic of submitting the request to all clients.

gilcu3 (approved):
Nice!

pbeza (approved):

PR #2930, opened 2026-04-16
refactor: replace tee launcher hash votes with generics

kevindeforth:
This is a draft, to showcase how to use the abstract voting structs.
Follow-up to #2739 

Currently failing due to:
- exceeding contract size limit.
- one new TODO, questioning the existing use of `ParticipantId`. To be addressed before merge.

### DIFF
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -70,7 +70,6 @@ use primitives::{
     votes::types::{ProposalHash, ProposalId},
 };
 use tee::measurements::{ContractExpectedMeasurements, MeasurementVoteAction};
-use tee::proposal::CodeHashesVotes;
 
 use state::{running::RunningContractState, ProtocolContractState};
 use tee::{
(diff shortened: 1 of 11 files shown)


kevindeforth:
Looking at this, I think we can simplify a bit by getting rid of the `ProposalRegistry` and storing all votes with respect to the hash of the proposal (or, directly the proposal, in case it is small).

So, while currently we have:

`VotesRegistry`:
- `BTreeMap<Voter, ProposalId>`
- `BTreeMap<ProposalId, VoterSet<Voter>>`

`ProposalRegistry:`
-  `BTreeMap<ProposalHash, ProposalId>`
- `IterableMap<ProposalId, (ProposalHash, Proposal)>`.

It would probably be enough to simply use:
`VotesRegistry`:
- `BTreeMap<Voter, ProposalHash>`
- `BTreeMap<ProposalHash, VoterSet<Voter>>`

If the proposal is small or already a hash (such as for some proposals of the TEE state), we can store the proposal directly:
- `BTreeMap<Voter, Proopsal>`
- `BTreeMap<Proposal, VoterSet<Voter>>`

kevindeforth:
closing this in favor of https://github.com/near/mpc/pull/2935

PR #2931, opened 2026-04-16
refactor: remove node dependency on the contract for the crypto types

### LINKED ISSUE #2897: Node should depend on contract-interface, not mpc-contract, for crypto/protocol types
### Background

Follow-up to near/mpc#2870, which moves the participants boundary off of `mpc-contract`. Many other non-test usages of `mpc-contract` types remain in the node and should also be migrated to `contract-interface`.

One of the remaining parts is

**Signature / request types**
- `types.rs` — `DomainId`, `Payload`, `Tweak`
- `indexer/types.rs` — `CKDResponse`, `Tweak`, `Payload`
- `indexer/handler.rs` — `CKDRequest`, `Payload`, `SignRequest`, `SignRequestArgs`
- `mpc_client.rs` — `derive_tweak`, `CKDResponse`, `Curve`, `DomainId`
- `providers/*/sign.rs` (4 files) — `Tweak`, `Payload`, `Bytes`

### Acceptance Criteria

The node no longer depends on `mpc-contract` for the crypto types

gilcu3:
Closes #2897 

this is meant to be a pure refactor, moving types from interface/contract to `mpc-primitives` or `mpc-crypto-types` so that they can be reused without duplication


crates/devnet/src/contracts.rs
```diff
@@ -90,7 +86,7 @@ pub fn make_actions(call: ContractActionCall) -> ActionCall {
             actions: vec![make_action(
                 method_names::SIGN,
                 &serde_json::to_vec(&SignArgsV2 {
-                    request: SignRequestArgs {
+                    request: LegacySignRequestArgs {
```
gilcu3: These were actually the same type, although the previous had a very misleading name that got corrected here

crates/near-mpc-crypto-types/src/sign.rs
```diff
@@ -0,0 +1,80 @@
...
+use serde::{Deserialize, Serialize};
+use std::fmt;
+
+use crate::Payload;
+use mpc_primitives::domain::DomainId;
+
+/// Backward-compatible sign request args. Accepts both old field names
+/// (payload, key_version) and new (payload_v2, domain_id).
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct LegacySignRequestArgs {
```
DSharifi: This is type is not legacy, it's also used for the "new" v2 payloads.

I see your comment above that you renamed this from `SignRequestArgs` -> `LegacySignRequestArgs`.

Unfortunately this is the type we use in the `sign` method in the API of the contract, note that it contains both an optional `payload_v2` attribute and an optional `deprecated_payload`. This type is a mix of both that we have to live with.

crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -135,34 +135,7 @@ pub struct AuthenticatedAccountId(pub AccountId);
...
-)]
-pub enum Curve {
-    Secp256k1,
-    // Accepts "Ed25519" for compat with pre-3.9.0 contracts. Remove after 3.9.0 deployment.
-    #[serde(alias = "Ed25519")]
-    Edwards25519,
-    Bls12381,
-    /// Robust ECDSA variant.
-    V2Secp256k1,
-}
```
DSharifi, line 165: Any reason to remove the contract variant and move this interface variant to the primitives crate?

DSharifi (changes requested):
Thanks for the cleanup looks quite good.

I have one concern with the renaming of the type `SignRequestArgs` to `LegacySignRequestArgs`

crates/near-mpc-contract-interface/src/types/state.rs
gilcu3: just to avoid duplication

crates/near-mpc-crypto-types/src/sign.rs
gilcu3: let me check my logic again, as the problem is we have two SignRequestArgs types and maybe I got them wrong, will get back to this

gilcu3: see [thread](https://nearone.slack.com/archives/C0912BTG51T/p1776408291518099?thread_ts=1776357498.754219&cid=C0912BTG51T) for context, I had a duplicate type with a misleading name that will need to be handled, will fix

crates/near-mpc-crypto-types/src/sign.rs
```diff
@@ -0,0 +1,436 @@
...
+#[derive(Deserialize)]
+struct SignRequestArgsCompat {
+    path: String,
+    payload_v2: Option<Payload>,
+    #[serde(rename = "payload")]
+    deprecated_payload: Option<[u8; 32]>,
+    domain_id: Option<DomainId>,
+    #[serde(rename = "key_version")]
+    deprecated_key_version: Option<u32>,
+}
```
gilcu3, line 46: added exactly the same existing struct to guarantee that no logic has changed

crates/near-mpc-crypto-types/src/snapshots/near_mpc_crypto_types__kdf__tests__derive_app_id__has_not_changed.snap
gilcu3, line 1: the larger change is just a snapshot :( we should find a way to make serialization of these nicer

DSharifi (approved):

crates/near-mpc-crypto-types/src/kdf.rs
```diff
@@ -0,0 +1,93 @@
...
+    hash.into()
+}
+
+fn derive_from_path(derivation_prefix: &str, predecessor_id: &AccountId, path: &str) -> [u8; 32] {
+    // TODO: Use a key derivation library instead of doing this manually.
+    // https://crates.io/crates/hkdf might be a good option?
+    //
+    // ',' is ACCOUNT_DATA_SEPARATOR from nearcore that indicate the end
+    // of the account id in the trie key. We reuse the same constant to
+    // indicate the end of the account id in derivation path.
```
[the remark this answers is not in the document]
gilcu3: true, but existing code, so will skip if no further changes are needed

crates/near-mpc-crypto-types/src/sign.rs
```diff
@@ -0,0 +1,436 @@
...
+#[cfg_attr(
+    all(feature = "abi", not(target_arch = "wasm32")),
+    derive(schemars::JsonSchema)
+)]
+pub struct SignRequestArgs {
+    pub path: String,
+    #[serde(rename = "payload_v2")]
+    pub payload: Payload,
+    pub domain_id: DomainId,
+}
```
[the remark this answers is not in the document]
gilcu3: that would be really cumbersome, and increase the abi size. New customers should just use the new format IMO

pbeza (approved):

Issue #1674, opened 2025-12-16
Contract: Vote for hash and separate upgrades from voting


netrome:
### Background

Currently contract upgrades happen automatically once enough votes have passed. However, since upgrading is a much more expensive operation than voting this means it can be very hard to predict the gas costs when making a vote.

It would be natural to separate the processes of voting for an upgrade and actually performing the upgrade. One benefit of this is also that the contract doesn't need to hold the full bytecode of the upgraded contract while votes are pending. Instead, people could vote for the hash of a contract, and once a hash of the contract has reached enough votes, anyone can submit the actual contract bytecode and perform the upgrade.

### User Story

As a node operator, I'd like to be able to vote for a contract upgrade without risking actually performing the upgrade, so that the operation can have more predictable gas costs.

### Acceptance Criteria

Voting for a contract upgrade and performing the upgrade happens in separate contract calls.

netrome:
Still relevant

kevindeforth:
different, but similar: https://github.com/near/intents/tree/5bb4fc83544ce88e36f4d77753bb7da156abe4e4/contracts/global-deployer

kevindeforth:
So, I had this as a branch for a while now and tried to reduce diff size, but it's always huge. Like +1k, -2k lines of code.
I will split this into multiple PRs to improve review experience.

PR #2934, opened 2026-04-17
feat(contract): vote generics structs

### LINKED ISSUE #1573: contract: abstract voting struct

kevindeforth:
resolves #1573 

The main differences to #2739 are:
- we do not keep an instance of the proposals in the contract state
- we rely on the proposal hash as an identifier (meaning, there is no `ProopsalId` here)
- we only have two maps instead of four and for both of these, we are now using `IterableMap` to keep gas costs low.

The main advantage is space saving: the contract binary size is smaller and we store less state on chain (c.f. https://github.com/near/mpc/issues/1674).

The main disadvantage is that viewing contract state is not human-friendly. For example, if someone queried the contract for all resharing proposals, they would no longer see a json serialization of the proposed participant set, but just a 32 byte hash.

For an example usage, c.f. https://github.com/near/mpc/pull/2935


netrome:
> For example, if someone queried the contract for all resharing proposals, they would no longer see a json serialization of the proposed participant set, but just a 32 byte hash.

To mitigate this, we should store and serve this information somewhere else. Either in the nodes or as a separate service. I'd picture the flow to be something like:

1. Post a proposal on-chain.
2. Post the proposal to the third-party service.
3. The third-party service verifies this proposal against the on-chain state.
4. Anyone can query the service for the proposal.

crates/contract/src/primitives/votes.rs
```diff
@@ -1,463 +1,607 @@
...
-        let votes = self
-            .vote_for(voter, proposal_id)
-            .expect("proposal id must exist");
-        (proposal_id, votes)
-    }
-
-    /// Registers a vote by `voter` for proposal with `proposal_id`
-    pub fn vote_for(
+    /// Returns the [`ProposalHash`] and votes for proposal.
+    pub fn vote<P: ProposalHashEncoding>(
```
netrome: I think we could simplify this further by just taking the hash as input directly. From what I can see the contract doesn't store anything beyond the actual hash anyway so I don't see a strong reason for taking the full proposal as input.

In either case I assume we need a separate step where the information in the proposal is revealed to conclude a vote. For example when we upgrade the participant set, once the voting is completed the contract does need the participant set to actually upgrade it once the vote has passed [^1].

What do you think?

[^1] Note that this is a feature we want to add for upgrades anyway https://github.com/near/mpc/issues/1674

netrome (commented):
Skimmed this, I think I understand the idea and I believe we could simplify this even further. Curious to hear your thoughts.

crates/contract/src/primitives/votes.rs
kevindeforth: > Note that this is a feature we want to add for upgrades anyway https://github.com/near/mpc/issues/1674

I was searching for that issue! Thanks!

> I think we could simplify this further by just taking the hash as input directly. From what I can see the contract doesn't store anything beyond the actual hash anyway so I don't see a strong reason for taking the full proposal as input.

I agree with this. `vote` is just a thin wrapper around one call to `Into<ProposalHash>` and one call to  `register`. This can be inlined if we want our API to take the full proposal when voting.

```rust
pub fn vote<P: ProposalHashEncoding>(
        &mut self,
        voter: V,
        proposal: P,
    ) -> (ProposalHash, &VoterSet<V>) {
        let proposal_hash = proposal.into();
        let votes = self.register(voter, proposal_hash);
        (proposal_hash, votes)
    }
```

crates/contract/src/primitives/votes.rs
```diff
@@ -1,135 +1,554 @@
...
+
+use near_sdk::near;
+use near_sdk::require;
+use near_sdk::store::IterableMap;
+use near_sdk::IntoStorageKey;
+
+/// Helper struct to keep track of submitted votes.
+/// Allows efficient look-up of votes by voter and votes by proposal.
+#[near(serializers=[borsh])]
+pub struct Votes<V>
```
pbeza, line 14: `V` is a bit confusing here — I keep reading it as `Value` (like in `HashMap<K, V>`) or `Vote` (the proposal). Maybe `Voter` instead? Feels clearer without having to look at the code below.

crates/contract/src/primitives/votes.rs
```diff
@@ -1,135 +1,553 @@
...
+where
+    V: BorshSerialize + Ord + BorshDeserialize + Clone,
+{
+    pub fn new(
+        proposal_by_voter: impl IntoStorageKey,
+        votes_by_proposal: impl IntoStorageKey,
+    ) -> Self {
+        Self {
+            proposal_by_voter: IterableMap::new(proposal_by_voter),
+            votes_by_proposal: IterableMap::new(votes_by_proposal),
```
pbeza, line 32: Curious what the minimum number of proposals/votes is where it becomes worth having both `proposal_by_voter` and `votes_by_proposal`, instead of just keeping `proposal_by_voter` and reconstructing `votes_by_proposal` when needed.

crates/contract/src/primitives/votes.rs
```diff
@@ -1,135 +1,553 @@
...
+
+    /// Removes any votes by voter V.
+    pub fn remove_vote(&mut self, voter: &V) {
+        let Some(proposal) = self.proposal_by_voter.get(voter) else {
+            return;
+        };
+        let casted_votes = self
+            .votes_by_proposal
+            .get_mut(proposal)
+            .expect("inconsistent votes registry");
```
pbeza: I was going to suggest making the `.expect()` messages different, but looks like they already are:

```
.expect("inconsistent voter registry")
```

vs

```
.expect("inconsistent votes registry")
```

vs

```
.expect("inconsistent vote registry")
```

😆

crates/contract/src/primitives/votes.rs
```diff
@@ -1,135 +1,553 @@
...
-    pub(crate) proposal_by_account: BTreeMap<AuthenticatedAccountId, ThresholdParameters>,
+use borsh::{BorshDeserialize, BorshSerialize};
+
+use near_sdk::near;
+use near_sdk::require;
+use near_sdk::store::IterableMap;
+use near_sdk::IntoStorageKey;
+
+/// Helper struct to keep track of submitted votes.
+/// Allows efficient look-up of votes by voter and votes by proposal.
```
pbeza, line 12: Might be worth stating this explicitly so devs don’t have to dig into the impl — each voter has only one active vote, and casting a new one just replaces the previous one. Although you already mention it here, so it’s optional:

https://github.com/near/mpc/blob/54c58109e9908222054727dd038c29bbc7028cdc/crates/contract/src/primitives/votes.rs#L36-L38

crates/contract/src/primitives/votes.rs
```diff
@@ -1,135 +1,553 @@
...
+        if let Some(existing_vote) = self.proposal_by_voter.get(&voter) {
+            // if the voter has an existing vote for a different proposal, remove that vote.
+            if *existing_vote != proposal {
+                self.remove_vote(&voter)
+            } else {
+                // voter has already voted for this proposal, just return the current voter set
+                return self
+                    .votes_by_proposal
+                    .get(&proposal)
+                    .expect("require consistent vote registry");
```
pbeza: When voter re-votes for the same proposal, we don't touch storage. This is fine if `V` is simple, but `V` is generic — and `BTreeSet` / `IterableMap` always keep the original key on equality match, not the new one. So if someone in the future uses a `V` where `Ord` ignores some fields but `Borsh` stores them, those fields will silently freeze to the values from the first insert.

Not a real bug today (our only caller's `Ord` matches its bytes exactly), but the generic does not enforce this.  A test showing the repro:

```rust

    /// Demonstrates that `Votes<V>` silently drops non-`Ord` fields of `V` when
    /// a voter re-votes for the same proposal. When a voter casts a vote, then
    /// casts a second vote for the same proposal with a value of `V` that
    /// compares `Ord::Equal` to the first but carries different payload bytes,
    /// the stored `V` is the first insertion's bytes. The newer payload is
    /// silently discarded because `vote()` early-returns for same-proposal
    /// re-votes without going through `remove_vote`, so the stored bytes are
    /// never refreshed
    #[test]
    #[expect(non_snake_case)]
    fn vote__silently_drops_non_ord_fields_when_revoting_same_proposal() {
        #[derive(Clone, Debug, BorshSerialize, BorshDeserialize)]
        struct PartiallyOrderedVoter {
            id: String,
            payload: u64,
        }

        impl PartialEq for PartiallyOrderedVoter {
            fn eq(&self, other: &Self) -> bool {
                self.id == other.id
            }
        }
        impl Eq for PartiallyOrderedVoter {}

        impl PartialOrd for PartiallyOrderedVoter {
            fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
                Some(self.cmp(other))
            }
        }
        impl Ord for PartiallyOrderedVoter {
            fn cmp(&self, other: &Self) -> std::cmp::Ordering {
                self.id.cmp(&other.id)
            }
        }

        // Given: voter type where `Ord` ignores `payload` but Borsh includes it
        let mut votes_registry = Votes::<PartiallyOrderedVoter>::new(
            TestStorageKey::ProposalByVoter,
            TestStorageKey::VotesByProposal,
        );

        let proposal = make_proposal_hash(1);
        let v1 = PartiallyOrderedVoter {
            id: "alice".into(),
            payload: 111,
        };
        let v2 = PartiallyOrderedVoter {
            id: "alice".into(),
            payload: 222,
        };

        assert_eq!(v1, v2);
        assert_ne!(v1.payload, v2.payload);

        // When: alice votes with v1, then re-votes the same proposal with v2
        votes_registry.vote(v1.clone(), proposal);
        votes_registry.vote(v2.clone(), proposal);

        // Then: `all()` returns v1's payload (111), not v2's (222)
        let all = votes_registry.all();
        let stored_voters = all.get(&proposal).expect("proposal should exist");
        let stored_voter = stored_voters.iter().next().expect("one voter");

        assert_eq!(stored_voter.id, "alice");
        assert_eq!(
            stored_voter.payload, 111,
            "BUG: payload should be 222 (last write), but is 111 (first write)"
        );
    }
```

We should probably either document it, tighten the trait bound, or always refresh the bytes on re-vote.

crates/contract/src/primitives/votes.rs
```diff
@@ -1,135 +1,553 @@
...
+    V: Ord,
+{
+    pub(super) fn new() -> Self {
+        Self(BTreeSet::new())
+    }
+
+    // counts all the votes for which `predicate` returns true
+    pub fn count_for(&self, predicate: impl Fn(&V) -> bool) -> usize {
+        self.0.iter().filter(|voter| predicate(voter)).count()
+    }
```
pbeza, line 159: It’s only used in tests right now — should we gate it with `#[cfg(test)]`?

crates/contract/src/primitives/votes.rs
```diff
@@ -1,135 +1,553 @@
...
+    static ALICE: LazyLock<TestVoter> = LazyLock::new(|| TestVoter("alice".to_string()));
+    static BOB: LazyLock<TestVoter> = LazyLock::new(|| TestVoter("bob".to_string()));
+
+    #[test]
+    #[expect(non_snake_case)]
+    fn votes_new__should_be_empty() {
+        let votes_registry = setup();
+        assert_eq!(votes_registry.all(), BTreeMap::new());
+        assert_eq!(votes_registry.proposal_by_voter.len(), 0);
+        assert_eq!(votes_registry.votes_by_proposal.len(), 0);
```
pbeza, line 236:
```suggestion
        assert!(votes_registry.proposal_by_voter.is_empty());
        assert!(votes_registry.votes_by_proposal.is_empty());
```

pbeza (changes requested):
One [blocker](https://github.com/near/mpc/pull/2934#discussion_r3110076118) and a couple of optional nits.

crates/contract/src/primitives/votes.rs
kevindeforth: no, we will use it in production _(note: all code added in this PR is only used in tests right now)_

crates/contract/src/primitives/votes.rs
kevindeforth: I think `assert_eq!(votes_registry.proposal_by_voter.len(), 0);` is more informative when it fails, as it will indicate the number of elements in the map.

crates/contract/src/primitives/votes.rs
kevindeforth: Do we have a note in a style-guide for this? @netrome mentioned in a different PR that he prefers single-letter generics.

pbeza: I don’t think we have. It’s your call how to name it, but I think the Rust community is generally fine with more meaningful names. Anyway, not a blocker.

crates/contract/src/primitives/votes.rs
pbeza: Make sense. 👍🏼

crates/contract/src/primitives/votes.rs
kevindeforth: That's an interesting observation.
Will change the behavior to always remove any existing vote before inserting a new one.
https://github.com/near/mpc/pull/2934/commits/4b6b5073982b7f544126fcb836a84455d95eefbc

pbeza: It’s a bit of an exotic edge case—typically `Ord` takes all fields of a struct into account. I’m fine with just documenting it so it doesn’t surprise anyone using it. If you’d rather always remove the existing vote before inserting a new one, I’m good with that too.

crates/contract/src/primitives/votes.rs
kevindeforth: https://github.com/near/mpc/pull/2934/commits/6fa6b1f162574c4fe4f2bf6aec087a17baa389f3

crates/contract/src/primitives/votes.rs
pbeza: @kevindeforth I think the current implementation aligns with standard Rust behavior: https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.insert

Quote: 
> If the map did have this key present, the value is updated, and the old value is returned. **The key is not updated, though**; this matters for types that can be == without being identical. 

So I don’t think anything needs to be changed here—just adding a similar comment would be nice. Sorry for the confusion.

kevindeforth: It's definitely constructed and I think we agree that the `PartialEq` in the example implementation is shady.

But I thought it was an interesting and valid concern, especially because we found some legacy code like that today (c.f. https://github.com/near/mpc/pull/2954#issuecomment-4278539494).

Also, the proposed fix is rather easy and removes a few lines of code. I don't think we need to optimize for gas here - if the caller cares, they can refrain from sending the same vote twice.

pbeza (dismissed):

crates/contract/src/primitives/threshold_votes.rs
```diff
@@ -0,0 +1,137 @@
...
+        self.proposal_by_account
+            .iter()
+            .filter(|&(acc, prop)| {
+                participants
+                    .participants()
+                    .iter()
+                    .any(|(acc_id, _, _)| acc.get() == acc_id)
+                    && prop == proposal
+            })
+            .count() as u64
```
netrome: Small but important - let's be explicit about the intended behavior of type conversions. This one should never fail on wasm32 so an explicit try from + expect would be preferable imo.

```suggestion
        u64::try_from(
            self.proposal_by_account
                .iter()
                .filter(|&(acc, prop)| {
                    participants
                        .participants()
                        .iter()
                        .any(|(acc_id, _, _)| acc.get() == acc_id)
                        && prop == proposal
                })
                .count(),
        )
        .expect("usize should never fail to convert to u64 on wasm32")
```

crates/contract/src/primitives/threshold_votes.rs
```diff
@@ -0,0 +1,137 @@
...
+            .proposal_by_account
+            .insert(participant, proposal.clone())
+            .is_some()
+        {
+            log!("removed one vote for signer");
+        }
+        self.proposal_by_account
+            .values()
+            .filter(|&prop| prop == proposal)
+            .count() as u64
```
netrome: This one should also be explicit. If you feel strongly about the panic-overhead I'm okay with a safety comment as well.

crates/contract/src/primitives/votes.rs
netrome: Yeah would be fun to benchmark this just to learn more. I bet it would be pretty different when running in a smart contract on-chain vs locally.

crates/contract/src/primitives/votes.rs
```diff
@@ -1,135 +1,542 @@
...
+        let mut bytes = [0u8; 32];
+        bytes[24..].copy_from_slice(&i.to_be_bytes());
+        ProposalHash::new(bytes)
+    }
+
+    static ALICE: LazyLock<TestVoter> = LazyLock::new(|| TestVoter("alice".to_string()));
+    static BOB: LazyLock<TestVoter> = LazyLock::new(|| TestVoter("bob".to_string()));
+
+    #[test]
+    #[expect(non_snake_case)]
```
netrome, line 231: Nice! I'm always using `#[allow(non_snake_case)]` but expect is better. I'll start doing this instead.

crates/contract/src/primitives/votes.rs
netrome: > @netrome mentioned in a different PR that he prefers single-letter generics.

Depends on the context. Would love if you could reference the PR. I would guess it was a struct like:

```rust
struct GenericThingamajig<Spam, Eggs> {
    spam: Spam,
    eggs: Eggs,
}
```

in which case single-letter generics definitely reads better
```rust
struct GenericThingamajig<S, E> {
    spam: S,
    eggs: E,
}
```

But in this case, I agree it's less obvious so it would be nice to convey some information in the generic declaration.

crates/contract/src/primitives/votes.rs
```diff
@@ -1,135 +1,542 @@
...
+
+/// Helper struct to keep track of submitted votes.
+/// Allows efficient look-up of votes by voter and votes by proposal.
+#[near(serializers=[borsh])]
+pub struct Votes<V>
+where
+    V: BorshSerialize + Ord,
+{
+    proposal_by_voter: IterableMap<V, ProposalHash>,
+    votes_by_proposal: IterableMap<ProposalHash, VoterSet<V>>,
```
netrome, line 19: Now as we touched on in the meeting today, now we have the full "voter" struct in both maps, both as key and value.

In practice, do we ever want to track a complex data structure here? It feels like we just want to keep track of voter identifiers right? So we could just call this `VoterId` perhaps?

Or we could even simplify this further by converging on using the concrete `AccountId` type for all votes. While we use `ParticipantId` for most voting flows, I don't see why we couldn't change this to converge on `AccountId` everywhere.

crates/contract/src/primitives/threshold_votes.rs
netrome: Noticed this is in moved code, so not a blocker.

netrome (dismissed):
Generally looks good to me. Shared some thoughts on the voting struct. I wonder if we can simplify the voter type further. I don't see any instance where this isn't either a `ParticipantId` or an `AccountId` (or `AuthenticatedAccountId`) - would love to hear your thoughts on this.

crates/contract/src/primitives/votes.rs
kevindeforth: Answered on slack. I am having a hard time interpolating a guideline for our style-guide based on these comments.

crates/contract/src/primitives/votes.rs
kevindeforth: > by converging on using the concrete `AccountId` type for all votes

Space-wise, `ParticipantId` is preferred, because it's a simple `u64`, whereas `AccountId` is a String.
We can safely use `ParticipantId` whenever the vote does not cross epoch boundaries (e.g. when voting for new `ThresholdParameters`).
But when the vote _does_ cross epoch boundaries, we start relying on invariants, which may not be ideal (c.f. https://github.com/near/mpc/issues/2955).

So, for now, I wouldn't want to force `AccountId` here.

> In practice, do we ever want to track a complex data structure here?

No, we don't plan on storing complex `Voter` structures and this isn't the most optimal struct to do so.

>  It feels like we just want to keep track of voter identifiers right? So we could just call this `VoterId` perhaps?

Are you suggesting to rename the generic `V` as `VoterId`?

crates/contract/src/primitives/threshold_votes.rs
kevindeforth: broken window fix: https://github.com/near/mpc/pull/2934/commits/37df95f134b92141773fa2e703c1a0eebfee6b75

crates/contract/src/primitives/votes.rs
kevindeforth: ~it's moved code, but added a broken window fix: https://github.com/near/mpc/pull/2934/commits/37df95f134b92141773fa2e703c1a0eebfee6b75~

edit: oops, wanted to respond to https://github.com/near/mpc/pull/2934#discussion_r3113636300

netrome (approved):

crates/contract/src/primitives/threshold_votes.rs
kevindeforth: it's moved code, but added a broken window fix: https://github.com/near/mpc/pull/2934/commits/37df95f134b92141773fa2e703c1a0eebfee6b75

pbeza (approved):

Issue #2889, opened 2026-04-15
test: port migration endpoint, foreign chain policy unanimity, lagging node cleanup, migration service, and foreign transaction validation from pytest to Rust e2e


anodar:
### Background

The MPC project is migrating its test suite from pytest to Rust-based e2e tests as part of the pytest deprecation effort. This task covers porting five test groups: `test_migration_endpoint` (migration HTTP debug endpoint), `test_foreign_chain_policy_auto_voting_requires_unanimity` (unanimity requirement for policy voting), `test_cleanup_lagging_node` (asset cleanup when a node falls behind in block ingestion), `test_migration_service` (full backup CLI migration workflow), and `test_foreign_transaction_validation` (Bitcoin, Abstract/EVM, and Starknet foreign transaction verification).

### Acceptance Criteria

- [x] Port `test_migration_endpoint`: verify `/debug/migrations` endpoint tracks migration state
  - Initial migration state is empty
  - After setting backup service info on a node, migration state updates accordingly
  - After initiating node migration, migration state includes destination node info
  - Contract and endpoint migration state remain synchronized

- [x] Port `test_foreign_chain_policy_auto_voting_requires_unanimity`: unanimity requirement for foreign chain policy
  - 3-node cluster; nodes 0 and 1 configured with identical foreign chains config (e.g. Solana), node 2 has none
  - After 2 nodes vote, verify votes appear in proposals but policy is NOT applied to the contract
  - After the third node votes (unanimity achieved), verify policy is applied
  - Once applied, verify all proposal votes are cleared

- [x] Port `test_cleanup_lagging_node`: asset cleanup when a node falls behind in block ingestion
  - 3-node cluster with presignatures buffered; disable block ingestion on one node
  - Verify indexer lag reaches `INDEXER_MAX_HEIGHT_DIFF` threshold
  - Verify offline presignatures go to 0 (all cleaned up)
  - Verify online presignatures remain at buffer amount
  - Verify nodes can still handle signature requests after cleanup

- [x] Port `test_migration_service`: full end-to-end node migration via backup CLI
  - 2 participating nodes + 2 target nodes; compile backup CLI
  - Backup service registration on participating nodes
  - GET keyshares from source node via backup service
  - Initiate node migration
  - PUT keyshares to target node via backup service
  - Verify target node becomes active participant (contract state updated)
  - Verify migration state clears from contract
  - Kill old node; verify new node handles sign requests successfully

[...]

PR #2946, opened 2026-04-17
test: port test_cleanup_lagging_node from pytest to Rust e2e

anodar:
Port the lagging node asset cleanup test. Verifies that when a node falls behind in block ingestion past the threshold, other nodes clean up all presignatures involving it and can still handle sign requests.

Renaming `listen_blocks` → `listen_blocks.flag` is a bugfix for the existing test utility, not just incidental to the new test. The node reads [listen_blocks.flag](https://github.com/near/mpc/blob/5614c68fa12159a1e315351567c30a24fe4a23bf/crates/node/src/config.rs#L429), so any previous e2e test that called set_block_ingestion would have written to the wrong path and silently had no effect.

Work towards: #2889

### DIFF
--- a/crates/e2e-tests/src/mpc_node.rs
+++ b/crates/e2e-tests/src/mpc_node.rs
@@ -19,7 +19,7 @@ use crate::port_allocator::E2ePortAllocator;
 const DUMMY_IMAGE_HASH: &str =
     "sha256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef";
 
-const LISTEN_BLOCKS_FILE: &str = "listen_blocks";
+const LISTEN_BLOCKS_FILE: &str = "listen_blocks.flag";
 
 const TEMP_KEYS_FILE: &str = "temporary_keys";
 
--- a/crates/e2e-tests/tests/cleanup_lagging_node.rs
+++ b/crates/e2e-tests/tests/cleanup_lagging_node.rs
@@ -20,12 +20,6 @@ async fn cleanup_lagging_node_should_purge_offline_presignatures_and_keep_signin
     let (cluster, running) =
         common::setup_cluster(common::CLEANUP_LAGGING_NODE_PORT_SEED, |_| {}).await;
 
-    assert_eq!(cluster.nodes.len(), 3, "expected 3 nodes");
-    assert!(
-        !running.domains.domains.is_empty(),
-        "expected at least one domain"
-    );
-
     // Wait for all nodes to have presignatures buffered.
     common::wait_for_presignatures(&cluster, &[0, 1, 2], DEFAULT_PRESIGNATURES_TO_BUFFER).await;
 
@@ -109,9 +103,4 @@ async fn cleanup_lagging_node_should_purge_offline_presignatures_and_keep_signin
             );
         }
     }
-
-    // Re-enable block ingestion.
-    cluster
-        .set_block_ingestion(&[faulty_node_idx], true)
-        .expect("failed to re-enable block ingestion");
 }
--- a/crates/e2e-tests/tests/cleanup_lagging_node.rs
+++ b/crates/e2e-tests/tests/cleanup_lagging_node.rs
@@ -1,0 +1,117 @@
+use crate::common;
+
+use backon::Retryable;
+use e2e_tests::{CLUSTER_WAIT_TIMEOUT, DEFAULT_PRESIGNATURES_TO_BUFFER, metrics};
+use near_mpc_contract_interface::types::{Curve, DomainPurpose};
+use rand::SeedableRng;
+
+/// Maximum block delay before an MPC node is considered offline by its peers.
+/// Must match `MAX_HEIGHT_DIFF` in `crates/node/src/network.rs`.
+const INDEXER_MAX_HEIGHT_DIFF: i64 = 50;
+
+/// Verify that when a node falls behind in block ingestion past the
+/// `INDEXER_MAX_HEIGHT_DIFF` threshold, the other nodes clean up all
+/// presignatures that involved it (offline presignatures go to 0),
+/// retain their own presignatures, and can still handle signature requests.
+#[tokio::test]
+async fn cleanup_lagging_node_should_purge_offline_presignatures_and_keep_signing() {
+    // given
+    let mut rng = rand::rngs::StdRng::seed_from_u64(0);
+    let (cluster, running) =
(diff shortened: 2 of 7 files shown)


anodar:
@claude review

crates/e2e-tests/src/mpc_node.rs
```diff
@@ -19,7 +19,7 @@ use crate::port_allocator::E2ePortAllocator;
 const DUMMY_IMAGE_HASH: &str =
     "sha256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef";
 
-const LISTEN_BLOCKS_FILE: &str = "listen_blocks";
+const LISTEN_BLOCKS_FILE: &str = "listen_blocks.flag";
```
barakeinav1, line 22: nit: Worth calling out in the PR description that this rename (`listen_blocks` → `listen_blocks.flag`) is a bugfix for the existing test utility, not just incidental to the new test. The node reads `listen_blocks.flag` (see `crates/node/src/config.rs:429`), so any previous e2e test that called `set_block_ingestion` would have written to the wrong path and silently had no effect.

crates/e2e-tests/tests/cleanup_lagging_node.rs
```diff
@@ -0,0 +1,107 @@
...
+                heights[idx].unwrap_or(0) >= faulty + INDEXER_MAX_HEIGHT_DIFF,
+                "node {idx} not yet {} blocks ahead of faulty node (alive={:?}, faulty={faulty})",
+                INDEXER_MAX_HEIGHT_DIFF,
+                heights[idx],
+            );
+        }
+        Ok(())
+    })
+    .retry(
+        backon::ConstantBuilder::default()
```
barakeinav1: nit: The inline `backon::Retryable` block here is a custom poll loop that compares heights across multiple nodes — it's more elaborate than the existing `wait_metric_on_nodes` helper in `common.rs` because the condition depends on a dynamic comparison between the faulty node and the alive ones. Consider extracting to a `wait_for_indexer_lag` helper in `common.rs` if another test needs the same pattern. Not worth doing for a single caller.

crates/e2e-tests/tests/cleanup_lagging_node.rs
```diff
@@ -0,0 +1,107 @@
...
+        CLUSTER_WAIT_TIMEOUT,
+    )
+    .await;
+
+    // Verify nodes can still handle signature requests by depleting their asset stores.
+    if let Some(domain) = running
+        .domains
+        .domains
+        .iter()
+        .find(|d| matches!(d.purpose, DomainPurpose::Sign))
```
barakeinav1, line 67: If no domain matches `DomainPurpose::Sign` (or if only `Bls12381` domains exist — the inner `match` falls through to `break`), the `if let Some(domain) = ...` branch is skipped entirely and the test passes without exercising the "still can sign" assertion. The pytest original guards with `assert len(contract.get_running_domains()) \u003e 0`; consider adding an assertion here that at least one signable domain was found, so a misconfigured cluster setup can't silently skip the most important part of the test.

barakeinav1 (commented):
Nice port. Three non-blocking observations below.

crates/e2e-tests/tests/cleanup_lagging_node.rs
```diff
@@ -0,0 +1,107 @@
+use crate::common;
+
+use backon::Retryable;
+use e2e_tests::{CLUSTER_WAIT_TIMEOUT, DEFAULT_PRESIGNATURES_TO_BUFFER, metrics};
+use near_mpc_contract_interface::types::{Curve, DomainPurpose};
+use rand::SeedableRng;
+
+/// Maximum block delay before an MPC node is considered offline by its peers.
+/// Must match `MAX_HEIGHT_DIFF` in `crates/node/src/network.rs`.
+const INDEXER_MAX_HEIGHT_DIFF: i64 = 50;
```
gilcu3: would be hood to move this var to a shared crate in that case

gilcu3 (dismissed):
Thank you!

barakeinav1:
Broader question (not specific to this PR, but came up while reviewing): do we have a documented guideline for running these e2e tests locally? Specifically around build prerequisites — this test silently depends on the `mpc-node` binary being built with `--features network-hardship-simulation`. If a contributor runs `cargo nextest run` after a plain `cargo build -p mpc-node --release`, `test_cleanup_lagging_node` will hang until timeout and panic with "alive nodes did not pull ahead of faulty node" — not an obvious failure mode for someone new to the codebase.

I looked through the repo:
- `AGENTS.md` / `CLAUDE.md` quick reference has `cargo build -p mpc-node --release` — no features mentioned
- `pytest/readme.md` does mention `--features=network-hardship-simulation`, but that's for the deprecated pytest flow
- `crates/e2e-tests/README.md` describes architecture but has no "running locally" or build-prerequisites section

Would it make sense to add a short "Running locally" section to `crates/e2e-tests/README.md` documenting the `cargo build -p mpc-node --release --features network-hardship-simulation,test-utils --locked` prerequisite (matching what CI does)?

Also thinking longer-term: could we add a runtime check so tests that need the feature fail fast with a clear error instead of timing out?

barakeinav1 (dismissed):
approved with some comments.
comments are not blocking, though.

barakeinav1:
Thanks, I missed that section — `cargo make e2e-tests` covers it cleanly. Sorry for the noise on the first question.

On the runtime check — agreed, there's no clean way to introspect binary features without extra machinery. Maybe overkill for now; `cargo make e2e-tests` is the right default path. Could revisit if this becomes a recurring source of confusion.

crates/e2e-tests/tests/cleanup_lagging_node.rs
anodar: Moved it to node-config as both node and tests depend on it already.

crates/e2e-tests/tests/cleanup_lagging_node.rs
anodar: Great observation, thanks! Addressed.

crates/e2e-tests/tests/cleanup_lagging_node.rs
anodar: Factored it out in common.rs

crates/e2e-tests/src/mpc_node.rs
anodar: Updated description.

gilcu3 (approved):

barakeinav1 (approved):

PR #2953, opened 2026-04-19
test: port foreign chain policy and transaction validation from pytest to Rust e2e

### LINKED ISSUE #2996: test: port foreign chain policy unanimity and foreign transaction validation from pytest to Rust e2e
### Background

Part of the pytest → Rust e2e migration effort tracked in #2889.

Ports the foreign chain policy auto-voting unanimity test and the three `verify_foreign_transaction` tests (Bitcoin, Abstract/EVM, Starknet). Requires mock JSON-RPC servers and per-node foreign chains configuration in the e2e cluster infrastructure.

### Acceptance Criteria

**Port `test_foreign_chain_policy_auto_voting_requires_unanimity`:**
- [ ] 3-node cluster; nodes 0 and 1 configured with identical foreign chains config (e.g. Solana), node 2 has none
- [ ] After 2 nodes vote, verify votes appear in proposals but policy is NOT applied to the contract
- [ ] After the third node votes (unanimity achieved), verify policy is applied
- [ ] Once applied, verify all proposal votes are cleared

**Port `test_verify_foreign_transaction_bitcoin`:**
- [ ] 2-node cluster with mock Bitcoin JSON-RPC server
- [ ] Submit `verify_foreign_transaction` with Bitcoin tx_id, confirmations, and `["BlockHash"]` extractor
- [ ] Assert response contains `SuccessValue` with `payload_hash` (64 hex chars) and valid `signature` (scheme `Secp256k1`, `big_r`, `s`, `recovery_id`)

**Port `test_verify_foreign_transaction_abstract`:**
- [ ] Same cluster setup with mock EVM JSON-RPC server
- [ ] Submit `verify_foreign_transaction` with Abstract tx_id, finality `Finalized`, and `["BlockHash", {"Log": {"log_index": 0}}]` extractors
- [ ] Same assertion structure as Bitcoin test

**Port `test_verify_foreign_transaction_starknet`:**
- [ ] Same cluster setup with mock Starknet JSON-RPC server
- [ ] Submit `verify_foreign_transaction` with Starknet tx_id, finality `AcceptedOnL1`, and `["BlockHash"]` extractor
- [ ] Same assertion structure as Bitcoin test

### Resources

- PR: #2953
- Parent: #2889

anodar:
* Implements mock JSON-RPC servers
* Ports test_foreign_chain_policy_auto_voting_requires_unanimity (unanimity requirement for policy voting)
* Portsverify_foreign_transaction tests (Bitcoin, Abstract, Starknet) using shared cluster

Adds per-node foreign chains configuration support to the e2e cluster infrastructure and contract view/vote helpers for foreign chain policy.

Closes: https://github.com/near/mpc/issues/2996
Work towards: #2889


crates/e2e-tests/src/cluster.rs
```diff
@@ -661,6 +665,61 @@ impl MpcCluster {
...
+        self.contract
+            .view(method_names::GET_FOREIGN_CHAIN_POLICY_PROPOSALS)
+            .await
+    }
+
+    /// Vote for a foreign chain policy from a specific node.
+    pub async fn vote_foreign_chain_policy(
+        &self,
+        node_index: usize,
+        policy: serde_json::Value,
```
gilcu3: we should avoid passing base serde types. Don't we have a proper type for the policy?

crates/e2e-tests/src/cluster.rs
```diff
@@ -661,6 +665,61 @@ impl MpcCluster {
...
+                method_names::VOTE_FOREIGN_CHAIN_POLICY,
+                json!({ "policy": policy }),
+            )
+            .await
+    }
+
+    /// Send a verify_foreign_transaction request from the default user account.
+    pub async fn send_verify_foreign_transaction(
+        &self,
+        request: serde_json::Value,
```
gilcu3: same here, lets make sure we reuse the types from the rest of the codebase

crates/e2e-tests/src/foreign_chain_mock.rs
```diff
@@ -0,0 +1,213 @@
...
+            "jsonrpc": "2.0",
+            "result": {
+                "blockhash": MOCK_BLOCK_HASH,
+                "confirmations": 10,
+            },
+            "id": request_id,
+        })
+    } else {
+        jsonrpc_error(request_id, method)
+    }
```
gilcu3: not sure how useful it is for these tests to return the jsonrpc_error instead of just crashing there, will continue to see if we have tests that depend on the error being nice

crates/e2e-tests/tests/common.rs
```diff
@@ -25,6 +25,9 @@ pub const CKD_VERIFICATION_PORT_SEED: u16 = 9;
 pub const LOST_ASSETS_PORT_SEED: u16 = 10;
 pub const CKD_PV_VERIFICATION_PORT_SEED: u16 = 11;
 pub const CLEANUP_LAGGING_NODE_PORT_SEED: u16 = 12;
+pub const FOREIGN_CHAIN_POLICY_PORT_SEED: u16 = 13;
+// 12, 14, 15 reserved for remaining #2889 subtasks (cleanup_lagging_node, migration_endpoint, migration_service)
+pub const FOREIGN_TX_VALIDATION_PORT_SEED: u16 = 16;
```
gilcu3: 12 seems already in use, or at least defined

crates/e2e-tests/tests/foreign_chain_policy.rs
```diff
@@ -0,0 +1,135 @@
...
+///
+/// 1. After nodes 0 and 1 auto-vote, proposals should show 2 votes but
+///    the policy should NOT be applied to the contract.
+/// 2. After node 2 manually votes (achieving unanimity), the policy is applied.
+/// 3. Once applied, all proposal votes are cleared.
+#[tokio::test]
+#[expect(non_snake_case)]
+async fn foreign_chain_policy__should_require_unanimity_for_auto_voting() {
+    // given — 3-node cluster with foreign chains on nodes 0 and 1 only
+    let (cluster, _running) = common::setup_cluster(common::FOREIGN_CHAIN_POLICY_PORT_SEED, |c| {
```
gilcu3, line 41: I think this is going to change very soon, but I guess it could be transformed for the new requirements without much trouble

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
gilcu3, line 1: slightly better name for the file could be `foreign_chain_tx_validation`

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
```diff
@@ -0,0 +1,275 @@
...
+/// BNB, and Starknet and verifies the MPC nodes return valid signed responses.
+#[tokio::test]
+#[expect(non_snake_case)]
+async fn verify_foreign_transaction__should_sign_all_supported_chains() {
+    let env = setup_foreign_tx_cluster().await;
+
+    verify_bitcoin(&env).await;
+    verify_abstract(&env).await;
+    verify_bnb(&env).await;
+    verify_starknet(&env).await;
```
gilcu3, line 334: so we are checking the happy case for all chains here. I wonder if there is any interesting failure case we could check.

gilcu3 (approved):
Thank you!

Left some minor comments

gilcu3 (changes requested):
Ups found a small blocker, it seems we already have base support in pytests, but this was not added here

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
anodar: Added some failure cases.

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
anodar: Renamed.

crates/e2e-tests/tests/foreign_chain_policy.rs
anodar: Sounds good.

crates/e2e-tests/tests/common.rs
anodar: Dropped. It was remnant from merge with pr that implemented cleanup_lagging_node.

crates/e2e-tests/src/foreign_chain_mock.rs
anodar: Thinking behind it was to have it similar to real rpc server, which won't panic but return error instead. Changed to panic though, I don't have strong opinion on it.

crates/e2e-tests/src/cluster.rs

crates/e2e-tests/src/cluster.rs

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
```diff
@@ -0,0 +1,365 @@
...
+
+/// Sets up a single 2-node cluster with mock RPC servers for all chains,
+/// then submits verify_foreign_transaction requests for Bitcoin, Abstract,
+/// BNB, Base, and Starknet and verifies the MPC nodes return valid signed responses.
+/// Also verifies rejection for unsupported chains and non-existent domains.
+#[tokio::test]
+#[expect(non_snake_case)]
+async fn verify_foreign_transaction__should_sign_all_supported_chains() {
+    // Given — 2-node cluster with Bitcoin, Abstract, BNB, Base, and Starknet configured
+    let env = setup_foreign_tx_cluster().await;
```
gilcu3, line 327: nit: technically we discourage testing too many things in the same test (although that pattern had not been fully respected in some pytests). Here maybe we could use rstest to reduce the code, although it would be best to reuse `env`.

gilcu3 (dismissed):
Thank you!

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
anodar: I agree in principle, although here I want to avoid spawning separate cluster for each chain and instead share them.

crates/e2e-tests/src/cluster.rs
```diff
@@ -60,6 +60,9 @@ pub struct MpcClusterConfig {
     /// An empty vec means all nodes are participants. Set to a subset to start
     /// extra non-participant nodes (useful for resharing and attestation tests).
     pub initial_participant_indices: Vec<usize>,
+    /// Per-node foreign chains configuration. If empty, all nodes get the default
+    /// (empty) config. If non-empty, must have exactly `num_nodes` entries.
+    pub node_foreign_chains_configs: Vec<mpc_node_config::ForeignChainsConfig>,
```
DSharifi, line 65: Tip:
you can use const generics here to maintain the variant in the comment with the type system instead.
```rust
pub struct MpcClusterConfig<const N: usize> {
    pub num_nodes: N,
    pub node_foreign_chains_configs: Option<[mpc_node_config::ForeignChainsConfig; N]>,
}
```

crates/e2e-tests/src/foreign_chain_mock.rs
```diff
@@ -0,0 +1,202 @@
...
+    handler: fn(serde_json::Value) -> serde_json::Value,
+) -> MockServerGuard {
+    let app = axum::Router::new().route(
+        "/",
+        post(move |Json(body): Json<serde_json::Value>| async move { Json(handler(body)) }),
+    );
+
+    let listener = TcpListener::bind(format!("127.0.0.1:{port}"))
+        .await
+        .unwrap_or_else(|e| panic!("failed to bind mock server on port {port}: {e}"));
```
DSharifi: wouldn't expect give you the same message here?

crates/e2e-tests/src/foreign_chain_mock.rs
DSharifi: A panic here will not crash the test which I think what @gilcu3 wants and is nice to have.

A panic in the method handler on axum will just terminate the connection. The panic is isolated to the task handling the request, it's not propagated.

In that case let's revert back to returning proper errors @anodar

crates/e2e-tests/src/foreign_chain_mock.rs
```diff
@@ -0,0 +1,202 @@
...
+    tokio::spawn(async move {
+        axum::serve(listener, app)
+            .with_graceful_shutdown(async {
+                rx.await.ok();
+            })
+            .await
+            .ok();
+    });
+
+    MockServerGuard { _shutdown: tx }
```
DSharifi: We can use cancellation tokens, instead of manually implementing it with oneshot.

https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html

crates/e2e-tests/src/foreign_chain_mock.rs
DSharifi: Alternatively, we can add some mechanism to fail the tests if a bad request comes in.

crates/e2e-tests/src/foreign_chain_mock.rs
```diff
@@ -0,0 +1,202 @@
...
+                    "l1_data_gas": 512,
+                },
+            },
+            "jsonrpc": "2.0",
+            "id": request_id,
+        })
+    } else {
+        panic!("unexpected Starknet RPC method: {method}");
+    }
+}
```
DSharifi: Instead of using Axum, we could use [httpmock](https://docs.rs/httpmock/latest/httpmock/) for the mocked http server.

That way we can set expectations for the request and have the server cause a panic on the main thread if any of the expectations are not met.

We already use it  in the integration tests for the foreign chain inspector crate, so the test code can be reused here

https://github.com/near/mpc/blob/4652b86eaad05ef481213cda80773cbabd6d44d0/crates/foreign-chain-inspector/tests/starknet_inspector.rs#L304-L331

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
DSharifi: Too bad we run these tests with cargo-nextest, otherwise we could have had injected a shared cluster for the proccess

DSharifi (changes requested):
Thanks for doing this porting, looks very good!

Hate to block here, but let's revert the change with the panics in the axum handlers. This will just panic silently.

crates/e2e-tests/src/foreign_chain_mock.rs
anodar: Done.

crates/e2e-tests/src/foreign_chain_mock.rs
anodar: Reverted to jsonrpc errors.

crates/e2e-tests/src/foreign_chain_mock.rs
anodar: Done.

crates/e2e-tests/src/cluster.rs
anodar: I thought about this, although I prefer instead not to change signature and have these enforced on runtime. That way it's more flexible: to add nodes, or have nodes in standby (e.g. for migration tests), and avoid overhead of changing every function signature that takes this as an argument.

crates/e2e-tests/src/foreign_chain_mock.rs
anodar: Done.

DSharifi (approved):
Thank you!

crates/e2e-tests/src/foreign_chain_mock.rs
```diff
@@ -0,0 +1,206 @@
...
+        when.method(POST).path("/");
+        then.respond_with(move |req: &HttpMockRequest| {
+            let body: serde_json::Value =
+                serde_json::from_slice(req.body().as_ref()).expect("valid json-rpc request");
+            let id = body["id"].clone();
+            let method = body["method"].as_str().expect("method field");
+
+            if method != "getrawtransaction" {
+                return jsonrpc_error(id, method);
+            }
```
DSharifi, line 38: I think httpmock has convenient apis for these expectations to cause panics in the test driver instead of returning errors to the client.

anodar: Good point, will follow up.

gilcu3 (approved):

Issue #2932, opened 2026-04-17
Node should not depend on mpc-contract for migration and TEE types


gilcu3:
### Background

Follow-up to #2897. The node still imports these types from `mpc-contract`:

**Node migrations**
- `BackupServiceInfo`
- `DestinationNodeInfo`
- `NodeMigrations`

**TEE**
- `NodeImageHash` (already in `mpc-primitives`)
- `LauncherDockerComposeHash` (already in `mpc-primitives`)
- `NodeId`

Files:
- `indexer/migrations.rs`
- `indexer/tee.rs`
- `tee/allowed_image_hashes_watcher.rs`
- `tee/remote_attestation.rs`
- `migration_service/onboarding.rs`
- `migration_service/types.rs`
- `migration_service/web.rs`
- `migration_service/web/server.rs`
- `migration_service/web/test_utils.rs`
- `tests/onboarding.rs`

### Acceptance Criteria

The node no longer depends on `mpc-contract` for migration and TEE types.

Issue #1710, opened 2025-12-23
Make `account_public_key` non optional


DSharifi:
### Background

The field `account_public_key` was made optional, because the contract did not have the participants' near account public key, thus we needed a transition period where nodes could submit their public key as part of their attestation.

https://github.com/near/mpc/blob/738cb206acc36ae4250c09ad545cabc6c162ee96/crates/contract/src/tee/tee_state.rs#L24-L30

Now that all nodes have submitted their mock attestations with this public key registered on chain, we can remove the optionality of the field.

### Acceptance Criteria

Make `account_public_key` non optional

netrome:
Let's prioritize this cleanup. Marked as P1

gilcu3:
yeah I will get to it as soon as I get the other refactors into main, would appreciate fast reviews for those :)

PR #2954, opened 2026-04-20
refactor: unify NodeId from contract and dto

### LINKED ISSUE #2942: Unify NodeId public key types with DTO (requires state migration)
### Background

`NodeId` is currently defined twice:

- **Contract-internal** (`crates/contract/src/tee/tee_state.rs`): `tls_public_key: near_sdk::PublicKey`, `account_public_key: Option<near_sdk::PublicKey>`
- **DTO** (`crates/near-mpc-contract-interface/src/types/tee.rs`): `tls_public_key: Ed25519PublicKey`, `account_public_key: Option<PublicKey>`

The internal version is stored on chain inside `NodeAttestation`, held in `TeeState.stored_attestations: BTreeMap<near_sdk::PublicKey, NodeAttestation>` (and used as a key elsewhere).

We should use our on types defined in the dto crate instead of `near_sdk` types, that way we fully control serialization, and are able to unify these types with the interface to reduce contract binary size.

### Acceptance Criteria

- Single `NodeId` definition (reuse the DTO in `near-mpc-contract-interface`).
- `stored_attestations` and any other maps keyed by `near_sdk::PublicKey` (e.g. TLS key → attestation) migrated to the new layout.
- Contract internal type migration

gilcu3:
Closes #2942


crates/contract/src/snapshots/mpc_contract__tests__mpc_contract_borsh_schema_has_not_changed.snap
gilcu3, line 1: The change in this file is hard to parse because the name change created a reorder

crates/contract/src/v3_9_0_state.rs
```diff
@@ -43,13 +83,58 @@ pub struct MpcContract {
...
+        let stored_attestations = old
+            .stored_attestations
+            .into_iter()
+            .filter_map(|(tls_pk, old_attestation)| {
+                let new_tls_key = dtos::Ed25519PublicKey::try_from(&tls_pk).ok()?;
+                let account_public_key = old_attestation
+                    .node_id
+                    .account_public_key
+                    .as_ref()
+                    .and_then(|pk| dtos::Ed25519PublicKey::try_from(pk).ok());
```
gilcu3, line 110: All stored attestations are `Ed25519PublicKey` already, but this code takes into account that if they were not, they would simply be dropped

crates/contract/src/tee/tee_state.rs
```diff
@@ -254,17 +217,25 @@ impl TeeState {
...
             .filter(|(account_id, _, participant_info)| {
-                let tls_public_key = participant_info.sign_pk.clone();
+                // A non-Ed25519 sign_pk can never have a matching attestation
+                // (the attestation map is keyed by Ed25519), so treat the
+                // participant as invalid.
+                let Some(tls_public_key) =
+                    Ed25519PublicKey::try_from(&participant_info.sign_pk).ok()
+                else {
+                    return false;
+                };
```
gilcu3, line 227: will not be needed once `participant_info.sign_pk` is migrated to `Ed25519PublicKey`

scripts/check-contract-wasm-size.sh
```diff
@@ -12,7 +12,7 @@ set -euo pipefail
 WASM_PATH="${1:-target/near/mpc_contract/mpc_contract.wasm}"
 
 # NEAR max_transaction_size = 1572864; keep some headroom
-HARD_LIMIT=1490000
+HARD_LIMIT=1500000
```
gilcu3: should be recovered after migration is deleted in the next release

gilcu3:
> I understand this pattern was inherited from the old NodeId in the contract, but moving it into the shared interface crate increases the blast radius.

> Suggested fix: manually implement Ord/PartialOrd to also ignore account_public_key, consistent with Eq/Hash:

Will leave it as is to keep past behavior. Anyway these trait implementations will be removed in #1710

crates/contract/src/lib.rs
```diff
@@ -817,14 +817,24 @@ impl MpcContract {
...
+
         // Add the participant information to the contract state
         let attestation_insertion_result = self
             .tee_state
             .add_participant(
                 NodeId {
-                    account_id: account_id.clone(),
-                    tls_public_key: tls_public_key.into(),
-                    account_public_key: Some(account_key),
+                    account_id: dtos::AccountId(account_id.to_string()),
```
kevindeforth: I am not sure why we would prefer storing a string over the actual type in the contract state.

crates/contract/src/primitives/participants.rs
```diff
@@ -228,8 +228,11 @@ impl Participants {
         self.participants()
             .iter()
             .map(|(account_id, _, p_info)| NodeId {
-                account_id: account_id.clone(),
-                tls_public_key: p_info.sign_pk.clone(),
+                account_id: near_mpc_contract_interface::types::AccountId(account_id.to_string()),
+                tls_public_key: near_mpc_contract_interface::types::Ed25519PublicKey::try_from(
+                    &p_info.sign_pk,
+                )
+                .expect("sign_pk must be Ed25519"),
```
kevindeforth: IIRC, these are just strings. What guarantees does the interface crate give, that the provided strings will be valid `AccountId`s and public keys?

crates/contract/src/lib.rs
gilcu3: I was not sure either, just decided for this to avoid the full struct duplication. If using the account-id type from near did not require duplicating the whole struct, I would have kept it

crates/contract/src/primitives/participants.rs
gilcu3: valid account ids not much, for public keys we get the same we had before (the near-sdk type also does not check for valid keys, just valid size). Not having a valid account-id would eventually lead to failures when comparing to the actual account-id, so I don't think we lose much by doing this. Let me know if that's not the case

kevindeforth: True, I would suggest the following:
- refactor `sign_pk` in [`ParticipantInfo`](https://github.com/near/mpc/blob/e05e85679fb7833f58121adf45ee01f5b09fd88f/crates/contract/src/primitives/participants.rs#L19) to our own `Ed25519PublicKey` type.
_Note: this is the tls key, we should have an issue somewhere for renaming that field, but I didn't find it._
- replace our `contract_interface::AccountId` with `near_account_id::AccountId`. I would think that near sdk is extremely careful introducing any breaking changes there.

gilcu3: fixed in [e2df603](https://github.com/near/mpc/pull/2954/commits/e2df6033ce32a27e80b717a3cf75a85d83d59be9)

> refactor sign_pk in [ParticipantInfo](https://github.com/near/mpc/blob/e05e85679fb7833f58121adf45ee01f5b09fd88f/crates/contract/src/primitives/participants.rs#L19) to our own Ed25519PublicKey type.
Note: this is the tls key, we should have an issue somewhere for renaming that field, but I didn't find it.

as we discussed will handle this in a follow-up

scripts/check-contract-wasm-size.sh
kevindeforth: Do we still need this with the latest changes?

crates/near-mpc-contract-interface/src/types/tee.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub account_id: AccountId,
+    /// TLS public key used by the node for peer-to-peer communication.
+    pub tls_public_key: Ed25519PublicKey,
+    /// Full-access Ed25519 public key of the operator account.
+    pub account_public_key: Option<Ed25519PublicKey>,
+}
+
+impl PartialEq for NodeId {
+    fn eq(&self, other: &Self) -> bool {
+        self.account_id == other.account_id && self.tls_public_key == other.tls_public_key
```
kevindeforth, line 26: What prevents us from requiring a match on `account_public_key` here?

crates/node/src/indexer/participants.rs
```diff
@@ -355,10 +355,7 @@ pub fn convert_participant_infos(
         let p2p_public_key = ed25519_dalek::VerifyingKey::try_from(&info.sign_pk)
             .with_context(|| format!("Invalid sign_pk for peer: {:?}", info.sign_pk))?;
 
-        let near_account_id: AccountId = account_id
-            .0
-            .parse()
-            .with_context(|| format!("Invalid account id: {}", account_id.0))?;
+        let near_account_id: AccountId = account_id.clone();
```
kevindeforth, line 358: Not related to this exact line, but I think we might be breaking something, because the currently deployed contract has a different serialization format than what this node expects.

crates/contract/src/lib.rs
```diff
@@ -2252,14 +2260,25 @@ impl MpcContract {
             }
             .into());
         }
-        // ensure that this node has a valid TEE quote
+        // ensure that this node has a valid TEE quote. Both keys must be
+        // Ed25519 — a non-Ed25519 TLS key could never have an attestation
+        // stored (the map is keyed by Ed25519), and a non-Ed25519 signer key
+        // could never sign this very transaction.
+        let account_public_key =
+            dtos::Ed25519PublicKey::try_from(&expected_destination_node.signer_account_pk)
```
kevindeforth, line 2268: Suggestion: we should change `DestinationNodeInfo` do use the dto types as well. This should b a rather trivial change, since we don't need backwards-compatibility for them.

kevindeforth (commented):
Partial review, as it's quite a large PR.
Need to check that the resulting node is still compatible with the currently deployed contract.

crates/near-mpc-contract-interface/src/types/tee.rs
gilcu3: this was moved as is, I did not want to try changing that behavior in the PR. Will fix when account_public_key becomes non-optional

crates/node/src/indexer/participants.rs
gilcu3: which struct do you have in mind that changed for serde serialization?

crates/contract/src/lib.rs
gilcu3: we have a follow-up planned for the migration types. Actually that's what led to this PR in the first place https://github.com/near/mpc/issues/2932

scripts/check-contract-wasm-size.sh
gilcu3: luckily can be reverted indeed:

```
Size:       1481635 bytes
```

will do after the first round of reviews is done

crates/contract/src/primitives/participants.rs
```diff
@@ -229,7 +229,10 @@ impl Participants {
             .iter()
             .map(|(account_id, _, p_info)| NodeId {
                 account_id: account_id.clone(),
-                tls_public_key: p_info.sign_pk.clone(),
+                tls_public_key: near_mpc_contract_interface::types::Ed25519PublicKey::try_from(
+                    &p_info.sign_pk,
+                )
+                .expect("sign_pk must be Ed25519"),
```
anodar, line 235: Why not propagate this as an error instead of panic? Technically this could be `secp256k1` and panic no?

Alternatively if you don't want to change the signature here, you could validate in `insert_with_id` and return error there as it already returns `Result<(), Error>` and add comment here that this expect should never happen.

crates/contract/src/tee/tee_state.rs
```diff
@@ -405,15 +376,17 @@ impl TeeState {
     /// Removes TEE information and stale votes for nodes that are not in the provided
     /// participants list. Used to clean up storage after a resharing concludes.
     pub fn clean_non_participants(&mut self, participants: &Participants) {
-        // Collect all allowed TLS public keys from current participants
-        let active_tls_keys: HashSet<&near_sdk::PublicKey> = participants
+        // Collect all allowed TLS public keys from current participants.
+        // Non-Ed25519 `sign_pk`s can't appear in `stored_attestations` (it's
+        // keyed by Ed25519), so they're silently dropped here.
```
anodar, line 381: optional nit: here and in functions below, consider logging those that's not Ed25519.

anodar (dismissed):

crates/contract/src/primitives/participants.rs
gilcu3: we plan changing the type of `p_info.sign_pk` in #2961, so this code will disappear quite soon. We could also propagate it, but we don't really expect it to be hit as all nodes submit only ed25519 public keys anyway.

crates/contract/src/tee/tee_state.rs
gilcu3: similar to the previous comment, this case will be enforced by the type system after https://github.com/near/mpc/issues/2961. In general I think we are avoiding logging in the contract because it tends to increase its size

crates/node/src/indexer/participants.rs
kevindeforth: The `AccountId` is used in the `Participants` type of the contract-interface and changes from `AccountId(String)` to the near-implementation.

This means that the upgraded node won't be able to deserialize the contract state in the indexer: https://github.com/near/mpc/blob/c2ea01b397c4436e832ada1f0da29cbce5d475ec/crates/node/src/indexer.rs#L301

kevindeforth (commented):
looks good, but we need to ensure the node can deserialize the currently deployed contract. C.f. https://github.com/near/mpc/pull/2954#discussion_r3111475769

crates/contract/src/primitives/test_utils.rs
```diff
@@ -82,6 +82,14 @@ pub fn bogus_ed25519_near_public_key() -> near_sdk::PublicKey {
...
 }
 
+/// Generates a random [`near_mpc_contract_interface::types::Ed25519PublicKey`]
+/// suitable for use as the `account_public_key` field of a
+/// [`near_mpc_contract_interface::types::NodeId`]. The node's signer key is
+/// always Ed25519, so this mirrors real submissions.
+pub fn bogus_ed25519_account_public_key() -> near_mpc_contract_interface::types::Ed25519PublicKey {
+    bogus_ed25519_public_key()
+}
+
```
SimonRastikian: Why is this function necessary if you already have `bogus_ed25519_public_key`?

crates/contract/src/tee/tee_state.rs
```diff
@@ -14,40 +14,12 @@ use mpc_attestation::{
...
-
-impl Eq for NodeId {}
-
-impl Hash for NodeId {
-    fn hash<H: Hasher>(&self, state: &mut H) {
-        self.account_id.hash(state);
-        self.tls_public_key.hash(state);
-        // intentionally ignoring account_public_key
-    }
-}
```
SimonRastikian, line 50: Gosh so good to see this!

crates/contract/src/tee/tee_state.rs
```diff
@@ -108,36 +78,41 @@ pub struct TeeState {
...
     pub(crate) fn with_mocked_participant_attestations(participants: &Participants) -> Self {
         let mut participants_attestations = BTreeMap::new();
 
-        participants
-            .participants()
-            .iter()
-            .for_each(|(account_id, _, participant_info)| {
-                let node_id = NodeId {
-                    account_id: account_id.clone(),
-                    tls_public_key: participant_info.sign_pk.clone(),
```
SimonRastikian, line 127: Does this mean that we supported more types of keys in the past and with the current changes we don't? Or was this a bug?

crates/contract/src/tee/tee_state.rs
```diff
@@ -108,36 +78,41 @@ pub struct TeeState {
...
-                    account_public_key: None,
-                };
+        for (account_id, _, participant_info) in participants.participants() {
+            let Some(tls_public_key) = Ed25519PublicKey::try_from(&participant_info.sign_pk).ok()
+            else {
+                continue;
+            };
```
SimonRastikian, line 99: Question above relates to this change

SimonRastikian (commented):
Only a partial review... duty calls...

crates/contract/src/primitives/test_utils.rs
gilcu3: just to distinguish the usage for account_public_key and tls_public_key in tests. Not strictly needed

crates/contract/src/tee/tee_state.rs
gilcu3: We never supported secp256k1 keys here, but the near-sdk type included those as well. So here we are fixing that, by making the requirement uniform

crates/contract/src/tee/tee_state.rs
gilcu3: responded above

crates/node/src/indexer/participants.rs
gilcu3: Both the old `AccountId(pub String)` newtype and `near_account_id::AccountId` serialize as a bare string "alice.near"

crates/contract/src/primitives/test_utils.rs
gilcu3: removed

crates/node/src/indexer/participants.rs
kevindeforth: Thanks for having taken the time to verify this. As we discussed on [slack](https://nearone.slack.com/archives/C0912BTG51T/p1776756424645189?thread_ts=1776669145.501399&cid=C0912BTG51T) we have a gap in our test coverage because we changed our upgrade order since https://github.com/near/mpc/issues/1185.
We will add test coverage separately (https://github.com/near/mpc/issues/2970) to avoid having to scrutinize these things during review.

kevindeforth (approved):
Thank you!

pbeza (approved):
Only had time to skim it (sorry—lots of other reviews).

crates/near-mpc-contract-interface/src/types/tee.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+impl Eq for NodeId {}
+
+impl Hash for NodeId {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.account_id.hash(state);
+        self.tls_public_key.hash(state);
+    }
+}
```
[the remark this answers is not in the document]
gilcu3: we discussing this already, will be fixed soon

PR #2956, opened 2026-04-20
fix: cleanup of stored attestations

gilcu3:
Use lazy struct to store attestations, and add a public cleanup function to eventually delete expired attestations

### DIFF
--- a/.github/contract-size-baseline.toml
+++ b/.github/contract-size-baseline.toml
@@ -3,4 +3,4 @@
 max_transaction_size = 1572864
 
 # Expected contract WASM size in bytes (reproducible builds must match exactly)
-expected_size = 1481703
+expected_size = 1490000
--- a/crates/contract/src/snapshots/mpc_contract__tests__mpc_contract_borsh_schema_has_not_changed.snap
+++ b/crates/contract/src/snapshots/mpc_contract__tests__mpc_contract_borsh_schema_has_not_changed.snap
@@ -51,12 +51,6 @@ BorshSchemaContainer {
                 "NonEmptyBTreeSet<RpcProvider>",
             ],
         },
-        "(PublicKey, NodeAttestation)": Tuple {
-            elements: [
-                "PublicKey",
-                "NodeAttestation",
-            ],
-        },
         "AccountId": Struct {
             fields: UnnamedFields(
                 [
@@ -183,11 +177,6 @@ BorshSchemaContainer {
             length_range: 0..=4294967295,
             elements: "(ForeignChain, NonEmptyBTreeSet<RpcProvider>)",
         },
-        "BTreeMap<PublicKey, NodeAttestation>": Sequence {
-            length_width: 4,
-            length_range: 0..=4294967295,
-            elements: "(PublicKey, NodeAttestation)",
-        },
         "BTreeSet<AuthenticatedParticipantId>": Sequence {
             length_width: 4,
             length_range: 0..=4294967295,
@@ -252,6 +241,10 @@ BorshSchemaContainer {
                     ),
                     (
                         "clean_tee_status_tera_gas",
+                        "u64",
+                    ),
+                    (
+                        "clean_invalid_attestations_tera_gas",
                         "u64",
                     ),
                     (
@@ -418,20 +411,6 @@ BorshSchemaContainer {
             fields: UnnamedFields(
                 [
                     "u64",
-                ],
-            ),
-        },
-        "ExpectedMeasurements": Struct {
-            fields: NamedFields(
-                [
-                    (
-                        "rtmrs",
-                        "Measurements",
-                    ),
-                    (
-                        "key_provider_event_digest",
-                        "[u8; 48]",
-                    ),
                 ],
             ),
         },
@@ -772,28 +751,6 @@ BorshSchemaContainer {
                 ],
             ),
         },
-        "Measurements": Struct {
-            fields: NamedFields(
-                [
-                    (
-                        "mrtd",
-                        "[u8; 48]",
-                    ),
-                    (
-                        "rtmr0",
-                        "[u8; 48]",
-                    ),
-                    (
-                        "rtmr1",
-                        "[u8; 48]",
-                    ),
-                    (
-                        "rtmr2",
-                        "[u8; 48]",
-                    ),
-                ],
-            ),
-        },
         "Metrics": Struct {
             fields: NamedFields(
                 [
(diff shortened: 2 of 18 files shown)


gilcu3:
> ⚠️ (minor issues found — none strictly blocking, but the gas concerns warrant validation before mainnet deployment)

issues are valid, but should not affect our current deployment (because we have very few stored attestations), nor future deployments (because migration is only needed once)

barakeinav1:
**Design question:** We already have `verify_tee`, which nodes call periodically and which runs `reverify_and_cleanup_participants`. That endpoint re-verifies attestations for current participants but never touches outsider entries in `stored_attestations` — which is exactly the gap this PR fills.

Did you consider extending `verify_tee` (or `reverify_and_cleanup_participants`) to also sweep invalid outsider entries instead of adding a new public endpoint?

Pros of folding in:
- One less public endpoint and one less gas config knob
- Nodes already call `verify_tee` regularly — outsider cleanup would happen automatically, no need to remember to call `clean_invalid_attestations` or schedule it post-reshare

Cons (why a separate endpoint might still be right):
- `verify_tee` has richer semantics (can trigger resharing if valid-participant count drops below threshold) — mixing sweep logic into that code path is a bigger change
- `verify_tee` is gated to `voter_or_panic()` (participant-only); `clean_invalid_attestations` is public. Separate endpoints keep those access controls cleanly distinct
- Bounded scan (`max_scan`) makes gas cost predictable; adding unbounded sweep to `verify_tee` could make that call's gas unpredictable

Totally fine to keep them separate if that's a deliberate choice — just wanted to make sure the alternative was weighed. Not blocking for the 3.9.2 patch.

.github/contract-size-baseline.toml
```diff
@@ -3,4 +3,4 @@
 max_transaction_size = 1572864
 
 # Expected contract WASM size in bytes (reproducible builds must match exactly)
-expected_size = 1481703
+expected_size = 1490000
```
barakeinav1, line 6: is the limit 1500000!?

barakeinav1 (approved):
thanks

barakeinav1:
**Question on `get_tee_accounts`** (`tee_state.rs:436`):

This function enumerates the full `stored_attestations` map:

```rust
pub fn get_tee_accounts(&self) -> Vec<NodeId> {
    self.stored_attestations
        .values()
        .map(|node_attestation| node_attestation.node_id.clone())
        .collect()
}
```

Post-fix, the map can still grow unboundedly (just doesn't brick state-changing calls anymore). An attacker could keep submitting/resubmitting attestations to inflate the list.

It's a view method though, so I'm not sure what the actual risk is — are there gas/response size limits on view calls that would make this a real problem at scale? Or is this fine to leave as-is?

.github/contract-size-baseline.toml
gilcu3: luckily is higher, around 1570000 :)

gilcu3:
> Post-fix, the map can still grow unboundedly (just doesn't brick state-changing calls anymore). An attacker could keep submitting/resubmitting attestations to inflate the list.
> 
> It's a view method though, so I'm not sure what the actual risk is — are there gas/response size limits on view calls that would make this a real problem at scale? Or is this fine to leave as-is?

I think we don't depend on this view method not failing for anything. So having the failure would be helpful, make us run a cleanup

crates/contract/src/tee/tee_state.rs
```diff
@@ -95,28 +93,40 @@ pub(crate) struct NodeAttestation {
     pub(crate) verified_attestation: VerifiedAttestation,
 }
 
-#[derive(Default, Debug, BorshSerialize, BorshDeserialize)]
-#[cfg_attr(
-    all(feature = "abi", not(target_arch = "wasm32")),
-    derive(borsh::BorshSchema)
-)]
```
netrome, line 102: Why do we change to use the near macro here instead of the feature gated derive?

crates/contract/src/tee/tee_state.rs
```diff
@@ -402,33 +409,47 @@ impl TeeState {
...
-            .map(|(_, _, p_info)| &p_info.sign_pk)
-            .collect();
+    /// Drops votes cast by nodes that are no longer participants. Used after a resharing
+    /// concludes. Attestation cleanup is handled separately by
+    /// [`TeeState::clean_invalid_attestations`].
+    pub fn clean_non_participant_votes(&mut self, participants: &Participants) {
+        self.votes = self.votes.get_remaining_votes(participants);
+        self.launcher_votes = self.launcher_votes.get_remaining_votes(participants);
+        self.measurement_votes = self.measurement_votes.get_remaining_votes(participants);
+    }
```
netrome, line 419: I think we discussed having this one call `clean_invalid_attestations` at the end. Any reason for not doing this here to align closer with the previous behavior?

crates/contract/src/tee/tee_state.rs
```diff
@@ -402,33 +409,47 @@ impl TeeState {
...
 
-        // Remove stale votes from non-participants
-        self.votes = self.votes.get_remaining_votes(participants);
-        self.launcher_votes = self.launcher_votes.get_remaining_votes(participants);
-        self.measurement_votes = self.measurement_votes.get_remaining_votes(participants);
+        let removed = invalid_tls_keys.len() as u32;
+        for tls_pk in invalid_tls_keys {
+            self.stored_attestations.remove(&tls_pk);
+        }
+        removed
```
netrome, line 452: So a few things here:

1. There's an `as` conversion that I think should be an explicit `.expect()` for clarity.
2. We don't need to collect two vectors from what I can see.
3. There's no need to clone `node_id`, we only need the TLS key to be cloned.

Only the first of these is a blocker imo.

I'd suggest changing the function to be something like:

```suggestion
        let has_invalid_attestation = |node_id: &NodeId| {
            !matches!(
                self.reverify_participants(node_id, tee_upgrade_deadline_duration),
                TeeQuoteStatus::Valid
            )
        };

        // Materialize candidates before any mutation to avoid iterator invalidation.
        let invalid_tls_keys: Vec<near_sdk::PublicKey> = self
            .stored_attestations
            .iter()
            .take(max_scan)
            .filter(|(_, node_attestation)| has_invalid_attestation(&node_attestation.node_id))
            .map(|(tls_pk, _)| tls_pk.clone())
            .collect();

        let removed = u32::try_from(invalid_tls_keys.len())
            .expect("u32 should always be convertible from usize on wasm32");

        for tls_pk in invalid_tls_keys {
            self.stored_attestations.remove(&tls_pk);
        }
        removed
```

netrome (changes requested):
A few questions and one blocker

crates/contract/src/tee/tee_state.rs
gilcu3: the near annotation is the "more" correct way of doing it. We dont use it only when it is not available, like in the interface crate

crates/contract/src/tee/tee_state.rs
gilcu3: we still call it when cleaning up tee state, just not in the clean votes function

crates/contract/src/tee/tee_state.rs
netrome: Okay but in this case we gate the implementation based on our feature flags, I assume the near macro is unaware of these - so I assume this changes when we support the schema and when we don't.

gilcu3: I cannot really explain why it works, but I am sure it does because we are still generating abi in tests and with cargo-near without issues

crates/contract/src/snapshots/mpc_contract__tests__mpc_contract_borsh_schema_has_not_changed.snap
netrome, line 1: Was surprised to see this schema change. But Claude gave a good explanation for it:

```
  1. stored_attestations field type
  - BTreeMap<PublicKey, NodeAttestation> → IterableMap (from near_sdk::store)
  - This is the main change in the PR: moving attestations to lazy, storage-backed map access so the whole collection isn't deserialized on every contract call.
  - Side effect: the schema no longer enumerates the inner types (NodeAttestation, NodeId, VerifiedAttestation, MockAttestation, ExpectedMeasurements, related Option<...> and [u8; 48])
  because IterableMap's BorshSchema impl just emits its own name — unlike BTreeMap which recurses into K and V. That's why all those entries vanished from the snap; the types still exist and
   are still borsh-serialized per-entry, they're just no longer reachable from the root schema via TeeState.
```

netrome (approved):
Thanks!

PR #2964, opened 2026-04-20
refactor: unify participant types from the contract/node and interface

### LINKED ISSUE #2933: Node should not depend on mpc-contract for participant and threshold types
### Background

Follow-up to #2897. The node still imports these types from `mpc-contract`:

**Participants**
- `ParticipantId`, `ParticipantInfo`, `Participants`

**Thresholds**
- `Threshold`, `ThresholdParameters`

These all have equivalents in `contract-interface` already. The contract defines its own internal versions with `#[near]` macros and `From`/`Into` conversions — similar to how `DomainId`/`Curve` were handled in #2897.

Alongside this, `YieldIndex` should also moved to `mpc-primitives`. JSON and borsh wire formats are already compatible (`[u8; 32]` on both sides), so no state migration is needed for that part.

### Acceptance Criteria

- The node no longer depends on `mpc-contract` for participant and threshold types.
- `YieldIndex` is moved to `mpc-primitives` and re-exported through `contract-interface`.

gilcu3:
Closes #2933

### DIFF
--- a/crates/contract/src/dto_mapping.rs
+++ b/crates/contract/src/dto_mapping.rs
@@ -26,9 +26,9 @@ use crate::{
             AttemptId, AuthenticatedAccountId, AuthenticatedParticipantId, EpochId, KeyEventId,
             KeyForDomain, Keyset,
         },
-        participants::Participants,
+        participants::{ParticipantInfo, Participants},
         threshold_votes::ThresholdParametersVotes,
-        thresholds::{Threshold, ThresholdParameters},
+        thresholds::ThresholdParameters,
     },
     state::{
         initializing::InitializingContractState,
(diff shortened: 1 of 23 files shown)


crates/contract/src/dto_mapping.rs
```diff
@@ -198,6 +198,34 @@ impl TryIntoContractType<TcbInfo> for dtos::TcbInfo {
...
+            .collect();
+        Participants::init(self.next_id, participants)
+    }
+}
+
+impl IntoContractType<ThresholdParameters> for dtos::ThresholdParameters {
+    fn into_contract_type(self) -> ThresholdParameters {
+        ThresholdParameters::new_unvalidated(self.participants.into_contract_type(), self.threshold)
+    }
+}
```
gilcu3, line 227: this uses unvalidated to remain in line with existing code which uses an explicit `validate` call for these parameters

netrome (approved):
Nice stuff!

DSharifi (approved):

PR #2974, opened 2026-04-21
chore: make account public key non-optional

### LINKED ISSUE #1710: Make `account_public_key` non optional

gilcu3:
Closes #1710

### DIFF
--- a/crates/contract/src/tee/tee_state.rs
+++ b/crates/contract/src/tee/tee_state.rs
@@ -97,10 +97,15 @@ impl TeeState {
             else {
                 continue;
             };
+            // `Participants` does not carry the operator's account public key,
+            // so a mocked entry cannot record the real one. The all-zero key is
+            // a placeholder that never matches a live signer; it keeps the
+            // participant from being kicked out of an empty `TeeState` until a
+            // real `submit_participant_info` call replaces it (keyed by TLS).
             let node_id = NodeId {
                 account_id: account_id.clone(),
                 tls_public_key: tls_public_key.clone(),
-                account_public_key: None,
+                account_public_key: Ed25519PublicKey::from([0u8; 32]),
             };
 
             participants_attestations.insert(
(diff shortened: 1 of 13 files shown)


crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
```diff
@@ -2950,12 +2951,9 @@ expression: abi
             },
             "account_public_key": {
               "description": "Full-access Ed25519 public key of the operator account.",
-              "anyOf": [
+              "allOf": [
```
DSharifi, line 2954: Wonder why it's allof instead of the concrete type?

DSharifi (dismissed):
Thanks for this cleanup

crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
gilcu3: good question, maybe related to the flakiness I found with abi generation, that will be solved in #2976

crates/contract/src/tee/tee_state.rs
```diff
@@ -97,10 +97,15 @@ impl TeeState {
...
+            // `Participants` does not carry the operator's account public key,
+            // so a mocked entry cannot record the real one. The all-zero key is
+            // a placeholder that never matches a live signer; it keeps the
+            // participant from being kicked out of an empty `TeeState` until a
+            // real `submit_participant_info` call replaces it (keyed by TLS).
             let node_id = NodeId {
                 account_id: account_id.clone(),
                 tls_public_key: tls_public_key.clone(),
-                account_public_key: None,
+                account_public_key: Ed25519PublicKey::from([0u8; 32]),
```
netrome: This is giving me quite a bit of a pause. If we allow `account_public_key` to be unset, it's better to just retain it as an option to make this explicit.

I also don't recall why we need the account public key to identify a node, nor the `account_id`. Isn't the `tls_public_key` sufficient to be a node identifier?

netrome (changes requested):
I don't like the idea of us using a zero key as a placeholder for an unset key. If we want to represent the possibility of this key being unset we should retain it as an Option imo.

crates/contract/src/tee/tee_state.rs
gilcu3: It is sufficient to have `tls_public_key` as identity. I will try to change the init method signatures to receive `account_public_key` as well, as both are anyway available when a node starts.

netrome:
Had some discussions back and forth on this. This PR will be on hold over the next few days and @gilcu3 will pick it up again next week, exploring implications and potential related refactors we need to enable this.

DSharifi (approved):

netrome (approved):
Approved (after more discussions we agreed to live with the sentinel values for a short time)

Issue #2961, opened 2026-04-20
Rename ParticipantInfo.sign_pk to tls_public_key and change type to Ed25519PublicKey


gilcu3:
### Background

Follow-up to #2933. The contract's `ParticipantInfo` currently has:

```rust
pub struct ParticipantInfo {
    pub url: String,
    pub sign_pk: near_sdk::PublicKey,
}
```

Two issues:
- The field name `sign_pk` is misleading — it's actually the TLS key used for P2P, matching `NodeId.tls_public_key` (already `Ed25519PublicKey` after #2942). The rest of the codebase consistently calls this `tls_public_key`.
- The type `near_sdk::PublicKey` is a tagged 33-byte borsh encoding; `Ed25519PublicKey` is the plain 32-byte form used everywhere else in the node and DTOs. Participant TLS keys are always Ed25519.

Renaming the field to `tls_public_key` and switching to `Ed25519PublicKey` changes both the JSON key and the on-chain borsh layout, so a one-time state migration is required — same pattern as `crates/contract/src/v3_9_0_state.rs` (PR #2942).

### Acceptance Criteria

- `ParticipantInfo.sign_pk: near_sdk::PublicKey` is renamed to `tls_public_key: Ed25519PublicKey`.
- A versioned state migration (`v3_X_X_state.rs`) converts existing on-chain `ParticipantInfo` values to the new layout.
- Call sites across contract and node are updated.

PR #2976, opened 2026-04-21
refactor: remove node contract dependency for TEE and migration types in the node

### LINKED ISSUE #2932: Node should not depend on mpc-contract for migration and TEE types

### LINKED ISSUE #2977: Flaky test_abi_has_not_changed from duplicate ParticipantInfo schema name
### Background

`cargo nextest run --cargo-profile=test-release --locked --all-features -p mpc-contract test_abi_has_not_changed` is flaky. The failing diff is always `ParticipantInfo.sign_pk`:

```
-              "$ref": "#/definitions/Ed25519PublicKey"
+              "description": "The public key used for verifying messages.",
+              "type": "string"
```

Root cause: two `ParticipantInfo` structs both derive `schemars::JsonSchema` and return `schema_name() == "ParticipantInfo"`:

- `crates/contract/src/primitives/participants.rs:13` — `sign_pk: near_sdk::PublicKey` (inlines as `type: string` because `near_sdk::PublicKey::is_referenceable() == false`).
- `crates/near-mpc-contract-interface/src/types/participants.rs:25` — `sign_pk: near_mpc_crypto_types::Ed25519PublicKey` (emits `$ref: Ed25519PublicKey`).

`schemars` keys its `/definitions/` table by `schema_name`, so whichever struct it visits first via contract endpoint schemas wins and the other silently collapses to a `$ref`. Visit order depends on codegen / compile-unit ordering, so it's non-deterministic. The committed snapshot happens to capture the DTO variant; CI hits the other ordering more often.

Introduced by #2964 (`refactor: unify participant types from the contract/node and interface`), not by any recent PR.

### Acceptance Criteria

- `test_abi_has_not_changed` passes consistently.
- The fix makes the two `ParticipantInfo` structs distinguishable to `schemars` (e.g. `#[schemars(rename = "ParticipantInfoDto")]` on the DTO, a custom `schema_name()` override, or by not exposing both types in the ABI surface).

gilcu3:
Closes #2932 

Also appears to close #2977 (just by chance)

### DIFF
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -63,7 +63,7 @@ use near_sdk::{
     store::{IterableMap, LookupMap},
     AccountId, CryptoHash, Gas, GasWeight, NearToken, Promise, PromiseError, PromiseOrValue,
 };
-use node_migrations::{BackupServiceInfo, DestinationNodeInfo, NodeMigrations};
+use node_migrations::NodeMigrations;
 use primitives::{
     domain::DomainRegistry,
     key_state::{AuthenticatedAccountId, AuthenticatedParticipantId, EpochId, KeyEventId, Keyset},
@@ -1535,7 +1535,7 @@ impl MpcContract {
     /// Returns all accounts that have TEE attestations stored in the contract.
     /// Note: This includes both current protocol participants and accounts that may have
     /// submitted TEE information but are not currently part of the active participant set.
-    pub fn get_tee_accounts(&self) -> Vec<NodeId> {
+    pub fn get_tee_accounts(&self) -> Vec<dtos::NodeId> {
         log!("get_tee_accounts");
         self.tee_state.get_tee_accounts()
     }
@@ -2130,7 +2130,13 @@ impl MpcContract {
 impl MpcContract {
     pub fn migration_info(
         &self,
-    ) -> BTreeMap<AccountId, (Option<BackupServiceInfo>, Option<DestinationNodeInfo>)> {
+    ) -> BTreeMap<
+        AccountId,
+        (
+            Option<dtos::BackupServiceInfo>,
+            Option<dtos::DestinationNodeInfo>,
+        ),
+    > {
         log!("migration_info");
         self.node_migrations.get_all()
     }
@@ -2145,7 +2151,7 @@ impl MpcContract {
     #[handle_result]
     pub fn register_backup_service(
         &mut self,
-        backup_service_info: BackupServiceInfo,
+        backup_service_info: dtos::BackupServiceInfo,
     ) -> Result<(), Error> {
         let account_id = Self::assert_caller_is_signer();
         log!(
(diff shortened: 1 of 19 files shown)


crates/contract/src/lib.rs
```diff
@@ -2256,32 +2262,23 @@ impl MpcContract {
...
+        // ensure that this node has a valid TEE quote.
         let node_id = NodeId {
             account_id: account_id.clone(),
-            account_public_key,
-            tls_public_key,
+            account_public_key: expected_destination_node.signer_account_pk,
+            tls_public_key: expected_destination_node
+                .destination_node_info
+                .sign_pk
+                .clone(),
```
gilcu3, line 2281: this misleading name (`sign_pk`) change will get finally fixed in #2961 , right after this PR

gilcu3:
@claude review

yes, we are aware of the borsh incompatibility, and we guarantee that ongoing_migrations is empty, as no migrations have yet started in prod

crates/near-mpc-contract-interface/src/types/node_migrations.rs
```diff
@@ -0,0 +1,25 @@
+use crate::types::participants::ParticipantInfo;
+use borsh::{BorshDeserialize, BorshSerialize};
+use near_mpc_crypto_types::Ed25519PublicKey;
```
kevindeforth: Q: should we replace this with the public re-export in `src/lib`?
I think this would simplify future refactors.

crates/contract/src/node_migrations.rs
```diff
@@ -153,10 +141,10 @@ mod tests {
     fn test_ongoing_migrations() {
         let mut migrations = NodeMigrations::default();
         let (account_id, participant_info) = gen_participant(0);
-        let signer_account_pk = bogus_ed25519_near_public_key();
         let destination_node_info = DestinationNodeInfo {
-            signer_account_pk,
-            destination_node_info: participant_info,
+            signer_account_pk: Ed25519PublicKey::try_from(&bogus_ed25519_near_public_key())
+                .unwrap(),
```
kevindeforth: nit: Would be nice to have a helper method that directly returns the required key, to avoid these wrappers.

crates/contract/src/lib.rs
```diff
@@ -3666,15 +3667,15 @@ mod tests {
         let url_id: usize = rand::random();
         let (_, participant_info) = gen_participant(url_id);
         DestinationNodeInfo {
-            signer_account_pk: node_signer_pk,
-            destination_node_info: participant_info,
+            signer_account_pk: Ed25519PublicKey::try_from(&node_signer_pk).unwrap(),
```
kevindeforth: nit: We avoid the conversion here if we call `bogus_ed25519_public_key` instead of `bogus_ed25519_near_public_key`,

kevindeforth (dismissed):
many thanks!

kevindeforth (approved):

anodar (approved):

PR #2978, opened 2026-04-21
chore: bump nearcore to 2.11.1

gilcu3:
Related to #2975

### DIFF
--- a/crates/e2e-tests/src/cluster.rs
+++ b/crates/e2e-tests/src/cluster.rs
@@ -21,7 +21,7 @@ use crate::mpc_node::{MpcNode, MpcNodeSetup, MpcNodeSetupArgs, NodePorts};
 use crate::near_sandbox::NearSandbox;
 use crate::port_allocator::E2ePortAllocator;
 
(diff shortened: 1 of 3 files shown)


crates/e2e-tests/src/cluster.rs
```diff
@@ -21,7 +21,7 @@ use crate::mpc_node::{MpcNode, MpcNodeSetup, MpcNodeSetupArgs, NodePorts};
 use crate::near_sandbox::NearSandbox;
 use crate::port_allocator::E2ePortAllocator;
 
-const DEFAULT_SANDBOX_VERSION: &str = "2.11.0";
```
netrome, line 24: Nice touch, do we have any mechanism to remember this?

netrome (approved):
Thanks!

crates/e2e-tests/src/cluster.rs
gilcu3: yes, the fast checks fail, we have a script that take care of it

PR #2984, opened 2026-04-21
refactor: create generic config type for all foreign chains

### LINKED ISSUE #2866: API variant should not be specific per chain
### Background

Currently we have API variant per chain for the foreign chain config on the nodes. Given that we are shifting the foreign chain configuration to not require consensus on the RPC provider, but instead only require nodes to submit their URL/name of the provider, it does not make sense to have an enum for the RPC provider name for every chain.

See thred: https://github.com/near/mpc/pull/2823/changes#r3078656414

### User Story

_No response_

### Acceptance Criteria

remove [CHAIN_NAME]ApiVariant enum in the config for nodes

_No response_

DSharifi:
closes #2866


DSharifi:
@claude review

crates/node-config/src/foreign_chains.rs
```diff
@@ -529,175 +510,6 @@ foreign_chains:
...
     }
 
-    #[test]
-    fn config_parsing__should_fail_when_max_retries_is_zero_for_solana() {
-        // Given
-        let yaml = r#"
-my_near_account_id: test.near
-near_responder_account_id: test.near
-number_of_responder_keys: 1
-web_ui:
```
DSharifi, line 539: These suite of tests are now covered by the type system since we use NonZeroU64 instead of u64

DSharifi:
@claude review

DSharifi:
> Cross-chain duplicate RPC URL validation is now stricter than before (potential config breakage)

This is intentional. We shouldn't allow the same URL for RPC providers across chains.

crates/node-config/src/foreign_chains.rs
```diff
@@ -1,83 +1,91 @@
...
-            && self.ethereum.is_none()
-            && self.abstract_chain.is_none()
-            && self.starknet.is_none()
-            && self.bnb.is_none()
-            && self.base.is_none()
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct ForeignChainConfig {
+    pub timeout_sec: NonZeroU64,
+    pub max_retries: NonZeroU64,
+    // TODO: what is the key here?
```
netrome: Yeah I'd like to know this as well. I assume this is something you intended to resolve?

netrome: Ah I see, this is based on the previous setup where providers was a map from the provider name (e.g. "abstract") and the config of the provider, such as the API variant of the provider.

crates/node-config/src/foreign_chains.rs
```diff
@@ -1,83 +1,91 @@
...
+    pub providers: NonEmptyBTreeMap<String, ForeignChainProviderConfig>,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct ForeignChainProviderConfig {
+    pub rpc_url: String,
+    pub api_variant: RpcProvider,
```
netrome: This looks strange. It seems like we're confusing the API variant and RPC provider concepts here. I see the existing code confuses them a bit already, but they are separate concepts. The RPC provider denotes "who is the provider of the RPC", whereas the "API variant" denotes which type of an API does this provider support.

Most chains typically have pretty standard API interfaces, but these can vary and there can be different common flavors of it. For example for bitcoin, both the standard bitcoin core RPC format and the esplora API format are common.

I don't have more examples top-of-mind, so if the EVM chain RPC providers all use basically the same API perhaps we could consider simplifying this model, but the way I read this it looks very confusing.

netrome (changes requested):
Two blockers:

1. A dangling todo.

2. It seems like we're confusing the concepts of RPC providers and API variants in the current change. Those should be disentangled so it's possible to reuse API variants across RPC providers if they coincide, or configure different API variants for the same provider for any provider supporting multiple.

Happy to jump on a call and discuss this tomorrow.

crates/node-config/src/foreign_chains.rs
netrome: Discussing this further, it seems like all providers provide "standard" APIs. As we don't have an immediate need to differentiate between API variants, let's just drop this field for now and add it when it's needed.

crates/node-config/src/foreign_chains.rs
DSharifi: Discussed sync. Currently all rpc providers we work with have the same schema per chain, so we the API variant is not needed.

If we ever stumble on a problem with RPC providers not supporting a standard schema we will revisit

netrome (dismissed):
So much red! 🤩

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -360,10 +360,11 @@ mod tests {
     use crate::indexer::MockReadForeignChainPolicy;
     use assert_matches::assert_matches;
     use mpc_node_config::{
-        BitcoinApiVariant, BitcoinChainConfig, BitcoinProviderConfig, ForeignChainsConfig,
+        foreign_chains::RpcProviderName, ForeignChainConfig, ForeignChainProviderConfig,
+        ForeignChainsConfig,
     };
     use near_mpc_bounded_collections::NonEmptyBTreeSet;
-    use std::collections::BTreeMap;
+    use std::{collections::BTreeMap, num::NonZeroU64};
```
gilcu3, line 367: nice I did not know that type was there

gilcu3 (dismissed):
Thank you!

netrome (dismissed):

crates/node/src/providers/verify_foreign_tx/sign.rs
netrome: It's a nice one :)

gilcu3 (approved):

netrome (approved):

PR #2988, opened 2026-04-22
test: port migration endpoint and migration service from pytest to Rust e2e

### LINKED ISSUE #2997: test: port migration endpoint and migration service from pytest to Rust e2e
### Background

Part of the pytest → Rust e2e migration effort tracked in #2889.

Ports the `/debug/migrations` endpoint state-tracking test and the full end-to-end node migration workflow via the backup CLI.

### Acceptance Criteria

**Port `test_migration_endpoint`:** verify `/debug/migrations` endpoint tracks migration state
- [ ] Initial migration state is empty
- [ ] After setting backup service info on a node, migration state updates accordingly
- [ ] After initiating node migration, migration state includes destination node info
- [ ] Contract and endpoint migration state remain synchronized

**Port `test_migration_service`:** full end-to-end node migration via backup CLI
- [ ] 2 participating nodes + 2 target nodes; compile backup CLI
- [ ] Backup service registration on participating nodes
- [ ] GET keyshares from source node via backup service
- [ ] Initiate node migration
- [ ] PUT keyshares to target node via backup service
- [ ] Verify target node becomes active participant (contract state updated)
- [ ] Verify migration state clears from contract
- [ ] Kill old node; verify new node handles sign requests successfully

### Resources

- PR: #2988
- Parent: #2889

anodar:
Closes: https://github.com/near/mpc/issues/2997
Work towards: https://github.com/near/mpc/issues/2889


crates/e2e-tests/src/cluster.rs
```diff
@@ -157,9 +183,20 @@ impl MpcCluster {
...
 
+        // Pre-generate keys for migration target nodes.
+        // Migration targets share the source's NEAR account, so node_keys
+        // (which stores NEAR signer keys) gets the source's key.
+        for (i, &source_idx) in config.migration_targets.iter().enumerate() {
+            let target_idx = config.num_nodes + i;
+            node_keys.push(node_keys[source_idx].clone());
+            operator_keys.push(generate_deterministic_key(
+                KEY_SEED_MIGRATION_OPERATOR + target_idx as u64,
+            ));
```
netrome: Why do we push new operator keys? Do we need the operator keys to differ for the new and old node? I'd expect the typical production scenario that the same operator manages both the old and new node, so it would be nice to mirror that in the test.

netrome: Note - not a blocker since I don't think the current behavior is incorrect.

crates/e2e-tests/tests/migration_endpoint.rs
```diff
@@ -0,0 +1,167 @@
...
+/// Verify the `/debug/migrations` endpoint tracks migration state.
+///
+/// For each node:
+///   1. Initial migration state is empty
+///   2. After registering a backup service, migration state updates accordingly
+///   3. After initiating node migration, migration state includes destination node info
+///   4. Contract and endpoint migration state remain synchronized
+#[tokio::test]
+#[expect(non_snake_case)]
+async fn migration_endpoint__should_track_migration_state() {
```
netrome, line 27: Hmm, this test is a bit convoluted and seems to lack a clear separation between the action being performed and the expected outcome. Is there a strong need to continuously wait for the endpoints matching the expected migrations?

Also, if the purpose of the test is to verify the migration endpoint serving consistent migration state with the contract, do we need to migrate several nodes? Shouldn't it be sufficient to initiate a node migration, query the migration endpoint, finish the migration, query the endpoint and assert the expected results of these queries?

crates/e2e-tests/tests/migration_service.rs
```diff
@@ -0,0 +1,463 @@
...
+            .kill_nodes(&[source_idx])
+            .expect("failed to kill source node");
+
+        common::send_sign_request(&cluster, &running, &mut rng, cluster.default_user_account())
+            .await;
+        common::send_ckd_request(&cluster, &running, &mut rng, cluster.default_user_account())
+            .await;
+
+        tracing::info!(source_idx, target_idx, "migration completed successfully");
+    }
```
netrome, line 469: This test also lacks a clear asserted outcome, Shouldn't we query the contract state after the migration and assert that the TLS keys of the active participants match the migrated nodes?

netrome (changes requested):
Haven't done a full review, most of what I've seen looks pretty good but the test cases seem to lack clear asserted outcomes.

I really like the "//Given //When //Then" framework as it forces us to reflect over what we're actually testing. It would be nice if we could make the new tests fit this framework better, even though I know some of the legacy pytests also have very unclear separation between setup, action and outcome. But I hope we can hold the new tests to a somewhat higher standard.

netrome (commented):
Edit, my comments on the test are subjective so it's not a hard blocker. I'll continue reviewing later today or tomorrow.

crates/e2e-tests/tests/migration_service.rs
anodar: `wait_for_migration_completion` above verifies that with:
```
anyhow::ensure!(
            running_state_matches_participant_key(&state, &source_account_id, &target_p2p_key),
            "target node not yet active participant"
        );
```

crates/e2e-tests/src/cluster.rs
anodar: This was following pytest. Changed to using same key.

crates/e2e-tests/tests/migration_endpoint.rs
anodar: Discussed offline. Added more comments to make it explicit it's series of given/when/then's. Left migration of several nodes as switching to single doesn't simplify it much.

crates/e2e-tests/src/cluster.rs
```diff
@@ -157,9 +182,18 @@ impl MpcCluster {
...
-        let (node_keys, node_near_keys, node_p2p_keys, operator_keys) =
+        let (mut node_keys, node_near_keys, node_p2p_keys, mut operator_keys) =
             generate_signing_keys(u64::try_from(config.num_nodes).unwrap());
 
+        // Pre-generate keys for migration target nodes.
+        // Migration targets share the source's NEAR account and operator,
+        // mirroring the production scenario of one operator managing both
+        // the old and new node.
+        for &source_idx in &config.migration_targets {
+            node_keys.push(node_keys[source_idx].clone());
```
kevindeforth: What are `node keys`? Are those the keys the node uses to sign transactions on chain?

crates/e2e-tests/src/cluster.rs
```diff
@@ -666,6 +694,36 @@ impl MpcCluster {
...
+            .client_for(node.account_id().as_ref(), &self.operator_keys[node_index])
+    }
+
+    /// Register backup service info for a node.
+    pub async fn register_backup_service(
+        &self,
+        node_index: usize,
+        backup_service_info: serde_json::Value,
+    ) -> anyhow::Result<near_kit::FinalExecutionOutcome> {
+        let client = self.client_for(node_index)?;
```
kevindeforth: does `client_for` sign with the operator key, or with the node key?

kevindeforth (commented):
partial review, some questions

crates/e2e-tests/src/cluster.rs
anodar: With operator key. In multiple places I needed to do:

```
let node = &self.nodes[node_index];
        self.blockchain
            .client_for(node.account_id().as_ref(), &self.operator_keys[node_index])

```


So defined client_for on cluster itself instead to avoid code duplication.

crates/e2e-tests/src/cluster.rs
anodar: Yes node key is used for signing transactions whereas operator keys is used for governance (registering backup service, voting, etc..)

crates/e2e-tests/src/cluster.rs
kevindeforth: Good! Orthogonal to this PR: how would you feel about renaming this to `operator_client` or something else that would make this clearer?

crates/e2e-tests/src/cluster.rs
kevindeforth: Hmm, in that case, we should have new keys for the migration targets. A new node is expected to have a new key as well.

We can re-use the operator key.

kevindeforth (commented):
looks pretty good. Just a comment for key re-use.

crates/e2e-tests/src/cluster.rs
anodar: Renamed to `operator_client_for`.

crates/e2e-tests/src/cluster.rs
anodar: Sounds good, done.

crates/e2e-tests/tests/migration_service.rs
```diff
@@ -0,0 +1,470 @@
...
+        .await
+        .expect("failed to get contract state");
+    backup_service.set_contract_state(&contract_state);
+
+    let target_p2p_key = cluster.nodes[target_idx].p2p_public_key_str();
+    backup_service.put_keyshares(
+        &target_migration_addr,
+        &target_p2p_key,
+        cluster.nodes[target_idx].backup_encryption_key_hex(),
+    );
```
kevindeforth, line 343: out of scope nit: would be nice to get the return value of these ops (to e.g. assert success / failure)

kevindeforth (approved):
Thank you!

crates/e2e-tests/tests/migration_service.rs
anodar: Makes sense, will follow up separately.

crates/e2e-tests/src/cluster.rs
```diff
@@ -157,9 +183,22 @@ impl MpcCluster {
...
+        // Migration targets share the source's NEAR account and operator
+        // (mirroring the production scenario of one operator managing both
+        // the old and new node), but get a fresh NEAR signer key so the new
+        // node's transactions are distinguishable from the source's.
+        for (i, &source_idx) in config.migration_targets.iter().enumerate() {
+            let target_idx = config.num_nodes + i;
+            node_keys.push(generate_deterministic_key(
+                KEY_SEED_MIGRATION_NEAR_SIGNER + target_idx as u64,
+            ));
+            operator_keys.push(operator_keys[source_idx].clone());
```
netrome, line 199: Funny how we flipped which one we generate and which one we copy since yesterday 😅

crates/e2e-tests/src/cluster.rs
```diff
@@ -954,6 +1056,34 @@ fn start_mpc_nodes(
         })?;
         nodes.push(MpcNodeState::Running(setup.start()?));
     }
+
+    // Start migration target nodes alongside the participants so their
+    // near-indexers sync from the same early point in the chain.
```
netrome, line 1061: Is this important? The migration target nodes could be started separately afaik.

crates/e2e-tests/tests/migration_endpoint.rs
```diff
@@ -0,0 +1,165 @@
...
+            outcome.is_success(),
+            "register_backup_service failed: {:?}",
+            outcome.failure_message()
+        );
+        let backup_info = BackupServiceInfo {
+            public_key: node_state.p2p_public_key(),
+        };
+        expected_migrations.insert(account_id.clone(), (Some(backup_info.clone()), None));
+
+        // Then: contract and endpoint both expose the backup registration.
```
netrome, line 66: I wasn't sure how this would pan out after our discussion, but reading this now I find these comments very helpful. Thanks for adding them 🙏

netrome (approved):
Thank you! 🙏

crates/e2e-tests/src/cluster.rs
anodar: It is important otherwise we need to increase timeout so that node catches up with the blocks.

crates/e2e-tests/tests/migration_endpoint.rs
anodar: Glad to hear! 👍

PR #3001, opened 2026-04-22
feat!: nodes don't require unanimous conensus on RPC providers

### LINKED ISSUE #2913: nodes use new API for submitting which foreign chains they support
### Background

In https://github.com/near/mpc/pull/2784 we added a new API on the contract for nodes to submit which chains they support.

The node should use this new endpoint.

A draft PR for this is already implemented in https://github.com/near/mpc/pull/2779

### User Story

_No response_

### Acceptance Criteria

The node uses the new contract methods introduced in https://github.com/near/mpc/pull/2784.

_No response_

DSharifi:
closes #2913 

This PR builds on top of github.com/near/mpc/pull/2784 which added a new API for nodes to submit the configuration of RPC providers they support.

This PR updates the node to use the new API and deprecates the old "policy" based API for foreign chains on the contract.

The breaking changes are fine, since we upgrade the nodes before the contract, and we have the new API already rolled out to the contract in the previous 3.9.1 release.


DSharifi:
@claude reviewe

DSharifi:
@claude review

crates/node/src/coordinator.rs
```diff
@@ -728,81 +729,6 @@ where
...
-    fn is_supported_foreign_chain(chain: &dtos::ForeignChain) -> bool {
-        matches!(
-            chain,
-            dtos::ForeignChain::Abstract
-                | dtos::ForeignChain::Solana
-                | dtos::ForeignChain::Bitcoin
-                | dtos::ForeignChain::Ethereum
-                | dtos::ForeignChain::Starknet
-        )
-    }
```
DSharifi, line 805: If it can be configured then it must be supported to avoid checks like this. I'll follow up removing all chains from the config struct that are not supported.

Adding new chains in the future to the config should be a one liner thanks to https://github.com/near/mpc/pull/2984

DSharifi: Added tracking issue https://github.com/near/mpc/issues/3002

crates/node/src/coordinator.rs
```diff
@@ -379,14 +377,17 @@ where
...
         };
 
-        if let Err(err) = Self::maybe_vote_foreign_chain_policy(
-            &config_file,
-            &foreign_chain_policy_reader,
-            &chain_txn_sender,
-        )
-        .await
+        // Register locally supported foreign chains with the contract.
+        let foreign_chain_configuration = config_file.foreign_chains.configured_chains();
```
anodar, line 381: Why do we do this unconditionally? Why not fetch it first, see if it's already registered and skip like it was before?

crates/contract/tests/sandbox/foreign_chain_policy.rs
```diff
@@ -4,202 +4,11 @@ use crate::sandbox::common::SandboxTestSetup;
...
-use near_mpc_contract_interface::method_names::{
-    GET_FOREIGN_CHAIN_POLICY_PROPOSALS, REGISTER_FOREIGN_CHAIN_CONFIG, VOTE_FOREIGN_CHAIN_POLICY,
-};
+use near_mpc_contract_interface::method_names::REGISTER_FOREIGN_CHAIN_CONFIG;
 use near_mpc_contract_interface::types as dtos;
 use serde_json::json;
 use std::collections::BTreeMap;
 
-#[tokio::test]
-async fn vote_foreign_chain_policy__should_reject_empty_rpc_providers() {
```
anodar, line 15: why did we drop these tests?

anodar (dismissed):

pbeza (approved):

pbeza (changes requested):
I think some docs still still refer to the old design:

https://github.com/near/mpc/blob/a66faad719b48aa1870a6fd0b4d507678cf8242f/docs/foreign-chain-transactions.md?plain=1#L358-L364

Might be worth asking Claude to help spot more outdated bits too — I only skimmed the docs since this PR is pretty large. But it looks like we may still be referencing some old structs there.

Possibly also here:
https://github.com/near/mpc/blob/a66faad719b48aa1870a6fd0b4d507678cf8242f/docs/foreign-chain-transactions.md?plain=1#L292-L294

pytest/tests/test_foreign_chain_policy_auto_voting.py
```diff
@@ -68,54 +67,49 @@ def foreign_chain_policy_cluster():
...
+    assert "Solana" not in supported, (
+        "Solana should not be supported when not all participants registered it"
+    )
 
+    # Have node 3 register Solana support directly on the contract.
     cluster.call_contract_function_with_account_assert_success(
         mpc_nodes[2],
-        "vote_foreign_chain_policy",
-        {"policy": EXPECTED_POLICY},
+        "register_foreign_chain_config",
```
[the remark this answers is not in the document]
DSharifi: it's 0 based indexing

crates/contract/tests/sandbox/foreign_chain_policy.rs
DSharifi: ~~The test shouldn't have been there in the first place so I decided to delete it instead of porting. It's testing that the contract rejects policies that are empty, which should have been a contract sandbox test.~~

~~Anyway, in the new setup we want a view over all participants' configuration, even if it's empty.~~

Edit:
My bad I commented on wrong file. These tests are testing the API that is being deprecated. The tests are testing the API which is being deleted in this PR. The new API is already tested below these tests in the same file

crates/node/src/coordinator.rs
DSharifi: Because it's very very cheap to submit this transaction, and is only done once at startup. It's and idempotent operation so the effect is the same.

This saves us ~100 lines of logic and having to poll the contract.

pbeza (approved):

anodar (approved):
Thank you!

PR #3010, opened 2026-04-23
feat(tee-authority): support multiple PCCS endpoints with fallback

### LINKED ISSUE #2945: Support multiple PCCS endpoints with fallback
### Background

After #2944 (replacing Phala's collateral endpoint with direct PCCS calls), the node supports a single configurable `pccs_url`. If that PCCS server is unavailable, attestation fails until it recovers.

Supporting multiple PCCS endpoints with fallback would improve resilience — the node could try a local Intel PCCS first, then fall back to `pccs.phala.network` or `api.trustedservices.intel.com`.

Raised in https://github.com/near/mpc/issues/1545#issuecomment-4260871740.

### Acceptance Criteria

- [ ] `pccs_url` config accepts a list of PCCS URLs (or a new `pccs_urls` field alongside the existing one)
- [ ] Node tries each endpoint in order until one succeeds
- [ ] Existing single-URL config remains backward-compatible

- `dcap_qvl::collateral::get_collateral(pccs_url, quote)` accepts a single URL — fallback logic would wrap multiple calls
- Possible PCCS sources: local Intel PCCS, `pccs.phala.network`, `api.trustedservices.intel.com`

pbeza:
Closes #2945.

Replace the single `pccs_url` with a `NonEmptyVec<Url>` so the collateral fetcher can try multiple PCCS endpoints in order, returning the first success. With one URL the behavior is unchanged.

## Breaking config change

Rename `pccs_url: Url` → `pccs_urls: NonEmptyVec<Url>`. Operators must migrate:

```diff
- pccs_url  = "https://pccs.phala.network"
+ pccs_urls = ["https://pccs.phala.network"]
```

Multi-URL example:

```toml
pccs_urls = [
    "http://localhost:8081",
    "https://pccs.phala.network",
    "https://api.trustedservices.intel.com",
]
```

This drops issue #2945's backward-compatibility requirement in favor of a single API with no `#[serde(untagged)]` shim.


crates/tee-authority/src/tee_authority.rs
```diff
@@ -164,6 +184,42 @@ impl TeeAuthority {
...
+    let mut last_err: Option<anyhow::Error> = None;
+    for url in pccs_urls.iter() {
+        match fetcher(url.as_str().to_owned()).await {
+            Ok(collateral) => return Ok(collateral),
+            Err(err) => {
+                warn!(
+                    ?err,
+                    %url,
+                    "failed to fetch collateral from PCCS, trying next endpoint"
+                );
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/pull/3010/commits/37f2442a346d5e276646cab308cc3200648b372d

crates/node/src/run.rs
```diff
@@ -59,7 +59,7 @@ pub async fn run_mpc_node(config: StartConfig) -> anyhow::Result<()> {
             launcher_interface::types::TeeAuthorityConfig::Local => "local",
         },
         image_hash = %config.tee.image_hash,
-        pccs_url = %config.pccs_url,
+        pccs_urls = ?config.pccs_urls,
```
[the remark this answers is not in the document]
pbeza: Done in ea31e187 — switched to a Display-joined comma-separated list.

crates/node-config/src/start.rs
```diff
@@ -199,4 +204,97 @@ mod tests {
...
+
+    /// An empty vec is explicitly rejected by [`NonEmptyVec`]'s bound.
+    #[test]
+    fn pccs_urls__should_reject_empty_vec() {
+        // Given
+        let empty: Vec<url::Url> = vec![];
+
+        // When
+        let err = NonEmptyVec::<url::Url>::try_from(empty).unwrap_err();
+
```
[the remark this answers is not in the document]
pbeza: The substring-on-toml-error version is fragile (`serde::de::Error::custom` only exposes a string). The typed `LowerBoundError` version got reviewed as testing `NonEmptyVec` rather than our wiring, so I dropped it in 547b256d. The remaining three `pccs_urls__*` tests cover what we own (single-element parse, multi-element order preservation, default-when-omitted).

deployment/cvm-deployment/user-config.toml
```diff
@@ -11,8 +11,10 @@ port_mappings = [
 
 [mpc_node_config]
 home_dir = "/data"
-# Override the PCCS URL for TDX attestation collateral. Defaults to Phala's PCCS if omitted.
-# pccs_url = "https://pccs.phala.network"
+# PCCS URLs for TDX attestation collateral, tried in order until one succeeds.
+# At least one URL is required. Defaults to Phala's PCCS if omitted.
```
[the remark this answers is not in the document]
pbeza: Reworded as suggested in ea31e187.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -164,6 +184,43 @@ impl TeeAuthority {
...
+                    "failed to fetch collateral from PCCS; {}",
+                    if is_last_endpoint { "no more endpoints remain" } else { "trying next endpoint" }
+                );
+                last_err = Some(err);
+            }
+        }
+    }
+    // NonEmptyVec guarantees at least one iteration above, which either
+    // returns Ok(...) or populates last_err.
+    Err(last_err.expect("NonEmptyVec guarantees at least one URL"))
```
barakeinav1, line 221: When all endpoints fail, the returned error is just the last URL's error. Suggest aggregating into a single error that lists every endpoint tried with its own failure, e.g.:

```
all 3 PCCS endpoints failed:
  - http://localhost:8081/: connection refused
  - https://pccs.phala.network/: 503 Service Unavailable
  - https://api.trustedservices.intel.com/: DNS resolution failed
```

So a structured-error consumer (alerting, tests, support tickets) sees the full picture without having to scrape warn logs. Non-blocking.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -164,6 +184,43 @@ impl TeeAuthority {
...
+    Fetcher: Fn(Url) -> Fut,
+    Fut: Future<Output = anyhow::Result<Collateral>>,
+{
+    let mut last_err: Option<anyhow::Error> = None;
+    let total_endpoints = pccs_urls.len();
+    for (index, url) in pccs_urls.iter().enumerate() {
+        let attempt = index + 1;
+        let is_last_endpoint = attempt == total_endpoints;
+        match fetcher(url.clone()).await {
+            Ok(collateral) => return Ok(collateral),
```
barakeinav1, line 205: A successful fetch on attempt 2+ produces no log line — only the warn from the failed primary, which an operator has to read in the negative ("no further warns ⇒ fallback worked"). That makes silent primary degradation invisible to log-based alerting: a primary that fails 100% of the time, masked by a healthy fallback, looks identical to a healthy system. Suggest a single `info!` on success when `attempt > 1` to give that case a positive signal:

```rust
Ok(collateral) => {
    if attempt > 1 {
        info!(%url, attempt, total_endpoints, "fetched collateral via PCCS fallback");
    }
    return Ok(collateral);
}
```

Common path (attempt == 1) stays silent. Non-blocking.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -148,9 +153,24 @@ impl TeeAuthority {
...
+    ) -> anyhow::Result<Collateral> {
+        try_each_pccs_endpoint(pccs_urls, async |url: Url| {
+            Self::fetch_collateral_from(&url, quote).await
+        })
+        .await
+    }
+
+    /// Fetches attestation collateral from a single PCCS endpoint, with the
+    /// usual per-request timeout and a single retry via exponential backoff.
+    async fn fetch_collateral_from(pccs_url: &Url, quote: &[u8]) -> anyhow::Result<Collateral> {
```
barakeinav1, line 172: Non-blocking — surfacing a choice rather than pushing for it.

Retry is currently per-URL (the `get_with_backoff(..., Some(1))` lives inside `fetch_collateral_from`), so URL 1 is tried twice with backoff before URL 2 is touched. A transient blip on URL 1 delays the fallback by ~21s.

Alternative: hoist the retry to wrap the whole cycle, so each URL gets one shot per cycle and the cycle itself is retried.

```rust
async fn fetch_collateral(
    pccs_urls: &NonEmptyVec<Url>,
    quote: &[u8],
) -> anyhow::Result<Collateral> {
    get_with_backoff(
        async || try_each_pccs_endpoint_once(pccs_urls, quote).await,
        "fetch collateral from any PCCS endpoint",
        Some(1),
    )
    .await
}
```

Trade-off: faster failover for transient blips (~10s vs ~21s), at the cost of not retrying a slow primary before falling through. For `[local, phala, intel]` style lists where fallbacks are different vendors, the list diversity *is* the redundancy and per-URL retry is largely redundant on top of it.

Worst-case total latency is the same order either way. Happy to leave the current shape if there's a reason for the per-URL retry I'm missing.

barakeinav1:
Heads-up: this PR conflicts with #3026 (PCCS self-signed TLS) — both touch the same config surface across `start.rs`, `tee_authority.rs`, `run.rs`, `cli.rs`, the e2e fixture, and `user-config.toml`. Suggest this PR lands first and I will submit #3026 after.

barakeinav1:
Breaking config change worth flagging: `pccs_url` → `pccs_urls` is a rename plus a type change (string → array), and the schema isn't backward-compatible. Failure mode is silent — serde ignores unknown fields by default, so an operator who keeps `pccs_url = "..."` and forgets to migrate gets the Phala default with no error or log line.

We're OK with this for now: one TEE node on testnet, under team control, migration burden is nil.

One ask: **update #2945's acceptance criteria** to drop "Existing single-URL config remains backward-compatible" — it's currently still listed as a requirement, and amending the issue records the deviation against the original spec rather than just in the PR body.

barakeinav1 (approved):

gilcu3 (approved):
LGTM

About the order of urls and retries. Which variant is better depends on url list semantic. If it is ordered by priority/preference, then it makes sense to keep as we have. If they are pretty much equivalent, then it would make sense to optimize for latency as mentioned by @barakeinav1

crates/tee-authority/src/tee_authority.rs
pbeza: Good insight, but I agree with @gilcu3’s [take](https://github.com/near/mpc/pull/3010#pullrequestreview-4186554650) on this, so I’d prefer to keep it as-is for now.

crates/tee-authority/src/tee_authority.rs
pbeza: Good point. I’ll follow up on this in a separate PR: https://github.com/near/mpc/pull/3048.

crates/tee-authority/src/tee_authority.rs
pbeza: Good point. I’ll follow up on this in a separate PR: https://github.com/near/mpc/pull/3048.

PR #3017, opened 2026-04-24
refactor: don't panic in test helpers

### LINKED ISSUE #3011: Don't panic in test helpers
### Background

Refactor test helpers to return Results and assert success/failures at call sites.

Reference: https://github.com/near/mpc/pull/2988#discussion_r3131210944

### User Story

_No response_

### Acceptance Criteria

No panic in test helpers

_No response_

anodar:
Closes: https://github.com/near/mpc/issues/3011

### DIFF
--- a/crates/e2e-tests/tests/ckd_verification.rs
+++ b/crates/e2e-tests/tests/ckd_verification.rs
@@ -1,5 +1,6 @@
 use crate::common;
 
+use anyhow::Context;
 use blstrs::{G1Projective, G2Projective, Scalar};
 use group::Group as _;
 use group::ff::Field as _;
@@ -24,16 +25,16 @@ fn verify_ckd(
     private_key: Scalar,
     big_y: &Bls12381G1PublicKey,
     big_c: &Bls12381G1PublicKey,
-) -> bool {
-    let big_y = G1Projective::try_from(big_y).expect("invalid big_y G1 point");
-    let big_c = G1Projective::try_from(big_c).expect("invalid big_c G1 point");
-    let mpc_pk = G2Projective::try_from(mpc_public_key).expect("invalid MPC G2 key");
+) -> anyhow::Result<bool> {
+    let big_y = G1Projective::try_from(big_y).context("invalid big_y G1 point")?;
+    let big_c = G1Projective::try_from(big_c).context("invalid big_c G1 point")?;
+    let mpc_pk = G2Projective::try_from(mpc_public_key).context("invalid MPC G2 key")?;
 
     let mpc_vk = VerifyingKey::new(mpc_pk);
     let confidential_key = CKDOutput::new(big_y, big_c).unmask(private_key);
     let app_id = derive_app_id(account_id, path);
 
-    verify_signature(&mpc_vk, app_id.as_ref(), &confidential_key).is_ok()
+    Ok(verify_signature(&mpc_vk, app_id.as_ref(), &confidential_key).is_ok())
 }
 
 /// Verify that a CKD response (AppPublicKey variant) is mathematically correct.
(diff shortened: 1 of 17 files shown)


crates/e2e-tests/tests/ckd_verification.rs
```diff
@@ -91,8 +94,9 @@ async fn ckd_response__passes_cryptographic_verification() {
 #[expect(non_snake_case)]
 async fn ckd_pv_response__passes_cryptographic_verification() {
     // given
-    let (cluster, running) =
-        common::setup_cluster(common::CKD_PV_VERIFICATION_PORT_SEED, |_| {}).await;
+    let (cluster, running) = common::setup_cluster(common::CKD_PV_VERIFICATION_PORT_SEED, |_| {})
```
kevindeforth: I think it would be slightly more ergonomic if `setup_cluster` would still panic. If it fails, we can't do any actual test logic, so panicking seems okay.

kevindeforth (commented):
Responded on slack, tldr is that I think we want to panic for everything that is related to test setup (for example, failure to submit a signature request, failure to fetch contract binaries), but that returning an error is preferred for everything that tests the actual system (for example, if we fail to get a signature in response to a request - only the actual test knows if this is expected or not).

crates/e2e-tests/tests/migration_endpoint.rs
```diff
@@ -136,19 +161,19 @@ async fn assert_contract_matches(cluster: &e2e_tests::MpcCluster, expected: &Mig
...
     )
     .await
-    .expect("timed out waiting for contract migration state to match");
+    .context("timed out waiting for contract migration state to match")
 }
 
 /// Assert the node's `/debug/migrations` endpoint eventually equals `expected`.
-/// Retries to absorb indexer lag, then panics on timeout.
+/// Retries to absorb indexer lag, then returns an error on timeout.
 async fn assert_endpoint_matches(
```
kevindeforth: nit: I usually don't care about theses things, because I feel like devs should have some flexibility about this.
But intuitively, I would expect a method named `assert_*` to panic if the assertion is wrong. Otherwise, `validate_*`, `check_*` or `ensure_` might be better prefixes. Wdyt?

crates/e2e-tests/tests/migration_service.rs
```diff
@@ -19,60 +20,69 @@ struct BackupService {
 }
 
 impl BackupService {
-    fn new(binary_path: PathBuf) -> Self {
-        Self {
-            home_dir: tempfile::tempdir().expect("failed to create backup service home dir"),
+    fn new(binary_path: PathBuf) -> anyhow::Result<Self> {
+        Ok(Self {
+            home_dir: tempfile::tempdir().context("failed to create backup service home dir")?,
```
kevindeforth: I would make this a panic. This is test-setup error, not our own logic.

crates/e2e-tests/tests/migration_service.rs
```diff
@@ -19,60 +20,69 @@ struct BackupService {
...
-            .args([
-                "--home-dir",
-                self.home_dir.path().to_str().unwrap(),
-                "generate-keys",
-            ])
+            .args(["--home-dir", self.home_dir_str()?, "generate-keys"])
             .output()
-            .expect("failed to run backup-cli generate-keys");
-        assert!(
+            .context("failed to run backup-cli generate-keys")?;
```
kevindeforth: Same here. I think an error at this point just means we couldn't run the binary (file not found, permission denied, invalid argument,...). It doesn't meant he binary didn't work as expected. So, a panic seems preferred.

kevindeforth (commented):
Thanks a lot!
Spotted a few places where I think panic is still preferred. Wondering about the `asset_*` methods, curious to hear your thoughts on those.

crates/e2e-tests/tests/migration_endpoint.rs
anodar: Changed to ensure_.

crates/e2e-tests/tests/migration_service.rs
anodar: Changed to panic with few more methods.

crates/e2e-tests/tests/migration_service.rs
anodar: Addressed.

kevindeforth (approved):
Thanks!

DSharifi (approved):

PR #3018, opened 2026-04-24
ci: factor out pytests into daily workflow

### LINKED ISSUE #2838: Move pytests to run daily rather than on every PR
### Background

As we're migrating them to Rust we want to have some grace period where we observe both Rust and Pytests to make sure we have same or more coverage in Rust.

### User Story

_No response_

### Acceptance Criteria

Pytests aren't required in Pr's anymore.

_No response_

anodar:
First step towards dropping pytests, submodule and workflow entirely. We'll have grace period (e.g. week) where we observe parity between pytest and rust e2e tests.

Before this is merged we should disable pytests as required, and enable `mpc-e2e-tests` as required.

Closes: https://github.com/near/mpc/issues/2838

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -321,10 +321,8 @@ jobs:
       - name: Build mpc-node
         run: cargo make build-mpc-node-network-hardship-simulation
 
-      # We run this build step on main as well to cache the contract build.
-      # The non reproducible build will not be used on main since it will be overwritten by the reproducible
-      # step below.
       - name: Build mpc-contract
+        if: github.ref != 'refs/heads/main'
         run: cargo make build-mpc-contract-optimized
 
       - name: Build mpc-contract reproducibly
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,133 +157,6 @@ jobs:
       - name: Run cargo-nextest
         run: cargo nextest run --cargo-profile=test-release --all-features --locked ${{ matrix.nextest-args }}
 
-  mpc-e2e-tests-build:
-    name: "MPC E2E tests: build"
-    runs-on: warp-ubuntu-2404-x64-16x
-    timeout-minutes: 60
-    permissions:
-      contents: read
-    env:
-      MPC_E2E_TESTS_BINARIES_CACHE_KEY: mpc-e2e-tests-binaries-${{ github.run_id }}
-
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
-        with:
-          persist-credentials: false
-
-      - name: Initialize submodules
-        run: git submodule update --init --recursive
-
-      - run: echo "WEEK=$(date -u +%Y-%U)" >> "$GITHUB_ENV"
-      - name: Cache Rust dependencies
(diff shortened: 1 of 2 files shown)


anodar:
@copilot review

.github/workflows/ci.yml
```diff
@@ -157,14 +157,14 @@ jobs:
       - name: Run cargo-nextest
         run: cargo nextest run --cargo-profile=test-release --all-features --locked ${{ matrix.nextest-args }}
 
-  mpc-pytest-build:
-    name: "MPC pytests: build"
+  mpc-e2e-tests-build:
```
netrome: Why do we need a separate build and cache cleanup step for the Rust E2E tests? Afaik we needed it for the pytests since they are executed in parallel (using `strategy.matrix`), but I don't see us using that for the Rust E2E tests yet.

Can we declare everything as a single job instead or is there any reason to have a separate build job that saves the binaries to a cache, and a cleanup job to clear it?

.github/workflows/pytests.yml
```diff
@@ -0,0 +1,239 @@
...
+          - group: "migration"
+            tests: "tests/test_migration_service.py"
+            runner: warp-ubuntu-2404-x64-2x
+          - group: "participant-info"
+            tests: "tests/test_submit_participant_info.py"
+            runner: warp-ubuntu-2404-x64-2x
+          - group: "lost-assets"
+            tests: "tests/test_lost_assets.py"
+            runner: warp-ubuntu-2404-x64-2x
+
```
netrome: If we run this as a nightly job I think we can just run all the pytests in a single job instead of having a matrix. The only reason for doing this is to speed up CI for PR reviews by parallelizing the tests. This is not a problem for a daily job, and we'd save some duplicated computations.

netrome (dismissed):
A few concerns, but not strictly blocking from my end.

.github/workflows/ci.yml
anodar: Good point. I'm assuming you mean why are they separate jobs (not separate steps). Merged it into single job like it was before pytests were parallelized.

.github/workflows/pytests.yml
anodar: I didn't bother with that as this workflow should be dropped entirely soon but makes sense, merged them all.

.github/workflows/ci.yml
```diff
@@ -531,28 +283,101 @@ jobs:
...
+      # step below.
+      - name: Build mpc-contract
+        run: cargo make build-mpc-contract-optimized
+
+      - name: Build mpc-contract reproducibly
+        if: github.ref == 'refs/heads/main'
+        run: |
+          cargo near build reproducible-wasm \
+            --manifest-path crates/contract/Cargo.toml \
+            --out-dir target/wasm32-unknown-unknown/release-contract
```
gilcu3: I do not understand why would we want to cache the non reproducible build in main

gilcu3 (dismissed):
LGTM

.github/workflows/ci.yml
anodar: I'm not sure it's needed, it was introduced in https://github.com/near/mpc/pull/2518/changes around pytest parallelization work. Dropped it.

netrome (approved):
Thanks!

gilcu3 (approved):

PR #3021, opened 2026-04-24
build(nix): add reproducible build flake for mpc-node

### LINKED ISSUE #3022: add reproducible build flake for mpc-node with nix
### Background

add reproducible build flake for mpc-node with nix

### User Story

As an MPC developer and nix user I'd like to be able create reproducible builds of the node binary with nix.

### Acceptance Criteria

add reproducible build flake for mpc-node with nix

_No response_

DSharifi:
closes https://github.com/near/mpc/issues/3022

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,8 @@ venv
 
 # Nix
 .direnv/
+result
+result-*
 
 .envrc
 
--- a/flake.nix
+++ b/flake.nix
@@ -8,6 +8,8 @@
       url = "github:oxalica/rust-overlay";
       inputs.nixpkgs.follows = "nixpkgs";
     };
+
+    crane.url = "github:ipetkov/crane";
   };
 
   outputs =
@@ -15,6 +17,7 @@
       self,
       nixpkgs,
       rust-overlay,
+      crane,
     }:
     let
       lib = nixpkgs.lib;
@@ -44,6 +47,12 @@
 
     in
     {
+      packages = forAllSystems (pkgs: {
+        mpc-node = pkgs.callPackage ./nix/mpc-node.nix {
+          inherit crane;
+        };
+      });
+
       devShells = forAllSystems (
         pkgs:
         let
--- a/flake.nix
+++ b/flake.nix
@@ -8,8 +8,6 @@
       url = "github:oxalica/rust-overlay";
       inputs.nixpkgs.follows = "nixpkgs";
     };
-
-    crane.url = "github:ipetkov/crane";
   };
 
   outputs =
@@ -17,7 +15,6 @@
       self,
       nixpkgs,
       rust-overlay,
-      crane,
     }:
     let
       lib = nixpkgs.lib;
@@ -47,12 +44,6 @@
 
     in
     {
-      packages = forAllSystems (pkgs: {
-        mpc-node = pkgs.callPackage ./nix/mpc-node.nix {
-          inherit crane;
-        };
-      });
-
       devShells = forAllSystems (
         pkgs:
         let
--- a/nix/mpc-node.nix
+++ b/nix/mpc-node.nix
@@ -2,6 +2,7 @@
   lib,
   pkgs,
   stdenv,
+  rustPlatform,
   rust-bin,
   llvmPackages_19,
   pkg-config,
@@ -14,61 +15,74 @@
   bzip2,
   udev,
   dbus,
-  apple-sdk_14,
-  darwin,
+  # Darwin-only. Defaulted to null so evaluating this package on Linux does
+  # not require a nixpkgs attr that may not exist there.
+  apple-sdk_14 ? null,
   crane,
 }:
 
 let
   llvmPkgs = llvmPackages_19;
+
+  # Pin the Rust toolchain to rust-toolchain.toml so Nix builds match what
+  # `cargo build` uses locally.
   rustToolchain = (rust-bin.fromRustupToolchainFile ../rust-toolchain.toml).override {
     extensions = [ "rust-src" ];
   };
 
   craneLib = (crane.mkLib pkgs).overrideToolchain rustToolchain;
-  clangResourceInclude =
-    let
-      clangVersion = lib.versions.major llvmPkgs.clang-unwrapped.version;
-    in
-    "${llvmPkgs.clang-unwrapped.lib}/lib/clang/${clangVersion}/include";
-  libcDev = lib.getDev stdenv.cc.libc;
+
   isX86 = stdenv.hostPlatform.isx86_64;
 
-  cargoToml = lib.importTOML ../Cargo.toml;
+  # Take the version from [workspace.package.version] so this file stays in
+  # sync on every release bump.
+  workspaceCargoToml = lib.importTOML ../Cargo.toml;
   pname = "mpc-node";
-  version = cargoToml.workspace.package.version;
+  version = workspaceCargoToml.workspace.package.version;
 
+  # Source filter. `filterCargoSources` keeps `.rs`, `Cargo.toml`, `Cargo.lock`
+  # and a couple of other Cargo-relevant files — anything else that is pulled
+  # in by `include_str!` / `include_bytes!` at compile time must be allow-
+  # listed here, otherwise the build fails with "file not found".
   src = lib.cleanSourceWith {
     src = craneLib.path ../.;
     filter =
       path: type:
-      # Keep standard Rust/Cargo files (.rs, .toml, .lock, etc.)
       (craneLib.filterCargoSources path type)
-
-      # license file
-      || (lib.hasSuffix "hird-party-licenses/licenses.html" path)
-
-      # README files that are included in docs
-      || (lib.hasSuffix "crates/contract-interface/README.md" path)
+      # `#![doc = include_str!("../README.md")]` on these two crates:
       || (lib.hasSuffix "crates/contract/README.md" path)
-
-      # TODO: --release should not need these assets.
-      || (lib.hasInfix "assets/" path);
+      || (lib.hasSuffix "crates/near-mpc-contract-interface/README.md" path)
+      # `include_str!` for the rendered third-party license bundle served by
+      # the node's web handler (crates/node/src/web.rs).
+      || (lib.hasSuffix "third-party-licenses/licenses.html" path)
+      # `include_bytes!` for historical signer contract snapshots
+      # (crates/contract-history/src/lib.rs).
+      || (lib.hasInfix "/crates/contract-history/archive/" path)
+      # Every `crates/*/assets/` directory — picked up by various
+      # `include_str!` / `include_bytes!` calls across the workspace.
+      || (lib.hasInfix "/assets/" path);
   };
 
+  # Pre-compute the vendored cargo deps so we can pass the exact same path to
+  # --remap-path-prefix below. Without an explicit handle, crane picks its own
+  # path and we cannot reference it for the remap.
   cargoVendorDir = craneLib.vendorCargoDeps { inherit src; };
 
   commonArgs = {
     inherit pname version src cargoVendorDir;
+
     strictDeps = true;
     cargoProfile = "reproducible";
     cargoExtraArgs = "-p mpc-node --bin mpc-node --locked";
 
     nativeBuildInputs = [
       pkg-config
-      pkgs.rustPlatform.bindgenHook
+      # Sets LIBCLANG_PATH and the base BINDGEN_EXTRA_CLANG_ARGS (clang
+      # resource dir, libc headers) so rust-bindgen works inside the Nix
+      # sandbox. Anything we add to BINDGEN_EXTRA_CLANG_ARGS below is appended
+      # to the hook's value, not replaced.
+      rustPlatform.bindgenHook
       llvmPkgs.clang
-      llvmPkgs.libclang
     ];
 
     buildInputs = [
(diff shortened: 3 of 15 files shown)


DSharifi:
Running on my local desktop + Patrick's machine:

```
❯ nix build .#mpc-node && sha256sum ./result/bin/mpc-node
warning: Git tree '/home/dsharifi/Dev/mpc' is dirty
66c45e0b6f2f36186861d0ebda4d3c753b22a62c7fcf7b94a11f97b9c1762b35  ./result/bin/mpc-node
```

Running on TDX server Alice:
```
d32ddada569d9005c45b29ae1c314e05d71eaad1632f00ed2a39541f386df2e2
```

DSharifi:
Managed to get it reproducible on Alice and produce identical hash. Fix was to remap the build directory at build time since nix on alice is user installed https://github.com/near/mpc/pull/3021/changes/39d0864698df4e0fd1f9691a674d2a167544d014

DSharifi:
@claude review

nix/mpc-node.nix
```diff
@@ -0,0 +1,259 @@
...
+  #      Some git deps (e.g. nearcore's `near-jsonrpc`) pull files from
+  #      sibling directories via `include_bytes!("../../../...")`; those get
+  #      stripped by cargo's packaging rules. `fetchgit` copies the entire
+  #      git checkout, preserving siblings.
+  #
+  # `allowBuiltinFetchGit = true` uses `builtins.fetchGit`, which is
+  # reproducible: the revision fully determines content, no `sha256` needed.
+  importedVendorDir = rustPlatform.importCargoLock {
+    lockFile = ../Cargo.lock;
+    allowBuiltinFetchGit = true;
```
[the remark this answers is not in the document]
DSharifi: We pin our git dependencies to specific tags/revisions so this is non issue.

nix/mpc-node.nix
```diff
@@ -0,0 +1,259 @@
...
+      # the build host's CPU and output bytes vary by machine.
+      #
+      # x86-64-v3 ≈ Haswell (2013) and newer: AVX2, BMI2, FMA. Deployment
+      # targets MUST be at least this CPU level; older hardware will SIGILL.
+      CFLAGS = lib.optionalString isX86 "-march=x86-64-v3";
+      CXXFLAGS = "-include cstdint" + lib.optionalString isX86 " -march=x86-64-v3";
+
+      RUSTFLAGS = lib.concatStringsSep " " (
+        lib.optionals isX86 [ "-C target-cpu=x86-64-v3" ]
+        ++ [
```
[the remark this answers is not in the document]
DSharifi: That's fine. Most modern servers are all long on v3 instruction set.

flake.nix
```diff
@@ -44,6 +47,12 @@
 
     in
     {
+      packages = forAllSystems (pkgs: {
+        mpc-node = pkgs.callPackage ./nix/mpc-node.nix {
+          inherit crane;
+        };
```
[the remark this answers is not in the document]
DSharifi: it's not default. We have other binaries such as the contract and launcher. We can reconsider this in the future.

nix/mpc-node.nix
```diff
@@ -0,0 +1,259 @@
...
+  cargoVendorDir = pkgs.runCommand "vendor-cargo-deps-crane" { } ''
+    mkdir -p $out/vendor $out/chain
+
+    cp -aL ${importedVendorDir}/near-jsonrpc-2.11.1 $out/chain/jsonrpc
+    chmod -R u+w $out/chain/jsonrpc
+    ln -s ../chain/jsonrpc $out/vendor/near-jsonrpc-2.11.1
+
+    for entry in ${importedVendorDir}/*; do
+      name=$(basename "$entry")
+      if [ "$name" != "near-jsonrpc-2.11.1" ]; then
```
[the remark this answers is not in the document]
DSharifi: Good point. I was vary of this because it adds a lot of logic, but strictly needed to avoid painful upgrades. Ideally we get nearcore folks to not include hardcoded dependencies in their source graph for crates outside of  it.

CC: @stedfn

pbeza (dismissed):
I got different SHA sums on Linux and Mac — I assume that’s expected.

Mac:
```
$ git rev-parse HEAD
d223511bad6d1c5fe1a655251cb50aa2faad5c37
$ nix build .#mpc-node && sha256sum ./result/bin/mpc-node
a2f59bf435e08e655cd0276dac448e0ffaa1c6ed1c040228cf721c5138e3da7e  ./result/bin/mpc-node
```

Linux:
```
$ git rev-parse HEAD
d223511bad6d1c5fe1a655251cb50aa2faad5c37
$ nix build .#mpc-node && sha256sum ./result/bin/mpc-node
66c45e0b6f2f36186861d0ebda4d3c753b22a62c7fcf7b94a11f97b9c1762b35  ./result/bin/mpc-node
```

DSharifi:
> I got different SHA sums on Linux and Mac — I assume that’s expected.
> 
> Mac:
> 
> ```
> $ git rev-parse HEAD
> d223511bad6d1c5fe1a655251cb50aa2faad5c37
> $ nix build .#mpc-node && sha256sum ./result/bin/mpc-node
> a2f59bf435e08e655cd0276dac448e0ffaa1c6ed1c040228cf721c5138e3da7e  ./result/bin/mpc-node
> ```
> 
> Linux:
> 
> ```
> $ git rev-parse HEAD
> d223511bad6d1c5fe1a655251cb50aa2faad5c37
> $ nix build .#mpc-node && sha256sum ./result/bin/mpc-node
> 66c45e0b6f2f36186861d0ebda4d3c753b22a62c7fcf7b94a11f97b9c1762b35  ./result/bin/mpc-node
> ```

Yes that is expected. The target build is for your local os and architecture. We can extend it in a follow up to allow for cros-compiling to linux x86

netrome:
On my Debian server I get the following on the latest commit (https://github.com/near/mpc/pull/3021/commits/09867208173d2c6be57ce4934cad89a335f66498):

```
nix build .#mpc-node && sha256sum ./result/bin/mpc-node
09f1d153fd73bb83d03759f79485ae2d3680a9ed94a9dbca163668f299481640  ./result/bin/mpc-node
```

@DSharifi what do you get here?

I'll try on the same commit as you and Patrick posted before as well.

DSharifi:
> On my Debian server I get the following on the latest commit ([0986720](https://github.com/near/mpc/commit/09867208173d2c6be57ce4934cad89a335f66498)):
> 
> ```
> nix build .#mpc-node && sha256sum ./result/bin/mpc-node
> 09f1d153fd73bb83d03759f79485ae2d3680a9ed94a9dbca163668f299481640  ./result/bin/mpc-node
> ```
> 
> @DSharifi what do you get here?
> 
> I'll try on the same commit as you and Patrick posted before as well.

Yes:
```log
❯ nix build .#mpc-node && sha256sum ./result/bin/mpc-node
09f1d153fd73bb83d03759f79485ae2d3680a9ed94a9dbca163668f299481640  ./result/bin/mpc-node
```

netrome:
Reproduced on commit `d223511bad6d1c5fe1a655251cb50aa2faad5c37`:

```
nix build .#mpc-node && sha256sum ./result/bin/mpc-node
66c45e0b6f2f36186861d0ebda4d3c753b22a62c7fcf7b94a11f97b9c1762b35  ./result/bin/mpc-node
```

DSharifi:
Alice is also good on last commit

```log
daniel@alice ~/mpc (dsharifi/add-repro-build-for-mpc-node)> nix build .#mpc-node && sha256sum ./result/bin/mpc-node
09f1d153fd73bb83d03759f79485ae2d3680a9ed94a9dbca163668f299481640  ./result/bin/mpc-node
```

.gitignore
```diff
@@ -22,6 +22,8 @@ venv
 
 # Nix
 .direnv/
+result
```
netrome, line 25: Why does the nix builds output to this directory? (Though I see it's a symlink to the nix store) - could we configure this to be set somewhere in the `target` directory?

netrome (dismissed):
The comments are very helpful. I'm not very familiar with Nix so not following everything that's going on here, but this looks reasonable from what I can see - and the build seems reproducible.

Though we previously concluded Nix wasn't reproducible as per https://linderud.dev/blog/nixos-is-not-reproducible/ - so I'm curious to learn more why this works. I guess while Nix doesn't guarantee reproducibility, it provides a very strong foundation and with the right control on top of it (setting the source date epoch etc) our builds become reproducible.

As next steps, if this works reliably, we should consider replacing our existing docker based build scripts with this one. But that's a larger effort, and it's probably good to see if this script stands the test of time for a bit before that.

netrome:
As I understand it, this build is also sandboxed so no dependency would be able to make an HTTP request to a random URL during build time?

.gitignore
DSharifi: It's the default directory for nix build commant. You can use -o flag to control this.

DSharifi:
> As I understand it, this build is also sandboxed so no dependency would be able to make an HTTP request to a random URL during build time?

Exactly

.gitignore
```diff
@@ -22,6 +22,8 @@ venv
...
+result-*
```
gilcu3, line 26: nit: could be reduced to 
```suggestion
result*
```

nix/mpc-node.nix
```diff
@@ -0,0 +1,272 @@
...
+    cp -aL "$near_jsonrpc_entry" $out/chain/jsonrpc
+    chmod -R u+w $out/chain/jsonrpc
+    ln -s ../chain/jsonrpc "$out/vendor/$near_jsonrpc_name"
+
+    for entry in ${importedVendorDir}/*; do
+      name=$(basename "$entry")
+      if [ "$name" != "$near_jsonrpc_name" ]; then
+        ln -s "$entry" "$out/vendor/"
+      fi
+    done
```
gilcu3, line 139: ouch this is a nasty hack

nix/mpc-node.nix
```diff
@@ -0,0 +1,272 @@
...
+      SOURCE_DATE_EPOCH = "0";
+
+      # Link against nixpkgs' openssl instead of openssl-sys' vendored copy.
+      # The vendored copy varies with the host; the nixpkgs one is pinned
+      # through flake.lock.
+      OPENSSL_NO_VENDOR = "1";
+
+      # Prevents rocksdb's build.rs from probing /proc/cpuinfo and baking
+      # host-specific ISA choices into its object files.
+      PORTABLE = "1";
```
gilcu3, line 194: was this one actually needed? We were not doing this, and still getting repro

nix/mpc-node.nix
```diff
@@ -0,0 +1,272 @@
...
+    env = {
+      # Reproducibility knobs ------------------------------------------------
+
+      # Fixed epoch for any build-script that stamps timestamps into output.
+      SOURCE_DATE_EPOCH = "0";
+
+      # Link against nixpkgs' openssl instead of openssl-sys' vendored copy.
+      # The vendored copy varies with the host; the nixpkgs one is pinned
+      # through flake.lock.
+      OPENSSL_NO_VENDOR = "1";
```
gilcu3, line 190: Is this better than enforcing a pure nix shell for the build? Similar to:
```
nix develop --ignore-environment
```

nix/mpc-node.nix
```diff
@@ -0,0 +1,272 @@
...
+          "--remap-path-prefix=${src}=/build/source"
+          "--remap-path-prefix=${cargoVendorDir}=/cargo-vendor"
+        ]
+      );
+
+      # Extra bindgen flags — paths are already provided by bindgenHook.
+      BINDGEN_EXTRA_CLANG_ARGS = lib.concatStringsSep " " (
+        lib.optionals isX86 [ "-march=x86-64-v3" ]
+        ++ [ "-fno-stack-protector" ]
+      );
```
gilcu3: why are we disabling the stack protector?

Also, why using `x86-64-v3`? I think we should only restrict the supported CPUs if we have optimization requirements that would be met with this. I am aware that most server CPUs support v3 already, but you never know what we would find

nix/mpc-node.nix
```diff
@@ -0,0 +1,272 @@
...
+    preBuild = ''
+      export RUSTFLAGS="$RUSTFLAGS --remap-path-prefix=$NIX_BUILD_TOP/source=/build/source --remap-path-prefix=$NIX_BUILD_TOP=/build"
+    '';
+
+    doCheck = false;
+  };
+
+  # Build deps in a separate derivation so that they're cached across
+  # mpc-node source changes.
+  cargoArtifacts = craneLib.buildDepsOnly commonArgs;
```
gilcu3, line 253: are we sure this does not break repro? In general it is advised to disable cache for such builds

gilcu3 (changes requested):
Looks good overall.

Requested a small change to remove the x86-v3 requirement. It does not seem necessary here, and is not in line with our previous builds. We can always add it later (in all our builds) if we obtain noticeable performance improvements with it

pbeza (dismissed):

nix/mpc-node.nix
DSharifi: I needed it a while ago when building neard. Keeping for now and we can experiment removing it later.

nix/mpc-node.nix
DSharifi: Yes, this forces crates to use the system (nix provided) openssl installation

nix/mpc-node.nix
DSharifi: We are only caching by the inputs. If toolchain, deps, version, a source file, etc. changes, then the cache key changes and Nix rebuilds.

Rebuilding multiple times on same machine should give same result.

nix/mpc-node.nix
DSharifi: removed v3, and compiling without the stack protector works fine.

netrome (approved):

pbeza (approved):

gilcu3 (approved):
Thank you!

Issue #2923, opened 2026-04-16
docs: add local PCCS setup + trust-policy section to operator deployment guide


barakeinav1:
### Background

Originally filed alongside the (since-deleted) `pccs-proxy` crate. The proxy was superseded by #2944 (direct `dcap_qvl` calls) and #3026 (operator-controlled TLS-trust knobs), so this issue's premise has shifted but its goal is the same: the operator deployment guide should explain how to point a node at a non-Phala PCCS.

The current operator guide (`docs/running-an-mpc-node-in-tdx-external-guide.md`) does not mention the `pccs_url` config field or the `pccs_ca_cert_pem` / `pccs_tls_insecure` trust knobs. Operators following the guide default to Phala's hosted PCCS without knowing they have alternatives, and operators who do try to point at a local PCCS hit a surprising trap: the cert that ships in the Ubuntu `sgx-dcap-pccs` package is X.509 v1 / expired / no SAN, so `pccs_ca_cert_pem` set to that cert still fails TLS validation.

### User Story

As an MPC node operator, I want the deployment guide to explain my PCCS options and the trust-policy knobs, so I can pick the right one for my deployment (centralized via Phala, centralized via Intel, or self-hosted local PCCS) and avoid the cert-format gotcha.

### Acceptance Criteria

- [ ] In `docs/running-an-mpc-node-in-tdx-external-guide.md`, add a "PCCS configuration" section covering:
  - The three real choices for `pccs_url`:
    - `https://pccs.phala.network` — default; Let's Encrypt cert; works out of the box
    - `https://api.trustedservices.intel.com` — Intel's public PCS direct; works out of the box; no centralized cache between you and Intel
    - `https://<your-host>:8081` — operator-run local PCCS; requires trust-policy config (next bullet)
  - The two trust-policy knobs from #3026 and when to use each:
    - `pccs_ca_cert_pem` (production-recommended): pin the operator's PCCS server cert as a trust anchor. Requires a properly-formed cert (v3, valid dates, SAN matching the host).
    - `pccs_tls_insecure` (dev-only): disable TLS verification entirely. Refused at startup for non-loopback hosts.
    - Note that the two are mutually exclusive.
[...]

PR #3026, opened 2026-04-26
feat(tee)!: allow PCCS with self-signed TLS certs via operator config

### LINKED ISSUE #2963: Make MPC nodes allow PCCS with self-signed certificates
### Background

We don't want to require operators to set up TLS with externally signed certificates for their self-hosted PCCS, but currently the MPC node rejects endpoints with self-signed certificates. We should change this.

Requires upstream change, already in progress in https://github.com/Phala-Network/dcap-qvl/pull/146

### User Story

As a node operator I don't want to have to manage certificates for my self-hosted PCCS.

### Acceptance Criteria

MPC nodes tolerate PCCS endpoints with self-signed certificates.

_No response_

### LINKED ISSUE #3027: Drop reqwest 0.12 pin in tee-authority once dcap-qvl ships reqwest 0.13
### Background

Code change introduced in #3026: `crates/tee-authority` depends on `reqwest = "0.12.24"` directly, while the rest of the workspace uses `reqwest = "0.13.2"`. The split is forced because `tee-authority` constructs a `reqwest::Client` and passes it into `dcap_qvl::collateral::CollateralClient::new(...)`, and `dcap-qvl 0.4.x` depends on `reqwest 0.12`. Cargo treats `reqwest 0.12.x` and `reqwest 0.13.x` as different crates, so the types must match.

The workaround works (#3026 verified end-to-end) but has costs:
- Both reqwest majors compiled into the final binary (~600KB).
- Manual-maintenance comment in `crates/tee-authority/Cargo.toml`: must track whatever version `dcap-qvl` uses.
- Confusing for contributors who add features in `tee-authority` and hit cryptic type errors against newer reqwest API.

### Upstream resolution

Initial plan was to upstream `pub use reqwest;` so we could track `dcap-qvl`'s reqwest version automatically (see closed Phala-Network/dcap-qvl#155). Upstream took a better path instead:

- **Phala-Network/dcap-qvl#156** — abstracted HTTP transport behind an `HttpClient` trait. `CollateralClient` is now generic over `H: HttpClient` (defaulting to `reqwest::Client`, so existing call sites are unchanged). Reqwest is no longer part of the public API.
- **Phala-Network/dcap-qvl#157** — bumped `dcap-qvl`'s internal reqwest from 0.12 → 0.13. Possible because of #156.

Both are merged on `Phala-Network/dcap-qvl@master`, awaiting a crates.io release.

This makes our cleanup strictly simpler than originally planned: no source-code changes are needed, just a workspace dep bump and a deletion in `tee-authority/Cargo.toml`. The `CollateralClient::new(reqwest::Client, url)` call site keeps working unchanged thanks to the back-compat default `H = reqwest::Client`.

### User Story

As a workspace maintainer, I want a single `reqwest` version compiled into the node binary, so we don't ship duplicate copies of an HTTP stack and don't carry hidden coupling between our `Cargo.toml` and Phala's release schedule.

### Trigger

Do this work once `Phala-Network/dcap-qvl` cuts a crates.io release containing both #156 and #157 (i.e. reqwest 0.13 and the `HttpClient` trait). As of this issue: merged on `master`, not yet released.

### Acceptance Criteria

- [ ] Bump the `dcap-qvl` pin in the workspace `Cargo.toml` to the new release.
- [ ] In `crates/tee-authority/Cargo.toml`, remove the explicit `reqwest = { version = "0.12.24", … }` line and the explanatory comment. Add `reqwest = { workspace = true }` (or rely on the workspace dep wherever it's already declared).
- [ ] No source changes expected: `reqwest::Client::builder()...` and `reqwest::Certificate::from_pem(...)` continue to work against reqwest 0.13.
- [ ] Verify `cargo tree -i reqwest` shows a single version in the output.
- [ ] Verify `cargo test -p tee-authority --lib` and `cargo clippy -p tee-authority --all-targets --locked -- -D warnings` still pass.
- [ ] Optional follow-up: migrate `tee-authority` to the `HttpClient` trait directly (decoupling us from `reqwest` entirely). Not required for the version-pin cleanup.

- Code that introduced the split: #3026 (search for `reqwest = { version = "0.12.24"` in `crates/tee-authority/Cargo.toml`).
- Closed alternative: Phala-Network/dcap-qvl#155 (re-export approach, superseded by #156 + #157).

barakeinav1:
Closes #2963.

Builds on Phala-Network/dcap-qvl#146 (merged in `dcap-qvl 0.4.0`) and integrates with #3010 (multi-PCCS) and #3048 (typed PCCS errors), both already on `main`.

## Summary

Lets operators point the node at a self-signed local PCCS — and combine it with public-CA fallbacks in the same multi-endpoint chain — by adding an optional per-URL TLS trust override to each `[[mpc_node_config.pccs_endpoints]]` entry.

The configuration:

```toml
# Default — single public endpoint, default trust roots
[[mpc_node_config.pccs_endpoints]]
url = "https://pccs.phala.network/"

# Per-URL TLS override only where needed
[[mpc_node_config.pccs_endpoints]]
url = "https://localhost:8081/"
tls = { mode = "insecure" }

[[mpc_node_config.pccs_endpoints]]
url = "https://pccs.phala.network/"

[[mpc_node_config.pccs_endpoints]]
url = "https://api.trustedservices.intel.com/"
```

Two override modes:

- **`mode = "ca_cert_pem"`** — inline PEM-encoded certificate added as an additional trust anchor for that endpoint. Default trust roots stay active; pinning is additive. Production-recommended when a properly-formed cert is available.
- **`mode = "insecure"`** — disable TLS certificate verification for that endpoint. **Loopback-only**: startup validation rejects it for non-loopback URLs (allowed hosts: `localhost`, `127.0.0.0/8`, `[::1]`, and the QEMU slirp gateway `10.0.2.2`). Acceptable on loopback because the host is the trust boundary; an attacker capable of swapping the cert already has host-level access.

Both modes default to off (no `tls` line on an entry → default reqwest+rustls trust roots, the bundled Mozilla `webpki-roots`). Existing deployments using only public endpoints are unchanged.

## Why per-endpoint and not a single global setting

Original drafts of this PR had `pccs_ca_cert_pem` / `pccs_tls_insecure` as two global `StartConfig` fields. That worked when there was a single PCCS URL, but #3010 (now merged) made `pccs_urls` a list — a global trust setting can't express, for example, "self-signed insecure local primary, public-CA Phala fallback, public-CA Intel last-resort." Per-URL trust resolves that cleanly: each entry says exactly what trust to apply, and the loopback gate becomes per-URL too.

## Implementation

- **`crates/launcher-interface/src/types.rs`**: new `PccsEndpointConfig { url, tls: Option<PccsTlsTrust> }` and `PccsTlsTrust` (tagged enum: `CaCertPem { ca_cert_pem }` | `Insecure`). Lives next to `TeeConfig` / `TeeAuthorityConfig` because it's a shared config type.
- **`crates/node-config/src/start.rs`**: `pccs_urls: NonEmptyVec<Url>` becomes `pccs_endpoints: NonEmptyVec<PccsEndpointConfig>`. `default_pccs_endpoints()` returns a one-element list with the Phala URL and no TLS override.
- **`crates/tee-authority/src/tee_authority.rs`**:
  - `DstackTeeAuthorityConfig.pccs_endpoints: NonEmptyVec<PccsEndpointConfig>`.
  - `validate_pccs_endpoints` walks the list once at startup: for each `Insecure` entry, the URL must be loopback; for each `CaCertPem`, the PEM must round-trip through `reqwest::Certificate::from_pem`. Per-`Insecure` WARN log.
  - `fetch_collateral_from(endpoint, quote)` honours `endpoint.tls`: `None` uses dcap-qvl's `with_default_http`; a set `tls` builds a custom `reqwest::Client` (`add_root_certificate` for PEM mode, `danger_accept_invalid_certs` + `danger_accept_invalid_hostnames` for insecure) and hands it to `CollateralClient::new`.
  - `try_each_pccs_endpoint` iterates `&NonEmptyVec<PccsEndpointConfig>`. Per-iteration logs include the endpoint URL.
- **`crates/node`** plumbing: `into_tee_authority` takes `NonEmptyVec<PccsEndpointConfig>`, `run.rs` and `cli.rs` pass it through.
- **`crates/tee-authority/Cargo.toml`**: explicit `reqwest = "0.12.24"` direct dep (matching dcap-qvl's reqwest major). Workspace pin is on `0.13` for everything else; this is the cross-major bridge required because we hand a `reqwest::Client` directly into `dcap_qvl::collateral::CollateralClient::new`. Tracked for cleanup in #3027 once `dcap-qvl` exposes the `HttpClient` trait upstream.
- **`deployment/cvm-deployment/user-config.toml`**: documented examples for default trust, multi-endpoint fallback, insecure local primary + public fallbacks, and cert-pinning. Stock `sgx-dcap-pccs` cert-regen heads-up included.

## Guardrails

- Per-endpoint validation at config conversion time (`validate_pccs_endpoints`):
  - `Insecure` requires loopback URL — rejected at startup with a clear error otherwise.
  - `CaCertPem` requires the PEM to parse — typo'd cert fails at startup, not at first attestation.
- One `WARN` log per `Insecure` endpoint at startup so operators see the security signal up front.
- Mutual-exclusivity of cert-pinning vs insecure is type-level: `PccsTlsTrust` is a tagged enum, you can pick at most one variant per endpoint.

## Operator note: stock PCCS cert

For `mode = "ca_cert_pem"` against an Ubuntu `sgx-dcap-pccs` install: the cert that ships in that package (`/opt/intel/sgx-dcap-pccs/ssl_key/file.crt`) is X.509 v1, expired, has no SAN. Pinning it directly will fail TLS validation regardless of the trust override — rustls still checks expiry / hostname / version. Operators must regenerate as v3 with valid dates and a SAN matching the endpoint's URL host first (one `openssl req -x509 ... -addext "subjectAltName=..."` command, full text in the deployment user-config.toml). Only `mode = "insecure"` bypasses these checks. Tracked for the operator deployment guide in #2923.

## Tests

19 unit tests in `tee-authority` + 3 in `node-config`:

**`validate_pccs_endpoints` — per-URL loopback gate (the new behavior):**
- `__should_accept_insecure_on_loopback` (4 cases: `localhost`, `127.0.0.1`, `[::1]`, `10.0.2.2`).
- `__should_reject_insecure_on_non_loopback` (3 cases: Phala, Intel, `0.0.0.0`).
- `__should_accept_mixed_loopback_insecure_with_public_fallbacks` — exercises the new per-URL gate. (A pre-this-PR global gate would have rejected this.)
- `__should_reject_insecure_non_loopback_even_in_mixed_list` — symmetric: a non-loopback insecure entry rejects regardless of what else is in the list.

**`pccs_endpoints` deserialization (`mpc-node-config`):**
- `__should_parse_single_bare_url_entry`
- `__should_preserve_order_with_mixed_tls_overrides`
- `__should_default_to_phala_when_omitted`

**Plus all preexisting `try_each_pccs_endpoint__*` and `all_pccs_endpoints_failed__*` tests** updated for the new endpoint-shape signature.

`cargo clippy --all-targets --locked -- -D warnings` and `cargo fmt --check` clean across `tee-authority` / `mpc-node` / `mpc-node-config` / `launcher-interface`.

## Test plan

- [x] Unit tests + clippy + fmt across affected crates.
- [ ] E2E re-verification against the local PCCS on the dev server with the **new per-endpoint TOML shape**. (Earlier E2E was done against the previous two-field design; the underlying fetch logic is structurally the same but should be re-verified end-to-end before merge.)

## Related

- Multi-PCCS fallback: #3010 (merged) — the per-URL TLS override only makes sense given the URL list this added.
- Typed PCCS errors: #3048 (merged) — `validate_pccs_endpoints` produces `anyhow::Error` at startup but per-URL fetch failures still flow through `PccsEndpointError` / `AllPccsEndpointsFailed`.
- Upstream: Phala-Network/dcap-qvl#146 (merged, in `dcap-qvl 0.4.0`).
- Reqwest pin cleanup: #3027 (gated on `dcap-qvl` exposing the `HttpClient` trait upstream).
- Research: near/mpc-private#292 (closed).
- Operator-guide docs follow-up: #2923.


crates/tee-authority/src/tee_authority.rs
```diff
@@ -164,6 +207,31 @@ impl TeeAuthority {
...
+) -> anyhow::Result<reqwest::Client> {
+    use anyhow::Context;
+    let mut builder = reqwest::Client::builder().timeout(PCCS_REQUEST_TIMEOUT);
+
+    if pccs_tls_insecure {
+        tracing::warn!(
+            "pccs_tls_insecure=true: PCCS TLS certificate verification is DISABLED. \
+             Only safe for a loopback PCCS in development."
+        );
+        builder = builder.danger_accept_invalid_certs(true);
```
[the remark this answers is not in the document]
barakeinav1: Addressed in 6384a4a — added `danger_accept_invalid_hostnames(true)` alongside `danger_accept_invalid_certs(true)`. rustls's NoVerifier covers both today, but the explicit second flag documents intent and is stable across reqwest backends.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -164,6 +207,31 @@ impl TeeAuthority {
     }
 }
 
+/// Build the `reqwest::Client` used to talk to the PCCS server, applying the
+/// operator's TLS-trust knobs. By default returns the standard `reqwest`
+/// client with system CAs only — same trust posture as `with_default_http`.
```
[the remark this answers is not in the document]
barakeinav1: Addressed in 6384a4a (comment fix only). Kept the `rustls-tls` feature to match dcap-qvl's own trust posture; updated the comment to say "the trust anchors built into the selected reqwest TLS feature (currently rustls-tls, which uses the bundled Mozilla webpki-roots)" — accurate description without changing the resolved roots.

crates/tee-authority/Cargo.toml
```diff
@@ -16,6 +16,10 @@ dstack-sdk = { workspace = true }
 hex = { workspace = true }
 launcher-interface = { workspace = true }
 mpc-attestation = { workspace = true }
+# Pinned to match the version used by `dcap-qvl` so we can pass a
+# `reqwest::Client` directly into `CollateralClient::new`. The workspace
+# pin is on a newer major (0.13) used by other crates.
+reqwest = { version = "0.12.24", default-features = false, features = ["rustls-tls"] }
```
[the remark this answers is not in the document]
barakeinav1: Skipped this specific feature change. Switching to `rustls-tls-native-roots` would diverge from dcap-qvl's own trust posture (they use `rustls-tls` with webpki-roots). Took the comment fix from your other suggestion (3143365053) instead so the description matches reality without changing the roots.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -164,6 +223,31 @@ impl TeeAuthority {
...
+    use anyhow::Context;
+    let mut builder = reqwest::Client::builder().timeout(PCCS_REQUEST_TIMEOUT);
+
+    if pccs_tls_insecure {
+        tracing::warn!(
+            "pccs_tls_insecure=true: PCCS TLS certificate verification is DISABLED. \
+             Only safe for a loopback PCCS in development."
+        );
+        builder = builder.danger_accept_invalid_certs(true);
+    } else if let Some(pem) = pccs_ca_cert_pem {
```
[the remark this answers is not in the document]
barakeinav1: Addressed in 6384a4a — moved the WARN from `build_pccs_http_client` to `validate_pccs_tls_config`, which runs once during config conversion. Now emits at startup, never on per-fetch.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -71,10 +78,47 @@ impl Default for DstackTeeAuthorityConfig {
...
+    pccs_tls_insecure: bool,
+) -> anyhow::Result<()> {
+    anyhow::ensure!(
+        !(pccs_tls_insecure && pccs_ca_cert_pem.is_some()),
+        "pccs_tls_insecure=true and pccs_ca_cert_pem are mutually exclusive: \
+         pin a specific cert or disable verification entirely, not both",
+    );
+    if !pccs_tls_insecure {
+        return Ok(());
+    }
```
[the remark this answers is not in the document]
barakeinav1: Addressed in 6384a4a — `validate_pccs_tls_config` now runs `reqwest::Certificate::from_pem(pem)?` so a malformed PEM fails at startup with a clear message instead of at the first attestation attempt.

crates/node-config/src/start.rs
```diff
@@ -28,6 +28,21 @@ pub struct StartConfig {
...
+    #[serde(default)]
+    pub pccs_ca_cert_pem: Option<String>,
+
+    /// Disable TLS certificate verification for the PCCS server. **Dev-only**:
+    /// only honored when `pccs_url` host is `localhost` / `127.0.0.1` /
+    /// `10.0.2.2`. The node startup rejects this flag combined with any other
+    /// host so a stray copy-paste cannot silently disable TLS validation
+    /// against a real network endpoint.
+    #[serde(default)]
+    pub pccs_tls_insecure: bool,
```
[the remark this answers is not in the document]
barakeinav1: Addressed in 6384a4a — both field rustdocs now end with a "Mutually exclusive with [other-field] — startup rejects setting both." line.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -71,10 +80,47 @@ impl Default for DstackTeeAuthorityConfig {
...
+    let host = pccs_url.host_str().unwrap_or("");
+    anyhow::ensure!(
+        LOOPBACK_PCCS_HOSTS.contains(&host),
+        "pccs_tls_insecure=true is only allowed for loopback PCCS URLs (one of {:?}); got host {host:?}",
```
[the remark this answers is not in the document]
barakeinav1: Addressed in 6384a4a — replaced the string-match list with `pccs_url.host()` + `IpAddr::is_loopback()` (plus the `10.0.2.2` slirp gateway as a special case). Catches all of `127.0.0.0/8` and avoids the bracket-handling fragility for IPv6.

crates/node/src/config/start.rs
```diff
@@ -3,20 +3,35 @@ use launcher_interface::types::{TeeAuthorityConfig, TeeConfig};
...
-    fn into_tee_authority(self, pccs_url: url::Url) -> anyhow::Result<TeeAuthority> {
+    fn into_tee_authority(
+        self,
+        pccs_url: url::Url,
+        pccs_ca_cert_pem: Option<String>,
+        pccs_tls_insecure: bool,
+    ) -> anyhow::Result<TeeAuthority> {
+        validate_pccs_tls_config(&pccs_url, pccs_ca_cert_pem.as_deref(), pccs_tls_insecure)?;
         Ok(match self.authority {
             TeeAuthorityConfig::Local => LocalTeeAuthorityConfig::default().into(),
```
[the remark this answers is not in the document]
barakeinav1: Skipped. The current behavior (validate unconditionally) fails fast on operator config mistakes — if someone sets `pccs_tls_insecure=true` with a non-loopback URL while accidentally on Local mode, the more useful signal is "your insecure flag is wrong" surfaced at startup, vs silently accepting it and surprising them later if/when they switch to Dstack. Both are defensible, leaning toward the strict behavior.

crates/node-config/src/start.rs
```diff
@@ -199,4 +220,65 @@ mod tests {
...
+        assert_eq!(
+            probe_default,
+            Probe {
+                pccs_ca_cert_pem: None,
+                pccs_tls_insecure: false,
+            }
+        );
+
+        // Lock in that StartConfig still uses these exact names — a rename
+        // would silently break operator configs.
```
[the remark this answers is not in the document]
barakeinav1: Addressed in 6384a4a — reworded the test docstring to be honest about what's covered: `offset_of!` catches plain field renames at compile time; `#[serde(rename = ...)]` would slip past, and the fixture cost to catch that (constructing a full `StartConfig`) outweighs the marginal value. Documented as a known scope limit in the test header.

crates/node-config/src/start.rs
```diff
@@ -28,6 +28,27 @@ pub struct StartConfig {
...
+    /// Set this when pointing at a local PCCS that uses a self-signed cert.
+    /// The cert is added as an additional trust anchor; system CAs still
+    /// work for other endpoints.
+    #[serde(default)]
+    pub pccs_ca_cert_pem: Option<String>,
+
+    /// Disable TLS certificate verification for the PCCS server. **Loopback
+    /// only**: startup rejects this flag for `pccs_url` hosts other than
+    /// `localhost` / `127.0.0.1` / `10.0.2.2`, so it cannot silently disable
+    /// TLS validation against a real network endpoint. Acceptable for
```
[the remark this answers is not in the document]
barakeinav1: Addressed in 6384a4a — the `pccs_tls_insecure` rustdoc now lists `localhost`, `127.0.0.0/8`, IPv6 loopback `::1` (with the `https://[::1]:<port>/` URL form spelled out), and `10.0.2.2`. Matches what the runtime validator now accepts.

crates/node-config/src/start.rs
```diff
@@ -199,4 +220,65 @@ mod tests {
...
+
+        // Lock in that StartConfig still uses these exact names — a rename
+        // would silently break operator configs.
+        let serialized = toml::to_string(&Probe {
+            pccs_ca_cert_pem: Some("x".into()),
+            pccs_tls_insecure: true,
+        })
+        .unwrap();
+        assert!(serialized.contains("pccs_ca_cert_pem"));
+        assert!(serialized.contains("pccs_tls_insecure"));
```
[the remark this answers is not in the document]
barakeinav1: Skipped this specific code suggestion — the proposed `toml::from_str(toml_input)` would fail because `StartConfig` has many non-default fields (home_dir, secrets, tee, node, log, …) that aren't in the minimal `toml_input`. Took the docstring-rewording approach (your other comment 3143497502) instead, documenting that the test pins probe-struct field names + uses `mem::offset_of!` to catch StartConfig field renames at compile time. The full StartConfig round-trip would need a sizeable fixture rewrite for marginal value over the offset_of! check.

deployment/cvm-deployment/user-config.toml
```diff
@@ -14,6 +14,41 @@ home_dir = "/data"
...
+# ...
+# -----END CERTIFICATE-----
+# """
+#
+#   pccs_tls_insecure: disable TLS certificate verification entirely.
+#   Loopback only; the node startup rejects this for non-loopback PCCS
+#   hosts (allowed: localhost, 127.0.0.1, ::1, 10.0.2.2). Acceptable for
+#   production loopback-PCCS deployments — the host is the trust
+#   boundary, so TLS authentication is redundant — but `pccs_ca_cert_pem`
+#   is preferred when a properly-formed cert is available.
```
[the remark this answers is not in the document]
barakeinav1: Addressed in 6384a4a — the user-config.toml comment now explicitly spells out the four allowed-host categories (localhost, `127.0.0.0/8`, `https://[::1]:<port>/` for IPv6, `10.0.2.2` slirp gateway) so operators don't have to reverse-engineer the bracket convention.

barakeinav1:
@claude review

crates/tee-authority/src/tee_authority.rs
```diff
@@ -235,6 +379,157 @@ mod tests {
...
+        // build_pccs_http_client__should_reject_invalid_pem).
+        let url: Url = url.parse().expect("valid URL");
+        let pem = "-----BEGIN CERTIFICATE-----\n...\n-----END CERTIFICATE-----";
+
+        // When
+        let result = validate_pccs_tls_config(&url, Some(pem), false);
+
+        // Then
+        assert!(result.is_ok(), "expected ok for {url}, got {result:?}");
+    }
```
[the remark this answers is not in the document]
barakeinav1: Already addressed in 695e0f62 — the affected tests (`validate_pccs_tls_config__should_accept_pem_only_for_any_host` and `…__should_reject_pem_and_insecure_combined`) now use `test_cert_pem()` (rcgen-generated), and a new `validate_pccs_tls_config__should_reject_invalid_pem` covers the malformed case.

crates/tee-authority/Cargo.toml
```diff
@@ -16,13 +16,21 @@ dstack-sdk = { workspace = true }
 hex = { workspace = true }
 launcher-interface = { workspace = true }
 mpc-attestation = { workspace = true }
+# Pinned to match the version used by `dcap-qvl` so we can pass a
+# `reqwest::Client` directly into `CollateralClient::new`. The workspace
+# pin is on a newer major (0.13) used by other crates.
+# TODO(#3027): drop this explicit pin once `dcap-qvl` re-exports
+# `reqwest` upstream — then we can use `dcap_qvl::reqwest` and the
+# workspace converges on a single reqwest version.
```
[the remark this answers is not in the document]
barakeinav1: Good catch — `version = "0.12.24"` is a caret requirement (Cargo.lock resolves to 0.12.28), not a strict pin. Caret is the right semantic here (we just need to track whatever 0.12.x `dcap-qvl` resolves to), so reworded the comment to "Constrained to the 0.12 series" in 5ab311ba rather than tightening the version.

crates/node-config/src/start.rs
```diff
@@ -28,6 +28,36 @@ pub struct StartConfig {
...
+    /// attacker capable of swapping the cert already has host-level access).
+    /// Use [`pccs_ca_cert_pem`](Self::pccs_ca_cert_pem) instead when a
+    /// properly-formed cert is available — same security posture on
+    /// loopback, plus a positive operational check that the cert matches
+    /// what the operator expects.
+    ///
+    /// Mutually exclusive with `pccs_ca_cert_pem` — startup rejects setting
+    /// both.
+    #[serde(default)]
+    pub pccs_tls_insecure: bool,
```
pbeza: I think `pccs_ca_cert_pem` alone covers the legitimate cases, so we don’t need this boolean.

There are three meaningful TLS states:

1. Default trust (bundled cert roots) — for public Phala/Intel endpoints, which is what every existing operator uses.
2. Default + pinned self-signed cert — for the new local-PCCS case.
3. No trust at all — `pccs_tls_insecure`, which I don’t think we need.

IMHO, having both `pccs_tls_insecure` and `pccs_ca_cert_pem` needlessly complicates the logic.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -71,10 +82,87 @@ impl Default for DstackTeeAuthorityConfig {
...
+/// - IPv6 loopback `::1`
+/// - the QEMU slirp gateway `10.0.2.2`
+fn pccs_host_is_loopback(pccs_url: &Url) -> bool {
+    match pccs_url.host() {
+        Some(url::Host::Domain(d)) => LOOPBACK_PCCS_DOMAINS.contains(&d),
+        Some(url::Host::Ipv4(ip)) => ip.is_loopback() || ip.octets() == [10, 0, 2, 2],
+        Some(url::Host::Ipv6(ip)) => IpAddr::V6(ip).is_loopback(),
+        None => false,
+    }
+}
```
pbeza: Checking against a specific set of IPs is very brittle. You can easily remap those IPs using `iptables`, so this ends up being security through obscurity. I think the operator should instead be able to specify the set of certificates they trust, regardless of which IP the connection uses.

pbeza (changes requested):

crates/node-config/src/start.rs
barakeinav1: the down side of only supporting a pinned self-signed cert, is that is complicated the operator setup, 
they will need to create a self sign cert and provision there PCCS with it.

I'm ok with supporting only this option, but let's get other people feedback on this.

crates/tee-authority/src/tee_authority.rs
barakeinav1: this is for pccs_tls_insecure unsecure only. 
probably need local admin privilege to do this, and in this case, you can also provision any self sign cert as well to the PCCS and into the MPC node.

pbeza: I mean, with the current implementation, an operator could bypass this check pretty easily by using `iptables`/`/etc/hosts` to remap the IPs we check to whatever IP they want. That effectively allows connecting to any insecure PCCS server, making this check ineffective:

https://github.com/near/mpc/blob/5ab311ba5a1c606e9a24df2362b23af2dc9ba0ad/crates/tee-authority/src/tee_authority.rs#L150-L155

crates/node-config/src/start.rs
pbeza: [Discussed offline](https://nearone.slack.com/archives/C0912BTG51T/p1777285674128319?thread_ts=1777213338.170469&cid=C0912BTG51T).

pbeza: I think we can make that mutual exclusivity more explicit in Rust. Something like this, perhaps (vibe coded):

```rust
#[derive(Serialize, Deserialize)]
#[serde(untagged, deny_unknown_fields)]
pub enum PccsTlsTrust {
    CaCertPem { pccs_ca_cert_pem: String },
    Insecure { #[serde(deserialize_with = "deserialize_true")] pccs_tls_insecure: bool },
}

// on StartConfig:
#[serde(default, flatten)]
pub pccs_tls_trust: Option<PccsTlsTrust>,
```

or

```rust
#[derive(Serialize, Deserialize)]
#[serde(tag = "mode", rename_all = "snake_case", deny_unknown_fields)]
pub enum PccsTlsTrust {
    CaCertPem { pem: String },
    Insecure,
}

// on StartConfig:
#[serde(default)]
pub pccs_tls: Option<PccsTlsTrust>,

[mpc_node_config.pccs_tls]
mode = "ca_cert_pem"
pem = "..."

# or
[mpc_node_config.pccs_tls]
mode = "insecure"
```

It’d probably make this logic less convoluted too:

https://github.com/near/mpc/blob/5ab311ba5a1c606e9a24df2362b23af2dc9ba0ad/crates/tee-authority/src/tee_authority.rs#L287-L308

barakeinav1: Done in 4a17a40e — went with your tagged-enum suggestion. Operator TOML now looks like:

```toml
[mpc_node_config.pccs_tls]
mode = "ca_cert_pem"
ca_cert_pem = "..."

# or
[mpc_node_config.pccs_tls]
mode = "insecure"

# omit the table entirely → system trust roots (default)
```

The enum (`PccsTlsTrust`) lives in `launcher-interface/types.rs`, parallel to the existing `TeeConfig` and `TeeAuthorityConfig`. Mutual exclusivity is now a type-level invariant — serde rejects ill-formed configs at parse time, and the `validate_pccs_tls_config` runtime check + the both-set tests went away with it. `build_pccs_http_client` collapses to a 3-arm exhaustive match over the enum.

Went with the tagged form (`mode = "..."`) over the untagged variant for the reasons we discussed: cleaner serde errors, no `deserialize_true` hack, and a natural extension point if a third trust mode shows up. Renamed `pem` → `ca_cert_pem` to keep the existing field-name spelling — operators familiar with the original config will recognize it.

crates/launcher-interface/src/types.rs
```diff
@@ -22,6 +22,36 @@ pub enum TeeAuthorityConfig {
...
+/// added later (just another variant).
+///
+/// `None` on `StartConfig` (i.e. no `[mpc_node_config.pccs_tls]` table)
+/// means "use the system trust roots" — the default behaviour.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "mode", rename_all = "snake_case", deny_unknown_fields)]
+pub enum PccsTlsTrust {
+    /// Trust an operator-pinned PEM certificate as an additional root
+    /// anchor. Use when the PCCS uses a self-signed cert. System trust
+    /// roots remain active for any other endpoint.
```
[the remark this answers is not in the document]
barakeinav1: Addressed in 8fa888e2 — the doc-comment now says "default reqwest+rustls trust roots — bundled Mozilla webpki-roots" instead of "system trust roots." The actual TLS backend is webpki-roots (not OS-native), and operators reading the old wording could be misled.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -62,6 +65,12 @@ pub struct DstackTeeAuthorityConfig {
     dstack_endpoint: PathBuf,
     /// Base URL of the PCCS server used to fetch TDX attestation collateral.
     pccs_url: Url,
+    /// Operator's PCCS TLS trust policy. `None` means "use the system trust
+    /// roots" — the default for public PCCS endpoints. The two non-default
+    /// modes (pinned-cert vs. insecure) are encoded as variants of
```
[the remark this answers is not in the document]
barakeinav1: Addressed in 8fa888e2 — the doc-comment now says "default reqwest+rustls trust roots — bundled Mozilla webpki-roots" instead of "system trust roots." The actual TLS backend is webpki-roots (not OS-native), and operators reading the old wording could be misled.

crates/node-config/src/start.rs
```diff
@@ -28,6 +28,28 @@ pub struct StartConfig {
     /// Defaults to Phala's PCCS if not set in config.
     #[serde(default = "default_pccs_url")]
     pub pccs_url: url::Url,
+
+    /// PCCS TLS trust policy. When omitted, the node uses the system
+    /// trust roots — the default behaviour for `pccs.phala.network`,
+    /// `api.trustedservices.intel.com`, etc. Set the `[mpc_node_config.pccs_tls]`
+    /// sub-table (with `mode = "ca_cert_pem"` or `mode = "insecure"`)
+    /// when pointing at a local PCCS that uses a self-signed cert.
```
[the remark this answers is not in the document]
barakeinav1: Addressed in 8fa888e2 — the doc-comment now says "default reqwest+rustls trust roots — bundled Mozilla webpki-roots" instead of "system trust roots." The actual TLS backend is webpki-roots (not OS-native), and operators reading the old wording could be misled.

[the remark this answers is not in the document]
barakeinav1: Addressed in 8fa888e2 — the PR body has been rewritten to reflect the current `[[mpc_node_config.pccs_endpoints]]` array-of-tables shape with per-URL `tls = ...` overrides. The earlier two-field form (`pccs_ca_cert_pem` / `pccs_tls_insecure`) is gone; per-URL overrides on each endpoint replace it (per the design discussion captured on this PR).

crates/tee-authority/src/tee_authority.rs
barakeinav1: according to our discussion this is not a real security feature but instead the loopback gates is a misconfiguration guardrail, not a security barrier. 
added a comment here https://github.com/near/mpc/pull/3026/changes/12c04d96b3d9e0627815bde3e85184e9e63388a2

barakeinav1:
@claude review

barakeinav1:
> **One minor non-blocking observation (not raised elsewhere):**
> 
> * `build_pccs_http_client` sets `reqwest::Client::builder().timeout(PCCS_REQUEST_TIMEOUT)` (10s), and the caller in `fetch_collateral_from` also wraps `client.fetch(quote)` in `tokio::time::timeout(PCCS_REQUEST_TIMEOUT, ...)`. The double timeout is harmless (reqwest's fires per individual HTTP request, tokio's wraps the entire fetch operation), but worth noting since they use the same constant — if the intent differs, separate constants would make that clearer. Not blocking.



Acknowledged the double-timeout observation in be5b98eb — the `PCCS_REQUEST_TIMEOUT` doc-comment now explains the two layers (`reqwest::Client::builder().timeout` per HTTP request, `tokio::time::timeout` around the whole fetch) and notes this is intentional given each fetch is a single HTTP request today. If we ever split that (e.g. retries inside a fetch needing different per-attempt vs total bounds), separate constants would make the layering explicit.

crates/launcher-interface/src/types.rs
```diff
@@ -22,6 +23,27 @@ pub enum TeeAuthorityConfig {
...
+}
+
+/// Per-endpoint TLS trust override. `None` = default reqwest+rustls trust
+/// roots (bundled Mozilla webpki-roots).
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "mode", rename_all = "snake_case", deny_unknown_fields)]
+pub enum PccsTlsTrust {
+    /// Pin a PEM certificate as an additional trust anchor. Default
+    /// trust roots stay active.
+    CaCertPem { ca_cert_pem: String },
```
pbeza, line 42: Nit: how about using a more proper type here, if possible? Maybe `reqwest::Certificate` or some PEM-native type, so we don’t need to parse it later somewhere deep in the call stack.

crates/node-config/src/start.rs
```diff
@@ -25,19 +26,26 @@ pub struct StartConfig {
...
-    /// Tried in order on every fetch; the first one to succeed wins, and the
-    /// rest are used only as fallbacks when earlier entries fail. At least
-    /// one URL is required. Defaults to Phala's PCCS if the field is omitted.
-    #[serde(default = "default_pccs_urls")]
-    pub pccs_urls: NonEmptyVec<url::Url>,
+    /// PCCS servers used to fetch TDX attestation collateral. Each entry
+    /// is a URL plus an optional per-URL TLS trust override. Tried in
+    /// order on every fetch; the first one to succeed wins, the rest are
+    /// fallbacks. At least one entry is required. Defaults to Phala's
+    /// PCCS if the field is omitted.
```
pbeza, line 33: Why do we use Phala as the default instead of Intel’s PCCS if it’s available?

crates/launcher-interface/src/types.rs
```diff
@@ -22,6 +23,27 @@ pub enum TeeAuthorityConfig {
     Dstack { dstack_endpoint: PathBuf },
 }
 
+/// One PCCS endpoint: a URL plus an optional per-URL TLS trust override.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct PccsEndpointConfig {
+    pub url: Url,
+    #[serde(default)]
+    pub tls: Option<PccsTlsTrust>,
```
pbeza, line 32: Nit: I think we could get rid of the `Option` altogether and add an extra `PccsTlsTrust` enum variant representing the default TLS certs. It’d be more self-explanatory, idiomatic, and clean.

crates/node-config/src/start.rs
```diff
@@ -205,77 +215,93 @@ mod tests {
...
         // When
         let parsed: Wrapper = toml::from_str(&toml_input).unwrap();
-        let urls: Vec<url::Url> = parsed.pccs_urls.into_iter().collect();
+        let entries: Vec<PccsEndpointConfig> = parsed.pccs_endpoints.into_iter().collect();
 
         // Then
-        assert_eq!(urls, expected);
+        assert_eq!(entries.len(), 1);
+        assert_eq!(entries[0].url.as_str(), URL);
+        assert!(entries[0].tls.is_none());
```
pbeza: Nit: we typically prefer comparing `entries` as a whole rather than individual elements accessed by index. It’s a bit more robust—makes sure the whole structure matches, not just a couple fields here and there.

crates/node-config/src/start.rs
```diff
@@ -205,77 +215,93 @@ mod tests {
...
+        assert_eq!(entries.len(), 3);
+        assert_eq!(entries[0].url.as_str(), "https://localhost:8081/");
+        assert_matches!(entries[0].tls, Some(PccsTlsTrust::Insecure));
+        assert_eq!(entries[1].url.as_str(), "https://pccs.phala.network/");
+        assert!(entries[1].tls.is_none());
+        assert_eq!(
+            entries[2].url.as_str(),
+            "https://api.trustedservices.intel.com/"
+        );
+        assert!(entries[2].tls.is_none());
```
pbeza: Same comment as [here](https://github.com/near/mpc/pull/3026/changes#r3159738362).

crates/node-config/src/start.rs
```diff
@@ -205,77 +215,93 @@ mod tests {
...
         // When
         let parsed: Wrapper = toml::from_str("").unwrap();
-        let urls: Vec<url::Url> = parsed.pccs_urls.into_iter().collect();
+        let entries: Vec<PccsEndpointConfig> = parsed.pccs_endpoints.into_iter().collect();
 
         // Then
-        assert_eq!(urls, expected);
+        assert_eq!(entries.len(), 1);
+        assert_eq!(entries[0].url, expected_url);
+        assert!(entries[0].tls.is_none());
```
pbeza: Same comment as [here](https://github.com/near/mpc/pull/3026/changes#r3159738362).

crates/tee-authority/src/tee_authority.rs
```diff
@@ -120,12 +130,103 @@ impl Default for DstackTeeAuthorityConfig {
...
+        None => false,
+    }
+}
+
+/// Validate the PCCS endpoint list at startup. Runs once during config
+/// conversion to fail fast on operator config mistakes:
+///
+/// - Each `PccsTlsTrust::CaCertPem` must contain a parseable PEM
+///   certificate (caught here so a typo'd cert fails at startup with a
+///   clear message rather than at the first attestation attempt).
```
pbeza: We can remove this function—and the functions it relies on—if you address this comment: https://github.com/near/mpc/pull/3026/changes#r3159691390.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -120,12 +130,103 @@ impl Default for DstackTeeAuthorityConfig {
...
+/// - Each `PccsTlsTrust::Insecure` is only honored for loopback-ish URLs
+///   so a copy-pasted dev config cannot silently disable TLS validation
+///   against a real network endpoint. The loopback gate is a
+///   misconfiguration guardrail, not a security barrier.
```
pbeza: I still think this is a very brittle check — more like security through obscurity — and not something we really need, as discussed here: https://github.com/near/mpc/pull/3026/#discussion_r3145908731.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -199,42 +300,64 @@ impl TeeAuthority {
...
+                        url: endpoint.url.clone(),
+                        source,
+                    }
+                })?;
+                dcap_qvl::collateral::CollateralClient::<dcap_qvl::configs::DefaultConfig>::new(
+                    http,
+                    endpoint.url.as_str(),
+                )
+            }
+        };
```
pbeza: I feel you could put all of this into the `build_pccs_http_client` function and have it return `PccsEndpointError::ClientConstruction` on error, so we don’t need extra error-mapping code here.

crates/tee-authority/Cargo.toml
```diff
@@ -17,6 +17,13 @@ hex = { workspace = true }
 launcher-interface = { workspace = true }
 mpc-attestation = { workspace = true, features = ["dstack-conversions"] }
 near-mpc-bounded-collections = { workspace = true }
+# Constrained to the 0.12 series to match `dcap-qvl`'s `reqwest`. `dcap-qvl`'s
+# `CollateralClient::new(reqwest::Client, _)` requires us to hand it a client
+# typed by the same major. The workspace pin is on a newer major (0.13) used
+# by other crates, so we declare an explicit `0.12.x` dep here.
+# TODO(#3027): drop this once `dcap-qvl` exposes the `HttpClient` trait
+# upstream, then we can implement it for any client and stop pinning.
```
pbeza: Can you open a relevant GitHub issue in the Phala repo, please?

pbeza (dismissed):
I have quite a few nits — nothing blocking, but since there are quite a few, it’d be great to address them if other reviewers feel the same.

crates/node-config/src/start.rs
barakeinav1: I didn't change the order in this PR. 
but I don't have strong opinion on using intel PCCS as a defualt

crates/launcher-interface/src/types.rs
barakeinav1: this will force every operator to add  tls = { mode = "default" } for each entry in the config.
you still think we should do it?

crates/tee-authority/src/tee_authority.rs
barakeinav1: do you have suggestion?

crates/node-config/src/start.rs
barakeinav1: Done in f81bd0c4 along with the sibling tests — see r3159738362.

crates/node-config/src/start.rs
barakeinav1: Done in f81bd0c4 along with the sibling tests — see r3159738362.

crates/node-config/src/start.rs
barakeinav1: Done in f81bd0c4 — rewrote this and the two sibling tests to assert against a single `Vec<PccsEndpointConfig>` literal. Required deriving `PartialEq, Eq` on `PccsEndpointConfig` and `PccsTlsTrust`.

crates/tee-authority/src/tee_authority.rs
barakeinav1: Done in f81bd0c4. `build_pccs_http_client` now takes `&PccsEndpointConfig` and returns `Result<reqwest::Client, PccsEndpointError>` directly. Both call sites (`fetch_collateral_from`, `validate_pccs_endpoints`) shed the URL-attaching `.map_err(...)` boilerplate.

crates/tee-authority/Cargo.toml
barakeinav1: kevin already did this-  [PR #157]([https://github.com/Phala-Network/dcap-qvl/pull/157](vscode-file://vscode-app/c:/Users/barak/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html)) (reqwest 0.12→0.13 bump) is merged at Phala but not yet in a published release

crates/node-config/src/start.rs
barakeinav1: do you think we should have both intel and phala as defaults? or just intel

crates/launcher-interface/src/types.rs
barakeinav1: Agree the per-fetch parse is wasteful — but typing this as `reqwest::Certificate` in `launcher-interface` would pull `reqwest` into a schema crate that's deliberately thin (just `serde` + `url` + `thiserror` today). Worth noting too: the launcher binary never references `PccsEndpointConfig` / `PccsTlsTrust` — these types only landed in `launcher-interface` by association with the other TEE config, not because the launcher boundary needs them.

Plan ([#3068](https://github.com/near/mpc/issues/3068)): relocate `PccsEndpointConfig` / `PccsTlsTrust` / `DEFAULT_PCCS_URL` to `mpc-node-config`, and parse the PEM **once** at `DstackTeeAuthorityConfig` construction (storing the built `reqwest::Client` per endpoint). That solves the parse-once concern at the right layer without dragging `reqwest` into the schema crate, and it also resolves r3159779606 by making `validate_pccs_endpoints` redundant. Tracking as a follow-up to keep this PR scoped to the TLS feature.

crates/tee-authority/src/tee_authority.rs
barakeinav1: Right — and it's broader than just `build_pccs_http_client`'s fate. Tracked in [#3068](https://github.com/near/mpc/issues/3068): once the per-endpoint `reqwest::Client` is built once at `DstackTeeAuthorityConfig` construction, the validator's responsibilities (PEM parse error, loopback gate, startup WARN for `insecure`) all collapse into the constructor, and `validate_pccs_endpoints` deletes. Same follow-up as r3159691390.

crates/node-config/src/start.rs
pbeza: I’m OK with using Phala — was just wondering how we ended up with it as the default. I think Intel might have some rate limits.

crates/tee-authority/src/tee_authority.rs
pbeza: Yeah, I’d just remove that extra check since it might be ineffective anyway — but if other reviewers are fine with it, it can stay as is.

crates/tee-authority/Cargo.toml
pbeza: You can just pin to the latest `dcap-qvl` commit—we’re alread pinning it this way anyway:

https://github.com/near/mpc/blob/f830d890be90527b312b4c5b3b9338fb26972aef/Cargo.toml#L103-L110

crates/node-config/src/start.rs
barakeinav1: we started only with phala, so this became the default

crates/tee-authority/Cargo.toml
barakeinav1: Done in 877730eb.

Bumped the workspace `dcap-qvl` git pin from 9c21fa52 (2026-04-24) to 63e9d4ae (2026-04-27), bringing in [#156](https://github.com/Phala-Network/dcap-qvl/pull/156) (HttpClient trait) and [#157](https://github.com/Phala-Network/dcap-qvl/pull/157) (reqwest 0.13). Deleted the cross-major `reqwest = "0.12.24"` direct dep + the TODO comment; tee-authority now inherits `reqwest = { workspace = true }` like every other consumer. Added a small `PccsHttpClient` adapter (~30 LOC) wrapping a `reqwest::Client` so we can use our workspace reqwest version with the new `HttpClient` API. `cargo tree -i reqwest:0.13.2` shows tee-authority as a direct caller; `cargo tree -i reqwest:0.12.28` is now exclusively transitive deps from `dstack-sdk` / `near-client` / `near-jsonrpc-client` (out of our control). Closes #3027.

crates/launcher-interface/src/types.rs
```diff
@@ -22,6 +23,27 @@ pub enum TeeAuthorityConfig {
...
+}
+
+/// Per-endpoint TLS trust override. `None` = default reqwest+rustls trust
+/// roots (bundled Mozilla webpki-roots).
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "mode", rename_all = "snake_case", deny_unknown_fields)]
```
netrome: Nit: "mode" feels a bit generic. Perhaps "override" is a better tag name here since this configures the override of the trust roots. 
```suggestion
#[serde(tag = "override", rename_all = "snake_case", deny_unknown_fields)]
```

netrome (dismissed):
Note that this is a breaking config change, which is fine since this config isn't configured for any nodes in production except our TEE node iirc. But I'll update the title to mark it as breaking (by adding an exclamation mark)

crates/tee-authority/src/tee_authority.rs
barakeinav1: @netrome what do you think?

barakeinav1: Done in 60102476 — gate removed.

Took your call: deleted `pccs_host_is_loopback`, `LOOPBACK_PCCS_DOMAINS`, the `anyhow::ensure!` in `validate_pccs_endpoints`, and the four rstest functions covering the gate (~110 LOC removed). The startup WARN log for `Insecure` stays — that remains the operator-facing signal.

Replaced the runtime check with an operator-facing note in `deployment/cvm-deployment/user-config.toml`: `mode = "insecure"` is recommended only for a local PCCS reachable over loopback, with a recommendation to prefer `ca_cert_pem` whenever the cert can be pinned. Doc comment on `PccsTlsTrust::Insecure` updated to match.

crates/launcher-interface/src/types.rs
barakeinav1: Done in 0ce24ddb. Renamed across the serde attribute, doc comments, the runtime WARN log message, the `user-config.toml` reference, and the test TOML. Reads cleaner — `override = "ca_cert_pem"` matches the field's actual semantics.

pbeza (approved):
I have quite a few nits — nothing blocking, but since there are quite a few, it’d be great to address them if other reviewers feel the same.

netrome (approved):
Thanks for the updates 🙏

Issue #2890, opened 2026-04-15
Debug page for submitted transactions


netrome:
### Background

One big limitations of our current architecture is that nodes do "fire and forget" submission of transactions. This means that if transactions fail to be picked up by the blockchain for any reason (e.g. signature failure, invalid nonce etc.) the nodes are completely oblivious to this.

Unfortunately fixing this problem completely requires bigger architectural changes, but in the meantime, we should make it extremely easy to debug and track which transactions a node has submitted so an operator can manually check the status of these transactions on-chain.

### User Story

As a node operator I'd like it to be easy to debug if my node's transactions are failing or missing on-chain.

### Acceptance Criteria

We have a dedicated debug page that lists recently submitted transactions from the node with as much relevant info as possible (txid, nonce, signature etc.)

### Resources & Additional Notes

Note: a problem a partner encountered was that when you submit several transactions from the same account in a short time period they may be processed out-of-order leading to transactions with lower nonces being rejected at times because a higher-nonce transaction was already included. This may explain some sporadic signature failures we're experiencing on-chain.

netrome:
Note: We have for a long time seen sporadic signing failures on mainnet that are not reported by signature leader metrics. They seem to correlate with traffic (not entirely sure), and the leading theory is that when the same leader submits multiple responses in close proximity, they may be processed out of order leading to the lower nonce transaction being rejected.

Once we've got this debug page in place, we should have better ways of confirming if this is the case or not.

netrome:
I've created an internal project page to track signing reliability work. Let's not jump too eagerly on this but finish per-domain thresholds and pytest deprecation first and then look into this. https://docs.nearone.org/doc/signing-reliability-jhej0lU3NA

Lowering the priority of this issue for now.

netrome:
Network traffic has increased and we're seeing more failures in mainnet as a consequence. Moreover, pytest deprecation is closed now - so let's prioritize this work.

netrome:
Bumping priority to P1

pbeza:
Related: https://github.com/near/mpc/issues/3167.

PR #3029, opened 2026-04-27
docs: recommend unlimited function-call allowance for the MPC operator account

### LINKED ISSUE #3028: External operator guide: bump function-call allowance recommendation (silent-failure mode)
### Background

A silent signing-failure incident on the NearOne testnet node (2026-04-25 → 2026-04-26) traced back to the **function-call access key on the operator account running out of NEAR allowance** and being unable to submit `Respond` transactions. Because the node uses fire-and-forget broadcast (`broadcast_tx_async`), the RPC's pre-validation rejects under-funded txs and the node never sees an error — it keeps reporting "succeeded" rounds while every response is rejected before reaching a validator. The internal post-mortem is in [near/mpc-private#307](https://github.com/near/mpc-private/pull/307) (private).

The `1 NEAR` allowance value used during the migration came from the external operator guide, which is the canonical reference any operator follows when granting function-call access to their MPC node's signer key. With sustained signing load, that budget drains in ~2 days, after which the operator's node silently stops contributing — exactly the failure mode we just hit.

The per-key allowance does **not** add meaningful safety in this configuration: the key is already scoped to a fixed set of methods on `v1.signer*` via `--contract-account-id` + `--function-names`, so it cannot burn NEAR for anything else. Capping the allowance just creates a slow-burn outage that's hard to detect from the node logs.

### User Story

As an MPC node operator following the external guide to grant function-call access on my operator account, I want the recommended `add-key` command to use a large or unlimited allowance — and to understand why — so my node doesn't silently stop submitting `Respond` transactions a few days after I bring it online.

### Acceptance Criteria

- [ ] Update `docs/running-an-mpc-node-in-tdx-external-guide.md` so the `near account add-key … grant-function-call-access …` example uses a large or unlimited allowance (e.g. `--allowance unlimited` if the CLI version supports it; otherwise `--allowance '1000 NEAR'` or similar).
- [ ] Apply the same update to any other doc that reproduces this `add-key` snippet (at minimum check `docs/node-migration-guide.md`).
- [ ] Add a short note next to the example explaining that the per-key allowance is *not* a safety boundary — the key is already restricted by `--contract-account-id` + `--function-names` — and a small allowance is purely a slow-burn failure mode.
- [ ] Recommend that operators monitor / alert on the remaining allowance of the function-call key on the operator account, since the failure mode is silent at the node level today.

- Internal incident post-mortem: [near/mpc-private#307](https://github.com/near/mpc-private/pull/307) (private).
- Related: [#2890](https://github.com/near/mpc/issues/2890) — track on-chain status of `Respond` transactions in the node, so this class of failure stops being silent at the node level.
- Related: [#2102](https://github.com/near/mpc/issues/2102) — remove unused `number_of_responder_keys` config (a distraction during this incident).
- A companion issue in `near/mpc-private` covers the corresponding internal-doc updates (`migration_log.md`, `tdx-testnet-node.md` runbook) without rewriting the historical migration log: [near/mpc-private#309](https://github.com/near/mpc-private/issues/309).

barakeinav1:
Closes #3028.

## Why

A 2026-04-26 testnet signing-failure incident (post-mortem in [near/mpc-private#307](https://github.com/near/mpc-private/pull/307), private) traced back to the function-call access key on the operator account running out of NEAR allowance and being unable to submit `Respond` transactions. With sustained signing load, the `1 NEAR` budget that this guide previously recommended drains in roughly two days, after which the RPC silently rejects every `respond*` tx — and because the node uses `broadcast_tx_async` (fire-and-forget), the rejection is invisible in node logs. The node keeps reporting "succeeded" rounds while every response is rejected before reaching a validator.

The per-key allowance is *not* a meaningful safety boundary in this configuration: the key is already restricted to a fixed set of methods on the MPC contract via `--contract-account-id` + `--function-names`, so it cannot burn NEAR for anything else. Capping the allowance just creates a slow-burn outage that's hard to detect.

## Changes

- Adds an explicit `ALLOWANCE` row to the *Parameters* section, recommending `unlimited` (or a very large value) and explaining why per-key capping doesn't help here, plus a reminder to monitor remaining allowance via Grafana since the failure mode is silent at the node level today.
- Updates the *Example Command* literal from `--allowance '1 NEAR'` to `--allowance unlimited`.
- Updates the *Sample Bash Script* default from `ALLOWANCE="1 NEAR"` to `ALLOWANCE="unlimited"`, with an inline comment pointing back at the new Parameters explanation.

## Verification

The same change is being made in the corresponding internal docs in `near/mpc-private` (companion follow-up [near/mpc-private#309](https://github.com/near/mpc-private/issues/309)) so the migration playbook + on-call runbook stay aligned with this guide.

## Out of scope

- Adding the actual Grafana alert on remaining allowance — separate follow-up.
- Surfacing on-chain `Respond` transaction status inside the node, which would have made this incident loud at the node level instead of silent — tracked in [#2890](https://github.com/near/mpc/issues/2890).

### DIFF
--- a/docs/running-an-mpc-node-in-tdx-external-guide.md
+++ b/docs/running-an-mpc-node-in-tdx-external-guide.md
@@ -918,6 +918,23 @@ This section shows how to add the MPC node's public key (from the previous secti
 
   > **Note:** This must be a single comma-separated string with no spaces or newlines.
 
+* **`ALLOWANCE`** → The total NEAR budget the function-call key can spend on
+  gas across all calls signed with it. **Recommended: `unlimited`** (or a very
+  large value, e.g. `'1000 NEAR'`).
+
+  > **Why unlimited?** The per-key allowance is *not* a meaningful safety
+  > boundary here — the key is already restricted to specific methods on
+  > `$MPC_CONTRACT_ID` via `--contract-account-id` + `--function-names`, so
+  > it cannot burn NEAR for anything else. A small finite allowance only
+  > creates a slow-burn failure mode: once the budget runs out, the RPC's
+  > pre-validation silently rejects every `respond*` transaction, and
+  > because the node uses `broadcast_tx_async` (fire-and-forget), the
+  > rejection is invisible in node logs — the node will keep reporting
+  > "succeeded" rounds while every response is dropped before reaching a
+  > validator. **Whatever value you pick, monitor and alert on remaining
+  > allowance** (e.g. via Grafana on `view_access_key`) so you don't
+  > rediscover this the same way we did.
+
 ---
 
 #### Example Command
@@ -925,7 +942,7 @@ This section shows how to add the MPC node's public key (from the previous secti
 ```bash
 ./target/release/near account add-key $ACCOUNT_ID \
   grant-function-call-access \
-  --allowance '1 NEAR' \
+  --allowance unlimited \
   --contract-account-id $MPC_CONTRACT_ID \
   --function-names $METHOD_NAMES \
   use-manually-provided-public-key $MPC_NODE_PUBLIC_KEY \
@@ -945,7 +962,7 @@ This section shows how to add the MPC node's public key (from the previous secti
 ACCOUNT_ID="your-node-account.testnet"
 MPC_CONTRACT_ID="v1.signer-prod.testnet"    # use "v1.signer" for mainnet
 MPC_NODE_PUBLIC_KEY="ed25519:YOUR_PUBLIC_KEY_HERE"
-ALLOWANCE="1 NEAR"
+ALLOWANCE="unlimited"   # see the Parameters section above for why
 NETWORK="testnet"   # or "mainnet"
 
 # Methods the MPC node is allowed to call
--- a/docs/running-an-mpc-node-in-tdx-external-guide.md
+++ b/docs/running-an-mpc-node-in-tdx-external-guide.md
@@ -918,22 +918,9 @@ This section shows how to add the MPC node's public key (from the previous secti
 
   > **Note:** This must be a single comma-separated string with no spaces or newlines.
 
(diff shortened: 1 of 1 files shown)


docs/running-an-mpc-node-in-tdx-external-guide.md
```diff
@@ -918,14 +918,18 @@ This section shows how to add the MPC node's public key (from the previous secti
 
   > **Note:** This must be a single comma-separated string with no spaces or newlines.
 
+* **`ALLOWANCE`** → Use `unlimited`. A finite allowance just means the node
+  will eventually stop being able to submit `respond*` transactions once the
+  budget runs out, with no error visible in node logs.
+
```
kevindeforth, line 924: Q: is there a way to reset the allowance?

kevindeforth (approved):
Thanks!

docs/running-an-mpc-node-in-tdx-external-guide.md
netrome: No, we had to remove the key and add it again with a new allowance.

netrome (approved):
Thanks!

docs/running-an-mpc-node-in-tdx-external-guide.md
```diff
@@ -918,14 +918,31 @@ This section shows how to add the MPC node's public key (from the previous secti
 
   > **Note:** This must be a single comma-separated string with no spaces or newlines.
 
+* **`ALLOWANCE`** → The total NEAR budget the function-call key can spend on
+  gas across all calls signed with it. **Recommended: `unlimited`** (or a very
+  large value, e.g. `'1000 NEAR'`).
```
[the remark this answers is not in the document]
barakeinav1: will be fix in a follow-up PR

Issue #3031, opened 2026-04-27
batch adding requests to the RocksDB


kevindeforth:
### Background

We currently add each request individually to the database
https://github.com/near/mpc/blob/c8f8186b7e9631a505ceac559b960a0b70f552b9/crates/node/src/mpc_client.rs#L260

This has two minor disadvantages:
- in case there is an indexer or node restart mid block update, we may not have added all requests from the same block. It might be desirable to have atomic updates instead (the entire block, or no block).
- performance wise, batching would make sense. If the node's indexer lags behind and has `M` threads waiting on these broadcasts, then for `N` signatures per block, we have `M*N` wake-ups per block, whereas it could be `M` or less. Also, note that we do `N` RocksDB updates, whereas a single one would suffice. This is not an issue right now, but may become a concern if we start to see heavier loads.

### User Story

As a dev, I like easy to reason code.

### Acceptance Criteria

Batch the database updates such that we do a single database commit per block per column. Change the subscription logic.
Optional: Instead of broadcasting every newly received request to all waiters, have a map `request_hash <--> oneshot_channel` to do targeted wake-ups.

PR #3034, opened 2026-04-27
refactor: `mpc_client.rs` and `PendingRequests` unit tests

### LINKED ISSUE #3033: Simplify mpc_client and clean up queue unit tests
### Background

We currently have a lot of code in a select statement in `mpc_client`.
Future refactor will touch on that code and review would be simpler if the conversion of request types from chain to the request types stored inside the queue were done in helper methods.

Additionally, we have vest duplication inside `crate/node/src/requests/queue.rs`. The `PendingRequest` struct is generic and the unit tests cover logic not depending on the generics itself. We can thus safely delete one of the two test types (they seem to be copy-paste either way).

### User Story

As a reviewer, I like easy to review code.

### Acceptance Criteria

simplify the following code, by moving conversion of requests to helper methods.

https://github.com/near/mpc/blob/c8f8186b7e9631a505ceac559b960a0b70f552b9/crates/node/src/mpc_client.rs#L239-L331

_No response_

kevindeforth:
resolves #3033

follow-up issue: #3031

### DIFF
--- a/crates/node/src/requests/queue.rs
+++ b/crates/node/src/requests/queue.rs
@@ -2,10 +2,8 @@ use super::debug::{CompletedRequest, CompletedRequests};
 use crate::indexer::types::ChainRespondArgs;
 use crate::primitives::ParticipantId;
 use crate::requests::metrics;
-use crate::requests::recent_blocks_tracker::{
-    BlockViewLite, CheckBlockResult, RecentBlocksTracker,
-};
-use crate::types::{self, Request, RequestId};
+use crate::requests::recent_blocks_tracker::{CheckBlockResult, RecentBlocksTracker};
+use crate::types::{self, Request, RequestId, RequestsFromBlock};
 use k256::sha2::Sha256;
 use near_indexer_primitives::types::NumBlocks;
 use near_indexer_primitives::CryptoHash;
@@ -32,7 +30,7 @@ pub const CHECK_EACH_REQUEST_INTERVAL: Duration = Duration::seconds(1);
 const STALE_PARTICIPANT_THRESHOLD: NumBlocks = 10;
 /// The number of blocks after which a request is assumed to have timed out.
 /// This is equal to the yield-resume timeout on the blockchain.
-const REQUEST_EXPIRATION_BLOCKS: NumBlocks = 200;
+pub(super) const REQUEST_EXPIRATION_BLOCKS: NumBlocks = 200;
 /// The maximum time we'll wait, after a transaction is submitted to the chain, before we decide
 /// that the transaction is lost and that we should retry.
 const MAX_LATENCY_BEFORE_EXPECTING_TRANSACTION_TO_FINALIZE: Duration = Duration::seconds(10);
@@ -280,14 +278,14 @@ impl<RequestType: Request + Clone, ChainRespondArgsType: ChainRespondArgs>
     /// This must be called for every block that comes from the indexer.
     /// These are the requests successfully submitted in the block, and the
     /// completed_requests are the requests whose responses are included in the block.
-    pub fn notify_new_block(
-        &mut self,
-        requests: Vec<RequestType>,
-        completed_requests: Vec<RequestId>,
-        block: &BlockViewLite,
-    ) {
-        let add_result = match self.recent_blocks.add_block(
+    pub(crate) fn notify_new_block(&mut self, requests: RequestsFromBlock<RequestType>) {
+        let RequestsFromBlock::<RequestType> {
             block,
+            requests,
+            completed_requests,
+        } = requests;
+        let add_result = match self.recent_blocks.add_block(
+            &block,
             BufferedBlockData {
                 requests: requests.iter().map(|r| r.get_id()).collect(),
                 completed_requests,
@@ -550,15 +548,15 @@ impl<RequestType: Request + Clone, ChainRespondArgsType: ChainRespondArgs>
 #[cfg(test)]
 mod tests {
     use super::{NetworkAPIForRequests, PendingRequests, QueuedRequest};
-    use crate::indexer::types::{ChainCKDRespondArgs, ChainSignatureRespondArgs};
+    use crate::indexer::types::ChainSignatureRespondArgs;
     use crate::primitives::ParticipantId;
     use crate::requests::queue::{
         CHECK_EACH_REQUEST_INTERVAL, MAX_ATTEMPTS_PER_REQUEST_AS_LEADER,
-        MAX_LATENCY_BEFORE_EXPECTING_TRANSACTION_TO_FINALIZE,
+        MAX_LATENCY_BEFORE_EXPECTING_TRANSACTION_TO_FINALIZE, REQUEST_EXPIRATION_BLOCKS,
     };
-    use crate::requests::recent_blocks_tracker::tests::TestBlockMaker;
+    use crate::requests::recent_blocks_tracker::tests::{TestBlock, TestBlockMaker};
     use crate::tests::into_participant_ids;
-    use crate::types::{CKDRequest, SignatureRequest};
+    use crate::types::{RequestsFromBlock, SignatureRequest};
     use mpc_primitives::domain::DomainId;
     use near_indexer_primitives::CryptoHash;
     use near_mpc_contract_interface::types::{Payload, Tweak};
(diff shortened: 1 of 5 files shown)


crates/node/src/requests/queue.rs
```diff
@@ -675,260 +639,224 @@ mod tests {
         }
     }
 
-    #[test_log::test]
-    fn test_pending_ckd_requests_leader_retry() {
```
kevindeforth, line 679: The ckd tests are redundant. The `PendingRequests` struct is generic over the request and response type and these unit tests do not rely on the internal logic of these types.

crates/node/src/requests/queue.rs
```diff
@@ -675,260 +639,234 @@ mod tests {
...
-        let req2 = test_sign_request(&participants, &[1]);
-        let b1 = t.block(100);
-        pending_requests.notify_new_block(
-            vec![req1.clone(), req2.clone()],
-            vec![],
-            &b1.to_block_view(),
-        );
-
-        // req1 is not attempted because we're not the leader. req2 is attempted.
+    fn test_pending_requests_requests_leader_retry() {
```
anodar: typo?

crates/node/src/types.rs
```diff
@@ -56,6 +111,30 @@ pub struct SignatureRequest {
     pub domain: DomainId,
 }
 
+pub(crate) trait FromChain<T> {
```
anodar: nit: consider moving trait before impl.

crates/node/src/types.rs
```diff
@@ -3,10 +3,46 @@ use std::fmt;
...
+    pub(crate) block: BlockViewLite,
+    pub(crate) requests: Vec<T>,
+    pub(crate) completed_requests: Vec<RequestId>,
+}
+
+impl<T> RequestsUpdate<T> {
+    pub(crate) fn from_chain<U>(
+        block: &BlockViewLite,
+        new_requests: Vec<U>,
+        completed_requests: Vec<CryptoHash>,
```
anodar: Not clear why is this CryptoHash and not RequestId.

If there's a good reason for it, please consider adding some comment.

crates/node/src/requests/queue.rs
```diff
@@ -675,260 +639,234 @@ mod tests {
...
+    impl TestSetup {
+        const MY_INDEX: usize = 1;
+        fn new() -> (PendingRequests<TestRequest, TestRequestRespondArgs>, Self) {
+            let clock = FakeClock::default();
+            let participants = into_participant_ids(&generate_participants(4));
+            let my_participant_id = participants[Self::MY_INDEX];
+            let network_api = Arc::new(TestNetworkAPI::new(&participants));
+
+            let pending_requests =
+                PendingRequests::<SignatureRequest, ChainSignatureRespondArgs>::new(
```
anodar: nit: PendingRequests::<TestRequest, TestRequestRespondArgs>, since you are aliasing those types, although I don't see benefit of aliasing them.

I'm guessing it's to show that test is generic and doesn't depend on internal logic of those types but that can be a comment.

crates/node/src/requests/queue.rs
```diff
@@ -32,7 +30,7 @@ pub const CHECK_EACH_REQUEST_INTERVAL: Duration = Duration::seconds(1);
 const STALE_PARTICIPANT_THRESHOLD: NumBlocks = 10;
 /// The number of blocks after which a request is assumed to have timed out.
 /// This is equal to the yield-resume timeout on the blockchain.
-const REQUEST_EXPIRATION_BLOCKS: NumBlocks = 200;
+pub(super) const REQUEST_EXPIRATION_BLOCKS: NumBlocks = 200;
```
anodar: nit: This doesn't seem to be used outside queue.rs, do we need this to be public?

crates/node/src/requests/queue.rs
```diff
@@ -675,260 +639,234 @@ mod tests {
...
-            drop(to_attempt);
-            clock.advance(CHECK_EACH_REQUEST_INTERVAL);
-        }
+        // Then: we should not have any requests to attempt
+        setup.advance_clock(CHECK_EACH_REQUEST_INTERVAL);
         assert_eq!(pending_requests.get_requests_to_attempt().len(), 0);
     }
 
     #[test_log::test]
     fn test_pending_signature_requests_abort_after_maximum_attempts() {
```
anodar: nit: this isn't changed in your pr, but to note, these test names don't really follow engineering-standards.md guide. It should be something like: 

`pending_signature_requests__should_abort_after....`
`pending_signature_requests__should_discard_old....`

crates/node/src/requests/queue.rs
```diff
@@ -675,260 +639,234 @@ mod tests {
...
-
-        for participant in &participants {
-            network_api.set_height(*participant, 100);
-        }
-
-        let t = TestBlockMaker::new();
+    type TestRequest = SignatureRequest;
+    type TestRequestRespondArgs = ChainSignatureRespondArgs;
+
+    struct TestSetup {
```
anodar, line 644: Why not make pending requests to be a field of this struct? That way you don't have to keep reference of it in test setup and just call methods on TestEnvSetup, less bookkeeping on callsites.

crates/node/src/requests/queue.rs
```diff
@@ -1250,101 +1003,46 @@ mod tests {
...
-        clock.advance(near_time::Duration::seconds(1));
-
+        // Given: a request queue with a request as leader
+        let (mut pending_requests, mut setup) = TestSetup::new();
+        setup.advance_clock(Duration::seconds(1));
+        let req1 = setup.add_request_leader();
+        let requests = setup.update();
+        pending_requests.notify_new_block(requests);
+        // When: we are lagging behind
+        setup.advance_clock(Duration::microseconds(2432123));
```
anodar, line 1034: Oddly specific number :D Can we just do `Duration::seconds(2.5)` or something like that ?

anodar (dismissed):

crates/node/src/requests/queue.rs
kevindeforth: > I'm guessing it's to show that test is generic and doesn't depend on internal logic of those types 

Yeah, exactly.

> but that can be a comment.

comments are easy to miss.

crates/node/src/requests/queue.rs
kevindeforth: I am not fond of hiding the struct and calls to methods that the unit test intends to test behind a wrapper.

crates/node/src/requests/queue.rs
kevindeforth: Existing code, would prefer not to change. Right now, it also takes account of rounding values. The benefit of that is arguably limited, but changing it to something simpler is not obviously better to me either.

anodar (approved):
Thank you!

crates/node/src/requests/queue.rs
```diff
@@ -675,260 +638,235 @@ mod tests {
...
-        let b2 = b1.child(101);
-        pending_requests.notify_new_block(
-            vec![req3.clone(), req4.clone()],
-            vec![],
-            &b2.to_block_view(),
-        );
+        // When: a new request is issues for which we are a leader
+        let req3 = setup.add_request_leader();
+        let _req4 = setup.add_request_follower();
+        let requests = setup.update();
```
gilcu3, line 799: I guess req4 is materialized inside `setup.update` somehow?

crates/node/src/requests/queue.rs
gilcu3: seems similar to the number below, `2s432ms` plus an extra 123 ns

gilcu3 (approved):
Thank you!

crates/node/src/requests/queue.rs
kevindeforth: yeah, I was not able to find a good name for this method. Had `add_block()` at some point, but that also didn't seem correct.

The .update() method:
- extends the chain for the block tracker 
- takes all generated requests and responses since the last call to `.update` and generates the request update for them.

I played around with different ergonomics, nothing was really nice. My feeling is that this is due to not having good separation of concerns and should get better as we continue the refactors.

gilcu3: Thanks for the long explanation!

PR #3048, opened 2026-04-28
refactor(tee-authority): add typed PCCS endpoint errors and fallback logging

### LINKED ISSUE #3046: tee-authority: improve PCCS-fallback observability and error reporting (follow-ups to #3010)
### Background

PR #3010 added support for trying multiple PCCS endpoints in order with the first success winning. Two follow-up review comments on that PR were marked non-blocking and deferred to a separate PR per https://github.com/near/mpc/pull/3010#discussion_r3152749328:

1. https://github.com/near/mpc/pull/3010#discussion_r3151863673 — when every endpoint fails, `try_each_pccs_endpoint` returns only the last URL's `anyhow::Error`, dropping the other failures on the floor. Operators have to reconstruct the full picture by reading preceding `warn` lines, and structured-error consumers (alerting, support tickets, tests) can't see the per-endpoint detail without scraping logs.
2. https://github.com/near/mpc/pull/3010#discussion_r3151864848 — a successful fetch on attempt 2+ produces no log line. That makes silent primary degradation invisible to log-based alerting: a primary that fails 100% of the time, masked by a healthy fallback, looks identical to a healthy system.

### User Story

As an MPC node operator, I want every per-endpoint PCCS failure recorded in the structured error returned when collateral fetch fails, and a positive log signal whenever the primary PCCS is silently degraded but a fallback succeeds, so that alerts and support tickets reflect the real state of the PCCS chain instead of a single last-attempt message.

### Acceptance Criteria

- [ ] `try_each_pccs_endpoint` returns a typed `Result<Collateral, AllPccsEndpointsFailed>` aggregating every per-endpoint failure in attempt order (replacing today's `anyhow::Result<Collateral>` that surfaces only the last error).
- [ ] A typed `PccsEndpointError` enum carries the URL that was tried and distinguishes the failure modes the body actually distinguishes today (`ClientConstruction`, `Timeout`, `Fetch`).
- [ ] `AttestationError::CollateralFetch` wraps the new aggregate type directly (no `anyhow::Error` adapter at the boundary).
- [ ] On the success arm of `try_each_pccs_endpoint`, when `attempt > 1`, emit a single structured `info!` naming the URL that won and the attempt number. The common path (attempt == 1) stays silent.
- [ ] Existing per-iteration `warn!` is preserved — real-time visibility during a slow fallback chain is unchanged.
- [ ] Unit tests cover: every-endpoint-failure aggregation preserves order and variants; the aggregate `Display` renders one line per failure (matching the reviewer's example shape).
- [ ] Downstream consumers in `crates/node` (`run.rs`, `tee/remote_attestation.rs`) compile unchanged — they only `?e`-format the error, so the variant payload type change is invisible to them.

- Origin PR: #3010
- Reviewer comments to address: https://github.com/near/mpc/pull/3010#discussion_r3151863673, https://github.com/near/mpc/pull/3010#discussion_r3151864848
- Author's commitment to follow up in a separate PR: https://github.com/near/mpc/pull/3010#discussion_r3152749328
- Out of scope: the four other `AttestationError` variants (`DstackClientInfo`, `TcbInfoConversion`, `QuoteGeneration`, `QuoteDecode`) still wrap `anyhow::Error` because their upstream APIs (`dstack_sdk`, `dcap_qvl`'s quote-decode path) are themselves untyped. Wrapping those is a separate refactor.

pbeza:
Closes #3046.

Two non-blocking follow-ups from #3010, addressed in a separate PR per https://github.com/near/mpc/pull/3010#discussion_r3152749328:

- **Typed aggregate error for PCCS-fallback failures.** `try_each_pccs_endpoint` now returns `Result<Collateral, AllPccsEndpointsFailed>` listing every endpoint's failure in attempt order, instead of dropping all but the last. Addresses https://github.com/near/mpc/pull/3010#discussion_r3151863673.
- **`info!` log on successful fallback.** A success on attempt 2+ now emits one structured log line so silent primary degradation is visible to alerting; common path (attempt == 1) stays silent. Addresses https://github.com/near/mpc/pull/3010#discussion_r3151864848.

See #3046 for the full scope and acceptance criteria.

### DIFF
--- a/crates/tee-authority/src/tee_authority.rs
+++ b/crates/tee-authority/src/tee_authority.rs
@@ -69,11 +69,18 @@ impl PccsEndpointError {
 }
 
 /// Returned when every configured PCCS endpoint failed. Owns the full list
-/// of per-endpoint failures in the order they were tried.
+/// of per-endpoint failures in the order they were tried. The collection is
+/// non-empty by construction: this error only ever fires after a loop over a
+/// `NonEmptyVec<Url>`, so the type-level invariant matches the value-level
+/// one and rules out nonsense renderings like "all 0 PCCS endpoints failed".
 #[derive(Debug, Error)]
-#[error("all {} PCCS endpoints failed:\n{}", failures.len(), format_pccs_failures(failures))]
+#[error(
+    "all {} PCCS endpoints failed:\n{}",
+    failures.len(),
+    format_pccs_failures(failures.as_slice())
+)]
 pub struct AllPccsEndpointsFailed {
-    pub failures: Vec<PccsEndpointError>,
+    pub failures: NonEmptyVec<PccsEndpointError>,
 }
 
 fn format_pccs_failures(failures: &[PccsEndpointError]) -> String {
@@ -290,6 +297,11 @@ where
             }
         }
     }
+    // Sound by construction: the loop iterates `pccs_urls`, which is itself a
+    // `NonEmptyVec<Url>`, so any path that exits the loop without an early
+    // `Ok(...)` return must have pushed at least one failure.
+    let failures = NonEmptyVec::try_from(failures)
+        .expect("loop over NonEmptyVec<Url> guarantees at least one failure");
     Err(AllPccsEndpointsFailed { failures })
 }
 
@@ -653,12 +665,15 @@ mod tests {
 
     /// `Display` for the aggregate error renders one line per failure, so the
     /// reviewer's example shape ("all 3 PCCS endpoints failed: …") shows up
-    /// intact in `tracing::error!(?e)` and support-ticket pastes.
+    /// intact in `tracing::error!(error = %e, …)` (Display) and
+    /// support-ticket pastes. Note: `?e` (Debug) prints the derived
+    /// `AllPccsEndpointsFailed { failures: [...] }` shape instead, so call
+    /// sites that want the multi-line summary must format with `%e`.
     #[tokio::test]
     async fn all_pccs_endpoints_failed__should_render_each_failure_on_its_own_line() {
         // Given
         let err = AllPccsEndpointsFailed {
-            failures: vec![
+            failures: NonEmptyVec::try_from(vec![
                 PccsEndpointError::Timeout {
                     url: "https://first.example/".parse().unwrap(),
                     timeout: Duration::from_secs(10),
@@ -667,7 +682,8 @@ mod tests {
                     url: "https://second.example/".parse().unwrap(),
                     source: anyhow::anyhow!("503 Service Unavailable"),
                 },
-            ],
+            ])
+            .expect("two-element vec is non-empty"),
         };
 
         // When
--- a/crates/tee-authority/src/tee_authority.rs
+++ b/crates/tee-authority/src/tee_authority.rs
@@ -634,6 +634,28 @@ mod tests {
     /// attempt.
     #[tokio::test]
     async fn try_each_pccs_endpoint__should_collect_every_endpoint_failure_in_order() {
+        // PccsEndpointError can't derive PartialEq (anyhow::Error: !PartialEq),
+        // so we project to a comparable shape that drops the source field but
+        // keeps the variant + URL — the only properties this test asserts.
+        #[derive(Debug, PartialEq)]
+        enum FailureShape {
+            Fetch(Url),
+            Timeout(Url),
+            ClientConstruction(Url),
+        }
+
+        impl From<&PccsEndpointError> for FailureShape {
+            fn from(err: &PccsEndpointError) -> Self {
+                match err {
+                    PccsEndpointError::Fetch { url, .. } => Self::Fetch(url.clone()),
+                    PccsEndpointError::Timeout { url, .. } => Self::Timeout(url.clone()),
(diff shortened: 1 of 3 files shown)


crates/tee-authority/src/tee_authority.rs
```diff
@@ -28,13 +28,62 @@ pub enum AttestationError {
...
+        }
+    }
+}
+
+/// Returned when every configured PCCS endpoint failed. Owns the full list
+/// of per-endpoint failures in the order they were tried.
+#[derive(Debug, Error)]
+#[error("all {} PCCS endpoints failed:\n{}", failures.len(), format_pccs_failures(failures))]
+pub struct AllPccsEndpointsFailed {
+    pub failures: Vec<PccsEndpointError>,
```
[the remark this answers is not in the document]
pbeza: Done in f40a5985. Switched the field to `NonEmptyVec<PccsEndpointError>` and built it from the loop's `Vec` via `NonEmptyVec::try_from(...).expect(...)` — sound by construction because the loop iterates `pccs_urls: NonEmptyVec<Url>`, so any path reaching the `Err` branch must have pushed at least one failure. Display renders via `failures.as_slice()`.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -28,13 +28,62 @@ pub enum AttestationError {
...
     QuoteDecode(#[source] anyhow::Error),
 
-    #[error("collateral fetch failed: {0:#}")]
-    CollateralFetch(#[source] anyhow::Error),
+    #[error("collateral fetch failed: {0}")]
+    CollateralFetch(#[source] AllPccsEndpointsFailed),
 
     #[error("dstack_endpoint path is not valid UTF-8")]
     InvalidEndpoint,
 }
```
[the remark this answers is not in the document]
pbeza: Not addressing this one — the crate isn't externally consumed under semver. The workspace has internal versioning (every crate moves in lockstep with the workspace version) and the only consumer of `AttestationError` is `crates/node` in this same workspace. There's no published downstream that could pattern-match the variant, so a typed-payload change isn't a breaking-change concern in practice.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -542,23 +617,69 @@ mod tests {
...
+            err.failures
+                .iter()
+                .all(|e| matches!(e, PccsEndpointError::Fetch { .. }))
+        );
+    }
 
-        assert_eq!(err, "https://third.example/ is down");
+    /// `Display` for the aggregate error renders one line per failure, so the
+    /// reviewer's example shape ("all 3 PCCS endpoints failed: …") shows up
+    /// intact in `tracing::error!(?e)` and support-ticket pastes.
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/commit/fd4eef9171f9a04c3c154d1ec984fe021a4eaa82

crates/tee-authority/src/tee_authority.rs
```diff
@@ -28,13 +28,69 @@ pub enum AttestationError {
...
+    #[error("collateral fetch failed for {url}: {source:#}")]
+    Fetch {
+        url: Url,
+        #[source]
+        source: anyhow::Error,
+    },
+}
+
+impl PccsEndpointError {
+    pub fn url(&self) -> &Url {
```
barakeinav1: nit: not used anywhere in the codebase. Drop until there's a caller.

barakeinav1:
AC #7 in #3046 ("downstream consumers compile unchanged") doesn't quite hold — the PR flips `?e` → `%e` in `run.rs` and `remote_attestation.rs`, which is the right call given the new Display impl. Suggest removing AC #7 from the issue.

barakeinav1 (dismissed):

crates/tee-authority/src/tee_authority.rs
pbeza: Done in 6ffbbc02 — dropped the impl block. Confirmed no callers anywhere in the workspace; `PccsEndpointError` itself stays public so future consumers can still pattern-match the variants directly.

pbeza: https://github.com/near/mpc/pull/3048/commits/6ffbbc029dc1439a0aca332ab04a58c50caea5a7

crates/tee-authority/src/tee_authority.rs
```diff
@@ -542,23 +618,88 @@ mod tests {
...
-        assert_eq!(err, "https://third.example/ is down");
+        // When
+        let rendered = err.to_string();
+
+        // Then
+        assert_eq!(
+            rendered,
+            "all 2 PCCS endpoints failed:\n\
+             \x20 - timed out fetching collateral from https://first.example/ after 10s\n\
+             \x20 - collateral fetch failed for https://second.example/: 503 Service Unavailable"
```
gilcu3: nit: would be nice to match against constants

gilcu3 (dismissed):
LGTM

crates/tee-authority/src/tee_authority.rs
pbeza: Done in 0bb25441.

gilcu3 (approved):

barakeinav1 (approved):

PR #3049, opened 2026-04-28
feat(tee-authority): reject PCCS collateral older than 7 days

### LINKED ISSUE #3042: Reject PCCS collateral older than 7 days on the node
### Background

DCAP collateral verification today only enforces Intel's own `nextUpdate` windows (~30 days for TCB Info and QE Identity). Within that window, a compromised or stalled PCCS endpoint can serve stale-but-Intel-signed collateral and hide newly-revoked PCK certs or downgraded TCB levels. The contract's 7-day attestation expiry forces re-verification but does not bound collateral age at each verification.

Research and decision recorded in `near/mpc-private#293` (design doc). 2026-04-27 team meeting concluded: hard-code a node-side check that rejects collateral older than 7 days, not exposed as configuration. Per [netrome's review](https://github.com/near/mpc-private/pull/293#discussion_r3159261432), the same 7-day MAX_AGE applies to all three periodically re-signed pieces that share Intel's 30-day validity window — `tcb_info.issueDate`, `qe_identity.issueDate`, and PCK CRL `thisUpdate`. Long-lived certs and the yearly Root CA CRL are intentionally outside the policy; see the design doc's [Gap 1 table](https://github.com/near/mpc-private/blob/research/dcap-collateral-freshness/resources/design-docs/dcap-collateral-freshness.md#gap-1-max-age-policy-is-entirely-intels) for the validity-window matrix.

### User Story

As a relying party on the MPC network, I want the node to reject collateral whose Intel signature is older than 7 days, so that a compromised or stalled PCCS endpoint can hide revocations or TCB downgrades for at most 7 days rather than up to 30.

### Acceptance Criteria

- [ ] Define `MAX_COLLATERAL_AGE = 7 days` as a hard-coded constant (no config exposure).
- [ ] After `dcap_qvl::verify` returns successfully, the node rejects the attestation if any of the following exceeds `MAX_AGE`:
  - `now - tcb_info.issueDate`
  - `now - qe_identity.issueDate`
  - `now - pck_crl.thisUpdate` (parsed from the X.509 CRL DER bytes)
- [ ] Rejection produces a clear, structured error including the offending field (`tcb_info`, `qe_identity`, or `pck_crl`), its issuance time, and the elapsed delta.
- [ ] **Log a structured `WARN` event on every freshness rejection** with at minimum: which field is stale, the issuance time, the elapsed delta, the configured `MAX_AGE`, and the PCCS URL that served the collateral. This is the operator's primary signal that "my PCCS isn't refreshing" — without it, a stalled-cron PCCS looks identical to other attestation failures.
- [ ] Unit tests cover, for each of the three checked fields: collateral within window accepts; field too old rejects; boundary case (exactly at MAX_AGE). Plus an unparseable-PCK-CRL case so garbage DER doesn't silently pass.

- Design doc: `near/mpc-private#293`
- Research issue: `near/mpc-private#288`
- Related: `#3010` (multi-PCCS fallback, merged 2026-04-28) — with both this and #3010 in place, the URL list acts as a freshness ladder (stale primary → falls through to fallback).
- **Future direction**: contract-side enforcement is the more robust security model but is deferred until on-chain attestation verification lands or the operator set grows beyond the team. See the design doc's Recommendation section for the full rationale.

barakeinav1:
Closes #3042.

Implements the node-side freshness check decided in the 2026-04-27 team meeting (research and rationale in `near/mpc-private#293`).

## What

Adds a hard-coded 7-day max-age check on PCCS collateral, applied inside `fetch_collateral_from` right after `dcap-qvl` returns successfully verified collateral. The check covers all three periodically re-signed pieces that share Intel's 30-day window: `tcb_info.issueDate`, `qe_identity.issueDate`, and PCK CRL `thisUpdate` (parsed from the X.509 CRL via `x509-parser`). Long-lived certs and the yearly Root CA CRL stay outside the policy — see the design doc's [Gap 1 table](https://github.com/near/mpc-private/blob/research/dcap-collateral-freshness/resources/design-docs/dcap-collateral-freshness.md#gap-1-max-age-policy-is-entirely-intels) for the validity-window matrix.

```rust
const MAX_COLLATERAL_AGE: time::Duration = time::Duration::days(7);
```

A rejection here propagates up as a fetch failure, so `try_each_pccs_endpoint` falls through to the next URL in `pccs_urls` — the multi-PCCS list (#3010) naturally acts as a freshness ladder (stale primary → fallback URL).

## Why hard-coded, not configurable

Per the team decision: a malicious operator should not have an obvious bypass knob. The trade-off is that the value can only be changed via a node release.

## Why 7 days

Aligns with the contract's existing `DEFAULT_EXPIRATION_DURATION_SECONDS` so any attestation in the contract is ≤7 days old *and* backed by collateral whose Intel signature is ≤7 days old. Daily-refresh sources (Phala, stock local PCCS) lag at most ~24h, well within the 7× margin.

## Errors

The freshness check is pure-logic and returns a typed `FreshnessError` enum so callers can match on specific failure modes (rather than parsing strings out of `anyhow::Error`):

```rust
pub enum FreshnessError {
    JsonParseFailed { field, source: serde_json::Error },
    IssueDateRfc3339Failed { field, raw, source: time::error::Parse },
    CrlParseFailed { message },
    FutureTimestamp { field, issued_at, skew_seconds, grace_seconds },
    TooStale { field, issued_at, elapsed_seconds, max_age_seconds },
    TimestampOutOfRange { field, issued_at },  // defensive, see below
}
```

The URL is attached one layer up at the `PccsEndpointError::FreshnessCheck { url, source: FreshnessError }` wrap site — the inner check itself doesn't take a URL parameter, since URL is irrelevant to the pure-logic computation.

`TimestampOutOfRange` is a defensive variant that fires only if the `now`-vs-`issued_at` `i64::checked_sub` overflows. Cannot occur for any realistic Intel collateral or system clock, but the typed branch keeps arithmetic total — no panicking subtraction anywhere in the check.

## Logging

Each rejection emits a structured WARN at the `fetch_collateral_from` layer (where the URL is in scope) via `log_freshness_failure`, with explicit fields rather than `Debug`-formatted blobs:

- `TooStale` → `field`, `issued_at`, `elapsed_seconds`, `max_age_seconds`, `pccs_url`
- `FutureTimestamp` → `field`, `issued_at`, `skew_seconds`, `grace_seconds`, `pccs_url`
- Parsing / overflow variants → `error` (Display), `pccs_url`

The `field` value is `"tcb_info"`, `"qe_identity"`, or `"pck_crl"`. JSON-log scrapers can index on the structured keys directly.

## Tests

10 new tests in `tee_authority.rs`:
- `check_collateral_freshness__should_accept_within_window` — 6-day-old collateral accepted
- `check_collateral_freshness__should_reject_when_tcb_info_too_old` — 8-day TCB Info, fresh QE Identity → typed `TooStale { field: "tcb_info", .. }`
- `check_collateral_freshness__should_reject_when_qe_identity_too_old` — symmetric
- `check_collateral_freshness__should_accept_at_exact_boundary` — exactly `MAX_AGE` accepted (comparison is `>`, not `>=`)
- `check_collateral_freshness__should_reject_one_second_past_boundary` — `MAX_AGE + 1` rejected
- `check_collateral_freshness__should_accept_slight_future_within_grace` — 5-min future-dating accepted (clock-skew tolerance)
- `check_collateral_freshness__should_reject_future_beyond_grace` — past 5-min grace → `FutureTimestamp`
- `check_collateral_freshness__should_reject_when_pck_crl_too_old` — 9-day-old PCK CRL with fresh JSON fields → `TooStale { field: "pck_crl", .. }`
- `parse_issue_date_json__should_error_on_invalid_json` — garbage in `tcb_info` → `JsonParseFailed`
- `parse_issue_date_json__should_error_on_invalid_rfc3339` — `{"issueDate": "yesterday"}` → `IssueDateRfc3339Failed` carrying the raw value
- `parse_pck_crl_thisupdate__should_error_on_garbage_bytes` — gibberish DER → `CrlParseFailed`

All assertions use `assert_matches!` against the typed enum variants rather than string-matching on Display output. All 21 tee-authority tests pass; clippy (incl. `--all-features`) + fmt clean.

## Test plan

- [x] `cargo test -p tee-authority --lib`
- [x] `cargo clippy -p tee-authority --all-targets --locked -- -D warnings`
- [x] `cargo fmt --check`
- [ ] E2E with a real PCCS (verifier passes for fresh collateral, fails clearly for forced-stale)

## Related

- Decision: `near/mpc-private#293`
- Research issue: `near/mpc-private#288`
- Implementation issue: #3042
- Multi-PCCS interaction: #3010 (merged) — the freshness ladder behavior is enabled by this combo
- Future direction: contract-side enforcement is the more robust security model, deferred until on-chain attestation verification lands or the operator set grows beyond the team


crates/tee-authority/src/tee_authority.rs
```diff
@@ -180,8 +198,88 @@ impl TeeAuthority {
...
+            issue_date = %issue_date,
+            elapsed_seconds,
+            max_age_seconds = MAX_COLLATERAL_AGE_SECONDS,
+            pccs_url,
+            "PCCS collateral rejected as stale (older than max age)"
+        );
+        anyhow::bail!(
+            "PCCS collateral too stale: {field} issued {issue_date} \
+             ({elapsed_seconds}s ago, max {MAX_COLLATERAL_AGE_SECONDS}s)"
+        );
```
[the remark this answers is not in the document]
barakeinav1: Fixed in 72e0a6a2 — added an explicit future-date check with a 5-minute grace window before the `saturating_sub`. Beyond the grace, rejection emits a structured WARN and an error message naming the field and the skew. Tests cover within-grace acceptance + beyond-grace rejection. Picked 5 minutes to absorb routine clock skew between Intel's PCS and our node without false-positive on normal NTP drift; tighter values risked false rejections, looser ones add no real security value.

crates/tee-authority/Cargo.toml
```diff
@@ -17,7 +17,10 @@ hex = { workspace = true }
 launcher-interface = { workspace = true }
 mpc-attestation = { workspace = true, features = ["dstack-conversions"] }
 near-mpc-bounded-collections = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
 thiserror = { workspace = true }
+time = { workspace = true }
```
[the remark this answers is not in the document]
barakeinav1: Fixed in 72e0a6a2 — added `features = ["parsing", "formatting"]` explicitly on the `time` workspace dep.

barakeinav1:
Fixed in 72e0a6a2 — same change as the Copilot inline thread on this point. Added explicit future-date rejection with a 5-minute grace window (covers routine clock skew between Intel's PCS and our node). Two new tests cover the new behavior.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -180,10 +199,92 @@ impl TeeAuthority {
...
+            time::OffsetDateTime::now_utc(),
+        )?;
+        Ok(collateral)
     }
 }
 
+/// Inner shape of `tcb_info` and `qe_identity` JSON blobs that we care
+/// about — only `issueDate`, the rest is parsed/validated by `dcap-qvl`.
+#[derive(Deserialize)]
+struct CollateralIssueDate {
```
barakeinav1, line 522: Why a custom struct rather than reusing a dcap-qvl type:

- `dcap_qvl::tcb_info::TcbInfo` is `pub` in 0.4.0, but `qe_identity::QeIdentity` is not (the module isn't in dcap-qvl's `pub mod` list). Mixing the two would mean two parsing patterns and two types for symmetric fields.
- `dcap_qvl::verify::verify` parses both internally during signature validation but doesn't return the parsed values via `VerifiedReport`, so we can't reuse them post-call.
- Using `TcbInfo` (~12 fields, nested `Vec<TcbLevel>`, etc.) just to read `issue_date` is wasteful.
- `serde_json::Value` + `.get("issueDate")` works but loses the compile-time field-name check.

The 5-line struct gives us symmetric type-safe parsing for both fields with zero extra overhead. Happy to revisit if dcap-qvl exposes `QeIdentity` or returns parsed types from `verify`.

barakeinav1:
@claude review

barakeinav1:
> **Minor (non-blocking):** The freshness unit tests are `async fn` with `#[tokio::test]` but never `await` — plain `#[test]` would work and avoids spawning a runtime unnecessarily.

Fixed in fc774e17 — converted the seven check_collateral_freshness__* tests from #[tokio::test] async fn to plain #[test] fn. The freshness check is synchronous, the tests never .await, so there's nothing to spawn a runtime for. All 7 still pass.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -239,10 +271,121 @@ impl TeeAuthority {
...
+/// `pccs_urls` acts as a freshness ladder, the stale primary is bypassed
+/// in favor of a fresher fallback (e.g. Intel as the last entry).
+///
+/// The `now` argument is parameterized so tests can drive specific
+/// boundary cases without depending on the wall clock.
+fn check_collateral_freshness(
+    collateral: &Collateral,
+    pccs_url: &str,
+    now: time::OffsetDateTime,
+) -> anyhow::Result<()> {
```
netrome: This is a pure-logic check afaik, we should be able to provide a well defined error enum outlining exactly what failed during this check.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -56,6 +60,13 @@ pub enum PccsEndpointError {
         #[source]
         source: anyhow::Error,
     },
+
+    #[error("collateral failed freshness check at {url}: {source:#}")]
+    FreshnessCheck {
```
netrome, line 70: This error variant is pretty opaque to me. Generally it's good if errors help users understand exactly what went wrong. It would be better with for example an `CollateralTcbInfoTooOld{issue_date, age}`. I don't see why the URL needs to be part of this error, since the URL is not used for the check. Also the `source` field feels weird, since we're packing a type-erased anyhow error from our own logic into this more well-typed error.

The collateral freshness check function should be able to provide an explicit error enum that we either could wrap or expand here.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -239,10 +271,121 @@ impl TeeAuthority {
...
+    pccs_url: &str,
+    now: time::OffsetDateTime,
+) -> anyhow::Result<()> {
+    check_json_issue_date_freshness("tcb_info", &collateral.tcb_info, pccs_url, now)?;
+    check_json_issue_date_freshness("qe_identity", &collateral.qe_identity, pccs_url, now)?;
+    check_pck_crl_freshness(&collateral.pck_crl, pccs_url, now)?;
+    Ok(())
+}
+
+fn check_json_issue_date_freshness(
```
netrome: Nit: I'd consider separating json parsing from this freshness check logic. It looks like this function could take a `CollateralIssueDate` as input, saving us from json parsing at this level.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -239,10 +271,121 @@ impl TeeAuthority {
...
+    check_age_within_window("pck_crl", issued_at, pccs_url, now)
+}
+
+fn check_age_within_window(
+    field: &'static str,
+    issued_at: time::OffsetDateTime,
+    pccs_url: &str,
+    now: time::OffsetDateTime,
+) -> anyhow::Result<()> {
+    let elapsed = now - issued_at;
```
netrome: Blocker: This function could panic if `issued_at` is in the future, as per the docs: https://docs.rs/time/0.3.47/time/struct.OffsetDateTime.html#impl-Sub-for-OffsetDateTime

Please use `checked_sub` instead and return an error if the `issued_at` field is in the future.

netrome (changes requested):
Started looking at this now. There's an unsafe subtraction that could be fixed. Beyond that just a couple of nits. In particular I think the error handling could be improved. We shouldn't use `anyhow` for this kind of functions imo.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -239,10 +271,121 @@ impl TeeAuthority {
...
+
+    if elapsed < -FUTURE_ISSUE_DATE_GRACE {
+        let skew = -elapsed;
+        warn!(
+            field,
+            issued_at = %issued_at,
+            skew_seconds = skew.whole_seconds(),
+            grace_seconds = FUTURE_ISSUE_DATE_GRACE.whole_seconds(),
+            pccs_url,
+            "PCCS collateral rejected: issuance time is in the future beyond grace window"
```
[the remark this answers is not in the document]
barakeinav1: Resolved in f6dcb6c1 — the `warn!` calls inside `check_age_within_window` were removed entirely as part of the typed-error refactor. The check is now pure logic and returns `Err(FreshnessError::...)`; the outer `try_each_pccs_endpoint` loop already logs each per-endpoint failure via `warn!(?err, %url, attempt, ...)` — no duplicate logging, and the typed `FreshnessError` renders cleanly via its `Display` impl. Shorthand-vs-explicit field syntax is therefore moot for these sites.

crates/tee-authority/src/tee_authority.rs
barakeinav1: Done in f6dcb6c1. Replaced `source: anyhow::Error` with `source: FreshnessError` (a new typed enum). Variants:

- `JsonParseFailed { field, source: serde_json::Error }`
- `IssueDateRfc3339Failed { field, raw, source: time::error::Parse }`
- `CrlParseFailed { message }`
- `FutureIssueDate { field, issued_at, skew_seconds, grace_seconds }`
- `TooStale { field, issued_at, elapsed_seconds, max_age_seconds }`
- `TimestampOutOfRange { field, issued_at }` (defensive, see your r3169091381 reply)

URL is no longer threaded into the inner check — it's attached one layer up at the `PccsEndpointError::FreshnessCheck { url, source: FreshnessError }` wrap site, where it's actually known. Operator-facing logs (via `try_each_pccs_endpoint`'s existing `warn!(?err, %url, ...)`) still get the URL, just composed by the caller. Display impl on each variant preserves the operator-facing wording (still says "too stale", "future issuance time", names the field, etc.).

crates/tee-authority/src/tee_authority.rs
barakeinav1: Done in f6dcb6c1 — the check function now returns `Result<(), FreshnessError>` (typed enum, see r3169054626). No more `anyhow::Result` for the pure-logic path.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -239,10 +271,121 @@ impl TeeAuthority {
...
+    }
+
+    if elapsed > MAX_COLLATERAL_AGE {
+        warn!(
+            field,
+            issued_at = %issued_at,
+            elapsed_seconds = elapsed.whole_seconds(),
+            max_age_seconds = MAX_COLLATERAL_AGE.whole_seconds(),
+            pccs_url,
+            "PCCS collateral rejected as stale (older than max age)"
```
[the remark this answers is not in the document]
barakeinav1: Same as r3162308195 — the `warn!` site here was removed in f6dcb6c1 along with the rest of the inline logging in the freshness check. The typed `FreshnessError` flows up to `try_each_pccs_endpoint`, which logs the per-endpoint failure with explicit fields.

crates/tee-authority/src/tee_authority.rs
barakeinav1: Fixed in f6dcb6c1. Switched the arithmetic to `now.unix_timestamp().checked_sub(issued_at.unix_timestamp())` — i64-on-i64 subtraction with an explicit checked path. The `OffsetDateTime - OffsetDateTime` Sub impl is gone. Added a defensive `FreshnessError::TimestampOutOfRange` variant for the (practically unreachable) overflow case so the path is total — no panicking subtraction anywhere in the check.

crates/tee-authority/src/tee_authority.rs
barakeinav1: Done in f6dcb6c1. Split into two functions:

- `parse_issue_date_json(field, raw_json) -> Result<OffsetDateTime, FreshnessError>` — JSON + RFC3339 parsing
- `check_age_within_window(field, issued_at, now) -> Result<(), FreshnessError>` — pure-logic age check, takes a pre-parsed `OffsetDateTime`

`check_collateral_freshness` orchestrates: parse → check → parse → check → parse → check.

barakeinav1:
@claude review

crates/tee-authority/src/tee_authority.rs
```diff
@@ -239,8 +329,118 @@ impl TeeAuthority {
...
         };
 
-        get_with_backoff(fetch, "fetch collateral from PCCS", Some(1)).await
+        let collateral = get_with_backoff(fetch, "fetch collateral from PCCS", Some(1)).await?;
+        check_collateral_freshness(&collateral, time::OffsetDateTime::now_utc()).map_err(
+            |source| PccsEndpointError::FreshnessCheck {
+                url: pccs_url.clone(),
+                source,
+            },
+        )?;
```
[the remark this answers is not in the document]
barakeinav1: Done in 4a7a3c7d — fair point, the AC explicitly calls for structured fields and the outer `try_each_pccs_endpoint` warn was logging the typed error via `Debug`, which puts the structured fields inside a quoted blob.

Added a `log_freshness_failure(pccs_url, &FreshnessError)` helper that fires from `fetch_collateral_from` (where the URL is in scope) and decomposes the typed enum into explicit structured fields:

- `TooStale` → `field`, `issued_at`, `elapsed_seconds`, `max_age_seconds`, `pccs_url`
- `FutureTimestamp` → `field`, `issued_at`, `skew_seconds`, `grace_seconds`, `pccs_url`
- Parsing / overflow variants → `error = %FreshnessError` (Display), `pccs_url`

Inner `check_collateral_freshness` stays pure-logic per netrome's earlier feedback. The WARN sits at the boundary where URL is known.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -239,8 +329,118 @@ impl TeeAuthority {
...
+    })
+}
+
+/// PCK CRL is DER-encoded X.509 v2 CRL. Extract `thisUpdate` from the
+/// TBS structure.
+fn parse_pck_crl_thisupdate(pck_crl_der: &[u8]) -> Result<time::OffsetDateTime, FreshnessError> {
+    let (_, crl) = x509_parser::revocation_list::CertificateRevocationList::from_der(pck_crl_der)
+        .map_err(|e| FreshnessError::CrlParseFailed {
+        message: e.to_string(),
+    })?;
```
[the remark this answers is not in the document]
barakeinav1: Fixed in 4a7a3c7d — `cargo fmt` cleaned up the misindented `map_err` block.

barakeinav1:
@claude review

crates/tee-authority/src/tee_authority.rs
```diff
@@ -239,8 +382,121 @@ impl TeeAuthority {
...
+    let max_age_seconds = MAX_COLLATERAL_AGE.whole_seconds();
+
+    if elapsed_seconds < -grace_seconds {
+        let skew_seconds = -elapsed_seconds;
+        return Err(FreshnessError::FutureTimestamp {
+            field,
+            issued_at,
+            skew_seconds,
+            grace_seconds,
+        });
```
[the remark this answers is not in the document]
barakeinav1: Fixed in 61607c04 — `elapsed_seconds.checked_neg().ok_or(FreshnessError::TimestampOutOfRange { .. })?`. Practically unreachable (would require `elapsed_seconds == i64::MIN`, which itself requires the outer `checked_sub` to have already produced `i64::MIN`), but matches the "no panicking arithmetic" posture the rest of this function already has.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -239,8 +382,121 @@ impl TeeAuthority {
...
-        get_with_backoff(fetch, "fetch collateral from PCCS", Some(1)).await
+        let collateral = get_with_backoff(fetch, "fetch collateral from PCCS", Some(1)).await?;
+        if let Err(source) =
+            check_collateral_freshness(&collateral, time::OffsetDateTime::now_utc())
+        {
+            log_freshness_failure(pccs_url, &source);
+            return Err(PccsEndpointError::FreshnessCheck {
+                url: pccs_url.clone(),
+                source,
+            });
```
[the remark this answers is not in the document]
barakeinav1: Fixed in 61607c04. Caught it — `log_freshness_failure` was emitting a structured warn from `fetch_collateral_from` and then `try_each_pccs_endpoint`'s per-attempt `warn!(?err, %url, ...)` was firing as well, producing two log lines per freshness failure. The per-attempt warn now skips `PccsEndpointError::FreshnessCheck` (already structure-logged); other variants (timeout, fetch, client construction) keep their per-attempt warn since they have no inner structured log.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -239,8 +382,121 @@ impl TeeAuthority {
...
+    })
+}
+
+/// PCK CRL is DER-encoded X.509 v2 CRL. Extract `thisUpdate` from the
+/// TBS structure.
+fn parse_pck_crl_thisupdate(pck_crl_der: &[u8]) -> Result<time::OffsetDateTime, FreshnessError> {
+    let (_, crl) = x509_parser::revocation_list::CertificateRevocationList::from_der(pck_crl_der)
+        .map_err(|e| FreshnessError::CrlParseFailed {
+        message: e.to_string(),
+    })?;
```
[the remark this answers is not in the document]
barakeinav1: Verified — this is what `cargo fmt` produces. I ran `cargo fmt -p tee-authority` and `--check` is clean. The visual oddity (struct literal body indented at the closure-body column rather than 4 spaces deeper) is rustfmt's chosen style for struct-literals-inside-closures; my earlier "manual" indent fix actually got reverted by rustfmt back to this shape. No further action.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -274,14 +532,19 @@ where
...
+                    warn!(
+                        ?err,
+                        %url,
+                        attempt,
+                        total_endpoints,
+                        "failed to fetch collateral from PCCS; {}",
+                        if is_last_endpoint { "no more endpoints remain" } else { "trying next endpoint" }
+                    );
+                }
                 failures.push(err);
```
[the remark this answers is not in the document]
barakeinav1: False alarm — this code compiles and tests/clippy pass against the latest commit (`61607c04`).

`matches!(err, PccsEndpointError::FreshnessCheck { .. })` doesn't move `err`. The `{ .. }` pattern uses `..` which discards all fields without binding any of them, so no field is moved out. `err` remains owned and usable in the subsequent `warn!(?err, ...)` and `failures.push(err)`.

If the pattern were `PccsEndpointError::FreshnessCheck { url, source }` (binding fields by value), you'd be right — that would move them out. But `{ .. }` is the discard-everything form. Verified locally: `cargo build -p tee-authority --tests` ✓, `cargo clippy --all-targets --locked -- -D warnings` ✓.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -35,9 +38,10 @@ pub enum AttestationError {
     InvalidEndpoint,
 }
 
-/// One PCCS endpoint's failure. Carries the URL it was tried against and
-/// the underlying cause. The cause stays as `anyhow::Error` because the
-/// upstream `dcap_qvl::CollateralClient` itself returns `anyhow::Result`.
+/// One PCCS endpoint's failure. The URL is attached at this layer
+/// because the inner per-step errors (collateral fetch, freshness
+/// check) don't all need it — we add it once here for operator-facing
+/// log/error messages.
```
pbeza: Nit: I think we can make this comment a bit clearer:

```suggestion
/// Wraps a single PCCS endpoint attempt's failure with the URL of the
/// endpoint that failed. Each variant carries a `url` field so that
/// per-endpoint log/error output can name which endpoint it's about
/// without the inner error types (e.g. `FreshnessError`) needing to
/// know about URLs themselves.
```

crates/tee-authority/src/tee_authority.rs
```diff
@@ -56,6 +60,123 @@ pub enum PccsEndpointError {
...
+}
+
+/// Emit a structured WARN for a failed PCCS freshness check, with all
+/// the fields the issue's acceptance criteria call for (field, issued
+/// timestamp, the relevant delta, the configured limit, and the
+/// `pccs_url`). Fires at the [`fetch_collateral_from`] layer where the
+/// URL is known; the inner check is intentionally pure-logic and
+/// returns a typed [`FreshnessError`] which this function decomposes
+/// into separate structured fields rather than logging via `Debug`.
+fn log_freshness_failure(pccs_url: &Url, error: &FreshnessError) {
```
pbeza, line 150: I don’t think we need a separate function for that. We can just call:
```rust
warn!(pccs_url = %pccs_url, error = %source, "PCCS collateral failed freshness check");
```
instead.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -87,6 +208,23 @@ const MAX_BACKOFF_DURATION: Duration = Duration::from_secs(60);
 /// Per-request timeout for fetching collateral from PCCS.
 const PCCS_REQUEST_TIMEOUT: Duration = Duration::from_secs(10);
 
+/// Maximum age accepted for PCCS collateral. Hard-coded at 7 days per
+/// the team decision in `near/mpc-private#293`; stricter than Intel's
```
pbeza: We don’t typically reference issue numbers in docstrings. Instead, we should briefly explain the reason for this change inline here.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -274,14 +532,19 @@ where
...
-                    %url,
-                    attempt,
-                    total_endpoints,
-                    "failed to fetch collateral from PCCS; {}",
-                    if is_last_endpoint { "no more endpoints remain" } else { "trying next endpoint" }
-                );
+                // Freshness rejections already emit a structured WARN with
+                // explicit fields via `log_freshness_failure`; a generic
+                // `?err` warn here would just duplicate it.
+                if !matches!(err, PccsEndpointError::FreshnessCheck { .. }) {
```
pbeza, line 659: Nit: It's a bit hacky/over-engineered. Is there a way to simplify it?

crates/tee-authority/src/tee_authority.rs
```diff
@@ -709,6 +977,231 @@ mod tests {
...
+
+    /// Format an `OffsetDateTime` as RFC3339 (UTC) the way Intel PCS would
+    /// emit it, so test fixtures stay close to real-world inputs.
+    fn rfc3339(t: time::OffsetDateTime) -> String {
+        t.format(&Rfc3339)
+            .expect("RFC3339 format always succeeds for a valid datetime")
+    }
+
+    /// Fixed `now` for boundary tests so they don't depend on the wall
+    /// clock. Chosen ~43 min after the fixture PCK CRL's `thisUpdate`
```
pbeza: Did you implicitly hardcode the 43 minutes in the date (`"2026-03-30T12:00:00Z"`) here? If so, let’s make it explicit by including the `43` somewhere.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -239,10 +382,125 @@ impl TeeAuthority {
...
+
+/// PCK CRL is DER-encoded X.509 v2 CRL. Extract `thisUpdate` from the
+/// TBS structure.
+fn parse_pck_crl_thisupdate(pck_crl_der: &[u8]) -> Result<time::OffsetDateTime, FreshnessError> {
+    let (_, crl) = x509_parser::revocation_list::CertificateRevocationList::from_der(pck_crl_der)
+        .map_err(|e| FreshnessError::CrlParseFailed {
+        message: e.to_string(),
+    })?;
+    Ok(crl.last_update().to_datetime())
+}
```
pbeza, line 582: I’m a bit concerned about how many custom, free-floating parsing functions we have here. Maybe we could ask Phala to expose some structs so we can hide some of this complexity?

Ideally these validity checks would be closer to a one-liner — the amount of boilerplate here feels a bit heavy for what we’re doing.

pbeza (commented):
If Intel only re-signs on content changes, then during quiet periods (no new TCB advisory, revocation, or QE binary), the `issueDate` could legitimately remain unchanged for weeks—and an honest operator would be rejected. Are we sure it’s always re-signed every 7 days? If not, the collateral would be rejected.

crates/tee-authority/src/tee_authority.rs
netrome: I both disagree and agree. This references a private issue which doesn't make sense in a public code base, since anyone should be able to read and understand the code and our decisions.

However, I am quite okay with referencing issues or PRs in the code still. It is a tight coupling between the issue tracking system and our code, so it's not a perfect model, but in practice I'm quite used to it and it tends to work well as long as you don't migrate frequently between issue tracking systems.

crates/tee-authority/src/tee_authority.rs
netrome: Hmm this suggestion means we use the display representation of the `source` though. Spontaneously it feels nice to retain more of the structure like we get with this function.

I dived into this rabbit hole a bit and seems like there's no good standard way to express how a type should be expanded into some nested structure for logging with `tracing`. They have experimental support for [`valuable`](https://docs.rs/tracing/latest/tracing/field/index.html#using-valuable) though. Though I wouldn't want to go that route.

So with that in mind, I think the helper function is pretty okay.

netrome (changes requested):
Thanks for fixing the error handling. I second most of Patrick's nits and comments. The reference to our private repo is the only one I see as a blocker though.

barakeinav1:
> If Intel only re-signs on content changes, then during quiet periods (no new TCB advisory, revocation, or QE binary), the `issueDate` could legitimately remain unchanged for weeks—and an honest operator would be rejected. Are we sure it’s always re-signed every 7 days? If not, the collateral would be rejected.

I checked, and even if there are no updated at all intel PCCS will re-sign that same collateral overy 2 hours

crates/tee-authority/src/tee_authority.rs
barakeinav1: 1. let me add explanation this is the doc string instead of referencing MPC-private.
2. I'll update are design doc (in our MPC repor with those detail as well, but as a separate PR), my guess there are many things that should be updated.

crates/tee-authority/src/tee_authority.rs
```diff
@@ -337,47 +442,54 @@ impl TeeAuthority {
...
         })
         .await
     }
 
-    /// Fetches attestation collateral from a single PCCS endpoint, with the
-    /// usual per-request timeout and a single retry via exponential backoff.
+    /// Fetches attestation collateral from a single PCCS endpoint, with
+    /// the usual per-request timeout and a single retry via exponential
+    /// backoff. Honors the endpoint's per-URL TLS trust override: with
+    /// `tls = None` the client uses default reqwest+rustls trust roots
```
barakeinav1: "TLS = none" is a bit misleading.
I think the field was changed to TLS override.
update comment

barakeinav1: Fixed in 63a4fe97. Reworded around the TOML shape so the operator's mental model lines up:

> Honors the endpoint's per-URL TLS trust override (the `tls = { override = ... }` part of each `[[pccs_endpoints]]` entry): when the operator omits `tls`, the client uses default reqwest+rustls trust roots (bundled Mozilla webpki-roots); with `tls.override` set, the client reflects it (`add_root_certificate` for `ca_cert_pem`, `danger_accept_invalid_certs` for `insecure`).

crates/tee-authority/src/tee_authority.rs
barakeinav1: Done in 636ca030 — replaced the `"2026-03-30T12:00:00Z"` literal with `parse_pck_crl_thisupdate(&fixture_pck_crl()) + TEST_NOW_OFFSET_FROM_FIXTURE_CRL`, where the new `TEST_NOW_OFFSET_FROM_FIXTURE_CRL = time::Duration::minutes(43)`. The 43-min number is now visible in code, and `test_now` no longer encodes the fixture CRL's `thisUpdate` silently in a string literal.

crates/tee-authority/src/tee_authority.rs
barakeinav1: ok,  keeping `log_freshness_failure`.

crates/tee-authority/src/tee_authority.rs
barakeinav1: Done in 636ca030. Dropped the `near/mpc-private#293` link and pulled the rationale inline — the doc now explains *why* 7 days (hard-coded so a malicious operator can't relax it; 7 days is stricter than Intel's 30-day window, more permissive than ~daily refresh schedules, and aligns with the contract's `DEFAULT_EXPIRATION_DURATION_SECONDS` so contract-accepted attestations are also ≤7 days collateral-fresh).

Also planning a follow-up to update the design doc itself (in the public repo) — that's separate from this PR.

crates/tee-authority/src/tee_authority.rs
barakeinav1: Out of scope for this PR — agree the right long-term fix is upstream Phala exposing typed structs (so we don't have to re-parse JSON / X.509 from raw bytes) but that's an upstream-coordination workstream. For this PR, the three parse helpers (~30 LOC across `parse_issue_date_json`, `parse_pck_crl_thisupdate`, and the CollateralIssueDate shape) are the smallest local landing for the AC.

crates/tee-authority/src/tee_authority.rs
barakeinav1: Agreed on the private-issue concern — addressed in 636ca030 by dropping the `near/mpc-private#293` reference and inlining the rationale. Public-repo issue/PR refs are still in the file (e.g. `PR #3026` in the `PCCS_REQUEST_TIMEOUT` doc) and stay.

crates/tee-authority/src/tee_authority.rs
barakeinav1: Going to push back on this one. The `if !matches!(err, PccsEndpointError::FreshnessCheck { .. })` block exists because freshness rejections already emit a structured WARN via `log_freshness_failure` at the inner layer (per the issue's AC, and per Copilot r3169492961 which explicitly flagged double-logging when both fire). Removing this guard reintroduces double-logging — every freshness failure produces two WARN lines: a structured one with explicit fields, and a generic `?err` blob.

The alternatives I considered:
- Drop `log_freshness_failure` and rely on the per-attempt `?err` blob alone — loses the structured fields scrapers want (this is r3169868296's path, where netrome and I landed on keeping the helper).
- Move all per-failure logging into the per-fetch path — symmetric but the per-attempt warn for non-freshness errors (timeout, fetch, client construction) lives most naturally at the loop layer where attempt/total counters are in scope.

So the matches! filter is the smallest amount of glue that makes both layers' logging coexist without duplication. Open to a cleaner phrasing if you have one in mind, but the dispatch logic itself is doing real work.

crates/tee-authority/src/tee_authority.rs
barakeinav1: Applied your suggestion verbatim in 636ca030 — agreed it reads cleaner.

crates/tee-authority/src/tee_authority.rs
barakeinav1: Follow-up tracked in #3094 (broader scope: refresh PCCS-collateral design docs across #3010 / #3026 / #3049 in one place, in the public repo).

barakeinav1:
@claude review

crates/tee-authority/src/tee_authority.rs
```diff
@@ -345,8 +502,123 @@ impl TeeAuthority {
...
+
+/// Reject collateral whose Intel-issued timestamp is older than
+/// [`MAX_COLLATERAL_AGE`]. Checks all three periodically re-signed pieces:
+/// `tcb_info.issueDate`, `qe_identity.issueDate`, and PCK CRL `thisUpdate`.
+/// All three share Intel's 30-day validity window, so a single MAX_AGE
+/// covers them.
+///
+/// Combined with the multi-PCCS support from [`try_each_pccs_endpoint`], a
+/// rejection here causes the helper to fall through to the next URL — i.e.
+/// `pccs_urls` acts as a freshness ladder, the stale primary is bypassed
```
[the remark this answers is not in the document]
barakeinav1: Done in d0d3e375 — `pccs_urls` → `pccs_endpoints` (renamed in #3026).

crates/tee-authority/src/tee_authority.rs
```diff
@@ -345,10 +495,125 @@ impl TeeAuthority {
...
+    })
+}
+
+/// PCK CRL is DER-encoded X.509 v2 CRL. Extract `thisUpdate` from the
+/// TBS structure.
+fn parse_pck_crl_thisupdate(pck_crl_der: &[u8]) -> Result<time::OffsetDateTime, FreshnessError> {
+    let (_, crl) = x509_parser::revocation_list::CertificateRevocationList::from_der(pck_crl_der)
+        .map_err(|e| FreshnessError::CrlParseFailed {
+        message: e.to_string(),
+    })?;
```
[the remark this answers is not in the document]
barakeinav1: False positive — verified for the third time on this PR. Running `cargo fmt -p tee-authority -- --check` against $SHA passes cleanly. The visual oddity (struct-literal body sitting at the closure-body column rather than 4 spaces deeper) is rustfmt's default for `map_err(|e| StructLiteral { ... })` patterns. My earlier "manual" indent fix was reverted by `cargo fmt` back to this exact shape. CI's `cargo fmt --check` step is also green on this PR. No change needed.

barakeinav1:
> One minor item already flagged by Copilot in an existing thread but still present in the diff:
> 
> * **Doc comment refers to `pccs_urls` instead of `pccs_endpoints`** — `check_collateral_freshness` doc comment (line ~538 in the diff) says _"`pccs_urls` acts as a freshness ladder"_ but the actual config/field name is `pccs_endpoints`. Should be updated for consistency with the codebase.

fix in d0d3e375

crates/tee-authority/src/tee_authority.rs
```diff
@@ -345,8 +502,123 @@ impl TeeAuthority {
...
```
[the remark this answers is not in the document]
barakeinav1: not a real issue. accepted by crago fmt

crates/tee-authority/src/tee_authority.rs
barakeinav1: Filed a follow-up tracking issue at near/mpc-private#316 — *Ask Phala to expose typed structs for PCCS collateral parsing*. Marking it as **low priority** since the current ~30 LOC of parse helpers work and the upstream coordination loop with Phala is the rate-limiting step. We can delete the helpers cleanly once Phala exposes typed accessors, but no urgency from this PR's side.

netrome (approved):

crates/tee-authority/src/tee_authority.rs
```diff
@@ -827,6 +1109,240 @@ mod tests {
...
+        let collateral = collateral_with_issue_dates(&issued, &issued);
+
+        let err = check_collateral_freshness(&collateral, test_now()).unwrap_err();
+
+        assert_matches!(err, FreshnessError::FutureTimestamp { .. });
+    }
+
+    /// PCK CRL `thisUpdate` past `MAX_COLLATERAL_AGE` is rejected even
+    /// when the JSON-shaped fields are fresh — covers the case netrome
+    /// raised on `near/mpc-private#293`: PCK CRL has the same 30-day
```
netrome: Oh we still have a reference to `mpc-private` here

netrome (changes requested):

crates/tee-authority/src/tee_authority.rs
barakeinav1: Good catch — the earlier `mpc-private#293` cleanup in 636ca030 covered the `MAX_COLLATERAL_AGE` doc-comment but missed this test docstring. Removed in 442f97d1; reworded to keep the rationale inline without the contributor-name + private-link framing.

netrome (approved):

pbeza (approved):

PR #3050, opened 2026-04-28
refactor: Separate keyshare related contract-internal types/logic from node-facing interface types

### LINKED ISSUE #2881: Separate keyshare related contract-internal types/logic from node-facing interface types
### Background

Types like `Keyset`, `Keyshares`, `Epoch`, `KeyEventId` currently live in `mpc-contract` but are also used by `mpc-node`. They carry both data and logic, and the logic is entangled on both sides.

This blocks further progress on the contract vs. node separation

Discussion summary:
- Node should not depend on contract-internal types; they are internal to the contract.
- Node already has an internal contract representation; it only needs to know how to deserialize `state()` and convert to that internal representation.
- Proposed shape: declare the type in `contract-interface` (plain data), keep the logic as a trait + impl inside `mpc-contract`. E.g.:
  - `contract-interface`: `pub struct MyThingamajig { ... }`
  - `mpc-contract`: `trait MyLogic { ... }` + `impl MyLogic for MyThingamajig`
- Some types may instead fit better in `mpc-primitives`.

### Acceptance Criteria

- [ ] Move data definitions for `Keyset`, `Keyshares`, `Epoch`, `KeyEventId` out of `mpc-contract` to `contract-interface` (or `mpc-primitives` where more appropriate).
- [ ] Keep contract-internal logic in `mpc-contract` via traits/impls on those types.
- [ ] `mpc-node` no longer depends on `mpc-contract` for these types or their logic.

gilcu3:
Closes #2881


crates/node/Cargo.toml
```diff
@@ -34,7 +34,6 @@ k256 = { workspace = true }
 launcher-interface = { workspace = true }
 lru = { workspace = true }
 mpc-attestation = { workspace = true }
-mpc-contract = { workspace = true, features = ["compat"] }
```
gilcu3, line 37: node no longer depends on the contract, except for tests

crates/backup-cli/Cargo.toml
```diff
@@ -8,7 +8,7 @@ edition = { workspace = true }
 anyhow = { workspace = true }
 clap = { workspace = true }
 ed25519-dalek = { workspace = true }
-mpc-contract = { workspace = true, features = ["dev-utils"] }
+mpc-contract = { workspace = true, features = ["compat", "dev-utils"] }
```
gilcu3, line 11: we still need to remove this, #3051

DSharifi (approved):
thank you!

kevindeforth (approved):
Thank you!

PR #3056, opened 2026-04-28
fix: retry conclude_node_migration until contract reflects completion, not until tx is applied

anodar:

`retry_conclude_onboarding` previously treated tx_sender::send returning Ok as success, but that only signals the tx was applied — NEAR rolls back state when a contract method returns Err, so a contract-side rejection (e.g. InvalidTeeRemoteAttestation racing the new node's attestation submission) silently left the migration record in place and the onboarding node waiting forever on a participant list change that would never happen. 
Now we verify the migration actually completed by waiting for active_migration to flip false on the local MigrationInfo watch — set when the contract removes our migration record on success — and retry the conclude tx if it doesn't.

Verified by running all tests 100 times in https://github.com/near/mpc/pull/3024, specifically this run: https://github.com/near/mpc/actions/runs/25051806158/job/73381273673

### DIFF
--- a/crates/node/src/migration_service/onboarding.rs
+++ b/crates/node/src/migration_service/onboarding.rs
@@ -37,7 +37,7 @@ pub(crate) async fn onboard(
     tracing::info!(?my_near_account_id, "starting onboarding");
     let (cancel_monitoring_task, mut onboarding_job_receiver) = start_onboarding_monitoring_task(
         contract_state_receiver,
-        my_migration_info_receiver,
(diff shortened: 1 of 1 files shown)


kevindeforth (dismissed):
Thanks a lot! I remember skipping this part when implementing the migration logic.

_edit: would be curious to know why the transactions fail though. I don't think this was an issue in the pytests._

netrome (approved):
Nice!

kevindeforth (approved):

PR #3057, opened 2026-04-28
refactor: move block height check out of `RecentBlocksTracker`, into `PendingRequests`

### LINKED ISSUE #3053: move block height check out of `RecentBlocksTracker`
### Background

The `RecentBlocksTracker` struct currently keeps track of the indexer height seen by other participants.
We aim to move `RecentBlocksTracker` into the chain gateway crate, at which point, the notion of _external indexer views_ does not make too much sense anymore. The `RecentBlocksTracker` will become a local view tracker only and the height comparison between the block and the network view can be handled by the `PendingRequests` struct.

### User Story

As a dev, I like clear separation of concerns.

### Acceptance Criteria

- remove `notify_maximum_height` from `RecentBlocksTracker`
https://github.com/near/mpc/blob/c717dd9f495ed58f7fa6d985675a9739fd31df4e/crates/node/src/requests/recent_blocks_tracker.rs#L301

**_Acceptance Criteria out of scope and moved to #3061:_** 
- remove the `block_height` argument from `classify_block` and inline the block height comparison with the maximum indexer height to the caller.
https://github.com/near/mpc/blob/c717dd9f495ed58f7fa6d985675a9739fd31df4e/crates/node/src/requests/recent_blocks_tracker.rs#L422

kevindeforth:
resolves #3053

### DIFF
--- a/crates/node/src/requests/queue.rs
+++ b/crates/node/src/requests/queue.rs
@@ -11,6 +11,7 @@ use near_time::Duration;
 use sha3::Digest;
 use std::collections::{HashMap, HashSet};
 use std::fmt::Debug;
+use std::ops::Add;
 use std::sync::{Arc, Mutex, Weak};
 use time::ext::InstantExt as _;
 
@@ -428,8 +429,6 @@ impl<RequestType: Request + Clone, ChainRespondArgsType: ChainRespondArgs>
 
         let (eligible_leaders, maximum_height) = self.eligible_leaders_and_maximum_height();
         tracing::debug!(target: "request", "Eligible leaders: {:?}", eligible_leaders);
-        self.recent_blocks
-            .notify_maximum_height_available(maximum_height);
 
         let mut result = Vec::new();
 
@@ -445,6 +444,26 @@ impl<RequestType: Request + Clone, ChainRespondArgsType: ChainRespondArgs>
                 tracing::debug!(target: "request", "Skipping request {:?} from block {} because there's already an active attempt", request.request.get_id(), request.block_height);
                 continue;
             }
+            // check it against the network height
+            if maximum_height
+                .saturating_sub(REQUEST_EXPIRATION_BLOCKS)
+                .add(1)
+                > request.block_height
+            {
+                // this request expired for at least one participant in the network. We can already
+                // remove it from the queue.
+                tracing::debug!(target: "request", "Discarding {} request {:?} from block {}", RequestType::get_type(), request.request.get_id(), request.block_height);
+                // Increment metric for timeout (only for signature requests)
+                if matches!(RequestType::get_type(), types::RequestType::Signature) {
+                    metrics::MPC_CLUSTER_FAILED_SIGNATURES_COUNT
+                        .with_label_values(&["timeout"])
+                        .inc();
+                }
+                // This request is definitely not useful anymore, so discard it.
+                requests_to_remove.push(*id);
+                continue;
+            }
+
             match self
                 .recent_blocks
                 .classify_block(request.block_hash, request.block_height)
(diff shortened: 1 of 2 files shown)


crates/node/src/requests/queue.rs
```diff
@@ -445,6 +444,26 @@ impl<RequestType: Request + Clone, ChainRespondArgsType: ChainRespondArgs>
                 tracing::debug!(target: "request", "Skipping request {:?} from block {} because there's already an active attempt", request.request.get_id(), request.block_height);
                 continue;
             }
+            // check it against the network height
+            if maximum_height
+                .saturating_sub(REQUEST_EXPIRATION_BLOCKS)
+                .add(1)
+                > request.block_height
+            {
```
DSharifi, line 462: I'm all for safety, but isn't `maximum_height` an u64?

crates/node/src/requests/queue.rs
```diff
@@ -445,6 +444,26 @@ impl<RequestType: Request + Clone, ChainRespondArgsType: ChainRespondArgs>
...
+                // this request expired for at least one participant in the network. We can already
+                // remove it from the queue.
+                tracing::debug!(target: "request", "Discarding {} request {:?} from block {}", RequestType::get_type(), request.request.get_id(), request.block_height);
```
DSharifi: can you use structured logging here?

DSharifi (dismissed):

crates/node/src/requests/queue.rs
```diff
@@ -445,6 +444,26 @@ impl<RequestType: Request + Clone, ChainRespondArgsType: ChainRespondArgs>
...
+                // Increment metric for timeout (only for signature requests)
+                if matches!(RequestType::get_type(), types::RequestType::Signature) {
+                    metrics::MPC_CLUSTER_FAILED_SIGNATURES_COUNT
+                        .with_label_values(&["timeout"])
+                        .inc();
+                }
+                // This request is definitely not useful anymore, so discard it.
+                requests_to_remove.push(*id);
+                continue;
+            }
```
anodar, line 481: This seems almost duplicate of code below:

```
         CheckBlockResult::NotIncluded | CheckBlockResult::OlderThanRecentWindow => {
                    tracing::debug!(target: "request", "Discarding {} request {:?} from block {}", RequestType::get_type(), request.request.get_id(), request.block_height);
                    // Increment metric for timeout (only for signature requests)
                    if matches!(RequestType::get_type(), types::RequestType::Signature) {
                        metrics::MPC_CLUSTER_FAILED_SIGNATURES_COUNT
                            .with_label_values(&["timeout"])
                            .inc();
                    }

                    // This request is definitely not useful anymore, so discard it.
                    requests_to_remove.push(*id);
                }
            }
        }
```

Do we really need this ? Or did you mean to drop this below and get rid of block height argument?

anodar (commented):
In the issue you say we should drop block height argument from `classify_block` and instead do checks at callsites, but this PR doesn't seem to be doing it.

crates/node/src/requests/queue.rs
kevindeforth: fixed for all the logs: https://github.com/near/mpc/pull/3057/commits/5d259dfa116b37b1c48c25ae4a77cc460eef7f97

crates/node/src/requests/queue.rs
kevindeforth: I think this is required for localnet. Maximum height may be less than `REQUEST_EXPIRATION_BLOCKS` in that case.

crates/node/src/requests/queue.rs
kevindeforth: We still need this.
The tracker can still respond with a `OlderThanRecentWindow`, in case both of the following are true:
1. it's recency window is greater than what the queue uses;
2. the block in question is `OlderThanRecentWindow`, but has not yet been cleaned in the tracker.

Point 1. is not the case with the current implementation, but having independent recent block windows is not something that would be inherently wrong. So we shouldn't assume this to always hold.
Point 2 only happens if the block in question is the last final block.

kevindeforth:
> In the issue you say we should drop block height argument from `classify_block` and instead do checks at callsites, but this PR doesn't seem to be doing it.

Apologies, I made an ill-defined issue. Will adjust the acceptance criteria.
The main goal of the issue is to not have the network block height leak into the `RecentBlocksTracker`. This is achieved by moving this check here, instead of having `RecentBlocksTracker` concern itself with the network view:
https://github.com/near/mpc/blob/5d259dfa116b37b1c48c25ae4a77cc460eef7f97/crates/node/src/requests/queue.rs#L458

Removing the `block_height` argument from the call to `classify_blocks` is a secondary goal, but one that requires more care, as it also requires us to properly handle`Unknown` bock results. I made a follow-up issue #3061 for that.

anodar (approved):
Thank you!

DSharifi (approved):

kevindeforth:
Failed CI check is due to handshake timeout. Nothing to do with this PR, probably due to slow runner.
https://nearone.slack.com/archives/C07UW93JVQ8/p1777453498062959?thread_ts=1777452705.114429&cid=C07UW93JVQ8

PR #3059, opened 2026-04-28
feat: add jemalloc pprof endpoint on the node

### LINKED ISSUE #3020: add jemalloc pprof endpoint on the node
### Background

We should add this, https://github.com/polarsignals/rust-jemalloc-pprof/, tool for profiling memory allocations as an endpoint similar to the rust-pprof endpoint we have for profiling CPU time.

### User Story

_No response_

### Acceptance Criteria

node serves rust-jemalloc-pprof flamegraphs on a debug endpoint

_No response_

DSharifi:
closes #3020


DSharifi:
Example from testing with localnet:

jemalloc_flamegraph

DSharifi:
Works with pprof graph view as well now to look at diffs between snapshots, so we can see where memory grows/shrinks over time

image

DSharifi:
@claude review

DSharifi:
@claude review

crates/node/src/profiler/jemalloc.rs
```diff
@@ -0,0 +1,173 @@
+use axum::{
+    http::{header, StatusCode},
+    response::{IntoResponse, Response},
+};
+use std::{ffi::c_void, fs::File, io::BufReader};
+
+const CONTENT_TYPE_SVG: &str = "image/svg+xml";
```
anodar: nit: this constant already exists in web_server.rs, consider deduplicating.

crates/node/src/profiler/jemalloc.rs
```diff
@@ -0,0 +1,173 @@
...
+
+// pprof sample type: bytes currently allocated and not yet freed (live heap).
+// Tuple is (name, unit) per the pprof proto spec.
+const PPROF_SAMPLE_TYPE: (&str, &str) = ("inuse_space", "bytes");
+// pprof period type: the unit between successive samples, matching the sample
+// type for a heap profile.
+const PPROF_PERIOD_TYPE: (&str, &str) = ("space", "bytes");
+
+const MSG_PROFILING_UNAVAILABLE: &str =
+    "jemalloc heap profiling is not available; ensure the binary was built with the `profiling` feature \
```
anodar: nit: This message seems misleading, (unless I'm missing something) in case of Linux it's hardcoded to be started with that already. Consider rewording saying it's not available besides linux.

anodar (dismissed):

crates/node/src/profiler/jemalloc.rs
DSharifi: This code path is only active on linux. The issue is that for whatever reason `jemalloc_pprof::PROF_CTL` can evaluate to None, in that case meaning the profiling is not activated. Agreed though, the current error message is not accurate.

anodar (approved):

crates/node/src/profiler/jemalloc.rs
```diff
@@ -0,0 +1,173 @@
...
+
+fn render_flamegraph_svg(dump_file: File) -> anyhow::Result<Vec<u8>> {
+    let stack_profile = pprof_util::parse_jeheap(BufReader::new(dump_file), None)?;
+
+    let mut collapsed_lines: Vec<String> = Vec::with_capacity(stack_profile.stacks.len());
+    for (stack, _annotation) in &stack_profile.stacks {
+        // Heap dumps weight each sample by allocated bytes (a float). Inferno
+        // expects a non-negative integer count, so round up to ensure samples
+        // are never silently dropped.
+        let weight = stack.weight.max(0.0).ceil() as u64;
```
kevindeforth, line 138: Q: why `ceil` and not `round`?

kevindeforth (approved):
Nice! Thanks!

crates/node/src/profiler/jemalloc.rs
DSharifi: Because we want to round up. 0.3 should for example become 1, not 0.

PR #3069, opened 2026-04-29
refactor: recent blocks tracker relies on internal block height

### LINKED ISSUE #3061: mpc node should not act on signatures in `Unknown` blocks
### Background

Currently, the node acts on signatures that appear in a block that the `RecentBlocksTracker` has not yet seen.

https://github.com/near/mpc/blob/79cc28751e5155a451a2cb7b3513674961d115f1/crates/node/src/requests/queue.rs#L450-L490

This behavior does not make much sense in the current code logic. We first add blocks to the `RecentBlocksTracker` and only then add the requests to the queue.
That means that an `Unknown` block Result is only possible if the `RecentBlocksTracker` has removed the block from its state (currently, not possible).

https://github.com/near/mpc/blob/79cc28751e5155a451a2cb7b3513674961d115f1/crates/node/src/requests/queue.rs#L287-L301

https://github.com/near/mpc/blob/79cc28751e5155a451a2cb7b3513674961d115f1/crates/node/src/requests/queue.rs#L376-L387

### User Story

As a dev, I like correct code.

### Acceptance Criteria

Do not compute signature responses from blocks `Unknown` to the `RecentBlocksTracker`.

Remove the `block_height` argument from the call to `classify_blocks` and treat `Unknown` the same as `OlderThanRecentWindow`

Follow-up to https://github.com/near/mpc/pull/3057, related to https://github.com/near/mpc/issues/2680

kevindeforth:
resolves #3061

With this PR, we treat treat `Unknown` the same as `OlderThanRecentWindow`.

With our current code, the queue contains a request from block `B`, only if block `B` has been added to the tracker (c.f. [notify_new_block](https://github.com/near/mpc/blob/22249473f051a572b377f2afa840bf200b0967dd/crates/node/src/requests/queue.rs#L282)).

That means, if we have a request from block `B` in the queue, but block `B` is unknown to the tracker, then the tracker removed it. That happens only if the block is outside of the trackers recency window.

### DIFF
--- a/crates/node/src/requests/queue.rs
+++ b/crates/node/src/requests/queue.rs
@@ -467,7 +467,7 @@ impl<RequestType: Request + Clone, ChainRespondArgsType: ChainRespondArgs>
                     request_type = %RequestType::get_type(),
                     request_id = %request.request.get_id(),
                     block_height = request.block_height,
-                    "disarding expired request"
+                    "discarding expired request"
                 );
                 // Increment metric for timeout (only for signature requests)
                 if matches!(RequestType::get_type(), types::RequestType::Signature) {
@@ -479,14 +479,9 @@ impl<RequestType: Request + Clone, ChainRespondArgsType: ChainRespondArgs>
                 requests_to_remove.push(*id);
                 continue;
             }
-
-            match self
-                .recent_blocks
-                .classify_block(request.block_hash, request.block_height)
-            {
-                CheckBlockResult::RecentAndFinal
-                | CheckBlockResult::OptimisticAndCanonical
-                | CheckBlockResult::Unknown => {
+            let block_classification = self.recent_blocks.classify_block(request.block_hash);
+            match block_classification {
+                CheckBlockResult::RecentAndFinal | CheckBlockResult::OptimisticAndCanonical => {
                     if let Some(leader) = request.current_leader(&eligible_leaders) {
                         tracing::debug!(
                             target: "request",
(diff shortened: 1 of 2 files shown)


crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -434,9 +438,8 @@ impl<T: Clone> RecentBlocksTracker<T> {
                 CheckBlockResult::OptimisticButNotCanonical
             }
             None => {
-                // At this point, the block is recent enough but we have not seen it yet.
-                // We could do a few more checks to narrow down the case, but it's not really
-                // worth the complexity. So just return Unknown.
+                // We don't know this block: either it is too old and we removed it, or we have
+                // geninely never seen it.
```
anodar: nit: typo s/geninely/genuinely

crates/node/src/requests/queue.rs
```diff
@@ -555,15 +550,21 @@ impl<RequestType: Request + Clone, ChainRespondArgsType: ChainRespondArgs>
                         "ignoring non-canonical request",
                     );
                 }
-                CheckBlockResult::NotIncluded | CheckBlockResult::OlderThanRecentWindow => {
+                CheckBlockResult::NotIncluded
+                | CheckBlockResult::OlderThanRecentWindow
+                | CheckBlockResult::Unknown => {
                     // note: We will not receive "OlderThanRecentWindow" if the `RecentBlocksTracker`
                     // has the same recencly window as the queue.
```
anodar: nit: typo s/recencly/recency

anodar (dismissed):

gilcu3 (approved):
Thanks!

anodar (approved):

PR #3073, opened 2026-04-29
chore: add status badges, skip contract-history tests in PR's

anodar:
sample of how badges will look:
image

Why skip contract-history in PR's: before it's updated it fails CI and we have pages of red CI which can mask real issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,12 @@
 # MPC
 
+[![CI](https://github.com/near/mpc/actions/workflows/ci.yml/badge.svg?branch=main)](https://github.com/near/mpc/actions/workflows/ci.yml?query=branch%3Amain)
+[![External Services Tests](https://github.com/near/mpc/actions/workflows/external-services-tests.yml/badge.svg?branch=main)](https://github.com/near/mpc/actions/workflows/external-services-tests.yml?query=branch%3Amain)
+[![nearcore](https://img.shields.io/badge/nearcore-2.11.1-blue)](https://github.com/near/nearcore/releases/tag/2.11.1)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,5 +1,12 @@
 # MPC
 
+[![CI](https://github.com/near/mpc/actions/workflows/ci.yml/badge.svg?branch=main)](https://github.com/near/mpc/actions/workflows/ci.yml?query=branch%3Amain)
+[![External Services Tests](https://github.com/near/mpc/actions/workflows/external-services-tests.yml/badge.svg?branch=main)](https://github.com/near/mpc/actions/workflows/external-services-tests.yml?query=branch%3Amain)
+[![nearcore](https://img.shields.io/badge/nearcore-2.11.1-blue)](https://github.com/near/nearcore/releases/tag/2.11.1)
```
gilcu3, line 5: badges look quite need. Could we do the nearcore one also automatically, like the rust one?

netrome:
Badges looks great. As we discussed on Slack I'll draft an issue for removing the `contract-history` tests in the PR completely. Our CI should not be concerned with the current on-chain state. These tests were just added as a reminder for us to update our contract history after making releases, but we could make it much simpler by just adding a small check in our normal CI to ensure the version we define in our Cargo.toml is the same as the version in our contract history tests.

Taking this further, and our upgrade tests should not be tied to a "single" version at all either. It could be the case that we'd want to guarantee that the next contract version is upgradable from several past versions. Currently we don't have any way of expressing that in our tests, but we should.

netrome:
Created https://github.com/near/mpc/issues/3077 to eliminate the need for the current contract-history tests, and improve our contract upgrade compatibility tests in the same go.

anodar:
Closing this in favor of https://github.com/near/mpc/issues/3077

PR #3076, opened 2026-04-29
chore: remove pytests, nearcore submodule, and pytests workflow, update docs

### LINKED ISSUE #3030: Delete pytests, nearcore submodule dependency, and workflow that runs pytests
### Background

Delete pytests, nearcore submodule dependency, and workflow that runs pytests.

### User Story

_No response_

### Acceptance Criteria

Above things are all deleted.

_No response_

anodar:
Closes: https://github.com/near/mpc/issues/3030


DSharifi (approved):

crates/e2e-tests/README.md
```diff
@@ -316,37 +315,6 @@ CI runs the same task via the `mpc-e2e-tests` job.
 
 ---
 
-## Ported tests
-
-The following tests have been ported from pytest to the Rust framework and
-live in `crates/e2e-tests/tests/`:
-
-- `cancellation_of_resharing`
```
gilcu3, line 324: I think we could leave this for historical reasons

libs/nearcore
gilcu3, line 1: will this also delete the folder?

README.md
```diff
@@ -78,12 +75,12 @@ The Rust toolchain version is pinned in [`rust-toolchain.toml`](rust-toolchain.t
...
 - _unit test_ -> a rust test in `/src` folder (per crate)
 - _integration test_ -> a rust test in `/tests` folder (per crate)
-- _system test_ -> a pytest in the `/pytest` folder
+- _e2e test_ -> a rust test in the `crates/e2e-tests` crate
 
 ### Run tests
 
 - **Unit and integration tests**: Run with `cargo nextest run --cargo-profile=test-release`
-- **System tests**: See the README in the `/pytest` directory.
+- **E2E tests**: See the README in [`crates/e2e-tests`](crates/e2e-tests/README.md).
```
gilcu3, line 83: IMO we should still call them system tests, as that naming was the result of long discussions, and still makes sense wrt unit and integration

gilcu3 (approved):
I know this has not been advertised yet, but it seems we are eager to do it, so here you go :) 

Left some minor comments

crates/e2e-tests/README.md
DSharifi: I disagree on this one. I don't see why it's relevant to keep pointers to deleted test code in the docs.

README.md
DSharifi: +1 since they are synonymous and rest of team also is familiar with it. But no strong opinion/

DSharifi: But the rewrite has consistently used E2E test, so makes sense to use that now in the readme

crates/e2e-tests/README.md
gilcu3: my idea is that if we find some discrepancy in the future we could still find what was the original test. After some time the value would be diminished ofc. No strong opinion though

README.md
gilcu3: I am not sure if we plan to have more system tests in the future (outside of e2e crate), where e2e test might no longer be appropriate for the whole set.

crates/e2e-tests/README.md
DSharifi: The source of truth for the tests would be the git history though where we'd browse the test files?

libs/nearcore
anodar: It's not marking for deletion but since it will be empty after deleting nearcore inside git shouldn't track empty dir anymore I believe.

README.md
anodar: I'd argue this is closer to E2E than system tests, although line can be blurred. As Daniel mentioned I've referred as E2E throughout in the rewrite, so I can follow up renaming to system tests instead but would rather not include that renaming in this PR as this just drops pytests.

gilcu3: yeah what I was hinting at was not renaming everything but really keeping that classification as:
- unit
- integration
- system

and then the E2E tests are part of system (right now the only part)

PR #3080, opened 2026-04-30
[Do Not Merge] Explore contract size reduction via dummy dstack verification

kevindeforth:
moving dstack attestation verification out of the MPC contract leads to ~360 kB of binary size savings (non-reproducible build):

| state   | size    |
|---------|---------|
| before: | 1439075 |
| now     | 1075766 |

Note: tests are expected to fail. This is not functional. The intent is to showcase the binary size savings of moving the dstack attestation verification out of this contract.
Specifically, the proposal would be to have the attestation verified by an external contract and get the result in a callback.

edit:
After adding back a few methods we would still likely need (https://github.com/near/mpc/pull/3080/commits/0f6bed54e38d679221d0d4e9f12b90f8cd1b4597), the binary size is still considerable at ~340 kB

| state   | size    |
|---------|---------|
| before: | 1439075 |
| now     | 1096331 |

### DIFF
--- a/crates/attestation/src/attestation.rs
+++ b/crates/attestation/src/attestation.rs
@@ -125,20 +125,63 @@ impl DstackAttestation {
     /// On success, returns the matched measurements.
     pub fn verify(
         &self,
-        _expected_report_data: ReportData,
+        expected_report_data: ReportData,
         _timestamp_seconds: u64,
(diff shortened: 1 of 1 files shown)


kevindeforth:
Will close this, as it served its purpose.

Issue #3079, opened 2026-04-30
nodes should not upload the configuration of foreign chains, just what chain they support


DSharifi:
### Background

We don't want nodes to advertise the URLs and specific RPC providers they are using. From contracts perspective it only needs to know if a a node supports the foreign chain or not, which the node can itself determine by fetching the quorum size and list of supported RPC providers and do validation locally.

### Acceptance Criteria

- mpc nodes no longer send rpc identifier or url to the contract
- remove all methods deprecated in #3086

Issue #2712, opened 2026-04-06
deprecate configuration votes in version 3.10 for foreign TX


DSharifi:
### Background

In #2648 we added configuration of foreign chain TXs to locally decide what list of foreign RPC providers to use which is planned to be released in `3.9.0`.

For backwards compatibility of the contract version `3.9.0` with nodes on `3.8.0` we kept the , we kept the interface types `ForeignChainPolicy` `ForeignChainPolicyVotes` and the respective contract methods.

All these types marked as deprecated as part of #2648 can be deleted for the release of `3.10.0`.

### Acceptance Criteria

remove all items that are marked as deprecated in #2648 after `3.9.0` release

PR #3086, opened 2026-04-30
feat(contract): add API to submit foreign support instead of configuration

### LINKED ISSUE #3087: contract stores only supported foreign chains, not the nodes' config
### Background

For foreign chain feature we want nodes to submit which chains they support to the contract. The contract does not need to know which RPC providers or urls the node is uing.

This is a prequisite for #3079 

### User Story

_No response_

### Acceptance Criteria

the contract no longer stores the rpc url and rpc identifier for foreign chains for node setup

_No response_

DSharifi:
closes #3087 

Pytests are intentionally not updated as they are not on CI and being deleted in #3076


DSharifi:
@claude review

crates/contract/src/lib.rs
```diff
@@ -1891,8 +1909,8 @@ impl MpcContract {
             .into()
     }
 
-    pub fn get_foreign_chain_configurations(&self) -> dtos::NodeForeignChainConfigurations {
-        self.node_foreign_chain_configurations.to_dto()
+    pub fn get_foreign_chain_support_by_node(&self) -> dtos::ForeignChainSupportByNode {
```
barakeinav1, line 1912: Quick question on the deprecation strategy: write side has a backward-compat shim (`register_foreign_chain_config` → `register_foreign_chain_support`), but read side removes `get_foreign_chain_configurations` entirely with no shim. Was the asymmetry deliberate?

My guess is: a read shim would have to fabricate empty URL lists, misleading callers about what's actually stored. But worth confirming, and either way a one-line note in the PR description would help future readers.

Also — any external consumers (dashboards, monitoring) currently calling `get_foreign_chain_configurations` that need a heads-up before deployment?

crates/contract/src/lib.rs
```diff
@@ -963,13 +964,30 @@ impl MpcContract {
...
+            .insert(account_id, foreign_chain_support);
 
         Ok(())
     }
 
+    #[deprecated(
+        note = "https://github.com/near/mpc/issues/3079. Node will be upgraded to use register_foreign_chain_support instead"
+    )]
+    #[handle_result]
+    pub fn register_foreign_chain_config(
```
barakeinav1, line 978: Worth opening a follow-up issue to track removal of this deprecated shim once #3079 ships and node binaries are fully upgraded — otherwise it'll quietly accumulate as tech debt. The team did this for the previous deprecation cycle ([#2712](https://github.com/near/mpc/issues/2712) tracked removing the previous deprecated foreign-chain API).

Alternatively, the `#[deprecated]` attribute could specify a target removal version (e.g., 3.10) so the cleanup expectation lives next to the code.

barakeinav1 (dismissed):
Looks good to me. I added 2 comments (deprecation strategy question + follow-up issue suggestion), but neither is blocking. Approving.

crates/contract/src/lib.rs
DSharifi: I'll update #3079's accpetance criteria to also require removing these shim methods

crates/contract/src/lib.rs
DSharifi: This read is only used for debugging and is not depended on by the nodes, so it's fine.

netrome (dismissed):
Thanks!

barakeinav1 (approved):

netrome (approved):

PR #3088, opened 2026-04-30
build(contract): switch reproducible build to wasm-opt -Os

barakeinav1:
## Summary

Switches the contract's reproducible build to a custom image
(`nearone/cargo-near-mpc`) that has `binaryen` on PATH, then replaces
cargo-near's bundled `wasm-opt -O` with `wasm-opt -Os --strip-debug
--strip-producers --strip-target-features --vacuum`.

**Result:** contract WASM goes from **1,499,976 → 1,435,586 bytes**
(**−64 KB / −4.3%**), restoring headroom under the repo's 1,490,000-byte
safety margin (`scripts/check-contract-wasm-size.sh`) and the protocol's
1,572,864-byte `max_transaction_size`. No runtime tradeoff: `-O` and `-Os`
produce byte-identical output on this contract; the savings come from the
strip-flags. End-to-end reproducibility is preserved.

## What's in this PR

- `docker/cargo-near-mpc/Dockerfile` — extends the pinned
  `sourcescan/cargo-near:0.17.0-rust-1.86.0` base with `binaryen=120-4`.
- `.github/workflows/cargo-near-image.yml` — builds & pushes the image to
  `nearone/cargo-near-mpc:<sha>` + `:latest` on pushes to `main` that touch
  the Dockerfile or the workflow itself, plus `workflow_dispatch`. Prints
  the published manifest digest in the run log.
- `crates/contract/Cargo.toml` — `[package.metadata.near.reproducible_build]`
  now points at `nearone/cargo-near-mpc:latest` and chains
  `bash -c "cargo near build … --no-wasmopt && wasm-opt -Os …"` as the
  `container_build_command`.
- `docs/contract-build-size-optimization.md` — full rationale, measurements
  for `-O` / `-Os` / `-Oz`, reproducibility argument.

## Why `-Os` (and not `-O` or `-Oz`)

Measured on the same raw wasm with the same strip-flags, wasm-opt 120:

| Pipeline | Size | Δ vs production | Runtime tradeoff |
|---|---|---|---|
| Current production (`-O`, no extra strips) | 1,499,976 | — | baseline |
| Our pipeline `-O` + strip-flags | 1,435,586 | **−64 KB** | none |
| Our pipeline `-Os` + strip-flags | 1,435,586 | **−64 KB** | none |
| Our pipeline `-Oz` + strip-flags | 1,413,072 | −87 KB | possibly slower |

`-O` and `-Os` are byte-identical on this contract — wasm-opt's `-O` already
runs the size-favoring passes that `-Os` enables. Most of the win comes from
`--strip-producers` / `--strip-target-features`, sections cargo-near's
bundled wasm-opt does not strip. `-Oz` saves a further 22 KB but at the cost
of potentially slower generated code, which would require gas-parity
measurements before adopting.

## Verification done locally

- Built the custom image; `wasm-opt 120` available on PATH; cargo-near 0.17.0
  intact.
- Ran the new `container_build_command` inside the container twice — output
  is byte-identical (sha256 stable across runs).
- Final size: 1,435,586 bytes with `-Os`; 1,413,072 with `-Oz`.

## Why this is a draft

`image_digest` in `crates/contract/Cargo.toml` is a placeholder
(`sha256:0000…`). It needs to be updated to the real manifest digest after
the new `cargo-near-image` workflow publishes the first build. The workflow
log prints the digest for easy copy-paste into `Cargo.toml`.

Once the digest is filled in:
1. `cargo near build reproducible-wasm --manifest-path crates/contract/Cargo.toml`
   should produce a deterministic 1,435,586-byte artifact,
2. CI's `mpc-contract-reproducible-build` job should pass the
   `check-contract-wasm-size.sh` gate again.

## Test plan

- [ ] Merge order: this PR can land in two phases.
  - Phase 1 (after CI merges): the `cargo-near-image` workflow runs on
    `main` and publishes `nearone/cargo-near-mpc:<sha>` + `:latest`. Capture
    the digest from the workflow log.
  - Phase 2: a small follow-up PR (or amendment to this one) replaces the
    placeholder `image_digest` with the captured digest.
- [ ] Verify locally: `cargo near build reproducible-wasm --manifest-path crates/contract/Cargo.toml`,
      confirm sha256 stable across two clean runs.
- [ ] Verify size gate passes: `bash scripts/check-contract-wasm-size.sh`.
- [ ] Run `cargo make e2e-tests` against the optimized wasm to validate
      runtime correctness on the NEAR runtime (covers the small risk that
      stripping `target_features` or `producers` interacts with the
      runtime's wasm validator).
- [ ] Confirm the embedded ABI section is preserved through the
      `--strip-*` pass (`cargo near contract abi` against the new wasm, or
      equivalent extraction).
- [ ] Communicate to consumers that the contract sha256 changes on first
      release after this lands.

### DIFF
--- a/.github/workflows/cargo-near-image.yml
+++ b/.github/workflows/cargo-near-image.yml
@@ -1,0 +1,58 @@
+name: Build cargo-near MPC image
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'docker/cargo-near-mpc/Dockerfile'
+      - '.github/workflows/cargo-near-image.yml'
+  workflow_dispatch:
+
+jobs:
+  build-and-push:
+    name: "Build and push reproducible-build image"
+    runs-on: warp-ubuntu-2404-x64-2x
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          persist-credentials: false
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build and push image
+        id: push
+        uses: docker/build-push-action@v6
+        with:
+          context: docker/cargo-near-mpc
+          file: docker/cargo-near-mpc/Dockerfile
+          push: true
+          provenance: false
+          tags: |
+            nearone/cargo-near-mpc:${{ github.sha }}
+            nearone/cargo-near-mpc:latest
+          platforms: linux/amd64
+
+      - name: Print published digest
+        run: |
+          echo "Image: nearone/cargo-near-mpc@${{ steps.push.outputs.digest }}"
+          echo "Tag:   nearone/cargo-near-mpc:${{ github.sha }}"
+          echo
+          echo "Update crates/contract/Cargo.toml [package.metadata.near.reproducible_build]:"
+          echo "  image = \"nearone/cargo-near-mpc:${{ github.sha }}\""
+          echo "  image_digest = \"${{ steps.push.outputs.digest }}\""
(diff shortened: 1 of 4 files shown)


barakeinav1:
Closing without merging.

Follow-up isolated measurements showed the PR's premise was wrong:

- The "production size = 1,499,976" baseline was a **stale local host build** (cargo-near 0.20.0), not what `cargo near build reproducible-wasm` actually produces. Real reproducible-build output is **~1,436,167 bytes** — 54 KB *under* the self-imposed safety margin, not 10 KB over it.
- All four `--strip-*` flags (debug, producers, target-features, vacuum) save **0 bytes** on this contract. The release profile's `strip = true` already removes debug info; the other sections are absent or dropped by `-O`.
- `-O` and `-Os` produce **byte-identical** output (same sha256). The PR's `-Os` change saves nothing.
- Only `-Oz` saves real bytes (~23 KB, 1.6%), at possible runtime cost. Would need gas validation on a representative `sign()` call before merging.

Since there is no actual size pressure today, no change is shipped. The investigation log is preserved in `docs/contract-build-size-optimization.md` (last commit on this branch) so future readers see the corrected analysis instead of repeating the same measurement-design error.

Revisit if the artifact grows past ~1.45 MB or specific feature work threatens the `HARD_LIMIT=1490000` gate. The Dockerfile and pipeline approach in this branch are correct in shape — only the optimization-level flag and the urgency framing were wrong.

PR #3091, opened 2026-05-01
fix: EVM inspector checks that transaction is on the canonical chain

### LINKED ISSUE #3090: EVM inspector doesn't verify that the transaction is on the canonical chain
### Background

Currently, our EVM inspector finality check only compares the difference between block numbers for our transaction block and the latest block on the canonical chain:

```rust
        let finality_is_ok =
            latest_block_with_finality_level.number >= transaction_receipt.block_number;
```

this assumes the transaction happens to exist on the canonical chain. While in practice, RPCs wouldn't return a transaction receipt response if the transaction wasn't on the canonical chain, we should not bet our security on all RPC providers upholding this guarantee.

Instead, we should add a check to ensure the block hash of the transaction receipt exists on the canonical chain according to the RPC.

### User Story

As an MPC user I want to be certain foreign tx requests are secure

### Acceptance Criteria

The EVM inspector verifies that transactions exist on the canonical chain before making the finality check,

_No response_

netrome:
closes #3090

### DIFF
--- a/crates/e2e-tests/src/foreign_chain_mock.rs
+++ b/crates/e2e-tests/src/foreign_chain_mock.rs
@@ -66,7 +66,22 @@ pub fn setup_evm_mock(server: &MockServer) {
 
             let result = match method {
                 "eth_getBlockByNumber" => {
-                    serde_json::json!({ "number": "0x16740f3" })
+                    // First param is either a finality tag (e.g. "finalized") for the
+                    // finality-head lookup, or a `0x`-prefixed block number for the
+                    // canonical-chain lookup. Return a hash matching the receipt's
+                    // blockHash for the canonical lookup so the check passes.
+                    let block_id = body["params"][0].as_str().expect("block id param");
+                    if block_id.starts_with("0x") {
+                        serde_json::json!({
+                            "number": block_id,
+                            "hash": format!("0x{MOCK_BLOCK_HASH}"),
+                        })
+                    } else {
+                        serde_json::json!({
+                            "number": "0x16740f3",
+                            "hash": format!("0x{MOCK_BLOCK_HASH}"),
+                        })
+                    }
                 }
                 "eth_getTransactionReceipt" => {
                     serde_json::json!({
--- a/crates/foreign-chain-inspector/src/evm/inspector.rs
+++ b/crates/foreign-chain-inspector/src/evm/inspector.rs
@@ -64,7 +64,9 @@ where
         )
         .await?;
 
-        if transaction_receipt.status != U64::one() {
+        let transaction_success = transaction_receipt.status != U64::one();
+
+        if transaction_success {
             return Err(ForeignChainInspectionError::TransactionFailed);
         }
 
--- a/crates/foreign-chain-inspector/src/evm/inspector.rs
+++ b/crates/foreign-chain-inspector/src/evm/inspector.rs
@@ -64,9 +64,9 @@ where
         )
         .await?;
 
-        let transaction_success = transaction_receipt.status != U64::one();
+        let transaction_success = transaction_receipt.status == U64::one();
 
-        if transaction_success {
+        if !transaction_success {
             return Err(ForeignChainInspectionError::TransactionFailed);
         }
 
--- a/crates/foreign-chain-inspector/src/evm/inspector.rs
+++ b/crates/foreign-chain-inspector/src/evm/inspector.rs
@@ -133,7 +133,11 @@ where
             .await?;
 
         if canonical.hash != receipt_block_hash {
-            return Err(ForeignChainInspectionError::NonCanonicalBlock);
+            return Err(ForeignChainInspectionError::NonCanonicalBlock {
+                block_number: receipt_block_number,
+                receipt_hash: receipt_block_hash,
+                canonical_hash: canonical.hash,
+            });
         }
         Ok(())
     }
--- a/crates/foreign-chain-inspector/src/evm/inspector.rs
+++ b/crates/foreign-chain-inspector/src/evm/inspector.rs
@@ -6,12 +6,12 @@ use jsonrpsee::core::client::ClientT;
 use crate::{EthereumFinality, ForeignChainInspectionError, ForeignChainInspector};
 
 use foreign_chain_rpc_interfaces::evm::{
-    FinalityTag, GetBlockByNumberArgs, GetBlockByNumberResponse, GetTransactionReceiptARgs,
-    GetTransactionReceiptResponse, Log, ReturnFullTransactionObjects,
+    BlockNumberOrTag, FinalityTag, GetBlockByNumberArgs, GetBlockByNumberResponse,
+    GetTransactionReceiptARgs, GetTransactionReceiptResponse, Log, ReturnFullTransactionObjects,
 };
 
 const GET_TRANSACTION_RECEIPT_METHOD: &str = "eth_getTransactionReceipt";
-const GET_BLOCK_BY_FINALITY_METHOD: &str = "eth_getBlockByNumber";
+const GET_BLOCK_BY_NUMBER_METHOD: &str = "eth_getBlockByNumber";
 
 /// Marker trait for EVM-compatible chain type parameters.
 ///
@@ -50,13 +50,15 @@ where
             EthereumFinality::Safe => FinalityTag::Safe,
             EthereumFinality::Latest => FinalityTag::Latest,
         };
-        let get_latest_block_by_finality_args =
-            GetBlockByNumberArgs::new(finality_tag, ReturnFullTransactionObjects::from(false));
+        let get_latest_block_by_finality_args = GetBlockByNumberArgs::new(
+            BlockNumberOrTag::Tag(finality_tag),
+            ReturnFullTransactionObjects::from(false),
+        );
 
         let latest_block_with_finality_level: GetBlockByNumberResponse = self
             .client
             .request(
-                GET_BLOCK_BY_FINALITY_METHOD,
+                GET_BLOCK_BY_NUMBER_METHOD,
                 &get_latest_block_by_finality_args,
             )
             .await?;
@@ -78,6 +80,23 @@ where
 
         if !finality_is_ok {
             return Err(ForeignChainInspectionError::NotFinalized);
+        }
+
+        // Defense in depth: a number-only check accepts any receipt whose block height is at or
+        // below the finalized head, but a malicious or out-of-sync RPC could still return a
+        // receipt from a side block. Re-fetch the canonical block at that height and reject
+        // when its hash does not match the receipt's block hash.
+        let get_canonical_block_args = GetBlockByNumberArgs::new(
+            BlockNumberOrTag::Number(transaction_receipt.block_number),
+            ReturnFullTransactionObjects::from(false),
+        );
+        let canonical_block: GetBlockByNumberResponse = self
+            .client
+            .request(GET_BLOCK_BY_NUMBER_METHOD, &get_canonical_block_args)
+            .await?;
+
+        if canonical_block.hash != transaction_receipt.block_hash {
+            return Err(ForeignChainInspectionError::NonCanonicalBlock);
         }
 
         let transaction_success = ethereum_types::U64::one() == transaction_receipt.status;
--- a/crates/foreign-chain-inspector/src/evm/inspector.rs
+++ b/crates/foreign-chain-inspector/src/evm/inspector.rs
@@ -7,7 +7,8 @@ use crate::{EthereumFinality, ForeignChainInspectionError, ForeignChainInspector
 
 use foreign_chain_rpc_interfaces::evm::{
     BlockNumberOrTag, FinalityTag, GetBlockByNumberArgs, GetBlockByNumberResponse,
-    GetTransactionReceiptARgs, GetTransactionReceiptResponse, Log, ReturnFullTransactionObjects,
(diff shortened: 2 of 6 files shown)


netrome:
Based on https://github.com/near/mpc/pull/3093 - should resolve it before asking for reviews and merging this one.

crates/foreign-chain-rpc-interfaces/src/evm.rs
```diff
@@ -23,20 +23,44 @@ pub struct GetTransactionReceiptARgs {
     pub transaction_hash: H256,
 }
```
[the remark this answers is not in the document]
netrome: Unrelated to this PR, won't fix now.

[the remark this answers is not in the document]
netrome: Unrelated.

crates/foreign-chain-rpc-interfaces/src/evm.rs
```diff
@@ -23,20 +23,44 @@ pub struct GetTransactionReceiptARgs {
...
-/// Partial RPC arguments for `eth_getTransactionReceipt`.
-/// <https://ethereum.org/developers/docs/apis/json-rpc/#eth_gettransactionreceipt>
+/// Partial RPC arguments for `eth_getBlockByNumber`.
+/// <https://ethereum.org/developers/docs/apis/json-rpc/#eth_getBlockByNumber>
 #[derive(
     Constructor, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize,
 )]
-pub struct GetBlockByNumberArgs(FinalityTag, ReturnFullTransactionObjects);
+pub struct GetBlockByNumberArgs(BlockNumberOrTag, ReturnFullTransactionObjects);
+
```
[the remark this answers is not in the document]
netrome: This is only used within the workspace for our RPC calls, so it's fine.

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -65,6 +65,10 @@ pub enum ForeignChainInspectionError {
     },
     #[error("transaction has not reached expected finality level")]
     NotFinalized,
+    #[error(
+        "transaction receipt's block_hash does not match the canonical chain at that block number"
+    )]
+    NonCanonicalBlock,
```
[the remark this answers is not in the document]
netrome: Fixed in https://github.com/near/mpc/pull/3091/commits/0a98257dd785711eadf0501ad1fae658f5554408

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -64,6 +64,10 @@ pub enum ForeignChainInspectionError {
...
```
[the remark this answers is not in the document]
netrome: Outdated.

crates/foreign-chain-inspector/src/evm/inspector.rs
```diff
@@ -73,16 +56,15 @@ where
...
-        let transaction_success = ethereum_types::U64::one() == transaction_receipt.status;
+        self.verify_finality_level(transaction_receipt.block_number, finality)
+            .await?;
+        self.verify_block_is_canonical(
+            transaction_receipt.block_number,
+            transaction_receipt.block_hash,
+        )
+        .await?;
 
-        if !transaction_success {
```
DSharifi, line 85: Why remove the variable name which explains what the check does?

crates/foreign-chain-inspector/src/evm/inspector.rs
```diff
@@ -6,12 +6,13 @@ use jsonrpsee::core::client::ClientT;
...
-    FinalityTag, GetBlockByNumberArgs, GetBlockByNumberResponse, GetTransactionReceiptARgs,
-    GetTransactionReceiptResponse, Log, ReturnFullTransactionObjects,
+    BlockNumberOrTag, FinalityTag, GetBlockByNumberArgs, GetBlockByNumberResponse,
+    GetTransactionReceiptARgs, GetTransactionReceiptResponse, H256, Log,
+    ReturnFullTransactionObjects, U64,
 };
 
 const GET_TRANSACTION_RECEIPT_METHOD: &str = "eth_getTransactionReceipt";
-const GET_BLOCK_BY_FINALITY_METHOD: &str = "eth_getBlockByNumber";
+const GET_BLOCK_BY_NUMBER_METHOD: &str = "eth_getBlockByNumber";
```
DSharifi, line 15: Thanks

crates/foreign-chain-inspector/tests/evm_inspector.rs
```diff
@@ -155,15 +176,22 @@ macro_rules! evm_inspector_tests {
...
                 };
+                // not reached: extract returns NotFinalized before the canonical lookup
+                let canonical_block_response = GetBlockByNumberResponse {
+                    number: tx_response.block_number,
+                    hash: tx_response.block_hash,
+                };
 
-                let mock_client = mock_evm_client(block_response, tx_response);
+                let mock_client =
+                    mock_evm_client(block_response, tx_response, canonical_block_response);
```
DSharifi: Nit:
can you make this a builder to avoid boilerplate args that are not needed/dead in the test logic?

crates/foreign-chain-inspector/src/evm/inspector.rs
```diff
@@ -73,16 +56,15 @@ where
...
-        }
-
-        let transaction_success = ethereum_types::U64::one() == transaction_receipt.status;
+        self.verify_finality_level(transaction_receipt.block_number, finality)
+            .await?;
+        self.verify_block_is_canonical(
+            transaction_receipt.block_number,
+            transaction_receipt.block_hash,
+        )
+        .await?;
```
DSharifi, line 65: To avoid two roundtrips here, we could batch these requests. There's good support for it with the jsonrpsee crate we are using. https://docs.rs/jsonrpsee/latest/jsonrpsee/async_client/struct.Client.html#method.batch_request

crates/foreign-chain-rpc-interfaces/src/evm.rs
```diff
@@ -23,20 +23,44 @@ pub struct GetTransactionReceiptARgs {
...
     pub number: U64,
+    /// the canonical block hash at this block number
+    pub hash: H256,
 }
 
-/// Partial RPC arguments for `eth_getTransactionReceipt`.
-/// <https://ethereum.org/developers/docs/apis/json-rpc/#eth_gettransactionreceipt>
+/// Partial RPC arguments for `eth_getBlockByNumber`.
+/// <https://ethereum.org/developers/docs/apis/json-rpc/#eth_getBlockByNumber>
 #[derive(
```
DSharifi, line 38: Thanks for updating this

DSharifi (dismissed):
Thank you. Took me a while to understand that this doesn't help against malicious RPC providers, but only if they decide t return receipts for transactions that aren't on the canonical chain.

netrome:
> Thank you. Took me a while to understand that this doesn't help against malicious RPC providers, but only if they decide t return receipts for transactions that aren't on the canonical chain.

Ah yeah apologies if that wasn't crystal clear. Yeah this is only for protecting against if honest RPC providers would happen to serve data from the non-canonical chain (which they typically don't).

crates/foreign-chain-inspector/src/evm/inspector.rs
netrome: Thanks for spotting. This was AI slop slipping through, I'll fix.

crates/foreign-chain-inspector/src/evm/inspector.rs
netrome: Hmm the barch request functionality seems to require using some type erasure, so it might turn out a bit ugly. But I'll give it a go and see if I can make it pretty, Otherwise I'd prioritize readability over performance in this case.

crates/foreign-chain-inspector/tests/evm_inspector.rs
netrome: Yeah good idea, will fix

crates/foreign-chain-inspector/src/evm/inspector.rs
netrome: Done in https://github.com/near/mpc/pull/3091/commits/45341c9c10ea426f849989a01353c8b3db0945a2

crates/foreign-chain-inspector/tests/evm_inspector.rs
netrome: Hmm though I'd assume the args are still needed in the test, since the inspector now makes the two separate calls for the finality check and canonical chain check

netrome: Done in https://github.com/near/mpc/pull/3091/commits/59ae36cb3bbf9488c76767d95e9cfd71936c96b6 - I think this looks nicer and there were a few tests where we didn't have to configure all responses.

crates/foreign-chain-rpc-interfaces/src/evm.rs
netrome: Gotta give Claude cred for spotting this, though I was happy to see it as well :)

crates/foreign-chain-inspector/src/evm/inspector.rs
netrome: Always nice with minor cleanups :blush:

crates/foreign-chain-inspector/src/evm/inspector.rs
```diff
@@ -73,16 +56,17 @@ where
...
 
-        let transaction_success = ethereum_types::U64::one() == transaction_receipt.status;
+        let transaction_success = transaction_receipt.status != U64::one();
 
-        if !transaction_success {
+        if transaction_success {
             return Err(ForeignChainInspectionError::TransactionFailed);
         }
```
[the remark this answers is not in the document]
netrome: Ah right, good observation. I was a bit sloppy when I fixed this. Updated in https://github.com/near/mpc/pull/3091/commits/73106e68b91c7ced03029ec4ecbaab10f841497a

crates/foreign-chain-inspector/src/evm/inspector.rs
```diff
@@ -73,16 +56,17 @@ where
...
-        if !finality_is_ok {
-            return Err(ForeignChainInspectionError::NotFinalized);
-        }
+        self.verify_finality_level(transaction_receipt.block_number, finality)
+            .await?;
+        self.verify_block_is_canonical(
+            transaction_receipt.block_number,
+            transaction_receipt.block_hash,
+        )
+        .await?;
```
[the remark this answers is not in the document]
netrome: The order doesn't matter, as long as both checks are executed. Though I appreciate you checking the acceptance criteria.

crates/foreign-chain-inspector/src/evm/inspector.rs
netrome: Okay this ended up being pretty nice. Done in https://github.com/near/mpc/pull/3091/commits/d6609077552cb204a2baab9c625bed5b625c592f

crates/foreign-chain-inspector/src/evm/inspector.rs
```diff
@@ -73,14 +56,15 @@ where
...
```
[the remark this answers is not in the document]
netrome: Already commented, this is ok

crates/foreign-chain-inspector/src/evm/inspector.rs
netrome: Edit: Got compiler errors that weren't trivial to fix due to higher rank trait bounds that I didn't want to dive deep into now, caused by lifetimes in the batch call. Skipping this for now.

crates/e2e-tests/src/foreign_chain_mock.rs
```diff
@@ -66,7 +66,22 @@ pub fn setup_evm_mock(server: &MockServer) {
...
+                            "number": block_id,
+                            "hash": format!("0x{MOCK_BLOCK_HASH}"),
+                        })
+                    } else {
+                        serde_json::json!({
+                            "number": "0x16740f3",
+                            "hash": format!("0x{MOCK_BLOCK_HASH}"),
+                        })
+                    }
                 }
```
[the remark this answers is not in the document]
netrome: Reverted batch request anyway

crates/foreign-chain-inspector/tests/common.rs
```diff
@@ -63,3 +75,43 @@ pub struct JsonRpcResponse<T> {
...
     pub id: u64,
 }
+
+/// Builds a mock RPC client that returns pre-configured responses in call order.
+///
+/// Each call to the resulting client returns the response at the corresponding call index,
+/// regardless of method name or params. Useful for tests of flows that issue a
+/// deterministic sequence of RPC calls. Panics if the client is called more times than
+/// responses were configured. Configured responses that are never used are left unused.
+#[derive(Default)]
```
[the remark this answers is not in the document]
netrome: Yeah I considered this, but it would be quite some more work since it would need to be a dedicated type with a custom Drop impl instead of reusing the existing `FixedResponseRpcClient`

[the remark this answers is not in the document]
DSharifi: This would have been nice. I can add it in a follow up

DSharifi (approved):

anodar (approved):

PR #3102, opened 2026-05-04
chore: make schemars crate excluded from compiling under wasm32 target

### LINKED ISSUE #759: Make abi-schema related crates be excluded from compiling under `wasm32` target
### Background

In order to generate abi, we include additional dependencies, which doesn't need to be enabled when compiling for wasm32 target
https://github.com/near/mpc/pull/751#discussion_r2247457570

### User Story

The size of the contract has increased by 1kb out of 953, still we want to avoid it.

### Acceptance Criteria

Exclude schema related crates from wasm32 target compilation

_No response_

gilcu3:
Closes #759

### DIFF
--- a/crates/primitives/Cargo.toml
+++ b/crates/primitives/Cargo.toml
@@ -5,7 +5,7 @@ license = { workspace = true }
 edition = { workspace = true }
 
 [features]
-abi = ["borsh/unstable__schema", "near-account-id/schemars-v0_8", "schemars"]
+abi = ["borsh/unstable__schema", "schemars"]
 
 [dependencies]
 borsh = { workspace = true }
@@ -16,6 +16,7 @@ serde = { workspace = true }
 thiserror = { workspace = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+near-account-id = { workspace = true, features = ["borsh", "serde", "schemars-v0_8"] }
 schemars = { workspace = true, optional = true }
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


crates/primitives/Cargo.toml
```diff
@@ -16,6 +16,7 @@ serde = { workspace = true }
 thiserror = { workspace = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+near-account-id = { workspace = true, features = ["borsh", "serde", "schemars-v0_8"] }
```
barakeinav1, line 19: nit (cosmetic, non-blocking): `near-account-id` is declared twice — once in `[dependencies]` with `["borsh", "serde"]` and again in the wasm32-excluded block with `["borsh", "serde", "schemars-v0_8"]`. Cargo correctly unions features per target, but it's slightly less readable than the alternative pattern of listing the dep once in regular `[dependencies]` and only adding the *extra* feature in the wasm32-excluded section, e.g.:

```toml
[dependencies]
near-account-id = { workspace = true, features = ["borsh", "serde"] }

[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
near-account-id = { workspace = true, features = ["schemars-v0_8"] }
```

The two forms are equivalent functionally. Just a minor readability call.

barakeinav1 (approved):
Looks good. Posted one cosmetic nit on the duplicated `near-account-id` declaration in `primitives/Cargo.toml`, but it's non-blocking. Approving.

BTW — did you measure the actual contract size saving?

gilcu3:
> BTW — did you measure the actual contract size saving?

Yeah it was around 1Kb as mentioned in the issue

crates/primitives/Cargo.toml
gilcu3: For me it actually feels more readable in the current form, as the other one only works in wasm32 bc the presence of feature unification.

DSharifi (approved):

PR #3106, opened 2026-05-04
feat: accept foreign tx validation requests on contract

### LINKED ISSUE #2801: Arbitrum support in mpc contract
### Background

smart contract accepts Arbitrum foreign chain validation requests

### User Story

_No response_

### Acceptance Criteria

smart contract accepts Arbitrum foreign chain validation requests

_No response_

DSharifi:
closes #2801

### DIFF
--- a/crates/contract/tests/sandbox/foreign_chain_request.rs
+++ b/crates/contract/tests/sandbox/foreign_chain_request.rs
@@ -1,7 +1,8 @@
 #![allow(non_snake_case)]
 
 use crate::sandbox::common::{
-    abstract_evm_request, await_pending_foreign_tx_request_observed_on_contract, base_evm_request,
+    abstract_evm_request, arbitrum_evm_request,
+    await_pending_foreign_tx_request_observed_on_contract, base_evm_request,
     bitcoin_extracted_values, bitcoin_request, bnb_evm_request, ethereum_evm_request,
     evm_block_hash_extracted_values, register_foreign_chain_configuration,
     sign_foreign_tx_response, starknet_extracted_values, starknet_request, SandboxTestSetup,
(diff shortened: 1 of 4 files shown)


pbeza (approved):

crates/contract/tests/sandbox/foreign_chain_request.rs
```diff
@@ -24,6 +25,7 @@ const SIGNATURE_TIMEOUT_BLOCKS: u64 = 200;
 #[case::starknet(starknet_request(), starknet_extracted_values())]
 #[case::bnb(bnb_evm_request(), evm_block_hash_extracted_values())]
 #[case::base(base_evm_request(), evm_block_hash_extracted_values())]
+#[case::arbitrum(arbitrum_evm_request(), evm_block_hash_extracted_values())]
 #[tokio::test]
 async fn verify_foreign_transaction__should_succeed(
```
gilcu3, line 30: side comment: I wonder if we need to test each chain separately here. These tests are unfortunately getting a noticeable time inside the sandbox tests (just my experience running them locally sometimes). If the behavior in the contract is actually the same for all chains, maybe we could just test one, or use some mechanism (trait?) to ensure that testing one is enough

gilcu3 (approved):
Thank you!

crates/contract/tests/sandbox/foreign_chain_request.rs
DSharifi: I wouldn't want to lose test coverage. But we could instead collapse them all to one test case, and run requests against all of them in one test similar to what @anodar did with the e2e tests https://github.com/near/mpc/blob/9c8accae15a26d6a745b11c1ba6c070e111c7296/crates/e2e-tests/tests/foreign_chain_tx_validation.rs#L369-L389

gilcu3: yeah that would also fix the problem indeed

PR #3108, opened 2026-05-04
fix: recent blocks tracker memory leak

### LINKED ISSUE #3103: Recent Blocks Tracker Memory Leak
### Background

We have a memory leak in `RecentBlocksTracker`.
Each `BlockNode` holds a strong reference to its parent and to its children:

https://github.com/near/mpc/blob/13804b7edbce47baa973d751a941bb930ebffd2d/crates/node/src/requests/recent_blocks_tracker.rs#L106-L119

Together, these hold a reference cycle.

The tracker prunes blocks that are no longer required in `prune_old_blocks`. The method:

1. Walks from `root_children`, classifying nodes as kept (`new_root_children`) or pruned (`old_nodes`).
2. Sets `parent = None` on kept nodes **so they no longer point up at pruned ancestors.**
3. Replaces `self.root_children` with `new_root_children` (drops `Arc`s to pruned roots).
4. Removes pruned nodes from `self.hash_to_node` (drops the per-block `Arc`).

https://github.com/near/mpc/blob/13804b7edbce47baa973d751a941bb930ebffd2d/crates/node/src/requests/recent_blocks_tracker.rs#L375-L409

The issue is that while step 2 ensures no node that is kept will point to a node that is pruned, we do not break any cycles that may exist in pruned nodes.

Image

### User Story

As a node operator, I like not paying for avoidable RAM.

### Acceptance Criteria

Ensure pruned blocks are removed from memory.

For example: 
```rust
for old_node in old_nodes {
    self.hash_to_node.remove(&old_node.hash);
    *old_node.children.lock().unwrap() = Vec::new();
}
```

Also, add a unit test for it:

1. Adds a chain of N blocks where the latest is finalized.
2. Calls `Arc::downgrade` on one of the early blocks (held via `hash_to_node`) and saves the `Weak`.
3. Adds enough additional blocks past the window to force a prune of that early block.
4. Asserts `weak.upgrade().is_none()`.

_No response_

kevindeforth:
resolves #3103

test fails without the fix: https://github.com/near/mpc/pull/3109

### DIFF
--- a/crates/node/src/requests/recent_blocks_tracker.rs
+++ b/crates/node/src/requests/recent_blocks_tracker.rs
@@ -928,4 +928,61 @@ pub mod tests {
         assert_eq!(t.check(&b110), CheckBlockResult::NotIncluded);
         assert_eq!(t.check(&b111), CheckBlockResult::NotIncluded);
     }
+
+    #[test]
+    #[expect(non_snake_case)]
+    fn prune_old_blocks__should_drop_connected_pruned_block_nodes() {
+        // Given a tree where finality is stalled by a height gap (b3 is skipped),
+        // so b1 and b2 sit in the tracker without any finality advancement.
+        // The first finality event will jump straight to b4, pruning b1 and b2
+        // together in a single prune_old_blocks call.
+        let mut tester = Tester::new(2);
+        let b1 = tester.block(1);
+        let b2 = b1.child(); // h=2
+        let b4 = b2.descendant(4); // h=4, parent=b2 (h=3 intentionally skipped)
+        let b5 = b4.child(); // h=5
+        let b6 = b5.child(); // h=6 → first block whose grandparent-parent-self
+                             // forms 3 consecutive heights (b4-b5-b6), so its
+                             // last_final_block is b4
+
+        tester.add(&b1, "1");
+        tester.add(&b2, "2");
+        tester.add(&b4, "4");
+        tester.add(&b5, "5");
+
+        // Capture Weak refs to b1 and b2 while they are still in the tree.
+        // After add(b6), final_head will jump from None to b4 (h=4), and
+        // prune_old_blocks will evict both b1 (h=1) and b2 (h=2) in one call.
+        // b1 and b2 are connected by the parent<->children Arc cycle:
+        //   b1.children = [Arc(b2)]
+        //   b2.parent   = Some(Arc(b1))
+        // Without breaking the cycle, neither can be freed.
+        let weak_b1 = Arc::downgrade(
+            tester
+                .tracker
+                .hash_to_node
+                .get(&b1.hash)
+                .expect("b1 present before prune"),
+        );
+        let weak_b2 = Arc::downgrade(
+            tester
+                .tracker
+                .hash_to_node
+                .get(&b2.hash)
+                .expect("b2 present before prune"),
+        );
+
+        // When the finality jump triggers a multi-node prune
+        tester.add(&b6, "6");
+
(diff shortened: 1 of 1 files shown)


pbeza:
@kevindeforth Curious how you spotted the leak — just by studying the code? I’ll treat this fix as a benchmark for future AI tools.

gilcu3 (approved):
LGTM

I would love to take some time to understand this part of the code, but the one liner fix seems good :)

pbeza (approved):
@kevindeforth, **great** catch! 👏🏼

Do you know why we're using `Arc` for both `children` and `parent`? IIUC the more natural choice would be `Weak<BlockNode>` for `parent`. Trees and DLLs in Rust are usually implemented so node back-edges aren't owning — sometimes via `Weak` ([html5ever's `RcDom`](https://github.com/servo/html5ever/blob/201534ee4f4c3db10dc558869555768d864278bf/rcdom/lib.rs#L109-L117)), sometimes via raw `NonNull` ([`std::BTreeMap`'s `parent`](https://github.com/rust-lang/rust/blob/main/library/alloc/src/collections/btree/node.rs#L53), or both ends in [`std::LinkedList`](https://github.com/rust-lang/rust/blob/master/library/alloc/src/collections/linked_list.rs#L50-L66) with ownership [held by the container](https://github.com/rust-lang/rust/blob/1d72d7e8136faaebad3a85eeed432e6ea1b2ffab/library/alloc/src/collections/linked_list.rs#L58) via `PhantomData<Box<Node<T>>>`).

kevindeforth:
> Curious how you spotted the leak — just by studying the code?

That was actually a lot of Claude. I am still trying to find the best way for refactoring this code. One of the options is to rely on `Weak::upgrade() == None`, for assessing whether a block was removed from the `RecentBlocksTracker`.

When exploring this option with Claude, **it** suggested:

> Today's prune_old_blocks would need an audit before relying on this. If any cycle leaks, prune becomes silent and the queue never sees the tombstone

So, Claude (Opus 4.7) flagged the possibility of there being a bug and then found the bug itself. Pretty nice.

Claude also wrote the unittest and it was the first time I didn't feel the need to change any of it.


> Do you know why we're using `Arc` for both `children` and `parent`? IIUC the more natural choice would be `Weak<BlockNode>` for `parent`. Trees and DLLs in Rust are usually implemented so node back-edges aren't owning — sometimes via `Weak` ([html5ever's `RcDom`](https://github.com/servo/html5ever/blob/201534ee4f4c3db10dc558869555768d864278bf/rcdom/lib.rs#L109-L117)), sometimes via raw `NonNull` ([`std::BTreeMap`'s `parent`](https://github.com/rust-lang/rust/blob/main/library/alloc/src/collections/btree/node.rs#L53), or both ends in [`std::LinkedList`](https://github.com/rust-lang/rust/blob/master/library/alloc/src/collections/linked_list.rs#L50-L66) with ownership [held by the container](https://github.com/rust-lang/rust/blob/1d72d7e8136faaebad3a85eeed432e6ea1b2ffab/library/alloc/src/collections/linked_list.rs#L58) via `PhantomData<Box<Node<T>>>`).

That's a great question! I don't know, but my guess is this was just an oversight. The author left this comment:

https://github.com/near/mpc/blob/13804b7edbce47baa973d751a941bb930ebffd2d/crates/node/src/requests/recent_blocks_tracker.rs#L47-L48

But I don't see why we couldn't use `Weak<BlockNode>`. Will open a follow-up for this!

edit: https://github.com/near/mpc/issues/3114

Issue #3113, opened 2026-05-04
Remove `curve` field from `DomainConfig`


gilcu3:
### Background

After #2442 (Protocol enum + V2Secp256k1 removal) `Curve` is fully derivable from `Protocol` via `From<Protocol> for Curve`. The `curve` field on `DomainConfig` is now redundant; keeping both creates a validation burden (`validate_domain_consistency`, `InconsistentCurveProtocol`) and risks divergence. Per the [domain-separation design](docs/design/domain-separation.md) §2.1, `Curve` should be derived, not stored.

This is the `curve`-removal slice of design PR 4; splitting it out lets it land independently of `ReconstructionThreshold`.

### Acceptance Criteria

- Remove `curve` from the internal contract `DomainConfig` (`crates/contract/src/primitives/domain.rs`).
- Remove `curve` from the DTO `DomainConfig` in `near-mpc-contract-interface`.
- Add a Borsh `migrate()` converting old `{id, curve, protocol, purpose}` to new `{id, protocol, purpose}`.
- Drop `validate_domain_consistency` and the `InconsistentCurveProtocol` error variant.
- Update all `domain.curve` reads to `Curve::from(domain.protocol)` (node, sandbox tests, e2e tests, devnet).

Issue #2970, opened 2026-04-21
Add e2e test for contract/node upgrade compatibility


gilcu3:
### Background

The pytest `test_update_to_current` was removed in #1186 (tracked by #1185) and replaced with sandbox integration tests in the contract crate. At the time, the upgrade workflow was: upgrade the contract first, then the node — so the sandbox tests were sufficient.

Since then, the upgrade order has been inverted (node first, then contract). This creates a coverage gap: sandbox tests only exercise contract DTOs, not the node's actual deserialization behavior. If contract serialization and node deserialization change together such that the new node is compatible with the new contract but not the old one, this will only be caught at deployment time.

### Acceptance Criteria

- An e2e test covers the upgrade scenario where nodes of the new version run against a contract of the previous version, verifying the node remains functional (signing, indexing, etc.) until the contract is upgraded.

PR #3115, opened 2026-05-04
refactor: remove V2Secp256k1 variant

### LINKED ISSUE #2442: Add Protocol separation in the Smart Contract

gilcu3:
Closes #2442

The `curve` will be removed from `DomainConfig` in a follow-up #3113


crates/node/src/coordinator.rs
```diff
@@ -547,29 +547,51 @@ where
...
+                                "Keyshare data does not match the domain protocol",
+                            );
+                            routing_invariant_violated = true;
                         }
                     }
                 }
+                if routing_invariant_violated {
+                    tracing::error!("Keyshare/domain routing invariant violated; halting until contract state changes.");
+                    return Ok(MpcJobResult::HaltUntilInterrupted);
+                }
```
gilcu3: An alternative to this would be to log the errors and continue, let me know if you think that would be better

gilcu3:
@claude review

please check again, we know no node in testnet or devnet has the `V2Secp256k1` variant stored

crates/contract/src/lib.rs
```diff
@@ -2427,7 +2429,7 @@ mod tests {
         rng: &mut impl CryptoRngCore,
     ) -> (dtos::PublicKey, SharedSecretKey) {
         match domain_curve {
-            Curve::Secp256k1 | Curve::V2Secp256k1 => {
+            Curve::Secp256k1 => {
```
kevindeforth, line 2450: nit: I would name this method `make_public_key_for_curve`

crates/node/src/coordinator.rs
kevindeforth: Panicking is fine I think. This codepath is not really reachable.
But a better place for these checks would be during construction of the `ContractRunningState` in `participants.rs`.

kevindeforth (dismissed):
compared it against https://github.com/near/mpc/issues/2442#issuecomment-4369158576

These refactors are pretty hard to follow and considering https://github.com/near/mpc/issues/2970 is still open, I just hope we don't miss anything.

gilcu3:
> compared it against [#2442 (comment)](https://github.com/near/mpc/issues/2442#issuecomment-4369158576)
> 
> These refactors are pretty hard to follow and considering #2970 is still open, I just hope we don't miss anything.

Indeed I think I should tackle #2970 asap, I was thinking on it yesterday :) Thanks for the reminder!

crates/node/src/coordinator.rs
gilcu3: Will switch to `bail!` then. Doing it in `ContractRunningState` might not help much, in the sense that we still need to match against the `Keyshares`, so will leave in coordinator for now.

gilcu3:
@claude review (I want to see if we get better reviews now)

kevindeforth (approved):
Thanks!

gilcu3:
@claude review (lets see how the new model compares to the old one)

anodar (approved):
Looks like `docs/foreign-chain-transactions.md` and `docs/design/domain-separation.md` still mention `V2Secp256k1`.

gilcu3:
> Looks like `docs/foreign-chain-transactions.md` and `docs/design/domain-separation.md` still mention `V2Secp256k1`.

for the latter it makes sense to keep it, for the former we need to check, but given that it is anyway not used there we can solve that later when that doc is updated after the latest design changes

PR #3124, opened 2026-05-05
test: flaky migration service test

### LINKED ISSUE #3123: Fix flaky migration service test
### Background

Scenario that causes this race condition:

  1. Test `calls start_node_migration` on the contract. The contract now records a destination pointing at target's p2p key.
  2. Test calls `put_keyshares` to target's migration web server. TCP connect → TLS handshake completes.
  3. Target's indexer ingests the `start_node_migration` block → target's local `MigrationInfo.active_migration` flips false → true → cancellation fires → the in-flight TLS connection from step 2 is dropped mid-request → backup-cli reports failure → test panics.

Logs collected from the failed run (https://github.com/near/mpc/actions/runs/25104589273/job/73562427323)
```

  11:35:01.031984  my migration state changed   active_migration: false
                   ↑ register_backup_service indexed; only backup_service_info changed.

  11:35:02.021526  Handle connection            ← PUT keyshares TCP accepted (1)
  11:35:02.027360  Handle connection            ← second TCP accept (2)

  11:35:02.028117  contract migration state changed (block 168)
                   ↑ start_node_migration just got applied;
                     contract now has destination_node_info → target.

  11:35:02.028130  my migration state changed   active_migration: true  ← the flip
  11:35:02.028138  TLS handshake complete, peer authenticated …
  11:35:02.028142  TLS handshake complete, backup service authenticated and encrypted channel established
  11:35:02.028412  dropping connection due to cancellation (change in migration info …)
                   ↑ exactly what `expected_peer.cancelled.cancelled()` arm logs.

```

### User Story

_No response_

### Acceptance Criteria

Query the nodes debug endpoint to make sure we put the keyshares only when it has a correct view of the migration service

_No response_

anodar:
Closes: https://github.com/near/mpc/issues/3123

start_migration_and_wait now also polls the target node's /debug/migrations endpoint until it reflects the new destination_node_info before returning, to make sure an in-flight PUT keyshares connection isn't torn down by the migration web server's cancellation token when the target's indexer first sees the migration state change.

### DIFF
--- a/crates/e2e-tests/tests/migration_service.rs
+++ b/crates/e2e-tests/tests/migration_service.rs
@@ -330,7 +330,43 @@ async fn start_migration_and_wait(
             ),
     )
     .await
-    .context("timed out waiting for contract to reflect node migration")
+    .context("timed out waiting for contract to reflect node migration")?;
+
+    // Wait until the target node's own indexer has caught up to the
+    // start_node_migration block. Without this, the target's
+    // `MigrationInfo.active_migration` flips false → true mid-PUT, which
+    // fires the migration web-server's per-connection cancellation token
+    // and tears down backup-cli's in-flight TLS stream
+    // (see `migration_service::web::server::handle_stream`). Mirrors the
+    // source-side poll in `register_backup_service_and_wait`.
+    let target_web_addr = match &cluster.nodes[target_idx] {
+        MpcNodeState::Running(n) => n.web_address(),
+        _ => bail!("target node not running"),
+    };
+    let http_client = reqwest::Client::new();
+    (|| async {
+        let resp = http_client
+            .get(format!("http://{target_web_addr}/debug/migrations"))
+            .send()
+            .await?;
+        let body = resp.text().await?;
+        anyhow::ensure!(
+            body.contains(&target_p2p_key),
+            "target node debug endpoint doesn't reflect destination_node_info yet"
+        );
+        Ok(())
+    })
+    .retry(
+        ConstantBuilder::default()
+            .with_delay(common::POLL_INTERVAL)
+            .with_max_times(
+                (INDEXER_SYNC_TIMEOUT.as_millis() / common::POLL_INTERVAL.as_millis()) as usize,
+            ),
+    )
+    .await
+    .context("timed out waiting for target node to index node migration")?;
+
+    Ok(())
 }
 
 /// PUT keyshares to the target node via the backup CLI.
--- a/crates/e2e-tests/tests/migration_service.rs
+++ b/crates/e2e-tests/tests/migration_service.rs
@@ -4,11 +4,22 @@ use std::path::PathBuf;
 use std::process::Command;
 use std::time::Duration;
 
+use std::collections::BTreeMap;
+
 use anyhow::{Context, bail};
 use backon::{ConstantBuilder, Retryable};
 use e2e_tests::MpcNodeState;
-use near_mpc_contract_interface::types::ProtocolContractState;
+use near_mpc_contract_interface::types::{
+    AccountId, BackupServiceInfo, DestinationNodeInfo, ProtocolContractState,
+};
 use rand::SeedableRng;
+
+/// Mirror of the production `node::indexer::migrations::ContractMigrationInfo`
+/// type. Used to deserialize the `/debug/migrations` response strictly so the
+/// readiness asserts compare typed fields rather than substring-match the
+/// JSON body.
+type ContractMigrationInfo =
+    BTreeMap<AccountId, (Option<BackupServiceInfo>, Option<DestinationNodeInfo>)>;
 
 const MIGRATION_PORT_TIMEOUT: Duration = Duration::from_secs(120);
 const INDEXER_SYNC_TIMEOUT: Duration = Duration::from_secs(30);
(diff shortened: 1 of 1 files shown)


crates/node/src/migration_service/web/server.rs
```diff
@@ -71,19 +71,34 @@ async fn handle_stream(
...
         }
         _ = expected_peer.cancelled.cancelled() => {
-            tracing::info!("dropping connection due to cancellation (change in migration info or cancellation of web server)");
+            // Don't kill the in-flight request — that surfaces to the
+            // backup-cli client as "peer closed connection without sending
+            // TLS close_notify" because hyper drops the TLS stream without
+            // sending close_notify. Instead, signal hyper to stop accepting
+            // new requests on this connection and let the current one (e.g.
+            // the PUT keyshares whose receipt itself triggered the migration
+            // state change) finish cleanly.
```
kevindeforth: That seems dangerous. If the migration info changes, that is sufficient reason to cancel all incoming requests, as we might be communicating with a backup service that is no longer in use.

kevindeforth (changes requested):
Left a comment in the issue: https://github.com/near/mpc/issues/3123#issuecomment-4377732699

crates/e2e-tests/tests/migration_service.rs
```diff
@@ -330,7 +362,44 @@ async fn start_migration_and_wait(
...
+    (|| async {
+        let info = fetch_debug_migration_info(&http_client, &target_web_addr).await?;
+        let actual = info
+            .get(&source_account)
+            .and_then(|(_, destination)| destination.as_ref())
+            .map(|d| String::from(&d.destination_node_info.tls_public_key));
+        anyhow::ensure!(
+            actual.as_deref() == Some(target_p2p_key.as_str()),
+            "target debug endpoint destination_node_info.tls_public_key={actual:?}, expected {target_p2p_key}"
+        );
```
kevindeforth, line 387: Q: why are we doing string comparisons here?

kevindeforth (dismissed):
Thanks!

crates/e2e-tests/tests/migration_service.rs
anodar: Because one of them was already a string, I was returning string from `public_key(&self)` but for no good reason. Changed it.

kevindeforth (approved):

crates/e2e-tests/tests/migration_service.rs
```diff
@@ -330,7 +360,44 @@ async fn start_migration_and_wait(
...
+    let target_web_addr = match &cluster.nodes[target_idx] {
+        MpcNodeState::Running(n) => n.web_address(),
+        _ => bail!("target node not running"),
+    };
+    let source_account: AccountId = source_account_id
+        .parse()
+        .with_context(|| format!("invalid source account id: {source_account_id}"))?;
+    let http_client = reqwest::Client::new();
+    (|| async {
+        let info = fetch_debug_migration_info(&http_client, &target_web_addr).await?;
```
barakeinav1, line 379: nit: this wait block duplicates the structure of `register_backup_service_and_wait` (line 266) — same `fetch_debug_migration_info` → typed field check → `anyhow::ensure!` → retry pattern. Could be DRY'd into a shared helper, e.g.:

```rust
async fn wait_until_debug_migration_matches<F>(
    web_addr: &str,
    timeout: Duration,
    description: &str,
    check: F,
) -> anyhow::Result<()>
where
    F: Fn(&ContractMigrationInfo) -> Result<(), String>,
{ ... }
```

Each call site provides only the field-check closure. Not blocking, but team conventions favor extracting shared retry/poll helpers.

crates/e2e-tests/tests/migration_service.rs
```diff
@@ -330,7 +360,44 @@ async fn start_migration_and_wait(
             ),
     )
     .await
-    .context("timed out waiting for contract to reflect node migration")
+    .context("timed out waiting for contract to reflect node migration")?;
+
+    // Wait for the target's own indexer to ingest start_node_migration
```
barakeinav1, line 365: nit: the comment explains *why* we wait (avoid mid-PUT cancellation), but it'd help to also note that `destination_node_info.tls_public_key` is being used as a **proxy** for `MigrationInfo.active_migration`. They're correlated today because both come from indexing the same `start_node_migration` block, but if production logic ever decoupled them, the test could become flaky again without anyone noticing.

A one-line addition like:

```
// Note: we check destination_node_info.tls_public_key because it's populated by
// indexing the same start_node_migration block that flips active_migration to true.
```

…makes the intent clear and protects against silent regressions if the production logic changes.

crates/e2e-tests/tests/migration_service.rs
```diff
@@ -330,7 +360,44 @@ async fn start_migration_and_wait(
...
+            "target debug endpoint destination_node_info.tls_public_key={actual:?}, expected {target_p2p_key:?}"
+        );
+        Ok(())
+    })
+    .retry(
+        ConstantBuilder::default()
+            .with_delay(common::POLL_INTERVAL)
+            .with_max_times(
+                (INDEXER_SYNC_TIMEOUT.as_millis() / common::POLL_INTERVAL.as_millis()) as usize,
+            ),
```
barakeinav1, line 395: nit: even after this wait, there's still a residual race window between "test confirms target's view" and "PUT keyshares request completes." If *any* migration-info-affecting block lands during that window (e.g., a concurrent test or an unrelated migration), the migration service's cancellation token fires and tears down the connection.

In this test there's only one migration in flight, so the assumption holds. But it's worth a brief comment noting the assumption — if the test is ever extended to run alongside other migration activity, this assumption breaks silently.

barakeinav1 (commented):
Three non-blocking suggestions on the wait-for-indexer-sync change. Approve in principle.

barakeinav1 (approved):
thanks!

crates/e2e-tests/tests/migration_service.rs
kevindeforth: > If any migration-info-affecting block lands during that window (e.g., a concurrent test or an unrelated migration), the migration service's cancellation token fires and tears down the connection.

Oh, that would be a proper bug. We should only cancel in case migration info relevant to the node in question has changed.

PR #3126, opened 2026-05-05
test: flaky key_resharing test

### LINKED ISSUE #3137: Fix flaky test_key_resharing
### Background

In tests key_event_timeout_blocks from default 30 during contract initialization. On mainnet we produce blocks with ~1.66 block per second (once in ~600ms) whereas sandbox does ~8 blocks per second, so key resharing on mainnet has ~5x longer timeout. When there's enough resource contention this test fails because timeout is too short.

This change:

increases (8x) key_event_timeout_blocks to reduce (and hopefully eliminate) test flakiness
additionally allows tolerance on transient errors in test assertion while still checking that key resharing worked

### User Story

_No response_

### Acceptance Criteria

Increase key_event_timeout_blocks accordingly and additionally allow tolerance on transient errors to retry key resharing.

_No response_

anodar:
Increase `key_event_timeout_blocks` from default 30 to 240 blocks during contract initialization. On mainnet we produce blocks with ~1.66 block per second (once in ~600ms) whereas sandbox does ~8 blocks per second, so key resharing on mainnet has ~5x longer timeout.

This change:

-  increases (8x) `key_event_timeout_blocks` to reduce (and hopefully eliminate) test flakiness
- additionally allows tolerance on transient errors in test assertion while still checking that key resharing worked

Close: https://github.com/near/mpc/issues/3137


crates/e2e-tests/src/cluster.rs
```diff
@@ -970,8 +970,19 @@ async fn init_contract(
...
         "initializing contract"
     );
+    // The contract's default `key_event_timeout_blocks = 30` is ~18 s on
+    // mainnet (~600 ms blocks). The e2e sandbox runs ~8 blocks/s, so the
+    // same 30 collapses to ~3.7 s — too tight for the resharing
+    // round-trip (start_reshare → indexer × N → MPC compute →
+    // vote_reshared) once concurrent e2e tests put the runner under
+    // load. Override to 240 blocks (~30 s in sandbox) as a comfortable
+    // budget over mainnet's effective headroom.
+    let init_config = json!({ "key_event_timeout_blocks": 240u64 });
```
netrome: Also I'm questioning why we have this timeout in blocks rather than in wall clock time. The timeout should not be dependent on the block production pace afaik. It happens that block production speeds up or slows down for different reasons. For example, I think they were down to 1 block per minute on testnet last fall during an incident.

netrome (dismissed):
Just an unrelated reflection on this timeout.

DSharifi (dismissed):

DSharifi:
@claude review

crates/e2e-tests/src/cluster.rs
DSharifi: The wall clock can have drift of up to 2 minutes, so it's not reliable to use that on chain

DSharifi (dismissed):

crates/e2e-tests/src/cluster.rs
netrome: Oh, 2 minutes for a chain that posts blocks every second. That's interesting. The question then is if it's less reliable than block count? :thinking:

DSharifi (approved):

netrome (approved):

PR #3132, opened 2026-05-05
refactor: RecentBlocksTracker use weak pointers for parent

### LINKED ISSUE #3114: RecentBlockTracker: Use `Weak<BlockNode>` for pointing to partent node
### Background
In #3108, we shipped a fix for pruning cycles. The root cause is because in the tree stored in `RecentBlocksTracker`, we use `Arc<BlockNode>` for pointing to parents and children of nodes.

### User Story

As a dev, I like it when the type system takes care of things for me.

### Acceptance Criteria

`Weak<BlockNode>` for `parent` in `RecentBlocksTracker`

### Additional Resources
Do you know why we're using `Arc` for both `children` and `parent`? IIUC the more natural choice would be `Weak<BlockNode>` for `parent`. Trees and DLLs in Rust are usually implemented so node back-edges aren't owning — sometimes via `Weak` ([html5ever's `RcDom`](https://github.com/servo/html5ever/blob/201534ee4f4c3db10dc558869555768d864278bf/rcdom/lib.rs#L109-L117)), sometimes via raw `NonNull` ([`std::BTreeMap`'s `parent`](https://github.com/rust-lang/rust/blob/main/library/alloc/src/collections/btree/node.rs#L53), or both ends in [`std::LinkedList`](https://github.com/rust-lang/rust/blob/master/library/alloc/src/collections/linked_list.rs#L50-L66) with ownership [held by the container](https://github.com/rust-lang/rust/blob/1d72d7e8136faaebad3a85eeed432e6ea1b2ffab/library/alloc/src/collections/linked_list.rs#L58) via `PhantomData<Box<Node<T>>>`).

_Originally posted by @pbeza in https://github.com/near/mpc/pull/3108#pullrequestreview-4221546389_

kevindeforth:
resolves #3114 

Using a weak ref instead of an `Arc` allows us not only to leverage the type system cleanup process, but also:
- to get rid of a no longer necessary mutex.
- to distinguish between _"The parent of this node has never been seen by `RecentBlocksTracker`"_ and _"The parent of this node was removed by `RecentBlocksTracker`"_. This will be useful for a future refactor.

### DIFF
--- a/crates/node/src/requests/recent_blocks_tracker.rs
+++ b/crates/node/src/requests/recent_blocks_tracker.rs
@@ -5,7 +5,7 @@ use std::collections::HashMap;
 use std::fmt::Debug;
 use std::ops::Add;
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, Weak};
 
 /// Tracks the topology of the recent blocks, using the blocks given by the indexer.
 ///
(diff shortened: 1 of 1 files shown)


kevindeforth:
> * Adds `prune_old_blocks__should_drop_connected_pruned_block_nodes` which captures `Weak`s to two pruned nodes connected via a parent/children chain and asserts both are dropped after a prune jump.

Claude, you are confused. This was added in a previous PR.

barakeinav1 (approved):
LGTM

pbeza (approved):

PR #3133, opened 2026-05-05
test(perf): measure testnet signing performance using devnet

gilcu3:
Part of https://github.com/near/mpc-private/issues/320

### DIFF
--- a/crates/devnet/scripts/loadtest-scenarios.sh
+++ b/crates/devnet/scripts/loadtest-scenarios.sh
@@ -1,0 +1,131 @@
+#!/usr/bin/env bash
+# Drive `mpc-devnet loadtest run` through a fixed set of load-shape scenarios
+# against an MPC contract (defaults to the testnet `v1.signer-prod.testnet`).
+#
+# Each scenario is just a `mpc-devnet loadtest <name> run` invocation with a
+# different `--qps` / `--duration`.  Output is streamed to stdout — pipe to
+# `tee` yourself if you want to keep a transcript.
+#
+# Prerequisites:
+#   * `mpc-devnet` is installed (see crates/devnet/README.md).
+#   * A loadtest setup has been created with `mpc-devnet loadtest <name> new ...`
+#     and has enough access keys to sustain the highest QPS used below
+#     (a single account with 16 keys is usually enough up to ~20 QPS).
+#   * The parallel-sign helper contract is deployed
+#     (`mpc-devnet loadtest <name> deploy-parallel-sign-contract`).
+#
+# Usage:
+#   ./loadtest-scenarios.sh <loadtest-name> [options]
+#
+# Options:
+#   --contract <id>    MPC contract to target (default: v1.signer-prod.testnet)
+#   --domain <id>      Domain id to sign against (default: 0)
+#   --batch <n>        Sigs per parallel-sign tx (default: 10). Each scenario's
+#                      `--qps` is the target signature rate; the loadtest sends
+#                      `qps/batch` RPC txs/sec, so a higher batch lowers RPC
+#                      pressure for the same signature throughput.
+#   --scenario <name>  One of: sustained, high, spike, all (default: all)
+#   -h, --help         Show this message
+
+set -euo pipefail
+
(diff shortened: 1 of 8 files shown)


crates/devnet/src/devnet.rs
```diff
@@ -28,17 +30,25 @@ impl OperatingDevnetSetup {
...
+            client.with_retry(5, |rpc| {
+                async move {
+                    rpc.call(methods::block::RpcBlockRequest {
+                        block_reference: BlockReference::Finality(Finality::Final),
+                    })
+                    .await
+                    .map_err(anyhow::Error::from)
+                }
+                .boxed()
             }),
```
pbeza, line 47: Could RPC rate limiting skew the test results if we get rate-limited and keep retrying here?

pbeza (dismissed):

crates/devnet/src/rpc.rs
```diff
@@ -25,7 +25,12 @@ struct NearRpcClient {
 
 impl NearRpcClient {
     fn new(config: RpcConfig) -> Self {
-        let client = JsonRpcClient::connect(config.url);
+        let client = JsonRpcClient::connect(&config.url);
+        let client = match config.api_key {
+            Some(key) => client
+                .header(auth::Authorization::bearer(key).expect("api_key is not a valid token")),
+            None => client,
+        };
```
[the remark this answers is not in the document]
gilcu3: this is just for testing so I think it is fine

crates/devnet/src/contracts.rs
```diff
@@ -135,7 +149,7 @@ pub fn make_actions(call: ContractActionCall) -> ActionCall {
                     },
                 })
                 .unwrap(),
-                300,
+                CKD_TGAS,
```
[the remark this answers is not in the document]
gilcu3: yeah this is pre-existing, the load test should probably have both versions of CKD, did not want to get this done here

crates/devnet/src/devnet.rs
gilcu3: probably yes, although I did not observe RPC failures [yesterday](https://github.com/near/mpc-private/issues/320#issuecomment-4399081147)

crates/devnet/scripts/loadtest-scenarios.sh
netrome, line 1: Hmm thinking about the placement of this script. Perhaps this makes sense, though we do have the top-level `scripts` folder as well which might also be a good place for this - although that folder is quite messy 🤔

crates/devnet/scripts/loadtest-scenarios.sh
```diff
@@ -0,0 +1,131 @@
...
+        --mpc-contract "$CONTRACT" \
+        --parallel-sign-calls-per-domain "${DOMAIN}=${BATCH}" \
+        --qps "$qps" \
+        --duration "$duration"
+    echo "[$(date -Is)] <<< scenario '$label' done"
+    echo
+}
+
+# Acceptance criterion from near/mpc-private#320 is "sustained load of at
+# least 100 transactions per minute".  2 sigs/s = 120 tpm; 10 minutes gives
```
netrome: Feels weird to refer to the acceptance criteria here. Also we should not refer to a private repo in our public code base.

```suggestion
# 2 sigs/s = 120 tpm; 10 minutes gives
```

netrome (changes requested):
Small blocker, we should not refer to mpc-private in this code. Otherwise LGTM, thanks!

crates/devnet/scripts/loadtest-scenarios.sh
gilcu3: ups this slept through :|

crates/devnet/scripts/loadtest-scenarios.sh
gilcu3: the problem is that the mpc-devnet binary also assumes it is launcher from `crates/devnet/` so having the script elsewhere would also be weird

netrome (approved):
Thanks!

pbeza (approved):

PR #3128, opened 2026-05-05
ci: enforce all `.editorconfig` rules via `editorconfig-checker`

pbeza:
Closes #3127

## Summary

- Adds `editorconfig-checker` to the Nix dev shell.
- Replaces `scripts/check-trailing-whitespace.sh` (added in #3112, one rule) with a new `[tasks.editorconfig]` cargo-make task that runs `editorconfig-checker` on `git ls-files -z` output. Catches violations of every rule declared in `.editorconfig`: `trim_trailing_whitespace`, `insert_final_newline`, `end_of_line = lf`, `charset = utf-8`.
- Cleans up 32 tracked files missing a final newline.
- Extends `.editorconfig` exemptions for files that genuinely can't have a final newline / LF (`third-party-licenses/licenses.html` ships CRLF from `cargo about`; `crates/threshold-signatures/LICENSE` is vendored; four byte-exact test fixtures in `crates/test-utils/assets/` consumed via `include_str!`).

`git ls-files -z | xargs -0 editorconfig-checker` rather than `editorconfig-checker .` — the latter walks the working tree and produces false positives for contributor-local artefacts (`.ruff_cache/`, `.pytest_cache/`, etc.).

PR #3140, opened 2026-05-05
chore(editorconfig): drop LICENSE exemption, add final newline

pbeza:
Follow-up to #3128, addressing https://github.com/near/mpc/pull/3128#discussion_r3189531378.

## Summary

- Drop the `crates/threshold-signatures/LICENSE` exemption from `.editorconfig`.
- Append a final newline to the file so it satisfies `insert_final_newline = true`.


pbeza:
Reopening from a clean branch — see follow-up PR.

PR #3146, opened 2026-05-06
chore: remove mpc-contract dep from devnet

### LINKED ISSUE #658: Unify `rand` dependency version in workspace and `devnet` crate.

### LINKED ISSUE #3142: Remove mpc-contract dependency from devnet
### Background

`mpc-devnet` depends on `mpc-contract` and reaches into contract-internal types (`ProtocolContractState`, `ParticipantInfo`, etc.). This couples devnet to the on-chain implementation and means it misses backwards-compatibility behavior that lives only in `near-mpc-contract-interface` DTOs (e.g. the `#[serde(alias = "sign_pk")]` on `ParticipantInfo::tls_public_key`).

Mirrors #381 for the node side.

### Acceptance Criteria

- `mpc-contract` is removed from `crates/devnet/Cargo.toml`.
- All current usages route through `near-mpc-contract-interface`.

gilcu3:
Closes #3142

Also closes #658 coincidentally


crates/devnet/Cargo.toml
```diff
@@ -6,25 +6,26 @@ edition = "2021"
...
 near-mpc-contract-interface = { workspace = true, features = [
+    "blstrs",
     "ed25519-dalek",
     "near",
 ] }
 near-sdk = { workspace = true }
 node-types = { workspace = true }
-# TODO(#658): use workspace version
-rand = "0.9.3"
+rand = { workspace = true }
```
gilcu3, line 28: this change was needed to solve an incompatibility problem with generating CKD public keys

pbeza (approved):
Approved, although IIUC we now copy-pasted some code from the contract crate, e.g.:

https://github.com/near/mpc/blob/4b01c65ac3a665e9c6d5abc7a31854917a47e721/crates/contract/src/state.rs#L304-L329

vs

https://github.com/near/mpc/blob/ad0f11f212b1390626da911dca3588878c464610/crates/devnet/src/loadtest.rs#L448-L456

Would it make sense to have a common shared place (crate?) for some of these helper functions?

gilcu3:
> Approved, although IIUC we now copy-pasted some code from the contract crate, e.g.:
> https://github.com/near/mpc/blob/ad0f11f212b1390626da911dca3588878c464610/crates/devnet/src/loadtest.rs#L448-L456
> 
> Would it make sense to have a common shared place (crate?) for some of these helper functions?

The problem is that code exists in the internal type of the contract, but not on the dto. So unless the types are unified, we would still live with code duplication.

crates/devnet/src/mpc.rs
```diff
@@ -339,22 +329,25 @@ impl MpcInitContractCmd {
...
-                    mpc_account_to_participant_info(setup.accounts.account(account_id), i),
-                )
-                .unwrap();
+            let info = mpc_account_to_participant_info(setup.accounts.account(account_id), i);
+            participant_entries.push((account_id.clone(), next_id, info));
+            next_id = next_id.next();
         }
-        let parameters =
-            ThresholdParameters::new(participants, Threshold::new(self.threshold)).unwrap();
+        let parameters = ThresholdParameters {
```
barakeinav1, line 344: The old code used `Participants::new()` + `.insert()` from the contract crate, which enforced invariants (no duplicate accounts/IDs, ID assignment correct). The new code constructs `Participants` directly from a Vec with no validation — duplicate accounts or out-of-sync `next_id` could silently slip through.

For devnet tooling this is probably acceptable (only the operator constructs these), but it's a subtle safety regression worth being aware of. Same pattern in `MpcVoteNewParametersCmd` (line 725) where `.insert()` becomes manual `.push()`.

Could either:
- Add a sanity check after manual construction (e.g., `assert!(no duplicates)`)
- Or factor the construction into a small helper that mimics `Participants::insert`'s validation

Not blocking, but flagging.

crates/devnet/src/contracts.rs
```diff
@@ -199,6 +203,11 @@ fn make_payload(protocol: Protocol) -> Payload {
     }
 }
 
+fn random_app_public_key() -> Bls12381G1PublicKey {
```
barakeinav1, line 206: `random_app_public_key` is now duplicated — this local copy plus the one in `mpc_contract::utils` (still needed there). They're equivalent today (`G1Projective::random` and `generator() * Scalar::random` are the same algorithm under the hood), but independent implementations can drift.

If `mpc_contract::utils::random_app_public_key` will be retained for the contract crate's own tests, consider moving it to a shared crate (`mpc-primitives` or test utilities) so both consumers reference the same implementation. Not blocking; just noting the duplication.

barakeinav1 (commented):
Two non-blocking observations on the dependency removal.

barakeinav1 (approved):
LGTM. Devnet now routes through interface DTOs, matching the #381 effort on the node side. Posted two non-blocking observations.

Approving.

BTW — who's the author of #658?

crates/devnet/src/mpc.rs
gilcu3: yeah, Claude mentioned it above. I think for devnet is fine not to have all invariants. If they are broken we will just notice it later.

crates/devnet/src/contracts.rs
gilcu3: yeah, I think we have it in some other places as well. Not sure if it is worth having a new crate just for it, given that it is a relatively simple operation

gilcu3:
> LGTM. Devnet now routes through interface DTOs, matching the #381 effort on the node side. Posted two non-blocking observations.
> 
> Approving.
> 
> BTW — who's the author of #658?

Thanks! #658 was created by Daniel using his old account that was then deleted :)

PR #3147, opened 2026-05-06
test: add e2e for contract upgrade compatibility

### LINKED ISSUE #2970: Add e2e test for contract/node upgrade compatibility

gilcu3:
Closes #2970


crates/e2e-tests/src/cluster.rs
```diff
@@ -1140,6 +1259,50 @@ async fn create_user_accounts(
...
+#[derive(borsh::BorshSerialize)]
+struct ProposeUpdateArgsBorsh<'a> {
+    code: Option<&'a [u8]>,
+    config: Option<()>,
+}
+
+/// JSON mirror of the contract's `UpdateId`. `propose_update` returns it and
+/// `vote_update` consumes it via the `id` field; both wire it as a bare u64.
+#[derive(Debug, Clone, Copy, serde::Serialize, serde::Deserialize)]
+struct UpdateId(u64);
```
gilcu3, line 1319: these two could be de-duplicated, but that would require a refactor. If you agree I can add a todo for a follow-up

crates/e2e-tests/src/cluster.rs
```diff
@@ -36,6 +36,10 @@ const SIGN_GAS: near_kit::Gas = near_kit::Gas::from_tgas(15);
 // which costs significantly more than a plain CKD or sign request.
 const CKD_PV_GAS: near_kit::Gas = near_kit::Gas::from_tgas(100);
 const SIGN_DEPOSIT: near_kit::NearToken = near_kit::NearToken::from_yoctonear(1);
+// Mirrors `CURRENT_CONTRACT_DEPLOY_DEPOSIT` from the contract sandbox tests.
```
kevindeforth: Q: How is this comment relevant?

crates/e2e-tests/src/cluster.rs
```diff
@@ -36,6 +36,10 @@ const SIGN_GAS: near_kit::Gas = near_kit::Gas::from_tgas(15);
...
+// The deposit must cover storage for the contract WASM payload (~17 NEAR for
+// current binaries); the contract panics if the attached deposit is short.
```
kevindeforth: nit: This risks to become a stale comment. The `~17 Near` does not add relevant information, as it's implied in the constant defined just underneath
 
```suggestion
// The deposit must cover storage for the contract WASM payload;
// the contract panics if the attached deposit is short.
```

crates/e2e-tests/src/cluster.rs
```diff
@@ -79,6 +83,45 @@ pub struct MpcClusterConfig {
     /// cluster so their indexers sync before blocks accumulate (`start_near_node`
     /// blocks until synced).
     pub migration_targets: Vec<usize>,
+    /// Wire format used when calling `init` on the contract. Defaults to
+    /// `Current`; switch to a `Legacy*` variant when `contract_wasm` is an
+    /// older production binary whose `init` signature is incompatible with
+    /// the current `ThresholdParameters` JSON.
```
kevindeforth: nit: I have a dislike for comments of this sort. If we change the default value of `ContractInitFormat`, we also need to change this comment, which is very easy to miss.

crates/e2e-tests/src/cluster.rs
```diff
@@ -79,6 +83,45 @@ pub struct MpcClusterConfig {
...
+/// - **When a new wire-breaking change to `init` lands**: add a new variant
+///   (e.g. `Legacy3_10_X`) that emits the now-old shape, and update the
+///   `init_contract` helper in `cluster.rs` to branch on it.
+/// - **After the breaking change has rolled out to Mainnet/Testnet**: remove
+///   the obsolete variant and any tests that pin to it. The `current_*()`
+///   pointers in `contract-history` will already reference a binary that
+///   speaks the new format, so `Current` is enough.
+/// - **`init_format` should always be `Current`** for tests that deploy the
+///   *current* contract WASM (the default). Only override it when the test
+///   pins `contract_wasm` to an older `contract_history::*` binary.
```
kevindeforth: Why do I need to "override" it? Can't I directly select an older version?

crates/e2e-tests/src/cluster.rs
```diff
@@ -835,6 +880,73 @@ impl MpcCluster {
...
+            config: None,
+        };
+        let proposer_client = self.operator_client_for(0)?;
+        let outcome = self
+            .contract
+            .call_from_borsh_with_deposit(
+                &proposer_client,
+                method_names::PROPOSE_UPDATE,
+                propose_args,
+                near_kit::Gas::from_tgas(300),
```
kevindeforth: nit: lets make a constant for this.

crates/e2e-tests/src/cluster.rs
```diff
@@ -835,6 +880,73 @@ impl MpcCluster {
...
+
+        // `propose_update` is borsh-deserialized on the contract side. We use
+        // a local mirror struct (`ProposeUpdateArgsBorsh`) to avoid pulling
+        // the contract crate into the e2e-tests dependency tree for two
+        // fields.
+        let propose_args = ProposeUpdateArgsBorsh {
+            code: Some(new_wasm),
+            config: None,
+        };
+        let proposer_client = self.operator_client_for(0)?;
```
kevindeforth: nit: lets make a constant for the index of the proposer

crates/e2e-tests/src/cluster.rs
```diff
@@ -835,6 +880,73 @@ impl MpcCluster {
...
+    pub async fn propose_and_vote_contract_update(&self, new_wasm: &[u8]) -> anyhow::Result<()> {
+        anyhow::ensure!(
+            !self.nodes.is_empty(),
+            "cannot propose contract update with no nodes"
+        );
+
+        // `propose_update` is borsh-deserialized on the contract side. We use
+        // a local mirror struct (`ProposeUpdateArgsBorsh`) to avoid pulling
+        // the contract crate into the e2e-tests dependency tree for two
+        // fields.
```
kevindeforth: nit: I am not sure this comment is needed. The code reads very cleanly already.

crates/e2e-tests/src/cluster.rs
```diff
@@ -835,6 +880,73 @@ impl MpcCluster {
...
+            outcome.is_success(),
+            "propose_update failed: {:?}",
+            outcome.failure_message()
+        );
+        let proposal_id: UpdateId = outcome
+            .json()
+            .context("propose_update did not return a JSON UpdateId")?;
+
+        // Cast votes one-by-one. `vote_update` returns `true` once the
+        // threshold has been reached and the new code is in place.
```
kevindeforth: > // and the new code is in place.

This is not accurate. It returns true as soon as the threshold is reached, independent if the code deployment is successful or not.

crates/e2e-tests/src/cluster.rs
```diff
@@ -835,6 +880,73 @@ impl MpcCluster {
...
+            anyhow::ensure!(
+                vote_outcome.is_success(),
+                "vote_update from node {i} failed: {:?}",
+                vote_outcome.failure_message()
+            );
+            let update_applied: bool = vote_outcome
+                .json()
+                .with_context(|| format!("vote_update from node {i} returned non-bool"))?;
+            if update_applied {
+                tracing::info!(votes = i + 1, "contract code update applied");
```
kevindeforth: Would be nice to add an assert here that the number of votes corresponds to the expected threshold. Or return this and do it in the caller.

crates/e2e-tests/src/cluster.rs
```diff
@@ -961,6 +1074,7 @@ async fn init_contract(
     threshold: usize,
     participant_indices: &[usize],
     ports: &E2ePortAllocator,
+    init_format: ContractInitFormat,
```
kevindeforth: nit: might be worth to make a struct for these args

crates/e2e-tests/src/cluster.rs
```diff
@@ -971,10 +1085,17 @@ async fn init_contract(
     tracing::info!(
         threshold,
         num_participants = participant_indices.len(),
+        ?init_format,
         "initializing contract"
     );
+    let parameters_json = match init_format {
+        ContractInitFormat::Current => serde_json::to_value(&params)?,
+        ContractInitFormat::Legacy3_9_1 => {
+            serde_json::to_value(LegacyThresholdParameters::from(&params))?
```
kevindeforth: Did you consider something like this?

```suggestion
    let parameters_json = init_format.make_json(&params);
```

crates/e2e-tests/src/cluster.rs
```diff
@@ -1140,6 +1261,70 @@ async fn create_user_accounts(
...
+        Self {
+            threshold: params.threshold,
+            participants: LegacyParticipants {
+                next_id: params.participants.next_id,
+                participants,
+            },
+        }
+    }
+}
+
```
kevindeforth: nit: I think I would move these to a new file and probably module.

crates/e2e-tests/src/cluster.rs
```diff
@@ -79,6 +83,45 @@ pub struct MpcClusterConfig {
...
+///   `init_contract` helper in `cluster.rs` to branch on it.
+/// - **After the breaking change has rolled out to Mainnet/Testnet**: remove
+///   the obsolete variant and any tests that pin to it. The `current_*()`
+///   pointers in `contract-history` will already reference a binary that
+///   speaks the new format, so `Current` is enough.
+/// - **`init_format` should always be `Current`** for tests that deploy the
+///   *current* contract WASM (the default). Only override it when the test
+///   pins `contract_wasm` to an older `contract_history::*` binary.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Default)]
+pub enum ContractInitFormat {
```
kevindeforth, line 107: I am curious how we would extend this if we change other function call arguments.
Adding one of these for every argument that changes, and ensuring it is called at the right place would be a bit cumbersome.

kevindeforth (approved):
Thanks a lot!
It does what it's supposed to do. I have quite a few nits, things to consider:
- I found some of the comments distracting and not adding much value. Comments may be helpful, but in general, I much prefer to have the code be self-explanatory.
- The current PR only provides a code path for changing arguments to the init method call. It is not unlikely that we will have other contract calls that will require different argument formats in the future. If we follow this, similar changes will be required for each one of those, which seems like a lot of boilerplate.

That being said, the above are not blocking issues for me, though the second point is something we probably want to align on before merging.

crates/e2e-tests/tests/contract_upgrade_compatibility.rs
```diff
@@ -0,0 +1,89 @@
...
+    common::send_ckd_request(&cluster, &running, &mut rng, &user)
+        .await
+        .expect("ckd request failed against production contract");
+
+    // When: we propose and vote in an update to the current contract WASM.
+    cluster
+        .propose_and_vote_contract_update(&current_contract_wasm)
+        .await
+        .expect("contract upgrade to current WASM failed");
+
```
kevindeforth, line 74: oh, actually, here is a blocker. We should compare the actual with the expected hash of deployed code. Otherwise, we risk missing a failed migration.

kevindeforth (changes requested):
Realized we never check that the update was successful. This is a blocking issue.

crates/e2e-tests/src/cluster.rs
gilcu3: IMO we should have a single value, so it was about mentioning where it came from. Dont mind removing though

crates/e2e-tests/src/cluster.rs
gilcu3: true, the `ThresholdParameters` should not be mentioned there (I guess I missed it), will remove

crates/e2e-tests/src/cluster.rs
gilcu3: The way it is now `Current` is the default. Only in case the old one differs (that hopefully is not a frequent situation) then we need to define a new one, and therefore override `Current`

crates/e2e-tests/src/cluster.rs
gilcu3: I can remove it, hopefully no one adds the contract dependency again for this reason :smiling_face_with_tear:

crates/e2e-tests/src/cluster.rs
gilcu3: yep, that's what I forgot and led to the blocker, will fix

crates/e2e-tests/src/cluster.rs
gilcu3: seems better to do it here

crates/e2e-tests/src/cluster.rs
gilcu3: not really, but does look slightly better, will take it :)

crates/e2e-tests/src/cluster.rs
gilcu3: I hesitated a bit on it. It is true that cluster.rs is getting a bit bigger, will move to a new file

crates/e2e-tests/src/cluster.rs
gilcu3: I will try to apply something similar to your idea on slack:


> One thing that might already be an improvement is to store the not just the deployed contract in MpcCluster:
```rust
pub contract: DeployedContract,
```
> but to store a DeployedMpcContract struct:
```rust
pub contract: DeployedMpcContract
```
> where we have something like:
```rust
pub struct DeployedMpcContract {
    contract: DeployedContract,
    version: MpcContractVersion
}
```
> Then, we can slowly add methods to it:
```rust
impl DeployedMpcContract {
    fn make_init_args(...) -> ,
}
```

crates/e2e-tests/tests/contract_upgrade_compatibility.rs
gilcu3: true, will fix

gilcu3: fixed in https://github.com/near/mpc/pull/3147/commits/7806aa2c096a80ba49842d5c3cbdde44676615c3

crates/e2e-tests/src/cluster.rs
gilcu3: Opened a follow up, as the rename suggested in the thread will touch many more files https://github.com/near/mpc/issues/3156

kevindeforth (dismissed):
Nice, thank you!

crates/e2e-tests/tests/contract_upgrade_compatibility.rs
```diff
@@ -0,0 +1,93 @@
+//! Verifies that current `mpc-node` binaries remain functional when paired
+//! with the contract WASM currently deployed on Mainnet/Testnet, then
+//! continue working after the contract is upgraded to the current build.
+
+#![expect(non_snake_case)]
```
anodar, line 5: nit: consider moving this to where it's expected instead of doing it over the whole file.

crates/e2e-tests/src/cluster.rs
```diff
@@ -835,6 +891,98 @@ impl MpcCluster {
...
+                );
+                tracing::info!(votes = i + 1, "contract code update vote threshold reached");
+                return Ok(());
+            }
+        }
+        anyhow::bail!("contract code update was not applied after votes from every node")
+    }
+
+    /// Wait until the deployed contract code hash matches `sha256(expected_wasm)`.
+    pub async fn assert_deployed_code(&self, expected_wasm: &[u8]) -> anyhow::Result<()> {
```
anodar: This method doesn't seem panic, only returns an error, but name assert_* comes off as if it could panic. Consider changing to ensure_*

Related discussions about this convention in https://github.com/near/mpc/pull/3017

anodar (dismissed):

crates/e2e-tests/tests/contract_upgrade_compatibility.rs
gilcu3: yep, this is a good point. The problem was `rstest` which does not play nice with clippy and this. If I move it to the function, clippy just complains :) So this is just a workaround

kevindeforth (approved):

anodar (approved):
Thank you!

PR #3151, opened 2026-05-06
build: use x86-64-v3 instruction set

### LINKED ISSUE #3150: use x86-64-v3 instruction set for node binary
### Background

Production nodes all run on hardware supporting x86-64-v3 (Intel Haswell / AMD Excavator and newer, ~2013+), but until now the binary was compiled at the original AMD64 baseline (2003 ISA). That meant rustc and gcc couldn't emit AVX2, BMI2, FMA, F16C, or MOVBE for the bulk of the workspace — only crates that hand-roll runtime feature dispatch (ring, blake3, sha2, curve25519-dalek, ...) were using AVX2. Flamegraphs from production confirm this: AVX2 frames are present, but only inside crypto libraries that opt in at function granularity via #[target_feature(enable = "avx2")].

Bumping the global target-cpu to x86-64-v3 lets the compiler:

Emit AVX2/BMI2 across the whole binary, including coordinator/network/indexer code that has no hand-written SIMD
Auto-vectorize loops that LLVM previously left scalar
Drop the runtime-dispatch branches in libraries that have v3 fast paths (small but nonzero)
Use BMI2's mulx/pdep/pext and FMA in arithmetic-heavy code (relevant for the threshold-ECDSA hot path)

### User Story

_No response_

### Acceptance Criteria

- mpc-node binary is built with x86-v3 instruction set for reproducible builds
- CI is running tests with this instruction set as well

_No response_

DSharifi:
closes #3150

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -80,13 +80,22 @@
 
           isX86 = stdenv.hostPlatform.isx86_64;
 
+          # Production ISA: x86-64-v3 plus PCLMUL and AES. The v3 micro-arch
+          # level (per System V psABI) covers AVX2/BMI2/F16C/FMA/LZCNT/MOVBE
+          # but NOT PCLMUL or AES — we add those explicitly so rocksdb's
+          # PCLMUL-accelerated CRC32C path is compiled in. See
+          # .cargo/config.toml, nix/mpc-node.nix, and deployment/build-images.sh
+          # for the matching production-build settings. Host fleet is all
+          # v3-capable (Haswell / Excavator and newer).
+          prodCFlags = "-march=x86-64-v3 -mpclmul -maes";
+
           envCommon = {
-            # Match the production ISA baseline (see .cargo/config.toml and
-            # nix/mpc-node.nix) so C/C++ deps (rocksdb, snappy, zstd, jemalloc)
-            # built for tests use the same target-cpu as the shipped binary.
-            # Production nodes all run on x86-64-v3-capable hardware.
-            CFLAGS = lib.optionalString isX86 "-march=x86-64-v3";
-            CXXFLAGS = "-include cstdint" + lib.optionalString isX86 " -march=x86-64-v3";
+            # `-include cstdint` is needed by neard's rocksdb C++ build
+            # regardless of host. Production ISA flags are scoped to the
+            # x86_64 Linux host target below so wasm cross-compilation
+            # (e.g. the contract WASM build via blst) isn't polluted —
+            # `-march=x86-64-v3` is invalid for the wasm32 target.
+            CXXFLAGS = "-include cstdint";
 
             # WASM Toolchain
             CC_wasm32_unknown_unknown = "${llvmPkgs.clang-unwrapped}/bin/clang";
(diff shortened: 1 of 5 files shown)


DSharifi:
@claude review

DSharifi:
> **Non-blocking** (suggestions / things to keep in mind):
> 
> * `.cargo/config.toml:9` — Plain `cargo build` outside Nix/repro-env (e.g. `nightly_build.yml`'s `cargo build -p mpc-node`) now gets `-C target-cpu=x86-64-v3` for Rust but cc-crate deps (rocksdb/snappy/zstd/jemalloc) still compile with the host default `-march`. Cargo doesn't propagate config-file rustflags to env vars. Not a runtime hazard (the v3 Rust binary still runs fine on a v3 host built against a v1 rocksdb), and it doesn't matter for reproducible artefacts since those go through repro-env or Nix — just an asymmetry to be aware of when comparing local-build perf to production.

This is fine. CI runs with the nix profile.

> * `docs/nix-dev-environment.md:8` — The note correctly warns about pre-v3 hardware running with `SIGILL`. Worth confirming the WarpBuild `warp-ubuntu-2404-x64-*` runners used by `ci.yml`/`nightly_build.yml` are all v3-capable; if any older instance class slips into the pool, every CI job that runs the resulting binary (unit tests, e2e) would crash on startup. Almost certainly fine in practice (Cascade Lake / EPYC Rome+), but a single sentence in the PR description confirming you've checked would help future maintainers.

Confirmed warp runners support v3

> * `flake.nix:111-132` — Worth a quick smoke test on `aarch64-darwin` (Apple Silicon) that the dev shell still builds the contract WASM and that `cargo build` for the host works, since this PR moves a non-trivial amount of env-var plumbing.
> 

Sanity checked that building the contract and the node, plus running tests work as expected.

DSharifi:
reproducibility is intact comparing local desktop to CI:

Nix :heavy_check_mark: 
```
❯ nix build .#mpc-node && sha256sum ./result/bin/mpc-node
944dcec76a677db6d019fbdd8f2fd1fa758f506916965ae49ddaa237e26672c1  ./result/bin/mpc-node
```

repro-env: :heavy_check_mark: : : 
Local;
```
commit hash: 37539bae5952926fe31bcd9c17a255a8091823a4
SOURCE_DATE_EPOCH used: 0
node binary hash: 5dc06ff7c4646f172a60f3e1feede18b1efe0424bc324d15db98b6b1834e2e07
node manifest digest: sha256:dc966ed5400f60a09a5790e0227a03db52f6b135d1b138c8420e62586ded113f
```


CI:
```
commit hash: 37539bae5952926fe31bcd9c17a255a8091823a4
SOURCE_DATE_EPOCH used: 0
node binary hash: 5dc06ff7c4646f172a60f3e1feede18b1efe0424bc324d15db98b6b1834e2e07
node manifest digest: sha256:dc966ed5400f60a09a5790e0227a03db52f6b135d1b138c8420e62586ded113f
```

netrome (approved):

flake.nix
```diff
@@ -93,11 +109,17 @@
...
+            # `__FMA__`, `__BMI2__`, `__PCLMUL__`, `__AES__`) so rust-bindgen
+            # generates the same Rust bindings as `nix/mpc-node.nix` does.
+            BINDGEN_EXTRA_CLANG_ARGS = lib.concatStringsSep " " (
+              [
+                "-I${clangResourceInclude llvmPkgs}"
+                "-I${libcDev}/include"
+                "-fno-stack-protector"
+              ]
+              ++ lib.optional isX86 prodCFlags
+            );
```
gilcu3, line 122: is `fno-stack-protector` needed? And is it in the other build `mpc-node`?

gilcu3 (approved):
LGTM

flake.nix
DSharifi: It's orthogonal to this PR. It was originally added way back because it failed to build neard for me, but this can have changed.

Issue #3070, opened 2026-04-29
`PendingRequests` queue should keep track of block content


kevindeforth:
### Background

Currently, `RecentBlocksTracker` is a generic struct, keeping track of block contents

https://github.com/near/mpc/blob/f830d890be90527b312b4c5b3b9338fb26972aef/crates/node/src/requests/recent_blocks_tracker.rs#L70

That's not ideal. It forces us to have one instance of `RecentBlocksTrcker` per `PendingRequests` queue

https://github.com/near/mpc/blob/f830d890be90527b312b4c5b3b9338fb26972aef/crates/node/src/requests/queue.rs#L73

A better way would be to keep track of the block contents on the queue side. This will allow us to share the `RecentBlocksTracker` instance between all queues.

### User Story

As a dev, I like clear separation of concerns and responsibilities.

### Acceptance Criteria

`RecentBlocksTracker` should only track the block headers, it should not track block content.
note: the same signature request / response can appear in multiple blocks. This implementation should account for that, or handled in a follow-up issue.

Issue #3135, opened 2026-05-05
`RecentBlocksTracker`: replace `classify_blocks` with atomic u8


kevindeforth:
### Background

`RecentBlocksTracker` classifies blocks using `classify_block` method:

https://github.com/near/mpc/blob/74476c3d1a256efacbd224ea07ec44bf6a3b75f6/crates/node/src/requests/recent_blocks_tracker.rs#L413-L448

Besides a comparison to the head of the canonical and finalized chain stored by the recent blocks tracker, it relies on  two atomic bools of each Node:

https://github.com/near/mpc/blob/74476c3d1a256efacbd224ea07ec44bf6a3b75f6/crates/node/src/requests/recent_blocks_tracker.rs#L106-L119

It would be a nicer API if the block itself could tell us its finality status, without requiring external data like the height of the final chain. This is absolutely possible, as there is only one instant at which a block's finality status might change, that is when new blocks are added. It is at that point, that other blocks may get removed form the tracker or be finalized.

This would allow a consumer to only store a pointer to the block, not requiring a reference to the full `RecentBlocksTracker` to determine the finality status of a block.

We could replace the two atomic bools by a single atomic u8.

### User Story

As a dev, I like efficient and easy to reason code.

### Acceptance Criteria

Replace the two atomic bools in `BlockNode` with a single atomic u8 for its finality status.
note: this will address https://github.com/near/mpc/pull/3159#discussion_r3209795803, but linking for reference.

PR #3159, opened 2026-05-06
fix: `RecentBlocksTracker` removes dead blocks instead of classifying them as `OptimisticAndCanonical`

### LINKED ISSUE #3136: Remove `NotIncluded` variant from `RecentBlocksTracker`
### Background

The `RecentBlocksTracker` currently keeps some blocks that are `NotIncluded`:
 

https://github.com/near/mpc/blob/74476c3d1a256efacbd224ea07ec44bf6a3b75f6/crates/node/src/requests/recent_blocks_tracker.rs#L82

https://github.com/near/mpc/blob/74476c3d1a256efacbd224ea07ec44bf6a3b75f6/crates/node/src/requests/recent_blocks_tracker.rs#L433

`RecentBlocksTracker` classifies a block at height `h` as `NotIncluded`, in case:
- it is not final
- there exists a different final block of height greater or equal `h`.

This is pointless:
- `PendingRequests`, the only consumer, does not distinguish between `NotIncluded` or `Unknown`, so there is no requirement to keep the separation.
- There is no benefit for the `RecentBlocksTracker` to keep track of `NotIncluded` blocks.
- it makes refactor for #3135

  

### User Story

_No response_

### Acceptance Criteria

Remove the `NotIncluded` variant and discard blocks as soon as they are not useful anymore.

_No response_

### LINKED ISSUE #3139: Bug: `RecentBlocksTracker` classifies blocks wrongly as `OptimisticAndCanonical`
### Background

Consider this unit test of `RecentBlocksTracker`:

https://github.com/near/mpc/blob/f6078e85d009e20cb18d49d7f868e51c245aa0b0/crates/node/src/requests/recent_blocks_tracker.rs#L864-L912

Specifically, this comment:
>         // Note: b000 has no chance of being included in the canonical chain due to b10 being final.
>        // However, checking that case is not worth the complexity. In practice, the check we use of
>        // checking the height against the final head's height is likely good enough.

What's happening is that we are classifying some blocks that should be labelled `NotIncluded` as `OptimisticAndCanonical`. This is a genuine bug. It means that our nodes may be attempting signature requests that we know will never be included on chain.

Now, the original author obviously knew about that and deemed it not necessary. Their reasoning was that the complexity of the added check would not be worth it.
That makes sense due to how the `classify_blocks` method was structured at the time. However, with the modifications in #3132 it becomes much easier to set finality status of `BlockNode` when calling `add_block` (c.f. #3136).

### User Story

As a node operator, I like when my node uses its resources efficiently.

### Acceptance Criteria

Do not classify blocks on discarded sub-trees as `Optimistic`. Or simply put, these should compare to `NotIncluded` or `Unknown`:

https://github.com/near/mpc/blob/f6078e85d009e20cb18d49d7f868e51c245aa0b0/crates/node/src/requests/recent_blocks_tracker.rs#L895-L898

_No response_

kevindeforth:
resolves #3136 and #3139

This PR introduces the following changes:
- refactor (#3136): `RecentBlocksTracker` no longer classifies blocks as `NotIncluded`. This classification applied to blocks that were:
    -  not finalized
    - older than the latest final block
    These block are essentially useless. The consumer of `RecentBlocksTracker` treats them the same as `Unknown` blocks, so the distinction does't matter.
- fix (#3139): `RecentBlocksTracker` no longer classifies blocks that are known to not be included as `Optimistic`. This is a bug of low impact. It's primary effect was waste of computation resources.

The solution proposed in this PR has the same asymptotic complexity as the existing code.

### DIFF
--- a/crates/node/src/requests/recent_blocks_tracker.rs
+++ b/crates/node/src/requests/recent_blocks_tracker.rs
@@ -80,7 +80,7 @@ pub enum CheckBlockResult {
     /// The block is optimistically included in the chain, and it is on the canonical chain,
     /// but it is not yet part of the final chain. It is also recent enough.
     /// Note that if two chains tie for canonical height, the first one seen is considered the
-    /// canonical chain (c.f. [`RecentBlocksTracker::update_canonical_head`]).
+    /// canonical chain (c.f. `RecentBlocksTracker::update_canonical_head`).
     OptimisticAndCanonical,
     /// The block is optimistically included in the chain, but it is not on the canonical chain.
     /// It is also recent enough.
--- a/crates/node/src/requests/recent_blocks_tracker.rs
+++ b/crates/node/src/requests/recent_blocks_tracker.rs
@@ -481,33 +481,41 @@ impl<T: Clone> RecentBlocksTracker<T> {
     /// If block 8 is finalized, we can remove the subtrees at (5), (6) and (7):
     ///
     /// ```text
-    ///  ------------------------┐
-    ///     These blocks will    ┆
-    ///      never finalize      ┆
-    ///                          ┆
-    ///   root 1         root 2  ┆   root 3
-    ///   (1)                    ┆             remove because too old
-    ///    │              (2)    ┆             remove because too old
-    ///   (3)              │     ┆             remove because too old
-    ///    │               │     ┆     (4F)    remove because too old
-    ///  ┌─┴─┐             │     ┆    ┌─┴─┐
-    /// ═════════════════════════┆═══════════ ← cutoff (h ≥ 4). Remove blocks older than this.
-    /// (5)  │             │     ┆    │   │    remove because can't finalize
-    ///  │  (6)            │     ┆    │   │    remove because can't finalize
-    ///  │   │            (7)    ┆    │   │    remove because can't finalize
-    ///  │   │             │     ┆  (8F)  │    new root
-    ///  │   │             │     ┆    │  (9)   new root
-    /// -------------------------┆----------- ← below this line are children of new roots
-    ///  │   │             │     ┆    │   │
-    ///  ..  ..            ..    ┆    ..  ..
+    ///  ---------------------------┐
+    ///     These blocks will       ┆
+    ///     never be included       ┆
+    ///                             ┆
+    ///                             ┆
+    ///   root 1   root 2           ┆ root 3
+    ///   (1)                       ┆              Dead-fork prune at (1)
+    ///    │        (2)             ┆              Dead-fork prune at (2)
+    ///   (3)        │              ┆
+    ///    │         │              ┆ (4F)         Recency window prune node (4F)
+    ///  ┌─┴─┐       │       ┌──────────┴─┐
+    /// ════════════════════════════════════════ ← cutoff (h ≥ 4). Remove blocks older than this.
+    /// (5)  │       │       │      ┆     │
+    ///  │   │       │       │      ┆    (7F)      (7F) becomes new root
+    ///  │   │       │       │   ┌────────┴─┐
+    ///  │  (6)      │       │   │  ┆       │
+    ///  │   │       │       │  (8) ┆       │      Dead-fork prune at (8)
+    ///  │   │       │       │   │  ┆     (9F)
+    ///  │   │       │      (10) │  ┆       │      Dead-fork prune at (10)
+    ///  │   │       │       │   │  ┆     (11F)    ← latest final block height.
+    ///  │   │       │       │   │  ┆       │         Remove subtrees of older,
+    ///  │   │       │       │   │  ┆       │         non-final blocks
+    ///  │   │       │       │   │  ┆       │
+    ///  │   │       │       │   │  ┆     (12)
+    ///  │   │       │      (13) │  ┆       │
+    ///  │   │       │       │   │  ┆       │
+    ///  ..  ..      ..      ..  .. ┆       ..
     ///
     /// ```
-    /// After prune, we have two new roots:
+    /// After prune, we have one new root:
     /// ```text
-    ///  root 1  root 2
-    ///   (8F)
-    ///    │       (9)
-    ///    ..       │
+    ///  root 1
+    ///   (7F)
+    ///    │
+    ///    ..
     ///             ..
     ///  ```
     fn prune_old_blocks(&mut self) {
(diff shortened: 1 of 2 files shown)


crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -234,7 +328,7 @@ pub struct BlockViewLite {
     pub timestamp_nanosec: u64,
 }
 
-impl<T: Clone> RecentBlocksTracker<T> {
+impl<T: Clone + Debug> RecentBlocksTracker<T> {
```
kevindeforth, line 333: This will be reverted when we resolve https://github.com/near/mpc/issues/3070

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -33,12 +33,12 @@ use std::sync::{Arc, Mutex, Weak};
 ///    height 12 and then another block of height 10 that belongs to a different fork.
 ///  - The indexer, upon startup, may start giving blocks from any position in the blockchain,
 ///    including giving blocks from multiple forks without giving us any common parents.
+///  - The indexer may jump block heights, as long as the `prev_hash` exists and has been provided.
```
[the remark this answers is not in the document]
kevindeforth: > The doc comment says the indexer may "jump block heights" (line 36), which reads like allowing gaps in the ancestor chain, but earlier this same doc explicitly assumes there are no gaps (lines 28-29)

Did you try to finish reading the sentence? :smile: 

> The indexer may jump block heights, **as long as the `prev_hash` exists and has been provided.**

But I get it. Why explain something in one sentence, if it can be explained in three? :melting_face:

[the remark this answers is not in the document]
gilcu3: I did try to read it but I found some ambiguity :) What does it mean that "prev_hash exists and has been provided", does that mean that it corresponds to another block the indexer saw? If yes, then what does it mean that "the indexer may jump block heights"?

crates/node/src/requests/recent_blocks_tracker.rs
gilcu3, line 89: IIUC this comment is not longer accurate

crates/node/src/requests/recent_blocks_tracker.rs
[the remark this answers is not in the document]
kevindeforth: > What does it mean that "prev_hash exists and has been provided", does that mean that it corresponds to another block the indexer saw

Yes, exactly. it's a repetition of this statement:

> /// if A < B and B < C, and both A and C are given, then B must also be given.

So, if we receive block `B` referencing parent `A`, we need to have seen `A` beforehand (modulo starting the indexer. We don't need to sync from genesis, but we will have issues if we skip blocks randomly).

> what does it mean that "the indexer may jump block heights"?

It means we can skip block heights, as long as we don't skip blocks. 

So, This is okay for arbitrary m:

A at height h
B at height h+m, with prev_hash A

We don't care if block height increases by 1 or more.

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -375,39 +469,94 @@ impl<T: Clone> RecentBlocksTracker<T> {
         )
     }
 
-    /// Remove old blocks that are neither needed for the `classify_block` query or for providing
-    /// the stream of finalized blocks.
+    /// Remove blocks that are no longer needed:
+    ///  - **Recency Window prune:** Drop nodes below `minimum_height_to_keep`,
+    ///    their children become new roots
```
gilcu3, line 476: is the second part accurate? Wouldn't it be possible that the children are also too old, and therefore pruned?

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -120,27 +117,124 @@ struct BlockNode {
...
-    /// Find the closest descendants of this node (including itself), whose height is at least
-    /// `height`. These descendants are appended to `descendants_output`. The nodes from the
-    /// subtree that are lower than the height are appended to `old_nodes`.
-    fn closest_descendants_with_height_at_least(
-        self: &Arc<BlockNode>,
-        height: u64,
-        descendants_output: &mut Vec<Arc<BlockNode>>,
-        old_nodes: &mut Vec<Arc<BlockNode>>,
+    /// Walk the subtree of this node and identify nodes to prune, as well as nodes that become
+    /// orphaned:
```
gilcu3, line 123: what is an orphaned node here? it is not mentioned in any of the points below.

gilcu3 (commented):
Added some comments as I need to go and I cannot yet understand the change. Will try to continue the review as soon as I get some time, unless it is already approved and merged by then

crates/node/src/requests/recent_blocks_tracker.rs
kevindeforth: Yeah, it should state:
"We may have seen the block and removed it, or we may never have seen it."

Is fixed in https://github.com/near/mpc/pull/3180/files

crates/node/src/requests/recent_blocks_tracker.rs
kevindeforth: > Wouldn't it be possible that the children are also too old, and therefore pruned?

Such children would also be pruned yes. Their children will then become the new roots (or be dropped as well, it's recursive).

crates/node/src/requests/recent_blocks_tracker.rs
kevindeforth: "become orphaned" means that their parent is to be removed from the tree.
Orphaned nodes = nodes that will become the new roots.

crates/node/src/requests/recent_blocks_tracker.rs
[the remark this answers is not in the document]
gilcu3: might be worth being more verbose here to make this sentence understandable

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -77,11 +76,9 @@ pub enum CheckBlockResult {
     /// The block is within the recent window, and also finalized by the blockchain
     /// (it is an ancestor (including self) of the latest final block).
     RecentAndFinal,
-    /// The block is recent enough, but belongs to a different fork so has no chance
-    /// of being included in any canonical chain from this point on.
-    NotIncluded,
     /// The block is optimistically included in the chain, and it is on the canonical chain,
     /// but it is not yet part of the final chain. It is also recent enough.
+    /// On a same-height tie the canonical chain is "first-seen wins" — see `update_canonical_head`.
     OptimisticAndCanonical,
```
gilcu3, line 84: Here I cannot also make sense of the last sentence. The jump from "It is also recent enough" to "On a same-height tie..." was unexpected.

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -120,27 +117,124 @@ struct BlockNode {
...
+                self.prune_dead_children(fh, blocks_to_prune);
+            }
+        }
+    }
+
+    /// This block is kept, but it may have descendants that sit on dead forks
+    fn prune_dead_children(&self, final_head_height: u64, blocks_to_prune: &mut Vec<CryptoHash>) {
+        let mut children = self.children.lock().unwrap();
+        children.retain(|child| {
+            if child.height <= final_head_height && !child.is_final.load(Ordering::Relaxed) {
```
gilcu3, line 222: would be nice to have a helper function for `child.is_final.load(Ordering::Relaxed)` similar to `get_parent`

gilcu3 (dismissed):
Thank you!

If the assumptions about how blocks are given by the indexer are valid, this LGTM.

kevindeforth:
> Thank you!
> 
> If the assumptions about how blocks are given by the indexer are valid, this LGTM.


Note: this PR does not change the assumptions made about the indexer's behavior, compared to what we currently have on main.

crates/node/src/requests/recent_blocks_tracker.rs
kevindeforth: yeah, this will be reworked in #3135

crates/node/src/requests/recent_blocks_tracker.rs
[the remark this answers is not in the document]
kevindeforth: rephrased in https://github.com/near/mpc/pull/3159/commits/c0127d46b1587c8a03ba510970d3b15665706ae1

crates/node/src/requests/recent_blocks_tracker.rs
kevindeforth: rephrased in https://github.com/near/mpc/pull/3159/commits/c0127d46b1587c8a03ba510970d3b15665706ae1

pbeza (approved):
This code is pretty tricky, and I wish I had more time to study it properly and look for simplifications, but I can’t really afford that right now. Overall LGTM. I’ll open a follow-up PR if I find a good way to simplify it — based on the [vibe-coded experiments](https://github.com/near/mpc/pull/3196), I think there _might_ be some room to simplify the algorithm.

gilcu3 (approved):

Issue #1238, opened 2025-10-06
Add Attestation Quote Verification Restriction to Known PPIDs


barakeinav1:
### Background
Not mandatory for first TEE launch 

Recent research disclosed vulnerabilities in DDR4-based SGX and SEV-SNP platforms ([wiretap.fail](https://wiretap.fail/)
 and [batteringram.eu](https://batteringram.eu/)
). While these attacks currently target SGX systems using DDR4, they demonstrate potential for future adaptations to DDR5 and may eventually affect TDX.

Although TDX-capable systems use DDR5 (and are therefore not currently impacted), one critical takeaway is that extraction of the attestation key could allow forging arbitrary quotes for a given chip and firmware 

It was noted that verifiers should restrict accepted quotes to known PPIDs to prevent forgery from unapproved nodes. The current implementation does not enforce this restriction.

### Acceptance Criteria

Implement verification logic that restricts attestation quote validation to a predefined list of known PPIDs.

 Add a governance mechanism for approving and voting on new PPIDs (e.g., as part of node onboarding).

 Ensure that contract verification logic uses this approved PPID list to validate attestation quotes.

 Document the PPID registration and validation flow for node operators.

### Resources & Additional Notes

Discussion summary: insights from @barak einav, @Mårten Blankfors, and @Tjaden Hess.

Tjaden Hess’s Twitter breakdown of the vulnerabilities: https://x.com/tjade273/status/1973168405738815851

Vulnerability disclosures:

https://wiretap.fail/

https://batteringram.eu/

DDR4 attacks currently require physical access and do not directly impact MPC nodes today, but proactive hardening is recommended.

Consider periodic review as DDR5-based variants of these attacks emerge.

netrome:
Still very important but we need to focus on the rust based launcher this week - so demoting this from the tip

netrome:
Keeping this as a P2. This is a relatively high priority, but given the current contract size challenges we'll hold this for a short while.

PR #3160, opened 2026-05-06
docs: design doc for attestation verifier contract breakout

pbeza:
## Summary

- Adds [docs/design/attestation-verifier-contract.md](docs/design/attestation-verifier-contract.md): a design doc for breaking TEE attestation verification out of the MPC contract into a thin shared global contract.
- v1 ships a stateless `tee-verifier.near` that wraps `dcap_qvl::verify()`; MPC contract calls it via async Promise + callback. No Borsh state migration.
- v2 sketches a three-contract architecture (global verifier + per-team `admin_id`-governed policy contract + TEE-agnostic application contract) so Defuse and Proximity can adopt the same verifier with their own policies.

Closes near/mpc-private#303


pbeza:
@claude review

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+
+Only the cold path needs the heavyweight verifier; the hot path stays cheap.
+
+MPC binds report-data as `sha3_384(tls_pk || account_pk)` (see [`crates/mpc-attestation/src/report_data.rs`][mpc-report-data]). It runs RTMR3 event-log replay, MPC image-hash whitelisting, launcher-compose-hash whitelisting, and app_compose JSON validation as post-DCAP checks. Allowlists are governed by a threshold-of-participants vote.
+
+### Other teams
+
+[**Proximity**][proximity-shade-attestation]:
+
+- PPID whitelist check (MPC has no equivalent).
```
barakeinav1: PPID whitelisting is something we want to add for the MPC contract see https://github.com/near/mpc/issues/1238

barakeinav1:
partial review (V1 part only)
Reviewed the V1 approach. this seems like a straightforward approach mimicking what we have now (e.g keeping the same call hierarchy as we have today) . With an easy and simple upgrade mechanism for the verification contract. looks good.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+`Collateral` and `VerifiedReport` are Borsh-stable wire types matching `dcap_qvl::verify::QuoteCollateralV3` and `dcap_qvl::verify::VerifiedReport` respectively, exposed via a new `tee-verifier-interface` crate so callers can decode the response without depending on `dcap-qvl` directly. The returned `VerifiedReport` carries the parsed quote (RTMRs, MRTD, report_data, advisory IDs, TCB statuses); each consumer's downstream policy then decides what counts as acceptable for its application.
+
+The verifier does only what `dcap_qvl::verify::verify` does. No advisory-ID check, no report-data binding, no RTMR3 replay, no measurement match — those checks differ across teams (RTMR3 inputs, app-compose policy, report-data binding all diverge), so they live in each consumer's policy/application code.
+
+### MPC contract changes
+
+`TeeState` (the MPC contract's stored attestations and allowlists) is unchanged in v1. Cross-contract calls into a Global Contract are normal NEAR async Promises; the MPC contract becomes a Promise + callback flow on the cold path:
+
+```
+submit_participant_info(attestation, tls_pk):
```
barakeinav1, line 110: is the flow illustrated here, the same flow as we have today? (except for Sync call to the contract), or is there some diviation?

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+
+This generalizes naturally: non-TEE policies (admin-set ACL), multi-mechanism policies (TEE or DAO membership), or delegated policies that forward to another ACL contract — all conform to the same interface.
+
+This matches the preference Defuse has expressed for hiding TEEs behind an authorization layer rather than baking TEE concepts into application contracts.
+
+**On-chain PCCS caching.** Mirroring [Automata's design][automata-pccs]: a separate `pccs.near` contract stores fresh collateral per FMSPC, governed by an `admin_id` (or permissionless if root-CA validation is replayed on insert). The verifier grows a `verify_quote_via_pccs(quote, pccs_account_id, fmspc, now)` companion that reads collateral by cross-contract call instead of accepting it inline. This removes the obligation for off-chain components to fetch and bundle collateral on every call.
```
barakeinav1, line 226: we have 2 possible way to fetch the collateral:
1. verifier fetches it
2. The policy contract fetch it and send to the verifier.

I don't have a preference , just mentioning the options.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+   └──────────▲─────┘  └───────────▲──────┘  └───────────▲──────┘
+              │ is_attested?       │                     │
+              │ (hot path)         │                     │
+   ┌──────────┴─────┐  ┌───────────┴──────┐  ┌───────────┴──────┐
+   │ mpc-contract   │  │ defuse-app       │  │ proximity-app    │
+   │ TEE-agnostic   │  │ TEE-agnostic     │  │ TEE-agnostic     │
+   └────────────────┘  └──────────────────┘  └──────────────────┘
+```
+
+Per-team policy contract (e.g., `mpc-tee-policy.near`) holds: allowlists, the team's extra-check logic, and `stored_attestations`. Each team's policy contract has its own `admin_id` for governance — the `admin_id` is per-team, not shared. That admin can be a DAO, a proxy, a multisig, or a single account. Different teams can pick different governance schemes behind the same interface.
```
barakeinav1, line 188: what is the main benefit(s) of have it as a separate contract and not part of the main contract? (since we don't have code sharing).

here is what I can think of:
abstraction of TEE? (e.g can use different TEEs etc..) 
different voting DAO

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+}
+```
+
+`NodeId` is opaque to the trait — each team can use whatever shape it needs (MPC's `(account_id, tls_pubkey, account_pubkey)`, Proximity's account ID, etc.). The interface is currently TEE-flavoured (`is_attested`, `node_id`); v3 generalizes it (see below).
+
+### v3: ACL-shaped interface and on-chain PCCS
+
+**ACL generalization.** Most TEE-using applications don't actually want to know about TEEs — they want to know "is the caller authorized for this action?". TEE attestation just populates the answer.
+
+Replacing `is_attested(node_id) -> bool` with `is_authorized(account_id, action) -> bool` lets the application contract drop every TEE type. Internally the policy maps `is_authorized` to "is `account_id` in `stored_attestations` AND does it re-verify AND does the action satisfy any role-based restriction?". TEE machinery becomes an implementation detail of the policy.
```
barakeinav1, line 220: nice idea

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+
+This generalizes naturally: non-TEE policies (admin-set ACL), multi-mechanism policies (TEE or DAO membership), or delegated policies that forward to another ACL contract — all conform to the same interface.
+
+This matches the preference Defuse has expressed for hiding TEEs behind an authorization layer rather than baking TEE concepts into application contracts.
+
+**On-chain PCCS caching.** Mirroring [Automata's design][automata-pccs]: a separate `pccs.near` contract stores fresh collateral per FMSPC, governed by an `admin_id` (or permissionless if root-CA validation is replayed on insert). The verifier grows a `verify_quote_via_pccs(quote, pccs_account_id, fmspc, now)` companion that reads collateral by cross-contract call instead of accepting it inline. This removes the obligation for off-chain components to fetch and bundle collateral on every call.
+
+[mpc-repo]: https://github.com/near/mpc
+[defuse-site]: https://near-intents.org
```
netrome: Intents has moved to near.com 

Image

```suggestion
[defuse-site]: https://near.com
```

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
+# Shared TEE Attestation Verifier Contract
+
+[NEAR's MPC][mpc-repo], [Defuse][defuse-site], and [Proximity][proximity-shade-attestation] all run on-chain Intel TDX attestation verification, each carrying its own copy of `dcap_qvl::verify()` and its dependency closure (`ring`/`webpki`, X.509 parsing, SHA-2/SHA-3). Every fix to `dcap-qvl` (TCB cert rotation, advisory-ID handling) gets applied N times. Compounding this, the MPC contract WASM is approaching the 1.5 MB transaction-size limit imposed by [NEP-509][nep-509]; the attestation crates are the heaviest dependency surface in that contract.
+
+Goals:
+
+- Provide a shared, generic verifier serving any Intel TDX-based NEAR contract — `dcap-qvl` bugfixes ship once.
+- Shrink the MPC contract WASM below the NEP-509 limit by breaking attestation verification out into a separate contract.
+- Roll out incrementally: v1 keeps state migration to zero. Architectural improvements follow in v2 (per-team policy contracts) and v3 (ACL-shaped interfaces, on-chain PCCS caching).
```
netrome: I wouldn't phrase the roll-out strategy as a design goal. Either roll-out considerations are features of different design choices, or just a separate consideration once we've settled on the target architecture.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
+# Shared TEE Attestation Verifier Contract
+
+[NEAR's MPC][mpc-repo], [Defuse][defuse-site], and [Proximity][proximity-shade-attestation] all run on-chain Intel TDX attestation verification, each carrying its own copy of `dcap_qvl::verify()` and its dependency closure (`ring`/`webpki`, X.509 parsing, SHA-2/SHA-3). Every fix to `dcap-qvl` (TCB cert rotation, advisory-ID handling) gets applied N times. Compounding this, the MPC contract WASM is approaching the 1.5 MB transaction-size limit imposed by [NEP-509][nep-509]; the attestation crates are the heaviest dependency surface in that contract.
+
+Goals:
+
+- Provide a shared, generic verifier serving any Intel TDX-based NEAR contract — `dcap-qvl` bugfixes ship once.
```
netrome: Intel TDX-based NEAR contract sounds strange. Also the "`dcap-qvl` bugfixes ship once" sentence feels redundant here.

How about something like this?

```suggestion
- Provide shared on-chain TEE attestation verification functionality.
```

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+- Shrink the MPC contract WASM below the NEP-509 limit by breaking attestation verification out into a separate contract.
```
netrome: We're mixing the goal and the solution here. The goal is just to shrink the contract. Breaking out attestation verification is how we achieve that goal.

```suggestion
- Shrink the MPC contract WASM below the NEP-509 limit.
```

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+- Roll out incrementally: v1 keeps state migration to zero. Architectural improvements follow in v2 (per-team policy contracts) and v3 (ACL-shaped interfaces, on-chain PCCS caching).
+
+Design choices that follow from these goals, applied throughout the design below:
+
+- The verifier is the thinnest possible wrapper around `dcap_qvl::verify()`. Anything team-specific lives outside it.
+- State stays where it's governed: per-team allowlists and stored attestations live in per-team contracts, not in the shared verifier.
+- The hot path doesn't change. Re-verification is hash comparisons today; v1 does not regress that.
```
netrome: This feels out-of-place. Why does it follow from these goals how we organize state? That feels like a separate discussion. Perhaps we should have a separate design choices section, containing some supporting motivation for the design choices?

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+## Current State
+
+### MPC contract today
+
+The MPC contract has two attestation flows:
+
+- **Initial verification** (`submit_participant_info`): runs `dcap_qvl::verify` plus all post-DCAP checks once per node onboarding. Cold path.
+- **Re-verification** (`verify_tee`, post-reshare cleanup, `clean_invalid_attestations`): re-checks each stored `ValidatedDstackAttestation` against current allowlists using only hash comparisons. This path does **not** invoke `dcap_qvl::verify`.
+
+Only the cold path needs the heavyweight verifier; the hot path stays cheap.
```
netrome: Both of these paths are cold imo. None of these are direct parts of signature request flows.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+- PPID whitelist check (MPC has no equivalent).
+- No launcher concept: app images are stateless by design, so the [launcher pattern][mpc-launcher] isn't needed; app-image gating is via the app-compose hash whitelist.
+- Report-data binding is the caller's account ID (vs. MPC's `sha3_384(tls_pk || account_pk)`).
+- Single `owner_id` field gates all governance.
+
+[**Defuse**][defuse-site]: no on-chain attestation contract yet. Their proposed design:
+
+- Verifier is global; callbacks the application via `application.on_tee_attested(sender, msg)`.
+- Application is fully TEE-agnostic — gates by `predecessor == verifier.near` plus its own ACL.
+- Two contracts: collateral governance + measurements governance, each with a single `admin_id`.
```
netrome, line 48: This section seems to outline design choices different teams are pushing for. I'd be very interested to know if we could shed some light on what their needs are? Which products do this power, and how does TEE verification fit into their needs?

I guess some of this may be sensitive, so if we can't share anything this is fine. But if we could more cleanly separate what they need from what design they propose it would make for much better reading.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Short-Term Design (v1)
+
+The MPC team needs the contract shrunk fast, so v1 is deliberately small: extract `dcap_qvl::verify()` into a shared verifier, point the MPC contract at it, change nothing else. Per-team policy contracts, on-chain PCCS caching, and ACL generalization wait for v2 and v3.
```
netrome: This makes it sound like we're desperate for a solution. We're not under high pressure to cram this out super fast, as we can always add support for multi-part uploads if we'd hit contract size limitations.

But yeah, the sooner we break this out the better.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+       │ Promise        │ VerifiedReport
+       │ (verify_quote) │ (callback)
+       │                ▼
+┌──────┴────────────────────────┐
+│  mpc-contract                 │
+│  TeeState unchanged           │
+│  stored_attestations cache    │
+│  re_verify hot path local     │
+└───────────────────────────────┘
+```
```
netrome, line 71: Also note, while this sounds simple, properly implementing this in the code would force us to re-architecture our low-level attestation verification crate and primitives. So this will still not be a trivial change.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+
+### Verifier contract
+
+The verifier WASM is published as a [NEP-591][nep-591] **Global Contract by CodeHash**. CodeHash makes the published code immutable on the network and auditable by hash — the security story we want for an attestation-verification dependency. Upgrades happen by publishing a new hash and migrating consumers to a new versioned account; we never edit deployed code in place.
+
+A versioned account (e.g., `tee-verifier-v1.near`) issues `UseGlobalContractAction(Global(<code_hash>))` to make itself a callable instance of the verifier. Consumers Promise into that account. When v2 of the verifier ships, a fresh account (e.g., `tee-verifier-v2.near`) issues `UseGlobalContractAction` against the new hash and consumers update their Promise targets explicitly. There is no admin on either the published code or the version account.
```
netrome, line 77: How do we ensure the versioned account doesn't deploy a new arbitrary contract? If this is deployed by code hash I assume we'd need to require this account to be locked (i.e. have all access keys removed), before the MPC network can start using it for TEE verification.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+
+- RTMR3 replay needs `tcb_info.event_log`, a separate input with a Dstack-specific shape we don't want to standardize at the verifier layer.
+- App-compose validation is opinionated (MPC's `kms_enabled == false`, etc.); other teams may want different rules.
+- Report-data binding differs across teams.
+
+### Re-verification doesn't call the verifier
+
+`re_verify`, `verify_tee`, post-reshare cleanup, and `clean_invalid_attestations` are unchanged in v1. They operate on cached `ValidatedDstackAttestation` blobs and do hash comparisons against the current allowlists. None of them call the verifier; they don't even need a Promise. Same property as today.
+
+In v2/v3, this logic moves into the per-team policy contract (see "Long-Term Direction" below). At every version, re-checking already-stored attestations against current allowlists never calls the verifier — it's a local read of the cache (`mpc-contract`'s in v1, `mpc-tee-policy`'s in v2/v3). The verifier only runs at first submission, never on re-checks.
```
netrome, line 149: There's a lot of foreshadowing to v2/v3. I think it would be easier to read if the initial design was presented as a stand-alone design, and any envisioned extensions could be added as an appendix or some separate document outlining the long-term vision but clearly separating that from the actionable next steps we plan to take.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+    collateral: Collateral,
+    now_seconds: u64,
+) -> Result<VerifiedReport, VerificationError>;
+```
+
+`Collateral` and `VerifiedReport` are Borsh-stable wire types matching `dcap_qvl::verify::QuoteCollateralV3` and `dcap_qvl::verify::VerifiedReport` respectively, exposed via a new `tee-verifier-interface` crate so callers can decode the response without depending on `dcap-qvl` directly. The returned `VerifiedReport` carries the parsed quote (RTMRs, MRTD, report_data, advisory IDs, TCB statuses); each consumer's downstream policy then decides what counts as acceptable for its application.
+
+The verifier does only what `dcap_qvl::verify::verify` does. No advisory-ID check, no report-data binding, no RTMR3 replay, no measurement match — those checks differ across teams (RTMR3 inputs, app-compose policy, report-data binding all diverge), so they live in each consumer's policy/application code.
+
+### MPC contract changes
```
netrome, line 97: How do we intend to configure the verifier contract? Is this governed by a separate voting mechanism, or just part of the existing contract configuration struct (of which we already have voting flows)?

I don't see anything on this here.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+- Two contracts: collateral governance + measurements governance, each with a single `admin_id`.
+
+## Short-Term Design (v1)
+
+The MPC team needs the contract shrunk fast, so v1 is deliberately small: extract `dcap_qvl::verify()` into a shared verifier, point the MPC contract at it, change nothing else. Per-team policy contracts, on-chain PCCS caching, and ACL generalization wait for v2 and v3.
+
+```
+┌──────────────────────────────┐
+│  tee-verifier-v1.near        │
+│  Global Contract by CodeHash │
```
netrome, line 57: Given how simple this is, how many instantiations of this global contract do we envision? The docs recommend not using global contracts unless we picture more than 10 instantiations of it https://docs.near.org/smart-contracts/global-contracts#when-to-use-global-contracts

What is the data an instantiation would add to this global contract? If instantiating the contract doesn't add anything, it seems like it might as well just be a concrete contract instead of a global one.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+### Re-verification doesn't call the verifier
+
+`re_verify`, `verify_tee`, post-reshare cleanup, and `clean_invalid_attestations` are unchanged in v1. They operate on cached `ValidatedDstackAttestation` blobs and do hash comparisons against the current allowlists. None of them call the verifier; they don't even need a Promise. Same property as today.
+
+In v2/v3, this logic moves into the per-team policy contract (see "Long-Term Direction" below). At every version, re-checking already-stored attestations against current allowlists never calls the verifier — it's a local read of the cache (`mpc-contract`'s in v1, `mpc-tee-policy`'s in v2/v3). The verifier only runs at first submission, never on re-checks.
+
+### What v1 saves
+
+- The `mpc-contract` WASM no longer transitively depends on `dcap-qvl`, `ring`, `webpki`, or X.509 parsing. This is the load-bearing reduction for the NEP-509 size constraint.
+- No on-chain state migration: existing `stored_attestations` and allowlists keep their Borsh layout, all governance entry points keep working. The only behavioural change is that `submit_participant_info` becomes a two-step transaction.
```
netrome, line 154: I wouldn't be surprised if we wanted to change something about the attestation types though.

netrome (commented):
Gone through the first bits now. Some comments, questions and thoughts from my side. I'll continue catch up on this during the week.

docs/design/attestation-verifier-contract.md
PiVortex: It's worth considering the trade-off between using ppid and device_id. device_id in dstack is supposed to be a platform-independent solution for identifying devices if / when dstack supports other hardware than just TDX. 
Note that device_id is calculated differently depending on the version of dstack used. 

For TDX
dstack >= 0.5.6,  device_id = sha256(ppid) https://github.com/Dstack-TEE/dstack/blob/v0.5.6/dstack-attest/src/attestation.rs#L486
dstack <0.5.6, device_id = sha256(quote.header.user_data) https://github.com/Dstack-TEE/dstack/blob/v0.5.5/ra-tls/src/attestation.rs#L165

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+- The hot path doesn't change. Re-verification is hash comparisons today; v1 does not regress that.
+
+## Current State
+
+### MPC contract today
+
+The MPC contract has two attestation flows:
+
+- **Initial verification** (`submit_participant_info`): runs `dcap_qvl::verify` plus all post-DCAP checks once per node onboarding. Cold path.
+- **Re-verification** (`verify_tee`, post-reshare cleanup, `clean_invalid_attestations`): re-checks each stored `ValidatedDstackAttestation` against current allowlists using only hash comparisons. This path does **not** invoke `dcap_qvl::verify`.
```
PiVortex, line 25: Proximity verifies that the measurements of a node still match the accepted measurements each time they request an action

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+MPC binds report-data as `sha3_384(tls_pk || account_pk)` (see [`crates/mpc-attestation/src/report_data.rs`][mpc-report-data]). It runs RTMR3 event-log replay, MPC image-hash whitelisting, launcher-compose-hash whitelisting, and app_compose JSON validation as post-DCAP checks. Allowlists are governed by a threshold-of-participants vote.
+
+### Other teams
+
+[**Proximity**][proximity-shade-attestation]:
+
+- PPID whitelist check (MPC has no equivalent).
+- No launcher concept: app images are stateless by design, so the [launcher pattern][mpc-launcher] isn't needed; app-image gating is via the app-compose hash whitelist.
+- Report-data binding is the caller's account ID (vs. MPC's `sha3_384(tls_pk || account_pk)`).
+- Single `owner_id` field gates all governance.
```
PiVortex, line 42: The contract given is a minimal example with a single owner; the main user of the shade-attestation, STP, has multiple accounts for governance. There is flexibility here.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+
```
PiVortex, line 43: Another key difference is that shade-attestation has no restrictions on the fields of the app compose like the MPC has, it just requires that the app compose hash in the attestation matches the approved hash. Users may want to turn off public_logs for example

docs/design/attestation-verifier-contract.md
PiVortex: shade-attestation is being used for STP (not live yet) https://x.com/proximityfi/status/2008584528613519688?s=46&t=sPrc7IMmwC_3Wog4EoUSRw 

Let me know what information you would like, and I'll be happy to provide what I can

PiVortex: Outlayer from fastNEAR is worth adding to this conversation. They allow customers to run arbitrary executions in TDX via yield and resume and are now powering the native NEAR Oracle. 

https://outlayer.fastnear.com/

docs/design/attestation-verifier-contract.md
barakeinav1: the meaning by hot vs cold is for attestation flow only, not for  signature requests.
 
cold - > first time submission of an attestation, full verification done 
hot -> e.g verfiy_TEE, checking existing attestation, partial verification only.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+A versioned account (e.g., `tee-verifier-v1.near`) issues `UseGlobalContractAction(Global(<code_hash>))` to make itself a callable instance of the verifier. Consumers Promise into that account. When v2 of the verifier ships, a fresh account (e.g., `tee-verifier-v2.near`) issues `UseGlobalContractAction` against the new hash and consumers update their Promise targets explicitly. There is no admin on either the published code or the version account.
+
+The verifier exposes one method:
+
+```rust
+pub fn verify_quote(
+    &self,
+    quote: Vec<u8>,
+    collateral: Collateral,
+    now_seconds: u64,
```
PiVortex: Is there a reason to allow the caller to control now_seconds? Instead of fetching the block timestamp inside the method?

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,219 @@
...
+   │ mpc-contract   │  │ defuse-app       │  │ proximity-app    │
+   │ TEE-agnostic   │  │ TEE-agnostic     │  │ TEE-agnostic     │
+   └────────────────┘  └──────────────────┘  └──────────────────┘
+```
+
+Per-team policy contract (e.g., `mpc-tee-policy.near`) holds: allowlists, the team's extra-check logic, and `stored_attestations`. Each team's policy contract has its own `admin_id` for governance — the `admin_id` is per-team, not shared. That admin can be a DAO, a proxy, a multisig, or a single account. Different teams can pick different governance schemes behind the same interface.
+
+Application contract becomes TEE-agnostic and asks the policy "is this node attested?". Re-verification on allowlist change moves with `TeeState` into `mpc-tee-policy`; the verifier remains uninvolved.
+
+### v2 application/policy interface
```
PiVortex, line 192: For v2/v3 I'd be keen to explore aligning on the fundamentals of what we verify — measurement matching, RTMR3 replay, PPID checks, app_compose validation, and sharing them as a crate (with a reference contract) or a shared contract.

The motivation is further code reuse and expanding the upgrade surface from dcap_qvl alone to the whole attestation crate. A future TDX exploit might land in the attestation crate rather than dcap_qvl, meaning each team needs to patch and audit independently. The same N× argument can apply here. 

We can be flexible in what we verify and how we verify it to allow for this.

docs/design/attestation-verifier-contract.md
pbeza: https://github.com/near/mpc/pull/3160/changes/e6effff88ded3e531deba7ffd1334a3dba0a9148

docs/design/attestation-verifier-contract.md
pbeza: It’s similar. I expanded on it a bit here: https://github.com/near/mpc/pull/3160/changes/029076adbeaf520c4d73bff5917d3c62834b653c. ✅

docs/design/attestation-verifier-contract.md
pbeza: Good point — both are valid.

To set expectations on scope: on-chain PCCS caching is a v3 idea sketched in the doc as future work; v1 doesn't change how collateral is fetched (off-chain MPC node still bundles it inline, same as today). When we do tackle v3, the choice is:
1. Verifier fetches — verifier takes (`pccs_account_id`, `fmspc`) and reads from `pccs.near` itself (what the doc currently sketches).
1. Policy contract fetches and forwards inline — keeps the verifier interface unchanged.

I'll flag this as an open question when v3 lands — no need to pick now.

docs/design/attestation-verifier-contract.md
pbeza: > here is what I can think of:
> abstraction of TEE? (e.g can use different TEEs etc..)
> different voting DAO

Yeah, both of yours are valid; there are a couple more concrete ones too:

- **TEE-agnostic app contract.** The MPC contract currently contains a lot of TEE-specific logic. Moving that out keeps `mpc-contract` focused purely on MPC: signing, key events, and governance. That gives us a smaller audit surface and makes the contract easier to reason about.
- **Further WASM size reduction.** v1 only drops `dcap-qvl` from the contract WASM. v2 moves the rest of the TEE-related code (allowlists, vote tables for `vote_code_hash`/`vote_add_launcher_hash`/`vote_add_os_measurement`, `stored_attestations`, `re_verify`) into the policy contract.
- **Independent governance**. Your "different voting DAO." Today both MPC governance (resharing, key events) and TEE governance (allowlists) live under the same `voter_or_panic` flow. Splitting them lets the TEE-policy authority be a different account / DAO / voting threshold from the MPC's signing-quorum governance. Useful when, e.g., an emergency measurement revocation shouldn't require the same threshold as a resharing vote.
- **TEE abstraction.** Your other one. If/when we add SEV-SNP or non-Dstack-TDX, the post-DCAP check stack lives in the policy contract; the MPC app doesn't import any TEE-specific type. Could even support multiple TEEs in parallel under one policy contract.
- **Reusable reference policy.** Other teams (Defuse, Proximity) can fork mpc-tee-policy or write their own conforming to the same interface; the v3 ACL-shaped variant (`is_authorized(account_id, action)`) makes that even more general.

docs/design/attestation-verifier-contract.md
pbeza: https://github.com/near/mpc/pull/3160/changes/d4571c9d894d81ac6f9acc98405779531d47e17f

docs/design/attestation-verifier-contract.md
pbeza: https://github.com/near/mpc/pull/3160/changes/b7d30b4bf33176635d69f521509b0c274b6a1c32

docs/design/attestation-verifier-contract.md
pbeza: https://github.com/near/mpc/pull/3160/commits/e962f6f4

docs/design/attestation-verifier-contract.md
pbeza: https://github.com/near/mpc/pull/3160/changes/05e456dba322f9ef4bd7558aaa3097c2eb39242c

docs/design/attestation-verifier-contract.md
pbeza: https://github.com/near/mpc/pull/3160/changes/e03ee80c41782110527b8ee0a195ee046c621559

docs/design/attestation-verifier-contract.md
pbeza: Yep, that’s the distinction I meant. I clarified it here: https://github.com/near/mpc/pull/3160/changes/db6e140e30129d81205e46a22a2672a4240a12db.

docs/design/attestation-verifier-contract.md
pbeza: @PiVortex Who at fastNEAR should I loop in — anyone you'd specifically recommend?

docs/design/attestation-verifier-contract.md
pbeza: Rephrased: https://github.com/near/mpc/pull/3160/changes/7c888ce7a245573b39a53a0f1b6162d7ddb7a5c1.

docs/design/attestation-verifier-contract.md
pbeza: Agreed: covered later in [§"What v1 saves"](https://github.com/near/mpc/blob/7c888ce7a245573b39a53a0f1b6162d7ddb7a5c1/docs/design/attestation-verifier-contract.md?plain=1#L144):

https://github.com/near/mpc/blob/7c888ce7a245573b39a53a0f1b6162d7ddb7a5c1/docs/design/attestation-verifier-contract.md?plain=1#L150

The crate split is required to get `dcap-qvl` out of the `mpc-contract` WASM.

docs/design/attestation-verifier-contract.md
pbeza: IIUC, with [NEP-616](https://github.com/near/NEPs/blob/master/neps/nep-0616.md) there’s no lock step because deterministic accounts are created without access keys in the first place. The sequence is:

1. Publish verifier WASM as a Global Contract by CodeHash.
1. Anyone triggers a `StateInit` action with `StateInit { code: GlobalCodeHash(<verifier_hash>) };` the protocol creates the account at the deterministic ID derived from that.
1. Consumers check off-chain that the on-chain account ID matches `derive_account_id(...)` for the expected hash, then start using it.

The verifier WASM exposes no method that calls `AddKey` / `DeployContract`, so the binding between account ID and code hash is enforced by protocol.

See also:

https://github.com/near/NEPs/blob/7ea639f0ada35dac706035e19c549229d57eaf80/neps/nep-0616.md?plain=1#L188-L204

docs/design/attestation-verifier-contract.md
pbeza: Good catch. Addressed in https://github.com/near/mpc/pull/3160/commits/54320e14.

tl;dr
- **Verifier itself**: no on-chain config. Each verifier instance is a NEP-616 deterministic account whose code (and therefore everything embedded in it — `dcap-qvl` version, Intel root certs) is bound to the code hash. Changes mean publishing a new hash and a new instance, not modifying the existing one.
- **Policy** (which measurements/image hashes/launcher hashes are acceptable): lives in the consumer contract. For MPC, the existing `vote_code_hash` / `vote_add_launcher_hash` / `vote_add_os_measurement` flows on `mpc-contract` are what govern it.

docs/design/attestation-verifier-contract.md
pbeza: For v1 a non-sharded regular contract at `tee-verifier.near` is probably enough, as long as it stays stateless — just a thin wrapper around `dcap_qvl::verify(quote, collateral, now)` that returns the parsed report and leaves measurement allowlists to each consumer. That's what lets one shared instance work for MPC, Proximity, and Defuse even though their allowlists differ.

A Global Contract starts making sense once the verifier itself holds state — specifically the RTMRs baked in per instance (Arseny's sharded direction). At that point you want the same code at many addresses, one per approved build, and Global Contracts fit naturally: audit once, reuse across teams in the NEAR ecosystem.

@mitinarseny might be able to add more on why he sees a sharded verifier with per-instance measurements as preferable to a stateless one where measurements live in the application contract.

docs/design/attestation-verifier-contract.md
pbeza: Thanks for clarifying. Fixed it here: https://github.com/near/mpc/pull/3160/commits/2d2269067c2a6d4ab078d728948c5427b1286dad.

docs/design/attestation-verifier-contract.md
pbeza: I took `owner_id` to be just an `AccountId` — the contract has one field, but the account behind it can be a DAO, multisig, or any governance contract.

docs/design/attestation-verifier-contract.md
pbeza: Thanks for the clarification. Fixed here: https://github.com/near/mpc/pull/3160/commits/01d9eb6ba69cdf59f71e2cbafd8ef3cd9a74221d.

docs/design/attestation-verifier-contract.md
pbeza: Good catch. Today MPC reads now from `env::block_timestamp_ms()` inside [`tee_state::current_time_seconds()`](https://github.com/near/mpc/blob/8d5d5bfad072f8141d72ef839eb3aaaa2db2d4da/crates/contract/src/tee/tee_state.rs#L136-L139) caller never supplies it. Fixed in 8d5d5bfa.

pbeza:
@mitinarseny, I've been thinking about the design you [sketched](https://nearone.slack.com/archives/C0B12RKBSAV/p1777897902903889) — one verifier instance per approved measurement, addressed by `H(verifier_code_hash || measurement_bundle)`, where the application gates on `predecessor_id ∈ trusted_verifiers`:

design

The main win of this design, as I understand it, is governance-time auditability: when a new `AccountId` is proposed for the trusted set, voters (and anyone else) can read its `account.contract = GlobalByCodeHash(...)` and `StateInit.data` on-chain, recompute `H(code_hash || measurement_bundle)`, and confirm the account really runs the audited verifier code against the claimed measurement. The on-chain runtime check stays simple — just `predecessor_id ∈ trusted_set` — but each entry in that set has a publicly verifiable provenance, and NEP-616 immutability guarantees the account can't change behavior after being added. This works cleanly when the measurement bundle fully identifies the approved deployment.

The catch for MPC: our RTMR3 _isn't_ set at build time. The Dstack launcher extends RTMR3 at runtime with deployment-specific events — MPC image hash, `app_compose` hash, key-provider digest. So an MPC verifier instance can only bake (`MRTD`, `RTMR0..2`) into its `StateInit.data` — the OS layer. The application-layer policy (which MPC image, which launcher, which `app_compose` JSON, plus our opinionated rules like `kms_enabled == false`) lives one level above what the address commits to.

That means MPC would need to track attestation policy in two places:
1. `HashSet<AccountId>` of trusted verifier instances, one per approved OS layer.
1. The existing MPC-specific allowlists: image hashes, launcher hashes, `app_compose rules`, TCB checks, etc.

The verifier instance would still need to return the parsed `VerifiedReport` to our callback, and MPC would still need to replay RTMR3 and match the hashes against its allowlists.

We already have three vote types today: `vote_code_hash`, `vote_add`/`remove_launcher_hash`, and `vote_add`/`remove_os_measurement`. With this design, we’d add a fourth one for trusted verifier instance `AccountIds` without really retiring any of the others.

What I’m proposing instead is a stateless `dcap_qvl` verifier deployed as a regular contract, basically exposing a single method that wraps `dcap_qvl::verify()` and accepts whatever inputs that function needs. It would return `VerifiedReport`, so the downstream contract can run all `post-dcap_qvl::verify()` checks itself, like replaying RTMR3 as MPC does today.

That feels like the most flexible primitive to me: a thin wrapper around the function we all eventually need to call. And if we ever want a global verifier contract later, IIUC, it could still reuse this regular stateless verifier under the hood.

mitinarseny:
> The catch for MPC: our RTMR3 isn't set at build time. The Dstack launcher extends RTMR3 at runtime with deployment-specific events — MPC image hash, app_compose hash, key-provider digest. So an MPC verifier instance can only bake (MRTD, RTMR0..2) into its StateInit.data — the OS layer. The application-layer policy (which MPC image, which launcher, which app_compose JSON, plus our opinionated rules like kms_enabled == false) lives one level above what the address commits to.

What prevents you from putting `RTMR3` in Verifier's `StateInit`, so that you store only `approved_verifiers: HashSet<AccountId>` in your Application contract and MPC participants vote for account_ids being added to/removed from this set?

pbeza:
> > The catch for MPC: our RTMR3 isn't set at build time. The Dstack launcher extends RTMR3 at runtime with deployment-specific events — MPC image hash, app_compose hash, key-provider digest. So an MPC verifier instance can only bake (MRTD, RTMR0..2) into its StateInit.data — the OS layer. The application-layer policy (which MPC image, which launcher, which app_compose JSON, plus our opinionated rules like kms_enabled == false) lives one level above what the address commits to.
> 
> What prevents you from putting `RTMR3` in Verifier's `StateInit`, so that you store only `approved_verifiers: HashSet<AccountId>` in your Application contract and MPC participants vote for account_ids being added to/removed from this set?

The thing is we _don't_ store allowed RTMR3 onchain explicitly. Instead we replay it from `tcb_info.event_log` (filtered to `imr == 3`) and check the folded digest matches the `rt_mr3` in the quote — see [`verify_event_log_rtmr3`](https://github.com/near/mpc/blob/8d5d5bfad072f8141d72ef839eb3aaaa2db2d4da/crates/attestation/src/attestation.rs#L153-L188) and [`verify_rtmr3`](https://github.com/near/mpc/blob/8d5d5bfad072f8141d72ef839eb3aaaa2db2d4da/crates/attestation/src/attestation.rs#L333-L341). Then we pull named events out of that log and check their payloads against what *is* stored on-chain:

https://github.com/near/mpc/blob/a5e83224b4febc139f609fd8666c4a9bf21c1bad/crates/contract/src/tee/tee_state.rs#L68-L82

https://github.com/near/mpc/blob/a5e83224b4febc139f609fd8666c4a9bf21c1bad/crates/mpc-attestation/src/attestation.rs#L135-L216

mitinarseny:
I see, thanks for explaining! But still, can this logic be a part of verifier contract then? And these event_names and their allowed value(s) be part of its StateInit?

PiVortex:
> I see, thanks for explaining! But still, can this logic be a part of verifier contract then? And these event_names and their allowed value(s) be part of its StateInit?

I believe the idea is to make the verifier implementation agnostic due to differences in implementations. For example, the MPC checks MPC_IMAGE_HASH_EVENT, where proximity does not, it instead checks the app_compose_hash. Another difference in implementations is that the MPC is opinionated in the app compose fields, requiring certain fields to be certain values, whereas proximity says, as long as the hash of the app compose matches the expected hash, that's fine, you decide the values each time you approve measurements. 

This logic is implemented at the higher level in v2, with each team having its own implementations. This being said, I think it could be possible to come to agreement on what we all verify here and possibly settle on a shared implementation.

pbeza:
> I see, thanks for explaining! But still, can this logic be a part of verifier contract then? And these event_names and their allowed value(s) be part of its StateInit?

We already covered this offline [here](https://nearone.slack.com/archives/C0B12RKBSAV/p1778746660285559?thread_ts=1777897902.903889&cid=C0B12RKBSAV), but just to repeat: some parts of that attestation verification logic are specific to how we, the MPC team, implemented our flow.

The global verifier contract is meant to be shared across teams, so I think it should probably stay closer to the common denominator of attestation verification — meaning it only needs to call `dcap_qvl::verify()`.

Different teams may have quite different flows. For example, teams that don’t use the launcher pattern may build very different logic on top of vanilla `dcap_qvl::verify()`. Attestation verification is not really a one-size-fits-all process.

> This logic is implemented at the higher level in v2, with each team having its own implementations. This being said, I think it could be possible to come to agreement on what we all verify here and possibly settle on a shared implementation.

💯

We could try to align on one shared attestation flow across teams. But that would mean Proximity, Defuse, and potentially others would need to use the [launcher pattern](https://github.com/near/mpc/blob/main/docs/securing-mpc-with-tee-design-doc.md#launcher-pattern) we use. It’s quite flexible, but also very custom, so I wouldn’t expect other teams in the ecosystem to adopt it without us guiding them through it. @PiVortex @mitinarseny Would you be up for exploring this direction?

https://github.com/near/mpc/blob/9febd3e82c0af1a4a401ca266e5aeab9bdd0094c/docs/securing-mpc-with-tee-design-doc.md?plain=1#L138-L181

PiVortex:
> Would you be up for exploring this direction?

Let me take a look more into our differences and see if there's any reason we couldn't adopt it

PiVortex:
> > Would you be up for exploring this direction?
> 
> Let me take a look more into our differences and see if there's any reason we couldn't adopt it

Following up on this @pbeza 

Adopting the launcher pattern doesn't make sense for us now because we don't store any data on disk that needs to be hidden from non-approved image upgrades; it would add unnecessary complexity, and we choose KMS over local-sgx for convenience. As such, it seems neither possible nor appropriate to share the whole attestation contract. 

However, I would be interested in sharing a published `attestation` crate if you're willing to explore this? 

Since adding PPID / device_id checks is in your roadmap, our only major difference is checks surrounding the app_compose. How our implementation differs: 

### MPC 
`attestation` requires strict values for some fields of the app_compose, then checks the launcher Docker compose hash matches an approved launcher Docker compose hash in `mpc-attestation`. 

### shade-attestation 
Checks the app compose hash against an approved app compose in the attestation crate. Fields are set by the hash.

---

We cannot adopt the MPC approach for a few reasons: we require environment variables, require KMS, require a pre-launch script, and we have not decided concretely on public vs private logs, etc.

Would you be willing to either:
1) Move `validate_app_compose_config` out of `attestation` into `mpc-attestation`.

or 

2) Adopt an appComposePolicy argument into `attestation` that determines the field requirements.

or

3) Implement an app_compose hash check in the `attestation` crate and remove `validate_app_compose_config`. 
 
@mitinarseny I don't know much about Defuse's planned architecture. Would a shared attestation crate help your team? Are you intending to adopt the launcher pattern?

netrome:
Thanks for all the good discussions. We're proceeding with a simpler initial design for the first few bits. It would be lovely if we could converge on some conventions that would allow us to increase the shared surface on-chain for TEE verification - but as it seems `dcap_qvl::verify` looks like the minimal shared reusable surface. Our current direction is already merged in https://github.com/near/mpc/pull/3311 - happy to continue discuss how we can grow and improve this over time.

Closing this PR for now to keep our list of open PRs clean.

PR #3161, opened 2026-05-06
feat: add ReconstructionThreshold to DomainConfig

### LINKED ISSUE #3143: Add ReconstructionThreshold in DomainConfig
### Background

See parent #2675

### User Story

_No response_

### Acceptance Criteria

The `ReconstructionThreshold` is added to `DomainConfig`. Current domains are initialized with the existing threshold, future domains will require this parameter as part of the vote.

https://github.com/near/mpc/blob/main/docs/design/domain-separation.md

gilcu3:
Closes #3143

For most tests the specific `ReconstructionThreshold` value was not relevant, so I just added a value that coincided with the already used threshold to keep the status quo. We might want to rethink this and figure if it makes sense to use different thresholds for some tests now that we can.

Notice this does not wire yet the `ReconstructionThreshold` in the node. That will need to be done in a follow-up #3164, as it is a complicated change by itself.

There is yet another piece of the equation here that is not yet handled by this PR, which is resharing (now handled in #3169). After this PR lands, when resharing occurs, the per domain thresholds are checked against the new number of participants, but they cannot be modified as part of the vote, as was the case before. This is not a problem because the thresholds per domains are not wired yet in the node, but would be after that. So probably this needs to be sorted out in tandem with #3164. The idea would be to add per domain thresholds to the resharing vote parameters, which seems enough of a change to deserve its own PR. So I am not sure if it would be better to have this as part of the next release, and then do both this resharing and #3164 in the next. The only problem with this idea is that the thresholds in domains, that are now part of the `add_domain` votes would be fake in the sense that the nodes would still use the governance threshold for everything.

### DIFF
--- a/crates/contract/src/errors.rs
+++ b/crates/contract/src/errors.rs
@@ -1,4 +1,5 @@
 use crate::crypto_shared::kdf::TweakNotOnCurve;
+use crate::primitives::domain::MIN_RECONSTRUCTION_THRESHOLD;
 use crate::primitives::key_state::{EpochId, Keyset};
 use near_account_id::AccountId;
 use near_mpc_contract_interface::types as dtos;
@@ -204,7 +205,10 @@ pub enum DomainError {
         protocol: Protocol,
         purpose: DomainPurpose,
     },
-    #[error("Reconstruction threshold must be at least 2.")]
+    #[error(
+        "Reconstruction threshold must be at least {}.",
+        MIN_RECONSTRUCTION_THRESHOLD
+    )]
     ReconstructionThresholdTooLow,
     #[error("Reconstruction threshold {threshold} exceeds participant count {participants}.")]
     ReconstructionThresholdExceedsParticipants { threshold: u64, participants: u64 },
@@ -216,6 +220,10 @@ pub enum DomainError {
         required: u64,
         participants: u64,
     },
+    #[error(
+        "Reconstruction threshold {threshold} overflowed when computing the DamgardEtAl bound."
+    )]
+    ReconstructionThresholdOverflow { threshold: u64 },
 }
 
 /// A list specifying general categories of MPC Contract errors.
--- a/crates/contract/src/errors.rs
+++ b/crates/contract/src/errors.rs
@@ -204,6 +204,18 @@ pub enum DomainError {
         protocol: Protocol,
         purpose: DomainPurpose,
     },
+    #[error("Reconstruction threshold must be at least 2.")]
+    ReconstructionThresholdTooLow,
+    #[error("Reconstruction threshold {threshold} exceeds participant count {participants}.")]
+    ReconstructionThresholdExceedsParticipants { threshold: u64, participants: u64 },
+    #[error(
+        "Protocol {protocol:?} requires at least {required} participants, found {participants}."
+    )]
+    InsufficientParticipantsForProtocol {
+        protocol: Protocol,
+        required: u64,
+        participants: u64,
+    },
 }
 
 /// A list specifying general categories of MPC Contract errors.
--- a/crates/contract/src/primitives/domain.rs
+++ b/crates/contract/src/primitives/domain.rs
@@ -1,11 +1,14 @@
 use super::key_state::AuthenticatedParticipantId;
 use crate::errors::{DomainError, Error};
 use crate::primitives::participants::Participants;
-use near_mpc_contract_interface::types::{
-    Curve, DomainConfig, DomainId, DomainPurpose, Protocol, ReconstructionThreshold,
-};
+use near_mpc_contract_interface::types::{Curve, DomainConfig, DomainId, DomainPurpose, Protocol};
 use near_sdk::{log, near};
 use std::collections::BTreeMap;
+
+/// Lower bound on a domain's reconstruction threshold. `t = 1` would mean a
+/// single share is sufficient to reconstruct the secret, defeating the point
+/// of threshold cryptography.
+pub const MIN_RECONSTRUCTION_THRESHOLD: u64 = 2;
 
 /// Returns whether the given protocol is valid for the given purpose.
 pub fn is_valid_protocol_for_purpose(purpose: DomainPurpose, protocol: Protocol) -> bool {
@@ -50,7 +53,7 @@ pub fn validate_domain_threshold(
     num_participants: u64,
 ) -> Result<(), Error> {
     let t = domain.reconstruction_threshold.inner();
-    if t < 2 {
+    if t < MIN_RECONSTRUCTION_THRESHOLD {
         return Err(DomainError::ReconstructionThresholdTooLow.into());
     }
     if t > num_participants {
@@ -64,7 +67,7 @@ pub fn validate_domain_threshold(
         let required = t
             .checked_mul(2)
             .and_then(|x| x.checked_sub(1))
-            .unwrap_or(u64::MAX);
+            .ok_or(DomainError::ReconstructionThresholdOverflow { threshold: t })?;
         if required > num_participants {
             return Err(DomainError::InsufficientParticipantsForProtocol {
                 protocol: domain.protocol,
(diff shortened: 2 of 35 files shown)


crates/contract/src/primitives/domain.rs
```diff
@@ -55,21 +77,21 @@ impl DomainRegistry {
         &self.domains
     }
 
-    /// Migration from legacy: creates a DomainRegistry with a single ecdsa key.
-    pub fn new_single_ecdsa_key_from_legacy() -> Self {
-        let mut registry = Self::default();
-        registry.add_domain(Curve::Secp256k1, Protocol::CaitSith, DomainPurpose::Sign);
-        registry
-    }
```
gilcu3, line 63: this was dead code

crates/primitives/src/threshold.rs
```diff
@@ -35,3 +35,42 @@ impl Threshold {
...
+    pub const fn new(value: u64) -> Self {
+        Self(value)
+    }
+
+    pub fn inner(self) -> u64 {
+        self.0
+    }
+}
+
+impl From<Threshold> for ReconstructionThreshold {
```
gilcu3, line 72: Notice this is not done here, but this type `Threshold` will be renamed to `GovernanceThreshold` in the future

crates/primitives/src/threshold.rs
```diff
@@ -35,3 +35,42 @@ impl Threshold {
...
+
+impl ReconstructionThreshold {
+    pub const fn new(value: u64) -> Self {
+        Self(value)
+    }
+
+    pub fn inner(self) -> u64 {
+        self.0
+    }
+}
```
gilcu3, line 70: this was set this way in the design, don't mind changing it as it does not feel really rust ergonomic

crates/contract/src/v3_9_1_state.rs
```diff
@@ -313,81 +319,33 @@ impl From<OldThresholdParametersVotes> for ThresholdParametersVotes {
...
-impl From<OldResharingContractState> for ResharingContractState {
-    fn from(old: OldResharingContractState) -> Self {
-        ResharingContractState {
-            previous_running_state: old.previous_running_state.into(),
-            reshared_keys: old.reshared_keys,
-            resharing_key: old.resharing_key.into(),
-            cancellation_requests: old.cancellation_requests,
-        }
-    }
-}
```
gilcu3, line 362: realized these were not needed, for migration contract must be in Running state anyway

crates/contract/src/state/running.rs
```diff
@@ -372,4 +384,100 @@ pub mod running_tests {
             "Expected InvalidProtocolPurposeCombination, got: {err}"
         );
     }
+
+    fn proposal_with_threshold(
+        state: &super::RunningContractState,
```
netrome: Nit: let's not use relative imports this way. Either import the crate by name or the type directly.

crates/contract/src/v3_9_1_state.rs
netrome: Nice

netrome (dismissed):
Just a nit. Thank you!

crates/contract/src/state/running.rs
```diff
@@ -137,6 +137,15 @@ impl RunningContractState {
...
         self.parameters.validate_incoming_proposal(proposal)?;
 
+        // TODO(#3169): re-enable once resharing votes carry per-domain `t`
+        // and the node honors per-domain reconstruction thresholds (#3164).
+        // Until both are in place this check would block legitimate
+        // resharings that the node currently signs at the cluster threshold.
+        // let new_num_participants = proposal.participants().len() as u64;
+        // for domain in self.domains.domains() {
+        //     crate::primitives::domain::validate_domain_threshold(domain, new_num_participants)?;
+        // }
```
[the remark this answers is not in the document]
gilcu3: the threshold per-domain will only become effective in later PRs

crates/contract/src/v3_9_1_state.rs
```diff
@@ -716,18 +662,14 @@ mod tests {
...
         let running = old_running_state(participants, 2);
-        let old_state = OldProtocolContractState::Running(running);
-        let bytes = borsh::to_vec(&old_state).unwrap();
+        let bytes = borsh::to_vec(&running).unwrap();
 
         // When borsh-decoding and migrating to the new layout
-        let decoded: OldProtocolContractState = borsh::from_slice(&bytes).unwrap();
-        let migrated: ProtocolContractState = decoded.into();
+        let decoded: OldRunningContractState = borsh::from_slice(&bytes).unwrap();
+        let migrated: RunningContractState = decoded.into();
```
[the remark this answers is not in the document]
gilcu3: migration only happens if protocol state is Running, so this name is also fine

crates/primitives/src/threshold.rs
```diff
@@ -35,3 +35,42 @@ impl Threshold {
...
+    Deserialize,
+    BorshSerialize,
+    BorshDeserialize,
+)]
+#[cfg_attr(
+    all(feature = "abi", not(target_arch = "wasm32")),
+    derive(schemars::JsonSchema, borsh::BorshSchema)
+)]
+#[serde(transparent)]
+pub struct ReconstructionThreshold(u64);
```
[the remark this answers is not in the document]
gilcu3: this required a custom schemars implementation, which did not seem worth the effort, given that the values anyway require more runtime checks

crates/contract/src/primitives/domain.rs
```diff
@@ -40,6 +42,41 @@ pub fn validate_domain_consistency(domain: &DomainConfig) -> Result<(), Error> {
...
 
+/// Validates the per-domain reconstruction threshold against the participant
+/// count. Universal bound `2 <= t <= n` plus, for `DamgardEtAl`, the
+/// honest-majority bound `2t - 1 <= n`.
+pub fn validate_domain_threshold(
+    domain: &DomainConfig,
+    num_participants: u64,
+) -> Result<(), Error> {
+    let t = domain.reconstruction_threshold.inner();
+    if t < 2 {
```
DSharifi: Can you define `2` as a constant?

crates/contract/src/primitives/domain.rs
```diff
@@ -40,6 +42,41 @@ pub fn validate_domain_consistency(domain: &DomainConfig) -> Result<(), Error> {
...
+            threshold: t,
+            participants: num_participants,
+        }
+        .into());
+    }
+    if domain.protocol == Protocol::DamgardEtAl {
+        let required = t
+            .checked_mul(2)
+            .and_then(|x| x.checked_sub(1))
+            .unwrap_or(u64::MAX);
```
DSharifi: Why unwrap_or, instead of map to an error and return so explicitly failing for over/underflows?

crates/contract/src/primitives/domain.rs
```diff
@@ -84,8 +121,12 @@ impl DomainRegistry {
...
         for domain in domains {
             validate_domain_consistency(&domain)?;
-            let new_domain_id =
-                new_registry.add_domain(domain.curve, domain.protocol, domain.purpose);
+            let new_domain_id = new_registry.add_domain(
+                domain.curve,
+                domain.protocol,
+                domain.reconstruction_threshold,
+                domain.purpose,
+            );
```
DSharifi: Why not change the signature to take the whole domain struct instead of 4 fields of it?

crates/contract/src/errors.rs
```diff
@@ -204,6 +204,18 @@ pub enum DomainError {
         protocol: Protocol,
         purpose: DomainPurpose,
     },
+    #[error("Reconstruction threshold must be at least 2.")]
```
DSharifi: Would be good if you reference the constant I suggested adding instead of `2`

crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -436,4 +594,63 @@ mod tests {
...
+            "keyset": { "epoch_id": 0, "domains": [] },
+            "parameters": {
+                "participants": { "next_id": 0, "participants": [] },
+                "threshold": 5
+            },
+            "parameters_votes": { "proposal_by_account": {} },
+            "add_domains_votes": { "proposal_by_account": {} },
+            "previously_cancelled_resharing_epoch_id": null
+        }"#
+    }
```
DSharifi: Nit:
Could we use the json macro instead of this?

DSharifi (dismissed):
No hard blockers, but I'd strongly prefer adding the suggested constant and explicit failing if there's under/overflow instead of unwrap_or.

crates/contract/src/primitives/domain.rs
DSharifi: You can use `is_some_and` as an alternative to check if the predicate holds.

crates/contract/src/primitives/domain.rs
gilcu3: need to check if there is any reason for not passing the domain, cannot see any rn

crates/contract/src/primitives/domain.rs
```diff
@@ -1,12 +1,15 @@
...
-    Curve, DomainConfig, DomainId, DomainPurpose, Protocol, ReconstructionThreshold,
-};
+use near_mpc_contract_interface::types::{Curve, DomainConfig, DomainId, DomainPurpose, Protocol};
 use near_sdk::{log, near};
 use std::collections::BTreeMap;
 
+/// Lower bound on a domain's reconstruction threshold. `t = 1` would mean a
+/// single share is sufficient to reconstruct the secret, defeating the point
+/// of threshold cryptography.
+pub const MIN_RECONSTRUCTION_THRESHOLD: u64 = 2;
```
netrome, line 11: I know we reject 1 in the algorithms, but I'd think it would be interesting to allow singlesig domains (i.e. threshold 1) for low-latency use cases. There may be other implications of this ofc so it makes sense for now to explicitly disable it.

netrome (approved):
Thanks for the update!

DSharifi (approved):

crates/contract/src/primitives/domain.rs
gilcu3: > I know we reject 1 in the algorithms, but I'd think it would be interesting to allow singlesig domains (i.e. threshold 1) for low-latency use cases.

the problem is that for that case the threshold signature math simply doesn't work. So we would need to get creative just to get there

PR #3177, opened 2026-05-07
feat(config): support query-param auth and optional header scheme

### LINKED ISSUE #3181: Support AuthConfig::Query and optional Header scheme for RPC providers
### Background

  The MPC node supports multiple RPC providers per foreign chain via the AuthConfig enum — none / header / path / query
  (crates/node-config/src/foreign_chains/auth.rs). These four variants cover every mainstream provider's auth shape (Alchemy,
  Ankr, Blockdaemon, BlockPi, Chainstack, dRPC, Helius, Dwellir, GetBlock, Infura, QuickNode, Tatum, NowNodes, OnFinality).

  Two gaps in the implementation prevent operators from actually configuring several of them:

  1. AuthConfig::Query is declared but unimplemented — auth_config_to_rpc_auth in crates/node/src/config.rs:472 bails at
  runtime with "this is not supported yet". This blocks any provider whose API key is a query parameter: Helius (?api-key=),
  OnFinality (?apikey=), dRPC (?dkey=).
  2. Header.scheme silently defaults to "Bearer" when omitted — config.rs:459 does scheme.unwrap_or_else(||
  "Bearer".to_string()), producing the wire value Bearer <token> even when the operator sets scheme: ~. This breaks providers
  that expect a raw token: Tatum (x-api-key: <token>), NowNodes (api-key: <token>).

### User Story

_No response_

### Acceptance Criteria

  - AuthConfig::Query produces a working jsonrpsee::HttpClient: token is appended to rpc_url as a query parameter, with proper
  URL-encoding and correct ? vs & separator
  - AuthConfig::Header with scheme: ~ emits the raw token as the header value (no implicit Bearer prefix); scheme: Bearer
  continues to emit Bearer <token>
  - Unit tests for both fixes — at minimum: query-on-clean-URL, query-on-URL-with-existing-params,
  query-encoding-special-chars, header-without-scheme
  - All 14 listed providers can be wired up using only the existing four AuthConfig variants — no new enum variant or
  per-provider code path required

_No response_

anodar:
Closes: https://github.com/near/mpc/issues/3181

### DIFF
--- a/crates/node/src/config.rs
+++ b/crates/node/src/config.rs
@@ -474,10 +474,17 @@ pub fn auth_config_to_rpc_auth(
         }
         AuthConfig::Query { name, token } => {
             let token_value = token.resolve()?;
-            let mut parsed = url::Url::parse(rpc_url)
-                .with_context(|| format!("invalid RPC URL: `{rpc_url}`"))?;
-            parsed.query_pairs_mut().append_pair(&name, &token_value);
-            *rpc_url = parsed.into();
+            // Validate the URL but preserve original bytes — using `Url::parse` then
+            // writing back its `to_string()` would normalize host casing, append a
+            // trailing `/`, etc. The other arms leave the URL byte-identical aside
+            // from placeholder substitution, so do the same here.
+            url::Url::parse(rpc_url).with_context(|| format!("invalid RPC URL: `{rpc_url}`"))?;
+            let encoded_pair = url::form_urlencoded::Serializer::new(String::new())
+                .append_pair(&name, &token_value)
+                .finish();
+            let separator = if rpc_url.contains('?') { '&' } else { '?' };
+            rpc_url.push(separator);
+            rpc_url.push_str(&encoded_pair);
             Ok(RpcAuthentication::KeyInUrl)
         }
     }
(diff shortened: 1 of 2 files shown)


crates/node/src/config.rs
```diff
@@ -469,7 +472,21 @@ pub fn auth_config_to_rpc_auth(
...
+            // writing back its `to_string()` would normalize host casing, append a
+            // trailing `/`, etc. The other arms leave the URL byte-identical aside
+            // from placeholder substitution, so do the same here.
+            url::Url::parse(rpc_url).with_context(|| format!("invalid RPC URL: `{rpc_url}`"))?;
+            let encoded_pair = url::form_urlencoded::Serializer::new(String::new())
+                .append_pair(&name, &token_value)
+                .finish();
+            let separator = if rpc_url.contains('?') { '&' } else { '?' };
+            rpc_url.push(separator);
+            rpc_url.push_str(&encoded_pair);
```
DSharifi: Instead of manually adding the query parameter, you can just work with the parsed url type.

```suggestion
            let mut parsed_rpc_url = url::Url::parse(rpc_url)
                .with_context(|| format!("invalid RPC URL: `{rpc_url}`"))?;
            parsed_rpc_url
                .query_pairs_mut()
                .append_pair(&name, &token_value);
            *rpc_url = parsed_rpc_url.as_str().to_string();
```

DSharifi (changes requested):
Looks good, thanks for fixing. Can you change the parsing of the query parameters to use the URL crate, since it's already used to parse the URL instead of discarding the parsed result?

crates/node/src/config.rs
anodar: I've done it like that to preserve host casing, there was a comment on it above:

>  Validate the URL but preserve original bytes — using `Url::parse` then
            writing back its `to_string()` would normalize host casing
            
But I guess it doesn't matter much so changed to your suggestion.

DSharifi (approved):

netrome (approved):
Nice, thanks!

PR #3185, opened 2026-05-08
fix: Don't serve rpc providers information over public debug endpoint

### LINKED ISSUE #3183: Don't serve RPC provider informations in public debug endpoint
### Background

`/debug/node_config` endpoint implemented in https://github.com/near/mpc/pull/2329 serves information about RPC providers.

We don't want to have this publicly available for security reasons.

### User Story

_No response_

### Acceptance Criteria

Don't serve RPC provider info over debug endpoint from the node.

_No response_

anodar:
Closes: https://github.com/near/mpc/issues/3183

### DIFF
--- a/crates/node/src/web.rs
+++ b/crates/node/src/web.rs
@@ -305,6 +305,27 @@ mod tests {
     use std::num::NonZeroU64;
     use std::str::FromStr;
 
+    const PROVIDER_ALCHEMY: &str = "alchemy";
+    const PROVIDER_ANKR: &str = "ankr";
+    const PROVIDER_BLAST: &str = "blast";
+    const PROVIDER_PUBLIC: &str = "public";
+
+    const SOLANA_RPC_URL: &str = "https://solana-mainnet.g.alchemy.com/v2/";
+    const BITCOIN_RPC_URL: &str = "https://rpc.ankr.com/btc/{api_key}";
+    const ETHEREUM_RPC_URL: &str = "https://eth-mainnet.g.alchemy.com/v2/";
+    const ABSTRACT_RPC_URL: &str = "https://api.testnet.abs.xyz";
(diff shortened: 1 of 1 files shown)


DSharifi:
Agree with claude, it's not a refactor removing functionality

anodar:
> Agree with claude, it's not a refactor removing functionality

Claude's suggestion was refactor (or fix) :D Changed to fix. Although Fix doesn't really fit either, it's not fixing anything it worked as intended before, now our requirements changed and we change the code. Maybe chore or feat is best, but Claude didn't like chore that I initially had either.

crates/node/src/web.rs
```diff
@@ -511,183 +448,54 @@ mod tests {
...
-
-        // Then
-        let solana = response.foreign_chains.solana.unwrap();
-        assert_eq!(solana.timeout_sec.get(), 30);
-        assert_eq!(solana.max_retries.get(), 3);
-    }
-
-    #[test]
-    fn node_config_response_json__does_not_contain_auth() {
+    fn node_config_response_json__does_not_leak_foreign_chain_info() {
```
barakeinav1: nit: per CLAUDE.md, new test names must use the `<system_under_test>__should_<assertion>` form. Non-blocking; the rest of the file pre-dates the convention.

```suggestion
    fn node_config_response_json__should_not_leak_foreign_chain_info() {
```

barakeinav1 (dismissed):
LGTM, approving, with one small nit.

crates/node/src/web.rs
anodar: Done.

DSharifi (approved):
Thank you!

barakeinav1 (approved):
thanks

PR #3186, opened 2026-05-08
feat: fan out to all rpc providers

### LINKED ISSUE #3197: Foreign chain inspector crate should querying multiple inspectors
### Background

We want to change the foreign chain validation logic on the node side such that every node queries every RPC provider it is configured for. Since that is the case, there is no point for the node to create multiple inspector instances anymore where each instance only has one client handle.

### User Story

_No response_

### Acceptance Criteria

All RPC providers for a given chain is used during a foreign chain validation request on the node.

_No response_

DSharifi:
closes #3197

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3605,6 +3605,7 @@ dependencies = [
  "httpmock",
  "jsonrpsee",
  "mpc-primitives",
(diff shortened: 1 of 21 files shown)


DSharifi:
@claude review

PR #3187, opened 2026-05-09
feat: Fan-out responses to all duplicate requests

### LINKED ISSUE #1306: don't sleep in integration tests

### LINKED ISSUE #3184: Duplicate requests should be processed and responded to
### Background

Currently, if the same predecessor sends the same request twice in a short time frame, our contract overwrites the yield index of the first request making the first request time out and only responding to the second request.

### User Story

As an MPC caller I expect to get signature responses for each of my requests, even if some requests happens to be duplicate.

### Acceptance Criteria

The MPC system returns responses to all requests even when requests are duplicated and in close proximity.

_No response_

netrome:
Almost ready: Done several iterations of self-review, updates and local LLM reviews. Now ready for public LLM reviews.

closes #3184 

This PR fans out node responses to all matching requests in the presence of duplicate requests.

Note: This doesn't prevent nodes from processing identical requests, but the first successful response would resolve all identical requests. As a follow-up optimization we could make nodes ignore duplicate requests, but we'd need to be very careful with how we implement that to not risk missing any signature requests.


crates/contract/src/lib.rs
```diff
@@ -138,39 +141,20 @@ impl Default for MpcContract {
...
+    pending_signature_requests: LookupMap<SignatureRequest, Vec<YieldIndex>>,
+    pending_ckd_requests: LookupMap<CKDRequest, Vec<YieldIndex>>,
+    pending_verify_foreign_tx_requests: LookupMap<VerifyForeignTransactionRequest, Vec<YieldIndex>>,
     proposed_updates: ProposedUpdates,
     node_foreign_chain_support: SupportedForeignChainsByNode,
     config: Config,
     tee_state: TeeState,
     accept_requests: bool,
     node_migrations: NodeMigrations,
-    stale_data: StaleData,
```
netrome, line 150: Note: At first Claude extended the `StaleData` struct but the docs of it was misaligned with the current use case so I found it cleaner to use a dedicated struct instead. I chose to remove the `StaleData` since it's currently unused and risks leading to confusion if misused. It's easy to re-add it if/when we'd need it.

crates/contract/src/pending_requests.rs
```diff
@@ -0,0 +1,164 @@
...
+        for YieldIndex { data_id } in yield_indices {
+            env::promise_yield_resume(&data_id, response_bytes.clone());
+        }
+        Ok(())
+    } else if let Some(YieldIndex { data_id }) = legacy_map.remove(request) {
+        // Legacy entries never accepted duplicates, so there is at most one yield.
+        env::promise_yield_resume(&data_id, response_bytes);
+        Ok(())
+    } else {
+        Err(InvalidParameters::RequestNotFound.into())
```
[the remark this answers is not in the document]
netrome: Fixed in https://github.com/near/mpc/pull/3187/commits/fa1cc830a9757eeaaa3c9bf41468e3f37094b945

crates/contract/src/pending_requests.rs
```diff
@@ -0,0 +1,164 @@
...
+pub(crate) fn pop_one_yield_for<K>(
+    new_map: &mut LookupMap<K, Vec<YieldIndex>>,
+    legacy_map: &mut LookupMap<K, YieldIndex>,
+    request: &K,
+) where
+    K: BorshSerialize + BorshDeserialize + Clone + Ord,
+{
+    if !pop_oldest_pending_yield(new_map, request) {
+        legacy_map.remove(request);
+    }
```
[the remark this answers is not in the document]
netrome: Fixed in https://github.com/near/mpc/pull/3187/commits/fa1cc830a9757eeaaa3c9bf41468e3f37094b945

crates/contract/src/pending_requests.rs
```diff
@@ -0,0 +1,203 @@
...
+    },
+    storage_keys::StorageKey,
+};
+
+/// Maximum number of concurrent yield-resume promises that can be queued for a single
+/// request key (i.e. the number of duplicate submissions whose responses fan out from
+/// one MPC reply). This is a somewhat arbitrary bound, set high enough so that we should
+/// never hit it in practice (famous last words...), while still providing a ceiling to
+/// keep the theoretical number of resumed promises bounded.
+pub(crate) const MAX_PENDING_REQUEST_FAN_OUT: u8 = 128;
```
barakeinav1: nit: could we add a sandbox test that fills the queue to `MAX_PENDING_REQUEST_FAN_OUT` and asserts `respond*` fits in 300 Tgas? The worst case is cheaply reachable by an attacker (1 yocto + std sign gas per dup), so empirical verification would be nice to have.

crates/contract/src/pending_requests.rs
```diff
@@ -0,0 +1,203 @@
...
+/// keys. These fields are rooted at the *previous* storage keys with the *previous*
+/// singular value type, so requests that were already yielded before the upgrade
+/// can still be answered through `respond*` (or cleaned up by their timeout
+/// handler) until they expire.
+///
+/// Unlike a write-once cleanup buffer, these maps are read and written on every
+/// `respond*`, every timeout handler, and every `get_pending_*` call — they are
+/// operationally live for the lifetime of the legacy window. They should be
+/// dropped (along with the now-unused storage keys) in the next upgrade after
+/// this has been released — same lifecycle as the V2→V3 cleanup in #2940.
```
barakeinav1, line 72: nit: could we open a tracking issue for dropping `LegacyPendingRequests` + the old storage keys in the next-but-one upgrade and reference it as `TODO(#NNNN)` here, so it doesn't get forgotten between releases?

barakeinav1:
nit: doc drift in `crates/contract/src/v3_9_1_state.rs:365-367` — the `OldStaleData` doc still says "the new `StaleData` is empty", but the new contract no longer has a `StaleData` field at all. Suggested replacement:

```rust
/// Previous `StaleData` layout — held a [`LookupMap`] of pre-upgrade signature requests.
/// After the v3.9 migration is fully deployed those requests have been resolved or timed
/// out, so the field is dropped here. The new contract no longer has a `StaleData` field;
/// pre-upgrade pending requests are surfaced through `LegacyPendingRequests` instead.
```

crates/contract/src/lib.rs
```diff
@@ -2761,6 +2816,406 @@ mod tests {
...
+            .pending_signature_requests
+            .get(&signature_request)
+            .is_none());
+    }
+
+    #[test]
+    fn return_signature_and_clean_state_on_success__should_prefer_new_map_when_both_have_entries() {
+        // Given: the same request key has yields in both the new fan-out map and the
+        // legacy fallback (an edge case where a caller resubmitted post-upgrade
+        // while a legacy yield is still outstanding).
```
barakeinav1: nit: post-`fa1cc830` this state is unreachable from normal contract use — `push_pending_yield` migrates legacy entries on first push. Worth noting so the next reader doesn't think this documents a live edge case.

```suggestion
        // Given: the same request key has yields in both the new fan-out map and the
        // legacy fallback. This state cannot arise in production (`push_pending_yield`
        // migrates legacy entries into the new queue on first push) — the test sets it
        // up via direct insert to guard the single-map invariant against future regressions.
```

crates/contract/src/pending_requests.rs
netrome: Fair point, will add in ~1 hour (going out to take the kid to preschool first)

crates/contract/src/pending_requests.rs
netrome: Sure thing, I'll do it.

netrome:
> nit: doc drift in `crates/contract/src/v3_9_1_state.rs:365-367` — the `OldStaleData` doc still says "the new `StaleData` is empty", but the new contract no longer has a `StaleData` field at all. Suggested replacement:
> 
> ```rust
> /// Previous `StaleData` layout — held a [`LookupMap`] of pre-upgrade signature requests.
> /// After the v3.9 migration is fully deployed those requests have been resolved or timed
> /// out, so the field is dropped here. The new contract no longer has a `StaleData` field;
> /// pre-upgrade pending requests are surfaced through `LegacyPendingRequests` instead.
> ```

Thanks for spotting, will fix

crates/contract/src/lib.rs
netrome: Fair point, will update

barakeinav1 (dismissed):
I'm not familiar enough with this area to challenge the underlying design too much or propose a better solution, but from what I see it solves the issue from a user's perspective — all duplicate requests will be handled correctly. Approving with a few nits.

crates/contract/src/lib.rs
netrome: Actually since this is unreachable in production now let's remove it altogether.

netrome: Removed in https://github.com/near/mpc/pull/3187/commits/f8bc81bb20aa0d26bcc16de6a02f80fb10e2e811

netrome:
> > nit: doc drift in `crates/contract/src/v3_9_1_state.rs:365-367` — the `OldStaleData` doc still says "the new `StaleData` is empty", but the new contract no longer has a `StaleData` field at all. Suggested replacement:
> > ```rust
> > /// Previous `StaleData` layout — held a [`LookupMap`] of pre-upgrade signature requests.
> > /// After the v3.9 migration is fully deployed those requests have been resolved or timed
> > /// out, so the field is dropped here. The new contract no longer has a `StaleData` field;
> > /// pre-upgrade pending requests are surfaced through `LegacyPendingRequests` instead.
> > ```
> 
> Thanks for spotting, will fix

Fixed in https://github.com/near/mpc/pull/3187/commits/d73c22912266ffb8fb7219e275803f55f5155cec

crates/contract/src/pending_requests.rs
netrome: Looked into this, it will be more complicated than I anticipated to test this due to limitations in our testing setup.

Since this is a very unlikely scenario, I'd prefer to push the system test for this scenario to a follow-up. I'll write up an issue for it.

netrome: Created https://github.com/near/mpc/issues/3206

crates/contract/src/pending_requests.rs
netrome: Created https://github.com/near/mpc/issues/3207

crates/contract/tests/sandbox/foreign_chain_request.rs
```diff
@@ -99,6 +99,98 @@ async fn verify_foreign_transaction__should_succeed(
...
+        .transact_async()
+        .await
+        .unwrap();
+    let status_bob = bob
+        .call(
+            setup.contract.id(),
+            method_names::VERIFY_FOREIGN_TRANSACTION,
+        )
+        .args_json(json!({ "request": request_args }))
+        .deposit(NearToken::from_yoctonear(1))
```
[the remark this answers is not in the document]
netrome: This does compile.

crates/contract/src/pending_requests.rs
```diff
@@ -0,0 +1,203 @@
...
+/// and time out in that same order — so the timing-out yield is always the head.
+fn pop_oldest_pending_yield<K>(map: &mut LookupMap<K, Vec<YieldIndex>>, request: &K) -> bool
+where
+    K: BorshSerialize + BorshDeserialize + Clone + Ord,
+{
+    let became_empty = match map.get_mut(request) {
+        Some(queue) if !queue.is_empty() => {
+            queue.remove(0);
+            queue.is_empty()
+        }
```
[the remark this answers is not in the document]
netrome: Not worth it for small vectors.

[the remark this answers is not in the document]
netrome: Okay this is getting annoying. Turning off automatic re-reviews.

crates/contract/src/lib.rs
```diff
@@ -328,9 +316,7 @@ impl MpcContract {
             .expect("read_register failed")
             .try_into()
             .expect("conversion to CryptoHash failed");
-        if insert(self, return_id) {
-            log!("request already present, overriding callback.")
-        }
+        insert(self, return_id);
 
         env::promise_return(promise_index);
```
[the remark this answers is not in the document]
netrome: Duplicate comment, already chosen not to do this.

barakeinav1 (dismissed):
Re-approving — all comments addressed. Nit fixes applied cleanly (one verbatim, one by removing the test outright), and the deferred items are now tracked in #3206 (gas-budget sandbox test) and #3207 (legacy cleanup). Thanks!

crates/contract/src/pending_requests.rs
kevindeforth: This seems too high.
We append 7 Tgas for a call to `return_signature_and_clean_state_on_success`, which would put a response fulfilling all of these requests at 896 Tgas, much higher than what the NEAR blockchain currently supports.

crates/contract/src/pending_requests.rs
```diff
@@ -0,0 +1,203 @@
...
+        for YieldIndex { data_id } in yield_indices {
+            env::promise_yield_resume(&data_id, response_bytes.clone());
+        }
+        Ok(())
+    } else if let Some(YieldIndex { data_id }) = legacy_map.remove(request) {
+        // Legacy entries never accepted duplicates, so there is at most one yield.
+        env::promise_yield_resume(&data_id, response_bytes);
+        Ok(())
+    } else {
+        Err(InvalidParameters::RequestNotFound.into())
```
kevindeforth: We should make sure that the first request always gets resolved. Subsequent requests may run out of gas.
I would suggest to resolve the first one here, and spawn a promise to resolve subsequent requests (which may run out of gas).

kevindeforth (commented):
Thanks for this.
I would suggest to lower the supported number of parallel requests, because we risk running out of gas with the current logic.

crates/contract/src/pending_requests.rs
netrome: Good point, let's lower this to 4 for the immediate fix and add a test case to ensure we don't run out of gas then.

crates/contract/src/pending_requests.rs
netrome: Good idea, let's do it.

crates/contract/src/pending_requests.rs
netrome: Okay I think there's a flaw in the logic here, got this explanation from Claude which seems to check out:

TL;DR - the 7 Tgas comes from the caller of the `sign` method, not the caller of the `resume` method. So this should not block us from resuming 128 sign requests.

```
  Where the 7 Tgas actually comes from. The 7 Tgas is attached at sign() time via env::promise_yield_create (crates/contract/src/lib.rs:307, called from lib.rs:374-385). The user attaches
  gas to their sign() call, and 7 Tgas of it is reserved for their yielded callback's receipt. When respond() resolves the queue, each promise_yield_resume
  (crates/contract/src/pending_requests.rs:174) just unblocks an already-pre-paid callback — it doesn't add 7 Tgas onto the responder's transaction. So 128 × 7 = 896 Tgas isn't a
  single-transaction figure; it's spread across 128 receipts, each prepaid by a different signer. There's no NEAR limit being blown by that.
```

netrome: I'll validate this further and get the test case in place to test it.

netrome: (potentially just adopting the test in https://github.com/near/mpc/pull/3218/changes)

kevindeforth: Oh, yeah, I completely forgot about that. That's very cool!

crates/contract/Cargo.toml
```diff
@@ -40,6 +40,11 @@ crate-type = ["cdylib", "lib"]
 test-utils = ["rand", "threshold-signatures", "near-mpc-contract-interface/blstrs"]
 # WASM-compatible benchmark endpoints for sandbox gas testing
 bench-contract-methods = []
+# Sandbox-only view methods that expose internal contract state for tests that need to
+# assert invariants that aren't visible through the production API surface (e.g. the
+# length of a fan-out queue). Distinct from `bench-contract-methods` because these are
+# behavioral hooks, not gas-measurement hooks.
+sandbox-test-methods = []
```
kevindeforth, line 47: Q: What is the reasoning for introducing a new feature-flag, instead of re-using the `test-utils`?

crates/contract/src/errors.rs
```diff
@@ -114,6 +114,8 @@ pub enum InvalidParameters {
     CallerNotSigner,
     #[error("Requested foreign chain, {requested:?}, is not supported.")]
     ForeignChainNotSupported { requested: ForeignChain },
+    #[error("Pending-request queue is full for this request key (limit: {limit}). Try again once an in-flight response or timeout has cleared room.")]
+    PendingRequestQueueFull { limit: u8 },
```
kevindeforth: I am not sure `InvalidParameters` is the right place for this error. How about adding it to the `RequestError` enum?

crates/contract/README.md
```diff
@@ -259,8 +259,8 @@ stateDiagram-v2
...
+| `sign(request: SignRequestArgs)`                                                             | Submits a signature request to the contract. Requires a deposit of 1 yoctonear. Duplicate submissions of the same request (same caller, domain, path, and payload) while an earlier one is still pending are queued and all receive the same response when the MPC nodes reply; the queue is bounded — concurrent duplicates beyond that bound are rejected with `PendingRequestQueueFull`.             | deferred to promise        | `10 Tgas`       | `~7 Tgas`          |
+| `request_app_private_key(request: CKDRequestArgs)`                                           | Submits a confidential key derivation (ckd) request to the contract. Requires a deposit of 1 yoctonear. Duplicate submissions of the same request (same caller, domain, derivation path, and app public key) while an earlier one is still pending are queued and all receive the same response when the MPC nodes reply; the queue is bounded — concurrent duplicates beyond that bound are rejected with `PendingRequestQueueFull`. | deferred to promise        | `10 Tgas`       | `~7 Tgas`          |
```
kevindeforth, line 263: Are we missing the foreign transaction methods in this README?

crates/contract/src/pending_requests.rs
```diff
@@ -0,0 +1,211 @@
...
+        Some(queue) if !queue.is_empty() => {
+            queue.remove(0);
+            queue.is_empty()
+        }
+        _ => return false,
+    };
+    if became_empty {
+        map.remove(request);
+    }
+    true
```
kevindeforth: _edit: suggested changes in https://github.com/near/mpc/pull/3232_


While there doesn't seem to be a valid code path for it right now, I would prefer if this method removed empty queues eagerly:

```suggestion
    let Some(queue) = map.get_mut(request) else {
        return false;
    };
    if queue.is_empty() {
        map.remove(request);
        return false;
    }
    queue.remove(0);
    if queue.is_empty() {
        map.remove(request);
    }
    true
```

crates/contract/src/pending_requests.rs
```diff
@@ -0,0 +1,211 @@
...
+            env::promise_yield_resume(&data_id, response_bytes.clone());
+        }
+        Ok(())
+    } else if let Some(YieldIndex { data_id }) = legacy_map.remove(request) {
+        // Legacy entries never accepted duplicates, so there is at most one yield.
+        env::promise_yield_resume(&data_id, response_bytes);
+        Ok(())
+    } else {
+        Err(InvalidParameters::RequestNotFound.into())
+    }
```
kevindeforth: _edit: suggested changes in https://github.com/near/mpc/pull/3232_

If we always empty both maps, then we don't need to migrate entries from the old to the new map, simplifying `push_pending_yield`

```suggestion
let resumed = new_map.remove(request).unwrap_or_default()
    .into_iter()
    .chain(legacy_map.remove(request))
    .map(|YieldIndex { data_id }| {
        env::promise_yield_resume(&data_id, response_bytes.clone());
    })
    .count() > 0;

if resumed { Ok(()) } else { Err(InvalidParameters::RequestNotFound.into()) }
```

crates/contract/src/pending_requests.rs
kevindeforth: Closing this, as it is no longer relevant, given that the gas is pre-paid. (c.f. https://github.com/near/mpc/pull/3187#discussion_r3239810936)

crates/contract/src/pending_requests.rs
```diff
@@ -0,0 +1,211 @@
...
+            );
+        }
+        queue.push(new_yield);
+    } else {
+        let queue = match legacy_map.remove(&request) {
+            Some(legacy_yield) => vec![legacy_yield, new_yield],
+            None => vec![new_yield],
+        };
+        new_map.insert(request, queue);
+    }
```
kevindeforth, line 118: _edit: added to suggested changes in https://github.com/near/mpc/pull/3232_

We can simplify this and simplifying `resolve_yields_for` by only adding to the new queue:

```suggestion
pub(crate) fn push_pending_yield<K>(
    new_map: &mut LookupMap<K, Vec<YieldIndex>>,
    request: K,
    data_id: CryptoHash,
) where
    K: BorshSerialize + BorshDeserialize + Clone + Ord,
{
    let new_yield = YieldIndex { data_id };
    let queue = new_map.entry(request).or_default();
    if queue.len() >= usize::from(MAX_PENDING_REQUEST_FAN_OUT) {
        env::panic_str(
            &InvalidParameters::PendingRequestQueueFull {
                limit: MAX_PENDING_REQUEST_FAN_OUT,
            }
            .to_string(),
        );
    }
    queue.push(new_yield);
```

crates/contract/src/pending_requests.rs
```diff
@@ -0,0 +1,211 @@
...
+pub(crate) fn pop_one_yield_for<K>(
+    new_map: &mut LookupMap<K, Vec<YieldIndex>>,
+    legacy_map: &mut LookupMap<K, YieldIndex>,
+    request: &K,
+) where
+    K: BorshSerialize + BorshDeserialize + Clone + Ord,
+{
+    if !pop_oldest_pending_yield(new_map, request) {
+        legacy_map.remove(request);
+    }
```
kevindeforth: _edit: added to suggested changes in: https://github.com/near/mpc/pull/3232_

and this would invert:
```suggestion
   if legacy_map.remove(request).is_none() {
       pop_oldest_pending_yield(new_map, request);
   }
```

kevindeforth (commented):
It's a lot to review, am halfway through. Taking a short break.

crates/contract/src/lib.rs
```diff
@@ -2761,6 +2823,369 @@ mod tests {
...
+    #[test]
+    fn sign__should_queue_duplicate_requests_and_drain_all_on_respond() {
+        // Given
+        let (context, mut contract, secret_key) =
+            basic_setup_with_protocol(Protocol::CaitSith, DomainPurpose::Sign, &mut OsRng);
+        let SharedSecretKey::Secp256k1(secret_key) = secret_key else {
+            unreachable!();
+        };
+        let payload = Payload::from_legacy_ecdsa([7u8; 32]);
+        let path = "m/44'\''/60'\''/0'\''/0/0".to_string();
```
kevindeforth: Q: Is there a reason not to use a human-readable path, such as `"duplicate_requests_test"`?

crates/contract/src/lib.rs
```diff
@@ -2761,6 +2823,369 @@ mod tests {
...
+        contract
+            .respond(signature_request.clone(), signature_response)
+            .expect("respond should succeed");
+
+        // Then: the entire queue is drained — both queued yields received the response.
+        assert!(contract
+            .pending_signature_requests
+            .get(&signature_request)
+            .is_none());
+    }
```
kevindeforth, line 2887: _edit: added to suggested changes in: https://github.com/near/mpc/pull/3232_

A lot of this seems to be copied from `test_signature_common`. Would be nice to make refactor the relevant parts (request and response generation, respectively), into their own methods to reduce code duplication.

crates/contract/tests/sandbox/sign.rs
```diff
@@ -68,55 +67,6 @@ async fn test_contract_request_all_schemes() -> anyhow::Result<()> {
...
-        worker
-            .fast_forward(NUM_BLOCKS_BETWEEN_REQUESTS)
-            .await
-            .unwrap();
-        let status_2 = req
-            .submit_request_ensure_included(&alice, &contract)
-            .await?;
-
-        // unfortunately, we still can't completely get rid of this sleep
-        // TODO(#1306): remove the need to sleep
```
kevindeforth, line 102: Can we close #1306 now?

kevindeforth (commented):
Nice, thank you.

crates/contract/Cargo.toml
kevindeforth: I guess it's to be explicit and so we don't compile all test features into the wasm binary.

pbeza:
From what I can see, reviewing this PR and addressing the comments will take a bit more time, so I guess we should hold off on the release until that’s done.

crates/contract/src/lib.rs
kevindeforth: I didn't find one, addressed in https://github.com/near/mpc/pull/3232

crates/contract/README.md
kevindeforth: added to suggested changes in https://github.com/near/mpc/pull/3232

crates/contract/src/errors.rs
kevindeforth: added to suggested changes in https://github.com/near/mpc/pull/3232

kevindeforth (commented):
Thanks a lot for this.
I collected the suggested changes in #3232, containing the remaining blockers for approval.

Note that this PR should also close #1306, as we remove the only remaining `sleep` in the sandbox tests: https://github.com/near/mpc/pull/3187#discussion_r3241157836

crates/contract/Cargo.toml
netrome: Yeah, this is analogous to the `bench-contract-methods` but for sandbox tests and not benchmarks. Iirc we did use the `test-utils` feature flag for these methods in the original bench PR, but would need to fact check this as I may misremember.

In either case, I don't see any problem with having a few pretty explicit and narrow feature flags but I have no strong opinion against unifying test/bench helpers under a single one either.

crates/contract/src/errors.rs
netrome: Good point, will have a look :eyes:

crates/contract/README.md
netrome: Thanks!

crates/contract/src/lib.rs
netrome: Yeah probably not, thanks!

crates/contract/src/lib.rs
netrome: Definitely. I wanted to be mindful about refactor noise in this PR, but this seems worth it.

crates/contract/tests/sandbox/sign.rs
netrome: image

netrome: ...after this PR is merged

crates/contract/src/lib.rs
netrome: Fixed in https://github.com/near/mpc/pull/3187/commits/bf58148145601675858f767cbf0322148479fdd0

crates/contract/src/pending_requests.rs
netrome: Fixed in https://github.com/near/mpc/pull/3187/commits/bf58148145601675858f767cbf0322148479fdd0

crates/contract/src/pending_requests.rs
netrome: Fixed in https://github.com/near/mpc/pull/3187/commits/bf58148145601675858f767cbf0322148479fdd0

crates/contract/src/pending_requests.rs
netrome: Fixed in https://github.com/near/mpc/pull/3187/commits/bf58148145601675858f767cbf0322148479fdd0

crates/contract/src/pending_requests.rs
netrome: Fixed in https://github.com/near/mpc/pull/3187/commits/bf58148145601675858f767cbf0322148479fdd0

crates/contract/README.md
netrome: Fixed in https://github.com/near/mpc/pull/3187/commits/bf58148145601675858f767cbf0322148479fdd0

crates/contract/src/errors.rs
netrome: Fixed in https://github.com/near/mpc/pull/3187/commits/bf58148145601675858f767cbf0322148479fdd0

kevindeforth (approved):
Thank you!

crates/contract/src/sandbox_test_methods.rs
```diff
@@ -0,0 +1,52 @@
...
+
+#[near]
+impl MpcContract {
+    /// Returns the number of yields queued under `request` in the fan-out map.
+    ///
+    /// Used by the duplicate-request sandbox test to poll until the full fan-out queue
+    /// has landed before calling `respond`, replacing a previous wall-clock sleep. The
+    /// legacy single-yield map is not consulted: post-upgrade pushes drain it into the
+    /// head of the new queue, so the new map's length is authoritative for the test's
+    /// purposes.
```
barakeinav1, line 27: nit: this paragraph is stale — the dual-map drain refactor (#3232, merged in here) removed the legacy→new migration, so the original reasoning no longer applies. The view is still correct for sandbox use; just the *why* needs a tweak.

```suggestion
    /// The legacy single-yield map is not consulted; for sandbox tests starting from
    /// fresh state the new map's length is authoritative.
```

barakeinav1 (approved):

crates/contract/src/lib.rs
gilcu3: The idea with `StaleData` since it was introduced by @DSharifi was to keep it there, and extend the struct when something needed to be cleaned. Not sure why the docs were misaligned with current use?

PR #3188, opened 2026-05-10
feat(node): enable decentralised state sync (DSS)

### LINKED ISSUE #1734: bug: In TEE CVM mode, the MPC node is not able to use DSS
### Background

Related to [#1724](https://github.com/near/mpc/issues/1724).

after disabling bucket state sync. and also opening port 24567, the MPC mode. is not able to state sync via DSS (decentralized state sync)

note - this maybe also be the behavior for MPC node not running in CVM. (but haven't check this yet).

see slack. 

### User Story

As an MPC operator or user, and I'l the node to be able to state sync in a more decentralized manager, avoiding a single point of failure.

### Acceptance Criteria

DSS is enabled and working 

_No response_

barakeinav1:
DSS state sync is currently disabled in practice because `apply_near_config_patches` hardcoded `external_storage_fallback_threshold = 0` (bucket-only); on multi-IP hosts neard's auto-discovery also picks the wrong outbound IP for Tier3 replies. This PR adds two optional fields so operators can opt into DSS and override the advertised Tier3 address.

- Adds `tier3_public_addr` and `external_storage_fallback_threshold` to `NearInitConfig` (under `[mpc_node_config.near_init]` in the launcher TOML).
- Threads both through `deploy-tee-cluster.sh`, the testnet TOML template, and the example `user-config.toml`; lists them in the TDX operator guide.
- 3 new unit tests covering the new fields.
- Validated end-to-end on a 2-node TDX testnet cluster: DSS sync, full DKG, on-chain ECDSA signature returned ([tx](https://explorer.testnet.near.org/transactions/24jPGKYiGSSsE8kHNSr9VnrcCdVoKFU7TkpHmbu4pLCW)).

Stacked on #3189.

Closes #1734.

### DIFF
--- a/crates/node-config/src/start.rs
+++ b/crates/node-config/src/start.rs
@@ -108,6 +108,19 @@ pub struct NearInitConfig {
     /// Override the NEAR node network (indexer) listen address (e.g. "0.0.0.0:24568").
     /// Useful when running multiple nodes on the same machine.
     pub network_addr: Option<String>,
+    /// Override the public address advertised for Tier3 state-sync responses
+    /// (e.g. "203.0.113.5:24567"). Required on multi-IP hosts where outbound
+    /// source IP differs from the bound IP — auto-discovery picks the
+    /// outbound IP and DSS times out. Patches into nearcore's
+    /// `network.experimental.tier3_public_addr` config field.
+    pub tier3_public_addr: Option<String>,
+    /// Override how many P2P (DSS) state-sync attempts to make before falling
+    /// back to the external storage bucket. `0` (current default) means
+    /// "go straight to bucket, never use DSS." A moderate value enables
+    /// DSS-first with bucket as a safety net; a very large value effectively
+    /// disables the bucket fallback. Patches into nearcore's
+    /// `state_sync.sync.ExternalStorage.external_storage_fallback_threshold`.
+    pub external_storage_fallback_threshold: Option<u64>,
 }
 
 /// Encryption keys needed at startup.
--- a/crates/node/src/config/start.rs
+++ b/crates/node/src/config/start.rs
@@ -162,8 +162,9 @@ fn apply_near_config_patches(
     if near_init.chain_id.is_localnet() {
         config["state_sync_enabled"] = serde_json::Value::Bool(false);
     } else {
+        let storage_fallback_threshold = near_init.external_storage_fallback_threshold.unwrap_or(0);
         config["state_sync"]["sync"]["ExternalStorage"]["external_storage_fallback_threshold"] =
-            serde_json::json!(0);
+            serde_json::json!(storage_fallback_threshold);
     }
 
     // Track the shard that hosts the MPC contract.
@@ -175,6 +176,10 @@ fn apply_near_config_patches(
     }
     if let Some(network_addr) = &near_init.network_addr {
         config["network"]["addr"] = serde_json::Value::String(network_addr.clone());
+    }
+    if let Some(tier3) = &near_init.tier3_public_addr {
+        config["network"]["experimental"]["tier3_public_addr"] =
+            serde_json::Value::String(tier3.clone());
     }
 }
 
@@ -196,6 +201,8 @@ mod tests {
             download_genesis_records_url: None,
             rpc_addr: None,
             network_addr: None,
+            tier3_public_addr: None,
+            external_storage_fallback_threshold: None,
         }
     }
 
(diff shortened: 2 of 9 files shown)


barakeinav1:
@claude review

crates/node/src/config/start.rs
```diff
@@ -159,12 +178,171 @@ fn patch_near_config(
...
+            serde_json::Value::String(tier3.clone());
+    }
+}
 
-    let patched =
-        serde_json::to_string_pretty(&config).context("failed to re-serialize NEAR config.json")?;
-    std::fs::write(config_path, patched)
-        .with_context(|| format!("failed to write {}", config_path.display()))?;
+#[cfg(test)]
+#[allow(non_snake_case)] // tests follow `<system_under_test>__should_<assertion>` convention
```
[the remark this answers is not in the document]
barakeinav1: fixed

crates/node-config/src/start.rs
```diff
@@ -108,6 +108,19 @@ pub struct NearInitConfig {
     /// Override the NEAR node network (indexer) listen address (e.g. "0.0.0.0:24568").
     /// Useful when running multiple nodes on the same machine.
     pub network_addr: Option<String>,
+    /// Override the public address advertised for Tier3 state-sync responses
+    /// (e.g. "203.0.113.5:24567"). Required on multi-IP hosts where outbound
+    /// source IP differs from the bound IP — auto-discovery picks the
+    /// outbound IP and DSS times out. Patches into nearcore's
+    /// `network.experimental.tier3_public_addr` config field.
+    pub tier3_public_addr: Option<String>,
```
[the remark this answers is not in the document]
barakeinav1: fixed

barakeinav1:
Thanks for the reviews. Addressed in `25d7b037`:

- **`${VAR}` unguarded substitution** → switched to `${VAR:?…}` in `node.conf.testnet.toml.tpl`. A missing env var now fails the render with a clear pointer to `deploy-tee-cluster.sh`, instead of silently producing invalid TOML.
- **Default-threshold change for the rust-launcher** → tightened: `TIER3_PUBLIC_ADDR` and `EXTERNAL_STORAGE_FALLBACK_THRESHOLD` are now only exported when `MODE=testnet`. Localnet disables state sync entirely (`apply_near_config_patches` localnet branch sets `state_sync_enabled = false`), so the exports were dead code there anyway.
- **Patch-only-at-init operator footgun** → added a one-line note in `running-an-mpc-node-in-tdx-external-guide.md` explaining that `tier3_public_addr` and `external_storage_fallback_threshold` are applied at first init only; existing nodes need a hand-edit of `config.json` or a wipe.

Skipping (with rationale):

- **Default-`0`-when-`None` test missing.** Already covered on the base branch (#3189) as `should_set_external_storage_fallback_threshold_for_non_localnet`. Stacked-PR diffs only show the top of the stack, so it looks absent on this PR's view.
- **`tier3_public_addr` as `Option<String>` vs `SocketAddr`.** Pre-existing pattern — `rpc_addr` and `network_addr` are also raw strings. Validating one without the others would be inconsistent; out of scope for this PR.
- **Legacy `deployment/start.sh:46` still hardcodes `= 0`.** Intentional — that path is being superseded by `start-with-config-file`, which is the codepath this PR's fix flows through.
- **Earlier Copilot threads.** Both already addressed in current code: e2e-tests struct literal updated; `#[allow(non_snake_case)]` → `#[expect(non_snake_case)]` on the test module.

crates/node/src/config/start.rs
```diff
@@ -176,6 +177,10 @@ fn apply_near_config_patches(
     if let Some(network_addr) = &near_init.network_addr {
         config["network"]["addr"] = serde_json::Value::String(network_addr.clone());
     }
+    if let Some(tier3) = &near_init.tier3_public_addr {
+        config["network"]["experimental"]["tier3_public_addr"] =
+            serde_json::Value::String(tier3.clone());
```
anodar, line 182: nit: (I see that for other fields we also just pass without validation but) consider validating (maybe with simple regexp) the value before propagating to fail fast.

localnet/tee/scripts/rust-launcher/node.conf.testnet.toml.tpl
```diff
@@ -18,6 +18,8 @@ chain_id = "testnet"
 boot_nodes = "${NEAR_BOOT_NODES}"
 download_genesis = true
 download_config = "rpc"
+tier3_public_addr = "${TIER3_PUBLIC_ADDR:?TIER3_PUBLIC_ADDR must be set; deploy-tee-cluster.sh sets it from MODE=testnet}"
```
anodar: Claude claims "envsubst does not support ${VAR:?msg} or ${VAR:-default} syntax — only $VAR and ${VAR}."

Seems to be correct as `man envsubst` says:

```
In normal operation mode, standard input is copied to standard output, with references to environment variables of the form $VARIABLE or ${VARIABLE} being replaced with the corresponding values.
```

anodar (dismissed):
Please have a look at env variable substitution comment. Otherwise LGTM

barakeinav1:
Fix in `dc24203f` — see inline replies.

localnet/tee/scripts/rust-launcher/node.conf.testnet.toml.tpl
barakeinav1: Confirmed — envsubst leaves `${VAR:?msg}` literal in output (`echo 'foo=\"${VAR:?msg}\"' | VAR=hi envsubst` prints the `:?msg` part verbatim). Fixed in `dc24203f`: template reverted to plain `${VAR}`; fail-fast moved into `deploy-tee-cluster.sh` as `: "${VAR:?msg}"` before `envsubst`, where bash actually handles the syntax.

crates/node/src/config/start.rs
barakeinav1: Tracked as a follow-up in #3210, covering `rpc_addr`, `network_addr`, and `tier3_public_addr` together (validating only one here would be inconsistent with the existing two).

anodar (approved):

localnet/tee/scripts/rust-launcher/node.conf.testnet.toml.tpl
barakeinav1: fix in dc24203f

netrome (approved):
Thanks!

PR #3209, opened 2026-05-12
refactor: replace classify-blocks with atomic u8 in `RecentBlocksTracker`

### LINKED ISSUE #3135: `RecentBlocksTracker`: replace `classify_blocks` with atomic u8

kevindeforth:
resolves #3135


kevindeforth:
> * `crates/node/src/requests/recent_blocks_tracker.rs:101-105` / `recent_blocks_tracker.rs:312-316` — Carried over from the prior review and not addressed: `is_canonical()` returns `true` for `Final`, so `BlockNode::debug_print` now renders `Final` blocks as `C F` even when, in pathological scenarios, the block is final but not on the current canonical chain (previously rendered `   F`). The test comment in `test_complex_forks` was updated to match the new rendering, which masks rather than fixes the diagnostic regression. Consider letting `debug_print` read the raw stored discriminant so `C` and `F` remain independent display bits.

This is not worth addressing, as it is taken care of in the following point.

> * `crates/node/src/requests/recent_blocks_tracker.rs:113,124` — `make_canonical`/`make_non_canonical` log at `error!` level when invoked on a `Final` node. The crate-level comment in `test_complex_forks` ("we want reasonable behavior and no crashes") explicitly admits indexer-supplied pathological inputs, and the second loop of `update_canonical_head` can walk past a now-`Final` block on the _old_ canonical chain in exactly such a scenario. An `error!` here will page operators for an intentionally-tolerated case. Suggest `warn!` (or `debug!`), since the call is already a benign no-op.

This is not something we expect to see in production, so if it actually does happen, we want to know about it asap.

> * `crates/node/src/requests/queue.rs:487-491` — On `Weak::upgrade()` failure the log says _"indicating that it sat on a dead fork."_ and the metric label is `block_not_found`. The Arc also drops when a `Final` block ages out of the recency window (case 2 of `partition_subtree`), so this message will fire for a block that finalized but whose request lingered in the queue past pruning. Old `CheckBlockResult::Unknown` collapsed both cases under the same label without claiming the dead-fork interpretation; consider softening the wording (e.g., _"tracker no longer has the block (pruned as dead fork or aged out of recency window)"_).

No. We have a height check before we compraing the block status. If it it timed out, we will catch it before this log.

> * `crates/node/src/requests/queue.rs:502-505` — Cosmetic only: the `tracing::debug!(..., "processing request")` call has a stray blank line and inconsistent indentation between `target:` and `leader = %leader,`. Worth a `cargo fmt` pass before merge.

Believe it or not, I did run cargo fmt like a million times and it din't help with these macros. Had to fix most of them myself.
fixed in https://github.com/near/mpc/pull/3209/commits/e54dcad4855693fe8630ab7d1a0bf036a9aea69d

pbeza (dismissed):

crates/node/src/requests/queue.rs
```diff
@@ -482,102 +484,68 @@ impl<RequestType: Request + Clone, ChainRespondArgsType: ChainRespondArgs>
...
-                        target: "request",
-                        request_type = %RequestType::get_type(),
-                        request_id= %request.request.get_id(),
-                        block_height= request.block_height,
-                        "ignoring non-canonical request",
-                    );
+            let Some(status) = request.status.upgrade() else {
+                tracing::debug!(
+                    target: "request",
+                    "discarding request because the tracker removed the block, indicating that it sat on a dead fork."
```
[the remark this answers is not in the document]
kevindeforth: Glad to see we AI has not yet reached overlord status.
Your brother claude already (wrongly) complained about this: https://github.com/near/mpc/pull/3209#issuecomment-4431790946

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -4,13 +4,13 @@ use near_indexer_primitives::CryptoHash;
...
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicU8, Ordering};
 use std::sync::{Arc, Mutex, Weak};
 
 /// Tracks the topology of the recent blocks, using the blocks given by the indexer.
 ///
 /// This class provides two important functionalities:
-///  - For any given block, it classifies it into one of the categories in `CheckBlockResult`.
+///  - For any given block, it classifies it into one of the categories in `BlockStatus`.
 ///    See the documentation of that enum for the classifications.
```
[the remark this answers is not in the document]
kevindeforth: addressed in https://github.com/near/mpc/pull/3209/commits/60c36a9c4c794c84a28ce83fb69e228740df9176

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -216,7 +256,7 @@ impl BlockNode {
     fn prune_dead_children(&self, final_head_height: u64, blocks_to_prune: &mut Vec<CryptoHash>) {
         let mut children = self.children.lock().unwrap();
```
barakeinav1, line 257: nit (suggested by claude) : `partition_subtree` emits hashes into `new_roots: Vec<CryptoHash>`, then this block re-resolves them via `hash_to_node.get(hash)` with a defensive `or_else` + `tracing::error!` for a "should NEVER happen" case. If `partition_subtree` took `self: &Arc<BlockNode>` and emitted `Vec<Arc<BlockNode>>` directly, the re-lookup and the dead branch both disappear — at the cost of a cheap `Arc::clone` per kept root.

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -70,20 +70,65 @@ pub struct RecentBlocksTracker<T: Clone + 'static> {
...
+        status as u8
+    }
+}
+
+pub struct AtomicBlockStatus(AtomicU8);
+
+impl AtomicBlockStatus {
+    pub fn is_final(&self) -> bool {
+        self.0.load(Ordering::Relaxed) == u8::from(BlockStatus::Final)
+    }
```
barakeinav1, line 99: nit: no test asserts the "no downgrade from `Final`" invariant guarded here (and at L123 for `make_non_canonical`) — but enforcing this in code rather than convention was a primary motivation for the bool→u8 swap. A small unit test would lock the invariant against future refactors:

```rust
#[test]
#[expect(non_snake_case)]
fn atomic_block_status__should_not_downgrade_from_final() {
    // Given
    let s = AtomicBlockStatus(AtomicU8::new(BlockStatus::Final.into()));

    // When / Then
    s.make_canonical();
    assert!(s.is_final());
    s.make_non_canonical();
    assert!(s.is_final());
}
```

barakeinav1 (dismissed):
Not very familiar with this code, but reviewed it with Claude's help. Left a few nits, but the design looks solid and the test coverage is reasonable. Approving.

crates/node/src/requests/recent_blocks_tracker.rs
kevindeforth: yeah, this will be fixed very soon. @pbeza has this one out: https://github.com/near/mpc/pull/3196 and I am working on some modifs which will take care of that. But deferred to a follow-up PR.

crates/node/src/requests/recent_blocks_tracker.rs
kevindeforth: good point, will add.

kevindeforth: Maybe just adding a `debug_assert` will also be sufficient. 

edit:
> but enforcing this in code rather than convention was a primary motivation for the bool

Nit: the main motivation was to enable efficient querying of the block status, without requiring access to `RecentBlocksTracker`.

edit 2:
Just re-read the unit test and feel compelled to clarify something:
The test generated by claude doesn't make much sense, because we won't change the state of something that's already final.

The existing unit tests should capture this invariant, but if they don't then, then a `debug_assert` seems like a good idea to put.

edit 3: ok, of course, I can't add debug statements and unit test. Given that we are already logging errors in case we encounter that situation, I chose to add the unit test over the debug asserts: https://github.com/near/mpc/pull/3209/commits/b7117179d97062e1bdc594f927b1ce9f082a24ad

DSharifi (dismissed):

DSharifi (approved):

barakeinav1 (approved):

Issue #1691, opened 2025-12-18
test_parallel_sign_calls for robust_ecdsa


gilcu3:
### Background

The `robust_ecdsa` `test_parallel_sign_calls` test introduced in https://github.com/near/mpc/pull/1682 included 

```
# TODO: this values should be the same, but for some reason are not.
# This means that something strange is happening in the queue, not sure
# if it is expected when there are failures, although the metric for failures
# stayed at 0 for those cases
if led_requests < num_parallel_requests:
    metrics_good = False
if metrics_good:
    break
```
For the original `test_parallel_sign_calls` , we had `led_requests == num_parallel_requests`, but for some reason for robust-ecdsa led_requests usually ends up being higher, maybe provoked by some timeouts happening, as requests seem noticeably slower.

### Acceptance Criteria

We improve the situation and achieve `led_requests == num_parallel_requests` in that test

anodar:
There's now an umbrella issue for all flaky tests (for now only has flaky parallel requests as well) https://github.com/near/mpc/issues/2898. Added this issue as a context to that one.


@gilcu3 will close this as duplicate unless you object.

gilcu3:
please check out that TODO I reference in the code, if you did not need it in rust tests then no information is lost by closing this

anodar:
So far we have same in Rust test as well, so keeping this issue open: https://github.com/near/mpc/blob/136c7c2bf5906475c972bd0ec93b8876c341f864/crates/e2e-tests/tests/parallel_sign_calls.rs#L138

PR #3211, opened 2026-05-12
feat: metric for counting failed requests

### LINKED ISSUE #3217: Add metric for tracking request failures
### Background

We currently don't have a reliable metric for tracking request failures. We have a `mpc_cluster_failed_signatures_count` metric, but that one is not a good indicator for network health:

1. It only tracks signature requests, omitting CKD and foreign chain transaction requests.
3. It may have false positives as it does not distinguish between signatures that ended up being included on chain and signatures that were not included on chain.
4. It may have false positives in case our indexer is lagging behind peers:
     1.  the networks maximum_height exceeds the timeout height
     2. Our indexer, lagging, has not yet seen the response transaction on chain.
5. It may have false negatives, due to the flow outlined below:   
    1. We scan blocks for transactions that call `return_signature_and_clean_state_on_success` in `indxer/handle.rs` and append those to the list of `completed_signatures`: [`ingexer/handler.rs:229`](https://github.com/near/mpc/blob/218af33061f3a27886b20f9c460a56fa754cc8a4/crates/node/src/indexer/handler.rs#L229), which will get removed from the queue. **In case the request timed out, it will also get added to the list of completed requests.**
    2. This gets sent to the `PendingRequestQueue`, where we then remove it from the queue [`requests/queue.rs:350`](https://github.com/near/mpc/blob/218af33061f3a27886b20f9c460a56fa754cc8a4/crates/node/src/requests/queue.rs#L350).
    3. Only if the signature is till in the queue will the metric fire in [`requests/queue.rs:515`](https://github.com/near/mpc/blob/218af33061f3a27886b20f9c460a56fa754cc8a4/crates/node/src/requests/queue.rs#L515) and [`requests/queue.rs:477`](https://github.com/near/mpc/blob/218af33061f3a27886b20f9c460a56fa754cc8a4/crates/node/src/requests/queue.rs#L477). Note that in `monitor_block_updates`, we add the new block _before_ we call `get_requests_to_attempt()`: [`mpc_client.rs`](https://github.com/near/mpc/blob/218af33061f3a27886b20f9c460a56fa754cc8a4/crates/node/src/mpc_client.rs#L245)

So, we need a good metric for measuring signature timeouts. The proposal is to simply index any `fail_on_timeout` calls to our mpc contract. That won't be a perfect metric either, because we may have false positives due to counting timeouts on forks that end up never getting included, and it does not allow us to compare between the different request types, but, it will be a better indicator than what we currently have.

### User Story

As a node operator, I want reliable metrics.

### Acceptance Criteria

Add a metric, counting the number of calls to `FAIL_ON_TIMEOUT`

_No response_

kevindeforth:
resolves https://github.com/near/mpc/issues/3217

### DIFF
--- a/crates/e2e-tests/tests/timeout_metric.rs
+++ b/crates/e2e-tests/tests/timeout_metric.rs
@@ -1,8 +1,16 @@
 use crate::common;
 
 use e2e_tests::{CLUSTER_WAIT_TIMEOUT, metrics};
-use near_mpc_contract_interface::types::{Curve, DomainPurpose};
-use rand::SeedableRng;
+use near_mpc_contract_interface::types::{Curve, DomainConfig, DomainPurpose};
+use rand::{RngCore, SeedableRng};
+
+fn must_get_payload_for_domain(domain: &DomainConfig, rng: &mut impl RngCore) -> serde_json::Value {
+    match domain.curve {
+        Curve::Secp256k1 => common::generate_ecdsa_payload(rng),
+        Curve::Edwards25519 => common::generate_eddsa_payload(rng),
+        c => panic!("unsupported curve in test: {c:?}"),
+    }
+}
 
 /// When a sign request can't be answered (because too many participants are
 /// down), the contract calls `fail_on_timeout` and each alive node's indexer
@@ -27,33 +35,10 @@ async fn timeout_metric__should_increment_when_signature_times_out() {
         .find(|d| matches!(d.purpose, DomainPurpose::Sign))
         .expect("cluster must have a signable domain");
 
-    let payload = match domain.curve {
-        Curve::Secp256k1 => common::generate_ecdsa_payload(&mut rng),
-        Curve::Edwards25519 => common::generate_eddsa_payload(&mut rng),
-        c => panic!("unsupported curve in test: {c:?}"),
-    };
-
-    let outcome = cluster
-        .send_sign_request(domain.id, payload, cluster.default_user_account())
-        .await
-        .expect("expected success");
-
-    // sanity check
-    assert!(
-        outcome.is_success(),
-        "sign request for domain {:?} failed: {:?}",
-        domain.id,
-        outcome.failure_message()
-    );
-
     // when — kill node 0, then submit a request no one can answer
     cluster.kill_nodes(&[0]).expect("failed to kill node 0");
 
-    let payload = match domain.curve {
-        Curve::Secp256k1 => common::generate_ecdsa_payload(&mut rng),
-        Curve::Edwards25519 => common::generate_eddsa_payload(&mut rng),
-        c => panic!("unsupported curve in test: {c:?}"),
-    };
+    let payload = must_get_payload_for_domain(domain, &mut rng);
 
     // then — node 1's indexer must observe fail_on_timeout. We can't `.await`
     // the second sign request: the contract's yield + auto-timeout completes
@@ -67,9 +52,11 @@ async fn timeout_metric__should_increment_when_signature_times_out() {
             &cluster,
             &[1],
             metrics::TIMEOUTS_INDEXED,
-            |v| v == 1,
+            |v| v >= 1, // note: we would prefer to have strict equality, but we use inequality, in
+            // case the send below has retry mechanism. C.f.
+            // https://github.com/near/mpc/pull/3211#discussion_r3233189801
             CLUSTER_WAIT_TIMEOUT,
-        ) => res.expect("mpc_num_timeouts_indexed did not reach 1 on node 1"),
+        ) => res.expect(&format!("{} did not reach 1 on node 1", metrics::TIMEOUTS_INDEXED)),
         _ = cluster.send_sign_request(domain.id, payload, cluster.default_user_account()) =>
             panic!(
                 "sign request future returned before timeout metric — test wiring is wrong \
--- a/crates/e2e-tests/tests/timeout_metric.rs
+++ b/crates/e2e-tests/tests/timeout_metric.rs
@@ -55,22 +55,13 @@ async fn timeout_metric__should_increment_when_signature_times_out() {
         c => panic!("unsupported curve in test: {c:?}"),
     };
 
-    // Race the request against the metric: we don't care about the request's
-    // outcome (it will fail), only that the surviving node observes
-    // fail_on_timeout. `tokio::join!` lets the indexer keep advancing while
-    // the request is in flight.
     let outcome = cluster
         .send_sign_request(domain.id, payload, cluster.default_user_account())
         .await
         .expect("expected success");
 
     // sanity check
-    assert!(
-        outcome.is_success(),
-        "sign request for domain {:?} failed: {:?}",
-        domain.id,
-        outcome.failure_message()
-    );
+    assert!(outcome.is_failure(), "expected sign request to fail",);
 
     // Then: expect metric to kick in
     common::wait_metric_on_nodes(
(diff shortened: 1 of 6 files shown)


crates/e2e-tests/tests/timeout_metric.rs
```diff
@@ -0,0 +1,80 @@
...
+    // indexer bumps the counter the test passes; the sign-request future is
+    // cancelled (the tx is already on chain by then).
+    tokio::select! {
+        res = common::wait_metric_on_nodes(
+            &cluster,
+            &[1],
+            metrics::TIMEOUTS_INDEXED,
+            |v| v == 1,
+            CLUSTER_WAIT_TIMEOUT,
+        ) => res.expect("mpc_num_timeouts_indexed did not reach 1 on node 1"),
```
anodar: nit: use TIMEOUTS_INDEXED constant that's already defined in crates/e2e-tests/src/metrics.rs

crates/e2e-tests/tests/timeout_metric.rs
```diff
@@ -0,0 +1,80 @@
...
+        outcome.failure_message()
+    );
+
+    // when — kill node 0, then submit a request no one can answer
+    cluster.kill_nodes(&[0]).expect("failed to kill node 0");
+
+    let payload = match domain.curve {
+        Curve::Secp256k1 => common::generate_ecdsa_payload(&mut rng),
+        Curve::Edwards25519 => common::generate_eddsa_payload(&mut rng),
+        c => panic!("unsupported curve in test: {c:?}"),
```
anodar: nit: this is duplicate of above, consider refactoring into `must_get_payload_from_curve()` or something like that.

crates/node/src/metrics.rs
```diff
@@ -162,6 +162,14 @@ pub static MPC_NUM_VERIFY_FOREIGN_TX_RESPONSES_INDEXED: LazyLock<prometheus::Int
         .unwrap()
     });
 
+pub static MPC_NUM_TIMEOUTS_INDEXED: LazyLock<prometheus::IntCounter> = LazyLock::new(|| {
+    prometheus::register_int_counter!(
+        "mpc_num_timeouts_indexed",
```
anodar: nit: This sounds pretty generic, consider more verbose metric name maybe mpc_num_fail_on_timeout_indexed, as changing it later may not be trivial.

crates/e2e-tests/tests/timeout_metric.rs
```diff
@@ -0,0 +1,80 @@
...
+
+use e2e_tests::{CLUSTER_WAIT_TIMEOUT, metrics};
+use near_mpc_contract_interface::types::{Curve, DomainPurpose};
+use rand::SeedableRng;
+
+/// When a sign request can't be answered (because too many participants are
+/// down), the contract calls `fail_on_timeout` and each alive node's indexer
+/// bumps `mpc_num_timeouts_indexed`. This test stands up a 2-of-2 cluster,
+/// kills one node, submits a sign request, and verifies the surviving node
+/// observes the timeout.
```
netrome, line 19: Hmm, this is definitely fine for this PR but it feels strange to have to rely on killing a node to trigger a signing failure. It would be nice with the ability to define a more isolated test setup, for example by spinning up a modified MPC node designed to explicitly fail signing requests but otherwise operate normally. Just thinking loosely here, this is definitely beyond the scope of the current PR.

crates/e2e-tests/tests/timeout_metric.rs
```diff
@@ -0,0 +1,80 @@
...
+            |v| v == 1,
+            CLUSTER_WAIT_TIMEOUT,
+        ) => res.expect("mpc_num_timeouts_indexed did not reach 1 on node 1"),
+        _ = cluster.send_sign_request(domain.id, payload, cluster.default_user_account()) =>
+            panic!(
+                "sign request future returned before timeout metric — test wiring is wrong \
+                 (request unexpectedly succeeded or near_kit retries exhausted before \
+                 the on-chain timeout)"
+            ),
+    }
```
netrome, line 66: Hmm, I don't have any better suggestion but it feels like this could lead to flaky failures if there for any reason would be any internal delay in reporting the metric in the node. It feels unlikely but it's definitely not ideal.

crates/node/src/indexer/handler.rs
```diff
@@ -237,6 +237,9 @@ async fn handle_message(
                             completed_verify_foreign_txs.push(request_id);
                             metrics::MPC_NUM_VERIFY_FOREIGN_TX_RESPONSES_INDEXED.inc();
                         }
+                        FAIL_ON_TIMEOUT => {
+                            metrics::MPC_NUM_TIMEOUTS_INDEXED.inc();
+                        }
```
netrome, line 242: Would it be possible to correlate this with the request type (i.e. label this by signing/CKD or foreign tx validation)? Or are we lacking that information in this context?

crates/e2e-tests/tests/timeout_metric.rs
```diff
@@ -0,0 +1,80 @@
...
+    // future is doomed to return Err from near_kit retries before the on-chain
+    // outcome materializes. Race the call against the metric — once node 1's
+    // indexer bumps the counter the test passes; the sign-request future is
+    // cancelled (the tx is already on chain by then).
+    tokio::select! {
+        res = common::wait_metric_on_nodes(
+            &cluster,
+            &[1],
+            metrics::TIMEOUTS_INDEXED,
+            |v| v == 1,
```
anodar: Just heads up, in case this becomes flaky (v == 1 instead of v >= 1, that is if for whatever reason transaction is retried), we have an open issue that we haven't gotten to yet: https://github.com/near/mpc/issues/1691

netrome (dismissed):
Some thoughts and questions but nothing blocking. Thanks for fixing this, this will be a much better way of tracking failures than we currently have.

crates/node/src/indexer/handler.rs
kevindeforth: That's not possible here. We would need a more elaborate approach, tracking the transaction over multiple receipts.

crates/e2e-tests/tests/timeout_metric.rs
kevindeforth: Hmm, doesn't this apply to all tests relying on `wait_metric_on_nodes`? Or are you talking about the race between the send_sign_request and awaiting the metric?

crates/e2e-tests/tests/timeout_metric.rs
```diff
@@ -0,0 +1,80 @@
...
+        .await
+        .expect("expected success");
+
+    // sanity check
+    assert!(
+        outcome.is_success(),
+        "sign request for domain {:?} failed: {:?}",
+        domain.id,
+        outcome.failure_message()
+    );
```
anodar: nit: `must_setup_cluster` already waits for running state and presignatures. Not sure this adds much value to the test.

anodar (dismissed):

crates/e2e-tests/tests/timeout_metric.rs
kevindeforth: Ok, will refactor to >= 1 and add a comment.

crates/e2e-tests/tests/timeout_metric.rs
kevindeforth: This made sense to me, as I wanted to test that successful signatures don't contribute to the metric.
However, with https://github.com/near/mpc/pull/3211#discussion_r3233324885, there doesn't seem to be a good way to test for this now.

Thinking of it, it's also not that critical (to test for this). If we have successful requests contributing to this metric, we will soon realize this in production.
So I agree, we can omit this.

anodar (approved):
Thank you!

netrome (approved):

PR #3213, opened 2026-05-12
feat: Handle duplicate requests through unique request IDs

### LINKED ISSUE #3184: Duplicate requests should be processed and responded to

netrome:
Draft: Alternative to https://github.com/near/mpc/pull/3187

Closes #3184


netrome:
Not relevant anymore.

PR #3216, opened 2026-05-12
feat: implement foreign chain RPC providers in contract

### LINKED ISSUE #3214: Design and implement schema how contract stores RPC providers
### Background

Contract needs to store allowlisted RPC providers.

### User Story

_No response_

### Acceptance Criteria

Have structure in the contract for storing allowlisted RPC providers.

_No response_

anodar:
Note: this bumps up contract size hard limit from 1490000 -> 1495000
Closes: https://github.com/near/mpc/issues/3214

### Deviations from the original design

  The original design tracked **which chains** each participant supports, leaving the **provider URL** picked locally:

  ```rust
  pub struct ForeignChainSupportByNode {
      pub foreign_chain_support_by_node: BTreeMap<AccountId, SupportedForeignChains>,
  }
  pub struct SupportedForeignChains(pub BTreeSet<ForeignChain>);
```

  That covers chain enablement but doesn't constrain which RPC providers the network agrees to trust — a TEE-attested node binary still consumes its operator's config file unchecked. The new model layers a per-chain provider whitelist on top, with these specific deviations:

  1. The contract owns the connection config, not the operator. Instead of nodes picking arbitrary URLs and the contract only knowing chain enablement, the contract stores (base_url, auth_scheme, chain_routing) per voted-in provider. Operators reference providers by provider_id only and supply just the token. Closes the "operator points the binary at a malicious RPC" gap.
  2. Chain identity is structured, not stringly typed. ChainRouting is an enum (Embedded | PathSegment | QueryParam) rather than a free-form URL string. This makes "exactly one
  encoding per entry" a type-system property and removes the syntactic surface an operator could use to inject extra path/query components.
  3. Whitelist is keyed (chain, provider_id), not just chain. The same provider_id (e.g. "ankr") appearing under multiple chains is expected — each entry carries chain-specific
  base_url / chain_routing. Per-chain duplication is the natural cross-chain marker, not a redundancy to dedupe.
  4. Multi-vote per participant per target. Diverges from the TEE single-vote-per-category pattern (vote_code_hash, vote_add_launcher_hash). Pending votes are partitioned by target
  (chain, provider_id), so a participant can simultaneously vote Add(Ethereum, alchemy) and Add(Ethereum, ankr) without losing one. PR 2 will land the matching vote endpoints.
  5. Per-chain voting threshold, not the global protocol threshold. chain_thresholds: BTreeMap<ForeignChain, u64> ships with PR 1 (empty); lookups fall back to
  DEFAULT_PROVIDER_VOTE_THRESHOLD = 2 until a future vote_set_foreign_chain_provider_threshold endpoint lets the network override it per chain.
  6. Testnets get separate ForeignChain enum variants (when added) rather than an environment: Mainnet | Testnet field. Same information through the type system at no extra cost. Adding the variants themselves is out of scope here.
  7. Validation at runtime, not just lookup. PR 3 adds a chain-identity probe at node startup that calls the chain's self-identifying RPC (eth_chainId, getGenesisHash,
  starknet_chainId, a checkpointed Bitcoin block hash) and asserts the response matches a constant hardcoded in the inspector. Catches the case where a vote slipped a testnet URL into  a mainnet slot, even if URL shape and routing/auth collisions passed contract-side validation.

  The original register_foreign_chain_support API and the per-node intersection model are kept — the whitelist is a new layer on top, not a replacement. A chain is still only  "supported" once every active participant has registered it after applying the whitelist filter.

### CODE REVIEW

File: crates/contract/src/foreign_chain_rpc.rs [hunk 1]
```diff
@@ -0,0 +1,285 @@
+//! On-chain whitelist of RPC providers for foreign-chain transaction validation.
+//!
+//! The MPC network agrees, by vote, on which RPC providers each node may use to verify
```
DSharifi: Feel like it sounds like we are whitelisting RPCs per node

```suggestion
//! The MPC network agrees, by vote, on which RPC providers any node may use to verify
```

crates/contract/src/foreign_chain_rpc.rs
```diff
@@ -0,0 +1,285 @@
...
+/// Default vote threshold used by every chain until `vote_set_foreign_chain_provider_threshold`
+/// (future PR) lets the network override it per chain. Two votes is the smallest threshold
+/// that still requires more than one party to agree, which is the meaningful security floor.
+pub const DEFAULT_PROVIDER_VOTE_THRESHOLD: u64 = 2;
+
+/// Per-chain set of voted-in providers. Within a chain `provider_id` is unique; uniqueness
+/// is enforced on [`AllowedProviders::add`]. The same `provider_id` (e.g. `"ankr"`) showing
+/// up under different chains is expected — each entry carries chain-specific connection
+/// details (`base_url`, `chain_routing`), so two `"ankr"` entries are per-chain configs,
+/// not duplicates.
```
DSharifi: Can we instead just change the hash/eq/partialeq implementations of the ProviderEntry to use the `provider_id` field so we don't have to maintain this invariant at runtime?

Then the Vec can be changed to BTreeSet

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
```diff
@@ -952,6 +952,141 @@ impl ForeignTxSignPayload {
...
+pub enum ProviderVoteAction {
+    Add {
+        chain: ForeignChain,
+        entry: ProviderEntry,
+    },
+    Remove {
+        chain: ForeignChain,
+        provider_id: ProviderId,
+    },
+}
```
DSharifi: Voting on diffs (Add/Remove) is nice, but I'm a bit worried about the operational complexity it will entail. Let's say we want to support 5 chains and 5 RPCs for these chains at launch, i.e. 5x5 = 25 actions.

It will require 13 node operators to all vote in sync 25 times. This will be very painful and difficult to orchestrate/manage in practice.



I do like the concept of voting on diffs though. Could we instead vote on a batch of `ProviderVoteAction`? Meaning the contract accepts voting on an ordered sequence of multiple `ProviderVoteAction` entries, and once the vote is accepted the contract applies the actions in that order.

docs/foreign-chain-transactions.md
DSharifi, line 1: I think this file will also need updating:

https://github.com/near/mpc/blob/ea15bfc62e01048ae09dd76d39f86acd57d012ad/docs/design/allowing-per-node-foreign-chain-rpc-configuration.md

anodar:
@claude review

docs/foreign-chain-transactions.md
anodar: Done.

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
anodar: Got rid of it in this PR. Will batch in the PR that will implement voting.

crates/contract/src/foreign_chain_rpc.rs
anodar: Changed to`entries: BTreeMap<ForeignChain, BTreeMap<ProviderId, ProviderEntry>>`, that way we don't need custom eq.

crates/contract/src/foreign_chain_rpc.rs
```diff
@@ -0,0 +1,213 @@
...
+/// invariant). The same `provider_id` (e.g. `"ankr"`) showing up under different chains
+/// is expected — each entry carries chain-specific connection details (`base_url`,
+/// `chain_routing`), so two `"ankr"` entries in different chains are per-chain configs,
+/// not duplicates.
+//
+// Borsh-only: `AllowedProviders` is `pub(crate)` and only ever lives in contract state.
+// There is no view function returning it in PR 1 — that ships in the follow-up PR
+// alongside the vote endpoint that actually populates the whitelist. Keeping the type
+// borsh-only here trims the contract WASM size by skipping the serde JSON-serializer
+// monomorphizations that would otherwise be emitted for a perpetually-empty view.
```
DSharifi: subjective nit:
I feel like these docs add more noise than information. The code is quite simple and self explanatory. Seems like we have many more lines of docs than code now.

crates/contract/src/v3_9_1_state.rs
```diff
@@ -834,6 +837,12 @@ mod tests {
         }
     }
 
+    // The `MpcContract` migration assigns `foreign_chain_rpc_whitelist: Default::default()`
+    // in its `From` impl; if anyone removes that line the contract fails to compile
+    // ("missing field" in the struct literal), and the borsh schema snapshot test
+    // (`mpc_contract_borsh_schema_has_not_changed`) catches any layout drift, so a
+    // dedicated migration test for that single line would be redundant.
+
```
DSharifi:
```suggestion
```

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
```diff
@@ -952,6 +952,85 @@ impl ForeignTxSignPayload {
...
+/// auth shape that lets them inject extra path or query components.
+// `Hash` / `Ord` / `PartialOrd` are intentionally omitted — PR 1 doesn't need them
+// (the storage is `BTreeMap<ProviderId, ProviderEntry>` so only the key needs `Ord`).
+// PR 2 can re-add if the batched vote-action shape requires them.
+#[derive(Debug, Clone, Eq, PartialEq, BorshSerialize, BorshDeserialize)]
+// `Serialize` and `Deserialize` are only needed by node-side / external SDK consumers;
+// the contract itself never parses these from JSON args nor returns them from a view fn
+// in PR 1 (the view fn lands in the follow-up PR alongside the vote endpoint). Gating
+// the serde derives off wasm avoids several KB of monomorphizations per type in the
+// contract WASM.
```
DSharifi: Please remove these comments
```suggestion
#[derive(Debug, Clone, Eq, PartialEq, BorshSerialize, BorshDeserialize)]
```

DSharifi (changes requested):
Looks good overall, but there are many LLM comments I'd like to remove which are adding a noise to the code, but also not ideally placed.

For example a comment like this in the interface referring to node and contract implementation;
```rust
// `Serialize` and `Deserialize` are only needed by node-side / external SDK consumers;
// the contract itself never parses these from JSON args nor returns them from a view fn
// in PR 1 (the view fn lands in the follow-up PR alongside the vote endpoint). Gating
// the serde derives off wasm avoids several KB of monomorphizations per type in the
// contract WASM.Collapse comment
```

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
anodar: Done.

crates/contract/src/v3_9_1_state.rs
anodar: Done.

crates/contract/src/foreign_chain_rpc.rs
anodar: Done.

DSharifi (dismissed):

barakeinav1:
> ### Deviations from the original design

thanks for this, was really helpful when reviewing 

Would also be great to have [`docs/design/allowing-per-node-foreign-chain-rpc-configuration.md`](https://github.com/near/mpc/blob/main/docs/design/allowing-per-node-foreign-chain-rpc-configuration.md) updated to reflect this PR's changes — can be a follow-up PR, though preferably design-doc updates should land *before* implementation PRs rather than after.

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
```diff
@@ -952,6 +952,75 @@ impl ForeignTxSignPayload {
     }
 }
 
+/// Stable label for an RPC provider entry (e.g. `"alchemy"`, `"ankr"`, `"drpc"`).
+/// Unique within a chain in the on-chain foreign-chain RPC whitelist.
+pub type ProviderId = String;
```
barakeinav1: nit: with a type alias, `ProviderId` is interchangeable with any `String` at call sites — passing a `base_url` (also `String`) where a `provider_id` is expected would compile. A newtype `pub struct ProviderId(String)` is cheap insurance and matches the typed-boundary pattern used elsewhere in this file.

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
```diff
@@ -952,6 +952,75 @@ impl ForeignTxSignPayload {
...
+/// Where the operator's API key/token gets injected into the assembled RPC URL.
+/// Lives on the contract (not in operator yaml) so the operator can't pick a custom
+/// auth shape that lets them inject extra path or query components.
+#[derive(Debug, Clone, Eq, PartialEq, BorshSerialize, BorshDeserialize)]
+#[cfg_attr(not(target_arch = "wasm32"), derive(Serialize, Deserialize))]
+#[cfg_attr(
+    all(feature = "abi", not(target_arch = "wasm32")),
+    derive(schemars::JsonSchema, borsh::BorshSchema)
+)]
+pub enum AuthScheme {
```
barakeinav1, line 975: nit: consider `#[non_exhaustive]` on `AuthScheme` (and `ChainRouting` below) — the neighboring `ForeignChain` enum already uses it, and future variants are plausible (new auth shapes, new routing styles). Forces downstream matches to keep a default arm so adding a variant later isn't a breaking change for consumers.

barakeinav1 (dismissed):
Approving — clean, well-scoped data shape. Two small nits inline (newtype `ProviderId`, `#[non_exhaustive]` on the new enums), plus a note above on keeping the design doc in sync with the PR series.

anodar:
Thanks for the review! 

> Would also be great to have [docs/design/allowing-per-node-foreign-chain-rpc-configuration.md](https://github.com/near/mpc/blob/main/docs/design/allowing-per-node-foreign-chain-rpc-configuration.md) updated to reflect this PR's changes — can be a follow-up PR, though preferably design-doc updates should land before implementation PRs rather than after.

That design doc file is already updated in this PR, let me know if you want me to elaborate more on something specifically there.

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
anodar: Done.

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
anodar: Done.

anodar:
Reworded changes in design doc a bit.

barakeinav1 (dismissed):
Re-approving — nits addressed, doc rollout updated.

barakeinav1 (approved):

DSharifi (approved):

PR #3225, opened 2026-05-13
feat: fan out requests to all configured RPC providers

### LINKED ISSUE #3197: Foreign chain inspector crate should querying multiple inspectors

DSharifi:
closes #3197


barakeinav1:
This makes one malicious or unavailable RPC enough to DoS a node out of signing. That's a design choice, not an implementation detail, and should be reflected in the design doc — which currently says the opposite:

[`docs/design/allowing-per-node-foreign-chain-rpc-configuration.md#L23-L31`](https://github.com/near/mpc/blob/main/docs/design/allowing-per-node-foreign-chain-rpc-configuration.md#L23-L31) — Requirement 3 says "a *quorum* of the RPC providers must all agree" and lists single-RPC unavailability as the problem to solve, not introduce.

barakeinav1 (changes requested):
This makes one malicious or unavailable RPC enough to DoS a node out of signing. That's a design choice, not an implementation detail, and should be reflected in the design doc.

I'd like to see a design doc update as well to reflect this.

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -27,7 +28,68 @@ pub trait ForeignChainInspector {
...
+            .collect::<Result<_, _>>()?;
+
+        let mut iter = responses.into_iter();
+        let first = iter
+            .next()
+            .expect("inspectors is a `NonEmptyVec`, so there is at least one response");
+
+        for other in iter {
+            if other != first {
+                return Err(ForeignChainInspectionError::InspectorResponseMismatch);
```
barakeinav1, line 153: When `FanOut::extract` returns `InspectorResponseMismatch`, there's no way to tell which provider disagreed — the error has no fields and nothing is logged. A `tracing::warn!` with the per-provider responses before the early-return would fix it; those logs are operator-only, so including provider identity there is fine.

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -27,7 +28,68 @@ pub trait ForeignChainInspector {
...
+
+            let inspector = inspector.clone();
+            join_set.spawn(async move { inspector.extract(tx_id, finality, extractors).await });
+        }
+
+        let responses: Vec<Vec<Self::ExtractedValue>> = join_set
+            .join_all()
+            .await
+            .into_iter()
+            .collect::<Result<_, _>>()?;
```
barakeinav1: Worth considering: ignore unresponsive/erroring RPCs instead of failing the node. Right now any inner `Err` short-circuits the whole fan-out; tolerating failures and running the agreement check only over successful responses would let a single down provider degrade the node without taking it out of signing.

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
```diff
@@ -145,7 +165,16 @@ async fn setup_foreign_tx_cluster() -> anyhow::Result<ForeignTxTestEnv> {
...
     setup_evm_mock(&arbitrum_server);
     setup_evm_mock(&hyper_evm_server);
-    setup_evm_mock(&polygon_server);
+
+    // Polygon is configured with three RPC providers so the test can assert
+    // that `FanOut` queries every one of them.
+    let polygon_mocks: Vec<MockServerExt> = (0..3)
+        .map(|_| {
+            let server = MockServer::start();
+            let mock_id = setup_evm_mock(&server);
```
barakeinav1, line 174: The 3 Polygon mocks all return identical data, so the e2e only exercises the happy path. Worth adding an e2e where one mock returns different data and asserting the signing request fails — that locks in the disagreement-detection guarantee end-to-end, which is the security point of the PR.

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
```diff
@@ -415,6 +444,22 @@ async fn verify_hyper_evm(env: &ForeignTxTestEnv) -> anyhow::Result<()> {
...
+/// at least one HTTP request during the preceding `verify_polygon` call.
+///
+/// A regression in `FanOut` (e.g. routing each verify request to a single
+/// provider instead of fanning out to all of them) would leave at least one
+/// mock untouched and this assertion would fail.
+fn assert_fan_out_queried_every_polygon_provider(env: &ForeignTxTestEnv) {
+    for (i, polygon) in env.polygon_mocks.iter().enumerate() {
+        let calls = polygon.calls();
+        assert!(
+            calls > 0,
```
kevindeforth, line 456: nit: Would be nice to pass the minimum number of expected calls as an argument to this method, just in case we do more complicated unit tests where would expect a higher count.

Q: in the below unit test, do we expect 2 calls? (one for each node in the cluster)

kevindeforth (dismissed):
Thank you!

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
DSharifi: It's a very good point, but with the current MVP we don't have a way to cancel these requests other waiting for the 300 block timeout.

I added two tracking issues which are needed before we can have such a test.

https://github.com/near/mpc/issues/3261
https://github.com/near/mpc/issues/3262

crates/foreign-chain-inspector/src/lib.rs
DSharifi: As mentioned on slack, currently we only use 1 RPC provider anyway, and the feature is actively developed. This will be addressed in follow ups.

crates/foreign-chain-inspector/src/lib.rs
DSharifi: Good point, will fix.

DSharifi: Added logs now, and also more test coverage

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
DSharifi: It might be flaky if we set an exact amount for expected calls. For example the nodes will/might call these endpoints periodically as health checks, which will also increment the count.

barakeinav1 (approved):
Re-reviewed — the two inline asks are addressed: mismatch logging ✓ and transient tolerance via `is_transient()` ✓. Approving.

One follow-up before merge, please: update the design doc. `docs/design/allowing-per-node-foreign-chain-rpc-configuration.md:110-112` (from #3249) still describes a **quorum** model where a node verifies iff `ChainEntry.threshold` providers agree, but `FanOut` implements **unanimity among responders** and never reads a threshold. Please reconcile those lines with the current unanimity + transient-tolerance behavior and mark the quorum model as future work.

kevindeforth (approved):

PR #3228, opened 2026-05-14
feat(node-config): validate NearInitConfig address fields as SocketAddr

### LINKED ISSUE #3210: Validate address fields in NearInitConfig at config-load (rpc_addr, network_addr, tier3_public_addr)
### Background

Three address fields in `NearInitConfig` (in `crates/node-config/src/start.rs`) are `Option<String>` and passed through to neard's `config.json` without checking:

- `rpc_addr`
- `network_addr`
- `tier3_public_addr` (added in #3188)

A typo (missing port, hostname instead of IP) only surfaces later when neard fails to start, with a less helpful error.

### Acceptance Criteria

- [ ] Validate each of the three fields as a `SocketAddr` at config-load.
- [ ] On a malformed value, fail with an error that names the field.
- [ ] Existing valid configs still load.

### Notes

Raised as non-blocking nits on [#3188](https://github.com/near/mpc/pull/3188); deferred here so all three fields are validated together.

barakeinav1:
- `NearInitConfig.{rpc_addr, network_addr, tier3_public_addr}` change from `Option<String>` to `Option<SocketAddr>`. `std::net::SocketAddr` validates at deserialize, so a typo fails at TOML load with field name + bad value instead of when neard later fails to bind. TOML wire format unchanged; output to neard's `config.json` byte-identical for IPv4-with-port (canonical for any IPv6).
- 6 unit tests covering each field's happy path, missing port, non-numeric port, and the all-absent default.
- **Backward-compat note:** hostname-style values like `"localhost:3030"` now hard-fail at config load (they previously passed through as `String` and failed later at neard bind). No in-repo config uses this form; operators with hand-edited configs should ensure `IP:port` shape.

Closes #3210.

### DIFF
--- a/crates/e2e-tests/src/mpc_node.rs
+++ b/crates/e2e-tests/src/mpc_node.rs
@@ -1,3 +1,4 @@
+use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::process::{Child, Command, Stdio};
 
@@ -401,8 +402,8 @@ impl MpcNodeSetup {
                 download_genesis: false,
                 download_genesis_url: None,
                 download_genesis_records_url: None,
-                rpc_addr: Some(format!("0.0.0.0:{}", self.ports.near_rpc)),
-                network_addr: Some(format!("0.0.0.0:{}", self.ports.near_network)),
+                rpc_addr: Some(SocketAddr::from(([0, 0, 0, 0], self.ports.near_rpc))),
+                network_addr: Some(SocketAddr::from(([0, 0, 0, 0], self.ports.near_network))),
                 tier3_public_addr: None,
                 external_storage_fallback_threshold: None,
             }),
--- a/crates/node-config/src/start.rs
+++ b/crates/node-config/src/start.rs
@@ -4,6 +4,7 @@ use clap::ValueEnum;
 use launcher_interface::types::PccsEndpointConfig;
 use near_mpc_bounded_collections::NonEmptyVec;
 use serde::{Deserialize, Serialize};
+use std::net::SocketAddr;
 use std::path::PathBuf;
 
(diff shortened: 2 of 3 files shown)


barakeinav1:
@claude review

barakeinav1:
> * `crates/node-config/src/start.rs:107-110` — doc comments still describe IPv4-only examples (`"0.0.0.0:3031"`, `"0.0.0.0:24568"`). Since `SocketAddr` accepts bracketed IPv6 (`"[::1]:3031"`), consider mentioning that form so operators on dual-stack hosts know it's supported. Not a behavioral issue.

> * `crates/node-config/src/start.rs:328-414` — tests do not cover an IPv6 input (e.g., `"[::1]:24567"`). Optional, but cheap to add and would lock in the IPv6 contract you implicitly took on by switching to `SocketAddr`.

observations are valid but skipping: rejection coverage is type-equivalent across the three fields, and adding IPv6 examples/tests would imply support that no other layer in the stack is tested for.

> * Behavioral note (not a defect): any pre-existing config that set one of these fields to a hostname-style value (e.g., `"localhost:3030"`) will now hard-fail at config load instead of further downstream. This appears intentional and matches the PR description, but worth flagging in release notes for operators upgrading.

Reviewed. Backward-compat note about hostname-style values added to the description. The symmetry / IPv6

netrome (approved):
Nice. I assume this is compatible with existing configurations, since I don't see any config fixture changed in this PR.

gilcu3 (approved):
LGTM

barakeinav1:
> Nice. I assume this is compatible with existing configurations, since I don't see any config fixture changed in this PR.

yes, this is just a bit of input validation hardening

PR #3232, opened 2026-05-14
refactor: suggested changes for #3187

kevindeforth:
Suggesting changes for #3187


crates/contract/src/pending_requests.rs
```diff
@@ -177,35 +160,38 @@ pub(crate) fn resolve_yields_for<K>(
 where
     K: BorshSerialize + BorshDeserialize + Clone + Ord,
 {
-    if let Some(yield_indices) = new_map.remove(request) {
-        for YieldIndex { data_id } in yield_indices {
+    let resumed = new_map
+        .remove(request)
+        .unwrap_or_default()
+        .into_iter()
+        .chain(legacy_map.remove(request))
```
barakeinav1, line 167: The deleted migration test was the only coverage for "key K in both maps at once." That scenario still happens here (pre-upgrade caller + post-upgrade duplicate), but the new drain-both behavior — and the legacy-first ordering in `pop_one_yield_for` below — have no replacement test. Worth adding: insert a legacy entry, push a new-map entry for the same key, assert `respond*` drains both and timeout pops the legacy one first.

crates/contract/src/pending_requests.rs
```diff
@@ -92,82 +93,64 @@ impl LegacyPendingRequests {
...
-        }
-        queue.push(new_yield);
-    } else {
-        let queue = match legacy_map.remove(&request) {
-            Some(legacy_yield) => vec![legacy_yield, new_yield],
-            None => vec![new_yield],
-        };
-        new_map.insert(request, queue);
+    let queue = new_map.entry(request).or_default();
+    if queue.len() >= usize::from(MAX_PENDING_REQUEST_FAN_OUT) {
```
barakeinav1, line 111: This now only counts the new-map queue, so a key with a stale legacy entry can fan out to `MAX + 1` total during the legacy window (the old migration path counted the legacy yield against the cap). Probably intentional given the bounded legacy window — worth confirming.

barakeinav1 (approved):
Approving — clean simplification. Two non-blocking notes inline (test gap on the dual-map drain, and a `MAX_PENDING_REQUEST_FAN_OUT` nuance).

crates/contract/src/pending_requests.rs
kevindeforth: added in https://github.com/near/mpc/pull/3232/commits/9569f60ea1bc8f186b34c88f9779ad02e3a7a566

crates/contract/src/pending_requests.rs
netrome: Good observation. Should be fine though. I'd assume the test case we added with 128 duplicate requests should cover this also.

netrome (approved):
Thank you for the updates! 🙏 let's merge this into the base PR.

PR #3235, opened 2026-05-14
feat(tee-verifier-interface): Borsh DTOs for the verifier contract boundary

### LINKED ISSUE #3265: Introduce `tee-verifier-interface` crate (Borsh wire DTOs)
### Background

Prerequisite for the standalone verifier contract in #3266 (part of the #3264 effort to move Intel TDX quote verification out of the MPC contract). Once verification lives in a separate contract, the MPC contract reaches it by sending a quote and getting back a result — and both sides must agree on the shape of that data without pulling in each other's heavy dependencies.

This issue defines exactly those shared types (DTOs): a tiny, dependency-light crate holding only the data passed to and from the verifier. It has no logic of its own — it just gives #3266 and its caller a common vocabulary.

### User Story

As an MPC contract maintainer, I want the data exchanged with the verifier contract to be defined in one small, self-contained crate, so that the two contracts stay in sync and the verifier's heavyweight cryptographic dependencies never leak into the MPC contract.

### Acceptance Criteria

- [ ] New crate `tee-verifier-interface` under `crates/`.
- [ ] Defines the request and response types exchanged with the verifier contract.
- [ ] Carries none of the cryptographic verification dependencies — it is intentionally minimal so any contract can depend on it cheaply.
- [ ] Tests confirm each type survives a serialize/deserialize round-trip unchanged (the guarantee both sides rely on).

PR: #3235. Prerequisite for #3266; first step in the stack tracked by #3264.

pbeza:
Closes #3265

## Overview

Adds `tee-verifier-interface`: the Borsh DTOs spoken at the `tee-verifier` contract boundary.

They are field-for-field mirrors of the `dcap-qvl` input and output types, owned here so the Borsh wire layout is independent of upstream.

This crate is the *only* DTO crate a consumer (`mpc-contract`, future Proximity / Defuse contracts) needs in order to talk to the verifier — without re-linking the `dcap-qvl` / `ring` / `webpki` / `x509-cert` closure into its own WASM.

Prerequisite for the verifier contract in #3266 (PR #3237); part of the stack tracked by #3264.


pbeza:
@claude review

crates/tee-verifier-interface/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+        let original = status;
+
+        // When
+        let bytes = borsh::to_vec(&original).expect("Borsh serialization should succeed");
+        let decoded: TcbStatus =
+            borsh::from_slice(&bytes).expect("Borsh deserialization should succeed");
+
+        // Then
+        assert_eq!(original, decoded);
+    }
```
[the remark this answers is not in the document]
pbeza: Addressed in #3237.

gilcu3 (dismissed):
LGTM

Would be good to mention why we use borsh only (where we were previously using also serde). Also as many of these structs will then be removed from the contract-interface crate, might be good to point that out in some issue or comment.

A more general question, if we plan to keep this in sync with the dcap types, would it make sense to add some tests to enforce that?

crates/tee-verifier-interface/src/lib.rs
```diff
@@ -0,0 +1,364 @@
+//! Borsh DTOs spoken at the `tee-verifier` contract boundary.
+//!
+//! Field-for-field mirrors of the `dcap_qvl` input and output types,
+//! owned here so the Borsh wire layout is independent of upstream.
+//!
+//! This crate is the *only* DTO crate a consumer (`mpc-contract`, future
```
barakeinav1, line 6: The design doc (`docs/design/attestation-verifier-contract.md`, §Crate layout and §API Proposal) lists `VerifierError` among the wire DTOs owned by this crate, but #3237 defines it in `tee-verifier` as a `FunctionError` — meaning callers only ever see `PromiseError::Failed`, not a decodable error. Is that intentional? If so, the design doc's callback signature (`Result<Result<VerifiedReport, VerifierError>, PromiseError>`) and DTO list should be updated; if not, the type belongs here so #3247 can distinguish "quote rejected" from "verifier unavailable" for the refund path.

barakeinav1 (dismissed):

barakeinav1:
LGTM, I agree with [Reynaldo's comments](https://github.com/near/mpc/pull/3235#pullrequestreview-4425935590).

I added one [doc-vs-implementation drift comment](https://github.com/near/mpc/pull/3235#discussion_r3369023591).

crates/tee-verifier-interface/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+
+/// Quote collateral, mirroring `dcap_qvl::QuoteCollateralV3`.
+///
+/// Field-for-field copy so the wire layout matches the upstream Borsh
+/// encoding of `QuoteCollateralV3`.
+#[derive(Debug, Clone, BorshSerialize, BorshDeserialize, PartialEq, Eq)]
+#[cfg_attr(feature = "borsh-schema", derive(borsh::BorshSchema))]
+pub struct Collateral {
+    pub pck_crl_issuer_chain: String,
+    pub root_ca_crl: Vec<u8>,
```
netrome, line 56: Shouldn't we use HexVec here like in the `mpc-contract-interface`?

```
#[derive(
    Debug,
    Clone,
    Eq,
    PartialEq,
    Ord,
    PartialOrd,
    Hash,
    Serialize,
    Deserialize,
    BorshSerialize,
    BorshDeserialize,
)]
#[cfg_attr(
    derive(schemars::JsonSchema)
)]
pub struct Collateral {
    pub pck_crl_issuer_chain: String,
    pub root_ca_crl: HexVec,
    pub pck_crl: HexVec,
    pub tcb_info_issuer_chain: String,
    pub tcb_info: String,
    pub tcb_info_signature: HexVec,
    pub qe_identity_issuer_chain: String,
    pub qe_identity: String,
    pub qe_identity_signature: HexVec,
    pub pck_certificate_chain: Option<String>,
}
```

Also I wonder if we want to reuse the types between the DTO crates. We could introduce a shared crate with TEE DTOs that is re-exported in both this and the MPC contract crate. Alternatively have the MPC contract re-export the types from this crate directly.

This requires some thinking about potential architecture risks, but similar to how we sometimes reuse contract DTOs in the contract state (although that's technically a separate DTO boundary), I think it might be fine to reuse these DTOs in the contract interface.

netrome (dismissed):
Left a few comments, not blocking but I think it would be nice to use vectors that serialize to hex encoding on the wire.

crates/tee-verifier-interface/src/lib.rs
pbeza: Yeah, good catch, the doc got ahead of the implementation here. `verify_quote` returns its error via `#[handle_result]` (so the caller just sees `PromiseError::Failed`), which is how we do errors everywhere else in `mpc-contract`. None of our cross-contract callbacks decode a typed error off the wire, they all key off `PromiseError`. So I'd rather not make this one special.

Updated the design doc in https://github.com/near/mpc/commit/b9dfed6674dcf07fae370c8922321d8c7da5e611 to match.

pbeza:
> Would be good to mention why we use borsh only (where we were previously using also serde). (...)

Good point! Added a note about that here: https://github.com/near/mpc/pull/3235/commits/bb226c37a05b75b13f84f09fa962e49d29934c65.

> (...) Also as many of these structs will then be removed from the contract-interface crate, might be good to point that out in some issue or comment.

Created an issue: https://github.com/near/mpc/issues/3494 and referenced it in the comment here: https://github.com/near/mpc/pull/3235/commits/314a535e542ec903c4617c90b4de6dc80da74407.

> A more general question, if we plan to keep this in sync with the dcap types, would it make sense to add some tests to enforce that?

Great catch! #3237 (the PR stacked on top of this one) does exactly that: it adds the `From<dcap_qvl::…>` conversions plus tests asserting `borsh::to_vec(interface) == borsh::to_vec(dcap)` per type, so an upstream field reorder/rename fails CI. Has to live there rather than here, since this crate stays `dcap-qvl`-free and only the verifier crate links both. See:

https://github.com/near/mpc/blob/cf72a5ab610bf540343e5f0a88a9d01a101ded7e/crates/tee-verifier/src/conversions.rs#L148-L339

crates/tee-verifier-interface/src/lib.rs
pbeza: > Shouldn't we use HexVec here like in the mpc-contract-interface?

Went Borsh-only on purpose here (same reason I gave gilcu3 in https://github.com/near/mpc/pull/3235#pullrequestreview-4425935590 +https://github.com/near/mpc/commit/bb226c37a05b75b13f84f09fa962e49d29934c65): `HexVec` only changes the JSON wire, and there's no JSON wire here, so plain `Vec<u8>` keeps these as clean Borsh mirrors of `dcap_qvl`.

If you think JSON serialization would still be useful (e.g. for calling the verifier manually), happy to add a follow-up. Just keep in mind the input of the verifier is mostly binary, so I think Borsh is the more natural fit.

> Also I wonder if we want to reuse the types between the DTO crates. We could introduce a shared crate with TEE DTOs that is re-exported in both this and the MPC contract crate. Alternatively have the MPC contract re-export the types from this crate directly.

Once `mpc-contract` switches to these mirrors, the duplicates in `near-mpc-contract-interface` go away and this crate becomes the single owner (see #3494).

gilcu3 (dismissed):

pbeza:
@barakeinav1 @netrome I need one more approval. Please re-review when you get a chance. Thanks! 🙏🏼

crates/tee-verifier-interface/src/lib.rs
barakeinav1: Thanks — the code blocks are correct now. 
But the prose/diagrams didn't fully follow, though: the sequence 
diagram (line 92), the entry-removal list (line 149), the testing enumeration (line 574), and the flow prose 
 (line 62) still describe resolve_verification handling the verifier-failure case with an immediate refund+resume. Under the corrected code that path returns early and the ~200-block timeout callback does the 
cleanup — only post-DCAP failure resolves immediately. 

Could you align those too so the doc's failure-handling control flow matches?

barakeinav1: actualy more important,


the TEE verifier split (#3235/#3237/#3247): today a bad attestation fails submit_participant_info
  synchronously with a reason. After the split, verify_quote returns its error via #[handle_result], so a dcap
  rejection looks identical to "verifier down" (both PromiseError::Failed) and resolve_verification defers
  cleanup to the ~200-block yield timeout.

  Net: an accepted quote (or one that fails our own policy checks) resolves immediately, but a quote rejected by
  the verifier only gets its error + refund at the timeout — ~3–4 min, deterministically.

  The node polls state and resubmits hourly so I don't think it actually hurts us, but it's a real regression vs
  the current sync behavior.
  
   Are we fine with it, or do we want the rejection path to resume immediately?

pbeza: Yeah you're right, that'd be a real regression. Fixed it in fa44ba9b.

`verify_quote` returns a `VerificationResult { Verified, Rejected }` value now instead of going through `#[handle_result]`. So a dcap rejection comes back as `Rejected(..)` on a successful receipt, and the caller can actually tell it apart from `PromiseError::Failed` ("verifier down"); rejections resolve right away with the reason, and only a real no-verdict failure waits for the ~200-block timeout.

crates/tee-verifier-interface/src/lib.rs
netrome: Ah right, makes sense.

barakeinav1 (approved):
Re-reviewed `fa44ba9b`. The `VerificationResult { Verified, Rejected }` change resolves the regression cleanly — a rejected quote now resumes immediately with its reason, and only a genuinely unreachable verifier falls through to the timeout, which is the correct behavior. Doc is fully in sync now. Thanks for the quick turnaround. 👍

gilcu3 (approved):

PR #3236, opened 2026-05-14
refactor: improve `RecentBlocksTracker` invariants

### LINKED ISSUE #3248: Improve recent blocks tracker `prune_dead_children`
### Background

The current method `RecentBlocksTracker::prune_dead_children` on main is quite expensive. It goes from the minimum height to keep, all the way to the latest final height (so, almost 200 blocks). That's a non-negligible recursion depth.

A better approach would be to use a Fifo stack and not recursively call functions.

Additionally, we expose ourselves to a pathological case where `RecentBlocksTracker::canonical_head` may point to a node that is no longer present in the `RecentBlocksTracker::hash_to_node` map. While this is not something we expect to see in production, it does not feel good. 

If we were to refactor `maybe_update_final_head`, such that we get the following invariant:
*If a node is `final` at height `h`, then there is no non-final node of height less or equal to `h` in our tree.*

then this would improve readability of the pruning methods and result in easier to reason code.

Specifically, the algorithm could look like this:

1. Walk the tree from `potential_final_head` up to the first parent that was marked final:
    1. Label all nodes encountered as final (current behavior);
    3. Prune the subtrees of any siblings that have not finalized.
2. Prune any root nodes that are not final and of height less or equal to the current final height

If the indexer behaves as expected and is not skipping nay blocks, then we should be guaranteed to have a well behaved tree. And even if the indexer is skipping blocks (which it is not supposed to do), `prune_old_blocks` and `maybe_update_final_head` together would eventually clean up dangling root nodes.

### Acceptance criteria

- remove the recursive function call in `prune_dead_children`
- make `canonical_head` a weak pointer

### Additional Resources: 

 https://github.com/near/mpc/pull/3196#pullrequestreview-4270399101

kevindeforth:
This PR has `RecentBlocksTracker` provide the following invariants:
- _every_ block tracked by `RecentBlocksTracker` has the potential to become final;
- a final block will never become non-final.

resolves https://github.com/near/mpc/issues/3248.

### DIFF
--- a/crates/node/src/requests/recent_blocks_tracker.rs
+++ b/crates/node/src/requests/recent_blocks_tracker.rs
@@ -218,7 +218,7 @@ impl BlockNode {
     ) {
         let mut children = self.children.lock().expect("lock must not be poisoned");
         for child in children.iter() {
-            if child.hash != child_to_retain.hash {
+            if !Arc::ptr_eq(child, child_to_retain) {
                 subtrees_to_remove.push_back(child.clone());
             }
         }
@@ -427,7 +427,7 @@ impl<T: Clone + Debug> RecentBlocksTracker<T> {
         new_final_blocks
     }
 
-    /// Any root children that sit on dead branches get removed from `self.root_children` and add
+    /// Any root children that sit on dead branches get removed from `self.root_children` and added
     /// to `subtrees_to_remove`
     fn update_roots(
         &mut self,
--- a/crates/node/src/requests/recent_blocks_tracker.rs
+++ b/crates/node/src/requests/recent_blocks_tracker.rs
@@ -14,7 +14,7 @@ use std::sync::{Arc, Mutex, Weak};
 ///    blocks. The content of each block in the finalized stream is specified via the `T`
 ///    type parameter.
 ///  - For each block added via `add_block`, it returns a `Weak<AtomicBlockStatus>` that can be
-///    used to observe that block's current `BlockStatus`.
+///    used to observe that block's current `BlockStatus` (non-canonical, canonical or final).
 ///
 /// We have certain expectations of the order of blocks that come from the indexer.
 /// First, let's define a partial order for blocks. For any two blocks A and B:
@@ -43,19 +43,18 @@ use std::sync::{Arc, Mutex, Weak};
 ///  - We keep track of the canonical chain as well as the final chain.
 ///
 /// Despite the assumptions we make on the indexer's behavior, this class guarantees not to panic
-/// even if the indexer violates these assumptions in arbitrary ways. In particular, if the indexer
-/// feeds blocks such that the canonical chain ends up on a fork that a later final-block advance
-/// discards, the next `add_block` re-elects `canonical_head` over the remaining tree rather than
-/// holding a stale reference. As a side effect, downstream consumers (e.g. the request queue) will
-/// not be invited to act on requests that sit on a permanently-dead fork: the block_ref handed out
-/// by `add_block` upgrades to `None` as soon as the finality advance condemns the fork.
+/// even if the indexer violates these assumptions in arbitrary ways.
 ///
-/// **Cleanup overview.** Two cleanup paths run after every `add_block`:
+/// Note that the `RecentBlocksTracker` is removing blocks aggressively. A block is removed if one
+/// of the following conditions is met:
+///  - The block sits on a dead fork and can't ever be finalized;
+///  - The block is outside of the recency window `RecentBlocksTracker::window_size`
 ///
+/// Cleanup takes place after every `add_block` in two methods:
 ///  - `maybe_update_final_head` owns **dead-fork cleanup**. When a new final block is established,
 ///    every subtree that can't be part of the final chain — non-final siblings of the final chain
 ///    at any height, and any `root_children` subtree not on the final chain with height ≤ the new
-///    final-head height — is dropped entirely. This is the "Dead-fork prune at X" annotation in
+///    final-head height — is dropped. This is the "Dead-fork prune at X" annotation in
 ///    the picture below.
 ///  - `prune_old_blocks` owns **recency-window prune**. Any node below `min_height_to_keep` is
 ///    dropped; its in-window descendants become new roots. This is the "Recency window prune"
@@ -91,22 +90,14 @@ use std::sync::{Arc, Mutex, Weak};
 ///  │   │       │       │   │  ┆       │
 ///  ..  ..      ..      ..  .. ┆       ..
 /// ```
-///
-/// Note: The use of Arc is a little unfortunate. Logically we only need Rc, but this code needs
-/// to work in futures, and Rc is not Send. So we use Arc instead.
 pub struct RecentBlocksTracker<T: Clone + 'static> {
     window_size: u64,
     /// By "root", we mean the blocks whose parents we don't know or are older than the window.
     /// The children of the root are the earliest blocks we are keeping who do not have any order
     /// with each other.
     root_children: Vec<Arc<BlockNode>>,
-    /// The head of the canonical chain. Held as `Weak` so that when finality
-    /// cleanup sweeps a subtree containing the current canonical head, the
-    /// `BlockNode` drops naturally and the next `add_block` re-elects from the
-    /// remaining tree instead of clinging to a dangling node. `Weak::new()` at
-    /// construction time (which never upgrades) stands in for "no canonical
-    /// head yet" — the same `None`-on-upgrade behavior we already get when the
-    /// target is dropped, so the two cases collapse cleanly at every call site.
+    /// The head of the canonical chain. This is the chain of the highest-height block we've seen.
+    /// This may be None if we have no block at all.
     canonical_head: Weak<BlockNode>,
     /// The head of the final chain. This is determined by recovering information from the
     /// last_final_block fields of the block headers given to us. This may be None if we have not
@@ -187,7 +178,7 @@ pub struct AddBlockResult<T> {
     /// The list of newly finalized blocks, in ascending height order. Each entry is a tuple of
     /// the block height and the data passed to us when adding the block.
     /// It is guaranteed that the new final blocks returned from multiple calls to add_block are
-    /// contiguous, thus forming the stream of finalized blocks.
+    // contiguous, thus forming the stream of finalized blocks.
     pub new_final_blocks: Vec<(u64, T)>,
     pub block_ref: Weak<AtomicBlockStatus>,
 }
@@ -212,6 +203,20 @@ struct BlockNode {
 impl BlockNode {
     fn get_parent(&self) -> Option<Arc<BlockNode>> {
         self.parent.as_ref().and_then(Weak::upgrade)
+    }
+
+    fn keep_only_child(
+        &self,
+        child_to_retain: &Arc<BlockNode>,
+        subtrees_to_remove: &mut VecDeque<Arc<BlockNode>>,
+    ) {
+        let mut children = self.children.lock().expect("lock must not be poisoned");
+        for child in children.iter() {
+            if child.hash != child_to_retain.hash {
+                subtrees_to_remove.push_back(child.clone());
+            }
+        }
+        *children = vec![child_to_retain.clone()];
     }
 
     fn debug_print(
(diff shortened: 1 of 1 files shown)


crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -1040,23 +969,11 @@ pub mod tests {
...
-        //    │ ├─[8]     GounuZUfMmxdVequL65iUm7D94sKYg67Q34Z5cjRjZ71 "b100"
-        //    │ ├─[8]     7aXqE7cPZt6FnVcpytqW3oy5y2E17X9Gpgs6C7praap5 "b101"
-        //    │ └─[7]     FK5PX18pxwwtaB6AvYjNkWZ4Jvnn2HeNx4tknkM5g7VP "b102"
-        //    │   [8]     AQif6GckVVDt4L4rUeAN43PNComjsa7fPYAF5NHW7oX8 "b1020"
-        //    └─[7]     2xhND7PwiNZckwCnXFa25wMG1G8JWfKN2Sinmbz3W6xK "b11"
-        //      ├─[9]     6E2vqjZLbuUY2y6VK571Ai3pk43EUHXuNxibJuBGh44T "b110"
-        //      └─[10]     Cq9uMZqNZr6zuF7nT6yfGms1ptaVVu5dAdiRJ6pqCTN8 "b111"
-        //
-        // Note above: the canonical head is not a descendant of final head. This can't happen in
-        // the real blockchain, but here we fed in a pathological scenario.
```
kevindeforth, line 1059: Note that in the pathological case, we are pruning more aggressively now compared to before.

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -53,9 +100,14 @@ pub struct RecentBlocksTracker<T: Clone + 'static> {
...
-    /// This may be None if we have no block at all.
-    canonical_head: Option<Arc<BlockNode>>,
+    /// The head of the canonical chain. Held as `Weak` so that when finality
+    /// cleanup sweeps a subtree containing the current canonical head, the
+    /// `BlockNode` drops naturally and the next `add_block` re-elects from the
+    /// remaining tree instead of clinging to a dangling node. `Weak::new()` at
+    /// construction time (which never upgrades) stands in for "no canonical
+    /// head yet" — the same `None`-on-upgrade behavior we already get when the
+    /// target is dropped, so the two cases collapse cleanly at every call site.
+    canonical_head: Weak<BlockNode>,
```
kevindeforth, line 107: This is another change. I can't find it, but someone (or AI) commented on the possibility that in a pathological case, we would remove the `BlockNode` from the hash to node map, but keep a reference to it here.
This is now no longer possible.

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -506,95 +500,30 @@ impl<T: Clone + Debug> RecentBlocksTracker<T> {
...
-                final_head_height,
-                &mut new_roots,
-                &mut blocks_to_prune,
-            );
-        }
-
-        self.root_children = new_roots
-            .iter()
-            .filter_map(|hash| {
-                self.hash_to_node.get(hash).cloned().or_else(|| {
```
kevindeforth, line 579: note that we resolve https://github.com/near/mpc/pull/3209#discussion_r3239959517

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -487,7 +485,23 @@ impl<T: Clone + Debug> RecentBlocksTracker<T> {
...
+        let mut best: Option<Arc<BlockNode>> = None;
+        while let Some(node) = queue.pop_front() {
+            if best.as_ref().is_none_or(|b| b.height < node.height) {
+                best = Some(node.clone());
+            }
+            let children = node.children.lock().expect("lock must not be poisoned");
+            queue.extend(children.iter().cloned());
+        }
+        best
     }
```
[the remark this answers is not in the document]
kevindeforth: > highest_tracked_node walks the entire kept tree (BFS) every time the canonical head was swept by finality cleanup

that is correct but this shouldn't happen very often.

> preferring the just-inserted node if its height equals or exceeds any other in-tree node 

How do you figure out the "height of any other in-tree node", without a BFS walk?

> tracking the highest-height node incrementally so re-election is O(1).

not following. How do you do that across all branches of a tree efficiently?

> Additionally, BFS-order tie-breaking makes the re-elected canonical head dependent on insertion order rather than recency: in the updated test, the newly added b1020 (h=8) loses to the earlier-inserted b100 (h=8), which is a subtle behavioral change worth a comment on the function explaining the tie-break semantics.

This is not a behavioral change. This was already the case before.


_Bad bot!_

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -158,127 +207,22 @@ struct BlockNode {
...
+            if child.hash != child_to_retain.hash {
+                subtrees_to_remove.push_back(child.clone());
+            }
         }
-    }
-
-    fn get_parent(&self) -> Option<Arc<BlockNode>> {
-        self.parent.as_ref().and_then(Weak::upgrade)
+        *children = vec![child_to_retain.clone()];
     }
```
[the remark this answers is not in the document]
kevindeforth: Ok, I accept the suggestion of `ptr_eq`, but prefer readability over a minor efficiency gain when it comes to the vector https://github.com/near/mpc/pull/3236/commits/d086b0cbaee118e51189d3df7dba7c11b4980788

crates/node/src/requests/recent_blocks_tracker.rs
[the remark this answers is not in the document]
pbeza: > tracking the highest-height node incrementally so re-election is O(1).

@kevindeforth I think we could use a [`MaxHeap`](https://en.wikipedia.org/wiki/Heap_(data_structure)#Comparison_of_theoretic_bounds_for_variants) for that, but I’m not quite sure it’s worth it given that we’re running BFS on a tree with a 200-block window size, so the time complexity is bounded and effectively constant. Also, forks are probably rare anyway.

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -14,9 +14,15 @@ use std::sync::{Arc, Mutex, Weak};
 ///    blocks. The content of each block in the finalized stream is specified via the `T`
 ///    type parameter.
 ///  - For each block added via `add_block`, it returns a `Weak<AtomicBlockStatus>` that can be
-///    used to observe that block's current `BlockStatus`.
+///    used to observe that block's current `BlockStatus` (non-canonical, canonical or final).
```
pbeza, line 17: Nit: Perhaps worth adding that it’s _canonical_, but not final?


```suggestion
///    used to observe that block's current `BlockStatus` (non-canonical, canonical-but-not-final, or final).
```

crates/node/src/requests/recent_blocks_tracker.rs
pbeza: > (...) I can't find it (...)

https://github.com/near/mpc/pull/3196#discussion_r3218396995.

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -431,36 +380,85 @@ impl<T: Clone + Debug> RecentBlocksTracker<T> {
...
-                node.status.make_final();
-                new_final_blocks.push(node.clone());
-                let Some(parent) = node.get_parent() else {
-                    break;
-                };
-                node = parent;
+        let final_head_node = self.hash_to_node.get(&potential_final_head).cloned();
+        let Some(final_head_node) = final_head_node else {
+            return Vec::new();
+        };
```
pbeza, line 395: I’m not sure what “returning empty” means semantically here — or more specifically, *when* we hit this case. Is it during indexer cold start? Might be worth adding a short comment explaining it.

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -431,36 +380,85 @@ impl<T: Clone + Debug> RecentBlocksTracker<T> {
...
+        let mut new_final_blocks: Vec<Arc<BlockNode>> = Vec::new();
+        let mut node = final_head_node.clone();
+        let mut subtrees_to_remove: VecDeque<Arc<BlockNode>> = VecDeque::new();
+        loop {
+            // The node we previously visited is the only final child of `node`.
+            if let Some(final_child_of_node) = new_final_blocks.last() {
+                node.keep_only_child(final_child_of_node, &mut subtrees_to_remove);
```
pbeza, line 402: To me, passing mutable variables as args feels like a bit of a code smell, since they act as a kind of implicit return and make the code harder to reason about. I think we should try to structure the code so we don’t need that, if possible. How about:

```rust
fn keep_only_child(&self, child_to_retain: &Arc<BlockNode>) -> Vec<Arc<BlockNode>> {
    let mut children = self.children.lock().expect("lock must not be poisoned");
    let dead = children.iter()
        .filter(|c| c.hash != child_to_retain.hash)
        .cloned()
        .collect();
    *children = vec![child_to_retain.clone()];
    dead
}
```

and then call it here:

```
subtrees_to_remove.extend(node.keep_only_child(final_child_of_node));
```

BTW, it’s a bit unfortunate that we need to collect these subtrees manually instead of just having them dropped automatically on this assignment:

```rust
*children = vec![child_to_retain.clone()];
```

that would be great.

Claude claims it might be possible if we had a structure like this:

```
struct BlockNode {
    hash: CryptoHash,
    height: u64,
    status: Arc<AtomicBlockStatus>,
    parent_hash: Option<CryptoHash>,   // ← plain data, just a hash
    // no `children` field at all
    // no `parent: Weak<BlockNode>` — replaced by parent_hash
}
```

Apparently, we don’t access children through this field anywhere except in the debug impl, and cleanup could be automatic if we didn’t keep `Arc`s to children.

pbeza (dismissed):
I think the cleanup logic [could potentially be simplified even further](https://github.com/near/mpc/pull/3236#discussion_r3250527989), as it’s still quite nuanced, but overall this is a nice improvement over what we have now, so approving. Thank you!

crates/node/src/requests/recent_blocks_tracker.rs
pbeza: @kevindeforth, this is a vibe-coded PoC: https://github.com/near/mpc/pull/3251. It implements the idea a bit differently than by dropping the `children` field, but still simplifies pruning. It’s probably not ideal yet, but with a bit more thought, I think we could simplify this further. Not a blocker here.

crates/node/src/requests/recent_blocks_tracker.rs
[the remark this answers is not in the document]
kevindeforth: My issue with copilots suggestion was that it optimizes for a rare event at the expense of common events.
While a heap would make identifying the canonical head O(1), insertion would now be O(log(n)).
That doesn't seem like a good trade-off for our use-case.

edit: To be clear this here:
> every time the canonical head was swept by finality cleanup

should be pretty rare and is not something I want to optimize for at the expense of:
- code clarity
- complexity of adding a new block

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -431,36 +380,85 @@ impl<T: Clone + Debug> RecentBlocksTracker<T> {
...
+        self.root_children = new_root_children;
+    }
+
+    fn remove_subtrees(&mut self, mut subtrees_to_remove: VecDeque<Arc<BlockNode>>) {
+        while let Some(node) = subtrees_to_remove.pop_front() {
+            subtrees_to_remove.extend(
+                node.children
+                    .lock()
+                    .expect("lock must not be poisoned")
+                    .clone(),
```
barakeinav1: nit: `.drain(..)` instead of `.clone()` here matches what `prune_old_blocks` does and saves a strong-count bump per child — the parent's children Vec is about to be dropped this iteration anyway, so emptying it eagerly is equivalent in behavior.

barakeinav1 (dismissed):
Approving — nice refactor, the eager-prune-at-finality model with stronger invariants reads much cleaner than the previous lazy sweep. One small `.drain(..)` nit left inline.

crates/node/src/requests/recent_blocks_tracker.rs
kevindeforth: Will skip this one - it's already stated in the `BlockStatus` enum.

crates/node/src/requests/recent_blocks_tracker.rs
kevindeforth: Good suggestion. Addressed in https://github.com/near/mpc/pull/3236/commits/fa8d5f40dd47a7dd40ff722aaf0cf037789b8c80

crates/node/src/requests/recent_blocks_tracker.rs
kevindeforth: Good suggestion! Addressed in https://github.com/near/mpc/pull/3236/commits/fa8d5f40dd47a7dd40ff722aaf0cf037789b8c80

crates/node/src/requests/recent_blocks_tracker.rs
kevindeforth: Oh, interesting. #3251 looks pretty complicated. If you don't mind, I will merge like this and look at it in detail once we have more time.

barakeinav1 (approved):

pbeza (approved):

PR #3237, opened 2026-05-14
feat(tee-verifier): stateless `dcap_qvl::verify` wrapper contract

### LINKED ISSUE #3266: Introduce `tee-verifier` contract (stateless `dcap_qvl::verify` wrapper)
### Background

Today the MPC contract performs Intel TDX quote verification itself, which bakes a large cryptographic library — [`dcap-qvl`](https://github.com/Phala-Network/dcap-qvl) and its `ring` / `webpki` / `x509-cert` dependencies — into its on-chain binary, bloating its size and forcing every other team that wants quote verification to bundle the same library.

This issue introduces a small, standalone verifier contract whose only job is to verify a quote and return the result. Moving that one step into its own contract lets the MPC contract (and anyone else) request verification over a cross-contract call instead of carrying `dcap-qvl`, and keeps the verifier independently auditable and upgradeable.

### User Story

As an MPC contract maintainer, I want a dedicated, stateless verifier contract that performs quote verification, so the MPC contract can delegate that step instead of embedding the cryptographic library in its own binary.

### Acceptance Criteria

- [ ] New crate `tee-verifier` under `crates/`.
- [ ] A single entry point that takes a quote and returns the verification result, using the shared types from #3265.
- [ ] Stateless — the contract holds no on-chain state and behaves as a pure function of its inputs.
- [ ] Actual deployment of the contract is tracked separately; merging this work closes the issue.

PR: #3237. Depends on #3265. Tracked by #3264.

pbeza:
Closes #3266

## Overview

Adds `tee-verifier`: a stateless TEE attestation verifier contract.

It wraps `dcap_qvl::verify::verify` in a single `verify_quote` method. The contract holds no state and has no admin — verifier-internal policy (the `dcap-qvl` version, Intel root certs, etc.) is bound to the deployed code hash. Per-team allowlists, report-data binding, and other post-DCAP checks live in the caller, not here.

This moves the heavy `dcap-qvl` / `ring` / `webpki` / `x509-cert` closure out of any contract that needs quote verification and into this one contract, reached over a cross-contract call using the DTOs from #3235.

Builds on #3235 (PR #3235); part of the stack tracked by #3264. See `docs/design/attestation-verifier-contract.md` for the design.

### DIFF
--- a/crates/tee-verifier/Cargo.toml
+++ b/crates/tee-verifier/Cargo.toml
@@ -1,0 +1,48 @@
+[package]
+name = "tee-verifier"
+version = { workspace = true }
+license = { workspace = true }
+edition = { workspace = true }
+repository = "https://github.com/near/mpc"
+
+[package.metadata.cargo-shear]
+ignored = ["borsh"]
+
+[package.metadata.near.reproducible_build]
+image = "sourcescan/cargo-near:0.17.0-rust-1.86.0"
+image_digest = "sha256:1784ca6310f3496f0048356ce420921c8f5fdf71ee8124d43a2e1ceb1f70db8a"
+passed_env = []
+container_build_command = [
+    "cargo",
+    "near",
+    "build",
+    "non-reproducible-wasm",
+    "--locked",
+    "--features",
+    "abi",
+]
+
+[lib]
+crate-type = ["cdylib", "lib"]
+
+[features]
+# Enabled by `cargo near build` (via `--features near-sdk/__abi-generate`)
+# and required for ABI / Borsh schema generation. Pulls in
+# `borsh/unstable__schema` and `BorshSchema` derives on the wire types.
+abi = ["borsh/unstable__schema", "tee-verifier-interface/borsh-schema"]
+
+[dependencies]
+borsh = { workspace = true }
+dcap-qvl = { workspace = true }
+near-sdk = { workspace = true }
+tee-verifier-interface = { workspace = true }
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+getrandom = { workspace = true, features = ["custom"] }
+
+[dev-dependencies]
+hex = { workspace = true }
+test-utils = { workspace = true }
+
+[lints]
+workspace = true
--- a/crates/tee-verifier/Cargo.toml
+++ b/crates/tee-verifier/Cargo.toml
@@ -45,6 +45,7 @@ getrandom = { workspace = true, features = ["custom"] }
 
 [dev-dependencies]
 hex = { workspace = true }
+rstest = { workspace = true }
 tee-verifier = { path = ".", features = ["test-utils"] }
 test-utils = { workspace = true }
 
--- a/crates/tee-verifier/Cargo.toml
+++ b/crates/tee-verifier/Cargo.toml
@@ -4,9 +4,6 @@ version = { workspace = true }
 license = { workspace = true }
 edition = { workspace = true }
 
-[package.metadata.cargo-shear]
-ignored = ["borsh"]
-
 [package.metadata.near.reproducible_build]
 image = "sourcescan/cargo-near:0.21.1-rust-1.93.0"
 image_digest = "sha256:79b789c81ba19ec30794a6b1046b02dcaf5c055994b22cc7c9af67ab5096b095"
--- a/crates/tee-verifier/Cargo.toml
+++ b/crates/tee-verifier/Cargo.toml
@@ -8,8 +8,8 @@ edition = { workspace = true }
 ignored = ["borsh"]
 
 [package.metadata.near.reproducible_build]
-image = "sourcescan/cargo-near:0.17.0-rust-1.86.0"
-image_digest = "sha256:1784ca6310f3496f0048356ce420921c8f5fdf71ee8124d43a2e1ceb1f70db8a"
+image = "sourcescan/cargo-near:0.21.1-rust-1.93.0"
+image_digest = "sha256:79b789c81ba19ec30794a6b1046b02dcaf5c055994b22cc7c9af67ab5096b095"
 passed_env = []
 container_build_command = [
     "cargo",
--- a/crates/tee-verifier/Cargo.toml
+++ b/crates/tee-verifier/Cargo.toml
@@ -14,6 +14,7 @@ container_build_command = [
     "build",
     "non-reproducible-wasm",
     "--locked",
+    "--profile=release-contract",
     "--features",
     "abi",
 ]
--- a/crates/tee-verifier/Cargo.toml
+++ b/crates/tee-verifier/Cargo.toml
@@ -38,7 +38,6 @@ borsh = { workspace = true }
 dcap-qvl = { workspace = true }
 near-sdk = { workspace = true }
 tee-verifier-interface = { workspace = true }
-thiserror = { workspace = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 getrandom = { workspace = true, features = ["custom"] }
--- a/crates/tee-verifier/Cargo.toml
+++ b/crates/tee-verifier/Cargo.toml
@@ -3,7 +3,6 @@ name = "tee-verifier"
 version = { workspace = true }
 license = { workspace = true }
 edition = { workspace = true }
-repository = "https://github.com/near/mpc"
 
 [package.metadata.cargo-shear]
 ignored = ["borsh"]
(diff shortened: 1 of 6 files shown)


crates/tee-verifier/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+
+#[near]
+impl TeeVerifier {
+    /// Verify a TDX quote against Intel collateral.
+    ///
+    /// Calls `dcap_qvl::verify::verify` with the current block timestamp
+    /// and returns the parsed `VerifiedReport` on success. The caller is
+    /// responsible for any post-DCAP policy (RTMR3 replay, report-data
+    /// binding, measurement allowlist matching, etc.).
+    #[handle_result]
```
barakeinav1: Since DCAP rejections panic (18aca60c), the caller only sees `PromiseError::Failed` — indistinguishable from "verifier down", and the dcap error detail no longer reaches the submitter's tx outcome (today `submit_participant_info` returns it synchronously via `InvalidTeeRemoteAttestation { reason }`; after the split it's only on the verifier's failed receipt). Also, according to Claude's explanation, a rejected quote would then take the `Err(promise_err)` arm of the design doc's `resolve_verification` draft and wait out the ~200-block yield timeout (~3–4 min) for its refund, instead of the immediate refund the `Ok(Err(verifier_err))` arm was meant to provide — am I reading that right? If so, is that acceptable for #3247, or will `resolve_verification` refund immediately on `PromiseError::Failed`? Either way the design doc still describes the decodable-error flow (with `VerifierError` as an interface-crate DTO), so it needs realigning — could you do that in this PR?

barakeinav1 (commented):
Verified locally: the Borsh pinning tests cover every mirrored type, `verify_quote` is the same `dcap_qvl::verify::verify` invocation as today's `attestation.rs` call site, and tests/clippy/wasm32 are all green. +1 on adding a negative-path test asserting `Err(DcapVerification)` (e.g. flipped quote byte or stale timestamp) — direct calls can exercise the `Err` branch without the panic wrapper.

Holding off on approval until the inline design-doc question is sorted.

crates/tee-verifier/tests/verify_quote.rs
```diff
@@ -0,0 +1,124 @@
+//! Integration test for the stateless `tee-verifier` contract.
+//!
+//! Calls `TeeVerifier::verify_quote` directly (no Promise round-trip)
+//! with a real Dstack quote+collateral fixture taken from `test-utils`,
+//! and asserts the returned `VerifiedReport` matches the fixture's known
+//! value in full.
+
+#![allow(non_snake_case)]
```
barakeinav1: nit: `#![expect(non_snake_case)]` over `#![allow(...)]` — repo convention (`conversions.rs` in this PR uses `#[expect]`).

barakeinav1 (commented):
left a comment/question [here](https://github.com/near/mpc/pull/3237#pullrequestreview-4444798348)

crates/tee-verifier/src/lib.rs
pbeza: Addressed in #3235.

crates/tee-verifier/tests/verify_quote.rs
pbeza: Fixed in https://github.com/near/mpc/pull/3237/changes/06cf823b39545804b9538d66dda3eb7d2ba807f7.

crates/tee-verifier/tests/verify_quote.rs
```diff
@@ -0,0 +1,177 @@
...
+}
+
+fn make_quote_bytes() -> QuoteBytes {
+    QuoteBytes(Vec::from(quote()))
+}
+
+/// Sets a VM context whose block timestamp makes the fixture's collateral
+/// current, so DCAP verification of the valid fixture succeeds.
+fn set_valid_timestamp_context() {
+    let block_timestamp_ns = Duration::from_secs(VALID_ATTESTATION_TIMESTAMP).as_nanos() as u64;
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/pull/3237/commits/b612db828a6e6d1096b201e2ec0a9112cae6e178

crates/tee-verifier/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+license = { workspace = true }
+edition = { workspace = true }
+
+[package.metadata.cargo-shear]
+ignored = ["borsh"]
+
+[package.metadata.near.reproducible_build]
+image = "sourcescan/cargo-near:0.17.0-rust-1.86.0"
+image_digest = "sha256:1784ca6310f3496f0048356ce420921c8f5fdf71ee8124d43a2e1ceb1f70db8a"
+passed_env = []
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/pull/3237/commits/b612db828a6e6d1096b201e2ec0a9112cae6e178

crates/tee-verifier/Cargo.toml
```diff
@@ -0,0 +1,52 @@
+[package]
+name = "tee-verifier"
+version = { workspace = true }
+license = { workspace = true }
+edition = { workspace = true }
+
+[package.metadata.cargo-shear]
+ignored = ["borsh"]
```
barakeinav1: nit: cargo-shear flags this as a redundant ignore. Your layout-pinning tests in `conversions.rs` use `borsh::` directly, so unlike the contract-crate template you copied this from (`chain-gateway-test-contract`, `test-parallel-contract`, which don't touch borsh in tests), this crate doesn't need the ignore — safe to drop the `[package.metadata.cargo-shear]` block.

barakeinav1 (dismissed):
LGTM, aligned to the changes we agreed upon in #3235. `verify_quote` returning `VerificationResult` resolves the failure-distinguishability concern, conversions are field-for-field correct and pinned against `dcap-qvl`, and the negative-path tests are a good add. 👍

One nit inline.

crates/tee-verifier/Cargo.toml
pbeza: Fixed in https://github.com/near/mpc/pull/3237/commits/38774c5ee6a2bed0eddd16b33e9e6f27fc9e9b43. ✅

barakeinav1 (dismissed):

crates/tee-verifier/tests/verify_quote.rs
```diff
@@ -0,0 +1,179 @@
...
+        tcb_info_signature: hex::decode(v["tcb_info_signature"].as_str().unwrap()).unwrap(),
+        qe_identity_issuer_chain: v["qe_identity_issuer_chain"].as_str().unwrap().to_string(),
+        qe_identity: v["qe_identity"].as_str().unwrap().to_string(),
+        qe_identity_signature: hex::decode(v["qe_identity_signature"].as_str().unwrap()).unwrap(),
+        pck_certificate_chain: v
+            .get("pck_certificate_chain")
+            .and_then(|s| s.as_str())
+            .map(str::to_string),
+    }
+}
```
gilcu3, line 44: I am surprised we need such manual conversion here. I would expect we should have that conversion code somewhere already

gilcu3 (approved):
Thank you!

crates/tee-verifier/Cargo.toml
```diff
@@ -0,0 +1,49 @@
...
+passed_env = []
+container_build_command = [
+    "cargo",
+    "near",
+    "build",
+    "non-reproducible-wasm",
+    "--locked",
+    "--features",
+    "abi",
+]
```
gilcu3, line 20: ups small blocker here, we should use `--profile=release-contract`

gilcu3 (changes requested):
ups found a small blocker

crates/tee-verifier/tests/verify_quote.rs
pbeza: Yeah, looks like there are no existing conversions for this, but we can easily add one (see demo: https://github.com/near/mpc/pull/3521).

The issue is that we don’t want to add a dependency on MPC-specific crates to `tee-verifier`, since it’s supposed to stay MPC-agnostic and potentially be used by external teams too.

crates/tee-verifier/Cargo.toml
pbeza: https://github.com/near/mpc/pull/3237/commits/6082844530a482cc9ff2983182f6cd86162f96a7 ✅

crates/tee-verifier/tests/verify_quote.rs
gilcu3: yeah, maybe in a future refactor this will become easier, thank you for giving it a try

gilcu3 (approved):

barakeinav1 (approved):

PR #3249, opened 2026-05-15
feat: implement voting for RPC providers

### LINKED ISSUE #3215: Implement voting mechanism for adding / removing allowed RPC provider
### Background

MPC contract will store allowed RPC providers. We need to have mechanism for adding / removing RPC providers via node provider voting.

### User Story

_No response_

### Acceptance Criteria

Node providers are able to vote for adding / removing RPC provider.

_No response_

anodar:
Note: each vote is a full snapshot rather than diff.

Closes: https://github.com/near/mpc/issues/3215


anodar:
@claude review

anodar:
@claude review

anodar:
@claude review

crates/contract/src/foreign_chain_rpc.rs
```diff
@@ -1,179 +1,791 @@
...
-            return false;
-        };
-        let removed = bucket.remove(provider_id).is_some();
-        if bucket.is_empty() {
-            self.entries.remove(&chain);
-        }
-        removed
+    #[cfg(test)]
+    pub fn get(&self, chain: ForeignChain) -> Option<&ChainEntry> {
+        self.entries.get(&chain)
```
kevindeforth: we have `snapshot`. Would feel cleaner to me to write a test helper that implements `get` on `snapshot` and lives in the test module.

crates/contract/src/foreign_chain_rpc.rs
```diff
@@ -1,179 +1,791 @@
...
+            chain,
+            providers,
+            threshold: response_quorum,
+        } in votes
+        {
+            if chains_in_batch.contains(&chain) {
+                return Err(InvalidParameters::MalformedPayload {
+                    reason: format!("duplicate chain {chain:?} in vote batch"),
+                }
+                .into());
```
kevindeforth: This seems like a map `chain <--> ProvidersConfig` could be a better suited data structure, where
```rust
ProvidersConfig {
    providers: BTreeSet<ProviderEntry>,
    threshold: ProvidersThreshold
}

struct ProvidersThreshold(u64);
```


packing `providers` into a `BTreeSet` also removes the need for the `canonicalize` function. We may even be able to use a `NonEmptyBTreeSet` from near-mpc-bounded-collections, though I don't know if they behave well in contract code (cc @DSharifi).

crates/contract/src/foreign_chain_rpc.rs
kevindeforth, line 1: Did you consider using the generic votes struct?
So, instead of:
```rust
pub struct ProviderVotes {
    pub pending: BTreeMap<(AuthenticatedParticipantId, ForeignChain), ChainEntry>,
}
```    

We would have
```rust
pub struct ProviderVotes {
    pub pending: Votes<(AuthenticatedParticipantId, ForeignChain)>,
}
``` 

_edit: to be explicit about the motivation: `Votes` already implements a lot of the logic needed for this file, is robust and relatively light-weight (though, disclaimer: I would expect contract binary size to increase compared to what is currently proposed)._

https://github.com/near/mpc/blob/9dc8c46b165cc15777844e31df626a1b19595e57/crates/contract/src/primitives/votes.rs#L22-L117

crates/contract/src/foreign_chain_rpc.rs
```diff
@@ -1,179 +1,791 @@
...
+        self.entries.clone()
+    }
+}
+
+// Flat `(participant, chain)` key (rather than a nested map) halves the BTreeMap
+// monomorphizations the contract WASM has to pay for.
+#[near(serializers=[borsh])]
+#[derive(Debug, Default, PartialEq, Eq)]
+pub struct ProviderVotes {
+    pub pending: BTreeMap<(AuthenticatedParticipantId, ForeignChain), ChainEntry>,
```
kevindeforth: `IterableMap` is preferred in contract code, as its entries are lazy-loaded (we get this for free with the generic votes structs).

kevindeforth (commented):
Thank you for this!
Left a few questions regarding the data structures used in the contract.

I would [suggest](https://github.com/near/mpc/pull/3249#discussion_r3264051399) to use the generic votes struct.

Regarding the high-level API, I might have missed it, but is there a way for  a user to withdraw / cancel a vote?

crates/contract/src/foreign_chain_rpc.rs
anodar: Done.

crates/contract/src/foreign_chain_rpc.rs
anodar: Change to Votes<>

crates/contract/src/foreign_chain_rpc.rs
anodar: Done.

crates/contract/src/foreign_chain_rpc.rs
anodar: Changed to NonEmptyBTreeMap

crates/contract/src/foreign_chain_rpc.rs
```diff
@@ -1,179 +1,770 @@
...
+            if segment.contains('/') {
+                return Err(InvalidParameters::MalformedPayload {
+                    reason: format!(
+                        "ChainRouting::PathSegment.segment for provider_id {:?} must not contain '/'",
+                        id.0
+                    ),
+                }
+                .into());
+            }
+        }
```
kevindeforth: This might be a dumb question, but why do we need this check in the contract?

crates/contract/src/foreign_chain_rpc.rs
```diff
@@ -1,179 +1,770 @@
...
+        {
+            if routing_name == auth_name {
+                return Err(InvalidParameters::MalformedPayload {
+                    reason: format!(
+                        "ChainRouting::QueryParam.name collides with AuthScheme::Query.name {:?} for provider_id {:?}",
+                        auth_name, id.0
+                    ),
+                }
+                .into());
+            }
```
kevindeforth: same here, why do we check for this specifically?

crates/contract/src/foreign_chain_rpc.rs
```diff
@@ -1,179 +1,770 @@
...
+/// Validate an entry. `entry.providers` is a `NonEmptyBTreeMap<ProviderId, ProviderConfig>`
+/// at the type level (enforced at borsh-deserialize), which gives us three things for
+/// free: non-empty, no duplicate `provider_id`, and ordered iteration — so the
+/// canonical hash matches across voters without an explicit sort.
+///
+/// `quorum` is the RPC response quorum (`ChainEntry.quorum`), not the protocol signing
+/// threshold. Errors on: `quorum == 0` or exceeding `providers.len()`, `PathSegment`
+/// containing a literal `/`, or `QueryParam` whose name collides with the entry's
+/// `AuthScheme::Query` name.
+fn canonicalize(entry: ChainEntry) -> Result<ChainEntry, Error> {
```
kevindeforth: nit: _to canonicalize_ means to normalize data, changing its shape.

Here, I think we are just validating the data and returning an error in case the content itself is invalid.

Have you considered making this `ChainEntry::validate(&self) -> Result<(), Error>`?

crates/contract/src/foreign_chain_rpc.rs
```diff
@@ -1,179 +1,770 @@
...
+    /// The input batch is a `NonEmptyBTreeMap<ForeignChain, ChainEntry>`, so two
+    /// invariants are enforced at borsh-deserialize time and don't need to be re-checked
+    /// here: the batch is non-empty, and each chain appears at most once.
+    ///
+    /// Returns the chains applied this call.
+    pub fn vote(
+        &mut self,
+        participant: AuthenticatedParticipantId,
+        votes: NonEmptyBTreeMap<ForeignChain, ChainEntry>,
+        protocol_threshold: u64,
```
kevindeforth: Instead of passing `participant` and `protocol_threshold` separately, we can pass `ThresholdParameters`

crates/contract/src/lib.rs
```diff
@@ -1470,6 +1471,63 @@ impl MpcContract {
...
+
+        let threshold_parameters = self
+            .protocol_state
+            .threshold_parameters()
+            .unwrap_or_else(|ContractNotInitialized| {
+                env::panic_str(
+                    "Contract is not initialized. Cannot vote for foreign chain providers before initialization. \
+                     This should be unreachable — voter_or_panic() above already errors on NotInitialized.",
+                )
+            });
```
kevindeforth: nit: unreachable seems more fitting in this case:

```suggestion
let threshold_parameters = self
    .protocol_state
    .threshold_parameters()
    .unwrap_or_else(|_| unreachable!(
        "voter_or_panic() above already errors on NotInitialized"
    ));
```

crates/contract/src/lib.rs
```diff
@@ -1470,6 +1471,63 @@ impl MpcContract {
...
+        &mut self,
+        #[serializer(borsh)] votes: near_mpc_bounded_collections::NonEmptyBTreeMap<
+            dtos::ForeignChain,
+            dtos::ChainEntry,
+        >,
+    ) -> Result<(), Error> {
+        log!(
+            "vote_update_foreign_chain_providers: signer={}, n_votes={}",
+            env::signer_account_id(),
+            votes.len(),
```
kevindeforth, line 1488: Why do we only print the length of votes? Printing the hash would seem more helpful.

crates/contract/src/lib.rs
```diff
@@ -1785,6 +1847,17 @@ impl MpcContract {
...
+        // PR 3216 shipped `foreign_chain_rpc_whitelist` with a nested-map layout
+        // that PR 3249 reshaped. Guards against an environment where PR 3216 was
+        // deployed standalone before PR 3249 lands.
+        match try_state_read::<pr_3216_state::MpcContract>() {
+            Ok(Some(state)) => return Ok(state.into()),
+            Ok(None) => return Err(InvalidState::ContractStateIsMissing.into()),
+            Err(err) => {
+                log!("failed to deserialize state into PR 3216 state: {:?}", err);
+            }
+        };
```
kevindeforth: We do not use PR numbers for contract versions, that's what we have tags for.

PR numbers are quite confusing, as the merge order may not match the PR number order. Additionally, it's easy to arrive at a broken state.

This is a blocker.

crates/contract/src/foreign_chain_rpc.rs
```diff
@@ -1,179 +1,770 @@
...
+}
 
-    /// All providers currently whitelisted for `chain`, in `provider_id` order.
-    pub fn get(&self, chain: ForeignChain) -> impl Iterator<Item = &ProviderEntry> {
-        self.entries
-            .get(&chain)
-            .into_iter()
-            .flat_map(|bucket| bucket.values())
+impl ProviderVotes {
+    pub fn retain_only(&mut self, current: &Participants) {
```
kevindeforth: nit: why not just call it `retain`?

crates/contract/src/foreign_chain_rpc.rs
```diff
@@ -1,179 +1,770 @@
...
+impl AllowedProviders {
+    fn replace(&mut self, chain: ForeignChain, entry: ChainEntry) {
+        self.entries.insert(chain, entry);
+    }
+
+    /// Snapshot of the whole whitelist. Collected so the caller can ship it across
+    /// the contract boundary without holding a borrow on `self`.
+    pub fn snapshot(&self) -> BTreeMap<ForeignChain, ChainEntry> {
+        self.entries.iter().map(|(c, e)| (*c, e.clone())).collect()
+    }
```
kevindeforth: Why is holding a reference a problem? The caller can still decide to clone if they need.

crates/contract/src/foreign_chain_rpc.rs
```diff
@@ -1,179 +1,770 @@
...
+            let count =
+                u64::try_from(count_usize).map_err(|e| ConversionError::DataConversion {
+                    reason: format!("vote count {count_usize} does not fit in u64: {e}"),
+                })?;
+            if count >= protocol_threshold {
+                self.entries.replace(chain, entry);
+                // Drop ALL pending rows for this chain regardless of which proposal
+                // they held — matches the previous `clear_chain` semantics.
+                self.votes.pending.retain_votes(|(_, c)| *c != chain);
+                applied.push(chain);
```
kevindeforth: nit:
Might read nicer if we moved this logic to a method `ProviderVotes::vote(hash, threshold, participant_id, participants) -> bool`:

```rust
    if self.votes.vote(hash, threshold, participant_id, participants) {
        self.entries.insert(chain, entry);
        applied.push(chain);
    }
   
```

crates/contract/src/lib.rs
```diff
@@ -1470,6 +1471,63 @@ impl MpcContract {
...
+            participant,
+            votes,
+            protocol_threshold,
+            participants,
+        )?;
+        log!(
+            "vote_update_foreign_chain_providers: applied chains={:?}",
+            applied,
+        );
+        Ok(())
```
kevindeforth: Q: Is there a reason we wouldn't want to return the bool?

crates/contract/src/lib.rs
```diff
@@ -1470,6 +1471,63 @@ impl MpcContract {
...
     }
 
+    /// Vote on per-chain RPC provider whitelist state. The input is keyed by
+    /// `ForeignChain`; each `ChainEntry` value carries the proposed full provider list
+    /// and the RPC response quorum for that chain. The chain's stored state is replaced
+    /// once the protocol's signing threshold of participants has voted the same
+    /// `(providers, quorum)` pair. `NonEmptyBTreeMap` enforces a non-empty batch and
+    /// at-most-one entry per chain at borsh-deserialize time.
+    #[handle_result]
+    pub fn vote_update_foreign_chain_providers(
```
kevindeforth, line 1475: this method is missing a unit test

kevindeforth (commented):
Some comments and questions.
Partial review only, didn't look at the tests yet.

crates/contract/src/lib.rs
DSharifi: If we are doing nits, I think expect is the best here to panic. We have an expectation for it to not error.

crates/contract/src/foreign_chain_rpc.rs
anodar: This is to check the invariant: for URL construction, segment shouldn't contain `/`, e.g. `eth/sepolia` which is invariant for all RPC providers schemes we checked. We do this sanity check before accepting a vote so that invalid config doesn't get stored in allowlisted providers.

crates/contract/src/foreign_chain_rpc.rs
anodar: Some RPC providers take chain and auth parameter as query parameter, something like: `?network=ethereum`, `?dkey=<TOKEN>`. They can't be the same param obviously, it's a trivial sanity check for copy/pasting errors.

kevindeforth: Makes sense, but are these checks exhaustive?
I.e. if a config passes these checks, is that a sufficient condition for the config to be valid?

If not, then the voter will still need to make some manual verification and we might as well remove these.

crates/contract/src/foreign_chain_rpc.rs
kevindeforth: lets continue in the other thread: https://github.com/near/mpc/pull/3249#discussion_r3275221463

crates/contract/src/foreign_chain_rpc.rs
anodar: It's exactly so the caller doesn't have to clone,  `allowed_foreign_chain_providers` returns borsch serialized entries which requires an owned object, this is a helper for cloning it.

crates/contract/src/foreign_chain_rpc.rs
anodar: Done.

crates/contract/src/lib.rs
anodar: Changed to 3.10

crates/contract/src/lib.rs
anodar: Technically they could all vote different thing and still agree on some chain providers, hash on whole messages will all be different. But in practice it's probably all same so makes sense to print hash.
Alternatively we could print hash per chain rpc provider entries but leaving it to be on whole message.

crates/contract/src/lib.rs
anodar: Changed to expect.

crates/contract/src/foreign_chain_rpc.rs
anodar: I'm assuming you mean `participants` (as list of mpc participants) not `participant` (AuthenticatedParticipantId). Changed it.

crates/contract/src/foreign_chain_rpc.rs
anodar: The name is remnant from the code that was indeed changing the data, specifically sorting it, which is no longer the case. Validate is more suitable now, changed to that.

crates/contract/src/foreign_chain_rpc.rs
anodar: Done.

crates/contract/src/lib.rs
anodar: Added `vote_update_foreign_chain_providers__should_apply_chain_and_return_it_when_threshold_reached`

crates/contract/src/lib.rs
anodar: Yes because vote can technically apply some chains but not all. Changed to return `Vec<dtos::ForeignChain>`.

crates/contract/src/foreign_chain_rpc.rs
anodar: It's not exhaustive but I'm not sure that's a good reason for dropping sanity checks entirely. These are checks that can be easily checked, other validity checks can't be easily checked from the contract.

anodar:
> Regarding the high-level API, I might have missed it, but is there a way for a user to withdraw / cancel a vote?

There's no explicit withdraw vote, but you can submit different vote for that chain and it will override your current one.

crates/contract/src/foreign_chain_rpc.rs
kevindeforth: Interesting. Then, I have two follow-up questions:
- where will the other validity checks take place?
- what guarantees do the validity checks on the contract provide for the data stored on-chain?

crates/contract/src/foreign_chain_rpc.rs
anodar: Reworded comment above method to make it clearer.

crates/contract/src/foreign_chain_rpc.rs
anodar: > where will the other validity checks take place?

Full validity check will be manual, when we add a new RPC provider we have to manually check how does one construct a query (how is chain specified, how is authentication done), then define a config according to it and do simple probe test. We can add e2e tests where one adds new config for new RPC provider and expected response to validate config of new RPC before proposing it. Although I'd rather do that in separate PR to not bloat this further.

> what guarantees do the validity checks on the contract provide for the data stored on-chain?

It's more sanity check than validations. It currently guarantees that you don't make simple copy/paste error e.g. where you pass same query param name for auth and chain, or that you don't have backslash in chain name.

crates/contract/src/v3_10_state.rs
```diff
@@ -25,45 +26,87 @@ use crate::{
...
+    pub(crate) proposed_updates: ProposedUpdates,
+    pub(crate) node_foreign_chain_support: SupportedForeignChainsByNode,
+    pub(crate) config: Config,
+    pub(crate) tee_state: TeeState,
+    pub(crate) accept_requests: bool,
+    pub(crate) node_migrations: NodeMigrations,
+    pub(crate) legacy_pending_requests: LegacyPendingRequests,
+    pub(crate) metrics: dtos::Metrics,
+    pub(crate) foreign_chain_rpc_whitelist: OldForeignChainRpcWhitelist,
+}
```
DSharifi: Why do we need pub(crate) on all these fields?

DSharifi (dismissed):

crates/contract/src/v3_10_state.rs
anodar: You're right, we don't. Dropped it and also moved `dead_code` annotation to field rather than whole struct.

crates/contract/src/lib.rs
```diff
@@ -1470,6 +1470,57 @@ impl MpcContract {
...
+        &mut self,
+        #[serializer(borsh)] votes: near_mpc_bounded_collections::NonEmptyBTreeMap<
+            dtos::ForeignChain,
+            dtos::ChainEntry,
+        >,
```
DSharifi, line 1480: Seems like my comment in previous review got dropped, thanks github.

Any reason we need to use borsh serializer for this argument?

DSharifi (dismissed):

scripts/check-contract-wasm-size.sh
```diff
@@ -11,8 +11,12 @@ set -euo pipefail
 
 WASM_PATH="${1:-target/near/mpc_contract/mpc_contract.wasm}"
 
-# NEAR max_transaction_size = 1572864; keep some headroom
-HARD_LIMIT=1496000
+# NEAR max_transaction_size = 1572864; keep some headroom.
+# Raised from 1495000 to fit the on-chain RPC provider whitelist voting
+# machinery (issue #3215), then bumped again to fit the `Votes<V>` +
+# `IterableMap` migration which adds generic-monomorphization overhead.
+HARD_LIMIT=1570000
```
kevindeforth: Do we still need this high of a limit, now that we merged https://github.com/near/mpc/pull/3280?

crates/contract/src/lib.rs
anodar: `ChainEntry` does not have serde JSON derived for wasm32 target to keep the generated wasm size down (that's when I was trying to squeeze it under the previous limit), so this has to be borsh unless we add serde for wasm32 target as well.

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
```diff
@@ -1029,6 +1028,94 @@ pub struct ProviderEntry {
...
+        "ChainRouting::PathSegment.segment for provider_id {provider_id:?} must not contain '/'"
+    )]
+    PathSegmentContainsSlash { provider_id: String },
+    #[error(
+        "ChainRouting::QueryParam.name collides with AuthScheme::Query.name {name:?} for provider_id {provider_id:?}"
+    )]
+    QueryParamCollidesWithAuth { provider_id: String, name: String },
+    #[error("providers.len() {len} does not fit in u64: {reason}")]
+    ProvidersLenOverflow { len: usize, reason: String },
+}
```
kevindeforth: What is the reason for defining these errors in the contract-interface crate? 
 
Until now, convention in the repo is to define contract errors in `contract/src/errors.rs`.

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
```diff
@@ -1029,6 +1028,94 @@ pub struct ProviderEntry {
...
+
+impl ChainEntry {
+    /// Validate shape invariants. See [`ChainEntryValidationError`] for the rules.
+    pub fn validate(&self) -> Result<(), ChainEntryValidationError> {
```
kevindeforth: Hmm, two things come to mind:
- I think we are missing unit tests for this method;
- I don't think this is the right place for the method. The contract-interface crate is supposed to be pretty light-weight and not hold custom logic. The way we solved this for other types in the contact that require custom logic is to have two types: one light-weight in the contract interface crate, one with the validation logic in the contract crate.

crates/contract/tests/sandbox/utils/consts.rs
```diff
@@ -27,15 +27,25 @@ pub const GAS_FOR_INIT: Gas = Gas::from_tgas(300);
...
 /// This vote is more expensive because it deploys the new contract code and executes
-/// the migration function.
-pub const MAX_GAS_FOR_THRESHOLD_VOTE: Gas = Gas::from_tgas(185);
+/// the migration function. Raised from 185 → 190 → 200 TGas as the foreign-chain
+/// RPC providers feature (issue #3215) grew the WASM: first to add the whitelist +
+/// `v3_10_state` try-read step, then again when the `Votes<V>` + `IterableMap` +
+/// `NonEmptyBTreeMap` reshape pushed the binary slightly larger. Revert toward 185
+/// once `v3_10_state` is retired and `Votes<V>` monomorphization overhead has been
+/// amortized across other consumers.
+pub const MAX_GAS_FOR_THRESHOLD_VOTE: Gas = Gas::from_tgas(200);
```
kevindeforth: Please verify if this is still required after removing the migration logic.

crates/contract/tests/sandbox/utils/consts.rs
```diff
@@ -27,15 +27,25 @@ pub const GAS_FOR_INIT: Gas = Gas::from_tgas(300);
 /// TODO(#1571): Gas cost for voting on contract updates. Reduced somewhat after
 /// optimization (#1617) by avoiding full contract code deserialization; there’s likely still
 /// room for further optimization.
-pub const GAS_FOR_VOTE_UPDATE: Gas = Gas::from_tgas(260);
+/// Bumped 260 → 290 TGas after the `Votes<V>` + `IterableMap` migration grew the
+/// reproducible WASM. The threshold vote spawns `deploy_contract(code)` whose
+/// action cost scales with code size; the prior 260 TGas budget was failing with
+/// `Exceeded the prepaid gas` once `burnt + promises_gas` crossed the limit
+/// (`test_propose_update_contract_many` in particular).
+pub const GAS_FOR_VOTE_UPDATE: Gas = Gas::from_tgas(290);
```
kevindeforth: idem

crates/contract/src/v3_10_state.rs
```diff
@@ -1,18 +1,19 @@
-//! ## Overview
-//! This module stores the previous contract state—the one you want to migrate from.
-//! The goal is to describe the data layout _exactly_ as it existed before.
+//! State-migration shim from contract release `3.10.0` to the current state shape.
 //!
-//! ## Guideline
-//! In theory, you could copy-paste every struct from the specific commit you're migrating from.
-//! However, this approach (a) requires manual effort from a developer and (b) increases the binary size.
-//! A better approach: only copy the structures that have changed and import the rest from the existing codebase.
```
kevindeforth, line 8: I find the existing comments pretty helpful, as they provide a strategy for new devs.
Arguably, everyone in the team should be aware of it by now, but I would prefer to preserve them.

crates/contract/tests/sandbox/user_views.rs
```diff
@@ -65,3 +70,66 @@ async fn test_derived_public_key() -> anyhow::Result<()> {
...
+        .view(method_names::ALLOWED_FOREIGN_CHAIN_PROVIDERS)
+        .args_json(json!({}))
+        .await?
+        .borsh()?;
+    let stored = whitelist
+        .get(&ForeignChain::Ethereum)
+        .expect("Ethereum entry should be present after 6 matching votes");
+    assert_eq!(stored.providers.len(), 1);
+    assert_eq!(stored.providers.get(&proposed_id), Some(&proposed_config));
+    assert_eq!(stored.quorum, 1);
```
kevindeforth, line 158: nit: sanity check that the entry is not present before the last vote is cast would be sensible.

crates/contract/src/lib.rs
```diff
@@ -6341,4 +6394,63 @@ mod tests {
         // Then
         assert!(result.is_empty());
     }
+
+    #[test]
+    fn vote_update_foreign_chain_providers__should_apply_chain_and_return_it_when_threshold_reached(
```
kevindeforth, line 6236: Would be nice to have a regression test that non-participants are not allowed to vote.

Can be a follow-up

kevindeforth (commented):
Thanks for the hard work!

Some things to check before merging:
- error placement seems not consistent with the rest of the crate
- missing unit test for validation method (& its placement)
- we should check if the gas constants improved after merging #3280. They seem pretty high to me.

crates/contract/src/lib.rs
anodar: Added `vote_update_foreign_chain_providers__should_panic_when_caller_is_not_a_participant`

crates/contract/tests/sandbox/user_views.rs
anodar: Done.

crates/contract/src/v3_10_state.rs
anodar: Reverted.

scripts/check-contract-wasm-size.sh
anodar: Lowered the limit after merging, both contract size and gas.

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
anodar: I'd expect any type that's returned from Contract to be in contract-interface crate, including error. Although this is used internally and not returned from a contract.

Moved it to `errors.rs`.

crates/contract/tests/sandbox/utils/consts.rs
anodar: Reverted, not required after merge.

crates/contract/tests/sandbox/utils/consts.rs
anodar: Same here.

anodar:
Thanks for the thorough review!

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
anodar: Got rid of validate and validation is now in ` TryFrom<dtos::ChainEntry> for ChainEntry` and it's tested. 

Also introduced other type as you described.

kevindeforth (approved):
Thanks for the changes!

DSharifi (approved):

PR #3234, opened 2026-05-14
docs: fix operator deployment guide

barakeinav1:
## Summary

Closes #3233. Fixes the blocking errors an operator hits when following the deployment docs literally — TOML schema drift in the `user-config.toml` sample, an undefined `FUTURE_PORT` env var the deploy script depends on, wrong paths/OS-image in `deploy-launcher-guide`, `get_tee_accounts` called as a transaction instead of a view, the `DSTACK_ENDPOINT` contradiction in the non-TEE doc, and a `migration_web_ui` bind that didn't match the documented `:8079` port-forward.

Adjacent typo/cleanup fixes landed in the same files; broader (non-blocking) items from the review are deferred.

## Test plan

- [x] `bash -n` on the touched shell scripts
- [ ] One end-to-end `deploy-launcher.sh` against testnet
- [ ] `deploy-tee-cluster.sh` two-node localnet smoke
- [ ] Markdown render check on GitHub

PR #3263, opened 2026-05-18
chore(cvm-deployment): align user-config.toml example with prod values

### LINKED ISSUE #3257: Align example configuration values with our current testnet/mainnet configuration
### Background

Currently, when new operators follow the guide at https://github.com/near/mpc/blob/main/docs/running-an-mpc-node-in-tdx-external-guide.md#prepare-mpc-node-configuration they will by default configure the node based on this file https://github.com/near/mpc/blob/main/deployment/cvm-deployment/user-config.toml.

This is quite different to how our nodes are configured in production. For example we typically set `desired_triples_to_buffer` to `16384` in our production nodes, whereas anyone following this file would set it to `128`.

### User Story

As an onboarding node operator I'd like to end up with a similar configuration to the other production nodes.

### Acceptance Criteria

We've updated the example configuration to match our current production values.

_No response_

barakeinav1:
Closes #3257.

The previous `user-config.toml` was a CI smoke-test config the TDX operator guide pointed operators at. This PR splits the two consumers:

- **`ci-config.toml`** (new) — keeps the previous minimal values; the non-TEE CI smoke test mounts it.
- **`user-config.toml`** — real testnet operator template, values cross-validated against `/debug/node_config` on every live mainnet and testnet operator. Inline `# mainnet: …` comments show what to swap for mainnet.
- **`docs/running-an-mpc-node-in-tdx-external-guide.md`** — small additions on top of #3234 (now in main).
- **`docs/using-the-launcher-in-nontee-setup.md`** — points non-TEE / local-dev users at `ci-config.toml` rather than the testnet `user-config.toml`.

### Why two files
CI wants a tiny config that boots cheaply on a small runner; the chain doesn't matter. Operators want production-aligned values and a real chain. CI's trick — `chain_id = "mainnet"` + `download_genesis = false` — only works because nearcore embeds the mainnet genesis in its binary; testnet has no equivalent and downloading it OOMs the runner (PR #2720). One file can't serve both.

### `cores = 6` (live mainnet operators have 12)
CVM VCPU is attested at 8. The MPC tokio runtime is deliberately separate from the indexer's so the indexer isn't starved (`coordinator.rs:238`). 6 leaves 2 vCPUs of headroom.

For reference, all 8 live mainnet operators currently report `cores = 12` via `/debug/node_config` — but those are 12-vCPU legacy GCP VMs (also matches `start.sh:115`'s hardcoded default), not 8-vCPU TDX CVMs. The template targets the launcher / TDX path (attested `VCPU=8` from `deploy-launcher.sh:135`), so `cores = 12` would oversubscribe; `cores = 6` is the right scaled-down value for this audience.

### `foreign_chains` schema notes
- Dropped legacy `api_variant = "standard"` lines. The field is silently accepted but no longer functional in the current node-config schema — see the test `config_parsing__should_succeed_with_legacy_field__api_variant` at `crates/node-config/src/foreign_chains.rs:564`.
- **Forward-looking heads-up:** the design at `docs/design/allowing-per-node-foreign-chain-rpc-configuration.md` (referenced by #3216 / cf0c430f) describes a future shift where operators will write `provider_id` + a token env reference instead of raw `rpc_url`s — the contract will hold `base_url` / `auth_scheme` / `chain_routing`. The contract-side types landed in cf0c430f but the node-side migration hasn't yet, so this PR uses today's `rpc_url` shape. A future PR will need to update this section of `user-config.toml`.

### Operator-guide additions
Update the `user-config.toml` example in `running-an-mpc-node-in-tdx-external-guide.md` to better match the production-aligned template (on top of the schema fixes #3234 already merged).

### Open questions
- `presignature.concurrency 4 → 16` — outside #3257's stated AC; drop if undesired.
- DSS hints stay commented-out (`n1-multichain.near` runs bucket-only; `deploy-tee-cluster.sh` defaults DSS-first — separate decision).
- Mainnet `foreign_chains.*.rpc_url` inline hints not yet added — happy to add (URLs surveyed: `https://bitcoin-rpc.publicnode.com`, `https://api.mainnet.abs.xyz`, `https://starknet-rpc.publicnode.com`).

### Out of scope (separate follow-ups)
- `start.sh:98`, `deployment/testnet/frodo.toml`, `deployment/testnet/sam.toml` still carry stale `desired_triples_to_buffer = 1000000` — live mainnet config is `16384` (verified via `/debug/node_config` on all 8 mainnet operators). Worth cleaning up in a separate PR.

### Effect on existing nodes
No hot-reload. Operators unaffected until they re-pull and restart. Raising buffer ceilings is safe.

### DIFF
--- a/deployment/cvm-deployment/ci-config.toml
+++ b/deployment/cvm-deployment/ci-config.toml
@@ -1,0 +1,78 @@
+# Minimal config for non-TEE CI check.
+# Only needs enough for the launcher to start and pull the MPC node image.
+# Uses mainnet chain_id with embedded genesis to avoid downloading the large
+# testnet genesis file, which requires ~20 GiB of RAM.
+
+[launcher_config]
+image_reference = "nearone/mpc-node"
+port_mappings = [
+    { host = 8080, container = 8080 },
+]
+
+[mpc_node_config]
+home_dir = "/data"
+
+# PCCS endpoints for TDX attestation collateral. The node tries each
+# entry in order until one succeeds. Defaults below: Phala primary,
+# Intel direct as fallback.
+#
+# To add a self-hosted local PCCS (e.g. for an operator running their
+# own), prepend an entry with a `tls` override. See the operator
+# deployment guide's "Self-hosting a local PCCS" appendix for the cert
+# setup. Example (commented):
+#
+#   [[mpc_node_config.pccs_endpoints]]
+#   url = "https://10.0.2.2:8081/"
+#   tls = { override = "ca_cert_pem", ca_cert_pem = "..." }
+
+[[mpc_node_config.pccs_endpoints]]
+url = "https://pccs.phala.network/"
+
+[[mpc_node_config.pccs_endpoints]]
+url = "https://api.trustedservices.intel.com/"
+
+[mpc_node_config.log]
+format = "plain"
+filter = "info"
+
+[mpc_node_config.near_init]
+chain_id = "mainnet"
+boot_nodes = "ed25519:ERguu7jQuYk8pxNsRC6FdezvNsegBPva1GRGqjmtD7i2@10.10.10.10:24567"
+download_genesis = false
+download_config = "rpc"
+
+[mpc_node_config.secrets]
+secret_store_key_hex = "BD399143F5B3126098B0EAA023A0E730"
+backup_encryption_key_hex = "0000000000000000000000000000000000000000000000000000000000000000"
+
+[mpc_node_config.node]
+my_near_account_id = "test_image.near"
+near_responder_account_id = "test_image.near"
+number_of_responder_keys = 1
+web_ui = "0.0.0.0:8080"
+migration_web_ui = "0.0.0.0:8078"
+cores = 4
+
+[mpc_node_config.node.indexer]
+validate_genesis = false
+sync_mode = "Latest"
+concurrency = 1
+mpc_contract_id = "v1.signer"
+finality = "optimistic"
+
+[mpc_node_config.node.triple]
+concurrency = 2
+desired_triples_to_buffer = 128
+timeout_sec = 60
+parallel_triple_generation_stagger_time_sec = 1
+
+[mpc_node_config.node.presignature]
+concurrency = 4
+desired_presignatures_to_buffer = 64
+timeout_sec = 60
+
+[mpc_node_config.node.signature]
+timeout_sec = 60
+
+[mpc_node_config.node.ckd]
+timeout_sec = 60
--- a/deployment/cvm-deployment/launcher_docker_compose_nontee.yaml
+++ b/deployment/cvm-deployment/launcher_docker_compose_nontee.yaml
@@ -10,7 +10,7 @@ services:
 
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
-      - ./user-config.toml:/tapp/user_config:ro
+      - ./ci-config.toml:/tapp/user_config:ro
       - shared-volume:/mnt/shared:rw
       - mpc-data:/data
 
--- a/deployment/cvm-deployment/user-config.toml
+++ b/deployment/cvm-deployment/user-config.toml
@@ -1,12 +1,26 @@
-# Minimal config for non-TEE CI check.
-# Only needs enough for the launcher to start and pull the MPC node image.
-# Uses mainnet chain_id with embedded genesis to avoid downloading the large
-# testnet genesis file, which requires ~20 GiB of RAM.
+# Example user-config.toml for new MPC node operators.
+# Defaults target testnet and mirror what NEAR One runs on its live
+# testnet node. Inline `# mainnet: ...` comments below show the value
+# to use when running on mainnet instead.
+#
+# MUST be replaced with real values before deploying:
+#   - my_near_account_id / near_responder_account_id
+#   - secrets.secret_store_key_hex
+#   - secrets.backup_encryption_key_hex
+# The placeholders below are intentionally non-hex so a forgotten swap
+# fails loudly at node startup.
 
 [launcher_config]
 image_reference = "nearone/mpc-node"
+# Matches the canonical port list in
+# docs/running-an-mpc-node-in-tdx-external-guide.md#required-ports
+# (80 P2P, 8080 web/metrics, 8079 migration, 3030 debug, 24567 DSS).
 port_mappings = [
-    { host = 8080, container = 8080 },
+    { host = 80,    container = 80 },     # MPC P2P (port_override convention)
+    { host = 8080,  container = 8080 },   # web_ui / metrics / /public_data
+    { host = 8079,  container = 8079 },   # migration_web_ui
+    { host = 3030,  container = 3030 },   # debug / telemetry
+    { host = 24567, container = 24567 },  # decentralized state sync
 ]
 
 [mpc_node_config]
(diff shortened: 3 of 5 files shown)


barakeinav1:
> **PR title type suggestion:** This PR updates configuration files in the deployment directory, not documentation. The type prefix should be `chore:` instead of `docs:`.

Title has been changed ✓

barakeinav1:
Walking through the findings:

> **Blocking** — `deployment/cvm-deployment/user-config.toml:10-12` — `port_mappings` only publishes `8080`, but line 82 sets `port_override = 80`, which forces every peer URL (listen + dial) onto port 80. […] An operator copying this template as-is will boot with broken P2P.

Fixed. Cross-checked against the operator guide's example block (`docs/running-an-mpc-node-in-tdx-external-guide.md:642-648`) and the required-ports table (`docs/running-an-mpc-node-in-tdx-external-guide.md:791-799`) — both list the same 5 ports, and Bob's `node_config.toml` in `mpc-private` is identical. Template now publishes:

```toml
port_mappings = [
    { host = 80,    container = 80 },     # MPC P2P (port_override convention)
    { host = 8080,  container = 8080 },   # web_ui / metrics / /public_data
    { host = 8079,  container = 8079 },   # migration_web_ui
    { host = 3030,  container = 3030 },   # debug / telemetry
    { host = 24567, container = 24567 },  # decentralized state sync
]
```

Also aligned `migration_web_ui` from `0.0.0.0:8078` → `0.0.0.0:8079` so the binding port matches the exposed port (Bob, start.sh, and the operator guide all use 8079).

> **Non-blocking** — `secret_store_key_hex = "BD399143F5B3126098B0EAA023A0E730"` is the same publicly-known CI fixture key. […] Consider `"REPLACE_ME_BEFORE_DEPLOY"` (or any value that fails hex-decoding) so a forgotten swap fails loudly at startup.

Fixed. Both secrets are now non-hex sentinels so startup fails loudly if forgotten:

```toml
secret_store_key_hex = "REPLACE_ME_WITH_16_BYTE_HEX_KEY_"
backup_encryption_key_hex = "REPLACE_ME_WITH_32_BYTE_HEX_KEY_REPLACE_ME_WITH_32_BYTE_HEX_KEY_"
```

Inline comments now reference `openssl rand -hex 16` / `openssl rand -hex 32` for generation, and the header comment lists these as MUST-replace fields.

> **Non-blocking** — `api_variant = "standard"` is a legacy field that is silently ignored (see the regression test `config_parsing__should_succeed_with_legacy_field__api_variant` at `crates/node-config/src/foreign_chains.rs:565-612`). Recommend dropping the three `api_variant` lines.

Already dropped in the current commit (this review was on an earlier branch state). The legacy-field finding is cited in the PR description under "`foreign_chains` schema notes".

> **Non-blocking** — `docs/using-the-launcher-in-nontee-setup.md:57` still points readers at `deployment/cvm-deployment/user-config.toml` as the example for the non-TEE setup, but that file is now testnet with `download_genesis = true` — exactly the configuration the header of `ci-config.toml` says will OOM small runners.

Fixed. That section now points non-TEE / local-dev users at `ci-config.toml` and explicitly calls out the chain/RAM trade-off with a pointer to #2720.

deployment/cvm-deployment/user-config.toml
```diff
@@ -1,7 +1,9 @@
-# Minimal config for non-TEE CI check.
-# Only needs enough for the launcher to start and pull the MPC node image.
-# Uses mainnet chain_id with embedded genesis to avoid downloading the large
-# testnet genesis file, which requires ~20 GiB of RAM.
+# Example user-config.toml for new MPC node operators.
+# Defaults target testnet and mirror what NEAR One runs on its live
+# testnet node. Inline `# mainnet: ...` comments below show the value
+# to use when running on mainnet instead.
+# Replace the placeholder account IDs, secret_store_key_hex and
```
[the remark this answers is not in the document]
barakeinav1: > The header describes this as an operator-ready template, but `launcher_config.port_mappings` still only exposes 8080. Operators generally need additional inbound ports exposed (at least the MPC P2P port 80 and state-sync 24567; the TDX operator guide also lists 3030 and the migration/web UI ports).

Fixed in the latest commit on the branch. `port_mappings` now exposes the full canonical 5-port list from `docs/running-an-mpc-node-in-tdx-external-guide.md#required-ports` (80 / 8080 / 8079 / 3030 / 24567), with inline purpose comments on each entry. Matches the operator-guide example block (`:642-648`) and Bob's `node_config.toml` in `mpc-private` exactly. Also aligned `migration_web_ui` from `8078` → `8079` so the binding port matches the exposed port.

deployment/cvm-deployment/launcher_docker_compose_nontee.yaml
```diff
@@ -10,7 +10,7 @@ services:
 
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
-      - ./user-config.toml:/tapp/user_config:ro
+      - ./ci-config.toml:/tapp/user_config:ro
```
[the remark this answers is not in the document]
barakeinav1: > This compose file now mounts `./ci-config.toml` instead of `./user-config.toml`. Existing docs in the repo still instruct non-TEE users to create/mount `user-config.toml`; following those instructions will now fail with a missing file.

Fixed in the latest commit. `docs/using-the-launcher-in-nontee-setup.md:57` now points non-TEE / local-dev users at `ci-config.toml` (the embedded-mainnet-genesis CI smoke-test config), and explicitly calls out that `user-config.toml` is the sibling testnet operator template — which would download the multi-GB testnet genesis at startup and OOM small dev runners (citing #2720 for context).

deployment/cvm-deployment/ci-config.toml
```diff
@@ -0,0 +1,78 @@
...
+
+[[mpc_node_config.pccs_endpoints]]
+url = "https://pccs.phala.network/"
+
+[[mpc_node_config.pccs_endpoints]]
+url = "https://api.trustedservices.intel.com/"
+
+[mpc_node_config.log]
+format = "plain"
+filter = "info"
```
barakeinav1, line 36: on bob testnet I put "mpc=debug,info"
what do you prefer?

deployment/cvm-deployment/user-config.toml
```diff
@@ -53,37 +70,72 @@ download_config = "rpc"
...
-near_responder_account_id = "test_image.near"
+my_near_account_id = "your-mpc-account.testnet"
+near_responder_account_id = "your-mpc-account.testnet"
 number_of_responder_keys = 1
 web_ui = "0.0.0.0:8080"
 migration_web_ui = "0.0.0.0:8079"
-cores = 4
+# MPC tokio runtime worker threads (separate from the indexer's). Kept
+# below the CVM's attested VCPU=8 so the indexer isn't starved.
+cores = 6
```
barakeinav1, line 89: see PR description for more details

deployment/cvm-deployment/user-config.toml
```diff
@@ -53,37 +70,72 @@ download_config = "rpc"
...
 concurrency = 2
-desired_triples_to_buffer = 128
+desired_triples_to_buffer = 16384
 timeout_sec = 60
 parallel_triple_generation_stagger_time_sec = 1
 
 [mpc_node_config.node.presignature]
-concurrency = 4
-desired_presignatures_to_buffer = 64
+concurrency = 16
```
barakeinav1, line 108: taken from 
http://multichain-mainnet-0.nearone.org:8080/debug/node_config

deployment/cvm-deployment/user-config.toml
netrome: Makes sense

netrome (approved):
Thank you! This will be very helpful. We need to think how we communicate and manage foreign rpc configuration, but that's currently in flux so let's leave it open for now.

deployment/cvm-deployment/user-config.toml
gilcu3: it seems that naming the var as just `cores` is therefore a bit misleading.

deployment/cvm-deployment/user-config.toml
```diff
@@ -53,37 +70,72 @@ download_config = "rpc"
...
+
+[mpc_node_config.node.foreign_chains.abstract.providers.abstract-testnet]
+rpc_url = "https://api.testnet.abs.xyz"
+
+[mpc_node_config.node.foreign_chains.starknet]
+timeout_sec = 30
+max_retries = 3
+
+[mpc_node_config.node.foreign_chains.starknet.providers.publicnode]
+rpc_url = "https://starknet-sepolia-rpc.publicnode.com"
```
gilcu3, line 141: shouldn't we use fake placeholders here, so that operators are forced to add the values? same as we did for many values above?

gilcu3 (approved):
LGTM

deployment/cvm-deployment/user-config.toml
barakeinav1: to be hosent, I'm not really sure what should we have here. I think we should provide a list of all supported RPC,
but I'm not sure if we should have it here with real values, or just place a holder. ( @DSharifi  to comment on this)
Let create a follow up PR for foreign_chains section.

barakeinav1: follow up item: https://github.com/near/mpc/issues/3278

PR #3272, opened 2026-05-18
fix: correctly label succeeded and failed transactions

### LINKED ISSUE #3182: bug: `mpc_outgoing_transaction_outcomes` has wrong labels
### Background

Our metrics claim the respond transaction didn't execute in case we do not find the signature request in the queue. This should be the other way around. If the respond transaction was successful, the request will have been removed from the state.

https://github.com/near/mpc/blob/0800d67aca1b25cbb0e22300f56d90150057c129/crates/node/src/indexer/tx_sender.rs#L183

However, this metric is unreliable in other ways: the node never checks that the transaction was in the state before sending the transaction. So, in case the node sends the response before the request was included on-chain, this metric wouldn't catch that.

### User Story

As a node operator, I want meaningful metrics.

### Acceptance Criteria

Fix the semantics of the metrics for respond transactions (applies to signature, ckd and foreign tx responses)

https://nearone.slack.com/archives/C0912BTG51T/p1778237569083459?thread_ts=1778158255.862259&cid=C0912BTG51T

DSharifi:
closes #3182

### DIFF
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -1842,7 +1842,7 @@ impl MpcContract {
     /// only correct interpretation is presence: `Some(_)` vs `None`.
     ///
     /// The `Option<YieldIndex>` shape is retained for JSON wire compatibility with
-    /// out-of-tree consumers; the in-tree caller (`tx_sender::compute_transaction_status`)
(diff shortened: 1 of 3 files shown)


SimonRastikian (dismissed):
Thanks

SimonRastikian (approved):

pbeza (approved):

PR #3275, opened 2026-05-18
ci: don't run CI tests on doc only change prs

### LINKED ISSUE #695: [Task] Conditional CI test running
### Background

Suppose you change `readme.md`. You don't want to run the whole test suite

### User Story

As a developer, when I do changes only to the, for example, `tee_launcher`, I don't want to run and **wait** for the tests in the different parts of the project, i.e. mpc and contract. Only e2e/integratio tests should be run, if the respective source has changed 

### Acceptance Criteria

Only run tests that depend on changed logic

Decide on dependancy graph between CI jobs. Check what sources have changed (e.g. via `git diff`) and trigger respective jobs

anodar:
Closes: https://github.com/near/mpc/issues/695

Note: we can't just use `paths-ignore` in `pull_request` directly as jobs below won't run then, and one won't be able to merge doc only changes as these jobs are marked as required in our branch protection.

### DIFF
--- a/.github/workflows/changes.yml
+++ b/.github/workflows/changes.yml
@@ -1,0 +1,41 @@
+# Reusable workflow: detects whether a PR changes.
+#
+# Currently it returns a single `code_changed` boolean used to skip the full CI
+# suite on docs-only PRs. Later this can grow into per-area filters (e.g.
+# `node_changed`, `contract_changed`) so each test job runs only when its
+# sources actually changed.
+
+name: Changes
+
+on:
+  workflow_call:
+    outputs:
+      code_changed:
+        description: "'true' if any non-docs file changed in the PR; 'true' on non-PR events"
+        value: ${{ jobs.detect.outputs.code_changed }}
+
+jobs:
+  detect:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      code_changed: ${{ steps.filter.outputs.code || 'true' }}
+    steps:
+      - name: Checkout repository
+        if: github.event_name == 'pull_request'
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          persist-credentials: false
+      - name: Detect changed paths
+        if: github.event_name == 'pull_request'
+        id: filter
+        uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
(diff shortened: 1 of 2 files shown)


anodar:
@claude review

anodar:
@claude 

> Branch protection still needs the "Detect non-docs changes" job added to the required-checks list so docs-only PRs can satisfy required-check rules via the gated jobs' skipped conclusions. Without this, branch protection will block docs-only PRs since the existing required checks now resolve to skipped. This is a settings change, not a code change — confirming it's been done (or queued) would be worth a one-line note on the PR.


"non-docs changes" jobs are already marked as required. Docs only jobs will be skipped, which satisfied required-check rule.

.github/workflows/changes.yml
```diff
@@ -0,0 +1,42 @@
...
+on:
+  workflow_call:
+    outputs:
+      code_changed:
+        description: "'true' if any non-docs file changed in the PR, or on non-PR events; 'false' only on docs-only PRs"
+        value: ${{ jobs.detect.outputs.code_changed }}
+
+jobs:
+  detect:
+    runs-on: ubuntu-latest
```
netrome: Why don't we use a warpbuilds runner like the other jobs? e.g. `warp-ubuntu-2404-x64-2x`

netrome (dismissed):
Nice stuff - just one question but not a hard blocker

.github/workflows/changes.yml
anodar: Good point, changed.

netrome (approved):

.github/workflows/ci.yml
```diff
@@ -163,12 +183,16 @@ jobs:
...
+    needs: changes
+    if: ${{ always() && needs.changes.outputs.code_changed != 'false' }}
     permissions:
       contents: read
     uses: ./.github/workflows/nix-build-mpc-node.yml
 
   fast-ci-checks:
     name: "Fast CI checks"
+    needs: changes
+    if: ${{ always() && needs.changes.outputs.code_changed != 'false' }}
```
netrome: Let's still run the fast checks though

```suggestion
```

netrome (changes requested):
Minor change requested

.github/workflows/ci.yml
anodar: Done.

gilcu3 (approved):
Thank you!

netrome (approved):
Thanks for updating!

PR #3284, opened 2026-05-19
build(nix): remove stale neard flake

### LINKED ISSUE #3283: Remove orphaned nix/neard.nix
### Background

`nix/neard.nix` is not referenced anywhere in the repo — `flake.nix` does not `callPackage` it, and a repo-wide grep finds no other importers. It was added in #1812 but its consumer has since been removed, so it's dead code.

It currently declares `version = "2.10.4"` and pins commit `3c4442e48af2e0ad12e84e10fec269d1b25638fc` as a workaround for nearcore PR near/nearcore#14902, while the workspace `Cargo.toml` is on nearcore tag `2.11.1` (which already includes that PR). Rather than refresh the rev/hash, the file should just be removed.

### Acceptance Criteria

- `nix/neard.nix` is deleted.
- `nix build` and the flake's dev shell still work.

gilcu3:
Closes #3283 

This is now possible because pytests are gone.


barakeinav1 (approved):
LGTM

DSharifi (approved):

PR #3285, opened 2026-05-19
fix: make robust-ecdsa tasks infallible

### LINKED ISSUE #2134: robust-ecdsa background tasks should be infallible
### Background

In https://github.com/near/mpc/pull/2133 we made Cait-Sith ECDSA background tasks fallible. We should do the same for robust-ecdsa, as it might suffer from the same problems we have seen recently in production where tasks die

### User Story

_No response_

### Acceptance Criteria

The background tasks for robust-ecdsa are infallible

_No response_

gilcu3:
Closes #2134


crates/node/src/providers/robust_ecdsa/presign.rs
```diff
@@ -70,7 +70,7 @@ pub(super) async fn run_background_presignature_generation(
     domain_id: DomainId,
     presignature_store: Arc<PresignatureStorage>,
     keygen_out: KeygenOutput,
-) -> anyhow::Result<()> {
+) {
```
netrome: I realize we should use the never type here to ensure we don't have any return statements that kill the task at all: https://doc.rust-lang.org/reference/types/never.html

We should do the same for the other infallible tasks, as they should not return.

```suggestion
)  -> ! {
```

crates/node/src/providers/robust_ecdsa/presign.rs
```diff
@@ -87,14 +87,20 @@ pub(super) async fn run_background_presignature_generation(
...
+        running_participants.len(),
+    ) {
+        Ok(values) => values,
+        Err(err) => {
+            tracing::error!(
+                "Aborting robust-ecdsa presignature generation for domain {:?}: invalid threshold configuration: {}",
+                domain_id,
+                err
+            );
+            return;
```
netrome: This return would kill the task, which is what we want to avoid. Could we break out this to a separate function and take the robust ECDSA threshold as input to the task?

Otherwise, this function still needs to return an error since the caller should be able to decide what to do if this threshold is invalid - e.g. if it's a panic or something to recover from.

netrome (changes requested):
Hmm we need to think about how to handle if the threshold is invalid. It's not obvious to me right now, but the current `return` statement defeats the purpose of this change.

crates/node/src/providers/robust_ecdsa/presign.rs
gilcu3: Could be, although if the task enters in an impossible state, then returning seems strictly better than keep trying without hope

crates/node/src/providers/robust_ecdsa/presign.rs
gilcu3: >Could we break out this to a separate function and take the robust ECDSA threshold as input to the task?

We could, but the end result would be the same, we would not have a task to begin with. I think returning an error would be better, unless we can ofc guarantee that the task cannot fail except for transient errors. I am not sure we strictly guarantee that for the other tasks though.

crates/node/src/providers/robust_ecdsa/presign.rs
gilcu3: we agreed on taking the never type and making the task panic in case parameter is wrong

crates/node/src/providers/robust_ecdsa/presign.rs
gilcu3: we no killing the task through a panic.

netrome (approved):
Thanks for updating 🙏

crates/node/src/providers/robust_ecdsa/presign.rs
netrome: Had a synchronous discussion - we agreed panicking on these states that should be impossible are appropriate

SimonRastikian (approved):
Nice thanks

PR #3291, opened 2026-05-19
fix: Release workflow env and secrets

andrei-near:
Re-organising release GHA workflow secrets handling.
each reusable workflow has the environment already setup. This means we don't need to pass the secrets to them.


gilcu3 (approved):
LGTM

netrome (approved):
Thanks!

PR #3299, opened 2026-05-20
feat: split ecdsa triple pool per threshold

### LINKED ISSUE #3292: Split ECDSA triple pool per threshold and migrate existing triples on startup
### Background

Spun out of #3164. After #3143, each `DomainConfig` carries its own `reconstruction_threshold`. A CaitSith triple's Shamir degree is `t − 1`, so a triple is only valid when combined with keyshares of matching `t`. The node currently keeps a single shared `TripleStorage` and a single `run_background_triple_generation` task; #3143 worked around the divergence risk by reading `t` from the first CaitSith domain in `per_domain_data` (`EcdsaSignatureProvider::spawn_background_tasks` / `run_triple_generation_follower`). Once domains can choose different `reconstruction_threshold` values, that shortcut is unsound.

In our current system triples implicitly contain the threshold, as the number of participants computing a triple always equals the current threshold. Therefore, it is simpler to arrange triples in pools where each pools contains triples generated by the same number of participants.

### Acceptance Criteria

- The node should be backwards and forwards compatible. To achieve this, a new database might be created, and write operations might use both the old and the new database columns.
- `TripleStorage` and `run_background_triple_generation` are partitioned per threshold — one pool and one background generator per threshold `t`.
- Presign draws triples from the compatible's pool only (with the same `t`).
- The temporary "pick any one CaitSith domain's `t`" code in `EcdsaSignatureProvider::spawn_background_tasks` and `run_triple_generation_follower` is removed.
- Test coverage for the migration path and for per-domain isolation of generation and consumption.

gilcu3:
Closes #3292


crates/node/src/assets.rs
```diff
@@ -346,7 +345,17 @@ where
 {
     db: Arc<SecretDB>,
     col: DBCol,
-    domain_id: Option<DomainId>,
+    /// Byte prefix prepended to every key written under `col`. Empty `Vec` means
+    /// no prefix (i.e. the original layout where keys were just
+    /// `borsh(UniqueId)`).
+    prefix: Vec<u8>,
```
gilcu3, line 351: Given that we dont plan to use `DomainId` as the universal prefix, it seemed natural to use `Vec<u8>` instead. An alternative would be a proper Enum for this, but then we would need to guarantee some backward compatibility or make migrations.

crates/node/src/providers/ecdsa/triple.rs
gilcu3, line 1: Most code changes in this file are tests

gilcu3:
>crates/node/src/providers/ecdsa/triple.rs:83 — migrate_legacy_triples_to_v2 queues delete_all(DBCol::TripleV2) plus one put(DBCol::TripleV2, ...) per legacy row into a single WriteBatch, then commits at the end. With the documented buffer target of 1M triples per node and ~hundreds of bytes (post-JSON+AES-GCM-encryption) per row, this batch is on the order of hundreds of MB to >1 GB in memory before commit 

number of triples in prod is just 16k, @claude review it should be fine no? I will take care of the other non-blocking issues

crates/node/src/providers/ecdsa/triple.rs
```diff
@@ -53,6 +64,67 @@ impl Deref for TripleStorage {
...
+        // Both halves of a cait-sith paired triple are generated with the same
+        // participant set, so either one yields the correct `t`.
+        let threshold: u64 = triple.0 .1.participants.len().try_into()?;
+        let mut new_key = Vec::with_capacity(std::mem::size_of::<u64>() + legacy_key.len());
+        new_key.extend_from_slice(&threshold.to_be_bytes());
+        new_key.extend_from_slice(&legacy_key);
+        update.put(DBCol::TripleV2, &new_key, &value_ser);
+        migrated += 1;
+    }
+    update.commit()?;
```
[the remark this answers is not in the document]
gilcu3: for the current sizes of triples, 16K, should be fine.

crates/node/src/assets.rs
```diff
@@ -346,7 +345,17 @@ where
...
+    /// Optional companion column mirrored on every write/delete. Keys in
+    /// `legacy_col` are written WITHOUT the `prefix` (just `borsh(UniqueId)`).
+    /// Used during the per-`t` triple migration so a downgrade can still read
+    /// triples this binary wrote.
```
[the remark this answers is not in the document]
gilcu3: for this reason in the transition period we will enforce a single threshold, see https://github.com/near/mpc/issues/3304

gilcu3:
> ### Reviewed changes
> Copilot reviewed 12 out of 13 changed files in this pull request and generated 2 comments.

I am surprised you only produced 2 comments out of this @copilot :|

crates/node/src/assets/cleanup.rs
```diff
@@ -23,6 +23,7 @@ pub fn delete_stale_triples_and_presignatures(
     current_epoch_data: EpochData,
     my_participant_id: primitives::ParticipantId,
     ecdsa_domain_ds: Vec<DomainId>,
+    triple_thresholds: Vec<ReconstructionThreshold>,
```
netrome, line 26: Why is this passed separately instead of being part of the `EpochData` struct?

netrome: Started making sense of the epoch data struct and I saw a lot of legacy code path that we didn't remove last time we upgraded them, making this review a bit more complex than it needs to.

I created https://github.com/near/mpc/issues/3348 - it would be nice to do it as a preparatory refactor before this PR, but I'm okay with doing it after as well.

netrome (commented):
Started reviewing, but took some time to make sense of some stale legacy code paths. Will continue tomorrow.

crates/node/src/assets/cleanup.rs
gilcu3: In epoch data we currently don't have per domain data, and we had already the `ecdsa_domain_ds` parameter just above. So it seemed less intrusive to just add a new parameter to the function. Could be changes as part of https://github.com/near/mpc/issues/3348 as you mention, but I would prefer to get this merged first as it has been blocking other important issues for a few days now

netrome: Sure, yeah feels like both of these parameters would belong there - but yeah, not blocking.

crates/node/src/assets/cleanup.rs
```diff
@@ -43,25 +44,37 @@ pub fn delete_stale_triples_and_presignatures(
...
-            // cleanup presignatures:
+            for t in &triple_thresholds {
+                clean_db::<PairedTriple>(
+                    db,
+                    DBCol::TripleV2,
+                    &persitent_participants,
+                    my_participant_id,
+                    &t.inner().to_be_bytes(),
+                )?;
+            }
```
netrome, line 74: Shouldn't these cleanups be atomic? Or is it fine if the node crashes half-way because we'd just wipe the remaining triplets on startup again?

We still do the full cleanup in a single atomic update so I don't see why we wouldn't do it here as well.

crates/node/src/providers/ecdsa/triple.rs
```diff
@@ -53,6 +64,67 @@ impl Deref for TripleStorage {
...
+/// The whole rebuild commits in one atomic batch.
+///
+/// SAFETY: must be called before any [`TripleStorage`] is constructed for this
+/// DB, and exactly once per process. The unconditional [`DBCol::TripleV2`]
+/// wipe would race with a live `TripleStorage` (vending the same `UniqueId`
+/// to two consumers across the wipe) — so this is invoked from `run.rs` right
+/// after [`crate::db::SecretDB::new`], before the coordinator can construct
+/// any provider. Do not call it from `EcdsaSignatureProvider::new` or any
+/// other path that may run multiple times per process (e.g., across
+/// coordinator Running ↔ Resharing transitions).
```
netrome, line 92: Very helpful, and makes sense. Thanks!

crates/node/src/assets.rs
netrome: Makes sense.

netrome (dismissed):
Didn't go as deep as I wanted, but this looks good to me. One question about the epoch based cleanup function but not blocking.

anodar (dismissed):

crates/node/src/assets/cleanup.rs
gilcu3: addressed in [ecb07aa](https://github.com/near/mpc/pull/3299/commits/ecb07aae80b4565c69a939e011d3e5c545107269)

anodar (approved):

crates/node/src/assets/cleanup.rs
netrome: Thanks! 🙏

netrome (approved):
Thanks for updating 🙏

Issue #2825, opened 2026-04-10
replace ThresholdParametersVotes with generic votes struct


kevindeforth:
### Background

-

### User Story

-

### Acceptance Criteria

-

SimonRastikian:
I reworked the whole issue above and figured some really interesting tech-dept

PR #3311, opened 2026-05-21
docs(design): attestation verifier contract breakout

### LINKED ISSUE #3313: Write design doc: attestation verifier contract breakout
### Background

`mpc-contract` is at 1,459,158 bytes, against NEP-509's 1,490,000-byte hard limit. The `dcap-qvl` closure used for on-chain TDX quote verification accounts for ~310 KB of that and leaves no headroom for future state-migration changes. Breaking it out into a standalone verifier contract is the natural fix, but the shape of that split (Promise flow, crate layout, deployment, testing) needs to be agreed before implementation work starts.

This issue tracks writing the design doc only. Implementation will be tracked separately once the design is merged.

### Relationship to #3160

This work supersedes the earlier [#3160](https://github.com/near/mpc/pull/3160), which proposed a broader three-contract architecture unifying TEE verification across MPC, Defuse, Proximity, and other teams. That direction was deferred because it presumes shared lifecycle conventions (the launcher pattern) the other teams haven't aligned on yet. This issue narrows the scope to extracting just the stateless DCAP verification primitive — the one piece that benefits every team — and leaves policy in `mpc-contract`.

### Acceptance Criteria

- [ ] Design doc landed under `docs/design/` covering: motivation and WASM-size numbers, proposed crate layout, `submit_participant_info` Promise+callback flow, deployment ordering, and test plan.
- [ ] Doc reviewed and merged.

pbeza:
Closes: #3313

This PR is a narrower, more focused design than the earlier #3160. #3160 proposed a three-contract architecture (shared verifier + per-team policy contract + TEE-agnostic application contract) aimed at unifying TEE verification across MPC, Defuse, Proximity, and other teams. That direction was deferred because a shared policy contract presumes shared lifecycle conventions (the launcher pattern) that the other teams haven't aligned on yet.

This PR narrows the scope to the one piece that benefits every team — extracting the stateless DCAP verification primitive into a standalone `tee-verifier` contract — and leaves policy in `mpc-contract`.


docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,471 @@
...
+```
+
+Each `PendingAttestation` holds:
+
+- **The submitter's `Attestation::Dstack` payload** — the RTMR3 event log, app-compose, and report-data the post-DCAP checks consume.
+- **The submitter's TLS public key** — the callback hashes it with the submitter's account public key and compares to the quote's `report_data` field, proving the enclave produced the quote for this specific submitter.
+- **The attached deposit** — covers storage staking on success, must be refunded on failure. `env::attached_deposit()` is not visible from the callback receipt.
+- **`expires_on: BlockHeight`** — TTL for orphan recovery; see [§Handling failures](#handling-failures).
+
+Entries are removed by the callback regardless of outcome; the only way one outlives its callback is the out-of-gas case in [§Handling failures](#handling-failures), where the TTL guarantees the orphan is bounded.
```
barakeinav1: in the out of gas case, will the Entries be cleared?

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,471 @@
...
+
+Notably absent from `PendingAttestation`: the **allowed-image-hash list**. The callback re-reads it from contract state, so any governance vote that adds or removes an entry mid-Promise applies to verifications it overlaps. Snapshotting at request time would freeze each submission against stale policy — wrong default for a security control, where removing a compromised hash should take effect immediately.
```
barakeinav1: >**allowed-image-hash list*

I assume the same goes for allowed measurements and allowed launcher hashes as well

barakeinav1: > wrong default for a security control, where removing a compromised hash should take effect immediately

not critical, but better safe than sorry.
does it complicate the implementation?

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,471 @@
...
+
+```mermaid
+flowchart TD
+    MPCC[mpc-contract WASM]
+    EXT[Defuse / Proximity / other callers]
+    CLI[attestation-cli, tee-authority, mpc-node]
+
+    MPCA[mpc-attestation]
+    ATT[attestation]
+    VER[tee-verifier WASM]
```
barakeinav1, line 195: not clear from the graph that "tee-verifier WASM" implements the interface. (it look like it is a simple dependency)

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,471 @@
...
+2. RPC `view_account(candidate_account_id).contract` → `CodeHash(H_deployed)`.
+3. RPC `view_access_key_list(candidate_account_id)` → empty (locked).
+4. `H_source == H_deployed`.
+
+
+### Who chooses, and why NEP-591
+
+Active MPC participants vote, through `vote_tee_verifier_change` (see [§Voting on the trusted verifier](#voting-on-the-trusted-verifier)). External callers (Defuse, Proximity) run their own equivalent vote on their own contract — independent decisions, no shared governance surface.
+
+The reasons to prefer NEP-591 globals over a plain locked contract:
```
barakeinav1: not sure I follow the reasons for using a global contract over a locked contract.
seem like the 3 first bullets can be achieved with both.
only the Cross-shard code dedup seem like a difference. 
please correct me if I'm wrong about this.

There is also an additional price for using global contracts: 

https://github.com/near/NEPs/blob/master/neps/nep-0591.md#costs

Would be ~30 NEAR per contract. (instead of ~3 NEAR )

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,471 @@
...
+
+The contract is stateless. The wire DTOs (`QuoteBytes`, `Collateral`, `VerifiedReport`, `VerifierError`, and the nested report types) are field-for-field Borsh mirrors of the corresponding `dcap_qvl` types, defined in `tee-verifier-interface`.
+
+The method is named `verify_quote`, not `verify`: even on a single-method contract, an unqualified `verify` reads ambiguously in caller code (`tee_verifier.verify(...)` vs `tee_verifier.verify_quote(...)`), and leaves no room for future siblings (e.g. a batched `verify_quote_batch`).
```
barakeinav1: good idea

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,471 @@
...
+
+### Voting on the trusted verifier
+
+Voting reuses the contract's generic [`Votes<V>`](../../crates/contract/src/primitives/votes.rs) primitive — the same one [`ProviderVotes`](../../crates/contract/src/foreign_chain_rpc.rs) uses for the foreign-chain-RPC whitelist. `Votes<V>` already implements per-voter bookkeeping, `vote` / `remove_vote`, and `voters_for(proposal_hash)`; we just provide the proposal payload and a `ProposalHashEncoding` impl so it can be addressed by hash. Rolling a bespoke `PendingVerifierChange { … votes: HashSet }` struct would re-implement what's already in-tree and grow the contract WASM for no gain.
```
barakeinav1: do we use this for other voting (like node/launcher/measurements) ?

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,471 @@
...
+
+```rust
+/// Proposal payload. Two voters arrive at the same `ProposalHash` iff they
+/// borsh-serialize the same `(candidate_account_id, expected_code_hash)` —
+/// description is excluded from the hash so cosmetic differences don't fork
```
barakeinav1: >description is excluded from the hash so cosmetic differences don't fork

I didn't understand this sentence

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,471 @@
...
+    ///
+    /// When the threshold is reached, the contract writes
+    /// `self.tee_verifier_account_id = candidate_account_id`, emits an event
+    /// carrying `(candidate_account_id, expected_code_hash, description)`,
+    /// and clears the proposal.
+    pub fn vote_tee_verifier_change(
+        &mut self,
+        candidate_account_id: AccountId,
+        expected_code_hash: CryptoHash,
+        description: String,
```
barakeinav1: I didn't understand why we need expected_code_hash

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,471 @@
...
+`VERIFIER_GAS_TGAS` and `CALLBACK_GAS_TGAS` are placeholders until benchmarked against the PoC in [#3247](https://github.com/near/mpc/pull/3247). The verifier-side cost is dominated by ECDSA verifications and X.509-chain walking inside `dcap_qvl::verify::verify`; the callback-side cost is dominated by RTMR3 replay and the four post-DCAP checks. Both need measurement, not estimation.
+
+The contract gains the following state fields:
+
+```rust
+pub struct MpcContract {
+    // ... existing fields ...
+    pending_attestations: LookupMap<AccountId, PendingAttestation>,
+    tee_verifier_account_id: AccountId,
+    // pending_tee_verifier_changes is described in §Voting on the trusted verifier
```
barakeinav1: I thought of something, after upgrading to a new verifier, existing attestation can potentially become invalid, but we never verify existing attestations against the verifier. 

This is not something new, this is the behaviour of the current contract AFAIK. (but I'm not sure, need to check verify_tee)

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,384 @@
...
+        MPC-->>Op: error
+    end
+```
+
+Failure-mode invariants implied by the diagram:
+
+1. **Verifier panic.** `PromiseResult::Failed` reaches the callback. The pending entry is cleared, the deposit is refunded, an error is returned. Contract state is otherwise unchanged.
+2. **Post-DCAP check failure** (allowlist miss, RTMR3 mismatch, app-compose mismatch, binding mismatch). Pending entry cleared, deposit refunded, specific reason logged. No permanent state change.
+3. **Out-of-gas in the callback.** The callback panics; the pending entry stays. Caller can retry by submitting again — `AccountId`-keyed state means the resubmit overwrites the stale entry on next success.
+4. **Concurrent allowlist vote.** The callback re-reads the allowlist from state, so a governance vote landing between the Promise and the callback takes effect on the verification in flight. No request-time snapshot.
```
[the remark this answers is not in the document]
pbeza: Addressed in https://github.com/near/mpc/commit/78f90622 and https://github.com/near/mpc/commit/6ae0c647. The pending-entry lifecycle is now consistent: each `PendingAttestation` carries an `expires_on: BlockHeight` (TTL). A second submit from the same `AccountId` is rejected only while `block_height() < expires_on`; once the TTL elapses the entry is treated as stale and overwritten by the next submit. For accounts that never resubmit, an anyone-callable sweep `clean_pending_attestations(max_scan)` removes expired entries and refunds their deposits.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,471 @@
...
+                log!("verifier promise failed for {account_id}: {promise_err:?}");
+                return;
+            }
+            Ok(Ok(report)) => report,
+        };
+
+        // Post-DCAP checks operate on the verified report plus state held here.
+        // The allowlist is read fresh — governance votes mid-flight take effect.
+        if let Err(reason) = finish_verify(&pending, &verified_report, self.allowlist_fresh()) {
+            refund_deposit(&account_id, pending.attached_deposit);
```
barakeinav1: to whom do we return the deposit?

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,384 @@
...
+In this section we propose the on-chain interface for `tee-verifier`, the wire DTOs, and the changes to `mpc-contract`'s `submit_participant_info`.
+
+### The Verifier Contract
+
+The verifier exposes exactly one method:
+
+```rust
+#[near]
+impl TeeVerifier {
+    /// Verify a TDX / SGX quote against Intel collateral.
```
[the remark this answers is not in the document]
pbeza: Addressed in https://github.com/near/mpc/commit/78f90622 — the rustdoc on `verify_quote` now says "Verify a TDX quote against Intel collateral" (no SGX).

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,424 @@
...
+    ///
+    /// On verification failure, panics with the upstream error rendered as
+    /// a string. Callers observe this as `PromiseResult::Failed` in the
+    /// callback.
+    #[result_serializer(borsh)]
+    pub fn verify_quote(
+        &self,
+        #[serializer(borsh)] quote: QuoteBytes,
+        #[serializer(borsh)] collateral: Collateral,
+    ) -> VerifiedReport;
```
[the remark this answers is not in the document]
pbeza: Addressed in https://github.com/near/mpc/commit/78f90622. The verifier API now returns a structured `Result<VerifiedReport, VerifierError>` (Borsh-serialized) instead of panicking. `VerifierError` lives in `tee-verifier-interface` with one variant per `dcap_qvl::verify::verify` failure category. The callback in `mpc-contract` branches on the actual `Err` variant instead of getting an opaque `PromiseError::Failed`.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,384 @@
...
+                env::panic_str("dcap verification failed");
+            }
+        };
+
+        // Post-DCAP checks operate on the verified report plus state held here.
+        // The allowlist is read fresh — governance votes mid-flight take effect.
+        if let Err(reason) = finish_verify(&pending, &verified_report, self.allowlist_fresh()) {
+            refund_deposit(&account_id, pending.attached_deposit);
+            env::panic_str(&format!("post-DCAP check failed: {reason}"));
+        }
```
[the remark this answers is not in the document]
pbeza: Addressed in https://github.com/near/mpc/commit/78f90622. The callback no longer panics on verifier or post-DCAP failures. All failure branches do their state mutation, schedule the refund via `refund_deposit`, log a structured reason, and `return` normally — no `env::panic_str` on the failure paths. The rustdoc on `on_attestation_verified` explicitly explains why panicking would defeat both the pending-entry-cleanup and refund-Promise invariants. (Out-of-gas mid-callback still aborts the receipt, but that's covered by `clean_pending_attestations`.)

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,424 @@
...
+        let Some(pending) = self.pending_attestations.remove(&account_id) else {
+            env::panic_str("no pending attestation for this account");
+        };
+
+        let verified_report = match result {
+            Ok(report) => report,
+            Err(_) => {
+                refund_deposit(&account_id, pending.attached_deposit);
+                env::panic_str("dcap verification failed");
+            }
```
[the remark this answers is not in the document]
pbeza: Same fix as [the duplicate above](https://github.com/near/mpc/pull/3311#discussion_r3280765242): addressed in https://github.com/near/mpc/commit/78f90622 — the callback no longer panics on logical failures. `log!` + `return` on each failure branch, with `refund_deposit` scheduled before the return.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,424 @@
...
+
+The existing E2E setup (the `near-sandbox`-backed tests in `crates/e2e-tests`) gets one new deployment step: alongside `mpc-contract`, the test harness deploys either the real `tee-verifier` (for tests that want to exercise real `dcap-qvl` against a fixture quote) or the stub verifier (for everything else). No new test framework — same crate, one extra `deploy` call in the setup helper.
+
+Once the stub exists, `Attestation::Mock`'s role in tests is largely superseded: the stub covers skipping `dcap-qvl`, running on non-TDX machines, and exercising post-DCAP policy in isolation. The first iteration of this design keeps `Attestation::Mock`; a later iteration can remove it once the stub is the established path.
+
+[submit-participant-info]: https://github.com/near/mpc/blob/efe49230bb66854c55bba080e7610e42f9221506/crates/contract/src/lib.rs#L754-L782
+[launcher-pattern]: https://github.com/near/mpc/blob/efe49230bb66854c55bba080e7610e42f9221506/docs/tee-lifecycle.md#upgrade
+[slack-launcher-discussion]: https://nearone.slack.com/archives/C0B12RKBSAV/p1777897902903889
```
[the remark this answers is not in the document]
pbeza: Kept intentionally for now: there's no public artifact yet that summarizes the launcher-pattern alignment discussion this points at, and the link is meant to give internal readers a paper trail for why #3160 was deferred. Open to dropping it if the public-durability concern outweighs the internal-context one — happy to replace with a one-line inline summary.

docs/design/attestation-verifier-contract.md
pbeza: Good catch — this was originally a gap and is now addressed in https://github.com/near/mpc/commit/78f90622 (TTL recovery) and https://github.com/near/mpc/commit/6ae0c647 (`clean_pending_attestations` sweep). In the out-of-gas case the entry is *not* cleared by the callback (the receipt aborts), but it's recoverable two ways:

1. **Passive**: each `PendingAttestation` carries an `expires_on: BlockHeight` set to `env::block_height() + PENDING_ATTESTATION_TIMEOUT_BLOCKS`. Once the TTL elapses, the next `submit_participant_info` from the same `AccountId` overwrites the stale entry. Mirrors the timeout-on-read pattern `KeyEventInstance::cleanup_if_timed_out` already uses.
2. **Explicit**: `clean_pending_attestations(max_scan)` is anyone-callable, sweeps expired entries, and refunds their deposits. Mirrors the existing `clean_invalid_attestations` (same shape, same `max_scan` budget pattern).

See _§Handling failures_ item 4 and the `clean_pending_attestations` snippet later in §`mpc-contract::submit_participant_info`:

https://github.com/near/mpc/blob/89b14ffa917a87ff225a6d3d98d16d58b0037b10/docs/design/attestation-verifier-contract.md?plain=1#L111

https://github.com/near/mpc/blob/89b14ffa917a87ff225a6d3d98d16d58b0037b10/docs/design/attestation-verifier-contract.md?plain=1#L438-L452

docs/design/attestation-verifier-contract.md
pbeza: Addressed in https://github.com/near/mpc/commit/6ae0c647: the `description: String` field was dropped from the proposal entirely. The whole "cosmetic differences fork the vote" worry is gone since the proposal payload is now just `(candidate_account_id, expected_code_hash)`.

docs/design/attestation-verifier-contract.md
pbeza: Addressed in https://github.com/near/mpc/commit/6ae0c647 — the rationale is now spelled out in _§Voting on the trusted verifier_:

https://github.com/near/mpc/blob/89b14ffa917a87ff225a6d3d98d16d58b0037b10/docs/design/attestation-verifier-contract.md?plain=1#L240

The contract itself only ever consumes `candidate_account_id` from the payload (used as the Promise target); `expected_code_hash` is there purely to bind each voter's endorsement to a specific code snapshot.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,471 @@
...
+    MPC->>State: read fresh allowlist
+    MPC->>MPC: finish_verify (post-DCAP checks)
+
+    alt Verifier ok and post-DCAP ok
+        MPC->>State: store VerifiedAttestation
+        MPC->>State: remove PendingAttestation
+        MPC-->>Op: success
+    else Verifier failed or post-DCAP failed
+        MPC->>State: remove PendingAttestation
+        MPC->>Op: refund attached deposit
```
barakeinav1, line 102: > refund attached deposit
Is this something we do today?

docs/design/attestation-verifier-contract.md
pbeza: No, it's actually simpler than snapshotting. `finish_verify` in the callback already takes the live allowlist by reference from `mpc-contract` state:

https://github.com/near/mpc/blob/89b14ffa917a87ff225a6d3d98d16d58b0037b10/docs/design/attestation-verifier-contract.md?plain=1#L420-L426

The reason this works: the callback runs in `mpc-contract`'s own execution context, so it can read contract state directly; the only thing `PendingAttestation` needs to carry is data not derivable from current state (the submitter's `Attestation::Dstack` payload, their TLS key, the attached deposit, the TTL).

docs/design/attestation-verifier-contract.md
pbeza: Good question!

`Votes<V>` is currently used by [`ProviderVotes`](https://github.com/near/mpc/blob/main/crates/contract/src/foreign_chain_rpc.rs) for the foreign-chain-RPC whitelist — that's the precedent this design mirrors. Other voting flows go through different mechanisms:

- `propose_update` / `vote_update` (contract code upgrades): separate state in `update.rs`.
- Threshold-change votes: `crates/contract/src/primitives/threshold_votes.rs`, which carries a TODO ([#2825](https://github.com/near/mpc/issues/2825)) to migrate to `Votes<V>`.
- Node/launcher/measurement allowlist updates: go through the contract's existing `propose_*` / `vote_*` methods in `tee/proposal.rs` and `tee/measurements.rs`, not `Votes<V>` directly.

So the verifier-change vote would be the second user of `Votes<V>` after `ProviderVotes`. IIUC, in the long run we want to converge on using the `Votes<V>` struct more broadly. The only reason we haven’t done it yet for other voting mechanisms is that we’re blocked by contract size, which grew too much to adopt this approach everywhere.

docs/design/attestation-verifier-contract.md
barakeinav1: ok, so this is more a sanity check to make sure voter know what they are voting for.
and the reason we need account_id is since the contract can't compute it by itself?

pbeza: > ok, so this is more a sanity check to make sure voter know what they are voting for.

Exactly! I wasn’t 100% sure if we wanted this, since technically operators could just vote for the verifier contract’s account ID. But since it’s a critical component of the system, I thought it would be good to have an extra check that they know which contract hash they’re voting for.

We would be the ones telling them which hash is valid/trusted, but technically they could also compile the verifier contract themselves and check that the hash matches.

> and the reason we need account_id is since the contract can't compute it by itself?

Right — the contract can't derive the account from the `expected_code_hash` (same WASM can be deployed to many accounts). `account_id` is what the contract actually uses at runtime (the Promise target for `verify_quote`); `expected_code_hash` is the off-chain audit anchor that voters commit to. The contract needs both because they answer different questions: "where do we call?" vs "what code did you audit?".

docs/design/attestation-verifier-contract.md
pbeza: Good catch — done in https://github.com/near/mpc/commit/59d00979. All three (image hashes, launcher compose hashes, accepted measurements) are re-read fresh in the callback; updated the paragraph in _§Contract state changes_ to call out all three explicitly:

https://github.com/near/mpc/blob/59d00979ecce8cd4b5fcfaf4c55d3940ec3a56b2/docs/design/attestation-verifier-contract.md?plain=1#L130

docs/design/attestation-verifier-contract.md
pbeza: You're right that the graph is about Cargo `[dependencies]` edges only — mixing in an "implements" arrow would conflate two different relations. Kept the graph as-is and clarified the caption in https://github.com/near/mpc/commit/59d00979.

docs/design/attestation-verifier-contract.md
pbeza: > not sure I follow the reasons for using a global contract over a locked contract.
seem like the 3 first bullets can be achieved with both.
only the Cross-shard code dedup seem like a difference.
please correct me if I'm wrong about this.

Excellent point! I actually relaxed the requirements for the contract while you were reviewing this. I just hadn’t pushed it at the time, but it’s pushed now:

https://github.com/near/mpc/blob/59d00979ecce8cd4b5fcfaf4c55d3940ec3a56b2/docs/design/attestation-verifier-contract.md?plain=1#L197

Only the audit-side `view_account` win and cross-shard dedup are NEP-591-specific; the other properties (caller-side immutability, atomic rollback, multi-version coexistence) are equally achievable with a plain locked contract, which is why _§Requirements on the verifier account_ now says "regular contract or a NEP-591 global contract — both satisfy the requirements once locked".

> There is also an additional price for using global contracts:
>
> https://github.com/near/NEPs/blob/master/neps/nep-0591.md#costs
>
> Would be ~30 NEAR per contract. (instead of ~NEAR )

On the ~30 NEAR publication cost: agreed, and it only applies if the deployer actively picks the global option, so the doc as-is keeps the decision (and the cost) with whoever does the deploy.

docs/design/attestation-verifier-contract.md
pbeza: Confirmed your read — checked `re_verify` in `mpc-attestation/src/attestation.rs`:

https://github.com/near/mpc/blob/cf179467124203b0187ef0e80b429885b9a51627/crates/mpc-attestation/src/attestation.rs#L93-L134

and it only re-checks the four post-DCAP allowlist invariants (expiry, MPC image hash, launcher compose hash, measurements). It never re-runs `dcap_qvl::verify::verify`, so stored attestations carry their original DCAP verdict forward across verifier upgrades. This is intentional (re-running DCAP on every stored entry would undo the WASM-size win) and is the same behavior as today's contract, as you suspected. In practice the node's `periodic_attestation_submission` task (1-hour cadence) re-verifies against the new verifier naturally.

docs/design/attestation-verifier-contract.md
pbeza: The signer of the original `submit_participant_info` transaction. The chain:

1. At submit time, `let account_id = Self::assert_caller_is_signer();` binds `account_id` to the signer (not the predecessor — the comment a few lines above the snippet says "caller must be the signer of this transaction, not a relayer or proxy"). [`assert_caller_is_signer` at `crates/contract/src/lib.rs:2158`](https://github.com/near/mpc/blob/59d00979/crates/contract/src/lib.rs#L2158) returns `env::signer_account_id()` and panics if the predecessor differs.
2. That `account_id` gets captured into the callback Promise via `.on_attestation_verified(account_id)`.
3. The callback receives it as its `account_id` parameter and passes it to `refund_deposit(&account_id, pending.attached_deposit)`.

So on any of the three failure paths (verifier `Err`, infrastructure failure, post-DCAP fail), the original signer gets their attached deposit back. `clean_pending_attestations` also refunds to each entry's stored submitter `account_id`.

Tightened the wording in https://github.com/near/mpc/commit/6d72f69d.

docs/design/attestation-verifier-contract.md
pbeza: Yes, behaviorally the same as today — but today it's implicit, in the new design it has to be explicit. Today's [`submit_participant_info`](https://github.com/near/mpc/blob/6d72f69d6f3c35a0801a904bc25b4d0d8f969e1a/crates/contract/src/lib.rs#L749) verifies synchronously and returns `Err` on failure; the protocol rolls the receipt back, so the attached deposit is automatically refunded as part of the rollback. There's also an explicit excess-refund for the success path (line 833: `Promise::new(account_id).transfer(diff).detach()`) — anything attached above the storage cost goes back.

In the new design the verifier round-trip happens in a Promise + callback, so the callback runs in a *separate* receipt. A logical failure (verifier rejection, post-DCAP fail) in the callback doesn't roll back the original submit receipt's state — which means the pending entry sits in `pending_attestations` and the attached deposit is still in the contract account. So the callback has to do both things explicitly: `pending_attestations.remove(&account_id)` to clean up the entry, and `refund_deposit(&account_id, pending.attached_deposit)` to schedule the deposit refund to the original signer. Behaviorally equivalent to today's implicit refund, just done with an explicit Promise because the failure is now async.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,488 @@
...
+   | `main` baseline   | 1,459,158  | —                         |
+   | After this design | ~1,149,708 | **−309,450 (−21.2%)**     |
+
+   Sizes after this design are measured on the PoC branch in [#3247](https://github.com/near/mpc/pull/3247), which strips `dcap-qvl` out of `mpc-contract`'s dependency graph.
+
+2. **Non-reusable verification primitive.** Other NEAR teams (Proximity, Defuse, anyone building on Intel TDX) cannot call `dcap_qvl::verify` on-chain without re-linking the entire dependency tree into their own contract.
+
+## Design goal
+
+The primary goal is to bring the `mpc-contract` WASM safely under NEP-509's 1,490,000-byte limit by extracting `dcap_qvl::verify` into a standalone, stateless `tee-verifier` contract.
```
gilcu3: I believe the NEP-509 limit is luckily a bit higher than that, around 1570 KBs

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,488 @@
...
+flowchart LR
+    Op[Operator] --> MPC[mpc-contract WASM]
+    MPC -- "verify_quote(quote, collateral)" --> VER[tee-verifier WASM]
+    VER -- "VerifiedReport" --> MPC
+    VER --> DCAP[dcap-qvl library]
+```
+
+### Submission flow
+
+`mpc-contract`'s [`submit_participant_info`][submit-participant-info] becomes asynchronous for Dstack attestations. The method extracts the quote bytes and collateral from the submitted `Attestation::Dstack`, schedules a Promise to `tee-verifier::verify_quote`, and chains a private callback (`on_attestation_verified`) onto the Promise. The Promise yields control; the receipt executes in a later block; the callback runs after the verifier returns. The post-DCAP checks (RTMR3 replay, app-compose validation, measurement allowlist matching, report-data binding) all run in the callback against the `VerifiedReport` the verifier returns, and against state held by `mpc-contract`.
```
netrome: What's the point of highlighting "The Promise yields control"? Isn't this always the case for cross contract calls?

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,488 @@
...
+    participant Op as Operator
+    participant MPC as mpc-contract
+    participant State as State
+    participant Ver as tee-verifier
+    participant DCAP as dcap-qvl
+
+    Op->>MPC: submit_participant_info(Dstack, tls_pk)
+    MPC->>State: insert PendingAttestation
+    MPC->>Ver: Promise: verify_quote
+    Note over MPC: yield
```
netrome: What does `yield` mean in the context of this sequence diagram? I feel this causes more confusion than help, or I might be missing something here?

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,488 @@
...
+
+The voting flow lives on `mpc-contract`, not on the verifier itself. It reuses `mpc-contract`'s existing generic [`Votes<V>`](../../crates/contract/src/primitives/votes.rs) primitive.
+
+The proposal payload is the pair `(candidate_account_id, expected_code_hash)`. `candidate_account_id` is the address whose `verify_quote` method `mpc-contract` will invoke on every subsequent `submit_participant_info` call once the vote passes; that's all the contract actually consumes from the payload. `expected_code_hash` is included to make every voter explicitly commit to the hash they checked off-chain: without it, two voters could converge on the same `account_id` while disagreeing about what code that account runs. Both fields feed `ProposalHashEncoding`, so two voters submitting the same `account_id` with different hashes land in different vote buckets and neither reaches threshold on its own — that's how the contract enforces "everyone who voted yes endorsed the same code," without needing a separate validation step. When the winning bucket crosses threshold the contract clears *all* pending proposals for that `candidate_account_id` (including losing-hash buckets).
```
netrome, line 266: How does the contract verify the account runs the expected code hash? Do we assume the use of deterministic accounts? https://github.com/near/NEPs/blob/master/neps/nep-0616.md - though if we use that I'd assume it wouldn't be possible to disagree on the code that actually runs for the same account ID.

Or is this just as an extra precaution, the voters include the code hash they expect the account to run just to ensure they all agree on what code they think will be executed, but there's no mechanism for the contract to enforce this actually being the case?

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,488 @@
...
+
+### `mpc-contract::submit_participant_info`
+
+The method splits into two halves with a Promise between them — see [§Submission flow](#submission-flow) above for the architecture. The return type is [`PromiseOrValue<()>`](https://docs.rs/near-sdk/5.26.1/near_sdk/enum.PromiseOrValue.html), `near-sdk`'s "sometimes synchronous, sometimes a Promise chain" type: `Mock` attestations return `Value(())` immediately, and `Dstack` attestations return a `Promise` that resolves once `tee-verifier::verify_quote` returns and the callback finishes. Draft implementation:
+
+```rust
+impl MpcContract {
+    pub fn submit_participant_info(
+        &mut self,
+        attestation: Attestation,
```
netrome, line 357: Right so this remains just a DTO, and we never convert this to any types holding any dcap-qvl. So our main contract would not depend on any of our attestation crates.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,488 @@
...
+```
+
+## Crate layout
+
+Two new crates, plus an existing one that picks up a new dependency:
+
+- **`tee-verifier-interface`** (new). Wire DTOs only — `QuoteBytes`, `Collateral`, `VerifiedReport`, and the nested report / TCB-status types — as Borsh-serializable mirrors of the corresponding `dcap-qvl` types. No `dcap-qvl` dependency, no MPC-specific types. This is what every caller of the verifier links against.
+- **`tee-verifier`** (new). The verifier contract WASM. Exposes a single method, `verify_quote`, which wraps `dcap_qvl::verify::verify`.
+- **`attestation`** (existing). TDX domain types and the post-DCAP verification logic. Picks up `tee-verifier-interface` to re-export `Collateral` and `QuoteBytes` instead of duplicating them.
+- **`mpc-attestation`** (existing). MPC-specific framing on top of `attestation`: the `Attestation { Dstack, Mock }` enum, the `(tls_pk, account_pk)` binding, mock attestation verification.
```
netrome: Where would this be used? If this is on top of our attestation crate and used in the MPC contract, we'd drag in `dcap-qvl` as a transitive dependency afaik. So this is primarily intended for the MPC node?

netrome (dismissed):
Some questions and thoughts but this looks very good. Eager to get this implemented.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,488 @@
...
+    ATT --> TVI
+
+    EXT --> TVI
+    VER --> TVI
+    VER --> DCAP
+```
+
+## Governance and upgrades
+
+The verifier contract is stateless and has no admin methods or on-chain configuration. For security, every verifier instance is deployed to a locked account — a NEAR account with no full-access keys, so the protocol refuses any future redeploy. The deployed bytes are frozen for the lifetime of the account; there is no in-place upgrade path. Changing the verifier means voting in a different, separately-deployed instance at a new locked account. The only governance decision is on `mpc-contract`, choosing which instance to trust through a vote of active MPC participants. External callers (Defuse, Proximity) run their own equivalent vote on their own contract.
```
barakeinav1, line 212: Worth being explicit about the security trade-off here — what happens to attestations validated by the previous verifier isn't currently spelled out. Can you add this to the doc. Suggested wording:

> **Security trade-off — stored attestations keep their old verdict on verifier change.** The contract does not re-run DCAP under the new verifier. Acceptable because attestations expire after 7 days (the node's next resubmission re-verifies under the new verifier) and `dcap-qvl` updates are rare.

barakeinav1 (dismissed):
thanks for address my comments,
one comment left,
looks good.

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,488 @@
...
+    Ver-->>MPC: VerifiedReport
+    MPC->>MPC: read allowlist (sees H)
+    MPC->>MPC: finish_verify against fresh allowlist
+```
+
+## Crate layout
+
+Two new crates, plus an existing one that picks up a new dependency:
+
+- **`tee-verifier-interface`** (new). Wire DTOs only — `QuoteBytes`, `Collateral`, `VerifiedReport`, and the nested report / TCB-status types — as Borsh-serializable mirrors of the corresponding `dcap-qvl` types. No `dcap-qvl` dependency, no MPC-specific types. This is what every caller of the verifier links against.
```
gilcu3, line 180: I thought we had exactly this somewhere, but maybe we removed it and use the upstream dstack types instead. Worth checking though

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,488 @@
...
+            pending.tls_pk.clone(),
+            VerifiedAttestation::from((pending, verified_report)),
+        );
+    }
+}
+```
+
+`VERIFIER_GAS_TGAS` and `CALLBACK_GAS_TGAS` are placeholders until benchmarked. The verifier-side cost is dominated by ECDSA verifications and X.509-chain walking inside `dcap_qvl::verify::verify`; the callback-side cost is dominated by RTMR3 replay and the four post-DCAP checks.
+
+For accounts that never resubmit after an out-of-gas callback, an anyone-callable sweep recovers the stranded deposit and the storage. It mirrors the existing [`clean_invalid_attestations`](https://github.com/near/mpc/blob/cf179467124203b0187ef0e80b429885b9a51627/crates/contract/src/lib.rs#L1636) for `stored_attestations` — same shape, same `max_scan` budget pattern:
```
gilcu3: isn't it better to have a timeout, similar for what we have for pending signatures? (and this timeout would be relatively short, given that it only depends on the blockchain, and not external computation)

docs/design/attestation-verifier-contract.md
```diff
@@ -0,0 +1,488 @@
...
+```rust
+pub struct MpcContract {
+    // ... existing fields, including tee_verifier_account_id and
+    // tee_verifier_votes from §Voting on the trusted verifier ...
+    pending_attestations: LookupMap<AccountId, PendingAttestation>,
+}
+
+pub struct PendingAttestation {
+    pub dstack: DstackAttestation,
+    pub tls_pk: PublicKey,
```
gilcu3: better make this the appropriate `Ed25519PublicKey` type

gilcu3 (dismissed):
Thank you! Added some comments, overall LGTM

docs/design/attestation-verifier-contract.md
pbeza: Checked `origin/main` — half right:

- For the inputs, yeah we kind of have it: `attestation::collateral::Collateral` is a newtype around `dcap_qvl::QuoteCollateralV3`:
   https://github.com/near/mpc/blob/5e47bfe93b398cb2343681fa2c0f2691d02c7285/crates/attestation/src/collateral.rs#L16-L23
   and `QuoteBytes` is just `Vec<u8>` with Borsh — that one's actually standalone:

   https://github.com/near/mpc/blob/5e47bfe93b398cb2343681fa2c0f2691d02c7285/crates/attestation/src/quote.rs#L6-L22

   But `Collateral` still drags in `dcap-qvl` because the inner type is from upstream.

- For the verifier outputs (`VerifiedReport`, `TDReport10`, `Report`, `EnclaveReport`, `TcbStatus`) there's nothing — `attestation::attestation` and `attestation::measurements` use them straight from `dcap_qvl::...`.

So `tee-verifier-interface` ends up doing two things: turning the `Collateral` wrapper into a real standalone mirror, and adding the `VerifiedReport` family that doesn't exist yet.

docs/design/attestation-verifier-contract.md
pbeza: > How does the contract verify the account runs the expected code hash? Do we assume the use of deterministic accounts?

The contract doesn't verify it — there’s no on-chain mechanism for that, as far as I know. See:

https://github.com/near/mpc/blob/6d72f69d6f3c35a0801a904bc25b4d0d8f969e1a/docs/design/attestation-verifier-contract.md?plain=1#L199-L210

> Do we assume the use of deterministic accounts? https://github.com/near/NEPs/blob/master/neps/nep-0616.md - though if we use that I'd assume it wouldn't be possible to disagree on the code that actually runs for the same account ID.

No. NEP-616 requires the code to be deployed as a NEP-591 global contract, and the current design doesn’t require that.

In the future, we could consider moving the verifier deployment to NEP-591 global + a NEP-616 deterministic account, and then drop `expected_code_hash` from the proposal entirely, since the `0s…` account ID would already cryptographically commit to the code.

Should be a fairly small change if we ever see a benefit and are OK with the extra cost of deploying a global contract.

> Or is this just as an extra precaution, the voters include the code hash they expect the account to run just to ensure they all agree on what code they think will be executed, but there's no mechanism for the contract to enforce this actually being the case?

Exactly. `expected_code_hash` is a voter commitment with no on-chain enforcement.

docs/design/attestation-verifier-contract.md
pbeza: Yeah, exactly. `tee-verifier` only depends on `tee-verifier-interface` for the DTOs, plus `dcap-qvl` itself. It never touches `attestation` or `mpc-attestation`.

The `quote` + `collateral` come in as DTOs, get handed straight to `dcap_qvl::verify::verify`, and the resulting `VerifiedReport` goes back out as a DTO.

That’s what keeps the verifier MPC-agnostic: Defuse/Proximity can call it without inheriting any of our domain types.

docs/design/attestation-verifier-contract.md
pbeza: https://github.com/near/mpc/pull/3311/commits/d97d3acd

docs/design/attestation-verifier-contract.md
pbeza: https://github.com/near/mpc/pull/3311/commits/7e69f10d

docs/design/attestation-verifier-contract.md
pbeza: https://github.com/near/mpc/pull/3311/commits/af761ee3

docs/design/attestation-verifier-contract.md
pbeza: Yeah, feels redundant. Fixed: https://github.com/near/mpc/pull/3311/commits/41604163. ✅

docs/design/attestation-verifier-contract.md
pbeza: Yeah, feels redundant. Fixed: https://github.com/near/mpc/pull/3311/commits/bf550e41. ✅

docs/design/attestation-verifier-contract.md
pbeza: `mpc-attestation` would stay where it is today — used by both `mpc-contract` and `mpc-node`.

It doesn’t pull in `dcap-qvl` itself; on `main`, there’s no `dcap-qvl` in its `[dependencies]`. The transitive edge is coming through `attestation`, so that’s the bit we’d clean up: replace the `dcap_qvl::*` stuff there — collateral re-export, post-DCAP helper signatures, `Attestation::verify`, etc. — with DTOs from `tee-verifier-interface`.

Once that’s done, `attestation` drops `dcap-qvl`, and the WASM stops pulling it in as well.

I expanded on it here: https://github.com/near/mpc/pull/3311/changes/57b29935f94eb792a8269c3eeab8c352f0aa547c.

pbeza: https://github.com/near/mpc/pull/3311/commits/57b29935

docs/design/attestation-verifier-contract.md
pbeza: Great suggestion! 🚀 I wasn't aware of `env::promise_yield_create` / `env::promise_yield_resume`. Pivoted to yield-resume in https://github.com/near/mpc/pull/3311/commits/4543a130267d050706cc8bf117835fddced9a795: `pending_attestations` now uses the same primitive as `sign` / CKD / verify-foreign-tx, `expires_on` and `clean_pending_attestations` are gone, and the runtime owns the ~200-block timeout.

pbeza: Restructured in https://github.com/near/mpc/pull/3311/commits/bed21bb1b8f04cd9873418bc957ccce825d6cca3 to minimize the risk of `on_attestation_verified` running out of gas and orphaning the entry: post-DCAP checks, `stored_attestations` insert, refund, and pending-entry removal moved into `resolve_verification` (the `.then` bridge). The yield-callback shrinks to a sign-shaped body — return value on resume, refund + remove on timeout — small enough that OOG is implausible at any reasonable gas setting, the same argument the sign-request callback already relies on.

@gilcu3 pls re-review when you get a chance.

gilcu3 (approved):
Thank you!

barakeinav1 (approved):

PR #3315, opened 2026-05-21
chore: remove legacy pending requests after 3.10

### LINKED ISSUE #3279: Drop `legacy_pending_requests` from contract state
### Background

`MpcContract::legacy_pending_requests` was added in the 3.9.1 → 3.10 migration as a read-through bridge so requests yielded against the v3.9.1 contract could still be answered or timed out after the upgrade. See the doc on `LegacyPendingRequests` in `crates/contract/src/pending_requests.rs:59-72`:

> They should be dropped (along with the now-unused storage keys) in the next upgrade after this has been released — same lifecycle as the V2→V3 cleanup in #2940.

Now that 3.10 is the production binary, the legacy window has closed; the field and its fallback paths can be removed in the next upgrade.

### Acceptance Criteria

- [ ] Remove the `legacy_pending_requests` field from `crate::MpcContract` (and the 3 init points in `new`, `init_running`, and the test helper)
- [ ] Drop the `legacy_map` parameter from `pending_requests::{resolve_yields_for, pop_one_yield_for}` and inline `pop_oldest_pending_yield` (no longer used externally)
- [ ] Delete the `LegacyPendingRequests` struct
- [ ] Remove the legacy-fallback branches in `get_pending_request`, `get_pending_ckd_request`, `get_pending_verify_foreign_tx_request`, and the three `return_*_and_clean_state_on_success` timeout callbacks
- [ ] Rename the 3 legacy storage keys (`PendingSignatureRequestsV3`, `PendingCKDRequestsV2`, `PendingVerifyForeignTxRequests`) to `_Deprecated*` to preserve `BorshStorageKey` discriminants
- [ ] Drop the 3 legacy-specific tests: `respond__should_resolve_legacy_yield`, `get_pending_request__should_fall_back_to_legacy_map`, `return_signature_and_clean_state_on_success__should_remove_legacy_entry_on_timeout`
- [ ] In the next `v3_XX_state.rs` migration scaffold, keep `legacy_pending_requests: LegacyPendingRequests` inlined so storage from the 3.10 contract still deserializes; drop it in the `From` impl

gilcu3:
Closes #3279


crates/contract/src/pending_requests.rs
```diff
@@ -177,21 +87,26 @@ where
...
-/// oldest-first FIFO timeout semantics across the legacy window.
-pub(crate) fn pop_one_yield_for<K>(
-    new_map: &mut LookupMap<K, Vec<YieldIndex>>,
-    legacy_map: &mut LookupMap<K, YieldIndex>,
-    request: &K,
-) where
+/// Yields are removed in FIFO order because they were appended in submission order
+/// and time out in that same order — so the timing-out yield is always the head.
+/// If the queue empties (or was already empty), the map entry itself is removed.
+pub(crate) fn pop_oldest_pending_yield<K>(new_map: &mut LookupMap<K, Vec<YieldIndex>>, request: &K)
```
netrome: `new_map` is not a meaningful name without the presence of `old_map`. Can you rename this to just `map`? Or even better, a meaningful name like `requests`?

netrome (changes requested):
Minor naming bug (yes I consider this a blocker since the name is technically incorrect now - sue me 😅)

netrome (approved):
Thanks for updating 🙏

anodar (approved):

Issue #3169, opened 2026-05-07
Carry per-domain ReconstructionThreshold in resharing votes


gilcu3:
### Background
`DomainConfig.reconstruction_threshold` is set when a domain is created (via `vote_add_domains`) and is currently immutable through resharing. Per the design (`docs/design/domain-separation.md` §6.4), per-domain thresholds should be changed via resharing — there is no separate vote function. To get there, `vote_new_parameters` (or its successor) needs to accept a per-domain threshold for each existing domain, validate each one against the proposed participant count, and apply the new values to the `DomainRegistry` on resharing success.

This unblocks relaxing the "no node enforcement of per-domain `t`" constraint: once resharing votes carry per-domain thresholds and the node honors them (#3164), the per-domain proposal-validation guard in `process_new_parameters_proposal` can be re-enabled.

### Acceptance Criteria
- Resharing proposal carries one `ReconstructionThreshold` per existing domain.
- `ThresholdParametersVotes` keying includes the per-domain thresholds so distinct proposals don't collapse into a single tally.
- Successful resharing updates each domain's stored `reconstruction_threshold`.
- The per-domain validation in `process_new_parameters_proposal` is re-enabled.

PR #3317, opened 2026-05-21
feat: lock all CaitSith domains to the same threshold

### LINKED ISSUE #3304: Lock CaitSith threshold during 3.11 transition
### Background

#3292 partitions ECDSA triples per `t` in `DBCol::TripleV2` while dual-writing to the legacy `DBCol::Triple` (unprefixed key) for downgrade safety until #3298 drops the legacy column. If, during this transition window, a single node ever runs with two `TripleStorage`s — i.e. two CaitSith domains whose `reconstruction_threshold` values differ — their dual-writes to the
unprefixed legacy mirror can collide and silently overwrite each other. The consequence is data loss rather than nonce reuse, but it's a real liveness hazard.

The cleanest way to keep that scenario unreachable in 3.11 is for the contract to refuse any operation that would introduce a second CaitSith threshold on the network. Once #3298 lands and the legacy column is gone,
the restriction can be relaxed.

### Acceptance Criteria

- `vote_add_domains` rejects a CaitSith `DomainConfig` whose `reconstruction_threshold` differs from the network's current CaitSith threshold (i.e. from any already-registered CaitSith domain.
- Resharing votes are rejected if they would allow CaitSith domain with different `reconstruction_threshold`
- Comment / inline doc on the new check pins the rationale and references #3298 as the trigger for removing it.

gilcu3:
Closes #3304

### DIFF
--- a/crates/contract/src/state/resharing.rs
+++ b/crates/contract/src/state/resharing.rs
@@ -206,7 +206,7 @@ pub mod tests {
         state::test_utils::gen_resharing_state,
     };
     use near_account_id::AccountId;
-    use near_mpc_contract_interface::types::DomainId;
+    use near_mpc_contract_interface::types::{DomainId, Protocol};
     use rstest::rstest;
     use std::collections::BTreeSet;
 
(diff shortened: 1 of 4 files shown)


crates/contract/src/state/resharing.rs
```diff
@@ -465,4 +465,76 @@ pub mod tests {
...
+    /// Pins the 3.11 CaitSith threshold invariant across the resharing path:
+    /// after a reshare, all CaitSith domains must share a single
+    /// `reconstruction_threshold`. They may move in lockstep, but must never
+    /// diverge from one another. Today this holds trivially because the
+    /// proposal carries only `ThresholdParameters` and leaves per-domain `t`
+    /// untouched, but once #3169 carries per-domain `t`s through resharing the
+    /// contract will need to keep them uniform — and this test is the tripwire.
+    #[test]
+    #[expect(non_snake_case)]
+    fn resharing__should_keep_all_caitsith_domains_at_one_reconstruction_threshold() {
```
netrome: I'm not following what we're testing here. Even if we don't enforce the thresholds to be the same, this test would still pass right?

netrome (changes requested):
The resharing test seems like LLM slop, or am I missing something?

crates/contract/src/state/resharing.rs
gilcu3: this test is meant to fail if the resharing function is modified in an incorrect way, it is not doing much now but that was the objective

netrome: Fair enough, had a call about this. I'd still prefer not to make too much assumptions about future code but it's a decent reminder at least to test that path after #3169 lands - so I won't block on this.

netrome (dismissed):
Thanks for discussing the test case, approved.

netrome (approved):

crates/contract/src/state/resharing.rs
gilcu3: ended up deleting in [c064ad9](https://github.com/near/mpc/pull/3317/commits/c064ad9db18668db5bfc23387a3ed515f4995c2c) to avoid confusion

anodar (approved):

Issue #3316, opened 2026-05-21
share one `RecentBlocksTracker` instance across all queues


kevindeforth:
### Background

We currently have one instance for each queue (signatures, ckd, foreign tx). That's wasteful, they all do the same changes. We should have a single instance, shared between all queues.

### User Story

As a node operaor, I like when my node is not wasting resources.

### Acceptance Criteria

Share one `RecentBlocksTracker` across all queues.

Issue #2680, opened 2026-04-01
RecentBlocksTracker usage cleanup


kevindeforth:
### Background

The purpose of `RecentBlocksTracker` is to ensure performance and reliability targets of signature computations. Specifically, it handles forks in the blockchain and makes estimates which blocks (read: signature requests) become final or not. Based on that estimate, signature requests are either selected for computation, put on hold for a short while, or are completely dropped if already timed out (or if they were on a fork that has not finalized).

Now, there are a few issues with the current set-up:
1. We currently have three separate instances of `RecentBlocksTracker` (one for each of our queues). We don't have benchmarks, but it is reasonable to suspect that a significant amount of compute time is spent on the same thing between these three instances (ordering of blocks and handling of forks).
2. The struct clearly states that it expect blocks to arrive in [partial order](https://github.com/near/mpc/blob/82d384f71138dddde0a6571d4f2a1a395aeb2502/crates/node/src/requests/recent_blocks_tracker.rs#L19). Our current code does not give any such guarantees, because we allow [concurrent processing](https://github.com/near/mpc/blob/eaed70b3b21b41699cefec1ac6425b05d91f47e3/crates/node/src/indexer/handler.rs#L139) of blocks, which makes the order subject to a race condition.
3. We already have a known race condition in our code related to block finality.  We may attempt to submit a signature response _before_ the block in which the signature was requested is finalized. We can observe this in production, it's the random "signaure timed out" error messages we see (I think we should also have an issue for this).

With https://github.com/near/mpc/pull/2625, we introduced the `BlockEventSubscriber` API that was agreed to during the design phase in https://github.com/near/mpc/pull/2103. One of the main reasons why we chose to collect block events on a per-block basis, was to make integration with the mpc node easier.

Now, after investigating the `RecentBlocksTracker` struct, it becomes clear that there is a nicer way to solve this.

1. The `RecentBlocksTracker` should be refactored, such that it is only concerned with keeping track of the blocks and deciding what blocks are considered optimistic, final or dropped. This struct should be part of the chain gateway crate. The MPC node should not really be concerned about forks in the blockchain.
2. The `BlockEventSubscriber` should directly return a stream for the block event in question.
3. Each block event should hold a pointer to the block it was observed in. That pointer then allows to quickly check the finality status of the event itself.

These things combined, will solves issues 1, 2, and 3 identified above.

### User Story

As a dev, I like separation of concerns an I like performance.

### Acceptance Criteria

Implement the plan outlined above.
[...]

Issue #3318, opened 2026-05-21
Track block time to measure signature response time


kevindeforth:
### Background

Each queue has it's own internal `clock`, used to measure wall-time between first seeing the request and when the response is indexed on-chain.

This measure can be useful and is currently used on-chain.

However, from a user-perspective, the timestamps of the blocks that hold the request and the response respectively, are more meaningful.

### User Story

As a node operator, I like meaningful metrics.

### Acceptance Criteria

Add a histogram metric for tracking on-chain response times.

### Resources & Additional Notes

Related to #3229 

Please take note of this https://github.com/near/mpc/pull/3319#discussion_r3286519843

Issue #3229, opened 2026-05-14
Establish meaningful metrics


kevindeforth:
### Background

In #3217, we outline the shortcomings of one of our current metrics used to tracking signature failures. The issue proposes to introduce a new, hopefully more reliable metric, which was merged in #3211.
As mentioned in the issue, the new metric is not ideal. It does not discriminate between the varying types of requests and it may still incur false positives due to counting timeouts happening on dead forks.

We need much more robust metrics for monitoring our production system. Specifically, for each type of request (ckd / foreign chain / ecdsa signature / eddsa signature), we want to:

- count the number of requests;
- count the number of requests that finalized;
- count the number of requests that ended up not getting included;
- measure the latency between receiving the request and submitting the response on-chain;
- count the number of finalized requests that ended up timing out;

With the changes in #2680 and the introduction of the Chain gateway crate #2587, establishing these metrics reliably _shouldn't be too hard anymore™._

Note that this is a larger issue that can be split into multiple sub-issues. Each metric will probably be it's own issue. Additionally, we should integrate the new metrics with the grafana dashboard and construct alerts from them.

### User Story

As a node operator, I like when my system has meaningful metrics that allow me to assess the state of it.

### Acceptance Criteria

This issue can be considered closed once all of the metrics outlined above have been implemented, tested, included in the grafana dashboard and hooked up with new alerts.
It is advisable to split this task into sub-issues.

- [ ] have a metric for number of dropped blocks https://github.com/near/mpc/pull/3616#issuecomment-4739357656

PR #3319, opened 2026-05-21
refactor: `RecentBlocksTracker` does no longer track block contents

### LINKED ISSUE #3070: `PendingRequests` queue should keep track of block content

kevindeforth:
Finally, we can resolve #3070!

This is the main refactor we wanted to achieve and have been working towards with the previous 9 PRs, outlined in https://github.com/near/mpc/issues/2680.

With this refactor, block content is stored on the queue side, no longer in the `RecentBlocksTracker`. This allows us to:
- share one `RecentBlocksTracker` instance between all queues (follow-up PR)
- move `RecentBlocksTracker` to the chain-gateway crate.

Follow-up issues:
#3320: test refactor
#3316: share one recent blocks tracker instance between all queues.


crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -180,13 +179,19 @@ impl AtomicBlockStatus {
...
+    /// test builds for assertion convenience — production paths read the counter that
+    /// `add_block` increments instead. Carrying the list this way avoids any production
+    /// allocation overhead while still letting tests reconstruct the finalized-name
+    /// string after pruning may have dropped strong refs to ancestors.
+    // TODO(#3320): drop this cfg(test) field. The test helper should observe finality by
+    // holding a strong `Arc<AtomicBlockStatus>` (upgraded from `block_ref` at `add_block`
+    // time), which survives even after `prune_old_blocks` drops the `BlockNode`.
+    // https://github.com/near/mpc/issues/3320
+    #[cfg(test)]
+    pub new_final_blocks: Vec<(BlockHeight, CryptoHash)>,
```
kevindeforth, line 194: This is to reduce the diff size. Could have changed the tests in this PR, but there is no harm in deferring it to a follow-up #3320

crates/node/src/requests/queue.rs
```diff
@@ -907,43 +990,86 @@ mod tests {
         let to_attempt3 = pending_requests.get_requests_to_attempt();
         assert_eq!(to_attempt3.len(), 1);
         assert_eq!(to_attempt3[0].request.id, req2.id);
+    }
 
-        // When: this attempt submits a response, but it is not yet recorded on the blockchain.
-        to_attempt3[0]
+    #[test_log::test]
+    #[expect(non_snake_case)]
+    fn test_pending_requests__should_wait_for_response_to_finalize() {
```
kevindeforth, line 1005: This is a behavior change compared to what we had before. It is deliberate. 

if we have a response sitting on the canonical chain, we wait for that response to finalize or be demoted to a non-canonical chain before attempting agani.

crates/node/src/requests/metrics.rs
```diff
@@ -68,7 +72,17 @@ pub static MPC_PENDING_VERIFY_FOREIGN_TXS_QUEUE_MATCHING_RESPONSES_INDEXED_TOTAL
...
+    )
+    .unwrap()
+});
+pub static MPC_PENDING_VERIFY_FOREIGN_TXS_QUEUE_FINALIZED_RESPONSES_INDEXED_TOTAL: LazyLock<
+    prometheus::IntCounter,
+> = LazyLock::new(|| {
+    prometheus::register_int_counter!(
+        "mpc_pending_verify_foreign_txs_queue_finalized_responses_indexed_total",
+        "Number of verify foreign tx responses that have been finalized,
+                 indexed by the pending verify foreign tx queue."
```
kevindeforth, line 85: Broken window fix: we were missing this metric.

kevindeforth:
@claude Good blocker, addressed. The saturating sub is addressed too.

As for the other nits:
> * `crates/node/src/requests/recent_blocks_tracker.rs:314` — taking `&'static prometheus::IntCounter` as a constructor argument bakes a global metric into otherwise pure tree-bookkeeping logic and complicates testing (the test helper at `queue.rs:1121-1127` papers over this by piping production `MPC_PENDING_CKDS_QUEUE_FINALIZED_BLOCKS_INDEXED` in as a sink, which couples test-runs that share process state). The TODO([share one `RecentBlocksTracker` instance across all queues #3316](https://github.com/near/mpc/issues/3316)) acknowledges this — fine to defer.

Yeah, this is very ugly, but also not really an issue. Will be removed soon.

> so a stuck-but-canonical response can keep a request in the queue indefinitely if the chain never advances finality past it.


Do you expect this to happen? I agree we have some additional risk exposure here, because we don't know to what degree the indexer sanitizes blocks before passing them to us. Like, this could maybe be an attack vector, but the attacker would need to produce blocks that are way ahead in height and somehow be able to have them not expire. I think this is pretty constructed.

> * `crates/node/src/requests/recent_blocks_tracker.rs:182-195` — `#[cfg(test)]` field on a public production struct is unusual; TODO([Remove new_final_blocks from AddBlockResult #3320](https://github.com/near/mpc/issues/3320)) already tracks removal.

Yeah, also very ugly, but what can you do?

kevindeforth:
> I re-traced the IndexedRespondTxs::status() early-return logic at crates/node/src/requests/queue.rs:104-124 for the multi-respond case (e.g., the same request indexed in two distinct blocks). The early return PendingOptimistic on a canonical-but-not-final entry would, in isolation, mask a Final entry that appears later in the vec — but because notify_new_block pushes in indexer/height order and a Final block must be at a lower or equal height than any still-canonical-only block, the Final entry is always at an equal or earlier index. Correctness holds, but it's an invariant that's not documented or tested. A short comment near the is_canonical() branch noting this dependence on insertion-order would harden the code against a future refactor that changes how/when entries are added.

The invariant that makes this work is actually a different one: we can't have, simultaneously, a "final" and a different "OptimisticAndCanonical" response transaction.
Reason for this is that we track calls to `MpcContract::return_signature_and_clean_state_on_success`. This is a method that can only be called by the contract and gets called _exactly once_ per signature.
BFT rules prohibit us from having a final transaction and, on the same chain, an optimistic transaction for the same response.

But, I will add a comment. Have some other comments to fix too.

crates/e2e-tests/tests/web_endpoints.rs
```diff
@@ -112,7 +112,10 @@ async fn test_web_endpoints() {
             &client,
             i,
             &format!("http://{web_addr}/debug/blocks"),
-            &["Recent blocks:", "reqs:"],
+            // `/debug/blocks` only renders the block-tree structure now; the per-block
+            // "reqs:" / "completed:" lines were dropped together with BufferedBlockData,
+            // since the queue no longer needs the reverse block→requests index.
+            &["Recent blocks:"],
```
gilcu3, line 115: I didnt even know we had this endpoint. Were we using it for anything?

crates/node/src/requests/debug.rs
```diff
@@ -10,7 +10,8 @@ use std::sync::{Arc, Mutex};
 
 const NUM_COMPLETED_REQUESTS_TO_KEEP: usize = 100;
 
-/// A completed request, for exporting to /debug/requests.
+/// A completed request, kept for surfacing on the queue's debug
+/// endpoints (`/debug/signatures`, `/debug/ckds`).
```
gilcu3, line 14: what do you think about merging these two endpoints? Not sure if it makes sense to have signatures and ckd separate at this level. And we probably still need to add foreign chain tx

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -231,7 +236,6 @@ impl BlockNode {
         indents: &mut Vec<u8>,
         final_head: Option<CryptoHash>,
         canonical_head: Option<CryptoHash>,
-        content_printer: &impl Fn(&CryptoHash, &mut std::fmt::Formatter<'_>) -> std::fmt::Result,
```
gilcu3, line 234: this change is related to the change in the debug endpoint, right?

crates/node/src/requests/queue.rs
```diff
@@ -79,57 +77,84 @@ pub struct PendingRequests<RequestType: Request, ChainRespondArgsType: ChainResp
...
+    /// Weak ref to the block's atomic status; lets us observe canonical/final changes the
+    /// tracker makes after we recorded the response.
+    block_status: Weak<AtomicBlockStatus>,
+    /// Wall-clock time at which our node observed the block containing the response.
+    received_at: near_time::Instant,
+    // TODO(#3318): We could share the `block_height` throug the same weak pointer as
+    // `block_status`, however, that's a larger refactor and this change will only truly make sense
+    // once we start improving the metrics with #3318, so we defer it to later and accept to hold a
+    // u64 for each response.
+    block_height: u64,
```
gilcu3: is this really meant to be solved/changed in #3318? It seems it actually belongs to some other issue (probably not yet written) for which #3318 is a premise

crates/node/src/requests/queue.rs
```diff
@@ -79,57 +77,84 @@ pub struct PendingRequests<RequestType: Request, ChainRespondArgsType: ChainResp
...
-        if i < hashes.len() - 1 {
-            result.push_str(", ");
-        }
-    }
-    if hashes.len() > 3 {
-        result.push_str("...");
+impl IndexedRespondTxs {
+    fn new() -> Self {
+        Self(Vec::new())
     }
```
gilcu3: maybe we could use some derive with [Constructor](https://docs.rs/derive_more/latest/derive_more/derive.Constructor.html) for this?

crates/e2e-tests/tests/web_endpoints.rs
gilcu3, line 115: this explanation is fine, but only for a comment in the PR, not really relevant to have it in the source code IMO

crates/node/src/requests/queue.rs
```diff
@@ -79,57 +77,84 @@ pub struct PendingRequests<RequestType: Request, ChainRespondArgsType: ChainResp
...
-    requests: Vec<RequestId>,
-    completed_requests: Vec<RequestId>,
-    timestamp_received: near_time::Instant,
+struct IndexedRespondTx {
+    /// Weak ref to the block's atomic status; lets us observe canonical/final changes the
+    /// tracker makes after we recorded the response.
+    block_status: Weak<AtomicBlockStatus>,
+    /// Wall-clock time at which our node observed the block containing the response.
+    received_at: near_time::Instant,
+    // TODO(#3318): We could share the `block_height` throug the same weak pointer as
```
gilcu3:
```suggestion
    // TODO(#3318): We could share the `block_height` through the same weak pointer as
```

crates/node/src/requests/queue.rs
```diff
@@ -79,57 +77,84 @@ pub struct PendingRequests<RequestType: Request, ChainRespondArgsType: ChainResp
...
+                // response transaction.
+                // Reason for this is that we track `return_signature_and_clean_state_on_success`.
+                // This is a method that can only be called by the contract and gets called
+                // _exactly once_ per signature.
+                //
+                // BFT rules thus guarantee that there won't be a different final transaction response,
+                // which makes this early return ok.
+                return ResolutionStatus::PendingOptimistic;
+            }
+            result = ResolutionStatus::PendingNonCanonical;
```
gilcu3: the classification logic here is a bit hard to follow, could you add a pointer to some reference that would make sense of the classification, and the note? If we still don't have that, I think we need to add some doc for it. Atm I am most of the time guessing what seems correct and what not, but it is mostly on a best effort basis

crates/node/src/requests/queue.rs
```diff
@@ -397,12 +465,24 @@ impl<RequestType: Request + Clone, ChainRespondArgsType: ChainRespondArgs>
         my_participant_id: ParticipantId,
         network_api: Arc<dyn NetworkAPIForRequests>,
     ) -> Self {
+        let finalized_blocks_indexed = match RequestType::get_type() {
```
gilcu3: nit: a metric suffix would probably make it easier to read this var in the constructor below

crates/node/src/requests/queue.rs
```diff
@@ -907,43 +1010,86 @@ mod tests {
         let to_attempt3 = pending_requests.get_requests_to_attempt();
         assert_eq!(to_attempt3.len(), 1);
         assert_eq!(to_attempt3[0].request.id, req2.id);
+    }
+
+    #[test_log::test]
+    #[expect(non_snake_case)]
+    fn test_pending_requests__should_wait_for_response_to_finalize() {
```
gilcu3, line 1005: This is a very enlightening test indeed, I should have probably started here :)

crates/node/src/requests/queue.rs
```diff
@@ -907,43 +1010,86 @@ mod tests {
...
-        pending_requests.notify_new_block(requests);
-        // Then: we should re-attempt the response
+        // When: the response now lands on the canonical chain, but is not yet final.
+        drop(to_attempt2);
+        setup.add_indexed_respond_tx(req.id);
+        pending_requests.notify_new_block(setup.update());
+
+        // Then: we do NOT re-attempt — the response is on the canonical chain, so we wait for
+        // it to finalize.
         setup.advance_clock(CHECK_EACH_REQUEST_INTERVAL);
```
gilcu3, line 1040: should this be something like `2 * CHECK_EACH_REQUEST_INTERVAL` so that no race condition is possible?

crates/node/src/requests/queue.rs
```diff
@@ -907,43 +1010,86 @@ mod tests {
...
-        pending_requests.notify_new_block(requests);
-        let requests = setup.update();
-        pending_requests.notify_new_block(requests);
+        // When: enough subsequent blocks arrive to finalize the response.
+        pending_requests.notify_new_block(setup.update());
+        pending_requests.notify_new_block(setup.update());
 
-        // Then: we should not have any requests to attempt
+        // Then: we still don't re-attempt — the request is resolved.
         setup.advance_clock(CHECK_EACH_REQUEST_INTERVAL);
```
gilcu3, line 1048: probably same here?

gilcu3 (commented):
Thank you, looks very good to me!

 I still don't understand some parts, and those are the only ones holding me back from approving now. I can give it another try if we need to get this merged. 

Left a bunch of comments and questions.

crates/e2e-tests/tests/web_endpoints.rs
kevindeforth: Yeah, it also caught me by surprise. It's kind of cool, but I don't think we use it for anything right now.

image

crates/node/src/requests/recent_blocks_tracker.rs
kevindeforth: yes, exactly. `debug/blocks` prints the `RecentBlocksTracker`. Since we no loner track block content in there, we don't need this printer anymore.
the endpoints /debug/signatures and /debug/cds are better suited for displaying what requests arrived.

crates/node/src/requests/queue.rs
kevindeforth: It would make sense to solve this at the same time as #3318.
If we followed the existing (naive) pattern, then for #3318, we would add the block timestamp here and hold one copy with every instance.
Since the block timestamp and block height are part of the block itself, we might as well just store a `BlockReference` in `RecentBlocksTracker` and just store a pointer to the reference here. Seems cleaner.

crates/node/src/requests/debug.rs
kevindeforth: We could merge them, but I wonder what the practical benefit of that would be.

crates/node/src/requests/queue.rs
gilcu3: Makes sense, then it only remains to rewrite #3318 a bit so that this is also in scope

crates/e2e-tests/tests/web_endpoints.rs
netrome: Ah me neither. Looks pretty nice! 😊 

Image

crates/node/src/requests/metrics.rs
netrome: Nice, thanks!

netrome: And I assume the same goes for the same metric for CKD and signing requests added in this PR.

crates/node/src/requests/queue.rs
netrome: This also feels like a case where implementing `Default` is sensible, and the new should just be `Self::default()`.

kevindeforth: Wouldn't that change the constructor to take a `Vec` as argument?

crates/node/src/requests/queue.rs
kevindeforth: I am hesitant to make it an acceptance criteria of #3318, as it seems like an implementation detail.
This comment is intended as a suggestion. CI will force the author to remove this TODO, at which point, it is reasonable to assume that they have seen and read it. It is kind of up to them to decide how they want to structure the code. They may agree with the suggestion, or they may find a more elegant solution.

gilcu3: yeah I think that's fine, but if nothing about it is mentioned in the issue, then only the failure of CI would be noticed by the implementor, which is not optimal. In my view a TODO issue should always mention what the problem is at least, without needing a narrow acceptance criteria

crates/node/src/requests/queue.rs
gilcu3: maybe it was `Default` then :) , but that one wouldn't use `new` but `default`

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -340,20 +343,28 @@ impl<T: Clone + Debug> RecentBlocksTracker<T> {
...
-        // We must do this lookup before calling prune_old_blocks or else we may no longer have
-        // some of the blocks. Note: filter_map should not really filter anything out, but we're
-        // doing this defensively to not crash just in case we have a bug.
-        let new_final_blocks = new_final_blocks
-            .into_iter()
-            .filter_map(|node| Some((node.height, self.node_to_content.get(&node.hash)?.clone())))
+        // TODO(#3316): collapse the per-queue counter into a single shared metric on the
+        // tracker itself.
+        self.finalized_blocks_indexed
+            .inc_by(new_final_blocks.len() as u64);
```
netrome: Nit: I prefer to explicitly explain why these conversions are okay.

```suggestion
            .inc_by(u64::try_from(new_final_blocks.len()).expect("usize should always fit into a u64 on our targets"));
```

crates/node/src/requests/queue.rs
kevindeforth: > the classification logic here is a bit hard to follow, 

Are you referring to the classification that we receive from the tracker, or how we map that classification to `ResolutionStatus`?

It may help to spell out what the code does:

This method looks at the status of the block in which a response to this request has been seen.
If a response was seen in a finalized block, it means the response itself is finalized. The return value of this method indicates this by returning `ResolutionStatus::Final`.

If a response was seen in a canonical, but not in a final block, it means that the block (and thus the response) is likely to become final, so we return `PendingOptimistic`. To state naively: The response is still pending, and it could be reverted (not finalize), but since it sits on the canonical chain, we are optimistic it will become final.

If neither of these are true, it means that the block the response is in is not in the canonical chain. I.e. there is a longer chain out there, that this block (and the response) is not part of. The response is pending, but it is not on the canonical chain and likely to never finalize.

> could you add a pointer to some reference that would make sense of the classification and the note

The note is justifying the early return. Imagine you have a response vector where the first response is `canonical` and the second is `final`. This method would return early with `PendingOptimstic`, whereas the correct behavior would be to return `Final`.

The note is stating that we won't ever see such a response vector from on-chain data. It's impossible due to the BFT rules, specifically, the finality rule: once something is declared final, it won't ever be reverted.

To make sense of this, the reader needs to understand how `return_signature_and_clean_state_on_success` is invoked in the contract. The note states: _This is a method that can only be called by the contract and gets called _exactly once_ per signature_.

That means, if you have a transaction invoking that method for a specific signature, and if that transaction is final, you won't _ever_ see another transaction invoking that method for the exact same signature response in the same chain. Such transactions are invalid and the NEAR node would reject them.

Now, one may ask: "but what about duplicate transactions?". `return_signature_and_clean_state_on_success` is indexed by the yield index, which is a unique identifier for a signature request.

crates/node/src/requests/queue.rs
kevindeforth: This is a nearcore Fake clock and code execution in these tests is synchronous and serial. I don't see where we could have a race condition.

crates/node/src/requests/queue.rs
```diff
@@ -79,57 +77,84 @@ pub struct PendingRequests<RequestType: Request, ChainRespondArgsType: ChainResp
...
+/// An on-chain `respond` transaction observed in an indexed block.
 #[derive(Clone)]
-struct BufferedBlockData {
-    requests: Vec<RequestId>,
-    completed_requests: Vec<RequestId>,
-    timestamp_received: near_time::Instant,
+struct IndexedRespondTx {
+    /// Weak ref to the block's atomic status; lets us observe canonical/final changes the
+    /// tracker makes after we recorded the response.
+    block_status: Weak<AtomicBlockStatus>,
```
netrome, line 89: Hmm using a weak pointer here makes the code hard to reason about, but this seems to open the door to some nasty timing related edge cases. I chatted with Claude about this and got the following observation:

```
1. Tick starvation can lose resolutions. queue.rs:556-564 only records a Weak<AtomicBlockStatus> for each response. The transition Final → "request removed" now requires
  get_requests_to_attempt to actually run. The Weak holds no strong ref to the BlockNode; once the block falls outside the recency window, prune_old_blocks drops the Arc and
  Weak::upgrade returns None.

     So if get_requests_to_attempt is starved for ~200 blocks (≈200 s, since REQUEST_EXPIRATION_BLOCKS = 200 and the same constant is the tracker window), the sequence is:
  - response observed at H
  - block H becomes Final, then drifts out of the window and is pruned
  - next tick: IndexedRespondTxs::status() returns RemovedByTracker (queue.rs:107)
  - this is treated identically to None (queue.rs:357), so the request falls through to the cutoff/leader path
  - eventually is_older_than(cutoff_block) drops it as RequestTimedOut

     In main, that resolution was applied synchronously inside add_block, so even a stalled get_requests_to_attempt couldn't lose it. This is the most direct "weird timing" bug —
  under normal coordinator cadence it never triggers, but tick starvation now silently flips successes into MPC_CLUSTER_FAILED_SIGNATURES_COUNT{reason="timeout"} (queue.rs:660-664).
```

Afaik this seems very unlikely and does not seem to impact signing correctnes. I have not gone deep enough to have concrete better suggestions, but it seems there's still room for further refactors in these parts of the code.

All in all, I don't consider this a hard blocker.

netrome (dismissed):
Some parts of this code is giving me a bit of a pause but I don't spot anything I consider a hard blocker, so from my perspective we should be good to merge and we can keep improving the code from there.

crates/node/src/requests/queue.rs
gilcu3: right, I thought there must be another part of the code doing the re-attempts, but if that uses the same fake clock it should be fine

crates/node/src/requests/queue.rs
kevindeforth: > Hmm using a weak pointer here makes the code hard to reason about, but this seems to open the door to some nasty timing related edge cases.

The sentence structure is a bit off with the "hard to reason about, but ...".

I actually think using a weak pointer makes reasoning about the code easier. The problem here is that this is not the right place to track request timeouts, and that the queue is not in a good position to establish timeout metrics (which is why we have  #3229). From the queues perspective, all of these are valid reasons for dropping a request:
- it timed out
- it was resolved
- it sits on a dead fork and will never finalize

But the queue doesn't need to know what the reason for dropping a request is. It just needs to know whether it should be dropped or not.

Using a weak pointer is making this a more honest situation. We need to track signature success elsewhere, but not here.

> I chatted with Claude about this [...] Afaik this seems very unlikely and does not seem to impact signing correctnes. 

This is super constructed. If what claude suggests would actually happen (`get_requests_to_attempt` is starved for 200 blocks), then for the past 200 blocks, this node didn't check if it needed to act as leader for requests.
That would be a huge issue. If that was the case, I would suspect that the entire `monitor_block_updates` method would be starved, in which case, even with the current code, those metrics wouldn't advance

https://github.com/near/mpc/blob/457effe72aa74f0aca75e510e1a295b27587eecc/crates/node/src/mpc_client.rs#L188

crates/node/src/requests/queue.rs
kevindeforth: addressed in https://github.com/near/mpc/pull/3319/commits/9a48c21f78b8e1968552615eadd67687318f19a3

crates/node/src/requests/queue.rs
kevindeforth: addedin https://github.com/near/mpc/pull/3319/commits/ec7988b4fa7e9877bd7308cea4f45fc3b7cc7223

crates/node/src/requests/recent_blocks_tracker.rs
kevindeforth: addressed in https://github.com/near/mpc/pull/3319/commits/bcd194ecdfdce1ada7e1bab38e125fe49c67d67c

crates/node/src/requests/metrics.rs
kevindeforth: yep :+1:

crates/node/src/requests/queue.rs
kevindeforth: added a note to the issue, will close this discussion.

crates/node/src/requests/queue.rs
```diff
@@ -315,60 +349,82 @@ impl<RequestType: Request + Clone, ChainRespondArgsType: ChainRespondArgs>
...
+            return RequestStatus::Wait("check not due");
         }
         if self.has_active_attempt() {
-            return RequestStatusResult::Wait("active attempt ongoing");
+            return RequestStatus::Wait("active attempt ongoing");
+        }
+        // Classify any observed respond txs before doing other timeout/leader work below.
+        match self.indexed_respond_txs.status() {
+            AggregateResponseStatus::None => {
+                // Note that a respone on the non-canonical chain is treated the same as no
```
gilcu3: nit:
```suggestion
                // Note that a response on the non-canonical chain is treated the same as no
```

crates/node/src/requests/queue.rs
kevindeforth: Discussed on [slack](https://nearone.slack.com/archives/C0912BTG51T/p1779382833801739).
Source of the confusion was due to naming issues. We noticed in the process that we can simplify the `ResolutionStatus`. Fixed in https://github.com/near/mpc/pull/3319/commits/2ec9ee3b3f12a92d7d5d6aa84b16c3b5115cb62e

gilcu3 (dismissed):
Thank you!

gilcu3 (approved):

crates/node/src/requests/queue.rs
netrome: > The sentence structure is a bit off with the "hard to reason about, but ...".

Yeah it wasn't the best phrasing. I had to cut the review short unfortunately. It should be "and also".

> This is super constructed.

I agree, which is why I don't consider it a hard blocker. But it means that reasoning about the safety of this code requires a lot of assumptions on the surrounding system, which is code smell and also makes maintaining and reviewing code touching these places very time consuming if done properly.

I don't have any good suggestions on how to improve this right now, but it warrants further thinking about how we can improve how this component works and is integrated in the node as a whole. This can be saved for later though.

netrome (approved):
Thanks for the updates

PR #3325, opened 2026-05-22
chore: remove new final blocks from add block result and use snapshots for test

### LINKED ISSUE #3320: Remove new_final_blocks from AddBlockResult
### Background

In #3070, we feature-gated newly finalized blocks as a return value for test purposes, to simplify PR review.
A follow-up PR should remove the `new_final_blocks` entirely and refactor the tests.

### User Story

As a dev, I like when test code matches production code.

### Acceptance Criteria

Remove `new_final_blocks` from `AddBlockResult`.

_No response_

kevindeforth:
resolves #3320

### DIFF
--- a/crates/node/src/requests/recent_blocks_tracker.rs
+++ b/crates/node/src/requests/recent_blocks_tracker.rs
@@ -185,17 +185,6 @@ impl AtomicBlockStatus {
 
 pub struct AddBlockResult {
     pub block_ref: Weak<AtomicBlockStatus>,
-    /// Newly-finalized blocks captured before `prune_old_blocks` runs. Populated only in
-    /// test builds for assertion convenience — production paths read the counter that
-    /// `add_block` increments instead. Carrying the list this way avoids any production
-    /// allocation overhead while still letting tests reconstruct the finalized-name
-    /// string after pruning may have dropped strong refs to ancestors.
-    // TODO(#3320): drop this cfg(test) field. The test helper should observe finality by
-    // holding a strong `Arc<AtomicBlockStatus>` (upgraded from `block_ref` at `add_block`
-    // time), which survives even after `prune_old_blocks` drops the `BlockNode`.
-    // https://github.com/near/mpc/issues/3320
-    #[cfg(test)]
-    pub new_final_blocks: Vec<(BlockHeight, CryptoHash)>,
 }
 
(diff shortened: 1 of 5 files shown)


crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -747,7 +721,12 @@ pub mod tests {
         maker: Arc<TestBlockMaker>,
         tracker: RecentBlocksTracker,
         parents_of_added_blocks: HashSet<CryptoHash>,
-        content: HashMap<CryptoHash, String>,
+        /// When adding a block to the tracker via [`Tester::add`], we debug-print the tracker and
+        /// append the output to [`Tester::evolution`]. This allows us to inspect how adding a block
+        /// influences finality and canonical status of tracked blocks.
+        /// The resulting string can be compared against snapshots, allowing the unittests to catch
+        /// regressions.
+        evolution: String,
```
gilcu3, line 729: ouch, this seems hacky, but I guess it is good enough for now

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -821,34 +796,14 @@ pub mod tests {
...
-        );
-        assert_eq!(tester.check(&b16), None);
+
+        tester.add(&b11);
+        tester.add(&b12);
+        tester.add(&b13);
+        tester.add(&b14);
+        tester.add(&b15);
+
+        insta::assert_snapshot!(tester.evolution);
```
gilcu3, line 806: hmm I see this as a neat way to compare all the strings, but it makes it more complicated to check what the test is supposed to assert. Could you add an explanatory comment about that? If anyone modifies these tests they would need to be really careful.

gilcu3 (approved):
Thank you!

I am not fully sold for using snapshots this way, but it works and looks neater :)

crates/node/src/requests/recent_blocks_tracker.rs
kevindeforth: What would you consider an improvement?

gilcu3: nothing really, the status quo also seemed hacky (but I only realized that much after writing this)

pbeza (approved):

Issue #3306, opened 2026-05-21
Lift CaitSith threshold restriction after 3.11 transition


gilcu3:
### Background

#3304 added a temporary contract restriction in 3.11: `vote_add_domains` and resharing votes are rejected if they would land the network in a state with multiple CaitSith `reconstruction_threshold` values. The restriction is what guarantees that 3.11 nodes — which still dual-write triples to the unprefixed legacy `DBCol::Triple` mirror — never run with two `TripleStorage`s and therefore never hit the legacy-key collision described in #3292.

Once #3298 lands and removes the legacy column, the collision surface is gone and the restriction is no longer load-bearing. This issue tracks relaxing it so that CaitSith domains can again carry independent `reconstruction_threshold` values (which #3164 needs for real per-domain thresholds).

### Acceptance Criteria

- Drop the threshold-equality check that #3304 added to `vote_add_domains` for CaitSith domains.
- Drop the threshold-equality check #3304 added to the resharing vote path.
- Remove or update the inline comments that reference the transitional restriction and #3298 / #3304.
- Sequencing: this issue must land *after* #3298

Issue #3298, opened 2026-05-20
Remove legacy Triple column family after 3.11 release


gilcu3:
### Background

#3292 split the ECDSA triple pool per `t` and moved storage into `DBCol::TripleV2`. To keep rollbacks safe during the 3.11 rollout, every triple write also mirrors into the legacy `DBCol::Triple` column (unprefixed key), and startup migrates anything still in `Triple` into `TripleV2`. The mirror is write-amplification we only need until every production node has upgraded past 3.11.

Once 3.11 is in the field across all participants, the legacy column, dual-write, and migration step should go away in 3.12+.

### Acceptance Criteria

- Drop `DBCol::Triple` from the enum, from `DBCol::all()`, and from the  `as_str` map (`crates/node/src/db.rs`).
- Remove the `legacy_col` parameter from `DistributedAssetStorage::new` and the  mirror branches in `add_owned`, `add_unowned`, `take_owned`,  `take_unowned_inner`, and `maybe_discard_owned`  (`crates/node/src/assets.rs`). Update callers  (`TripleStorage`, `PresignatureStorage`, robust-ecdsa `PresignatureStorage`).
- Delete `migrate_legacy_triples_to_v2` and its call in  `EcdsaSignatureProvider::new` (`crates/node/src/providers/ecdsa/triple.rs`, `crates/node/src/providers/ecdsa.rs`).
- Remove the legacy `clean_db::<PairedTriple>(... DBCol::Triple ...)` arm and  the `DBCol::Triple` line in the `DeleteAll` branch of  `delete_stale_triples_and_presignatures`  (`crates/node/src/assets/cleanup.rs`).
- Update or delete the dual-write / migration tests in `crates/node/src/providers/ecdsa/triple.rs::tests`.

PR #3326, opened 2026-05-22
feat: adding resharing per-domain

### LINKED ISSUE #3169: Carry per-domain ReconstructionThreshold in resharing votes

SimonRastikian:
Closes #3169


crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -139,6 +257,9 @@ pub use near_mpc_crypto_types::{KeyForDomain, Keyset};
 pub struct ThresholdParameters {
     pub participants: Participants,
     pub threshold: Threshold,
+    // The following skip_serializing_if should be deleted after release 3.11.0
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub per_domain_thresholds: BTreeMap<DomainId, ReconstructionThreshold>,
```
SimonRastikian: A proposal should contain the domains along with their new thresholds. The BTreeMap remains empty if the proposal is not a reshare

crates/contract/src/dto_mapping.rs
```diff
@@ -215,6 +215,7 @@ impl IntoContractType<Participants> for dtos::Participants {
 impl IntoContractType<ThresholdParameters> for dtos::ThresholdParameters {
     fn into_contract_type(self) -> ThresholdParameters {
         ThresholdParameters::new_unvalidated(self.participants.into_contract_type(), self.threshold)
```
SimonRastikian, line 224: I did not want to change that but question to the public: Why is this a new_unvalidated instead of the normal new (with validation)?

SimonRastikian (commented):
Not sure if I need to add requirements for `validate_incoming_proposal`. In fact the requirements there feel to me very randomly chosen and without clear basis. I wouldn't know if something is overdone or something is missing. However, I am pretty sure there is good logic behind them but it's not clear to me what.
Would need someone's wisdom on that.

crates/contract/src/primitives/thresholds.rs
```diff
@@ -12,26 +13,64 @@ const MIN_THRESHOLD_ABSOLUTE: u64 = 2;
...
     participants: Participants,
     threshold: Threshold,
+    /// Proposed per-domain `ReconstructionThreshold` updates for this
+    /// resharing. Empty map means "keep current per-domain thresholds";
+    /// populated map must cover every existing domain (validated in
+    /// [`super::super::state::running::RunningContractState::process_new_parameters_proposal`]).
+    /// `skip_serializing_if`+`default` preserve the legacy `state()` wire
+    /// shape when no resharing is in flight.
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    per_domain_thresholds: BTreeMap<DomainId, ReconstructionThreshold>,
```
gilcu3: why do we want to keep the legacy state() wire shape? It will probably be simpler to add compatible serialization.

crates/contract/src/primitives/thresholds.rs
```diff
@@ -12,26 +13,64 @@ const MIN_THRESHOLD_ABSOLUTE: u64 = 2;
...
+    pub fn per_domain_thresholds(&self) -> &BTreeMap<DomainId, ReconstructionThreshold> {
+        &self.per_domain_thresholds
+    }
+
+    /// Clears the per-domain overlay. Called when a resharing proposal has
+    /// been applied to the [`super::domain::DomainRegistry`] — the proposal
+    /// map is no longer meaningful as part of the running state's parameters.
+    pub fn clear_per_domain_thresholds(&mut self) {
+        self.per_domain_thresholds.clear();
+    }
```
gilcu3: So this is needed because `ThresholdParameters` is used both for voting and for storing the current parameter set. Wouldn't it make sense to separate the types then?

crates/contract/src/v3_10_state.rs
```diff
@@ -7,23 +7,107 @@
...
+            add_domains_votes: old.add_domains_votes,
+            previously_cancelled_resharing_epoch_id: old.previously_cancelled_resharing_epoch_id,
+        }
+    }
+}
+
+/// Pre-3.11 layout of `ProtocolContractState`. Only the `Running` variant
+/// has a verified shadow — Initializing/Resharing reuse current types and
+/// would fail to deserialize old data, which matches the pre-existing
+/// "migration panics if not Running" policy.
```
gilcu3: there is a small inconsistency here where this comment mentions "pre-3.11 layout" while the one below mentions "3.10.0 Contract layout"

crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
```diff
@@ -4549,7 +4549,7 @@ expression: abi
           "minimum": 0.0
         },
         "ThresholdParameters": {
-          "description": "Threshold parameters for distributed key operations.",
+          "description": "Threshold parameters for distributed key operations.\n\n`per_domain_thresholds` carries a proposed update for each domain's [`ReconstructionThreshold`] when this struct flows into `vote_new_parameters`. An empty map means \"keep current per-domain thresholds\"; a populated map must cover every existing domain (validated by the contract). Outside of resharing proposals the map is empty.\n\nBackwards-compat for the legacy `{ participants, threshold }` wire shape is intrinsic: `serde(default)` parses old JSON without `per_domain_thresholds` as an empty map, and `skip_serializing_if` omits the field from `state()` output when no overlay is in flight.",
```
gilcu3: lets keep the descriptions that leak into the contract abi as simple as possible. This can be achieved by separating doc comments `///` from dev comment `//`

crates/near-mpc-contract-interface/src/types/state.rs
gilcu3: I understand the point, but almost all proposals are actually reshares, as we normally vary both the domain's and governance thresholds when adding or removing participants.

crates/contract/src/primitives/domain.rs
```diff
@@ -177,6 +179,41 @@ impl DomainRegistry {
...
         self.next_domain_id
     }
+
+    /// Returns a new registry whose domains have their
+    /// `reconstruction_threshold` rewritten from `overlay`. Domain IDs in
+    /// `overlay` that are not present in the registry are rejected with
+    /// [`DomainError::UnknownDomainInProposal`]. Domains absent from
+    /// `overlay` retain their existing threshold. An empty overlay returns a
+    /// structurally identical clone.
+    pub fn with_overlaid_thresholds(
```
gilcu3: This might make sense with the proactive resharing, but not with our more typical resharing where thresholds change most of the time. I am torn between not having an overlay (always demanding explicit thresholds for each domain) and the current solution

crates/contract/src/state/running.rs
```diff
@@ -137,14 +137,28 @@ impl RunningContractState {
...
+        for domain in self.domains.domains() {
+            let effective_threshold = overlay
+                .get(&domain.id)
+                .copied()
+                .unwrap_or(domain.reconstruction_threshold);
+            let proposed = DomainConfig {
+                reconstruction_threshold: effective_threshold,
+                ..domain.clone()
+            };
+            crate::primitives::domain::validate_domain_threshold(&proposed, new_num_participants)?;
```
gilcu3: nit: we normally avoid this and import the function first

crates/contract/src/state/running.rs
```diff
@@ -473,4 +487,74 @@ pub mod running_tests {
             "Expected InsufficientParticipantsForProtocol, got: {err}"
         );
     }
+
+    // ----- #3169: per-domain threshold proposals in resharing -----
```
gilcu3:
```suggestion
```

gilcu3 (changes requested):
Thank you!

I have two main problems with the current solution.

- The first is that it tries to reuse `ThresholdParameters`, while I think it now makes sense to have one type for the running state:
```
pub struct ThresholdParameters {
    participants: Participants,
    threshold: Threshold,
}
```
and another for the voting:
```
pub struct ThresholdParametersVote {
    participants: Participants,
    threshold: Threshold,
    per_domain_thresholds: BTreeMap<DomainId, ReconstructionThreshold>, 
}
```

- The second is that the way to achieve backward compatibility in this PR seems hacky, as it uses:
```
#[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
```
It should be better to define `Compat` versions of the modified struct in the interface crate, as we did with the previous migrations.

I am also torn about the overlay concept. Most manual resharing operations will anyway change the thresholds per domain (as they add or remove participants). On the other hand, automatic resharing operations (in the future) should have no problem creating the correct parameters. Therefore, it seems strictly better, to have a simpler system, to require that the votes always contain a `ReconstructionThreshold` per domain.

crates/near-mpc-contract-interface/src/types/state.rs
SimonRastikian: If you're suggesting that I should make two proposals then this should be a big (now draft PR) rework of all the infrastructure of the propose & vote mechanism

crates/contract/src/dto_mapping.rs
SimonRastikian: @gilcu3 do you have an idea?

crates/contract/src/primitives/thresholds.rs
SimonRastikian: You're right this skip_serializing_if is not necessary. A reader either reads it or ignores it

SimonRastikian: Will take away this part

crates/contract/src/primitives/thresholds.rs
SimonRastikian: I agree it would be cleaner to keep ThresholdParameters  as is and add something like 
  `ProposedThresholdParameters { parameters, per_domain_thresholds }` used only by vote_new_parameters, the votes map, and KeyEvent.

This will require some more refactor I prefer to push into a newer PR.

crates/contract/src/primitives/domain.rs
SimonRastikian: Let's keep it this way as proactive resharing is on our plate. It was a blocker for 3.11.0 but no more

crates/contract/src/state/resharing.rs
```diff
@@ -465,4 +472,81 @@ pub mod tests {
...
+            // Pick a new value that is achievable for the proposed participant
+            // count: the current proposal's threshold is a safe lower-bound.
+            let new_value = state
+                .resharing_key
+                .proposed_parameters()
+                .threshold()
+                .value();
+            assert!(new_value >= 2);
+            let new_threshold = ReconstructionThreshold::new(new_value);
+            assert_ne!(new_threshold, original_threshold);
```
anodar: If my understanding is correct, `original_threshold` is always 2 for all domains (https://github.com/near/mpc/blob/a2a53f3c20b8dbfc3ecbae4eb20485691e701736/crates/contract/src/primitives/test_utils.rs#L31), but new threshold value can also land on 2 depending on randomness picked in gen_resharing_state, so this can become flaky no ? 


Can't `gen_valid_params_proposal` generate 2 or 3 participants in which case new threshold ( ceil(0.6 * new_n)  ) be 2 ?

crates/contract/src/state/running.rs
```diff
@@ -137,19 +137,28 @@ impl RunningContractState {
         // ensure the proposal is valid against the current parameters
         self.parameters.validate_incoming_proposal(proposal)?;
 
-        // TODO(#3169): once resharing votes carry per-domain `t` (and the node
-        // honors per-domain reconstruction thresholds, #3164), this path must
-        // (1) reject proposals that would leave CaitSith domains with differing
```
anodar, line 142: First part of the TODO is dropped but where is this 3.11-transition lock enforced now that CaitSith domains must have same reconstruction threshold ? Do we not need this invariant anymore?

crates/contract/src/primitives/thresholds.rs
anodar: +1 for separate structs for voting and storing parameters.

crates/contract/src/state/running.rs
```diff
@@ -137,19 +137,28 @@ impl RunningContractState {
...
-        // would wrongly block legitimate resharings.
-        // let new_num_participants = proposal.participants().len() as u64;
-        // for domain in self.domains.domains() {
-        //     crate::primitives::domain::validate_domain_threshold(domain, new_num_participants)?;
-        // }
+        // Validate effective per-domain thresholds against the proposed new
+        // participant count. Domains not present in the overlay keep their
+        // existing threshold; overlay entries override. An overlay entry
+        // referencing an unknown domain ID is rejected.
+        let new_num_participants = proposal.participants().len() as u64;
```
anodar: nit: consider using checked conversion instead of as.

crates/contract/src/dto_mapping.rs
anodar: I think to allow low threshold values in tests that would otherwise not pass in production (e.g. >= 60%)

crates/contract/src/primitives/domain.rs
```diff
@@ -177,6 +179,41 @@ impl DomainRegistry {
...
+        &self,
+        overlay: &BTreeMap<DomainId, ReconstructionThreshold>,
+    ) -> Result<DomainRegistry, Error> {
+        for id in overlay.keys() {
+            if !self.domains.iter().any(|d| d.id == *id) {
```
anodar, line 222: This check is already done in `with_overlaid_thresholds`, is this really needed here?

Fine to leave (better safe) but consider adding comment.

SimonRastikian:
@anodar Validation is not the job of the DTO mapping, the contract that accepts the parameter will call a certain `validate` and `validate_incoming_proposal` down the stream. I added a comment for it

crates/contract/src/state/running.rs
SimonRastikian: [fca347b](https://github.com/near/mpc/pull/3326/commits/fca347bcc887db0c833edf13b573cfb889bcb252)

SimonRastikian: Thank you!

crates/contract/src/state/resharing.rs
SimonRastikian: The test is now deterministic 
[cdae919](https://github.com/near/mpc/pull/3326/commits/cdae919012d0d23ebf10d28168c44fe4ffbc6049)

gilcu3:
@claude review

Pay attention to what in the changes is making the contract binary bigger, would be nice to keep the current limits but not a problem if we need to bump them a bit

crates/contract/src/primitives/domain.rs
```diff
@@ -567,4 +636,176 @@ pub mod tests {
         assert_eq!(remaining.proposal_by_account[&auth_ids[0]], proposal_a);
         assert_eq!(remaining.proposal_by_account[&auth_ids[1]], proposal_b);
     }
+
+    #[expect(non_snake_case)]
+    mod with_overlaid_thresholds {
```
gilcu3: we generate use `mod tests` for tests, I don't see any reason for using another name here

crates/contract/src/primitives/threshold_votes.rs
```diff
@@ -100,6 +106,45 @@ mod tests {
...
+
+        let base = gen_proposed_threshold_params(30);
+        let mut overlay_a = BTreeMap::new();
+        overlay_a.insert(DomainId(0), ReconstructionThreshold::new(2));
+        let mut overlay_b = BTreeMap::new();
+        overlay_b.insert(DomainId(0), ReconstructionThreshold::new(3));
+        let proposal_a = base.clone().with_per_domain_thresholds(overlay_a);
+        let proposal_b = base.with_per_domain_thresholds(overlay_b);
+
+        // When each voter casts a different overlay
```
gilcu3: nit: overlays are not the right keyword here
```suggestion
        // When each voter casts a different proposal
```

crates/contract/src/v3_10_state.rs
```diff
@@ -102,3 +184,33 @@ impl From<MpcContract> for crate::MpcContract {
...
 }
+
+#[cfg(test)]
+#[expect(non_snake_case)]
+mod tests {
+    use super::*;
+    use crate::primitives::test_utils::{NUM_PROTOCOLS, gen_participants};
+    use crate::primitives::thresholds::Threshold;
+
+    /// Borsh round-trip: write a `ThresholdParametersVotes` in the OLD layout
```
anodar: It's not really testing it though is it? Seems to serialize ThresholdParameters and deserialize it into ThresholdParameters again. I don't see OldThresholdParametersVotes used here.

anodar (dismissed):
LGTM, please fix the test as part of fixing conflict with main.

crates/contract/src/primitives/thresholds.rs
```diff
@@ -165,6 +168,68 @@ impl ThresholdParameters {
     }
 }
```
gilcu3, line 167: Isnt this wrong? I thought this type was the governance threshold after this PR?

I also don't understand the text:

> so this type can never hold a meaningless overlay

The comment should just mention what it is

crates/contract/src/primitives/thresholds.rs
```diff
@@ -165,6 +168,68 @@ impl ThresholdParameters {
...
+
+    /// Builder-style helper: replace the per-domain reconstruction-threshold
+    /// overlay. Convenient for constructing proposals (notably in tests).
+    pub fn with_per_domain_thresholds(
+        mut self,
+        per_domain_thresholds: BTreeMap<DomainId, ReconstructionThreshold>,
+    ) -> Self {
+        self.per_domain_thresholds = per_domain_thresholds;
+        self
+    }
```
gilcu3, line 202: if it happens to be used only in tests, we should feature gate it to save contract space.

crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
```diff
@@ -3833,6 +3833,31 @@ expression: abi
             }
           }
         },
+        "ProposedThresholdParameters": {
+          "description": "A proposed set of threshold parameters submitted to `vote_new_parameters`. Carries the new [`ThresholdParameters`] plus an optional per-domain `ReconstructionThreshold` overlay for the resharing it would trigger.",
```
gilcu3: What do you mean by "new [`ThresholdParameters`]" this is not accurate

crates/devnet/src/cli.rs
```diff
@@ -298,6 +298,21 @@ pub struct MpcVoteNewParametersCmd {
...
     #[clap(long, value_delimiter = ',')]
     pub voters: Vec<usize>,
+    /// Optional per-domain reconstruction-threshold overlay, given as
+    /// repeated `DOMAIN_ID:THRESHOLD` pairs (e.g.
+    /// `--per-domain-threshold 0:6 --per-domain-threshold 1:5`).
+    /// Omitting it preserves every domain's existing threshold; supplying it
+    /// changes those listed and re-validates the rest against the new
+    /// participant count.
+    #[clap(long = "per-domain-threshold", value_parser = parse_per_domain_threshold)]
+    pub per_domain_thresholds: Vec<(u64, u64)>,
```
gilcu3: why not using the same format we have for the voters above? Instead of multiple `--per-domain-threshold`, a single `--per-domain-threshold` with values coma separated. It would be more in line with the rest of the cli

crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -241,6 +269,10 @@ pub struct ResharingContractState {
     pub reshared_keys: Vec<KeyForDomain>,
     pub resharing_key: KeyEvent,
     pub cancellation_requests: HashSet<AuthenticatedAccountId>,
+    /// Per-domain `ReconstructionThreshold` overlay carried from the accepted
+    /// proposal. Applied to the `DomainRegistry` when resharing completes.
+    #[serde(default)]
+    pub per_domain_thresholds: BTreeMap<DomainId, ReconstructionThreshold>,
```
gilcu3: if this `serde(default)` is needed for backward compatibility reasons we should add a TODO to remove it after the 3.12 release

crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -141,6 +144,31 @@ pub struct ThresholdParameters {
     pub threshold: Threshold,
 }
```
gilcu3, line 145: isn't this type no longer speaking about crypto thresholds after this PR?

crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -141,6 +144,31 @@ pub struct ThresholdParameters {
...
+#[cfg_attr(
+    all(feature = "abi", not(target_arch = "wasm32")),
+    derive(schemars::JsonSchema)
+)]
+pub struct ProposedThresholdParameters {
+    #[serde(flatten)]
+    pub parameters: ThresholdParameters,
+    #[serde(default)]
+    pub per_domain_thresholds: BTreeMap<DomainId, ReconstructionThreshold>,
+}
```
gilcu3, line 168: I don't see how we would remove `#[serde(flatten)]` after 3.12.  Please make sure we have a migration path forward, using compat structs if needed

gilcu3, line 168: Also, what's the reason for having this type duplicate? We are trying to avoid that to keep contract size low. I see that `ThresholdParameters` is also duplicate (not in this PR), but wondering if that's the only reason, because in that case we could solve it first

crates/contract/src/v3_11_2_state.rs
gilcu3, line 1: Another PR needs to land so that these changes are made against the 3.11 state instead

crates/contract/src/primitives/domain.rs
```diff
@@ -69,6 +71,30 @@ pub fn validate_domain_threshold(
     Ok(())
 }
 
+/// Enforces the 3.11-transition lock: every CaitSith domain (ForeignTx
+/// included) must share a single `reconstruction_threshold` so the legacy
+/// unprefixed `DBCol::Triple` mirror (#3292) can't collide. Domains using
+/// other protocols are unconstrained. The first CaitSith domain encountered
+/// fixes the expected value; any later CaitSith domain with a different
+/// threshold is rejected. Remove once #3298 drops the mirror.
+pub fn validate_caitsith_uniform_threshold(domains: &[DomainConfig]) -> Result<(), Error> {
```
gilcu3: #3298 already landed, so this is a bit outdated, but I think we have an issue for it #3306

crates/contract/src/state/resharing.rs
```diff
@@ -462,7 +479,98 @@ pub mod tests {
...
             .unwrap_err();
     }
+
+    #[expect(non_snake_case)]
+    mod per_domain_threshold_overlay {
+        use super::*;
+        use crate::state::key_event::tests::Environment;
+        use crate::state::test_utils::gen_running_state;
+        use near_mpc_contract_interface::types::ReconstructionThreshold;
+        use std::collections::BTreeMap;
```
gilcu3: why using a different mod for these tests?

gilcu3 (commented):
Partial review, dropped quite a few comments. 

The most concerning for me are the current backward compatibility approach, and the interpretation of `ThresholdParamters` as cryptographic (while I believe it is only about governance now).

As I had mentioned in slack, we need to cleanup the 3.10 migrations before this can land.

crates/contract/src/primitives/domain.rs
SimonRastikian: This is a mod inside mod tests. I should not have nested mod tests

SimonRastikian: [5511ec9](https://github.com/near/mpc/pull/3326/commits/5511ec9aed83eb8eed6f0fb5b65b60238b104bbc)

crates/contract/src/state/resharing.rs
SimonRastikian: [5511ec9](https://github.com/near/mpc/pull/3326/commits/5511ec9aed83eb8eed6f0fb5b65b60238b104bbc)

crates/contract/src/primitives/domain.rs
SimonRastikian: [c7b6123](https://github.com/near/mpc/pull/3326/commits/c7b61231e44e29a4f27ff6fd014896c629ab9a65)

crates/contract/src/v3_11_2_state.rs
SimonRastikian: Done!

crates/contract/src/state/resharing.rs
SimonRastikian: [304a3d2](https://github.com/near/mpc/pull/3326/commits/304a3d2a44fd8a24195077e08c9a0c68728c59a6)

crates/contract/src/primitives/thresholds.rs
SimonRastikian: Done

crates/near-mpc-contract-interface/src/types/state.rs
SimonRastikian: So ThresholdParameters is a duplicate itself which itself calls a duplicate Participants struct. I think the problem is much more fundamental. I opened an issue for it  https://github.com/near/mpc/issues/3492

SimonRastikian: [aff981c](https://github.com/near/mpc/pull/3326/commits/aff981c5d20d685edf742a70c7941c11e2941f2f)

crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
SimonRastikian: Thanks for the finding! [c4a5a37](https://github.com/near/mpc/pull/3326/commits/c4a5a37ef849fc6f3f7ef2fd2e956c86f1d7ea36)

crates/near-mpc-contract-interface/src/types/state.rs
SimonRastikian: [7ea2180](https://github.com/near/mpc/pull/3326/commits/7ea21801873514ad6fe2e1f0273ab39bcc20ee97)

crates/near-mpc-contract-interface/src/types/state.rs
SimonRastikian: [1695875](https://github.com/near/mpc/pull/3326/commits/1695875ec0577ca3dba9ac467ad91fe7f322d356)

crates/contract/src/primitives/thresholds.rs
SimonRastikian: [7ea2180](https://github.com/near/mpc/pull/3326/commits/7ea21801873514ad6fe2e1f0273ab39bcc20ee97)

crates/devnet/src/cli.rs
SimonRastikian: [cc40c21](https://github.com/near/mpc/pull/3326/commits/cc40c217e453806e87c3c15e06b27969eb008882)


Thanks

crates/contract/src/primitives/domain.rs
gilcu3: Please add this comment as a proper TODO for #3306

crates/contract/src/v3_11_2_state.rs
```diff
@@ -17,19 +19,99 @@ use crate::{
...
+    add_domains_votes: AddDomainsVotes,
+    previously_cancelled_resharing_epoch_id: Option<EpochId>,
+}
+
+impl From<OldRunningContractState> for RunningContractState {
+    fn from(old: OldRunningContractState) -> Self {
+        RunningContractState {
+            domains: old.domains,
+            keyset: old.keyset,
+            parameters: old.parameters,
```
netrome, line 83: Since we did perform a resharing on testnet recently, we need to handle the case where the old parameters mismatch. For example, currently the global threshold in testnet is 7 although each domain has a reconstruction bound of 6.

I'd suggest updating this PR so that we can overwrite all reconstruction bounds with the global domain during the migration so we can self-heal this discrepancy.

netrome (changes requested):
A small but important change requested

crates/contract/src/primitives/domain.rs
```diff
@@ -177,6 +203,49 @@ impl DomainRegistry {
...
+    /// The resulting registry is re-checked against the 3.11-transition lock
+    /// (see [`validate_caitsith_uniform_threshold`]): because the overlay can
+    /// rewrite per-domain thresholds, it could otherwise leave CaitSith
+    /// domains with differing thresholds. This is the authoritative chokepoint
+    /// — no resharing transition can produce a registry that violates the
+    /// invariant.
+    pub fn with_overlaid_thresholds(
+        &self,
+        overlay: &BTreeMap<DomainId, ReconstructionThreshold>,
+    ) -> Result<DomainRegistry, Error> {
```
gilcu3: did you try doing this mutation in place, for better performance?

crates/contract/src/primitives/thresholds.rs
gilcu3, line 21:
```suggestion
/// Stores the threshold key parameters: the owners of key shares
/// (`participants`) and the cryptographic `threshold`. This is the stored,
/// always-current shape.
#[near(serializers=[borsh, json])]
#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
pub struct ThresholdParameters {
participants: Participants,
threshold: Threshold,
}
```

crates/contract/src/primitives/thresholds.rs
```diff
@@ -165,6 +168,69 @@ impl ThresholdParameters {
...
+/// An empty overlay means "keep current per-domain thresholds"; a populated map
+/// must reference only existing domains (validated in
+/// `RunningContractState::process_new_parameters_proposal`).
+#[near(serializers=[borsh, json])]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub struct ProposedThresholdParameters {
+    parameters: ThresholdParameters,
+    #[serde(default)]
+    per_domain_thresholds: BTreeMap<DomainId, ReconstructionThreshold>,
+}
```
gilcu3: I am not sure how the reader of this doc is supposed to figure what an overlay is. This comment needs some improvement, maybe an inline comment on `per_domain_thresholds` would help

crates/contract/src/primitives/thresholds.rs
gilcu3, line 1: We have a lot of tests in this file for validating proposals, and in this PR we are extending what a proposal is. Therefore, I would expect that we also extend or add new tests for all the new cases. The new reconstruction threshold probably require very similar tests

crates/contract/src/state/resharing.rs
```diff
@@ -31,6 +32,10 @@ pub struct ResharingContractState {
     pub reshared_keys: Vec<KeyForDomain>,
     pub resharing_key: KeyEvent,
     pub cancellation_requests: HashSet<AuthenticatedAccountId>,
+    /// Per-domain `ReconstructionThreshold` overlay carried from the accepted
+    /// proposal. Applied to the [`DomainRegistry`](crate::primitives::domain::DomainRegistry)
+    /// when resharing completes; empty means "keep current per-domain thresholds".
+    pub per_domain_thresholds: BTreeMap<DomainId, ReconstructionThreshold>,
```
gilcu3: why not computing the actual `per_domain_thresholds` before entering the resharing state? That way we don't need to keep explaining that the value is just an overlay, which adds complexity

crates/contract/src/v3_11_2_state.rs
```diff
@@ -17,19 +19,99 @@ use crate::{
...
+                .map(|(acc, params)| {
+                    (
+                        acc,
+                        ProposedThresholdParameters::new(params, BTreeMap::new()),
+                    )
+                })
+                .collect(),
+        }
+    }
+}
```
gilcu3, line 63: this is technically not true. According to the old resharing, the new thresholds per domain would be the old governance threshold, not the same they had before. Here we should either do that, or simplify by deleting the old votes (@netrome operationally wise this should be fine)

crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -141,6 +144,61 @@ pub struct ThresholdParameters {
...
+//
+// That is byte-identical for JSON, borsh, and the generated ABI: borsh is
+// positional and ignores serde attributes, and the inlined JSON keys match. So
+// it needs no compat struct and no migration — only the conversion impls in
+// `dto_mapping.rs` that read `.parameters` must follow the field move. The
+// `proposed_threshold_parameters__*` wire-lock tests below pin this shape so a
+// non-equivalent change fails loudly. Changing the *shape itself* (e.g. nesting
+// the params under a `parameters` key) WOULD be wire-breaking and would instead
+// require a compat deserializer accepting both the old flat and new nested
+// forms across a transition window.
```
gilcu3: this comment is overly verbose (LLM verbosity level). This should be simplified before merging

crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -141,6 +144,61 @@ pub struct ThresholdParameters {
...
+pub struct ProposedThresholdParameters {
+    #[serde(flatten)]
+    pub parameters: ThresholdParameters,
+    // TODO(#3495): drop `serde(default)` after the 3.12 release. It exists only
+    // so pre-3.11 `vote_new_parameters` payloads (which omit this field) still
+    // deserialize as an empty overlay; once all callers populate it explicitly
+    // the field should be mandatory so legacy/malformed payloads fail loudly.
+    #[serde(default)]
+    pub per_domain_thresholds: BTreeMap<DomainId, ReconstructionThreshold>,
+}
```
gilcu3, line 168: I still don't see how to remove the `#[serde(flatten)]` in the next transition. It seems much easier to define a Compat struct.

crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -413,4 +476,96 @@ mod tests {
...
+        // Given a proposal carrying a non-empty per-domain overlay.
+        let proposal = sample_proposal();
+
+        // When borsh round-tripped.
+        let bytes = borsh::to_vec(&proposal).unwrap();
+        let decoded: ProposedThresholdParameters = borsh::from_slice(&bytes).unwrap();
+
+        // Then it survives unchanged.
+        assert_eq!(decoded, proposal);
+    }
```
gilcu3, line 564: why do we need the borsh representation to remain the same? In this case we do not care about it. Maybe that's the reason the code above does not match previous migrations

crates/contract/src/state/running.rs
```diff
@@ -132,24 +142,57 @@ impl RunningContractState {
...
+            .collect();
+        for domain in &effective_domains {
+            validate_domain_threshold(domain, new_num_participants)?;
+        }
+        // 3.11-transition lock: the overlay can rewrite per-domain thresholds,
+        // so it could leave CaitSith domains with differing thresholds — which
+        // `vote_add_domains` forbids and the legacy `DBCol::Triple` mirror
+        // (#3292) requires. Fail fast here; `with_overlaid_thresholds` re-runs
+        // this same check at the final resharing transition.
+        validate_caitsith_uniform_threshold(&effective_domains)?;
```
gilcu3: Please add a proper TODO for this one, which should be eliminated at some point

crates/contract/src/state/running.rs
```diff
@@ -167,7 +210,12 @@ impl RunningContractState {
 
         // finally, vote.
         let n_votes = self.parameters_votes.vote(proposal, candidate);
-        Ok(proposal.participants().len() as u64 == n_votes)
+        let num_participants = u64::try_from(proposal.participants().len()).map_err(|e| {
+            ConversionError::DataConversion {
+                reason: format!("participant count does not fit in u64: {e}"),
+            }
+        })?;
```
gilcu3: IIUC we already computed this variable in line 155, no need to repeat here

crates/contract/src/state/running.rs
```diff
@@ -186,32 +234,20 @@ impl RunningContractState {
...
-        // later CaitSith — in the same proposal or in future calls — must
-        // match it.
-        let mut expected_caitsith_t = self
+        // 3.11-transition lock: all CaitSith domains (ForeignTx included) must
+        // share one `reconstruction_threshold` so the legacy unprefixed
+        // `DBCol::Triple` mirror (#3292) can't collide. If no CaitSith domain
+        // exists yet the first one is free to pick any valid `t`; any later
+        // CaitSith — already present or in this proposal — must match it.
+        // Tracked for removal in #3306.
+        let existing_and_new: Vec<DomainConfig> = self
```
gilcu3, line 236: please convert this to a proper TODO so that we can easily track it (we have CI steps to check stale TODOs, that's why I am insisting on this)

gilcu3 (changes requested):
Thank you for all the fixes. I think we still have a few remaining:

- There are quite a few verbose comments with LLM smell, we should make them to the point to reduce the noise
- We should write proper TODOs for comments that must be handled in the future
- The contract migration seems to have a problem in the reconstruction threshold handling
- The backward compatibility in the dto crate seems to be assuming we need to keep borsh compatibility, which is not the case. A much cleaner solution is possible without that constraint.

crates/contract/src/primitives/thresholds.rs
SimonRastikian: [828df6c](https://github.com/near/mpc/pull/3326/commits/828df6cffa72ffe4f910e3986aa5ba15ca43719c)

crates/contract/src/primitives/thresholds.rs
SimonRastikian: [5ee8a8f](https://github.com/near/mpc/pull/3326/commits/5ee8a8fcc6633646ca5c3ab658d2f0d98b46cfae)

SimonRastikian: I renamed overlay to threshold_updates

crates/contract/src/primitives/domain.rs
SimonRastikian: Done https://github.com/near/mpc/pull/3326/commits/5ee8a8fcc6633646ca5c3ab658d2f0d98b46cfae

crates/contract/src/dto_mapping.rs
kevindeforth: > I did not want to change that but question to the public: Why is this a new_unvalidated instead of the normal new (with validation)?

That's an excellent question. I don't see a good reason why this needs to be infallible.
Generally speaking, if we receive the interface type and convert it to the contract internal type, it means we have something external communicating with the contract, so adding these checks makes sense.

The comment higher up in the code states:

>        // every contract entry point that accepts these parameters
>        // (`init`, `init_running`, `vote_new_parameters`) calls `validate()` /
>        // `validate_incoming_proposal` downstream, so production proposals are still
>        // rejected if they violate the threshold bounds.

If they get rejected later, those methods already allow for an error path, so I think we should prefer to reject sooner.

> I think to allow low threshold values in tests that would otherwise not pass in production (e.g. >= 60%)

Why do we need tests with sub-production thresholds that use the `dtos` type? Those sound like contract unit tests and we can rely on the contract internal type there, directly constructing with `new_unvalidated`.

@SimonRastikian This is definitely out of scope of this PR, but if you agree, I think it might be worth opening a tracking issue for this.

crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -130,7 +130,10 @@ pub use near_mpc_crypto_types::{KeyForDomain, Keyset};
 // Threshold/Participants Types
 // =============================================================================
 
-/// Threshold parameters for distributed key operations.
+/// Threshold parameters for distributed key operations: the current
+/// participant set and the governance threshold. This is the stored,
+/// always-current shape; per-domain reconstruction-threshold *proposals* live
+/// on [`ProposedThresholdParameters`].
```
kevindeforth, line 136: Is the plan to deprecate and remove `ThresholdParameters` eventually?

crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -141,6 +144,39 @@ pub struct ThresholdParameters {
...
+    all(feature = "abi", not(target_arch = "wasm32")),
+    derive(schemars::JsonSchema)
+)]
+pub struct ProposedThresholdParameters {
+    #[serde(flatten)]
+    pub parameters: ThresholdParameters,
+    // TODO(#3495): drop `serde(default)` after the 3.12 release. It exists only
+    // so pre-3.11 `vote_new_parameters` payloads (which omit this field) still
+    // deserialize as an empty map; once all callers populate it explicitly
+    // the field should be mandatory so legacy/malformed payloads fail loudly.
```
kevindeforth: We don't really care about pre-3.11 `vote_new_parameters` payloads. That call is initiated by node operators and we can just tell them to use the new format.

The reason we absolutely need it though, is because otherwise, the nodes won't be able to de-serialize the old contract state.

Please update the doc accordindly.

crates/contract/src/primitives/thresholds.rs
```diff
@@ -165,6 +166,70 @@ impl ThresholdParameters {
...
+///
+/// These threshold updates are meaningful only while a proposal is in flight —
+/// they are applied to the [`super::domain::DomainRegistry`] when resharing
+/// completes and the stored [`ThresholdParameters`] is taken from
+/// [`Self::parameters`], so they are structurally dropped rather than scrubbed
+/// at runtime.
+///
+/// An empty map means "keep current per-domain thresholds"; a populated map
+/// must reference only existing domains (validated in
+/// `RunningContractState::process_new_parameters_proposal`).
```
kevindeforth: This comment smells of LLM. 

> /// A proposed set of threshold parameters submitted to `vote_new_parameters`:

This is a thing we have throughout our codebase and it's not ideal. If someone modifies the `vote_new_parameters` method and forgets to change this comment, then it will become stale, inaccurate and lead to confusion.

> /// These threshold updates are meaningful only while a proposal is in flight —
> /// they are applied to the [`super::domain::DomainRegistry`] when resharing
> /// completes and the stored [`ThresholdParameters`] is taken from
> /// [`Self::parameters`], so they are structurally dropped rather than scrubbed
> /// at runtime.

What do "structurally dropped" and "scrubbed at runtime" mean and why do I, as someone who is reading this comment, need to care about their difference?

crates/near-mpc-contract-interface/src/types/state.rs
kevindeforth: Yeah, lets inline. I don't see a benefit in delaying that.
I would also advocate for explicitly naming it `governance_threshold` instead of `threshold`. But I understand that will add some complexity on the interface migration.

crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -413,4 +454,96 @@ mod tests {
         // Then
         assert_eq!(output, "Contract is not initialized\n");
     }
+
+    /// A proposal carrying non-empty per-domain threshold updates, used by the
+    /// `ProposedThresholdParameters` wire-lock tests.
+    fn sample_proposal() -> ProposedThresholdParameters {
```
kevindeforth: note: feels like a snapshot file would serve us better here.

crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -413,4 +454,96 @@ mod tests {
...
+    /// `serde(default)` it guards.
+    #[test]
+    #[expect(non_snake_case)]
+    fn proposed_threshold_parameters__legacy_payload_omitting_field__parses_as_empty() {
+        // Given a legacy proposal value with the per_domain_thresholds field absent.
+        let mut legacy = serde_json::to_value(sample_proposal()).unwrap();
+        legacy
+            .as_object_mut()
+            .unwrap()
+            .remove("per_domain_thresholds");
```
kevindeforth: Yeah, with snapshots, this would be cleaner. It's an easy task for Claude.

crates/contract/src/state/test_utils.rs
```diff
@@ -11,11 +11,12 @@ use crate::primitives::{
...
     participants::{ParticipantId, Participants},
     test_utils::{gen_participant, gen_threshold_params},
-    thresholds::{Threshold, ThresholdParameters},
+    thresholds::{ProposedThresholdParameters, Threshold, ThresholdParameters},
 };
 use rand::Rng;
+use std::collections::BTreeMap;
 
-pub fn gen_valid_params_proposal(params: &ThresholdParameters) -> ThresholdParameters {
+pub fn gen_valid_params_proposal(params: &ThresholdParameters) -> ProposedThresholdParameters {
```
kevindeforth, line 21: Lets add a comment here, stating that the generated thresholds per domain is always empty.

crates/contract/src/lib.rs
```diff
@@ -1604,7 +1604,11 @@ impl MpcContract {
                 )
                 .expect("Require valid threshold parameters"); // this should never happen.
                 current_params.validate_incoming_proposal(&threshold_parameters)?;
-                let res = running_state.transition_to_resharing_no_checks(&threshold_parameters);
+                // TEE-driven resharing only changes the participant set, so the
+                // per-domain reconstruction-threshold updates map is empty.
```
kevindeforth: I don't understand this comment. What is "TEE-driven resharing"?

crates/contract/src/lib.rs
kevindeforth, line 1: **Blocker**: I see that we don't have a unit test for verifying that per-domain thresholds are accepted and applied on the contract-level. We only unit test for having same threshold over all domains.

crates/contract/src/primitives/domain.rs
```diff
@@ -69,6 +71,30 @@ pub fn validate_domain_threshold(
     Ok(())
 }
 
+/// TODO(#3306): Enforces the 3.11-transition lock: every CaitSith domain (ForeignTx
+/// included) must share a single `reconstruction_threshold` so the legacy
+/// unprefixed `DBCol::Triple` mirror (#3292) can't collide. Domains using
+/// other protocols are unconstrained. The first CaitSith domain encountered
+/// fixes the expected value; any later CaitSith domain with a different
+/// threshold is rejected. Remove once #3298 drops the mirror.
```
kevindeforth: Is this still relevant? #3298 has been merged.

crates/contract/src/primitives/domain.rs
```diff
@@ -177,6 +203,50 @@ impl DomainRegistry {
...
+            }
+        }
+        let domains: Vec<DomainConfig> = self
+            .domains
+            .iter()
+            .map(|d| {
+                let reconstruction_threshold = threshold_updates
+                    .get(&d.id)
+                    .copied()
+                    .unwrap_or(d.reconstruction_threshold);
```
kevindeforth, line 233: Hmm, so non-specified domains default to the existing threshold, not to the proposed governance threshold.
That seems counter-intuitive.

crates/contract/src/lib.rs
```diff
@@ -873,10 +873,10 @@ impl MpcContract {
     pub fn vote_new_parameters(
```
kevindeforth, line 876: The docstring of this method needs to be updated.

crates/near-mpc-contract-interface/src/types/state.rs
SimonRastikian: After thinking about this, even though it would introduce  non-negligible machinery, I kinda agree with you. One could define a Compat struct and get done with this especially that these types are used by  signers and not users

crates/contract/src/v3_11_2_state.rs
kevindeforth: Yeah, lets just delete the votes and save us the headache. It's a minor inconvenience, if at all.

kevindeforth (commented):
That's a big one. Thank you very much @SimonRastikian that's a lot of work!

I did a first pass and it looks correct, but it's a non-trivial PR, so I want to spend a bit more time with it.

The main blocker I have right now is that we need to add unit and integration tests for this (in `contract/src/lib.rs`, our in-process and our sandbox tests).

As a follow-up, I would like to see a e2e test where we actually use the different reconstruction thresholds.

Regarding the API, I understand that if the user does not specify a domain threshold, the _existing_ domain threshold will be applied. I would have found it somewhat more intuitive if the governance threshold would be applied instead.

I am not sure why we want to keep both `ThresholdParameters` and `ProposedThresholdParameters`, I think we only need one of these.

crates/near-mpc-contract-interface/src/types/state.rs
SimonRastikian: tried handling things here [166de83](https://github.com/near/mpc/pull/3326/commits/166de838dfd2862def7d7df96507fdf7a41e7e42)

crates/contract/src/state/running.rs
SimonRastikian: [2134cb2](https://github.com/near/mpc/pull/3326/commits/2134cb2665d44b2b4d76158e8ab310dd68a1acc0)

crates/contract/src/state/running.rs
SimonRastikian: Done

crates/contract/src/primitives/thresholds.rs
SimonRastikian: Is this enough you think? [73af9b0](https://github.com/near/mpc/pull/3326/commits/73af9b05e223e1092073ef9da6e23671ac7d1a5e)

crates/contract/src/state/resharing.rs
SimonRastikian: I do not get your comment

crates/contract/src/dto_mapping.rs
SimonRastikian: https://github.com/near/mpc/issues/3541

crates/contract/src/v3_11_2_state.rs
SimonRastikian: Ok now I drop pending votes [0f6fcc0](https://github.com/near/mpc/pull/3326/commits/0f6fcc06c427bf600a88a93103328657fbf08f5f)

crates/near-mpc-contract-interface/src/types/state.rs
SimonRastikian: FIxed in [f622eb5](https://github.com/near/mpc/pull/3326/commits/f622eb50ab7630222e7529ec5aaa978a1aac79fd) -- this and a little more

crates/contract/src/state/running.rs
SimonRastikian: [9edce36](https://github.com/near/mpc/pull/3326/commits/9edce36e8628b7af5529a0b4174f98b7db4667ac)

SimonRastikian:
> I am not sure why we want to keep both ThresholdParameters and ProposedThresholdParameters, I think we only need one of these.

@kevindeforth I think the answer to this statement is somewhere up where Reynaldo and Nodar thought it would be better to have them split. I think in the long run, non of this should exist when we make the "vote hashing" discussion we had in the past with different types of votes

crates/near-mpc-contract-interface/src/types/state.rs
SimonRastikian: IMO, the whole voting mechanism should change with the hash then open proposal we discussed about with @netrome

crates/near-mpc-contract-interface/src/types/state.rs
SimonRastikian: I have proposed a ProposedThresholdParametersCompat struct  in one of the following commits, would you be so kind to revisit this review and tell me now what you think?

crates/contract/src/primitives/thresholds.rs
SimonRastikian: [a38dc0e](https://github.com/near/mpc/pull/3326/commits/a38dc0ec03a06c8c58e479e10e62cde432a84c98)

crates/near-mpc-contract-interface/src/types/state.rs
SimonRastikian: [e7bfdaf](https://github.com/near/mpc/pull/3326/commits/e7bfdaf61a978f050b92ff56d506bda4b5efdadb)

crates/near-mpc-contract-interface/src/types/state.rs
SimonRastikian: Not sure here what would you snapshot?

crates/contract/src/state/test_utils.rs
SimonRastikian: [cdc1361](https://github.com/near/mpc/pull/3326/commits/cdc136182226c76b3ce7b6f49b8ed5beee1c714e)

crates/contract/src/lib.rs
SimonRastikian: Updated to this resharing: [547f5d0](https://github.com/near/mpc/pull/3326/commits/547f5d061e96908fe5b4f54e02309a4dc2d7d3aa)

crates/contract/src/lib.rs
SimonRastikian: [18973ce](https://github.com/near/mpc/pull/3326/commits/18973ce46ce9c41fd9449d4eae089d8cff7cafac)

crates/contract/src/lib.rs
SimonRastikian: I did add some tests check above, but then I realized maybe one more is missing with two domains [2687edd](https://github.com/near/mpc/pull/3326/commits/2687eddb61b9fd440f343b2d56548256236ddbf7)

crates/contract/src/primitives/domain.rs
SimonRastikian: Yup exactly, I think your intuition goes with the old behaviour: old resharing resets every domain's threshold but here it should no more be the case I guess

crates/contract/src/primitives/domain.rs
SimonRastikian: It will go away anyway in 3306 which should be soon

crates/contract/src/v3_11_2_state.rs
netrome: Given that this PR is large already as-is we could do this as a follow up as well.

crates/contract/src/primitives/domain.rs
```diff
@@ -69,6 +71,30 @@ pub fn validate_domain_threshold(
     Ok(())
 }
 
+/// TODO(#3306): Enforces the 3.11-transition lock: every CaitSith domain (ForeignTx
+/// included) must share a single `reconstruction_threshold` so the legacy
+/// unprefixed `DBCol::Triple` mirror (#3292) can't collide. Domains using
+/// other protocols are unconstrained. The first CaitSith domain encountered
+/// fixes the expected value; any later CaitSith domain with a different
+/// threshold is rejected. Remove once #3298 drops the mirror.
+pub fn validate_caitsith_uniform_threshold(domains: &[DomainConfig]) -> Result<(), Error> {
```
gilcu3, line 76: with the recent comment guidelines we should probably reduce this to
```suggestion
/// TODO(#3306): Enforces the 3.11-transition lock: every CaitSith domain (ForeignTx
/// included) must share a single `reconstruction_threshold`.
pub fn validate_caitsith_uniform_threshold(domains: &[DomainConfig]) -> Result<(), Error> {
```

crates/near-mpc-contract-interface/src/types/state.rs
netrome: Yeah, though I'm not sure if this type would vanish entirely. At some point we still need to transmit this information one way or another afaik.

netrome (commented):
Started reviewing, so far so good - will continue in 30 minutes

crates/contract/src/lib.rs
```diff
@@ -1604,7 +1606,11 @@ impl MpcContract {
...
                 current_params.validate_incoming_proposal(&threshold_parameters)?;
-                let res = running_state.transition_to_resharing_no_checks(&threshold_parameters);
+                // This resharing only changes the participant set, so the
+                // per-domain reconstruction-threshold updates map is empty.
+                let proposed_parameters =
+                    ProposedThresholdParameters::new(threshold_parameters, BTreeMap::new());
+                let res = running_state.transition_to_resharing_no_checks(&proposed_parameters);
                 if let Some(resharing) = res {
                     self.protocol_state = ProtocolContractState::Resharing(resharing);
                 }
```
gilcu3, line 1678: I think we have a big problem here. This is a resharing that is triggered automatic. Just keeping the thresholds as before might not work at all

gilcu3 (commented):
Partial review, but found a problematic situation that I had not noticed before, we probably need to sync on this

crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
```diff
@@ -4579,7 +4610,7 @@ expression: abi
           "minimum": 0.0
         },
         "ThresholdParameters": {
-          "description": "Threshold parameters for distributed key operations.",
+          "description": "Threshold parameters for distributed key operations: the current participant set and the governance threshold. This is the stored, always-current shape; per-domain reconstruction-threshold *proposals* live on [`ProposedThresholdParameters`].",
```
gilcu3, line 4613: I cannot make sense of this comment, specially "This is the stored, always-current shape"

crates/contract/src/state/running.rs
```diff
@@ -132,24 +142,54 @@ impl RunningContractState {
...
+        })?;
+        let threshold_updates = proposal.per_domain_thresholds();
+        // Reject unknown domain IDs: the loop below iterates existing domains, so
+        // an unknown ID would otherwise be silently ignored here (it's caught at
+        // the resharing transition, but we fail fast at vote acceptance).
+        for id in threshold_updates.keys() {
+            if self.domains.get_domain_by_domain_id(*id).is_none() {
+                return Err(DomainError::UnknownDomainInProposal { domain_id: *id }.into());
+            }
+        }
```
netrome, line 166: Nit/question: Did you consider adding this check to `validate_incoming_proposal`?

crates/contract/src/lib.rs
SimonRastikian: [d3ed117](https://github.com/near/mpc/pull/3326/commits/d3ed1173d3652fd96fe7f04d515291cca6219c24) a [78b841a](https://github.com/near/mpc/pull/3326/commits/78b841a4236e31a4dcceaced22efd49da17b70d0)

crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -141,6 +144,79 @@ pub struct ThresholdParameters {
...
+// carries none of it. borsh and schemars are intentionally NOT derived — the
+// bridge is serde-only, and the ABI schema is generated from the public type's
+// own fields instead.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+struct ProposedThresholdParametersCompat {
+    #[serde(flatten)]
+    parameters: ThresholdParameters,
+    #[serde(default)]
+    per_domain_thresholds: BTreeMap<DomainId, ReconstructionThreshold>,
+}
```
netrome: Why do we need wire format backwards compatibility for the votes? This is something we manually coordinate, so changing the command after we've upgraded the contract shouldn't be any problem.

crates/contract/src/lib.rs
```diff
@@ -927,6 +932,46 @@ impl MpcContract {
...
+
+        let domains = self.protocol_state.domain_registry()?;
+        let updates = proposal.per_domain_thresholds();
+        for domain in domains.domains() {
+            let effective = updates
+                .get(&domain.id)
+                .copied()
+                .unwrap_or(domain.reconstruction_threshold)
+                .inner();
+            if effective > num_participants {
```
SimonRastikian, line 965: I decided to compare only reconstruction_threshold with num_participants here (not reconstruction_threshold with gov_threshold) because the latter is not even an enforcement in validate_domain_threshold which would be very inconsistent right now.
The whole correlation will be handled in #3499.

crates/contract/src/lib.rs
```diff
@@ -927,6 +932,46 @@ impl MpcContract {
...
+        let num_participants = proposal.participants().len() as u64;
+
+        let threshold = proposal.threshold().value();
+        if threshold > num_participants {
+            return Err(InvalidThreshold::MaxRequirementFailed {
+                max: num_participants,
+                found: threshold,
+            }
+            .into());
+        }
```
SimonRastikian, line 955: Making sure gov_threshold <= num_participants

crates/contract/src/lib.rs
```diff
@@ -925,6 +932,46 @@ impl MpcContract {
         }
     }
 
+    /// Defense-in-depth guard for [`Self::vote_new_parameters`]: rejects a
+    /// proposal whose participant set is smaller than the proposed signing
+    /// threshold or any domain's effective reconstruction threshold (proposed
+    /// override if present, else the domain's current value). Such a set would
+    /// leave a key un-signable or un-reconstructible. Redundant with
+    /// `RunningContractState::process_new_parameters_proposal`.
+    fn assert_proposal_meets_all_thresholds(
```
netrome, line 941: Nice, thanks for adding

crates/contract/src/lib.rs
```diff
@@ -925,6 +932,46 @@ impl MpcContract {
...
+        &self,
+        proposal: &ProposedThresholdParameters,
+    ) -> Result<(), Error> {
+        let num_participants = proposal.participants().len() as u64;
```
netrome: Classic nit:

```suggestion
        let num_participants = u64::try_from(proposal.participants().len()).expect("participant list should be wayyyy smaller than u64::MAX");
```

netrome (dismissed):
Nice stuff. No hard blockers but I think we could simplify the PR by removing the backwards compatibility wiring for the parameter votes - unless I missed something.

Note that we probably need to update some docs/examples when doing this, but Claude should be able to search and find these easily.

crates/contract/src/lib.rs
SimonRastikian: I added the check for both the tee and the vote_update [78b841a](https://github.com/near/mpc/pull/3326/commits/78b841a4236e31a4dcceaced22efd49da17b70d0) [d3ed117](https://github.com/near/mpc/pull/3326/commits/d3ed1173d3652fd96fe7f04d515291cca6219c24)

crates/contract/src/lib.rs
SimonRastikian: [058bb41](https://github.com/near/mpc/pull/3326/commits/058bb41ffd042090dd5d8365fc3f6d04fcd2defe)

crates/contract/src/state/running.rs
SimonRastikian: I do not think it should be added there. It could be that the proposal of adding a participant gets voted on before removing a participant. I do not want things to happen dynamically as we could potentially(?) end up discarding proposals that are not too wrong

crates/near-mpc-contract-interface/src/types/state.rs
SimonRastikian: Unless you want to discard existing proposals then I guess this is essential no?

netrome (dismissed):
Thanks for updating and adding the kickout test

gilcu3 (approved):
Thank you for all the hard work on this one!

netrome (approved):
Thanks for updating

PR #3327, opened 2026-05-22
feat: add foreign-chain whitelist verifier

### LINKED ISSUE #3324: Have nodes verify their configuration matches whitelisted RPC providers in the contract
### Background

We are implementing voting for whitelisting RPC providers in the contract. After that's implemented we should add checks in the node to validate that RPC providers from the config are whitelisted.

### User Story

_No response_

### Acceptance Criteria

Have nodes verify that RPC providers from the config is whitelisted.

This could be initially just error logs while still using RPC providers from config, and as we vote in RPC providers  we can start enforcing them in following release.

_No response_

anodar:
Closes: https://github.com/near/mpc/issues/3324


anodar:
@claude review

crates/node/src/foreign_chain_whitelist_verifier.rs
```diff
@@ -0,0 +1,1132 @@
...
+            DedupOutcome::Matching
+        } else {
+            DedupOutcome::Changed(current.clone())
+        };
+        self.previous = Some(current);
+        action
+    }
+}
+
+pub(crate) async fn run(indexer_state: Arc<IndexerState>, local: ForeignChainsConfig) {
```
netrome: Nit: This function seems like the main entrypoint to this module. I'd find it easier to read if it was declared first.

crates/node/src/foreign_chain_whitelist_verifier.rs
```diff
@@ -0,0 +1,1132 @@
...
+    let mut ticker = tokio::time::interval(VERIFY_INTERVAL);
+    let mut dedup = Dedup::new();
+    loop {
+        ticker.tick().await;
+        let whitelist = match indexer_state
+            .view_client()
+            .get_allowed_foreign_chain_providers(indexer_state.mpc_contract_id().clone())
+            .await
```
netrome: Hmm, there's no good way to subscribe to the latest foreign chain providers? A watch channel sounds like it would be ideal here. I'm not a fan of polling loops.

crates/node/src/foreign_chain_whitelist_verifier.rs
```diff
@@ -0,0 +1,1132 @@
...
+
+    fn observe(&mut self, current: Vec<Diagnostic>) -> DedupOutcome {
+        let action = if self.previous.as_ref() == Some(&current) {
+            DedupOutcome::Unchanged
+        } else if current.is_empty() {
+            DedupOutcome::Matching
+        } else {
+            DedupOutcome::Changed(current.clone())
+        };
+        self.previous = Some(current);
```
netrome: `Dedup` feels like a weird name. This type doesn't seem to deduplicate entries in the vector but rather emits the whole vector if anything has changed.

From the current structure it seems like it would make more sense to take a reference to a slice as input and emit `Changed` without any contained data. Or what is your thought about this type?

crates/node/src/foreign_chain_whitelist_verifier.rs
```diff
@@ -0,0 +1,1132 @@
...
+
+fn auth_scheme_variant_name(scheme: &AuthScheme) -> &'static str {
+    match scheme {
+        AuthScheme::None => "None",
+        AuthScheme::Header { .. } => "Header",
+        AuthScheme::Path { .. } => "Path",
+        AuthScheme::Query { .. } => "Query",
+        _ => "Unknown",
+    }
+}
```
netrome, line 239: What's the purpose of stringifying these variants?

netrome (dismissed):
The code seems functionally correct, but there's room for some improvements. I'm curious if we couldn't make this module assume a `tokio::watch` channel or equivalent to subscribe to the latest chain state instead of having a polling loop. Iirc, the new chain gateway will provide a more subscriber friendly interface and, if this doesn't work well with the old one we can always add an adapter outside of this module to keep the main run function here cleaner and easier to test.

Beyond that I have a few more nits and considerations, but nothing strictly blocking. Haven't reviewed particularly deep.

crates/node/src/foreign_chain_whitelist_verifier.rs
DSharifi: I assume it's to avoid leaking the keys

crates/node/src/foreign_chain_whitelist_verifier.rs
DSharifi: One of the many things I'm happy to have learned from you @netrome ! I'm still not bought in on the `use::` statements at the bottom, but love the push for function ordering

crates/node/src/foreign_chain_whitelist_verifier.rs
DSharifi: I agree. A watcher here would be sufficient.

DSharifi (dismissed):

crates/node/src/foreign_chain_whitelist_verifier.rs
anodar: Moved it to the top.

crates/node/src/foreign_chain_whitelist_verifier.rs
anodar: I changed it to follow patterns of other watchers (e.g. monitor_allowed_hashes). 

Although we're not actually getting rid of polling loop rather outsourcing it to the whatever sends in that channel.

crates/node/src/foreign_chain_whitelist_verifier.rs
anodar: Yes it's to avoid leaking secrets in `Diagnostic`  logs.

crates/node/src/foreign_chain_whitelist_verifier.rs
anodar: Not applicable anymore as it's dropped now (following pattern in monitor_allowed_hashes).

DSharifi (approved):

netrome (approved):
Thanks for updating 🙏

crates/node/src/foreign_chain_whitelist_verifier.rs
netrome: Yep but outsourcing is strictly better since it can be removed later without touching the inner workings of this module.

crates/node/src/foreign_chain_whitelist_verifier.rs
netrome: > I'm still not bought in on the use:: statements at the bottom, but love the push for function ordering

Lovely to hear that. I felt the same from the engineer I learned it from.

Use statements on the bottom is :fire: though. But I've unfortunately deviated from it in later side projects since it tends to confuse LLMs a bit more :melting_face:

anodar:
Thanks both! 🙏

PR #3330, opened 2026-05-22
refactor: share blocks tracker instance between queues

### LINKED ISSUE #3316: share one `RecentBlocksTracker` instance across all queues

kevindeforth:
resolves #3316


crates/node/src/types.rs
```diff
@@ -12,18 +12,20 @@ use crate::{
...
-    requests::recent_blocks_tracker::BlockViewLite,
+    requests::recent_blocks_tracker::{AtomicBlockStatus, BlockViewLite},
 };
 
 pub(crate) struct RequestsUpdate<T> {
-    pub(crate) block: BlockViewLite,
     pub(crate) requests: Vec<T>,
     pub(crate) completed_requests: Vec<RequestId>,
+    pub(crate) block_height: u64,
+    pub(crate) block_ref: Weak<AtomicBlockStatus>,
```
gilcu3, line 22: nit: having a small type for `block_height` and `block_ref` and keeping it as the first field seems nicer

gilcu3 (approved):
Thank you!

crates/node/src/types.rs
kevindeforth: Yeah, we will have both shortly. BlockRef will be resolved with #3451, block height we will have once we use the chain-gateway types.

crates/node/src/requests/metrics.rs
```diff
@@ -26,28 +26,6 @@ pub static MPC_PENDING_VERIFY_FOREIGN_TX_QUEUE_ATTEMPTS_GENERATED_TOTAL: LazyLoc
         )
         .unwrap()
 });
-// TODO(#3316): remove — collapses into a single shared block-stream counter once the
-// tracker is shared across queues. https://github.com/near/mpc/issues/3316
-pub static MPC_PENDING_VERIFY_FOREIGN_TXS_QUEUE_BLOCKS_INDEXED_TOTAL: LazyLock<
-    prometheus::IntCounter,
-> = LazyLock::new(|| {
-    prometheus::register_int_counter!(
-        "mpc_pending_verify_foreign_txs_queue_blocks_indexed_total",
```
anodar, line 35: optional nit: we don't seem to be using these metrics as far as I can tell, but probably worthy of noting in the PR description that we're getting rid of it (in favor of mpc_blocks_indexed_total/mpc_finalized_blocks_indexed_total)

anodar (approved):
Thank you!

PR #3339, opened 2026-05-25
docs: documenting eddsa with presign

SimonRastikian:
Closes #https://github.com/near/threshold-signatures/issues/329

I also updated a little bit the API of presigning to the strict necessary. This is not a breaking change as this is not really linked currently to the node


barakeinav1:
@copilot review (review only, do not make any code changes)

crates/threshold-signatures/benches/bench_utils/frost_eddsa.rs
```diff
@@ -33,7 +33,7 @@ pub fn ed25519_build_presign_protocols<R: CryptoRngCore + SeedableRng + Send + '
             participants,
             *p,
             &eddsa::PresignArguments {
-                keygen_out: keygen_out.clone(),
+                private_share: keygen_out.private_share,
```
gilcu3: it looks a bit weird to update this in a `docs` PR

crates/threshold-signatures/src/frost/presign.rs
gilcu3, line 1: Does this change belong to this PR? I cannot understand why was this file added here, this should only entail docs changes

crates/threshold-signatures/src/frost.rs
```diff
@@ -24,8 +24,8 @@ pub mod redjubjub;
 
 /// The necessary inputs for the creation of a presignature.
 pub struct PresignArguments<C: Ciphersuite> {
-    /// The output of key generation, i.e. our share of the secret key, and the public key package.
-    pub keygen_out: KeygenOutput<C>,
+    /// The output of key generation, i.e. our share of the secret key.
+    pub private_share: SigningShare<C>,
```
gilcu3: how is it possible that changing this does not require changes in the node? Should this `struct` be internal then?

gilcu3 (changes requested):
Partial review. I cannot understand why file `crates/threshold-signatures/src/frost/presign.rs` was added here. It is not mentioned in the issue, therefore I think it should be removed. I can re-review again once that is done.

crates/threshold-signatures/src/frost/presign.rs
SimonRastikian: Orphan file introduced in cargo fmt. will be deleted

crates/threshold-signatures/src/frost.rs
SimonRastikian: I changed the sign_v2 stuff which is not used in the smart contract or the node currently, this is why it did not introduce changes there. The goal is to push for v2 to be used

gilcu3:
@claude review

Focus on the alignment between the eddsa docs and the comments on the implementation

crates/threshold-signatures/benches/bench_utils/frost_eddsa.rs
SimonRastikian: Done no more updated

crates/threshold-signatures/docs/eddsa/signing.md
```diff
@@ -93,20 +95,133 @@ $$
...
+
+
+## Implementation Version 2: Signing
+
+In this phase, the same set of parties $\mathcal{P}_2 = \mathcal{P}_1$
+wishes to generate an EdDSA signature. Following the
+[RFC9591](https://datatracker.ietf.org/doc/html/rfc9591) we will use
+domain separated hash functions $H_1, H_2, H_4$.
+
+*Note: The signer set $\mathcal{P}_2$ must be the same set used at presigning time; the implementation does not validate this and signing will fail (signature aggregation or commitment lookup) if the sets diverge.*
```
gilcu3: We already mentioned that P2 = P1, so we should avoid repeating that in this sentence. Something like:


```suggestion
*Note: If the signer set $\mathcal{P}_2$ is not the same set used at presigning time $\mathcal{P}_1$; then signing will fail (signature aggregation or commitment lookup).*
```

crates/threshold-signatures/docs/eddsa/signing.md
```diff
@@ -93,20 +95,133 @@ $$
...
+
+$$
+s\gets \sum_j s_j
+$$
+
+1.5 $\blacktriangle$ The coordinator asserts that $(R, s)$ is a valid EdDSA signature for message $m$
+
+
+
+*Note: We do not make use of the cheater detection feature which requires additional computation and potentially an extra round of communicating the cheater to the rest of the participants.*
```
gilcu3:
```suggestion
*Note: We do not make use of the cheater detection feature which requires additional computation and potentially an extra round to communicate the cheater to the rest of the participants.*
```

crates/threshold-signatures/src/frost/eddsa/sign.rs
```diff
@@ -118,24 +118,26 @@ async fn do_sign_coordinator_v1(
...
+    // --- Round 2
+    // * Receive others' commitments, then send the signing package.
 
-    // Step 1.4
+    let commit_waitpoint = chan.next_waitpoint();
+    // Step 2.1
     for (from, commitment) in recv_from_others(&chan, commit_waitpoint, &participants, me).await? {
+        // Step 2.2
         commitments_map.insert(from.to_identifier()?, commitment);
     }
```
gilcu3, line 148: Seems more natural:


```suggestion
    let commit_waitpoint = chan.next_waitpoint();
    // Step 2.1 and 2.2
for (from, commitment) in recv_from_others(&chan, commit_waitpoint, &participants, me).await? {
commitments_map.insert(from.to_identifier()?, commitment);
}
```

crates/threshold-signatures/src/frost/eddsa/sign.rs
```diff
@@ -203,6 +205,9 @@ async fn do_sign_coordinator_v2(
     message: Vec<u8>,
 ) -> Result<SignatureOption, ProtocolError> {
     // --- Round 1
+    // * Compute my signature share.
+    // * Receive others' signature shares.
+    // * Output the signature.
```
gilcu3: feels a bit weird to add this explanation here for round 1, while other comments do not have them. Better just remove

gilcu3 (dismissed):
Thank you!

Left some minor comments for improvement

crates/threshold-signatures/src/frost/eddsa/sign.rs
netrome: Alternatively add similar explanations to other round comments for consistency. Not sure if these are overly verbose or helpful. Though generally, whenever you need to comment code like "// Compute my signature share" it's worth questioning if you can break out that code to a function: `let res = compute_my_signature_share(params)`

crates/threshold-signatures/docs/eddsa/signing.md
```diff
@@ -10,7 +10,9 @@ of detecting which party deviated from the protocol.
 
 ### Note:  We denote $\mathcal{P}$ the set of participants included the DKG and the threshold $t = \mathsf{MaxMalicious}+1$
 
-## Signing
+### Note: We introduce two signing versions both formally proven secure in the original paper. The first version performs the entirety of FROST protocol during the signing. It relies heavily on the coordinator and thus takes 3 non-symmetric rounds to complete. The second version has been implemented later on, it splits the FROST scheme into a presigning round and a signing round. This relies less on the coordinator in the presigning phase by treating all participants symmetrically. It thus completes in two rounds.
```
netrome: This is a very long header. I'd consider making it a blockquote:
```suggestion
> **Note:** We introduce two signing versions both formally proven secure in the original paper. The first version performs the entirety of FROST protocol during the signing. It relies heavily on the coordinator and thus takes 3 non-symmetric rounds to complete. The second version has been implemented later on, it splits the FROST scheme into a presigning round and a signing round. This relies less on the coordinator in the presigning phase by treating all participants symmetrically. It thus completes in two rounds.
```

crates/threshold-signatures/docs/eddsa/signing.md
```diff
@@ -49,21 +51,21 @@ $$
...
+### Round 3
 
-2.1 $\bullet$ Each $P_i$ waits to receive $(\mathit{commits}, m^*)$ sent by the coordinator.
+3.1 $\bullet$ Each $P_i$ waits to receive $(\mathit{commits}, m^*)$ sent by the coordinator.
 
-2.2 Each $P_i$ verifies that $m = m^*$
+3.2 Each $P_i$ verifies that $m = m^*$
 
-2.3 Each $P_i$ computes a signature share using following [RFC9591](https://datatracker.ietf.org/doc/html/rfc9591#name-round-two-signature-share-g).
+3.3 Each $P_i$ computes a signature share using following [RFC9591](https://datatracker.ietf.org/doc/html/rfc9591#name-round-two-signature-share-g).
```
netrome: Typo (not introduced in this PR though)

```suggestion
3.3 Each $P_i$ computes a signature share following [RFC9591](https://datatracker.ietf.org/doc/html/rfc9591#name-round-two-signature-share-g).
```

crates/threshold-signatures/docs/eddsa/signing.md
```diff
@@ -93,20 +95,133 @@ $$
...
+* Compute the following binding and hiding points:
+
+$$
+\begin{aligned}
+A_i&\gets a_i \cdot G\cr
+B_i &\gets b_i \cdot G
+\end{aligned}
+$$
+
+1.2 $\star$ Each $P_i$ sends $(A_i, B_i)$ **to every party**.
```
netrome, line 151: Can you refresh my memory on why we don't do this in the old implementation? I.e. why are all nonce commitments distributed through the coordinator in that implementation?

crates/threshold-signatures/docs/eddsa/signing.md
netrome: Rendering bug

Image

netrome (changes requested):
Nice stuff, just a rendering problem that should be fixed.

crates/threshold-signatures/docs/eddsa/signing.md
SimonRastikian: It's because in the old implementation I followed Zcash who themselves followed the main part of the paper not the note in appendix. IMO: the note in the appendix is the most important for the efficiency and it affects absolutely not at all the security

SimonRastikian: This kind of design is good if you do not want to reveal the participants. Each participant does not know the others so rely on a single coordinator that forwards the messages to the right party

crates/threshold-signatures/src/frost/eddsa/sign.rs
SimonRastikian: I relaxed maybe the explanation. I went through all the // --- Round  in eddsa implementation (related to this PR) and all of them have similar explanation. So this seems consistent.

netrome (approved):
Looks good now, thanks for updating

gilcu3 (approved):

Issue #3340, opened 2026-05-25
Cleanup the threshold signature crate


SimonRastikian:
### Background

The threshold signature crate is a little messy currently. 

More concretely frost.rs, ecdsa.rs, test_utils.rs have accumulated logic that belongs in their submodules, and the crate's visibility is too loose (almost nothing is pub(crate)).

Additionally, tests/ are supposedly some integration tests but most (if not all) of the tests are redundant inside the crate and shall be deleted

### Acceptance Criteria

  - frost.rs — move misplaced functions (e.g. presign, assert_sign_inputs) into appropriate submodules under frost/; keep only module declarations and types that genuinely belong at the root.
  - ecdsa.rs — same: relocate logic that doesn't belong at the module root into ecdsa/ submodules.
  - test_utils.rs — tighten visibility (pub(crate) where possible) so test helpers stop forming part of the public API, and move helpers into the relevant submodules.
  - tests/ — keep it. These are intentional integration tests of the public API. Only remove a test if it is provably redundant and not exercising the public surface; otherwise leave as-is.

netrome:
tests/ contains integration tests, see https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html - what's the problem with that?

gilcu3:
adding a bit to the previous comment

> frost.rs and ecdsa.rs files should not exist

why? If you mean they should contain only sub-module definitions, then that might be, but what's wrong with keeping the most important definitions of the rust module there? Did we change some guidelines regarding this?

SimonRastikian:
@netrome since we do not use pub(crate) in threshold signatures then all functions in the files tests.rs could be considered as integration testings.
Additionally, I argue that the tests in tests/ are absolutely useless as they are redundant to some tests in each of the corresponding tests.rs.

@gilcu3 as for the frost.rs, I originally thought I can replace things with mod.rs but this is not the case as clippy yells. Independently it needs a rework as it contains functions that should not belong there but somewhere else (check the corresponding PR)

gilcu3:
> [@netrome](https://github.com/netrome) since we do not use pub(crate) in threshold signatures then all functions in the files tests.rs could be considered as integration testings. Additionally, I argue that the tests in tests/ are absolutely useless as they are redundant to some tests in each of the corresponding tests.rs.

[...]

PR #3343, opened 2026-05-25
refactor: cleanup frost interface

SimonRastikian:
Deals with part of #3340 namely the one related to frost


netrome (dismissed):
Quick look, looks like a pure refactor - thanks 🙏

crates/threshold-signatures/src/frost/presign.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// The secret signing nonces.
+    pub nonces: SigningNonces<C>,
+    #[zeroize(skip)]
+    pub commitments_map: BTreeMap<Identifier<C>, SigningCommitments<C>>,
+}
+
+impl_secret_debug!({C: Ciphersuite} PresignOutput<C> { show: [commitments_map], redact: [nonces] });
+
+/// Maximum incoming buffer entries for the FROST presign protocol.
+pub const FROST_PRESIGN_MAX_INCOMING_BUFFER_ENTRIES: usize = 1;
```
barakeinav1: nit: This is only used within this module now, so it can be private.

```suggestion
const FROST_PRESIGN_MAX_INCOMING_BUFFER_ENTRIES: usize = 1;
```

crates/threshold-signatures/src/frost/presign.rs
```diff
@@ -0,0 +1,164 @@
...
+
+/// Runs Presigning of either `EdDSA` or `RedDSA`
+pub fn presign<C, R>(
```
barakeinav1, line 48: There's an opportunity for some cleanup here: this function's participant/threshold validation (`len < 2`, `ParticipantList::new`, `contains(me)`, threshold check) largely duplicates `assert_sign_inputs` in `sign_utils.rs`, so the shared checks could be factored out. Better as a follow-up issue.

barakeinav1 (dismissed):
LGTM, approved with 2 comments

crates/threshold-signatures/src/frost/presign.rs
SimonRastikian: Agreed I will open an issue for it

netrome (approved):

barakeinav1 (approved):

PR #3350, opened 2026-05-25
feat: Enforcing strong typing and one explicit threshold: ReconstructionThr…

### LINKED ISSUE #348: Unify the threshold type used in all algorithms
### Background

Historically we've used the term `threshold` inconsistently, sometimes referring to the max number of malicious parties (max malicious) and sometimes referring to the amount of parties needed to reconstruct the private key (reconstruction bound).

We've already introduced strong types in the code for this, but some algorithms takes a reconstruction bound as input and some takes max malicious as input.

It would be nice to have all algorithms use the same threshold type.

### User Story

As a crypto developer, I'd like us to use consistent threshold terms to make it easier to work with different algorithms with minimal context-switching overhead.

### Acceptance Criteria

We consistently use either max malicious or the reconstruction bound as input to all signing and DKG algorithms.

Note: We want to update the MPC nodes to use the strong types before changing the interface, to ensure this doesn't break the nodes in surprising ways.

SimonRastikian:
Closes https://github.com/near/threshold-signatures/issues/348
Our long lived issue :)


crates/node/src/indexer/fake.rs
```diff
@@ -418,7 +418,11 @@ fn participants_config_to_threshold_parameters(
             )
             .expect("Failed to insert participant");
     }
-    ThresholdParameters::new(participants, Threshold::new(participants_config.threshold)).unwrap()
+    ThresholdParameters::new(
+        participants,
+        Threshold::new(participants_config.threshold.inner()),
```
netrome, line 423: Hmm, this type has not been renamed. It feels weird to just cast the reconstruction threshold to this other type here. Shouldn't this also be the `ReconstructionThreshold` so we don't have any ambiguous "threshold" concepts remaining? Long term I guess this should become the governance threshold, but if this is the case, we should already have the governance threshold in `participants_config` imo where we now have `ReconstructionThreshold`.

So I suggest one of two things:

1. Use `ReconstructionThreshold` everywhere.
2. Rename the existing primitive`Threshold` type used in the contract for voting operations to `GovernanceThreshold` and separate the governance and reconstruction thresholds.

The second option seems like it deserves its own issue, so I'd lean on 1. here.

crates/node/src/config.rs
```diff
@@ -62,11 +63,24 @@ impl MpcConfig {
 
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub struct ParticipantsConfig {
-    /// The threshold for the MPC protocol.
-    pub threshold: u64,
+    /// The threshold for the MPC protocol — number of share-holders required
+    /// to reconstruct the secret. Wire format is a plain integer (the inner
+    /// type is `#[serde(transparent)]`), so this is wire-compatible with the
+    /// previous `u64` field.
+    pub threshold: ReconstructionThreshold,
```
netrome, line 70: Isn't this the governance threshold now? Do we want to have a separate notion between governance and reconstruction threshold? Otherwise, could we get rid of the existing `Threshold` type in our primitives crate favor of just having the `ReconstructionThreshold` everywhere?

netrome (changes requested):
Nice stuff, though it seems like there are some ambiguous thresholds remaining. Gotta catch em all - or what are your thoughts here?

SimonRastikian:
This will be replaced by a series of smaller more compact PRs that better affect the implementation. See #3571

PR #3321, opened 2026-05-21
chore: bump nearcore to 2.12 rc 1, bumps rust to 1.93

gilcu3:
Closes #3282 and #972

PR #3354, opened 2026-05-26
chore: unify sanbox version across workspace

### LINKED ISSUE #3353: Pin sandbox version in contract tests, mirroring e2e-tests
### Background

`crates/contract/tests/**` calls `near_workspaces::sandbox()` (e.g. `crates/contract/tests/sandbox/common.rs:105`), which has no version arg and falls through to upstream `near-sandbox`'s default.

The e2e-tests crate handles this explicitly: `crates/e2e-tests/src/cluster.rs:24` defines `DEFAULT_SANDBOX_VERSION` and passes it into `Sandbox::start_sandbox_with_config_and_version`. `near_workspaces` exposes the same knob as `near_workspaces::sandbox_with_version(version)`, so contract tests can be aligned the same way — keeping the sandbox binary in lockstep with the workspace's nearcore tag rather than whatever upstream `near-sandbox` happens to default to.

### Acceptance Criteria

- Introduce a sandbox version constant in `crates/contract/tests/` mirroring `e2e-tests/src/cluster.rs:24`.
- Replace each `near_workspaces::sandbox()` call in `crates/contract/tests/**` with `near_workspaces::sandbox_with_version(<constant>)` (9 call sites across `abi.rs`, `sandbox/common.rs`, `sandbox/participants_gas.rs`, `sandbox/upgrade_to_current_contract.rs`).
- Update `scripts/check-sandbox-image-version.sh` so it also validates the new contract-tests constant against the workspace's nearcore tag (today it only checks `DEFAULT_SANDBOX_VERSION` in `cluster.rs`).

gilcu3:
Closes #3353 

This is a premise to #3321

### DIFF
--- a/.github/scripts/bump-nearcore.sh
+++ b/.github/scripts/bump-nearcore.sh
@@ -36,11 +36,11 @@ git remote set-url origin "https://x-access-token:${GH_TOKEN}@github.com/${GITHU
 git push origin --delete "$BRANCH" 2>/dev/null || true
 
 sed -i "s|tag = \"$CURRENT_TAG\"|tag = \"$NEW_TAG\"|g" Cargo.toml
-sed -i "/DEFAULT_SANDBOX_VERSION/s|\"$CURRENT_TAG\"|\"$NEW_TAG\"|g" crates/e2e-tests/src/cluster.rs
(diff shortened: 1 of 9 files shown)


netrome (approved):

anodar (approved):
Thank you!

PR #3356, opened 2026-05-26
ci: use nix for contract reproducible builds

### LINKED ISSUE #3355: Switch mpc-contract reproducible build from cargo-near Docker to Nix
### Background

The reproducible WASM build for `mpc-contract` currently runs via `cargo near build reproducible-wasm`, which spawns the `sourcescan/cargo-near` Docker image declared in `crates/contract/Cargo.toml`'s `[package.metadata.near.reproducible_build]` block. That image's Rust toolchain is controlled upstream by sourcescan, so we cannot bump the in-image Rust on our own schedule — we have to wait for a new image to be published.

A Nix-based reproducible build for the contract already exists at `nix/mpc-contract.nix` (exposed as `packages.<system>.mpc-contract` in `flake.nix`). It produces a reproducible WASM via the Nix sandbox rather than the cargo-near Docker container, removing the upstream dependency.

Switching means the contract WASM will no longer carry sourcescan-compatible NEP-330 build metadata, so third-party verifiers (sourcescan.io, nearblocks) cannot automatically replay the build. Verification becomes "anyone with a checkout and Nix can rebuild and compare hashes". This trade-off needs to be documented for operators and contract consumers.

### Acceptance Criteria

- Replace the cargo-near Docker invocation with `nix build .#mpc-contract` in:
  - `mpc-contract-reproducible-build` job in `.github/workflows/ci.yml`
  - The "Build mpc-contract reproducibly" step of the `mpc-e2e-tests` job in `.github/workflows/ci.yml`
  - The `build-contract` job in `.github/workflows/release.yml` (also drop the now-unneeded `cargo-binstall` / `cargo-near` install steps and the `liblzma-dev`/`libudev-dev` apt installs)
- Update `scripts/check-contract-wasm-size.sh` callers to read from `result/mpc_contract.wasm`.
- Decide what to do with `[package.metadata.near.reproducible_build]` in `crates/contract/Cargo.toml` — either remove it (no embedded metadata) or rewrite it to describe the Nix derivation; pick whichever is least misleading to downstream verifiers.
- Add operator-facing docs explaining how to reproduce the contract WASM via Nix and compare the SHA-256 against the release artifact.

gilcu3:
Closes #3355 

Also a premise to #3321


.github/workflows/release.yml
gilcu3, line 1: I am not using the nix cache here on purpose, given that this is a release build. The downside is that making the release will take a few more minutes because of that. Lmk if you think it should be used.

.github/workflows/release.yml
```diff
@@ -113,46 +113,30 @@ jobs:
...
           echo "sha256=$CONTRACT_HASH" >> "$GITHUB_OUTPUT"
 
       - name: Package contract artifacts
         env:
           VERSION: ${{ needs.release-info.outputs.version }}
         run: |
-          cd target/near/mpc_contract/
-          mv mpc_contract.wasm "mpc-contract-v${VERSION}.wasm"
+          cp result/mpc_contract.wasm "mpc-contract-v${VERSION}.wasm"
           tar -czf "mpc-contract-v${VERSION}.tar.gz" "mpc-contract-v${VERSION}.wasm"
```
barakeinav1, line 136: Quick question — is the removal of `mpc_contract_abi.json` from the release archive intentional? The previous archive included it alongside the WASM; the Nix build still produces it (uses `--features abi`), but the derivation's `installPhase` only copies the WASM. If intentional, we should call it out in `RELEASES.md` since downstream consumers may rely on it.

gilcu3: yes we had removed it a while ago because we embed the abi in the contract

barakeinav1 (dismissed):
LGTM. One question on the ABI removal left inline.

.github/workflows/release.yml
DSharifi: What's the reason to not use the cache if it's a release build?

DSharifi (dismissed):
Approved, but I'm curious why we can't reuse/read the cache when making release builds

barakeinav1 (approved):

.github/workflows/release.yml
gilcu3: added it in [5b20bfc](https://github.com/near/mpc/pull/3356/commits/5b20bfc7480738799a7bcdc76a2aef6bdcbe553a)

DSharifi (approved):

PR #3359, opened 2026-05-26
refactor: presign args private share

SimonRastikian:
Tiny fix: modifies the api to the strictly necessary.
The fix does not impact the node as the scheme is not even integrated in the nodes.


netrome (approved):
Nice

barakeinav1 (approved):
LGTM

PR #3361, opened 2026-05-26
feat: support contract upgrades in batches

### LINKED ISSUE #2987: Two-part contract uploads
### Background

We're starting to face problems with contract size limitations. Nearcore has a limitation of 4MB on receipts and 1.5 MB on transactions. These were introduced in  [NEP-0509](https://github.com/near/NEPs/blob/425f4652dccf2a5dc33206408aba181a921fb459/neps/nep-0509.md#state-witness-size-limits), and implemented in nearcore in https://github.com/near/nearcore/pull/11701.

As our contract is close to 1.5 MB we cannot grow it further without hitting these limitations. The past weeks we've made a few optimizations to get some margin, but we're still close to the limit.

We've observed that we can make the contract significantly smaller by breaking out the TEE verification logic, which we'll prioritize for the coming months.

However, to ensure we're not constrained by this in the near future we can implement two-part uploads where we upload the contract in two transactions before the migration.

### User Story

As an MPC developer I want to be able to keep developing our planned features and ship a contract larger than 1.5 MB.

### Acceptance Criteria

The MPC contract can be uploaded in two parts so we can upload and migrate to a larger contract than 1.5 MB.

_No response_

DSharifi:
closes https://github.com/near/mpc/issues/2987


DSharifi:
@claude review

DSharifi:
@claude review

crates/contract/src/update.rs
```diff
@@ -87,32 +88,120 @@ pub enum Update {
     derive(schemars::JsonSchema, borsh::BorshSchema)
 )]
 pub struct ProposeUpdateArgs {
```
netrome: Since this is only used for config updates now, I'd suggest rename this to `ProposedConfigUpdateArgs`.

crates/contract/src/lib.rs
```diff
@@ -1222,6 +1234,198 @@ impl MpcContract {
...
+        }
+
+        let staged =
+            self.staged_uploads
+                .get_mut(&caller)
+                .ok_or(InvalidParameters::MalformedPayload {
+                    reason: "no staged upload found; call start_contract_upload first".into(),
+                })?;
+
+        let chunk_index = staged.record_chunk(chunk_len as u64)?;
```
netrome: Nit: I'm sure you'll miss my comments on `as` conversions in your next role ;)

```suggestion
        let chunk_index = staged.record_chunk(u64::try_from(chunk_len).expect("chunk length cannot be close to u64::MAX due to transaction size limitations"))?;
```

crates/contract/src/lib.rs
```diff
@@ -1222,6 +1234,198 @@ impl MpcContract {
...
+        }
+
+        let num_chunks = staged.num_chunks;
+        let total_size = staged.total_size;
+        let deposited = staged.deposited;
+
+        // Read (don't remove) the chunks first so the deposit can be validated before
+        // any storage is mutated: if the deposit is short we reinstate the staged entry
+        // and bail with the chunks still in place.
+        let mut code = Vec::with_capacity(total_size.get() as usize);
```
netrome:
```suggestion
        let mut code = Vec::with_capacity(usize::try_from(total_size.get()).expect("total size should fit in a usize on wasm32"));
```

crates/contract/src/lib.rs
```diff
@@ -1222,6 +1234,198 @@ impl MpcContract {
...
+        let mut staged = StagedContractUpload::new(args.total_size);
+        staged.deposited = env::attached_deposit();
+        self.staged_uploads.insert(caller, staged);
+
+        log!(
+            "start_contract_upload: signer={}, total_size={}",
+            env::signer_account_id(),
+            args.total_size,
+        );
+
```
netrome: What do we need this log for?

crates/contract/src/lib.rs
```diff
@@ -1222,6 +1234,198 @@ impl MpcContract {
...
+        if staged.deposited > NearToken::from_yoctonear(0) {
+            Promise::new(caller).transfer(staged.deposited).detach();
+        }
+
+        log!(
+            "clear_staged_contract: signer={}, num_chunks_cleared={}",
+            env::signer_account_id(),
+            staged.num_chunks,
+        );
```
netrome: Same here, curious why we need to log these in the contract. Do we need to pick up these logs somewhere?

netrome (dismissed):
Mostly looks good to me. Nothing strictly blocking, but a few things I would like to change:

1. I'd like to rename the existing `propose_update` method to `propose_config_update` as it no longer is used for anything else than config updates. This should be fine since it's a method we call.
2. I don't see the point of emitting logs in the new methods. I suggest skipping it (YAGNI).
3. There are some `as` conversion. You know my preference on these.

crates/contract/src/v3_10_state.rs
gilcu3, line 1: small blocker: This file cannot be modified until https://github.com/near/mpc/issues/3473 lands

crates/test-large-contract/build.rs
```diff
@@ -0,0 +1,26 @@
+//! Generates a non-compressible padding blob that `src/lib.rs` embeds at build time.
+//!
+//! The blob must be both large enough to push the assembled WASM above the RPC's
+//! ~1.5 MiB single-transaction payload limit (so chunked-upload tests actually
+//! exercise the chunked path) and incompressible enough that `wasm-opt`/`gzip`
+//! can't shrink it back below that limit. Pseudo-random bytes from a tiny LCG
+//! satisfy both at near-zero build cost.
```
gilcu3: I am not sure this is correct. Why do we care that gzip cannot put it back? Doesn't the limit apply to the binary size of the WASM contract?

crates/test-large-contract/build.rs
```diff
@@ -0,0 +1,26 @@
...
+    let out = Path::new("src/padding.bin");
+    if !out.exists() {
+        let mut padding = vec![0u8; PADDING_BYTES];
+        let mut state: u64 = 0xDEAD_BEEF;
+        for byte in padding.iter_mut() {
+            state = state.wrapping_mul(LCG_MULTIPLIER).wrapping_add(1);
+            *byte = (state >> 33) as u8;
+        }
+        fs::write(out, &padding).expect("failed to write padding.bin");
+    }
```
gilcu3, line 15: we could just use some rng but yeah this is also fine

scripts/check-contract-wasm-size.sh
```diff
@@ -16,7 +16,7 @@ WASM_PATH="${1:-result/mpc_contract.wasm}"
 # ~12 KB over the pre-feature baseline; this limit gives a bit of headroom
 # above the post-feature size without leaving the contract free to creep up
 # to the protocol boundary.
-HARD_LIMIT=1450000
+HARD_LIMIT=1520000
```
gilcu3: small blocker:
This is no longer needed. From the last CI in this PR:
https://github.com/near/mpc/actions/runs/26729451451/job/78795487883

we got:
```
Contract WASM size report
=========================
  Binary:     result/mpc_contract.wasm
  Size:       1458023 bytes
  Hard limit: 1520000 bytes
  Headroom:   61977 bytes
```

crates/contract/tests/sandbox/common.rs
```diff
@@ -337,16 +337,57 @@ impl SandboxTestSetupBuilder {
...
+        hash(&contract_binary_post_upgrade),
+        "Code hash post upgrade is not matching the proposed binary."
+    );
+}
+
+/// Upgrades the given contract to `new_contract_binary` using the **legacy inline**
+/// `propose_update` flow, where the whole code blob is sent in a single transaction.
+///
+/// This exists only to drive upgrades *from* a production contract that predates
+/// the chunked-upload endpoints (the tests pin the 3.10.1 binary): those binaries
```
gilcu3: the 3.11.2 contract will land before this is released, so these comments are a bit stale atm

crates/contract/tests/sandbox/common.rs
```diff
@@ -337,16 +337,57 @@ impl SandboxTestSetupBuilder {
...
     contract: &Contract,
     new_contract_binary: &[u8],
 ) {
     let propose_update_execution = accounts[0]
         .call(contract.id(), method_names::PROPOSE_UPDATE)
-        .args_borsh(ProposeUpdateArgs {
-            code: Some(new_contract_binary.to_vec()),
-            config: None,
-        })
+        .args_borsh((Some(new_contract_binary.to_vec()), Option::<()>::None))
```
gilcu3: isnt that code equivalent to the previous one? Or is this a patch because `ProposeUpdateArgs` changed?

crates/contract/tests/sandbox/common.rs
```diff
@@ -381,18 +413,102 @@ pub async fn propose_and_vote_contract_binary(
...
+    // which for multi-MiB binaries comfortably exceeds the `GAS_FOR_VOTE_UPDATE`
+    // budget used by the cheaper sub-threshold path.
     for voter in accounts {
         let execution = voter
             .call(contract.id(), method_names::VOTE_UPDATE)
             .args_json(serde_json::json!({
                 "id": proposal_id,
             }))
-            .gas(GAS_FOR_VOTE_UPDATE)
+            .max_gas()
```
gilcu3, line 507: still we shouldn't use max gas, but update the parameter `GAS_FOR_VOTE_UPDATE` so that we know how far are we from limits

crates/contract/tests/sandbox/update_votes_cleanup_after_resharing.rs
```diff
@@ -145,6 +139,102 @@ async fn update_votes_from_kicked_out_participants_are_cleared_after_resharing()
...
+        .call(contract.id(), method_names::START_CONTRACT_UPLOAD)
+        .args_borsh(StartContractUploadArgs {
+            total_size: std::num::NonZeroU64::new(chunk.len() as u64).unwrap(),
+        })
+        .max_gas()
+        .deposit(NearToken::from_yoctonear(1))
+        .transact()
+        .await?
+        .into_result()
+        .map_err(|e| anyhow::anyhow!("start_contract_upload failed: {e}"))?;
```
gilcu3, line 178: we should probably measure and set a gas constant here instead of max_gas

crates/contract/tests/sandbox/upgrade_from_current_contract.rs
```diff
@@ -529,12 +468,14 @@ async fn only_one_vote_from_participant() {
...
 
+    // This vote crosses the threshold and triggers deploy_contract + migrate
+    // on a multi-MiB binary; use max_gas so chunk reassembly + deploy fit.
     let execution = mpc_signer_accounts[1]
         .call(contract.id(), method_names::VOTE_UPDATE)
         .args_json(serde_json::json!({
             "id": proposal_b,
         }))
-        .gas(GAS_FOR_VOTE_UPDATE)
+        .max_gas()
```
gilcu3, line 479: as I said in some comments above, we should still track how much gas is that, by keeping the constant

crates/contract/tests/sandbox/upgrade_from_current_contract.rs
```diff
@@ -569,7 +510,409 @@ async fn update_from_current_contract_to_migration_contract() {
...
+        contract,
+        mpc_signer_accounts,
+        ..
+    } = SandboxTestSetup::builder()
+        .with_protocols(ALL_PROTOCOLS)
+        .build()
+        .await;
+
+    let code = current_contract();
+    assert!(code.len() > 1024, "contract binary should be non-trivial");
```
gilcu3: this is a strange restriction. Why would the test care about that?

crates/contract/src/lib.rs
```diff
@@ -1618,8 +1822,31 @@ impl MpcContract {
...
+        for (account, num_chunks, deposited) in orphaned {
+            for i in 0..num_chunks {
+                self.staged_chunks.remove(&(account.clone(), i));
+            }
+            self.staged_uploads.remove(&account);
+            if deposited > NearToken::from_yoctonear(0) {
+                Promise::new(account).transfer(deposited).detach();
+            }
+        }
+
```
gilcu3, line 1832: how are we ensuring that this does not consume more gas than we are paying for it?

crates/contract/src/lib.rs
gilcu3, line 1799: isn't this comment a bit misleading? It mentions online participants can call it, but it seems anyone can actually call it

gilcu3 (changes requested):
Thank you very much! Looks like it should work!

I have a few small blockers:
- contract size check (the bump seems no longer needed)
- the contract state migration assumes 3.10 (needs to be fixed by another PR)
- the lack of update_gas constants, which serve as an upper bound in tests.

I am also a bit concerned about the deposit logic, which does not seem trivial. Would be good to get that sorted out

I hoped we could do this change so that we could revert back to a single chunk easily when we get our contract size down, but the current implementation does not seem trivial to revert.

crates/contract/src/lib.rs
DSharifi: baa80a6c069d6dabc9fbb7e375c7b1480d52cb81

crates/contract/src/lib.rs
DSharifi: baa80a6c069d6dabc9fbb7e375c7b1480d52cb81

crates/contract/tests/sandbox/common.rs
DSharifi: I'll just change the comment to say current production contract

crates/contract/tests/sandbox/common.rs
DSharifi: It's because args changed, so this is inlining the borsh.. I'll add a legacy variant. so the code is easier to read.

crates/contract/src/update.rs
DSharifi: good spot. Changed the args and method name

crates/contract/src/lib.rs
DSharifi: Removed. Was meant for visibility if we wanted to debug.

crates/contract/src/lib.rs
DSharifi: No, they are just for visibility/debuggin

DSharifi: Removed them

scripts/check-contract-wasm-size.sh
DSharifi: reverted

crates/test-large-contract/build.rs
DSharifi: 9b23ea939a21cbf3b68e5d974364f6d8ed755ad1

crates/test-large-contract/build.rs
DSharifi: You're right, this is AI slop when I asked it to generate a big contract binary. Changed it to just embed some arbitrary bytes, which wasm-opt does not optimize away.

crates/contract/tests/sandbox/upgrade_from_current_contract.rs
DSharifi: Added constants

crates/contract/tests/sandbox/update_votes_cleanup_after_resharing.rs
DSharifi: Added constants

crates/contract/tests/sandbox/common.rs
DSharifi: Added constants

crates/contract/src/lib.rs
DSharifi: I'm not sure I follow what you mean here. We are only sending back a deposit the participant paid themselves.

And for the transfer cost, this should be part of the gas cost for whoever is calling the method?

crates/contract/tests/sandbox/upgrade_from_current_contract.rs
DSharifi: AI slop. Removed.

crates/contract/src/lib.rs
DSharifi: Yeah, seems like it is a stale comment. Since it's already stale on main, I'll update it in a separate PR/issue.

The fact that anyone can call this function doesn't matter for the code added.

DSharifi: https://github.com/near/mpc/issues/3489

crates/contract/src/lib.rs
gilcu3: What I meant here is that when we call this cleanup function `remove_non_participant_update_votes` via promise, we need to know how much gas it will spend. Unless we have some tests that measures it, I am not sure it can fit for a big contract which requires several upload parts. Specially this line:
```
self.staged_chunks.remove(&(account.clone(), i));
```
might cost a lot

netrome:
Pausing this for now. It seems like we might have enough margin after the Rust 1.93 upgrade to survive long enough to ship the attestation contract breakout. If not, we'll resurrect this PR.

PR #3362, opened 2026-05-26
fix(node): refresh attestation before concluding back-migration (#2121)

### LINKED ISSUE #2121: feat: enable migration of key-shares into a populated keyshare storage
### Background

[Latest testing](https://github.com/near/mpc-private/blob/main/misc/2026_2_15_migration_testing/mpc-migration-backward-issue.md)
 of the migration flow revealed that migrating node A → node B → back to node A does not work out of the box.
The second migration (back to the original node) appears to succeed at the backup-cli and TLS layer, but the MPC contract state does not update and the migration does not complete without manual intervention (see full reproduction steps and workaround in the link).

This flow is useful in at least the following scenarios:

An operator wants to keep an operational (dormant) backup node at all times and occasionally fail back to the original node.

An operator wants to reuse the original node after migration (e.g., to avoid re-provisioning and setup time for a fresh node).

Request: the migration flows (MPC node, CLI) supports migration to an existing old node. 

### User Story

_No response_

### Acceptance Criteria

migration flow is updates and tested

Please note https://github.com/near/mpc/issues/2121#issuecomment-4590063729

barakeinav1:
Closes #2121.

## Summary

Back-migration (A → B → A) can leave the destination's stored on-chain attestation past expiry by the contract's `current_time_seconds` at conclude time. When that happens, `reverify_participants` rejects `conclude_node_migration` with `InvalidTeeRemoteAttestation`, and `retry_conclude_onboarding` loops forever on the same stale state — the migration never completes.

This PR fixes that by having the destination submit a fresh attestation **before** entering `retry_conclude_onboarding`, and adds contract-sandbox tests that pin both the failure mode and the recovery path.

## The fix

`crates/node/src/migration_service/onboarding.rs::execute_onboarding` now calls a new helper `submit_attestation_before_concluding_migration` between keyshare import and `retry_conclude_onboarding`. The helper generates a fresh quote via `TeeAuthority` and submits it via `submit_remote_attestation`, which blocks until `TransactionStatus::Executed` — so by the time `retry_conclude_onboarding` runs, the contract holds a fresh attestation under the destination's TLS key.

Helper failure is logged but non-fatal: in the non-back-migration path the existing on-chain attestation is typically still valid, so falling through to `retry_conclude_onboarding` is safe.

Plumbing: `TeeAuthority` and `account_public_key` are threaded `run.rs → spawn_recovery_server_and_run_onboarding → onboard → execute_onboarding`.

## Contract-sandbox tests (`crates/contract/tests/sandbox/tee_back_migration.rs`)

Two paired tests sharing a `setup_stale_back_migration` helper:

1. **`conclude_node_migration__rejects_when_destination_attestation_is_stale`** — reproduces the bug. Submits an attestation with `expiry = now + 5s`, fast-forwards 100 blocks, calls `conclude_node_migration` directly; asserts `InvalidTeeRemoteAttestation`.
2. **`conclude_node_migration__succeeds_when_destination_submits_fresh_attestation_before_conclude`** — spec-by-test for the fix. Same setup, but the destination submits a fresh attestation for the same TLS key before calling conclude; asserts success. This is exactly the recovery path the new node-side helper drives.

Both pass in ~12s.

## What's covered vs out of scope

| Layer | Covered |
|---|---|
| Contract rejects when destination attestation is past expiry by `current_time_seconds` | ✅ contract test 1 |
| Contract accepts when a fresh attestation is submitted just before conclude | ✅ contract test 2 |
| Node submits a fresh attestation before `retry_conclude_onboarding` | ✅ unit test `submit_attestation_before_concluding_migration__should_submit_one_fresh_attestation` |
| Mocked Rust E2E A → B → A regression | ✅ already merged in #3346 |
| Real-TDX localnet smoke test for A → B → A | ✅ already merged in #3338 |
| 7-day PCCS-collateral TTL real-TDX repro | ❌ operationally impractical; mock substitutes |

## Test plan

- [x] `cargo test -p mpc-node --lib submit_attestation_before_concluding_migration` — passes.
- [x] `cargo test -p mpc-contract --test test --profile test-release conclude_node_migration` — both contract tests pass.
- [x] `cargo check -p mpc-node --lib --tests` — clean.
- [x] `cargo fmt --check` — clean.
- [ ] CI green.

## Related

- Bug: #2121.
- Mocked Rust E2E coverage: PR #3346 (merged).
- Real-TDX localnet coverage: PR #3338 (merged).
- Investigation thread: <https://github.com/near/mpc/issues/2121#issuecomment-4545873485>.

## Follow-up

- #3368 — Surface `conclude_node_migration` failure cause in node logs (observability gap uncovered while investigating this bug).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3288,6 +3288,7 @@ dependencies = [
  "hex",
  "httpmock",
  "launcher-interface",
+ "libc",
  "mpc-node-config",
  "mpc-primitives",
  "near-account-id 2.6.0",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6044,6 +6044,7 @@ dependencies = [
  "near-mpc-crypto-types",
  "near-o11y 2.12.0",
  "near-sdk",
+ "near-store",
  "near-time 2.12.0",
  "node-types",
  "num_enum",
--- a/crates/e2e-tests/Cargo.toml
+++ b/crates/e2e-tests/Cargo.toml
@@ -15,6 +15,7 @@ futures = { workspace = true }
 hex = { workspace = true }
 httpmock = { workspace = true }
 launcher-interface = { workspace = true }
+libc = "0.2"
 mpc-node-config = { workspace = true }
 near-account-id = { workspace = true }
 near-indexer-primitives = { workspace = true }
--- a/crates/e2e-tests/src/cluster.rs
+++ b/crates/e2e-tests/src/cluster.rs
@@ -344,6 +344,33 @@ impl MpcCluster {
         Ok(())
     }
 
+    /// Send SIGTERM to a running node and wait up to `grace` for it to exit on
+    /// its own. Returns the exit status — `status.code().is_some()` indicates
+    /// the process exited cleanly via its own main() (i.e. our SIGTERM handler
+    /// ran), while `status.signal().is_some()` indicates the OS terminated it
+    /// (i.e. there was no handler).
+    pub fn terminate_node_with_sigterm(
+        &mut self,
+        idx: usize,
+        grace: std::time::Duration,
+    ) -> anyhow::Result<std::process::ExitStatus> {
+        anyhow::ensure!(
+            idx < self.nodes.len(),
+            "node index {idx} out of bounds (have {} nodes)",
+            self.nodes.len()
+        );
+        let state = self.nodes.remove(idx);
+        let (status, setup) = match state {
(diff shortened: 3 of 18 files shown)


netrome:
Closing as stale. This PR revealed that implementing this is quite complex, and the current implementation raised concerns in the team about maintainability and readability. For this feature we don't want to drag on that kind of tech debt, but rather take this as an opportunity to clean up the relevant parts of the code since this should be possible to express more elegantly.

PR #3374, opened 2026-05-27
test: showcase contention during asset generation, analyse problem and solutions

### LINKED ISSUE #1175: Investigate why asset generation impacts signing performance
### Background

During our last signer set upgrade, we saw degraded performance on mainnet with increased CPU utilization and signing latency for three days. This correlated with the nodes generating new assets (presignatures & triplets). However these should run on a separate thread with lower priority. We need to understand how increased asset generation load can impact our signing performance.

### User Story

As an operator I'd like the system to keep producing signatures while assets are generated.

### Acceptance Criteria

We understand why asset generation can cause degraded performance.

One theory is that the near core process starts lagging due to the asset generation load.

gilcu3:
Closes #1175 

Notice here we don't deal with the indexer being part of the problem, which we can somehow safely assume as we only face this right after resharings

### DIFF
--- a/crates/node/src/tests/asset_generation_signing_contention.rs
+++ b/crates/node/src/tests/asset_generation_signing_contention.rs
@@ -1,0 +1,304 @@
+//! Reproduction for issue #1175 — "asset generation impacts signing performance".
+//!
+//! Background asset generation (triples + presignatures) and signing both run on
+//! the same `cores`-limited per-epoch MPC runtime (see
+//! [`crate::coordinator::Coordinator`]'s `create_runtime_and_run`) with no priority
+//! separation between them. The cait-sith poke loop in
+//! [`crate::protocol::run_protocol`] is CPU-bound and does not yield between
+//! network rounds, and the follower/passive side of generation
+//! ([`crate::mpc_client`]'s `monitor_passive_channels_inner`) is unbounded. After a
+//! resharing every node's triple/presignature stores are empty at once, so the
+//! whole network refills simultaneously and each node is flooded with generation
+//! work — which starves signing.
+//!
+//! This test reproduces that locally *without* nearcore, isolating the MPC-runtime
+//! contention from the "nearcore lags" theory. It runs a 4-node in-process cluster
+//! and compares signing latency in two states:
+//!   - steady state — a small desired triple buffer that fills with a single batch
+//!     and then idles;
+//!   - post-resharing — the mainnet desired triple buffer (2^14), which starts
+//!     empty and so triple generation runs flat out.
+//!
+//! Two knobs are not literal mainnet values, both because this in-process cluster
+//! has near-instant networking (see `CORES_PER_NODE` and the `*_TRIPLE_CONCURRENCY`
+//! constants): mainnet reaches CPU saturation because real cross-node latency
+//! keeps many generations in flight at once, whereas here protocols finish too
+//! fast to pile up that way. To recreate the same saturation cheaply we (a) give
+//! MPC a single core, and (b) drive triple generation harder in the post-resharing
+//! scenario (higher concurrency, no stagger) while leaving the steady scenario
+//! with a gentle top-off concurrency. The buffer *size* that defines
+//! "post-resharing" is the real mainnet 2^14.
+//!
+//! It is timing-sensitive, so it is `#[ignore]`d and excluded from CI. Run it
+//! manually and read the printed report. `RUST_LOG=off` silences all `tracing`
+//! output so the `[#1175]` summary lines aren't buried; the report goes through
+//! `println!`, not tracing, so this only mutes the noise. Drop `RUST_LOG=off` if
+//! you need the logs to diagnose a failure.
+//! ```text
+//! RUST_LOG=off cargo nextest run --cargo-profile=test-release -p mpc-node \
+//!     signing_latency__should_degrade_under_concurrent_asset_generation \
+//!     --run-ignored all --no-capture
+//! ```
+//! A handful of `panicked at crates/node/src/indexer/fake.rs … Result::unwrap()
+//! on an Err value: Closed` lines may also appear on stderr at the end — these
+//! are a pre-existing teardown race in the fake indexer's channels (the cluster
+//! drops while background tasks are still using them) and don't affect the
+//! result. The `[#1175]` summary still sits near the top of the test output.
+
+use crate::indexer::participants::ContractState;
+use crate::p2p::testing::PortSeed;
+use crate::tests::{
+    DEFAULT_BLOCK_TIME, DEFAULT_MAX_PROTOCOL_WAIT_TIME, IntegrationTestSetup,
+    request_signature_and_await_response,
+};
+use crate::tracking::AutoAbortTask;
+use average::{Estimate, Max, Mean, Quantile};
+use mpc_node_config::{PresignatureConfig, TripleConfig};
+use mpc_primitives::domain::DomainId;
+use near_mpc_contract_interface::types::{
+    DomainConfig, DomainPurpose, Protocol, ReconstructionThreshold,
+};
+use near_time::Clock;
+use std::time::Duration;
+
+const NUM_PARTICIPANTS: usize = 4;
+const THRESHOLD: usize = 3;
+const TXN_DELAY_BLOCKS: u64 = 1;
+
+// The only knob that differs between the two scenarios is the desired triple
+// buffer: the post-resharing run targets the mainnet size (2^14) and so generates
+// continuously, while the steady run targets a tiny buffer that fills in a couple
+// of batches and then leaves triple generation idle. Everything else is identical.
+const CORES_PER_NODE: usize = 1;
+/// Mainnet triple buffer (2^14): empty after a resharing, so a node refills toward
+/// it continuously — the post-resharing state.
+const REFILL_TRIPLES_TO_BUFFER: usize = 16_384;
+/// Small enough to fill in a couple of batches and then idle — steady state, where
+/// the buffer is already full.
+const STEADY_TRIPLES_TO_BUFFER: usize = 128;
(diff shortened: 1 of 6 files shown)


crates/node/src/tests/asset_generation_signing_contention.rs
```diff
@@ -0,0 +1,304 @@
...
+/// the post-resharing concurrency here would let the initial buffer fill spawn 6
+/// concurrent CPU-heavy batches on the single MPC core, overshooting the small
+/// target and producing a multi-second latency tail on early measured signatures.
+/// One batch at a time is enough to refill the small steady buffer.
+const STEADY_TRIPLE_CONCURRENCY: usize = 1;
+const TRIPLE_STAGGER_SEC: u64 = 0;
+/// Presignatures are buffered modestly and identically in both scenarios, so they
+/// stay available for signing without themselves being the variable under test.
+const PRESIGNATURE_CONCURRENCY: usize = 2;
+const PRESIGNATURES_TO_BUFFER: usize = 64;
```
gilcu3, line 96: this test focuses on the triples, but the presignatures might have some part in the problem, as they are also computed aggressively after resharing.

crates/node/src/tests/asset_generation_signing_contention.rs
gilcu3, line 1: Notice this would not be executed in CI. The plan is to have it so that once the fixes are applied we can check if the behavior changes

crates/node/src/tests/asset_generation_signing_contention.rs
```diff
@@ -0,0 +1,304 @@
...
+        )
+        .await;
+        attempts.push(latency);
+    }
+
+    LatencyReport::from_attempts(label, &attempts)
+}
+
+#[test_log::test(tokio::test(flavor = "multi_thread", worker_threads = 4))]
+#[ignore = "timing-sensitive reproduction for #1175; run manually with --run-ignored"]
```
netrome, line 276: Oh didn't know until now the `MetaNameValueStr` syntax was supported for ignore attributes https://doc.rust-lang.org/reference/attributes/testing.html#r-attributes.testing.ignore.reason

Not sure what the benefit over a normal comment is 🤔

netrome: (I also didn't know this was called `MetaNameValueStr` until now when I looked it up 😅 )

docs/design/signing-starvation-solution.md
```diff
@@ -0,0 +1,130 @@
...
+
+Three architectural facts, plus two that make them fire after a resharing:
+
+1. **Shared runtime, equal priority.** Signing, leader gen, and follower gen
+   all run as peer `tracking::spawn` tasks on the same `cores`-limited
+   per-epoch tokio runtime (`coordinator.rs::create_runtime_and_run`). The
+   `cores` limit exists to protect the indexer, not signing.
+2. **CPU-bound, non-yielding poke loop.** `run_protocol` in `protocol.rs`
+   runs `protocol.poke()` until `Action::Wait`. A 64-batch triple gen burst
+   is tens-to-hundreds of ms between awaits.
```
netrome, line 20: Oh this sounds exactly like the kind of work `tokio::spawn_blocking` is for. Alternatively we should consider inserting more await points.

docs/design/signing-starvation-solution.md
```diff
@@ -0,0 +1,130 @@
...
+3. **Unbounded follower fan-out.**
+   `mpc_client.rs::monitor_passive_channels_inner` spawns one task per
+   incoming peer channel with no cap, so a node has no way to bound how much
+   follower work peers can push onto it.
```
netrome, line 24: This is per network peer? Since we typically have ~10 participants this doesn't sound like it should be any problem.

docs/design/signing-starvation-solution.md
```diff
@@ -0,0 +1,130 @@
...
+4. **Resharing erases the asset buffers.** Triples reference the old
+   participant set and get cleaned up (`assets::cleanup`); presignatures
+   embed the old keyshare and are also wiped. Every node empties at once and
+   refills toward `desired_*_to_buffer`, so the load spike is synchronized
+   network-wide.
+5. **Mainnet runs two CaitSith domains.** Presignature generation runs
+   per-domain (one background loop per `(provider, domain)` in
+   `spawn_background_tasks`), so the per-node presig pipeline doubles.
+
```
netrome, line 33: Sure, but this problem was observed before we ran two CaitSith domains.

docs/design/signing-starvation-solution.md
```diff
@@ -0,0 +1,130 @@
...
+becomes 4× shorter; gen threads idle more often, opening more chances for
+higher-priority threads to grab a core. **Protocol-affecting**: followers
+validate `count == BATCH_SIZE` in `triple.rs`, so leader and follower
+must agree — coordinated network-wide rollout. Worth doing only if a
+protocol bump is already on the table.
+
+### E — `spawn_blocking` or dedicated rayon pool
+
+Move CPU-bound compute off async worker threads entirely. Overlaps with what
+A achieves through runtime separation, but at a deeper restructuring layer.
```
netrome, line 118: Yes, this sounds like something we should have added already. Should be a no-brainer for any CPU kind of work.

docs/design/signing-starvation-solution.md
```diff
@@ -0,0 +1,130 @@
...
+### C — `yield_now()` in the poke loop
+
+A single `tokio::task::yield_now().await` in `run_protocol`'s outer loop
+shortens the maximum CPU burst between cooperative yield points
+(a comment in `protocol.rs` already documents the hazard).
+
+With A in place this mainly helps fairness *within* `gen_runtime`; gen tasks
+yielding to each other does not free a core for signing. Cheap, harmless,
+defensible independently of #1175.
+
```
netrome, line 94: Yeah definitely worth exploring, unless we go with E at which point this shouldn't matter as much.

docs/design/signing-starvation-solution.md
```diff
@@ -0,0 +1,130 @@
...
+   embed the old keyshare and are also wiped. Every node empties at once and
+   refills toward `desired_*_to_buffer`, so the load spike is synchronized
+   network-wide.
+5. **Mainnet runs two CaitSith domains.** Presignature generation runs
+   per-domain (one background loop per `(provider, domain)` in
+   `spawn_background_tasks`), so the per-node presig pipeline doubles.
+
+## Solution options
+
+### A — Lower-OS-priority gen runtime
```
netrome, line 36: I would much prefer not to add additional runtimes, but if not E and possibly C helps with this we should definitely consider something like this.

crates/node/src/tests/asset_generation_signing_contention.rs
```diff
@@ -0,0 +1,304 @@
...
+#[expect(non_snake_case)]
+async fn signing_latency__should_degrade_under_concurrent_asset_generation() {
```
netrome, line 278: Hmm I tried to run this on my machine now and I'm getting this error:

```

    thread 'tests::asset_generation_signing_contention::signing_latency__should_degrade_under_concurrent_asset_generation' (419399) panicked at crates/node/src/tests/asset_generation_signing_contention.rs:292:5:
    assertion `left == right` failed: steady-state baseline should not time out; harness is unhealthy
```

anything you recognize?

netrome (approved):
Thanks for investigating and writing this up. I'm experiencing some issues running the test, but since it's ignored I don't consider it a hard blocker.

docs/design/signing-starvation-solution.md
gilcu3: > Oh this sounds exactly like the kind of work tokio::spawn_blocking is for.

Yes, that should at least fix the contention problem on the tokio runtime.

Are we considering adding waiting points inside the crypto implementation?

docs/design/signing-starvation-solution.md
gilcu3: I think it is, because each peer launches 2 batch triple computations (64 triples each batch) at the same time. So when we compute the load after resharing, it scales linearly with the number of participants.

Basically with 10 nodes, right after resharing, each node now will start concurrently computing 64*10*2 triples, and 16 (presignature concurrency) * 10 presignatures (as soon as triples becomes available). That seems like a lot, but the important bit is that due to work started by peers, it scales linearly in each node.

docs/design/signing-starvation-solution.md
gilcu3: Right, that just maybe made it a bit worse, although as they share triples, and we also decreased the # triples from 1M to 16K, probably we never really observed the diff.

docs/design/signing-starvation-solution.md
gilcu3: The one things I am not 100% sure of is if triple generation is just CPU bound, or the network has also a big influence.

crates/node/src/tests/asset_generation_signing_contention.rs
gilcu3: we discussed in the thread, probably due to the use of `test-release` mode

crates/node/src/tests/asset_generation_signing_contention.rs
anodar: Ideally we'd have this manually dispatchable, run in CI (from this PR), and add that benchmark as a comment and later compare after we fix it.

If not, consider at least adding summarized benchmark results from your local testing.

anodar (approved):
Thanks for looking into this!

crates/node/src/tests/asset_generation_signing_contention.rs
gilcu3: As we discussed in slack having it in CI is a bit out of scope, as this was just tuned for my local env. For the future we plan to do benchmarks in real clusters so that the solutions are tested there.

> If not, consider at least adding summarized benchmark results from your local testing.

The test atm fails if what I found on my local env does not happen, so I think in that sense that counts as a "result".

PR #3376, opened 2026-05-27
docs(design): clarify verifier-rotation eviction in `attestation-verifier-contract.md`

### LINKED ISSUE #3375: docs(design): clarify verifier-rotation eviction in `attestation-verifier-contract.md`
### Background

In [near/mpc@af761ee3#r186633780](https://github.com/near/mpc/commit/af761ee3a1570b9ef3be1064503c45f26b338e6c#r186633780), @barakeinav1 noted that the paragraph added to [`docs/design/attestation-verifier-contract.md`](https://github.com/near/mpc/blob/main/docs/design/attestation-verifier-contract.md) on stored-attestation behavior across a verifier change does not make the security limitation explicit. Voting in a fixed verifier does not retroactively evict entries the old verifier accepted (`re_verify` re-checks post-DCAP allowlist invariants only; the original quote + collateral are not retained), and an attacker holding a wrongly-accepted entry has no incentive to resubmit on the 1-hour cycle.

This is a pre-existing property; the design doc inherits it but does not currently call it out.

### Acceptance Criteria

Update the *Governance and upgrades* section of [`docs/design/attestation-verifier-contract.md`](https://github.com/near/mpc/blob/main/docs/design/attestation-verifier-contract.md) to:

- Lead with the verifier-bug scenario and why a routing-only rotation is insufficient.
- Describe the design that closes the gap: stamp every `VerifiedAttestation` with a `verified_by: AccountId` (from `env::predecessor_account_id()`), check it in `re_verify`, and have `vote_tee_verifier_change` schedule a self-`Promise` calling `clean_invalid_attestations` on threshold-crossing.
- State the brief `accept_requests = false` window honestly, and clarify scope is containment, not remediation.
- Preserve the existing allowlist-drop fast path.

Doc-only. Code work tracked separately.

Review thread: [near/mpc@af761ee3#r186633780](https://github.com/near/mpc/commit/af761ee3a1570b9ef3be1064503c45f26b338e6c#r186633780)

pbeza:
Closes #3375


docs/design/attestation-verifier-contract.md
```diff
@@ -211,7 +211,15 @@ flowchart TD
...
+
+The cost is a brief service-availability gap: right after the sweep, `verify_tee` flips [`accept_requests = false`][accept-requests-flag] until enough honest nodes have re-attested through the new verifier. The contract should refuse new requests until it has fresh verdicts from a verifier the protocol still trusts. Note also what this mechanism does *not* do: it stops future trust in entries the old verifier produced, but does not undo whatever MPC operations a node holding such an entry already participated in. Recovering from past damage is out of scope for this document.
```
barakeinav1: I think this proposed solution is very dangerous.
`verify-tee ` will evict all  (down to the threshold minimum) the nodes (and participants) from contract. and we don't want that.  

also - in case there is a temporary problem with PCCS,  node are blocked out until it is fixed.

barakeinav1 (changes requested):
I actually did not expect it to be fixed, but rather documented.
I also think the proposed solution is dangerous see my response [here](https://github.com/near/mpc/pull/3376/changes#r3311114969).

docs/design/attestation-verifier-contract.md
pbeza: Yeah, that’s a good point. I was also a bit undecided on whether we actually want to invalidate all attestations once we realize the DCAP verifier was buggy or untrusted.

What else can we do in that case? How would reducing the expiration time to `1` day help (this is what you [suggested](https://nearone.slack.com/archives/C0912BTG51T/p1779889505409679?thread_ts=1779882625.964659&cid=C0912BTG51T) offline)? My solution is effectively setting the expiration time to `0`, so I’m not sure if that’s fundamentally different.

pbeza:
@gilcu3 @netrome could you review when you get a chance? Thanks!

gilcu3 (changes requested):
> The cost is a brief service-availability gap: right after the sweep, `verify_tee` flips [`accept_requests = false`][accept-requests-flag] until enough honest nodes have re-attested through the new verifier. The contract should refuse new requests until it has fresh verdicts from a verifier the protocol still trusts.

I think we need to find a better solution than this, a brief service-availability gap for a potentially routine update seems unacceptable. I don't have a clear solution in mind, so maybe we should setup a sync to discuss about this next week

pbeza:
> (...) I don't have a clear solution in mind, so maybe we should setup a sync to discuss about this next week

Sent out the invite for Monday at 1pm CEST. Feel free to move it if that doesn’t work for anyone. cc @barakeinav1 @netrome

pbeza:
@gilcu3 @barakeinav1 I updated the doc to align with what we discussed during today’s call. Feel free to re-review.

docs/design/attestation-verifier-contract.md
```diff
@@ -211,7 +211,15 @@ flowchart TD
...
+The expected trigger for a verifier rotation is a discovered bug in the existing verifier — for example, a DCAP signature-chain flaw that let a non-genuine TEE submit an attestation the contract then stored. A rotation vote re-points `tee_verifier_account_id` at a patched, separately-deployed instance, so every subsequent `submit_participant_info` is verified by the new verifier. It does *not* reach back and delete the entries the broken verifier already accepted: the contract keeps neither the original quote nor the collateral, and [`re_verify`][re-verify] re-checks only post-DCAP allowlist invariants, never DCAP itself.
```
barakeinav1, line 214: nit: I would try shorten this section.
too much text become hard to read.
focus on what is the problem,why we want to fix it.
solution decided, and the security/availability trade off.

barakeinav1 (approved):
approved with one nit.
thanks

gilcu3 (approved):
Thank you!

docs/design/attestation-verifier-contract.md
pbeza: Thanks for the review!

I think we could probably shorten some of the other paragraphs too, but I merged it as-is since I’d rather move on and start working on the implementation. To me, this doc is mainly meant to align us on the implementation direction, not to be a standalone post-implementation write-up of the contract breakout. We can prepare that after we actually implement it.

PR #3386, opened 2026-05-28
chore: use release branches instead of tags as basis for release workflows

### LINKED ISSUE #3342: Support making patch releases from release branches
### Background

In recent changes (https://github.com/near/mpc/pull/3287 and https://github.com/near/mpc/pull/3291) we restricted workflows with access to secrets. While this enhances security, it also means we cannot easily make patch releases from release branches, as we don't have access to push to docker hub from those branches.

Additionally, our release scripts and workflows doesn't handle release branches well.

We should figure out a nice flow to allow releasing from dedicated release branches. Most likely we should add some branch protection rule for `release/vX.X.X` branches and update our CI and scripts to release from explicit release branches.

### User Story

As a release manager I want to be able to backport fixes and make patch releases on top of old versions.

### Acceptance Criteria

We have first-class support for making patch releases, not directly based on latest main.

_No response_

netrome:
closes #3342 

This PR changes our release workflow from being triggered on tag pushes to an explicit release workflow that can be triggered on `main` or `release` branches, both of which are now protected branches.


.github/workflows/docker_release_legacy.yml
netrome, line 1: legacy cleanup

.github/workflows/release.yml
```diff
@@ -2,210 +2,230 @@ name: Release
...
+          echo "sha=$SHA" >> "$GITHUB_OUTPUT"
+          echo "branch-tag=$BRANCH_TAG" >> "$GITHUB_OUTPUT"
 
-      - name: Compute contract hash
-        id: contract-hash
-        run: |
-          CONTRACT_HASH=$(sha256sum result/mpc_contract.wasm | awk '{print $1}')
-          echo "sha256=$CONTRACT_HASH" >> "$GITHUB_OUTPUT"
+          echo "Releasing version $VERSION from $REF_NAME at $SHORT_SHA"
+          echo "Source image tag: $BRANCH_TAG"
```
netrome, line 53: We should consider breaking out this logic to some script later - but for now let's just keep this to unblock the 3.11.0 release.

netrome:
> `.github/workflows/release.yml:43` — the awk extractor picks the first `^version = "X.Y.Z` line in `Cargo.toml`. It happens to be the workspace version today, but is not anchored to `[workspace.package]`. If a future edit reorders sections this could silently grab the wrong field; a two-pass awk that first locates the `[workspace.package]` header would make the intent explicit.

Good point but I think this is fine for now. Will be easy to spot and fix if we hit this later.

.github/workflows/release.yml
```diff
@@ -2,195 +2,208 @@ name: Release
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: false
```
gilcu3, line 5: is this just in case we want to make several releases at the same time? Or just because it seems overly restrictive?

RELEASES.md
```diff
@@ -1,166 +1,204 @@
...
+### 5. Promote to operator floating tags (optional)
 
-1. Retags the Docker images (`mpc-launcher`, `mpc-node`, `mpc-node-gcp`) from `main-<short-sha>` to the release version.
-2. Builds the contract reproducibly and computes its SHA-256 digest.
-3. Creates a **draft** GitHub release with the changelog, Docker image digests, and the contract artifact attached.
+Some operators consume floating tags like `nearone/mpc-node-gcp:testnet-release`
+and `:mainnet-release`. Promote with the retag workflows:
 
-Once the workflow completes, go to the [releases page](https://github.com/near/mpc/releases), review and edit the draft as needed, then publish it.
+- [Release Node Docker Image](.github/workflows/docker_node_release.yml) — run twice (once per repository)
```
gilcu3: this reads a bit ambiguous to me. Might be better to be explicit about the gcp variant

RELEASES.md
```diff
@@ -1,166 +1,204 @@
...
 
-Before pushing the tag, verify that the Docker images for the tagged commit have already been published by the CI pipeline (e.g. `main-<short-sha>` tags on Docker Hub). The Release workflow retags these existing images, so it will fail if they don't exist yet.
+### 4. Edit and publish the draft release
 
-You can create the tag directly in GitHub, but I prefer to do it locally:
+When the workflow finishes, a draft release named `MPC 3.11.0` appears on
+the [releases page](https://github.com/near/mpc/releases). The draft
+includes the changelog section, Docker image manifest digests, and the
+contract `.tar.gz`. **The git tag does not yet exist** — it is created
+by GitHub when the draft is published.
```
gilcu3, line 124: Oh I did not know that. How is that achieved?

scripts/prepare-release.sh
```diff
@@ -82,28 +84,8 @@ if ! git diff --quiet || ! git diff --cached --quiet; then
...
-git checkout -b "$BRANCH"
-
-# --- Push to remote ---
-
-echo "==> Pushing '${BRANCH}' to ${REMOTE}..."
-git push -u "$REMOTE" "$BRANCH"
-
 # --- Generate changelog ---
 
-# Branch must be pushed first so git-cliff can resolve PR links.
```
gilcu3, line 106: are we sure this is not needed anymore?

.github/workflows/release.yml
```diff
@@ -2,195 +2,208 @@ name: Release
...
+            echo "::error::Could not find workspace 'version = \"X.Y.Z\"' line in Cargo.toml"
+            exit 1
+          fi
+          if [[ ! "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9.]+)?$ ]]; then
+            echo "::error::Invalid semver in Cargo.toml: $VERSION"
+            exit 1
+          fi
+
+          SHA=$(git rev-parse HEAD)
+          SHORT_SHA="${SHA::7}"
```
anodar, line 45: As discussed offline, we should probably do: `git rev-parse --short HEAD` here.

.github/workflows/release.yml
```diff
@@ -2,195 +2,208 @@ name: Release
...
-        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
-        with:
-          name: mpc-contract
-          path: mpc-contract-v${{ needs.release-info.outputs.version }}.tar.gz
+      - name: Extract changelog section
+        env:
+          VERSION: ${{ steps.inputs.outputs.version }}
+        run: |
+          set -euo pipefail
+          awk "/^## \\[${VERSION}\\]/{found=1; next} /^## \\[/{if(found) exit} found" CHANGELOG.md > release-notes-raw.md
```
gilcu3, line 83: nit: would be slightly better to do this with grep instead, just because it is better known

.github/workflows/release.yml
```diff
@@ -2,195 +2,208 @@ name: Release
...
-      short-sha: ${{ steps.version.outputs.short-sha }}
+    environment: production
+    permissions:
+      contents: write
+      actions: read
 
     steps:
-      - name: Extract version
-        id: version
+      - name: Validate trigger ref
```
andrei-near: this step is unnecessary. 
workflow will fail at start if it cannot access required env.
Removing this also avoids hard-coding branch names in workflow

.github/workflows/release.yml
```diff
@@ -2,195 +2,208 @@ name: Release
...
-          VERSION: ${{ needs.release-info.outputs.version }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERSION: ${{ steps.inputs.outputs.version }}
         run: |
-          cp result/mpc_contract.wasm "mpc-contract-v${VERSION}.wasm"
-          tar -czf "mpc-contract-v${VERSION}.tar.gz" "mpc-contract-v${VERSION}.wasm"
+          set -euo pipefail
+          if gh release view "${VERSION}" --repo "${GITHUB_REPOSITORY}" >/dev/null 2>&1; then
+            echo "::error::A GitHub release (draft or published) for ${VERSION} already exists."
+            echo "::error::Delete it on the releases page before re-running."
```
andrei-near, line 74: releases are immutable org wide. this advice is not applicable

.github/workflows/release.yml
```diff
@@ -2,195 +2,208 @@ name: Release
...
+        run: |
+          set -euo pipefail
+          RUN_ID=$(gh run list \
+            --repo "$GITHUB_REPOSITORY" \
+            --workflow=build_contract.yml \
+            --commit="$SHA" \
+            --status=success \
+            --limit=1 \
+            --json databaseId \
+            -q '.[0].databaseId')
```
gilcu3, line 127: do we need this because of some security restriction in the new setting? It feels more natural that the output on one action is the input of another action, as we had before (or as we have in other parts of the CI)

gilcu3 (dismissed):
LGTM, although I am not super confident it will work (github actions have bitten me several times already).

Left some comments, maybe some would be worth addressing before merging.

RELEASES.md
netrome: Alledgedly, Github will auto-create the tag once you publish the release. Will see if that actually happens

anodar (dismissed):

.github/workflows/release.yml
netrome: Or rather the opposite. `git rev-parse --short HEAD` is not guaranteed to be the same size. But yeah we should be consistent. Updated in https://github.com/near/mpc/pull/3386/commits/2d3be888ad209186df3f8baaced28fe5514f9351

.github/workflows/release.yml
netrome: Fair point, I'll remove it.

.github/workflows/release.yml
netrome: Seems like I can delete releases - also this would primarily happen for release drafts which definitely should be deletable.

Image

.github/workflows/release.yml
netrome: Yeah concurrent workflows shouldn't happen frequently, but since this one has mid-flight side effects it's nice to have it complete even if another job is queued.

No strong opinion here though if there's any reason to set this to true.

.github/workflows/release.yml
netrome: I like the idea of publishing build artifacts that we can consume later to decouple the "build release candidate artifacts" from the "assemble release bundle" steps. This makes the actual release step very quick, and would allow us to easily smoke test the release artifacts before making the release down the line.

Happy to revisit this if you have other preferences, but let's try it this way for this release unless you feel strongly against it.

.github/workflows/release.yml
netrome: Claude pushed back, I haven't vetted the response so lmk if you think I should take this with a pinch of salt:

```
grep instead of awk — I'd push back. Awk is the right tool for "extract lines between two markers." grep doesn't have a between-patterns mode; doing this with grep would actually
  require sed -n '/start/,/end/p' (i.e. sed, not grep) plus extra logic to exclude the boundary line. The substitution would be more complex than the current awk, not less. The "better
  known" argument is weak — awk has been standard Unix since 1977 and any reviewer who can read sed can read this awk one-liner. Recommend leaving as-is.
```

.github/workflows/release.yml
netrome: Removed in https://github.com/near/mpc/pull/3386/commits/59644a3859a491cde5dd06d09c1c03235d2638c1

RELEASES.md
netrome: Makes sense, elaborated in https://github.com/near/mpc/pull/3386/commits/59644a3859a491cde5dd06d09c1c03235d2638c1

.github/workflows/release.yml
gilcu3: ok if claude says it cannot be done with grep alone then I believe it :) I thought we were simply searching for a pattern there, for which grep should be more than enough

.github/workflows/release.yml
gilcu3: It feels a bit hackier in the sense that this workflow has little guarantees about the other workflows, but no strong opinion if you think the decoupling is worth it

gilcu3 (approved):

scripts/prepare-release.sh
netrome: I'm not entirely sure, but we can easily fix this later if this fails. Claude seems to think this is accurate:

```
Yes, it's fine. The old comment was misleading.

  git-cliff resolves PR links by querying the GitHub API per commit SHA: GET /repos/<owner>/<repo>/commits/<sha>/pulls. The commits it looks up are the ones being summarized in the changelog
   range — i.e., the commits between the previous release tag and HEAD. Those commits all came from previously-merged PRs and are already on GitHub (that's how they landed on main or
  release/vX.Y in the first place).

  The operator's working branch existing remotely doesn't enter into git-cliff's lookups. What matters is that the historical commits are reachable from some ref on GitHub, which they are by
   virtue of being on main / release/vX.Y.

  The old script's push was actually doing something slightly different: it created and published release/vX.Y.Z as the release branch. The # Branch must be pushed first so git-cliff can
  resolve PR links comment was a post-hoc rationalization — true in the old flow (because you wanted the branch to exist before tagging happened on it later), but not literally about
  git-cliff's lookup behavior.

  In the new flow, the working branch is just a scratch branch the operator hasn't pushed yet. git-cliff doesn't care. It'll resolve PR links for every commit in the range correctly.

  If you want a quick smoke test on a real release prep, look at the output of git-cliff --unreleased -t X.Y.Z on a fresh working branch off main before pushing — you should see PR links in
  the output. That's all the script does.
 ```

netrome: Also apologies for the Claude ping-ponging. I want to get this out so we can make the 3.11.0 release. Normally I would spend more time understanding everything myself. In this case, since the script is optional, I prefer to not go too deep right now.

gilcu3: yeah I think it would not work because we added that comment precisely because it did not work :) But yeah we can always try, maybe there is something different now that I am missing

andrei-near (approved):

anodar (approved):

.github/workflows/release.yml
netrome: True, the same goes for the retagging workflows though.

PR #3387, opened 2026-05-28
chore: use workspace edition for threshold-signatures crate

### LINKED ISSUE #3369: Migrate threshold-signatures crate to Rust edition 2024
### Background

While unifying the workspace Rust edition (#1035), every crate was moved to inherit the workspace edition (2024) except `threshold-signatures`, which remains pinned to `edition = "2021"`.

Migrating it requires invasive changes across its public protocol API. Under Rust 2024 the `impl Trait` lifetime-capture rules changed, so every function returning `impl Protocol` (`keygen`, `reshare`, `refresh`, `presign`, `sign`, `ckd`, `generate_triple`/`generate_triple_many`, …) needs precise-capturing `use<…>` bounds. That in turn requires converting their argument-position `impl Trait` parameters to named generics and updating all call sites in `node`, tests, and benches.

```
error[E0597]: `participants_list` does not live long enough
note: this call may capture more lifetimes than intended, because Rust 2024
      has adjusted the `impl Trait` lifetime capture rules
```

### Acceptance Criteria

- `threshold-signatures` inherits the workspace edition (`edition = { workspace = true }`) instead of pinning `edition = "2021"`.
- The crate and the full workspace build under edition 2024.

gilcu3:
Closes #3369

### DIFF
--- a/crates/test-parallel-contract/Cargo.toml
+++ b/crates/test-parallel-contract/Cargo.toml
@@ -22,7 +22,7 @@ sha2 = { workspace = true }
 
 # k256 won't compile without settings the custom rand flag for `getrandom`
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-getrandom = { version = "0.2.12", features = ["custom"] }
+getrandom = { workspace = true, features = ["custom"] }
 
 [lints]
 workspace = true
--- a/crates/threshold-signatures/Cargo.toml
+++ b/crates/threshold-signatures/Cargo.toml
@@ -3,10 +3,7 @@ name = "threshold-signatures"
 description = "Threshold Signatures"
 repository = "https://github.com/near/threshold-signatures"
 version = "0.1.0"
-# TODO(#3369): inherit the workspace edition (2024). Pinned to 2021 for now
-# because the migration needs precise-capturing `use<>` changes across the
-# public protocol API.
-edition = "2021"
+edition = { workspace = true }
 license.workspace = true
 
 [lib]
--- a/crates/threshold-signatures/src/confidential_key_derivation/protocol_pv.rs
+++ b/crates/threshold-signatures/src/confidential_key_derivation/protocol_pv.rs
@@ -1,15 +1,15 @@
+use crate::Protocol;
 use crate::confidential_key_derivation::ciphersuite::{
-    check_valid_point_g1, check_valid_point_g2, multi_miller_loop, BLS12381SHA256,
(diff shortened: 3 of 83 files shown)


crates/test-parallel-contract/Cargo.toml
```diff
@@ -22,7 +22,7 @@ sha2 = { workspace = true }
 
 # k256 won't compile without settings the custom rand flag for `getrandom`
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-getrandom = { version = "0.2.12", features = ["custom"] }
+getrandom = { workspace = true, features = ["custom"] }
```
gilcu3, line 25: broken window fix

crates/threshold-signatures/Cargo.toml
```diff
@@ -3,10 +3,7 @@ name = "threshold-signatures"
 description = "Threshold Signatures"
 repository = "https://github.com/near/threshold-signatures"
 version = "0.1.0"
-# TODO(#3369): inherit the workspace edition (2024). Pinned to 2021 for now
-# because the migration needs precise-capturing `use<>` changes across the
-# public protocol API.
-edition = "2021"
+edition = { workspace = true }
```
gilcu3, line 6: this is the main change that triggers all the rest, that are strictly mechanical changes as rustfmt and clippy rules have changed

anodar (approved):

crates/threshold-signatures/src/confidential_key_derivation/protocol_pv.rs
```diff
@@ -84,15 +84,19 @@ pub const CKD_MAX_INCOMING_PARTICIPANT_ENTRIES: usize = 0;
...
-    app_id: impl Into<AppId>,
+    app_id: A,
     app_pk: PublicVerificationKey,
-    rng: impl CryptoRngCore + Send + 'static,
-) -> Result<impl Protocol<Output = CKDOutputOption>, InitializationError> {
+    rng: R,
+) -> Result<impl Protocol<Output = CKDOutputOption> + use<A, R>, InitializationError>
+where
+    A: Into<AppId>,
+    R: CryptoRngCore + Send + 'static,
```
netrome, line 98: Wait clippy complains about `impl Trait` in function args now?

netrome (approved):
Nice stuff!

crates/threshold-signatures/src/confidential_key_derivation/protocol_pv.rs
gilcu3: In this case it did, not sure if it complains about all cases or only some cases

PR #3388, opened 2026-05-28
test(e2e): add no-restart back-migration variant + refactor shared scaffolding

barakeinav1:
## Summary

Adds a second back-migration test that does *not* kill+restart A0 between the forward and back directions, and refactors the shared scaffolding so the two tests share their setup/teardown.

### Why a no-restart variant

`migration_service__should_handle_back_migration_a_to_b_to_a` (existing) kill+restart's A0 to model the disk-persisted-state scenario investigated in #2121. That kill+restart is also the trigger for the upstream nearcore indexer panic surfaced via #3366 — see [the diagnostic report](https://github.com/near/mpc/issues/3366) for the `chain/indexer/src/streamer/mod.rs:207` `expect("receipt must be present at this moment")` panic. On branches that produce more sandbox chain activity than plain main, A0's post-restart indexer catch-up hits that panic and the test flakes.

Until the upstream nearcore fix lands, this PR adds **`migration_service__should_handle_back_migration_a_to_b_to_a_no_restart`** — a stable sibling that exercises the back-migration code path through A0's in-memory state transitions, with no kill+restart and no restart-time catch-up.

### What each test covers

| Test | Coverage | Triggers nearcore panic? |
|---|---|---|
| `…_a_to_b_to_a` (existing) | back-migration code path + #2121 disk-persisted-state scenario | yes, on branches with high chain activity |
| `…_a_to_b_to_a_no_restart` (new) | back-migration code path (in-memory state transitions only) | no |

### Refactor

The two tests previously had ~85% identical bodies (cluster setup, sanity asserts, forward A→B, post-forward sign+ckd, back B→A, kill B0, final sign+ckd). I factored that into `run_back_migration_round_trip(port_seed, between)` with a small `BetweenDirections` enum (`None` vs `KillAndRestartA`). Each `#[tokio::test]` is now a one-line wrapper that picks the port seed and variant.

The kill+restart-specific block (keyshare snapshots, indexer-height snapshot, kill/start/health/indexer-wait, keyshare-preserved assert) lives inside the `KillAndRestartA` match arm.

### Test plan

- [x] `cargo check --tests -p e2e-tests` clean
- [ ] Both tests pass in CI (the new no-restart should be stable; the existing one's flake behavior is unchanged — still tracked under #3366)
- [ ] If the no-restart variant goes flaky too, we've learned something important about whether the issue is fully kill-specific

### DIFF
--- a/crates/e2e-tests/tests/common.rs
+++ b/crates/e2e-tests/tests/common.rs
@@ -35,6 +35,7 @@ pub const CONTRACT_UPGRADE_COMPATIBILITY_MAINNET_PORT_SEED: u16 = 18;
 pub const CONTRACT_UPGRADE_COMPATIBILITY_TESTNET_PORT_SEED: u16 = 19;
 pub const TIMEOUT_METRIC_PORT_SEED: u16 = 20;
 pub const MIGRATION_BACK_PORT_SEED: u16 = 21;
+pub const MIGRATION_BACK_NO_RESTART_PORT_SEED: u16 = 22;
 
 /// Start a cluster, wait for Running state and presignatures to buffer.
 ///
--- a/crates/e2e-tests/tests/migration_service.rs
+++ b/crates/e2e-tests/tests/migration_service.rs
@@ -628,42 +628,47 @@ async fn migration_service__should_migrate_nodes_via_backup_cli() {
     }
 }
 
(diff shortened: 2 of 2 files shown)


netrome:
Closing as stale.

PR #3382, opened 2026-05-27
feat(foreign-chain-inspector): canonical-chain check for Bitcoin and Starknet

pbeza:
## Summary

Closes #3335.

The EVM inspector already protects against an RPC returning a transaction receipt anchored to a *side block* by re-fetching the canonical block at the receipt's height and comparing hashes (added in #3091). This PR applies the same defense-in-depth check to the Bitcoin and Starknet inspectors, which previously relied on the RPC honoring "no receipt for non-canonical txs" as a security property.

## Approach

- **Bitcoin**: two-step lookup — `getblock(receipt.blockhash) → height`, then `getblockhash(height) → canonical_hash`, compare to `receipt.blockhash`. Mirrors the EVM pattern closely.
- **Starknet**: single extra RPC — `starknet_getBlockWithTxHashes({block_number: receipt.block_number}) → canonical_hash`, compare to `receipt.block_hash`. The receipt already exposes the block number, so no second lookup is needed.
- **Error variant**: generalized `ForeignChainInspectionError::NonCanonicalBlock` to chain-agnostic primitive types (`u64` + `Vec<u8>`) so Bitcoin's 32-byte `TransportBitcoinBlockHash` newtype can populate it without coercion. EVM call site and EVM test updated accordingly.
- **Docs**: `docs/foreign-chain-transactions.md` updated to describe the new RPC call plan for all three inspectors.
- **e2e mocks**: `crates/e2e-tests/src/foreign_chain_mock.rs` dispatches on method and handles the new `getblock`/`getblockhash`/`starknet_getBlockWithTxHashes` calls.

## Verification

- `cargo nextest run -p foreign-chain-rpc-interfaces -p foreign-chain-inspector` — 118 tests pass
- `cargo clippy --all-targets -D warnings` clean on changed crates
- `cargo fmt -- --check` clean
- e2e foreign-chain validation suite still needs a manual run on CI

PR #3389, opened 2026-05-28
feat(foreign-chain-inspector): canonical-chain check for Starknet

pbeza:
## Summary

Adds the defense-in-depth canonical-chain check (already present on the EVM inspector) to the Starknet inspector. After fetching the receipt and verifying its finality status, the inspector now independently calls `starknet_getBlockWithTxHashes` for the canonical block at the receipt's height and rejects the receipt if the block hashes disagree.

## Notes for reviewers

Split out from #3382 for easier review. A separate stacked PR will add the equivalent check for Bitcoin on top of this.


crates/e2e-tests/src/foreign_chain_mock.rs
```diff
@@ -135,12 +135,32 @@ pub fn setup_starknet_mock(server: &MockServer) {
...
             let method = body["method"].as_str().expect("method field");
 
-            if method != "starknet_getTransactionReceipt" {
-                return jsonrpc_error(id, method);
-            }
+            let result = match method {
+                "starknet_getTransactionReceipt" => starknet_receipt_result(),
+                "starknet_getBlockWithTxHashes" => serde_json::json!({
+                    "block_hash": format!("0x{MOCK_BLOCK_HASH}"),
+                    "block_number": 6868546,
```
gilcu3: we should probably also have a constant for this number, same as `MOCK_BLOCK_HASH`

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -46,6 +45,29 @@ pub enum RpcAuthentication {
...
+        }
+        Ok(())
+    }
+}
+
+impl std::fmt::Debug for HexBytes {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        std::fmt::Display::fmt(self, f)
+    }
+}
```
gilcu3: don't we have some nice derive so that this uses hex? I think we use it in other parts of the codebase

crates/foreign-chain-inspector/src/starknet/inspector.rs
```diff
@@ -74,6 +78,42 @@ where
...
+            .request(GET_BLOCK_WITH_TX_HASHES_METHOD, &args)
+            .await?;
+
+        let hash_matches = canonical.block_hash == receipt_block_hash;
+        let height_matches = canonical.block_number == receipt_block_number;
+        if !hash_matches || !height_matches {
+            return Err(ForeignChainInspectionError::NonCanonicalBlock {
+                block_number: receipt_block_number,
+                receipt_hash: receipt_block_hash.as_bytes().to_vec().into(),
+                canonical_hash: canonical.block_hash.as_bytes().to_vec().into(),
```
gilcu3: would be nice to have proper `Into` conversions here. Feel free to add them if they don't clutter the PR

crates/foreign-chain-rpc-interfaces/src/starknet.rs
```diff
@@ -161,4 +199,45 @@ mod tests {
...
+        assert_eq!(serialized, serde_json::json!([{ "block_number": 842_750 }]));
+    }
+
+    #[test]
+    fn deserialize_get_block_with_tx_hashes_response__should_accept_short_hex_block_hash() {
+        let json = r#"
+        {
+            "block_hash": "0x5",
+            "block_number": 842750
+        }
```
gilcu3: nit: why the different formats for the block_number? Could we use a constant?

gilcu3 (changes requested):
Thank you! Looks good overall, left some comments. 

I have only a small request change: we should test this new behavior in the manual test `crates/foreign-chain-inspector/tests/starknet_rpc_manual.rs`. If that does not make sense I am ready to approve.

crates/foreign-chain-rpc-interfaces/src/starknet.rs
pbeza: https://github.com/near/mpc/pull/3389/commits/e093db60 + https://github.com/near/mpc/pull/3389/commits/c6d7f46991d6f1f29606805dc4ab804502589dcf

crates/foreign-chain-inspector/src/starknet/inspector.rs
pbeza: https://github.com/near/mpc/pull/3389/commits/734974d1

crates/e2e-tests/src/foreign_chain_mock.rs
pbeza: https://github.com/near/mpc/pull/3389/commits/9fdd5a7a

crates/foreign-chain-inspector/src/lib.rs
pbeza: I don’t think we have any, but I still managed to simplify it using `#[display("0x{}", hex::encode(_0))]`: https://github.com/near/mpc/pull/3389/commits/279b6fc8. ✅

pbeza:
> I have only a small request change: we should test this new behavior in the manual test `crates/foreign-chain-inspector/tests/starknet_rpc_manual.rs`. If that does not make sense I am ready to approve.

Test added: https://github.com/near/mpc/pull/3389/commits/c23d4af9307c055ff2c4e4e59f31826af47a3568. Note that it only covers the happy path, since a correctly behaving Starknet node won’t return a side-block receipt, so we can’t really hit the canonical-chain failure path through a live RPC. The check is tested for failure — but only with a mock RPC, in `extract__should_return_non_canonical_block_when_receipt_block_hash_differs_from_canonical`:

https://github.com/near/mpc/blob/2a0a44cf94c53f3e30b9d06cf608cbbf933da7e2/crates/foreign-chain-inspector/tests/starknet_inspector.rs#L396-L444

DSharifi (dismissed):

DSharifi (approved):

gilcu3 (approved):
Thank you!

PR #3394, opened 2026-05-29
chore: bump cargo-near/cargo-near-build to 0.20.3, remove workaround

### LINKED ISSUE #3363: Remove `--skip-rust-version-check` CLI override once cargo-near releases #420
### Background

Two places in the workspace bypass cargo-near's refusal to build contracts with rustc >= 1.87 (the bulk-memory incompatibility with the nearcore contract VM that motivated this check has been resolved upstream):

1. `crates/test-utils/src/contract_build.rs` injects `--skip-rust-version-check` into the `cargo-near` subprocess via a custom `CliDescription` prefix. `BuildOpts::skip_rust_version_check` is currently honored only by the in-process build path, not by `build_with_cli`. [near/cargo-near#420](https://github.com/near/cargo-near/pull/420) (commit [60335d1](https://github.com/near/cargo-near/commit/60335d11825bdab41b1f2ce5ddcf795f1c52e079)) fixes that — once cargo-near publishes a release that includes it, the `BuildOpts` field will work through the CLI path and the prefix override becomes unnecessary.

2. `nix/mpc-contract.nix` passes `--skip-rust-version-check` to the `cargo near build` invocation that produces the reproducible contract WASM. This stays needed until cargo-near drops the obsolete rustc 1.87+ refusal entirely.

### Acceptance Criteria

- [ ] cargo-near-build bumped to a release containing [60335d1](https://github.com/near/cargo-near/commit/60335d11825bdab41b1f2ce5ddcf795f1c52e079); `skip_rust_version_check_cli_description` helper and its `cli_description` override removed from `crates/test-utils/src/contract_build.rs`, replaced with `skip_rust_version_check: true` on `BuildOpts`
- [ ] `--skip-rust-version-check` removed from `cargoNearArgs` in `nix/mpc-contract.nix` once cargo-near drops the obsolete 1.87+ check

gilcu3:
Closes #3363 

Includes a broken window fix to cope with this [nix issue](https://www.reddit.com/r/NixOS/comments/1tpqnw1/403_on_cratesio/)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -168,10 +168,10 @@
           ];
 
           pythonTools = with pkgs; [
-            python311
-            python311Packages.keyring
-            python311Packages.tree-sitter
-            python311Packages.tree-sitter-rust
+            python313
+            python313Packages.keyring
+            python313Packages.tree-sitter
+            python313Packages.tree-sitter-rust
             ruff # linter and formatter
           ];
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1002,7 +1002,7 @@ dependencies = [
  "mpc-primitives",
  "near-mpc-crypto-types",
  "node-types",
- "reqwest 0.13.3",
+ "reqwest 0.13.4",
  "serde_json",
  "sha2 0.10.9",
  "tempfile",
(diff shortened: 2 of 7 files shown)


flake.nix
```diff
@@ -168,10 +168,10 @@
...
 
           pythonTools = with pkgs; [
-            python311
-            python311Packages.keyring
-            python311Packages.tree-sitter
-            python311Packages.tree-sitter-rust
+            python313
+            python313Packages.keyring
+            python313Packages.tree-sitter
+            python313Packages.tree-sitter-rust
```
gilcu3, line 174: we had an incompatibility here. We were forced to python 3.11 because of bls in pytests, so this is fine. I was thinking to actually remove this soon, by finding the equivalent rust alternative to avoid building python.

DSharifi (approved):

SimonRastikian (approved):

Issue #3395, opened 2026-05-29
Nearcore metrics missing in 3.11.0


netrome:
### Background

While upgrading we noticed that we're lacking the nearcore metrics in 3.11.0. We should restore them

### User Story

As an MPC engineer I want to have visibility on how the embedded nearcore node is working.

### Acceptance Criteria

The nearcore metrics are served again in the node, and we have tests to ensure that they are in place.

gilcu3:
ups found the problem, nearcore updated prometheus to 0.14 and we are still in 0.13. Should be an easy fix, going to do so against the 3.11 release branch and main

PR #3396, opened 2026-05-29
chore: bump prometheus, fix missing nearcore metric

gilcu3:
Solves part of #3395 

Confirmed working using localnet.

In the PR against main I can add the e2e tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -269,8 +269,7 @@ httpmock = "0.8.2"
 # This can be removed once the workspace `utilities` crate is removed
 near-account-id-v1 = { version = "1.1.4", package = "near-account-id" }
 
-# This cannot be upgraded because it conflicts the prometheus version required by nearcore
-prometheus = { version = "0.13.4", default-features = false }
+prometheus = { version = "0.14.0", default-features = false }
 
 # TODO(#690): This cannot be upgraded until nearcore does
 rocksdb = "0.21.0"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -269,8 +269,7 @@ httpmock = "0.8.2"
 # This can be removed once the workspace `utilities` crate is removed
 near-account-id-v1 = { version = "1.1.4", package = "near-account-id" }
 
-# This cannot be upgraded because it conflicts the prometheus version required by nearcore
-prometheus = { version = "0.13.4", default-features = false }
+prometheus = { version = "0.14.0", default-features = false }
```
netrome, line 272: Curious how the prometheus version change caused metrics to vanish - why is that?

netrome (dismissed):
Thanks!

Cargo.toml
gilcu3: the crates were incompatible, for that reason we had not upgraded, as they silently drop metrics

netrome: Right because we just use a global mutable registry for the metrics. It's really weird that we rely on nearcore metrics leaking into our registry. We should serve those separate. But perhaps we can live with this a bit longer until we cleanly separate the binaries.

gilcu3: yeah I think we discussed exactly that possibility a while ago

netrome: I must have missed or forgotten that conversation :sweat_smile:  but yeah we have known for a long while that we're relying on incidental behavior for our nearcore metrics.

pbeza (approved):

netrome (approved):

PR #3399, opened 2026-05-29
test: flaky test for DamgardEtAl

### LINKED ISSUE #3398: Flaky test_onboarding_jobs
### Background

While running CI I encountered a flaky test generated following #3161. In short the PR added the n >= 2t − 1 enforcement for DamgardEtAl and patched gen_threshold_params (lower bound 2→3), but missed gen_valid_params_proposal. It's a flaky test that fails with probablity 1/1120.  The test is indeed probabilistic and uses rand::thread_rng();.

### User Story

_No response_

### Acceptance Criteria

The test currently fails when  n_old_participants = 2 and max_added = 0.
Update gen_valid_params_proposal

_No response_

SimonRastikian:
Closes #3398

### DIFF
--- a/crates/contract/src/state/test_utils.rs
+++ b/crates/contract/src/state/test_utils.rs
@@ -47,11 +47,11 @@ pub fn gen_valid_params_proposal(params: &ThresholdParameters) -> ThresholdParam
     // consistent. Clamping after sampling keeps the original `[0, 10)` range
     // and only nudges the lowest values up when they'd produce an invalid n.
     let min_added = if n_old_participants < 3 {
-        3 - n_old_participants
+        3_i32.saturating_sub(n_old_participants.try_into().unwrap())
     } else {
         0
     };
-    let max_added = max_added.max(min_added);
+    let max_added = max_added.max(min_added.try_into().unwrap());
     let mut next_id = current_participants.next_id();
     for i in 0..max_added {
         let (account_id, info) = gen_participant(i);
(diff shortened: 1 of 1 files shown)


DSharifi (dismissed):
Thanks!

crates/contract/src/state/test_utils.rs
```diff
@@ -40,6 +40,18 @@ pub fn gen_valid_params_proposal(params: &ThresholdParameters) -> ThresholdParam
...
+    // `n >= 3` even at the minimum `t = 2`. `gen_threshold_params` enforces
+    // the same floor for the initial params; this keeps the proposal
+    // consistent. Clamping after sampling keeps the original `[0, 10)` range
+    // and only nudges the lowest values up when they'd produce an invalid n.
+    let min_added = if n_old_participants < 3 {
+        3_i32.saturating_sub(n_old_participants.try_into().unwrap())
+    } else {
+        0
+    };
+    let max_added = max_added.max(min_added.try_into().unwrap());
```
gilcu3: if DamgardEtAl requires $n \geq 3$ then how can we have `n_old_participants < 3`? I think the fix proposed is very convoluted and does not really solve part of the issue, which is that this test is not deterministic

gilcu3 (changes requested):
I think we should start by making the test generator deterministic, and then ensure that it generates a valid set of parameters for all our schemes more naturally, without having to patch values after they were generated

crates/contract/src/state/test_utils.rs
SimonRastikian: n_old_participants is computed as follows.
Where

 ```rust
    let current_k = params.threshold().value() as usize;
    let current_n = params.participants().len();
   let n_old_participants: usize = rng.gen_range(current_k..current_n + 1);
```

It could happen that 
current_k = t = 2 and current_n = n = 3. As you see this can indeed imply that n_old_participants = 2.

gilcu3 (approved):
Thank you!

DSharifi (approved):

PR #3410, opened 2026-05-31
fix(node): handle SIGTERM for graceful shutdown on operator stop

### LINKED ISSUE #3409: mpc-node: handle SIGTERM for graceful shutdown on operator stop
### Background

`mpc-node` does not install a SIGTERM handler today. When an operator stops the process via `docker stop`, `kubectl delete`, `systemctl stop`, or dstack's CVM stop command, the orchestrator sends SIGTERM first and then SIGKILL after a grace period (10s for Docker default, 30s for Kubernetes, 90s for systemd). Because we have no handler installed, **SIGTERM has the same effect as SIGKILL** — the OS terminates the process immediately, the embedded `near-indexer` thread is killed mid-write, and the next start can land on inconsistent RocksDB state.

This issue was surfaced while investigating [`docs/investigation/2121-back-migration-e2e-flake.md`](https://github.com/near/mpc/blob/main/docs/investigation/2121-back-migration-e2e-flake.md), where a CI test SIGKILLs `mpc-node` mid-flight and the next start panics ~65–80% of the time inside `near-indexer` (see [`docs/investigation/nearcore-indexer-sigkill-restart-panic.md`](https://github.com/near/mpc/blob/main/docs/investigation/nearcore-indexer-sigkill-restart-panic.md)). Production stops via dstack/Docker/Kubernetes/systemd take the same code path, so any production stop today carries the same restart-corruption risk as the test scenario.

### User Story

As an operator stopping an MPC node via my orchestrator (dstack CVM stop / Docker / Kubernetes / systemd), I want the node to receive SIGTERM, finish in-flight commits, and exit cleanly within the grace period — so that the next start finds RocksDB in a consistent state and doesn't trip an indexer restart panic.

### Acceptance Criteria

- [ ] `mpc-node` installs a SIGTERM handler that routes the signal into the existing internal shutdown channel (`shutdown_signal_sender`), so the same `tokio::select!` arm that handles TEE image-hash shutdowns also handles SIGTERM.
- [ ] After the main `select!` exits, `near_async::shutdown_all_actors()` is called so nearcore's actor system can commit any in-flight RocksDB batches before the process exits.
- [ ] `tracing::warn!("SIGTERM received, initiating graceful shutdown")` is emitted when the signal arrives, so operators can confirm the path was taken.
- [ ] Verified in CI: `mpc-node` exits gracefully (typically within 100 ms) after SIGTERM, vs SIGKILL fallback firing in the pre-fix state.

- The investigation that surfaced this issue and the test campaign data live in [`docs/investigation/2121-back-migration-e2e-flake.md`](https://github.com/near/mpc/blob/main/docs/investigation/2121-back-migration-e2e-flake.md). With a working handler, the e2e test passed 1/5 vs 0/5 without it on the same commit — confirming the handler is a real production improvement but **does not fully close the upstream nearcore restart panic**, which fires non-deterministically regardless of shutdown cleanliness.
- The upstream nearcore bug is documented separately in [`docs/investigation/nearcore-indexer-sigkill-restart-panic.md`](https://github.com/near/mpc/blob/main/docs/investigation/nearcore-indexer-sigkill-restart-panic.md). That issue needs to be fixed in nearcore; this issue is the orthogonal mpc-node-side fix that should land regardless.
- We considered also calling `near_store::db::RocksDB::block_until_all_instances_are_dropped()` (which neard's standalone binary does after `shutdown_all_actors`), but it **hangs indefinitely** in our embedding because our indexer thread's `block_on` never returns — the spawned monitor tasks hold `Arc<IndexerState>` → `Arc<RocksDB>` references that nothing currently cancels on shutdown. A proper fix for that hang would wire a `CancellationToken` through the indexer thread; out of scope for this issue.

barakeinav1:
Closes https://github.com/near/mpc/issues/3409.

## Summary

- Installs a Unix signal handler in `mpc-node` that catches **SIGTERM**, **SIGINT**, and **SIGHUP**, routing any of them into the existing internal shutdown channel — operators stopping the node via dstack CVM stop / `docker stop` / `kubectl delete` / `systemctl stop` (or a dev pressing Ctrl-C) now get graceful shutdown semantics where before these signals were effectively SIGKILL.
- After the main `select!` exits, calls `near_async::shutdown_all_actors()` so nearcore's actor system can commit any in-flight RocksDB batches before the process exits.
- Handler is installed **as the first thing** after the tokio runtime is built — before `spawn_real_indexer` and the expensive startup (attestation generation, contract state fetch) — so signals arriving during early startup are also handled gracefully.

Closes #3409. Does NOT close the upstream nearcore restart panic at `streamer/mod.rs:207` — that's filed upstream as [`near/nearcore#15867`](https://github.com/near/nearcore/issues/15867); investigation context in near/mpc#3409. Campaign data shows this handler gets the back-migration e2e from 0/5 to 1/5 pass — a real production improvement, but not a full fix for the upstream non-determinism.

## Notable non-action: no `RocksDB::block_until_all_instances_are_dropped()`

neard's standalone binary follows `shutdown_all_actors()` with `RocksDB::block_until_all_instances_are_dropped()`. This PR deliberately omits that next step because in our embedding it hangs indefinitely: the indexer runs in a `std::thread::spawn`'d closure whose `block_on` never returns — the spawned `monitor_*`/`indexer_logger` tasks hold `Arc<IndexerState>` → `Arc<RocksDB>` references that nothing currently cancels on shutdown. Calling `block_until_all_instances_are_dropped` would block the SIGTERM path past any reasonable grace period and the orchestrator would SIGKILL us anyway. A proper fix wires a `CancellationToken` through the indexer thread — that's [PR #3486](https://github.com/near/mpc/pull/3486), stacked on top of this one. The inline comment in `run.rs` captures the rationale.

## Test plan

- [x] e2e test added: `sigterm_handler__should_exit_cleanly_instead_of_default_terminating`. Sends SIGTERM to a running cluster node, asserts the process exits via its own `main()` (i.e. `status.code().is_some()`) rather than being default-terminated by the OS (`status.signal().is_some()`). Would fail under revert.
- [x] CI green (cargo check / clippy / fmt on the new code).
- [x] No regression: existing e2e tests pass.

## Known related issues outside this PR

- **`deployment/start.sh` doesn't `exec` mpc-node** — the non-TEE Docker `Dockerfile-node` uses `CMD ["/app/start.sh"]`, and `start.sh` runs `/app/mpc-node start` directly (not via `exec`). That means bash is PID 1 in the container; bash doesn't forward SIGTERM to its child, so for the non-TEE Docker deployment this handler still never receives the signal. The TEE deployment (via `crates/tee-launcher/mpc-node-docker-compose.tee.template.yml`) is unaffected — it uses `command: ["/app/mpc-node", ...]` directly, making mpc-node PID 1. The start.sh fix is tracked separately as #3487.

### DIFF
--- a/crates/e2e-tests/Cargo.toml
+++ b/crates/e2e-tests/Cargo.toml
@@ -15,7 +15,7 @@ futures = { workspace = true }
 hex = { workspace = true }
 httpmock = { workspace = true }
 launcher-interface = { workspace = true }
-libc = "0.2"
+libc = { workspace = true }
 mpc-node-config = { workspace = true }
 near-account-id = { workspace = true }
 near-indexer-primitives = { workspace = true }
--- a/crates/e2e-tests/Cargo.toml
+++ b/crates/e2e-tests/Cargo.toml
@@ -15,6 +15,7 @@ futures = { workspace = true }
 hex = { workspace = true }
 httpmock = { workspace = true }
 launcher-interface = { workspace = true }
+libc = "0.2"
 mpc-node-config = { workspace = true }
 near-account-id = { workspace = true }
 near-indexer-primitives = { workspace = true }
--- a/crates/e2e-tests/src/mpc_node.rs
+++ b/crates/e2e-tests/src/mpc_node.rs
@@ -39,10 +39,13 @@ impl MpcNode {
         self.setup
     }
 
-    /// Send SIGTERM to the node and wait up to `grace` for it to exit on its
-    /// own. Returns the child's `ExitStatus` on graceful exit; errors if the
-    /// grace period elapses (in which case `self.process`'s Drop will SIGKILL
-    /// it as a fallback). Used to exercise the production SIGTERM handler.
+    /// Send SIGTERM and wait up to `grace` for the node to exit. If grace
+    /// expires before the process exits on its own, send SIGKILL and wait
+    /// for it explicitly (instead of relying on `ProcessGuard::Drop` to do
+    /// the same). Either way, returns the child's `ExitStatus`:
+    /// `status.success()` ⇒ graceful exit via mpc-node's SIGTERM handler;
+    /// `status.signal() == Some(9)` ⇒ SIGKILL fallback fired. Used to
+    /// exercise the production SIGTERM handler.
     pub fn terminate_with_sigterm(
         mut self,
         grace: std::time::Duration,
@@ -65,9 +68,19 @@ impl MpcNode {
                 Some(status) => return Ok((status, self.setup)),
                 None => {
                     if start.elapsed() >= grace {
-                        anyhow::bail!(
-                            "mpc-node pid {pid} did not exit within {grace:?} after SIGTERM"
-                        );
+                        // Grace expired: explicit SIGKILL fallback, then wait
+                        // for the child to actually exit so we return a real
+                        // `ExitStatus` rather than bailing.
+                        self.process
+                            .0
+                            .kill()
+                            .context("SIGKILL fallback on grace-period expiry failed")?;
+                        let status = self
+                            .process
+                            .0
+                            .wait()
+                            .context("wait after SIGKILL fallback failed")?;
+                        return Ok((status, self.setup));
                     }
                     std::thread::sleep(poll_interval);
                 }
--- a/crates/e2e-tests/src/mpc_node.rs
+++ b/crates/e2e-tests/src/mpc_node.rs
@@ -39,6 +39,42 @@ impl MpcNode {
         self.setup
     }
 
+    /// Send SIGTERM to the node and wait up to `grace` for it to exit on its
+    /// own. Returns the child's `ExitStatus` on graceful exit; errors if the
+    /// grace period elapses (in which case `self.process`'s Drop will SIGKILL
+    /// it as a fallback). Used to exercise the production SIGTERM handler.
+    pub fn terminate_with_sigterm(
+        mut self,
+        grace: std::time::Duration,
+    ) -> anyhow::Result<(std::process::ExitStatus, MpcNodeSetup)> {
+        let pid = self.process.0.id() as libc::pid_t;
+        // SAFETY: `libc::kill` with `SIGTERM` and a pid Rust's Child reports
+        // for a live child is well-defined. We make no assumptions beyond
+        // that the pid is the child we spawned.
+        let rc = unsafe { libc::kill(pid, libc::SIGTERM) };
+        anyhow::ensure!(
+            rc == 0,
+            "libc::kill({pid}, SIGTERM) failed: {}",
+            std::io::Error::last_os_error()
+        );
+
+        let start = std::time::Instant::now();
+        let poll_interval = std::time::Duration::from_millis(100);
+        loop {
+            match self.process.0.try_wait()? {
+                Some(status) => return Ok((status, self.setup)),
+                None => {
+                    if start.elapsed() >= grace {
+                        anyhow::bail!(
+                            "mpc-node pid {pid} did not exit within {grace:?} after SIGTERM"
+                        );
+                    }
+                    std::thread::sleep(poll_interval);
+                }
+            }
+        }
+    }
+
     /// Kill then start. New process, same config and data directory.
     pub fn restart(self) -> anyhow::Result<MpcNode> {
         self.kill().start()
--- a/crates/e2e-tests/tests/sigterm_handler.rs
+++ b/crates/e2e-tests/tests/sigterm_handler.rs
@@ -3,15 +3,16 @@ use std::time::Duration;
 
 use crate::common;
 
-/// Verifies that mpc-node's SIGTERM handler initiates a graceful shutdown
-/// instead of letting the OS default-terminate the process.
+/// Verifies that mpc-node's SIGTERM handler drives a clean graceful shutdown
+/// (exit code 0), not just any process termination.
 ///
-/// Would fail under revert: without the handler in `crates/node/src/run.rs`,
-/// SIGTERM hits the process with no handler installed and the OS terminates
-/// the process directly. `status.signal()` is then `Some(15)` (SIGTERM) and
-/// `status.code()` is `None`, which the assertion below catches. With the
-/// handler installed the process exits via `main`'s normal return path,
-/// `status.code()` is `Some(_)`, and the assertion passes.
+/// Would fail under revert in two ways:
+/// - Without the SIGTERM handler at all, the OS default-terminates the
+///   process and `status.success()` is false (`code()` is `None`,
+///   `signal()` is `Some(15)`).
+/// - With a handler that returns an error from `run_mpc_node` on signal
+///   shutdown (the previous behavior, before the signal/image-hash split),
+///   `status.success()` is also false (exit code 1).
 #[tokio::test(flavor = "multi_thread")]
 #[expect(non_snake_case)]
 async fn sigterm_handler__should_exit_cleanly_instead_of_default_terminating() {
@@ -24,10 +25,11 @@ async fn sigterm_handler__should_exit_cleanly_instead_of_default_terminating() {
         .terminate_node_with_sigterm(0, Duration::from_secs(30))
         .expect("node did not exit within the SIGTERM grace period");
 
-    // Then: the process exited via its own main(), not by OS signal.
+    // Then: the process exited cleanly with code 0.
     assert!(
-        status.code().is_some(),
-        "mpc-node was terminated by signal {:?} instead of exiting cleanly via the SIGTERM handler",
+        status.success(),
+        "mpc-node did not exit cleanly after SIGTERM: code={:?} signal={:?}",
+        status.code(),
         status.signal()
     );
 }
--- a/crates/e2e-tests/tests/sigterm_handler.rs
+++ b/crates/e2e-tests/tests/sigterm_handler.rs
@@ -1,0 +1,33 @@
+use std::os::unix::process::ExitStatusExt;
+use std::time::Duration;
+
+use crate::common;
+
+/// Verifies that mpc-node's SIGTERM handler initiates a graceful shutdown
+/// instead of letting the OS default-terminate the process.
+///
+/// Would fail under revert: without the handler in `crates/node/src/run.rs`,
+/// SIGTERM hits the process with no handler installed and the OS terminates
+/// the process directly. `status.signal()` is then `Some(15)` (SIGTERM) and
+/// `status.code()` is `None`, which the assertion below catches. With the
+/// handler installed the process exits via `main`'s normal return path,
+/// `status.code()` is `Some(_)`, and the assertion passes.
+#[tokio::test(flavor = "multi_thread")]
+#[expect(non_snake_case)]
+async fn sigterm_handler__should_exit_cleanly_instead_of_default_terminating() {
+    // Given: a running cluster.
+    let (mut cluster, _running) =
+        common::must_setup_cluster(common::SIGTERM_HANDLER_PORT_SEED, |_| {}).await;
+
+    // When: SIGTERM is delivered to node 0 with a 30s grace period.
+    let status = cluster
+        .terminate_node_with_sigterm(0, Duration::from_secs(30))
+        .expect("node did not exit within the SIGTERM grace period");
+
+    // Then: the process exited via its own main(), not by OS signal.
+    assert!(
+        status.code().is_some(),
+        "mpc-node was terminated by signal {:?} instead of exiting cleanly via the SIGTERM handler",
+        status.signal()
+    );
+}
--- a/crates/node/src/run.rs
+++ b/crates/node/src/run.rs
@@ -54,22 +54,10 @@ async fn await_optional_signal(handle: Option<&mut Signal>) {
     }
 }
 
-/// Listen for SIGTERM / SIGINT / SIGHUP and forward the first arrival into
-/// `sender`. Operators stopping the node via dstack CVM stop / `docker stop`
-/// / `kubectl delete` / `systemctl stop` (or a dev pressing Ctrl-C) hit this
-/// path; routing into the same channel that `monitor_allowed_image_hashes`
-/// uses lets the main `select!` exit gracefully.
-///
-/// Logs a per-signal warning if any individual install fails; if none can be
-/// installed (rare), logs an error and exits without ever forwarding —
-/// graceful shutdown via signals is disabled in that case but the rest of
-/// the node keeps running.
-///
-/// Spawned as the first thing after the tokio runtime is built so signals
-/// arriving during early startup (indexer bootstrap, contract state fetch,
-/// attestation generation) are also handled gracefully — otherwise they'd
-/// hit the process with no handler installed and the OS would terminate it
-/// immediately, functionally identical to SIGKILL.
+/// Install handlers for SIGTERM / SIGINT / SIGHUP and forward the first
+/// signal that arrives into `sender` so the main `select!` can drive a
+/// graceful shutdown. Spawned right after the runtime is built so signals
+/// arriving during early startup are also handled.
 async fn await_and_forward_shutdown_signal(sender: mpsc::Sender<()>) {
     let install = |kind: SignalKind, name: &'static str| {
         signal(kind)
@@ -136,10 +124,13 @@ pub async fn run_mpc_node(config: StartConfig) -> anyhow::Result<()> {
 
     let _tokio_enter_guard = root_runtime.enter();
 
-    let (shutdown_signal_sender, mut shutdown_signal_receiver) = mpsc::channel(1);
-    root_runtime.spawn(await_and_forward_shutdown_signal(
-        shutdown_signal_sender.clone(),
-    ));
+    // Two independent shutdown sources, each owns its own channel so the main
+    // `select!` can give them distinct exit semantics: a graceful operator stop
+    // (SIGTERM/SIGINT/SIGHUP) exits cleanly with code 0; the image-hash watcher
+    // exits non-zero on the rare error conditions that cause it to bail.
+    let (signal_shutdown_sender, mut signal_shutdown_receiver) = mpsc::channel(1);
+    let (image_hash_shutdown_sender, mut image_hash_shutdown_receiver) = mpsc::channel(1);
+    root_runtime.spawn(await_and_forward_shutdown_signal(signal_shutdown_sender));
 
     // Load configuration and initialize persistent secrets
     let node_config = config.node.clone();
@@ -255,7 +246,7 @@ pub async fn run_mpc_node(config: StartConfig) -> anyhow::Result<()> {
         *config.tee.image_hash,
         allowed_hashes_in_contract,
         image_hash_storage,
-        shutdown_signal_sender.clone(),
+        image_hash_shutdown_sender,
     ));
 
     let home_dir = config.home_dir.clone();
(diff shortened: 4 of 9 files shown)


barakeinav1:
Thanks for the thorough review — addressed all five findings in the latest push (`95087d4c`).

> `crates/node/src/run.rs:195` — The inline comment points readers to `docs/investigation/nearcore-indexer-sigkill-restart-panic.md`, [...] Neither this PR nor `main` actually contains it [...] Either add the doc in this PR or drop the reference so future readers don't chase a broken link.

Dropped the reference. The inline comment now just explains *why* `block_until_all_instances_are_dropped()` is omitted; the longer context lives in #3409 (and the actual fix is in #3486).

> `crates/node/src/run.rs:184-210` — No test exercises the new path; engineering-standards.md asks that every change have a test that would fail under revert.

Added `sigterm_handler__should_exit_cleanly_instead_of_default_terminating`. It sends SIGTERM via `libc::kill` and asserts `status.code().is_some()` (clean main() exit) vs `status.signal().is_some()` (OS default-terminate). Fails under revert.

> `crates/node/src/run.rs:199` — Only SIGTERM is wired up. SIGINT (Ctrl-C) and SIGHUP still default-terminate.

All three are now caught in one `tokio::select!`, racing whichever arrives first. Each installs independently — a per-signal install failure logs a warning naming that signal but doesn't disable the others.

> `crates/node/src/run.rs:203` — `let _ = shutdown_for_sigterm.send(()).await;` is the right call [...] but a one-line note explaining that would save the next reader a moment.

Added a comment on that line: `// "send" returns Err only if the receiver has been dropped (the main "select!" already exited via a different arm); we're shutting down anyway, so it's fine to ignore.`

> `crates/node/src/run.rs:198` — Spawning the handler on `root_runtime` after the indexer thread is already spawned introduces a small race window where SIGTERM arriving early in `run_mpc_node` still default-terminates the process. Moving the handler install to the top of the function [...] closes that window.

Moved. The listener is now spawned immediately after the tokio runtime is built, before `spawn_real_indexer` / attestation generation / contract state fetch. The whole spawned task is also now an extracted top-level fn `await_and_forward_shutdown_signal`, so `run_mpc_node`'s body shrinks by ~25 lines and the signal logic gets a docstring.

crates/node/src/run.rs
```diff
@@ -41,6 +42,64 @@ use crate::tee::{
...
+/// Listen for SIGTERM / SIGINT / SIGHUP and forward the first arrival into
+/// `sender`. Operators stopping the node via dstack CVM stop / `docker stop`
+/// / `kubectl delete` / `systemctl stop` (or a dev pressing Ctrl-C) hit this
+/// path; routing into the same channel that `monitor_allowed_image_hashes`
+/// uses lets the main `select!` exit gracefully.
+///
+/// Logs a per-signal warning if any individual install fails; if none can be
+/// installed (rare), logs an error and exits without ever forwarding —
+/// graceful shutdown via signals is disabled in that case but the rest of
+/// the node keeps running.
```
anodar: nit: this could just be a single sentence saying that we're registering sigterm/sigint/sighup handlers and notify sender once we receive it. Or something a bit shorter than what currently is.

crates/node/src/run.rs
```diff
@@ -41,6 +42,64 @@ use crate::tee::{
...
+    if sigterm.is_none() && sigint.is_none() && sighup.is_none() {
+        tracing::error!(
+            "no shutdown-signal handlers could be installed; graceful shutdown disabled"
+        );
+        return;
+    }
+    let signal_name = tokio::select! {
+        _ = await_optional_signal(sigterm.as_mut()) => "SIGTERM",
+        _ = await_optional_signal(sigint.as_mut()) => "SIGINT",
+        _ = await_optional_signal(sighup.as_mut()) => "SIGHUP",
```
anodar: Just for my curiosity, do we really need sighup here? I don't see this getting SIGHUP other than explicitly sending that signal: `docker kill --signal=HUP`.

crates/node/src/run.rs
```diff
@@ -228,6 +291,22 @@ pub async fn run_mpc_node(config: StartConfig) -> anyhow::Result<()> {
...
+    // in-flight RocksDB batches before the process exits. We deliberately do
+    // NOT then call `near_store::db::RocksDB::block_until_all_instances_are_dropped()`
+    // (which neard's standalone binary does next) — our embedded indexer
+    // runs in a separate `std::thread::spawn`'d closure whose `block_on`
+    // never returns, because the spawned monitor tasks hold
+    // `Arc<IndexerState>` → `Arc<RocksDB>` references that nothing
+    // currently cancels. Calling `block_until_all_instances_are_dropped`
+    // here would hang the SIGTERM path past any reasonable grace period;
+    // the orchestrator would then SIGKILL us and we'd land in the exact
+    // uncommitted-RocksDB state this handler was supposed to prevent.
```
anodar: nit: there's too much detail, I think it would be more useful to explain high level reasoning that `block_until_all_instances_are_dropped` would just hang SIGTERM and would result node eventually being SIGKILLED.

crates/node/src/run.rs
anodar: Error message here isn't correct anymore, this signal can also come from `await_and_forward_shutdown_signal`.


I'd recommend to send different type of message in channel from the latter to differentiate the two and not just for error message but after graceful shutdown I'd expect it to exit (0) not with error code.

crates/e2e-tests/tests/sigterm_handler.rs
```diff
@@ -0,0 +1,33 @@
...
+        common::must_setup_cluster(common::SIGTERM_HANDLER_PORT_SEED, |_| {}).await;
+
+    // When: SIGTERM is delivered to node 0 with a 30s grace period.
+    let status = cluster
+        .terminate_node_with_sigterm(0, Duration::from_secs(30))
+        .expect("node did not exit within the SIGTERM grace period");
+
+    // Then: the process exited via its own main(), not by OS signal.
+    assert!(
+        status.code().is_some(),
```
anodar: This doesn't really check that shutdown was graceful, it could have been sigkilled as well. 

If you address suggestion here: https://github.com/near/mpc/pull/3410/changes#r3375791083, you can then check that status.success here instead, which then should tells you whether it was graceful or not.

anodar (dismissed):
Left some comments, but looks good otherwise.

crates/node/src/run.rs
barakeinav1: Done — trimmed to ~3 lines.

crates/node/src/run.rs
barakeinav1: Probably not strictly needed — added it for completeness after the Copilot reviewer flagged the SIGTERM/SIGINT/SIGHUP trio. Cost is one extra line and the behavior if we ever do receive a SIGHUP (graceful shutdown) is the right thing, so leaving it in.

crates/node/src/run.rs
barakeinav1: Done — trimmed.

crates/node/src/run.rs
barakeinav1: Good catch. Refactored to two independent channels — `signal_shutdown_sender` (owned by the signal handler) and `image_hash_shutdown_sender` (owned by the watcher). Each gets its own `select!` arm:

- Signal arm: logs `shutdown signal received; exiting cleanly` and returns `Ok(())` → process exits with code 0.
- Image-hash arm: returns `Err(anyhow!("TEE allowed-image-hashes watcher exited unexpectedly."))` → process exits non-zero, preserving the existing semantics for that path. The specific underlying error (storage I/O vs indexer-channel-closed) is already in the `info!(?exit_result, "Image hash watcher exited.")` log line that fires right after, so the anyhow! doesn't need to enumerate them.

crates/e2e-tests/tests/sigterm_handler.rs
barakeinav1: Tightened to `status.success()` now that the signal-shutdown path returns `Ok(())` (exit code 0). Updated the test docstring with the new failure modes it catches: SIGKILL, OS default-terminate by signal, AND clean-exit-but-with-error-code all fail it now; only exit code 0 passes.

crates/e2e-tests/src/mpc_node.rs
```diff
@@ -39,6 +39,42 @@ impl MpcNode {
         self.setup
     }
 
+    /// Send SIGTERM to the node and wait up to `grace` for it to exit on its
+    /// own. Returns the child's `ExitStatus` on graceful exit; errors if the
+    /// grace period elapses (in which case `self.process`'s Drop will SIGKILL
+    /// it as a fallback). Used to exercise the production SIGTERM handler.
+    pub fn terminate_with_sigterm(
```
gilcu3: The usual correct way to kill a process is to send SIGTERM, wait a bit, and if the process is still alive, then send SIGKILL. We should not depend on Drop behavior for that

crates/e2e-tests/Cargo.toml
```diff
@@ -15,6 +15,7 @@ futures = { workspace = true }
 hex = { workspace = true }
 httpmock = { workspace = true }
 launcher-interface = { workspace = true }
+libc = "0.2"
```
gilcu3: we must use the workspace for dependencies

crates/e2e-tests/src/mpc_node.rs
```diff
@@ -39,6 +39,42 @@ impl MpcNode {
...
+        mut self,
+        grace: std::time::Duration,
+    ) -> anyhow::Result<(std::process::ExitStatus, MpcNodeSetup)> {
+        let pid = self.process.0.id() as libc::pid_t;
+        // SAFETY: `libc::kill` with `SIGTERM` and a pid Rust's Child reports
+        // for a live child is well-defined. We make no assumptions beyond
+        // that the pid is the child we spawned.
+        let rc = unsafe { libc::kill(pid, libc::SIGTERM) };
```
gilcu3, line 57: An alternative to avoid having to use `unsafe` is `nix::sys::signal::kill` using the `nix` crate (not related to Nix)

crates/node/src/run.rs
```diff
@@ -41,6 +42,52 @@ use crate::tee::{
...
+        None => std::future::pending().await,
+    }
+}
+
+/// Install handlers for SIGTERM / SIGINT / SIGHUP and forward the first
+/// signal that arrives into `sender` so the main `select!` can drive a
+/// graceful shutdown. Spawned right after the runtime is built so signals
+/// arriving during early startup are also handled.
+async fn await_and_forward_shutdown_signal(sender: mpsc::Sender<()>) {
+    let install = |kind: SignalKind, name: &'static str| {
```
gilcu3: nit: could we define these functions in a new module? This run file is already a bit big

crates/node/src/run.rs
```diff
@@ -41,6 +42,52 @@ use crate::tee::{
...
+        tracing::error!(
+            "no shutdown-signal handlers could be installed; graceful shutdown disabled"
+        );
+        return;
+    }
+    let signal_name = tokio::select! {
+        _ = await_optional_signal(sigterm.as_mut()) => "SIGTERM",
+        _ = await_optional_signal(sigint.as_mut()) => "SIGINT",
+        _ = await_optional_signal(sighup.as_mut()) => "SIGHUP",
+    };
```
gilcu3: I think this could be improved quite a bit. If we are sure the handlers were installed correctly (are not None), then we don't really need `await_optional_signal` as here a simple `handle.recv().await` for each would suffice, unless I am missing something

crates/node/src/run.rs
```diff
@@ -228,6 +289,15 @@ pub async fn run_mpc_node(config: StartConfig) -> anyhow::Result<()> {
...
     info!(?exit_result, "Image hash watcher exited.");
 
+    // Stop nearcore's actor system so its tasks have a chance to commit any
+    // in-flight RocksDB batches before the process exits. We deliberately
+    // skip `RocksDB::block_until_all_instances_are_dropped()` here (which
+    // neard's standalone binary calls next): in this embedding it would
+    // hang past any reasonable grace period, and the orchestrator would
+    // SIGKILL us before the graceful path completes.
+    info!("Stopping nearcore actor system.");
+    near_async::shutdown_all_actors();
```
gilcu3, line 275: If that comment is true (and I hope it is not), wouldn't we be always leaving the rocksdb part in an unstable state? I think we should do a best effort and call that here. During tests we could have both tests for graceful shutdowns, and tests for nasty shutdowns (with SIGKILL), if needed

crates/node/src/run.rs
```diff
@@ -77,6 +124,14 @@ pub async fn run_mpc_node(config: StartConfig) -> anyhow::Result<()> {
 
     let _tokio_enter_guard = root_runtime.enter();
 
+    // Two independent shutdown sources, each owns its own channel so the main
+    // `select!` can give them distinct exit semantics: a graceful operator stop
+    // (SIGTERM/SIGINT/SIGHUP) exits cleanly with code 0; the image-hash watcher
+    // exits non-zero on the rare error conditions that cause it to bail.
+    let (signal_shutdown_sender, mut signal_shutdown_receiver) = mpsc::channel(1);
+    let (image_hash_shutdown_sender, mut image_hash_shutdown_receiver) = mpsc::channel(1);
```
gilcu3: this line is out of place now, should go to before

```
    let cancellation_token = CancellationToken::new();

    let allowed_hashes_in_contract = indexer_api.allowed_docker_images_receiver.clone();
```
where it is used

gilcu3 (dismissed):
Looks okaish to me, although I have a few concerns:

- Not shutting down the rocksdb part in nearcore correctly
- Handling extra signals for no good reason I can see. For our use case we only need SIGTERM, any process sending the others would probably be wrong
- Some parts of the code can be improved

crates/e2e-tests/src/mpc_node.rs
netrome: Not following, this is not a Drop implementation. I see we rely on the Drop behavior for the final SIGKILL - and yes, calling self.kill() explicitly would be nicer - but ProcessGuard already guarantees to kill the process on Drop so I don't think it's necessarily wrong per se.

crates/node/src/run.rs
netrome: Yeah I'm curious why `RocksDB::block_until_all_instances_are_dropped()` should hang here 🤔

crates/e2e-tests/src/mpc_node.rs
gilcu3: Yeah I didn't mean that it was wrong. What I meant is that if the function purpose is to kill a node, we should implement it properly, first SIGTERM, wait, then SIGKILL. That way we don't need to depend on Drop behavior for that, making it easier to understand what is happening

crates/node/src/run.rs
```diff
@@ -41,6 +42,52 @@ use crate::tee::{
...
+
+/// Install handlers for SIGTERM / SIGINT / SIGHUP and forward the first
+/// signal that arrives into `sender` so the main `select!` can drive a
+/// graceful shutdown. Spawned right after the runtime is built so signals
+/// arriving during early startup are also handled.
+async fn await_and_forward_shutdown_signal(sender: mpsc::Sender<()>) {
+    let install = |kind: SignalKind, name: &'static str| {
+        signal(kind)
+            .inspect_err(|e| tracing::warn!(error = %e, signal = name, "failed to install shutdown-signal handler"))
+            .ok()
```
netrome: I don't think these errors should be ignored. This in combination with `await_optional_signal` effectively ignores any signal we fail to install.

Is there any case we expect these to fail that we want to tolerate? Otherwise, if the node cannot install these handlers it feels appropriate to terminate the node.

I think the code could be simplified quite a bit if we relax this assumption. We could just add each signal separately to the top-level tokio select.

netrome (changes requested):
One blocker, I don't think we should tolerate failing to install signal handlers.

crates/node/src/run.rs
anodar: To my understanding it's because indexer thread builds different runtime for indexer, and tasks running there  holds references to rocks db (through indexer_state), e.g. spawnd indexer logger: https://github.com/near/mpc/blob/b7d90d166f7d8dca3db37a42bc1e6128a7c39ed1/crates/node/src/indexer/real.rs#L146

and at this point that runtime (hence db references) aren't dropped so block_until_... should hang.

@barakeinav1 may have a better explanation though.

gilcu3: In that case shouldn't we try to end that runtime properly then?

crates/e2e-tests/src/mpc_node.rs
barakeinav1: did you mean that our should not stop the node, but let the caller send SIGKILL?
or that we should send sigkill after some timeout?

gilcu3: > that we should send sigkill after some timeout?

this one. IMO the function should be a simple "kill this node" and the signals are just implementation details

barakeinav1: Applied. `terminate_with_sigterm` now does an explicit `self.process.0.kill()` + `wait()` when grace expires, instead of bailing and leaving the SIGKILL to `ProcessGuard::Drop`. Returns an `ExitStatus` either way — `status.success()` ⇒ graceful, `status.signal() == Some(9)` ⇒ SIGKILL fallback fired.

crates/e2e-tests/Cargo.toml
barakeinav1: Applied — `libc` is now in `[workspace.dependencies]` and the e2e-tests crate uses `libc = { workspace = true }`.

crates/e2e-tests/src/mpc_node.rs
barakeinav1: Considered but keeping `libc` — adding a new direct dep to avoid a 3-line documented `unsafe` block doesn't feel worth it (the SAFETY note covers the conditions). Easy to swap later if the team would rather take the dep.

crates/node/src/run.rs
barakeinav1: Moot after round 2 — the two helper functions are gone, SIGTERM is installed inline in `run_mpc_node`. No new module needed.

crates/node/src/run.rs
barakeinav1: Addressed by the round-2 simplification — dropped SIGINT/SIGHUP per the offline discussion. so only one signal remains. `await_optional_signal` is gone; the SIGTERM future is raced inline in the main `tokio::select!` via a small `async {}` block that handles the install-failed case by parking on `pending()`.

crates/node/src/run.rs
barakeinav1: anodar's reply above has the technical explanation: the indexer thread holds `Arc<RocksDB>` refs via `indexer_state` that nothing currently cancels, so `block_until_all_instances_are_dropped()` would loop forever waiting for the refcount to reach zero, and the orchestrator would SIGKILL us before we could return — landing us in exactly the uncommitted-state we were trying to prevent.

PR #3486 is the follow-up that actually drains the indexer thread (wraps `listen_blocks` in a `tokio::select!` race against a cancellation token, drops the runtime to release the refs) and re-adds the `block_until_all_instances_are_dropped()` call after `shutdown_all_actors()`. Stacks on this PR.

On the "unstable state" worry: RocksDB's WAL guarantees committed data survives a kill regardless of whether we call `block_until_all_instances_are_dropped()`. The call mostly closes a small flush window for batches that were in-flight at shutdown time. So in this PR we lose that window; #3486 reclaims it.

crates/node/src/run.rs
barakeinav1: Applied — moved the `image_hash_shutdown_sender` channel declaration down next to the `monitor_allowed_image_hashes` spawn that uses it.

crates/node/src/run.rs
barakeinav1: Discussed in Slack — keeping the install failure non-fatal. Now that the code only handles SIGTERM (per round 2), the failure mode if install fails is just the pre-PR baseline: SIGTERM falls through to OS-default-terminate. No regression vs main. 

I really prefer not to terminat on a "non critical" functionality failure

crates/node/src/run.rs
barakeinav1: removed SIGINT/SIGHUP per slack discussion

crates/node/src/run.rs
netrome: Fair, additionally we'd want to keep the allowed image hash watcher live even if this fails - so I'm okay with this.

netrome (approved):
Not a hard blocker from my side - only a shallow review though so would be good to await approvals from both Reynaldo and Nodar.

gilcu3 (approved):

anodar (approved):
Thanks for the update!

PR #3412, opened 2026-05-31
docs(operator-guide): require empty Pre-launch Script and Init Script on manual deploy

barakeinav1:
## What

Clarifies the **Web interface (manual)** deploy steps in the operator guide: both the **Pre-launch Script** and **Init Script** fields must be left empty, and explains why.

The dstack-vmm WebUI pre-fills the Pre-launch Script with a default launch-token check. The MPC attestation verifier binds the whole `app-compose.json` into RTMR3 and requires `pre_launch_script` to be absent (`validate_app_compose_config` → `pre_launch_script.is_none()`), so a node deployed with the pre-filled script fails attestation and is rejected by peers and the contract.

## Change

One line in `docs/running-an-mpc-node-in-tdx-external-guide.md` (step 3 of the Web-interface deploy settings), expanded from "Pre script - empty." to name both fields and state the consequence.

Closes #3411

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3414, opened 2026-05-31
fix(attestation): reject non-empty app-compose init_script

### LINKED ISSUE #3413: TEE attestation does not verify app-compose `init_script` — arbitrary root code in CVM passes verification
## Background

TEE attestation binds the whole `app-compose.json` into RTMR3 via the compose hash, but `validate_app_compose_config` (`crates/attestation/src/attestation.rs`) only *validates* a fixed set of fields. It deliberately requires `pre_launch_script` to be **absent**, because that field is arbitrary root code run before the workload.

dstack (>= 0.5.5) added a second such field, **`init_script`** — a bash script run as **root, before `dockerd` starts**. MPC does not verify it:

- The `AppCompose` struct (`crates/attestation/src/app_compose.rs`) does not model `init_script`, and has no `#[serde(deny_unknown_fields)]`, so it is silently ignored on parse.
- `validate_app_compose_config` has no rule for it.
- The only allow-listed compose value is `sha256(app_compose.docker_compose_file)` — just the `docker_compose_file` field, not the whole app-compose (`crates/mpc-attestation/src/attestation.rs:192`).
- The measurements allow-list pins only MRTD/RTMR0/RTMR1/RTMR2 — **not RTMR3** (`crates/attestation/src/measurements.rs`).

**Impact:** a participant can deploy a CVM with an arbitrary `init_script` that runs as root inside the TEE before the MPC node starts, and the attestation still passes contract- and peer-side verification. This defeats the guarantee that only approved code runs in the enclave. `init_script` runs *earlier* than the `pre_launch_script` we already forbid, so the missing check is at least as important.

## Acceptance Criteria

- [ ] `init_script` is modeled in `AppCompose` and verification rejects any attestation whose `init_script` is non-empty (mirroring `pre_launch_script`).
- [ ] Test coverage for the rejection.
- [ ] (Discussion) decide whether to make `AppCompose` fail-closed on unknown fields (`#[serde(deny_unknown_fields)]`) so a future dstack-added executable field can't silently bypass verification again. Note the parallel unused DTO `AppCompose` in `crates/near-mpc-contract-interface/src/types/attestation.rs` (also missing the field) — keep in sync or remove if dead.

## Resources

- `crates/attestation/src/app_compose.rs` — `AppCompose`
- `crates/attestation/src/attestation.rs` — `validate_app_compose_config`
- `crates/mpc-attestation/src/attestation.rs:192` — launcher compose hash = `sha256(docker_compose_file)`
- dstack security doc — `init_script` "executed prior to dockerd startup" (added 0.5.5)

barakeinav1:
## What

Verification rejected a non-empty `pre_launch_script` but ignored dstack's `init_script` (added in 0.5.5), a bash script run as **root before dockerd**. It was measured into the compose hash but never pinned or validated, so a CVM with an arbitrary `init_script` passed attestation — arbitrary root code in the enclave before the MPC node.

## Change

- `AppCompose` (`crates/attestation/src/app_compose.rs`) now models `init_script: Option<String>`.
- `validate_app_compose_config` requires `init_script.is_none()`, mirroring `pre_launch_script`.
- Added rejection tests for both `init_script` and `pre_launch_script`; existing real-fixture deserialization tests still pass (the field is optional → missing parses as `None`).
- Operator guide (`docs/running-an-mpc-node-in-tdx-external-guide.md`): one-line note that both the Pre-launch Script and Init Script fields must be empty on the manual/WebUI deploy path (folds in the superseded #3412).

## Notes / follow-up

- The launcher-compose allow-list only covers `sha256(docker_compose_file)` and the measurements allow-list does not pin RTMR3, which is why an unvalidated app-compose field could slip through. Whether to add `#[serde(deny_unknown_fields)]` so future dstack-added executable fields fail closed is tracked in #3415.
- There is a parallel, currently-unused DTO `AppCompose` in `crates/near-mpc-contract-interface` that also lacks the field; left untouched here.

Closes #3413

### DIFF
--- a/crates/attestation/src/app_compose.rs
+++ b/crates/attestation/src/app_compose.rs
@@ -24,6 +24,11 @@ pub struct AppCompose {
     pub no_instance_id: bool,
     pub secure_time: Option<bool>,
     pub pre_launch_script: Option<String>,
+    /// Bash script that dstack (>= 0.5.5) executes as root, before the docker
+    /// daemon starts. Like `pre_launch_script`, it is arbitrary pre-workload
+    /// code, so we model it explicitly and require it to be absent during
+    /// verification (see `validate_app_compose_config`).
+    pub init_script: Option<String>,
     // The following fields that don't have any security implication are omitted:
     //
     // - docker_config: JsonValue,
--- a/crates/attestation/src/app_compose.rs
+++ b/crates/attestation/src/app_compose.rs
@@ -24,10 +24,6 @@ pub struct AppCompose {
     pub no_instance_id: bool,
     pub secure_time: Option<bool>,
     pub pre_launch_script: Option<String>,
-    /// Bash script that dstack (>= 0.5.5) executes as root, before the docker
-    /// daemon starts. Like `pre_launch_script`, it is arbitrary pre-workload
-    /// code, so we model it explicitly and require it to be absent during
-    /// verification (see `validate_app_compose_config`).
     pub init_script: Option<String>,
     // The following fields that don't have any security implication are omitted:
     //
--- a/crates/attestation/src/attestation.rs
+++ b/crates/attestation/src/attestation.rs
@@ -388,6 +388,7 @@ impl DstackAttestation {
             && app_compose.allowed_envs.is_empty()
             && app_compose.no_instance_id
             && app_compose.pre_launch_script.is_none()
+            && app_compose.init_script.is_none()
     }
 
     /// Verifies local key-provider event digest matches the expected digest.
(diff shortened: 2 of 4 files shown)


netrome (dismissed):

netrome:
Thanks for spotting and fixing this! I feel we need to be very careful with this kind of vulnerabilities popping up in dstack version changes.

SimonRastikian (dismissed):
Thanks
Where can I find the discussion mentioned in the issue's acceptance criteria?

barakeinav1:
@claude thanks — addressed in 7accf0f6:
- **Parallel DTO**: synced `init_script` onto the public `AppCompose` in `near-mpc-contract-interface` (confirmed it's re-exported as public surface via `types/lib.rs`).
- **`deny_unknown_fields`**: agreed this is the fail-closed fix for the whole class; tracked as a follow-up in #3415 rather than this PR.
- **Test names**: renamed to `..._rejects_present_*` since `is_none()` rejects any present value (incl. `Some("")`), not just non-empty scripts.

barakeinav1:
@SimonRastikian the "(Discussion)" item in the issue's acceptance criteria — whether to make `AppCompose` fail closed via `#[serde(deny_unknown_fields)]` so future dstack-added fields can't silently bypass verification — is tracked in the follow-up investigation issue #3415. That's where that discussion lives.

barakeinav1:
@netrome fully agree. The systemic angle — making app-compose verification fail closed so a new dstack field can't silently slip through on a version bump — is captured in #3415.

gilcu3 (approved):
Thank you!

netrome (approved):

PR #3420, opened 2026-06-01
fix(attestation): fail closed on unknown app-compose fields; reject tproxy_enabled

### LINKED ISSUE #3415: Investigate/harden app-compose verification: unverified fields beyond `init_script` (deny_unknown_fields, tproxy_enabled, …)
> Investigation / hardening follow-up to #3413 (fixed by #3414, which added the `init_script` check). `init_script` was one instance of a broader class: app-compose fields that are **measured** but not **validated or pinned**. This issue captures the full audit and the open hardening decisions.

## Background

The contract-side TEE verification path is `Attestation::verify` (`crates/mpc-attestation/src/attestation.rs`) → `DstackAttestation::verify` → `verify_app_compose` / `validate_app_compose_config` / `verify_any_measurements` (`crates/attestation/`).

What it pins today:
- MPC image hash → `allowed_mpc_docker_image_hashes` (RTMR3 `mpc-image-hash` event)
- Launcher compose → `sha256(app_compose.docker_compose_file)` ∈ `allowed_launcher_compose_hashes` (`crates/mpc-attestation/src/attestation.rs:192`)
- Key provider → `key_provider_event_digest` (measurements)
- MRTD / RTMR0 / RTMR1 / RTMR2 → measurements allow-list (`crates/attestation/src/measurements.rs`)
- `report_data` (TLS + account key)
- Named app-compose toggles in `validate_app_compose_config`: `manifest_version==2`, `runner=="docker-compose"`, `kms_enabled==false`, `gateway_enabled==Some(false)`, `public_logs`, `public_sysinfo`, `local_key_provider_enabled`, `allowed_envs.is_empty()`, `no_instance_id`, `pre_launch_script.is_none()`, `init_script.is_none()` (added in #3414)

**Root cause of the `init_script` gap:** the whole `app-compose.json` is bound into RTMR3 via the compose hash, but RTMR3 is *not* pinned to an allowed value (instance-id/app-id are dynamic), and the launcher allow-list only covers the `docker_compose_file` field. So the only protection for the *rest* of app-compose is the explicit field list in `validate_app_compose_config`. Any field not in that list is silently ignored — `AppCompose` has no `#[serde(deny_unknown_fields)]`.

## Findings

| Field / aspect | Severity | Notes |
|---|---|---|
| No `#[serde(deny_unknown_fields)]` on `AppCompose` | **Medium (systemic)** | Any future dstack-added executable/security field bypasses verification silently, exactly as `init_script` did. Fail-closed is the right default for a verifier; trade-off is the struct must track dstack's field set across upgrades. |
| `tproxy_enabled` modeled but **not checked** | **Medium** | `tproxy` is the legacy name for the gateway. `gateway_enabled` is forced `false`, but `tproxy_enabled` is unchecked. Confirm against the dstack version in use whether `tproxy_enabled==true` can still expose the node; if so, require `None`/`Some(false)`. |
| `key_provider` enum + `key_provider_id` not validated | Low | Likely fine — the measured `key-provider` event digest is the real binding. Confirm. |
| `docker_config` ignored (comment: "no security implication") | Low | Not pinned. Image is digest-pinned, so registry swap is moot; but it is an unverified app-compose field. Covered automatically if `deny_unknown_fields` lands. |
| `secure_time` not required | Info | Intentional (explicit test allows insecure time). Re-confirm this is acceptable for the threat model. |
| RTMR3 not pinned in measurements | Info | By design (dynamic instance-id/app-id). This is *why* unmodeled fields are the blind spot. |
| `runner=="docker-compose"` is load-bearing | Info | It neutralizes the `bash_script` field (dstack only runs `bash_script` when `runner=="bash"`). Add a comment so it isn't relaxed accidentally. |
| Parallel unused DTO `AppCompose` | Info | `crates/near-mpc-contract-interface/src/types/attestation.rs` has a second `AppCompose` (no usages found, also missing `init_script`). Keep in sync or delete if dead. |

## Acceptance Criteria

- [ ] Decide on `#[serde(deny_unknown_fields)]` for `AppCompose` (fail-closed). If adopted, model all fields dstack emits for the target version(s) and add/update fixtures.
- [ ] Add a `tproxy_enabled` check (or document why it's unnecessary), confirmed against the deployed dstack version.
- [ ] Confirm `key_provider` / `key_provider_id` and `docker_config` carry no unverified security implication (or model them).
- [ ] Add a comment noting the `runner=="docker-compose"` check is what neutralizes `bash_script`.
- [ ] Resolve the duplicate/unused DTO `AppCompose` (sync or remove).

## Resources

- `crates/attestation/src/app_compose.rs` — `AppCompose`
- `crates/attestation/src/attestation.rs` — `validate_app_compose_config`, `verify_app_compose`, `verify_any_measurements`
- `crates/attestation/src/measurements.rs` — `Measurements` (pins MRTD/RTMR0/1/2 only)
- `crates/mpc-attestation/src/attestation.rs:192` — launcher compose hash = `sha256(docker_compose_file)`
- #3413 / #3414 — the `init_script` instance of this class

barakeinav1:
Closes https://github.com/near/mpc/issues/3415

follow up for #3414 
## What

Makes app-compose verification fail closed, and closes a second bypass of the `init_script` class.

- `#[serde(deny_unknown_fields)]` on `AppCompose` + the full dstack field set modeled, so a field added by a future dstack version halts verification instead of being silently ignored.
- Reject `tproxy_enabled == Some(true)` (legacy gateway alias — dstack enables the gateway on `gateway_enabled || tproxy_enabled`) and `bash_script`, alongside the existing `pre_launch_script`/`init_script` checks.
- Sync `init_script` onto the public `AppCompose` DTO in `near-mpc-contract-interface` (dedupe tracked in #3425).
- Drop unused Borsh derives on `AppCompose`.

Verified end-to-end on a real-TDX localnet: a real dstack 0.5.8 CVM attestation is accepted by the hardened contract (Running 2-node cluster, ECDSA sign).

### DIFF
--- a/crates/attestation/src/app_compose.rs
+++ b/crates/attestation/src/app_compose.rs
@@ -49,19 +49,20 @@ pub struct AppCompose {
     pub bash_script: Option<String>,
     // The fields below are absorbed only to satisfy `deny_unknown_fields`; they are never read, so
     // they deserialize as `IgnoredAny` to avoid allocating attacker-controlled JSON from the
-    // untrusted app-compose. `key_provider` integrity is pinned by the measured `key-provider`
-    // event digest (`verify_key_provider_digest`), not this field. `port_policy` is a dstack 0.5.11
-    // gateway field (0.5.8 never emits it), inert while the gateway is disabled.
+    // untrusted app-compose.
     #[serde(default)]
     pub features: Option<IgnoredAny>,
     #[serde(default)]
     pub public_tcbinfo: Option<bool>,
+    // Key-provider integrity is pinned by the measured `key-provider` event digest
+    // (`verify_key_provider_digest`), not this field.
     #[serde(default)]
     pub key_provider: Option<IgnoredAny>,
     #[serde(default)]
     pub storage_fs: Option<String>,
     #[serde(default)]
     pub swap_size: Option<IgnoredAny>,
+    // dstack 0.5.11 gateway field (0.5.8 never emits it); inert while the gateway is disabled.
     #[serde(default)]
     pub port_policy: Option<IgnoredAny>,
     #[serde(default)]
--- a/crates/attestation/src/app_compose.rs
+++ b/crates/attestation/src/app_compose.rs
@@ -1,8 +1,7 @@
 use alloc::{string::String, vec::Vec};
 use borsh::{BorshDeserialize, BorshSerialize};
 use derive_more::{Deref, From};
-use serde::{Deserialize, Serialize};
-use serde_json::Value;
+use serde::{Deserialize, Serialize, de::IgnoredAny};
 
 /// Helper struct to deserialize the `app_compose` JSON from TCB info. This is a workaround due to
 /// current limitations in the Dstack SDK.
@@ -15,7 +14,7 @@ use serde_json::Value;
 /// dstack's `AppCompose` (`dstack-types/src/lib.rs`) plus the script keys read directly via `jq`
 /// during boot (`pre_launch_script`, `init_script`, `bash_script`). Fields without a security
 /// implication are modeled only to absorb their key; they are not validated.
-#[derive(Debug, Deserialize, Serialize)]
+#[derive(Debug, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct AppCompose {
     pub manifest_version: u32,
@@ -48,23 +47,25 @@ pub struct AppCompose {
     pub init_script: Option<String>,
     #[serde(default)]
     pub bash_script: Option<String>,
-    // Modeled to absorb the key under `deny_unknown_fields`; not validated.
+    // The fields below are absorbed only to satisfy `deny_unknown_fields`; they are never read, so
+    // they deserialize as `IgnoredAny` to avoid allocating attacker-controlled JSON from the
+    // untrusted app-compose. `key_provider` integrity is pinned by the measured `key-provider`
+    // event digest (`verify_key_provider_digest`), not this field. `port_policy` is a dstack 0.5.11
+    // gateway field (0.5.8 never emits it), inert while the gateway is disabled.
     #[serde(default)]
-    pub features: Vec<String>,
+    pub features: Option<IgnoredAny>,
     #[serde(default)]
     pub public_tcbinfo: Option<bool>,
     #[serde(default)]
-    pub key_provider: Option<Value>,
+    pub key_provider: Option<IgnoredAny>,
     #[serde(default)]
     pub storage_fs: Option<String>,
     #[serde(default)]
-    pub swap_size: Option<Value>,
-    // Added in dstack 0.5.11 (0.5.8 never emits it); pre-modeled for future 0.5.11 support. Gateway
-    // port config, inert while the gateway is disabled, so no security implication.
+    pub swap_size: Option<IgnoredAny>,
     #[serde(default)]
-    pub port_policy: Option<Value>,
+    pub port_policy: Option<IgnoredAny>,
     #[serde(default)]
-    pub docker_config: Option<Value>,
+    pub docker_config: Option<IgnoredAny>,
 }
 
 /// A type that contains a docker compose the contents of a docker compose file as
--- a/crates/attestation/src/app_compose.rs
+++ b/crates/attestation/src/app_compose.rs
@@ -2,33 +2,69 @@ use alloc::{string::String, vec::Vec};
 use borsh::{BorshDeserialize, BorshSerialize};
 use derive_more::{Deref, From};
 use serde::{Deserialize, Serialize};
+use serde_json::Value;
 
 /// Helper struct to deserialize the `app_compose` JSON from TCB info. This is a workaround due to
 /// current limitations in the Dstack SDK.
 ///
 /// See: <https://github.com/Dstack-TEE/dstack/issues/267>
-#[derive(Debug, Deserialize, Serialize, BorshSerialize, BorshDeserialize)]
+///
+/// `deny_unknown_fields` makes verification fail closed: every key dstack can emit into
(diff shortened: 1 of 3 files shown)


crates/near-mpc-contract-interface/src/types/attestation.rs
```diff
@@ -253,6 +253,10 @@ impl fmt::Debug for DstackAttestation {
     BorshSerialize,
     BorshDeserialize,
 )]
+// NOTE: this is a hand-maintained mirror of the verified `attestation::AppCompose`. It is the
+// security-relevant subset and is NOT the type used during attestation verification. Keep its
+// security fields in sync with the verified struct, or it can drift (see #3425). Deduping the two
```
barakeinav1: I open a follow up issue #3425 for this

crates/attestation/src/app_compose.rs
```diff
@@ -2,33 +2,69 @@ use alloc::{string::String, vec::Vec};
...
+    #[serde(default)]
+    pub features: Vec<String>,
+    #[serde(default)]
+    pub public_tcbinfo: Option<bool>,
+    #[serde(default)]
+    pub key_provider: Option<Value>,
+    #[serde(default)]
+    pub storage_fs: Option<String>,
+    #[serde(default)]
+    pub swap_size: Option<Value>,
```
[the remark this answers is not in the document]
barakeinav1: Good call — fixed in a30f0f66. The absorb-only fields (`features`, `key_provider`, `swap_size`, `port_policy`, `docker_config`) now deserialize as `serde::de::IgnoredAny`, so untrusted JSON in keys we never read is consumed without allocation; dropped the now-unused `Serialize` derive on `AppCompose`.

crates/near-mpc-contract-interface/src/types/attestation.rs
```diff
@@ -253,6 +253,10 @@ impl fmt::Debug for DstackAttestation {
...
+// is pending a team decision.
 #[cfg_attr(
     all(feature = "abi", not(target_arch = "wasm32")),
     derive(schemars::JsonSchema)
```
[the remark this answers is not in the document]
barakeinav1: Fixed in a30f0f66 — added `bash_script` to the public DTO for parity with the verified struct. The full dedupe (single source of truth) is still tracked in #3425.

barakeinav1:
@claude addressed the non-blocking nits in a30f0f66:
- Absorb-only fields now use `IgnoredAny` instead of `Option<Value>` (no allocation of untrusted app-compose JSON), and added the "key-provider integrity is pinned by the `key-provider` event digest, not this field" note.
- `app_compose__rejects_unknown_field` now asserts the error contains `"unknown field"`, so it fails loudly if `deny_unknown_fields` is ever dropped.
- Added `bash_script` to the DTO mirror (full dedupe tracked in #3425).

crates/near-mpc-contract-interface/src/types/attestation.rs
gilcu3: could you make it a proper TODO?

crates/attestation/src/app_compose.rs
```diff
@@ -1,34 +1,72 @@
...
+    #[serde(default)]
+    pub features: Option<IgnoredAny>,
+    #[serde(default)]
+    pub public_tcbinfo: Option<bool>,
+    // Key-provider integrity is pinned by the measured `key-provider` event digest
+    // (`verify_key_provider_digest`), not this field.
+    #[serde(default)]
+    pub key_provider: Option<IgnoredAny>,
+    #[serde(default)]
+    pub storage_fs: Option<String>,
```
gilcu3: question: cant we use `IgnoredAny` for this one?

crates/attestation/src/app_compose.rs
```diff
@@ -1,34 +1,72 @@
...
+    #[serde(default)]
     pub tproxy_enabled: Option<bool>,
+    #[serde(default)]
     pub gateway_enabled: Option<bool>,
+    #[serde(default)]
     pub public_logs: bool,
+    #[serde(default)]
     pub public_sysinfo: bool,
+    #[serde(default)]
     pub local_key_provider_enabled: bool,
```
gilcu3: if this is not optional, and we expect the value is `true`, how does `serde(default)` fit here? Shouldn't we fail when required parameters are missing?

gilcu3 (dismissed):
Thank you!

Left a few questions

EDIT: afaik `serde(default)` is redundant for `Option<T>`

crates/near-mpc-contract-interface/src/types/attestation.rs
barakeinav1: Done in 1ad4934f — changed to `TODO(#3425): dedupe with attestation::AppCompose`.

crates/attestation/src/app_compose.rs
barakeinav1: Yes — fixed in 1ad4934f, `storage_fs` is now `Option<IgnoredAny>` like the other absorb-only structured fields.

crates/attestation/src/app_compose.rs
barakeinav1: Good question. There was no security gap: a missing flag defaulted to `false`, which `validate_app_compose_config` rejects (we require `true`) — so it failed validation rather than silently passing. But you're right it's clearer to fail at parse, so in 1ad4934f I dropped `#[serde(default)]` from the must-be-true flags (`public_logs`, `public_sysinfo`, `local_key_provider_enabled`, `no_instance_id`) → a missing one now fails to parse. `kms_enabled`/`allowed_envs` keep their default since the default *is* the safe value (disabled/empty). Also addressed your EDIT: `#[serde(default)]` is redundant on `Option<_>` fields, so I removed it from all of them.

gilcu3 (approved):

pbeza (approved):

PR #3423, opened 2026-06-01
docs: add SAFETY notice in bls scalar zeroize block

### LINKED ISSUE #3364: Add SAFETY comment to unsafe block in `ScalarWrapper::zeroize`
### Background

The `unsafe` block in `ScalarWrapper::zeroize` (`crates/threshold-signatures/src/confidential_key_derivation/scalar_wrapper.rs:28`) has no `// SAFETY:` comment explaining why the `ptr::write_volatile(&mut self.0, blstrs::Scalar::default())` call is sound.

Flagged as a broken-window fix in https://github.com/near/mpc/pull/3321#discussion_r3305340186.

### Acceptance Criteria

- Add a `// SAFETY:` comment above the `unsafe` block in `ScalarWrapper::zeroize` documenting the invariants that make the `ptr::write_volatile` call sound (valid, aligned, exclusive `&mut` pointer to an initialized `blstrs::Scalar`).

### Context

This would not be needed if #2242 was doable

gilcu3:
Closes #3364

### DIFF
--- a/crates/threshold-signatures/src/confidential_key_derivation/scalar_wrapper.rs
+++ b/crates/threshold-signatures/src/confidential_key_derivation/scalar_wrapper.rs
@@ -25,6 +25,10 @@ impl Zeroize for ScalarWrapper {
     // behavior in a routine version bump.
     #[allow(clippy::volatile_composites, clippy::borrow_as_ptr)]
     fn zeroize(&mut self) {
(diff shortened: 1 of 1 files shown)


netrome (approved):
Thanks! 🙏 I haven't verified the invariant actually holds, but this is still a strict improvement as it makes the assumption clear.

SimonRastikian (approved):

PR #3427, opened 2026-06-01
ci: use upstream version of cache-nix-action

### LINKED ISSUE #3401: nix cache stored in github
### Background

We recently realized that the nix cache was being stored in github instead of WB.

### User Story

_No response_

### Acceptance Criteria

Figure out if we can store it in WB, else roll back to using the upstream nix cache action and make sure we do not surpass the github limits for cache.

_No response_

gilcu3:
Closes #3401


SimonRastikian (approved):

DSharifi (approved):
Thank you!

PR #3435, opened 2026-06-01
feat(indexer): add `/debug/recent_transactions` page

### LINKED ISSUE #2890: Debug page for submitted transactions

pbeza:
Closes #2890

An MPC node constantly submits transactions to the NEAR blockchain — mostly `respond` (signature results) and `vote_pk` (key-generation votes). Sometimes those submissions fail or never land on-chain. The suspected culprit is out-of-order nonce rejections: every transaction carries a nonce (a counter), and the chain rejects one whose nonce is stale or out of sequence.

The problem was that when this happened, an operator had almost no visibility — failures were only aggregated into a Prometheus counter, so you'd see _"N transactions failed"_ but not which ones, with what nonce, signed by which key. You couldn't correlate a failure with the specific access key and nonce involved.

This PR adds a debug web page, `/debug/recent_transactions`, that lists every transaction the node recently submitted, each with its final outcome and the details you'd need to diagnose a nonce problem.

# Implementation

The flow, end to end:

1. **The indexer submits a transaction and observes its outcome.** Its transaction processor builds, signs, sends the tx, waits, then checks whether it took effect — producing one `SubmittedTransaction` record (all the fields above + the final status).

2. **It hands that record to a `LogTransaction`.** This is the key abstraction. Instead of the indexer knowing about web pages, channels, or buffers, it just holds something that implements an infallible trait:

   ```rust
   pub trait LogTransaction {
       fn log_transaction(&self, transaction: SubmittedTransaction);
   }
   ```

   "Fire-and-forget, can't fail, can't block." The indexer calls `log_transaction(record)` and moves on. It has no idea where the record goes. (This decoupling matters because the indexer is being reworked for a "chain-gateway" integration, and you don't want the debug page entangled in that.)

3. **The concrete implementation forwards over a channel.** The web module provides `RecentTransactionsLogger`, which wraps a bounded `mpsc` channel. Its `log_transaction` does a non-blocking `try_send`. If the channel is full or the receiver is gone, it logs a `warn` and drops the record — because a debug page must never slow down or break actual transaction processing.

4. **A drain task moves records into the buffer.** A background task reads from the channel and appends each record into `SharedRecentTransactions` — an in-memory ring buffer (`Arc<Mutex<VecDeque<…>>>`) that keeps the newest 2000 entries and evicts the oldest.

5. **The web handler reads the buffer.** When you GET `/debug/recent_transactions`, the handler takes a quick snapshot of the buffer (clones it under the lock, releases immediately) and renders the lines.

Visually: `indexer → log_transaction() → [mpsc channel] → drain task → buffer ← web page reads`


crates/node/src/indexer/tx_sender.rs
```diff
@@ -153,15 +170,22 @@ async fn submit_tx(
     );
 
     let tx_hash = transaction.get_hash();
+    let nonce = transaction.transaction.nonce().nonce();
     tracing::info!(
```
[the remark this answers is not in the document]
pbeza: > (...) is invalid and will not compile (...)

It does compile.

crates/node/src/indexer/recent_transactions.rs
```diff
@@ -0,0 +1,268 @@
...
+    pub nonce: Option<Nonce>,
+    /// The account the transaction was submitted from.
+    pub signer_account_id: AccountId,
+    /// The access key (public key) the transaction was signed with.
+    pub signer_public_key: Ed25519PublicKey,
+    /// The contract method invoked (e.g. `respond`, `respond_ckd`).
+    pub method: &'static str,
+    /// The height of the reference block the transaction was built against.
+    pub block_height: Option<BlockHeight>,
+    /// Wall-clock time at which the transaction was recorded as submitted.
```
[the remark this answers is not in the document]
pbeza: Good call — added the transaction signature to each record in 9a7f671fb. It's captured in `submit_tx` (cloned before the signed tx is moved into the RPC handler), carried through `SubmittedTxMetadata`, and rendered via `Display` (`ed25519:<base58>`) on the page; `None` on the submit-failed path.

crates/node/src/web.rs
```diff
@@ -189,6 +194,11 @@ async fn contract_state(state: State<WebServerState>) -> String {
     near_mpc_contract_interface::types::protocol_state_to_string(&protocol_state)
 }
 
+async fn debug_recent_transactions(State(state): State<WebServerState>) -> String {
+    let recent_transactions = state.recent_transactions.lock().unwrap();
+    format!("{:?}", &*recent_transactions)
+}
```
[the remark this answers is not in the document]
pbeza: Done in 85416c640 — added `RecentTransactions::snapshot()` which clones the entries under the lock, and the handler now formats via a `render()` helper after releasing the guard, so rendering no longer blocks `record_submitted`/`update_status`.

crates/node/src/indexer/tx_sender.rs
```diff
@@ -153,15 +161,25 @@ async fn submit_tx(
     );
 
     let tx_hash = transaction.get_hash();
+    let nonce = transaction.transaction.nonce().nonce();
+    // Clone the signature before the transaction is moved into `submit_tx`.
+    let signature = transaction.signature.clone();
```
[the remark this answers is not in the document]
pbeza: Fixed in 85416c640 — reworded to "before `transaction` is moved into `rpc_handler.submit_tx`".

crates/node/src/web.rs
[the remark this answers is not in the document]
pbeza: Leaving the `.unwrap()` as-is: a poisoned mutex only occurs if a thread panics while holding it, and the critical sections here (push to a deque, iterate-and-clone) are trivial and panic-free. This matches the other debug handlers, which also do not special-case lock poisoning. The snapshot change keeps the lock hold extremely short regardless.

crates/node/src/indexer/tx_sender.rs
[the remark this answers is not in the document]
pbeza: Fixed in https://github.com/near/mpc/commit/85416c64007ee1037831cafb842ff00f1ff6f9bb

crates/node/src/web.rs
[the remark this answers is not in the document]
pbeza: Addressed in the identical thread above: https://github.com/near/mpc/pull/3435#discussion_r3335068180

crates/node/src/indexer/recent_transactions.rs
netrome, line 1: Haven't gone deep yet, but given that this is a data structure for the `recent_transactions` page it feels weird to have it live under the `indexer`. It feels like this belongs closer to the `web` module - potentially as a submodule of it, or somewhere adjacent.

netrome (commented):
Started reviewing, will continue next week.

crates/node/src/indexer/tx_sender.rs
```diff
@@ -73,6 +79,7 @@ impl TransactionProcessorHandle {
                         indexer_state,
                         tx_request,
                         txn_json,
+                        recent_transactions,
                     )
                     .await;
```
kevindeforth, line 84: @pbeza With the integration of chain gateway in mind, I am motivated to keep the diff on the lower level as small as possible. Here, my suggestion would be to have `ensure_send_transaction` return any relevant data for the `RecentTransaction` struct.

Instead of passing `RecentTransaction` in here (wrapped in a mutex), I would suggest to send the return value of `ensure_send_transaction` via channel to the web-endpoint, where it can then proceed as is wants (seconding @netrome's suggestion of moving the `RecentTransactions` to the web endpoint component).

crates/node/src/web.rs
```diff
@@ -258,16 +276,35 @@ pub async fn start_web_server(
...
+    // completed transactions sent by the indexer's transaction processor in a
+    // background drain task, so the page stays available regardless of the
+    // node's running state.
+    let recent_transactions = Arc::new(Mutex::new(RecentTransactions::default()));
+    {
+        let recent_transactions = recent_transactions.clone();
+        tokio::spawn(async move {
+            while let Some(transaction) = recent_tx_receiver.recv().await {
+                recent_transactions.lock().unwrap().record(transaction);
+            }
```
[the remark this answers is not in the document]
pbeza: Moved the spawn to after `TcpListener::bind` succeeds in 60408a882, so nothing is spawned on the bind-failure path.

Kept it as a detached `tokio::spawn` deliberately: the drain is a process-lifetime singleton, like the indexer's other background monitors (`indexer/real.rs`, `run.rs`), and the web server future is itself a never-dropped `root_runtime.spawn` in production. `AutoAbortTask` is used here only for connection/job-scoped tasks tied to a droppable struct (`p2p.rs`, `network.rs`), not for these singletons. The task also can't use `tracking::spawn` — `start_web_server` runs outside a tracking scope in prod (`root_runtime.block_on`), so `current_task()` would panic. It exits cleanly once the sender side is dropped and the channel closes.

crates/node/src/indexer/tx_sender.rs
pbeza: Makes total sense. Addressed it here: https://github.com/near/mpc/pull/3435/commits/26df9da9d4a5ed9c9c126bb03dfcc89426f2a2cc. Please re-review when you get a chance. 🙏🏼

crates/node/src/indexer/recent_transactions.rs
pbeza: Good point. Fixed in https://github.com/near/mpc/pull/3435/commits/26df9da9d4a5ed9c9c126bb03dfcc89426f2a2cc. ✅

crates/node/src/web.rs
```diff
@@ -65,6 +68,13 @@ struct WebServerState {
     migration_state_receiver: watch::Receiver<(u64, ContractMigrationInfo)>,
     static_web_data: StaticWebData,
     node_config: NodeConfigResponse,
+    /// Buffer of recently submitted transactions, owned by the web server and
+    /// fed by the indexer's transaction processor over a channel (drained into
+    /// it by a background task spawned in [`start_web_server`]). Read directly
+    /// here rather than via the debug-request broadcast; see
+    /// [`recent_transactions`] for why this keeps the page available regardless
+    /// of the node's running state.
+    recent_transactions: Arc<Mutex<RecentTransactions>>,
```
netrome: Nit: I'd consider encapsulating this `Arc<Mutex<...>>` into a wrapper type. But not important here.

crates/node/src/web.rs
```diff
@@ -189,6 +199,13 @@ async fn contract_state(state: State<WebServerState>) -> String {
     near_mpc_contract_interface::types::protocol_state_to_string(&protocol_state)
 }
 
+async fn debug_recent_transactions(State(state): State<WebServerState>) -> String {
+    // Clone the entries under the lock, then format after releasing it, so we
+    // don't block the drain task's writes while rendering.
+    let snapshot = state.recent_transactions.lock().unwrap().snapshot();
```
netrome: Nit: I typically always prefer expect over `unwrap()`. In this case, the error is obvious in either case, but I just find the code easier to skim as "unwrap" and "expect" calls always catch my eye, and a quick reason like "poisoned mutex" makes it easier to skim past this.

```suggestion
    let snapshot = state.recent_transactions.lock().expect("poisoned mutex").snapshot();
```

netrome (dismissed):
Nice stuff. Skimmed the PR, looks good!

crates/node/src/indexer/tx_sender.rs
```diff
@@ -62,20 +67,43 @@ impl TransactionProcessorHandle {
...
+                            tracing::warn!(
+                                target: "mpc",
+                                "recent-transactions channel full; dropping debug record"
+                            );
+                        }
+                        Err(mpsc::error::TrySendError::Closed(_)) => {
+                            tracing::warn!(
+                                target: "mpc",
+                                "recent-transactions drain task gone; dropping debug record"
+                            );
```
kevindeforth: Not a blocker, more a discussion starter around appropriate log levels.
I think logging a `Warn`message works from the perspective of this MPC component: a failure to send the transaction through the channel is not something that prevents this component from working correctly.

But from the perspective of the whole system, I do wonder if `error` would be more appropriate.

- A full buffer is not something we expect to occur. It means the receiver is not draining the channel properly, which is a problem and indicates starvation on the runtime. It also means the debug page won't display the "actual" latest transactions, but older transactions and might not be accurate. I don't think there is a way to catch this on the receiver side, which is why I am inclined to think an "error" is more appropriate here.
- A dropped receiver is not something we would expect during normal operations. If that occurred, it would indicate a serious issue on the debug side, which is hard to know if it can be catched somewhere else. Exception is when the node shuts down and we drop everything (in which case, an "info" would be more appropriate). But I would think we currently have loads of error and warning messages firing when dropping the node, so I don't think this is something we need to optimize for.

So,to summarize: From the perspective of this component: "warn" makes sense, from the perspective of the system, "error" seems more appropriate, but only if we're not catching those errors in some other places. Wdyt?

crates/node/src/tests.rs
```diff
@@ -110,6 +111,10 @@ impl OneNodeTestConfig {
                 let (_, dummy_protocol_state_receiver) =
                     watch::channel(ProtocolContractState::NotInitialized);
                 let (_, dummy_migration_state_receiver) = watch::channel((0, BTreeMap::new()));
+                // The fake indexer never records, so the sender is dropped; the
+                // drain task just sees the channel close.
```
kevindeforth: **Out of Scope of this PR:**
Ok, so this is super awkward due to our current integration test architecture, but I would suggest to still hook this up with the fake transaction sender and add some sort of integration test.

I didn't spend too much time looking, but one possible place where I could see this getting integrated in the test node is in `fake.rs`:

https://github.com/near/mpc/blob/1116bb57fdb8b074b961f04802d8f8e0ce71b9b2/crates/node/src/indexer/fake.rs#L922

I would suggest to do this as a follow-up, as whoever does it may want to re-visit some of our testing architecture and this might become tricky.
For now, we can test the endpoint via e2e tests.

crates/node/src/web.rs
```diff
@@ -280,12 +310,28 @@ pub async fn start_web_server(
...
+            recent_transactions,
         });
 
     let tcp_listener = TcpListener::bind(&bind_address).await?;
 
     tracing::info!(address = %bind_address,"Successfully bound to address");
 
+    // Spawn the drain only after a successful bind so it can't leak on the error
+    // path above. Runs for the whole process; warns on exit.
+    tokio::spawn(async move {
```
kevindeforth: Ah, I was wondering why we spawn this only after the webserver :D 
Have you considered adding a cancellation token?
Alternatively, it might be even cleaner to spawn this task in `src/run.rs` and pass the `Arc<Mutex<RecentTransactions>>` as an argument to `spawn_web_server`. Wdyt?

crates/node/src/indexer/tx_sender.rs
```diff
@@ -3,12 +3,16 @@ use super::IndexerState;
 use super::tx_signer::{TransactionSigner, TransactionSigners};
 use crate::config::RespondConfig;
 use crate::metrics;
+use crate::web::recent_transactions::{
+    SignerContext, SubmittedTransaction, SubmittedTransactionStatus, SubmittedTxMetadata,
```
kevindeforth: Since these are shared between `tx_sender` and `web_endpoint` (which are two separate components of the node), I would move those types to `node/src/types.rs`.

Having `tx_sender.rs` depend on `web` seems odd.

crates/node/src/web/recent_transactions.rs
kevindeforth, line 1: note: As above, I suggest to move the types shared between web and indexer to `node/types.rs`, just so we don't have the indexer depend on the web module.

kevindeforth (commented):
Thank you very much, this looks really good!

I have some minor comments regarding integration tests and code org that I would like us to align on before merging, but nothing major.

crates/node/src/indexer/tx_sender.rs
netrome: Interesting. I can definitely see the arguments for both:

On one hand, this is just a debug page so I wouldn't want to wake someone up in the middle of the night[^1] to debug if just the debug page has problems.

On the other hand, on the system level something should be seriously wrong if this sender fails which would warrant a deeper investigation - hence error would be appropriate.

[^1]: We don't have any such alerts yet, but I think it would be nice to get to a point where we could alert on error logs.

Furthermore, if we make this an error today and we change assumptions in other parts of the code of how this channel is dropped we might end up with spammy error logs. Which would be easy to spot and fix, but still.

If we zoom out, I think this might be an indicator of code smell because the indexer should not be concerned with internals of the debug page backend - which it has to be in this piece of the code. One alternative here is to provide an infallible interface (e.g. trait) for how to record these transactions, and make the implementation of that trait handle this logic. This implementation could be entirely self-contained in the `web.rs` module, whereas the trait could live in either module (I'd lean towards have the trait defined in the indexer module, similar to how ports are defined in hexagonal architecture).

Anyway, this might be overkill and in practice I think both "warn" and "error" here would work fine since I would be very surprised to see us hit these errors in production and we'd notice both warning and error messages. (Funny thing, I think warnings stand out more than errors currently in production since we have some spammy error logs still)

kevindeforth: > One alternative here is to provide an infallible interface (e.g. trait) for how to record these transactions, and make the implementation of that trait handle this logic. This implementation could be entirely self-contained in the web.rs module, whereas the trait could live in either module


Ah, so something like:
```rust
trait TransactionLogger {
    fn log_transaction(transaction: RecentTansaction);
}
```

and we do:
```
 pub(crate) fn start_transaction_processor(
        owner_account_id: AccountId,
        owner_secret_key: SigningKey,
        config: RespondConfig,
        tx_logger: impl TransactionLogger,
)
```
instead of 
```rust
 pub(crate) fn start_transaction_processor(
        owner_account_id: AccountId,
        owner_secret_key: SigningKey,
        config: RespondConfig,
        recent_tx_sender: mpsc::Sender<SubmittedTransaction>,
)
```

This refactor could then also take care of https://github.com/near/mpc/pull/3435#discussion_r3386640825

This doesn't seem too big of a change to me. @pbeza are you up for it? It's not a blocker to merge the PR, so feel free to reject. But I understand this PR is also not super urgent, so I think it's fine if we refine it a bit more.

crates/node/src/web.rs
pbeza: Addressed in https://github.com/near/mpc/pull/3435/commits/d93d753752e46b9b5663065571d3d00a7cef470a + https://github.com/near/mpc/pull/3435/commits/33a3cec1f98dc61407c5e6d4d6e5d7531d3ad270.

crates/node/src/web.rs
pbeza: Addressed in https://github.com/near/mpc/pull/3435/commits/33a3cec1f98dc61407c5e6d4d6e5d7531d3ad270. ✅

crates/node/src/tests.rs
pbeza: > I would suggest to do this as a follow-up, as whoever does it may want to re-visit some of our testing architecture and this might become tricky.

Created an issue for that: https://github.com/near/mpc/issues/3522. ✅

> For now, we can test the endpoint via e2e tests.

Added an e2e test here: https://github.com/near/mpc/pull/3435/commits/90d8a811dbc4ba7bdb1d0c30f3c3e26a34955cf5. ✅

crates/node/src/indexer/tx_sender.rs
pbeza: Nice ideas! I went with the trait approach in the end. Liked that one the most: https://github.com/near/mpc/pull/3435/commits/4c5ab1717cd6c2cd0da58e622a2414f3e3e2e2e2. ✅

crates/node/src/web.rs
pbeza: > Alternatively, it might be even cleaner to spawn this task in src/run.rs and pass the Arc<Mutex<RecentTransactions>> as an argument to spawn_web_server. Wdyt?

I like the idea! Addressed it in https://github.com/near/mpc/pull/3435/commits/4c5ab1717cd6c2cd0da58e622a2414f3e3e2e2e2:

https://github.com/near/mpc/blob/4c5ab1717cd6c2cd0da58e622a2414f3e3e2e2e2/crates/node/src/run.rs#L186-L200

(see also https://github.com/near/mpc/pull/3435#discussion_r3382771860 on why not passing the raw `Arc<Mutex<RecentTransactions>>`)

crates/node/src/indexer/tx_sender.rs
pbeza: Addressed in https://github.com/near/mpc/pull/3435/commits/4c5ab1717cd6c2cd0da58e622a2414f3e3e2e2e2. ✅

crates/node/src/web/recent_transactions.rs
pbeza: Addressed in https://github.com/near/mpc/pull/3435/commits/4c5ab1717cd6c2cd0da58e622a2414f3e3e2e2e2. ✅

crates/node/src/indexer/tx_sender.rs
netrome: Nice! Though as usual I'd suggest naming it `LogTransaction` rather than `TransactionLogger` (not blocking)

crates/node/src/indexer/real.rs
```diff
@@ -12,6 +12,7 @@ use crate::indexer::tee::{
     monitor_allowed_launcher_compose_hashes, monitor_tee_accounts,
 };
 use crate::indexer::tx_sender::{TransactionProcessorHandle, TransactionSender};
+use crate::types::TransactionLogger;
```
netrome: Nit1: Feels a bit weird to declare a trait in a `types` module. We could consider moving it to a new `ports` module following more the ports and adapters (or hexagonal) approach.

Nit2: I feel the code reads better when traits are verbs and not nouns.

```suggestion
use crate::types::LogTransaction;
```

crates/node/src/web/recent_transactions.rs
```diff
@@ -0,0 +1,290 @@
...
+            }
+            Err(mpsc::error::TrySendError::Closed(_)) => {
+                tracing::warn!(
+                    target: "mpc",
+                    "recent-transactions drain task gone; dropping debug record"
+                );
+            }
+        }
+    }
+}
```
netrome: Yeah looks much better to have these errors here in this module. Feels like it requires no further system context to reason about it.

netrome (dismissed):
Quick review, looks good. Thanks for the updates 🙏

pbeza:
Ready for re-review. 🫡

crates/e2e-tests/tests/web_endpoints.rs
```diff
@@ -38,6 +38,61 @@ async fn ensure_body_contains(
...
+            "node {node}: {url} body did not match within {timeout:?}"
+        );
+        tokio::time::sleep(common::POLL_INTERVAL).await;
+    }
+}
+
+/// Regex matching one rendered recent-transactions row for a node's keygen
+/// `vote_pk` submission (which every node makes, so we can assert per-node).
+/// Stable tokens (status, method, signer, key) are matched exactly; run-specific
+/// ones (timestamp, txid, nonce, block, sig) are matched by pattern only.
```
kevindeforth: nit :
```suggestion
/// Regex matching one rendered recent-transactions row for a node's keygen
/// `vote_pk` submission.
/// Stable tokens (status, method, signer, key) are matched exactly; run-specific
/// ones (timestamp, txid, nonce, block, sig) are matched by pattern only.
```

kevindeforth (dismissed):
Thanks a lot!

crates/node/src/indexer/real.rs
pbeza: > Nit1: Feels a bit weird to declare a trait in a types module. We could consider moving it to a new ports module following more the ports and adapters (or hexagonal) approach.

Created a follow-up for this one (https://github.com/near/mpc/issues/3546), as addressing the suggestion consistently means moving the existing boundary traits too, instead of adding a one-off `ports` module with a single trait.

> Nit2: I feel the code reads better when traits are verbs and not nouns.

https://github.com/near/mpc/pull/3435/commits/131b7fcc58c65a7b7e9a17b21ab3cebe2e381fe5 ✅

netrome (approved):
Thanks for the name change 🙏

kevindeforth (approved):

PR #3436, opened 2026-06-01
docs: propose different way of calculating supported foreign chain set

### LINKED ISSUE #3474: Propose a design for preventing a single node disable foreign chain validation feature
### Background

### User Story

_No response_

### Acceptance Criteria

We have a design approved

_No response_

anodar:
Closes: https://github.com/near/mpc/issues/3474


docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,133 @@
...
+[`docs/foreign-chain-transactions.md`](../foreign-chain-transactions.md#on-chain-rpc-provider-whitelist)),
+and no per-node input can subtract a chain from it.
+
+**Every active node is required to support every supported chain.** A node
+*supports* `C` iff it has at least `quorum(C)` whitelisted RPC providers
+configured for `C` (the minimum it needs to ever reach its RPC response quorum at
+verification time). Requiring all nodes to support all chains is the option that
+deviates least from today's liveness and security guarantees — under the old
+intersection rule, every node already had to register a chain for it to count —
+while removing the single-operator DoS.
```
kevindeforth: > deviates least from today's liveness and security guarantees

This won't age well in docs. It would be more helpful to state the liveness and security guarantees that we expect today, and say why this option adheres to those.

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,133 @@
...
+
+A node that does **not** support a supported chain is treated **identically to a
+node that is down** for that chain: it abstains from `C`'s verification requests,
+and the request proceeds with the remaining signers (subject to the usual signing
+threshold). The accepted cost is wasted assets — and the waste is **not confined
```
kevindeforth: Only the triples are shared between domains. The presignatures are domain specific.

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,133 @@
...
+to chain `C`**. Pre-generated triples and presignatures are usable only while
+every participant associated with them is alive, so the ones co-owned with a node
+treated as offline get stranded and discarded as offline assets, exactly as a
+genuinely down node strands the assets generated with it. These are **shared
+signing assets**, not foreign-chain-specific: triples are domain-agnostic, and
+foreign-tx signing reuses the same ECDSA presignature pool as ordinary `sign()`
+(see `crates/node/src/providers/verify_foreign_tx.rs`). So the cost falls on the
```
kevindeforth: Oh really? I thought we had a separate domain for foreign chain verification requests.

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,133 @@
...
+| `whitelist(C)` | provider ids the network trusts for `C` | `AllowedProviders.entries[C].providers` |
+| `quorum(C)` | per-chain RPC response quorum | `AllowedProviders.entries[C].quorum` |
+
+## Verification behavior
+
+When a foreign-tx verification request is processed, each node fans out the query
+to its locally-configured whitelisted providers for `C`. If fewer than `quorum(C)`
+of them return the same response, the node **errors out the foreign-tx validation
+and does not retry** that request — it abstains for that request rather than
+re-querying or falling back. This keeps the trust property crisp: a result is only
```
kevindeforth: Is this AI?
> the node [...] does not retry** that request - it abstains for that request rather than
re-querying or falling back.

- What would it fall back to?
- What is the difference between retrying and re-qeurying?

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,133 @@
...
+accepted when an RPC response quorum genuinely agreed, and a node that can't reach
```
kevindeforth: > when an RPC response quorum genuinely agreed

as opposed to the quorum agreeing _disingenuously_?

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,133 @@
...
+that bar contributes nothing rather than a weaker observation.
```
kevindeforth: With _"bar"_, do we mean _threshold_? What does a _"weaker observation"_ mean?

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,133 @@
...
+Status: Proposed — supersedes the all-participant intersection rule described in
+[`docs/foreign-chain-transactions.md`](../foreign-chain-transactions.md). Tracked by
+[#3434](https://github.com/near/mpc/issues/3434).
+
+## Background
+
+For foreign-chain transaction verification, the contract exposes
+`get_supported_foreign_chains()` — the set of chains the network advertises as
+verifiable. A `verify_foreign_transaction` request is rejected on-chain unless
+its target chain is in this set.
```
kevindeforth: I don't think this is a sufficient condition. We still need at least `cryptographic_threshold` number of nodes to support a foreign chain, for a signing quorum to be possible.

kevindeforth (commented):
Thanks a lot for the doc and the discussions!

The modifications to the existing docs look good, the new doc `calculating-supported-foreign-chains.md` can, IMO, be more concise. It contains a bit of repetition and the main idea is a bit hard to identify.

The actual proposal is slightly insufficient right now. I would suggest to consider the following alternative:

_A foreign chain considered supported, if we have `threshold` distinct and active nodes that support that chain, where `threshold` corresponds to the cryptographic reconstruction threshold of the domain associated to foreign chain verification._

This means we will still inspect the `foreign_chain_support_by_node`, but instead of requiring a full match between the participant set and the nodes supporting that chain, we require that the intersection contains at least `threshold` nodes.

docs/design/calculating-supported-foreign-chains.md
anodar: Same as below, my understanding was wrong on this.

docs/design/calculating-supported-foreign-chains.md
anodar: We do have separate domains but I thought presignatures were shared, I was wrong: 

[crates/node/src/providers/verify_foreign_tx.rs](https://github.com/near/mpc/blob/main/crates/node/src/providers/verify_foreign_tx/sign.rs#L61)

docs/design/calculating-supported-foreign-chains.md
anodar: This is AI wording but vetted by me. As we agreed offline, if only less than quorum number of RPC providers return same response, we can't trust that response, and we should fail this foreign tx verification request.

- not re-querying: as part of processing this verification request it won't query RPC providers again to see if quorum number of providers match
- fall back: poorly worded, but fall back would be to re-try the request itself.

So difference is retry at node level (re-querying rpc), and retrying at protocol level (fall back). Will reword to make it clear.

docs/design/calculating-supported-foreign-chains.md
anodar: Bar = quorum not threshold, I'm trying to be consistent with these terms to differentiate per node quorum (least amount of number of RPC providers that need to return same response for it to be valid) vs protocol threshold (cryptographic reconstruction threshold).

Weaker observation =  less than quorum observed same value.

Agree though this is poorly worded. Rewording.

docs/design/calculating-supported-foreign-chains.md
anodar: Proposal is to have all chains supported that are voted in during RPC providers allowlisting. If some node reports config that misses some of these chains this is not normal and our alerts will start firing.

So by default everyone should be supporting all the chains, if not something is wrong and should be fixed.

docs/design/calculating-supported-foreign-chains.md
anodar: Reworded.

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
```diff
@@ -20,8 +20,8 @@ requiring all other nodes to have the exact same configuration.
...
+5. A foreign chain is considered supported by the MPC network iff it has an entry in the on-chain RPC whitelist (a voted-in `ChainEntry`). Every active node is required to support every supported chain (configure at least a per-chain quorum number of whitelisted RPC providers for it); a node that does not is treated like a node that is down for that chain — it does not participate, and the pre-generated triples/presignatures it co-owns become offline assets and are discarded unused (foreign-tx presignatures sit in their own `ForeignTx`-domain pool, but triples are shared per reconstruction threshold across all CaitSith domains, so the waste also hits ordinary `sign()`, not just that chain) — and is surfaced by alerting. The quorum threshold is per-chain. See [Calculating the supported foreign-chain set](calculating-supported-foreign-chains.md).
```
gilcu3: Yesterday we discussed this and somehow agreed that there is some waste. Today I looked at the implementation again, and I think the status quo for offline nodes is that as long as the leader knows the node is down, there is 0 waste in the queue (unless it remains down for a long time). So here I think if we implement it properly, we could say that instead of:

> the pre-generated triples/presignatures it co-owns become offline assets and are discarded unused

the pre-generated triples/presignatures it co-owns become offline assets. If an assert remains offline for a long period of time then it is eventually discarded, leading to a waste of resources.

docs/design/calculating-supported-foreign-chains.md
kevindeforth: Hmm, I might have a mis-understanding how the output of `get_supported_foreign_chains` is used. Here is my mental model:

I agree that by default, everyone should support all these chains, but if we arrive at a state where we don't have threshold number of nodes supporting a chain, we are effectively not supporting that chain.
Any requests sent by users will timeout, as we won't be able to serve them.

So, if the contract has a current view of the foreign chain configs of all nodes, then it seems a diligent to reject requests for non-supported chains, even if the non-support is only temporarily.

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,118 @@
...
+result only when ≥ `quorum(C)` return the same response. If fewer agree, the node
+errors out and produces no signature share.
+
+**This sub-quorum outcome must be terminal — the leader must not re-attempt the
+request.** Implementation requirement, not current behavior: the generic queue
+retries every request up to `MAX_ATTEMPTS_PER_REQUEST_AS_LEADER`
+(`requests/queue.rs:38`), so the foreign-tx path must special-case a sub-quorum
+result as non-retryable. (Open: whether a sub-quorum from purely *transient*
+failures — timeouts, finality not reached — should still retry, vs. only genuine
+disagreement being terminal.)
```
kevindeforth: Would be good to reference the issue here.

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
```diff
@@ -107,13 +107,13 @@ Since the nodes are running in a Trusted Execution Environment (TEE), this funct
...
 
 The quorum value comes from on-chain `ChainEntry.threshold`, voted in as part of the same `ChainVote` that voted the chain's provider list — so participants agree on both "which providers are trusted" and "how many of them must concur" in one round.
 
 ### Nodes submit the configured foreign chains on-chain
 
-Nodes submit their per-chain provider set on-chain so the network knows which chains they support. Functionality for this was added on the contract side in [#2784](https://github.com/near/mpc/pull/2784) and is kept — the whitelist is a new layer on top, not a replacement.
+Nodes submit their per-chain provider set on-chain so the network knows which chains they support. Functionality for this was added on the contract side in [#2784](https://github.com/near/mpc/pull/2784) and is kept — but its role is now **monitoring/alerting only**: the supported set is derived from the on-chain whitelist, while these registrations let us detect (and alert on) any active node that does not support a supported chain. See [Calculating the supported foreign-chain set](calculating-supported-foreign-chains.md).
```
gilcu3: nit: "support a supported chain" reads a bit bad, not sure how to improve

docs/foreign-chain-transactions.md
```diff
@@ -555,8 +555,8 @@ See "Contract State (Foreign Chain Configurations)" above.
 * Node config contains chain RPC providers and timeouts (API keys stay local).
 * On startup, each node submits a single `register_foreign_chain_config` transaction derived from its local configuration. The call is idempotent.
 * Nodes do **not** vote, poll, or wait for network-wide consensus — the transaction is sent and startup continues.
-* A chain appears in `get_supported_foreign_chains()` only once **every** active participant has registered it.
-* Per-participant registrations can be inspected with `get_foreign_chain_support_by_node()`.
+* The supported set is the on-chain RPC whitelist, not these registrations: a chain appears in `get_supported_foreign_chains()` once the network votes in a `ChainEntry` for it. Every active node is expected to support every supported chain; a node that does not is treated like a node that is down for that chain (it does not participate) and is surfaced by alerting. See [Calculating the supported foreign-chain set](design/calculating-supported-foreign-chains.md).
```
kevindeforth: Here, what order does this happen? Do the operators whitelist a chain, then add RPC endpoints to a nodes config?
Or is it the other way around? I.e., do they first change their local config and then vote for supporting that chain?
How does the contract make sure that a whitelisted chain is actually supported by all nodes?

kevindeforth: Is this referring to the `ForeignChainRpcWhitelist`? 
If so, then I thought this is just the whitelist of RPC providers (I.e. the list of RPC endpoints the nodes can query).
At what point do we consider a foreign chain supported? Having a single RPC provider for a foreign chain seems a bit risky.

kevindeforth (commented):
Thanks a lot, appreciate the care!

Can't approve yet, as I need to understand at what point a whitelisted chain becomes supported by all nodes and how the output of `get_supported_foreign_chains` is used in the contract and outside.

We may have a slightly different mental model around what it means for a chain to be "supported" and "whitelisted".

docs/design/calculating-supported-foreign-chains.md
gilcu3: I think only genuine disagreement should be terminal. Although this does not match well with the quorum. What I mean is that when the node makes the RPC requests, it needs quorum but will make more requests. If some of those give a different response, then we should probably fail (and log which one did, and what was the request). Timeout and other problems might be forgiven (and there retries would be fine)

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Monitoring and alerting
+
+A node silently treated as down erodes a chain's availability margin, so we must
+detect it. Per-node registration (`register_foreign_chain_config` /
+`get_foreign_chain_support_by_node`) is **retained solely for this** — it no longer
+feeds the supported set, but lets monitoring **alert when an active node does not
+support a supported chain**. That alert fires for *us* (maintainers), who then
```
gilcu3:
```suggestion
support a whitelisted chain**. That alert fires for *us* (maintainers), who then
```

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,118 @@
...
+nudge the operator. Ideally operators also run their own coverage alert and fix
+the gap first.
+
+## Guarantees preserved
+
+Two guarantees hold today, and this rule keeps both.
+
+**Security** — the network signs an observation only if ≥ `threshold` participants
```
gilcu3:
```suggestion
**Safety** — the network signs an observation only if ≥ `threshold` participants
```

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,118 @@
...
+
+- **Presignatures are per-domain.** Foreign-tx signing uses a dedicated
+  `ForeignTx` domain (`DomainPurpose::ForeignTx`), so stranded foreign-tx
+  presignatures stay in that pool.
+- **Triples are shared** per reconstruction threshold across all CaitSith
+  domains, so stranding them also dents ordinary `sign()` presignature
+  generation — this cost is **not** confined to `C`.
+
+The mitigation is operational: the alerting above keeps coverage high, and
+operators are expected to configure every node for every chain.
```
gilcu3: this reads slightly worse that it should be, given my first comment above

gilcu3 (dismissed):
LGTM

Left some comments about the assets loss

docs/foreign-chain-transactions.md
anodar: Order would be the following:
1. all nodes add in their local config new RPC providers for existing chains or new chains with new providers, or mixture of two
2. they all vote for allowlisted (chain - rpc provider) list. Before this step providers / chains from config that aren't in allowlist will be ignored


>How does the contract make sure that a whitelisted chain is actually supported by all nodes?

There's an implicit assumption, that all chains that are voted in whitelisted RPC provider step are supported. It doesn't make sense to vote for whitelisting chains itself since we're already voting for [chain - [rpc provider]] list.

anodar: No it's not `ForeignChainRpcWhitelist`, this is `SupportedForeignChains`. 

`ForeignChainRpcWhitelist` is set of `<chain, [rpc, providers]>`. Keys from this map will be the supported foreign chains list. 

> At what point do we consider a foreign chain supported?

Once that chain makes into the whitelisting vote, that includes this chain.

> Having a single RPC provider for a foreign chain seems a bit risky.

Vote has quorum for each chain for minimum number of RPC providers. We can enforce lower bound to whatever seems reasonable number.

docs/design/calculating-supported-foreign-chains.md
anodar: I'm not sure it makes sense to create issues in the design proposal at least until it's agreed and approved.

docs/design/calculating-supported-foreign-chains.md
anodar: I think we might be disagreeing on just terms we use.

>Any requests sent by users will timeout, as we won't be able to serve them.

We won't attempt to validate request for every chain from supported chains. We'll still check if there's enough nodes supporting that chain.

We have two lists:
1.  `supported chains`: explicit list, one that everyone must support
2.  `available chains`: implicit list, calculated dynamically based on config reports from nodes, that we can actually support. 

We'll still early reject (temporarily) unavailable chains.

docs/design/calculating-supported-foreign-chains.md
anodar: I'm trying to scope this proposal down as much as possible to get it approved and start working on it. That's why I left it as open point, but that's exactly what I'd expect, transient errors should be retried not considered as disagreement.

docs/design/calculating-supported-foreign-chains.md
anodar: Changed to cover.

docs/design/calculating-supported-foreign-chains.md
anodar: Cover.

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
anodar: Changed to cover.

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
anodar: Reworded accordingly.

docs/design/calculating-supported-foreign-chains.md
anodar: Reworded.

docs/foreign-chain-transactions.md
kevindeforth: > No it's not ForeignChainRpcWhitelist, this is SupportedForeignChains.

Ah, because the doc talks about a whitelist, that's a bit confuing.

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,157 @@
...
+non-covering participant produces no share and can stall the request.
+Implementation requirement, not current behavior: today the signing set is inherited
+from a presignature (`take_owned()` in `verify_foreign_tx/sign.rs`), whose
+participants were chosen for liveness, not chain coverage.
+
+## Per-node registration
+
+Per-node registration (`register_foreign_chain_config` /
+`get_foreign_chain_support_by_node`) reports which chains each node currently covers,
+and serves two roles:
```
anodar: Ideally we'd also rename these to `register_available_chain_config`/`get_available_chain_by_node` but that means introducing these 2 new methods, upgrading contract, then upgrade node again to use new methods and deprecate old ones, then upgrading contract to drop those old methods.

Not sure it's worth as we're already adding new methods to the contract as part of this proposal and I'd rather keep set of new methods small to ensure contract size doesn't hit limit.

kevindeforth: > Ideally we'd also rename these to register_available_chain_config/get_available_chain_by_node but that means introducing these 2 new methods, upgrading contract, then upgrade node again to use new methods and deprecate old ones, then upgrading contract to drop those old methods.

Yeah, but it's not like we won't need to update the contract or node if we omit the rename.
Considering that we do those updates either way, the overhead of the rename is minimal: we just add one-line methods that call the existing methods and add a todo with an issue

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
```diff
@@ -20,8 +20,8 @@ requiring all other nodes to have the exact same configuration.
...
+4. Every node partaking in a foreign signature verification request queries all its locally configured RPC providers for the relevant chain, independently of other nodes. A quorum of those RPC providers must agree on the verification; if fewer than the quorum agree, the node errors out that foreign-tx validation and does **not** retry the request.
+5. A foreign chain is **whitelisted** iff the on-chain RPC whitelist has a `ChainEntry` for it — `get_whitelisted_foreign_chains()` returns these, and no single node can change it. It is **available** (actually served) only while ≥ signing threshold active nodes cover it (each with at least the per-chain RPC quorum of whitelisted providers configured), as returned by `get_available_foreign_chains()`; `verify_foreign_transaction` early-rejects requests for a whitelisted-but-unavailable chain. Per-node registrations feed both this availability check and the alerting on coverage gaps. A node not covering a chain is treated as down for it — it does not participate, and its co-owned pre-generated assets may eventually be discarded. See [Calculating the whitelisted and available foreign-chain sets](calculating-supported-foreign-chains.md).
```
kevindeforth: This section is mixing definitions with requirements and expects the reader to have quite a bit of context that is not provided until here. It talks about "alerting on coverage gaps", without properly specifying what those are.

It's a bit more verbose, but I would suggest to spell out definitions separate from requirements:

Definitions:
- A foreign chain is **whitelisted** iff the on-chain RPC whitelist has a `ChainEntry` for it.
- A foreign chain is considered **supported by a node**, if the node has a local RPC config able to reach a quorum.
- A foreign chain is **available** iff it is whitelisted and supported by at least `signing_threshold` number of nodes (where `signing_threshold` refers to the cryptographic reconstruction threshold of the foreign chain validation domain).

Requirements:
- We expect all nodes to support all whitelisted foreign chains at all times, but there is no on-chain enforcement of this and **the system must be tolerant to temporary non-support.** As long as a the chain is _available_, the system must serve requests.
-  Monitoring and off-chain alert channels will be used to notify operators if a node does not support all whitelisted foreign chains.

This part here, I would omit and consider an implementation detail.

> A node not covering a chain is treated as down for it — it does not participate, and its co-owned pre-generated assets may eventually be discarded.

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
```diff
@@ -107,18 +107,18 @@ Since the nodes are running in a Trusted Execution Environment (TEE), this funct
...
 
 The quorum value comes from on-chain `ChainEntry.threshold`, voted in as part of the same `ChainVote` that voted the chain's provider list — so participants agree on both "which providers are trusted" and "how many of them must concur" in one round.
 
 ### Nodes submit the configured foreign chains on-chain
 
-Nodes submit their per-chain provider set on-chain so the network knows which chains they support. Functionality for this was added on the contract side in [#2784](https://github.com/near/mpc/pull/2784) and is kept — the whitelist is a new layer on top, not a replacement.
+Nodes submit their per-chain provider set on-chain so the network knows which chains they cover. Functionality for this was added on the contract side in [#2784](https://github.com/near/mpc/pull/2784) and is kept, now serving two roles: it feeds the **available**-set computation (which chains have ≥ signing threshold covering nodes) and the alerting that flags any active node not covering a whitelisted chain. See [Calculating the whitelisted and available foreign-chain sets](calculating-supported-foreign-chains.md).
```
kevindeforth, line 127: nit: I think "support" was better, but both are fine. Just make sure to be consistent and to define what it means for a node to "cover" a foreign chain (c.f. comment above).

docs/foreign-chain-transactions.md
```diff
@@ -287,7 +287,7 @@ general-purpose `sign()` keys, even if the same account and derivation path are
 
 ## Contract State (Foreign Chain Configurations)
 
-The contract stores a foreign-chain configuration **per participant** — there is no global, voted-on policy. The set of chains the network collectively supports is derived as the **intersection** of chains registered by every active participant.
+The **whitelisted** set is derived from the **on-chain RPC whitelist** (`foreign_chain_rpc_whitelist`): a chain is whitelisted iff the network has voted in a `ChainEntry` for it, exposed by `get_whitelisted_foreign_chains()`. The **available** set — the chains ≥ signing threshold active nodes currently cover — is computed from the per-participant registrations and exposed by `get_available_foreign_chains()`; `verify_foreign_transaction` gates on it. The per-participant registration also drives alerting (detecting an active node that does not cover a whitelisted chain). See [Calculating the whitelisted and available foreign-chain sets](design/calculating-supported-foreign-chains.md).
```
kevindeforth, line 325: nit: This is the first time in this doc that we mention "whitelisted".
The way the sentence is structured it makes the reader believe they should know what that means. I would suggest to define the terminology before using it (ideally, we only have one place for these definitions and then use the correct terminology across all three docs).

docs/foreign-chain-transactions.md
```diff
@@ -312,8 +312,9 @@ pub enum ForeignChain {
 Relevant contract methods:
 
 * `register_foreign_chain_config(foreign_chain_configuration: ForeignChainConfiguration)` — call method. The authenticated participant (re)registers its per-chain provider set. The call is idempotent.
-* `get_supported_foreign_chains() -> SupportedForeignChains` — view method. Returns the set of chains that appear in **every** active participant's registered configuration.
-* `get_foreign_chain_support_by_node() -> ForeignChainSupportByNode` — view method. Returns each participant's registered set of supported chains.
+* `get_whitelisted_foreign_chains() -> WhitelistedForeignChains` — view method. Returns the chains present in the on-chain RPC whitelist (`foreign_chain_rpc_whitelist`). (`get_supported_foreign_chains()`, the old intersection rule, is **to be deprecated**.)
```
kevindeforth: > the old intersection rule is ** to be deprecated**.

Pleas see [this comment](https://github.com/near/mpc/pull/3436#discussion_r3341531506). "old" and "new" are terms that are extremely hard to track in versioned documents. 

I would also question the need for mentioning this. Why would the reader need to know about there having been an "old" intersection rule? Where was that rule defined?

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,165 @@
...
+- **Whitelisted** — the explicit, vote-driven policy set every node is expected to
+  cover. `C` is whitelisted iff the on-chain RPC whitelist has a `ChainEntry` for
+  `C`. `get_whitelisted_foreign_chains()` returns exactly these keys. **No per-node
+  input can add or remove a chain**, so no single operator can change it.
+- **Available** — the set the network can actually serve right now, computed
+  dynamically from the per-node config reports. `C` is available iff
+  ≥ `signing_threshold` active participants cover `C`, where `signing_threshold` is
+  the reconstruction threshold of the `ForeignTx` domain and a participant *covers*
+  `C` iff it has ≥ `rpc_quorum(C)` whitelisted providers configured for `C`.
+  `get_available_foreign_chains()` returns this set.
```
kevindeforth: nice! I see we have this, it's just in a different place.

I would suggest to organize these definitions in the main doc and refer the reader to those.

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,165 @@
...
+expected to cover every whitelisted chain.
+
+**Available** avoids spending resources on a request the network can't fulfill: one
+for a chain without a signing threshold's worth of coverage is rejected up front, not
+attempted and left to time out. Because it requires `signing_threshold` — not all —
+participants, a chain leaves it only when more than `n − signing_threshold` nodes
+stop covering it.
+
+In a healthy network `available == whitelisted`. A gap means nodes are down or
+misconfigured for a chain — an operational anomaly that alerting surfaces (see
```
kevindeforth, line 53: note: would be good to open an issue for setting up the alerts.

docs/design/calculating-supported-foreign-chains.md
kevindeforth: I thought we had already agreed on this. c.f. [this summary](https://nearone.slack.com/archives/C0912BTG51T/p1780390812834059?thread_ts=1780321648.375279&cid=C0912BTG51T) from the call two days ago.

We can also use the issue to define better how exactly we want to fail, but I think it's important to open some sort of tracking issue, otherwise we risk forgetting about it.

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,165 @@
...
+
+Per-node registration (`register_foreign_chain_config` /
+`get_foreign_chain_support_by_node`) reports which chains each node currently covers,
+and serves two roles:
+
+- it **feeds the available set** — the contract counts, per chain, how many active
+  participants report coverage and compares against `signing_threshold`; and
+- it **drives alerting** — when an active node does not cover a whitelisted chain,
+  monitoring fires for *us* (maintainers), who nudge the operator. Ideally operators
+  run their own coverage alert and fix the gap first.
```
kevindeforth: We don't have that monitoring in place yet, right?

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,165 @@
...
+
+Registration reflects each node's *current* config.
+
+The existing method names (`register_foreign_chain_config` /
+`get_foreign_chain_support_by_node`) are kept even though the data now feeds the
+*available* set: renaming them (e.g. to `register_available_chain_config` /
+`get_available_chain_by_node`) would add two more methods plus a multi-step
+contract→node→contract migration to retire the old ones — not worth growing the
+new-method set against the contract size limit, given this proposal already adds two
+views.
```
kevindeforth: The contract size impact should be minimal, as we would wrap existing methods.
We update node and contract on a regular basis, so including these in the next updates is a minor inconvenience and the overhead is negligilbe.
I prefer prioritiziing a clean API here.

kevindeforth (dismissed):
Nice work, thanks a lot for the updates!

I left a few comments. They fall into three categories: form, issue tracking and implementation steps. None of them are blocking, as they are all orthogonal to the design put forward in this doc.

- form suggestions: We mix requirements and definitions, sometimes we don't define terminology properly before using it, sometimes we define the same terms in multiple docs. It would be nice to consolidate this. A lot of the terminology "available", "whitelisted", "chain supported / covered by node" is or will be part of the API, so it makes sense to have a single place where these things are defined. Note that this will get considerably harder the more we modify and add to these docs.
- issue tracking: I would like to see tracking issues related to monitoring and computation-retry. Those are things we know need to be addressed, as they relate to liveness and security (albeit, the latter only in case we change our signing protocol).
- API cleanliness: the doc states that we won't rename some API endpoints to reflect the proper terminology. The reasoning given is that the overhead is too large. The arguments seem LLM-generated, as the overhead is pretty negligible, considering we do node and contract updates on a regular basis. Left a comment with more details.

docs/foreign-chain-transactions.md
```diff
@@ -431,10 +432,10 @@ pub fn vote_update_foreign_chain_providers(
...
 - **One active vote per participant per chain.** Votes are keyed by `(participant, chain)`. Recasting for a chain overwrites that participant's slot for *only that chain*; chains the participant didn't touch in the new call keep their prior slot.
-- **Gated on the protocol signing threshold.** A chain applies when the count of participants holding the same canonical `(providers, threshold)` pair reaches `self.threshold()?.value()` — the same threshold used by `verify_tee` and `vote_add_os_measurement`. There is no separate per-chain *voting* threshold; the per-chain numeric on `ChainVote.threshold` is the *RPC response quorum* (a runtime concept consumed by nodes), not a voting parameter.
+- **Gated on the protocol signing threshold.** A chain applies when the count of participants holding the same canonical `(providers, quorum)` pair reaches `self.threshold()?.value()` — the same threshold used by `verify_tee` and `vote_add_os_measurement`. There is no separate per-chain *voting* threshold; the per-chain numeric on `ChainVote.quorum` is the *RPC quorum* (a runtime concept consumed by nodes), not a voting parameter.
```
gilcu3: "A chain applies" is a bit ambiguous. Maybe better to say a chain becomes approved or available (not sure what the correct terminology is)

docs/foreign-chain-transactions.md
```diff
@@ -518,7 +519,7 @@ This ensures different nodes query different providers for the same request whil
 
 ## Failure and Timeout Behavior
 
-* Nodes **abstain** if RPC queries fail or extraction fails.
+* Nodes **do not participate** if RPC queries fail or extraction fails.
 * A failed verification does **not** produce an on-chain failure response. The request eventually times out and fails with the standard timeout error.
```
gilcu3, line 558: this is something that we should actually change. Timing out when we know why it failed, and might be able to take action, seems suboptimal

docs/foreign-chain-transactions.md
```diff
@@ -614,11 +615,11 @@ providers require no auth at all.
...
-* **Rollout coordination**: A chain is only considered supported once **every** active participant has registered it; a single lagging operator can delay enabling a new chain.
+* **Incomplete chain coverage**: A chain is whitelisted as soon as the network votes in its whitelist entry, independent of any single operator. A node that hasn't configured a whitelisted chain is treated like a node that is down for it — it does not participate in that chain's verification requests, and the pre-generated triples/presignatures it co-owns become offline assets. As long as the leader knows the node is down there is no waste, but an asset that stays offline for a long period is eventually discarded. Foreign-tx presignatures live in their own dedicated `ForeignTx`-domain pool, but triples are shared across all CaitSith domains with the same signing threshold, so stranding them also reduces ordinary `sign()` presignature generation — the waste is not confined to that chain's availability. This is mitigated operationally by alerting on any active node that does not cover a whitelisted chain, rather than by the protocol.
```
gilcu3: in this case I think only the presignatures for that domain would become offline assets. Triples should only become offline if the participants are actually down (as they are shared with other domains).

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,165 @@
...
+long as the leader knows a node is down there is no waste, only an asset that stays
+offline for a long period is eventually discarded, wasting the work that produced it.
+Two surfaces where that lands:
+
+- **Presignatures are per-domain.** Foreign-tx signing uses a dedicated `ForeignTx`
+  domain (`DomainPurpose::ForeignTx`), so stranded foreign-tx presignatures stay in
+  that pool.
+- **Triples are shared** across all CaitSith domains with the same signing threshold,
+  so stranding them also dents ordinary `sign()` presignature generation — this cost
+  is **not** confined to `C`.
```
gilcu3: not sure if it is worth going into such details here. Depending on how it is implemented, we should be able to avoid losing triples for example.

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,165 @@
...
+doc-alignment rule):
+
+- `docs/foreign-chain-transactions.md` — "Contract State (Foreign Chain
+  Configurations)", the `get_supported_foreign_chains` description (now superseded by
+  `get_whitelisted_foreign_chains` / `get_available_foreign_chains`), the operator-flow
+  note about a chain appearing "only once **every** active participant has registered
+  it", and the rollout-coordination risk all describe the intersection rule.
+- `docs/design/allowing-per-node-foreign-chain-rpc-configuration.md` requirement #5
+  — restate it as the whitelisted/available split above (registration feeds the
+  available set; it is not monitoring-only).
```
gilcu3: is this LLM stuff that needs to be removed?

gilcu3 (dismissed):
Thank you, left some comments.

One helpful change for future reviewers would be to make sure lines are capped to 80 chars or less. Some files here had actually much bigger changes than I initially assumed because of this 🙃

docs/design/calculating-supported-foreign-chains.md
anodar: I was treating these `docs/` more like a conventional design doc, scoped to a proposal of some idea, tracking issues don't really belong in design itself.

With these markdowns line is blurred, but I still find it quite awkward to have tracking issues created for design that's not completely approved (even if part that issue covers is agreed). Would make more sense to have maybe a section for milestones or something like that and have it as one of the goals (e.g. have monitoring setup for available foreign chains).


Anyway, created issues and referenced them here.

docs/design/calculating-supported-foreign-chains.md
anodar: Introduced new methods as well.

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
anodar: Moved definitions to the main doc and added "coverage"

docs/foreign-chain-transactions.md
anodar: Dropped it.

docs/design/calculating-supported-foreign-chains.md
anodar: Done and referenced (https://github.com/near/mpc/issues/3476)

docs/design/calculating-supported-foreign-chains.md
anodar: Moved to main doc.

docs/design/calculating-supported-foreign-chains.md
anodar: No, added tracking issue and referenced it here as well.

docs/design/calculating-supported-foreign-chains.md
anodar: Done.

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
anodar: Added definition of "cover"

docs/foreign-chain-transactions.md
anodar: Reworeded.

docs/foreign-chain-transactions.md
anodar: Changed.

docs/foreign-chain-transactions.md
anodar: Addressed.

docs/design/calculating-supported-foreign-chains.md
anodar: Yes. Dropped.

docs/design/calculating-supported-foreign-chains.md
anodar: Left out the details.

anodar:
Summary of the agreements from meetings about this issue:
- we'll be expecting all nodes to support all whitelisted chains
- we'll have monitoring setup for available chains so that in healthy network all nodes can validate all whitelisted foreign chain
- we'll introduce new methods for listing available foreign chains (that will be whitelisted chains that have at least signing threshold of node providers supporting it)
- once we migrate the contract we'll deprecate old methods and drop them.

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
```diff
@@ -107,18 +118,18 @@ Since the nodes are running in a Trusted Execution Environment (TEE), this funct
...
+When a foreign TX verification request is processed by a set of nodes, every node individually queries its locally-configured RPC providers for that chain. A node considers the foreign TX verified iff at least the per-chain RPC quorum of providers agreed. If fewer than the RPC quorum agree, the node errors out and produces no signature share. This sub-quorum outcome must be **terminal** for the request — the leader does not re-attempt it, unlike the generic per-request retry in `crates/node/src/requests/queue.rs` (`MAX_ATTEMPTS_PER_REQUEST_AS_LEADER`). This is an implementation requirement, not current behavior. See [Calculating the whitelisted and available foreign-chain sets](calculating-supported-foreign-chains.md#verification-behavior).
 
 The quorum value comes from on-chain `ChainEntry.threshold`, voted in as part of the same `ChainVote` that voted the chain's provider list — so participants agree on both "which providers are trusted" and "how many of them must concur" in one round.
```
gilcu3:
```suggestion
The quorum value comes from on-chain `ChainEntry.quorum`, voted in as part of the same `ChainVote` that voted the chain's provider list — so participants agree on both "which providers are trusted" and "how many of them must concur" in one round.
```

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
```diff
@@ -107,18 +118,18 @@ Since the nodes are running in a Trusted Execution Environment (TEE), this funct
...
+- **PR 2** ([#3249](https://github.com/near/mpc/pull/3249)): contract-side voting on the whitelist. Adds `vote_update_foreign_chain_providers(votes: Vec<ChainVote>)`, the `ProviderVotes` pending-vote storage, canonicalization (`providers` sorted by `provider_id`; duplicate chain or `provider_id` in a batch rejected with `InvalidParameters::MalformedPayload`), and the `clean_tee_status` extension that drops votes from non-participants. Voting is **full-snapshot**: each `ChainVote` proposes the chain's complete state (provider list + RPC response quorum), and the chain's stored `ChainEntry` is replaced once the protocol's signing threshold of participants holds the same canonical `(providers, quorum)` pair (same gate as `vote_add_os_measurement`). Drops the original Add/Remove-ops design for two reasons: (1) snapshot semantics canonicalize trivially (sort the proposed list), avoiding the order-of-apply ambiguity Add/Remove batches introduced, and (2) bundling the RPC response quorum into `ChainVote.threshold` collapses what was originally going to be two separate vote endpoints (whitelist + quorum) into one.
```
gilcu3:
```suggestion
- **PR 2** ([#3249](https://github.com/near/mpc/pull/3249)): contract-side voting on the whitelist. Adds `vote_update_foreign_chain_providers(votes: Vec<ChainVote>)`, the `ProviderVotes` pending-vote storage, canonicalization (`providers` sorted by `provider_id`; duplicate chain or `provider_id` in a batch rejected with `InvalidParameters::MalformedPayload`), and the `clean_tee_status` extension that drops votes from non-participants. Voting is **full-snapshot**: each `ChainVote` proposes the chain's complete state (provider list + RPC response quorum), and the chain's stored `ChainEntry` is replaced once the protocol's signing threshold of participants holds the same canonical `(providers, quorum)` pair (same gate as `vote_add_os_measurement`). Drops the original Add/Remove-ops design for two reasons: (1) snapshot semantics canonicalize trivially (sort the proposed list), avoiding the order-of-apply ambiguity Add/Remove batches introduced, and (2) bundling the RPC response quorum into `ChainVote.quorum` collapses what was originally going to be two separate vote endpoints (whitelist + quorum) into one.
```

docs/foreign-chain-transactions.md
gilcu3:
```suggestion
Voting uses the protocol's existing signing threshold (`self.threshold()?.value()`), the same gate as `verify_tee` and `vote_add_os_measurement`. An earlier design proposed a separate per-chain *voting* threshold so mainnet and testnet could be voted in under different agreement requirements; that was dropped because (a) there's no setter that could safely populate it without itself being voted in, leaving a hardcoded default that's strictly weaker than the protocol threshold, and (b) the per-chain numeric on `ChainVote.quorum` already covers the *runtime* security knob — how many of N whitelisted providers must agree for a node to accept a response — which is what operators actually need to tune per chain.
```

gilcu3 (dismissed):
Thank you for the changes!

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
```diff
@@ -107,18 +118,18 @@ Since the nodes are running in a Trusted Execution Environment (TEE), this funct
 
 ### Individual node quorum of RPC providers for verification requests
 
-When a foreign TX verification request is processed by a set of nodes, every node individually queries its locally-configured RPC providers for that chain. A node considers the foreign TX verified iff at least a per-chain quorum number of providers agreed.
+When a foreign TX verification request is processed by a set of nodes, every node individually queries its locally-configured RPC providers for that chain. A node considers the foreign TX verified iff at least the per-chain RPC quorum of providers agreed. If fewer than the RPC quorum agree, the node errors out and produces no signature share. This sub-quorum outcome must be **terminal** for the request — the leader does not re-attempt it, unlike the generic per-request retry in `crates/node/src/requests/queue.rs` (`MAX_ATTEMPTS_PER_REQUEST_AS_LEADER`). This is an implementation requirement, not current behavior. See [Calculating the whitelisted and available foreign-chain sets](calculating-supported-foreign-chains.md#verification-behavior).
```
kevindeforth: **Remark:**

This is a pattern noticeable in quite a few of our docs: we sometimes concern the reader with things that are not relevant to the subject or the idea we are trying to convey:

> the leader does not re-attempt it, unlike the generic per-request retry in `crates/node/src/requests/queue.rs` (`MAX_ATTEMPTS_PER_REQUEST_AS_LEADER`)

Could be shortened to:
> the leader does not re-attempt it.

There is no reason to burden the reader with knowledge about a different component. Linking to it adds a lot of noise, as it implies: "Hey, this is relevant, better look at that file", whereas, it really isn't and it distracts from the main idea of the sentence.

This is not a blocker, but I would like us to get better at writing docs.

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,129 @@
...
+
+**Whitelisted** is a stable, operator-visible commitment: it changes only by vote,
+never flaps with node churn, and no single operator can change it. Every node is
+expected to cover every whitelisted chain.
+
+**Available** avoids spending resources on a request the network can't fulfill: one
+for a chain without a signing threshold's worth of coverage is rejected up front, not
+attempted and left to time out. Because it requires `signing_threshold` — not all —
+participants, a chain leaves it only when more than `n − signing_threshold` nodes
+stop covering it.
```
kevindeforth: **blocker:** This is incorrect, a chain is considered not supported if less than `signing_threshold` nodes support it, which is different to what is stated here.

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -0,0 +1,129 @@
...
+
+In a healthy network `available == whitelisted`. A gap means nodes are down or
+misconfigured for a chain — an operational anomaly that alerting surfaces (see
+[Per-node registration](#per-node-registration)), not a steady state.
```
kevindeforth: What is the *gap* referring to here?

kevindeforth (dismissed):
Thanks for the changes!

I think the doc still has potential regarding conciseness and language, but the proposed design is sound and our docs are currently of mixed quality. So I don't see a good reason to block this (there is one *blocker* remark, but it's impact is minimal, as it's quite obvious what the writer intents to say. So, it's not really a blocker).

Regarding Language:
Please take note that there are filler-words, unnecessary adjectives and non-relevant information, all of which add noise, making it harder to grasp the main idea of some paragraphs. I suspect those were LLM generated.

We don't have any policy or guidelines surrounding usage of LLMs for design docs, but I would advise _against_ having AI generate the doc or its structure from scratch. Personally, I found it best to write the doc myself and reserve the use of LLMs (in the context of design docs) for:
- adding references, inserting links to other docs or code
- fixing typos and grammar
- review the proposed design
- propose variable or function names
- selectively explore alternative wording

In the end, while more work up-front, I have the impression that it makes the review process smoother.

docs/foreign-chain-transactions.md
gilcu3: github rendering is horrible here! The only requested change is using `ChainVote.quorum`, same as the one above

docs/design/calculating-supported-foreign-chains.md
anodar: Refers the 2 sets mentioned in the sentence before. Gap = exists a node that's whitelisted but not available. Reworded to make it clear.

docs/design/calculating-supported-foreign-chains.md
anodar: I'm not sure I follow. Less than `signing_threshold` supporting it means more than `n - signing_threshold` not supporting it, which is what is stated here. Reworded to make it clear.

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
anodar: Dropped it here and in few more places.

gilcu3 (approved):

kevindeforth (approved):
Thanks!

docs/design/allowing-per-node-foreign-chain-rpc-configuration.md
```diff
@@ -17,11 +17,22 @@ As part of [#2648](https://github.com/near/mpc/issues/2648), we want to be able
...
+4. Every node partaking in a foreign signature verification request queries all its locally configured RPC providers for the relevant chain, independently of other nodes. A quorum of those RPC providers must agree on the verification; if fewer than the quorum agree, the node errors out that foreign-tx validation and does **not** retry the request.
+5. Every node is expected to **cover** every whitelisted foreign chain at all times. There is no
+   on-chain enforcement, so the system must tolerate temporary **incomplete coverage**: a chain is
+   served while it is **available** (≥ signing threshold of active nodes cover it), and
+   `verify_foreign_transaction` early-rejects a whitelisted-but-unavailable chain. Per-node
+   registrations feed the availability computation. See
+   [Calculating the whitelisted and available foreign-chain sets](calculating-supported-foreign-chains.md).
+6. Monitoring and off-chain alert channels notify operators of incomplete coverage (a node not
+   covering some whitelisted chain); there is no on-chain penalty.
```
netrome, line 35: I'm not following why we update this existing design doc? This thing reads quite weirdly now as the document is phrased as a proposal for a change (i.e. "currently the behavior is X, but we want to change it to Y")

Imo, this doc has served it's purpose, and the new update to calculate the supported foreign chain set is the other one you added in this PR. I think we should either remove this one, or think about rephrasing it if there's any information in it we want to keep right now. The current state of it is quite confusing to me.

docs/foreign-chain-transactions.md
```diff
@@ -285,9 +285,44 @@ pub fn derive_foreign_tx_tweak(predecessor_id: &AccountId, path: &str) -> Tweak
 This ensures key material used for validated foreign transactions is **always** distinct from
 general-purpose `sign()` keys, even if the same account and derivation path are reused.
 
+## Terminology
+
+Defined here once; the two design docs
+([calculating the whitelisted/available sets](design/calculating-supported-foreign-chains.md),
+[per-node RPC configuration](design/allowing-per-node-foreign-chain-rpc-configuration.md)) point
+back here rather than redefining them.
```
netrome, line 293: IMO, the design docs are a good venue to discuss implementations - but we should backport the conclusions so that this main document stays in sync. It feels like it has already started drifting.

netrome (commented):
Nice stuff. However, I feel the modifications to the per-node-foreign-chain-configuration design doc was a bit confusing, and it feels like we have some doc drift in our main feature document for foreign tx requests. It would be worth doing some spring cleaning of these docs next week.

PR #3439, opened 2026-06-01
fix: enforce signer=predecessor across all governance methods

### LINKED ISSUE #2383: Resolve inconsistent signer==predecessor requirement across governance methods
### Background

`voter_or_panic()` calls `assert_caller_is_signer()`, which requires `env::signer_account_id() == env::predecessor_account_id()`. This prevents `propose_update`, `vote_update`, `verify_tee`, and `vote_code_hash` from being called via a governance multisig contract (where `predecessor` would be the multisig contract address, not the participant's account).

By contrast, `vote_new_parameters` uses `AuthenticatedAccountId` directly and does **not** call `assert_caller_is_signer()`, allowing cross-contract calls where a multisig is the predecessor but the participant is the signer.

This creates an inconsistency: the most sensitive governance actions (contract upgrades via `vote_update`, TEE policy via `vote_code_hash`) require direct calls, while resharing proposals (`vote_new_parameters`) allow multisig. There is no clear security rationale for this difference — a compromised signing key poses the same threat regardless of whether the call is direct or via multisig.

### User Story

As a **node operator**, I want a consistent and documented policy on whether governance methods can be called via a multisig contract, so that I can set up appropriate governance workflows without encountering unexpected access control errors.

### Acceptance Criteria

Pick **one** of the following directions and implement it:

### Option A: Unify by also requiring direct calls for `vote_new_parameters`
- [ ] `vote_new_parameters` is updated to call `assert_caller_is_signer()` (or equivalent), matching the behavior of `voter_or_panic()`-gated methods.
- [ ] Document that all governance methods require direct calls from the participant's NEAR account.

### Option B: Unify by allowing multisig for all governance methods
- [ ] `voter_or_panic()` is updated to drop the `assert_caller_is_signer()` requirement, relying on `env::signer_account_id()` for participant authentication (consistent with how `vote_new_parameters` works).
- [ ] Evaluate and document the security implications of allowing cross-contract calls for `vote_update` and `vote_code_hash`.

### Option C: Keep the inconsistency but document it
- [ ] Add clear documentation in the governance guide explaining which methods require direct calls vs. allow multisig.
- [ ] Add code comments on `voter_or_panic()` and `vote_new_parameters` explaining the intentional difference and its rationale.

- **`voter_or_panic`** location: `crates/contract/src/lib.rs:1808-1814` — calls `assert_caller_is_signer()`.
- **`assert_caller_is_signer`** location: `crates/contract/src/lib.rs:1845-1855` — asserts `signer == predecessor`.
- **`vote_new_parameters`** location: `crates/contract/src/lib.rs:949-953` — does NOT call `assert_caller_is_signer`.
- **Methods gated by `voter_or_panic()`:** `propose_update`, `vote_update`, `verify_tee`, `vote_code_hash`.

https://github.com/near/mpc-private/pull/207

netrome:
closes #2383 

This PR adds predecessor checks for `vote_new_parameters`, `vote_add_domains`, `register_foreign_chain_support`, `vote_cancel_resharing` and `vote_cancel_keygen` which previously missed this check.

### DIFF
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -864,12 +864,19 @@ impl MpcContract {
     /// The epoch_id must be equal to 1 plus the current epoch ID (if Running) or prospective epoch
     /// ID (if Resharing). Otherwise the vote is ignored. This is to prevent late transactions from
     /// accidentally voting on outdated proposals.
+    ///
+    /// Like the other governance voting methods, this must be called directly from the
+    /// participant's own NEAR account: `assert_caller_is_signer()` requires
+    /// `signer_account_id == predecessor_account_id`, so calls forwarded through another
+    /// contract (e.g. a multisig) are rejected. This prevents a confused-deputy vote being
+    /// cast on a participant's behalf when they merely call into an untrusted contract.
     #[handle_result]
     pub fn vote_new_parameters(
         &mut self,
         prospective_epoch_id: EpochId,
         proposal: dtos::ThresholdParameters,
     ) -> Result<(), Error> {
+        Self::assert_caller_is_signer();
         let proposal: ThresholdParameters = proposal.into_contract_type();
         log!(
             "vote_new_parameters: signer={}, proposal={:?}",
@@ -2157,6 +2164,17 @@ impl MpcContract {
 
     /// Ensures the current call originates from the signer account itself.
     /// Panics if `signer_account_id` and `predecessor_account_id` differ.
+    ///
+    /// This enforces the network-wide policy that **all governance methods must be called
+    /// directly from the participant's own NEAR account**, never forwarded through another
+    /// contract such as a multisig. Authenticating on `signer_account_id` alone would be a
+    /// confused-deputy hazard: a malicious or buggy intermediate contract could cast a vote
+    /// on a participant's behalf whenever that participant merely calls into it. Requiring
+    /// `signer == predecessor` removes the intermediary from the trust path.
+    ///
+    /// Governance methods gated by this check (directly or via [`Self::voter_or_panic`]):
+    /// `vote_new_parameters`, `propose_update`, `vote_update`, `vote_code_hash`, `verify_tee`,
+    /// and the other `voter_or_panic`-gated votes.
     fn assert_caller_is_signer() -> AccountId {
         let signer_id = env::signer_account_id();
         let predecessor_id = env::predecessor_account_id();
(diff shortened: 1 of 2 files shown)


crates/contract/src/lib.rs
```diff
@@ -944,6 +951,7 @@ impl MpcContract {
         &mut self,
         foreign_chain_support: dtos::SupportedForeignChains,
     ) -> Result<(), Error> {
+        Self::assert_caller_is_signer();
```
barakeinav1: Why not use `voter_or_panic()` here instead? It does the same current-participant check, so you could replace both this `assert_caller_is_signer()` and the `AuthenticatedAccountId::new(...)` below with `let account_id = self.voter_or_panic();` — states the voter requirement up top like in `vote_update`.

crates/contract/src/lib.rs
```diff
@@ -927,6 +933,7 @@ impl MpcContract {
     /// must be the same as the `next_domain_id` returned by state().
     #[handle_result]
     pub fn vote_add_domains(&mut self, domains: Vec<DomainConfig>) -> Result<(), Error> {
+        Self::assert_caller_is_signer();
```
barakeinav1, line 936: Same thing I'd like here — use `voter_or_panic()` like in https://github.com/near/mpc/pull/3439#discussion_r3339006733. But the membership check for this one lives in `running.rs::vote_add_domains`, so it's a much bigger refactor (you'd have to pass the authenticated participant down) — I'm not sure it's worth it.

barakeinav1 (dismissed):
LGTM — verified the check now reaches every governance/signer-authenticated method (directly, via `voter_or_panic`, or the attested-participant path), and the permissionless/`#[private]`/user-facing methods are correctly left alone. Left two non-blocking comments about using `voter_or_panic()` for `register_foreign_chain_support`/`vote_add_domains` — your call.

crates/contract/src/lib.rs
gilcu3: there is a small diff in that `voter_or_panic` does not enforce the `RunningState`, while the current code maybe does on purpose

gilcu3 (dismissed):
Thank you!

kevindeforth (dismissed):
Thanks!

crates/contract/src/lib.rs
netrome: Done in https://github.com/near/mpc/pull/3439/commits/ae1d3c549dd4842bfa3cffe42c7422e34e44b08c - note this is a small functional change (error -> panic + allowing outside of resharing) but I think that is fine here. Ideally we'd allow prospective participants to register their foreign chain support during resharing but I didn't want to refactor too much to allow that.

crates/contract/src/lib.rs
netrome: Skipped for now, it's a larger refactor as you noted and it's not obvious to me that it's appropriate since this should only be called during running state.

crates/contract/src/lib.rs
netrome: Edit, small rustdoc fix - amended the fix commit as https://github.com/near/mpc/pull/3439/commits/ad61264560d373cada90b388c7af24143adccfec

barakeinav1: >there is a small diff in that voter_or_panic does not enforce the RunningState, while the current code maybe does on purpose

I'm not familiar enough with this. @DSharifi ,@kevindeforth  do you think this is ok?

barakeinav1 (approved):

gilcu3 (approved):

crates/contract/src/lib.rs
kevindeforth: Hmm, that's fine, but I wonder - shouldn't `register_foreign_chain_support` be called only by the node, not by the operator?

So, should we compare the `env::signer_account_pk()`, instead of just the account id?

gilcu3:
@netrome took the liberty to correct a typo and add to the merge queue

netrome:
> @netrome took the liberty to correct a typo and add to the merge queue

Thank you :pray:

crates/contract/src/lib.rs
netrome: Perhaps. Though allowing it to be called by the operator can be a nice escape-hatch if there's any problem with the node submission of the foreign chain support like we've had during recent incidents related to this.

PR #3470, opened 2026-06-04
chore: remove EpochData legacy cleanup behavior

### LINKED ISSUE #3348: Clean up upgrade logic from last epoch data format upgrade
### Background

A long time ago, in a galaxy far far away, issue #1032 introduced a new form of epoch data stored in the `EpochData` DB column. It introduced some upgrade logic to migrate from the old format. For example, the `EpochDataWrapper` enum was introduced to distinguish between the new and old epoch data, and the `cleanup_behavior_during_update` method was created to define the cleanup behavior during an update.

Now all known production nodes run at least 3.10.0, so it should be very safe to remove the legacy code paths.

### User Story

As a developer I want our code base clean from legacy migration glue.

### Acceptance Criteria

We've removed any code path relating to the upgrade from the previous epoch data format and the current one.

_No response_

gilcu3:
Closes #3348

### DIFF
--- a/crates/node/src/assets/cleanup.rs
+++ b/crates/node/src/assets/cleanup.rs
@@ -25,17 +25,9 @@ pub fn delete_stale_triples_and_presignatures(
     ecdsa_domain_ds: Vec<DomainId>,
     triple_thresholds: Vec<ReconstructionThreshold>,
 ) -> anyhow::Result<()> {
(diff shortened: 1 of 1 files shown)


netrome (approved):
Thanks 🙏

kevindeforth (approved):
Oh nice, completely forgot about that one.

PR #3478, opened 2026-06-05
test: verify participants below threshold are not kicked out when attestation fails

### LINKED ISSUE #1132: Integration test (Rust-only) to ensure nodes with expired attestations below threshold number are NOT kick out
### Background

this is a follow up test for  #1092 (and the PR https://github.com/near/mpc/pull/1099 - can be used as infrastructure) 
need to add a test for the following scenario:
In case removal of expired node will reduce the participant number below threshold number. 
The nodes should not be kickout.(since will permanently break our network).
 instead, node should be kept as participant, BUT the network should stop to accept signature request.

### User Story

as a developer, I'd like to have test converge for all flows.

### Acceptance Criteria

test is implemented and passing (or a new issue is create for a bug in the code) 

_No response_

gilcu3:
Closes #1132

### DIFF
--- a/crates/contract/tests/sandbox/tee.rs
+++ b/crates/contract/tests/sandbox/tee.rs
@@ -862,9 +862,9 @@ async fn verify_tee__should_keep_participants_and_stop_signing_when_kickout_drop
     let initial_participants = assert_running_return_participants(&contract).await?;
     assert_eq!(initial_participants.participants.len(), PARTICIPANT_COUNT);
 
-    // Expiring 2 of 3 leaves 1 valid participant, which is below the threshold of 2.
-    let expired_count = PARTICIPANT_COUNT - 1;
-    let remaining_valid = PARTICIPANT_COUNT - expired_count;
+    // Expire all but `threshold - 1` attestations, leaving the valid set exactly one
+    // below threshold regardless of the participant/threshold constants above.
+    let remaining_valid = threshold.0 as usize - 1;
     assert!(
         remaining_valid < threshold.0 as usize,
         "test precondition: surviving participants ({remaining_valid}) must be below threshold ({})",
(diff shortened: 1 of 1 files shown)


crates/contract/tests/sandbox/tee.rs
```diff
@@ -822,3 +825,144 @@ async fn test_verify_tee_expired_attestation_triggers_resharing() -> Result<()>
...
+        .with_protocols(&[Protocol::CaitSith])
+        .with_number_of_participants(PARTICIPANT_COUNT)
+        .build()
+        .await;
+
+    let threshold = assert_running_return_threshold(&contract).await;
+    let initial_participants = assert_running_return_participants(&contract).await?;
+    assert_eq!(initial_participants.participants.len(), PARTICIPANT_COUNT);
+
+    // Expiring 2 of 3 leaves 1 valid participant, which is below the threshold of 2.
```
anodar: optional nit: I'd just say expiring expiring all but one (or threshold - 1) participant's attestation so that his doesn't get outdated if those constants above change.

crates/contract/tests/sandbox/tee.rs
```diff
@@ -822,3 +825,144 @@ async fn test_verify_tee_expired_attestation_triggers_resharing() -> Result<()>
...
+
+    // Then: verify_tee reports the network is no longer accepting requests.
+    let accepting_requests: bool = verify_result.json()?;
+    assert!(
+        !accepting_requests,
+        "verify_tee should return false when fewer than threshold participants remain valid"
+    );
+
+    // Then: no participant is kicked out — the contract stays Running with all participants.
+    let state_after_verify = get_state(&contract).await;
```
anodar, line 931: Question: I'm assuming participants are kicked out either before state changes, or transactionally as the state changes, otherwise this could be a race condition right?

anodar (dismissed):

crates/contract/tests/sandbox/tee.rs
gilcu3: they are kicked out in the same transaction, the one started by `verify_tee`

anodar (approved):
Thanks for the clarifications!

kevindeforth (approved):
Thanks a lot!
Was wondering if we had some more helper methods to reduce the boiler plate in sandbox tests, but I didn't find any.

Might be good to add a unit test for this as well.

PR #3481, opened 2026-06-05
refactor: wrap `BlockStatus` in handle

### LINKED ISSUE #3451: refactor: wrap Weak<AtomicBlockStatus> in a guard so the tracker owns block-status lifetime
### Background                                                                                                                                              
                                                                                                                                                              
  `RecentBlocksTracker::add_block` returns `Weak<AtomicBlockStatus>` so callers can observe canonical/final transitions on a block they care about:           
                                                                                                                                                              
  https://github.com/near/mpc/blob/d78386ff33aadbf6824de065bb205dddb961860b/crates/node/src/requests/recent_blocks_tracker.rs#L186-L196                       
                                                                                                                                                              
  The tracker is supposed to own block-status lifetime: when `prune_old_blocks` drops a `BlockNode`, the strong `Arc<AtomicBlockStatus>` goes with it and the status reads `None` on the next `Weak::upgrade`.
That invariant holds today only by convention — nothing stops a caller from upgrading the `Weak` and keeping the `Arc` alive past the tracker's prune horizon, silently extending the status's lifetime.                                                         
                                                                                                                                                              
The two existing call sites are well-behaved — they upgrade transiently, read, and drop:                                                                    
                                                                                                                                                              
https://github.com/near/mpc/blob/d78386ff33aadbf6824de065bb205dddb961860b/crates/node/src/requests/queue.rs#L104-L108                                       
                                                                                                                                                              
  https://github.com/near/mpc/blob/d78386ff33aadbf6824de065bb205dddb961860b/crates/node/src/requests/queue.rs#L383-L389                                       
                                                                                                                                                              
  …and the only public methods on `AtomicBlockStatus` are `is_final()` / `is_canonical()`:                                                                    
                                                                                                                                                              
  https://github.com/near/mpc/blob/d78386ff33aadbf6824de065bb205dddb961860b/crates/node/src/requests/recent_blocks_tracker.rs#L142-L186                       
                                                                                                                                                              
  So we can enforce the invariant in the type system: wrap the `Weak` in a small guard type (e.g. `BlockStatusRef`) that hides the `Weak` and exposes only `is_final()` / `is_canonical()` returning `Option<bool>` (`None` when the tracker has already pruned the block). The guard upgrades the `Arc` for the duration of one read and drops it before returning. Callers can no longer accidentally pin the status's lifetime, and we get a single point to enforce additional invariants later (e.g. metrics, tracing).                                                                                                        
                                                                                                                                                              
  ### Acceptance Criteria                                                                                                                                     
                                                                                                                                                              
  - [ ] New type (e.g. `BlockStatusRef`) wraps `Weak<AtomicBlockStatus>` with a private inner field.                                                          
  - [ ] `AddBlockResult::block_ref` (and the `Weak` field on `QueuedRequest` / `IndexedRespondTxs`) is replaced with `BlockStatusRef`.                        
  - [ ] Public read API: `BlockStatusRef::is_final() -> Option<bool>` and `BlockStatusRef::is_canonical() -> Option<bool>` (`None` ⇔ tracker dropped the      
  block).                                                                                                                                                     
  - [ ] No public method returns or otherwise leaks an `Arc<AtomicBlockStatus>`.                                                                              
  - [ ] `AtomicBlockStatus` is no longer `pub` (only `BlockStatusRef` is).                                                                                    
  - [ ] Existing `Weak::upgrade` call sites in `queue.rs` migrate to the new API.

kevindeforth:
resolves #3451

We have an invariant that assumes pruning a block from the tracker drops its blocks status ref count to zero.
This PR introduces `BlockStatusHandle`, a thin wrapper around `Weak<AtomicBlockStatus>`, ensuring consumers of the block status don't accidentally violate the invariant.

### DIFF
--- a/crates/node/src/requests/recent_blocks_tracker.rs
+++ b/crates/node/src/requests/recent_blocks_tracker.rs
@@ -14,7 +14,7 @@ use super::metrics::{MPC_BLOCKS_INDEXED, MPC_FINALIZED_BLOCKS_INDEXED};
 /// This class provides two important functionalities:
 ///  - Converts a stream of optimistic blocks from the indexer into a stream of finalized
 ///    blocks.
-///  - For each block added via `add_block`, it returns a `BlockStatusGuard` that can be
+///  - For each block added via `add_block`, it returns a `BlockStatusHandle` that can be
 ///    used to observe that block's current `BlockStatus` (non-canonical, canonical or final).
 ///
 /// This class provides the following invariants (provided the requirements listed below are met):
@@ -137,14 +137,14 @@ impl From<BlockStatus> for u8 {
     }
 }
 
-pub struct AtomicBlockStatus(AtomicU8);
+struct AtomicBlockStatus(AtomicU8);
 
 impl AtomicBlockStatus {
-    pub fn is_final(&self) -> bool {
+    fn is_final(&self) -> bool {
         self.0.load(Ordering::Relaxed) == u8::from(BlockStatus::Final)
     }
 
-    pub fn is_canonical(&self) -> bool {
+    fn is_canonical(&self) -> bool {
         let status = self.0.load(Ordering::Relaxed);
         status == u8::from(BlockStatus::OptimisticAndCanonical)
             || status == u8::from(BlockStatus::Final)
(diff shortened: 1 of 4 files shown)


kevindeforth:
@claude review

gilcu3 (approved):
Thank you!

anodar (approved):

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -181,17 +181,37 @@ impl AtomicBlockStatus {
...
 }
 
+/// A handle to a block's status held by the [`RecentBlocksTracker`].
+///
+/// Consumers can read the current finality status but cannot hold a strong
+/// reference to the underlying atomic. Pruning a block from the tracker
+/// drops its ref count to zero.
+#[derive(Clone)]
+pub struct BlockStatusHandle(Weak<AtomicBlockStatus>);
+
```
netrome, line 191: Nit(type ordering): I'd find it more readable if this type was declared before `AtomicBlockStatus`, since this type references `AtomicBlockStatus` (i.e. more significant types first + public types before private types).

crates/node/src/requests/recent_blocks_tracker.rs
```diff
@@ -181,17 +181,37 @@ impl AtomicBlockStatus {
...
+    /// `None` if the block has been pruned from the tracker.
+    pub fn is_final(&self) -> Option<bool> {
+        self.0.upgrade().map(|s| s.is_final())
+    }
+
+    /// `None` if the block has been pruned from the tracker.
+    pub fn is_canonical(&self) -> Option<bool> {
+        self.0.upgrade().map(|s| s.is_canonical())
+    }
+}
```
netrome, line 202: Alright, so the purpose is to reduce the interface so consumers can't accidentally upgrade this handle?

netrome (approved):
Looks good to me. Noted that this is not a pure refactor but a small behavioral change as we now do two separate upgrades for the separate checks rather than keeping the same pointer. I don't see how that should be able to break anything in this context, and I currently don't have the time to dive into this rabbit hole.

crates/node/src/requests/recent_blocks_tracker.rs
kevindeforth: Yeah, exactly :+1:

From your main comment:

> as we now do two separate upgrades for the separate checks rather than keeping the same pointer. I don't see how that should be able to break anything in this context, and I currently don't have the time to dive into this rabbit hole.

I guess we could expose a method that returns both of these booleans simultaneously, but I don't expect this to become even remotely a performance issue.

Note that in a previous version, I exposed a `try_with` method that gave the caller access to the underlying `AtomicBlockStatus`. It seemed less clean from an API perspective, considering that we want to move all of this into the chain gateway crate.
But we can revisit this in the future.

crates/node/src/requests/recent_blocks_tracker.rs
kevindeforth: I also prefer the ordering you propose, but I already started work on moving this to the chain gateway crate.
I will go ahead and merge, but I keep this in mind.

crates/node/src/requests/recent_blocks_tracker.rs
netrome: Yeah, I'm not concerned with performance. Just if something is a pure refactor it's a no-brainer to review. If there's actually behavioral changes, it requires deeper understanding of the surrounding system to know if the change is okay or if it can introduce bugs.

PR #3483, opened 2026-06-05
chore: deployment of the contract to 3.11.2

### LINKED ISSUE #3473: chore: remove migrations after 3.11 release
### Background

After the v3.11 migration is fully deployed, migration-only code from v3.10.0 → v3.11.2 can be removed.

### User Story

_No response_

### Acceptance Criteria

- [ ] Remove `crates/contract/src/v3_10_state.rs` and the ` try_state_read::<v3_10_state::MpcContract>()` branch in `MpcContract::migrate` (`crates/contract/src/lib.rs`)
- [ ] Add v3.11.2 WASM to `crates/contract-history/archive/`, expose `version_3_11_2()` in `crates/contract-history/src/lib.rs`, and point `current_mainnet` / `current_testnet` at it so upgrade tests keep passing

_No response_

SimonRastikian:
Closes #3473


crates/contract-history/archive/mpc-contract-v3.11.0.wasm
gilcu3, line 1: we don't need these files (and given that git does not behave well with binaries, better not add them). Having 3.11.2 is enough

crates/contract/src/v3_11_2_state.rs
```diff
@@ -0,0 +1,71 @@
...
+    update::ProposedUpdates,
+};
+
+/// `3.11.2`'s `MpcContract` layout. Identical to the current `MpcContract`: no field
+/// has been added, removed, or reshaped since the `3.11.2` release, so every field type
+/// is imported from the live codebase and the `From` impl below is a straight 1:1 copy.
+///
+/// Keep this module in lock-step with [`crate::MpcContract`]: the moment a field's borsh
+/// layout diverges, shadow the old type here (see this module's history for examples) so
+/// state written by the `3.11.2` contract still deserializes during migration.
```
gilcu3: this comment is a bit more verbose than it should. We should only mention the second part

gilcu3 (changes requested):
Small change needed ( if CI passes). There is no need to store all contracts, just the ones we use for migrations.

SimonRastikian:
@claude review this PR

gilcu3 (approved):
Thank you!

nit: the PR title reads as if we were deploying the contract here. A name like:

> chore: contract state migration cleanup after 3.11.2 deployment

is as accurate as it can get

crates/contract/src/v3_11_2_state.rs
anodar, line 1: Nit: would be nice to instead rename previous file so that we can see the diff.

anodar (approved):

Issue #3475, opened 2026-06-04
Implement proposed design for available foreign chains for validation


anodar:
### Background

Currently we take intersection of supported chains from all the participants to calculate supported foreign chains. If single node provides empty list this results in none of the chains being supported.

### Acceptance Criteria

Design proposal from https://github.com/near/mpc/issues/3474 is implemented

netrome:
Setting as P1, let's prioritize this for next week.

anodar:
Assigning child tasks of this one to the paren of this task to get rid of unnecessary indirection. 

That is changing structure from this: 
```
- Prevent single node disabling foreign chain validation feature in the network
  - Propose design
  - Implement proposed design for available foreign chains for validation
    - Contract changes
    - Node wiring
```

To this:

```
- Prevent single node disabling foreign chain validation feature in the network
  - Propose design
  - Contract changes
  - Node wiring
```


Closing this one.

Issue #3434, opened 2026-06-01
Prevent single node disabling foreign chain validation feature in the network


anodar:
### Background

Currently we take intersection of supported chains from all the participants to calculate supported foreign chains. If single node provides empty list this results in none of the chains being supported.

### Acceptance Criteria

- have agreement on how should allowed foreign chain list be calculated
- have it implemented

gilcu3:
@anodar notice that `High` Priority is not the same as P0/P1/P2, which are project dependent priorities (yeah I agree it is confusing :| ). I just set this as P1 (but there is no visual queue for that in the log here).

kevindeforth:
Hmm, I would not necessarily label this a "bug". More like a missing feature.

Due to the current implementation, we require that _all_ participants in the network always support _all_ foreign chains, so the current behavior is technically correct.

Allowing heterogeneity will require changes on the lower level, specifically, how we select participants for computing a signature to a foreign chain request.

kevindeforth:
It does look like we have the same asset queue driven participant selection: https://github.com/near/mpc/blob/277842bb62af38ff74c404f59e5bea532dea0c0a/crates/node/src/providers/verify_foreign_tx/sign.rs#L62

gilcu3:
@anodar did you pin this issue on purpose or was it a miss click?

anodar:
@gilcu3 just noticed your reply here. No it was mis-click, didn't even know github has pinning :D

PR #3484, opened 2026-06-05
feat: Implement available foreign chain registrations (new API)

anodar:
Introduces new API (and corresponding DTOs)
  - register_available_foreign_chain_config
  - get_available_foreign_chains() -> AvailableForeignChains
  - get_available_foreign_chain_by_node() -> BTreeMap<AccountId, AvailableForeignChains>
  
Node will call both legacy and new method to register config, even before contract is upgraded, but gracefully handle method not found error.

Once whitelist is voted in and contract is upgraded (to include new get_* methods), we can switch to accept foreign tx validation requests based on `get_available_foreign_chains`.

Implementation of the design: https://github.com/near/mpc/issues/3434

Work towards: https://github.com/near/mpc/issues/3475


anodar:
@claude review

crates/contract/src/snapshots/mpc_contract__tests__mpc_contract_borsh_schema_has_not_changed.snap
gilcu3, line 1: ouch I hate to see this file growing. Is this a transitional state, which is expected to decrease in the near future because we are creating duplicates?

crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
```diff
@@ -577,6 +577,31 @@ expression: abi
           }
         }
       },
+      {
+        "name": "get_available_foreign_chain_by_node",
+        "doc": " Per-participant view of which foreign chains each node currently covers. Feeds the\n available-set computation ([`Self::get_available_foreign_chains`]) and coverage alerting.",
```
gilcu3:
```suggestion
        "name": "get_available_foreign_chains_by_node",
        "doc": " Per-participant view of which foreign chains each node currently covers. Feeds the\n available-set computation ([`Self::get_available_foreign_chains`]) and coverage alerting.",
```

crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
```diff
@@ -1081,6 +1106,28 @@ expression: abi
           }
         }
       },
+      {
+        "name": "register_available_foreign_chain_config",
+        "doc": " (Re)registers the foreign chains this participant currently covers, feeding the available\n set ([`Self::get_available_foreign_chains`]). Idempotent.\n\n Note: `voter_or_panic` accepts callers in `Initializing` and `Resharing` phases, so a\n registration in those phases will be written to `available_foreign_chains_by_node` but\n the cache (`available_foreign_chains`) is not refreshed until the next `Running`-state\n recompute (e.g. `clean_foreign_chain_data` after `vote_reshared`, or the next call to\n this method once the contract is `Running`).",
```
gilcu3: we publish this ABI in the contract. So for docs that are meant for devs, here from "Note: " onwards, it is better to avoid embedding it, but using `//` instead of `///`

crates/e2e-tests/tests/common.rs
```diff
@@ -35,6 +35,7 @@ pub const CONTRACT_UPGRADE_COMPATIBILITY_MAINNET_PORT_SEED: u16 = 18;
 pub const CONTRACT_UPGRADE_COMPATIBILITY_TESTNET_PORT_SEED: u16 = 19;
 pub const TIMEOUT_METRIC_PORT_SEED: u16 = 20;
 pub const MIGRATION_BACK_PORT_SEED: u16 = 21;
+pub const AVAILABLE_FOREIGN_CHAINS_BY_NODE_PORT_SEED: u16 = 22;
```
gilcu3: oh I think this will have a collision with #3410

crates/e2e-tests/tests/foreign_chain_configuration.rs
```diff
@@ -130,3 +132,62 @@ async fn supported_foreign_chains__should_require_all_participants_to_register()
...
+            "expected exactly 1 empty registration (node 2), got {empty}"
+        );
+        Ok(())
+    })
+    .retry(
+        ConstantBuilder::default()
+            .with_delay(common::POLL_INTERVAL)
+            .with_max_times(
+                (CLUSTER_WAIT_TIMEOUT.as_millis() / common::POLL_INTERVAL.as_millis()) as usize,
+            ),
```
gilcu3, line 189: this seems like a useful construction for many tests, would be nice to make it reusable

crates/node/src/indexer/fake.rs
```diff
@@ -52,6 +52,10 @@ pub struct FakeMpcContractState {
     pub pending_verify_foreign_txs: BTreeMap<dtos::ForeignChainRpcRequest, VerifyForeignTxId>,
     supported_foreign_chains: dtos::SupportedForeignChains,
     supported_foreign_chains_by_node: dtos::ForeignChainSupportByNode,
+    /// Per-node map for the new `register_available_foreign_chain_config` API.
+    available_foreign_chains_by_node: BTreeMap<AccountId, dtos::AvailableForeignChains>,
+    /// Cached available set recomputed via threshold semantics (mirrors the real contract).
+    available_foreign_chains: dtos::AvailableForeignChains,
```
gilcu3, line 58: nit: we had the opposite order above

crates/node/src/indexer/tee.rs
```diff
@@ -41,13 +42,10 @@ async fn monitor_allowed_hashes<Fetcher, T, FetcherResponseFuture>(
...
-                        if error_msg.contains(
-                            "wasm execution failed with error: MethodResolveError(MethodNotFound)",
-                        ) {
-                            tracing::info!(target: "mpc", "method  not found in contract: {error_msg}");
+                        if is_method_not_found(&e) {
+                            tracing::info!(target: "mpc", "method not found in contract: {e:?}");
                         } else {
-                            tracing::error!(target: "mpc", "error reading tee state from chain: {error_msg}");
+                            tracing::error!(target: "mpc", "error reading tee state from chain: {e:?}");
                         }
```
gilcu3, line 51: seems strange that this code lives in `crates/node/src/indexer/tee.rs`. Still, could we make it so that only the methods that we expect to fail in the next release are logged with info? And add a TODO to remove this after the next release?

crates/node/src/indexer/types.rs
```diff
@@ -186,7 +191,11 @@ pub enum ChainSendTransactionRequest {
     VotePk(ChainVotePkArgs),
     StartKeygen(ChainStartKeygenArgs),
     VoteReshared(ChainVoteResharedArgs),
+    // Legacy registration. Sent alongside `RegisterAvailableForeignChainConfig` so a node
+    // works against both the current contract and one upgraded for #3475. Remove once the
+    // contract is upgraded and the whitelist is voted in.
     RegisterForeignChainConfig(ChainRegisterForeignChainConfigArgs),
+    RegisterAvailableForeignChainConfig(ChainRegisterAvailableForeignChainConfigArgs),
```
gilcu3: lets add TODOs for these to avoid forgetting to remove them later

crates/node/src/coordinator.rs
```diff
@@ -398,7 +409,20 @@ where
...
+            .send(
+                ChainSendTransactionRequest::RegisterAvailableForeignChainConfig(
+                    ChainRegisterAvailableForeignChainConfigArgs {
+                        available_foreign_chains,
+                    },
+                ),
+            )
+            .await
+        {
+            tracing::warn!(error = ?err, "failed to send register_available_foreign_chain_config transaction");
```
gilcu3: please mention here that this is fine for contract version X

deny.toml
```diff
@@ -23,6 +23,7 @@ ignore = [
   "RUSTSEC-2022-0054", # wee_alloc is unmaintained
   "RUSTSEC-2025-0134", # rustls-pemfile is unmaintained
   "RUSTSEC-2024-0370", # proc-macro-error is unmaintained
+  "RUSTSEC-2026-0173", # proc-macro-error2 is unmaintained
```
gilcu3, line 26: nit: this is on main already :)

crates/node/src/indexer/tx_sender.rs
```diff
@@ -291,18 +291,56 @@ async fn observe_tx_result(
...
+                .view_client
+                .get_available_foreign_chain_by_node(&indexer_state.mpc_contract_id)
+                .await
+            {
+                if is_method_not_found(&err) {
+                    tracing::warn!(
+                        target: "mpc",
+                        "register_available_foreign_chain_config is not available on the contract yet \
+                         (MethodNotFound); the contract has likely not been upgraded for #3475. \
+                         Foreign-chain registration will take effect after the upgrade."
```
gilcu3: would be nicer for operators looking at the logs to mention in what contract version this is fine

crates/node/src/indexer/tx_sender.rs
```diff
@@ -291,18 +291,56 @@ async fn observe_tx_result(
...
+        RegisterForeignChainConfig(_)
+        | StartKeygen(_)
         | StartReshare(_)
         | VotePk(_)
         | VoteReshared(_)
         | VoteAbortKeyEventInstance(_)
         | VerifyTee()
-        | ConcludeNodeMigration(_)
-        | RegisterForeignChainConfig(_) => Ok(TransactionStatus::Unknown),
+        | ConcludeNodeMigration(_) => Ok(TransactionStatus::Unknown),
```
gilcu3: was this clippy forcing a re-order after adding a comment?

crates/node/src/indexer/tx_sender.rs
```diff
@@ -291,18 +291,56 @@ async fn observe_tx_result(
                 Ok(TransactionStatus::NotExecuted)
             }
         }
+        RegisterAvailableForeignChainConfig(_) => {
+            // Registration is idempotent and best-effort; we only probe whether the contract
+            // exposes the new methods — we do not assert the write landed.  All view errors are
+            // swallowed: MethodNotFound is expected before the contract is upgraded for #3475,
+            // and any other transient view failure (network blip, indexer lag) is non-actionable
+            // because the registration already fire-and-forgot.
```
gilcu3: never seen fire-and-forgot before :)
```suggestion
            // because the registration already fire-and-forget.
```

crates/contract/src/lib.rs
```diff
@@ -191,6 +198,32 @@ impl SupportedForeignChainsByNode {
     }
 }
 
+/// Per-node map of the chains each participant reports covering, feeding the available set.
+#[near(serializers=[borsh])]
+#[derive(Debug)]
+struct AvailableForeignChainsByNode {
```
kevindeforth: As per the terminology in the docs, shoult this be called `ForeignChainsCoveredByNode` or `CoveredForeginChainsByNode`?

crates/contract/src/lib.rs
```diff
@@ -969,6 +1002,79 @@ impl MpcContract {
...
+    /// (Re)registers the foreign chains this participant currently covers, feeding the available
+    /// set ([`Self::get_available_foreign_chains`]). Idempotent.
+    ///
+    /// Note: `voter_or_panic` accepts callers in `Initializing` and `Resharing` phases, so a
+    /// registration in those phases will be written to `available_foreign_chains_by_node` but
+    /// the cache (`available_foreign_chains`) is not refreshed until the next `Running`-state
+    /// recompute (e.g. `clean_foreign_chain_data` after `vote_reshared`, or the next call to
+    /// this method once the contract is `Running`).
+    #[handle_result]
+    pub fn register_available_foreign_chain_config(
```
kevindeforth: For this, `set_foreign_chain_coverage` would be more in line with the terminology defined in the docs.
(btw, still happy tor rename `covered` to `supported`, as it sounds less awkward in some places, but both are fine with me).

crates/node/src/indexer/tx_sender.rs
```diff
@@ -291,18 +291,56 @@ async fn observe_tx_result(
...
+            if let Err(err) = indexer_state
+                .view_client
+                .get_available_foreign_chain_by_node(&indexer_state.mpc_contract_id)
+                .await
```
gilcu3: I am not sure exactly what are we doing here, shouldn't we check that the returning value is something? Else what is the point of having this?

kevindeforth (commented):
intermediate review (need to jump on a call now, but can continue in the afternoon).
Two things I noticed:
- to reduce diff size, I would advice to split this in two PRs: one for contract changes, one for node changes
- I am under the impression that the terminology used for methods and functions does not reflect the one used in the design doc. `available` as reserved as a term to denote chains for which a signing quorum can be formed, but now, we are using it to describe what we defined as "covered". This will lead to confusion. I would suggest to be consistent.

crates/contract/src/lib.rs
```diff
@@ -159,6 +161,11 @@ pub struct MpcContract {
     // TODO(#2937): Remove via state migration.
     metrics: Metrics,
     foreign_chain_rpc_whitelist: ForeignChainRpcWhitelist,
+    available_foreign_chains_by_node: AvailableForeignChainsByNode,
+    /// Cached set of chains that at least the signing threshold of active participants support,
+    /// recomputed on every registration. `get_available_foreign_chains` intersects it with the
+    /// on-chain whitelist to produce the available set, keeping that read cheap.
+    available_foreign_chains: dtos::AvailableForeignChains,
```
gilcu3: might be worth to have this two in a single attribute, and lazily loaded to reduce contract costs in unrelated methods

crates/contract/src/lib.rs
```diff
@@ -191,6 +198,32 @@ impl SupportedForeignChainsByNode {
...
+
+impl AvailableForeignChainsByNode {
+    fn snapshot(&self) -> BTreeMap<dtos::AccountId, dtos::AvailableForeignChains> {
+        self.available_foreign_chain_by_node
+            .iter()
+            .map(|(account_id, available)| (account_id.clone(), available.clone()))
+            .collect()
+    }
+}
+
```
gilcu3: could we add this code in another module? This one is huge already

crates/contract/src/lib.rs
```diff
@@ -969,6 +1002,79 @@ impl MpcContract {
...
+    ) -> Result<(), Error> {
+        let account_id = self.voter_or_panic();
+
+        self.available_foreign_chains_by_node
+            .available_foreign_chain_by_node
+            .insert(account_id, available_foreign_chains);
+        self.recompute_available_foreign_chains();
+
+        Ok(())
+    }
```
gilcu3, line 1003: don't we need some checks here, or are the nodes allowed to submit foreign chains that are not whitelisted?

crates/contract/src/lib.rs
```diff
@@ -969,6 +1002,79 @@ impl MpcContract {
...
+                .collect::<BTreeSet<_>>();
+            (threshold, active)
+        };
+
+        let mut chain_to_supporter_count: BTreeMap<dtos::ForeignChain, u64> = BTreeMap::new();
+        // Count supported chains that are also whitelisted.
+        for (account_id, chains) in self
+            .available_foreign_chains_by_node
+            .available_foreign_chain_by_node
+            .iter()
```
gilcu3: that we need this repetition smells like there is something to improve somewhere

crates/contract/src/lib.rs
```diff
@@ -969,6 +1002,79 @@ impl MpcContract {
...
+                }
+            }
+        }
+
+        self.available_foreign_chains = chain_to_supporter_count
+            .into_iter()
+            .filter(|(_, count)| *count >= threshold)
+            .map(|(chain, _)| chain)
+            .collect::<BTreeSet<dtos::ForeignChain>>()
+            .into();
```
gilcu3: this looks like it could be optimized

gilcu3 (commented):
Had a hard time trying to figure out how the changes fit the design, second's Kevin's take that if this could be split in two, that would help a lot.

I did not find any blocker, but did have quite a few comments and suggestions. I hope to be able to make sense of everything in the next round of reviews

crates/contract/src/lib.rs
anodar: No we changed "supported" to "covered" to avoid tautology "to support supported chain" in wording. That was definition in terminology not in struct names.

As for struct names we defined:
```
* `get_available_foreign_chains() -> AvailableForeignChains` — view method. Returns the chains that ≥ signing threshold active nodes currently cover; `verify_foreign_transaction` gates on this set.
```

So one that's for per node makes sense to have the same name suffixed with ByNode, that is `AvailableForeignChainsByNode`. 


Edit: I see in docs I wrote:
```
* `get_available_foreign_chain_by_node() -> ForeignChainSupportByNode` — view method (formerly `get_foreign_chain_support_by_node`; old name kept as a deprecated wrapper). Returns each participant's registered set of covered chains. Feeds the available-set computation and the coverage alerting (does every active node cover every whitelisted chain?).
```

but that wasn't intended. I would avoid having multiple terms for same thing and use available (and not supported/covered  etc..).

crates/contract/src/lib.rs
anodar: This name is intentional mentioned throughout the doc several times. Same reasoning as above it doesn't introduce another term (coverage) and consolidates them into one "available".

crates/contract/src/snapshots/mpc_contract__tests__mpc_contract_borsh_schema_has_not_changed.snap
anodar: Yes we're creating duplicates for the sake of renaming things, which I initially proposed we'd avoid but argument against that was that we still must rename few things and we might as we rename rest to have clean API. Which makes sense to me.

crates/e2e-tests/tests/common.rs
anodar: I'll address whatever conflicts may arise by merging with main.

crates/e2e-tests/tests/foreign_chain_configuration.rs
anodar: Not sure I follow. We use this construct throughout e2e tests infra quite a lot already if you're referring to:

`( || async {} ).retry( ConstantBuilder... )` 

If you mean to factor it out as a helper where we pass this callback instead, we could do that but as mentioned lot of places use them so would make sense to refactor them separately.

crates/node/src/indexer/tee.rs
anodar: > seems strange that this code lives in crates/node/src/indexer/tee.rs.

I'm not following, this code lived here before, I just factored it out into a method (`is_method_not_found`).

> Still, could we make it so that only the methods that we expect to fail in the next release are logged with info? And add a TODO to remove this after the next release?


I'd rather keep it as error. We should upgrade contract soon after upgrading the node, and I'd rather keep this trace very obvious, easy to find. This is a single error line per node startup so won't be spamming it either.

crates/node/src/indexer/tx_sender.rs
anodar: You really read comments all the way :D 👍

crates/node/src/indexer/tx_sender.rs
anodar: This is the new contract method that's successor of `RegisterForeignChainConfig`.

It is handled same way as  `RegisterForeignChainConfig`, additionally by handling expected method not found error.

`We don't care. The contract state change will handle this.` this comment applies to this as well, once contract is upgraded, handling this error will go away, and this function will replace `RegisterForeignChainConfig` in that section below.

crates/node/src/coordinator.rs
anodar: Done.

crates/node/src/indexer/types.rs
anodar: Done.

crates/node/src/indexer/fake.rs
anodar: Done.

crates/node/src/indexer/tx_sender.rs
anodar: Done.

deny.toml
anodar: Will resolve whatever conflicts with main will be.

crates/contract/src/lib.rs
anodar: This is intentional, we allow nodes to register their config (unfiltered), but available foreign chains are recomputed by filtering out ones that are not allowlisted.

crates/node/src/indexer/tx_sender.rs
anodar: Wanted `RegisterForeignChainConfig` on the top to have comment right above it explaining that will get deprecated. Reverted order.

crates/contract/src/lib.rs
anodar: This follows existing structure https://github.com/near/mpc/blob/1c5a4628fe7fc84805060665a260f67579d38eea/crates/contract/src/lib.rs#L1703

`self.node_foreign_chain_support.foreign_chain_support_by_node` although I agree we can improve this. 

Will merge `available_foreign_chains_by_node` and `available_foreign_chains` as single attribute and get rid of wrapping of first.

crates/node/src/indexer/tee.rs
gilcu3: yeah the comment about code living here was a sidenote. About the log, normally operators check logs right after upgrading, so this would add noise for them. For that reason only is that making it look like all the other possible errors might not be the best

crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
anodar: Done.

crates/contract/src/lib.rs

crates/contract/src/lib.rs

crates/contract/src/lib.rs

crates/contract/src/lib.rs
anodar: Number of chains/rpc providers are pretty small. Not sure we gain much by optimizing anything here.
Do you have some specific concern?

gilcu3: no real performance concern, it just looked not very idiomatic (which in rust tends to be more performant)

anodar: I see. Got rid of .map() using .then_some(), should read a bit nicer now.

crates/contract/src/lib.rs
kevindeforth: Ah wait, but I thought `available` meant: the chain is covered by enough nodes that we can form a signing quorum.
Here, `available` is used as `covered`, no?

anodar: Available in the network : requires consensus
Available in the node: doesn't require consensus, it means it's supported/covered/available by that node.

crates/contract/src/lib.rs
kevindeforth: by definiton, one can't "register available chains". Availability of a chain is established from the configs:

https://github.com/near/mpc/blob/1c5a4628fe7fc84805060665a260f67579d38eea/docs/foreign-chain-transactions.md?plain=1#L310-L317

> This name is intentional mentioned throughout the doc several times. Same reasoning as above it doesn't introduce another term (coverage) and consolidates them into one "available".

According to our definitions, "available" can not be used interchangeably with "covered".

anodar: We have a section [Per-node registration](https://github.com/near/mpc/blob/2ea556589a7a2a6862237d11c0e0663927733719/docs/design/calculating-supported-foreign-chains.md?plain=1#L78) in the doc about that to "register available chains": 

```

Per-node registration (`register_available_foreign_chain_config` /
`get_available_foreign_chain_by_node`) reports which chains each node currently covers,
and serves two roles:
```

Definitions you linked are for explaining the design, not naming proposal.

crates/contract/src/lib.rs
gilcu3: I think you forgot to push this one, I couldn't find the code change

crates/node/src/indexer/tee.rs
kevindeforth: Is this branching still relevant? Feels like we only needed that in anticipation of upgrading the contract code way back last year.
Now, if that method is not found on the contract, I think that's a legitimate error.

crates/node/src/indexer/tx_sender.rs
kevindeforth: Hmm, some thoughts on this:
- is it supposed to be fire and forget? Is there a reason not to inspect the return value and check whether or not the nodes foreign chain config was registered on the contract?
- If the answer above is yes, that we never want to verify, then why do we need to query `get_availalbe_foreign_chains_by_node` here? Why not just return a `Ok(TransactionStatus::Unknown)`?
- Lastly, this seems like something we expect to be changed after the contract update, so a TODO with issue would be warranted.

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
```diff
@@ -979,6 +979,32 @@ pub struct ForeignChainConfiguration(BTreeMap<ForeignChain, NonEmptyBTreeSet<Rpc
...
+    BorshDeserialize,
+    derive_more::From,
+    derive_more::Deref,
+    derive_more::DerefMut,
+)]
+#[cfg_attr(
+    all(feature = "abi", not(target_arch = "wasm32")),
+    derive(schemars::JsonSchema, borsh::BorshSchema)
+)]
+pub struct AvailableForeignChains(BTreeSet<ForeignChain>);
```
kevindeforth: So here, `Available` means implicitly by node?

crates/node/src/coordinator.rs
```diff
@@ -398,7 +409,20 @@ where
...
+            .send(
+                ChainSendTransactionRequest::RegisterAvailableForeignChainConfig(
+                    ChainRegisterAvailableForeignChainConfigArgs {
+                        available_foreign_chains,
+                    },
+                ),
+            )
+            .await
+        {
+            tracing::warn!(error = ?err, "failed to send register_available_foreign_chains_config transaction, this is expected until contract version 3.12");
```
kevindeforth: Hmm, I don't think the `method not available` error gets propagated up here. The only error we would see here is a `TransactionProcessorError`,  which currently only has one variant https://github.com/near/mpc/blob/2095361ebf7d04508f2079618f24b039985626da/crates/node/src/indexer/tx_sender.rs#L101

crates/contract/src/available_foreign_chains.rs
```diff
@@ -0,0 +1,40 @@
...
+
+use near_mpc_contract_interface::types as dtos;
+use near_sdk::near;
+use near_sdk::store::IterableMap;
+
+use crate::storage_keys::StorageKey;
+
+/// Combined lazy state for foreign-chain availability: the cached available set and the
+/// per-node coverage map that feeds it. Stored behind a `Lazy` in `MpcContract` so it is
+/// only deserialized when methods that touch foreign-chain data are called.
```
kevindeforth: I don't understand what "combined lazy state" means. The Iterable map is lazy loaded, but the dtos is a BTreeSet.

Note that this part here:
> /// Stored behind a `Lazy` in `MpcContract` so it is
> /// only deserialized when methods that touch foreign-chain data are called.

risks becoming a stale commit. Someone changing the way this is stored in the `MpcContract` struct will most likely forget to change this comment.

crates/contract/src/lib.rs
```diff
@@ -159,6 +163,10 @@ pub struct MpcContract {
     // TODO(#2937): Remove via state migration.
     metrics: Metrics,
     foreign_chain_rpc_whitelist: ForeignChainRpcWhitelist,
+    /// Lazily-loaded foreign-chain availability state (cached available set + per-node coverage
+    /// map). Only deserialized on methods that touch foreign-chain data, keeping unrelated calls
+    /// cheap.
+    foreign_chain_availability: Lazy<ForeignChainAvailability>,
```
kevindeforth: Does it make sense to load this lazily?
The "big stuff" is behind an iterable map and thus, already lazy loaded.
What's left is a BTreeSet of an enum with probably at best double-digit entries.

crates/contract/src/lib.rs
```diff
@@ -969,6 +977,83 @@ impl MpcContract {
...
     }
 
+    /// (Re)registers the foreign chains this participant currently covers, feeding the available
+    /// set ([`Self::get_available_foreign_chains`]). Idempotent.
+    //
+    // Note: `voter_or_panic` accepts callers in `Initializing` and `Resharing` phases, so a
+    // registration in those phases will be written to `available_foreign_chains_by_node` but
+    // the cache (`available_foreign_chains`) is not refreshed until the next `Running`-state
+    // recompute (e.g. `clean_foreign_chain_data` after `vote_reshared`, or the next call to
+    // this method once the contract is `Running`).
```
kevindeforth: I think this comment would be more useful if it didn't burden the reader with details about `vote_or_panic` and just stated what the impact are for the caller.

crates/contract/src/lib.rs
```diff
@@ -969,6 +977,83 @@ impl MpcContract {
...
+    #[handle_result]
+    pub fn register_available_foreign_chains_config(
+        &mut self,
+        available_foreign_chains: dtos::AvailableForeignChains,
+    ) -> Result<(), Error> {
+        let account_id = self.voter_or_panic();
```
kevindeforth: We may want to allow only the node to call this, not the node operator.

crates/contract/src/available_foreign_chains.rs
```diff
@@ -0,0 +1,40 @@
...
+#[near(serializers=[borsh])]
+#[derive(Debug)]
+pub(crate) struct ForeignChainAvailability {
+    pub(crate) available_foreign_chains: dtos::AvailableForeignChains,
+    pub(crate) available_foreign_chains_by_node:
+        IterableMap<dtos::AccountId, dtos::AvailableForeignChains>,
```
kevindeforth: I think here, we need to store not by account id, but by the TLS key of the node, since a config is per node, not per participant.

kevindeforth (changes requested):
Discussed some of this on slack and we agreed to some renames, most notably:

* `register_foreign_chains_config` instead nof `register_available_foreign_chains_config`
* `get_foreign_chains_configs` instead of `get_available_foreign_chains_by_node`

and associated structs.

Didn't finish the full review, as changes are quite extensive and non trivial. I would still encourage to break this up into two PRs, one for the node, one for the contract.

Will continue tomorrow with review.

Right now, there are two blockers for me:
- naming should align with definition (as discussed on slack)
- We should store node configs not by account id of the operator, but by the node. Otherwise, we may get conflicts when operators migrate their node to new machines.

crates/contract/src/lib.rs
gilcu3: I suggested this because of this [thread](https://nearone.slack.com/archives/C07UW93JVQ8/p1780677458535149) TL;DR our users complaining about tiny increases in gas usage

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
anodar: `AvailableForeignChains` means as it says available foreign chain. If it's used to register node's available chain then it refers that, if you're querying network for available chains then it's that.

I don't see why would we define different objects for the same exact thing: list of chains.

crates/node/src/indexer/tx_sender.rs
anodar: * yes
* it was fire-and-forget previously, I added trivial check to see that new API exists to log warnings just for us. Got rid of it now
* added todo

crates/contract/src/lib.rs
anodar: Looks like I only changed it fake.rs, addressed it here as well now.

crates/node/src/coordinator.rs
anodar: What's the issue with it not being propagated ? We're just logging that we failed to call a method.

crates/contract/src/available_foreign_chains.rs
anodar: "combined lazy state" = available_foreign_chains and available_foreign_chains_by_node states combined behinde Lazy<>.


> risks becoming a stale commit

got rid of the comment. It's already covered where it's defined.

crates/contract/src/available_foreign_chains.rs
anodar: Done.

crates/contract/src/lib.rs

crates/contract/src/lib.rs
anodar: Reworded.

anodar:
@claude  review

anodar:
@claude review please

anodar:
@claude review

crates/contract/src/available_foreign_chains.rs
kevindeforth: > "combined lazy state" = available_foreign_chains and available_foreign_chains_by_node states combined behinde Lazy<>.

But this is not something the struct itself can guarantee. This only applies if we store this structs instance behind a `Lazy`.

> got rid of the comment. It's already covered where it's defined.

Thanks! Yeah, that's definitely the better place for it.

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
kevindeforth: My question arose because of to the comment on this struct: 

> /// The set of foreign chains a node reports it can currently serve (cover). The available-set
> /// analog of [`SupportedForeignChains`]; submitted by nodes via
> /// `register_available_foreign_chains_config`.

(The comment only discusses the use of this struct in the context of a nodes configuguration)

crates/node/src/coordinator.rs
kevindeforth: The issue is that the warning message says "this is expected", but this is not expected.

anodar:
Split this PR in two: https://github.com/near/mpc/pull/3511, https://github.com/near/mpc/pull/3512. Closing this in favor of those two.

Issue #3381, opened 2026-05-27
Auto-remove unused launcher image hashes from the contract


pbeza:
## Background

`allowed_launcher_image_hashes` currently accumulates entries forever — removal requires a unanimous vote — while MPC Docker image hashes auto-expire after 7 days. This asymmetry means stale launcher hashes pile up on the contract over time.

The design for usage-based expiry is approved in [docs/design/auto-remove-launcher-hashes-design.md](https://github.com/near/mpc/blob/main/docs/design/auto-remove-launcher-hashes-design.md) (PR #3488). This issue tracks its **implementation**.

Example testnet state with accumulated entries:

```
$ near contract call-function as-read-only v1.signer-prod.testnet allowed_launcher_image_hashes json-args '{}' network-config testnet now

[
  "e28cb0425db06255fe5fc7aadb79534ac63c94c7a721f75c1af1e934d2eb0701",
  "d07c86cb8c3804a78a961884295f006d465929f0449a3bc10fb9bcbd246e7193",
  "c004f02cb6c75a8203ac772d22fdb3d9bb2e05f2378a0a53a089e78372888c70"
]
```

## User Story

As a node operator, after I rotate to a new launcher and stop using an old one, the old launcher hash should disappear from the contract on its own — without coordinating a unanimous removal vote.

## Acceptance Criteria

- [ ] `AllowedLauncherImage` gains `added` and `last_attested` timestamps; entry is expired when `max(added, last_attested) + TTL < now`
- [ ] `last_attested` refreshed on every successful attestation in `submit_participant_info` (refresh-on-use)
- [ ] All reads of the allowed launcher set skip expired entries; if all are expired, the newest is still returned (fallback so the set never goes empty)
- [ ] New config field `launcher_hash_unused_ttl_seconds` (default 14 days), validated `>= DEFAULT_EXPIRATION_DURATION_SECONDS`
- [ ] `verify_tee` spawns a detached self-call to a new `#[private] clean_expired_launcher_hashes`; gas via new `clean_expired_launcher_hashes_tera_gas` config field; never removes the last entry
- [ ] `vote_add_launcher_hash` for an already-present hash refreshes its `added` timestamp (re-vote recovery — covers the corner case where a launcher is voted in before it is actively used)
- [ ] State migration initializes new fields to migration time for existing entries
- [ ] Tests: refresh-on-use, expiry + newest fallback, never-used expiry, re-vote refresh, sweep, migration

## Resources & Additional Notes

[...]

PR #3488, opened 2026-06-07
docs: design doc for auto-removal of unused launcher image hashes

barakeinav1:
Part of #3381

Design-only PR — proposes usage-based expiry for `allowed_launcher_image_hashes`:

- "Use" signal: a launcher hash is "used" whenever a node's hourly attestation resubmission (`submit_participant_info`) verifies against it — the contract refreshes that entry's `last_attested` timestamp there. Nodes already send these, so no node-side changes
- Expiry: `max(added, last_attested) + TTL < now`, TTL default 14 days (config field, validated `>= DEFAULT_EXPIRATION_DURATION_SECONDS`)
- Enforcement: every read of the allowed list filters out expired entries, so an expired hash is rejected immediately at TTL lapse
- Storage reclaimed by a detached self-call from `verify_tee` to a `#[private]` cleanup method (no new public API, cannot fail the host tx)
- Never-used hashes expire too (e.g. a newly voted launcher image that no node ever migrated to); recovery is a threshold re-vote
- Unanimous `vote_remove_launcher_hash` stays as manual early-removal override; measurements and MPC docker-image expiry out of scope

Remaining open question for reviewers: **TTL length** (14 days vs. 30 — see the Open questions section). Implementation follows in a separate PR.


docs/auto-remove-launcher-hashes-design.md
```diff
@@ -0,0 +1,121 @@
...
+1. **Refresh on use** — after a successful verify in `submit_participant_info`
+   (hourly per node), the contract sets `last_attested = now` on the entry
+   owning the validated compose hash. No node-side changes.
+2. **Read-time filtering** — all reads of the allowed set (verify, re-verify,
+   views) skip expired entries, so an expired hash is rejected the moment its
+   TTL lapses. If *all* entries are expired, the newest is still returned
+   (disaster-recovery fallback, mirrors `AllowedDockerImageHashes`).
+3. **Lazy sweep** — `verify_tee` deletes expired entries from storage
+   (keeping at least the newest). Housekeeping only; expiry is already
+   enforced by (2).
```
pbeza: I’m afraid we could potentially run out of gas if we inline it. Instead, we could introduce a public contract method, similar to this one, that cleans up the expired entries from storage:

https://github.com/near/mpc/blob/dd1695a5e92326a3b667608af73ce3b796ff2379/crates/contract/src/tee/proposal.rs#L195-L200

Alternatively, we could spawn it as a separate self-call via `Promise::new(...).function_call(CLEAN_EXPIRED_LAUNCHER_HASHES, ...).detach()`, similar to what we do here:

https://github.com/near/mpc/blob/cf72a5ab610bf540343e5f0a88a9d01a101ded7e/crates/contract/src/lib.rs#L1088-L1136

pbeza (changes requested):

docs/auto-remove-launcher-hashes-design.md
barakeinav1: I prefer not to add a new public API. We already have a lot.
I don't think that running out of gas is a real concern, since the launcher whitelist is very small (and we control the size via voting). but I'm ok with moving this to a private method and using a Promise if you think it is better.

Let get another reviewer on this.

docs/design/auto-remove-launcher-hashes-design.md
```diff
@@ -0,0 +1,121 @@
...
+
+`allowed_launcher_image_hashes` accumulates entries forever; removal requires a
+**unanimous** vote (`vote_remove_launcher_hash`). MPC Docker image hashes, by
+contrast, auto-expire 7 days after a newer image lands. Testnet carries 3
+launcher hashes with only the newest in use.
+
+The insight: **not using a launcher is itself a vote.** Every node already
+resubmits an attestation hourly (`submit_participant_info`), proving which
+launcher it runs. The contract can observe disuse and evict stale hashes
+without any vote.
```
netrome, line 17: Good point. I wonder if we should do the same for node image hashes. I don't think those should be auto-removed since we may have long periods during upgrades where we want to allow two concurrent node versions.

netrome (commented):
One thought so far, haven't read the full proposal. Will continue later in the afternoon.

docs/design/auto-remove-launcher-hashes-design.md
barakeinav1: If we do want add such a thing for nodes, then we will need to add voting mechanism to remove the nodes (same as we have for the launcher). To prevent an outdated node from staying forever.

netrome: Yeah exactly, which does make sense to me. In practice I'd expect all operators to update in due time and auto-eviction working well.

barakeinav1: We also need to think if this is in addition to or instead of the auto 7 day removall we have.

barakeinav1: I wouldn't change this for now, but we can think of this for the future.

docs/auto-remove-launcher-hashes-design.md
netrome: A private method with a promise seems like a manageable situation. If this starts accumulating and running out of gas it doesn't bring down any other functionality. I'm happy with that solution.

netrome (dismissed):
Looks good to me, as long as we're aligned the implementation of the lazy sweep should be through a promise. Could be worth highlighting in this doc - but we could also view it as an implementation detail. I'm fine with both.

docs/auto-remove-launcher-hashes-design.md
barakeinav1: Updated the doc to go with the detached `Promise` + `#[private]` method (your option b): `verify_tee` spawns a detached self-call to a new `#[private] clean_expired_launcher_hashes`, gas reserved via a `clean_expired_launcher_hashes_tera_gas` config knob like the existing post-resharing cleanups. Skipped the public-method option to avoid growing the public API. Since read-time filtering already enforces expiry, a failed/out-of-gas sweep is harmless and just retries next `verify_tee`.

netrome (approved):

pbeza (approved):

PR #3490, opened 2026-06-08
build: bring back support for NEP-330 contract repro builds

### LINKED ISSUE #3480: Restore cargo-near reproducible contract build now that required sourcescan images are available
### Background

We previously built the MPC contract WASM reproducibly via cargo-near's `[package.metadata.near.reproducible_build]` (NEP-330) using a `sourcescan/cargo-near` Docker image. In #3356 we removed that metadata and switched to a Nix derivation (`nix/mpc-contract.nix`) because no `sourcescan/cargo-near` image existed for the workspace's toolchain (now `rust-1.93.0`, bumped in #3321).

The Nix build is reproducible, but loses NEP-330 build metadata, so third-party verifiers like sourcescan.io and nearblocks can no longer replay and verify the on-chain contract (noted in `docs/reproducible-builds.md`).

The required images have now been released: https://hub.docker.com/r/sourcescan/cargo-near/tags — so we can restore the cargo-near-based reproducible build and regain automated third-party verification. `nix/mpc-contract.nix` stays; cargo-near is added back alongside it to provide the NEP-330 metadata third-party verifiers need.

### Acceptance Criteria

- Re-add `[package.metadata.near.reproducible_build]` to `crates/contract/Cargo.toml`, pinning a `sourcescan/cargo-near` image (with digest) matching `rust-toolchain.toml` (`1.93.0`).
- Keep `nix/mpc-contract.nix`; the cargo-near build coexists with it.
- Contract WASM carries NEP-330 metadata and is verifiable on sourcescan.io / nearblocks.
- `docs/reproducible-builds.md` updated to document both build paths and remove the third-party-verification caveat.

gilcu3:
Closes #3480

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,7 +87,7 @@ bs58 = { version = "0.5.1" }
 built = { version = "0.8", features = ["git2", "chrono"] }
 byteorder = "1.5.0"
 bytes = { version = "1.11.1" }
-cargo-near-build = "0.11.1"
+cargo-near-build = "0.11.3"
 clap = { version = "4.6.1", features = ["derive", "env"] }
 criterion = { version = "0.8.2", features = ["html_reports"] }
 curve25519-dalek = { version = "4.1.3", features = [
@@ -175,17 +175,17 @@ mappings = "0.5"
 mockall = "0.14.0"
 named-lock = "0.4.1"
 near-abi = "0.4.4"
-near-account-id = "2.5.0"
+near-account-id = "2.6.0"
 near-jsonrpc-client = "0.20.0"
 near-jsonrpc-primitives = "0.34.6"
-near-kit = { version = "0.9.1", default-features = false }
+near-kit = { version = "0.11.0", default-features = false }
 near-primitives = "0.34.6"
-near-sandbox = "0.3.9"
-near-sdk = { version = "5.26.1", features = [
+near-sandbox = "0.3.11"
+near-sdk = { version = "5.28.0", features = [
     "legacy",
     "unstable",
 ] }
-near-workspaces = { version = "0.22.1" }
+near-workspaces = { version = "0.22.2" }
 num_enum = { version = "0.7.6", features = ["complex-expressions"] }
 pairing = { version = "0.23.0" }
 pprof = { version = "0.15.0", features = [
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -156,23 +156,39 @@ args = [
   "--locked",
 ]
 
+# Build via cargo-near, the same tool that produces the deployed/reproducible
+# artifact. It sets `--cfg near` (required by near-sdk's host bindings — e.g.
+# `env::panic_str` -> `panic_utf8`) and runs wasm-opt. A plain `cargo build`
+# omits `--cfg near`, which makes contract panics compile to raw `unreachable`
+# traps instead of readable errors. Output: target/near/<crate>/<crate>.wasm.
 [tasks.build-mpc-contract-optimized]
 description = "Build the mpc-contract WASM used by the E2E tests"
 condition = { env_not_set = ["E2E_SKIP_BUILD"] }
-# `--enable-bulk-memory` and `--enable-nontrapping-float-to-int` mirror what
-# `cargo near build` passes to wasm-opt; rustc >= 1.87 emits both feature
-# sets by default, so wasm-opt must accept them to validate the input.
-script = [
-  "cargo build -p mpc-contract --target=wasm32-unknown-unknown --profile=release-contract --locked",
-  "wasm-opt -Oz --enable-bulk-memory --enable-nontrapping-float-to-int target/wasm32-unknown-unknown/release-contract/mpc_contract.wasm -o target/wasm32-unknown-unknown/release-contract/mpc_contract.wasm",
+command = "cargo"
+args = [
+  "near",
+  "build",
+  "non-reproducible-wasm",
+  "--no-abi",
+  "--profile=release-contract",
+  "--manifest-path",
+  "crates/contract/Cargo.toml",
+  "--locked",
 ]
 
 [tasks.build-test-parallel-contract-optimized]
 description = "Build the test-parallel-contract WASM used by the E2E tests"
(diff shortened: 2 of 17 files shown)


Makefile.toml
```diff
@@ -156,23 +156,39 @@ args = [
...
+args = [
+  "near",
+  "build",
+  "non-reproducible-wasm",
+  "--no-abi",
+  "--profile=release-contract",
+  "--manifest-path",
+  "crates/contract/Cargo.toml",
+  "--locked",
 ]
```
gilcu3, line 177: used cargo-near here because I got a compatibility issue with a new feature added there. I think we should trust cargo-near to manage our builds to avoid these issues in the future

docs/reproducible-builds.md
```diff
@@ -48,25 +49,40 @@ The script will output the image hashes and other build information, which can b
...
 ## mpc-contract
 
-The MPC contract WASM is built reproducibly via the Nix derivation at
-[`nix/mpc-contract.nix`](../nix/mpc-contract.nix). The Nix sandbox provides the
-hermetic toolchain (Rust pinned by `rust-toolchain.toml`, clang/LLVM, vendored
-cargo registry), so the build does not depend on a third-party Docker image.
+The MPC contract WASM is built reproducibly via two coexisting paths. They do
+**not** produce byte-identical output: the cargo-near build embeds an extra
+NEP-330 `build_info` metadata section that the Nix build omits, so their hashes
+differ by exactly that section. The released artifact is the cargo-near build.
```
barakeinav1: We now ship two reproducible build paths and document that they "differ by exactly that [`build_info`] section" — but nothing actually verifies that. The `cargo-near` and `nix-build-mpc-contract` jobs run independently and are never compared, so if the two ever diverged beyond the metadata (a real codegen difference, not just the section) nothing would catch it — and the "independent reproducible path" value of keeping Nix is then unverified.

Could we add a CI step that strips the `build_info` section from the cargo-near WASM and asserts the remainder equals the Nix output? That makes the "differ by exactly that section" claim true-by-construction and turns the Nix build into a genuine cross-check. I'd rather not ship an equivalence claim we don't check.

Cargo.toml
```diff
@@ -175,17 +175,17 @@ mappings = "0.5"
...
+near-account-id = "2.6.0"
 near-jsonrpc-client = "0.20.0"
 near-jsonrpc-primitives = "0.34.6"
-near-kit = { version = "0.9.1", default-features = false }
+near-kit = { version = "0.11.0", default-features = false }
 near-primitives = "0.34.6"
-near-sandbox = "0.3.9"
-near-sdk = { version = "5.26.1", features = [
+near-sandbox = "0.3.11"
+near-sdk = { version = "5.28.0", features = [
```
barakeinav1, line 184: nit: this PR also bumps `near-sdk` 5.26.1→5.28.0, `near-kit` 0.9.1→0.11.0, `near-sandbox` 0.3.9→0.3.11, `near-workspaces`, and `near-account-id` (the Cargo.lock is net −821 lines, mostly `near-sandbox`/`near-kit` dropping the embedded wasmtime/cranelift closure). Are all of these required by the cargo-near 0.21.1 / rust-1.93.0 toolchain, or did they come along for the ride? If they're not strictly needed for restoring NEP-330, a separate bump PR would keep this one focused and make a future bisect easier.

barakeinav1 (changes requested):
Nice to have automated third-party verification back. One blocking concern on the two-path equivalence, plus a scope nit on the dependency bumps — both inline.

docs/reproducible-builds.md
gilcu3: The diff is not only the embedded data, although that is certainly a major difference. There is no practical way to make them equal, as the environment in which they are built inherently different. Here we just keep the nix one in case we need it in the future, while using the NEP build for releases

Cargo.toml
gilcu3: I began facing incompatibility issues while working on this, so the simplest solution was to upgrade them all. Maybe some of them are not ultimately needed, but it doesn't hurt to bump them here in any case (except the lock diff, which should be safe)

crates/devnet/README.md
```diff
@@ -272,7 +272,7 @@ To upgrade the contract, first propose the upgrade (suppose we wish to use the n
 contract code for the upgrade):
 
 ```shell
-mpc-devnet mpc $MPC_NETWORK_NAME propose-update-contract --path ../target/wasm32-unknown-unknown/release-contract/mpc_contract.wasm
+mpc-devnet mpc $MPC_NETWORK_NAME propose-update-contract --path ../target/near/mpc_contract/mpc_contract.wasm
```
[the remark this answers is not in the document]
barakeinav1: +1, this is correct. The README (line 88) says the CLI must be run from the `devnet` directory, so `../target/...` resolves to `crates/target` which does not exist — should be `../../target/near/mpc_contract/mpc_contract.wasm`. Note `constants.rs` in this same PR already uses the correct `../../target/...`, so the README is internally inconsistent. The `../` depth is pre-existing, but since this PR touches the line anyway, worth fixing here.

docs/reproducible-builds.md
barakeinav1: Makes sense they can't be made equal — but then this line is wrong: it says they "differ by exactly that [`build_info`] section," which contradicts that. Could you reword to: independently reproducible but **not** bit-identical (different build environments), cargo-near is the released artifact, Nix a fallback not used for releases?

crates/devnet/README.md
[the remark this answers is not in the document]
gilcu3: yeah seems like a pre-existing issue. In any case we would fix it when revamping the devnet crate later, but can fix here as well

Cargo.toml
barakeinav1: Understood — but remember this next time I make such changes! :smile:

gilcu3: > Understood — but remember this next time I make such changes! 😄

I promise I will!

barakeinav1 (approved):
Thanks, doc reads accurately now. 👍

anodar (approved):

PR #3491, opened 2026-06-08
fix(contract): add auth check to `remove_non_participant_update_votes` 

### LINKED ISSUE #3489: `remove_non_participant_update_votes` has stale comment on authorization
### Background

https://github.com/near/mpc/blob/3dfa88ec98d87f5981d592e8d5cdf2d3cf642f83/crates/contract/src/lib.rs#L1792-L1796

Has a comment that only participants can call this method. That is not true, anyone can call the method as no authorization is being done.

See thread where found:
https://github.com/near/mpc/pull/3361/changes#r3361627007

### User Story

_No response_

### Acceptance Criteria

Either update the stale comment or add authorization requirement if needed.

_No response_

DSharifi:
closes #3489

### DIFF
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -1615,7 +1615,7 @@ impl MpcContract {
     }
 
     /// Cleans update votes from non-participants after resharing.
-    /// Can be called by any participant or triggered automatically via promise.
+    /// Can only be called by participants or by the contract itself.
     #[handle_result]
     pub fn remove_non_participant_update_votes(&mut self) -> Result<(), Error> {
         log!(
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -1630,6 +1630,16 @@ impl MpcContract {
             }
         };
 
+        // Authorize the caller: allow self-calls (the cleanup promise spawned after a
+        // successful resharing, where the predecessor is the contract account) and
+        // direct calls from a current participant. Reject everyone else so that
+        // non-participants cannot drive this cleanup.
+        let caller = env::predecessor_account_id();
+        let is_self_call = caller == env::current_account_id();
+        if !is_self_call && !participants.is_participant_given_account_id(&caller) {
+            return Err(InvalidState::NotParticipant { account_id: caller }.into());
+        }
+
         self.proposed_updates
             .remove_non_participant_votes(participants);
         Ok(())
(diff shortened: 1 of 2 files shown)


crates/contract/src/lib.rs
```diff
@@ -1630,6 +1630,16 @@ impl MpcContract {
             }
         };
 
+        // Authorize the caller: allow self-calls (the cleanup promise spawned after a
+        // successful resharing, where the predecessor is the contract account) and
+        // direct calls from a current participant. Reject everyone else so that
+        // non-participants cannot drive this cleanup.
+        let caller = env::predecessor_account_id();
+        let is_self_call = caller == env::current_account_id();
```
anodar, line 1638: optional nit: this seems useful, consider factoring out into helper.

anodar (approved):

crates/contract/src/lib.rs
```diff
@@ -1630,6 +1630,16 @@ impl MpcContract {
             }
         };
 
+        // Authorize the caller: allow self-calls (the cleanup promise spawned after a
+        // successful resharing, where the predecessor is the contract account) and
+        // direct calls from a current participant. Reject everyone else so that
+        // non-participants cannot drive this cleanup.
+        let caller = env::predecessor_account_id();
```
barakeinav1, line 1637: Worth a one-line comment that this authenticates the *predecessor* (not the signer, like the other participant-gated methods via `assert_caller_is_signer`) — it's deliberate so the post-resharing self-call passes, and it's safe since a routed call would have predecessor != participant. Also, since `vote_update` already filters non-participant votes (`update.rs:248`), this guard is hygiene rather than a security fix — both fine, just worth stating.

Related nit: the log a few lines up prints `signer` but the auth decision uses `predecessor`; for the self-call these differ, so logging the predecessor too would help debugging.

crates/contract/src/lib.rs
```diff
@@ -4948,60 +4958,108 @@ mod tests {
...
         let running_state = gen_running_state(NUM_DOMAINS);
         let participants = running_state.parameters.participants().clone();
         let mut contract =
             MpcContract::new_from_protocol_state(ProtocolContractState::Running(running_state));
 
-        // Propose an update with 2 non-participant votes (from propose_and_vote)
+        // propose_and_vote_code adds 2 non-participant votes.
         let update_id_u64 = 0;
-        let test_update = propose_and_vote_code(update_id_u64, &mut contract);
+        let _ = propose_and_vote_code(update_id_u64, &mut contract);
```
barakeinav1, line 4985: nit: the authorized and unauthorized tests share ~8 lines of setup (state + propose_and_vote + participant votes) — could factor into a small helper.

barakeinav1 (approved):
LGTM. A few non-blocking notes inline.

crates/contract/src/lib.rs
```diff
@@ -4948,60 +4958,108 @@ mod tests {
...
 
-        let actual_voters_before: HashSet<_> =
-            contract.proposed_updates.voters().into_iter().collect();
-        assert_eq!(actual_voters_before, expected_voters_before);
+        // When: the authorized caller invokes the cleanup directly.
+        testing_env!(
+            VMContextBuilder::new()
+                .current_account_id(env::current_account_id())
+                .predecessor_account_id(caller.clone())
+                .signer_account_id(caller)
```
barakeinav1, line 5006: nit: in a real self-call the promise carries `signer` = the original `vote_reshared` caller (a participant), not the contract — only `predecessor` is the contract. Setting `signer_account_id` to a participant in the `ContractItself` case would match production and keep the test honest if the auth ever starts consulting the signer.

PR #3497, opened 2026-06-08
refactor: RecentBlocksTracker uses types from chain gateway crate

### LINKED ISSUE #3496: Make recent blocks tracker depend on chain gateway types
### Background

preparation of moving `RecentBlocksTracker` to the chain gateway crate.

### User Story

As a reviewer, I like small scoped PRs

### Acceptance Criteria

Have `RecentBlocksTracker` depend on chain gateway types.

_No response_

kevindeforth:
resolves #3496 
will simplify diff when moving recent blocks tracker to chain gateway

### DIFF
--- a/crates/node/Cargo.toml
+++ b/crates/node/Cargo.toml
@@ -18,6 +18,7 @@ base64 = { workspace = true }
 borsh = { workspace = true }
 bs58 = { workspace = true }
 bytes = { workspace = true }
+chain-gateway = { workspace = true }
 clap = { workspace = true }
 derive_more = { workspace = true }
 ed25519-dalek = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6004,6 +6004,7 @@ dependencies = [
  "bs58 0.5.1",
  "built",
  "bytes",
+ "chain-gateway",
  "clap",
  "derive_more 2.1.1",
  "ed25519-dalek",
--- a/crates/chain-gateway/src/event_subscriber/block_events.rs
+++ b/crates/chain-gateway/src/event_subscriber/block_events.rs
@@ -2,7 +2,7 @@ use derive_more::{Deref, From};
 use near_account_id::AccountId;
 use near_indexer_primitives::CryptoHash;
 
-use crate::types::BlockHeight;
+use crate::types::{BlockEntropy, BlockHeight};
 
 /// The BlockUpdate returned by the Chain indexer.
 /// Similar to [`ChainBlockUpdate`](../../node/src/indexer/handler.rs) in the `mpc-node` crate.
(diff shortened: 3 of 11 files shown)


kevindeforth:
@claude

> - `crates/node/src/requests/queue.rs:605-606` — Now that `BlockHeight::saturating_add` exists, the trailing `+ 1` is still raw `u64` arithmetic and the only `+` left on a primitive height in this file. Per `docs/engineering-standards.md` ("Use safe arithmetic methods"), prefer:
>  ```rust
>  let cutoff_block: BlockHeight = BlockHeight::from(maximum_height)
>      .saturating_sub(REQUEST_EXPIRATION_BLOCKS)
>      .saturating_add(1);
>  ```
>  In practice this can't overflow with `REQUEST_EXPIRATION_BLOCKS = 200`, but using the new helper keeps the file consistent with the rest of the refactor.
  
This is current code and yes, in practice, this can't overflow.
Will address in case a human reviewer complains and states this as a blocker.
  
> - `crates/node/src/requests/queue.rs:528` — `eligible_leaders_and_maximum_height` still returns `(HashSet<ParticipantId>, u64)` and the call site immediately converts to `BlockHeight`. Returning `BlockHeight` directly here (and on `indexer_heights`) would finish the type unification this PR starts; reasonable follow-up.
> - `crates/node/src/requests/debug.rs:18` / `debug.rs:107` — `CompletedRequest.indexed_block_height` still uses `near_indexer_primitives::types::BlockHeight` (a `u64` alias), which is why `queue.rs:666` and `debug.rs:189` round-trip through `.into()`. Worth migrating in the same follow-up to avoid a second `BlockHeight` type drifting.
> - `crates/chain-gateway/src/types.rs:17-27` — The three new helpers don't have direct unit tests. They're trivial wrappers, but the project standard ("Add tests") would suggest a couple of `saturating_*` / `blocks_since` cases (especially the underflow saturation in `blocks_since`) so a future change can't silently break them.
> - `crates/chain-gateway/src/types.rs:29` — `BlockEntropy` derives `Debug`, which prints the raw `[u8; 32]`. Block `random_value` is public chain data, so this isn't a secret leak — flagging only because the same type could later be reused for non-public 32-byte material and the `Debug` impl would then be a problem.

Those are purposefully out of scope to reduce diff size. Most  of these will be resolved automatically when we fully transition to chain-gateway.

gilcu3:
@claude review

I think @kevindeforth forgot to add the correct keyword in the comment above. Please check again given his responses

netrome (approved):
Looks like a no-brainer pure refactor. Thanks! 🙏

crates/node/Cargo.toml
```diff
@@ -18,6 +18,7 @@ base64 = { workspace = true }
 borsh = { workspace = true }
 bs58 = { workspace = true }
 bytes = { workspace = true }
+chain-gateway = { workspace = true }
```
gilcu3, line 21: this is the important change :D

gilcu3 (approved):
Thank you!

PR #3498, opened 2026-06-08
refactor: move recent blocks tracker to chain-gateway crate

### LINKED ISSUE #3523: move recent blocks tracker to chain-gateway crate
### Background

We want to move `RecentBlocksTracker` to the chain-gateway crate

### User Story

As a dev, I like separation of concerns

### Acceptance Criteria

Move `RecentBlocksTracker` to the chain-gateway crate

_No response_

kevindeforth:
resolves #3523


crates/chain-gateway/src/event_subscriber/metrics.rs
```diff
@@ -0,0 +1,17 @@
...
+    prometheus::register_int_counter!(
+        "mpc_blocks_indexed_total",
+        "Number of blocks observed from the indexer and fed to the shared RecentBlocksTracker"
+    )
+    .unwrap()
+});
+
+pub static MPC_FINALIZED_BLOCKS_INDEXED: LazyLock<prometheus::IntCounter> = LazyLock::new(|| {
+    prometheus::register_int_counter!(
+        "mpc_finalized_blocks_indexed_total",
```
kevindeforth, line 13: To consider: We can rename those metrics and prefix with `chain_gateway_*` if desired.

crates/node/src/requests/metrics.rs
```diff
@@ -238,19 +238,3 @@ pub static MPC_CLUSTER_FAILED_SIGNATURES_COUNT: LazyLock<prometheus::IntCounterV
...
         .unwrap()
     });
-
-pub static MPC_BLOCKS_INDEXED: LazyLock<prometheus::IntCounter> = LazyLock::new(|| {
-    prometheus::register_int_counter!(
-        "mpc_blocks_indexed_total",
-        "Number of blocks observed from the indexer and fed to the shared RecentBlocksTracker"
-    )
-    .unwrap()
-});
```
gilcu3, line 248: could you add this metric to the web e2e test? It will be very cheap, and makes sure we don't silently lose it

crates/node/Cargo.toml
```diff
@@ -107,6 +107,7 @@ built = { workspace = true }
 assert_matches = { workspace = true }
 average = { workspace = true }
 blstrs = { workspace = true }
+chain-gateway = { workspace = true, features = ["test-utils"] }
```
gilcu3, line 110: shouldn't the test-utils feature be enabled only for dev dependencies?

ups: that's exactly the case :)

gilcu3 (approved):
Thank you!

crates/node/src/requests/metrics.rs
kevindeforth: Good thought, but is web endpoints e2e test really the place for such a test?
I don't see us testing metics there and I don't think we have a dedicated metrics test for this.

barakeinav1 (approved):
thanks

crates/node/src/requests/metrics.rs
kevindeforth: I think testing metrics makes sense for critical metrics, so I will add it as an acceptance criteria to https://github.com/near/mpc/issues/3229

PR #3504, opened 2026-06-09
feat: add a separate runtime for asset generation

### LINKED ISSUE #3500: Compute assets in a separate tokio runtime
### Background

We have been facing signing degradation right after resharing for a while #1175 . As explained in [signing-starvation-solution.md](https://github.com/near/mpc/blob/dd1695a5e92326a3b667608af73ce3b796ff2379/docs/design/signing-starvation-solution.md), one of the possible solutions in the short term is to have the assets generated using a separate tokio runtime with lower thread priority, so that signing does not degrade in the presence of heavy asset computations.

### User Story

Signing should not degrade at any point

### Acceptance Criteria

Assets are generated in a separate runtime with lower priority.

[thread](https://nearone.slack.com/archives/C07UW93JVQ8/p1780927443564329)

gilcu3:
Closes #3500

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -227,6 +227,7 @@ syn = "2.0"
 tempfile = "3.27.0"
 test-log = "0.2.20"
 thiserror = "2.0.18"
+thread-priority = "1.2.0"
 tikv-jemallocator = { version = "0.5.4", features = [
     "profiling",
     "unprefixed_malloc_on_supported_platforms",
--- a/crates/e2e-tests/src/mpc_node.rs
+++ b/crates/e2e-tests/src/mpc_node.rs
@@ -415,6 +415,7 @@ impl MpcNodeSetup {
                 migration_web_ui: format!("127.0.0.1:{}", self.ports.migration_web_ui).parse()?,
                 pprof_bind_address: format!("127.0.0.1:{}", self.ports.pprof).parse()?,
                 cores: Some(4),
+                separate_asset_generation_runtime: true,
                 indexer: IndexerConfig {
                     validate_genesis: true,
                     concurrency: std::num::NonZeroU16::new(1).unwrap(),
--- a/crates/node/src/metrics/tokio_runtime_metrics.rs
+++ b/crates/node/src/metrics/tokio_runtime_metrics.rs
@@ -1,51 +1,65 @@
-use prometheus::{Counter, IntGauge, register_counter, register_int_gauge};
+use prometheus::{CounterVec, IntGaugeVec, register_counter_vec, register_int_gauge_vec};
 use std::sync::LazyLock;
 use tokio_metrics::RuntimeMonitor;
 
 use crate::metrics::MONITOR_SAMPLE_DURATION;
 
-static TOKIO_RUNTIME_BUSY_DURATION_SECONDS_TOTAL: LazyLock<Counter> = LazyLock::new(|| {
-    register_counter!(
+/// Distinguishes the runtimes that each run their own monitor loop (e.g. the
+/// main MPC runtime vs. the lower-priority asset-generation runtime).
+const RUNTIME_LABEL: &str = "runtime";
+
+static TOKIO_RUNTIME_BUSY_DURATION_SECONDS_TOTAL: LazyLock<CounterVec> = LazyLock::new(|| {
+    register_counter_vec!(
         "tokio_runtime_busy_duration_seconds_total",
-        "Total time worker threads were busy since runtime start (seconds)."
+        "Total time worker threads were busy since runtime start (seconds).",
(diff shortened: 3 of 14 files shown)


crates/e2e-tests/src/mpc_node.rs
```diff
@@ -415,6 +415,7 @@ impl MpcNodeSetup {
                 migration_web_ui: format!("127.0.0.1:{}", self.ports.migration_web_ui).parse()?,
                 pprof_bind_address: format!("127.0.0.1:{}", self.ports.pprof).parse()?,
                 cores: Some(4),
+                separate_asset_generation_runtime: true,
```
gilcu3, line 418: this is just a temporarily added option to be able to evaluate if this change fixes the problem. Once we are 100% sure we can remove it.

crates/node/src/metrics/tokio_runtime_metrics.rs
gilcu3, line 1: changes here are not strictly needed, but it is nice to have the tokio runtime metrics for this new runtime as well

crates/node/src/tests/asset_generation_signing_contention.rs
gilcu3, line 1: a few tweaks were needed here in order to make it as close as possible to our deployments and at the same time to observe the PR results in a single machine

crates/node/src/metrics/tokio_runtime_metrics.rs
```diff
@@ -1,51 +1,65 @@
...
+        "Current tasks in global queue.",
+        &[RUNTIME_LABEL]
     )
     .unwrap()
 });
 
-static TOKIO_RUNTIME_LIVE_TASKS_COUNT: LazyLock<IntGauge> = LazyLock::new(|| {
-    register_int_gauge!(
+static TOKIO_RUNTIME_LIVE_TASKS_COUNT: LazyLock<IntGaugeVec> = LazyLock::new(|| {
+    register_int_gauge_vec!(
```
DSharifi, line 39: Just an FYI; This will break the prometheus dashboards, since the suffix of the metric will change, so keep in mind to update them afterwards.

DSharifi (approved):

crates/node/src/metrics/tokio_runtime_metrics.rs
gilcu3: Yeah I was aware, not sure which of our dashboards are using this metrics atm though. If you know let me know

DSharifi: ~~[Here](https://grafana.nearone.org/d/da7qsfd/tokio-metrics)~~

You're right, these metrics are unused on the tokio metrics dashboard

crates/e2e-tests/src/mpc_node.rs
netrome: Nice that you made it configurable

crates/node/src/metrics/tokio_runtime_metrics.rs
```diff
@@ -1,51 +1,65 @@
...
-        "Current number of alive tasks."
+        "Current number of alive tasks.",
+        &[RUNTIME_LABEL]
     )
     .unwrap()
 });
 
-pub(crate) async fn run_monitor_loop(runtime_monitor: RuntimeMonitor) {
+/// Samples `runtime_monitor` forever, publishing its metrics under the given
+/// `runtime` label so multiple runtimes don't conflate into one series.
```
netrome, line 48: Given the conversation yesterday I'm not sure I find this comment helpful 🤔

Cargo.toml
```diff
@@ -227,6 +227,7 @@ syn = "2.0"
 tempfile = "3.27.0"
 test-log = "0.2.20"
 thiserror = "2.0.18"
+thread-priority = "1.2.0"
```
netrome, line 230: Hmm I'm a bit torn if we should add a third party dependency for this over trying to use lower level primitives ourselves. Though given the small lockfile changes I assume this was already a transitive dependency.

netrome (approved):
Looks good to me, although I hope we can get rid of this dedicated lower priority runtime over using a dedicated synchronous work queue for CPU intensive jobs long term.

Cargo.toml
gilcu3: yep, it was already there :)

crates/node/src/metrics/tokio_runtime_metrics.rs
gilcu3: yeah it is a bit subjective if this one is helpful, although I do remember when I first read this function I could not understand what it was, so certainly could be useful for some people :)

PR #3507, opened 2026-06-09
feat: aptos support for foreign tx validation requests in mpc contract

### LINKED ISSUE #3503: Add foreign chain support for Aptos
### Background

Placeholder issue to add support for the Aptos chain. Feel free to break it down to separate the contract and node changes etc.

### User Story

_No response_

### Acceptance Criteria

We have foreign chain support for Aptos.

_No response_

frolvanya:
## Summary

Adds Aptos as a supported foreign chain for the contract's `verify_foreign_transaction`
flow — the **contract-interface (DTO) half** of Aptos support.

This is one of a two-PR split (see #3508 for the node/inspector side), keeping the
contract-interface changes reviewable independently from the inspector logic.

## What's included

- **New wire types** in `near-mpc-contract-interface`
  (`crates/near-mpc-contract-interface/src/types/foreign_chain.rs`):
  - `AptosTxId` — 32-byte SHA3-256 tx hash (hex), `AptosAddress` — 32-byte account address (hex)
  - `AptosRpcRequest { tx_id, finality, extractors }`
  - `AptosFinality::Committed` — Aptos has BFT instant finality (no confirmations / no reorgs),
    so a single "committed" level
  - `AptosExtractor::Event { event_index } = 1` — discriminant `1`, mirroring the log/event
    extractor on EVM/Starknet/TON; discriminant `0` is left as the conventional block-hash slot
  - `AptosEvent { account_address, sequence_number, type_tag, data }`
  - `AptosExtractedValue::Event(AptosEvent)`
  - variant wiring: `ForeignChainRpcRequest::Aptos`, `ExtractedValue::AptosExtractedValue`,
    `ForeignChain::Aptos`, and the `ForeignChainRpcRequest::chain()` arm
- **No `crates/contract/src` logic changes** — the contract is generic over `ForeignChain`
  (the provider whitelist is keyed by it; there is no per-chain matching), so it picks up Aptos
  transitively through the DTO enums. This matches the TON contract PR, which also touched no
  contract source.
- **Snapshots**:
  - regenerated `crates/contract/tests/snapshots/abi__abi_has_not_changed.snap` — now includes
    the Aptos types and the `ForeignChain__Aptos` variant
  - new `foreign_tx_sign_payload_v1_aptos__should_have_consistent_hash.snap` — pins the V1
    Aptos sign-payload hash
  - the borsh-schema snapshot is intentionally unchanged: adding a `ForeignChain` variant does
    not alter `MpcContract`'s borsh schema (verified, same as TON)
- **Sandbox tests**: `aptos_request()` / `aptos_extracted_values()` fixtures in
  `tests/sandbox/common.rs`, and `#[case::aptos(...)]` added to the existing parameterized
  `verify_foreign_transaction__*` tests in `tests/sandbox/foreign_chain_request.rs`

## Testing

- `near-mpc-contract-interface` unit tests (consistent-hash + chain mapping) pass.
- `test_abi_has_not_changed` passes against the regenerated snapshot.
- Contract WASM size: 1,445,894 bytes, within the 1,500,000-byte limit (no bump needed).

## Out of scope

- The node-side inspector, Aptos RPC client, and node wiring — see #3508.

### DIFF
--- a/crates/near-mpc-contract-interface/src/types/foreign_chain.rs
+++ b/crates/near-mpc-contract-interface/src/types/foreign_chain.rs
@@ -179,6 +179,7 @@ pub enum ForeignChainRpcRequest {
     Polygon(EvmRpcRequest),
     HyperEvm(EvmRpcRequest),
     Ton(TonRpcRequest),
+    Aptos(AptosRpcRequest),
 }
 
 impl ForeignChainRpcRequest {
@@ -195,6 +196,7 @@ impl ForeignChainRpcRequest {
             Self::Polygon(_) => ForeignChain::Polygon,
             Self::HyperEvm(_) => ForeignChain::HyperEvm,
             Self::Ton(_) => ForeignChain::Ton,
+            Self::Aptos(_) => ForeignChain::Aptos,
         }
     }
 }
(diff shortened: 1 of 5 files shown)


crates/near-mpc-contract-interface/src/types/foreign_chain.rs
```diff
@@ -561,6 +563,167 @@ pub enum TonExtractedValue {
...
+    Serialize,
+    Deserialize,
+    BorshSerialize,
+    BorshDeserialize,
+)]
+#[cfg_attr(
+    all(feature = "abi", not(target_arch = "wasm32")),
+    derive(schemars::JsonSchema, borsh::BorshSchema)
+)]
+pub struct AptosEvent {
```
anodar, line 698: Just a remark: This isn't marked as non-exhaustive but I see neither are `EvmLog` and other extracted value enum values from other chains.
If it changes, we can just introduce AptosEventV2 or something like that in AptosExtractedValue enum.

anodar (approved):
Thank you!

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
frolvanya: > I see neither are `EvmLog` and other extracted value enum values from other chains

Yep, I followed the same style as other structs here, so hope it won't be a problem

netrome: This is a struct, not an enum. non-exhaustive is just relevant for public facing enums to ensure we can add variants without breaking downstream code.

netrome (approved):
Looks good to me, thanks!

PR #3508, opened 2026-06-09
feat: aptos foreign-chain inspector and node wiring

### LINKED ISSUE #3503: Add foreign chain support for Aptos

frolvanya:
## Summary

Implements the **node-side** Aptos foreign-chain inspector. Given an Aptos transaction hash, it
fetches the transaction from the configured Aptos RPC providers, verifies it, and extracts the
locker event that the MPC network signs over (for Aptos → NEAR bridging via the omni-bridge
Aptos locker).

## Relationship to #3507 (please read first)

This PR depends on the Aptos DTOs added in #3507. Because it is opened from a fork against
`main`, GitHub won't let me base it on the #3507 branch, so **its diff currently also contains
#3507's commits**. Once #3507 merges, `feat/aptos-inspector` will be rebased onto `main` and this
diff will reduce to inspector-only changes.

The files that belong to *this* PR are:
- `crates/foreign-chain-rpc-interfaces/**`
- `crates/foreign-chain-inspector/**`
- `crates/node/**`, `crates/node-config/**`
- `Cargo.lock`

## What's included

- **`foreign-chain-rpc-interfaces/src/aptos.rs`** — `ReqwestAptosClient` and the
  `TransactionResponse` DTO + `normalize_event_data`. The Aptos node exposes a **REST** API
  (not JSON-RPC), so this uses a `reqwest` GET against
  `/v1/transactions/by_hash/{hash}` rather than the jsonrpsee `ClientT` the EVM/Starknet/Bitcoin
  inspectors share.
- **`foreign-chain-inspector/src/aptos/`** — `AptosInspector` implementing
  `ForeignChainInspector`: fetch by hash → hash-echo check → committed/finality check → extract
  the event at `event_index`.
- **Node wiring** — `build_aptos_fanout`, the `Aptos` arm in `execute_foreign_chain_request`,
  the `aptos` config field, and the contract-interface conversions.

## Key design decisions

- **REST, not JSON-RPC.** Hence a dedicated reqwest client and a new generic
  `ForeignChainInspectionError::RpcRequestFailed(String)` variant — Aptos can't reuse
  `ClientError`, which is typed to a jsonrpsee error.
- **Provider-tolerant response model.** On a vanilla fullnode the Aptos `Transaction` is a
  `type`-tagged union, but providers diverge (e.g. Alchemy returns a flat object with no `type`
  field). The DTO therefore depends only on the fields actually verified — `hash` (required),
  `success`, and `events` — and never on the `type` tag. Committed-vs-pending is inferred from
  the presence of `success` (a pending mempool tx carries no execution result → `NotFinalized`,
  which is transient and retried). This parses a fullnode response, an Alchemy-style flat
  response, and a pending transaction alike, all reducing to the same extracted event so the
  fan-out quorum agrees.
- **Lean top-level error enum.** Aptos errors map onto the shared `ForeignChainInspectionError`
  variants (`InconsistentRpcResponse` for a hash mismatch, `TransactionFailed`,
  `LogIndexOutOfBounds`) plus two generic additions (`TransactionNotFound`, `RpcRequestFailed`),
  consistent with how the other chains report.
- **Determinism across nodes.** Events are taken by `event_index` (the REST API returns them in
  accumulator order, so no sorting is needed), and event `data` JSON is normalized with a
  recursive key sort so every node hashes an identical payload.

## Trust model

Same as the existing inspectors (EVM/Bitcoin/Starknet/TON): trust the on-chain-whitelisted RPC
providers plus the fan-out quorum (all providers must return the same extracted value, else
`InspectorResponseMismatch` blocks signing). It does not perform Aptos light-client / BLS /
Merkle-accumulator proof verification — out of scope, consistent with the other chains.

## Testing

- Inspector unit tests: committed success, transaction failed, pending → `NotFinalized`,
  404 → `TransactionNotFound`, server error → transient `RpcRequestFailed`, event index out of
  bounds, hash mismatch.
- RPC DTO tests: fullnode (tagged) response, Alchemy-style flat response (no `type`), and a
  pending transaction.
- `cargo test`, `cargo clippy -- -D warnings`, `cargo fmt`, and the repo lint suite
  (`assert-matches`, `cargo-sort`, `use-in-fn`) all pass.


crates/foreign-chain-rpc-interfaces/src/aptos.rs
```diff
@@ -0,0 +1,264 @@
...
+///
+/// Since the signed payload is derived solely from `events`, we model only the fields we actually
+/// verify and treat each as optional. Committed-ness is inferred from the presence of `success`
+/// (a pending, still-in-mempool transaction carries no execution result), so we never rely on the
+/// `type` discriminator. This parses a fullnode response, an Alchemy-style flat response, and a
+/// pending transaction alike, all reducing to the same extracted event.
+///
+/// See <https://aptos.dev/en/build/apis/fullnode-rest-api>.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+pub struct TransactionResponse {
```
frolvanya, line 15: There was an issue with this structure, I've validated it against https://fullnode.mainnet.aptoslabs.com/v1/spec#/operations/get_transaction_by_hash and https://www.alchemy.com/docs/chains/aptos/aptos-api-endpoints/aptos-api-endpoints/v-1-transactions-by-hash-txn-hash

But these providers return different stuctures (e.g `type` is missing from alchemy, so it's better to validate this structure against all providers you're planning to use)

crates/node/src/providers/verify_foreign_tx.rs
```diff
@@ -58,6 +61,19 @@ impl ForeignChainInspectors<HttpClient> {
...
+        fn build_aptos_fanout(
+            chain_config: Option<&ForeignChainConfig>,
+        ) -> anyhow::Result<Option<FanOut<AptosInspector<ReqwestAptosClient>>>> {
+            let Some(c) = chain_config else {
+                return Ok(None);
+            };
+            let inspectors = c.providers.try_map_to_vec(|_, p| {
+                let client = ReqwestAptosClient::new(p.rpc_url.clone());
+                Ok::<_, anyhow::Error>(AptosInspector::new(client))
+            })?;
```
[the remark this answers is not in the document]
frolvanya: Fixed in https://github.com/near/mpc/pull/3508/commits/6c591d612061a94b0b1c34a0ad0ce0eb560f4c46

crates/foreign-chain-rpc-interfaces/src/aptos.rs
```diff
@@ -0,0 +1,264 @@
...
+impl AptosRpcClient for ReqwestAptosClient {
+    fn get_transaction_by_hash(
+        &self,
+        tx_hash_hex: &str,
+    ) -> impl Future<Output = Result<TransactionResponse, AptosRpcError>> + Send {
+        let url = format!(
+            "{}/v1/transactions/by_hash/{}",
+            self.base_url.trim_end_matches('/'),
+            tx_hash_hex
+        );
```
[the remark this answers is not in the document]
frolvanya: Fixed in https://github.com/near/mpc/pull/3508/commits/6c591d612061a94b0b1c34a0ad0ce0eb560f4c46

crates/foreign-chain-inspector/src/aptos/inspector.rs
```diff
@@ -0,0 +1,505 @@
...
+                        .parse()
+                        .map_err(|e: std::num::ParseIntError| {
+                            ForeignChainInspectionError::RpcRequestFailed(format!(
+                                "failed to parse event sequence_number: {e}"
+                            ))
+                        })?;
+
+                // Verbatim: providers are assumed to agree on the struct-tag's address form
+                // (fullnodes return canonical short form). Canonicalization deferred to BCS.
+                let type_tag = event.event_type.clone();
```
anodar: We should normalize `tag_type`, otherwise if some RPC providers return short form for address in there, and others canonical long form, there won't be a quorum.

crates/foreign-chain-rpc-interfaces/src/aptos.rs
anodar: The link from alchemy is using docs-demo endpoint that seems to not quite follow the specification. 

If you try alchemy endpoint with our API key it returns identical response (at least few transactions I've tested) e.g.:

```
curl --request GET \
  --url https://fullnode.mainnet.aptoslabs.com/v1/transactions/by_hash/0xadc6b85a0931fc7f0d7e3839b52d63105e22cec1cb1cdee48aa2065773098c3c \
  --header 'Accept: application/json, application/x-bcs' | jq .
```



```
curl --request GET  --url https://aptos-mainnet.g.alchemy.com/v2/$API_KEY/v1/transactions/by_hash/0xadc6b85a0931fc7f0d7e3839b52d63105e22cec1cb1cdee48aa2065773098c3c | jq .
```

To me it seems like https://fullnode.mainnet.aptoslabs.com/v1/spec is specification of the API, not just a documentation of one (their) implementation of node, so should I'd expect it to be consistent across providers.

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -253,7 +258,10 @@ impl ForeignChainInspectionError {
     pub fn is_transient(&self) -> bool {
         matches!(
             self,
-            Self::ClientError(_) | Self::NotFinalized | Self::NotEnoughBlockConfirmations { .. }
+            Self::ClientError(_)
+                | Self::RpcRequestFailed(_)
```
anodar, line 270: RpcRequestFailed is returned any error that's not 404. But that also includes 400, 403, 410 that we shouldn't retry either. 

Possible codes in responses according to the specification: https://fullnode.mainnet.aptoslabs.com/v1/spec#/operations/get_transaction

crates/foreign-chain-inspector/src/aptos/inspector.rs
```diff
@@ -0,0 +1,505 @@
...
+}
+
+impl<Client> AptosInspector<Client> {
+    pub fn new(client: Client) -> Self {
+        Self { client }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum AptosFinality {
```
anodar, line 22: We should probably mark this [non-exhaustive]. I understand Aptos doesn't have number of confirmations needed to be considered finalized, once it's commited that's it, but maybe we'll add Pending (when transaction is in mempool ?!). 

Regardless, it would be a good practice so that if one uses this code as a guide to implement it for different chain it will be more complete.

crates/foreign-chain-inspector/src/aptos/inspector.rs
```diff
@@ -0,0 +1,505 @@
...
+                }
+                other => ForeignChainInspectionError::RpcRequestFailed(other.to_string()),
+            })?;
+
+        ensure_hash_matches(&tx_hash_hex, &tx.hash)?;
+
+        match finality {
+            AptosFinality::Committed => {
+                // Committed iff the tx carries an execution result; a pending tx does not.
+                let Some(success) = tx.success else {
```
anodar, line 86: This doesn't look right. Status 200 also can return body with type: `pending_transaction` according to the specs: https://fullnode.mainnet.aptoslabs.com/v1/spec#/operations/get_transactions

Image

crates/foreign-chain-inspector/src/aptos/inspector.rs
```diff
@@ -0,0 +1,505 @@
...
+            requested_hash: hex_str_to_bytes(requested),
+            returned_hash: hex_str_to_bytes(returned),
+        });
+    }
+    Ok(())
+}
+
+/// Best-effort hex decode for error messages; empty on non-hex input.
+fn hex_str_to_bytes(s: &str) -> HexBytes {
+    HexBytes(hex::decode(s.strip_prefix("0x").unwrap_or(s)).unwrap_or_default())
```
anodar: Let's not panic here, instead propagate error.

crates/foreign-chain-rpc-interfaces/src/aptos.rs
```diff
@@ -0,0 +1,317 @@
...
+use std::time::Duration;
+
+/// Response from `GET /v1/transactions/by_hash/{txn_hash}`.
+///
+/// Modelled leniently and *without* the `type` discriminator: providers diverge on the envelope
+/// (a fullnode tags it and includes every field; gateways like Alchemy omit `type` and make
+/// fields optional), so we keep only the fields we verify and infer committed-ness from the
+/// presence of `success` — a pending mempool tx has no execution result.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+pub struct TransactionResponse {
```
anodar, line 15: Looks like there's plenty types of transactions: https://github.com/aptos-labs/aptos-core/blob/main/api/types/src/transaction.rs#L208

We can add string `transaction_type` and do something like:

```
match {
   "pending_transaction" => return Err(ForeignChainInspectionError::NotFinalized),
   "user_transaction" | "whatever_else_makes_sense" => {}
   _ => return Err(ForeignChainInspectionError::InspectorResponseMismatch ..)
}
```

crates/foreign-chain-rpc-interfaces/src/aptos.rs
```diff
@@ -0,0 +1,317 @@
...
+            let parsed = response.json::<TransactionResponse>().await?;
+            Ok(parsed)
+        }
+    }
+}
+
+/// Canonical string form of the event `data`: object keys sorted recursively so all nodes hash
+/// identically regardless of provider key ordering.
+///
+/// TODO: migrate to BCS-encoded responses (via Accept: application/x-bcs) to drop this dependency.
```
anodar: Not very familiar with aptos or their binary encoding but with if we were to actually migrate this to BCS we'd either have to pull in their crate as dependency (not sure we want to introduce that), or duplicate their structs. Neither choice seems plausible, so consider dropping this TODO.

crates/foreign-chain-rpc-interfaces/src/aptos.rs
```diff
@@ -0,0 +1,317 @@
...
+pub fn normalize_event_data(value: &serde_json::Value) -> String {
```
anodar, line 130: I don't think we need to normalize this, serde_json seems to be already sorted by keys:

https://docs.rs/serde_json/latest/serde_json/enum.Value.html: `By default the map is backed by a BTreeMap. `

crates/node/src/providers/verify_foreign_tx.rs
```diff
@@ -58,6 +62,29 @@ impl ForeignChainInspectors<HttpClient> {
...
+                    RpcAuthentication::CustomHeader {
+                        header_name,
+                        header_value,
+                    } => Some((header_name, header_value)),
+                };
+                let client = ReqwestAptosClient::new(url, auth_header, timeout);
+                Ok::<_, anyhow::Error>(AptosInspector::new(client))
+            })?;
+            Ok(Some(FanOut::new(inspectors)))
+        }
```
anodar: Why not use generic build_fanout<> instead like rest of the chains?

crates/node-config/src/foreign_chains.rs
```diff
@@ -619,4 +622,59 @@ foreign_chains:
...
+    max_retries: 3
+    providers:
+      nodereal:
+        rpc_url: "https://aptos-mainnet.nodereal.io/v1/"
+        auth:
+          kind: none
+"#;
+
+        // When
+        let config: ConfigFile =
```
anodar, line 671: Please add rpc config with auth as well: one that uses url path for api key, and one that's in header and verify those are correctly propagated.


Something like:

```
 providers:
    alchemy:
      rpc_url: "https://apt-mainnet.g.alchemy.com/v2"
      auth:
        kind: path
        key: "my-api-key"
```

and

```
  providers:
    nownodes:
      rpc_url: "https://aptos.nownodes.io/v1"
      auth:
        kind: header
        header_name: "api-key"
        key: "my-api-key"
```

anodar (commented):
Overall looks good to me! There are few things to address before merging.

crates/foreign-chain-rpc-interfaces/src/aptos.rs
frolvanya: That would be true, if we'd really used `serde_json` without any extra feature flags, but near-sdk-rs enables `preserve_order` flag [which preserves the insertion order](https://github.com/serde-rs/json/blob/a1ae73ac6a6940a4a57c673aebaa13ed4dfe3e8c/Cargo.toml#L62-L65): https://github.com/near/near-sdk-rs/blob/de270eb4a40cc1a6d345270cd6f5695adc908d9e/near-sdk/Cargo.toml#L24

You can verify that mpc repo uses `serde_json` with this flag using:
```bash
$ cargo tree -p mpc-node -i serde_json -e features | grep -B1 -A4 'preserve_order'
```

crates/foreign-chain-inspector/src/lib.rs
frolvanya: I might be wrong here, so please correct me. From what I see `is_transient` is not about retries, but rather showing if node's verdict is included in quorum or not:
https://github.com/near/mpc/blob/76bddd62a655f5c475cdd13377a8f2f4a06d79a3/crates/foreign-chain-inspector/src/lib.rs#L98-L105

So let's say we have 3 nodes and we must have 3/3 consensus to verify transaction and first node might have an expired API key. Then rest api will return 403 status code:
- with transient error (current implementation), 2nd and 3rd node will agree and verification works
- with a new proposed option to make `RpcRequestFailed` non transient, we'd block all nodes

Which option is correct here?

crates/foreign-chain-rpc-interfaces/src/aptos.rs
frolvanya: I think guarding by `user_transaction` doesn't make sense, because any user can deploy a Move module that emits arbitrary events inside a `user_transaction`. So I think that rejecting `pending_transaction` alone should be enough. Added filter in 6d24833f961b6701888f9ff4ae1add8b6a64dfef

crates/foreign-chain-inspector/src/aptos/inspector.rs
frolvanya: Rejected `pending_transaction` as `NotFinalized` in 6d24833f961b6701888f9ff4ae1add8b6a64dfef

crates/foreign-chain-rpc-interfaces/src/aptos.rs
frolvanya: My bad, should've verified against real node, instead of relying solely on docs. Looks like `type` indeed exists. Added it back to structure in 6d24833f961b6701888f9ff4ae1add8b6a64dfef

crates/foreign-chain-inspector/src/aptos/inspector.rs
frolvanya: Added normalization in 9ca8fc536f007e8d280f4fca9b67620ee0c8b33e

crates/node-config/src/foreign_chains.rs
frolvanya: Covered it with a test in d04dad4dfcefa0842da9aa6c1461c799f680e366

crates/node/src/providers/verify_foreign_tx.rs
frolvanya: Refactored in 218aaae602f8f434183a0ee943f78a205b2acd13
But honestly this generalization resulted in more code then before, so not sure if it's worth it or maybe there's a better way to generalize `jsonrpsee` and aptos inspectors here

crates/foreign-chain-rpc-interfaces/src/aptos.rs
frolvanya: Dropped TODO in 8285e3b046a9467f27e33a3a5969f99cf0657c02

crates/foreign-chain-rpc-interfaces/src/aptos.rs
frolvanya: Added a comment justifying sorting keys in 8285e3b046a9467f27e33a3a5969f99cf0657c02

crates/foreign-chain-inspector/src/aptos/inspector.rs
frolvanya: `unwrap_or_default` couldn't panic, but there was a chance of empty response. Changed hash matching logic and added error propagation in 162a71702f524804cf54eb2509ab9a1e94f55803

crates/foreign-chain-inspector/src/aptos/inspector.rs
frolvanya: Added `[non-exhaustive]` in 162a71702f524804cf54eb2509ab9a1e94f55803

crates/foreign-chain-inspector/src/aptos/inspector.rs
anodar: missed `_or_default` part, my bad.

crates/foreign-chain-rpc-interfaces/src/aptos.rs
anodar: I wasn't suggesting `user_transaction` alone, but 3 categories: 

- "pending", 
- "ones we consider finalized" ,  (that's what I meant by `"user_transaction" | "whatever_else_makes_sense" => {}`)
- "the rest".  

Basically explicitly listing 2nd category so that if aptos starts returning some new variant, we'll catch it in the 3rd arm (although it shouldn't break extractor itself anyway since that would be breaking API change from aptos api spec).

The current approach of `pending` -> NotFinalized and `everything else` -> OK, sounds good to me.

crates/foreign-chain-rpc-interfaces/src/aptos.rs
anodar: Good catch, thanks for clarification!

crates/foreign-chain-inspector/src/lib.rs
anodar: > From what I see is_transient is not about retries, but rather showing if node's verdict is included in quorum or not

That's correct. So in the current implementation any error that's not 404 will be considered (RpcRequestFailed ) transient and even if 1/3 RPC gets success and rest 2/3 get 400/403 or 410 or even timeout (which seems like a great attack vector) transaction will be validated (as long as error codes of transient errors also match).


> So let's say we have 3 nodes and we must have 3/3 consensus to verify transaction 

I assume here you mean single node 3 RPC's, and that RPC responses agreeing, not nodes.

> with a new proposed option to make RpcRequestFailed non transient, we'd block all nodes

I'm not proposing to make RpcRequestFailed non-transient, I'm proposing to restrict transient errors to only 500 and 503 (currently that includes: 400,403,410, 500, 503. 
So your specific example with expired API key: I'd say yes we should certainly not validate requests. We have work in progress that sets quorum (minimum number of RPC's inside node must agree so that node considers validation) of RPC's, and if we can't have that number of RPC's agreeing doesn't matter if it's because of the API key or otherwise we shouldn't validate unless we'll have quorum reached.

Moreover, while we don't currently retry transient errors, we should retry those (@DSharifi correct me if I'm wrong). Not in this pr of course.

crates/node/src/providers/verify_foreign_tx.rs
anodar: Looks good to me. Thank you!

anodar (dismissed):
Thanks for the update!

crates/foreign-chain-inspector/src/aptos/inspector.rs
DSharifi: This is not a crate we publish or have others depend on, so for the future I don't don't think there's any need to make enums here non exhaustive.

crates/foreign-chain-inspector/src/aptos/inspector.rs
```diff
@@ -0,0 +1,671 @@
...
+            .client
+            .get_transaction_by_hash(&tx_hash_hex)
+            .await
+            .map_err(|e| match e {
+                // 404 = definitively absent → a non-transient verdict, not a retry.
+                AptosRpcError::ApiError { status: 404, .. } => {
+                    ForeignChainInspectionError::TransactionNotFound
+                }
+                other => ForeignChainInspectionError::RpcRequestFailed(other.to_string()),
+            })?;
```
DSharifi: All server errors, `5xx` codes, are also transient.

crates/foreign-chain-inspector/src/aptos/inspector.rs
```diff
@@ -0,0 +1,671 @@
...
+        tx: &TransactionResponse,
+    ) -> Result<AptosExtractedValue, ForeignChainInspectionError> {
+        match self {
+            AptosExtractor::Event { event_index } => {
+                let event = tx
+                    .events
+                    .get(*event_index)
+                    .ok_or(ForeignChainInspectionError::LogIndexOutOfBounds)?;
+
+                // An unparseable field → transient, so one malformed provider drops from the
```
DSharifi: How are unparsable fields transient?

crates/foreign-chain-inspector/src/aptos/inspector.rs
```diff
@@ -0,0 +1,671 @@
...
+/// immediately followed by `::`; everything else (identifiers, primitive type args) is copied
+/// verbatim.
+fn normalize_type_tag(tag: &str) -> String {
+    let bytes = tag.as_bytes();
+    let mut out = String::with_capacity(tag.len());
+    let mut segment_start = 0;
+    let mut i = 0;
+    while i < bytes.len() {
+        let at_address_position = i == 0 || matches!(bytes[i - 1], b'<' | b',' | b' ');
+        if at_address_position && bytes[i] == b'0' && bytes.get(i + 1) == Some(&b'x') {
```
DSharifi: I find it quite hard to understand this function.

Can you add example inputs and outputs to the documentation so it's easier to follow what is happening here?

crates/foreign-chain-inspector/src/aptos/inspector.rs
```diff
@@ -0,0 +1,671 @@
...
+                }
+                segment_start = hex_end;
+                i = hex_end;
+                continue;
+            }
+        }
+        i += 1;
+    }
+    out.push_str(&tag[segment_start..]);
+    out
```
DSharifi: Let's not do direct inedxing. This can cause a panic on the mpc node, which can abort the whole process.

crates/foreign-chain-inspector/src/lib.rs
DSharifi: > Moreover, while we don't currently retry transient errors, we should retry those (@DSharifi correct me if I'm wrong). Not in this pr of course.


Yes, for sure

DSharifi: Shouldn't we consider all 5xx, server errors, as transient?

DSharifi (changes requested):
Looks quite good, and the test coverage is also good.

I have a couple of comments, but one hard blocker;
Let's not do direct indexing in the function `normalize_type_tag`. That function is not a trivial one, and it can be easy to shoot ourselves in the foot by indexing out of bounds causing a production panic.

crates/foreign-chain-inspector/src/lib.rs
frolvanya: Changed in d3009bb54e2d1aa0ba9f9e7ae363adf89b46f31b

crates/foreign-chain-inspector/src/aptos/inspector.rs
frolvanya: Done in d3009bb54e2d1aa0ba9f9e7ae363adf89b46f31b + 408/429 are transient now

crates/foreign-chain-inspector/src/aptos/inspector.rs
frolvanya: Changed to `MalformedRpcResponse` which is non-transient in d3009bb54e2d1aa0ba9f9e7ae363adf89b46f31b

crates/foreign-chain-inspector/src/aptos/inspector.rs
frolvanya: Refactored in 4b28f9222d6d20415e6891719d2e246c2a425da5

crates/foreign-chain-inspector/src/aptos/inspector.rs
frolvanya: Refactored and added examples in doc-comment in 4b28f9222d6d20415e6891719d2e246c2a425da5

DSharifi (approved):

anodar (approved):

netrome (commented):
Looks good to me, thanks!

Issue #1087, opened 2025-09-11
Figure out a good way to initialize the contract with attestations enforced for hard launch


DSharifi:
### Background

In #1067 we start enforcing TEE attestations for all participants to be in a running state. If any node is running without a valid TEE attestation it will be removed when [`verify_tee()`](https://github.com/near/mpc/blob/97cc792e028cd9bc49f0193c1b01f0bf3e84c314/libs/chain-signatures/contract/src/lib.rs#L1147) method of the contract is called.

A problem with the current implementation is that when the contract is initialized it transitions directly into the `Running` state. Since it's an initialization of the contract, by definition no node can have submitted their attestation yet.

On startup nodes will trigger the `verify_tee` method, where the contract detects that less than threshold nodes have valid attestation, and the contract will stall. There can also be a race condition where some nodes manage to submit their attestation before `verify_tee` is invoked. Either way, we need a robust way to initialize the contract in such a way that nodes are not getting kicked out right after an initialization.

A temporary solution has been to give all nodes a valid temporary attestation in `init()`:
https://github.com/near/mpc/blob/e870f199d513b98a55d04f7e000a936c7b57db88/libs/chain-signatures/contract/src/lib.rs#L158-L171

This solution is not viable post hard launch, as the contract won't accept mocked attestations anymore.

### Acceptance Criteria

The contract must be initializable for the hard launch without room for race conditions where nodes are being kicked out for not having submitted an attestation in time before `verify_tee()` is invoked.

Alternative solutions (non-exhaustive):
1. The contract should not transition directly into the `running` state. We could introduce a new state, for example`WaitingForAttestations`, that the contract goes into during initialization which all it does is to wait for all participants to submit their attestations. Once all attestations have been submitted the contract transitions into running.
2. The `init` function of the contract expects a valid attestation for every single participant of the contract.

I am wary of option 2. as it requires more manual work, and copy pasting multiple large attestations generated inside TDX enabled VMs  into a single CLI.

Option 1 will mean we can keep things automated to initialize the contract, and leave operators and humans with CLIs out of the loop to start the network.

netrome:
> In either case (if we keep the boolean or introduce an additional protocol state), we probably want to change the behavior of verify_tee. It should swap accept_requests from false to true (or exit WaitingForAttestations), only when all participants have a valid attestation.
This should not create any additional issues, as the network can always vote to enter a resharing and force some participants out.

Yeah this sounds like a good short-term solution.

[...]

PR #3511, opened 2026-06-09
feat: available foreign chains contract API

### LINKED ISSUE #3568: Implement new API in the contract for supporting new way of calculating available foreign chains
### Background

### User Story

_No response_

### Acceptance Criteria

Changes required in contract side for implementing design https://github.com/near/mpc/issues/3474 is implemented.

_No response_

anodar:
Contract side changes from: https://github.com/near/mpc/pull/3484

- Add `ForeignChainsConfig` struct for per-node reported chains
- Add `ForeignChainsMetadata` (lazy-loaded struct consolidating RPC whitelist, per-node configs, and cached available-chain set)
- Rename `register_available_foreign_chains_config` → `register_foreign_chains_config`
- Rename `get_available_foreign_chains_by_node` → `get_foreign_chains_configs`
- Key per-node configs by Ed25519 TLS public key; stale entries removed on key rotation
- Restrict `register_foreign_chains_config` to attested participants only
- Migrate `foreign_chain_rpc_whitelist` into `ForeignChainsMetadata`
- Update ABI and Borsh schema snapshots

Implementation of the design: https://github.com/near/mpc/issues/3434

closes: https://github.com/near/mpc/issues/3568


gilcu3:
@claude review

crates/contract/src/state.rs
```diff
@@ -238,6 +238,29 @@ impl ProtocolContractState {
...
+        let participants = match self {
+            ProtocolContractState::Initializing(state) => {
+                state.generating_key.proposed_parameters().participants()
+            }
+            ProtocolContractState::Running(state) => state.parameters.participants(),
+            ProtocolContractState::Resharing(state) => {
+                state.resharing_key.proposed_parameters().participants()
+            }
+            ProtocolContractState::NotInitialized => return None,
+        };
```
gilcu3: I am not totally sure about this. Is this method meant to be used in most contract states?

crates/contract/src/foreign_chains_metadata.rs
```diff
@@ -0,0 +1,137 @@
...
+        self.tls_key_by_account.remove(account_id);
+    }
+
+    pub(crate) fn snapshot_by_node(&self) -> dtos::ForeignChainsConfigs {
+        self.foreign_chains_configs
+            .iter()
+            .map(|(id, chains)| (id.clone(), chains.clone()))
+            .collect::<std::collections::BTreeMap<_, _>>()
+            .into()
+    }
```
gilcu3, line 89: a doc comment here would help understand what this is expected to return

crates/contract/src/foreign_chains_metadata.rs
```diff
@@ -0,0 +1,137 @@
...
+#[near(serializers=[borsh])]
+#[derive(Debug)]
+pub(crate) struct ForeignChainsMetadata {
+    pub(crate) rpc_whitelist: ForeignChainRpcWhitelist,
+    pub(crate) available_foreign_chains: dtos::AvailableForeignChains,
+    pub(crate) foreign_chains_configs:
+        IterableMap<dtos::Ed25519PublicKey, dtos::ForeignChainsConfig>,
+    /// Reverse map from account ID to its most-recently registered TLS key, used to remove
+    /// stale `foreign_chains_configs` entries when a node rotates its TLS key.
+    tls_key_by_account: LookupMap<dtos::AccountId, dtos::Ed25519PublicKey>,
```
gilcu3: any reason for not allowing access this attribute, but yes for the others?

crates/contract/src/foreign_chains_metadata.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+impl ForeignChainsMetadata {
+    /// Creates a new instance carrying over an existing `rpc_whitelist`; all other fields default.
+    pub(crate) fn with_rpc_whitelist(rpc_whitelist: ForeignChainRpcWhitelist) -> Self {
+        Self {
+            rpc_whitelist,
+            ..Default::default()
+        }
+    }
```
gilcu3: I see this method is only used in migration. IMO it should be defined there instead. Constructors for lazy structs from near-sdk are dangerous because these structs are stored in the trie and do not behave like normal rust data, for example when you overwrite one the old data is not cleaned up automatically.

gilcu3 (dismissed):
Thank you, LGTM

Is this solution fully aligned with the design, or do we need a minor update there?

crates/contract/src/foreign_chains_metadata.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    /// Registers `foreign_chains_config` for the given account, keyed by `tls_key`. If the
+    /// account previously registered with a different TLS key, the stale entry is removed first.
+    pub(crate) fn register(
```
kevindeforth: The point of storing foreign chain configs by TLS key is such that multiple nodes belonging to the same node operator can simultaneously register their foregin chain configuration.
This is to support a smooth migration of MPC nodes to new machines.

Here, the comment indicates that we only allow one entry per node operator at a time, which means the on-chain data may be inaccurate during migration and also, makes indexing by TLS key pointless.

crates/contract/src/foreign_chains_metadata.rs
```diff
@@ -0,0 +1,137 @@
+use near_mpc_contract_interface::types as dtos;
+use near_sdk::near;
+use near_sdk::store::{IterableMap, LookupMap};
+
+use crate::foreign_chain_rpc::ForeignChainRpcWhitelist;
+use crate::storage_keys::StorageKey;
+
+/// All foreign-chain state: the RPC provider whitelist, the per-node config reports, and the
+/// cached available-chain set derived from them. Stored behind `Lazy<>` in `MpcContract` so it
+/// is only deserialized when foreign-chain methods are called.
```
kevindeforth: This is another variant of [this comment](https://github.com/near/mpc/pull/3484#discussion_r3381759807). I encourage to be more critical of LLM generated comments. Good quality comments can be helpful. Low quality comments are a review and maintenance burden.

crates/contract/src/lib.rs
```diff
@@ -969,6 +975,65 @@ impl MpcContract {
...
+    pub fn register_foreign_chains_config(
+        &mut self,
+        foreign_chains_config: dtos::ForeignChainsConfig,
+    ) -> Result<(), Error> {
+        self.assert_caller_is_attested_participant_and_protocol_active();
+        let account_id = env::signer_account_id();
+        let tls_key = self
+            .protocol_state
+            .tls_key_for_account(&account_id)
+            .unwrap_or_else(|| env::panic_str("caller has no TLS key in participants list"));
```
kevindeforth: We should look up the TLS key using the attestation. With the current logic, we allow other access keys of to impersonate the currently active node.

https://github.com/near/mpc/blob/c697c119993b2fa3f65113125dfe22cc176d3cc0/crates/contract/src/tee/tee_state.rs#L449

kevindeforth (changes requested):
Unfortunately, I need to request changes to this PR.

**Access keys can impersonate currently active node:**
While we index the foreign chain configurations by node TLS keys, this PR only allows one entry per node operator. This defeats the purpose of indexing by TLS key and makes it impossible to have multiple nodes from the same operator submit their configs simultaneously (which is required during node migration).

Additionally, there is no guarantee that the node in possession of the TLS key is actually the one submitting the config. A node operator can override the configuration submitted by their node. A node that is about to onboard, or a node that has just been off-boarded, could still override the currently actively participating node. Essentially, any access key of the node operator account can override the registered foreign chain configuration of the currently active node.

We can take a look at [`is_caller_an_attested_participant`](https://github.com/near/mpc/blob/c697c119993b2fa3f65113125dfe22cc176d3cc0/crates/contract/src/tee/tee_state.rs#L449) for inspiration on how to map a caller of the smart contract to the nodes TLS key.

_edit: @anodar feel free to block some time in my calendar to discuss how we leverage TEE attestations to create a mapping from signer key to TLS key. I understand this is the first time you touch on the attestation logic in the contract,._

**Code comments:**
This PR makes extensive use of code comments. While I understand these are most likely are LLM-generated and thus low-cost for the PR author, they are a maintenance burden and require review effort. Given the quantity and lack of quality of many of these comments, it is currently the other blocker for this PR.

crates/contract/src/state.rs
anodar: Not applicable anymore got rid of this function. New equivalent `lookup_node_id_by_signer_pk` doesn't concern with states.

crates/contract/src/foreign_chains_metadata.rs
anodar: Addressed in [eef28b0](https://github.com/near/mpc/pull/3511/commits/eef28b02f0a6950ded040c1c51ac64711cf5c5e6).

crates/contract/src/foreign_chains_metadata.rs
anodar: Other fields are used in the same create in lib.rs, but this one is only used here in foreign_chains_metadata.rs

crates/contract/src/foreign_chains_metadata.rs
anodar: Good to know! Move it to v3_11_2_state.rs and initializing with `ForeignChainsMetadata::default()`, that should behave as expected right ?

crates/contract/src/foreign_chains_metadata.rs
anodar: As discussed offline this was intentional to allow only active nodes to register foreign chain config, hence why I used mapping of account -> config before (rather than tls key -> config).

I think now we should be aligned as per [eef28b0](https://github.com/near/mpc/pull/3511/commits/eef28b02f0a6950ded040c1c51ac64711cf5c5e6)

crates/contract/src/foreign_chains_metadata.rs
anodar: I overlooked this one, will pay attention to the comments, addressed now.

crates/contract/src/lib.rs
anodar: This was meant to get active node's tls key as intended as my implementation intended to allow single config per node operator (not per node).

Should be addressed now in per [eef28b0](https://github.com/near/mpc/pull/3511/commits/eef28b02f0a6950ded040c1c51ac64711cf5c5e6).

crates/contract/src/lib.rs
```diff
@@ -1677,7 +1766,13 @@ impl MpcContract {
...
+    /// Called after key resharing concludes (from vote_reshared).
+    /// TODO(anodar): add comment in commit explaining there is a race condition where
+    /// node operator wants to migrate a node to different node, spins up a new node
+    /// which registers new config, but then key resharing happens unrelated to that
+    /// that would cleanup new config. Node restart solves that.
+    /// Alternative is to not clean up but we probably don't want to accumulate
+    /// stale tls key -> config entries.
     #[private]
     #[handle_result]
     pub fn clean_foreign_chain_data(&mut self) -> Result<(), Error> {
```
anodar, line 1852: Tthere is a race condition where: node operator wants to migrate a node to different node, spins up a new node which registers new config, but then key resharing happens unrelated to that, that would cleanup new config. 

Solution: restart the new node so that registration is triggered again.


Alternative: not clean up but we probably don't want to accumulate stale tls key -> config entries.

crates/contract/src/tee/tee_state.rs
```diff
@@ -120,7 +120,8 @@ impl TeeState {
             let node_id = NodeId {
                 account_id: account_id.clone(),
                 tls_public_key: tls_public_key.clone(),
-                account_public_key: Ed25519PublicKey::from([0u8; 32]),
+                // TODO(anodar): add comment in PR explaining we need unique signer account pks.
+                account_public_key: tls_public_key.clone(),
```
anodar, line 128: Changing this as we need unique tls keys (for tests).

crates/contract/src/foreign_chains_metadata.rs
anodar: `register` now keeps mapping of `tls_key` -> `signer_pk`, but clears up old `tls_key` if same signer_pk is used to register config with different `tls_key`.

I've implemented it this way based on the assumption that when we migrate node, we don't reuse the same signer, is that correct @kevindeforth ? 

Edit: I guess it's correct at least for TEE since it's generated in TEE on first run.

anodar (commented):
Thanks both for the reviews. I've made changes to address your comments:

- Config registration is now per node, not per provider
- we cleanup stale (from inactive) configs after vote_reshared (which means there's race condition if node providers starts node migration right before it), see comment on that
- revisited comments, should hopefully be more concise now

anodar:
@claude review pay closer attention that these properties hold:

multiple nodes should be able to register their configuration even from the same node provider
when calculating available foreign chains, only the configs from active nodes should be taken into account
comments should be concise without describing too much detail

anodar:
@claude review pay closer attention that these properties hold:

- multiple nodes should be able to register their configuration even from the same node provider
- when calculating available foreign chains, only the configs from active nodes should be taken into account

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
```diff
@@ -1144,6 +1144,78 @@ pub struct ForeignChainConfiguration(BTreeMap<ForeignChain, NonEmptyBTreeSet<Rpc
...
+    derive_more::Deref,
+    derive_more::DerefMut,
+)]
+#[cfg_attr(
+    all(feature = "abi", not(target_arch = "wasm32")),
+    derive(schemars::JsonSchema, borsh::BorshSchema)
+)]
+pub struct ForeignChainsConfig(BTreeSet<ForeignChain>);
+
+/// Per-node foreign-chain configs as reported via `register_foreign_chains_config`, keyed by each
```
kevindeforth: For comments like these, it's always preferred to use [`import/path_to::register_foregin_chains_config`], such that they don't become stale. C.f. [slack thread](https://nearone.slack.com/archives/C07UW93JVQ8/p1781109113004619?thread_ts=1781104793.615029&cid=C07UW93JVQ8)

crates/contract/src/lib.rs
```diff
@@ -1677,7 +1773,7 @@ impl MpcContract {
 
     /// Private endpoint to clean up foreign chain policy votes and node configurations
     /// for non-participants after resharing.
-    /// This can only be called by the contract itself via a promise.
+    /// Called after key resharing concludes (from vote_reshared).
```
kevindeforth: > /// Called after key resharing concludes (from vote_reshared).

This risks becoming a stale comment, because it documents behavior of a different method.

crates/contract/src/lib.rs
```diff
@@ -1712,7 +1814,21 @@ impl MpcContract {
                 .remove(account);
         }
 
-        self.foreign_chain_rpc_whitelist.votes.retain(participants);
+        // Remove foreign chain configs and their reverse-map entries for departed nodes.
+        let active_tls_keys: std::collections::BTreeSet<_> =
+            participant_tls_keys.iter().cloned().collect();
```
kevindeforth: It looks like we are constructing a HashSet on line 1798, that is then converted to a `BTreeSet` here. I don't see a reason why we couldn't construct the BTreeSet driectly.

crates/contract/src/foreign_chains_metadata.rs
```diff
@@ -0,0 +1,181 @@
...
+        self.tls_key_by_signer_pk.insert(signer_pk, tls_key.clone());
+        self.foreign_chains_configs
+            .insert(tls_key, foreign_chains_config);
+    }
+
+    /// Removes all entries from `tls_key_by_signer_pk` and `foreign_chains_configs` whose TLS key
+    /// is not in `active_tls_keys`. Called during participant set cleanup.
+    pub(crate) fn remove_stale_configs(
+        &mut self,
+        active_tls_keys: &std::collections::BTreeSet<dtos::Ed25519PublicKey>,
```
kevindeforth: nit: I think `retain` would be a better suited method name.

crates/contract/src/lib.rs
```diff
@@ -151,14 +153,19 @@ pub struct MpcContract {
...
     tee_state: TeeState,
     accept_requests: bool,
     node_migrations: NodeMigrations,
     // TODO(#2937): Remove via state migration.
     metrics: Metrics,
-    foreign_chain_rpc_whitelist: ForeignChainRpcWhitelist,
+    /// Lazily-loaded foreign-chain state: RPC provider whitelist, per-node config reports, and
+    /// the cached available-chain set. Only deserialized on methods that touch foreign-chain data,
+    /// keeping unrelated calls cheap.
+    foreign_chains: Lazy<ForeignChainsMetadata>,
```
kevindeforth: I don't think we need to explain what `Lazy` does. It's a pretty standard concept.

I also doubt the usefulness of listing the data stored in `ForeginChainMetadata`. If it needs explanation, I encourage to find a better struct name.

crates/contract/src/lib.rs
```diff
@@ -969,6 +976,89 @@ impl MpcContract {
...
+        self.recompute_available_foreign_chains();
+
+        Ok(())
+    }
+
+    /// Recomputes [`Self::available_foreign_chains`]: chains supported by ≥ the signing threshold
+    /// of active participants (stale non-participant reports excluded).
+    ///
+    /// No-op outside `Running`: only that state has a single [`ThresholdParameters`] where
+    /// threshold and participants are consistent (during `Resharing` they disagree).
```
kevindeforth: blocker: This is an issue.
We have a running state embedded in the resharing state.
If an old node is changing their foreign chain config during a resharing, we should update the list of available chains, otherwise, we risk the nodes attempting to include participants that are not available.


The comment states:
> /// No-op outside `Running`: only that state has a single [`ThresholdParameters`] where
>    /// threshold and participants are consistent (during `Resharing` they disagree).

This is inaccurate. Please note that the initializing state has a single, well defined threshold.

crates/contract/src/lib.rs
```diff
@@ -969,6 +976,89 @@ impl MpcContract {
         Ok(())
     }
 
+    /// (Re)registers the foreign chains this node currently covers.
+    ///
+    /// Registrations are accepted from existing or prospective participant
+    /// in any active protocol state (Running, Resharing,
+    /// Initializing) but only takes effect when in Running-state.
```
kevindeforth: I don't see why we couldn't update the list of available chains in any other state.

crates/contract/src/lib.rs
```diff
@@ -969,6 +976,89 @@ impl MpcContract {
...
+                }
+            }
+        }
+
+        fc.available_foreign_chains = chain_to_supporter_count
+            .into_iter()
+            .filter_map(|(chain, count)| (count >= threshold).then_some(chain))
+            .collect::<BTreeSet<_>>()
+            .into();
+    }
```
kevindeforth: Feels like a lot of this logic should be moved to `ForeignChainsMetadata` behind a

```rust
impl ForeignChainsMetadata {
    fn update_cache(
        &mut self,
        active_node_tls_keys: BTreeSet<_>,
        signing_threshold: Threshold,
    );
}
```

kevindeforth (commented):
Thanks for the updates!
There is just one blocker / discussion point. I don't see a good reason for not updating the available chain cache during resharing and initializing states. In fact, I think it may lead to some issues if we don't allow that.

crates/contract/src/foreign_chains_metadata.rs
anodar: (since it's marked as nit, I assume it's optional) so I'll retain the current name (pun intended) as `remove_stale_configs` conveys intent clearer than `retain_active_nodes_configs` or something like that.

crates/contract/src/lib.rs
anodar: Addressed in [c24be8f](https://github.com/near/mpc/pull/3511/commits/c24be8f8904088d590e2d7dacbaf2d0b344d5b71)

crates/contract/src/lib.rs
anodar: Addressed.

crates/contract/src/lib.rs
anodar: Addressed in [e6e8a9e](https://github.com/near/mpc/pull/3511/commits/e6e8a9e0e43fa27d9679bdd32a4a75fe9c8a104d)

crates/contract/src/lib.rs
anodar: Dropped in [4788e5c](https://github.com/near/mpc/pull/3511/commits/4788e5c17bff7bd39762e331654e7b9e7a135ec2)

crates/contract/src/lib.rs
anodar: Only explanation I have is it possibly made sense in previous setting (when keys were stored by account) and slop to cleanup after switching to per tls key storage.

crates/contract/src/lib.rs

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
anodar: Cargo doc produces errors on that since this interface crate doesn't depend on mpc-contract (where `register_foregin_chains_config` is).

Either we add `#![allow(rustdoc::broken_intra_doc_links)]` to allow broken intra doc links or reword to not mention that symbol at all. Chose the latter.

crates/contract/src/v3_11_2_state.rs
```diff
@@ -61,7 +67,12 @@ impl From<MpcContract> for crate::MpcContract {
             accept_requests: old.accept_requests,
             node_migrations: old.node_migrations,
             metrics: old.metrics,
-            foreign_chain_rpc_whitelist: old.foreign_chain_rpc_whitelist,
+            // New in this revision; per-node configs self-populate on the first
+            // `register_foreign_chains_config` after upgrade.
+            foreign_chains: Lazy::new(
+                StorageKey::ForeignChainAvailability,
+                ForeignChainsMetadata::default(),
+            ),
```
gilcu3: nit: the "new in this revision" could be dropped 🤖

crates/contract/src/foreign_chains_metadata.rs
```diff
@@ -0,0 +1,207 @@
...
+        }
+        // Two different signer keys must never share a TLS key. TLS key uniquely identifies a node.
+        assert!(
+            !self.foreign_chains_configs.contains_key(&tls_key)
+                || self
+                    .tls_key_by_signer_pk
+                    .get(&signer_pk)
+                    .is_some_and(|k| *k == tls_key),
+            "TLS key already registered by a different signer"
+        );
```
gilcu3: I am not sure what's the best practice for near contracts now, but I remember something about using `near-sdk::env::panic_str`

crates/contract/src/foreign_chains_metadata.rs
```diff
@@ -0,0 +1,207 @@
...
+        self.tls_key_by_signer_pk.insert(signer_pk, tls_key.clone());
```
gilcu3: at which point does this set get cleaned if the signer remains a signer (in which case I believe `remove_stale_configs` below does not apply), but it has had several nodes in the past, with different TLS keys

crates/contract/src/lib.rs
```diff
@@ -969,6 +973,71 @@ impl MpcContract {
...
+            .protocol_state
+            .is_existing_or_prospective_participant(&node_id.account_id)?;
+        if !is_participant {
+            return Err(InvalidState::NotParticipant {
+                account_id: node_id.account_id.clone(),
+            }
+            .into());
+        }
+        let tls_key = node_id.tls_public_key.clone();
+        let signer_pk = node_id.account_public_key.clone();
```
gilcu3: I am a bit confused about this part. Here it appears the node will have a single tls_key, but in other parts it appears we want to keep the possibility of a node having several. Do we just have the last one as "active" and keep the rest for migrations?

crates/contract/src/lib.rs
```diff
@@ -1677,7 +1752,6 @@ impl MpcContract {
 
     /// Private endpoint to clean up foreign chain policy votes and node configurations
     /// for non-participants after resharing.
-    /// This can only be called by the contract itself via a promise.
     #[private]
     #[handle_result]
     pub fn clean_foreign_chain_data(&mut self) -> Result<(), Error> {
```
gilcu3, line 1852: was that line removed on purpose?

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
kevindeforth: file-path links also work, but if it can't be linked, then the comment is probably not that useful.

crates/contract/src/foreign_chains_metadata.rs
```diff
@@ -62,11 +64,35 @@ impl ForeignChainsMetadata {
             .insert(tls_key, foreign_chains_config);
     }
 
+    pub(crate) fn update_cache(
+        &mut self,
+        active_tls_keys: &[dtos::Ed25519PublicKey],
```
kevindeforth: I think passing a ref to a BTreeSet would be diligent, as we don't want to double-count the same tls key.

crates/contract/src/foreign_chains_metadata.rs
```diff
@@ -62,11 +64,35 @@ impl ForeignChainsMetadata {
...
+        threshold: u64,
```
kevindeforth, line 44: note that this should be the signing threshold of the domain we use for foreign chain transaction validations.

gilcu3 (commented):
Dropping a few non-blocking comments here, didn't get enough time today after another big PR review got in the way. Will finish it on Monday, good weekend!

crates/contract/src/lib.rs
```diff
@@ -1024,35 +1020,22 @@ impl MpcContract {
...
-            .into();
+        let active_tls_keys: Vec<_> = params
+            .participants()
+            .participants()
+            .iter()
+            .map(|(_, _, info)| info.tls_public_key.clone())
+            .collect();
+        self.foreign_chains
+            .get_mut()
+            .update_cache(&active_tls_keys, threshold);
```
kevindeforth: This should take as input not the governance threshold, but the domain threshold of the foreign transaction domain.

crates/contract/src/foreign_chains_metadata.rs
```diff
@@ -0,0 +1,207 @@
...
+impl ForeignChainsMetadata {
+    /// Registers `foreign_chains_config` for the node.
+    /// This means same node operator can register config per node.
+    /// If the same `signer_pk` previously registered with a different TLS key, the stale entry is
+    /// removed first.
+    pub(crate) fn register(
+        &mut self,
+        signer_pk: dtos::Ed25519PublicKey,
+        tls_key: dtos::Ed25519PublicKey,
+        foreign_chains_config: dtos::ForeignChainsConfig,
```
kevindeforth: ah, I missed this during the previous review. The same signer key shouldn't register multiple TLS keys.
This register function should just take as input the tls key and the config.

crates/contract/src/foreign_chains_metadata.rs
```diff
@@ -0,0 +1,207 @@
...
+#[derive(Debug)]
+pub(crate) struct ForeignChainsMetadata {
+    pub(crate) rpc_whitelist: ForeignChainRpcWhitelist,
+    pub(crate) available_foreign_chains: dtos::AvailableForeignChains,
+    // Stores tls key -> foreign chain config mapping.
+    pub(crate) foreign_chains_configs:
+        IterableMap<dtos::Ed25519PublicKey, dtos::ForeignChainsConfig>,
+
+    /// Map of node's tls key per signer account public key. Same node operator can have multiple nodes, therefore multiple tls key.
+    tls_key_by_signer_pk: IterableMap<dtos::Ed25519PublicKey, dtos::Ed25519PublicKey>,
```
kevindeforth: We don't need this, this information is stored in the TEE struct.

crates/contract/src/lib.rs
kevindeforth: > Here it appears the node will have a single tls_key, but in other parts it appears we want to keep the possibility of a node having several. Do we just have the last one as "active" and keep the rest for migrations?

yeah, good point. I skipped this in my last review.
We don't need multiple entries if we index by tls key.

crates/contract/src/lib.rs
kevindeforth: It's a repetition that doesn't add any value. The comment starts with "Private endpoint".

crates/contract/src/lib.rs
netrome: Oh this is a problem, since there could be multiple foreign transaction domains. This means we might need to rethink the design here. Perhaps the concept of "available foreign chain" has to be defined in the context of a domain.

E.g. for domain X, chain C might be available, but the same chain could be unavailable for domain Y if Y has a higher participant requirement than X.

Also note we're not interested in reconstruction thresholds here but rather the active participant count for the scheme. For example, if we use Robust ECDSA for a foreign tx domain we need 2t-1 signers to participate in a signing round.

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
anodar: Keeping it dropped.

crates/contract/src/foreign_chains_metadata.rs
anodar: Addressed in [db5dbd5](https://github.com/near/mpc/pull/3511/commits/db5dbd5c6f45e3a0eaa34bbf509b30b2138ce517)

crates/contract/src/foreign_chains_metadata.rs
anodar: Addressed in [db5dbd5](https://github.com/near/mpc/pull/3511/commits/db5dbd5c6f45e3a0eaa34bbf509b30b2138ce517)

crates/contract/src/v3_11_2_state.rs
anodar: Addressed in [db5dbd5](https://github.com/near/mpc/pull/3511/commits/db5dbd5c6f45e3a0eaa34bbf509b30b2138ce517)

crates/contract/src/foreign_chains_metadata.rs
anodar: This was enforced that same key doesn't register multiple TLS keys, we removed old key first.

Changed the logic now, register only takes tls key and config. We now rely in `stored_attestations` map we never have different TLS keys for the signer_pk. 

Addressed in [26af250](https://github.com/near/mpc/pull/3511/commits/26af2502fc2a3d0a1fbe3a52fb0e278601c6fb1c)

crates/contract/src/foreign_chains_metadata.rs
anodar: Got rid of it in [26af250](https://github.com/near/mpc/pull/3511/commits/26af2502fc2a3d0a1fbe3a52fb0e278601c6fb1c)

crates/contract/src/lib.rs
anodar: > I am a bit confused about this part. Here it appears the node will have a single tls_key, but in other parts it appears we want to keep the possibility of a node having several. 


I'm assuming you're referring to `register` function which was taking: `signer_pk`, `tls_key` and `foreign_chains_config`. That function checked if there already existed different `tls_key` for that `signer_pk` it removed it first, so it didn't allow storing multiple tls keys for the same node either. But since (from my understand) can't have attestation for same same pk_s with different tls keys, we don't need to pass signer pk.

So yes there should be a single tls key per node (/per signer pk), but possibly multiple per node operator. For threshold we only count ones from active participants.

Changed register function [26af250](https://github.com/near/mpc/pull/3511/commits/26af2502fc2a3d0a1fbe3a52fb0e278601c6fb1c) hopefully makes it bit more clear.

crates/contract/src/foreign_chains_metadata.rs
anodar: Addressed in [1973ee0](https://github.com/near/mpc/pull/3511/commits/1973ee0c8873bb2e765856bcdaa8f8b0ee1e8944).

crates/contract/src/lib.rs
anodar: As @kevindeforth mentioned in slack discussion so far we have single domain for foreign tx validation so we can just use that threshold.

If we decide to have multiple domains for different chains we'll introduce in the future, it should still be straightforward to extend to support it: we'll additionally keep mapping of which chain uses what domain and recompute available chains with thresholds from corresponding domain.

Addressed in [1973ee0](https://github.com/near/mpc/pull/3511/commits/1973ee0c8873bb2e765856bcdaa8f8b0ee1e8944).

crates/node/src/runtime.rs
anodar, line 1: Clippy complains about unused imports in mac os otherwise. I chose to skip entire mod tests module (which currently has single test) over annotating each use as well.

crates/contract/src/foreign_chains_metadata.rs
anodar: Not applicable anymore, I dropped this (`tls_key_by_signer_pk`) map in [26af250](https://github.com/near/mpc/pull/3511/commits/26af2502fc2a3d0a1fbe3a52fb0e278601c6fb1c)

crates/contract/src/lib.rs
netrome: > we'll additionally keep mapping of which chain uses what domain and recompute available chains with thresholds from corresponding domain.

Yeah I think this is the correct solution that we should adopt.

> Addressed in [1973ee0](https://github.com/near/mpc/pull/3511/commits/1973ee0c8873bb2e765856bcdaa8f8b0ee1e8944).

I don't think this solution is sufficent (as commented on slack). This can give inconsistent results if we were to add more domains. The fact that our deployed systems only support a single foreign chain domain is not an invariant we can rely on. We might change this just within the next few weeks. For example, after we conclude the TEE migration we want to regenerate new domains for all use cases.

netrome: While we don't need to pivot to the complete "per-domain-availability" solution now, we should at least make sure we have a consistent definition of available foreign chains in the presence of multiple domains. I added a stacked PR for this, we only need minor tweaks to the logic to use the max of all domains instead of the value of the first domain:

https://github.com/near/mpc/pull/3555

netrome (changes requested):
I don't think we should rely on assuming there will only be a single foreign tx validation domain. Here's a sufficient suggestion to fix this: https://github.com/near/mpc/pull/3555 although the more proper solution would be to change the model to define availability per domain.

crates/contract/src/tee/tee_state.rs
netrome: Blocking: We should update the comment above that references the sentinel value you removed here.

crates/node/src/runtime.rs
gilcu3: is it about this line?

> use thread_priority::get_current_thread_priority;

scripts/check-contract-wasm-size.sh
```diff
@@ -16,7 +16,10 @@ WASM_PATH="${1:-result/mpc_contract.wasm}"
 # ~12 KB over the pre-feature baseline; this limit gives a bit of headroom
 # above the post-feature size without leaving the contract free to creep up
 # to the protocol boundary.
-HARD_LIMIT=1500000
+# Bump up from 1500000 to 1520000 as part of #3475. TODO(#3475): reduce the limit
+# once we migrate to new endpoints after upgrading to 3.12 and drop deprecated
+# endpoints and DTO structs.
+HARD_LIMIT=1520000
```
gilcu3, line 22:
```suggestion
# TODO(#3475): reduce the limit
# once we migrate to new endpoints after upgrading to 3.12 and drop deprecated
# endpoints and DTO structs.
HARD_LIMIT=1520000
```

and we better do that fast :) Would love to get it back to 1450000

netrome (changes requested):
One small blocker, otherwise LGTM

crates/contract/src/foreign_chains_metadata.rs
```diff
@@ -0,0 +1,151 @@
...
+    ) {
+        self.foreign_chains_configs
+            .insert(tls_key, foreign_chains_config);
+    }
+
+    pub(crate) fn update_cache(
+        &mut self,
+        active_tls_keys: &BTreeSet<dtos::Ed25519PublicKey>,
+        threshold: u64,
+    ) {
```
gilcu3, line 45: a small comment on this function would help understanding when it is meant to be used, as it was not clear to me given `update_cache` reads very generic

crates/contract/src/lib.rs
```diff
@@ -152,14 +154,16 @@ pub struct MpcContract {
     pending_ckd_requests: LookupMap<CKDRequest, Vec<YieldIndex>>,
     pending_verify_foreign_tx_requests: LookupMap<VerifyForeignTransactionRequest, Vec<YieldIndex>>,
     proposed_updates: ProposedUpdates,
+    // TODO(#3475): drop this once we upgrade the contract and nodes start using
+    // the new API.
     node_foreign_chain_support: SupportedForeignChainsByNode,
```
gilcu3, line 159: does it make sense to add an underscore on this one? Or is it still wired in some places after this PR?

anodar:
`TODO(anodar)` was dropped while ago in [ab54e95](https://github.com/near/mpc/pull/3511/commits/ab54e95577bc494ae83bb8ad512b1c9a6a300df9).

I didn't add comment in `account_public_key` mentioning hardcoded value so that I wouldn't have "Using unique key instead of hardcoded because of xyz" (as I got feedback not to write that type of comments).

Added the comment now.  We still need to address https://github.com/near/mpc/issues/1087 separately though.

crates/contract/src/lib.rs
```diff
@@ -1790,7 +1883,19 @@ impl MpcContract {
...
+            .remove_stale_configs(&active_tls_keys);
+
+        self.foreign_chains
+            .get_mut()
+            .rpc_whitelist
+            .votes
+            .retain(participants);
+
+        // The active set just changed; refresh the cache against the new participants/threshold.
+        self.recompute_available_foreign_chains();
```
gilcu3: We might have a problem here, as this function might `clean_foreign_chain_data` fail after resharing (because it is a promise), and I am not sure about the system correctness if that's the case. It might be the case we need to do this in the reshare function itself, so that it cannot fail unless the original transaction fails

gilcu3 (changes requested):
Looks good to me. 

One minor blocker is that now that we did the branch cut for 3.12, we will need to update the state transition before this PR can land.

Another one is that if the cleanup promise fails after resharing, maybe the system will stay in an inconsistent state.

crates/contract/src/lib.rs
anodar: Good point, done in [9dfeaab](https://github.com/near/mpc/pull/3511/commits/9dfeaabe74560a72107ad211a31da879c7953176)

crates/contract/src/lib.rs
anodar: This is still the one that's being used and will be used until we land these changes in the contract, then make another release that includes update node to wire up new API.
Only then we can make a release where we can drop this.

crates/contract/src/foreign_chains_metadata.rs
anodar: Renamed to `update_available_chains_config_cache` to make it clear without comment in [f603960](https://github.com/near/mpc/pull/3511/commits/f60396040e818100b9458992e10755849f63b26a)

crates/node/src/runtime.rs
anodar: that and super::*; as well iirc

netrome (dismissed):
Thanks for updating 🙏

crates/contract/src/tee/tee_state.rs
```diff
@@ -106,11 +106,11 @@ impl TeeState {
 
         for (account_id, _, participant_info) in participants.participants() {
             let tls_public_key = participant_info.tls_public_key.clone();
-            // TODO(#1087): replace this sentinel with a real account public
+            // TODO(#1087): account_public_key with a real account public
```
netrome: Nit: Typo

```suggestion
            // TODO(#1087): replace account_public_key with a real account public
```

crates/contract/src/v3_11_2_state.rs
gilcu3, line 1: What we are doing here is not enough for the transition. We need to do something like https://github.com/near/mpc/pull/3483/changes (which includes making a clean file, adding the old 3.12 contract to history). Right now we had some problems with 3.12, so probably we will need to wait for that situation to be sorted out. Still would be good to do this in a separate PR

anodar: Dropped state transition changes from here and will address in separate PR once we sort out problems with 3.12.

crates/contract/src/tee/tee_state.rs
netrome: Thanks for fixing.

netrome (dismissed):
Thanks for the updates!

crates/contract/src/lib.rs
```diff
@@ -1017,6 +1021,84 @@ impl MpcContract {
...
+    pub fn register_foreign_chains_config(
+        &mut self,
+        foreign_chains_config: dtos::ForeignChainsConfig,
+    ) -> Result<(), Error> {
+        Self::assert_caller_is_signer();
+        let signer_account_id = env::signer_account_id();
+        let signer_account_pk = env::signer_account_pk();
+        let signer_account_ed25519_pk =
+            Ed25519PublicKey::try_from(&signer_account_pk).map_err(|_| {
+                InvalidParameters::InvalidTeeRemoteAttestation {
```
kevindeforth: nit: `InvalidTeeRemoveAttetsation` doesn't seem to be the correct error, as it's unrelated to TEE attestations.
I would think an `.expect` is fine here.

crates/contract/src/lib.rs
```diff
@@ -1017,6 +1021,84 @@ impl MpcContract {
...
+        }
+        let is_participant = self
+            .protocol_state
+            .is_existing_or_prospective_participant(&node_id.account_id)?;
+        if !is_participant {
+            return Err(InvalidState::NotParticipant {
+                account_id: node_id.account_id.clone(),
+            }
+            .into());
+        }
```
kevindeforth, line 1055: This is worth documenting. So, we only allow a new node to submit its configuration, _after_ we have started the resharing to include it in the network.

We might want to re-visit this in the future, as it makes it more complicated to onboard new node.

We could probably avoid this, if we required a deposit to pay for any storage costs.

crates/contract/src/lib.rs
```diff
@@ -1017,6 +1021,84 @@ impl MpcContract {
...
+        let Some(threshold) = self.protocol_state.domain_registry().ok().and_then(|r| {
+            r.domains()
+                .iter()
+                .filter(|d| d.purpose == DomainPurpose::ForeignTx)
+                .map(|d| d.reconstruction_threshold.inner())
+                .max()
+        }) else {
+            // No op if contract isn't in Running or Resharing state, or
+            // there is no foreign tx domain registered.
+            // Not panicing is intentional.
```
kevindeforth: nit: typo
```suggestion
            // Not panicking is intentional.
```

crates/contract/src/lib.rs
gilcu3, line 1055: would be good to add tests for these error paths, it seems we are only currently testing the happy path

gilcu3 (dismissed):
Thank you for all the work here!

As I mentioned on slack, the only thing remaining here before merging afaik is to patch the state transition.

crates/contract/src/lib.rs
```diff
@@ -1087,6 +1169,7 @@ impl MpcContract {
 
         if let Some(new_state) = self.protocol_state.vote_pk(key_event_id, extended_key)? {
             self.protocol_state = new_state;
+            self.recompute_available_foreign_chains();
```
kevindeforth: We don't need this here, this is for key generation only.

crates/contract/src/lib.rs
```diff
@@ -1132,6 +1215,9 @@ impl MpcContract {
         if let Some(new_state) = self.protocol_state.vote_reshared(key_event_id)? {
             // Resharing has concluded, transition to running state
             self.protocol_state = new_state;
+            // update_cache iterates active_tls_keys only, so departed nodes' configs are
+            // already excluded — correct result even before cleanup promises run.
+            self.recompute_available_foreign_chains();
```
kevindeforth, line 1216: Question: All other cleanup steps, we spawn in promises, to avoid running out of gas (c.f. the five promises spawned below).
Do we know how much additional gas this requires to conclud the resharing step?

I also notice that we spawn a promise to call `clean_foreign_chain_data`, which also recomputes the cache.

anodar:
Converting to draft so that it's not accidentally merged until preconditions are met.

crates/contract/src/lib.rs
anodar: Filed follow-up issue: https://github.com/near/mpc/issues/3585

kevindeforth (dismissed):
Thanks you!

Disclaimer: I didn't have time to review the unit tests or the contract migration.

My remaining concerns are regarding two items:
- we only allow current or prospective participants to register their foreign chain config. That seems to make onboarding new nodes tricky and we will likely have to re-visit in the future
- we update the cache of available foreign chain configs at the end of the resharing. This goes against existing convention, where we try to off-load as much as possible to follow-up promises. IMO, it would be fine to drop this and have possibly 1 block of stale contract data post resharing.

Other than that, API and data structs look good.

crates/contract/src/lib.rs
anodar: `clean_foreign_chain_data` no longer recomputes the cache. 
recomputing available foreign chains isn't spawned as a promise so that it's always recomputed when key resharing is complete (or whole transaction is failed otherwise).

Both of those changes were to address this comment: https://github.com/near/mpc/pull/3511#discussion_r3414278703

crates/contract/src/lib.rs
anodar: Filed followup task: https://github.com/near/mpc/issues/3586

crates/contract/src/lib.rs
anodar: Addressed in [5412b13](https://github.com/near/mpc/pull/3511/commits/5412b137a685257d2b3ef36e2c13163ce15dbce1).

crates/contract/src/lib.rs
anodar: [5412b13](https://github.com/near/mpc/pull/3511/commits/5412b137a685257d2b3ef36e2c13163ce15dbce1)

crates/contract/src/lib.rs
anodar: addressed in [5412b13](https://github.com/near/mpc/pull/3511/commits/5412b137a685257d2b3ef36e2c13163ce15dbce1)

But used panic_str instead of expect.

gilcu3 (dismissed):
Thank you!

anodar:
@claude review

netrome (approved):

crates/contract/src/lib.rs
anodar: Now that I think it probably should have stayed here. If we add another domain for Foreign Chain, this should be recomputed here as well as key generation completes. Although not yet applicable.

anodar:
@claude 

Regarding blocking: addressed in followup https://github.com/near/mpc/pull/3617

> crates/contract/src/lib.rs:1214-1216 — Inline comment refers to update_cache, but the helper was renamed to update_available_chains_config_cache in [f603960](https://github.com/near/mpc/pull/3511/commits/f60396040e818100b9458992e10755849f63b26a). The property the comment describes is still correct; only the symbol name is stale.

Addressed in followup: https://github.com/near/mpc/pull/3617

> crates/contract/src/foreign_chains_metadata.rs:33 — Doc on register adds little over IterableMap::insert: "Registers `foreign_chains_config` keyed by TLS key, so multiple nodes from the same operator can coexist." The second clause is a consequence of the containing type's schema (the map is keyed by Ed25519PublicKey), not of the register method. Either drop the doc or move the "multiple nodes from the same operator" sentence onto the foreign_chains_configs field, next to the schema decision. Per [engineering-standards.md](https://github.com/near/mpc/pull/docs/engineering-standards.md#write-helpful-code-comments) §Paraphrasing the code.

I think comment is valuable explaining why we store configs per tls key: so that we allow smooth migration of nodes


> crates/contract/src/tee/tee_state.rs:450-459 — lookup_node_id_by_signer_pk does a linear scan over stored_attestations per call. The map is keyed by TLS key, so the only way to resolve signer_pk → node_id is iteration. For long-lived deployments the map grows and each step is a storage read. Not a regression and the call site is rare (register_foreign_chains_config only), but worth a TODO if registration becomes a hot path — e.g. maintain a signer_pk → tls_pk reverse map.


Had that map, and got rid of it per @kevindeforth's feedback.

> crates/contract/src/lib.rs:1063-1066 — Separate from the blocking doc point above: the doc lists NotInitialized as the only no-op case, but the function is also a no-op when there's no ForeignTx domain in the registry. Worth a one-line addition so a reader doesn't infer "this always runs in Running."

It's obvious if there's no foreignTx domain nothing will be recomputed in foreign chains config. imho doesn't add much value adding comment on it.

> Prior-review carryover (acknowledged, not re-litigating): MpcContract Borsh snapshot now collapses foreign_chains to Lazy { storage_key: Vec<u8> }; a dedicated inner-schema snapshot would restore the structural-change safety net.

Lazy is used here as per @gilcu3's suggestion so that non foreign chain tx validation transactions don't consume more gas as a result. Added test in followup so that we detect schema change there.

gilcu3 (approved):

crates/contract/src/lib.rs
```diff
@@ -1224,6 +1308,9 @@ impl MpcContract {
 
         if let Some(new_state) = self.protocol_state.vote_cancel_keygen(next_domain_id)? {
             self.protocol_state = new_state;
+            // Cancel-keygen drops incomplete domains, which can change the max ForeignTx
+            // threshold — recompute to keep the cached set consistent.
+            self.recompute_available_foreign_chains();
```
kevindeforth, line 1313: note: For these cancellation paths, I prefer not having too much unnecessary bloat in them.
If we cancel a key generation, it's usually because something went wrong. 
Updating the contract code in a state other than running is either not supported, or really difficult to do.
Hence, being able to exit these states is quite important to stay operational.

In this case, I'd much rather have a mis-configured available foreign chain view for a few blocks, than risking to be stuck in a key generation due to unforeseen gas costs (AFAIK, we don't have any benches on `recompute_available_foreign_chain()`).

It's something that needs to be addressed in a follow-up.

kevindeforth (approved):
Thank you!
A few small items are left, but I see we already have a follow-up for those, so happy to approve.

crates/contract/src/lib.rs
anodar: Addressing this in followup: https://github.com/near/mpc/pull/3617

PR #3515, opened 2026-06-10
fix: add asset id checks for follower paths

### LINKED ISSUE #3514: Validate that follower-consumed presignature and triple ids are owned by the leader
### Background

Follower paths consume assets via `take_unowned(id)` using an id supplied by the
leader, without checking that the id is owned by the leader. Owned and unowned
assets share the same store key space, so a leader could supply an id embedding
the follower's own participant id and cause `take_unowned` to extract one of the
follower's owned assets — enabling asset double-use (a nonce-reuse-class risk).

The generation paths already guard this with `validate_owned_by`, but these did
not:

- `EcdsaSignatureProvider` sign follower (`presignature_id`)
- `RobustEcdsaSignatureProvider` sign follower (`presignature_id`)
- `EcdsaSignatureProvider` presignature follower (`paired_triple_id`)

### Acceptance Criteria

- Sign followers validate `presignature_id.validate_owned_by(leader)` before consuming the presignature.
- The ECDSA presignature follower validates `paired_triple_id.validate_owned_by(leader)` before consuming the triple pair.

gilcu3:
Closes #3514

### DIFF
--- a/crates/node/src/providers/ecdsa/presign.rs
+++ b/crates/node/src/providers/ecdsa/presign.rs
@@ -172,7 +172,11 @@ impl EcdsaSignatureProvider {
         domain_id: DomainId,
         paired_triple_id: UniqueId,
     ) -> anyhow::Result<()> {
-        id.validate_owned_by(channel.sender().get_leader())?;
+        let leader = channel.sender().get_leader();
+        // Both the new presignature id and the triples it consumes must be
+        // owned by the leader, never one of ours.
+        id.validate_owned_by(leader)?;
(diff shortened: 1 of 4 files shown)


netrome (approved):
Nice! I guess adding a test case where a non-leader tries to attempt to use a non-owned asset is non-trivial?

gilcu3:
> Nice! I guess adding a test case where a non-leader tries to attempt to use a non-owned asset is non-trivial?

Yeah, technically we can, but it is a horrible setup (we still need to work on making the provider part of the implementation easily testable)

anodar (approved):

PR #3518, opened 2026-06-10
feat: add yield point in CPU bound crypto code

### LINKED ISSUE #3501: Make sure the crypto code does not clog the runtime
### Background

We have been facing signing degradation right after resharing for a while #1175 . As explained in [signing-starvation-solution.md](https://github.com/near/mpc/blob/dd1695a5e92326a3b667608af73ce3b796ff2379/docs/design/signing-starvation-solution.md), one of the possible solutions in the short term is to avoid clogging the tokio runtime by inserting yield statements where appropriate.

### User Story

Lower latency is preferred

### Acceptance Criteria

We have inserted yield statements in the main crypto implementations to guarantee the runtime is not clogged by any task. We have measured task contention in the runtime before and after to verify there is a noticeable improvement.

[thread](https://nearone.slack.com/archives/C07UW93JVQ8/p1780927443564329)

gilcu3:
Closes #3501

### DIFF
--- a/crates/node/src/protocol.rs
+++ b/crates/node/src/protocol.rs
@@ -40,10 +40,11 @@ pub async fn run_protocol<T>(
     //    inefficient, so instead we put messages into queues, indexed by the recipient, and have
     //    a parallel task for each recipient that sends the messages.
     //  - We need the sending task to be a separate spawn from the computation task because while
-    //    we're computing, we would not be able to cooperatively run any other tasks, and that can
-    //    unnecessarily block sending. It is OK to have the receiving side blocked by computation,
-    //    because on the receiving side, the network channel already provides us with a buffer
-    //    dedicated to our task.
+    //    we're computing, we would not be able to cooperatively run any other tasks (computation
+    //    only yields at the protocol's explicit yield points), and that can unnecessarily block
+    //    sending. It is OK to have the receiving side blocked by computation, because on the
+    //    receiving side, the network channel already provides us with a buffer dedicated to our
+    //    task.
     let sending_handle = {
         let counters = counters.clone();
         let sender = channel.sender();
@@ -74,11 +75,21 @@ pub async fn run_protocol<T>(
     let participants = channel.participants().to_vec();
     let my_participant_id = channel.my_participant_id();
     let computation_handle = async move {
+        /// How a poke burst ended, deciding what to do after flushing messages.
+        enum PokeOutcome<T> {
+            Wait,
+            Yield,
+            Return(T),
+        }
+
         loop {
             let mut messages_to_send: HashMap<ParticipantId, _> = HashMap::new();
-            let done = loop {
+            let outcome = loop {
                 match protocol.poke()? {
-                    Action::Wait => break None,
+                    Action::Wait => break PokeOutcome::Wait,
+                    // Flush the accumulated messages before yielding, so peers can make
+                    // progress while we give other tasks a chance to run.
+                    Action::Yield => break PokeOutcome::Yield,
                     Action::SendMany(vec) => {
                         for participant in &participants {
                             if participant == &my_participant_id {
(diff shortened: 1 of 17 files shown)


crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/triples/batch_random_ot.rs
```diff
@@ -250,29 +250,28 @@ pub(super) async fn batch_random_ot_receiver(
...
+        // Step 5
+        out.push(hash(
+            i,
+            &big_x_i_verkey,
+            &big_y_verkey,
+            &CoefficientCommitment::new(big_y * x_i),
+        )?);
+
+        chan.yield_point().await;
+    }
```
gilcu3, line 274: small rewrite as a loop to be able to insert the yield

crates/node/src/tests/protocol_yielding.rs
```diff
@@ -0,0 +1,148 @@
...
+const TRIPLES_PER_BATCH: usize = SUPPORTED_TRIPLE_GENERATION_BATCH_SIZE;
+
+/// Upper bound on how long the canary may go unscheduled. Calibrated in
+/// test-release on an unloaded machine: ~130ms max gap with the yield points
+/// in place vs ~8.7s with them neutered (the canary wakes up only twice
+/// during the whole batch). Unoptimized builds are ~10x slower and will
+/// exceed this threshold; run with `--cargo-profile=test-release`.
+// TODO(#3517): the ~130ms floor comes from the OT fan-out in triple
+// multiplication; tighten this once per-poll child polling is budgeted.
+const CANARY_MAX_GAP: Duration = Duration::from_millis(300);
```
netrome, line 49: Oh 8.7 seconds? No wonder the MPC nodes gets bricked. 300 MS is already wayyyy too much for an async runtime. Ofc I understand this setup doesn't mimic production perfectly but still.

crates/node/src/protocol.rs
```diff
@@ -74,11 +75,21 @@ pub async fn run_protocol<T>(
     let participants = channel.participants().to_vec();
     let my_participant_id = channel.my_participant_id();
     let computation_handle = async move {
+        /// How a poke burst ended, deciding what to do after flushing messages.
+        enum PokeOutcome<T> {
+            Wait,
+            Yield,
+            Return(T),
+        }
+
```
netrome, line 84: This feels unnecessarily complex for the sake of inserting yield points. Hy not just insert a yield point first thing in the inner loop? Yields are cheap so I don't think we'd experience any issues from an extra yield when returning.

Or was there any particular reason for doing it this way?

netrome (approved):
Nice stuff! Feels potentially a bit over-engineered in the `run_protocol` loop. Curious about your thoughts here.

crates/node/src/protocol.rs
gilcu3: it is about the messages that are already waiting to be sent in a batch. When we yield we want to send them first, then actually yield, so that we dont delay them any further

crates/node/src/tests/protocol_yielding.rs
gilcu3: yeah locally the diff was huge, but we knew we were clogging the runtime, so the big surprise is the number

crates/node/src/protocol.rs
netrome: Would yielding before this send cause any problems? A single yield should not cost much, unless you share a runtime with a task that doesn't yield which is what we're trying to solve here anyway :sweat_smile:

gilcu3: I don't think it would cause any direct problem, but could risk it for sure, as the messages are ready to be sent and would need to wait for an arbitrary amount of time before they are actually sent

anodar (approved):

PR #3519, opened 2026-06-10
fix: outgoing connection hanging on unstable network

### LINKED ISSUE #3513: Outgoing connection attempt can hang indefinitely, permanently disconnecting a peer until restart
### Background

`OutgoingConnection::new` (`crates/node/src/p2p.rs`) performs three awaits when dialing a peer: `TcpStream::connect`, the TLS handshake, and the p2p handshake. Only the last one is wrapped in a timeout (`HANDSHAKE_TIMEOUT`, 2s) — the TCP connect and TLS handshake have none.

There is a realistic failure mode where neither ever completes nor fails: a middlebox (load balancer, firewall, NAT) in front of the peer accepts the TCP connection and ACKs our ClientHello, but never sends a byte back. TCP keepalive doesn't fire (probes are ACKed) and `TCP_USER_TIMEOUT` doesn't trigger (no un-ACKed outbound data), so the TLS handshake future waits forever. Since the `PersistentConnection` retry loop awaits `OutgoingConnection::new` with no overall deadline, a single such attempt wedges the loop permanently: no errors, no retries, no further log lines for that peer until the node is restarted.

We observed this in production. The outgoing connection to one peer had been flapping for ~2 days (`write operation timed out after 10s (connection may be half-open)` → reconnect about once per minute), then a burst of `tls handshake eof` / `Connection reset by peer` dial failures ended with the `Persistent connection to <id>` task going silent mid-retry-loop. From that moment, `mpc_network_live_connections{peer_participant_id="<id>"}` stayed at `0` for 14+ hours, while *incoming* connections from the same peer continued to succeed — confirming the peer was up and only our outgoing dial path was stuck. All signature and triple-generation tasks involving the peer stalled with `Waiting for connection to <id> (for 60s)`. Only a node restart recovers.

This silently degrades network redundancy while the node otherwise looks healthy — and it's hard to spot from logs precisely because the symptom is the *absence* of log lines.

### User Story

As a node operator, I want connection attempts to peers to always terminate (succeed or fail) within a bounded time, so that a misbehaving peer-side middlebox cannot permanently sever my node's connectivity to that peer.

### Acceptance Criteria

- [ ] `OutgoingConnection::new` is bounded by a timeout covering TCP connect and TLS handshake (either one overall deadline or per-step timeouts)
- [ ] On timeout, the persistent-connection loop logs the failure and retries as it does for other dial errors
- [ ] Test covering a dial that never completes (e.g., a listener that accepts TCP but never responds), asserting the loop keeps retrying

- Hang points: `crates/node/src/p2p.rs` — `TcpStream::connect` and `TlsConnector::connect` in `OutgoingConnection::new`; the retry loop in `PersistentConnection::new` awaits it with no deadline
- The 2s `HANDSHAKE_TIMEOUT` on the p2p handshake shows the intended pattern; it just doesn't cover the two earlier steps

gilcu3:
Closes #3513

### DIFF
--- a/crates/node/src/p2p.rs
+++ b/crates/node/src/p2p.rs
@@ -150,6 +150,12 @@ impl OutgoingConnection {
     /// the connection is considered not successful.
     const HANDSHAKE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(2);
 
+    /// Maximum time to establish the TCP connection and complete the TLS
+    /// handshake. Without this bound, a middlebox that accepts the TCP
+    /// connection but never sends a byte back would hang the dial — and the
+    /// persistent-connection retry loop awaiting it — forever.
+    const CONNECT_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+
     /// Makes a TLS/TCP connection to the given address, authenticating the
     /// other side as the given participant.
     async fn new(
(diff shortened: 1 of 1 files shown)


crates/node/src/p2p.rs
```diff
@@ -159,16 +165,20 @@ impl OutgoingConnection {
...
 
-        let mut tls_stream = tokio_rustls::TlsConnector::from(client_config)
-            .connect("dummy".try_into().unwrap(), tcp_stream)
-            .await
-            .context("failed to establish tls stream")?;
+            tokio_rustls::TlsConnector::from(client_config)
+                .connect("dummy".try_into().unwrap(), tcp_stream)
+                .await
+                .context("failed to establish tls stream")
+        })
```
netrome, line 179: Nit: The `new` function is already quite big. I wonder it we should consider breaking this out to a separate function for readability.

crates/node/src/p2p.rs
```diff
@@ -1430,4 +1444,108 @@ mod tests {
...
+        .expect("dial did not terminate within bounded time");
+
+        // Then the dial fails with a timeout instead of hanging.
+        let Err(error) = result else {
+            panic!("dial unexpectedly succeeded");
+        };
+        assert!(
+            format!("{error:#}").contains("timed out"),
+            "unexpected error: {error:#}"
+        );
```
netrome, line 1508: It's a shame we use `anyhow` here instead of returning a proper type so we could assert the actual return type. Would be a nice refactor in the future.

netrome (approved):
Quick review, looks good to me - thanks!

Just minor nits and thoughts, feel free to ignore.

crates/node/src/p2p.rs
gilcu3: indeed

pbeza (approved):

crates/node/src/p2p.rs
gilcu3: I was planning to fix this, but I don't think it is worth asking for re-reviews again, so will need to leave it for the next time.

Issue #2874, opened 2026-04-15
e2e test temp dirs are cleaned up on failure, hiding node logs


gilcu3:
### Background

When an e2e test fails because a node exits early, the panic message points to the stderr log:

```
failed to start cluster: mpc-node 0 exited early — check /tmp/.tmpXXXXX/node0/stderr.log
```

But the temp dir is cleaned up by the time the test process exits (`tempfile::TempDir` Drop), so the log file is gone and the failure cannot be diagnosed locally.

`ClusterConfig` has a `home_base` option but it's not wired to an env var, so there's no easy way to override this at runtime.

### Acceptance Criteria

- On test failure, the node logs (stderr, stdout, and home dir contents) are preserved in a known location, or
- An env var (e.g. `E2E_HOME_BASE` or `E2E_KEEP_TMP`) is supported that disables temp dir cleanup so logs can be inspected after failures.

PR #3520, opened 2026-06-10
test(e2e): dump node 0 stdout/stderr tails when sigterm_handler test fails

barakeinav1:
PR #3410 just merged, and Patrick spotted the `sigterm_handler` e2e test failing on his branch with `code=None signal=Some(6)` (SIGABRT) in [this CI run](https://github.com/near/mpc/actions/runs/27267382103/job/80527942157). 5/6 historical runs on the same branch passed, so it's a flake — mpc-node aborts during shutdown instead of exiting cleanly.

This PR adds inline log-tail capture so future failures show what mpc-node was doing right before the abort. Without it the test's tempdir gets cleaned up before we can read the panic backtrace.

Most likely root cause: a tokio task in the indexer's runtime panics when its underlying nearcore actor goes away during `shutdown_all_actors()`. PR #3486 (indexer-drain follow-up) should fix this by draining the runtime cleanly first. I'm running a 5-cycle campaign on #3486's branch now to verify.

## Test plan

- [ ] CI green

### DIFF
--- a/crates/e2e-tests/src/cluster.rs
+++ b/crates/e2e-tests/src/cluster.rs
@@ -1034,6 +1034,13 @@ impl MpcNodeState {
         }
     }
 
+    pub fn home_dir(&self) -> &Path {
+        match self {
+            MpcNodeState::Running(n) => n.setup().home_dir(),
+            MpcNodeState::Stopped(s) => s.home_dir(),
+        }
+    }
+
     pub fn p2p_public_key(&self) -> Ed25519PublicKey {
         match self {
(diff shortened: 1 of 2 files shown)


gilcu3:
Related to #2874

netrome:
Feels worth revisiting this. @barakeinav1 any reason for this getting stuck? Secondly, this approach feels quite hacky. Shouldn't the e2e test framework already provide a way to inspect/dump logs from failing tests? Sounds like something we should properly solve.

netrome:
@gilcu3 pointed out this is fixed in https://github.com/near/mpc/pull/4214

PR #3521, opened 2026-06-10
demo(tee-verifier): reuse mpc-attestation collateral parsing in the verify_quote test

pbeza:
Demonstrates one way to address @gilcu3's comment on #3237 ([r3387007113](https://github.com/near/mpc/pull/3237#discussion_r3387007113)): the `verify_quote` test's `make_collateral()` hand-maps the fixture JSON into the interface `Collateral` and hex-decodes the byte fields itself, duplicating logic that already exists.

This branch reuses `mpc_attestation::collateral::Collateral`'s `TryFrom<Value>` parser (which owns the hex-decoding) to produce a `dcap_qvl::QuoteCollateralV3`, then converts it to the interface mirror via a new public `IntoInterfaceType<Collateral>` impl (the inverse of the existing interface→dcap one). The test drops its hand-rolled JSON/hex mapping; all 21 `tee-verifier` tests still pass.

**Why this is a draft / for discussion, not folded into #3237:** it adds a dev-dependency on `mpc-attestation` to `tee-verifier`. Per `docs/design/attestation-verifier-contract.md`, `tee-verifier` is meant to be a standalone, MPC-agnostic, reusable verifier that other teams (Proximity, Defuse) can adopt without linking MPC-specific crates — and `mpc-attestation` is explicitly the MPC-specific framing. Even a dev-only dep works against that goal, so the hand-rolled mapping in #3237 may be the right trade-off for a shared crate. Opening this so the reviewer can weigh the reuse against the coupling.

Targets `feat/tee-verifier-contract` so the diff is just this change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11368,6 +11368,7 @@ dependencies = [
  "dcap-qvl",
  "getrandom 0.2.17",
  "hex",
+ "mpc-attestation",
  "near-sdk",
  "rstest",
  "tee-verifier",
--- a/crates/tee-verifier/Cargo.toml
+++ b/crates/tee-verifier/Cargo.toml
@@ -41,6 +41,7 @@ getrandom = { workspace = true, features = ["custom"] }
 
 [dev-dependencies]
 hex = { workspace = true }
+mpc-attestation = { workspace = true, features = ["test-utils"] }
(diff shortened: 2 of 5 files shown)


pbeza:
Demo only.

PR #3525, opened 2026-06-10
feat: Add debug info about the number of configured foreign chain RPC providers

### LINKED ISSUE #3524: Add debug information about the number of configured foreign chain RPC providers
### Background

We just updated our foreign chain configuration on our mainnet and testnet nodes by adding more RPC endpoints. However, we noted that there's no good way to verify this update on-chain or in debug endpoints since we removed the foreign chain RPC configuration from the public debug endpoints.

While we don't want to expose the full foreign chain RPC configuration, we'd like to show at least the number of configured providers per chain per node to help verify and create visibility on these config updates.

### User Story

As a user of foreign transaction verification I'd like to know how many independent RPCs each node has configured.

### Acceptance Criteria

It's possible to see debug information on the node how many RPC providers it has configured per chain.

_No response_

anodar:
Closes: https://github.com/near/mpc/issues/3524

### DIFF
--- a/crates/node/src/web.rs
+++ b/crates/node/src/web.rs
@@ -10,8 +10,8 @@ use futures::FutureExt;
 use futures::future::BoxFuture;
 use mpc_attestation::attestation::Attestation;
 use mpc_node_config::{
-    CKDConfig, ConfigFile, IndexerConfig, KeygenConfig, PresignatureConfig, SignatureConfig,
-    TripleConfig,
+    CKDConfig, ConfigFile, ForeignChainsConfig, IndexerConfig, KeygenConfig, PresignatureConfig,
+    SignatureConfig, TripleConfig,
 };
 use near_account_id::AccountId;
 use near_mpc_contract_interface::types::Ed25519PublicKey;
@@ -91,6 +91,7 @@ struct NodeConfigResponse {
     signature: SignatureConfig,
     ckd: CKDConfig,
     keygen: KeygenConfig,
+    foreign_chains_provider_counts: ForeignChainsProviderCounts,
     cores: Option<usize>,
 }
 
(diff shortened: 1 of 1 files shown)


anodar:
@claude review

anodar:
Unrelated test seems flaky: 
```
SIGABRT [  11.994s] (956/966) chain-gateway::test event_subscriber_integration::test_event_subscriber_receiver_error_if_non_private_call
```

Will rerun once workflows are finished.

crates/node/src/web.rs
```diff
@@ -109,11 +110,57 @@ impl From<ConfigFile> for NodeConfigResponse {
...
+
+#[derive(Clone, Serialize)]
+struct ForeignChainsProviderCounts {
+    #[serde(skip_serializing_if = "is_zero")]
+    solana: usize,
+    #[serde(skip_serializing_if = "is_zero")]
+    bitcoin: usize,
+    #[serde(skip_serializing_if = "is_zero")]
+    ethereum: usize,
+    #[serde(rename = "abstract", skip_serializing_if = "is_zero")]
```
netrome, line 131: Very curious why Claude initially created options here. Did you get any good explanation for it? That seemed below what I'd expect from it.

netrome (approved):
Thanks! 🙏

crates/node/src/web.rs
anodar: I didn't ask at the time, I assumed it was to differentiate chain not setup vs chain setup but has 0 providers (which are the same for us). Also that Option<usize> didn't need custom is_zero function because `Option::is_none` already exists (not a good reason though).

I asked it now and it says:
```
  The underlying config fields are already Option<ChainConfig> (each chain is optional). 
Option<usize> was a direct 1-to-1 mapping of that optionality into the output type 
— None meaning "chain not configured", Some(n) meaning "configured with n providers". It's the
  obvious mechanical translation.
```

DSharifi (approved):

PR #3529, opened 2026-06-11
perf: optimize CKD computations in contract

### LINKED ISSUE #3352: Optimize CKD response verification in the contract
### Background

Currently we are overly conservative in the CKD output verification:

https://github.com/near/mpc/blob/cf179467124203b0187ef0e80b429885b9a51627/crates/contract/src/primitives/ckd.rs#L57-L83

Given that internally `bls12381_pairing_check` already does the in group checks, we could make the code much more efficient, and potentially drop the `blstrs` dependency in the contract.

### User Story

_No response_

### Acceptance Criteria

The CKD output verification is simplified to about doing duplicate steps (decompression) and checks (in prime order group). Tests are added so that we guarantee that the host functions will do the needed checks.

[thread](https://nearone.slack.com/archives/C0B12RKBSAV/p1779726702054209)

gilcu3:
Closes #3352 

This is a relatively small PR with lots of defensive tests


crates/contract/tests/sandbox/sign.rs
```diff
@@ -443,11 +443,11 @@ async fn test_ckd_gas_regression() -> anyhow::Result<()> {
...
 
     // Gas thresholds (in TGas) with ~25% buffer over measured values.
-    // PV adds BLS12-381 pairing checks on both request and respond.
+    // PV adds BLS12-381 checks on both request and respond.
     const MAX_LEGACY_REQUEST_TGAS: u64 = 10;
     const MAX_LEGACY_RESPOND_TGAS: u64 = 7;
-    const MAX_PV_REQUEST_TGAS: u64 = 72;
-    const MAX_PV_RESPOND_TGAS: u64 = 75;
+    const MAX_PV_REQUEST_TGAS: u64 = 19;
+    const MAX_PV_RESPOND_TGAS: u64 = 30;
```
kevindeforth, line 450: nice!

crates/contract/src/primitives/ckd.rs
```diff
@@ -31,117 +31,161 @@ impl CKDRequest {
...
 pub(crate) fn app_public_key_check(app_public_key: &dtos::CKDAppPublicKeyPV) -> bool {
-    let Some(pk1) = G1Affine::from_compressed(&app_public_key.pk1).into_option() else {
-        return false;
-    };
-    let Some(pk2) = G2Affine::from_compressed(&app_public_key.pk2).into_option() else {
-        return false;
-    };
-    if !check_valid_point_g1(&pk1) || !check_valid_point_g2(&pk2) {
-        return false;
-    }
```
kevindeforth, line 44: Note: the main change in this PR is to remove this check here...

crates/contract/src/primitives/ckd.rs
```diff
@@ -31,117 +31,161 @@ impl CKDRequest {
...
 ) -> bool {
-    let Some(big_c) = G1Affine::from_compressed(&output.big_c).into_option() else {
-        return false;
-    };
-    let Some(big_y) = G1Affine::from_compressed(&output.big_y).into_option() else {
-        return false;
-    };
-    if !check_valid_point_g1(&big_c) || !check_valid_point_g1(&big_y) {
-        return false;
-    }
```
kevindeforth, line 71: ...and here.

crates/contract/src/primitives/ckd.rs
```diff
@@ -178,42 +223,308 @@ mod tests {
...
+        )
+        .into_option()
+        .expect("hash point is a valid G1 point");
+        let compressed = hex::encode(point.to_compressed());
         insta::assert_snapshot!(compressed);
     }
 
+    #[test]
+    #[expect(non_snake_case)]
+    fn bls12381_pairing_check__should_reject_g1_point_outside_prime_order_subgroup() {
```
kevindeforth, line 316: and we make sure that the host method covers those checks in these unit tests.

kevindeforth (dismissed):
Had to brush  up on some math, which was fun. This PR looks correct to me.

IIUC, we remove two explicit checks: we no longer check that the point is on the elliptic curve and we no longer check that the point is torsion free (where, torsion-free means the point lives in the prime-order subgroup G1).

The two checks are already covered by the host function `env::bls12381_pairing_check` and this PR adds unit tests to ensure the host function rejects cases of the above (point outside of prime order subgroup, point not on curve).

crates/contract/src/primitives/ckd.rs
gilcu3: hehe yeah, thanks for the explanatory comments, hopefully they are helpful for the next reviewer :)

crates/contract/src/primitives/ckd.rs
```diff
@@ -31,117 +31,161 @@ impl CKDRequest {
...
+    let minus_g2 = (-G2Projective::generator()).to_uncompressed();
+
+    let pairing_input = [
+        pk1.as_slice(),
+        minus_g2.as_slice(),
+        g1.as_slice(),
+        pk2.as_slice(),
+    ]
+    .concat();
     env::bls12381_pairing_check(&pairing_input)
```
SimonRastikian, line 73: It's not clear at all to me what this function does exactly and whether it can actually replace the previous checks. Could you point me to its actual implementation?

gilcu3: as I pointed a few days ago in slack, this is implemented in [link](https://github.com/near/nearcore/blob/093286305d23034e9aa3b167e22a1ebb9ee4ebb8/runtime/near-vm-runner/src/logic/bls12381.rs#L329)

crates/contract/src/primitives/ckd.rs
```diff
@@ -31,117 +31,161 @@ impl CKDRequest {
...
-
-    let g1 = G1Projective::generator().to_uncompressed().to_vec();
-    let minus_g2 = (-G2Projective::generator()).to_uncompressed().to_vec();
-
     let pk1 = env::bls12381_p1_decompress(&app_public_key.pk1);
     let pk2 = env::bls12381_p2_decompress(&app_public_key.pk2);
-
-    let pairing_input = [pk1, minus_g2, g1, pk2].concat();
+    let g1 = G1Projective::generator().to_uncompressed();
+    let minus_g2 = (-G2Projective::generator()).to_uncompressed();
```
SimonRastikian: Nit: these are constant terms so no need to compute them every time. Perhaps keep them stored and use the same ones?

SimonRastikian (dismissed):

SimonRastikian (approved):
Nice work thanks

kevindeforth (approved):
Thank you!

PR #3532, opened 2026-06-11
fix: make sure localnet launcher script kills processes successfully

### LINKED ISSUE #3531: launch-localnet.sh can no longer kill nodes after SIGTERM graceful-shutdown change
### Background

`scripts/launch-localnet.sh` tears down nodes via `kill_process`, which sends a
single plain `kill "$pid"` (SIGTERM) and neither waits nor escalates to SIGKILL.

Before #3410, `mpc-node` had no SIGTERM handler, so SIGTERM hit the OS default
and terminated the process immediately. After #3410, SIGTERM triggers a graceful
shutdown (`near_async::shutdown_all_actors()`), which is slow and can hang — the
node's own comment in `crates/node/src/run.rs` anticipates this:

> in this embedding it would hang past any reasonable grace period, and the
> orchestrator would SIGKILL us before the graceful path completes

The script is that orchestrator but has no SIGKILL fallback, so nodes linger
after teardown.

### Acceptance Criteria

- `kill_process` sends SIGTERM, waits a bounded grace period, then escalates to
  SIGKILL if the process is still alive.
- Running `launch-localnet.sh` and triggering cleanup leaves no orphaned
  `mpc-node` / `neard` processes.

gilcu3:
Closes #3531

### DIFF
--- a/scripts/launch-localnet.sh
+++ b/scripts/launch-localnet.sh
@@ -264,21 +264,36 @@ require_cmds() {
   [[ "${missing}" -eq 0 ]] || die "Please install the missing dependencies above."
 }
 
+# Send SIGTERM, wait up to GRACE seconds, then escalate to SIGKILL. mpc-node
+# now handles SIGTERM with a graceful shutdown that can be slow or hang, so a
+# single SIGTERM is no longer enough to reliably stop a node.
 kill_process() {
   local pid="$1"
+  local grace="${2:-5}"
 
   if [[ -z "$pid" ]]; then
     echo "Error: No PID provided." >&2
     return 1
   fi
 
-  if kill -0 "$pid" 2>/dev/null; then
-    echo "Killing process $pid..."
-    kill "$pid"
-    echo "Process $pid terminated." >&2
-  else
+  if ! kill -0 "$pid" 2>/dev/null; then
     echo "Warning: Process $pid does not exist. Skipping." >&2
+    return 0
   fi
+
+  echo "Sending SIGTERM to $pid (grace ${grace}s)..."
+  kill "$pid" 2>/dev/null
+
+  for ((s = 0; s < grace; s++)); do
+    if ! kill -0 "$pid" 2>/dev/null; then
+      echo "Process $pid exited gracefully." >&2
+      return 0
+    fi
+    sleep 1
+  done
+
+  echo "Process $pid still alive after ${grace}s; sending SIGKILL." >&2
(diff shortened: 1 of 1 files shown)


scripts/launch-localnet.sh
```diff
@@ -264,20 +264,39 @@ require_cmds() {
...
+      return 0
+    fi
+    sleep 1
+  done
+
+  echo "Process $pid still alive after ${grace}s; sending SIGKILL." >&2
+  kill -9 "$pid" 2>/dev/null || true
+
+  if kill -0 "$pid" 2>/dev/null; then
+    echo "Warning: Process $pid still alive after SIGKILL." >&2
```
anodar: Chuck-Norris process detected.

anodar (dismissed):
Thank you!

scripts/launch-localnet.sh
gilcu3: The worst part is that these are real Chuck-Norris processes :D

scripts/launch-localnet.sh
```diff
@@ -264,20 +264,39 @@ require_cmds() {
...
+  for ((s = 0; s < grace; s++)); do
+    if ! kill -0 "$pid" 2>/dev/null; then
+      echo "Process $pid exited gracefully." >&2
+      return 0
+    fi
+    sleep 1
+  done
+
+  echo "Process $pid still alive after ${grace}s; sending SIGKILL." >&2
+  kill -9 "$pid" 2>/dev/null || true
```
barakeinav1, line 296: nit: right after `kill -9` on a direct child, the process is a zombie until bash reaps it, so this `kill -0` succeeds and prints "still alive after SIGKILL" on every normal teardown (verified). Reaping it first makes the check report correctly and leaves nothing lingering. (The grace loop above is fine — its `sleep 1` gives bash time to reap.)

```suggestion
  kill -9 "$pid" 2>/dev/null || true
  wait "$pid" 2>/dev/null || true
```

barakeinav1 (dismissed):
LGTM — correctly escalates SIGTERM → grace → SIGKILL and meets the AC. One cosmetic nit inline.

scripts/launch-localnet.sh
gilcu3: as discussed in slack, this is a valid point, will handle by having a grace period before printing that the process did not die

barakeinav1 (approved):

anodar (approved):

PR #3533, opened 2026-06-11
fix(tdx): default state sync to Peers (DSS); require tier3_public_addr

### LINKED ISSUE #3526: Stop forcing ExternalStorage state-sync in mpc-node TDX init path; default to Peers (DSS-only)
### Background

Until recently NEAR supported two state-sync modes: **centralized** (download state parts from an S3/GCS bucket — `ExternalStorage`) and **decentralized / peer-to-peer** (DSS — `Peers`). In **nearcore 2.12.0** centralized state-sync was formally deprecated ([near/nearcore#15500](https://github.com/near/nearcore/pull/15500)) and the buckets are no longer served; nearcore's serde default for `state_sync.sync` is **`Peers`** (`client_config.rs:315-328`, `Default = Peers`).

mpc-node still actively patches the node's NEAR config (on the TDX init path) to **force `ExternalStorage`** and write `external_storage_fallback_threshold`. That needs to come out.

> **Scope: TDX init path only.** The non-TDX `deployment/start.sh` path is tracked separately in #3535.

`apply_near_config_patches` (`crates/node/src/config/start.rs:155-168`) does, for every non-localnet chain:

```rust
let storage_fallback_threshold = near_init.external_storage_fallback_threshold.unwrap_or(0);
config["state_sync"]["sync"]["ExternalStorage"]["external_storage_fallback_threshold"] = json!(...);
```

Two problems with this now:

1. **`threshold = 0` (the unset default) is a broken config.** In nearcore (`chain/client/src/sync/state/downloader.rs:76-83`), `0` means "skip peers, go straight to external storage on every attempt." With the bucket deprecated, external returns nothing and peers are never tried → **state sync never completes; the node is stuck.** (#3506 mitigates by flipping the template default to `100`, but the root cause is forcing `ExternalStorage` at all.)
2. **The patch is fragile.** It blind-writes the `ExternalStorage.external_storage_fallback_threshold` sub-key, assuming the RPC-downloaded config already carries a full `ExternalStorage { location, … }` block (`location` is required — `core/chain-configs/src/client_config.rs:191`). If NEAR moves their published config to `sync: "Peers"` or drops the block, this patch injects a location-less `ExternalStorage` (won't deserialize) or panics on the serde_json index.

The correct end-state is to **not force `ExternalStorage`** and let it default to `Peers`.

### Acceptance Criteria

docs and code are updated

- Non-TDX (`deployment/start.sh`) cleanup tracked in #3535.
- Verified against pinned nearcore: `SyncConfig` enum + `Default = Peers` (`client_config.rs:315-328`); `location` required (`:191`); native threshold default `3` (`:41`); `threshold=0` → bucket-only (`downloader.rs:76-83`).
- Related: #3505 / #3506 (flip default to DSS-first — interim), #3509 (fail closed when `tier3_public_addr` unset on multi-IP). This issue is the deeper cleanup that makes those partly unnecessary.
- Do this subject to testing — confirm a fresh TDX node state-syncs with `sync` defaulted to `Peers` and no external-storage config.

barakeinav1:
Closes #3526

remove support for deprecated centralized sync

> **Scope: TDX launcher path only.** The non-TDX `deployment/start.sh` path is split into #3535 / PR #3536.

📋 **Design / implementation plan:** [#3526 design comment](https://github.com/near/mpc/issues/3526#issuecomment-4678960620).

> ⚠️ **Operator-facing breaking change: `tier3_public_addr` is now mandatory.** A fresh TDX node **fails to start at first init** if `tier3_public_addr` is unset .

- Existing CVMs: (e.g upgrade path) are unaffected

**Testing gate:** confirm a fresh TDX node state-syncs end-to-end with `sync=Peers` and no external-storage config. Unit tests cover the Peers switch, ExternalStorage-block replacement, localnet untouched, and the tier3 guard.


crates/node/src/config/start.rs
```diff
@@ -57,6 +57,8 @@ impl StartConfigExt for StartConfig {
             return Ok(());
         }
 
+        require_tier3_public_addr(near_init)?;
```
barakeinav1, line 60: config is applied only during first initialization. So we don't need to worry about backward compatibility during a node upgrade

crates/node-config/src/start.rs
```diff
@@ -115,13 +115,6 @@ pub struct NearInitConfig {
     /// outbound IP and DSS times out. Patches into nearcore's
     /// `network.experimental.tier3_public_addr` config field.
     pub tier3_public_addr: Option<SocketAddr>,
-    /// Override how many P2P (DSS) state-sync attempts to make before falling
-    /// back to the external storage bucket. `0` (current default) means
-    /// "go straight to bucket, never use DSS." A moderate value enables
-    /// DSS-first with bucket as a safety net; a very large value effectively
-    /// disables the bucket fallback. Patches into nearcore's
-    /// `state_sync.sync.ExternalStorage.external_storage_fallback_threshold`.
-    pub external_storage_fallback_threshold: Option<u64>,
```
barakeinav1, line 124: safe to remove, even if field still exists in config.toml, it is just ignored silently with no error.

deployment/cvm-deployment/ci-config.toml
```diff
@@ -40,6 +40,7 @@ chain_id = "mainnet"
 boot_nodes = "ed25519:ERguu7jQuYk8pxNsRC6FdezvNsegBPva1GRGqjmtD7i2@10.10.10.10:24567"
 download_genesis = false
 download_config = "rpc"
+tier3_public_addr = "10.10.10.10:24567"
```
barakeinav1, line 43: just a placeholder. node never reaches state sync

crates/node/src/config/start.rs
barakeinav1: But note that any new TDX deployment will fail if this is not set.

crates/node/src/config/start.rs
```diff
@@ -125,6 +127,23 @@ fn convert_download_config(dt: DownloadConfigType) -> near_config_utils::Downloa
...
+             sync needs a reachable advertised IP:24567. Set tier3_public_addr in \
+             [mpc_node_config.near_init].",
+            near_init.chain_id
+        );
+    }
+    Ok(())
+}
+
 /// behaviour of `update_near_node_config()` in `start.sh`.
 fn patch_near_config(
```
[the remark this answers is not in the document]
barakeinav1: Good catch — fixed. My insertion had split `patch_near_config`'s two-line rustdoc. Restored its full header and removed the orphaned line above `require_tier3_public_addr`.

docs/running-an-mpc-node-in-tdx-external-guide.md
```diff
@@ -729,7 +728,7 @@ filter = "mpc=debug,info"
...
+> **⚠️ Set `tier3_public_addr` before first start.** State sync is decentralized (peer-to-peer) and requires the node to advertise a **publicly reachable** `IP:24567`. The template ships `tier3_public_addr` as a `REPLACE_WITH_…` placeholder and the node **fails to start if it's left unset** — replace it with the IP your dstack port-forward exposes for `:24567`. This matters most on hosts with more than one external IP or running multiple nodes, where auto-discovery would advertise an unreachable address and state sync would stall. It is applied at first init only, so getting it right up front avoids a CVM redeploy later.
```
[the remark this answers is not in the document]
barakeinav1: Fixed — now reads "unset **or left as the placeholder**". (The placeholder isn't a valid SocketAddr, so leaving it unchanged fails at config parse — still loud, just a different stage than the is_none() guard.)

docs/running-an-mpc-node-in-tdx-external-guide.md
```diff
@@ -738,8 +737,7 @@ Adjust the variables as per your environment.
...
-* `external_storage_fallback_threshold` *(under `[mpc_node_config.near_init]`)* — controls DSS (decentralized state sync). **Leave at `100` — do not change.** Applied at first init only, like `tier3_public_addr`.
+* `tier3_public_addr` *(under `[mpc_node_config.near_init]`)* — `IP:24567` the node advertises for decentralized (Tier3) state-sync responses. **Required — the template ships this as a `REPLACE_WITH_…` placeholder and the node fails to start if it's left unset** (intentionally, so state sync never silently runs with an unreachable advertised address). It is especially critical on any host with more than one external IP, or when running [multiple nodes on one host](./running-multiple-mpc-nodes-on-one-host.md): otherwise the node auto-discovers its advertised address as the host's default-route outbound IP, which peers may not be able to reach, and state sync stalls. Applied at first init only; changing later requires a CVM redeploy via the [Node Migration](./node-migration-guide.md) flow.
```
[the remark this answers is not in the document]
barakeinav1: Fixed with the same wording as above ("unset or left as the placeholder").

pbeza (approved):

pbeza:
@claude review

barakeinav1:
Thanks for the review. Responses to the three non-blocking findings:

> `localnet/tee/scripts/rust-launcher/node.conf.testnet.toml.tpl:22`, `deploy-tee-cluster.sh:828,869`, `how-to-run-deploy-tee-cluster.md:253-261` — the template + deploy script + how-to still write and require `external_storage_fallback_threshold` ... will mislead operators of the localnet TEE cluster.

Confirmed — dead plumbing after this PR (silently ignored, since `NearInitConfig` has no `deny_unknown_fields`). Filed as a follow-up: #3538.

> `crates/e2e-tests/src/mpc_node.rs:446` — e2e builds `ChainId::Custom("sandbox")`, which only round-trips to `ChainId::Sandbox` ... Constructing `ChainId::Sandbox` directly would make the invariant explicit.

Confirmed (pre-existing; blast radius grew with the new guard). Filed as a follow-up: #3539.

> `crates/node/src/config/start.rs:185` — `config["state_sync"]["sync"] = json!("Peers")` panics via serde_json's `IndexMut` if a future downloaded `config.json` ever ships without a `state_sync` key.

Not filing this one. serde_json's `IndexMut` **auto-vivifies** an absent key (and `Null`) into an object — it does not panic when `state_sync` is missing; it only panics if `state_sync` is present as a non-object scalar, which a real NEAR `config.json` never is. This line also uses a *shallower* path than the code already in `main` (`["state_sync"]["sync"]["ExternalStorage"][…]`), so it's strictly less panic-prone than what ships today.

anodar (approved):

PR #3548, opened 2026-06-12
chore: simplify ProposedThresholdParameters

gilcu3:


crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -510,84 +500,45 @@ mod tests {
...
-    /// Wire-format lock: the public contract for `ProposedThresholdParameters` is
-    /// the flat JSON object `{ participants, threshold, per_domain_thresholds }`,
-    /// not the `ProposedThresholdParametersCompat` bridge that currently produces
-    /// it. The snapshot pins the full shape, so deleting the compat struct later
-    /// (letting the type serialize to its native nested `parameters` sub-object)
-    /// surfaces as a snapshot diff rather than a silent wire break.
     #[test]
     #[expect(non_snake_case)]
-    fn proposed_threshold_parameters__serializes_to_flat_wire_shape() {
-        insta::assert_json_snapshot!(sample_proposal());
```
SimonRastikian, line 551: Why you do not want a snap?

crates/near-mpc-contract-interface/src/types/state.rs
```diff
@@ -510,84 +500,45 @@ mod tests {
...
-            .as_object_mut()
-            .unwrap()
-            .remove("per_domain_thresholds");
-
-        // When deserialized.
+        // When
         let parsed: ProposedThresholdParameters = serde_json::from_value(legacy).unwrap();
 
-        // Then the per-domain threshold updates default to empty.
+        // Then
```
SimonRastikian, line 520: Comments here are not useful

SimonRastikian (approved):

crates/near-mpc-contract-interface/src/types/state.rs
gilcu3: I don't see what this snap is for, now we don't have any flat variant

crates/near-mpc-contract-interface/src/types/state.rs
gilcu3: wdym?

SimonRastikian: The comment is //when 
// then

netrome: These are very useful in my opinion. They divide the test into three sections:

// Given
<setup code>

// When
<logic under test>

// Then
<assertions>

I've found test code written under this convention much easier to read and maintain in the past, which is why I've been pushing for adopting the convention here.

PR #3555, opened 2026-06-14
fix: Use max of all foreign tx thresholds for computing available chains

netrome:
Suggestion to aggregate all foreign tx domain thresholds to compute the available threshold.


anodar (approved):
LGTM, thank you!

PR #3438, opened 2026-06-01
test: Fixing flaky test

SimonRastikian:
Fixing a flaky test.

[Description to be added]

PR #3557, opened 2026-06-15
chore: remove 3.11 CaitSith restriction in the contract

### LINKED ISSUE #3306: Lift CaitSith threshold restriction after 3.11 transition

gilcu3:
Closes #3306

### DIFF
--- a/crates/contract/src/state/running.rs
+++ b/crates/contract/src/state/running.rs
@@ -3,10 +3,7 @@ use super::key_event::KeyEvent;
 use super::resharing::ResharingContractState;
 use crate::errors::{ConversionError, DomainError, Error, InvalidParameters, VoteError};
 use crate::primitives::{
-    domain::{
-        AddDomainsVotes, DomainRegistry, validate_caitsith_uniform_threshold,
-        validate_domain_threshold,
-    },
+    domain::{AddDomainsVotes, DomainRegistry, validate_domain_threshold},
     key_state::{AuthenticatedAccountId, AuthenticatedParticipantId, EpochId, Keyset},
     threshold_votes::ThresholdParametersVotes,
     thresholds::{ProposedThresholdParameters, ThresholdParameters},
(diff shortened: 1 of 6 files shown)


crates/e2e-tests/tests/web_endpoints.rs
```diff
@@ -55,7 +55,7 @@ async fn wait_for_body(
         }
         anyhow::ensure!(
             tokio::time::Instant::now() < deadline,
-            "node {node}: {url} body did not match within {timeout:?}"
+            "node {node}: {url} body did not match within {timeout:?}\nlast body:\n{body}"
```
gilcu3, line 58: this came from the debugging session that identified the time bomb, I guess leaving it there does not hurt

pbeza (approved):

crates/contract/src/state/running.rs
```diff
@@ -814,6 +775,21 @@ pub mod running_tests {
...
+            .iter()
+            .map(|(account_id, _, _)| account_id.clone())
+            .find(|account_id| {
+                state
+                    .parameters
+                    .participants()
+                    .is_participant_given_account_id(account_id)
+            })
+            .expect("proposal must retain at least one current participant");
+        env.set_signer(&signer);
```
SimonRastikian, line 792: This fixes a flaky test I was having here https://github.com/near/mpc/pull/3438/changes/9eb6b9d03f289be4376d8cc2cd99fe15e7594e8f

SimonRastikian: I will close the PR I have

SimonRastikian (approved):

anodar (approved):

PR #3572, opened 2026-06-15
chore: restore external_storage_fallback_threshold for pre-3.12.0 nodes

### LINKED ISSUE #3570: Docs drop external_storage_fallback_threshold ahead of 3.12.0, breaking state sync on released (3.11.x) nodes
### Background
PR #3533 removed `external_storage_fallback_threshold` from the TDX operator template (`deployment/cvm-deployment/user-config.toml`) and `docs/running-an-mpc-node-in-tdx-external-guide.md`, and switched the binary to default `state_sync.sync = "Peers"` (DSS). The binary change only ships in 3.12.0, but the doc/template change is already on `main`.

The released node (3.11.2) still defaults `external_storage_fallback_threshold` to `0` (bucket-only) when the field is unset, and nearcore no longer serves the centralized state-sync buckets. A node initialized from the current docs therefore crashes in a restart loop, with the last log line `running state sync shard_id=5`.

Hit during a testnet TDX migration on 2026-06-15 (stakin-mpc2.testnet). Adding `external_storage_fallback_threshold = 100` together with `tier3_public_addr` restored state sync.

### User Story
As an MPC node operator deploying the currently released (3.11.x) image, I need the TDX guide and `user-config.toml` template to produce a working state-sync configuration, so that my node syncs instead of crash-looping when I follow the docs exactly.

### Acceptance Criteria
- The TDX guide and `user-config.toml` template again set `external_storage_fallback_threshold = 100`, scoped as required for node versions before 3.12.0.
- The accompanying note makes clear the field is ignored on 3.12.0+ (which always uses peer-to-peer `Peers` state sync), so the config is safe for both releases.
- An operator following the docs with the released 3.11.x image can complete state sync.

- Caused by: #3533
- Surfaced in the Stakin testnet TDX migration Slack thread, 2026-06-15.

barakeinav1:
Closes #3570

### DIFF
--- a/deployment/cvm-deployment/user-config.toml
+++ b/deployment/cvm-deployment/user-config.toml
@@ -69,6 +69,14 @@ download_config = "rpc"
 # one host, where auto-discovery would otherwise advertise an unreachable IP
 # and state sync would stall silently.
 tier3_public_addr = "REPLACE_WITH_NODE_PUBLIC_IP:24567"
+# Required for node versions before 3.12.0: DSS (decentralized state sync)
+# attempts per state part before falling back to the external storage bucket.
+# Leave at 100 — released 3.11.x nodes default this to 0 (bucket-only), which
+# no longer works now that nearcore has stopped serving the centralized
+# buckets, and a node left at the default crashes in a state-sync restart
+# loop. Ignored on 3.12.0+, which always uses peer-to-peer ("Peers") state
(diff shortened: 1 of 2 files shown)


netrome (approved):
Thanks! Also a reminder that we should be more clear with the operators to check out the guide at the version they are configuring.

gilcu3 (approved):

PR #3573, opened 2026-06-15
fix(release): pin git-cliff to semver tags; restore dropped 3.12.0 changelog entries

pbeza:
The 3.12.0 changelog and release notes listed only 28 of 76 commits — ~47 were silently dropped, including the foreign-chain whitelist verifier (#3327).

**Cause:** `prepare-release.sh` runs `git-cliff --unreleased`, which treats *any* tag as a release boundary. A stray local tag near HEAD became the baseline, so only the commits after it were emitted.

**Fix:**
- `cliff.toml`: pin `tag_pattern` to semver, so only `X.Y.Z` tags count as release boundaries
- `CHANGELOG.md`: regenerate the complete `## [3.12.0]` block (76 commits); older sections unchanged
- `RELEASES.md`: correct the description of how `--unreleased` selects commits

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ This changelog is maintained using [git-cliff](https://git-cliff.org/) and [conv
 
 ### 🚀 Features
 
+- [#3327](https://github.com/near/mpc/pull/3327)(@anodar): Add foreign-chain whitelist verifier (#3327)
+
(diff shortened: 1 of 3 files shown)


netrome (approved):
Thanks for spotting this and fixing it 🙏

gilcu3 (approved):
Thank you!

PR #3577, opened 2026-06-15
chore: make the two threshold DTO conversions fallible & validating

### LINKED ISSUE #3541: Validate threshold parameters when converting proposal DTOs into contract types
### Background

Currently, ThresholdParameters can be converted from DTOs using new_unvalidated, allowing invalid threshold values to enter the contract type layer. The original rationale appears to support tests that use threshold values which would be rejected in production (e.g. values below the minimum allowed threshold).

However, when processing incoming proposals, the contract already expects invalid proposals to be rejected. Converting external input into contract types without validation weakens the boundary between DTOs and contract internals and delays failure until later execution paths.

For production proposal flows, it is preferable to reject invalid parameters as early as possible during DTO -> contract type conversion.

### User Story

_No response_

### Acceptance Criteria

* Replace usage of ThresholdParameters::new_unvalidated(...) in DTO -> contract type conversion paths used by incoming proposals with the validated constructor (new(...)) or equivalent validation logic.
* Invalid threshold values received via proposal DTOs are rejected during conversion and return an appropriate error.
* Existing proposal processing behavior remains unchanged for valid inputs.
* Tests relying on invalid threshold values are updated to construct contract-internal types directly rather than passing invalid values through DTO conversion layers.
* Any remaining uses of new_unvalidated are documented and justified, or tracked in a follow-up issue if they are required for test-only scenarios.

https://github.com/near/mpc/pull/3326#discussion_r3387002849

SimonRastikian:
Closes #3541

### DIFF
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -879,7 +879,7 @@ impl MpcContract {
         proposal: dtos::ProposedThresholdParameters,
     ) -> Result<(), Error> {
         Self::assert_caller_is_signer();
-        let proposal: ProposedThresholdParameters = proposal.into_contract_type();
+        let proposal: ProposedThresholdParameters = proposal.try_into_contract_type()?;
         log!(
             "vote_new_parameters: signer={}, proposal={:?}",
             env::signer_account_id(),
@@ -1805,7 +1805,7 @@ impl MpcContract {
         parameters: dtos::ThresholdParameters,
         init_config: Option<dtos::InitConfig>,
     ) -> Result<Self, Error> {
-        let parameters: ThresholdParameters = parameters.into_contract_type();
+        let parameters: ThresholdParameters = parameters.try_into_contract_type()?;
         // Log participant count and hash - full parameters exceed NEAR's 16KB log limit at ~100 participants
         let params_hash = env::sha256_array(borsh::to_vec(&parameters).unwrap());
         log!(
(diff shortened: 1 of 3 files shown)


crates/contract/src/primitives/thresholds.rs
```diff
@@ -143,7 +143,9 @@ impl ThresholdParameters {
         &self.participants
     }
 
-    /// For integration testing.
+    /// Test-only: builds parameters without [`Self::validate_threshold`], so tests can use
+    /// sub-production thresholds. Production proposal DTOs are validated during conversion
+    /// (see `TryIntoContractType` in `dto_mapping`) and use [`Self::new`] instead.
     pub fn new_unvalidated(participants: Participants, threshold: Threshold) -> Self {
```
kevindeforth: Lets gate this behind a compiler flag

kevindeforth (commented):
Thanks! The comment mentions that `ThresholdParameters::new_unvalidated` is now only used in tests. Lets gate it behind a compiler flag then.

crates/contract/src/lib.rs
```diff
@@ -1856,7 +1856,7 @@ impl MpcContract {
         parameters: dtos::ThresholdParameters,
         init_config: Option<dtos::InitConfig>,
     ) -> Result<Self, Error> {
-        let parameters: ThresholdParameters = parameters.into_contract_type();
+        let parameters: ThresholdParameters = parameters.try_into_contract_type()?;
```
barakeinav1, line 1859: `init_running` is the migration/state-init path, not external input like `init`/`vote_new_parameters` — are its `parameters` always already-valid (from a live running state)? Just want to make sure eager validation here can't reject a legitimate state-init that the old infallible conversion would've accepted.

barakeinav1 (dismissed):
Approved with a question (see the inline `init_running` comment).

crates/contract/src/primitives/thresholds.rs
SimonRastikian: [79ee2a0](https://github.com/near/mpc/pull/3577/commits/79ee2a0d490a830bdf3dd590dc682617bfa6d6a6)

crates/contract/src/lib.rs
SimonRastikian: The eager validation can't reject anything init_running would previously have accepted because the validation happens anyway in line 1872.

kevindeforth (approved):
Merci!

barakeinav1 (approved):

PR #3578, opened 2026-06-15
feat: correlate governancethreshold with reconstructionthreshold

### LINKED ISSUE #3499: Correlate GovernanceThreshold with ReconstructionThreshold
### Background

We need to preserve some relation between GovernanceThreshold and ReconstructionThreshold.

Currently one can vote for having a lower GovernanceThreshold than ReconstructionThreshold which makes the code vulnerable.

For this reason:
1. Lower bound: Governance should be lower bounded by the max of all reconstruction thresholds (care about the Robust ECDSA requirement that Governance > 2t+1). Trust assumptions have to stay consistent: otherwise it would be possible to vote for a reshare that drops drastically the crypto threshold  
2. Non-equality possibility: Governance could be larger than max Reconstruction (Robust ecdsa recon=49% but gov=60%)
3. Upper bound: Governance has to be upper-bounded with a fixed percentage that we feel comfortable with (say 60%). The reason is that we could have a problem if we set the governance too high (say 90%) then it is sufficient for 11% of our participants to stop serving us and our contract is locked: We cannot reshare, cannot vote for join, kick-out (and potentially cannot sign too?).

### User Story

_No response_

### Acceptance Criteria

1. Add an implementation to the conditions above to the smart contract and make a pass over existent thresholds, the pass must make sure that all the thresholds are validated. 
2. Make the smart contract prevent updates (reject proposals) in two cases:
    * If the ReconstructionThreshold vote update breaks the conditions above.
    * If the GovernanceThreshold vote update breaks the conditions above

https://nearone.slack.com/archives/C07UW93JVQ8/p1774362577840599

SimonRastikian:
Closes #3499


crates/contract/src/state/running.rs
```diff
@@ -180,6 +183,14 @@ impl RunningContractState {
             validate_domain_threshold(domain, new_num_participants)?;
         }
 
+        // The GovernanceThreshold must dominate every domain's effective ReconstructionThreshold;
+        // enforced here so the state transition is self-contained (single source of truth).
+        ThresholdParameters::validate_governance_against_reconstruction(
+            new_num_participants,
+            proposal.threshold(),
+            max_reconstruction_threshold(&effective_domains),
+        )?;
```
SimonRastikian, line 189: Here there is a tiny redundancy in checks namely that reconstruction <= num_participants. but otherwise 179-181 checks extra that reconstruction > 2 and that 2*reconstruction-1 < num_participants when in DamgardEtAl.

crates/contract/src/primitives/domain.rs
```diff
@@ -71,6 +71,19 @@ pub fn validate_domain_threshold(
...
+/// (an empty set imposes no cross-domain lower bound on the GovernanceThreshold).
+/// Feeds [`ThresholdParameters::validate_governance_against_reconstruction`](crate::primitives::thresholds::ThresholdParameters::validate_governance_against_reconstruction).
+pub fn max_reconstruction_threshold<'a>(
+    domains: impl IntoIterator<Item = &'a DomainConfig>,
+) -> u64 {
+    domains
+        .into_iter()
+        .map(|domain| domain.reconstruction_threshold.inner())
+        .max()
+        .unwrap_or(0)
```
SimonRastikian: Care the unwrap_or(0). If you have a nicer idea please let me know

crates/contract/src/v3_11_2_state.rs
```diff
@@ -145,6 +148,34 @@ impl From<MpcContract> for crate::MpcContract {
     }
 }
 
+/// One-time pass over migrated state: re-validate the GovernanceThreshold against
+/// the participant count and the largest ReconstructionThreshold under the current
+/// rules. A violation hard-blocks the migration (panic): we refuse to carry state
+/// that breaks the threshold relation into the new contract. Such state must be
+/// corrected (via `vote_new_parameters`) before upgrading.
+///
+/// TODO(#XXXX): remove together with this module once the 3.11.2 -> current
```
SimonRastikian: This todo is on purpose. It will be replaced with a proper issue (to be created) once the PR is approved

SimonRastikian (commented):
The reviewer please double (triple) check that I have covered all the necessary functions with these checks, namely all functions that:
add / remove / kick out participants
update the contract participant set
update the reconstruction threshold
update the governance threshold

kevindeforth (commented):
I am not sure we want to cap the threshold at 80% in the code. 
Makes testing harder and it's not obvious to me why we would impose such a requirement.

crates/contract/src/primitives/thresholds.rs
```diff
@@ -10,6 +10,36 @@ pub use near_mpc_contract_interface::types::Threshold;
...
+const MIN_THRESHOLD_DENOMINATOR: u64 = 5;
+
+/// Maximum fraction of participants the GovernanceThreshold may reach, expressed
+/// as `MAX_THRESHOLD_NUMERATOR / MAX_THRESHOLD_DENOMINATOR`. Currently set to 100%
+/// (5/5), so the relative upper cap is effectively disabled: the GovernanceThreshold
+/// may go all the way up to the participant count (the absolute `k <= n` check still
+/// applies). Kept as an explicit fraction so a stricter cap can be re-introduced by
+/// lowering the numerator should a future policy require it.
+const MAX_THRESHOLD_NUMERATOR: u64 = 5;
+const MAX_THRESHOLD_DENOMINATOR: u64 = 5;
```
SimonRastikian: I insist on having this along with the logic. I am 100% sure this is serving in the very near future, once we decide the upper bound, along with the associated logic

crates/contract/src/primitives/domain.rs
netrome: I think that's correct here. If there are no domains, the max is 0 - makes sense to me.

crates/contract/src/primitives/thresholds.rs
netrome: Sure, keeping this in the code will provoke further discussions on it.

crates/contract/src/primitives/thresholds.rs
```diff
@@ -10,6 +10,36 @@ pub use near_mpc_contract_interface::types::Threshold;
...
+
+/// Lower bound on the GovernanceThreshold for `n` participants: 60% rounded up.
+/// Single source of truth shared by validation and test fixtures.
+pub(crate) fn governance_threshold_lower_bound(n: u64) -> u64 {
+    (MIN_THRESHOLD_NUMERATOR * n).div_ceil(MIN_THRESHOLD_DENOMINATOR)
```
netrome: The multiplication is potentially overflowing. In general we should use checked arithmetic. The `div_ceil` function could panic although it's safe since `MIN_THRESHOLD_DENOMINATOR` won't be zero as it's a constant. But, we should still add a comment about this at least.

crates/contract/src/primitives/thresholds.rs
```diff
@@ -49,17 +84,47 @@ impl ThresholdParameters {
...
+    /// Validates the GovernanceThreshold `k` against both the participant count and the
+    /// largest ReconstructionThreshold across all domains. Layers the cross-domain rule
+    /// `GovernanceThreshold >= max(ReconstructionThreshold)` on top of [`Self::validate_threshold`]:
+    /// the network must never be able to govern with fewer parties than are required to
+    /// reconstruct any domain's key. Call this at every point where the GovernanceThreshold,
+    /// a ReconstructionThreshold, or the participant set changes.
+    pub fn validate_governance_against_reconstruction(
+        num_participants: u64,
+        governance: Threshold,
+        max_reconstruction_threshold: u64,
```
netrome: Why is this not also a `Threshold`? Or even better, a `ReconstructionThreshold`? In the latter case we'd make it harder to accidentally mix up the arguments.

crates/contract/src/v3_11_2_state.rs
netrome: That creates a catch-22 as the PR shouldn't be approved if there are dangling TODOs.

Also we've released 3.12 now, I just created https://github.com/near/mpc/issues/3598 to add support for migrating from 3.12 now.


```suggestion
/// TODO(#3598): remove together with this module once the 3.11.2 -> current
```

netrome (changes requested):
Only hard blocker is the dangling TODO. I would like to see checked arithmetic, but this is equivalent to the previous code and thus not technically a hard blocker.

crates/contract/src/primitives/thresholds.rs
SimonRastikian: You're right, ReconstructionThreshold implements Ord so I could use max. I thought this would not be possible due to  having to extract inner value

SimonRastikian: [7458432](https://github.com/near/mpc/pull/3578/commits/74584321fe521b084f45ae736556fb90a8c070e0)

crates/contract/src/primitives/thresholds.rs
SimonRastikian: [7458432](https://github.com/near/mpc/pull/3578/commits/74584321fe521b084f45ae736556fb90a8c070e0)

netrome (dismissed):
Thanks for updating 🙏

crates/contract/src/v3_11_2_state.rs
gilcu3, line 1: we cannot modify this file as we already published 3.12. This PR can only be merged after https://github.com/near/mpc/pull/3584

docs/design/domain-separation.md
```diff
@@ -343,11 +343,24 @@ pub fn validate_governance(governance: &GovernanceBody) -> Result<(), Error> {
...
+    // bound so the window is never empty) so a stricter cap can be re-introduced later.
+    // See §7.1.
+    let max_relative = (5 * n / 5).max(min_relative);
+    if t > max_relative {
+        return Err(Error::GovernanceThresholdAboveMaximumRelative);
+    }
+    // Cross-domain rule: GovernanceThreshold must dominate every ReconstructionThreshold,
+    // checked wherever governance, a reconstruction threshold, or the participant set
+    // changes (see §7.1). Implemented as
+    // `ThresholdParameters::validate_governance_against_reconstruction`.
```
gilcu3: I dont think keeping this here makes sense for the design doc. We can certainly add it in the future if we decide to have an upper bound.

crates/contract/src/errors.rs
```diff
@@ -183,6 +183,17 @@ pub enum InvalidThreshold {
     MinRelRequirementFailed { required: u64, found: u64 },
     #[error("Threshold must not exceed number of participants: max {max}, found {found}")]
     MaxRequirementFailed { max: u64, found: u64 },
+    #[error(
+        "GovernanceThreshold does not meet the maximum relative requirement: max {max}, found {found}"
```
kevindeforth: nit: this error message would be more helpful if it explained _relative to what_.

docs/design/domain-separation.md
```diff
@@ -919,11 +932,14 @@ let threshold = match dk.protocol {
 
 ## 7. Open Questions
 
-### 7.1 Governance Threshold Validation
+### 7.1 Governance Threshold Validation — RESOLVED (#3499)
+
+Resolved: the `GovernanceThreshold` is now constrained relative to the cryptographic `ReconstructionThreshold`. Concretely, for `n` participants and per-domain reconstruction thresholds `t_i`:
 
-Should the governance `GovernanceThreshold` be constrained relative to the cryptographic `ReconstructionThreshold`? For example, should we require `governance_threshold >= max(reconstruction_threshold for all configs)`?
+- `max(t_i) <= GovernanceThreshold` — a governance majority can never approve a reshare into a set smaller than what any domain needs to reconstruct its key (keeps trust assumptions consistent).
```
gilcu3, line 937: I don't see how this touches "keeps trust assumptions consistent". If we allowed that we would break correctness and liveness, as we would no longer be able to use the domain.

docs/design/domain-separation.md
```diff
@@ -919,11 +932,14 @@ let threshold = match dk.protocol {
...
+- `ceil(0.6*n) <= GovernanceThreshold <= n` — the existing 60% lower bound. A relative upper cap (`MAX_THRESHOLD_NUMERATOR / MAX_THRESHOLD_DENOMINATOR`) is retained in the code but currently set to 100% (5/5 = n), so it is effectively disabled and only the absolute `GovernanceThreshold <= n` check binds. An 80%-floor cap was considered (to stop a minority that stops serving from locking the contract) but not adopted; the fraction is kept explicit so a stricter cap can be re-introduced later.
```
gilcu3: to keep this document clean, we don't need to mention the upper cap until we actually decide if we need it or not

crates/contract/src/lib.rs
```diff
@@ -5461,19 +5422,20 @@ mod tests {
...
-        // Given: 5 participants, governance threshold 3, and one domain whose
-        // reconstruction threshold is 5 (every participant is needed to sign).
+        // Given: 5 participants, GovernanceThreshold 5, and one domain whose
+        // reconstruction threshold is 5 (every participant is needed to sign). Dropping
+        // to 4 participants would leave the GovernanceThreshold above the participant
+        // count, breaking the threshold relation.
         let participants = gen_participants(PARTICIPANT_COUNT);
-        let parameters = ThresholdParameters::new(participants.clone(), Threshold::new(3)).unwrap();
+        let parameters = ThresholdParameters::new(participants.clone(), Threshold::new(5)).unwrap();
         let domain_id = DomainId::default();
```
gilcu3, line 5604: why was this parameter change needed here?

crates/contract/src/lib.rs
```diff
@@ -3989,15 +3923,15 @@ mod tests {
...
+            setup_running_contract_with_domain(5, 4, 2);
+        // ...and a proposal raising the domain's reconstruction threshold to 4,
+        // which fits the 5 participants and does not exceed the GovernanceThreshold.
         let mut per_domain = BTreeMap::new();
-        per_domain.insert(domain_id, ReconstructionThreshold::new(3));
+        per_domain.insert(domain_id, ReconstructionThreshold::new(4));
         let proposal = ProposedThresholdParameters::new(
-            ThresholdParameters::new(participants, Threshold::new(2)).unwrap(),
+            ThresholdParameters::new(participants, Threshold::new(4)).unwrap(),
             per_domain,
```
gilcu3, line 4091: also here it seems changing the numbers was not needed. These changes were probably triggered by the enforcement of the upper bound, which is no longer there so we should simply revert them as well.

crates/contract/src/lib.rs
```diff
@@ -3939,33 +3900,6 @@ mod tests {
...
-        // participants, so the guard rejects it.
-        assert_matches!(
-            result.unwrap_err(),
-            Error::DomainError(DomainError::ReconstructionThresholdExceedsParticipants {
-                threshold: 3,
-                participants: 2,
-            })
-        );
-    }
-
```
gilcu3, line 4101: why are we deleting this test?

crates/contract/src/lib.rs
```diff
@@ -1628,24 +1582,25 @@ impl MpcContract {
...
-                if required > remaining {
+                // Defense in depth: the surviving participant set must keep the full
+                // threshold relation intact — the GovernanceThreshold must still sit
+                // within its bounds for the smaller set (in particular it must not
+                // exceed the remaining participant count or the upper cap) and must
+                // remain at least every domain's ReconstructionThreshold (the kickout
+                // keeps the existing per-domain thresholds). Otherwise we refuse and
+                // wait for manual intervention.
+                let max_reconstruction_threshold =
+                    crate::primitives::domain::max_reconstruction_threshold(
```
gilcu3: better import `max_reconstruction_threshold`

crates/contract/src/lib.rs
```diff
@@ -1628,24 +1582,25 @@ impl MpcContract {
             } => {
                 let threshold = current_params.threshold().value() as usize;
```
gilcu3: why do we need this variable if below we use `current_params.threshold()` directly instead?

crates/contract/src/state/running.rs
```diff
@@ -210,11 +221,20 @@ impl RunningContractState {
         if domains.is_empty() {
             return Err(DomainError::AddDomainsMustAddAtLeastOneDomain.into());
         }
-        let num_participants = self.parameters.participants().len() as u64;
+        let num_participants = u64::try_from(self.parameters.participants().len())
+            .expect("participant list should be wayyyy smaller than u64::MAX");
```
gilcu3: agree, but here we don't need the panic, the function is already returning a `Result`

crates/contract/src/lib.rs
```diff
@@ -886,10 +885,6 @@ impl MpcContract {
             proposal,
         );
 
-        // Defense in depth: never reshare into a participant set smaller than any
-        // threshold (see `assert_proposal_meets_all_thresholds`).
-        self.assert_proposal_meets_all_thresholds(&proposal)?;
-
```
gilcu3, line 896: I understand this check is removed from here because it is redundant, but at the same time not having it here makes it much harder to figure if the check is present at all possible execution branches. I wouldn't mind a small redundancy (if needed at all) to gain confidence that the check is present at all relevant entry points

crates/contract/src/primitives/test_utils.rs
```diff
@@ -158,8 +157,9 @@ pub fn gen_threshold_params(max_n: usize) -> ThresholdParameters {
     // every protocol — `DamgardEtAl` requires `n >= 2t - 1`, which forces
     // `n >= 3` even at the minimum `t = 2`.
     let n: usize = rand::thread_rng().gen_range(3..max_n + 1);
-    let k_min = min_thrershold(n);
-    let k = rand::thread_rng().gen_range(k_min..n + 1);
+    let k_min = governance_threshold_lower_bound(n as u64) as usize;
+    let k_max = governance_threshold_upper_bound(n as u64) as usize;
+    let k = rand::thread_rng().gen_range(k_min..k_max + 1);
```
gilcu3: this is existing, but now that we are modifying it here, we could as well make it deterministic

crates/contract/src/primitives/thresholds.rs
```diff
@@ -10,6 +10,38 @@ pub use near_mpc_contract_interface::types::Threshold;
...
+
+/// Upper bound on the GovernanceThreshold for `n` participants:
+/// `MAX_THRESHOLD_NUMERATOR / MAX_THRESHOLD_DENOMINATOR` floored (currently 100%, i.e.
+/// `n`), clamped up to the lower bound so the feasible window is never empty for small
+/// `n`. Single source of truth shared by validation and test fixtures.
+pub(crate) fn governance_threshold_upper_bound(n: u64) -> u64 {
+    (MAX_THRESHOLD_NUMERATOR.saturating_mul(n) / MAX_THRESHOLD_DENOMINATOR)
+        .max(governance_threshold_lower_bound(n))
+}
+
```
gilcu3: this seems overly verbose. afaik we only need something like:

```
pub(crate) fn governance_threshold_lower_bound(n: u64) -> u64 {
    n.saturating_mul(3).div_ceil(5)
}
```
which is already very readable

crates/contract/src/primitives/domain.rs
```diff
@@ -71,6 +71,18 @@ pub fn validate_domain_threshold(
...
+/// (an empty set imposes no cross-domain lower bound on the GovernanceThreshold).
+/// Feeds [`ThresholdParameters::validate_governance_against_reconstruction`](crate::primitives::thresholds::ThresholdParameters::validate_governance_against_reconstruction).
+pub fn max_reconstruction_threshold<'a>(
+    domains: impl IntoIterator<Item = &'a DomainConfig>,
+) -> Option<ReconstructionThreshold> {
+    domains
+        .into_iter()
+        .map(|domain| domain.reconstruction_threshold)
+        .max()
+}
```
gilcu3: while the current code is not wrong, the following variant is much simpler and still works

```suggestion
pub fn max_reconstruction_threshold(domains: &[DomainConfig]) -> Option<ReconstructionThreshold> {
    domains
        .iter()
        .map(|domain| domain.reconstruction_threshold)
        .max()
}
```

gilcu3 (changes requested):
Thank you for the work!

I have to request a few changes:

- There are many leftovers from the upper bound enforcement, we should clean them up to reduce the noise. Apparently some tests were changed to adapt to that, so it would be better to revert those to smooth the review and ensure we are not losing coverage. If we decide to go with it we can always add it back (adding code is very cheap nowadays).

- The state transition needs to wait for #3598

- Some existing threshold checks that were at the boundary (in the public contract functions) are now deeper in the call stack. This makes it harder to reason about the correctness of those public functions, which is important. Would be good to get other reviewers' input on this.

- We recently decided to keep redundant LLM comments to the minimum (#3589), and here we still have quite a few of those. Would be good to make a pass to reduce them.

crates/contract/src/primitives/thresholds.rs
gilcu3: indeed, it brings too many small changes here and there that make the review harder. In the LLM world, should be easy to remove them and add them back only if we actually decide we want them later

crates/contract/src/primitives/thresholds.rs
```diff
@@ -10,6 +10,38 @@ pub use near_mpc_contract_interface::types::Threshold;
...
+const MIN_THRESHOLD_DENOMINATOR: u64 = 5;
+
+/// Maximum fraction of participants the GovernanceThreshold may reach, expressed
+/// as `MAX_THRESHOLD_NUMERATOR / MAX_THRESHOLD_DENOMINATOR`. Currently set to 100%
+/// (1/1), so the relative upper cap is effectively disabled: the GovernanceThreshold
+/// may go all the way up to the participant count (the absolute `k <= n` check still
+/// applies). Kept as an explicit fraction so a stricter cap can be re-introduced by
+/// lowering the numerator should a future policy require it.
+const MAX_THRESHOLD_NUMERATOR: u64 = 1;
+const MAX_THRESHOLD_DENOMINATOR: u64 = 1;
```
kevindeforth: Lets remove these. If we need it later, it's easy to add. If not, it's better to not have it in the first place.

crates/contract/src/primitives/thresholds.rs
```diff
@@ -10,6 +10,38 @@ pub use near_mpc_contract_interface::types::Threshold;
...
+        .div_ceil(MIN_THRESHOLD_DENOMINATOR)
+}
+
+/// Upper bound on the GovernanceThreshold for `n` participants:
+/// `MAX_THRESHOLD_NUMERATOR / MAX_THRESHOLD_DENOMINATOR` floored (currently 100%, i.e.
+/// `n`), clamped up to the lower bound so the feasible window is never empty for small
+/// `n`. Single source of truth shared by validation and test fixtures.
+pub(crate) fn governance_threshold_upper_bound(n: u64) -> u64 {
+    (MAX_THRESHOLD_NUMERATOR.saturating_mul(n) / MAX_THRESHOLD_DENOMINATOR)
+        .max(governance_threshold_lower_bound(n))
```
kevindeforth: Happy to keep this method, but it can be simplified:

```rust
    n.max(governance_threshold_lower_bound(n))
```

crates/contract/src/primitives/thresholds.rs
```diff
@@ -10,6 +10,38 @@ pub use near_mpc_contract_interface::types::Threshold;
...
+pub(crate) fn governance_threshold_lower_bound(n: u64) -> u64 {
+    MIN_THRESHOLD_NUMERATOR
+        .saturating_mul(n)
+        .div_ceil(MIN_THRESHOLD_DENOMINATOR)
+}
+
+/// Upper bound on the GovernanceThreshold for `n` participants:
+/// `MAX_THRESHOLD_NUMERATOR / MAX_THRESHOLD_DENOMINATOR` floored (currently 100%, i.e.
+/// `n`), clamped up to the lower bound so the feasible window is never empty for small
+/// `n`. Single source of truth shared by validation and test fixtures.
```
kevindeforth: This is comment is not helpful. Please take a look at our [guidelines](https://github.com/near/mpc/blob/main/docs/engineering-standards.md#write-helpful-code-comments) on what constitutes a good / bad code comment.

kevindeforth (commented):
I agree with the comments expressed in @gilcu3's [review](https://github.com/near/mpc/pull/3578#pullrequestreview-4512897629). I think some things can be simplified, which will make review easier.

Also, please take a look at our engineering standards on permissible code comments:

crates/contract/src/primitives/thresholds.rs
SimonRastikian: [f168468](https://github.com/near/mpc/pull/3578/commits/f16846828edbf4c2b2556caf009118d4b2c2ae55)

crates/contract/src/primitives/domain.rs
SimonRastikian: [cd557e1](https://github.com/near/mpc/pull/3578/commits/cd557e14a1df8fecef9f76f9257e5c96f948252a)

crates/contract/src/v3_11_2_state.rs
SimonRastikian: [e769a53](https://github.com/near/mpc/pull/3578/commits/e769a538ade301dbee6416568f06da46c9258ae6)

docs/design/domain-separation.md
SimonRastikian: [df26736](https://github.com/near/mpc/pull/3578/commits/df267368b210861c2c997c1ef155e7bc5ca8ea71)

docs/design/domain-separation.md
SimonRastikian: If you assume on the cryptography side that you have 5 malicious participants but on the governance you assume only 2 malicious parties than this is not consistent in the sense that 2 malicious parties can drop the crypto threshold to 2 by voting themselves to change the threshold...

docs/design/domain-separation.md
SimonRastikian: I would still keep it but compacted the paragraph [876bcfe](https://github.com/near/mpc/pull/3578/commits/876bcfec2496a9afd7ab52d6f43428d3a50eae34)

crates/contract/src/lib.rs
SimonRastikian: because govthreshold >= max Reconstructionthreshold which here is 5

crates/contract/src/lib.rs
SimonRastikian: Also not: the test would fail as the reconstruction threshold would be set to 3 > governance =2

crates/contract/src/lib.rs
SimonRastikian: [484d40e](https://github.com/near/mpc/pull/3578/commits/484d40ed59e17344af0cd274e53dc126aca5cd98)

crates/contract/src/lib.rs
SimonRastikian: This is a redundant test now to should_reject_when_shrinking_below_unchanged_domain_threshold.
Both end up asserting the same ReconstructionThresholdExceedsParticipants error from the same validate_domain_threshold helper(one via over-large update and  the other via shrinking participants).

crates/contract/src/lib.rs
SimonRastikian: [ec871da](https://github.com/near/mpc/pull/3578/commits/ec871da602d4508df817a5c004020157ef6bb33b)

crates/contract/src/state/running.rs
SimonRastikian: [ee76a11](https://github.com/near/mpc/pull/3578/commits/ee76a110b9b56c9a23bc5b92127c0d33bd9fcf10)

crates/contract/src/primitives/test_utils.rs
SimonRastikian: [6a4fa9e](https://github.com/near/mpc/pull/3578/commits/6a4fa9eb9a651495c725b2fe1260de550e6c9ec3)

crates/contract/src/primitives/thresholds.rs
SimonRastikian: Done

crates/contract/src/primitives/thresholds.rs
SimonRastikian: I returned simply n. The ones who change it to a different cap shall be careful not making mistakes

crates/contract/src/primitives/thresholds.rs
SimonRastikian: Reworked the comment.
Thanks

crates/contract/src/errors.rs
SimonRastikian: [0ef09e5](https://github.com/near/mpc/pull/3578/commits/0ef09e5a2c70b997f0b643851afffd239ee58c9a)

crates/contract/src/primitives/thresholds.rs
```diff
@@ -33,12 +47,17 @@ impl ThresholdParameters {
...
     /// That is:
     /// - threshold must be at least `MIN_THRESHOLD_ABSOLUTE`
     /// - threshold can not exceed the number of shares `n_shares`.
     /// - threshold must be at least 60% of the number of shares (rounded upwards).
-    pub fn validate_threshold(n_shares: u64, k: Threshold) -> Result<(), Error> {
+    /// - threshold must not exceed `MAX_THRESHOLD_NUMERATOR / MAX_THRESHOLD_DENOMINATOR`
+    ///   of the number of shares (rounded downwards), clamped up to the 60% lower bound so
+    ///   the feasible window is never empty for small `n_shares`. This relative upper cap is
+    ///   currently set to 100%, so in practice it never binds below the absolute `k <= n`
+    ///   check above.
```
kevindeforth: stale commit

crates/contract/src/primitives/thresholds.rs
```diff
@@ -10,6 +10,20 @@ pub use near_mpc_contract_interface::types::Threshold;
 /// Minimum absolute threshold required.
 const MIN_THRESHOLD_ABSOLUTE: u64 = 2;
 
+/// Lower bound on the GovernanceThreshold for `n` participants: 60% rounded up.
+/// Single source of truth shared by validation and test fixtures.
+pub(crate) fn governance_threshold_lower_bound(n: u64) -> u64 {
```
kevindeforth: nit:

```suggestion
pub(crate) fn governance_threshold_lower_relative_bound(n: u64) -> u64 {
```

crates/contract/src/primitives/thresholds.rs
```diff
@@ -10,6 +10,20 @@ pub use near_mpc_contract_interface::types::Threshold;
...
+    3_u64.saturating_mul(n).div_ceil(5)
+}
+
+/// Upper bound on the GovernanceThreshold for `n` participants:
+/// Currently set to 100% of participants but would be a discussion subject
+/// to drop this upper bound down not to have problems with smart contract
+/// being locked if t = n and if an operator stops voting
+pub(crate) fn governance_threshold_upper_bound(n: u64) -> u64 {
```
kevindeforth: nit:

```suggestion
pub(crate) fn governance_threshold_upper_relative_bound(n: u64) -> u64 {
```

crates/contract/src/state/resharing.rs
```diff
@@ -416,12 +416,15 @@ pub mod tests {
...
+        // Threshold valid for the grown set (1.5x): old_len sits within
+        // [ceil(0.6 * 1.5n), 1.5n].
         let new_threshold = Threshold::new(old_participants.len() as u64);
         new_participants_1.add_random_participants_till_n((old_participants.len() * 3).div_ceil(2));
         let new_participants_2 = new_participants_1
             .subset(new_participants_1.len() - old_participants.len()..new_participants_1.len());
         let new_params_1 = ThresholdParameters::new(new_participants_1, new_threshold).unwrap();
-        let new_params_2 = ThresholdParameters::new(new_participants_2, new_threshold).unwrap();
+        let new_threshold_2 = Threshold::new((3 * new_participants_2.len() as u64).div_ceil(5));
+        let new_params_2 = ThresholdParameters::new(new_participants_2, new_threshold_2).unwrap();
```
kevindeforth: Q: why do we need to change the threshold to `new_threshold_2` here?

crates/contract/src/state/resharing.rs
```diff
@@ -569,27 +562,23 @@ pub mod tests {
...
     #[test]
     fn vote_reshared__final_transition__should_apply_distinct_thresholds_per_domain() {
-        // Given CaitSith (index 0) and Frost (index 1) domains, both at default
-        // threshold 2, and a proposal moving only the Frost domain to `n`.
+        // Given CaitSith (index 0) and Frost (index 1) at default threshold 2, and a
+        // proposal moving only Frost to the GovernanceThreshold.
         let mut env = Environment::new(Some(100), None, None);
         let mut running = gen_running_state(2);
+        running.parameters =
+            ThresholdParameters::new(gen_participants(5), Threshold::new(4)).unwrap();
```
kevindeforth: Q: why is this change necessary for this PR?

crates/contract/src/state/running.rs
```diff
@@ -210,11 +221,24 @@ impl RunningContractState {
         if domains.is_empty() {
             return Err(DomainError::AddDomainsMustAddAtLeastOneDomain.into());
         }
-        let num_participants = self.parameters.participants().len() as u64;
+        let num_participants =
+            u64::try_from(self.parameters.participants().len()).map_err(|e| {
+                ConversionError::DataConversion {
+                    reason: format!("participant count does not fit in u64: {e}"),
```
kevindeforth: we can use .expect() for these. Converting usize to u64 is expected to succeed.

crates/contract/src/state/running.rs
```diff
@@ -605,7 +674,11 @@ pub mod running_tests {
     fn vote_add_domains__should_accept_caitsith_threshold_differing_from_existing() {
         // Given a Running state already holding a CaitSith domain at t = 2
         // (the fixture default) and a proposal for a second CaitSith at t = 3.
+        // Require GovernanceThreshold >= 3 so a reconstruction threshold of 3 is allowed.
         let mut state = gen_running_state(1);
+        while state.parameters.threshold().value() < 3 {
+            state = gen_running_state(1);
+        }
```
kevindeforth: I guess this works xD.

`state` is mut, we can also modify the governance threshold.

crates/contract/src/state/running.rs
```diff
@@ -637,7 +712,11 @@ pub mod running_tests {
     fn vote_add_domains__should_accept_two_new_caitsith_with_differing_thresholds() {
         // Given a Running state with no existing CaitSith and a proposal
         // adding two CaitSith domains at different thresholds.
+        // Require GovernanceThreshold >= 3 so reconstruction thresholds 2 and 3 are both allowed.
         let mut state = gen_running_state(0);
+        while state.parameters.threshold().value() < 3 {
+            state = gen_running_state(0);
+        }
```
kevindeforth: Idem. Why don't we create a custom method for generating a running state with a minimum govrenance threshold?

crates/contract/src/state/running.rs
```diff
@@ -695,6 +774,10 @@ pub mod running_tests {
         // Given a running state with one CaitSith domain at the fixture default
         // t = 2.
         let mut state = gen_running_state(1);
+        // Pin a participant set so the generated proposal's GovernanceThreshold is >= 3;
+        // ReconstructionThreshold (3) must not exceed it.
+        state.parameters =
+            ThresholdParameters::new(gen_participants(5), Threshold::new(4)).unwrap();
```
kevindeforth: yeah, so given all of this, I think we may want to extend `gen_running_state` to take as input number of participants and governance thresholds.

crates/contract/src/lib.rs
```diff
@@ -1709,26 +1663,24 @@ impl MpcContract {
...
+                // threshold relation intact — the GovernanceThreshold must still sit
+                // within its bounds for the smaller set (in particular it must not
+                // exceed the remaining participant count or the upper cap) and must
+                // remain at least every domain's ReconstructionThreshold (the kickout
+                // keeps the existing per-domain thresholds). Otherwise we refuse and
+                // wait for manual intervention.
+                let max_reconstruction_threshold =
+                    max_reconstruction_threshold(running_state.domains.domains());
+                if let Err(err) = ThresholdParameters::validate_governance_against_reconstruction(
+                    remaining as u64,
```
kevindeforth: nit: use an explicit conversion and `.expect`

crates/contract/src/lib.rs
```diff
@@ -1743,7 +1695,7 @@ impl MpcContract {
                 //let n_participants_new = new_participants.len();
                 //let new_threshold = (3 * n_participants_new + 4) / 5; // minimum 60%
                 //let new_threshold = new_threshold.max(2); // but also minimum 2
-                let new_threshold = threshold;
+                let new_threshold = current_params.threshold().value() as usize;
```
kevindeforth: idem

crates/contract/src/lib.rs
```diff
@@ -5597,19 +5555,20 @@ mod tests {
...
 
-        // Given: 5 participants, governance threshold 3, and one domain whose
-        // reconstruction threshold is 5 (every participant is needed to sign).
+        // Given: 5 participants, GovernanceThreshold 5, and one domain whose
+        // reconstruction threshold is 5 (every participant is needed to sign). Dropping
+        // to 4 participants would leave the GovernanceThreshold above the participant
+        // count, breaking the threshold relation.
         let participants = gen_participants(PARTICIPANT_COUNT);
-        let parameters = ThresholdParameters::new(participants.clone(), Threshold::new(3)).unwrap();
+        let parameters = ThresholdParameters::new(participants.clone(), Threshold::new(5)).unwrap();
```
kevindeforth: nit:
```suggestion
        let parameters = ThresholdParameters::new(participants.clone(), Threshold::new(PARTICIPANT_COUNT)).unwrap();
```

crates/contract/src/v3_12_0_state.rs
```diff
@@ -69,3 +78,22 @@ impl From<MpcContract> for crate::MpcContract {
...
+        num_participants,
+        running.parameters.threshold(),
+        max_reconstruction_threshold,
+    ) {
+        env::panic_str(&format!(
+            "Migration aborted: existing state violates the GovernanceThreshold/ReconstructionThreshold relation ({err:?}). num_participants={}, governance_threshold={}, max_reconstruction_threshold={:?}. Correct it via vote_new_parameters before upgrading.",
+            num_participants,
+            running.parameters.threshold().value(),
+            max_reconstruction_threshold.map(|t| t.inner()),
+        ));
```
kevindeforth, line 95: Hmm, can't we recover from some inconsistent states by raising the govenance threshold?

kevindeforth (commented):
Thank you very much! 
I only have a few small things, main one is to update some of the unit test helpers, such that they stay relevant and useufl.

crates/contract/src/state/running.rs
SimonRastikian: [e72e801](https://github.com/near/mpc/pull/3578/commits/e72e801b7f8e9885c60fbbd0b22390e6d36069eb)

crates/contract/src/v3_12_0_state.rs
SimonRastikian: Not sure this will be the right solution as raising the gov threshold would impact the entire smart contract. I would be happy to keep the fix manual after potential discussions.

crates/contract/src/primitives/thresholds.rs
SimonRastikian: [7d46e93](https://github.com/near/mpc/pull/3578/commits/7d46e936050dfe2f29a1294ab1392b148aac489b)

crates/contract/src/state/running.rs
SimonRastikian: I think this is a dead code.. will update it

crates/contract/src/state/running.rs
SimonRastikian: Updated

SimonRastikian: [187de2a](https://github.com/near/mpc/pull/3578/commits/187de2a7efa9fe858677140ae10061f80a1e8efc)

crates/contract/src/state/running.rs
SimonRastikian: [187de2a](https://github.com/near/mpc/pull/3578/commits/187de2a7efa9fe858677140ae10061f80a1e8efc)

crates/contract/src/state/running.rs
SimonRastikian: I created a wrapper to gen_running_state as the latter is called over 50+ times.
[37465bd](https://github.com/near/mpc/pull/3578/commits/37465bd211ee120230910131d52e2970b39a1076)

crates/contract/src/state/resharing.rs
SimonRastikian: Reverted. This was due to adding an upper cap. [c9ab983](https://github.com/near/mpc/pull/3578/commits/c9ab9837baa5109b87dd61adc38f6d1f5d49decb)

crates/contract/src/state/resharing.rs
SimonRastikian: This is due to the newly inforced variant GovThreshold> max(ReconstructionThreshold). The gen_running_state(2) can land in [ceil(0.6n), n] and can be below n, so reconstruction  > governance would be rejected,

crates/contract/src/primitives/test_utils.rs
```diff
@@ -157,9 +156,11 @@ pub fn gen_threshold_params(max_n: usize) -> ThresholdParameters {
     // Lower bound is 3 (not 2) so the produced parameters are compatible with
     // every protocol — `DamgardEtAl` requires `n >= 2t - 1`, which forces
     // `n >= 3` even at the minimum `t = 2`.
-    let n: usize = rand::thread_rng().gen_range(3..max_n + 1);
-    let k_min = min_thrershold(n);
-    let k = rand::thread_rng().gen_range(k_min..n + 1);
+    let mut rng = StdRng::seed_from_u64(42);
```
gilcu3, line 159: technically this should be a parameter to the function, but for the sake of keeping this simple we can leave it here

crates/contract/src/v3_12_0_state.rs
```diff
@@ -69,3 +78,22 @@ impl From<MpcContract> for crate::MpcContract {
         }
     }
 }
+
+/// TODO(#3598): remove together with this module once the 3.12.0 -> current
+/// migration is retired.
+fn validate_threshold_relation_on_migration(running: &RunningContractState) {
```
gilcu3: I believe this TODO is wrong, as https://github.com/near/mpc/issues/3598 is already closed. Generally we don't add TODOs to remove code from this file, because it must always happen, but feel free to add an issue for the post 3.13 release migration cleanup, which would then fit here

crates/contract/src/lib.rs
gilcu3: I believe the two tests are not equivalent, appear to touch a different path of execution, so please restore this one. That they assert the same error does not mean they cover the same cases

crates/contract/src/state/test_utils.rs
```diff
@@ -84,8 +87,32 @@ pub fn gen_resharing_state(num_domains: usize) -> (Environment, ResharingContrac
...
+) -> RunningContractState {
+    let parameters = ThresholdParameters::new(
+        gen_participants(num_participants),
+        Threshold::new(governance_threshold),
+    )
+    .expect("valid threshold parameters");
+    gen_running_state_with_threshold_params(num_domains, parameters)
+}
+
+fn gen_running_state_with_threshold_params(
```
gilcu3: nit: it seems sufficient here to have two functions, `gen_running_state` and `gen_running_state_with_params`, because the second one should be enough to implement the first, without needing a third function `gen_running_state_with_threshold_params`

gilcu3 (changes requested):
Thank you!

Just a couple of minor changes requested

- There is a stale TODO
- There is a removed test that seems to cover an execution path no longer covered by other tests, see this [comment](https://github.com/near/mpc/pull/3578/changes#r3452925562) (not sure the link will work)

crates/contract/src/v3_12_0_state.rs
SimonRastikian: Yep sure I will remove the TODO

SimonRastikian: As you said it's useless

crates/contract/src/primitives/test_utils.rs
SimonRastikian: As you said, it's simpler this way

crates/contract/src/v3_12_0_state.rs
SimonRastikian: [afbe6b8](https://github.com/near/mpc/pull/3578/commits/afbe6b827cfbcf143cd6cc94e07f8b96dffaf078)

crates/contract/src/state/test_utils.rs
SimonRastikian: [f455d91](https://github.com/near/mpc/pull/3578/commits/f455d91f5e597826e25d655aa8caf0252403b8fb)

crates/contract/src/state/running.rs
```diff
@@ -891,7 +891,11 @@ pub mod running_tests {
     fn vote_add_domains__should_accept_non_caitsith_domain_with_differing_threshold() {
         // Given a Running state with a CaitSith domain at t = 2 and a Frost
         // proposal at a different threshold.
+        // Require GovernanceThreshold >= 3 so a reconstruction threshold of 3 is allowed.
         let mut state = gen_running_state(1);
+        while state.parameters.threshold().value() < 3 {
+            state = gen_running_state(1);
+        }
```
gilcu3: oh this is not a nice hack. We should be able to avoid it by calling `gen_running_state_with_params` instead

gilcu3 (dismissed):
Thank you!

left only a nit (over a hack)

crates/contract/src/state/running.rs
SimonRastikian: [0d6e0f6](https://github.com/near/mpc/pull/3578/commits/0d6e0f69c1b151b9c18887927fdc6bc823183c33)

crates/contract/src/primitives/thresholds.rs
```diff
@@ -49,17 +64,49 @@ impl ThresholdParameters {
...
+    /// Validates the GovernanceThreshold `k` against both the participant count and the
+    /// largest ReconstructionThreshold across all domains. Layers the cross-domain rule
+    /// `GovernanceThreshold >= max(ReconstructionThreshold)` on top of `validate_threshold`:
+    /// the network must never be able to govern with fewer parties than are required to
+    /// reconstruct any domain's key. Call this at every point where the GovernanceThreshold,
+    /// a ReconstructionThreshold, or the participant set changes.
+    pub fn validate_governance_against_reconstruction(
+        num_participants: u64,
+        governance: Threshold,
+        max_reconstruction_threshold: Option<ReconstructionThreshold>,
```
kevindeforth, line 95: Note (non-blocking): I don't like this being optional. I would prefer this to be a u64.
Reason for this is that as a reviewer, I get more suspicious when I see a function call passing a magical 0, compared to a function passing a magical `None`.

crates/contract/src/primitives/thresholds.rs
```diff
@@ -295,6 +354,55 @@ mod tests {
...
+        // The relative upper cap is clamped up to the ceil(0.6n) lower bound, so the
+        // feasible window must always hold at least one valid threshold.
+        for n in 2..=12u64 {
+            let lower = governance_threshold_lower_relative_bound(n);
+            let upper = governance_threshold_upper_relative_bound(n);
+            assert!(upper >= lower, "empty window at n={n}: [{lower}, {upper}]");
+            // The clamped boundary value must validate.
+            ThresholdParameters::validate_threshold(n, Threshold::new(upper)).unwrap();
+        }
+    }
```
kevindeforth, line 368: Note: This test seems redundant, as it is covered by `test_validate_threshold`, but we can keep it.

crates/contract/src/primitives/thresholds.rs
```diff
@@ -295,6 +354,55 @@ mod tests {
...
+            n,
+            governance,
+            Some(ReconstructionThreshold::new(6)),
+        )
+        .unwrap();
+        ThresholdParameters::validate_governance_against_reconstruction(
+            n,
+            governance,
+            Some(ReconstructionThreshold::new(5)),
+        )
```
kevindeforth, line 402: Note: this test would be nicer if it looped over a bunch of values instead of just one.

crates/contract/src/primitives/test_utils.rs
kevindeforth: I don't think this is an improvement without passing the RNG as a parameter, because this is a pretty low-level helper method, so we will now generate the exact same threshold params for any given `n` for most of our tests.

@SimonRastikian please modify in this PR or open a follow-up for this.

crates/contract/src/state/running.rs
```diff
@@ -623,8 +680,10 @@ pub mod running_tests {
         let mut state = gen_running_state(0);
         let mut env = Environment::new(None, None, None);
         env.set_signer(&state.parameters.participants().participants()[0].0);
-        let n = state.parameters.participants().len() as u64;
-        let proposal = single_domain_proposal(&state, Protocol::CaitSith, DomainPurpose::Sign, n);
+        // Use the GovernanceThreshold as the ReconstructionThreshold (the maximum allowed).
```
kevindeforth, line 683: Do our tests pass if we have more variety here and select a threshold in `[2, governance_threshold]`?

crates/contract/src/state/test_utils.rs
```diff
@@ -84,8 +87,31 @@ pub fn gen_resharing_state(num_domains: usize) -> (Environment, ResharingContrac
         resharing_state.expect("Should've transitioned into resharing"),
     )
 }
-/// Generates a Running state that contains this many domains.
+/// Generates a Running state that contains this many domains, with randomly
+/// generated threshold parameters.
```
kevindeforth, line 91: Please update the doc comment to reflect the new behavior `gen_threshold_params` will output the exact same threshold for the same `n`, it's not really "random" anymore.

crates/contract/src/state/test_utils.rs
```diff
@@ -84,8 +87,31 @@ pub fn gen_resharing_state(num_domains: usize) -> (Environment, ResharingContrac
...
+    let parameters = gen_threshold_params(max_n);
+    gen_running_state_with_params(
+        num_domains,
+        parameters.participants().len(),
+        parameters.threshold().value(),
+    )
+}
+
+/// Like [`gen_running_state`], but pins the participant count and
+/// GovernanceThreshold instead of randomizing them.
```
kevindeforth, line 103: Note: It's overhead for the reader to look-up what `gen_running_state` does.

kevindeforth (approved):
Thank you, looks good!

I have one concern about making a previously randomized helper method non-random, which means that many tests now us the exact same setup. Needs to be addressed, but can be done in a follow-up.

gilcu3 (approved):

PR #3580, opened 2026-06-16
fix: repro issues related to docker caching and skopeo versions

### LINKED ISSUE #3579: Reproducible docker build bakes stale context and couples skopeo to the docker daemon
### Background
On a developer machine, `deployment/build-images.sh` produced a node manifest digest that diverged from CI (`sha256:5ab74b6f…` vs `sha256:0f3e5721…`) even though the node binary hash matched. Two root causes in the script:

1. **Stale build context.** buildkit detects local-context changes by `(size, mtime)`, but the script's `find . -exec touch -d @"$SOURCE_DATE_EPOCH"` resets every mtime to `0`. A file whose content changed but size did not (here `deployment/localnet/genesis.json`, 3184 bytes in two versions) is seen as unchanged, so a stale cached copy is baked in. `--no-cache` does not help — it disables the layer cache, not the context filesync cache.
2. **skopeo coupled to the docker daemon.** The image was read via the `docker-daemon:` transport, tying the skopeo version to the host daemon's API. skopeo is not pinned — CI currently uses 1.13.3 (Ubuntu 24.04 apt) — and newer daemons are incompatible with that version, so it can't read the image from the daemon.

### Acceptance Criteria
- `build-images.sh` prunes only the local-context cache (`--filter type=source.local`) before building, forcing a fresh context read while preserving base-image/apt caches.
- `build-images.sh` builds to a `docker-archive` tar and compresses it via the `docker-archive:` transport, so skopeo never reads from the docker daemon.

gilcu3:
Closes #3579 

Note: I left the relatively long comments here because I think they are essential to understand and justify the added steps

We are not yet pinning the version of skopeo, as it is linked to the Ubuntu version anyway. Might be needed in the future though.

### DIFF
--- a/deployment/build-images.sh
+++ b/deployment/build-images.sh
@@ -100,28 +100,44 @@ buildkit_image_name="buildkit_${buildkit_version}"
 
 if ! docker buildx inspect ${buildkit_image_name} &>/dev/null; then
     docker buildx create --use --driver-opt image=moby/buildkit:v${buildkit_version} --name ${buildkit_image_name}
+else
+    # A reused builder may hold a stale local-context cache: buildkit keys
+    # context changes on (size, mtime), but the touch above resets mtime, so a
+    # changed file with unchanged size looks identical and the stale copy is
+    # reused. Prune only type=source.local to force a fresh read (base-image and
+    # apt caches are kept). A freshly created builder has no cache, so we only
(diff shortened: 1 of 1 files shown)


netrome (approved):
Thanks!

pbeza (approved):

PR #3584, opened 2026-06-16
chore: bump up state transition to be from 3_12_0

### LINKED ISSUE #3598: Support migrating the contract from 3.12.0
### Background

Currently, the contract supports migrating from 3.11.2. We've now released 3.12.0 so we should change the contract to support migrating from the 3.12.0 state, and we may drop support for migrating from 3.11.2.

### User Story

As an MPC operator I'd like to be able to upgrade the MPC contract from the latest deployed version when a new release is made.

### Acceptance Criteria

The contract can be upgraded from 3.12.0

_No response_

anodar:
Closes #3598


crates/contract-history/archive/signer-3_12_0.wasm
SimonRastikian, line 1: Is there a way to make sure from here that this file is indeed the one in the release?

SimonRastikian (dismissed):
Thank you

crates/contract-history/archive/signer-3_12_0.wasm
anodar: Yes we could, filed an issue for it with low priority: https://github.com/near/mpc/issues/3610

gilcu3: we already have that, there is a job in main that downloads the deployed wasm and compares, so I think #3610 is not needed

anodar: Are you referring to `external-services-tests.yml` ? If yes, that compares deployed wasm against commited wasm we have in archive right? But I don't see any test comparing released wasm agains archived one. Can you point me to it if that exists?

gilcu3: ahh sorry I thought you meant he same. Still, for released wasms do we need it? `external-services-tests`  seems to cover the case we care about for migrations, because we guarantee manually that deployed contract and released contract is the same. Although it is true that this extra test would allow early handling of this.

crates/contract/src/v3_12_0_state.rs
```diff
@@ -0,0 +1,65 @@
+//! ## Overview
+//! This module stores the previous contract state—the one you want to migrate from.
+//! The goal is to describe the data layout _exactly_ as it existed before.
+//!
+//! ## Guideline
+//! In theory, you could copy-paste every struct from the specific commit you're migrating from.
+//! However, this approach (a) requires manual effort from a developer and (b) increases the binary size.
+//! A better approach: only copy the structures that have changed and import the rest from the existing codebase.
+//!
+//! The 3.12.0 layout is identical to the current layout, so no shadow types are needed here.
```
gilcu3: nit: this line is not needed here, as it is only true for this PR

gilcu3 (changes requested):
Looks good overall but we are using the wrong contract

Notice for release 3.12 the one in the Release is wrong, we have a warning about it on top, the correct one should have digest `4f72fa9d73497fd1789c4d70860ebfbb6d6251ab77cc0ff8a59fd779a3a3c152`

crates/contract/src/v3_12_0_state.rs
anodar: yeah it was meant specifically for this PR explaining why we don't have any Old* struct. Dropped it in [ef04a97](https://github.com/near/mpc/pull/3584/commits/ef04a9798138cea3597b105aa4438f0e2f353549)

gilcu3 (dismissed):
Thank you!

gilcu3 (approved):

crates/contract-history/archive/signer-3_12_0.wasm
anodar: Yeah it would detect it earlier, it could be a test that once you're pushing new wasm into this archive folder, it will check that hash actually matches released one (which isn't the case here in this particular pr :D but generally it should).

SimonRastikian (approved):

PR #3588, opened 2026-06-16
chore: allow new foreign chain config before node upgrade

### LINKED ISSUE #3587: Don't fail node if foreign chain config has unknown fields
### Background

Currently the node would fail to start if foreign chain config contains unknown field: https://github.com/near/mpc/blob/8dab0ad20f0906388452bdb5b3d5070ea81438e5/crates/node-config/src/foreign_chains.rs#L17

We shouldn't be blocked on updating config if node doesn't yet support new chain.

### User Story

As an MPC team engineer, when working on supporting new chain I have to make sure we first roll out support for new chain and only then update config. 

Config update should not be blocked by node upgrade.

### Acceptance Criteria

Allow foreign chain config to contain new chain that's not yet supported in the node.

_No response_

anodar:
Closes: https://github.com/near/mpc/issues/3587

### DIFF
--- a/crates/node-config/src/foreign_chains.rs
+++ b/crates/node-config/src/foreign_chains.rs
@@ -14,8 +14,6 @@ pub use auth::{AuthConfig, TokenConfig};
 mod auth;
 
 #[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
-#[serde(deny_unknown_fields)]
-// TODO(#3002): only keep variants that are actually supported by the node binary
 pub struct ForeignChainsConfig {
     #[serde(default, skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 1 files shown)


crates/node-config/src/foreign_chains.rs
```diff
@@ -14,8 +14,6 @@ pub use auth::{AuthConfig, TokenConfig};
 mod auth;
 
 #[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
-#[serde(deny_unknown_fields)]
-// TODO(#3002): only keep variants that are actually supported by the node binary
```
anodar, line 18: This issue was closed as not planned, so dropping it here.

netrome (approved):

gilcu3 (approved):
Thank you!

PR #3589, opened 2026-06-16
docs: update engineering standards with section about code comments

### LINKED ISSUE #3581: update contribution guidelines with code comment quality
### Background

We have much more comments in our code base than we used to. This is likely due to LLMs making the generation of code comments much lower cost compare to two years ago.

The problem is that these comments often lack quality. They:
* describe how the piece of code that is being documented is used in other parts of the code-base, in ways that are not relevant to the component that is being documented ([^1] and [^2]);
* are overly verbose, contain repetitions or tautologies; ([^1] and [^3])
* burden the reader with unnecessary context they wouldn't even think of or don't need to be concerned about ([^1] and [^4]).
* describe something that the code self-documents (often, _what_ a function _does_ or _what data_ a struct holds ([^1] and [^2]). (Note: the exception here is public API methods that often benefit from such comments).

As a reviewer, it's not very enjoyable to read these. More often than not, they are a distraction, not a help.

A good comment:
* is concise;
* describes something that is not obvious from reading the code, such as invariants or constraints;
* draws in external context *only to the extent required and useful to the reader*
* doesn't risk getting stale (if it is temporary behavior, protect it with a `TODO(#issueNumber)`).

[^1]:
    ```rust
    /// returns None if the result is zero
    fn add_mod_10(x: MyCustomNonZeroUsize, y: MyCustomNonZeroUsize) -> Option<MyCustomNonZeroUsize>
    ```
    This is good enough. We don't need to have an explanation:
    ```rust
    /// Adds `y` and `x` modulo 10
    ///
    /// Returns `None` when the result is zero — i.e. whenever `x + y` is a
    /// multiple of 10 (`1 + 9`, `5 + 5`, ...) — since the result type cannot
    /// represent zero. Returns `Some` with the mod-10 sum otherwise.
    ///
    /// Helper used by `this_other_method` to implement decimal-digit arithmetic for `MyCustomNonZeroUsize`.
    /// Kept separate to make the carry logic easier to test.
    fn add_mod_10(x: MyCustomNonZeroUsize, y: MyCustomNonZeroUsize) -> Option<MyCustomNonZeroUsize>
    ```

[^2]:
    ```rust
    pub struct MyStruct {
        x: MyOtherStructX,
        y: MyOtherStructY,
    }
    ```
    This does *not* require the following explanation:
    ```rust
    /// Shared immutably across the codebase: `ThatOtherStruct` holds this
    /// behind an `Arc`, so instances are treated as read-only after
    /// construction and cloned cheaply by bumping the refcount rather than
    /// by deep-copying.
    pub struct MyStruct {
        /// An instance of `MyOtherStructX`
        x: MyOtherStructX,
        /// An instance of `MyOtherStructY`
        y: MyOtherStructY,
    }
    ```

[^3]:
    Consider this:

    ```rust
    /// idempotent
    fn add_the_thing(self, thing: Thing) {
    ```
    compared to this:
    ```rust
    /// Adds `thing`.
    ///
    /// Idempotent in `thing`: adding the same `thing` more than once has the
    /// same effect as adding it once. The second add is a no-op / overwrites
    /// the existing entry with an equal value — so callers can re-add without
    /// first checking whether it's already present.
    fn add_the_thing(self, thing: Thing) {
    ```

    The latter does not add any value to the reader (modulo saving a google search or an LLM prompt in case the reader does not know what "idempotent" means, but a code-base is not a lecture book.)

[^4]:
    All too often, claude tries to add brainstorming notes to the code. If I debate whether it makes more sense to create component `ComponentVersionA` or `ComponentVersionB` with differing properties, and then, we end up picking one of the two, I can be sure it tries to add a comment of the sort:

    ```rust
    /// This struct achieves <xyz>, using <insert_version_b_specific_way_of_doing_the_thing> instead of <insert_version_a_specific_way_of_doing_the_thing>.
    struct ComponentVersionB
    ```
    This sort of comment is not helpful. It explains _what_, but not _why_. If it's obvious that there are more than one ways to achieve the code and tradeoffs are non-obvious, then it makes sometimes sense to explain the _why_:
    ```rust
    // This struct achieves <xyz>, using <insert_version_b_specific_way_of_doing_the_thing>. We cold also have done it <insert_two_sentences_on_version_b>, but rejected that, because of <explain_invariant_or_constraint_from_codebase_and_explain_tradeoff_and_why_this_is_better>.
    struct ComponentVersionB
    ```

### User Story

As a dev and LLM, I like clear contribution guidelines.

### Acceptance Criteria

Add the above to our contribution guidelines and have our claude review prompt check for code comment quality.

[slack thread](https://nearone.slack.com/archives/C07UW93JVQ8/p1781104793615029)

kevindeforth:
resolves #3581

### DIFF
--- a/.github/prompts/pr-review.prompt.md
+++ b/.github/prompts/pr-review.prompt.md
@@ -46,11 +46,15 @@ PRIORITY CHECKS (report only if found):
    - Poor modularity (functions >100 lines, god objects)
    - Violated Single Responsibility Principle
 
+8. Code Comment Quality
+   - Enforce all standards defined in [engineering-standards.md] (paraphrasing of code, repetitions, explaining common terminology, leaking context, explanation that should be a stand-alone issue)
+
 REVIEW STYLE:
 - List only issues that should block the merge
 - Use bullet points, be direct and specific
 - Provide code suggestions for fixes when helpful
-- Do NOT comment on style, formatting, naming, or documentation unless it causes a bug
+- Flag code-comment quality issues per [engineering-standards.md]. The goal is to avoid comments that may become stale or add little value to the reader.
+- Do NOT comment on style, formatting, or naming unless it causes a bug.
 - Do NOT restate what the diff already shows
 - If no critical issues found: approve with a one-line summary
 - Sign off with: ✅ (approved) or ⚠️ (issues found)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -122,7 +122,7 @@ Do not suggest using `checked_add`, `checked_mul`, `checked_sub`, `saturating_ad
 Default to writing no comments. Add one only in case one of the following applies:
 - the *why* is non-obvious: an invariant, a constraint, a surprising behavior;
 - it is part of the public API. In this case, focus on information relevant to the _consumer_ of that code (provide brief example, explain complexity);
-- the code requires a follow-up. In this case, there should be an issue for it, linked via `TODO(#issue)`. The issue should do the heavy-lifting when it comes to explaining the desired behavior, not the code comment.
+- the code requires a follow-up. In this case, there should be an issue for it, linked via a `TODO(#example_issue_number): <description>` comment. The issue should do the heavy-lifting when it comes to explaining the desired behavior, not the code comment.
 
 Avoid comments that are:
 - paraphrasing the code;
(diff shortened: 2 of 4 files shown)


netrome (dismissed):
Thanks for adding this, this will make all future PRs a little bit better hopefully! 🙏

.github/prompts/pr-review.prompt.md
```diff
@@ -46,11 +46,15 @@ PRIORITY CHECKS (report only if found):
    - Poor modularity (functions >100 lines, god objects)
    - Violated Single Responsibility Principle
 
+8. Code Comment Quality
+   - Enforce all standards defined in [engineering-standards.md] (paraphrasing of code, repetitions, explaining common terminology, leaking context, explanation that should be a stand-alone issue)
+
```
gilcu3, line 51: hehe this one will be nasty, but granted :)

gilcu3 (dismissed):
Sounds good to me!

netrome (approved):

gilcu3 (approved):

Issue #3494, opened 2026-06-08
Remove serde TEE DTOs from `near-mpc-contract-interface` superseded by `tee-verifier-interface`


pbeza:
### Background

Once the verifier stack (#3235 / #3237, part of #3264) lands and `mpc-contract` consumes the Borsh DTOs from `tee-verifier-interface`, several serde-carrying TEE types in `near-mpc-contract-interface` become redundant. They duplicate the wire types now owned by `tee-verifier-interface` and only exist to serve the old in-contract JSON/serde path. Tracked as a standalone follow-up so it isn't lost after the verifier stack merges.

### User Story

As an MPC contract maintainer, I want the duplicated TEE DTOs removed from `near-mpc-contract-interface` once the verifier stack lands, so that the wire types have a single owner (`tee-verifier-interface`) and the interface crate stops carrying serde machinery it no longer needs.

### Acceptance Criteria

- [ ] Remove the TEE DTOs in `near-mpc-contract-interface` superseded by `tee-verifier-interface` (`Collateral`, `HexVec`, and the serde report / TCB-status types) once `mpc-contract` consumes the Borsh mirrors
- [ ] Re-point any remaining consumers to the `tee-verifier-interface` definitions
- [ ] Drop now-unused `serde_with` / `serde_bytes` deps from `near-mpc-contract-interface` if no other consumer remains

### Resources & Additional Notes

- Part of #3264 (move `dcap-qvl` out of the `mpc-contract` WASM)
- Related: the `tee-verifier-interface` <-> `attestation-types` `Collateral` / `QuoteBytes` dedup already noted in #3264's follow-ups (byte-identical Borsh layouts)
- Origin: review discussion on #3235

PR #3590, opened 2026-06-16
refactor(attestation): drop `dcap-qvl` dependency from the `attestation` crate

### LINKED ISSUE #3621: Make `dcap-qvl` optional in the `attestation` crates (prerequisite for dropping it from the contract WASM)
## Background

Make `dcap-qvl` (Intel TDX quote verification) an optional dependency of the `attestation` / `mpc-attestation` crates, so it is no longer in their default build. `dcap-qvl` is ~310 KB / ~21% of the `mpc-contract` WASM, pushing it toward NEAR's 1.5 MiB deployment cap ([NEP-509](https://github.com/near/NEPs/blob/master/neps/nep-0509.md)).

This is a prerequisite refactor: it does **not** move DCAP off-chain or shrink the WASM yet. The end goal is to stop compiling `dcap-qvl` into the contract, which requires the contract to verify an attestation **without** linking it — running only the post-DCAP policy checks (measurement allowlist, report-data binding, RTMR replay) and delegating the cryptographic DCAP step to the standalone `tee-verifier` contract over a cross-contract call. That was impossible before, because DCAP and the policy checks were welded into a single `Attestation::verify` that always called `dcap_qvl::verify::verify`, and `dcap-qvl` was non-optional.

This PR adds that `dcap-qvl`-free verification path and makes the dependency optional, so the follow-ups become mechanical:

1. **Move DCAP off the contract.** Instead of running DCAP in-WASM, the contract asks the `tee-verifier` contract to do it and returns a verified report; the contract then runs only the policy checks via the `verify_with_report` entry point added in this PR (which needs no `dcap-qvl`).
2. **Drop `dcap-qvl` from the contract.** With DCAP gone from the contract, it stops opting into the new `local-verify` feature — and `dcap-qvl` falls out of the build and the WASM.

The follow-up PRs do not exist yet; this is the first slice of a larger integration effort ([#3540](https://github.com/near/mpc/pull/3540)) being landed as smaller, self-contained PRs.

## Acceptance Criteria

- [ ] `Attestation::verify` is split into:
  - [ ] `verify_with_report(report, …)` — post-DCAP policy checks against an already-verified `VerifiedReport`; does not use `dcap-qvl`.
  - [ ] `verify_locally(…)` — full DCAP + post-DCAP in one call, behind a new `local-verify` feature.
- [ ] `dcap-qvl` is `optional`, gated behind `local-verify`, so it is absent from the default build of `attestation` and `mpc-attestation` (verifiable via `cargo tree`).
- [ ] The `dcap_qvl` ↔ `tee-verifier-interface` conversions live in a shared `tee-verifier-conversions` crate, with the byte-layout pin tests in one place.
- [ ] No behavior change: `mpc-contract` still enables `local-verify` and verifies locally exactly as before.

See the [design doc](docs/design/attestation-verifier-contract.md) for the full plan.

pbeza:
Closes #3621

## Why

`dcap-qvl` (Intel TDX quote verification) is ~310 KB / ~21% of the `mpc-contract` WASM, pushing it toward NEAR's 1.5 MiB deployment cap ([NEP-509]). The plan ([design doc](docs/design/attestation-verifier-contract.md)) offloads DCAP to a standalone `tee-verifier` contract so `mpc-contract` can drop `dcap-qvl`. This PR is the prerequisite refactor — it doesn't move DCAP or shrink the WASM yet.

This is the first reviewable slice of https://github.com/near/mpc/pull/3540, which integrates the standalone [`tee-verifier`](https://github.com/near/mpc/tree/main/crates/tee-verifier) contract into [`mpc-contract`](https://github.com/near/mpc/tree/main/crates/contract). That PR is too large to review in one pass, so it's being split into self-contained PRs that land one at a time.

## What changed

`Attestation::verify` is split so the cryptographic DCAP step is separable from the post-DCAP policy checks (measurement allowlist, report-data binding, RTMR replay):
- `verify_with_report(report, …)` — post-DCAP checks against an already-verified `VerifiedReport`; no `dcap-qvl`. The contract will call this once DCAP runs in the verifier contract.
  https://github.com/near/mpc/blob/231e48d0553d3d4c95dddfcedf39b993b1fcc39f/crates/attestation/src/attestation.rs#L131
- `verify_locally(…)` — full DCAP + post-DCAP in one call, behind the new `local-verify` feature; byte-identical signature to the old `verify`.
  https://github.com/near/mpc/blob/231e48d0553d3d4c95dddfcedf39b993b1fcc39f/crates/attestation/src/attestation.rs#L160

`dcap-qvl` becomes optional, so it's no longer in `attestation`'s default build.

Two supporting changes:
- The per-variant verification logic that was inlined in the old `Attestation::verify` now lives on each variant: `MockAttestation::verify(…)` (no DCAP, so a `Mock` is verifiable without a `VerifiedReport`) and `DstackAttestation::verify(…)` via a `DstackVerify` trait.
  https://github.com/near/mpc/blob/231e48d0553d3d4c95dddfcedf39b993b1fcc39f/crates/mpc-attestation/src/attestation.rs#L77
- The `dcap_qvl` ↔ `tee-verifier-interface` conversions move into a new `tee-verifier-conversions` crate (renamed from `tee-verifier/src/conversions.rs`), shared by the `tee-verifier` contract and the off-chain `verify_locally` path so the byte-layout pin tests live in one place instead of being duplicated.

## Why `local-verify` exists

Some callers still need full local DCAP:
1. the node verifies its own quote before submitting:
    https://github.com/near/mpc/blob/97170453f4e514420e92e06a01ac791a13aebb6d/crates/node/src/tee/remote_attestation.rs#L105-L113
1. `tee-authority`:
    https://github.com/near/mpc/blob/97170453f4e514420e92e06a01ac791a13aebb6d/crates/tee-authority/src/tee_authority.rs#L771-L776
1. `attestation-cli`:
    https://github.com/near/mpc/blob/97170453f4e514420e92e06a01ac791a13aebb6d/crates/attestation-cli/src/verify.rs#L73-L83

`local-verify` is the opt-in that re-adds `dcap-qvl` for them (gating `verify_locally`):
https://github.com/near/mpc/blob/231e48d0553d3d4c95dddfcedf39b993b1fcc39f/crates/mpc-attestation/src/attestation.rs#L266

## No behavior change (yet)

`mpc-contract` enables `local-verify` and still verifies locally via `verify_locally` (byte-identical to the old `verify`), so it links `dcap-qvl` and behaves exactly as before. The WASM only shrinks in the follow-up, when the contract switches to `verify_with_report` and drops `local-verify`.

[NEP-509]: https://github.com/near/NEPs/blob/master/neps/nep-0509.md


crates/tee-authority/src/tee_authority.rs
```diff
@@ -231,10 +231,10 @@ const PCCS_REQUEST_TIMEOUT: Duration = Duration::from_secs(10);
...
 /// more permissive than any default PCCS refresh schedule (Intel
 /// reference and Phala both refresh ~daily), so legitimate operators
-/// have ample headroom. The choice aligns with the contract's
-/// `DEFAULT_EXPIRATION_DURATION_SECONDS`: any attestation accepted by
-/// the contract is ≤7 days old *and* backed by collateral whose Intel
-/// signature is ≤7 days old.
+/// have ample headroom. This is the freshness bound on the collateral's
+/// Intel signature and is independent of the contract's attestation
+/// expiry (`DEFAULT_EXPIRATION_DURATION_SECONDS`, now 1 day); the two
+/// windows serve different purposes and are not required to match.
```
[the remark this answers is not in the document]
pbeza: Good catch — fixed in 2bc5d9921. Dropped the `, now 1 day` parenthetical; the constant is still 7 days in this PR (the 7d→1d change is a separate follow-up), and the point here is just that this collateral-freshness bound is independent of the contract's attestation-expiry window.

crates/mpc-attestation/src/attestation.rs
```diff
@@ -158,89 +170,208 @@ impl Attestation {
...
+        }
+    }
+
+    /// Full local verification: runs DCAP (`dcap_qvl::verify::verify`) and then
+    /// the post-DCAP checks. Off-chain only (the `local-verify` feature pulls
+    /// in `dcap-qvl`).
+    ///
+    /// Used by the node, `tee-authority`, and `attestation-cli`. On-chain,
+    /// `mpc-contract` instead calls the verifier contract for DCAP and then
+    /// [`verify_with_report`](Self::verify_with_report).
```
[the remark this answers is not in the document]
pbeza: Fixed in 2bc5d9921. Reworded to say `mpc-contract` calls `verify_locally` today (it enables `local-verify`), with the verifier-contract switch to `verify_with_report` noted as a planned follow-up.

crates/attestation/src/attestation.rs
```diff
@@ -163,6 +169,43 @@ impl DstackAttestation {
...
     }
 
+    /// Full local verification: runs `dcap_qvl::verify::verify` and then the
+    /// post-DCAP checks via [`verify_with_report`](Self::verify_with_report).
+    ///
+    /// Off-chain only (the `local-verify` feature pulls in `dcap-qvl`). Used by
+    /// the node, `tee-authority`, and `attestation-cli` to verify an
+    /// attestation end-to-end without the verifier contract. On-chain,
+    /// `mpc-contract` instead calls the verifier contract for the DCAP step and
+    /// then `verify_with_report` directly.
```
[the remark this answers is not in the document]
pbeza: Fixed in 2bc5d9921 — same reword as the sibling `mpc-attestation` doc: the contract performs local DCAP via `verify_locally` in this PR; moving the DCAP step to a separate verifier contract is the follow-up.

pbeza:
@claude review

crates/mpc-attestation/src/attestation.rs
```diff
@@ -144,103 +248,181 @@ pub fn default_measurements() -> &'static [ExpectedMeasurements] {
...
-                    accepted_measurements,
                     current_timestamp_seconds,
-                )?;
-
-                Ok(AcceptedAttestation {
-                    attestation: VerifiedAttestation::Mock(mock_attestation.clone()),
-                    advisory_ids: Vec::new(),
-                })
+                ))
             }
```
barakeinav1, line 370: I think we have a duplication here that can be resolved by routing this Dstack arm through `DstackVerify::verify` instead of re-inlining `verify_dstack_mpc_hashes` + the `AcceptedAttestation::dstack` assembly:

```rust
Self::Dstack(dstack_attestation) => {
    let report = dstack_attestation.verify_dcap_quote(current_timestamp_seconds)?;
    dstack_attestation.verify(
        &report,
        expected_report_data,
        current_timestamp_seconds,
        allowed_mpc_docker_image_hashes,
        allowed_launcher_docker_compose_hashes,
        accepted_measurements,
    )
}
```

This also gives the better ordering — DCAP first, then the post-DCAP checks (which conceptually depend on an already-verified report) — and matches what the async verifier flow will do anyway. The only change is which error surfaces first when a quote fails both DCAP and the hash checks; I don't think anything depends on that. Was the current hashes-first order intentional, or can we dedup this way?

barakeinav1: If we dedup this way, `verify_with_report__should_agree_with_verify_locally`  can be removed too.

barakeinav1 (commented):
wrote some comments

crates/mpc-attestation/src/attestation.rs
```diff
@@ -123,13 +226,14 @@ impl VerifiedAttestation {
...
-                timestamp_seconds,
-            ),
+            Self::Mock(mock_attestation) => mock_attestation
+                .verify(
+                    timestamp_seconds,
+                    allowed_mpc_docker_image_hashes,
+                    allowed_launcher_docker_compose_hashes,
+                    allowed_measurements,
+                )
+                .map(|_| ()),
```
barakeinav1: nit: `MockAttestation::verify` builds an `AcceptedAttestation::mock` (which clones the mock) only for it to be discarded by `.map(|_| ())` here. Since `re_verify` runs periodically on-chain, maybe extract the check into a small `fn check(...) -> Result<(), VerificationError>` that both `verify` and this path call, so the re-verify path doesn't allocate a throwaway `AcceptedAttestation`.

crates/attestation-cli/Cargo.toml
```diff
@@ -13,7 +13,7 @@ anyhow = { workspace = true }
 attestation = { workspace = true }
 bs58 = { workspace = true }
 clap = { workspace = true }
-mpc-attestation = { workspace = true }
+mpc-attestation = { workspace = true, features = ["local-verify"] }
```
netrome, line 16: Hmm, have you verified that this feature isn't accidentally enabled in production builds in the contract if we were to get rid of local verification there?

In my experience, feature unification can cause features to easily be accidentally enabled. I would expect this not to happen if we do contract-only builds but it could become a problem in e.g. integration tests if we're not careful.

netrome (dismissed):
Looks good to me. Didn't go deep as this doesn't seem to change any behavior. I second Barak's duplication concern but it's not a blocker imo.

crates/attestation-cli/Cargo.toml
pbeza: > Hmm, have you verified that this feature isn't accidentally enabled in production builds in the contract if we were to get rid of local verification there?

Checked: it's safe. Drop `local-verify` from the contract and `dcap-qvl` is gone from its tree:

```
$ cargo tree --manifest-path crates/contract/Cargo.toml -e normal | grep -ci dcap-qvl
0
```

The WASM builds as a single package, so its features only unify within its own dependency graph. Features enabled elsewhere in the workspace don't reach it. The follow-up PRs will take the contract fully async (cross-contract `verify_quote` + `verify_with_report` in the callback), so once that lands, the contract won't enable `local-verify` or link `dcap-qvl` at all.

crates/mpc-attestation/src/attestation.rs
pbeza: https://github.com/near/mpc/pull/3590/commits/8a4761cb79b80f0fe7138bfebae48309a01f4610

crates/mpc-attestation/src/attestation.rs
pbeza: https://github.com/near/mpc/pull/3590/commits/fc1129af9481179ea797c8cf94b8ca6f5b407451

crates/attestation-cli/Cargo.toml
netrome: Nice, that's good to know.

netrome: Thanks for verifying :pray:

pbeza:
All nits addressed, @barakeinav1 @netrome. ✅

Feel free to take re-review/approve. cc @gilcu3 @kevindeforth if you’re up for reviewing too.

netrome (approved):

crates/contract/src/dto_mapping.rs
```diff
@@ -135,7 +137,7 @@ impl IntoContractType<Collateral> for dtos::Collateral {
             qe_identity,
             qe_identity_signature: qe_identity_signature.into(),
             pck_certificate_chain,
-        })
+        }
```
gilcu3, line 140: not sure I understand that comment, what conversion it refers to? in the diff what I see is that we are removing a conversion

crates/test-utils/src/attestation.rs
```diff
@@ -98,7 +99,8 @@ pub fn near_account_key() -> near_sdk::PublicKey {
 pub fn mock_dstack_attestation() -> Attestation {
     let quote = quote();
     let collateral_json_string = include_str!("../assets/collateral.json");
-    let collateral = serde_json::from_str(collateral_json_string).unwrap();
+    let collateral = mpc_attestation::collateral::collateral_from_str(collateral_json_string)
+        .expect("collateral.json is valid collateral");
```
gilcu3, line 103: It is surprising to me that we lost the capability of deserializing the collateral here using `serde_json`. Could you briefly explain why it happened?

gilcu3 (approved):
Thank you!

crates/test-utils/src/attestation.rs
pbeza: Previously `attestation::Collateral` carried a custom `#[serde(try_from = "Value")]` to hex-decode the byte fields:

https://github.com/near/mpc/blob/ae8ca463df9ac8567923456889c884b03e1a3937/crates/attestation/src/collateral.rs#L16-L23

Now `Collateral` is the `tee-verifier-interface` type (foreign to this crate), which is the minimal Borsh wire-format mirror (no `hex`/`serde_json`):

https://github.com/near/mpc/blob/63fb5bec8299bb6997d6b22196a2b90e66b7d967/crates/tee-verifier-interface/src/lib.rs#L42-L62

Hex-string JSON is an off-chain fixture format, not the wire representation. The orphan rule means we can't attach that custom deserialize here, so the hex-decoding moved into the `collateral_from_str` free function. Same pattern as in `dto_conversions.rs`:

https://github.com/near/mpc/blob/a37bafa088d11d11ab8998191369d51f79bb4a35/crates/node/src/tests/dto_conversions.rs#L4-L7

crates/contract/src/dto_mapping.rs
pbeza: "This conversion" = the whole `impl` (the field-by-field mapping from the JSON DTO `dtos::Collateral` to the internal `Collateral`), not the line changed in the diff; that line just drops the old `QuoteCollateralV3` wrapper.

Right now there are two `Collateral` types: the serde JSON one in `near-mpc-contract-interface` (`dtos::Collateral`) and the internal one (`tee_verifier_interface::Collateral`), so we need this `impl` to convert between them. #3494 removes the duplicate DTO so only one `Collateral` remains, at which point this `impl` is dead and can be deleted.

PR #3596, opened 2026-06-16
refactor: importing reconstruction threshold from crypto into node

### LINKED ISSUE #3597: Importing reconstruction threshold from crypto into node
### Background

Currently there are two types of ReconstructionThreshold: one in the TSS repo and one in the node. 
It's good if we unify them.

### User Story

_No response_

### Acceptance Criteria

 * Change the inner type threshold_signatures::ReconstructionThreshold from usize to u64
 * Rename the appropriate functions such as .value() to .inner() to make it the same as the one in ReconstructionThreshold
 * Prevent using alias on the node and by calling the type directly from threshold_signatures crate and implementing necessary traits at the node on top of the imported ReconstructionThreshold

_No response_

SimonRastikian:
Closes #3597


anodar (approved):
is mpc_primitives::ReconstructionThreshold still used somewhere else ? If not should we drop it ?

anodar:
Reading the thread offline, looks like there's been discussions and it's deliberate why there's duplicate structs. I'm probably not a best person to review this as I'm missing quite a bit of context on it.

SimonRastikian:
Replaced with https://github.com/near/mpc/pull/3640

Issue #3551, opened 2026-06-14
Investigate: in-memory state cleanup on Done -> Inactive -> Onboard re-entry (follow-up to #3550)


barakeinav1:
### Background

#3550 made the migration onboarding state machine re-entrant — `OnboardingJob::Done → Onboard → Done` can now happen within a single process lifetime (a back-migration to a previously-active node, without restart).

Before that change, every entry into `Onboard(keyset)` came from a freshly-started process, so there was no in-memory state from any prior tenure to interfere. The re-entry path eliminates that implicit "process restart = state reset" guarantee. Restart still wipes all in-memory state by definition; re-entry does not.

Whether this is actually a problem depends on whether the coordinator (and the things it owns) cleanly drains in-memory state when the node transitions Active → Inactive. That reaction has never been exercised within a single process before. This issue is an **investigation**, not a known bug — we don't have evidence of staleness today, but we also don't have evidence the cleanup is complete.

### User Story

As an MPC node maintainer, I want to know whether re-entering `Onboard` after the node has been Active in this process risks corrupting protocol work with stale in-memory state, so that we can either close out the residual risk from #3550 or fix any gaps we find.

### Acceptance Criteria

- [ ] Identify every in-memory data structure owned by the coordinator or its child tasks that's keyed to (or scoped by) "I am a participant under key K".
- [ ] For each, determine whether it's properly drained / shut down when contract state transitions to make this node no longer a participant.
  - At minimum: P2P mesh / TLS connections (`network.rs`, `p2p.rs`), triple + presignature pools (`db.rs` + in-memory caches), pending sign-request queues, any spawned tokio task tied to participant status.
- [ ] Document findings inline (file + line refs); file separate fix issues for any gaps found.
- [ ] If a gap is non-trivial to close, add an integration or e2e test that exercises Done → Inactive → Onboard → Done in one process and asserts no stale state leaks (e.g. signatures produced after re-entry are correct against fresh-tenure inputs only).

### Resources & Additional Notes

- Triggering PR: [#3550](https://github.com/near/mpc/pull/3550) — re-entrant migration onboarding.
- Design doc with the risk flagged: [`docs/design/migration-onboarding-reentry.md`](https://github.com/near/mpc/blob/main/docs/design/migration-onboarding-reentry.md) (see the "Risk" section).
- Related issue: [#3406](https://github.com/near/mpc/issues/3406) (root cause).
[...]

PR #3601, opened 2026-06-17
feat(migration): back-migration without restart via state-driven dispatcher

### LINKED ISSUE #3406: Migration: migration without a restart is not supported
### Background

The migration web server's `import_keyshares` channel receiver is owned exclusively by the `onboard()` loop (`crates/node/src/migration_service/onboarding.rs:36`). When that loop exits with `OnboardingJob::Done` (line 55) — which happens immediately at startup for any node that is already an active participant — the receiver is dropped while the web server task keeps running with a dangling sender.

After that point, any `PUT /set_keyshares` to the node returns `500 keyshares receiver channel is closed` (`crates/node/src/migration_service/web/server.rs:185-191`). The node must be restarted to re-enter the onboarding loop and recreate the channel.

This surfaced when drafting a back-migration test that does not kill+restart A0 between forward and back directions (PR #3388): A0 cannot accept the back-migration PUT until its process restarts, because A0 was already an active participant at startup and `onboard()` returned before any migration was initiated.

### User Story

As an operator running a back-migration (B → A) onto a node A that is currently running and has previously been an active participant, I want the migration to succeed without restarting A — or, if a restart is required by design, for that requirement to be documented and the error message to say so.

### Acceptance Criteria

- [ ] Decide whether the current behavior is intentional (operator-driven restart as a state-reset boundary) or an oversight.
- [ ] If intentional: document the restart requirement in the migration runbook and update the 500 response body to name the cause + remediation (see #3407).
- [ ] If unintentional: keep the receiver alive for the lifetime of the migration web server (move it out of `onboard()`), or re-enter `onboard()` when the contract transitions back into a state where this node is an idle migration target.

- Code references:
  - `crates/node/src/migration_service.rs:42` — channel created
  - `crates/node/src/migration_service/onboarding.rs:45-87` — loop that exits on `Done`
  - `crates/node/src/migration_service/web/server.rs:185-191` — PUT handler that 500s when the receiver is gone
  - `crates/node/src/migration_service/types.rs:78-101` — `OnboardingJob::new` (`Active(Active) → Done`)
  - `crates/node/src/run.rs:335` — startup site that `.await`s `spawn_recovery_server_and_run_onboarding`
- Related: PR #3388 (no-kill back-migration test that exposed this), #2121 (production back-migration scenario), #3366 (upstream nearcore panic on the kill+restart path)

barakeinav1:
Closes #3406. Supersedes #3550.
followup -#3551 

See [`docs/design/migration-onboarding-dispatcher.md`](https://github.com/near/mpc/blob/barak/3406-migration-dispatcher/docs/design/migration-onboarding-dispatcher.md) for the design.

## Changes

- **`coordinator.rs`** — `Coordinator::run` takes a `CancellationToken`; new third arm in the outer `select!` returns `Ok(())` on cancel so the dispatcher can swap it out. `run()` is now `&mut self` so the dispatcher keeps the coordinator alive across role changes.
- **`migration_service.rs`** — splits the previous `spawn_recovery_server_and_run_onboarding` into:
  - `start_migration_web_server(...)` — binds the web server once and returns the keyshare-import receiver
  - `wait_until_role_change(...)` — watches contract_state + migration_info, returns when `OnboardingJob::new(...)` differs from the supplied current role
- **`run.rs`** — dispatcher loop replaces the linear startup. Each iteration: classify role via `OnboardingJob::new(migration_info, contract_state, my_id, tls_pub_key)`. `Done` → run coordinator until the role changes, then cancel. Otherwise → run `onboard()` (terminates on `Done`).
- **e2e tests** — extracts the existing back-migration test into a shared `run_back_migration_round_trip(port_seed, between)` helper with two variants:
  - `migration_service__should_handle_back_migration_a_to_b_to_a` (existing) — kill-and-restart A0 between directions
  - `migration_service__should_handle_back_migration_a_to_b_to_a_no_restart` (new) — A0 stays alive between directions; fails on the linear-init main path

## Test plan

- [x] `cargo check --workspace --all-features` — clean
- [x] `cargo clippy --all-targets --all-features --locked -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] `cargo test -p mpc-node --lib migration_service` — 26 / 26 pass
- [ ] CI green (in progress)
- [ ] Manual back-migration A → B → A in a localnet cluster with A0 alive throughout


crates/node/src/run.rs
```diff
@@ -410,26 +413,86 @@ where
...
+                        &tls_pub_key,
+                        OnboardingJob::Done,
+                    ) => {
+                        tracing::info!(
+                            "dispatcher: role changed away from active participant — \
+                             cancelling coordinator"
+                        );
+                        cancel.cancel();
+                    }
+                }
```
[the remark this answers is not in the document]
barakeinav1: Addressed in 4c9dd187: dropped the dead `cancel.cancel()`, removed the `CancellationToken` parameter from `Coordinator::run` entirely (and its third `select!` arm) — cleanup now flows exclusively through `Future::drop` → `drop_guard`, which is what was actually working before.

crates/node/src/migration_service.rs
```diff
@@ -30,38 +30,127 @@ impl From<&SecretsConfig> for MigrationSecrets {
...
+        let migration_info = my_migration_info_receiver.borrow_and_update().clone();
+        let job = OnboardingJob::new(migration_info, contract, my_near_account_id, tls_public_key);
+        if job != current {
+            return job;
+        }
+        tokio::select! {
+            _ = contract_state_receiver.changed() => {}
+            _ = my_migration_info_receiver.changed() => {}
+        }
+    }
```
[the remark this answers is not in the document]
barakeinav1: Addressed in d267c265: `wait_until_role_change` now returns `anyhow::Result<OnboardingJob>` and surfaces `changed()` errors with context.

crates/node/src/coordinator.rs
```diff
@@ -104,7 +104,14 @@ where
...
     ForeignChainPolicyReader: ReadSupportedForeignChain + Clone + Send + Sync + 'static,
 {
-    pub async fn run(mut self) -> anyhow::Result<()> {
+    /// Drives the MPC state machine for the lifetime of the coordinator role.
+    /// Returns `Ok(())` when `cancellation_token` fires (the dispatcher in
+    /// `run.rs` uses this to swap the coordinator out when the contract state
+    /// transitions out of the active-participant role — see
+    /// `docs/design/migration-onboarding-dispatcher.md`). On cancel, the
+    /// current in-flight MPC job's `AsyncDroppableRuntime` is dropped as
+    /// `self` borrows are released, terminating its spawned tasks.
```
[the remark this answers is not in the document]
barakeinav1: Addressed in 4c9dd187: dropped the `CancellationToken` parameter and rewrote the docstring to reflect actual semantics — `Future::drop` stops the in-flight MPC runtime and spawned tasks via the internal `drop_guard`; the coordinator's owned fields (`Arc<RwLock<KeyshareStorage>>`, etc.) stay allocated for the next run.

docs/design/migration-onboarding-dispatcher.md
```diff
@@ -0,0 +1,52 @@
...
+`run_coordinator` and `run_onboarding` are mutually exclusive — at most one is
+active at a time. When the contract state transitions out of the active
+subsystem's role, the dispatcher cancels it and re-enters the loop with the
+new role.
+
+## Requirements
+
+- **Coordinator: cancellable and restartable.** `Coordinator::run` takes a
+  `CancellationToken` and returns cleanly when it fires, releasing its
+  `Arc<RwLock<KeyshareStorage>>` handle and any in-flight protocol state.
```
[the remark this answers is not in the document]
barakeinav1: Addressed in 4c9dd187: the design doc now says "in-flight MPC runtime and spawned tasks are stopped; the coordinator's owned state is preserved for the next run" instead of the misleading Arc-release claim.

crates/node/src/migration_service.rs
```diff
@@ -30,38 +31,130 @@ impl From<&SecretsConfig> for MigrationSecrets {
...
 
-pub async fn spawn_recovery_server_and_run_onboarding(
+/// Binds the migration web server on `migration_web_ui` and returns the
+/// receiver side of the keyshare-import channel. The sender lives inside the
+/// web server's state, so the channel stays alive for as long as the server
+/// task does — i.e. the lifetime of the node process.
+///
+/// The dispatcher in `run.rs` clones this receiver into each invocation of
+/// `onboarding::onboard` so back-migrations can be served without a process
+/// restart. See `docs/design/migration-onboarding-dispatcher.md`.
```
kevindeforth: Please take a look at our code comment section in [engineering_standards.md](https://github.com/near/mpc/blob/main/docs/engineering-standards.md#write-helpful-code-comments).
This comment violates a few of the points outlined therein:
- it paraphrases the code, explaining "what" instead of "why"
- it explained where this code is used:
> // The dispatcher in `run.rs` clones this receiver into each invocation of
/// `onboarding::onboard` so back-migrations can be served without a process
/// restart. 
- it uses plain backticks instead of rustdoc intra-doc links:
> See `docs/design/migration-onboarding-dispatcher.md`.
- it explains what other parts of the code do. The following is not something this method can guarantee:
> Binds the migration web server on `migration_web_ui` and returns the
/// receiver side of the keyshare-import channel. The sender lives inside the
 /// web server's state, so the channel stays alive for as long as the server
/// task does — i.e. the lifetime of the node process.

crates/node/src/run.rs
```diff
@@ -410,26 +413,87 @@ where
...
+                    migration_info_receiver_for_dispatcher.clone(),
+                    my_near_account_id.clone(),
+                    tls_pub_key,
+                    tx_sender.clone(),
+                    keyshare_storage.clone(),
+                    import_keyshares_receiver.clone(),
+                )
+                .await?;
+            }
+        }
```
kevindeforth: This shouldn't be necessary. We should modify the `coordinator.run()` method such that it returns when it has nothing to coordinate, similar to how the onboard() method returns. Then, this will simplify to something along the lines of:

```rust
let current_job_receiver = decide_current_job(...);
while current_job_receiver.changed().await? {
    let current_job = current_job_receiver.borrow_and_update()?;
    match current_job {
        Job::Onboard => onboard(...).await?;
        Job::Participate => coordinator.run().await?;
    };
}
```

I will open a draft PR for this and see how it compares.

kevindeforth (commented):
Not what I had in mind. I will open a draft PR to see how the two solutions compare.

crates/node/src/run.rs
kevindeforth: Update: I played around with this and it's a pretty significant refactor. I think it's worth it, as it will remove _a lot_ of code. I will try to push something this week.

crates/node/src/migration_service.rs
barakeinav1: >Please take a look at our code comment section in [engineering_standards.md](https://github.com/near/mpc/blob/main/docs/engineering-standards.md#write-helpful-code-comments).

I see you added a new section for comments, I'll review it. thanks.

crates/node/src/run.rs
barakeinav1: > I will try to push something this week.
thanks

gilcu3:
@barakeinav1 we certainly want to get this done, putting it as draft so that we can retake it later

netrome:
Closing as stale

PR #3615, opened 2026-06-17
chore: drop post 3.12 migrations

### LINKED ISSUE #3456: Remove legacy triple column-family drop on startup after 3.12 rollout
### Background

#3298 removed the legacy `DBCol::Triple` column family, its dual-write, and the
startup migration. To reclaim disk on nodes upgrading from 3.11, `SecretDB::new`
now drops the on-disk `"triple"` column family once on startup if present
(`crates/node/src/db.rs`). This one-time cleanup is only needed until every
production node has run 3.12 at least once, after which no node still carries the
column on disk.

### Acceptance Criteria

- Remove the `legacy_triple` drop branch in `SecretDB::new` (`crates/node/src/db.rs`).
- Remove the `secret_db_new__should_drop_legacy_triple_column_family_on_startup` test.

### LINKED ISSUE #3495: Drop serde(default) on ProposedThresholdParameters::per_domain_thresholds after 3.12
### Background

per_domain_thresholds on ProposedThresholdParameters (crates/near-mpc-contract-interface/src/types/state.rs) is annotated #[serde(default)] so that pre-3.11 vote_new_parameters payloads — which predate the per-domain reconstruction-threshold
  overlay (#3169) and submit only { participants, threshold } — still deserialize, with the missing field defaulting to an empty (no-change) overlay.

  Once the 3.12 release has shipped and all clients/governance tooling are guaranteed to populate per_domain_thresholds explicitly (sending {} for a no-change vote), the backward-compat default is no longer needed. Remove #[serde(default)] so the
  field becomes mandatory and a legacy/malformed payload fails loudly instead of being silently treated as "no overlay."

### User Story

_No response_

### Acceptance Criteria

  - [ ] Remove #[serde(default)] from per_domain_thresholds.
  - [ ] Remove the corresponding TODO(#3495) comment.
  - [ ] Update the proposed_threshold_parameters__legacy_payload_omitting_overlay__parses_as_empty wire-lock test in state.rs — it asserts the exact legacy-omission behavior this change removes.
  - [ ] Regenerate the ABI snapshot (crates/contract/tests/snapshots/abi__abi_has_not_changed.snap) — per_domain_thresholds moves into the type's required list.

https://github.com/near/mpc/pull/3326#discussion_r3349190218

gilcu3:
Closes #3495 and #3456

This PR should be delete only, but it seemed we had missed a couple of places in the sandbox tests where the vote format needed to be updated. These only surface now because the backward compatibility is now gone.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6778,7 +6778,6 @@ dependencies = [
  "serde_repr",
  "serde_with",
  "sha2 0.10.9",
- "thiserror 2.0.18",
 ]
 
 [[package]]
--- a/crates/contract/tests/sandbox/common.rs
+++ b/crates/contract/tests/sandbox/common.rs
@@ -15,7 +15,7 @@ use mpc_contract::{
         key_state::{AttemptId, EpochId, KeyForDomain, Keyset},
         participants::{ParticipantInfo, Participants},
         test_utils::{bogus_ed25519_public_key, infer_purpose_from_protocol},
-        thresholds::{Threshold, ThresholdParameters},
+        thresholds::{ProposedThresholdParameters, Threshold, ThresholdParameters},
     },
     tee::tee_state::NodeId,
     update::{ProposeUpdateArgs, UpdateId},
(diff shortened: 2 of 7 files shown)


gilcu3:
@claude review

regarding the blocker, I just noticed it is not valid, the original issue was out-of-date. See my new comment there. The struct in the contract interface should keep the `#[serde(default)]` annotation to match the one internal to the contract

anodar (approved):
Thank you!

pbeza (approved):

PR #3616, opened 2026-06-17
feat: wire recent blocks tracker into event subscriber

### LINKED ISSUE #3608: feat(chain-gateway): wire `RecentBlocksTracker` into the event subscriber and omit empty block updates
### Background

[#3498](https://github.com/near/mpc/pull/3498) moved `RecentBlocksTracker` into the `chain-gateway` crate but didn't use it. The gateway's `BlockUpdate` carries no finality information, so consumers can't tell whether a block has finalized. Preparatory step for [#2112](https://github.com/near/mpc/issues/2112).

This will also allow us to avoid sending empty block updates over the channel.

### User Story

As a developer integrating `chain-gateway` into the node ([#2112](https://github.com/near/mpc/issues/2112)), I want the gateway to expose block finality on `BlockUpdate`, so that the node can drop its standalone `RecentBlocksTracker`.

### Acceptance Criteria

- [ ] The event-subscriber streamer owns a single `RecentBlocksTracker` and feeds every indexed block to it.
- [ ] `BlockUpdate` exposes a `BlockStatusHandle` consumers can poll for finality / canonicality.
- [ ] `chain-gateway` unit + integration tests pass.
- [ ] no longer send empty block updates to the event stream consumer.

The node-side migration (drop the standalone tracker in `mpc_client::monitor_block_updates`) is out of scope and lands as the next PR in the [#2112](https://github.com/near/mpc/issues/2112) chain.

kevindeforth:
resolves #3608 

Uses the `RecentBlocksTracker` inside the chain gateway crate, such that block events are accompanied by a block status handle (indicating their finality status).

Because `RecentBlocksTracker` now lives in the chain gateway, we can avoid sending empty block updates to the consumer, sending only in case a matching block event was found.

This PR adds two integration tests: one for checking back pressure works as expected, the other for verifying that finality status is eventually updated.


kevindeforth:
There is so much wrong with what Claude is spewing here, I don't even know where to start.

One thing that seems worth addressing is the blocking item: 
I agree that we would like some sort of metric for dropped messages. In this PR, we log a a warning. Will add this as an acceptance criteria to https://github.com/near/mpc/issues/3229

As for some other pearls of wisdom:
`u64::try_from(buffer_size).unwrap_or(u64::MAX)` this might be some new stanrdard I am not aware of, as Claude already tried to sneak that into the code before opening the PR. I'd rather panic than have a buffer of sixe `u64::MAX`, but maybe I am not up-to-date on the latest trends here...

> Localnet timing makes this likely fine in practice, but the test is vulnerable to flakiness; consider draining with a short recv timeout for the expected ones, then asserting try_recv is empty.

The event stream tracks optimistic events, whereas the viewer tracks finalized events. If this test is flaky due to that, then it means the finalized state view updates _before_ we see the optimistic event, which would be pretty bad and force us to re-think our architecture.
If we observe flakyness, it's a signal for bigger issues.

>  the removed comment explaining why this loop is sequential (vs. buffer_unordered in mpc-node) was a useful 'why' per engineering-standards.md (Write helpful code comments). Consider preserving a one-liner so a future reader does not reintroduce concurrency here.

Nah, thie usefulness of this comment kind of died when we moved the `RecentBlocksTracker` to the chain-gateway crate. It will become truly obsolete once we integrate the chain-gateway with the node.

crates/chain-gateway/src/event_subscriber/block_events.rs
```diff
@@ -2,13 +2,17 @@ use derive_more::{Deref, From};
...
+    event_subscriber::recent_blocks_tracker::BlockStatusHandle,
+    types::{BlockEntropy, BlockHeight},
+};
 
 /// The BlockUpdate returned by the Chain indexer.
 /// Similar to [`ChainBlockUpdate`](../../node/src/indexer/handler.rs) in the `mpc-node` crate.
 #[derive(Debug)]
 pub struct BlockUpdate {
     pub context: BlockContext,
+    pub status: BlockStatusHandle,
```
netrome, line 15: Why do we return a weak reference to a status here and what's the intended use case? I'd expect a block update to be a short-lived type, and as such I'd expect it to contain only plain data. It feels weird to receive a block update that explains we received a block with one status, only for this status to change later in the code.

On the other hand, if we expect users to actually want to hold this handle for a longer period of time to await a certain status I would suggest using a watch channel or something that can be awaited without polling instead.

crates/chain-gateway/tests/event_subscriber_integration.rs
```diff
@@ -361,3 +343,143 @@ async fn test_event_subscriber_receiver_error_if_non_private_call() {
...
+    let deadline = tokio::time::Instant::now() + Duration::from_secs(5);
+    while status.is_final() != Some(true) {
+        if tokio::time::Instant::now() >= deadline {
+            panic!(
+                "expected the block to be final once the state viewer confirmed the change; is_final={:?}",
+                status.is_final(),
+            );
+        }
+        tokio::time::sleep(Duration::from_millis(100)).await;
+    }
```
netrome, line 480: Building on my previous point, if we used a watch channel instead of atomics we wouldn't need a polling loop here.

crates/chain-gateway/src/event_subscriber/streamer/block_processor.rs
```diff
@@ -23,45 +24,64 @@ pub(super) async fn listen_blocks(
...
+        if !matched_events.is_empty() {
+            let permit = match block_update_sender.try_reserve() {
+                Ok(permit) => permit,
+                Err(tokio::sync::mpsc::error::TrySendError::Full(_)) => {
+                    tracing::warn!(
+                        block_context = ?context,
+                        "dropping block update because channel is full",
+                    );
+                    continue;
+                }
```
netrome, line 56: Why did we change to drop block updates instead of keeping the backpressure here?

I wouldn't want us to miss out on processing requests if we received a big burst of incoming requests.

netrome (dismissed):
Some questions from my side about the intended use case for the atomic. It feels like returning watch channels instead would make it much more ergonomic for users. 

Alternatively, since this is an event streaming model we might want to add behind a network boundary it would be nicer if we could use a model where we send plain data instead.

Anyway, these are non-blocking design considerations that I would love to discuss further since they could have a high impact on the usability of the chain gateway. I'll schedule a call on Monday to talk about this.

The only blocker I see is that we now can drop block events, which could lead to missing transactions. I would much prefer an architecture where we don't allow dropping block events in the chain gateway, that decision should be on the client side imo.

I see Claude also lifted the same point, and you only responded to the question about adding a metric - so I would be curious to understand more your reasoning behind this change. I understand that in the current structure, a straggling consumer would cause the whole machinery to pause including the recent blocks tracker. If in theory there would be a consumer waiting for a finality status, this could lead to a deadlock. But this to me suggests that we should separate the notion of subscribing to events from the ability to query for block status, not accept that a straggling consumer may miss out on events.

crates/chain-gateway/src/event_subscriber/block_events.rs
netrome: Adding a bit to this, top of mind I know two use cases for our subscriber:

1. Subscribe to requests. This is done optimistically and we don't need to await or poll for any changes in finality status here.
2. Subscribe to responses. Afaik we only care about responses once they are finalized, so ideally we should only subscribe to finalized responses. We are not interested in them before this point, so we don't need the ability to receive an optimistic response, and then poll or await it to be finalized.

crates/chain-gateway/tests/event_subscriber_integration.rs
kevindeforth: This will require extensive refactoring on the consumer side, as the node currently expects the atomic.

crates/chain-gateway/src/event_subscriber/streamer/block_processor.rs
kevindeforth: Because moving `RecenBlocksTracker` to the chain gateway removed the need to keep backpressure.

In case we experience it, it means there's something wrong on the consumer side and the MPC node is processing block updates slower than the indexer is producing blocks. 

I don't see an issue with dropping block updates, because:
1. This is already pathological and we should ensure this doesn't happen, so I am reluctant to implement elaborate work-arounds.
2. We decide the channel size and it will be set to 200 blocks. Any requests or responses older than 200 blocks are irrelevant for the MPC node.
3. In case of the node not draining the channel, we will experience signature timeouts long before this buffer is full.

So, in practice, there is no meaningful difference between the two options of _1) drop latest blocks_ and _2) drop newest blocks_.


That is why I opted for the solution with the least code complexity between the two (drop the newest blocks).

kevindeforth: > Because moving `RecenBlocksTracker` to the chain gateway removed the need to keep backpressure.

To add to this: the only reason we couldn't drop blocks before, was because `RecentBlocksTracker` lived in the node and required that the indexer send _all_ blocks to the node. Now, we no longer have that need.

So, we essentially have two options:
Option 1): keep the pre-PR status (keeping backpressure). This means that in case the node stalls, the indexer stalls and the block status of events already sent to the node will freeze.
Option 2): drop some blocks, but ensure that block status of events already sent to the node remain accurate.

Between these two, option 2) seemed more sensible.

> I wouldn't want us to miss out on processing requests if we received a big burst of incoming requests.

What do you mean by "big burst"? This comes from the indexer and would mean we have experienced a "big burst" in block production in the NEAR blockchain. Is that something we expect?

crates/chain-gateway/src/event_subscriber/block_events.rs
kevindeforth: > 1. Subscribe to requests. This is done optimistically and we don't need to await or poll for any changes in finality status here.


> Afaik we only care about responses once they are finalized, so ideally we should only subscribe to finalized responses

That's incorrect, the node is very much interested in tracking finality status of requests and responses (it wants to know they exist when they are optimistic and when they become final).

There are different ways of solving this than with the BlockStatusHandle, but they require larger rewrites of the code.

It is questionable if these other approaches would be as efficient as what the code currently does, so I am very hesitant to change the behavior, as it's a non-negligible effort (also on the reviewer side) and I don't see any meaningful benefit this would have to the product or code base.

> Why do we return a weak reference to a status here and what's the intended use case? I

If you look up references of that struct with your IDE, you will find it used in the node.

crates/chain-gateway/tests/event_subscriber_integration.rs
```diff
@@ -361,3 +343,143 @@ async fn test_event_subscriber_receiver_error_if_non_private_call() {
...
+    drop(watch_value);
+
+    for _ in 0..expected_received {
+        receiver
+            .try_recv()
+            .expect("expected a buffered block update");
+    }
+    receiver
+        .try_recv()
+        .expect_err("expected no further updates");
```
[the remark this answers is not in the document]
kevindeforth: addressed in https://github.com/near/mpc/pull/3616/commits/632602e1d322fb21693a1a39ec5ba407376e9d06

crates/chain-gateway/src/event_subscriber/streamer/block_processor.rs
netrome: > What do you mean by "big burst"? This comes from the indexer and would mean we have experienced a "big burst" in block production in the NEAR blockchain. Is that something we expect?

Ah right I was thinking about these events as individual requests, but this is containing all matched events for a single block. Makes sense, yeah then a burst would mean effectively a block storm which should be very unlikely on NEAR. That changes things.

netrome:
> This fails to take into account that the only prospective consumer of the block event stream (the MPC node) expects an atomic. Changing the API would require substantial refactor on the node side. It will also, most likely, result in a performance penalty.

Isn't the backup service also a prospective consumer of the block event stream?

But indeed, if it's too complicated to integrate on the node side we can keep the atomic. However this indicates that we have a much more complex refactor ahead of us when we want to break out the indexer to a separate binary.

crates/chain-gateway/src/event_subscriber/block_events.rs
netrome: > If you look up references of that struct with your IDE, you will find it used in the node.

The block status handle returned by the event subscriber isn't integrated in the node yet though, so it's currently unused. Yes, I've seen a lot of places where the type is used but it's not evident that the integration of the chain gateway maps 1:1 to the existing code.

netrome (approved):
@kevindeforth thanks for bearing with me on this. After exploring the code and discussing this more, I feel I understand the reasons for the tradeoffs much better now and I think it makes sense here.

Some highlights (for my own memory's sake):

- Why do we need to track status changes for events?
  - The node pauses/resumes retries for signature requests when it observes responses based on if they are on the canonical or non-canonical chain (i.e. if they are on the canonical chain we're not retrying computations).
  - This means we want an interface that allows tracking status updates, not just toggling between subscribing to events for finalized and non-finalized blocks.
- Why don't we want to keep the backpressure?
  - Because a straggling consumer would cause block updates to be stalled.
- Why don't we want to provide a separate interface for subscribing to block updates?
  - This would arguably be the better design choice, but it would be a more substantial refactor. We may consider it long term but it seems non-trivial to do this now.
  - If such a refactor was on the table, we'd likely primarily try to build a reactive system (e.g. using watch channels) rather than a polling based one (with atomics) to reduce latency for reacting to status updates. However, this again would be a massive refactor.
- Why is it okay to allow dropping block events?
  - For the node this happens only if the node falls behind and it should still not participate in signing rounds if this happens.
  - This means we may get false positive signing failure metrics if we miss responses. But that's already a possibility and even if we changed to an unbounded channel we'd still have problems with our metrics due to how they are computed.

Some things that would be nice to improve in this PR:

1. Make the lossy block event stream API more explicit, for example by using a tokio broadcast channel as it already provides very clear caller semantics for [lagging events](https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html#lagging).
2. Add clear error logs and metrics to make it easy for us to alert if a node lags behind.

pbeza (approved):
LGTM.

I did a relatively shallow review (~1h). Usually that would be enough for a decent/deep review, but this one is not trivial. The code is pretty tangled with indexer/nearcore internals, and it’s not an area I touch often, so I’d need more time before I could honestly say I reviewed it deeply.

I also slightly [overpromised](https://nearone.slack.com/archives/C091512F0DP/p1781850122666219) by saying I’d review it on Friday, so I don’t want to block you for too long, especially since @netrome already went quite deep on it. If you still need a second _deep_ review though, I’d need to dig more.

To review this, I spent some time with Claude refreshing my memory on [`RecentBlocksTracker`](https://github.com/near/mpc/blob/e46c70d141216c32eb1486e3f6fb68780a7b46d4/crates/chain-gateway/src/event_subscriber/recent_blocks_tracker.rs#L99) and its API. We have [good docs](https://github.com/near/mpc/blob/e46c70d141216c32eb1486e3f6fb68780a7b46d4/crates/chain-gateway/src/event_subscriber/recent_blocks_tracker.rs#L11-L98) on how it works internally, but understanding how it supposed to be wired with the indexer / nearcore, [channels](https://github.com/near/mpc/blob/632602e1d322fb21693a1a39ec5ba407376e9d06/crates/chain-gateway/src/event_subscriber/streamer/block_processor.rs#L25-L26), and the [subscription filters](https://github.com/near/mpc/blob/632602e1d322fb21693a1a39ec5ba407376e9d06/crates/chain-gateway/src/event_subscriber/streamer/block_processor.rs#L38) still takes a few rounds with Claude to piece together. Maybe adding some docs around [`listen_blocks`](https://github.com/near/mpc/blob/632602e1d322fb21693a1a39ec5ba407376e9d06/crates/chain-gateway/src/event_subscriber/streamer/block_processor.rs#L22) function would help people get the high-level picture faster (optional).

kevindeforth:
> Make the lossy block event stream API more explicit, for example by using a tokio broadcast channel as it already provides very clear caller semantics for [lagging events](https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html#lagging)

I remember that I had considered using a broadcast, but I didn't like the overhead of cloning the block events. We could put them behind an `Arc`, but that doesn't feel right. I am also not completely sold on the ring-buffer because of the implications on node behavior:

Assume the node falls behind consuming the blocks updates at time `t0` and recovers at time `t1 > t0`. Assume every block had a block event. Then, we have the following:

1. (this PR): Upon recovery, the channel contains blocks from `t0` until `t0 + 200 blocks`. Node works its way through those updates and discards blocks that are older than `t1-200`. It should catch-up fairly quickly to the newest block and start to be a useful participant in the network again.
2. (ring-buffer). Upon recovery, the channel contains blocks from `t1-200 blocks` until `t1`. None of these can be discarded by the node directly and it will need to work through all of these blocks. It may try to attempt signature generation for requests that already have a response in a later block. Overall, this takes the node longer to become useful again.

Comparing these two cases, I actually think dropping newest blocks in case the buffer is full is desirable.

> Add clear error logs and metrics to make it easy for us to alert if a node lags behind.

Note: this PR already logs this on a `warn` level. I can bump it to an `error` in a follow-up if need be, but we also have https://github.com/near/mpc/issues/3229, for adding a metric counting the number of dropped blocks.

PR #3617, opened 2026-06-17
fix: nits from available foreign chains work (contract changes)

anodar:
Followup on: https://github.com/near/mpc/pull/3511

- Rename Storage key `ForeignChainAvailability` to `ForeignChainMetadata` which is more suitable name for storing foreign chain metadata (shouldn't be a breaking change as we haven't deployed changes containing previous `ForeignChainAvailability` storage key)
- Add test that will detect changes in ForeignChainMetadata layout since that's stored with Lazy (raw bytes) so otherwise won't be detected until contract upgrade
- Drop recomputation of available foreign chains in vote_cancel_keygen avoid risk of running out of gas in this critical method, it now only happens 
when : 
 	- node registers foreign chains config (in register_foreign_chains_config)
 	- after key resharing concludes (in vote_reshared)
 	- when nodes vote to update/whitelist foreign chains providers (in vote_update_foreign_chain_providers)
 	- node concludes migration (conclude_node_migration)

### DIFF
--- a/crates/contract/src/foreign_chains_metadata.rs
+++ b/crates/contract/src/foreign_chains_metadata.rs
@@ -98,6 +98,14 @@ mod tests {
 
     use super::*;
 
+    #[cfg(all(feature = "__abi-generate", not(target_arch = "wasm32")))]
+    #[test]
+    #[expect(non_snake_case)]
+    fn foreign_chains_metadata_borsh_schema__should_not_change() {
+        let schema = borsh::schema::BorshSchemaContainer::for_type::<ForeignChainsMetadata>();
+        insta::assert_debug_snapshot!(schema);
+    }
+
     fn make_key(byte: u8) -> dtos::Ed25519PublicKey {
         dtos::Ed25519PublicKey([byte; 32])
     }
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -1284,8 +1284,6 @@ impl MpcContract {
 
         if let Some(new_state) = self.protocol_state.vote_cancel_resharing()? {
             self.protocol_state = new_state;
-            // Cancel-resharing reverts to the previous running state, so the
-            // cached available-chains set is still valid — no recompute needed.
         }
 
         Ok(())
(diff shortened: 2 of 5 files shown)


anodar:
@claude 
> crates/contract/src/lib.rs:1303 — Removing the recompute is fine for gas, but the prior comment's concern is still real: when Initializing was reached via vote_add_domains and at least one new ForeignTx domain with a higher reconstruction_threshold than any pre-existing ForeignTx domain completed before the cancel vote, the post-cancel Running state has a higher max ForeignTx threshold than the cache reflects — so get_available_foreign_chains will report chains as available that no longer meet the threshold until the next register / vote-update / reshare / conclude-migration recomputes. This is only a view (no safety impact — see lib.rs:2183) and it self-heals, but it's worth either (a) a one-line // in vote_cancel_keygen noting the cache may briefly be stale and self-heals on the next recompute trigger, or (b) a vote_cancel_keygen__should_* test pinning the current behavior. The PR description's gas-safety rationale isn't captured anywhere in code or tests.

We're not dealing with the case yet where we have multiple ForeignTx domains. So that's not applicable yet.

> crates/contract/src/lib.rs:1066 — Per docs/engineering-standards.md ("Use rustdoc intra-doc links instead of plain backticks"), the new doc comment should use `` [`ProtocolContractState::Running`] `` / `` [`ProtocolContractState::Resharing`] `` (or at least backticks) rather than single-quoted 'Running' / 'Resharing' so cargo doc checks the references. Also still doesn't cover the "no ForeignTx domain registered" no-op branch (already noted in the prior review).

Addressed in [1ee9b33](https://github.com/near/mpc/pull/3617/commits/1ee9b33f4e38b00991316189a2833e77b59cca1b)

> crates/contract/src/foreign_chains_metadata.rs:104 — The existing borsh-schema test at lib.rs:6744 is mpc_contract_borsh_schema_has_not_changed; the new one uses the project's <sut>__should_<assertion> pattern (good). Consider renaming the older one in a follow-up for consistency.

Unrelated to this PR. And would introduce another new snap file if I did rename it. I prefer to keep this PR scoped to addressing nits from previous pr.

gilcu3 (approved):
LGTM

netrome (approved):

PR #3624, opened 2026-06-18
feat(devnet): add localnet support in devnet

### LINKED ISSUE #3583: Add localnet-chain support to devnet for multi-machine benchmarking
### Background

The devnet CLI deploys MPC nodes across machines (GCP/Nomad), but only against testnet, so every cluster pays the testnet sync cost. The e2e harness runs against a localnet chain (no sync) but only on a single host. Neither covers the case we want for benchmarking: a localnet chain deployed *across* the cluster, giving real multi-machine P2P/CPU with zero chain sync and a controllable, non-bottleneck chain.

The MPC node already supports localnet chains (it keys off `chain_id`/`is_localnet()`), so the node side needs no changes — the work is in deployment and funding.

### Acceptance Criteria

- The devnet CLI can deploy a cluster running against a localnet chain (one or more `neard` localnet validators) instead of testnet, with MPC nodes still distributed across machines.
- A shared genesis and boot-node info are generated and distributed so each node's indexer peers with the localnet chain; a freshly deployed cluster has no chain sync wait.
- Account funding works without the testnet faucet (e.g. a genesis-funded master account), since `funding.rs`/`account.rs` currently rely on `helper.nearprotocol.com`.
- The devnet README documents the localnet deployment workflow.

gilcu3:
Closes #3583 

Corresponds to https://github.com/Near-One/infra-ops/pull/768


deployment/start.sh
```diff
@@ -95,7 +95,7 @@ migration_web_ui: 0.0.0.0:8079
 pprof_bind_address: 0.0.0.0:34001
 triple:
   concurrency: 2
-  desired_triples_to_buffer: 1000000
+  desired_triples_to_buffer: 16384
```
gilcu3, line 98: broken window fix, to match prod and avoid confusion

crates/devnet/README.md
```diff
@@ -423,3 +430,73 @@ running.
...
+
+By default the CLI deploys against **testnet**, so every cluster pays the testnet sync cost. For
+benchmarking you can instead deploy against a **localnet** chain (`mpc-localnet`): the cluster runs
+a single `neard` validator (on the base-infra machine, off the MPC node machines) and the MPC nodes
+peer with it as observers, so a freshly deployed cluster has **no chain sync wait**.
+
+The MPC node, its docker image, and the infra-ops Nomad job already support `mpc-localnet`; the
+shared genesis and keys come from the static assets in `deployment/localnet/`. The validator runs a
+**stock public `nearprotocol/nearcore` image** — the chain assets are delivered to it at runtime, so
+there is no custom image to build or publish. To use it:
```
kevindeforth: Please reword this paragraph to make it more useful to the reader.

> The MPC node, its docker image, and the infra-ops Nomad job already support `mpc-localnet`

Why would the reader not expect them to support this?

> the shared genesis and keys come from the static assets in `deployment/localnet/`.

It would be good to explain why that's relevant for the reader and if they can use these assets to tweak something.

> The validator runs a **stock public `nearprotocol/nearcore` image**

what is a "stock public" image?

> the chain assets are delivered to it at runtime, so there is no custom image to build or publish. To use it

How do I deliver the chain assets to the image? Why is this relevant to the reader?

crates/devnet/README.md
```diff
@@ -423,3 +430,73 @@ running.
...
+`deploy-chain` step starts the validator.
+
+1. **Images — nothing to build or publish.** The MPC nodes run a published image from Docker Hub
+   (pass it to `deploy-nomad` with `--docker-image nearone/mpc-node:<tag>`). The validator runs a
+   stock public `nearprotocol/nearcore` image, which you pass explicitly via `--neard-docker-image`
+   (use the tag matching the nearcore version the node embeds — see `nearcore` in the workspace
+   `Cargo.toml`, e.g. `nearprotocol/nearcore:2.12.0`); the CLI feeds it the genesis/config/keys at
+   runtime. Note: the published mpc-node image bakes in its own `genesis.json`, which must match
+   `deployment/localnet/genesis.json` in your checkout (the CLI sends that one to the validator) or
+   the nodes won't peer.
```
kevindeforth: I am confused what the instructions are. Do I need to change something in the `genesis.json` file?

It seems like I am instructed to run 

```
deploy-nomad --docker-image nearcore/mpc-node:<tag>
```
but I think I still need the network name for that, no?

crates/devnet/README.md
```diff
@@ -423,3 +430,73 @@ running.
...
+7. `mpc-devnet mpc $MPC_NETWORK_NAME update` — funds the participant accounts on the live chain (uses
+   the count from step 4).
+
+8. `mpc-devnet mpc $MPC_NETWORK_NAME deploy-contract --path ../../target/near/mpc_contract/mpc_contract.wasm`
+   (the `--path` is **required** on localnet — there is no testnet contract to fetch).
+
+9. `mpc-devnet mpc $MPC_NETWORK_NAME deploy-nomad --docker-image nearone/mpc-node:<tag>` — deploys the
+   MPC node jobs. The validator image is reused from step 6 automatically (override with
+   `--neard-docker-image` only if you intend to re-image the validator). Then proceed with the normal
+   `add-keys`, `init-contract`, and `vote-add-domains` steps — with no sync wait.
```
kevindeforth: I would prefer if we stayed consistent in the style. When we instruct the user to type a command, lets use the format from the previous section of this readme:

\```shell
<command_to_type>
\```

crates/devnet/README.md
```diff
@@ -423,3 +430,73 @@ running.
 ```
 
 Run `./scripts/loadtest-scenarios.sh --help` for the full list of options.
+
+## Deploying against a localnet chain
```
kevindeforth, line 434: This section needs a be refined to be useful, but the other changes to this file are good :+1:

crates/devnet/config.yaml.template
```diff
@@ -5,6 +5,11 @@ rpcs:
 
 infra_ops_path: /replace_me/infra-ops
 
+# Chain the RPCs talk to. Omit (or `testnet`) for testnet. Set to `mpc-localnet` to deploy and
+# operate against a localnet chain: funding then auto-derives the genesis master (no faucet), and
+# the terraform deploy brings up a neard validator. Point `rpcs` above at the validator's RPC.
```
kevindeforth: Why is this information:
>  funding then auto-derives the genesis master (no faucet), and the terraform deploy brings up a neard validator. Point `rpcs` above at the validator's RPC.

relevant to the reader in this file? Feels like we should link to some documentation instead.

crates/devnet/src/mpc.rs
```diff
@@ -247,9 +278,16 @@ impl MpcAddKeysCmd {
...
 impl MpcDeployContractCmd {
     pub async fn run(&self, name: &str, config: ParsedConfig) {
+        let funding_account = resolve_funding_account(&config);
         let (contract_data, contract_path) = match &self.path {
             Some(contract_path) => (std::fs::read(contract_path).unwrap(), contract_path.clone()),
             None => {
+                if config.is_localnet() {
+                    panic!(
+                        "On localnet there is no testnet contract to fetch; pass --path to a built \
+                         contract wasm (e.g. ../../target/near/mpc_contract/mpc_contract.wasm)."
```
kevindeforth, line 288: note: There would be ways around it, but I don't think it's worth it. Might be useful to refer to the contract-history crate: https://github.com/near/mpc/tree/main/crates/contract-history/archive

crates/devnet/src/mpc.rs
```diff
@@ -150,15 +174,22 @@ impl UpdateMpcNetworkCmd {
...
-            .unwrap_or(mpc_setup.participants.len());
+        // Default to the intended count stored at creation, so a bare `update` funds the network up
+        // to the size requested by `new` (relevant on localnet, where `new` deferred funding).
+        let num_participants = self.num_participants.unwrap_or(
+            mpc_setup
+                .desired_num_participants
+                .max(mpc_setup.participants.len()),
+        );
+        mpc_setup.desired_num_participants =
+            mpc_setup.desired_num_participants.max(num_participants);
```
kevindeforth, line 192: Q: What is the purpose of `desired_num_participants` and `num_participants`? Why do we need both?

crates/devnet/src/types.rs
```diff
@@ -63,6 +63,14 @@ pub struct MpcNetworkSetup {
     pub nomad_server_url: Option<String>,
     #[serde(default)]
     pub ssd: bool,
+    /// Intended participant count, stored so the cluster can be sized (`deploy-infra`) before the
+    /// participant accounts are funded. On localnet, funding is deferred until the chain is up.
+    #[serde(default)]
+    pub desired_num_participants: usize,
```
kevindeforth, line 69: Can't this be inferred from the `participants` vector?

crates/devnet/src/types.rs
```diff
@@ -63,6 +63,14 @@ pub struct MpcNetworkSetup {
...
+    /// The neard image the localnet validator was deployed with, recorded by `deploy-chain` so
+    /// `deploy-nomad` reuses the same tag without re-specifying `--neard-docker-image`.
+    #[serde(default)]
+    pub neard_docker_image: Option<String>,
```
kevindeforth, line 73: Hmm, interesting.

You probably added this here for the same reason I added the `ssd` bool above, which in hindsight, I wouldn't do anymore.

The issue is that these two things describe node-deployment specifics (do we use an ssd? what docker image is deployed?), whereas the rest is oblivious to those things and could be re-used for different deployments.
This means that it's harder to re-use generated accounts across deployments. For localnet, this is probably not an issue. For testnet, it's a bit of an inconvenience, as it takes time to generate and fund accounts.

In any case, not a blocker, but something I would be curious to hear your thoughts on.

kevindeforth (dismissed):
Very nice, thank you very much!

Main concern is the readme, which needs improvement (I won't block the PR for it, but I would like to see these addressed eventually).

I would prefer not to introduce the `desired_num_participants` variable in the types, but I am not sure I have good alternatives. Was thinking of an enum over `participants` maybe, but it's been a while since I have been in this part of the codebase (not a blocker).

crates/devnet/config.yaml.template
gilcu3: technically there is no need to explain the funding process here, it should work. Will just omit that part

crates/devnet/src/mpc.rs
gilcu3: yeah, although I think we should leave it as a manual option, so the current example should be fine

crates/devnet/src/mpc.rs
gilcu3: This change was to be able to create the infra before funding accounts, but let me check if I can workaround that without duplicating the variable

crates/devnet/src/types.rs
gilcu3: Not much thought on it from my side, I just needed a place to store it and this seemed to be the one with less friction. But from my experience lately funding accounts has been relatively fast, so not sure if it is worth refactoring if that's the only reason.

crates/devnet/src/mpc.rs
gilcu3: Ok now I remember well, this is a workaround to persist the number of participants so that a second round of deploy-infra is not needed. Technically we could do the participant persistence and its funding as separate steps (now they are a bit entangled) and then this variable would not be needed. We can certainly do that in a follow-up.

crates/devnet/src/account.rs
```diff
@@ -642,3 +687,73 @@ impl OperatingAccounts {
...
+            account_id: "funder.near".parse().unwrap(),
+            access_keys: vec![SigningKey::generate(&mut StdRng::seed_from_u64(42))],
+            kind: NearAccountKind::FundingAccount,
+        };
+
+        // When
+        let resolved = resolve_funding_account_inner(Some(&configured), true);
+
+        // Then
+        assert_eq!(resolved.unwrap().account_id.as_str(), "funder.near");
```
pbeza: Nit: it’s probably worth extracting `"funder.near"` into a const.

crates/devnet/src/account.rs
```diff
@@ -642,3 +687,73 @@ impl OperatingAccounts {
...
+
+    /// The localnet master must match `deployment/localnet/validator_key.json`, since its key
+    /// controls the genesis-funded `test.near` account used for faucet-free funding.
+    #[test]
+    fn localnet_genesis_master__should_load_test_near_with_genesis_key() {
+        // Given / When
+        let master = localnet_genesis_master();
+
+        // Then
+        assert_eq!(master.account_id.as_str(), "test.near");
```
pbeza:
```suggestion
        assert_eq!(master.account_id.as_str(), LOCALNET_MASTER_ACCOUNT_ID);
```

crates/devnet/src/account.rs
```diff
@@ -642,3 +687,73 @@ impl OperatingAccounts {
...
+    #[test]
+    fn resolve_funding_account__should_use_genesis_master_on_localnet_without_config() {
+        // Given
+        let configured = None;
+
+        // When
+        let resolved = resolve_funding_account_inner(configured, true);
+
+        // Then
+        assert_eq!(resolved.unwrap().account_id.as_str(), "test.near");
```
pbeza:
```suggestion
        assert_eq!(resolved.unwrap().account_id.as_str(), LOCALNET_MASTER_ACCOUNT_ID);
```

crates/devnet/src/account.rs
```diff
@@ -642,3 +687,73 @@ impl OperatingAccounts {
...
+        let NearAccountKind::FundingAccount = master.kind else {
+            panic!("expected FundingAccount, got {:?}", master.kind);
+        };
+        let public_key = format!(
+            "ed25519:{}",
+            bs58::encode(master.access_keys[0].verifying_key().as_bytes()).into_string()
+        );
+        assert_eq!(
+            public_key,
+            "ed25519:7E9jtdHWcn4eVZQPiHMCUMVVnMEvGeSG3HPVasisWooU"
```
pbeza: Nit: perhaps worth defining in `crates/devnet/src/constants.rs`?

crates/devnet/src/types.rs
```diff
@@ -284,3 +307,48 @@ pub mod near_crypto_compatible_serialization {
...
+    max_concurrency: 30
+infra_ops_path: /tmp/infra-ops
+chain_id: mpc-localnet
+";
+
+        // When
+        let config: Config = serde_yaml::from_str(yaml).unwrap();
+
+        // Then
+        assert_eq!(config.chain_id.as_deref(), Some("mpc-localnet"));
```
pbeza: Nit: extract to a const and reuse it.

crates/devnet/src/account.rs
```diff
@@ -642,3 +687,73 @@ impl OperatingAccounts {
...
 }
+
+#[cfg(test)]
+#[expect(non_snake_case)]
+mod tests {
+    use super::*;
+    use rand::SeedableRng;
+    use rand::rngs::StdRng;
+
+    /// The localnet master must match `deployment/localnet/validator_key.json`, since its key
```
pbeza:
```suggestion
    /// The localnet master must match [`LOCALNET_VALIDATOR_KEY_PATH`], since its key
```

pbeza (dismissed):

crates/devnet/src/types.rs
gilcu3: answered in the comment above https://github.com/near/mpc/pull/3624#discussion_r3442244772

crates/devnet/README.md
gilcu3: hopefully it is clear now

crates/devnet/README.md
gilcu3: reworded

kevindeforth (approved):
Thank you!

pbeza (approved):

PR #3632, opened 2026-06-22
fix(ci): make rust cache key depend on nix cache key

### LINKED ISSUE #3400: rust cache poisoned when nix is updated
### Background

We recently faced issues where all rust cache was poisoned.

When nix flake is updated, the rust cache should become stale. For that, we need to make sure that the nix cache key is part of the rust cache key.

### User Story

_No response_

### Acceptance Criteria

Make the rust cache key depend on the nix cache key, so that updates are handled gracefully.

_No response_

gilcu3:
Closes #3400

### DIFF
--- a/.github/actions/restore-nix-cache/action.yml
+++ b/.github/actions/restore-nix-cache/action.yml
@@ -7,13 +7,26 @@ inputs:
     description: Whether this job should save the cache after running.
     required: false
     default: "false"
+outputs:
+  cache-key-hash:
+    description: >
+      Hash of the Nix inputs (flake, toolchain, lockfiles) that key the /nix
+      cache. Dependent caches (e.g. the Rust build cache) should fold this into
+      their key so they are invalidated whenever the Nix-provided toolchain
+      changes.
+    value: ${{ steps.compute-key.outputs.hash }}
 runs:
   using: composite
   steps:
+    - name: Compute Nix cache key hash
+      id: compute-key
+      shell: bash
+      run: echo "hash=${{ hashFiles('flake.nix', 'flake.lock', 'nix/**', 'rust-toolchain.toml', 'Cargo.lock') }}" >> "$GITHUB_OUTPUT"
+
     - name: Restore /nix from cache
       uses: nix-community/cache-nix-action@7df957e333c1e5da7721f60227dbba6d06080569 # v7.0.2
       with:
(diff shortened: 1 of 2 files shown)


.github/actions/restore-nix-cache/action.yml
```diff
@@ -7,13 +7,26 @@ inputs:
...
+    - name: Compute Nix cache key hash
+      id: compute-key
+      shell: bash
+      run: echo "hash=${{ hashFiles('flake.nix', 'flake.lock', 'nix/**', 'rust-toolchain.toml', 'Cargo.lock') }}" >> "$GITHUB_OUTPUT"
+
     - name: Restore /nix from cache
       uses: nix-community/cache-nix-action@7df957e333c1e5da7721f60227dbba6d06080569 # v7.0.2
       with:
-        primary-key: nix-${{ runner.os }}-ci-${{ hashFiles('flake.nix', 'flake.lock', 'nix/**', 'rust-toolchain.toml', 'Cargo.lock') }}
+        primary-key: nix-${{ runner.os }}-ci-${{ steps.compute-key.outputs.hash }}
```
gilcu3, line 29: I am keeping the status quo here, but I cannot see why the nix cache should use `Cargo.lock`

pbeza (approved):

.github/actions/restore-nix-cache/action.yml
netrome: Isn't `Cargo.lock` part of the builds? Or how do we ensure in nix that we're building exactly the same versions of the different crates?

netrome (approved):
Shallow approval + a question, thanks for doing this 🙏

.github/actions/restore-nix-cache/action.yml
gilcu3: most of our builds are done by using cargo, so nix does not touch or see the `Cargo.lock`. Once exception is the `mpc-node nix build` action, but that one reuses cache from the other steps and recomputes the rest of the nix derivation (specific to the node, which does depend on Cargo.lock). So I think removing this would be correct, viewing this cache as the "base OS system" on top of which we can build rust crates, using `cargo` or `nix` itself

netrome: Ah right, so the cache could theoretically contain build artifacts for adjacent versions but that's not necessarily a problem. Sounds reasonable.

PR #3635, opened 2026-06-22
fix(bench): make sure benches do not fail by fixing determinism

### LINKED ISSUE #3516: Simulated-replay benches silently measure time-to-error instead of protocol runtime
### Background

The `advanced_*` benches drive a single real participant against recorded snapshot messages via `run_simulated_protocol`, but pass its `Result` straight to criterion, which discards it:

```rust
|preps| run_simulated_protocol(preps.participant, preps.protocol, preps.simulator),
```

A protocol that errors mid-run is simply timed as a fast sample. Adding an `.expect()` shows two benches are already failing on `main`:

- `advanced_ot_based_ecdsa` (triples group):
  `AssertionFailed("q check failed")` — the reported ~13.6ms is time-to-error, not triple generation time.
- `advanced_dkg`: replay also returns `Err`.

The replay appears to diverge from the recorded run, and any change to message interleaving moves where it trips an assertion (discovered while benchmarking #3501, where the broken triples bench reported a bogus −84% "improvement").

Six bench files share the discard pattern and should be audited: `advanced_ot_based_ecdsa`, `advanced_dkg`, `advanced_robust_ecdsa`, `advanced_eddsa_frost_sign_v1`, `advanced_eddsa_frost_sign_v2`, `ckd`.

### Acceptance Criteria

- All benches using `run_simulated_protocol` fail loudly (e.g. `.expect()`) when the protocol returns an error, so criterion can never time a truncated run.
- The replay divergence behind the `advanced_ot_based_ecdsa` triples and `advanced_dkg` failures is diagnosed and fixed, and those benches complete successfully.
- The `naive_*` benches assert success on their `run_protocol` results as well.

gilcu3:
Closes #3516

### DIFF
--- a/crates/threshold-signatures/benches/simulate_ecdsa.rs
+++ b/crates/threshold-signatures/benches/simulate_ecdsa.rs
@@ -295,6 +295,7 @@ fn ot_run_sign(
     }
 
     let (results, metrics) = run_simulation(protocols, latency);
+    assert_eq!(results.len(), participants.len());
     assert!(results.iter().any(|(_, sig)| sig.is_some()));
     metrics
 }
@@ -377,6 +378,7 @@ fn robust_run_sign(
     }
 
     let (results, metrics) = run_simulation(protocols, latency);
+    assert_eq!(results.len(), participants.len());
     assert!(results.iter().any(|(_, sig)| sig.is_some()));
     metrics
 }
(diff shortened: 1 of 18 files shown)


crates/threshold-signatures/benches/simulate_ecdsa.rs
```diff
@@ -377,6 +378,7 @@ fn robust_run_sign(
     }
 
     let (results, metrics) = run_simulation(protocols, latency);
+    assert_eq!(results.len(), participants.len());
```
gilcu3, line 381: these are broken window fixes. Not strictly needed, but better to have them in case a participant does not finish

anodar (approved):

netrome (approved):

PR #3640, opened 2026-06-22
feat: adding reconstruction threshold in node

SimonRastikian:
Closes #3164
Closes #1680  in [0b8b201](https://github.com/near/mpc/pull/3640/commits/0b8b20196f4b7d97d2e081b0c8e8147b43fbcceb)

Issue #3630, opened 2026-06-19
Drop deprecated foregin chain methods in contract once nodes are migrated to the new API


anodar:
### Background

We've introduced a new API for calculating available foreign chains in https://github.com/near/mpc/issues/3568.

We should drop deprecated methods once nodes switch to using them, that is once Contract is upgraded to 3.13, and then nodes are upgraded to 3.14 to use that new API.

### Acceptance Criteria

Following API from contract is dropped:

- get_supported_foreign_chains
- get_foreign_chain_support_by_node
- SupportedForeignChainsByNode struct

PR #3641, opened 2026-06-22
feat: wire new foreign chain API for available chains

### LINKED ISSUE #3569: Wire new API from contract in node for calculating available foreign chains
### Background

Instead we need to follow design from https://github.com/near/mpc/issues/3474

### User Story

_No response_

### Acceptance Criteria

Node uses new API from https://github.com/near/mpc/issues/3568 to register new chains and for available foreign chains.

_No response_

### Subtasks

 - [x] 1. #3828 — **refactor(node): make `VerifyForeignTxProvider` methods inherent**
   Drops the `SignatureProvider` impl (its keygen/resharing methods only bailed) so later slices can reshape the provider freely. No behavior change.
 - [x] 2. #3829 — **feat(node): register foreign-chain configs via the new contract API**
   Node sends the new `register_foreign_chains_config` alongside the deprecated registration (dual-write for the upgrade window). The fake contract gains the new model next to the legacy one: per-node configs plus available chains recomputed from ForeignTx-domain reconstruction thresholds. Read path untouched.
 - [x] 3. #3842
 - [x] 4. #3852
 - [x] 5. #3910
 - [x] 6. #3959 
 - [x] 7. #3961
 - [x] 8. #4221

anodar:
Notable changes:
- When selecting participants it now checks additionally that participants support requested chain (in `make_verify_foreign_tx_leader`)

- Switch supported chains to available chains: 
  - `get_supported_chains`: intersection of all reported suppored chains by node -> `get_available_chains`: chains that are available in at least threshold (of ForeignTx Domain) number of node (each node having at least quorum # of RPC providers for that chain)
  - `ReadSupportedForeignChain` trait -> `ReadAvailableForeignChains`

- Mark deprecated:
  - get_supported_foreign_chains
  - get_foreign_chain_support_by_node

- Extend `FakeMpcContractState` with Available Foreign Chains and Foreign Chains Configs:
  - Note: fake doesn't follow same flow for populating available foreign chains (in production that's via voting)
  - TODO(anodar): maybe worth changing ?! 

NOTE: these changes should not be deployed on the node until all of the below criteria is met:
- API changes in the contract from https://github.com/near/mpc/pull/3511 are deployed (on contract)
- RPC whitelist is voted on chain
- All nodes report all whitelisted chains (they have new providers in their configs, + aptos) & available_foreign_chains return all 4 chains

Closes: https://github.com/near/mpc/issues/3569


anodar:
@claude review

anodar:
@claude review

anodar:
@claude 

> crates/node/src/providers/verify_foreign_tx/sign.rs (new tests) — Comment casing is inconsistent: some tests use // Given / // When (capitalized) while several use // when, then: (lowercase). Per the AAA convention used elsewhere in the crate, the markers add no signal when the body is a single assert! — happy to drop them entirely or normalize to the capitalized form. Not a correctness concern; flagging only because the previous review removed the same markers from mock_policy_reader.

Addressed in: [58a4cb0](https://github.com/near/mpc/pull/3641/commits/58a4cb0e708c098cbe93688d5cd7db363774d558)

> crates/node/src/providers/verify_foreign_tx/sign.rs (tests) — There's a 3-participant negative case but no 3-participant positive case (all three support → true). Adding the dual would lock in the AND-over-participants semantics symmetrically; today a refactor that swapped all for any would still be caught only by the single-participant negatives.

Added test in: [58a4cb0](https://github.com/near/mpc/pull/3641/commits/58a4cb0e708c098cbe93688d5cd7db363774d558)


> crates/e2e-tests/src/cluster.rs and crates/e2e-tests/tests/common.rs — MpcClusterConfig::whitelisted_chains is added as a config field but immediately std::mem::take'd out of config in common::must_setup_cluster before MpcCluster::start(config) runs. Threading it through must_setup_cluster as a separate parameter (or as part of a small post-init step argument) would avoid the field-on-config-but-not-used-by-start pattern; minor design taste.

I prefer whitelist to be included in the config rather than fed as a separate parameter. Changed to clone though instead of mem:take. 


> crates/node/src/indexer/fake.rs (register_foreign_chains_config) — The available_foreign_chains recomputation uses unwrap_or_default() for participants without a registered config, so the intersection collapses to empty whenever any active participant hasn't yet registered. That's the prior semantic, but it now lives in two places (deprecated path + new path) and silently diverges from the contract's threshold-based rule. A one-line `// Fake-only: stricter than the contract's threshold-based rule (https://github.com/near/mpc/issues/3630)` near the reduce would make the divergence discoverable from the type rather than only from the PR description.

Available chains in node fake isn't calculated based on whitelist, but simply from reported available configs from nodes, that's correct. It's to simplify tests. Full flow is covered in e2e tests instead. Deprecated path will be dropped once contract 3.13 is deployed.

crates/e2e-tests/src/cluster.rs
```diff
@@ -859,7 +867,20 @@ impl MpcCluster {
             .await
     }
```
gilcu3, line 868: would it be possible to add this as a TODO, so that it is caught by CI later?

crates/e2e-tests/src/cluster.rs
```diff
@@ -889,6 +919,63 @@ impl MpcCluster {
...
+            provider("https://api.alchemy.com/v2/"),
+        );
+        providers.insert(
+            ProviderId("quicknode".to_string()),
+            provider("https://api.quicknode.com/"),
+        );
+        providers.insert(
+            ProviderId("public".to_string()),
+            provider("https://rpc.public.io/"),
+        );
```
gilcu3, line 948: hmm does this mean that these tests cannot run offline? I think we generally prefer that tests do not depend on third party services

EDIT: I checked and no, there is nothing online. To avoid future confusion could be better to use trivially wrong urls here, or adding a comment

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
```diff
@@ -199,6 +199,19 @@ async fn setup_foreign_tx_cluster() -> anyhow::Result<ForeignTxTestEnv> {
...
+        ForeignChain::Abstract,
+        ForeignChain::Bnb,
+        ForeignChain::Starknet,
+        ForeignChain::Base,
+        ForeignChain::Arbitrum,
+        ForeignChain::HyperEvm,
+        ForeignChain::Polygon,
+    ]
+    .into_iter()
+    .collect();
```
gilcu3, line 212: No Aptos here?

crates/node/src/indexer/fake.rs
```diff
@@ -51,25 +51,31 @@ pub struct FakeMpcContractState {
...
             .lock()
             .await
-            .supported_foreign_chains()
+            .available_foreign_chains()
             .clone())
     }
+
+    async fn get_foreign_chains_configs(&self) -> anyhow::Result<dtos::ForeignChainsConfigs> {
+        Ok(self.contract.lock().await.foreign_chains_configs().clone())
+    }
```
gilcu3: why is this returning a result? It appears it cannot fail

crates/node/src/indexer/fake.rs
```diff
@@ -163,14 +174,54 @@ impl FakeMpcContractState {
...
+            .iter()
+            .map(|(_, _, p_info)| {
+                foreign_chains_configs
+                    .get(&p_info.tls_public_key)
+                    .map(|c| c.iter().copied().collect::<BTreeSet<_>>())
+                    .unwrap_or_default()
+            })
+            .reduce(|a, b| a.intersection(&b).copied().collect())
+            .unwrap_or_default()
+            .into();
```
gilcu3, line 217: is this the place where a change will be needed, to make use of the threshold in foreign chain availability? If this is the only place a TODO seems granted

crates/node/src/coordinator.rs
```diff
@@ -952,6 +945,41 @@ fn stop_initializing(
...
+    if let Err(err) = chain_txn_sender
+        .send(ChainSendTransactionRequest::RegisterForeignChainConfig(
+            ChainRegisterForeignChainConfigArgs {
+                foreign_chain_configuration,
+            },
+        ))
+        .await
+    {
+        tracing::warn!(error = ?err, "failed to send register supported foreign chains transaction");
+    }
```
gilcu3, line 961: this is fine, but I wonder if we could make the node slightly smarter, in the sense that it can get the contract version, and do some matching around it to determine which transaction to send.

crates/node/src/providers/verify_foreign_tx/sign.rs
gilcu3, line 96: this is a bit of a blocker for me. I thought we had discussed to reuse the same mechanism that we use for when some nodes are down, which uses a fallback. Did you find any blocker for doing that?

gilcu3 (changes requested):
Thank you! 

Everything looks fine except for the way we are handling the new check in the provider. Let me know what's your take on this, as I thought we had agreed on a different solution.

crates/node/src/coordinator.rs
anodar: Probably, but this code will go away as soon as we deploy 3.13 so probably not worth it.

crates/node/src/indexer/fake.rs
anodar: No this is what I was referring to in the PR description by:

```
Extend FakeMpcContractState with Available Foreign Chains and Foreign Chains Configs:

Note: fake doesn't follow same flow for populating available foreign chains (in production that's via voting)
TODO(anodar): maybe worth changing ?!
```

To answer myself, I don't think this is worth changing although curious to hear your take on this. This whole FakeContractState is completely separate implementation and doesn't really test anything in the implementation of real contract state. So I figured there's no point duplicating voting logic in here rather just populate available chains with the most straightforward way (even if population differs from production behavior).  If we could somehow have contract state fields as interfaces and we could inject fakes/mocks there, then it would make sense. Production implementation of calculating available foreign chains is tested in e2e tests though. LMKWDYT

crates/node/src/indexer/fake.rs
anodar: Fake implements `ReadAvailableForeignChains` trait which has that signature which needs it because real one can fail.

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
anodar: Added in [f176e11](https://github.com/near/mpc/pull/3641/commits/f176e110e9043f597d41b5e695183f46d4f43ac3)

crates/e2e-tests/src/cluster.rs
anodar: Done.

crates/e2e-tests/src/cluster.rs
anodar: Done.

crates/contract/src/lib.rs
```diff
@@ -2173,9 +2173,7 @@ impl MpcContract {
         dtos::Config::from(&self.config)
     }
 
-    #[deprecated(
-        note = "use get_available_foreign_chains, see https://github.com/near/mpc/issues/3630"
-    )]
+    #[deprecated(note = "TODO(#3630): Drop after contract is upgraded to 3.13")]
     pub fn get_supported_foreign_chains(&self) -> dtos::SupportedForeignChains {
```
gilcu3: nit: the part "use get_available_foreign_chains" was useful to keep

crates/node/src/indexer/fake.rs
gilcu3: Yeah the idea here was to make the node testing much lighter than using a full blown contract, so I guess it makes sense to add anything where the node behavior matters. @kevindeforth what's your take on this?

gilcu3:
@claude review

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
gilcu3: ohh I see I was the one requesting aptos, lets fix this in a different PR, as I see it requires quite a few changes and not a one liner there

crates/contract/src/lib.rs

crates/node/src/requests/queue.rs
```diff
@@ -557,8 +557,15 @@ impl<RequestType: Request + Clone, ChainRespondArgsType: ChainRespondArgs>
...
+    ///
+    /// `refine_eligible_leaders` callback can be used to narrow the eligible
+    /// leader set (e.g. to those that support particular chain for this request).
     pub fn get_requests_to_attempt(
         &mut self,
+        refine_eligible_leaders: impl Fn(
+            &RequestType,
+            &HashSet<ParticipantId>,
+        ) -> HashSet<ParticipantId>,
     ) -> Vec<Arc<GenerationAttempt<RequestType, ChainRespondArgsType>>> {
```
gilcu3, line 589: we have a function called `eligible_leaders_and_maximum_height` defined in this file, and already being used inside this function. Unless it is not possible, it would be preferable to reuse than instead of inserting a callback here

crates/node/src/assets.rs
```diff
@@ -165,6 +165,23 @@ impl<T, CondVal: Default + Eq> ColdQueue<T, CondVal> {
...
+    pub(self) fn take_first_matching(&mut self, cond_val: &CondVal) -> Option<(UniqueId, T)> {
+        let pos = self
+            .cold_queue
+            .iter()
+            .position(|(_, value)| (self.condition)(cond_val, value))?;
+        let result = self.cold_queue.remove(pos);
+        self.cold_ready = 0;
+        self.cold_available = self.cold_queue.len();
+        result
+    }
```
gilcu3, line 184: I would prefer a solution that does not require modifying the internals of the queue. Was it not possible to add what we discussed without changing this? I don't have much time now to dig deeper, but at least initially I thought it was possible.

crates/node/src/assets.rs
```diff
@@ -504,6 +529,16 @@ where
...
+    pub fn take_owned_matching(&self, active: Vec<ParticipantId>) -> Option<(UniqueId, T)> {
+        let (id, asset) = self.owned_queue.take_owned_matching(active)?;
+        let mut update = self.db.update();
+        update.delete(self.col, &self.make_key(id));
+        update
+            .commit()
+            .expect("Unrecoverable error writing to database");
+        Some((id, asset))
+    }
+
```
gilcu3, line 556: same here, a solution where we do not modify this structs would be preferable, to avoid increasing the complexity here

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -60,8 +65,37 @@ where
...
+            .collect();
+
+        let (presignature_id, presignature) = domain_data
+            .presignature_store
+            .take_owned_matching(eligible_participants)
+            .with_context(|| {
+                format!(
+                    "no owned presignature whose participants all support chain {requested_chain:?}"
+                )
+            })?;
```
gilcu3: what happens here in the case where there is a good set of nodes that support the foreign chain, but we don't have (yet) a presignature for it? For standard signatures, we just try to generate it and wait till it is done. We should try to have the same here.

gilcu3 (commented):
Leaving some comments here, as anyway I hope this will move forward next week.

I am worried about the modifications to the queue. That part of the code is performance sensitive, so it would be good to avoid changing it in this PR, if possible. If it is required, then we will need to review that very carefully, to avoid introducing regressions.

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -60,8 +65,37 @@ where
         let domain_data = self
             .ecdsa_signature_provider
             .domain_data(foreign_tx_request.domain_id)?;
-        let (presignature_id, presignature) = domain_data.presignature_store.take_owned().await;
+        let foreign_chains_configs = self
+            .foreign_chain_policy_reader
+            .get_foreign_chains_configs()
```
kevindeforth: That's one request to the view client for every request we want to handle as a leader, I think that's too slow here.

Would be better to spawn a monitoring task and read this value from a watch channel. Whenever it changes, it gets updated.

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -119,7 +184,7 @@ where
         request: &dtos::ForeignChainRpcRequest,
         payload_version: dtos::ForeignTxPayloadVersion,
     ) -> anyhow::Result<dtos::ForeignTxSignPayload> {
-        chain_is_supported(&self.foreign_chain_policy_reader, request).await?;
+        is_chain_available(&self.foreign_chain_policy_reader, request).await?;
```
kevindeforth: and this is one per follower, but it is existing code...

kevindeforth (commented):
So, I haven't reviewed the full changes, because I wanted to focus on the performance remark made by @gilcu3 , and ended up tripping over the view client queries. Will do a full review later.

Querying the view client for every foreign chain transaction is not ideal. Main concern is contention, but it's also wasteful, because it's querying the same value multiple times per block. Those values do not change within one block. In fact, those values are expected to change very rarely.

I was surprised to see that this is a pattern that was previously introduced by the foreign chain transaction feature and not in this PR: https://github.com/near/mpc/blob/4fc15fa0caf72857e5f7d263acee9dc979c0ec13/crates/node/src/providers/verify_foreign_tx/sign.rs#L122

Before that, the pattern we used to follow in the node was pretty simple (albeit not ideal): We spawn a monitoring task and ensure we query at least once a block for the value. Any consumers hold a channel and in case the value changeds they get notified.

This made it quite easy to reason about the load we were putting on the view client. In the ForeignChainTx feature, the number of calls to the view client is now entirely driven by the number of requests, which is potentially many.

To put this into perspective:
If we have 200 requests per block, that translates to roughly 400 requests per second. That is 400 wasm executions per second, whereas it could be just 2.

We are currently far away from 200 requests per block, so I would be surprised if this lead to an actual problem. Additionally, the chain-gateway integration should take care of the frequent requests. So for me, this is not a blocker for this PR. But I don't quite understand the reason for this in the first place, so would be good to hear from @netrome why we chose to query once per request instead of following the previous pattern.

kevindeforth (commented):
_Please ignore, github glitch lead to my review being posted twice_

kevindeforth (changes requested):
Thank you very much.
Disclaimer: still didn't review everything, in this review, I am focusing on the queue logic, which seems to be the most complex piece of this implementation.

The high-level choices made in this PR seem sensible:
- keeping one queue for all foreign chains of the same domain 
- giving the option to refine the leader selection for the foreign transaction queue.

I think the way we go about the implementation can be optimized:
- In the current PR, we compute multiple times the participant set capable of engaging in the computation:
    - Once per request in the queue, in `PendingRequests::get_requests_to_attempt`:  https://github.com/near/mpc/blob/89522029971fe157637ee8f9a2144f860a7163ad/crates/node/src/requests/queue.rs#L624
    - Once per request as a foreign_tx_leader in `VerifyForeignTxProvider::make_verify_foreign_tx_leader`
      https://github.com/near/mpc/blob/89522029971fe157637ee8f9a2144f860a7163ad/crates/node/src/providers/verify_foreign_tx/sign.rs#L80
- There are multiple calls to the view client to get a view of current foreign chain configs and available foreign chains.
- there are unnecessary clones of participant sets for the non-foreign chain tx queues.

The following would be sufficient and arguably simpler:
1. We have a map `current_participants_by_foreign_chain`, `ForeiginChain --> {ParticipantId}`: mapping foreign chains to the participant set supporting them. Only chains that can reach a signing quorum are in that map.
2. We update that map if and only if the foreign chain config changes on-chain.
3. Before looping over all pending foreign chain requests, the queue computes the set of eligible participants by combining the above map with the network view of its participants (lets call this map `alive_participants_by_foreign_chain: ForeignChain -> {ParticipantId}`).
4. Now, for any given request, we can look-up the set of eligible leaders from `alive_participants_by_foreign_chain` without additional computation or view client queries and pass it to `QueuedRequest::process`. 
5. We use that map as an argument to `make_verify_foreign_tx_leader`, where it can is used to select the set of participants for the computation.


This way, we avoid unnecessary computation and view client queries.

anodar:
@claude review, check @kevindeforth's comment and validate whether all points are addressed.

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
```diff
@@ -39,6 +40,32 @@ struct MockServerUrls {
     polygon: Vec<String>,
 }
 
+fn build_providers_from_urls<'a>(
+    mut urls: impl Iterator<Item = (usize, &'a String)>,
+    chain_name: &str,
+) -> NonEmptyBTreeMap<RpcProviderName, ForeignChainProviderConfig> {
+    let (i, first_url) = urls
+        .next()
+        .expect("at least one polygon provider must be configured");
```
kevindeforth: If this is specifc to polygon, consider renaming the method `build_polygon_providers_from_url`. Or, make it generic and just state that at least one url must be provided.

crates/e2e-tests/src/cluster.rs
```diff
@@ -889,6 +919,63 @@ impl MpcCluster {
...
+        let placeholder = ChainEntry {
+            providers,
+            quorum: 1,
+        };
+
+        let mut iter = chains.iter().copied();
+        let first = iter
+            .next()
+            .context("whitelist_foreign_chains: chains must be non-empty")?;
```
kevindeforth, line 959: why does this method need to fail in case `chains` is empty? Couldn't it just return `Ok(())`?

crates/near-mpc-contract-interface/src/method_names.rs
```diff
@@ -78,7 +78,9 @@ pub const GET_PENDING_CKD_REQUEST: &str = "get_pending_ckd_request";
 pub const GET_PENDING_VERIFY_FOREIGN_TX_REQUEST: &str = "get_pending_verify_foreign_tx_request";
 pub const GET_TEE_ACCOUNTS: &str = "get_tee_accounts";
 pub const GET_ATTESTATION: &str = "get_attestation";
+#[deprecated(note = "TODO(#3630): Drop after contract is upgraded to 3.13")]
 pub const GET_SUPPORTED_FOREIGN_CHAINS: &str = "get_supported_foreign_chains";
+#[deprecated(note = "TODO(#3630): Drop after contract is upgraded to 3.13")]
 pub const GET_FOREIGN_CHAIN_SUPPORT_BY_NODE: &str = "get_foreign_chain_support_by_node";
```
kevindeforth, line 84: hmm, why can this be dropped only after upgrading the contract to 3.13? I thought we upgrade the nodes first, so after this PR, there shouldn't be anything calling these endpoints anymore.

crates/node/src/indexer/fake.rs
```diff
@@ -49,24 +49,35 @@ pub struct FakeMpcContractState {
     pub pending_signatures: BTreeMap<Payload, SignatureId>,
     pub pending_ckds: BTreeMap<dtos::CkdAppId, CKDId>,
     pub pending_verify_foreign_txs: BTreeMap<dtos::ForeignChainRpcRequest, VerifyForeignTxId>,
-    supported_foreign_chains: dtos::SupportedForeignChains,
+    available_foreign_chains: dtos::AvailableForeignChains,
+    // TODO(#3630): drop this once 3.13 contract is deployed.
     supported_foreign_chains_by_node: dtos::ForeignChainSupportByNode,
```
kevindeforth, line 54: Hmm, why do we keep testing the deprecated code path?

crates/node/src/indexer/handler.rs
```diff
@@ -278,6 +291,35 @@ async fn handle_message(
...
+        return false;
+    };
+    if !FOREIGN_CHAIN_POLICY_METHODS.contains(&method_name.as_str()) {
+        return false;
+    }
+    matches!(
+        execution_outcome.outcome.status,
+        ExecutionStatusView::SuccessValue(_) | ExecutionStatusView::SuccessReceiptId(_)
+    )
+}
```
kevindeforth: Hmm, this is very interesting. It's a deviation from the current pattern of monitoring finalized contract state via view client, but allows a natural notification mechanism, so it is not without merit. Issue is just that these transactions may not be finalized and pattern matching is a bit more expensive.

I need to give this a bit more thought.

_continued discussion in https://github.com/near/mpc/pull/3641#discussion_r3549368869_

crates/node/src/indexer/tx_sender.rs
```diff
@@ -318,7 +318,8 @@ async fn observe_tx_result(
         | VoteAbortKeyEventInstance(_)
         | VerifyTee()
         | ConcludeNodeMigration(_)
-        | RegisterForeignChainConfig(_) => Ok(TransactionStatus::Unknown),
+        | RegisterForeignChainConfig(_) // TODO(#3630): drop once contract is upgraded to 3.13
```
kevindeforth, line 321: same question, why can't we drop this already?

kevindeforth (commented):
intermediate review, will pick this up later.

crates/e2e-tests/src/cluster.rs
```diff
@@ -46,6 +47,7 @@ const SIGN_DEPOSIT: near_kit::NearToken = near_kit::NearToken::from_yoctonear(1)
 const KEY_EVENT_TIMEOUT_BLOCKS: u64 = 240;
 const CONTRACT_UPDATE_DEPOSIT: near_kit::NearToken = near_kit::NearToken::from_millinear(17_000);
 const CONTRACT_UPDATE_GAS: near_kit::Gas = near_kit::Gas::from_tgas(300);
+const VOTE_FOREIGN_CHAIN_GAS: near_kit::Gas = near_kit::Gas::from_tgas(30);
```
pbeza, line 50: Was this value benchmarked, or is it just an estimate?

crates/e2e-tests/src/cluster.rs
```diff
@@ -889,6 +919,63 @@ impl MpcCluster {
...
+
+        let mut iter = chains.iter().copied();
+        let first = iter
+            .next()
+            .context("whitelist_foreign_chains: chains must be non-empty")?;
+        let mut batch: NonEmptyBTreeMap<ForeignChain, ChainEntry> =
+            NonEmptyBTreeMap::new(first, placeholder.clone());
+        for chain in iter {
+            batch.insert(chain, placeholder.clone());
+        }
```
pbeza:
```suggestion
        let batch: NonEmptyBTreeMap<ForeignChain, ChainEntry> = chains
            .iter()
            .map(|&chain| (chain, placeholder.clone()))
            .collect::<BTreeMap<_, _>>()
            .try_into()
            .context("whitelist_foreign_chains: chains must be non-empty")?;
```

crates/e2e-tests/src/cluster.rs
```diff
@@ -889,6 +919,63 @@ impl MpcCluster {
             .await
     }
 
+    pub async fn whitelist_foreign_chains(
+        &self,
+        participant_indices: &[usize],
+        chains: &BTreeSet<ForeignChain>,
+    ) -> anyhow::Result<()> {
```
pbeza, line 926: Nit: Isn’t it a bit confusing that all chains get the same set of provider names and URLs? I think it’s fine for the test fixture, but it took me a while to realize we mock all chains the same way.

crates/e2e-tests/src/cluster.rs
```diff
@@ -889,6 +919,63 @@ impl MpcCluster {
             .await
     }
 
+    pub async fn whitelist_foreign_chains(
+        &self,
+        participant_indices: &[usize],
```
pbeza, line 924: I think passing around raw indices in this file is pre-existing, but it’s a bit unfortunate. We should not need to care here that operators, or their keys/state, are stored in an array/vector. Instead, we should be passing the items behind those indices/keys directly.

crates/e2e-tests/src/cluster.rs
```diff
@@ -889,6 +919,63 @@ impl MpcCluster {
...
+            ProviderId("quicknode".to_string()),
+            provider("https://example2.com/"),
+        );
+        providers.insert(
+            ProviderId("public".to_string()),
+            provider("https://example3.com/"),
+        );
+        let placeholder = ChainEntry {
+            providers,
+            quorum: 1,
```
pbeza, line 951: General question, not specific to this code, but about the RPC quorum we use for foreign chain verification in general: I assume the quorum means that this many RPCs must be available and return the same response. What happens if there are transient errors, like unavailable URLs, or invalid API tokens? In that case, can the effective quorum be lower while verification still passes?

So, to attack foreign chain verification for a particular chain, would someone need to compromise one RPC URL for that chain and DDoS all the other RPCs for that same chain, regardless of the quorum set in the contract?

pbeza: One more comment that came out of my convo with Claude: apparently Aptos is the exception to “transient errors are dropped.”

It classifies deterministic `4xx` responses, like `400`/`401`/`403`/`410`, for example a bad token, as a substantive `RpcRequestRejected`, while transport errors and `429`/`5xx` still stay transient.

On EVM/Bitcoin/Starknet, a bad token and a bad URL both collapse to transient errors and get dropped. So one bad-token Aptos provider can trigger `InspectorResponseMismatch` and block an otherwise healthy verification, while the same misconfig is silently masked on the other chains.

Something to double-check when you get a chance. Not a blocker since, IIUC, this was not introduced in this PR.

crates/e2e-tests/tests/common.rs
```diff
@@ -88,6 +89,13 @@ pub async fn must_setup_cluster(
         panic!("expected Running state");
     };
 
+    if !whitelisted_chains.is_empty() {
```
pbeza: It would be more natural to remove the guard and make `whitelist_foreign_chains` tolerate an empty set.

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
```diff
@@ -39,6 +40,32 @@ struct MockServerUrls {
     polygon: Vec<String>,
 }
 
+fn build_providers_from_urls<'a>(
+    mut urls: impl Iterator<Item = (usize, &'a String)>,
+    chain_name: &str,
+) -> NonEmptyBTreeMap<RpcProviderName, ForeignChainProviderConfig> {
```
pbeza:
```suggestion
fn build_providers_from_urls(
    urls: &[String],
    chain_name: &str,
) -> NonEmptyBTreeMap<RpcProviderName, ForeignChainProviderConfig> {
    let mut urls = urls.iter().enumerate();
```

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
```diff
@@ -39,6 +40,32 @@ struct MockServerUrls {
     polygon: Vec<String>,
 }
 
+fn build_providers_from_urls<'a>(
```
pbeza: I think we can shorten this method to something like:

```rust
fn build_providers_from_urls(
    urls: &[String],
    chain_name: &str,
) -> NonEmptyBTreeMap<RpcProviderName, ForeignChainProviderConfig> {
    let map: BTreeMap<_, _> = urls
        .iter()
        .enumerate()
        .map(|(i, url)| {
            let cfg = ForeignChainProviderConfig { rpc_url: url.clone(), auth: Default::default() };
            (format!("mock-{chain_name}-{i}").into(), cfg)
        })
        .collect();
    map.try_into()
        .unwrap_or_else(|_| panic!("at least one {chain_name} provider must be configured"))
}
```

pbeza (commented):
Partial review for now.

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
pbeza: ➕

I don’t think it’s Polygon-specific.

crates/near-mpc-contract-interface/src/method_names.rs
anodar: Because these aren't yet in contract version 3.12. So if we update node to have this changes and contract is still at 3.12 node won't be able to talk to the new API. 

Besides I'd like to switch to new API in one PR and get rid of deprecated API in another. It's already taking too long to get this merged as is, would like to keep changes at minimum.

crates/e2e-tests/src/cluster.rs
anodar: It's a design decision, if you call whitelist_foreign_chains you should pass non-empty list.

crates/node/src/indexer/fake.rs
anodar: Because it's not yet deprecated. That is what we currently use. They are marked for deprecation once we switch to the new API.

Also I'd like to get rid of all deprecated API at once only after nodes start to use new API.

crates/node/src/indexer/tx_sender.rs
anodar: Same answer as above.

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
anodar: Reworded in [3f7066d](https://github.com/near/mpc/pull/3641/commits/3f7066d043c6d84b3674df4761b9b0244ae6eb5b)

anodar:
Deleted some of the old claude comments as it's cluttering feed.

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
```diff
@@ -199,6 +207,19 @@ async fn setup_foreign_tx_cluster() -> anyhow::Result<ForeignTxTestEnv> {
...
+    let expected_chains: std::collections::BTreeSet<ForeignChain> = [
+        ForeignChain::Bitcoin,
+        ForeignChain::Abstract,
+        ForeignChain::Bnb,
+        ForeignChain::Starknet,
+        ForeignChain::Base,
+        ForeignChain::Arbitrum,
+        ForeignChain::HyperEvm,
+        ForeignChain::Polygon,
+    ]
```
pbeza, line 210: IIRC, I’ve seen this kind of test a couple of times, where we list some chains manually, similar to here, and then forget to update the list over time.

Maybe we can extract this into a more visible test config so it’s easier to spot? Alternatively, we could collect all the types and **explicitly** exclude _Solana_, _Ethereum_, and _TON_. BTW, why are we excluding those?

Vibe-coded PoC: https://github.com/near/mpc/pull/3753.

crates/node/src/indexer/fake.rs
```diff
@@ -161,14 +177,54 @@ impl FakeMpcContractState {
...
+            return;
+        };
+        self.foreign_chains_configs
+            .insert(info.tls_public_key.clone(), foreign_chains_config);
+
+        let foreign_chains_configs = &self.foreign_chains_configs;
+        self.available_foreign_chains = state
+            .parameters
+            .participants()
+            .participants()
```
pbeza: Copy-pasted `.participants().participants()` chain.

```suggestion
        let participants = state.parameters.participants().participants();
        let Some((_, _, info)) = participants.iter().find(|(id, _, _)| id == &account_id) else {
            tracing::info!(
                "register_foreign_chains_config transaction ignored because signer is not a participant"
            );
            return;
        };
        self.foreign_chains_configs
            .insert(info.tls_public_key.clone(), foreign_chains_config);

        let foreign_chains_configs = &self.foreign_chains_configs;
        self.available_foreign_chains = participants
```

crates/node/src/indexer/fake.rs
pbeza: I was about to raise a similar concern to @gilcu3, but just noticed it’s already been raised.

If we want to leave this as-is, can we add a comment explaining this non-trivial expression and why it deviates from how the real production contract computes available chains? I think it deserves a comment even without the deviation, since it’s pretty non-trivial.

crates/node/src/indexer/fake.rs
```diff
@@ -161,14 +177,54 @@ impl FakeMpcContractState {
...
+            .iter()
+            .map(|(_, _, p_info)| {
+                foreign_chains_configs
+                    .get(&p_info.tls_public_key)
+                    .map(|c| c.iter().copied().collect::<BTreeSet<_>>())
```
pbeza:
```suggestion
                    .map(|c| (**c).clone())
```

crates/node/src/mpc_client.rs
```diff
@@ -552,7 +565,52 @@ where
...
+                match self
+                    .verify_foreign_tx_provider
+                    .refresh_foreign_chain_policy()
+                    .await
+                {
+                    Ok(queried_height) => {
+                        if queried_height >= refresh_until_height {
+                            refresh_chain_policy_until_height = None;
+                        }
+                    }
```
pbeza: Optional nit.

```suggestion
                    Ok(queried_height) if queried_height >= refresh_until_height => {
                        refresh_chain_policy_until_height = None;
                    }
                    Ok(_) => {}
```

crates/node/src/requests/queue.rs
```diff
@@ -956,38 +972,56 @@ mod tests {
         setup.update(&mut pending_requests);
 
         // Then: req1 is not attempted because we're not the leader. req2 is attempted.
-        let to_attempt1 = pending_requests.get_requests_to_attempt();
+        let to_attempt1 =
+            pending_requests.get_requests_to_attempt(|_, eligible| Cow::Borrowed(eligible));
```
pbeza: By Claude:

> The "no refinement" closure `|_, eligible| Cow::Borrowed(eligible)` is repeated ~27× in `queue.rs` tests and 2× in `mpc_client.rs` production. Bind once as a `const`/`fn` (e.g. `fn no_refinement(...)` or a local) and reuse. Reduces the biggest source of repetition the PR introduced.

crates/node/src/mpc_client.rs
```diff
@@ -552,7 +565,52 @@ where
                 );
             }
 
-            let verify_foreign_tx_attempts = pending_verify_foreign_txs.get_requests_to_attempt();
+            if let Some(refresh_until_height) = refresh_chain_policy_until_height {
+                match self
+                    .verify_foreign_tx_provider
+                    .refresh_foreign_chain_policy()
+                    .await
```
kevindeforth: We should not await in this loop, otherwise we risk delaying the processing of block updates. This is a blocker.

crates/node/src/mpc_client.rs
```diff
@@ -552,7 +565,52 @@ where
...
+                            .collect()
+                    });
+                    Cow::Owned(
+                        alive_supporters
+                            .get(&req.request.chain())
+                            .cloned()
+                            .unwrap_or_default(),
+                    )
+                })
+            };
```
kevindeforth: I think the idea here is commendable - we want to avoid needless queryies to the view client. But there are serious downsides to this:
- it makes the code more complex. now, not only does the node need to know how to query the foreign transaction policy, but it also need to be aware of all the endpoints that may result in a state change on the contract. That is super hard to maintain, as it means the node will need to be aware of contract logic.
- it's yet another pattern that devs of this crate need to be familiar with. It's a pattern that is easy to miss. If I add a contract endpoint that changes the policy state, I now also need to be aware of doing node changes.
- It's optimizing for a problem we don't necessarily have. We don't know if view client queries are the bottleneck and how they compare to block event matching.

This is a blocker, I would prefer we stuck with the existing pattern and spawned a monitoring task for the view method.

crates/node/src/requests/queue.rs
kevindeforth: @anodar I think you missed this comment by @gilcu3.
Is there a way we can integrate this in the existing methods (maybe during construction of the queue?).
If not, then I would advise to make this an optional parameter, such that we can remove the cow.

kevindeforth (changes requested):
Lets stick with a monitoring task over the view client. Listening on block events is introducing a lot of dependency between node and contract that will be hard to maintain.

crates/node/src/mpc_client.rs
anodar: Changed to monitoring task in [7f20efe](https://github.com/near/mpc/pull/3641/commits/7f20efef2f0aa34c50013bc8075e1f98022a3daf)

crates/node/src/mpc_client.rs
anodar: No longer applicable as I switched to separate monitoring task for foreign chain config updates.

crates/node/src/requests/queue.rs
anodar: Addressed in [25e9ea1](https://github.com/near/mpc/pull/3641/commits/25e9ea1091ba3ea7cf7705bab4deaa9cef526448)

crates/node/src/assets.rs
anodar: I can't think of a way to do that without modifying this or alternatively having queues per chain which seems lot more significant change and then they wouldn't be sharing presignatures pool either.

crates/node/src/assets.rs
anodar: Same as above.

crates/e2e-tests/src/cluster.rs
anodar: Good observation, yes we do drop transient errors. This whole approach will changes as we start enforcing whitelist and require that at least quorum number of providers agree on the result. 

I'd like to keep that change outside this pr.

crates/node/src/providers/verify_foreign_tx/sign.rs
anodar: Modified to follow the same pattern. [82417c2](https://github.com/near/mpc/pull/3641/commits/82417c274590f5d3b2446858a8ebc0c70bc3202b)

crates/e2e-tests/src/cluster.rs
anodar: It's just estimate not a benchmark.

crates/e2e-tests/src/cluster.rs
anodar: Done in: [9a45ba4](https://github.com/near/mpc/pull/3641/commits/9a45ba4f178b6be30b0522768dbb5e29c94bbafe)

crates/e2e-tests/src/cluster.rs
anodar: Index of the node is also the ID of the node which is invariant in the cluster setup. I'd like to keep this changed aligned with that.

crates/e2e-tests/src/cluster.rs
anodar: Added comment about it in: [9a45ba4](https://github.com/near/mpc/pull/3641/commits/9a45ba4f178b6be30b0522768dbb5e29c94bbafe)

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
anodar: Created follow up task: https://github.com/near/mpc/issues/3762, I'd rather keep that separate from this changes.

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
anodar: Addressed in [21c44da](https://github.com/near/mpc/pull/3641/commits/21c44da4087aabf5daa2e5ce80d1e22dd54099d3)

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
anodar: Done in: [21c44da](https://github.com/near/mpc/pull/3641/commits/21c44da4087aabf5daa2e5ce80d1e22dd54099d3)

crates/node/src/requests/queue.rs
anodar: No longer applicable.

crates/node/src/mpc_client.rs
anodar: Addressed in [6e2f809](https://github.com/near/mpc/pull/3641/commits/6e2f809831a9d852ba5f56e72b7bd600ef67cead)

crates/node/src/indexer/fake.rs
anodar: Addressed in [6e2f809](https://github.com/near/mpc/pull/3641/commits/6e2f809831a9d852ba5f56e72b7bd600ef67cead)

crates/node/src/indexer/fake.rs
anodar: Added comment about it in [6e2f809](https://github.com/near/mpc/pull/3641/commits/6e2f809831a9d852ba5f56e72b7bd600ef67cead)

crates/e2e-tests/tests/common.rs
anodar: Addressed.

crates/node/src/indexer/fake.rs
anodar: Done.

pbeza (dismissed):
I spent a couple of hours doing a deeper review yesterday. Since you addressed most of the comments, and @kevindeforth already did a good deep review too, I only did a relatively shallow re-review today. LGTM.

In retrospect, I wish we could have made this PR smaller to make it easier to review. I also have a [similarly sized PR](https://github.com/near/mpc/pull/3714) that was hard to split up, so I guess this was a similar case here. We just need to live with it and review these huge PRs as-is sometimes.

crates/node-config/src/lib.rs
```diff
@@ -146,6 +146,11 @@ pub struct ConfigFile {
     pub keygen: KeygenConfig,
     #[serde(default)]
     pub foreign_chains: ForeignChainsConfig,
+    /// How often the node re-reads the contract's foreign-chain policy (per-node
+    /// configs and available chains). Until the next refresh, verify_foreign_tx
+    /// requests are checked against the previously cached policy.
+    #[serde(default = "default_foreign_chain_policy_refresh_interval_sec")]
+    pub foreign_chain_policy_refresh_interval_sec: u64,
```
kevindeforth: This should not be in the config, this is a near-blockchain specific value that needs to be tracked internally. We should default to the same value we use for tracking other contract state changes.

crates/node/src/assets.rs
gilcu3: I might give that a try later, as I always thought the change would not be very big.

> then they wouldn't be sharing presignatures pool either.

Afaik they cannot share the presignatures pool, as presignatures depend on the secret keys which are unique for each domain.

crates/node/src/indexer/fake.rs
kevindeforth: I completely missed this thread.

We have `ProtocolContractState` in the fake indexer, so we could, in a similar fashion, add  `ForeignChainsMetadata`.

https://github.com/near/mpc/blob/087f098f4c7b99ea3429b2c2cd1f380c1d9c5318/crates/node/src/indexer/fake.rs#L43

Now, I am not sure if that's completely sensible, because we prefer to avoid direct dependencies between the node and the contract, as otherwise, changes in the contract may require changes in the node, which slows down contract development speed.

That being said, we are too restrictive here. We definitely want an integration test to check if the nodes can sign with only partial signing quorum:
- Alice supporting chains A + B
- Bob supporting chains B + C
- Charlie supporting chains A +B + C

If this fails in the e2e test, its unclear whether the issue is on the contract or the node side.

kevindeforth (commented):
Still not done fully reviewing this PR. Just chiming in on an existing thread and objecting to a config value.

crates/node-config/src/lib.rs
anodar: This is in the config so that we can change it in e2e tests where we don't want it to be 60 second. We can't have it constant.

anodar:
> Now, I am not sure if that's completely sensible, because we prefer to avoid direct dependencies between the node and the contract, as otherwise, changes in the contract may require changes in the node, which slows down contract development speed.

not sure what you mean by that, but we're not introducing any new dependency.


> That being said, we are too restrictive here. We definitely want an integration test to check if the nodes can sign with only partial signing quorum:

> Alice supporting chains A + B
> Bob supporting chains B + C
> Charlie supporting chains A +B + C
> If this fails in the e2e test, its unclear whether the issue is on the contract or the node side.

Can follow up on this in subsequent PR.

crates/node-config/src/lib.rs
kevindeforth: Every 60 seconds is a bit rare, I think we query every 200-400 ms to make sure we get the state for each new block.
Expanding the config just for e2e tests is not a great solution. If it has to be 60 seconds in production and less in tests, we can use constants and compiler-flags.

kevindeforth:
> not sure what you mean by that, but we're not introducing any new dependency.

That's not what I was saying. I was saying that we already are using a struct from the smart contract and that already draws in a dependency we don't want. So depending on the foreign chain policy struct in the smart contract would be kind of fine, but not ideal, as it goes against our long-term goal of removing that dependency.

Given that you responded here instead of in the thread, I suspect you were missing the context, so linking the thread here: https://github.com/near/mpc/pull/3641#discussion_r3467128236

crates/e2e-tests/tests/common.rs
```diff
@@ -88,6 +89,11 @@ pub async fn must_setup_cluster(
         panic!("expected Running state");
     };
 
+    cluster
+        .whitelist_foreign_chains(&initial_participant_indices, &whitelisted_chains)
+        .await
+        .expect("failed to whitelist foreign chains");
+
```
kevindeforth, line 96: Does this line, together with [this](https://github.com/near/mpc/pull/3641/changes#r3542223117), mean that we need to have whitelisted foreign chains for all our test clusters?

If so, doesn't this make our testing setup unnecessarily restrictive?

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -86,6 +115,36 @@ where
...
+    /// Snapshot of the cached supporters-by-chain map.
+    pub(crate) fn participants_by_foreign_chain(
+        &self,
+    ) -> Arc<BTreeMap<contract_dtos::ForeignChain, HashSet<ParticipantId>>> {
+        self.foreign_chain_policy
+            .read()
+            .unwrap()
+            .participants_by_chain
+            .clone()
+    }
```
kevindeforth, line 146: These look out of place.
`provides/<name>/sign.rs` are generally not concerned with fetching ephemeral information from chain, but supposed to execute the signing logic.

I would suggest to move these to a different place.

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -60,8 +68,29 @@ where
...
+            .participants_by_chain
+            .get(&requested_chain)
+            .cloned()
+            .unwrap_or_default();
+        let eligible_participants: Vec<ParticipantId> = self
+            .ecdsa_signature_provider
+            .alive_participant_ids()
+            .into_iter()
+            .filter(|participant_id| chain_supporters.contains(participant_id))
+            .collect();
```
kevindeforth, line 86: some helper methods would be nice for these, just to have less boilerplate in these methods.

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -348,42 +410,85 @@ where
...
-    let on_chain_foreign_chains_support = policy_reader
-        .get_supported_chains()
-        .await
-        .map_err(ForeignChainSupportError::FailedToReadContract)?;
-
-    let requested_chain = request.chain();
-
-    if on_chain_foreign_chains_support.contains(&requested_chain) {
+) -> Result<(), ChainNotAvailableError> {
+    let requested = request.chain();
```
kevindeforth, line 424: nit: I think this method would read nicer if it took as input the chain, not the entire request.

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -60,8 +68,29 @@ where
         let domain_data = self
             .ecdsa_signature_provider
             .domain_data(foreign_tx_request.domain_id)?;
-        let (presignature_id, presignature) = domain_data.presignature_store.take_owned().await;
+        let requested_chain = foreign_tx_request.request.chain();
+
+        let chain_supporters = self
+            .foreign_chain_policy
+            .read()
+            .unwrap()
```
kevindeforth, line 76: We have a [don't-panic policy](https://github.com/near/mpc/blob/main/docs/engineering-standards.md#dont-panic). These are easy to miss. An `expect` with reasoning why we don't expect a panic is preferred.

crates/node/src/mpc_client.rs
```diff
@@ -180,6 +180,23 @@ where
...
+                    if let Err(err) = provider.refresh_foreign_chain_policy().await {
+                        tracing::warn!(
+                            ?err,
+                            "failed to refresh foreign-chain policy, keeping the previous view"
+                        );
+                    }
+                    sleep(refresh_interval).await;
+                }
+            })
+        };
```
kevindeforth, line 198: Hmm, this function is already quite cluttered: it spawns some monitoring tasks, asset generation background tasks, some comms stuff  and metrics.

Now, on top of that, we need it to be aware:
- of block-time constants in the indexer
- that the provider requires to call `refersh_foreign_chain_policy()` at a specific interval
- severity of errors when calling `refresh_foreign_chain_policy()`

That's not ideal. I think we should move most of this logic to a different place.

I will make a proposal.

crates/node/src/mpc_client.rs
```diff
@@ -247,6 +265,20 @@ where
             self.client.my_participant_id(),
             self.client.clone(),
         );
+        // Foreign-tx leader election is narrowed per request to the participants
+        // supporting the request's chain, per the provider's cached policy.
+        let eligible_leaders_refiner: EligibleLeadersRefiner<VerifyForeignTxRequest> = {
+            let provider = self.verify_foreign_tx_provider.clone();
```
kevindeforth, line 271: we don't need to pass the whole provider in here, I think the foreign chain policy reader should be sufficient.

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -86,6 +115,36 @@ where
...
+    /// Re-reads the contract's foreign-chain configs and available chains and updates
+    /// the cached [`ForeignChainPolicy`].
+    pub(crate) async fn refresh_foreign_chain_policy(&self) -> anyhow::Result<()> {
+        let (configs, available_chains) = tokio::try_join!(
+            self.foreign_chain_policy_reader
+                .get_foreign_chains_configs(),
+            self.foreign_chain_policy_reader.get_available_chains(),
+        )?;
+        let participants_by_chain = Arc::new(supporters_by_foreign_chain(
+            self.ecdsa_signature_provider.participants_config(),
```
kevindeforth, line 127: This is now the only place where we need `foreign_chain_policy_reader`. We can simplify this by moving this, together with the reader, into a new struct and then only going through `ForeignChainPolicy`. Will make a draft to show what I mean.

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -86,6 +115,36 @@ where
...
+            &configs,
+        ));
+        *self.foreign_chain_policy.write().unwrap() = ForeignChainPolicy {
```
kevindeforth, line 130: Here's another panic. `expect(...)` is [preferred](https://github.com/near/mpc/blob/main/docs/engineering-standards.md#dont-panic), but I think we can avoid the panic altogether. This is the only place where we write and we could move it out of here to not require the lock.

crates/node/src/providers/verify_foreign_tx.rs
```diff
@@ -128,12 +124,25 @@ impl ForeignChainInspectors<HttpClient> {
...
+/// Cached view of the contract's foreign-chain policy.
+#[derive(Default)]
+pub(crate) struct ForeignChainPolicy {
+    /// The contract's available chains: whitelisted and supported by a signing quorum.
+    /// Checked before executing any verify_foreign_tx request.
+    pub(crate) available_chains: dtos::AvailableForeignChains,
+    /// Participants supporting each foreign chain, derived from the per-node configs,
+    /// chains without a signing quorum are omitted. Drives leader election and leader
+    /// presignature selection. Behind `Arc` so snapshots don't clone the map.
+    pub(crate) participants_by_chain: Arc<BTreeMap<dtos::ForeignChain, HashSet<ParticipantId>>>,
```
kevindeforth, line 136: These code-comments should be removed, because they are paraphrasing the code or explaining where it is used.

crates/contract/src/lib.rs
```diff
@@ -2173,6 +2173,9 @@ impl MpcContract {
         dtos::Config::from(&self.config)
     }
 
+    #[deprecated(
+        note = "use get_available_foreign_chains, TODO(#3630): Drop after contract is upgraded to 3.13, "
+    )]
```
kevindeforth, line 2178: note: we are a bit inconsistent in this PR wrt deprecation and deletion - in the node, we delete the code, in e2e tests and contract, we deprecate.

I think deprecation in contract is fine, albeit unnecessary, but I don't see the point for the e2e tests.

crates/node/src/requests/queue.rs
```diff
@@ -614,7 +658,16 @@ impl<RequestType: Request + Clone, ChainRespondArgsType: ChainRespondArgs>
                 block_height = %request.block_height,
             )
             .entered();
-            match request.process(self.my_participant_id, &eligible_leaders, cutoff_block, now) {
+            let request_eligible_leaders = refined_eligibility
```
kevindeforth, line 661: then here, we could just access the once-computed map, without locking.

crates/node/src/requests/queue.rs
```diff
@@ -583,9 +610,26 @@ impl<RequestType: Request + Clone, ChainRespondArgsType: ChainRespondArgs>
...
+        // loop below borrows `self.requests` mutably. `None` when the queue has no
+        // refiner: every request then uses `eligible_leaders` as-is.
+        let refined_eligibility: Option<HashMap<RequestId, HashSet<ParticipantId>>> =
+            self.refine_eligible_leaders.is_some().then(|| {
+                self.requests
+                    .iter()
+                    .map(|(id, request)| {
+                        (
+                            *id,
+                            self.eligible_leaders_and_maximum_height(Some(&request.request))
```
kevindeforth, line 626: Note that each one of these calls locks the mutex, that's not ideal.

I think it might make sense to just fetch the map from chain id to participant here and then use that throughout.

(it's a bit more complicated, because this is a generic interface, but we could make the refinement method a struct implemening a trait, allowing to `update` and `fetch`).

crates/node/src/cli.rs
```diff
@@ -483,6 +483,7 @@ mod testing {
             my_near_account_id: participant.clone(),
             near_responder_account_id: responder.clone(),
             number_of_responder_keys: 1,
+            foreign_chain_policy_refresh_interval_sec: 1,
```
kevindeforth: we can delete this.

crates/node/src/assets.rs
kevindeforth, line 1: still need to look at this file.

kevindeforth (commented):
Some more comments. I will stack a PR on this with suggested changes.

crates/e2e-tests/tests/common.rs
kevindeforth: No, apologies, `whitelist_foreign_chains` actually returns `Ok())` if chains is empty.

kevindeforth:
Quick summary before the split, so we don't lose track of open items:

While my draft PR [#3803](https://github.com/near/mpc/pull/3803) addresses most of the issues, there are still three things that the split will need to address separately:

- the inefficiency in `queue.rs` - we shouldn't need to lock that often, instead, we should probably go with some sort of `.update_eligible_leabers()` mechanism, that we call before processing a batch [(comment)](https://github.com/near/mpc/pull/3803#discussion_r3579338010)
- the indexer API can be simplified to a single channel, not two [(comment)](https://github.com/near/mpc/pull/3803#discussion_r3580106174)
- the conflict with domain-thresholds PR [#3640](https://github.com/near/mpc/pull/3640)

Given that my draft is already ~1k lines of change and only addresses part of the issues, I agree that splitting is the right approach rather than merging as is.
Suggestion for the split: pull directly from my draft branch where it applies, so we don't redo work that's already reviewed.

anodar:
Closing this in favor of opening stacked prs.

PR #3647, opened 2026-06-23
fix: indexer getting stuck when syncing from genesis

### LINKED ISSUE #3623: MPC indexer wedges at genesis on a freshly state-synced node
### Background

On a node that state-syncs from scratch, the indexer never advances and no sign requests are captured. At first process start the embedded neard is at genesis, so the indexer — `sync_mode: Latest` (`LatestSynced`) with `await_for_node_synced: StreamWhileSyncing` (`crates/node/src/indexer/configs.rs`) — pins its cursor to the genesis height. neard then state-syncs to the chain tip, leaving the indexer stranded below the node's block tail, which it can never reach.

Observed on a fresh testnet node (neard fully synced, indexer frozen):

```
mpc_indexer_latest_block_height  42376888      # testnet genesis, delta 0 over time
near_block_height_head           255409058
near_tail_height                 255397255      # node has no blocks below this
```

The contract is read correctly via RPC (keygen succeeds); only request capture is affected. Restarting the task after neard is fully synced works around it: `LatestSynced` then re-evaluates to the current tip.

### Acceptance Criteria

- A node that state-syncs from scratch starts the indexer at (or after) the node's available block tail, with no manual restart.
- The indexer cursor advances toward the chain tip on first boot of such a node.

gilcu3:
Closes #3623

### DIFF
--- a/crates/chain-gateway/src/chain_gateway.rs
+++ b/crates/chain-gateway/src/chain_gateway.rs
@@ -252,8 +252,6 @@ struct StreamerSetup {
     near_config: NearConfig,
 }
 
-/// Races `sync` against `shutdown`. Returns `true` if `sync` completed first,
-/// `false` if `shutdown` resolved first.
 async fn await_sync_or_shutdown(
     sync: impl Future<Output = ()>,
     shutdown: impl Future<Output = ()>,
@@ -265,7 +263,7 @@ async fn await_sync_or_shutdown(
 }
 
 #[cfg(test)]
-#[expect(non_snake_case)] // tests follow `<system_under_test>__should_<assertion>` convention
+#[expect(non_snake_case)]
 mod tests {
     use super::await_sync_or_shutdown;
     use std::future::pending;
--- a/crates/chain-gateway/src/chain_gateway.rs
+++ b/crates/chain-gateway/src/chain_gateway.rs
@@ -1,3 +1,4 @@
+use std::future::Future;
 use std::path::Path;
 
 use near_account_id::AccountId;
@@ -153,7 +154,7 @@ fn run_node(
     ready_sender: tokio::sync::oneshot::Sender<RunNodeResult>,
     near_config: nearcore::NearConfig,
     home_dir: &Path,
-    shutdown_receiver: tokio::sync::oneshot::Receiver<()>,
+    mut shutdown_receiver: tokio::sync::oneshot::Receiver<()>,
     streamer_setup: Option<StreamerSetup>,
 ) {
     let rt = tokio::runtime::Builder::new_multi_thread()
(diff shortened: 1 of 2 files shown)


crates/node/src/indexer/real.rs
```diff
@@ -105,8 +106,6 @@ pub fn spawn_real_indexer(
 
             let indexer = Indexer::from_near_node(near_indexer_config, near_config, &near_node);
 
-            let stream = indexer.streamer();
```
kevindeforth, line 113: IIRC, a lot of the other methods are also depending on the indexer being in full sync. Did you encounter any issues when blocking here for full sync?

gilcu3: We discussed in slack that this could be done by blocking, leading to a simpler solution. Reworked the PR in [396c97b](https://github.com/near/mpc/pull/3647/commits/396c97b1efbba03f73f9bd081e911f77fc5398b7)

crates/chain-gateway/src/chain_gateway.rs
```diff
@@ -232,3 +251,50 @@ struct StreamerSetup {
     indexer_config: near_indexer::IndexerConfig,
     near_config: NearConfig,
 }
+
+/// Races `sync` against `shutdown`. Returns `true` if `sync` completed first,
+/// `false` if `shutdown` resolved first.
```
kevindeforth: Please remove this comment, it's not useful, because it explains in 2 lines _what_ the method does in 4 lines, which can be easily identified by reading the code. (c.f. our [engineering_standards](https://github.com/near/mpc/blob/main/docs/engineering-standards.md#write-helpful-code-comments))

crates/chain-gateway/src/chain_gateway.rs
```diff
@@ -232,3 +251,50 @@ struct StreamerSetup {
...
+    shutdown: impl Future<Output = ()>,
+) -> bool {
+    tokio::select! {
+        _ = sync => true,
+        _ = shutdown => false,
+    }
+}
+
+#[cfg(test)]
+#[expect(non_snake_case)] // tests follow `<system_under_test>__should_<assertion>` convention
```
kevindeforth: please remove the comment, this is a common pattern in our codebase and I wouldn't want this comment to be in every test module.

I am also not sure if we should have the `expect` above the entire module, or rather, above the test methods, but that's a nit.

kevindeforth (dismissed):
Nice, thank you!
Just some nits on excessive comments.

crates/chain-gateway/src/chain_gateway.rs
gilcu3: ups this one I did remove in other places, but this one slept through... not sure why Claude likes adding those now :|

> I am also not sure if we should have the expect above the entire module, or rather, above the test methods, but that's a nit.

we are pushing for always using tests with the naming convention, so keeping it at the mod level is probably the right choice

pbeza (approved):

kevindeforth (approved):

PR #3656, opened 2026-06-24
test: add support for wycheproof vectors verification for signatures

### LINKED ISSUE #2241: verify signatures implementations
### Background

By using a package such as https://appsec.guide/docs/crypto/wycheproof/ we could verify that our signature implementations are correct. It includes many test vectors of correct and incorrect signatures. That would allow us to figure if we are missing corner cases that have appeared in previous attacks.

### Acceptance Criteria

The package is integrated and our threshold signatures pass all the test-vector tests.

### Context

Moved from https://github.com/near/threshold-signatures/issues/272

gilcu3:
Closes #2241

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -386,7 +386,7 @@ dependencies = [
  "alloy-rlp",
  "alloy-serde",
  "alloy-sol-types",
- "itertools 0.13.0",
+ "itertools 0.14.0",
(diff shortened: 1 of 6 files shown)


netrome (approved):
Thanks 🙏

anodar (approved):

PR #3659, opened 2026-06-24
fix: node crash on malformed participant set in task

### LINKED ISSUE #3651: Remote node crash via Computation message from a peer outside the channel participant list
### Background

A single authenticated-but-Byzantine MPC participant can crash any honest node with two messages, dropping the network below threshold.

`MessageCounters` is keyed solely on `channel.participants()` (`crates/node/src/protocol.rs:17-20`), which is copied verbatim from the attacker-controlled `MpcStartMessage.participants` (`network.rs:312`). The receive path routes `Computation` messages by `channel_id` only — `receive_one` returns `from: message.from` with no check that `message.from ∈ self.sender.participants` (`network.rs:711-715`). The resulting `counters.received(msg.from, …)` does an unchecked lookup:

```rust
// crates/node/src/protocol.rs:211
self.counters.get(&participant).unwrap()
```

For a `from` absent from the (attacker-chosen) participant set this panics. The shipped `[profile.reproducible]` build sets `panic = "abort"` (`Cargo.toml:328-331`) and `deployment/Dockerfile-node:16` ships that binary, so the panic aborts the whole process.

Exploit: send a victim a `Start` for a fresh `channel_id` with `participants = [victim]` (omitting the sender), then send a `Computation` on that channel. Repeatable and aimable at every node.

The same unchecked `.get(..).unwrap()` exists in `queue_send` (`protocol.rs:193`) and `sent` (`protocol.rs:201`).

### Acceptance Criteria

- `receive_one` rejects a `Computation`/`Success`/`Abort` whose `message.from` is not in `self.sender.participants`, returning a structured per-peer error instead of propagating the out-of-set `from`.
- `MessageCounters` lookups in `received`, `queue_send`, and `sent` no longer panic on an unknown participant.
- A regression test feeds `run_protocol` a `Computation` whose `from` is absent from `channel.participants()` and asserts a structured error / no panic.

gilcu3:
Closes #3651

### DIFF
--- a/crates/node/src/network.rs
+++ b/crates/node/src/network.rs
@@ -714,7 +714,6 @@ impl NetworkTaskChannel {
     /// computation message.
     async fn receive_one(&mut self) -> anyhow::Result<Option<TaskChannelComputationData>> {
         let message = self.receive_raw().await?;
-        // An out-of-set `from` would propagate into per-participant state keyed on the set.
         if !self.sender.participants.contains(&message.from) {
             let kind = message.message.kind.variant_name();
             tracing::warn!(
--- a/crates/node/src/network.rs
+++ b/crates/node/src/network.rs
@@ -500,6 +500,14 @@ pub struct TaskChannelComputationData {
     pub data: Vec<Vec<u8>>,
 }
 
+/// A peer outside the channel's participant set sent us a message.
+#[derive(Debug, thiserror::Error)]
+#[error("received {kind} from participant {from}, who is not in the channel participant set")]
+pub struct ParticipantNotInChannelError {
+    pub from: ParticipantId,
+    pub kind: &'static str,
+}
+
 impl Drop for NetworkTaskChannel {
     fn drop(&mut self) {
         if let Some(drop) = self.drop.take() {
@@ -707,6 +715,14 @@ impl NetworkTaskChannel {
     /// computation message.
     async fn receive_one(&mut self) -> anyhow::Result<Option<TaskChannelComputationData>> {
         let message = self.receive_raw().await?;
+        // An out-of-set `from` would propagate into per-participant state keyed on the set.
+        if !self.sender.participants.contains(&message.from) {
+            return Err(ParticipantNotInChannelError {
+                from: message.from,
+                kind: message.message.kind.variant_name(),
+            }
+            .into());
+        }
         match message.message.kind {
             MpcMessageKind::Computation(data) => {
                 return Ok(Some(TaskChannelComputationData {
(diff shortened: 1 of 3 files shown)


crates/node/src/network.rs
```diff
@@ -707,6 +714,24 @@ impl NetworkTaskChannel {
     /// computation message.
     async fn receive_one(&mut self) -> anyhow::Result<Option<TaskChannelComputationData>> {
         let message = self.receive_raw().await?;
+        // An out-of-set `from` would propagate into per-participant state keyed on the set.
```
kevindeforth: What does this mean?
> propagate into per-participant state keyed on the set

crates/node/src/protocol.rs
```diff
@@ -189,31 +189,32 @@ impl MessageCounters {
     }
 
     pub fn queue_send(&self, participant: ParticipantId, num_messages: usize) {
-        self.counters
-            .get(&participant)
-            .unwrap()
-            .in_flight
-            .fetch_add(num_messages, std::sync::atomic::Ordering::Relaxed);
-        self.report_progress();
+        if let Some(counters) = self.counters.get(&participant) {
```
kevindeforth: as discussed on slack, `let Some (...) else { // log warning}` is preferred.

crates/node/src/protocol.rs
```diff
@@ -189,31 +189,32 @@ impl MessageCounters {
...
     pub fn sent(&self, participant: ParticipantId, num_messages: usize) {
-        let counters = self.counters.get(&participant).unwrap();
-        counters
-            .sent
-            .fetch_add(num_messages, std::sync::atomic::Ordering::Relaxed);
-        counters
-            .in_flight
-            .fetch_sub(num_messages, std::sync::atomic::Ordering::Relaxed);
-        self.report_progress();
+        if let Some(counters) = self.counters.get(&participant) {
```
kevindeforth: idem

crates/node/src/protocol.rs
```diff
@@ -189,31 +189,32 @@ impl MessageCounters {
...
+        }
     }
 
     pub fn received(&self, participant: ParticipantId, num_messages: usize) {
-        self.counters
-            .get(&participant)
-            .unwrap()
-            .received
-            .fetch_add(num_messages, std::sync::atomic::Ordering::Relaxed);
+        if let Some(counters) = self.counters.get(&participant) {
```
kevindeforth, line 217: idem

kevindeforth (commented):
Thank you!
Just a few comments on being more explicit that we don't actually expect the participant to not be in the participant set.

crates/node/src/network.rs
gilcu3: it actually made some draconian sense, but not worth keeping the comment, so will remove. The problem probably appeared because of the instruction: "make comments succinct" :cry:

kevindeforth (approved):
Thank you!

anodar (approved):

PR #3667, opened 2026-06-24
feat: add metrics to detect starvation and resolve CI flakyness

### LINKED ISSUE #3666: CI: CPU contention for chain-gateway integration tests
### Background

Each chain-gateway integration test stands up two nearcore nodes (~25 OS threads per test). Running them concurrently oversubscribes CI runners and starves the event-subscriber pipeline.

Note that this is more of a feature than a bug, as we may be susceptible to similar starvation in prod in case the machine is not powerful. We should thus also add a few metrics that can help us identify starvation.

Hence, it makes sense to limit the number of concurrent tests for the chain-gateway integration tests.

https://nearone.slack.com/archives/C07UW93JVQ8/p1782292921833029

### User Story

As a dev, I like when integration tests give meaningful signals and are not flaky
As a dev, I like when I have useful metrics.

### Acceptance Criteria

Limit number of parallel tests in CI.
Add metrics that would help us recognize starvation in the chain-gateway

#3229

kevindeforth:
resolves #3666

works towards #3229

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -18,5 +18,15 @@ max-threads = 4
 filter = 'package(e2e-tests)'
 test-group = 'e2e'
 
+# Each chain-gateway integration test stands up two nearcore nodes (~25 OS
+# threads per test). Running them concurrently oversubscribes CI runners and
+# starves the event-subscriber pipeline.
+[test-groups.chain-gateway-integration]
+max-threads = 1
+
 [profile.ci-other]
 default-filter = 'not package(mpc-node) and not package(mpc-contract) and not package(e2e-tests) and not package(contract-history) and not test(=tee_authority::tests::test_fetch_collateral_from_pccs) and not (test(/^registry::integration_tests::/) and package(tee-launcher))'
+
+[[profile.ci-other.overrides]]
+filter = 'package(chain-gateway) and binary(test)'
+test-group = 'chain-gateway-integration'
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -22,7 +22,7 @@ test-group = 'e2e'
 # threads per test). Running them concurrently oversubscribes CI runners and
 # starves the event-subscriber pipeline.
 [test-groups.chain-gateway-integration]
-max-threads = 1
+max-threads = 2
 
 [profile.ci-other]
 default-filter = 'not package(mpc-node) and not package(mpc-contract) and not package(e2e-tests) and not package(contract-history) and not test(=tee_authority::tests::test_fetch_collateral_from_pccs) and not (test(/^registry::integration_tests::/) and package(tee-launcher))'
--- a/crates/chain-gateway/src/event_subscriber/streamer/block_processor.rs
+++ b/crates/chain-gateway/src/event_subscriber/streamer/block_processor.rs
@@ -12,6 +12,7 @@ use crate::{
             BlockContext, BlockUpdate, EventData, ExecutorFunctionCallSuccessWithPromiseData,
             MatchedEvent, ReceiverFunctionCallData,
         },
+        metrics::{MPC_BLOCK_UPDATES_DROPPED, MPC_BLOCKS_RECEIVED_FROM_INDEXER},
         recent_blocks_tracker::{AddBlockResult, RecentBlocksTracker},
         stats::IndexerStats,
     },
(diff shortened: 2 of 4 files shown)


.config/nextest.toml
```diff
@@ -18,5 +18,15 @@ max-threads = 4
 filter = 'package(e2e-tests)'
 test-group = 'e2e'
 
+# Each chain-gateway integration test stands up two nearcore nodes (~25 OS
+# threads per test). Running them concurrently oversubscribes CI runners and
+# starves the event-subscriber pipeline.
+[test-groups.chain-gateway-integration]
+max-threads = 2
```
gilcu3, line 25: oh, really 2 nearcore nodes already use 25 OS threads?

This is very strict IMO, for e2e tests we are fine with 4 threads, and those are probably heavier than the chain gateway tests

crates/chain-gateway/src/event_subscriber/streamer/block_processor.rs
```diff
@@ -52,6 +54,7 @@ pub(super) async fn listen_blocks(
                         block_context = ?context,
                         "dropping block update because channel is full",
                     );
+                    MPC_BLOCK_UPDATES_DROPPED.inc();
```
gilcu3, line 57: Could be worth adding these also in the current indexer code (inside mpc-node). This is so that while the chain-gateway is not wired, we can already see how the old implementation behaved in this regard

.config/nextest.toml
```diff
@@ -18,5 +18,15 @@ max-threads = 4
...
+
 [profile.ci-other]
 default-filter = 'not package(mpc-node) and not package(mpc-contract) and not package(e2e-tests) and not package(contract-history) and not test(=tee_authority::tests::test_fetch_collateral_from_pccs) and not (test(/^registry::integration_tests::/) and package(tee-launcher))'
+
+[[profile.ci-other.overrides]]
+filter = 'package(chain-gateway) and binary(test)'
+test-group = 'chain-gateway-integration'
```
gilcu3, line 32: I don't understand the second part `binary(test)`, why do we need that?

docs/design/node-operator-metrics.md
gilcu3, line 1: seems reasonable to me

gilcu3 (approved):
Thank you!

.config/nextest.toml
netrome: Hmm I'm not fully following - didn't we say the starvation was rather a feature than a bug? I think it is nice if we test our code under harsh conditions, and we had alternatives for how to handle the CI flakyness in the backpressure test that was failing.

docs/design/node-operator-metrics.md
```diff
@@ -0,0 +1,40 @@
...
+| [`mpc_finalized_blocks_indexed_total`](../../crates/chain-gateway/src/event_subscriber/metrics.rs) | blocks the tracker has promoted to `Final` | should grow steadily a few seconds behind the received/indexed counters; if it freezes while those keep growing, finality is stalling |
+| [`mpc_block_updates_dropped_total`](../../crates/chain-gateway/src/event_subscriber/metrics.rs) | block updates that won't be received by the node (containing signature requests, responses, etc.) | should be zero or flat. If it increases, the consumer is starved or the MPC node is not working correctly |
+| [`mpc_num_fail_on_timeout_indexed`](../../crates/node/src/metrics.rs) | number of calls to `fail_on_timeout` in the MPC contract. Counts the number of failed requests (aggregate over all request types). May contain false positives if `mpc_finalized_blocks_indexed_total` diverges from `mpc_blocks_indexed_total`, as it may count transactions on non-finalized forks. | should be near zero in healthy operation. Sustained non-zero rate means the node (or the cluster) is missing the response deadline or the blockchain has a lot of forks. |
```
netrome, line 22: Very nice. I feel we should have one of these lists for all our metrics and alerts, that we also give to node operators.

crates/chain-gateway/src/event_subscriber/streamer/block_processor.rs
netrome: Afaik, this dropping of block events is new to the chain gateway - which was part of our long discussions last Monday 😅

netrome (approved):
Really love the metrics and how you documented them 🚀 I'm skeptical about the need for limiting the threads for the tests. I think gaining visibility on effects of CPU starvation in CI is a feature rather than a bug.

.config/nextest.toml
kevindeforth: So it only applies to the test binary (i.e. the integration test)

PR #3670, opened 2026-06-25
ci: remove noisy warnings in cargo-shear and cargo-deny

### LINKED ISSUE #3636: Make fast checks fail on `cargo-shear` warnings
### Background

`cargo shear` (run in `check-all-fast` via `[tasks.cargo-shear]` in `Makefile.toml`) currently emits warnings without failing the check, so they slip through CI. There is an outstanding warning that needs resolution:

```
⚠ dependency `borsh` only used in features
  ╭─[crates/tee-verifier/Cargo.toml:35]
  borsh = { workspace = true }   ← not used in code, only via `abi` feature
```

The fast checks should fail when `cargo shear` reports any such warning, so they get resolved instead of accumulating.

### Acceptance Criteria

- `check-all-fast` fails when `cargo shear` reports a warning.
- The outstanding `borsh` warning in `crates/tee-verifier` is resolved.

gilcu3:
Closes #3636

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -29,7 +29,7 @@ ignore = [
 ]
 
 [bans]
-multiple-versions = "warn"
+multiple-versions = "allow"
 skip = [{ name = "windows-sys" }]
 skip-tree = [
   { name = "axum" },
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -58,7 +58,7 @@ args = ["sort", "--workspace", "--grouped", "--check", "--check-format"]
 
 [tasks.cargo-shear]
 command = "cargo"
-args = ["shear"]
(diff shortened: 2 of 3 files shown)


deny.toml
```diff
@@ -29,7 +29,7 @@ ignore = [
 ]
 
 [bans]
-multiple-versions = "warn"
+multiple-versions = "allow"
```
gilcu3, line 32: this is a broken window fix, to remove another warning present when doing `cargo make check-all-fast`, which should be clean now. We had decided a while ago that we didn't care about duplicate crates (very hard to fix with so many dependencies), so this change is just to remove the noise

netrome (approved):

anodar (approved):

PR #3673, opened 2026-06-25
fix: make sure node's do not fail parsing the participant url

### LINKED ISSUE #3668: Scheme-less node migration URL stalls the whole network; harden URL parsing and fix migration docs
### Background

A node was migrated with a destination URL lacking a scheme (`<host>:80` instead of `http://<host>:80`), following the example in `docs/node-migration-guide.md` (lines 269 and 283: `"url": "new-node.example.com:80"`). `conclude_node_migration` wrote that scheme-less URL into the participant set (`crates/contract/src/lib.rs:2612-2623`).

**Every node** then logged the same error and got stuck — not just the migrated one:

```
ERROR mpc: error converting contract state obtained from chain: no host found in participant url <host>:80
```

In `convert_participant_infos` (`crates/node/src/indexer/participants.rs:344-348`), `Url::parse` treats a scheme-less `<host>:80` as a scheme, so `host_str()` returns `None` and the function bails. The conversion fails fast on the first bad URL, so a single participant's malformed URL prevents **all** nodes from building the participant set: `contract_state_receiver` is never updated, the coordinator never reaches Running, and the initial `fetch_contract_state().await` loops forever ("No root task has started yet"). One node's bad migration stalled the whole network.

### Acceptance Criteria

- `docs/node-migration-guide.md` `start_node_migration` example (lines 269, 283) uses a schemed URL (e.g. `http://new-node.example.com:80`), matching the already-correct examples (for resharing) in `running-an-mpc-node-in-tdx-external-guide.md`.
- `convert_participant_infos` tolerates a single malformed/scheme-less participant URL — skip-and-warn or default a missing scheme — instead of failing the entire conversion, so one node's bad entry cannot block others from reaching Running.
- Consider contract-side validation rejecting `start_node_migration` / `submit_participant_info` with an unparseable URL.

gilcu3:
Closes #3668 

The PR unifies the type `NodeAddress` so that parsing is defined once. The original sin was to represent `NodeUrl` with a strong type, that broke the underlying invariant of the nodes always being able to deserialize the contract state.


crates/node/src/indexer/participants.rs
```diff
@@ -335,19 +335,36 @@ pub async fn monitor_contract_state(
     contract_state_receiver
 }
 
+/// Sentinel address used when a participant's stored address cannot be parsed. Any value
+/// from RFC 5737 TEST-NET-1 (`192.0.2.0/24`) is guaranteed unroutable, so the participant
+/// is kept in the set but treated as unreachable rather than failing the whole conversion
+/// and stalling every node.
+const UNREACHABLE_PARTICIPANT_ADDRESS: &str = "192.0.2.0";
```
gilcu3: we could also use `something.invalid` but this avoids the DNS part

gilcu3:
@claude review

Addressed your comments, except:

> docs/node-migration-guide.md:296 — The note \"must contain the 'http://' prefix, please do not forget adding it\" directly contradicts the new behavior (and the new start_node_migration__should_accept_scheme_less_url test). With operators reading this guide for live migrations, the stale instruction will keep producing http://-only URLs at best, and confuse anyone debugging why their scheme-less URL was accepted. Either delete the note or soften it to "scheme-less host:port also works; we recommend the http:// prefix for clarity".

this was done on purpose for compatibility reasons, as operators might look at the doc from `main`, while still running older node versions

crates/contract/src/lib.rs
```diff
@@ -4653,6 +4665,54 @@ mod tests {
...
+            };
+            running.parameters.participants().participants()[0]
+                .0
+                .clone()
+        };
+        let mut test_env = Environment::new(None, None, None);
+        test_env.set_signer(&account_id);
+        let mut destination_node_info = gen_random_destination_info();
+        destination_node_info.destination_node_info.url = url.to_string();
+        contract.start_node_migration(destination_node_info)
```
kevindeforth, line 4683: There's a bit of code duplication between this test and the one above, might be nice to put it in some helper methods.

crates/contract/tests/sandbox/common.rs
```diff
@@ -59,7 +59,7 @@ pub async fn create_account_given_id(
 
 pub fn gen_participant_info() -> ParticipantInfo {
     ParticipantInfo {
-        url: "127.0.0.1".into(),
+        url: "http://127.0.0.1:3000".into(),
```
kevindeforth, line 62: Hmm, is this change needed? I thought scheme-less urls are ok...

crates/primitives/src/node_address.rs
```diff
@@ -0,0 +1,269 @@
+use alloc::string::{String, ToString};
+use core::net::SocketAddr;
+use core::str::FromStr;
```
kevindeforth, line 3: What's the impact of this on contract size?

crates/node/src/indexer/participants.rs
```diff
@@ -335,19 +335,36 @@ pub async fn monitor_contract_state(
...
+                    target: "mpc",
+                    participant = %account_id,
+                    url = %info.url,
+                    %error,
+                    "participant address unparseable; treating node as unreachable"
+                );
+                (
+                    UNREACHABLE_PARTICIPANT_ADDRESS.to_string(),
+                    port_override.unwrap_or(0),
+                )
```
kevindeforth, line 366: Proposal: Lets move this to a separate PR and merge to resolve the issue we had on testnet. URL validation is  a somewhat orthogonal concern that may require additional discussion.

kevindeforth (commented):
I need a bit more time for this, can get back tomorrow. My recommendation:
- open a separate PR for graceful handling of invalid URLs. That one is needed and should be in the next release
- adding a helper for unit tests would reduce duplication with existing code

I was a bit confused about this PR, it seems to allows host:port scheme, yet, in some instances in the tests we changed the participants info to always contain the authority.

I am not sure about the validation logic in the contract. It draws in yet another dependency and it's unclear we need that much logic in the contract in the first place. I would suggest one of the following approaches:
- Have very strict requirements in the contract on what type of url we allow and make a light-weight custom parser. Use that in the contract without bringing in external dependencies.
- Don't make URL validation the contract's problem and ensure nodes can recover from mis-configured urls.

crates/contract/tests/sandbox/common.rs
gilcu3: only the port was needed IIRC, the scheme was a tiny plus

crates/primitives/src/node_address.rs
gilcu3: Apparently very little:

now -> 1517018 bytes

before (on main now) -> 1509901 bytes

kevindeforth:
> * Don't make URL validation the contract's problem and ensure nodes can recover from mis-configured urls.

Specifically, I think we can do what @netrome proposed and expose a `change_participant_url()` method in the contract. Then, we just need to make sure that the nodes re-establish a network when the url changes.

Note however: we should time-gate that endpoint. A participant shouldn't be able to trigger it more than 3 times in a 24h window, as with our current implementation, re-establishing network connections is not a very smooth operation...
Also, it should only be allowed in running state.

crates/node/src/indexer/participants.rs
gilcu3: I wanted to solve the problem at the root, which was in part allowing anything on the contract, but don't mind separating that. What's your worry about URL validation? We just need that the contract and the node agree on what is a valid node address. I think keeping that feature makes sense, unless it breaks something else ofc

gilcu3:
dropped in favor of https://github.com/near/mpc/pull/3678

PR #3675, opened 2026-06-25
docs: clarify how to set backup encryption key

### LINKED ISSUE #3674: Update TDX node & migration docs after recent migration hiccups
### Background

We've had a couple of instances where migrating nodes followed our node migration guide and didn't add `http://` prefixes to their URLs. We should clarify this in the docs.

Additionally there was also some confusion about how to configure the BACKUP_ENCRYPTION_KEY in the TDX CVM which should be updated in the docs.

### User Story

_No response_

### Acceptance Criteria

We've updated the docs to more accurately describe the expected URL format and how to configure the backup encryption key

_No response_

netrome:
closes #3674


docs/node-migration-guide.md
```diff
@@ -134,21 +134,27 @@ You should see your account and registered backup_cli public key listed, somethi
...
+**Note:** If your node has been running without an encryption key configured, the node automatically generates one and stores it in a file called `backup_encryption_key.hex` in your home directory. On a **non-TEE** node you can retrieve it with:
 
 ```bash
 export BACKUP_ENCRYPTION_KEY=$(cat $MPC_HOME_DIR/backup_encryption_key.hex)
 ```
 
 Copy this key and set it as the `BACKUP_ENCRYPTION_KEY` environment variable for the backup-cli when running `get-keyshares`.
 
+> **⚠️ TDX / CVM source nodes:** you **cannot** retrieve an auto-generated key this way — there is no filesystem access into the CVM. If a CVM node was deployed without `backup_encryption_key_hex` set in its `user-config.toml`, the key it generated internally is unrecoverable and the node **cannot be migrated off**. Always set `backup_encryption_key_hex` explicitly when deploying a CVM node so you know the key. (The shipped `user-config.toml` template makes this field a required, non-hex placeholder so a forgotten value fails loudly at startup rather than silently auto-generating.)
```
gilcu3: oh this seems really dangerous. Shouldn't we fail instead if no key is provided?

gilcu3 (dismissed):
Thank you!

docs/node-migration-guide.md
netrome: Yeah that would be better, iirc this was just introduced for backwards compatibility when we introduced the field for the 3.0.0 release since we could not add new env variables at that time.

gilcu3 (dismissed):

docs/node-migration-guide.md
```diff
@@ -134,21 +134,27 @@ You should see your account and registered backup_cli public key listed, somethi
...
+- **Non-TEE node:** set the `MPC_BACKUP_ENCRYPTION_KEY_HEX` environment variable.
+
+**Important:** The backup encryption key must be the same between the backup-cli and the node it is currently communicating with (e.g., the old node when running `get-keyshares`, and the new node when running `put-keyshares`). The node requires exactly 64 hex characters (32 bytes) with no surrounding whitespace, and **fails to start** if the configured value is malformed.
 
 
 ### Retrieve a key from an existing node.
 
-**Note:** If your node has been running without the `MPC_BACKUP_ENCRYPTION_KEY_HEX` environment variable set, the node automatically generates an encryption key and stores it in a file called `backup_encryption_key.hex` in your `$MPC_HOME_DIR` directory. You can retrieve it with:
+**Note:** If your node has been running without an encryption key configured, the node automatically generates one and stores it in a file called `backup_encryption_key.hex` in your home directory. On a **non-TEE** node you can retrieve it with:
```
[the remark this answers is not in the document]
netrome: Makes sense, applied.

docs/node-migration-guide.md
```diff
@@ -134,21 +134,27 @@ You should see your account and registered backup_cli public key listed, somethi
...
 
-**Important:** The `MPC_BACKUP_ENCRYPTION_KEY_HEX` must be the same between the backup-cli and the node it is currently communicating with (e.g., the old node when running `get-keyshares`, and the new node when running `put-keyshares`).
+**How you configure the key on a node depends on the deployment type** (see [Step 5](#step-5-prepare-the-new-node) for the exact placement):
+- **TDX / CVM node:** set `backup_encryption_key_hex` under `[mpc_node_config.secrets]` in `user-config.toml`. There is **no** `MPC_BACKUP_ENCRYPTION_KEY_HEX` environment-variable pathway into a CVM — the `.env` used by `deploy-launcher.sh` is not read for this key.
+- **Non-TEE node:** set the `MPC_BACKUP_ENCRYPTION_KEY_HEX` environment variable.
+
+**Important:** The backup encryption key must be the same between the backup-cli and the node it is currently communicating with (e.g., the old node when running `get-keyshares`, and the new node when running `put-keyshares`). The node requires exactly 64 hex characters (32 bytes) with no surrounding whitespace, and **fails to start** if the configured value is malformed.
```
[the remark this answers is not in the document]
netrome: Nah that was resolved in another PR.

docs/node-migration-guide.md
netrome: Edit, just got back to this and this seems to be an LLM-hallucinated false claim (or maybe it's from me, don't remember) - anyway, seems like I confused this with the persistent secrets. This key is just transport-only and can be reconfigured with just a CVM restart so nothing to worry about here.

netrome: Removed this while I rebased the PR.

gilcu3 (approved):

pbeza (approved):

PR #3678, opened 2026-06-25
fix: a single participant with wrong url should not break the node

### LINKED ISSUE #3668: Scheme-less node migration URL stalls the whole network; harden URL parsing and fix migration docs

gilcu3:
Closes #3668

Slimmer alternative to #3673

### DIFF
--- a/crates/node/src/indexer/participants.rs
+++ b/crates/node/src/indexer/participants.rs
@@ -335,19 +335,48 @@ pub async fn monitor_contract_state(
     contract_state_receiver
 }
 
+/// Sentinel address used when a participant's stored address cannot be parsed. Any value
+/// from RFC 5737 TEST-NET-1 (`192.0.2.0/24`) is guaranteed unroutable, so the participant
+/// is kept in the set but treated as unreachable rather than failing the whole conversion
+/// and stalling every node.
+const UNREACHABLE_PARTICIPANT_ADDRESS: &str = "192.0.2.1";
+
+fn parse_participant_address(
+    url: &str,
+    port_override: Option<u16>,
+) -> anyhow::Result<(String, u16)> {
+    let parsed =
+        Url::parse(url).with_context(|| format!("could not parse participant url {url}"))?;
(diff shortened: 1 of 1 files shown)


netrome (approved):
Looks like we discussed, as much as I'd prefer to explicitly handle this case in the node - that's a bigger change and the sentinel value will be quick and easy to get through. Thank you 🙏

kevindeforth (approved):
Excellent, thank you!

PR #3684, opened 2026-06-26
refactor: unify gas type to not expose near indexer internals 

### LINKED ISSUE #3694: Do not expose indexer types outside of chain-gateway
### Background

We currently have three different types for Gas in our codebase:
- our own `TeraGas` in a test-crate
- `near_gas::NearGas` in contract calls
- `near_indexer_primitives::types::Gas` in the indexer and chain gateway crate

`near_indexer_primitives::types::Gas` is a wrapper around `near_gas::NearGas`.

We should not expose the `near_indexer_primitives::types::Gas` type outside of the chain-gateway crate, but use the `near_gas::NearGas` instead (c.f. also [this comment](https://github.com/near/mpc/pull/2625#discussion_r3007892532)).

Additionally, in preparation for #3693, we can introduce a `FunctionCallArgs` in a separate crate, such that it can be re-used across our code-base.

### User Story

As a dev, I like to avoid redundant types.

### Acceptance Criteria

Create a light-weight crate for unifying contract calls to NEAR smart contracts.

_No response_

kevindeforth:
resolves #3694 
- unify Gas Type
- introduce FunctionCallArgs type in preparation for extending mpc-contract interface type

Next PRs:
- update contract-interface crate
- hook this up in mpc node (indexer), to prepare migration of chain-gateway crate

### DIFF
--- a/crates/chain-gateway/src/transaction_sender/signer.rs
+++ b/crates/chain-gateway/src/transaction_sender/signer.rs
@@ -4,7 +4,6 @@ use near_indexer::near_primitives::account::AccessKey;
 use near_indexer_primitives::near_primitives::transaction::{
     FunctionCallAction, SignedTransaction, Transaction, TransactionV0,
 };
-use near_indexer_primitives::types::Gas;
 use std::sync::Mutex;
 
 use crate::types::{FunctionCallArgs, LatestFinalBlockInfo};
@@ -49,7 +48,7 @@ impl TransactionSigner {
         let action = FunctionCallAction {
             method_name: args.method_name,
             args: args.args,
-            gas: Gas::from_gas(args.gas.as_gas()),
+            gas: crate::types::to_action_gas(args.gas),
             deposit: args.deposit,
         };
 
@@ -88,11 +87,9 @@ impl TransactionSigner {
 #[cfg(test)]
 mod tests {
     use ed25519_dalek::SigningKey;
-    use mpc_call_args::{FunctionCallArgs, NearGas};
+    use mpc_call_args::{FunctionCallArgs, NearGas, NearToken};
     use near_account_id::AccountId;
     use near_indexer::near_primitives::{account::AccessKey, transaction::Transaction};
-    use near_indexer_primitives::types::Gas;
-    use near_token::NearToken;
     use rand::{SeedableRng, rngs::StdRng};
 
     use crate::{transaction_sender::TransactionSigner, types::LatestFinalBlockInfo};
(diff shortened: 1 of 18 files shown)


kevindeforth:
@claude review

pbeza (dismissed):

crates/chain-gateway/src/transaction_sender/traits.rs
```diff
@@ -1,19 +1,17 @@
...
 pub trait SubmitFunctionCall {
     fn submit_function_call_tx(
         &self,
         signer: &TransactionSigner,
         receiver_id: AccountId,
-        method_name: String,
-        args: Vec<u8>,
-        gas: Gas,
+        call: FunctionCallArgs,
     ) -> impl Future<Output = Result<CryptoHash, ChainGatewayError>> + Send;
```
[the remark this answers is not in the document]
kevindeforth: Have to love the confidence of these things :D 
Green checks everywhere: _"ThIS WiLl fAiL tO CoMpILE!!!"_

crates/chain-gateway/src/transaction_sender/signer.rs
```diff
@@ -178,9 +180,9 @@ mod tests {
         assert_eq!(tx.actions.len(), 1);
         match &tx.actions[0] {
             near_indexer_primitives::near_primitives::transaction::Action::FunctionCall(action) => {
-                assert_eq!(action.method_name, method_name);
-                assert_eq!(action.args, args);
-                assert_eq!(action.gas, gas);
+                assert_eq!(action.method_name, args.method_name);
+                assert_eq!(action.args, args.args);
+                assert_eq!(action.gas, crate::types::to_action_gas(args.gas));
```
anodar: nit: If `to_action_gas` at some point messes up conversion, this won't catch it. Consider hardcoding instead Gas::from_gas(1000....) or maybe add unit tests for to action_gas.

crates/mpc-call-args/src/lib.rs
anodar, line 1: nit: Not sure this single small struct is worth introducing a new package. It's also named in a way that it likely won't contain much more than this in the future either. Consider moving it into existing package, e.g. under node-types (or package "types" and sub-packages inside, node, function-call-args, etc.. ).

anodar (dismissed):

crates/chain-gateway/src/transaction_sender/signer.rs
kevindeforth: good catch! addressed in https://github.com/near/mpc/pull/3684/commits/d9dc608b73ad158338dfac0dfd5ad80fe1cf3326

crates/mpc-call-args/src/lib.rs
kevindeforth: Yeah, I was really hesitant about that. At this point, I am a bit unsure where the best place for this struct is, because I am not sure where it will end up being used. If we use it contracts, we want it to be lean and the node types are probably the wrong place.
If we use it exclusively in the node and chain-gateway, then maybe moving it to a different crate makes more sense.
Opened an issue to revisit this decision: https://github.com/near/mpc/issues/3696

anodar (approved):
Thank you!

pbeza (approved):

PR #3691, opened 2026-06-29
feat: support flag for wiping nearcore data

### LINKED ISSUE #3690: feat(node): add a recovery-mode parameter that wipes the nearcore DB and re-syncs on startup
### Background

Operators occasionally hit a state where the embedded `neard` chain database is wedged and cannot recover on its own — e.g. block sync stuck on a missing block header (`DBNotFoundErr("BLOCK HEADER: ...")`), or a corrupted/inconsistent DB after a sync interruption.

On non-TEE nodes, the easiest recovery is manual: get into the node's data volume and delete the nearcore chain data so it re-syncs from scratch. Wiping the chain DB and re-syncing is a well-understood, common recovery mechanism.

On TEE nodes this isn't possible, because by design the operator has no access to the internals of the CVM the node runs in.

Proposed solution: add a capability for the operator to erase the nearcore database. This must be restricted to the operator only (e.g. via authentication or physical access).

### User Story

As a node operator, when my node's nearcore database is wedged, I want to restart the node with a "recovery mode" parameter that wipes the nearcore chain database and re-syncs from scratch, so I can recover the node without manually editing the data volume.

### Acceptance Criteria

- A parameter is passed through to the node (in the same manner as other node parameters) that starts the node in "recovery mode".
- In recovery mode, the node wipes the embedded `neard` chain database and re-syncs from scratch on startup.
- The capability is restricted to the node operator only (e.g. gated by authentication or physical access).
- The wipe is **scoped to the nearcore chain data only**. It MUST NOT destroy:
  - keyshares (GCP-backed or local encrypted storage),
  - the MPC working database (triples / presignatures / epoch data),
  - node identity / network keys (e.g. `node_key.json`).
- The mechanism is safe against accidental repeated wipes (e.g. it does not silently re-wipe on every subsequent restart).

### Out of scope / left open (design)

The mechanism (CLI flag vs. config field vs. launcher passthrough), one-shot semantics, and exactly which paths are removed are intentionally left to the design/implementation. This issue captures the requirement, not the design.

### Resources

- Motivating failure mode: neard block sync wedging on a missing block header; see related stuck-node discussions (#3683, #3295).
- On-disk layout to respect when scoping the wipe: nearcore chain data lives under `<home_dir>/data`; keyshares and the MPC rocksdb are stored separately (`crates/node/src/db.rs`, `crates/node/src/keyshare/`).

anodar:
Closes: https://github.com/near/mpc/issues/3690

Also tested manually against localnet: 
- Setting 0 is no op (doesn't even store 0 as token)
- Setting any non-zero value that differs from last stored token wipes data folder and data folder only
- Setting non-zero value that's same as last stored token is no op


crates/node/src/indexer/near_data_wipe.rs
```diff
@@ -0,0 +1,140 @@
...
+            ?hot_store_path,
+            requested_counter,
+            "wipe_near_data_counter bumped but node is archival; ignoring"
+        );
+        return Ok(());
+    }
+
+    write_last_counter(&counter_path, requested_counter)?;
+
+    match std::fs::remove_dir_all(hot_store_path) {
```
anodar: This order is safer because if we first wipe the dir and writing counter keeps failing for whatever reason, this will keep wiping on every restart.

barakeinav1: I hope you are erasing the data/data and not data ;-)

crates/node/src/tests.rs
```diff
@@ -209,6 +209,7 @@ impl IntegrationTestSetup {
                     finality: Finality::Final,
                     mpc_contract_id: "test".parse().unwrap(),
                     port_override: None,
+                    wipe_near_data_counter: 0,
```
barakeinav1: can you add explanation for this to our documentation/samples

crates/node/src/indexer/real.rs
```diff
@@ -101,6 +102,25 @@ pub fn spawn_real_indexer(
...
+            // is opened.
+            let hot_store_path = home_dir.join(
+                near_config
+                    .config
+                    .store
+                    .path
+                    .as_deref()
+                    .unwrap_or_else(|| Path::new("data")),
+            );
+            wipe_near_data_if_requested(
```
barakeinav1, line 116: Is this called in the right place (e.g before this folder is used?)
also, do we need to restart the node, or should it work like this?

crates/node/src/indexer/near_data_wipe.rs
anodar: When have I ever done that? oh wait...

Jokes aside, it's whatever is configured as near store database path if set, otherwise `mpc_home_folder/data`, so should only wipe the near data.

crates/node/src/indexer/real.rs
anodar: Yes and yes. Added comment about it.

crates/node/src/tests.rs
anodar: Added explanation

anodar:
@claude review

crates/node/src/tests.rs
barakeinav1: thanks for adding, but I actually meant adding this to https://github.com/near/mpc/blob/main/docs/running-an-mpc-node-in-tdx-external-guide.md#troubleshooting, or somewhere equivalent.

crates/node/src/indexer/near_data_wipe.rs
```diff
@@ -0,0 +1,209 @@
...
+        Ok(()) => tracing::info!(
+            ?hot_store_path,
+            requested_token,
+            "wiped nearcore data dir (wipe_near_data_token)"
+        ),
+        // Fresh node: nothing to wipe.
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => {}
+        Err(err) => return Err(err),
```
barakeinav1: By the time we reach here the token is already recorded, so a wipe that fails partway (e.g. partial `remove_dir_all`) is skipped on the next restart and the node opens a half-deleted store. Do we log an error anywhere on this path? If not, can we add a `tracing::error!` here so operators can see why the wipe didn't complete?

crates/node/src/indexer/near_data_wipe.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+}
+
+/// Writes and fsyncs the token (and its parent dir) so it survives a crash before
+/// the impending restart.
+fn write_last_token(token_path: &Path, token: u64) -> std::io::Result<()> {
+    if let Some(parent) = token_path.parent() {
+        std::fs::create_dir_all(parent)?;
+    }
+    let mut file = std::fs::OpenOptions::new()
```
barakeinav1: nit: this writes in place (truncate + write). `tee/allowed_image_hashes_watcher.rs` already has a write-temp-then-`fs::rename` pattern — reusing it would make this atomic. As is, a crash mid-write leaves a truncated file that parses as 0 and triggers an extra wipe next boot. Not blocking.

crates/node/src/indexer/near_data_wipe.rs
```diff
@@ -0,0 +1,209 @@
...
+    Ok(())
+}
+
+#[cfg(test)]
+#[expect(non_snake_case)]
+mod tests {
+    use super::*;
+    use rstest::rstest;
+
+    fn read_recorded(home: &Path) -> Option<u64> {
```
barakeinav1: nit: this helper unwraps/panics — per engineering-standards `§must_ prefix`, panicking test helpers should be named `must_…` (e.g. `must_read_recorded`).

crates/node/src/indexer/near_data_wipe.rs
```diff
@@ -0,0 +1,209 @@
...
+        return Ok(());
+    }
+
+    // Guard against a misconfigured store.path (absolute, `..`, or the home dir
+    // itself) turning the wipe into a `remove_dir_all` outside the node's data tree.
+    ensure_within_home(home_dir, hot_store_path)?;
+
+    write_last_token(&token_path, requested_token)?;
+
+    match std::fs::remove_dir_all(hot_store_path) {
```
barakeinav1: This deletes `home_dir/data` wholesale, which is safe only because everything the node can't recreate — `permanent_keys/`, `assets/`, `key`, the backup key, and the token file itself — lives as a *sibling* of `data/`, not inside it. Nothing enforces that today, so a future change that relocates MPC state under `data/` would turn this into a keyshare-destroyer. Worth a tripwire test asserting those paths are disjoint from the wipe target, so it fails CI rather than in TEE.

barakeinav1 (commented):
thanks,
looks good in general , I added a few comments/ suggestions.

crates/node/src/tests.rs
anodar: Added in [43ee17d](https://github.com/near/mpc/pull/3691/commits/43ee17d02c9e5ecd0f42d5bb362e6426f337a76e)

crates/node/src/indexer/near_data_wipe.rs
anodar: Added in [f5935de](https://github.com/near/mpc/pull/3691/commits/f5935de4198f6a48ff4628fbb0b27d001046dfe6)

crates/node/src/indexer/near_data_wipe.rs
anodar: Changed in [f5935de](https://github.com/near/mpc/pull/3691/commits/f5935de4198f6a48ff4628fbb0b27d001046dfe6) to same pattern as in allowed_images_watcher.rs

crates/node/src/indexer/near_data_wipe.rs
anodar: Addressed in [b9abfc6](https://github.com/near/mpc/pull/3691/commits/b9abfc6d28d59f647313fdb1fb4ea7d0047edbe5).

docs/running-an-mpc-node-in-tdx-external-guide.md
```diff
@@ -1858,9 +1858,41 @@ Reviewers — please add here more scenarios (with or without solutions)
...
+Semantics:
+
+- `0` (the default) never wipes.
+- A non-zero value that **differs** from the last applied value wipes exactly once. Any change
+  triggers it — increasing *or* decreasing — so you can always wipe again by picking a
+  different value.
+- Restarting with the **same** value, or setting it back to `0`, is a no-op (the node records
+  the last value it acted on, so it won't re-wipe on every restart).
+- On **archival** nodes the wipe is skipped (a warning is logged each startup) so the archive
+  isn't destroyed.
```
gilcu3, line 1905: what is exactly an archival node in our system?

crates/node/src/indexer/near_data_wipe.rs
anodar: Factored out home dir subfolders into single source of truth package and added test in [eb3fd37](https://github.com/near/mpc/pull/3691/commits/eb3fd37a70b57e1e39916474fff92e9915e686f4)

crates/node/src/keyshare/local.rs
```diff
@@ -14,7 +14,7 @@ pub struct LocalPermanentKeyStorageBackend {
 
 impl LocalPermanentKeyStorageBackend {
     pub async fn new(home_dir: PathBuf, key: [u8; 16]) -> anyhow::Result<Self> {
-        let permanent_key_dir = home_dir.join("permanent_keys");
+        let permanent_key_dir = crate::home_paths::permanent_keys_dir(&home_dir);
```
gilcu3: nit: a `use` statement would help here

crates/node/src/home_paths.rs
gilcu3, line 1: fine for me, we usually use constants but for such a simple file it is an overkill IMO

crates/node/src/indexer/near_data_wipe.rs
```diff
@@ -0,0 +1,222 @@
...
+    home_dir: &Path,
+    hot_store_path: &Path,
+    requested_token: u64,
+    is_archival: bool,
+) -> std::io::Result<()> {
+    // 0 is the "off" value: never wipe.
+    if requested_token == 0 {
+        return Ok(());
+    }
+    let token_path = wipe_token_file(home_dir);
```
gilcu3, line 32: this was a bit confusing, I thought we were wiping the token file :|

crates/node/src/indexer/near_data_wipe.rs
```diff
@@ -0,0 +1,222 @@
...
+                ?hot_store_path,
+                requested_token,
+                ?err,
+                "failed to wipe nearcore data dir, the store may be partially deleted and the \
+                 wipe will NOT be retried automatically (token already recorded) — fix the \
+                 cause and set wipe_near_data_token to a new value to retry"
+            );
+            return Err(err);
+        }
+    }
```
gilcu3: nit: we are still removing the folder in one step here, which might not be atomic. Given that this should be fast and executed early I don't really see a problem. The other pattern of rename and delete would also work, with the caveat that data from older renames might remain around, and need more code for deletion (but we could do it with a trash folder, that always gets cleaned on startup, just a thought BTW)

gilcu3 (dismissed):
LGTM

docs/running-an-mpc-node-in-tdx-external-guide.md
anodar: Not sure if it's legacy stuff but we have https://github.com/near/mpc/blob/344ed8ba678147f8c2f687e49e719e17096cf5fc/crates/node-config/src/start.rs#L195. I don't think it's actually used but covered it for completeness.

crates/node/src/indexer/near_data_wipe.rs
anodar: I can see that, but if I rename it to get_wipe_token_file then it's inconsisten with the rest of the function naming in that package, and pre-pending get to all of them seems redundant.

crates/node/src/home_paths.rs
anodar: home dir is coming from node config so these values can't really be constants. If you mean just the child dir names (assets, permanent_keys, key, etc..) even if those are constants constructing the full path would still be scattered and harder to validate that there's no intersection between data folder and the rest.

crates/node/src/keyshare/local.rs
anodar: Addressed in [34845f5](https://github.com/near/mpc/pull/3691/commits/34845f5795158721905d9c1689e2c203951490b0)

crates/node/src/indexer/near_data_wipe.rs
anodar: Addressed in [5853f5e](https://github.com/near/mpc/pull/3691/commits/5853f5e805f9e5e56b0703aa38f03a1765810dd5).

crates/node/src/home_paths.rs
```diff
@@ -0,0 +1,92 @@
...
+            assert!(
+                !path.starts_with(&wipe_target),
+                "{path:?} is inside the near-data wipe target {wipe_target:?}; \
+                 a wipe would destroy it"
+            );
+        }
+    }
+
+    /// The wipe renames the data dir into the trash dir, which requires the two to
+    /// be siblings — neither can be nested inside the other.
```
anodar, line 80: > neither can be nested inside the other

This is the condition that we test, not that they are siblings. Can fix wording in followup.

docs/running-an-mpc-node-in-tdx-external-guide.md
barakeinav1: I think archival node need maintain all history from genesis so they can't wipe/erase anything. 
don't think it is relevant for us

barakeinav1 (dismissed):
thanks

gilcu3 (dismissed):

gilcu3 (approved):

barakeinav1 (approved):

Issue #3693, opened 2026-06-29
Unify mpc contract calls


kevindeforth:
### Background

We have many different places that call methods in the MPC contract - devnet, e2e test, sandbox tests and in the node.
Each one of these places follows slightly different conventions, which means that a developer working in the repo needs to know all of these different patterns.

It has the additional downside of requiring pretty substantial refactors in case the API changes. We already have the contract-intreface crate as a single source of truth for the contract API, but we should add function extend it to prepare function call arguments.

This issue is split into multiple tasks:

- [x]  https://github.com/near/mpc/issues/3728
- [x] #3844
- [x] rename `mpc-call-args` to `near-contract-transport` and implement generic traits. Use those traits in the chain-gateway crates test contract and `MockChainState`. Add it to the `TeeContext` and birth the `client` module with the first MPC methods (`submit_participant_info`, `verify_tee`).
    - [x] #3865
    - [x] #3866 
    - [x] #3867 
    - [x] #3868 
- [x] integrate new traits with e2e tests
    - [x] https://github.com/near/mpc/issues/3898
    - [x] #3913 
    - [x] #3915 
    - [x] #3917 
    - [x] #3919 
    - [x] #3921 
    - [x] #3923 
    - [x] #4204
- [x] integrate new traits in sandbox
    - [x] https://github.com/near/mpc/issues/3901
    - [x] https://github.com/near/mpc/issues/4120
- [x] integrate new traits in devnet
    - [x]  #4188
- [ ] integrate new traits in node (split in 2 PRs)
- [ ] migrate all contract functions to the new call-interface
    - [x] #3949
- [x] https://github.com/near/mpc/issues/4253
### User Story

As a dev, I prefer a consistent code base over one that has many different patterns to achieve the same thing.

### Acceptance Criteria

Extend contract-interface to be a single source of truth for creating function calls to the contract. Hide the functionality behind a feature-flag to avoid unnecessary contract-bloat.

### Resources & Additional Notes

Note: this is an enabler for https://github.com/near/mpc/issues/2112

PR #3697, opened 2026-06-29
[DNR]: extend contract interface for call args

kevindeforth:
do not review yet
#3693


kevindeforth:
Closing this, found a better approach with https://github.com/near/mpc/issues/3693

Issue #3702, opened 2026-06-30
Persist MPC node logs across upgrades: write to disk + dedicated local HTTP endpoint


barakeinav1:
### Background

Follow-up to #3695. The chosen fix for #3695 (stabilize the launcher's Compose
project — fixed compose path + `-p` project name, drop the `docker rm -f`) makes
the launcher **reuse** the MPC container instead of removing and recreating it,
so Docker's logs survive a **restart** and stay viewable through the existing
dstack log interface (public_logs / dstack agent on `:8090`).

That approach does **not** cover **upgrades**: a new image digest forces Compose
to recreate the container → new container ID → Docker's json logs reset. The
dstack log interface only ever reflects the *current* container, so upgrade
history is still lost.

Today the node logs only to stdout (`crates/node/src/tracing.rs` builds a
`tracing_subscriber::fmt()` subscriber with no file writer), and those lines are
captured by Docker's container-ID-keyed json logs — which is why they don't
outlive container recreation.

### User Story

As an operator I want MPC node logs to survive **upgrades** (not just restarts)
and to retrieve them without SSH into the CVM.

### What / How

1. **Persist logs to disk.** Add a file-writing layer to the node's tracing
   setup *in addition* to the existing stdout layer (keeps the dstack live-log
   view unchanged). Write rotated files onto the persistent `mpc-data` volume
   (`home_dir = /data`, e.g. `/data/logs`). Use `tracing-appender` (already a
   transitive dep) with a **retention cap** (`max_log_files`) so the encrypted
   disk cannot fill up. Extend `LogConfig` (`crates/node-config/src/start.rs:66`)
   with optional `log_dir` + retention settings.

2. **Dedicated local-only retrieval endpoint.** Serve a `GET /debug/logs`
   handler (list + fetch rotated files) reusing the axum handlers in
   `crates/node/src/web.rs:328`. It must **not** be mounted on the public
   `web_ui` listener (port 8080, forwarded `Public 0.0.0.0`, already serving
   `/public_data`). Instead bind a **dedicated loopback listener** and forward
   it `Local 127.0.0.1:<port>`, mirroring `pprof_bind_address`
   (`127.0.0.1:34001`, `crates/node-config/src/lib.rs:129`) and the neard
   `Local 127.0.0.1:3030` forward. This keeps logs operator-private by binding.

### Acceptance Criteria

- [ ] Node writes rotated logs to the persistent `/data` volume alongside stdout
      (stdout / dstack live-log view unchanged).
[...]

PR #3705, opened 2026-06-30
fix(launcher): reuse MPC container across restarts to preserve logs

### LINKED ISSUE #3695: add support for keeping MPC container logs (in dstack) after a restart
### Background

After a **restart** the MPC container logs are cleared while running inside a
dstack CVM, so we lose valuable debugging information.

This happens because the launcher removes and recreates the MPC container on
every boot (`docker rm -f` + `docker compose up -d` in
`crates/tee-launcher/src/compose.rs`). Docker's json logs are keyed by container
ID, so recreating the container drops them.

We want logs to survive a restart and stay viewable through the existing dstack
log interface (public_logs / dstack agent).

### User Story

As an operator I'd like MPC node logs from before a restart to be retained so I
can debug.

### Approach

Stabilize the launcher's Compose project so the container is **reused** instead
of removed and recreated: pass an explicit `-p <project>` and drop the
unconditional `docker rm -f`. A leftover container from a legacy launcher
(different/no compose project) is cleaned up once to avoid a name collision.

### Acceptance Criteria

- [ ] MPC container logs survive a restart/crash and remain viewable via the
      existing dstack log interface.
- [ ] Documentation: how/where to retrieve the logs.

### Scope & Follow-up

This issue covers the **restart** case only. Logs are still lost across an
**upgrade** — a new image digest forces container recreation regardless.
Upgrade-durable log retention (persist logs to the `/data` volume + a dedicated
local HTTP retrieval endpoint) is tracked in #3702.

- Design doc: `docs/design/persisting-tee-node-logs.md`.
- Follow-up: #3702.

barakeinav1:
Closes #3695.

The launcher recreated the MPC container on every boot (`docker rm -f` + `docker compose up -d`), wiping its logs. Run Compose under a stable project (`-p mpc`) and drop the `rm -f`, so the container is reused across restarts and its logs persist — viewable via the existing dstack log interface. No name-conflict cleanup is needed: a launcher change re-keys the CVM disk, so a new launcher always boots on a fresh disk.

Scope: restart only. Logs are still reset on **upgrade** (new image digest → container recreated) — tracked in #3702.

Verified on a dstack TDX localnet: restart → logs preserved; upgrade → logs reset.

### DIFF
--- a/crates/tee-launcher/src/compose.rs
+++ b/crates/tee-launcher/src/compose.rs
@@ -59,13 +59,11 @@ pub fn launch_mpc_container(
     let compose_file = render_compose_file(platform, port_mappings, image_name, manifest_digest)?;
     let compose_path = compose_file.path().display().to_string();
 
-    // Remove any existing container from a previous run (by name, independent of compose file)
-    let _ = Command::new("docker")
-        .args(["rm", "-f", MPC_CONTAINER_NAME])
-        .output();
-
+    // Use a stable compose project (`-p`) so the container is reused across
+    // restarts rather than recreated, preserving its logs. Compose
+    // recreates it only when the rendered config (e.g. image digest) changes.
     let run_output = Command::new("docker")
-        .args(["compose", "-f", &compose_path, "up", "-d"])
+        .args(["compose", "-p", "mpc", "-f", &compose_path, "up", "-d"])
         .output()
         .map_err(|inner| LauncherError::DockerRunFailed {
             image_hash: manifest_digest.clone(),
--- a/crates/tee-launcher/src/compose.rs
+++ b/crates/tee-launcher/src/compose.rs
@@ -59,11 +59,12 @@ pub fn launch_mpc_container(
     let compose_file = render_compose_file(platform, port_mappings, image_name, manifest_digest)?;
     let compose_path = compose_file.path().display().to_string();
 
-    // Use a stable compose project (`-p`) so the container is reused across
-    // restarts rather than recreated, preserving its logs. Compose
-    // recreates it only when the rendered config (e.g. image digest) changes.
+    // The compose file pins a stable top-level project `name`, so the
+    // container is reused across restarts rather than recreated, preserving its
+    // logs. Compose recreates it only when the rendered config (e.g. image
+    // digest) changes.
     let run_output = Command::new("docker")
-        .args(["compose", "-p", "mpc", "-f", &compose_path, "up", "-d"])
+        .args(["compose", "-f", &compose_path, "up", "-d"])
         .output()
         .map_err(|inner| LauncherError::DockerRunFailed {
             image_hash: manifest_digest.clone(),
@@ -262,5 +263,20 @@ mod tests {
 
         // then
         assert!(!rendered.contains("environment:"));
+    }
+
+    #[test]
+    fn includes_stable_project_name() {
+        // given
+        let port_mappings = empty_port_mappings();
+        let digest = sample_digest();
+
+        // when
+        let rendered = render(Platform::Tee, &port_mappings, &digest);
+
+        // then — a stable top-level project name keeps Compose reusing the
+        // container across restarts instead of recreating it (which would wipe
+        // its logs)
+        assert!(rendered.contains("name: mpc-node"));
     }
 }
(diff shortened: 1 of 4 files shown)


crates/tee-launcher/src/compose.rs
```diff
@@ -59,13 +59,11 @@ pub fn launch_mpc_container(
...
-        .args(["rm", "-f", MPC_CONTAINER_NAME])
-        .output();
-
+    // Use a stable compose project (`-p`) so the container is reused across
+    // restarts rather than recreated, preserving its logs. Compose
+    // recreates it only when the rendered config (e.g. image digest) changes.
     let run_output = Command::new("docker")
-        .args(["compose", "-f", &compose_path, "up", "-d"])
+        .args(["compose", "-p", "mpc", "-f", &compose_path, "up", "-d"])
         .output()
```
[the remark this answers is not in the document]
barakeinav1: Thanks — this is the right concern on a general Docker host, but it doesn't apply to our deployment, so the one-time cleanup is intentionally left out:

- **TEE (production):** the CVM disk is encrypted with a key derived from the launcher's docker-compose measurement (see `docs/securing-mpc-with-tee-design-doc.md`, Launcher Pattern, step 3). A launcher change ⇒ different measurement ⇒ different key ⇒ the old disk can't be decrypted, so a new launcher always boots on a **fresh disk** with no pre-existing `mpc-node` container to collide with.
- The only situation this would guard is an **in-place launcher upgrade that reuses the disk** (e.g. NonTee on a persistent host), which isn't a supported flow for us.
- Node-image upgrades keep the **same** launcher (same compose project label), so there's no conflict on that path either.

barakeinav1:
Thanks for the review. On the non-blocking nits — leaving the code as-is, with rationale:

- **(a) Compose project constant:** `"mpc"` is kept inline on purpose. Without a migration-cleanup step it would be a single-use constant, and the literal sits alongside the other inline docker args (`"compose"`, `"-f"`, …); the comment above the call explains the `-p` rationale.
- **(b) `-p mpc` regression test:** fair that the fix hinges on the flag, but the only clean way to assert it is re-introducing an arg-builder indirection we deliberately inlined. For a single, commented flag in a one-line `.args([...])`, a dedicated test felt like over-engineering, and a dropped flag would be caught in review. Noted as the thing to revisit if this area grows.
- **(c) disk-rekey comment:** intentionally omitted — that assumption holds only in TEE (not NonTee), so stating it in-code would be misleading; the PR description carries the reasoning.

crates/tee-launcher/src/compose.rs
```diff
@@ -59,13 +59,11 @@ pub fn launch_mpc_container(
...
-    let _ = Command::new("docker")
-        .args(["rm", "-f", MPC_CONTAINER_NAME])
-        .output();
-
+    // Use a stable compose project (`-p`) so the container is reused across
+    // restarts rather than recreated, preserving its logs. Compose
+    // recreates it only when the rendered config (e.g. image digest) changes.
     let run_output = Command::new("docker")
-        .args(["compose", "-f", &compose_path, "up", "-d"])
+        .args(["compose", "-p", "mpc", "-f", &compose_path, "up", "-d"])
```
gilcu3: nit: maybe better to say "mpc-node"

anodar: nit: you can instead having project name in yaml itself: `name: mpc` 

https://docs.docker.com/reference/compose-file/version-and-name/#name-top-level-element

anodar (dismissed):

gilcu3 (dismissed):
Thank you!

crates/tee-launcher/src/compose.rs
barakeinav1: Good call — done in b23ac3e5: moved the project name into the compose file as a top-level `name: mpc-node` and dropped the `-p` flag, so it's declarative and now covered by the render tests. Thanks!

barakeinav1: Done in b23ac3e5 — using `mpc-node` as the project name (set via the top-level `name:` in the compose file). Thanks!

gilcu3 (approved):

anodar (approved):
Thank you!

PR #3706, opened 2026-06-30
feat: add nearcore node config endpoint

### LINKED ISSUE #3703: Expose effective nearcore config via a debug endpoint
### Background

The MPC node embeds a NEAR indexer (neard) whose effective config is loaded at startup via `load_near_config()`. This config (network, RPC, tracked shards, sync mode, etc.) can come from a downloaded file, local `config.json`, and CLI overrides, so the values actually in effect aren't easy to verify from the outside.

We already expose the MPC node's own `ConfigFile` at `/debug/node_config`, but nothing exposes the nearcore config. Surfacing it lets operators confirm exactly what the indexer is running. As far as we know there's no private info in this config; any sensitive fields (e.g. keys) must be redacted as is done for the existing endpoint.

### Acceptance Criteria

- A debug endpoint (e.g. `/debug/nearcore_config`) returns the effective nearcore config the indexer is running with.
- Any sensitive/secret fields are redacted from the response.

gilcu3:
Closes #3703

### DIFF
--- a/crates/node/src/config/start.rs
+++ b/crates/node/src/config/start.rs
@@ -171,25 +171,14 @@ fn patch_near_config(
 }
 
 /// Reads and parses the NEAR node `config.json` under `home_dir`, for serving
-/// via the `/debug/nearcore_config` endpoint. Returns `None` (and logs) when
-/// the file is missing or malformed, so a debug-endpoint failure never blocks
-/// startup.
-pub(crate) fn read_near_config_json(home_dir: &Path) -> Option<serde_json::Value> {
+/// via the `/debug/nearcore_config` endpoint.
+///
+/// Panics if the file is missing or malformed: the embedded indexer loads the
+/// same file at startup, so it is guaranteed present here.
+pub(crate) fn read_near_config_json(home_dir: &Path) -> serde_json::Value {
     let path = near_config_file(home_dir);
-    let raw = match std::fs::read_to_string(&path) {
-        Ok(raw) => raw,
-        Err(error) => {
-            tracing::warn!(%error, path = %path.display(), "could not read NEAR config.json for debug endpoint");
-            return None;
-        }
-    };
-    match serde_json::from_str(&raw) {
-        Ok(value) => Some(value),
-        Err(error) => {
-            tracing::warn!(%error, path = %path.display(), "NEAR config.json is not valid JSON");
-            None
-        }
-    }
+    let raw = std::fs::read_to_string(&path).expect("NEAR config.json must be readable");
+    serde_json::from_str(&raw).expect("NEAR config.json must be valid JSON")
 }
 
 /// Pure JSON-manipulation half of [`patch_near_config`], extracted so it can
@@ -465,34 +454,6 @@ mod tests {
         let config = read_near_config_json(home_dir.path());
 
         // Then
-        assert_eq!(
-            config,
-            Some(serde_json::json!({ "genesis_file": "genesis.json" }))
-        );
-    }
-
-    #[test]
-    fn read_near_config_json__should_return_none_when_file_missing() {
-        // Given
-        let home_dir = tempfile::tempdir().unwrap();
-
-        // When
-        let config = read_near_config_json(home_dir.path());
-
-        // Then
-        assert_eq!(config, None);
-    }
-
-    #[test]
-    fn read_near_config_json__should_return_none_when_file_malformed() {
-        // Given
-        let home_dir = tempfile::tempdir().unwrap();
-        std::fs::write(near_config_file(home_dir.path()), "not json").unwrap();
-
-        // When
-        let config = read_near_config_json(home_dir.path());
-
-        // Then
-        assert_eq!(config, None);
+        assert_eq!(config, serde_json::json!({ "genesis_file": "genesis.json" }));
     }
 }
--- a/crates/node/src/config/start.rs
+++ b/crates/node/src/config/start.rs
@@ -174,9 +174,6 @@ fn patch_near_config(
 /// via the `/debug/nearcore_config` endpoint. Returns `None` (and logs) when
 /// the file is missing or malformed, so a debug-endpoint failure never blocks
 /// startup.
-///
-/// Only `config.json` is exposed — never the node/validator key files it
-/// references, which hold secrets.
 pub(crate) fn read_near_config_json(home_dir: &Path) -> Option<serde_json::Value> {
     let path = near_config_file(home_dir);
     let raw = match std::fs::read_to_string(&path) {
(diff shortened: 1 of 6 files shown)


gilcu3:
@claude review

responses to the issues you found:

1. Currently the nearcore config.json does not contain any secret value, so there is nothing to hide from there.
2. I checked how nearcore loads the file, and it does not modify anything except adding the defaults, so for debugging purposes what we expose now seems good enough. Notice we modify the config.json only before the node runs, so the on-disk file is the actual value we care about.

crates/node/src/config/start.rs
```diff
@@ -169,6 +169,31 @@ fn patch_near_config(
...
 
+/// Reads and parses the NEAR node `config.json` under `home_dir`, for serving
+/// via the `/debug/nearcore_config` endpoint. Returns `None` (and logs) when
+/// the file is missing or malformed, so a debug-endpoint failure never blocks
+/// startup.
+///
+/// Only `config.json` is exposed — never the node/validator key files it
+/// references, which hold secrets.
+pub(crate) fn read_near_config_json(home_dir: &Path) -> Option<serde_json::Value> {
+    let path = home_dir.join("config.json");
```
anodar: nit: consider adding this in crates/node/src/home_paths.rs instead.

Same in tests, instead of manually constructing path.

crates/node/src/web.rs
```diff
@@ -68,6 +68,9 @@ struct WebServerState {
     migration_state_receiver: watch::Receiver<(u64, ContractMigrationInfo)>,
     static_web_data: StaticWebData,
     node_config: NodeConfigResponse,
+    /// Parsed nearcore `config.json` served by `/debug/nearcore_config`.
+    /// `None` if the file could not be read or parsed at startup.
+    nearcore_config: Option<serde_json::Value>,
```
anodar: nit: Consider using struct here instead so that we know what's actually being served. 


From a quick look at my localnet it should probably be https://github.com/near/nearcore/blob/master/nearcore/src/config.rs#L227

anodar (dismissed):
Thank you!

crates/node/src/config/start.rs
gilcu3: true, that didn't exist but now makes sense

crates/node/src/web.rs
gilcu3: hmm that struct looks big and we would need to keep it in sync with them. If they expose it I am still not sure if it is better to use it, given that they might do their own processing on load, which we don't need to worry about for debugging purposes. The invariant we want to keep is the endpoint gives the node nearcore config, and that is really the json config, as the nearcore parsing is not something the operator can modify.

crates/node/src/config/start.rs
```diff
@@ -169,6 +169,31 @@ fn patch_near_config(
     Ok(())
 }
 
+/// Reads and parses the NEAR node `config.json` under `home_dir`, for serving
+/// via the `/debug/nearcore_config` endpoint. Returns `None` (and logs) when
+/// the file is missing or malformed, so a debug-endpoint failure never blocks
+/// startup.
+///
+/// Only `config.json` is exposed — never the node/validator key files it
+/// references, which hold secrets.
```
kevindeforth: Please take a look at our engineering standards on helpful comments:  https://github.com/near/mpc/blob/main/docs/engineering-standards.md#write-helpful-code-comments

Not sure how the validator key is relevant for the reader here

crates/node/src/config/start.rs
```diff
@@ -169,6 +169,31 @@ fn patch_near_config(
...
+        }
+    };
+    match serde_json::from_str(&raw) {
+        Ok(value) => Some(value),
+        Err(error) => {
+            tracing::warn!(%error, path = %path.display(), "NEAR config.json is not valid JSON");
+            None
+        }
+    }
+}
```
kevindeforth, line 182: We have some existing code that does similar work:

https://github.com/near/mpc/blob/b9d11af37ffdc408ea10ed7a3792f56d01eba0a8/crates/node/src/indexer/real.rs#L100-L104

https://github.com/near/mpc/blob/b9d11af37ffdc408ea10ed7a3792f56d01eba0a8/crates/chain-gateway/src/chain_gateway.rs#L116-L120

 I would prefer if we only needed to parse this once across the lifetime of the node and passed it down as arguments through the rest of the stack.

crates/node/src/web.rs
```diff
@@ -68,6 +68,9 @@ struct WebServerState {
     migration_state_receiver: watch::Receiver<(u64, ContractMigrationInfo)>,
     static_web_data: StaticWebData,
     node_config: NodeConfigResponse,
+    /// Parsed nearcore `config.json` served by `/debug/nearcore_config`.
+    /// `None` if the file could not be read or parsed at startup.
```
kevindeforth: please take a look at https://github.com/near/mpc/blob/main/docs/engineering-standards.md#write-helpful-code-comments

kevindeforth (commented):
Thank you very much, this is a nice additon!
Two blockers:
- (mis-)use of code comments - please take a look at our engineering standards guide: https://github.com/near/mpc/blob/main/docs/engineering-standards.md#write-helpful-code-comments
- I don't think we need all that boilerplate. There are existing helper functions that allow us to parse the nearcore and indexer configs.

crates/node/src/config/start.rs
gilcu3: > Not sure how the validator key is relevant for the reader here

it is, and that's why I left that reference there. So unless the node/validator keys cannot be there, I think this is a helpful comment

crates/node/src/web.rs
anodar: I didn't mean to duplicate struct here, but to use it directly from nearcore, as it's already public. Fine as is though.

crates/node/src/web.rs
gilcu3: I think the comment is helpful, if I read this struct and I want to know what is that without having to hunt down the code. I can see this is a bit subjective, but the comment is minimal, so I think it doesn't hurt

crates/node/src/config/start.rs
gilcu3: removed, apparently it is not possible to put the secrets there

crates/node/src/web.rs
gilcu3: anyway, removed

crates/node/src/config/start.rs
gilcu3: we discussed this on slack. Although this also makes sense, we agreed that exposing the actual json value without postprocessing by nearcore is also helpful while debugging with operators, so left as is.

anodar (approved):

kevindeforth (approved):
Thank you!

PR #3707, opened 2026-06-30
docs: clarify that OT setup cannot be reused in our implementation

### LINKED ISSUE #3652: Fix stale triple-generation docs claiming a reusable one-time base-OT setup
### Background

The module docs for triple generation describe a one-time setup that is run once and then reused across many triples:

> This protocol requires a setup protocol to be done once beforehand. After this setup protocol has been run, an arbitrary number of triples can be generated.

(`crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/triples.rs:25-27`)

The implementation does the opposite: each `multiplication_sender`/`multiplication_receiver` generates a fresh random package and runs a **fresh** `batch_random_ot` per call — the inline comments even say "First, run a fresh batch random OT ourselves" (`triples/multiplication.rs:63,122`). Nothing is persisted or reused across runs.

This freshness appears to matter for security, not just performance. The base OT establishes the per-run secret `delta` that the OT extension builds on, and the consistency-check abort in `random_ot_extension.rs` looks like a potential selective-failure surface: a malicious peer might be able to make the abort condition depend on a bit of `delta`, turning each run into a partial oracle on it. With a fresh `delta` per run any such leak is bounded to an ephemeral value that is then discarded; reusing `delta` across runs could plausibly let an attacker accumulate information about it over many runs, which — if the selective-failure path is real — could degrade OT-extension secrecy and, downstream, the triple shares. We have not confirmed an exploit; the point is that freshness seems to be what currently forecloses this class.

The stale docs are therefore a trap regardless of whether the attack is realizable: a maintainer who "optimizes" by hoisting base OT into a reusable setup, as the docs suggest, would change a security-relevant property based on documentation that is simply wrong about what the code does.

### Acceptance Criteria

- The triple-generation module docs (and `generate_triple`'s rustdoc) state that base OT is run fresh per multiplication and must never be persisted or reused across runs/aborts.
- The "base OT is fresh per run, never persist/reuse" invariant is recorded at the base-OT call sites in `triples/multiplication.rs`, so the property is documented where the optimization would be attempted.
- The removed "setup once, reuse" wording no longer appears anywhere in the crate's docs.

gilcu3:
Closes #3652

### DIFF
--- a/crates/threshold-signatures/docs/ecdsa/ot_based_ecdsa/triples.md
+++ b/crates/threshold-signatures/docs/ecdsa/ot_based_ecdsa/triples.md
@@ -66,9 +66,17 @@ as two points, providing a key derivation function $\mathbb{G} \to \mathbb{F}_2^
 
 # Setup Phase
 
+> **Implementation note.** This crate does **not** run the setup once and reuse
+> it. The base OT (`Batch-Random-OT`) is run fresh inside every `Multiplication`
+> call and is never persisted or reused across runs or aborts. A fresh $\Delta$
+> per run bounds any information a malicious peer might extract through the
+> `Random-OT-Extension` consistency check to that single ephemeral run.
+> Changing the base OT into a reusable one-time setup, as the abstract spec
+> below suggests, would be a security regression.
+
 The goal of the setup phase is for each unordered pair of parties $P_a$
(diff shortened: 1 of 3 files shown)


SimonRastikian (approved):
Thanks

barakeinav1 (approved):

PR #3713, opened 2026-06-30
fix: indexer wedging on startup

### LINKED ISSUE #3623: MPC indexer wedges at genesis on a freshly state-synced node

gilcu3:
Closes #3623 (II)

### DIFF
--- a/crates/chain-gateway/tests/event_subscriber_integration.rs
+++ b/crates/chain-gateway/tests/event_subscriber_integration.rs
@@ -21,9 +21,8 @@ use chain_gateway_test_contract::{
 };
 use rstest::rstest;
 
-// Generous: the gateway now waits for the node to catch up to its peers before
-// streaming (see `wait_for_full_sync`), which delays the first block update on a
-// loaded CI box.
+// Generous: the gateway waits for the node to catch up before streaming, which
+// delays the first block update on a loaded CI box.
 const EVENT_TIMEOUT: Duration = Duration::from_secs(30);
 
 async fn must_recv_block_update(
--- a/crates/chain-gateway/tests/event_subscriber_integration.rs
+++ b/crates/chain-gateway/tests/event_subscriber_integration.rs
@@ -21,9 +21,7 @@ use chain_gateway_test_contract::{
 };
 use rstest::rstest;
 
-// Generous: the gateway waits for the node to catch up before streaming, which
-// delays the first block update on a loaded CI box.
-const EVENT_TIMEOUT: Duration = Duration::from_secs(30);
+const EVENT_TIMEOUT: Duration = Duration::from_secs(10);
 
 async fn must_recv_block_update(
     receiver: &mut tokio::sync::mpsc::Receiver<BlockUpdate>,
--- a/crates/chain-gateway/tests/event_subscriber_integration.rs
+++ b/crates/chain-gateway/tests/event_subscriber_integration.rs
@@ -21,7 +21,10 @@ use chain_gateway_test_contract::{
 };
 use rstest::rstest;
 
-const EVENT_TIMEOUT: Duration = Duration::from_secs(10);
+// Generous: the gateway now waits for the node to catch up to its peers before
+// streaming (see `wait_for_full_sync`), which delays the first block update on a
+// loaded CI box.
+const EVENT_TIMEOUT: Duration = Duration::from_secs(30);
 
 async fn must_recv_block_update(
     receiver: &mut tokio::sync::mpsc::Receiver<BlockUpdate>,
--- a/crates/node/src/indexer.rs
+++ b/crates/node/src/indexer.rs
@@ -425,40 +425,58 @@ const INTERVAL: Duration = Duration::from_millis(500);
 const REQUIRED_STABLE_POLLS: u32 = 4;
 
 impl IndexerClient {
+    /// Polls sync status, yielding `(syncing, head_height)`, or `None` on a
+    /// failed request.
+    async fn sync_info(&self) -> Option<(bool, BlockHeight)> {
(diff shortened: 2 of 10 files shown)


crates/chain-gateway/tests/event_subscriber_integration.rs
```diff
@@ -21,7 +21,10 @@ use chain_gateway_test_contract::{
 };
 use rstest::rstest;
 
-const EVENT_TIMEOUT: Duration = Duration::from_secs(10);
+// Generous: the gateway now waits for the node to catch up to its peers before
+// streaming (see `wait_for_full_sync`), which delays the first block update on a
+// loaded CI box.
+const EVENT_TIMEOUT: Duration = Duration::from_secs(30);
```
gilcu3: the changes I am adding forced this constant to be bumped as tests would pass locally but not in CI

gilcu3:
@claude review

this is a new solution, so please disregard your previous review

barakeinav1 (dismissed):
Approving 

Testing this on a fresh state-synced testnet CVM in parallel, since this bug's failure mode is in neard's runtime behavior rather than the predicate itself — the unit tests can't cover that. Will report back.

crates/node/src/indexer.rs
```diff
@@ -421,28 +421,84 @@ struct IndexerClient {
...
-            let status_response = self
-                .client
-                .send_async(
-                    near_o11y::span_wrapped_msg::SpanWrappedMessageExt::span_wrap(status_request),
-                )
-                .await;
+    /// Returns once the node is confirmed following the chain tip. Used only at
+    /// startup: neard can report `syncing == false` while still behind (fresh
+    /// genesis, after downtime, or while disconnected from all peers), so we
+    /// confirm via [`SyncProgress`] before binding the streamer's cursor.
```
kevindeforth: please take a look at our engineering standards document regarding code comment quality.

_Used only at startup_ is a statement that can be invalidated by changing other parts of the code.

crates/node/src/indexer.rs
```diff
@@ -421,28 +421,84 @@ struct IndexerClient {
...
+        }
+    }
+}
 
-            let Ok(Ok(status)) = status_response else {
-                continue;
-            };
+/// Reports the node as caught up only after a run of non-syncing polls over
+/// which the head actually advances, filtering out the transient
+/// `syncing == false` a behind node reports before its head starts moving.
```
kevindeforth: This comment is a repetition of things that have already been stated in other places and it describes behavior of a component from a different repository.

Most of these comments can simply be replaced with a link to the relevant nearcore issue (c.f. our engineering standards document).

kevindeforth (commented):
Looks good, but I doubt the value of the LLM generated code comments.

crates/node/src/indexer.rs
gilcu3: trimmed in [103c58b](https://github.com/near/mpc/pull/3713/commits/103c58b333159d784a2df3bfc25cd33254398c90)

crates/node/src/indexer.rs
gilcu3: handled in [103c58b](https://github.com/near/mpc/pull/3713/commits/103c58b333159d784a2df3bfc25cd33254398c90)

kevindeforth (approved):
Thank you!

Note for future reference: this is a patch only that should be reverted once https://github.com/near/nearcore/issues/16004 is fixed.

barakeinav1 (approved):

barakeinav1:
### Tested on testnet TEE — fix confirmed ✅

Built this branch (`103c58b3`) reproducibly and deployed **5 fresh idxtest CVMs**, each state-syncing testnet from scratch — 5 independent rolls of the startup race.

**Result: all 5 caught up to head, 0 wedges.** The indexer streamer started at the **live tip**, not at genesis:

| node | head | indexer height | streamer started at |
|---|---|---|---|
| 15 | 257262010 | 257262010 | `#257260914` |
| 16 | 257262010 | 257262010 | `#257260853` |
| 17 | 257262011 | 257262011 | `#257260879` |
| 18 | 257262011 | 257262011 | `#257261024` |
| 19 | 257262011 | 257262011 | `#257260470` |

Throughout header→state→block sync the indexer correctly **deferred** (no `starting streamer`, no `mpc_indexer_latest_block_height`), then started at head only once fully synced. For contrast, on the old code a wedged node logged `starting streamer` ~1 s after `Indexer waiting` and pinned `mpc_indexer_latest_block_height` at genesis `42376888` while head advanced to ~257M.

Note: these are non-participant idxtest nodes, so this exercises the indexer/sync path only (not signing/attestation) — which is the scope of the fix.

Issue #2766, opened 2026-04-09
shrink contract code


kevindeforth:
### Background

The contract binary is too large. We are starting to hit the 1.5 Mb RPC payload limit during testing.
There are work-arounds, because the nearcoare contract limit is 4Mb, but lets first try to shed some code and make the binary smaller

### User Story

As a dev, I like to be able to make changes to the contract.

### Acceptance Criteria

Shrink the contract to below 1.3 Mb or less.

netrome:
I wonder if reducing code duplication like in this issue https://github.com/near/mpc/issues/1172 would reduce the binary size or if the compiler is already smart enough to optimize this away.

kevindeforth:
this has been achieved with the verifier contract breakout https://github.com/near/mpc/pull/3714

PR #3714, opened 2026-06-30
feat(contract): async TEE attestation verification, drop `dcap-qvl`

### LINKED ISSUE #3642: Wire async TEE attestation verification through the verifier contract and drop `dcap-qvl` from mpc-contract
## Background

Final slice of #3540, following #3590 (split DCAP from post-DCAP) and #3626 (verifier-account voting). The voted-in `tee_verifier_account_id` is currently stored but unused: `mpc-contract` still verifies attestations synchronously in-WASM via `verify_locally`, so it still links `dcap-qvl` (~310 KB / ~21% of the WASM) and stays under NEP-509 transaction-size pressure.

This slice makes the contract actually offload DCAP verification to the standalone `tee-verifier` contract over a cross-contract call, then drops `dcap-qvl` from the contract WASM. It follows the design doc's submission-flow and handling-failures sections.

## Goal

Offload Dstack DCAP attestation verification from `mpc-contract` to the standalone `tee-verifier` contract over a cross-contract call, then drop the `dcap-qvl` dependency from the contract WASM.

pbeza:
Closes #3642.

Splits the first half of #3664 out for focused review: the contract-side async TEE attestation flow and the removal of `dcap-qvl` from `mpc-contract`. The stub verifier, sandbox tests, and design doc land in a stacked PR on top of this branch.

Today the MPC contract verifies a node's Intel TDX attestation synchronously, inside its own WASM, by linking the `dcap-qvl` library. This PR moves that crypto verification out to a separate `tee-verifier` contract and makes `submit_participant_info` asynchronous (yield/resume via a cross-contract call, with refund/revert on failure or the ~200-block timeout). The MPC contract stops linking `dcap-qvl` entirely.

## WASM size impact

Dropping `dcap-qvl` shrinks the contract WASM by ~310 KB (-20.5%). Both binaries built identically (`cargo near build non-reproducible-wasm --features abi --profile=release-contract --locked`, same `wasm-opt -O` post-step):

| | Raw bytes | Gzipped |
| --- | ---: | ---: |
| `main` | 1,513,344 | 576,992 |
| this PR | 1,202,840 | 450,631 |
| delta | -310,504 (-20.5%) | -126,361 (-21.9%) |

`cargo tree -p mpc-contract -i dcap-qvl` returns no match on this branch (it resolves on `main`), confirming `dcap-qvl` is no longer in the contract's build graph.


crates/contract/src/tee/pending_attestation.rs
```diff
@@ -0,0 +1,53 @@
...
+use near_mpc_contract_interface::types::Ed25519PublicKey;
+use near_sdk::{CryptoHash, NearToken, near};
+
+/// One in-flight verification per submitter account.
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+pub struct PendingAttestation {
+    /// The submitted payload the post-DCAP checks consume once the verifier
+    /// returns its report.
+    pub dstack: DstackAttestation,
+    /// Checked against the quote's report-data during the post-DCAP checks.
```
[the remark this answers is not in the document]
pbeza: Good catch. Confirmed: after the `verify_quote` call, only `tcb_info` is read post-callback. Created a follow-up: near/mpc#3720.

crates/contract/src/lib.rs
```diff
@@ -792,62 +810,140 @@ impl MpcContract {
...
+        insertion: &ParticipantInsertion,
+        caller_is_participant: bool,
+        attached: NearToken,
+    ) -> Result<(), Error> {
+        let is_new_attestation =
+            matches!(insertion, ParticipantInsertion::NewlyInsertedParticipant);
+        // A participant refreshing an existing attestation is not charged.
+        if caller_is_participant && !is_new_attestation {
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
pbeza: Fixed in 16ce2263b — the participant-refresh early return now calls `refund_attestation_deposit(account_id, attached)` before returning, so an attached deposit is refunded in full rather than kept, matching the doc.

crates/contract/src/lib.rs
```diff
@@ -792,62 +811,137 @@ impl MpcContract {
...
+            Attestation::Dstack(dstack) => {
+                self.submit_dstack_attestation(node_id, dstack, caller_is_participant)
+            }
+        }
+    }
+
+    /// Async [`Attestation::Dstack`] submission: registers a yield, fires the
+    /// cross-contract verify-quote call, and resumes via
+    /// [`Self::resolve_verification`].
+    fn submit_dstack_attestation(
```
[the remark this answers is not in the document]
pbeza: Tests will follow in the [stacked PR](https://github.com/near/mpc/pull/3715).

Context – I split the original PR into two since it was too large: this one and the one stacked on top of it with tests: https://github.com/near/mpc/pull/3715.

That was the only sensible way I could find to split up the huge PR.

pbeza:
@claude review

crates/contract/src/lib.rs
```diff
@@ -2263,14 +2367,156 @@ impl MpcContract {
...
+            }
+        };
+
+        // Take the pending entry now. A late verifier response can arrive after the
+        // ~200-block yield timeout already fired and `on_attestation_verified` removed
+        // the entry and resolved the yield; there is then nothing to do.
+        let Some(pending) = self.pending_attestations.remove(&account_id) else {
+            log!(
+                "resolve_verification: no pending attestation for {account_id} (late response or already cleaned up); ignoring"
+            );
```
barakeinav1: nit(doc): can you please align the design doc (#3715) here — its `resolve_verification` does `.remove(..).expect(..)`, which would panic on exactly this late-response path (verifier answers after the ~200-block timeout already cleaned up). The graceful return here is the correct behavior; the doc should match.

crates/contract/src/lib.rs
```diff
@@ -2263,14 +2367,156 @@ impl MpcContract {
...
+
+        // Fail the submitter's transaction from a separate receipt so the
+        // cleanup above commits (a panic here would roll it back)
+        let promise = Promise::new(env::current_account_id()).function_call(
+            method_names::FAIL_ATTESTATION_SUBMISSION.to_string(),
+            borsh::to_vec(&reason).expect("borsh serialization of reason must succeed"),
+            NearToken::from_yoctonear(0),
+            Gas::from_tgas(self.config.fail_attestation_submission_tera_gas),
+        );
+        PromiseOrValue::Promise(promise.as_return())
```
barakeinav1: nit(doc): can you please align the design doc (#3715) — it has `on_attestation_verified -> Result<(), String>` returning `Err(reason)`, but the impl fails the caller's tx via a separate `fail_attestation_submission` receipt (with its own gas knob). Worth reflecting the actual mechanism so a future reader doesn't implement the `return Err` version.

crates/contract/src/lib.rs
```diff
@@ -2263,14 +2367,156 @@ impl MpcContract {
...
+
+        if matches!(attestation_result, AttestationResult::Err(_)) {
+            refund_attestation_deposit(&account_id, pending.attached_deposit);
+        }
+        // MUST be the last host call: anything after could panic and roll back
+        // the state mutations above
+        env::promise_yield_resume(
+            &pending.data_id,
+            serde_json::to_vec(&attestation_result)
+                .expect("json serialization of AttestationResult must succeed"),
```
barakeinav1: nit(doc): can you please align the design doc (#3715) — it resumes with `borsh::to_vec(&FinalOutcome)`, but the impl uses JSON `AttestationResult`. Reconcile the name + serializer.

crates/contract/src/lib.rs
```diff
@@ -792,62 +811,137 @@ impl MpcContract {
...
+        caller_is_participant: bool,
+    ) -> Result<(), Error> {
+        let account_id = node_id.account_id.clone();
+
+        if self.pending_attestations.contains_key(&account_id) {
+            return Err(TeeError::VerificationAlreadyPending.into());
+        }
+
+        let Some(verifier_account_id) = self.tee_verifier_account_id.clone() else {
+            return Err(TeeError::VerifierNotConfigured.into());
```
barakeinav1, line 851: nit(doc): can you please align the design doc (#3715) — a few names/fields drifted: `finish_verify` → `verify_post_dcap_and_store`, three → four gas knobs, `PendingAttestation.caller_is_participant` (absent from the doc's field list), and the guards here return `VerificationAlreadyPending`/`VerifierNotConfigured` rather than panicking as the doc describes.

barakeinav1:
Node-side follow-up (not this PR): the node polls `get_participant_attestation` `TRANSACTION_TIMEOUT` (10s) after submitting. The async round-trip (submit → `verify_quote` → `resolve_verification` stores) is ~3-4 blocks, so 10s usually covers the success/fast-fail path with no re-submit. But if the verifier round-trip ever creeps past 10s, the node re-submits into the in-flight entry and hits `VerificationAlreadyPending`. Once the verifier latency is benchmarked, we might consider bumping `TRANSACTION_TIMEOUT` a bit so the fast success/failure path never re-submits prematurely.

crates/contract/src/lib.rs
```diff
@@ -792,62 +811,137 @@ impl MpcContract {
...
+                        caller_is_participant,
+                        data_id,
+                    },
+                );
+            },
         );
 
-        let attestation_storage_must_be_paid_by_caller =
-            is_new_attestation || caller_is_not_participant;
+        Ok(())
```
barakeinav1: Using Claude for this one, so take it with a grain of salt — but it flagged something worth verifying.

**Issue:** `submit_participant_info` yields here (via `enqueue_yield_request` → `env::promise_return`) yet returns `Result<(), Error>`. On the `Ok(())` path, near-sdk's `#[handle_result]` codegen emits `env::value_return(b"null")` *after* the body's `promise_return`, and last-write-wins — so the yield gets clobbered and the caller's tx resolves immediately as success. The reject/timeout outcome computed later in `on_attestation_verified` then has nowhere to go. Note the other three yielding methods (`sign`, `request_app_private_key`, `verify_foreign_transaction`) all return `()` precisely to avoid this — a `()` method emits no `value_return`, so `promise_return` survives.

Net effect: a Dstack submission would always look like immediate success to the caller, regardless of the eventual verifier result.

**Suggested fix (mirror `sign`):** make `submit_participant_info` return `()` instead of `Result<(), Error>`, and `env::panic_str` on the early errors (`VerifierNotConfigured`, `VerificationAlreadyPending`, and the Mock-path `InsufficientDeposit`). A panic already fails the tx with a message — exactly what `sign` does ("It's important we fail here … This allows users to get the error message"). The `on_attestation_verified` → `fail_attestation_submission` half already mirrors `sign`'s `fail_on_timeout`, so no callback changes are needed.

**On the removed tests:** a unit test wouldn't have caught this — unit tests call the method directly and bypass the generated wasm entrypoint where `value_return`/`promise_return` actually live. This needs a sandbox test that submits a Dstack attestation and asserts the caller's tx parks and then *fails* on rejection/timeout. Worth confirming #3715 adds exactly that.

barakeinav1 (changes requested):
Requesting changes on one blocker (flagged with Claude's help, so sanity-check me).

`submit_participant_info` yields but returns `Result<(), Error>`. On `Ok(())`, near-sdk emits `value_return` after the yield's `promise_return` and clobbers it — so the caller's tx returns immediate success and never sees a reject/timeout. The other yield methods (`sign`, etc.) return `()` to avoid exactly this. Fix: mirror `sign` — return `()` and `panic` on the early errors. And add a sandbox test; a unit test can't catch this.

Details: https://github.com/near/mpc/pull/3714#discussion_r3524765190

crates/contract/src/v3_12_0_state.rs
SimonRastikian, line 1: Why did you have to modify this file? AFAIK We are already at v3_13_1. I do not think this change is meaningful as we always delete the previous states files.

crates/contract/src/v3_13_0_state.rs
SimonRastikian, line 1: Not sure if it is necessary to have this state file changed in this PR. We usually tend to change it right after the smart contract is upgraded. I would set this for another issue.

crates/mpc-attestation/src/attestation.rs
```diff
@@ -346,10 +346,9 @@ impl Attestation {
         }
     }
 
-    /// Full local verification: runs DCAP (`dcap_qvl::verify::verify`) and then
-    /// the post-DCAP checks. Behind the `local-verify` feature, which pulls in
-    /// `dcap-qvl`. Used by off-chain callers and, today, by `mpc-contract`.
-    // TODO(#3264): contract drops this once DCAP moves to the verifier contract.
```
kevindeforth, line 355: Would you want to keep this method in the code-base long-term? Or would you prefer we adjust our testing setup such that we test with the deployed verifier?

kevindeforth (commented):
still reviewing, just a question for now.

crates/mpc-attestation/src/attestation.rs
pbeza: Yeah, good question! I was also wondering if we can remove this method and get rid of the `local-verify` feature flag from `mpc-contract`.

Ideally, we would have a single entry point to the DCAP verifier, namely the verifier contract, so we could adjust our testing setup to test against the deployed verifier.

But we also have the offline [`attestation-cli`](https://github.com/near/mpc/tree/main/crates/attestation-cli), which is supposed to do exactly what the contract does on-chain when verifying attestations, so we still need that entry point somewhere. Maybe long-term we can move this logic to a separate lightweight crate used by both `attestation-cli` and the verifier contract?

crates/contract/src/lib.rs
```diff
@@ -792,62 +811,137 @@ impl MpcContract {
...
+    }
 
-        if attestation_storage_must_be_paid_by_caller {
```
kevindeforth, line 827: ah lol, so in the past, we just swallowed any deposits?

crates/contract/src/lib.rs
```diff
@@ -792,62 +811,137 @@ impl MpcContract {
...
+        &mut self,
+        node_id: NodeId,
+        dstack: DstackAttestation,
+        caller_is_participant: bool,
+    ) -> Result<(), Error> {
+        let account_id = node_id.account_id.clone();
+
+        if self.pending_attestations.contains_key(&account_id) {
+            return Err(TeeError::VerificationAlreadyPending.into());
+        }
```
kevindeforth: I just realized that if it takes more than 10 seconds for the entire call chain to conclude and the attestation to be included in the finalized contract state, the node might get antsy and re-submit its attestation.

crates/contract/src/lib.rs
```diff
@@ -792,62 +811,137 @@ impl MpcContract {
...
-                        format!("TeeQuoteStatus is invalid: {err}")
-                    }
-                    AttestationSubmissionError::TlsKeyOwnedByOtherAccount => err.to_string(),
-                };
-                InvalidParameters::InvalidTeeRemoteAttestation { reason }
-            })?;
+            .then(
+                Self::ext(env::current_account_id())
+                    .with_static_gas(Gas::from_tgas(self.config.resolve_verification_tera_gas))
+                    .resolve_verification(node_id),
```
kevindeforth: since we can pass in data through the callback, could we pass the entire `DstackAttestation` here and remove the pending attestation queue?
_edit: I guess the gas cost might be too high, but worth a try?..._

kevindeforth (commented):
Thank you. It looks correct, but I need a bit more time to digest this. Also, I am wondering if we could get rid of the pending attestation struct by changing the callback logic.

crates/contract/src/lib.rs
pbeza: Yes! It was originally caught by Copilot (good bot! 🤖): https://github.com/near/mpc/pull/3714#discussion_r3500625756. I created a follow-up issue to add a test for it: https://github.com/near/mpc/issues/3760.

crates/contract/src/lib.rs
pbeza: Good point! Fixed it in https://github.com/near/mpc/pull/3714/commits/5e4ea2828b99f97203bda246cf44d72509c36fcc. It now checks whether the attestation's still being verified and, if so, treats the submit as done instead of retrying and hitting `VerificationAlreadyPending` here:

https://github.com/near/mpc/blob/74e9f72ecab916cb5a7933ee9b636a2596c012aa/crates/contract/src/lib.rs#L849-L862

crates/contract/src/lib.rs
pbeza: Personally, I think the design docs have already served their purpose, and we don’t need to keep updating them after implementation. Anyway, I updated them in follow-up PR #3715, so I’ll defer to that PR if you don’t mind.

crates/contract/src/lib.rs
pbeza: After I addressed [this comment](https://github.com/near/mpc/pull/3714#discussion_r3549859405) in https://github.com/near/mpc/commit/5e4ea2828b99f97203bda246cf44d72509c36fcc by adding the `is_verification_pending` view, I don’t think it would be easy to change this without a larger reorg:

https://github.com/near/mpc/blob/5e4ea2828b99f97203bda246cf44d72509c36fcc/crates/contract/src/lib.rs#L965-L970

Also, I believe the same in-flight `pending_*` `LookupMap` pattern is used in the contract already:

https://github.com/near/mpc/blob/52b0c01ca82582c7263e0776c454b7af82ad1467/crates/contract/src/lib.rs#L153-L155

That’s also what we agreed on in the design doc:

https://github.com/near/mpc/blob/1c5a4628fe7fc84805060665a260f67579d38eea/docs/design/attestation-verifier-contract.md?plain=1#L128-L136

Ofc, we don’t need to align 100% with the design doc, but I think redesigning this in the current PR would be non-trivial now.

Perhaps we can follow up on this separately?

pbeza:
@kevindeforth pls re-review when you get a chance. Thanks! 🙏🏼

crates/node/src/indexer.rs
```diff
@@ -252,6 +252,43 @@ impl IndexerViewClient {
...
     }
 
+    /// Whether an async Dstack attestation verification is in flight for the
+    /// account, distinguishing "still verifying" from "submission never landed".
+    pub(crate) async fn is_verification_pending(
+        &self,
+        mpc_contract_id: &AccountId,
+        account_id: &AccountId,
+    ) -> anyhow::Result<bool> {
+        let args = serde_json::to_vec(&serde_json::json!({ "account_id": account_id }))
```
kevindeforth: lets add this to `near_mpc_contract_interface::contract_args`

crates/node/src/indexer/tx_sender.rs
```diff
@@ -186,6 +186,7 @@ async fn submit_tx(
 /// Confirms whether the intended effect of the transaction request has been observed on chain.
 async fn observe_tx_result(
     indexer_state: Arc<IndexerState>,
+    signer_account_id: &AccountId,
```
kevindeforth: We don't need to change the signature of this method, lets add ephemeral information to the `ChainSendTransactionRequest` type instead.

crates/contract/src/lib.rs
```diff
@@ -868,6 +962,13 @@ impl MpcContract {
             }))
     }
 
+    /// Whether the account has an [`Attestation::Dstack`] submission awaiting
+    /// async verification, so a submitter can tell "in flight" from "never
+    /// landed" rather than resubmit and hit a [`TeeError::VerificationAlreadyPending`].
+    pub fn is_verification_pending(&self, account_id: AccountId) -> bool {
```
kevindeforth: we should query by tls key, not by account_id (one account id may have multiple tls keys, but each tls key has at most one attestation).

crates/contract/src/lib.rs
```diff
@@ -165,11 +179,17 @@ pub struct MpcContract {
...
+    /// offloads quote verification to this account; while it is [`None`], such
+    /// submissions are rejected with [`TeeError::VerifierNotConfigured`].
     // TODO(#3639): once participants have voted a verifier in, make this
     // non-optional via a migration that requires it be set.
     tee_verifier_account_id: Option<AccountId>,
     tee_verifier_votes: TeeVerifierVotes,
+    /// In-flight [`Attestation::Dstack`] verifications, one entry per submitter
+    /// account, held between the cross-contract verify-quote call and its
+    /// resolution (or the yield timeout).
+    pending_attestations: LookupMap<AccountId, PendingAttestation>,
```
kevindeforth: One `AccountId` may have multiple attestations, so the key variable of this map needs to contain the tls key of the node.

kevindeforth (commented):
Found a few things related to the pending attestations - that struct does add complexity to the code, so I will try to find some time for investigating investigate a stateless approach and see how that compares.

crates/contract/src/lib.rs
kevindeforth: Yeah, it's probably not trivial. I will try to find some time to investigate this approach - if it's too cumbersome, I am okay using the pending attestation queue, but it does introduce a bit of complexity that I would prefer to avoid.

kevindeforth: > Also, I believe the same in-flight pending_* LookupMap pattern is used in the contract already:

Yeah, that is true, but there is a subtle difference: The logic for these requests happens off-chain and we need to account for the possibility of the network not responding in time (hence the yield-resume logic).
Here, the logic is entirely on-chain, so I don't see a reason for requiring a yield-resume, as the verifier contract is always expected to produce a callback result.

kevindeforth:
@pbeza please take a look at this https://github.com/near/mpc/pull/3766

pbeza:
> @pbeza please take a look at this #3766

Replied here: https://github.com/near/mpc/pull/3766#issuecomment-4934210196.

tl;dr Thank you! I like it a lot!

crates/node/src/indexer.rs
pbeza: This method was removed in #3766 (thank you!), which has already been merged into this PR. ✅

crates/node/src/indexer/tx_sender.rs
pbeza: Reverted in https://github.com/near/mpc/pull/3766 (thank you!), which has already been merged into this PR. ✅

crates/contract/src/lib.rs
pbeza: `pending_attestations` was removed in #3766 (thank you!), which has already been merged into this PR. ✅

crates/contract/src/lib.rs
pbeza: No longer relevant after #3766 was merged.

crates/contract/src/lib.rs
pbeza: Ditto: https://github.com/near/mpc/pull/3714#discussion_r3551103919.

crates/contract/src/lib.rs
pbeza: Ditto: https://github.com/near/mpc/pull/3714#discussion_r3551103919.

crates/contract/src/lib.rs
pbeza: Ditto: https://github.com/near/mpc/pull/3714#discussion_r3551103919.

crates/contract/src/v3_12_0_state.rs
pbeza: That change was initializing `pending_attestations` in the v3.12→current migration, but after we merged the no-yield alternative (#3766) the `pending_attestations` map is gone entirely, so `v3_12_0_state.rs` is back to being identical to `main`. Same for the `v3_13_0_state.rs` snapshot; it was removed too. Resolving.

crates/contract/src/v3_13_0_state.rs
pbeza: Ditto: https://github.com/near/mpc/pull/3714#discussion_r3558287869.

crates/contract/src/tee/tee_state.rs
```diff
@@ -44,9 +48,12 @@ pub(crate) enum AttestationSubmissionError {
 }
 
 #[derive(Debug)]
+#[expect(clippy::large_enum_variant)]
 pub(crate) enum ParticipantInsertion {
     NewlyInsertedParticipant,
-    UpdatedExistingParticipant,
+    /// Holds the overwritten entry so [`TeeState::revert_dstack_store`] can put
+    /// it back if the async store is rolled back.
+    UpdatedExistingParticipant(NodeAttestation),
```
kevindeforth: Being nice and reimbursing for deposits has so many annoying side effects 🥲

crates/attestation/src/tcb_info.rs
```diff
@@ -1,4 +1,6 @@
 use alloc::string::String;
+#[cfg(feature = "borsh-schema")]
+use alloc::string::ToString;
 use alloc::vec::Vec;
 use borsh::{BorshDeserialize, BorshSerialize};
```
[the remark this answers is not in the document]
pbeza: Removing the import breaks the build with:

```
error[E0599]: no method named `to_string` found for reference `&'static str`
  --> crates/attestation/src/tcb_info.rs:19  (and :36)
help: trait `ToString` ... is implemented but not in scope
```

crates/contract/src/tee/verification_context.rs
```diff
@@ -0,0 +1,15 @@
...
+
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+#[cfg_attr(
+    all(feature = "abi", not(target_arch = "wasm32")),
+    derive(borsh::BorshSchema)
+)]
+pub struct VerificationContext {
+    pub(crate) node_id: NodeId,
+    pub(crate) attestation: DstackAttestation,
+    pub(crate) caller_is_participant: bool,
```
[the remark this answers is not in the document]
pbeza: Tracked in #3720.

crates/contract/src/tee/tee_state.rs
```diff
@@ -143,39 +150,71 @@ impl TeeState {
...
+        )?;
+
+        log_informational_advisory_ids(&advisory_ids);
+
+        self.store_verified_attestation(node_id, verified_attestation)
+    }
 
+    /// Runs the post-DCAP checks for a [`DstackAttestation`] against the
+    /// [`VerifiedReport`] the verifier returned, then stores the result.
+    pub(crate) fn verify_and_store_dstack(
```
kevindeforth, line 176: note: I know this is an existing issue, but would be nice to have some test coverage for this one too.

crates/contract/src/tee/verification_context.rs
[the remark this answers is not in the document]
kevindeforth: oh, that's a good point. If there is no risk in shrinking this, we should do that to keep gas costs low!

crates/contract/src/config.rs
```diff
@@ -34,6 +36,11 @@ const DEFAULT_REMOVE_NON_PARTICIPANT_UPDATE_VOTES_TERA_GAS: u64 = 5;
 const DEFAULT_CLEAN_FOREIGN_CHAIN_DATA_TERA_GAS: u64 = 5;
 /// Prepaid gas for a `remove_non_participant_tee_verifier_votes` call
 const DEFAULT_REMOVE_NON_PARTICIPANT_TEE_VERIFIER_VOTES_TERA_GAS: u64 = 5;
+/// Gas attached to the cross-contract `verify_quote` call on the TEE verifier.
+const DEFAULT_VERIFIER_TERA_GAS: u64 = 100;
+/// Prepaid gas for the `resolve_verification` callback. Carries the bulk of the
+/// post-DCAP work (allowlist match, RTMR3 replay, app-compose validation, store).
+const DEFAULT_RESOLVE_VERIFICATION_TERA_GAS: u64 = 60;
```
kevindeforth, line 43: note: That's expensive. How does it compare to the current costs?

crates/contract/src/lib.rs
```diff
@@ -792,62 +806,109 @@ impl MpcContract {
...
-                return Err(InvalidParameters::InsufficientDeposit {
-                    attached: attached.as_yoctonear(),
-                    required: cost.as_yoctonear(),
-                }
-                .into());
-            }
+        if caller_is_participant && !is_new_attestation {
+            refund_deposit_to(account_id);
+            return Ok(());
+        }
```
kevindeforth: note: this is fine, but I don't think it's needed - if the caller is an existing participant with a previous attestation, then the additional memory they require will be much less than what a non-participant would need and the logic below will catch that.

crates/contract/src/lib.rs
```diff
@@ -5846,247 +6007,6 @@ mod tests {
...
-        let error_string = result.unwrap_err().to_string();
-        assert!(error_string
-        .contains("Invalid TEE Remote Attestation: TeeQuoteStatus is invalid: the submitted attestation failed verification, reason: Custom(\"the allowed mpc image hashes list is empty\")"), "Got error: {}", &error_string);
-    }
-
-    /// **TLS key validation** - Tests that TEE attestation fails when TLS key doesn't match the one in report data.
-    /// Similar to the successful test method case above, but uses a deliberately corrupted TLS key to verify
-    /// that attestation validation properly checks the TLS key embedded in the attestation report.
-    #[test]
-    fn test_tee_attestation_fails_with_invalid_tls_key() {
```
kevindeforth, line 6063: I suppose the remove tests are added back in the stacked PR?

crates/contract/src/lib.rs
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/pull/3714/commits/290b3dc22cae60efaf46eff864b953b2b981863d ✅

kevindeforth (dismissed):
Thank you, this looks good!

pbeza:
JFYI, this is ready for re-review, @gilcu3 @kevindeforth. I’ll hold off on further changes until then.

crates/contract/src/lib.rs
pbeza: Yes. They were moved to the stacked PR #3715 as sandbox tests (they can't stay unit tests now that Dstack verification is async across the verifier contract): https://github.com/near/mpc/blob/bc1f2558af4e60fe18b43cfb7fac0800c0ebfa51/crates/contract/tests/sandbox/tee_verifier.rs. It uses a stub verifier to cover verified→store, rejection→refund, verifier-not-configured, crash/timeout cleanup, and OOG rollback.

crates/contract/src/tee/tee_state.rs
pbeza: 💯

crates/contract/src/lib.rs
pbeza: Great catch! I’d rather avoid unnecessary complexity, especially since this is also fairer cost-wise, so I addressed it here: https://github.com/near/mpc/pull/3714/commits/9b0851ee53eb5881727ae6d0d2991b5604f548ef + https://github.com/near/mpc/pull/3714/commits/6d2765073742247bfa1fde0f9a8eece7867ed9e3. ✅

crates/contract/src/tee/tee_state.rs
pbeza: I’ll cover that in #3715 (noted [here](https://github.com/near/mpc/pull/3715#issuecomment-4936716796)).

crates/contract/src/config.rs
pbeza: I never benchmarked it, and I think we have quite a few other unbenchmarked constants too, for example here: https://github.com/near/mpc/pull/3641#discussion_r3543560094, although that one is in the e2e tests. I’d defer this to #3715 (noted [here](https://github.com/near/mpc/pull/3715#issuecomment-4937012551)), where I added the tests. Perhaps we can benchmark it there, or in a follow-up if it’s not trivial.

pbeza:
@kevindeforth I’ve addressed your latest comments. PTAL when you get a chance.

@gilcu3, could you prioritize reviewing this on Monday? It’s been in review for a while now, and since it’s quite a big PR, it could easily drift out of sync with `main`.

I’ll follow up with the stacked PR adding tests early next week (#3715). I just need to rebase it on top of the latest version of this one and fix a few things there.

crates/contract/src/lib.rs
pbeza: No longer relevant after merging #2766.

crates/contract/src/lib.rs
```diff
@@ -792,62 +799,85 @@ impl MpcContract {
...
+                method_names::VERIFY_QUOTE.to_string(),
+                borsh::to_vec(&(&attestation.quote, &attestation.collateral))
+                    .expect("borsh serialization of verify_quote args must succeed"),
+                NearToken::from_yoctonear(0),
+                Gas::from_tgas(self.config.verifier_tera_gas),
+            )
+            .then(
+                Self::ext(env::current_account_id())
+                    .with_static_gas(Gas::from_tgas(self.config.resolve_verification_tera_gas))
+                    .with_attached_deposit(env::attached_deposit())
```
barakeinav1, line 865: Non-blocking (very unlikely edge): the deposit is forwarded onto the `resolve_verification` callback receipt, whose predecessor is the contract. If `resolve_verification` itself OOGs/panics *before* its refund runs, the runtime refunds the attached deposit to the **contract**, not the submitter — the node silently loses it. The "verifier unavailable" case is fine (it lands in the `Err(PromiseError)` arm → refund), but a callback OOG isn't recoverable. Just worth confirming `resolve_verification_tera_gas` comfortably covers post-DCAP + store + charge so this stays implausible, or a one-line note to that effect.

crates/contract/src/lib.rs
```diff
@@ -792,62 +799,85 @@ impl MpcContract {
...
+        &self,
+        account_id: &AccountId,
+        initial_storage: u64,
+    ) -> Result<(), Error> {
+        // `saturating_sub`: if a re-submission shrinks the entry, charge nothing
+        // rather than underflow. Intentional asymmetry: we do not refund freed
+        // bytes either, since the caller already paid for the larger entry.
+        let attached = env::attached_deposit();
+        let storage_used = env::storage_usage().saturating_sub(initial_storage);
+        let cost = env::storage_byte_cost().saturating_mul(storage_used as u128);
```
barakeinav1: nit: `storage_used as u128` → `u128::from(storage_used)` (lossless, avoids the `as` cast).

barakeinav1:
Timing note (node-side): after `submit_participant_info`, the node confirms the submit landed by polling the stored attestation's expiry ([`submitted_attestation_landed`](https://github.com/near/mpc/blob/70c40b0f83143550c053a12cf446a87b3f324331/crates/node/src/indexer/tx_sender.rs#L199)), but only after a fixed [`TRANSACTION_TIMEOUT`](https://github.com/near/mpc/blob/70c40b0f83143550c053a12cf446a87b3f324331/crates/node/src/indexer/tx_sender.rs#L23) = 10s [sleep](https://github.com/near/mpc/blob/70c40b0f83143550c053a12cf446a87b3f324331/crates/node/src/indexer/tx_sender.rs#L365).

So the verifier round-trip (submit → `verify_quote` → `resolve_verification` store) should stay **under ~10s**. A few-block cross-contract round-trip (~4-6s at ~1.3s/block) fits comfortably — but it's now cross-contract and possibly cross-shard, so if it ever exceeds 10s the node just re-submits and fires a duplicate `verify_quote` (harmless, second deposit refunded — no error, since there's no pending guard). Worth benchmarking the round-trip against the 10s window once the stub verifier lands, or an e2e with an injected verifier delay to confirm it stays graceful.

barakeinav1 (dismissed):
Thanks, approved — with a few comments.

crates/contract/src/lib.rs
```diff
@@ -2262,14 +2290,107 @@ impl MpcContract {
                 let promise = Promise::new(env::current_account_id()).function_call(
                     method_names::FAIL_ON_TIMEOUT.to_string(),
                     vec![],
-                    NearToken::from_near(0),
+                    NearToken::from_yoctonear(0),
```
gilcu3: I guess this one was a nit change

crates/contract/src/lib.rs
pbeza: https://github.com/near/mpc/pull/3714/commits/ee9f82eb3de2faac8404b0be2ebf622f84ba116a ✅

crates/near-mpc-contract-interface/src/deposits.rs
```diff
@@ -0,0 +1,6 @@
+//! Deposit amounts to attach to contract methods, in yoctoNEAR. One shared value
+//! for node, tests, and e2e.
+
+/// Deposit for `submit_participant_info`. The contract charges the actual storage
+/// cost and refunds the excess.
+pub const SUBMIT_PARTICIPANT_INFO_DEPOSIT_YOCTONEAR: u128 = 1_000_000_000_000_000_000_000_000;
```
gilcu3: how much is this? 1 NEAR?

crates/contract/src/lib.rs
pbeza: Let’s address it in a follow-up, as discussed [here](https://nearone.slack.com/archives/C0912BTG51T/p1783946508058609?thread_ts=1782978028.783319&cid=C0912BTG51T).

crates/contract/src/tee/tee_state.rs
```diff
@@ -188,20 +227,44 @@ impl TeeState {
...
                 node_id,
                 verified_attestation,
             },
         );
 
-        Ok(match insertion {
-            Some(_previous_attestation) => ParticipantInsertion::UpdatedExistingParticipant,
+        // Materialize the insert before the storage-usage charge reads it;
+        // `IterableMap` otherwise defers the write to flush-on-Drop
+        self.stored_attestations.flush();
```
gilcu3: could be worth adding this as a doc to the function, as this behavior is relevant for the callers

gilcu3 (dismissed):
Thank you for the hard work!

Left some nits

crates/contract/src/lib.rs
pbeza: Yeah, Claude updated it and I missed that. From what I can see, `NearToken::from_yoctonear(0)` is way more common in the codebase than `NearToken::from_near(0)` (`22` vs `5`), so I changed two more occurrences too: https://github.com/near/mpc/pull/3714/commits/8e757458747c2b0fc86dea1833220c84f69aa002 (I only left the ones in the test helpers as they were).

crates/near-mpc-contract-interface/src/deposits.rs
pbeza: Yeah. I initially thought we should keep it fine-grained, but I think using whole NEAR values is enough here. I switched it back to NEAR across the codebase: https://github.com/near/mpc/pull/3714/commits/f2f470a91d0432dbcf6e512dd93bf8ed76b60c19. Sorry, this probably should’ve been a separate PR, but it’s a trivial change.

crates/contract/src/lib.rs
pbeza: Okay, I [changed](https://github.com/near/mpc/pull/3714#discussion_r3579134766) my mind. 🙃 I’m reverting the above for [readability](https://github.com/near/mpc/pull/3714#discussion_r3578134076). Sorry for the confusion!

crates/contract/src/tee/tee_state.rs
pbeza: Good point. Addressed here: https://github.com/near/mpc/pull/3714/commits/6d228ab914ce7b27ddf2015aa6ac32a5fb1a8203. ✅

pbeza:
@gilcu3 @kevindeforth @barakeinav1, thanks for your reviews! Feel free to re-review and re-approve when you get a chance. I’d like to make sure this PR is stable and approved, since I have a stacked PR with tests (#3715) that I’ll ask you to review in the coming days.

I discussed it with @netrome, and the plan is to merge the test PR into this one first, then merge everything to `main`. We’ll also need to make sure the verifier contract is deployed before cutting a release that includes this PR.

gilcu3 (dismissed):

crates/contract/src/tee/tee_state.rs
```diff
@@ -188,20 +231,43 @@ impl TeeState {
...
+        match insertion {
+            ParticipantInsertion::UpdatedExistingParticipant(previous) => {
+                self.stored_attestations
+                    .insert(tls_public_key.clone(), previous);
+            }
+            ParticipantInsertion::NewlyInsertedParticipant => {
+                self.stored_attestations.remove(tls_public_key);
+            }
+        }
+    }
```
kevindeforth: Given the complexity of this, I am considering if we should just swallow the deposit in case of failure.

We could require a minimum on the deposit (checked before insertion) and in case actual storage costs exceed that minimum, we fail and swallow the deposit.

kevindeforth (dismissed):
Thank you!

crates/contract/src/tee/tee_state.rs
pbeza: Sorry, I forgot to address this after [freezing](https://github.com/near/mpc/pull/3714#issuecomment-4969567605) the PR and before adding the tests in #3715. I’ll try to sneak it in if the review of #3715 takes a bit longer, since it shouldn’t be too invasive. Otherwise, I can handle it in a follow-up. I agree it’s a bit more complex, but it feels fairer cost-wise since we refund the cost this way.

pbeza: > Sorry, I forgot to address this after [freezing](https://github.com/near/mpc/pull/3714#issuecomment-4969567605) the PR and before adding the tests in #3715. I’ll try to sneak it in if the review of #3715 takes a bit longer, since it shouldn’t be too invasive. Otherwise, I can handle it in a follow-up. I agree it’s a bit more complex, but it feels fairer cost-wise since we refund the cost this way.

@gilcu3 @kevindeforth addressed in https://github.com/near/mpc/pull/3714/commits/2b4ea859a80a725567f12a486acfcc5abc7971d0. ✅ I’d appreciate a re-review so I can make sure the tests in #3715 are aligned correctly.

kevindeforth: Thank you, this looks good!
> worst-case ~520 bytes

Is this something we can guarantee?

kevindeforth (dismissed):
Thank you!
I think the (storage) deposit handling will need a follow-up, but lets do that in a future PR, not here.

Right now, we have the following behavior:
- under-funded (< 0.1 NEAR), we reject the submission
- funded (>= 0.1 NEAR) and success --> the entire deposit is kept
- funded (>= 0.1 NEAR)  and failure --> full refund

This means the following:
- re-submissions overpay. Re-submissions likely only result in very little memory change, yet pay the full deposit every time
- changes in attestation structure may lead to over- or underpayment.

A lot of this is acceptable in the short- and even in the medium-term, and all of it is easier to address in a follow-up PR than in this PR, so it's not a blocker, but we should open a follow-up issue and prioritize it for this week.

crates/contract/src/tee/tee_state.rs
pbeza: >> worst-case ~520 bytes
>
> Is this something we can guarantee?

It’s an estimate Claude helped me come up with, but I checked it and it made sense to me. The stored entry is `NodeAttestation = NodeId + VerifiedAttestation`, and every field is fixed-size except `account_id`, which NEAR’s `AccountId` caps at 64 bytes. The heavy DCAP data (`quote`, `collateral`, `advisory_ids`, `app-compose`) is only part of the input attestation and gets discarded after verification, so none of it is persisted.

I also quickly vibe-coded an explicit test for this in #3856:

```rust
    #[test]
    fn node_attestation__should_match_expected_max_borsh_size() {
        // Given: the largest entry each variant can produce (64-byte account id,
        // all fields present at their maximum).
        let max_dstack_borsh_size = 445;
        let max_mock_borsh_size = 450;
        let node_id = create_node_id(
            &"a".repeat(64).parse().unwrap(),
            &bogus_ed25519_public_key(),
        );
        let dstack = NodeAttestation {
            node_id: node_id.clone(),
            verified_attestation: VerifiedAttestation::Dstack(ValidatedDstackAttestation {
                mpc_image_hash: [0xff; 32].into(),
                launcher_compose_hash: [0xff; 32].into(),
                expiry_timestamp_seconds: u64::MAX,
                measurements: default_measurements()[0],
            }),
        };
        let mock = NodeAttestation {
            node_id,
            verified_attestation: VerifiedAttestation::Mock(
                mpc_attestation::attestation::MockAttestation::WithConstraints {
                    mpc_docker_image_hash: Some([0xff; 32].into()),
                    launcher_docker_compose_hash: Some([0xff; 32].into()),
                    expiry_timestamp_seconds: Some(u64::MAX),
                    expected_measurements: Some(default_measurements()[0]),
                },
            ),
        };

        // When
        let dstack_size = borsh::to_vec(&dstack).unwrap().len();
        let mock_size = borsh::to_vec(&mock).unwrap().len();

        // Then
        assert_eq!(dstack_size, max_dstack_borsh_size);
        assert_eq!(mock_size, max_mock_borsh_size);
    }
```

If you’d rather have it in this PR, feel free to merge it in here.

Also, we already have `minimum_attestation_storage_deposit__should_cover_worst_case_entry`, which builds the worst-case entry using a 64-character ID, all-`0xff` hashes, and a `u64::MAX` expiry.

https://github.com/near/mpc/blob/195d7b3e47e4e280a92cb6d666dcb57d1643fff5/crates/contract/src/lib.rs#L7881-L7920

It measures the actual `storage_usage` delta and fails if the flat fee doesn’t cover it, so any future schema change that makes the entry larger should break CI.

pbeza:
Thanks for the quick re-review and for re-approving it!

> (...) and all of it is easier to address in a follow-up PR than in this PR, so it's not a blocker, but we should open a follow-up issue and prioritize it for this week.

Created a follow-up issue: https://github.com/near/mpc/issues/3857. ✅

crates/contract/src/tee/tee_state.rs
```diff
@@ -33,8 +37,8 @@ pub enum TeeQuoteStatus {
     Invalid(String),
 }
 
-#[derive(Debug, Clone, thiserror::Error)]
-pub(crate) enum AttestationSubmissionError {
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
+pub enum AttestationSubmissionError {
```
gilcu3, line 41: so do we need these new trait in prod or only in tests? Just asking in case it was a rebase leftover

crates/contract/src/lib.rs
```diff
@@ -7887,4 +7877,44 @@ mod tests {
...
     }
+
+    // Catches only entry-size growth: fails if a schema change makes the stored entry
+    // cost more than the fee at today's storage_byte_cost. It cannot see a future
+    // storage_byte_cost increase on a live contract; the fee's margin covers that.
+    #[test]
+    fn minimum_attestation_storage_deposit__should_cover_worst_case_entry() {
+        // Given: the largest entry a submission can store. NEAR caps an account id
+        // at 64 bytes; every other field is fixed-size, so this is the worst case.
+        testing_env!(VMContextBuilder::new().build());
```
gilcu3, line 7965: nit: the comments explaining given/when/then might have slipped through

gilcu3 (dismissed):
Thank you!

crates/contract/src/config.rs
```diff
@@ -34,6 +36,11 @@ const DEFAULT_REMOVE_NON_PARTICIPANT_UPDATE_VOTES_TERA_GAS: u64 = 5;
 const DEFAULT_CLEAN_FOREIGN_CHAIN_DATA_TERA_GAS: u64 = 5;
 /// Prepaid gas for a `remove_non_participant_tee_verifier_votes` call
 const DEFAULT_REMOVE_NON_PARTICIPANT_TEE_VERIFIER_VOTES_TERA_GAS: u64 = 5;
+/// Gas attached to the cross-contract `verify_quote` call on the TEE verifier.
+const DEFAULT_VERIFIER_TERA_GAS: u64 = 100;
```
barakeinav1: Heads-up — this default looks too low for real on-chain DCAP.

Testing this branch end-to-end on a real localnet TDX cluster (real dstack quotes, stock release-v3.13 node, this contract + the real `tee-verifier`), attestation never gets through. The cross-contract `verify_quote` call runs out of gas — the receipt fails with `Exceeded the prepaid gas`, so real `dcap_qvl::verify` seems to need more than 100 Tgas. After it OOGs the contract hits `verifier did not respond within the yield-resume window`, the `pending_attestation` gets stuck, and every retry then fails with `already in flight` — so no node becomes attested and keygen stalls.

I don't know the exact figure yet — still testing to pin down how much DCAP actually needs (and whether it fits under the 300 Tgas tx cap once you add the resolve + callback gas). Flagging early. Note the sandbox tests in #3715 use the real verifier but only feed a malformed quote that's rejected at parse (≈no gas), so the expensive full-verification path that blows this budget isn't exercised.

barakeinav1 (changes requested):
see https://github.com/near/mpc/pull/3714#discussion_r3623683369

crates/contract/src/config.rs
barakeinav1: I measured 173 Tgas, but we should probably use a higher number (note, that we max tag is 300, and we need at least 70 for the rest of the method, so 200-220 tgas sounds good) in case we want to upgrade the verifier  without change the const.
just thinking out loud
 updating the const is a config update vote, correct? so it should be easy enough to update it.
 wandering if we should add this to the verifier contract id vote (so it can be done in one vote instead of 2), but I don't think this is crtical.

barakeinav1: I used 200 tgas- and the test passed!

pbeza: Bumped it here: https://github.com/near/mpc/pull/3714/commits/90693be7da1bb432c124b6c8f6468ad63b8b53e6. ✅

barakeinav1 (dismissed):
thanks for fixing

gilcu3 (dismissed):

kevindeforth (dismissed):
Thank you!

kevindeforth (approved):

gilcu3 (approved):

PR #3716, opened 2026-07-01
feat: add foreign chain tx config checker

### LINKED ISSUE #3721: Add a foreign-chain RPC config tester binary
### Background

Operators occasionally ship a broken foreign-chain RPC provider config (bad URL, wrong or expired API key/auth, or a provider pointed at the wrong network). Today this is only caught in production, once `verify_foreign_tx` signing starts failing.

The `#[ignore]`d "manual" tests in `crates/foreign-chain-inspector/tests/*_rpc_manual.rs` exercise live providers with fixed transactions, but they target hardcoded public endpoints rather than an operator's actual configured providers.

### Acceptance Criteria

- A standalone binary loads a node config via `--config <path>` and checks every configured foreign-chain RPC provider, printing a per-provider result (pass / fail-with-reason / skipped) and exiting non-zero if any provider fails.
- All providers are checked in a single run (no early exit on the first failure), each queried individually rather than via the node's consensus fan-out.
- Config loading supports YAML and TOML, including the dstack `user-config.toml` layout (`foreign_chains` under `mpc_node_config.node`), reusing the canonical `ForeignChainsConfig` type and the node's auth-resolution path.
- Each provider is verified by extracting a known value from a fixed reference transaction through the same inspector the node uses.
- Network-aware: mainnet/testnet reference transactions are selected per network, auto-detected from the config (`chain_id` / `mpc_contract_id`) and overridable with `--network`; a chain with no reference transaction for the selected network is skipped, not failed.

gilcu3:
Closes #3721

### DIFF
--- a/crates/foreign-chain-config-tester/src/checks.rs
+++ b/crates/foreign-chain-config-tester/src/checks.rs
@@ -1,0 +1,226 @@
+//! Per-provider golden checks: run a fixed request and verify the extracted value.
+
+use std::time::Duration;
+
+use anyhow::{Context, bail, ensure};
+use foreign_chain_inspector::{
+    BlockConfirmations, EthereumFinality, ForeignChainInspector,
+    aptos::{
+        AptosExtractedValue, AptosTransactionHash,
+        inspector::{AptosExtractor, AptosFinality, AptosInspector},
+    },
+    bitcoin::{
+        BitcoinExtractedValue, BitcoinTransactionHash,
+        inspector::{BitcoinExtractor, BitcoinInspector},
+    },
+    evm::inspector::{EvmChain, EvmExtractedValue, EvmExtractor, EvmInspector},
+    http_client::HttpClient,
+    starknet::{
+        StarknetExtractedValue, StarknetTransactionHash,
+        inspector::{StarknetExtractor, StarknetFinality, StarknetInspector},
+    },
+};
+use foreign_chain_rpc_interfaces::aptos::ReqwestAptosClient;
+use http::{HeaderName, HeaderValue};
+
+fn verify_block_hash(expected: [u8; 32], got: [u8; 32]) -> anyhow::Result<()> {
+    if got != expected {
+        return Err(anyhow::anyhow!(
+            "block hash mismatch: expected 0x{}, got 0x{} — is this provider on the expected network?",
+            hex::encode(expected),
+            hex::encode(got),
+        ));
+    }
+    Ok(())
+}
+
+pub async fn check_evm<Chain>(
+    client: HttpClient,
+    tx: [u8; 32],
+    expected_block_hash: [u8; 32],
+) -> anyhow::Result<()>
+where
+    Chain: EvmChain + Send + Sync,
+{
+    let inspector = EvmInspector::<HttpClient, Chain>::new(client);
+    let values = inspector
+        .extract(
+            Chain::TransactionHash::from(tx),
+            EthereumFinality::Finalized,
+            vec![EvmExtractor::BlockHash],
+        )
+        .await?;
+    match values.into_iter().next().context("RPC returned no value")? {
+        EvmExtractedValue::BlockHash(hash) => {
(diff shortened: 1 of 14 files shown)


crates/foreign-chain-config-tester/src/checks.rs
anodar, line 1: Not for this PR, but we should probably move this, backup CLI, ckd-example-cli (and whatever else CLI we may have) under `/crates/tools`.

crates/foreign-chain-config-tester/src/main.rs
```diff
@@ -0,0 +1,365 @@
...
+    name.as_str().to_owned()
+}
+
+fn prepare_jsonrpc(provider: &ForeignChainProviderConfig) -> anyhow::Result<HttpClient> {
+    let mut url = provider.rpc_url.clone();
+    let auth = auth_config_to_rpc_auth(provider.auth.clone(), &mut url)?;
+    build_http_client(url, auth).map_err(|e| anyhow::anyhow!("failed to build HTTP client: {e}"))
+}
+
+fn prepare_aptos(
```
anodar, line 140: We can probably code from `crates/node/src/providers/verify_foreign_tx.rs` here (`new_aptos_inspector` and few more I guess) instead for constructiong auth&client, possibly with some refactoring.


That way we also test production path too.

crates/foreign-chain-config-tester/src/config.rs
```diff
@@ -0,0 +1,262 @@
...
+
+const CONTRACT_ID_PATHS: &[&[&str]] = &[
+    &["mpc_node_config", "node", "indexer", "mpc_contract_id"],
+    &["node", "indexer", "mpc_contract_id"],
+    &["indexer", "mpc_contract_id"],
+];
+
+fn classify_network(chain_id: Option<&str>, contract_id: Option<&str>) -> Option<Network> {
+    match chain_id {
+        Some("mainnet") => return Some(Network::Mainnet),
```
anodar: nit: we can use ChainId enum instead of hardcoding this here.

https://github.com/near/mpc/blob/main/crates/node-config/src/start.rs#L158

anodar (dismissed):
Thank you! 🙏

crates/foreign-chain-config-tester/src/main.rs
gilcu3: given that the mechanism there uses the fanout, it seems it would require a sizable refactor, we better leave it for a follow-up

crates/foreign-chain-config-tester/src/config.rs
gilcu3: addressed

anodar (approved):

pbeza (approved):
LGTM, shallow review though.

crates/foreign-chain-config-tester/src/main.rs
```diff
@@ -0,0 +1,367 @@
...
+fn no_reference_reason(network: Network) -> String {
+    format!(
+        "no {} reference transaction for this chain",
+        network.label()
+    )
+}
+
+fn timeout_of(cfg: &ForeignChainConfig) -> Duration {
+    Duration::from_secs(cfg.timeout_sec.get())
+}
```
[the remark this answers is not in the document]
gilcu3: no need, this is just a cli for testing

crates/foreign-chain-config-tester/src/config.rs
```diff
@@ -0,0 +1,269 @@
...
+use anyhow::{Context, bail};
+use mpc_node_config::{ChainId, ForeignChainsConfig};
+use serde::Deserialize;
+use serde::de::IntoDeserializer;
+use serde::de::value::{Error as ValueError, StrDeserializer};
+
+use crate::golden::Network;
+
+/// Paths where `foreign_chains` may live, most-nested first so a wrapped config
+/// matches before a barer one.
```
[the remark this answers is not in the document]
gilcu3: probably this was existing, leaving as is

PR #3726, opened 2026-07-01
fix: resharing reconstruction threshold in the smart contract

### LINKED ISSUE #3725: Per domain SC resharing falsely providing old_threshold as the new threshold
### Background

The per domain implementation of the resharing on the contract side introduced a bug that was captured in the tests when integrating per-domain threshold on the node side.

The bug consists of the smart contract providing the domain's old non updated threshold as a reconstruction threshold for the new_threshold. This implies that the domain's Reconstruction threshold would update but the secret would not update it's threshold.

### User Story

_No response_

### Acceptance Criteria

Provide the resharing with the effective threshold of the domain.
Write unit tests for this

_No response_

SimonRastikian:
Closes #3725

### DIFF
--- a/crates/contract/src/state/resharing.rs
+++ b/crates/contract/src/state/resharing.rs
@@ -82,9 +82,8 @@ impl ResharingContractState {
                     self.prospective_epoch_id().next(),
                     self.previous_running_state
                         .domains
-                        .get_domain_by_index(0)
-                        .unwrap()
-                        .clone(),
+                        .effective_domain_by_index(0, proposal.per_domain_thresholds())
+                        .unwrap(),
                     proposal.parameters().clone(),
                 ),
                 cancellation_requests: HashSet::new(),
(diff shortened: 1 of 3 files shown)


crates/contract/src/state/resharing.rs
```diff
@@ -82,9 +82,8 @@ impl ResharingContractState {
                     self.prospective_epoch_id().next(),
                     self.previous_running_state
                         .domains
-                        .get_domain_by_index(0)
-                        .unwrap()
-                        .clone(),
+                        .effective_domain_by_index(0, proposal.per_domain_thresholds())
```
SimonRastikian, line 85: This used to output the domain itself, now it output the domain with a modified (overwritten) threshold that is collected from the proposal)

SimonRastikian:
Closed as the resharing issue was fixed by collecting the new threshold directly from the contract state at the indexer level

PR #3727, opened 2026-07-02
unify contract callsites

kevindeforth:
_Disclaimer: This is a proposal, subject to discussion, not intended for merge yet_

## Background

We currently have four different components that call the MPC contract:
1. devnet
2. e2e test
3. contract sandbox tests
4. MPC node

Each component follows its own, unique call-pattern and constructs the json function call arguments.

This has two downsides:

1. A developer in this repository needs to know four distinct patterns for achieving the same thing. That creates mental overhead not only during coding, but also during review.
2. If a contract API changes, it requires changes to all four repositories above, having a huge blast radius. Since a lot of the contract methods are not used by external users, but internal to node and operator calls, such changes do occur.

This PR is a proposal on how to unify the pattern across our crates. It centralizes the MPC contract API in contract-interface/src/call_args and introduces a new trait `CallContract`:

```rust
pub trait CallContract {
    /// Backend-specific successful call outcome.
    type Output;

    fn call_contract(
        &self,
        contract_id: &AccountId,
        call_args: FunctionCallArgs,
    ) -> impl Future<Output = Result<Self::Output, CallError>> + Send;
}
```
The idea is to have a central source of truth for constructing contract call arguments, and allowing each component (devnet, e2e test, contract sandbox tests and MPC node), to implement `CallContract` for their own types. This way, no matter the crate, a call to the contract will always read similar:

```rust
let sign_request_future = send_sign_request(contract_caller, contract_account_id, sign_request_args);
```

### Examples

Currently, each one of these components has their own call-pattern. Some examples:

#### Propose Contract Update:
devenet:
https://github.com/near/mpc/blob/cd067dd84a8ede25c8e2c6ff9ac9c63eb5d3a4d0/crates/devnet/src/mpc.rs#L505-L523

e2e test:
https://github.com/near/mpc/blob/cd067dd84a8ede25c8e2c6ff9ac9c63eb5d3a4d0/crates/e2e-tests/src/cluster.rs#L940-L951

sandbox:
https://github.com/near/mpc/blob/cd067dd84a8ede25c8e2c6ff9ac9c63eb5d3a4d0/crates/contract/tests/sandbox/upgrade_from_current_contract.rs#L59-L69

#### Signature Request:
devnet:

https://github.com/near/mpc/blob/0c217e081f76ad7de9903b6f8ed871bd9b226e3e/crates/devnet/src/loadtest.rs#L289-L333

e2e-test:
https://github.com/near/mpc/blob/cd067dd84a8ede25c8e2c6ff9ac9c63eb5d3a4d0/crates/e2e-tests/src/cluster.rs#L779-L789

sandbox:
https://github.com/near/mpc/blob/cd067dd84a8ede25c8e2c6ff9ac9c63eb5d3a4d0/crates/contract/tests/sandbox/sign.rs#L201-L206

The node follows an entirely different pattern, having a registry for the call patterns in indexer/types.rs

https://github.com/near/mpc/blob/cd067dd84a8ede25c8e2c6ff9ac9c63eb5d3a4d0/crates/node/src/indexer/types.rs#L104-L122

and the call submission logic in tx_sender:

https://github.com/near/mpc/blob/cd067dd84a8ede25c8e2c6ff9ac9c63eb5d3a4d0/crates/node/src/indexer/tx_sender.rs#L147-L184


kevindeforth:
Closing this. Much of this has been achieved and improved in https://github.com/near/mpc/issues/3693

PR #3729, opened 2026-07-02
refactor: establish contract-interfac/call_args.rs as source of truth

### LINKED ISSUE #3728: move Chain*Args from node to contract-interface crate
### Background

sub-task of #3693 - as a first step, we move the Chain* args constructs from node/indexer/types.rs to the contract-interface crate.

### User Story

As a dev, I like a single source of truth

### Acceptance Criteria

- Establish a call-args feature in the contract-interface crate, such that other modules can use it.
- The node gets the contract call argument structure from the contract-interface crate.

https://nearone.slack.com/archives/C07UW93JVQ8/p1782986596388519?thread_ts=1782985192.945249&cid=C07UW93JVQ8

kevindeforth:
resolves #3728 

ci blocked by #3730

### DIFF
--- a/crates/node/src/mpc_client.rs
+++ b/crates/node/src/mpc_client.rs
@@ -20,6 +20,7 @@ use crate::storage::{
     CKDRequestStorage, SignRequestStorage, VerifyForeignTransactionRequestStorage,
 };
 use crate::tracking::{self, AutoAbortTaskCollection};
+use crate::trait_extensions::convert_to_contract_dto::IntoContractInterfaceType;
 use crate::types::SignatureRequest;
 use crate::types::{CKDRequest, RequestsUpdate, VerifyForeignTxRequest};
 use crate::web::{DebugRequest, DebugRequestKind};
@@ -501,14 +502,7 @@ where
                                         .await??;
 
                                         let response = contract_args::CKDRespondArgs::new(
-                                            near_mpc_contract_interface::types::CKDRequest {
-                                                app_public_key: ckd_attempt
-                                                    .request
-                                                    .app_public_key
-                                                    .clone(),
-                                                app_id: ckd_attempt.request.app_id.clone(),
-                                                domain_id: ckd_attempt.request.domain_id,
-                                            },
+                                            (&ckd_attempt.request).into_contract_interface_type(),
                                             CKDResponse {
                                                 big_y: (&response.0.0).into(),
                                                 big_c: (&response.0.1).into(),
--- a/crates/node/src/mpc_client.rs
+++ b/crates/node/src/mpc_client.rs
@@ -501,11 +501,14 @@ where
                                         .await??;
 
                                         let response = contract_args::CKDRespondArgs {
-                                            request: contract_args::CKDRequest::new(
-                                                ckd_attempt.request.app_public_key.clone(),
-                                                ckd_attempt.request.app_id.clone(),
-                                                ckd_attempt.request.domain_id,
-                                            ),
+                                            request: contract_args::CKDRequest {
+                                                app_public_key: ckd_attempt
+                                                    .request
+                                                    .app_public_key
+                                                    .clone(),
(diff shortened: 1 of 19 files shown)


crates/node/src/mpc_client.rs
```diff
@@ -497,13 +500,17 @@ where
...
+                                            request: contract_args::CKDRequest::new(
+                                                ckd_attempt.request.app_public_key.clone(),
+                                                ckd_attempt.request.app_id.clone(),
+                                                ckd_attempt.request.domain_id,
+                                            ),
+                                            response: CKDResponse {
                                                 big_y: (&response.0.0).into(),
                                                 big_c: (&response.0.1).into(),
                                             },
-                                        )?;
```
kevindeforth, line 506: note: this method never required a failure path.

pbeza (dismissed):
LGTM.

---

FWIW, re [this](https://nearone.slack.com/archives/C07UW93JVQ8/p1782985192945249) and this PR: I had a short session with Claude to research how other teams handle this, since it seemed like a common problem. I found there’s a [`#[near_kit::contract]`](https://docs.rs/near-kit/latest/near_kit/contract/) macro that generates typed call builders from a trait. Since these structs (`SignatureRespondArgs`, `CKDRespondArgs`, `VotePkArgs`, `SubmitParticipantInfoArgs`, etc.) now live in one place here, I think they're exactly the input the macro would need:

```rust
#[near_kit::contract]
pub trait MpcSigner {
    #[call]
    fn respond(&mut self, args: SignatureRespondArgs);
    #[call]
    fn respond_ckd(&mut self, args: CKDRespondArgs);
    // ...vote_pk, submit_participant_info, etc.
}
```

NEAR Intents already does this for their `Wallet` contract, reused as-is by both their production relayer and sandbox tests ([client](https://github.com/near/intents/blob/f293f84076874048d8436d882ea034bf1bad43ba/crates/wallet/client/src/lib.rs#L7-L13), [sandbox reuse](https://github.com/near/intents/blob/f293f84076874048d8436d882ea034bf1bad43ba/crates/testing/sandbox/src/extensions/wallet.rs#L7)).

Just flagging it before the next layer (call construction / CallContract; https://github.com/near/mpc/pull/3727) gets built by hand, in case it saves us that work.

kevindeforth:
> I found there’s a [`#[near_kit::contract]`](https://docs.rs/near-kit/latest/near_kit/contract/) macro that generates typed call builders from a trait. Since these structs (`SignatureRespondArgs`, `CKDRespondArgs`, `VotePkArgs`, `SubmitParticipantInfoArgs`, etc.) now live in one place here, I think they're exactly the input the macro would need:

Oh, that's very cool! I need to check if we can make this work with all the other stuff we use. Do you know how stable `near_kit` is at this point?

pbeza:
> > I found there’s a [`#[near_kit::contract]`](https://docs.rs/near-kit/latest/near_kit/contract/) macro that generates typed call builders from a trait. Since these structs (`SignatureRespondArgs`, `CKDRespondArgs`, `VotePkArgs`, `SubmitParticipantInfoArgs`, etc.) now live in one place here, I think they're exactly the input the macro would need:
> 
> Oh, that's very cool! I need to check if we can make this work with all the other stuff we use. Do you know how stable `near_kit` is at this point?

Not totally sure, but if NEAR Intents [is](https://github.com/near/intents/blob/f293f84076874048d8436d882ea034bf1bad43ba/crates/wallet/client/src/lib.rs#L7-L13) using it, and that’s apparently one of NEAR’s core products, I’d guess it’s stable enough. Also looks like they’ve been releasing pretty regularly for the last ~5 months: https://crates.io/crates/near-kit/versions. We can ask Ricky, who seems to be `near-kit`’s release manager (cc @r-near).

anodar (dismissed):

anodar (approved):

pbeza (approved):

Issue #3734, opened 2026-07-02
Make attestation expiry 7 days by default, shorten to 1 day only on verifier rotation


barakeinav1:
## Background

Design doc [#3376](https://github.com/near/mpc/pull/3376) ([`docs/design/attestation-verifier-contract.md`](https://github.com/near/mpc/blob/main/docs/design/attestation-verifier-contract.md), §Governance and upgrades) — and the implementation in [#3626](https://github.com/near/mpc/pull/3626) — bound trust in attestations wrongly accepted by a buggy verifier by **permanently** lowering `DEFAULT_EXPIRATION_DURATION_SECONDS` from 7 days to 1 day (`crates/mpc-attestation/src/attestation.rs:28`). After a verifier rotation, a wrongly-accepted entry then ages out within ~1 day instead of 7.

The problem: a **permanent** 1-day expiry is a liveness risk. Attestation freshness depends on Intel's PCCS/PCS collateral being reachable at resubmit time. A *global* PCCS outage (Intel's servers down) lasting ≥ 1 day could expire every node's attestation roughly simultaneously and knock out the whole MPC cluster. The 7-day window exists precisely to give operators/nodes slack across transient connectivity issues, races, and upstream outages; 1 day is too tight to be the steady-state default.

The 1-day containment window is only *needed* right after a verifier rotation — the moment we stop trusting entries the old (possibly buggy) verifier produced. Outside that window, the extra availability margin of 7 days is worth keeping.

## User Story

As an MPC node operator, I need attestations to stay valid long enough to survive a transient global PCCS/collateral outage, so that my node (and the cluster) is not evicted during an Intel-side incident — while still allowing the protocol to quickly age out attestations produced by a verifier we've just rotated away from.

## Acceptance Criteria

Change the design **and** implementation so that:

- [ ] `DEFAULT_EXPIRATION_DURATION_SECONDS` is restored to **7 days** — the default expiry stamped on every newly submitted attestation, including under the new verifier contract.
- [ ] On each **verifier rotation** (a `vote_tee_verifier_change` that crosses threshold and updates `tee_verifier_account_id`), the contract shortens the expiry of all **currently stored** attestations to **1 day** from that point — i.e. cap each stored entry's `expiry_timestamp_seconds` at `min(existing, now + 1 day)` (never extend).
- [ ] Attestations submitted **after** the rotation are stamped with the normal 7-day expiry again (nodes resubmit hourly, so honest nodes refresh well within the shortened window).
[...]

Issue #1639, opened 2025-12-10
use expiration durations from the attestation certificate


DSharifi:
### Background

After #1629 we enforce a fixed timeout for attestation validity instead of storing the attestation on chain. As the contract only stores valid attestations, we assume the attestation only becomes invalid due to timeouts, as we currently don't have a mechanism to update the CRLs on chain.

Intel’s certificates and CRLs already include issueDate / nextUpdate fields that define their actual validity windows (≈30 days for PCK CRLs & TCB Info).
Per @barakeinav1’s analysis[ (see linked table)](https://github.com/near/mpc/issues/1629#issuecomment-3635983666), these TTLs are reliable and usable on-chain.

We should look into using these TTLs instead of hard coded values on the contract for expiration of attestations. I.e. how often a node should be enforced to submit attestations.

### User Story

As a contract developer, I want to use the TTLs embedded in Intel certificates and CRLs so that attestation expiry is enforced accurately without relying on a hard coded timeouts.

### Acceptance Criteria

The contract uses the timeout defined in the PCK CRL for expiry of attestations instead of an arbitrary value defined in the contract.

netrome:
Still makes sense. Would be nice to look into relatively soon. Keeping as P2

barakeinav1:
The mock path now also stamps `now + DEFAULT` at acceptance (`AcceptedAttestation::mock` → `MockAttestation::with_expiry`), mirroring the Dstack `now+DEFAULT` this issue is about — so when certificate-derived expiry lands here, the mock branch should be revisited too. (Mock currently only *caps* a submitted expiry rather than fully owning it like Dstack; full alignment is tracked in #4005. Introduced by #3785 / #3293.)

barakeinav1:
Design: #4292. This issue is the final step — the commit where behaviour changes. Prerequisites, each landable on its own:

| | Prerequisite | Depends on |
|---|---|---|
| A | #4515 — `verify_quote_with_claims` on `tee-verifier` | — |
| | near/mpc-private#625 — deploy that verifier and vote it in | #4515 |
| B | #4301 / #4500 — stored submission timestamp (plus #4498, #4504) | — |
| C | #4516 — evict launcher images by reference, not TTL alone | — |
| D | #4517 — refresh collateral early, refuse near-expiry | — |
| E | #4518 — pin the contract clock in sandbox tests | — |
| F | #4519 — size the verifier gas budget for `claims()` | #4515 |

[...]

PR #3736, opened 2026-07-05
fix(node): confirm submit_participant_info via advancing attestation expiry

### LINKED ISSUE #3686: Attestation freshness check breaks under node/contract version skew, causing a submit_participant_info retry storm
### Background

The post-submit confirmation for `submit_participant_info` reconstructs the attestation's creation time by subtracting a hard-coded constant from the contract-stored expiry (`crates/node/src/indexer/tx_sender.rs`, `TODO(#1639)`):

```rust
storage_time    = stored_expiry - DEFAULT_EXPIRATION_DURATION_SECONDS;
attestation_age = |now - storage_time|;
is_fresh        = attestation_age < MAX_ATTESTATION_AGE; // 120s
```

But `stored_expiry` is stamped by whoever **verifies and stores** the attestation — the **contract** (`AcceptedAttestation::dstack`, `crates/mpc-attestation/src/attestation.rs:71`) — using the *contract's* compiled-in `DEFAULT_EXPIRATION_DURATION_SECONDS`. The node then subtracts the *node's* constant. When the two differ, the reconstructed age is off by the difference.

This is live now: contract runs **3.12** (`DEFAULT_EXPIRATION_DURATION_SECONDS = 7 days`), nodes run **3.13** (changed to `1 day` in #3626). Reconstructed age = `7d − 1d = 6 days`, far over the 120s `MAX_ATTESTATION_AGE`, so the check never returns `Executed`:

```
tx_sender: node found dstack attestation on chain attestation_age=518390.57s attestation_is_fresh=false
ERROR periodic_attestation_submission: failed to submit attestation cause=attestation submission was not executed backoff_duration=60s
```

Each submission lands on-chain successfully, but the node treats it as `NotExecuted` and retries at the 60s backoff ceiling (+10s observe) — a ~70s `submit_participant_info` loop per node, indefinitely, each tx carrying a ~60 KB quote+collateral payload.

### Acceptance Criteria

- Attestation submission confirmation does not depend on node and contract sharing the same `DEFAULT_EXPIRATION_DURATION_SECONDS` (confirm via stored-vs-submitted identity, or an explicit creation timestamp, rather than `expiry − constant`).
- A node running a different version from the contract does not enter a permanent `submit_participant_info` resubmission loop.

Related: #1639 (extract timestamp from the certificate instead of the `now ± constant` heuristic).

barakeinav1:
Closes #3686

The node confirmed `submit_participant_info` by reconstructing the attestation's creation time as `stored_expiry − DEFAULT_EXPIRATION_DURATION_SECONDS`. That only works when node and contract share the constant; under version skew (contract 3.12 = 7d, node 3.13 = 1d) the reconstructed age is ~6 days, always exceeding the 120s freshness bound, so every submission is treated as `NotExecuted` and retried indefinitely.

Instead, confirm the submission by checking that the stored attestation expiry **advanced** past the value observed before submitting.

- Read the stored expiry before submitting; confirm it increased afterward
- Drop the `expiry − constant` age reconstruction, `MAX_ATTESTATION_AGE`, and the shared-constant import
- Extract the pure decision into `attestation_expiry_advanced` + unit tests

Expiry only ever advances via the node's own submit for its key, so this never yields a false positive. It also holds under the verifier-rotation expiry cap in #3734 (where stored expiry ≠ `creation + constant`), which lists #3686 as a prerequisite.

## Design discussion

- [Options considered](https://github.com/near/mpc/issues/3686#issuecomment-4886131005)
- [Chosen option (expiry-advanced delta)](https://github.com/near/mpc/issues/3686#issuecomment-4886131659)


crates/node/src/indexer/tx_sender.rs
```diff
@@ -338,6 +372,11 @@ async fn ensure_send_transaction(
         public_key: Ed25519PublicKey::from(&tx_signer.public_key()),
         method,
     };
+    // Baseline for the submit_participant_info confirmation: the attestation expiry stored
+    // before we submit. Captured before submit_tx so a successful submit is detected as an
+    // advance in observe_tx_result.
+    let pre_submit_expiry = read_pre_submit_attestation_expiry(&indexer_state, &request).await;
+
```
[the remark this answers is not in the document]
barakeinav1: Thanks — the mechanism is accurate (pre-read error → we still submit, then `observe_tx_result` returns `Unknown` → retry), but this is intentional and we'd prefer to keep it rather than skip the send:

- **Idempotent and bounded, not a storm.** `submit_participant_info` is keyed by `tls_public_key`, which is owned by a single account (`assert_caller_is_signer`), so a re-submit overwrites rather than duplicates, and these retries use exponential backoff. A pre-read failure is transient and rare — unlike the deterministic version-skew loop this PR fixes.
- **Skipping the send would hurt liveness.** The pre-read is only for *confirmation*; the submit itself is the goal. If we skip submitting whenever the baseline read fails (i.e. during an RPC/view-client hiccup), we stop refreshing the attestation and risk it expiring and the node being evicted — the exact outage-liveness concern behind #3734. Submitting while temporarily unable to confirm is the safer trade: worst case is a few backed-off re-submits, never a missed refresh.

So we'll leave it as-is: always submit, and treat an unreadable baseline as "can't confirm → retry".

barakeinav1:
Thanks for the review! Fixed the doc nit; the other three are intentional non-changes:

> Doc comment on `attestation_expiry_advanced` is slightly heavier than needed: the "None means nothing was stored before, so an entry existing now means it landed" clause is somewhat misleading, since in practice `None` can also arise from a stored `Mock` in tests.

✅ Fixed (f89aaf12) — tightened to the pure invariant: returns `true` iff `stored_expiry > pre_submit_expiry`, or there is no baseline (`None`).

> Passing `anyhow::Result<Option<u64>>` as a function parameter is awkward ... Unwrapping the `Result` at the call site in `ensure_send_transaction` and passing `Option<u64>` would keep the signature narrower ... Optional.

Leaving as-is. Narrowing to `Option<u64>` would need either a magic sentinel (`u64::MAX`) to force a retry on a read error, or moving the error branch into `ensure_send_transaction` — which returns a tuple, not a `Result`, so it can't `?`. The current threading mirrors how `observe_tx_result` already handles its own post-submit read via `?`, and keeps the "submit anyway, surface an unreadable baseline as can't-confirm" behavior in one place (see the reasoning in the Copilot thread).

> `stored_expiry` is computed only to feed the `tracing::info!` call and duplicates the discrimination already performed inside `submitted_attestation_landed`.

Leaving as-is. It's a single extra match producing an `Option<u64>` log field; the only "drift" risk is in an informational log line, and keeping the decision function returning a plain `bool` reads more clearly than returning a struct just to carry the observed expiry.

> there's no direct test for the Dstack branch of `submitted_attestation_landed` (matching kinds + advanced/unchanged expiry).

Leaving as-is for now. The Dstack decision is already covered through `attestation_expiry_advanced` (the delegate); a direct test would require constructing a full `VerifiedDstackAttestation` (image/launcher hashes + measurements) to exercise a one-line delegation. Noted as a possible follow-up if that matcher ever grows beyond delegating.

crates/node/src/indexer/tx_sender.rs
```diff
@@ -183,10 +181,74 @@ async fn submit_tx(
     })
 }
 
+/// Reads the attestation expiry currently stored for a `submit_participant_info` request, to
+/// use as the pre-submit baseline for the advanced-expiry confirmation in [`observe_tx_result`].
+/// Must be called before submitting. Returns `Ok(None)` for other request types or when no
+/// Dstack attestation is stored yet.
+async fn read_pre_submit_attestation_expiry(
```
kevindeforth: Why should this function only be called before submitting an attestation? 
Couldn't it also be useful in other situations?
Why is the "pre-submit-baseline" important?

This comment explains what the code does and where this method is called, but not _why_ it is done. This goes against our [engineering standards]( https://github.com/near/mpc/blob/main/docs/engineering-standards.md#write-helpful-code-comments) for useful code comments.

crates/node/src/indexer/tx_sender.rs
```diff
@@ -183,10 +181,74 @@ async fn submit_tx(
...
+}
+
 /// Confirms whether the intended effect of the transaction request has been observed on chain.
+///
+/// `pre_submit_expiry` is the attestation expiry read *before* submitting, used by the
+/// `submit_participant_info` confirmation (see [`read_pre_submit_attestation_expiry`]).
 async fn observe_tx_result(
     indexer_state: Arc<IndexerState>,
     request: &ChainSendTransactionRequest,
+    pre_submit_expiry: anyhow::Result<Option<u64>>,
```
kevindeforth: This is a generic function that handles multiple transaction requests. The `pre_submit_expiry` argument is only relevant for a single transaction type and can be added to the sub-type of the `ChainSendTransactionRequest` enum.

crates/node/src/indexer/tx_sender.rs
```diff
@@ -338,6 +373,11 @@ async fn ensure_send_transaction(
         public_key: Ed25519PublicKey::from(&tx_signer.public_key()),
         method,
     };
+    // Baseline for the submit_participant_info confirmation: the attestation expiry stored
+    // before we submit. Captured before submit_tx so a successful submit is detected as an
+    // advance in observe_tx_result.
+    let pre_submit_expiry = read_pre_submit_attestation_expiry(&indexer_state, &request).await;
```
kevindeforth: We shouldn't call this for all transactions, as its result is only used for a single one.
However, I would question if this is truly the correct place for this call. Instead, I think the previous expiration timestamp should be included in the `request` and passed to the `ensure_send_transaction` method (e.g. probably in the `submit_remote_attestation` level or even before that.)

kevindeforth (commented):
Thank you, requesting changes for now:
- querying the tee expiration status for all transactions is unnecessary overhead. I also think it's the wrong place, that information should be retrieved once, possibly before we call `submit_remote_attestation`.
- code comments are not very useful, please take a look at our [engineering standards]( https://github.com/near/mpc/blob/main/docs/engineering-standards.md#write-helpful-code-comments).

crates/node/src/indexer/tx_sender.rs
```diff
@@ -183,10 +181,74 @@ async fn submit_tx(
...
+
+/// Confirms a `submit_participant_info` landed by checking the stored expiry advanced past the
+/// pre-submit baseline: a successful submit re-stamps expiry forward, a failed one leaves it
+/// unchanged, and expiry only advances via our own submit for this key — so this never yields a
+/// false positive. Returns `true` iff `stored_expiry > pre_submit_expiry`, or there is no baseline
+/// (`None`).
+///
+/// Avoids reconstructing creation time as `expiry - constant`, which breaks under node/contract
+/// version skew and the verifier-rotation expiry cap (stored expiry is not `creation + constant`).
+// TODO(#1639): confirm via a creation timestamp read from the certificate itself.
```
pbeza: I think we could kill two birds with one stone by addressing #1639 here as well, but it’s not a blocker.

crates/node/src/indexer/tx_sender.rs
```diff
@@ -183,10 +181,74 @@ async fn submit_tx(
...
+fn attestation_expiry_advanced(pre_submit_expiry: Option<u64>, stored_expiry: u64) -> bool {
```
pbeza: I’m not sure it’s worth keeping this as a standalone function rather than just inlining it. I don’t think test coverage would suffer.

crates/node/src/indexer/tx_sender.rs
barakeinav1: Sure, will update the documentation.

crates/node/src/indexer/tx_sender.rs
barakeinav1: Agreed, will move it off the generic function. Heads up: the request enum is `#[serde(untagged)]` and serializes into the on-chain args, so I'll keep it as a `#[serde(skip)]`/internal field so it doesn't leak into the call. Is this what you had in mind?

crates/node/src/indexer/tx_sender.rs
barakeinav1: Agreed, will read it once before the retry loop instead of per-attempt. Only catch: that layer only has a `TransactionSender`, so I'll plumb in a contract-read handle to fetch it once and pass it down. Is this what you had in mind?

crates/node/src/indexer/tx_sender.rs
```diff
@@ -183,10 +181,74 @@ async fn submit_tx(
...
+        None => true,
+    }
+}
+
+/// Whether the attestation we submitted is the one now stored on chain. Dstack is confirmed via
+/// [`attestation_expiry_advanced`]; Mock (tests) by direct equality. Mismatched kinds never match.
+fn submitted_attestation_landed(
+    pre_submit_expiry: Option<u64>,
+    stored: &VerifiedAttestation,
+    submitted: &Attestation,
```
pbeza, line 202: It's a bit confusing that `submitted` is only used in the `Mock` arm. Intuitively I'd expect us to compare `submitted` against `stored`, since _the whole point of `ensure_send_transaction` is just to confirm the submitted args landed on-chain_. The Dstack arm can't do that directly: `submitted` (`DstackAttestation`) and `stored` (`VerifiedDstackAttestation`) are different types, and the stored form keeps no per-submission identity to match on (like creation time). Also, all the expiry math feels a bit undocumented, and it’s not clear why it’s needed without more context. Preexisting issue though, so not a blocker

pbeza (changes requested):
There’s a corner case we don’t handle properly, and I don’t think we document it anywhere: if we _shrink_ the contract’s attestation expiry window (`DEFAULT_EXPIRATION_DURATION_SECONDS`), this check breaks:

https://github.com/near/mpc/blob/f89aaf12e4df495fd5908cbeae46127b7ecdc084/crates/node/src/indexer/tx_sender.rs#L223

Right now, to confirm that a submit landed, we check that the stored expiry is later than it was before the submit. That works because every successful submit sets `expiry = block_now + DEFAULT_EXPIRATION_DURATION_SECONDS`, so a fresh submit should normally push the expiry forward, assuming the duration stays the same.

But if a contract upgrade makes `DEFAULT_EXPIRATION_DURATION_SECONDS` smaller, for example 7 days → 1 day, the next submit can set an expiry that is earlier than the old one already stored on-chain. So the submit actually succeeds, but the expiry goes down instead of up, and our check thinks it failed and keeps retrying forever.

So basically the same retry storm as #3686, but triggered by shrinking the expiry window instead of a node/contract version mismatch.

barakeinav1:
> But if a contract upgrade makes `DEFAULT_EXPIRATION_DURATION_SECONDS` smaller, for example 7 days → 1 day, the next submit can set an expiry that is earlier than the old one already stored on-chain. So the submit actually succeeds, but the expiry goes down instead of up, and our check thinks it failed and keeps retrying forever.

In this case, the first submission would succeed and shorten the expiry, but the node will think it failed and re-try.
If we **re-read** the submission (the new one that landed) and re-try, then we will succeed and the node will also also think he succeeded. 

so we will get only one false negative. which is fine by me.
but this does mean we need to re-read the submission before each re-try,


another option, is to still read once, but then ask the operators to just re-start the node.
but I don't like this, since, we have a future PR ([3734](https://github.com/near/mpc/issues/3734) ) that will cause the above scenario to have once each verifier upgrade

crates/node/src/indexer/tx_sender.rs
barakeinav1: Correction after more thought (and see my reply to Patrick re the expiry constant: https://github.com/near/mpc/pull/3736#issuecomment-4900361353): we actually need to re-read the baseline **before each retry**, not once. If a future upgrade lowers the expiration constant, a submit can land with a *lower* expiry than the old stored value; capturing once would then loop for days, while re-reading self-heals after a single false negative. Still keeping it at the submit layer and in the request — just per-attempt.

crates/node/src/indexer/tx_sender.rs
barakeinav1: I'd keep it — inlining means the expiry-advance cases can only be tested by constructing full Dstack attestation DTOs, whereas as a standalone fn they're tested with plain timestamps. So coverage would get more painful, not free.

crates/node/src/indexer/tx_sender.rs
barakeinav1: Update: went with `new_expiration != previous_expiration` instead of `>`, so correctness no longer depends on re-reading per attempt — a landed submit changes the expiry either way. Accepting the narrow false-positive corner (a rotation cap changing expiry mid-submit reads as landed); it self-corrects on the hourly resubmit. The stale-attestation retry-loop issue is tracked separately in #3746.

crates/node/src/indexer/tx_sender.rs
barakeinav1: Went with a dedicated `confirm_participant_info_submission` fn instead of a field on the enum variant — the request is `#[serde(untagged)]` and serialized as the on-chain args, so a field there would leak into the call. This keeps the baseline off the generic `observe_tx_result` signature (its actual concern) without touching serialization.

crates/node/src/indexer/tx_sender.rs
barakeinav1: I think this change is too big, let's do it separately.

crates/node/src/indexer/tx_sender.rs
```diff
@@ -409,8 +426,18 @@ async fn ensure_send_transaction(
...
+            confirm_participant_info_submission(
+                &indexer_state,
+                &args.tls_public_key,
+                &args.proposed_participant_attestation,
+                pre_submit_expiry,
+            )
+            .await
+        }
+        _ => observe_tx_result(indexer_state.clone(), &request).await,
+    };
```
kevindeforth: This looks bad. The existing pattern should be generic enough to be able to handle this, why can't we add it to the enum variant and call `observe_tx_result`?

barakeinav1: Done in a0572418 — the baseline now lives on the `SubmitParticipantInfo` request variant (`#[serde(flatten)]` args + `#[serde(skip)] pre_submit_expiry`, so the on-chain args are unchanged), read once per round by the caller. `observe_tx_result` / `ensure_send_transaction` are generic again — no per-transaction logic in there.

crates/node/src/indexer/tx_sender.rs
barakeinav1: Clarified in the `submitted_attestation_landed` doc (1b4c5f89): Mock carries a full identity so it's matched directly, while a Dstack entry has no stored per-submission identity — so it's confirmed via its expiry changing from the pre-submit baseline. So `submitted` is genuinely only needed for the Mock arm.

crates/node/src/tee/remote_attestation.rs
```diff
@@ -145,6 +152,7 @@ pub async fn periodic_attestation_submission<T: TransactionSender + Clone, I: Ti
     account_public_key: Ed25519PublicKey,
     allowed_image_hashes_in_contract: watch::Receiver<Vec<NodeImageHash>>,
     allowed_launcher_compose_hashes_in_contract: watch::Receiver<Vec<LauncherDockerComposeHash>>,
+    attestation_reader: std::sync::Arc<dyn crate::indexer::ReadAttestationExpiry>,
     mut interval_ticker: I,
```
kevindeforth: nit: we can clean this up by making a struct for the args.

crates/node/src/tee/remote_attestation.rs
```diff
@@ -180,13 +188,24 @@ pub async fn periodic_attestation_submission<T: TransactionSender + Clone, I: Ti
...
             allowed_launcher_compose_hashes_in_contract.borrow().clone();
+        let pre_submit_expiry = match attestation_reader
+            .read_stored_dstack_expiry(&tls_public_key)
+            .await
+        {
+            Ok(baseline) => baseline, // None just means nothing stored yet (e.g. first submit)
+            Err(error) => {
+                tracing::warn!(%error, "could not read pre-submit attestation baseline; skipping this round");
+                continue; // next tick. Do NOT submit with an unknown baseline.
+            }
```
kevindeforth, line 220: Here, I don't think we want to restart the loop. In case the contract read method is broken, or deserialization fails for unknown reasons, this would prevent the node from submitting its attestation.
I would suggest to log the error, but set `pre_submit_expiry` to `None`

crates/node/src/tee/remote_attestation.rs
```diff
@@ -286,13 +306,25 @@ pub async fn monitor_attestation_removal<T: TransactionSender + Clone>(
...
                 allowed_launcher_compose_hashes_in_contract.borrow().clone();
+            let pre_submit_expiry = match attestation_reader
+                .read_stored_dstack_expiry(&tls_public_key)
+                .await
+            {
+                Ok(baseline) => baseline, // None just means nothing stored yet (e.g. first submit)
+                Err(error) => {
+                    tracing::warn!(%error, "could not read pre-submit attestation baseline; skipping this round");
+                    was_available = is_available;
+                    continue; // next event. Do NOT submit with an unknown baseline.
```
kevindeforth: same here

kevindeforth (dismissed):
thank you! Just a nit and a comment regarding control flow

crates/node/src/tee/remote_attestation.rs
barakeinav1: This will re-create a submission storm again, but probably better than failing to submit at all.

crates/node/src/tee/remote_attestation.rs
barakeinav1: This will re-create a submission storm again, but probably better than failing to submit at all.

crates/node/src/tee/remote_attestation.rs
barakeinav1: Good catch — fixed in bae9ad89 (well, 67bfeaa2): on a read error we now submit anyway (baseline `None`) instead of skipping, so a broken read can't block submission. Added a test that submissions still happen when the read fails.

crates/node/src/tee/remote_attestation.rs
barakeinav1: Same fix applied here too (67bfeaa2).

crates/node/src/tee/remote_attestation.rs
barakeinav1: Done in 67bfeaa2 — bundled the shared args into an `AttestationSubmitter` struct, dropped the `too_many_arguments`.

kevindeforth (commented):
Thank you, this is very nice!

kevindeforth:
I took the liberty to resolve the merge conflict. Would be nice to get this merged today.

kevindeforth (approved):
Thank you!

pbeza (approved):

Issue #3751, opened 2026-07-08
remove legacy struct from mpc node and tee context


kevindeforth:
### Background

For #3735 we changed the contract interface types. For backwards compatibility, the node and TEE context still hold references to the legacy struct.

We can remove those with the next contract upgrade.

### User Story

As a dev, I like a clean code base

### Acceptance Criteria

Remove the legacy structs labeled with TODO(#3751)

Issue #3226, opened 2026-05-13
chore(contract): inconsistent ordering across allowed_docker_hashes, allowed_launcher_hashes  / allowed_os_measurements view methods


barakeinav1:
### Background

The contract has four public view methods that return vectors of hashes / measurements. Three return in insertion order (oldest first); one explicitly reverses to return newest first. Only one of the four has any doc comment about order at all.

| Method | Returns | Order |
|---|---|---|
| \`allowed_docker_image_hashes()\` | \`Vec<NodeImageHash>\` | **newest-first** (\`.reverse()\` at \`lib.rs:1831\`; documented) |
| \`allowed_launcher_image_hashes()\` | \`Vec<LauncherImageHash>\` | oldest-first (insertion order; undocumented) |
| \`allowed_launcher_compose_hashes()\` | \`Vec<LauncherDockerComposeHash>\` | oldest-first (insertion order; undocumented) |
| \`allowed_os_measurements()\` | \`Vec<ContractExpectedMeasurements>\` | oldest-first (insertion order; undocumented) |

Real-world impact:

- Operator docs in PR #3222 documented "the latest is first in each returned vector" — correct only for \`allowed_docker_image_hashes\`. Caught and corrected during Copilot review.
- \`scripts/fetch-allowed-launcher-hashes.sh\` in PR #3223 takes \`.[0]\` of both \`allowed_launcher_image_hashes\` and \`allowed_docker_image_hashes\` — wrong for the launcher arm.
- Anyone scripting against these views with a "latest first" mental model silently picks the wrong element for 3 of the 4.

### Acceptance Criteria

- [ ] All four hash-returning view methods return the same documented order.
- [ ] Each method has a doc comment that states the order explicitly.
- [ ] Update PR #3223's fetch script if its \`.[0]\` semantics change.

### Resources & Additional Notes

- Method definitions: [\`crates/contract/src/lib.rs:1821-1841\`](https://github.com/near/mpc/blob/main/crates/contract/src/lib.rs#L1821-L1841)
- Underlying storage iterators: [\`crates/contract/src/tee/proposal.rs:332-342\`](https://github.com/near/mpc/blob/main/crates/contract/src/tee/proposal.rs#L332-L342) (\`all_compose_hashes\`, \`launcher_hashes\`)
- The only docstring promising order: [\`lib.rs:1820\`](https://github.com/near/mpc/blob/main/crates/contract/src/lib.rs#L1820)
[...]

PR #3752, opened 2026-07-08
feat(metrics): expose the TEE image-hash allowlist expiry metrics

### LINKED ISSUE #3735: Create alert when current version is in the grace period and will expire soon
### Background

When the new mpc-node version is voted in there is a grace period (7 days) that current version is accepted in attestations.

We should add alerts that will fire when current version is expiring soon.

### User Story

_No response_

### Acceptance Criteria

We have alerts for current version expiring soon

_No response_

kevindeforth:
Closes #3735

The `allowed_docker_image_hashes` view now returns `{ image_hash, expiry_timestamp_seconds }` entries (descending expiry, newest first).

#### New metrics

| Metric | Meaning |
|---|---|
| `mpc_own_image_hash_allowed` | 1 if the node's image hash is on the contract allowlist, 0 if evicted |
| `mpc_own_image_hash_is_most_recent` | 1 if the node runs the newest allowed image hash |
| `mpc_own_image_hash_expiry_timestamp_seconds` | Block time (unix seconds) at which the node's hash is evicted; `-1` if no expiry applies (newest hash, or contract not yet upgraded), `0` if not allowed |
| `mpc_indexer_latest_block_timestamp_seconds` | Timestamp of the latest indexed block — the chain "now", so remaining grace time is `expiry − block_ts` in Grafana |

#### Migration strategy

Node and tee-context are the only consumers of the modified endpoint. They can parse both the previous and the new return values. After we upgrade the contract, we can remove the logic related to the legacy contract types (c.f. #3751).

Please take note that the metric `mpc_own_image_hash_expiry_timestamp_seconds` will only be useful once we upgrade the contract.


kevindeforth:
### Blocking

#### `crates/node/src/tee/image_expiry_metrics.rs:32`

> Doc comment reads "Expects allowed_images to be in increasing order of expiration timestamp", but every real caller passes the list newest-first (as the contract's view returns it after entries.reverse() in crates/contract/src/lib.rs:2101). The function itself only relies on .first() being the newest entry (the local is even named newest on image_expiry_metrics.rs:44). The current wording is the opposite of the invariant callers rely on and will mislead the next reader. Please rewrite it, e.g.: "Expects allowed_images to be newest-first (as returned by the contract view).".

Fixed — now reads "Expects `allowed_images` to be newest-first, as returned by the contract view."

#### `crates/node/src/tee/allowed_image_hashes_watcher.rs:172-192` / `image_expiry_metrics.rs:38-40`

> When the contract-reported list becomes empty, the watcher logs a warning and returns early before calling update_own_image_hash_gauges, and the defensive is_empty() check inside infer_image_status returns None (so gauges are not touched either). Net effect: mpc_own_image_hash_allowed remains at its previous value (potentially still 1) even though the node's image is objectively no longer allowed. For a safety-oriented metric this is a footgun for on-call. Either set ALLOWED=0, IS_MOST_RECENT=0, EXPIRY=0 on the empty-list branch (documented as "no images allowed → node's image is not allowed"), or explicitly document why staleness is acceptable. Also: since the caller short-circuits, the is_empty() guard inside infer_image_status becomes dead — pick one place to handle it.

Fixed. The gauge update now runs before the `NonEmptyVec` guard in the watcher, so an empty list sets `allowed=0, is_most_recent=0, expiry=0` (the disk write is still skipped). The `is_empty` guard and the `Option` return were removed from `infer_image_status` — an empty list falls through the existing `find → None` arm to `(false, false, 0)`, so it's handled in exactly one place. Test case updated accordingly.

### Non-blocking

#### `crates/contract/src/tee/proposal.rs:163-166` / `time.rs:28-33`

> The new add_or_panic will env::panic_str inside a view call if next.added + tee_upgrade_deadline_duration overflows Duration. This is safe for the contract itself, but it turns any query of allowed_docker_image_hashes (indexer, tee-context, dashboards) into a hard failure. Since tee_upgrade_deadline_duration_seconds is governance-configurable, a bad governance value could brick the view. Note the existing cutoff_index handles this by logging and returning true for the position (proposal.rs:224-228). Consider mirroring that (e.g. treat overflow as expiry_timestamp_seconds: None) rather than panicking; if panic is intentional, add a comment explaining the invariant.

Good catch — switched to `checked_add`, mapping overflow to `expiry_timestamp_seconds: None` (a grace period that overflows the time range never ends, so "no expiry" is the honest value, consistent with `cutoff_index`'s tolerance). `add_or_panic` had no other callers and is deleted.

#### `crates/contract/src/tee/proposal.rs:185-186`

> Doc says "sorted by expiration timestamp in ascending order", but the last entry has expiry_timestamp_seconds: None. Consider "sorted oldest-first; the newest entry has no expiry" to avoid the interpretive ambiguity of "None in ascending order".

Fixed — reworded to "oldest first; the newest entry has no expiry".

#### `crates/node/src/indexer.rs:1049-1068` / `crates/tee-context/src/lib.rs:31-53`

> AllowedDockerImageHashesResponse has a legacy-shape unit test but no matching test for WithExpiry. Symmetric coverage would prevent the untagged-enum precedence from silently regressing. Same nit applies to crates/tee-context/src/lib.rs:31-53 (no direct unit test of into_entries for the WithExpiry branch).

Deliberate scope cut: the two shapes (array of objects vs array of strings) can't shadow each other under `#[serde(untagged)]`, so variant precedence can't silently regress, and the with-expiry shape is exercised end-to-end by the contract inprocess/sandbox tests. Only the legacy fallback is load-bearing during rollout, hence the single test. Both enums are TODO(#3751) for removal anyway.

#### `crates/node/src/tee/image_expiry_metrics.rs:42-44`

> .first().is_some_and(...) is defensive after the is_empty guard on line 38; allowed_images[0].image_hash == *current_image reads cleaner and matches the invariant already established locally.

Kept `.first()`: with the empty-list fix above the `is_empty` guard is gone, so `.first()` is now load-bearing, not defensive — and we avoid a panicking index.

#### `crates/node/src/tee/image_expiry_metrics.rs:9`

> /// Indicates if no expiration date is known. on the NO_EXPIRY const paraphrases the identifier. Per engineering-standards §Write helpful code comments, consider dropping it.

Dropped.

#### `crates/node/src/indexer/handler.rs:250-253`

> i64::try_from(u64_secs).unwrap_or(i64::MAX): u64::MAX / 1_000_000_000 ≈ 1.8e10 always fits in i64, so the fallback is unreachable. Fine as defensive code; consider lifting into a helper for consistency with the same pattern in MPC_OWN_IMAGE_HASH_EXPIRY_TIMESTAMP_SECONDS.

Leaving as-is per your "fine as defensive code" — a shared helper for two sites doesn't pay for itself.

#### `crates/near-mpc-contract-interface/src/types/tee.rs:44-45`

> The DTO docstring says "None if expiration date is not yet known", which is a weaker guarantee than the metric's "newest entry never expires". Consider tightening to "None iff this is the newest allowed hash — the contract does not yet know its successor.".

Keeping the weaker wording deliberately: it would be wrong as "iff" — clients normalize legacy (pre-upgrade) contract responses into this same type with `None` on *every* entry, so client-side `None` means "unknown", not "newest". The strict iff only holds for what the upgraded contract emits.

kevindeforth:
@claude review

crates/contract/src/tee/proposal.rs
```diff
@@ -159,19 +176,48 @@ pub struct AllowedMpcDockerImage {
...
-    fn valid_entries(&self, tee_upgrade_deadline_duration: Duration) -> Vec<AllowedMpcDockerImage> {
-        let current_time = Timestamp::now();
-        // get the index of the most recently enforced docker image
-        let cutoff_index = self
+impl StoredDockerImageHashes {
+    /// Returns the list of currently allowed docker image hashes, oldest first; the newest entry
+    /// has no expiry.
+    pub fn allowed_images(
+        &self,
+        tee_upgrade_deadline_duration: Duration,
```
SimonRastikian, line 190: Why do we not have this as a public constant?

crates/contract/src/tee/proposal.rs
```diff
@@ -159,19 +176,48 @@ pub struct AllowedMpcDockerImage {
...
+                    unreachable!("windows(2) always yields two-element slices")
+                };
+                make_dtos_docker_image_hash(prev, next, tee_upgrade_deadline_duration)
+            })
+            .collect();
+        res.push(dtos::AllowedMpcDockerImageHash {
+            image_hash: latest.image_hash,
+            expiry_timestamp_seconds: None,
+        });
+        res
```
SimonRastikian, line 214: This should be nicer and less code. What do you think?

```suggestion
valid.iter()
    .map_windows(|[prev, next]| {
        make_dtos_docker_image_hash(prev, next, tee_upgrade_deadline_duration)
    })
    .chain(std::iter::once(dtos::AllowedMpcDockerImageHash {
        image_hash: latest.image_hash,
        expiry_timestamp_seconds: None,
    }))
    .collect()
```

crates/contract/src/tee/proposal.rs
kevindeforth: because it is not a constant, but a configurable value

crates/contract/src/tee/proposal.rs
kevindeforth: I would love that, but I thought `map_windows` wasn't stable on our rust version. did you try to compile it?
_edit: confirmed as unstable https://doc.rust-lang.org/std/iter/struct.MapWindows.html_

SimonRastikian: Unfortunately no I hadn't compile it.
You can use instead zip but makes the line a bit less tidy:
```rust    
.zip(valid.iter().skip(1))
    .map(|(prev, next)| {
        make_dtos_docker_image_hash(prev, next, tee_upgrade_deadline_duration)
    })
```

crates/contract/src/snapshots/mpc_contract__tests__mpc_contract_borsh_schema_has_not_changed.snap
```diff
@@ -1130,6 +1106,16 @@ BorshSchemaContainer {
                 ],
             ),
         },
+        "StoredDockerImageHashes": Struct {
```
barakeinav1, line 1109: why did you change the name?

crates/contract/src/snapshots/mpc_contract__tests__mpc_contract_borsh_schema_has_not_changed.snap
```diff
@@ -1269,6 +1250,11 @@ BorshSchemaContainer {
             length_range: 0..=4294967295,
             elements: "LauncherDockerComposeHash",
         },
+        "Vec<WhitelistedMpcDockerImageHash>": Sequence {
```
barakeinav1, line 1253: why the name change?

barakeinav1:
Deploy ordering: this must go out **node-first, then contract**. New nodes parse both shapes, but a pre-#3752 node hitting the new view shape from #3751 can't deserialize it and its allowlist fetch breaks. Matches our usual order — just flagging.

barakeinav1:
Related: #3226 — our view methods return inconsistent order  (existing issue).
mentioning this in case it affect your design

crates/contract/src/snapshots/mpc_contract__tests__mpc_contract_borsh_schema_has_not_changed.snap
kevindeforth: To make the difference between this type and the interface type explicit.

crates/contract/src/snapshots/mpc_contract__tests__mpc_contract_borsh_schema_has_not_changed.snap
kevindeforth: Because the previous name was misleading. If a hash was allowed or not was dependent on ephemeral data. `Whitelisted` makes it similar to how we name things in the foreign chain transaction feature (where we distinguish between whitelisted and available foreign chains, here, we distinguish between whitelisted and allowed docker image hashes.

barakeinav1:
Opened #3761 against this branch with the consumer fixups for the new `allowed_docker_image_hashes` return shape — the two launcher localnet scripts (`jq select(. == $h)` → `select(.image_hash == $h)`, they'd otherwise never match the new objects) plus the return type in `contract/README.md` and the securing-mpc design doc. Merge it in or cherry-pick, whichever you prefer.

crates/node/src/indexer.rs
```diff
@@ -548,7 +570,31 @@ pub struct IndexerAPI<TransactionSender, ForeignChainPolicyReader> {
...
 #[expect(non_snake_case)]
 mod tests {
-    use super::{BlockHeight, REQUIRED_STABLE_POLLS, SyncProgress};
+    use super::{
+        AllowedDockerImageHashesResponse, BlockHeight, REQUIRED_STABLE_POLLS, SyncProgress,
+    };
+    use mpc_primitives::hash::NodeImageHash;
+
+    #[test]
+    fn allowed_docker_image_hashes_response__should_deserialize_legacy_bare_hashes() {
```
barakeinav1, line 590: nit: this covers the `Legacy` (bare-hash) shape — worth a symmetric test for the `WithExpiry` object shape too, since that's the primary post-upgrade path. The `submit_participant_info` e2e does exercise it end-to-end (nodes read the new-shape allowlist to validate + submit), but only indirectly — a parse regression there surfaces as "nodes never attested" rather than a clear deserialization failure. A one-liner keeps the signal direct:

```rust
#[test]
fn allowed_docker_image_hashes_response__should_deserialize_with_expiry_objects() {
    let json = r#"[
        { "image_hash": "1111111111111111111111111111111111111111111111111111111111111111", "expiry_timestamp_seconds": 42 },
        { "image_hash": "2222222222222222222222222222222222222222222222222222222222222222", "expiry_timestamp_seconds": null }
    ]"#;
    let response: AllowedDockerImageHashesResponse = serde_json::from_str(json).unwrap();
    assert_matches!(response, AllowedDockerImageHashesResponse::WithExpiry(entries) if entries.len() == 2);
}
```

crates/contract/src/tee/proposal.rs
```diff
@@ -147,9 +148,25 @@ impl LauncherHashVotes {
     all(feature = "abi", not(target_arch = "wasm32")),
     derive(borsh::BorshSchema)
 )]
-pub struct AllowedMpcDockerImage {
-    pub(crate) image_hash: NodeImageHash,
-    pub(crate) added: Timestamp,
+struct WhitelistedMpcDockerImageHash {
+    image_hash: NodeImageHash,
+    added: Timestamp,
```
SimonRastikian: Nit: I would rename this to as currently it could also be a boolean.
```suggestion
    added_at: Timestamp,
```

crates/node/src/tee/image_expiry_metrics.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    match allowed_images
+        .iter()
+        .find(|entry| entry.image_hash == *current_image)
+    {
+        Some(entry) => match entry.expiry_timestamp_seconds {
+            Some(expires_at) => (
+                true,
+                is_most_recent,
+                i64::try_from(expires_at).unwrap_or(i64::MAX),
```
SimonRastikian, line 47: I know it's a bit "tire par les cheveux" but expires_at is a u64 but the outcome is i64.

SimonRastikian (dismissed):
This looks good to me. Thanks!

barakeinav1 (dismissed):
approved, one nit about the unit test, but I think the e2e do cover it

crates/node/src/tee/image_expiry_metrics.rs
kevindeforth: Yeah I know, it's not ideal :D 
But I don't think grafana supports unsigned integers.

crates/node/src/indexer.rs
kevindeforth: good point, addressed in https://github.com/near/mpc/pull/3752/commits/e123ef8d219c7b3547f18e57b60a09fb7282610b

crates/contract/src/tee/proposal.rs
kevindeforth: included in https://github.com/near/mpc/pull/3752/commits/e123ef8d219c7b3547f18e57b60a09fb7282610b

barakeinav1 (approved):

SimonRastikian (approved):
Thanks much appreciated my friend

PR #3754, opened 2026-07-08
feat(foreign-tx): sui support in contract interface

frolvanya:
## Summary

Adds Sui as a supported foreign chain for the contract's `verify_foreign_transaction`
flow — the **contract-interface (DTO) half** of Sui support.

This is one of a two-PR split (see the follow-up inspector PR for the node side), keeping
the contract-interface changes reviewable independently from the inspector logic — same
split as Aptos.

## What's included

- **New wire types** in `near-mpc-contract-interface`
  (`crates/near-mpc-contract-interface/src/types/foreign_chain.rs`):
  - `SuiTxId` — 32-byte transaction digest (Blake2b-256 of the signed tx data). Sui APIs
    display it base58-encoded; the DTO carries raw bytes (hex in JSON) like every other
    chain's tx id, and the inspector converts at the RPC boundary
  - `SuiAddress` — 32-byte address / object id (hex)
  - `SuiRpcRequest { tx_id, finality, extractors }`
  - `SuiFinality::Checkpointed` — Sui has no reorgs; a transaction is final once included
    in a committee-certified checkpoint, so a single finality level
  - `SuiExtractor::Event { event_index } = 1` — discriminant `1`, mirroring the log/event
    extractor on EVM/Starknet/TON/Aptos; discriminant `0` is left as the conventional
    block-hash slot
  - `SuiEvent { package_id, transaction_module, sender, type_tag, bcs }` — carries the
    event's **BCS bytes** rather than the API's `parsedJson` rendering: BCS is the
    protocol-certified encoding (covered by the effects' `eventsDigest`), while
    `parsedJson` is produced by version-dependent node code and could differ across
    providers running different node versions, breaking the fan-out quorum. `type_tag`
    is normalized by the inspector to canonical long-form addresses (`0x` + 64 hex)
  - `SuiExtractedValue::Event(SuiEvent)`
  - variant wiring: `ForeignChainRpcRequest::Sui`, `ExtractedValue::SuiExtractedValue`,
    `ForeignChain::Sui`, and the `ForeignChainRpcRequest::chain()` arm — all appended
    last, preserving existing borsh discriminants
- **No `crates/contract/src` logic changes** — the contract is generic over
  `ForeignChain`, so it picks up Sui transitively through the DTO enums (same as the TON
  and Aptos contract PRs).
- **Snapshots**:
  - regenerated `crates/contract/tests/snapshots/abi__abi_has_not_changed.snap` — now
    includes the Sui types and the `ForeignChain__Sui` variant (additive only)
  - regenerated `foreign_chains_metadata` borsh-schema snapshot — additive only:
    variant `(12, "Sui", "ForeignChain__Sui")` (this snapshot test postdates the Aptos
    PRs, hence a change here where #3507 had none)
  - new `foreign_tx_sign_payload_v1_sui__should_have_consistent_hash.snap` — pins the V1
    Sui sign-payload hash
- **Sandbox tests**: `sui_request()` / `sui_extracted_values()` fixtures in
  `tests/sandbox/common.rs`, and `#[case::sui(...)]` added to the three parameterized
  `verify_foreign_transaction__*` tests in `tests/sandbox/foreign_chain_request.rs`

## Testing

- `near-mpc-contract-interface` unit tests (consistent-hash + chain mapping) pass.
- `test_abi_has_not_changed` and `foreign_chains_metadata_borsh_schema__should_not_change`
  pass against the regenerated snapshots.
- Contract WASM size: 1,516,004 bytes vs the 1,520,000-byte limit — **only ~4 KB of
  headroom left** (main is at 1,508,697; the Sui DTOs add ~7.3 KB). Within the limit, but
  the next DTO addition will likely need the #3475 cleanup or a limit bump.

## Out of scope

- The node-side inspector, Sui RPC client, node wiring, config-tester golden vectors, and
  localnet docs — see the follow-up inspector PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2474,9 +2474,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-epoch"
-version = "0.9.18"
+version = "0.9.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b82ac4a3c2ca9c3460964f020e1402edd5753411d7737aa39c3714ad1b5420e"
+checksum = "2d6914041f254d6e9176c01941b21115dcfb7089e55135a35411081bd106ef3f"
 dependencies = [
  "crossbeam-utils",
 ]
--- a/crates/contract/src/snapshots/mpc_contract__foreign_chains_metadata__tests__foreign_chains_metadata_borsh_schema__should_not_change.snap
+++ b/crates/contract/src/snapshots/mpc_contract__foreign_chains_metadata__tests__foreign_chains_metadata_borsh_schema__should_not_change.snap
@@ -90,6 +90,11 @@ BorshSchemaContainer {
                     "Aptos",
                     "ForeignChain__Aptos",
                 ),
+                (
+                    12,
+                    "Sui",
+                    "ForeignChain__Sui",
+                ),
             ],
         },
         "ForeignChainRpcWhitelist": Struct {
@@ -137,6 +142,9 @@ BorshSchemaContainer {
             fields: Empty,
         },
         "ForeignChain__Starknet": Struct {
+            fields: Empty,
+        },
+        "ForeignChain__Sui": Struct {
             fields: Empty,
         },
         "ForeignChain__Ton": Struct {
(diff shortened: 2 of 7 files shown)


anodar (approved):
Thank you!

gilcu3 (approved):
Thank you!

PR #3755, opened 2026-07-08
feat(foreign-tx): sui inspector

frolvanya:
## Summary

Implements the **node-side** Sui foreign-chain inspector. Given a Sui transaction digest,
it fetches the transaction from the configured Sui providers over **gRPC
(`sui.rpc.v2.LedgerService`)**, verifies it is checkpointed and succeeded, and extracts
the event the MPC network signs over (for Sui → NEAR bridging via the omni-bridge Sui
locker).

gRPC is used from the start rather than JSON-RPC: Sui deprecated JSON-RPC with a July
2026 deactivation, and the deactivation is already underway — as of 2026-07-08,
`fullnode.testnet.sui.io` (sui-node 1.75.0) no longer serves JSON-RPC at all, while
serving gRPC on the same endpoint. gRPC is the fullnode-native replacement (every
fullnode operator is a provider) and returns the same canonical bytes: the event BCS
payload was verified byte-for-byte identical between both transports on live mainnet.

## Scope

Builds on the Sui DTOs from the contract-interface PR #3754 (merged), so the base is
`main`. To keep this PR to the inspector and node wiring, the tests, the Sui config-tester
support, and the localnet docs are split into the **stacked PR #3791** (based on this
branch). This PR touches:

- `crates/foreign-chain-rpc-interfaces/**`
- `crates/foreign-chain-inspector/**` (the `SuiInspector`; its integration tests live in #3791)
- `crates/node/**`, `crates/node-config/**`
- `Cargo.toml`, `Cargo.lock`

## What's included

- **`foreign-chain-rpc-interfaces/src/sui.rs`** — a `SuiRpcClient` trait (I/O boundary,
  per the engineering standards) and `GrpcSuiClient` implementing it over the official
  `sui-rpc` crate (Mysten's published bindings; its own node server consumes the same
  crate). Requests carry a field mask selecting exactly the verified fields (digest,
  checkpoint, execution status, and the five event fields) — the events' `json`
  rendering is never transferred. Dependency impact is small:
  tonic/prost 0.14 were already in the lock via `gcloud-sdk`; `sui-rpc` adds six new
  packages.
- **`foreign-chain-inspector/src/sui/`** — `SuiInspector` implementing
  `ForeignChainInspector`: fetch by digest → digest-echo check → checkpoint (finality)
  check → execution-status check → extract the event at `event_index`.
- **Node wiring** — the `sui` fan-out slot with a dedicated `new_sui_inspector` builder
  (header auth maps to gRPC metadata; the shared jsonrpsee path does not apply), the
  `Sui` arm in `execute_foreign_chain_request`, the `sui` node-config field (with
  header-auth parse tests; config validation rejects path/query auth for Sui, since
  gRPC carries credentials in request metadata and a token substituted into the URL
  would silently never be sent), the `/debug` provider count, and the contract-interface
  conversions.

Split into the stacked **#3791**: the mocked-client integration tests, the Sui
config-tester support (permanent-anchor check + live verification), and the localnet
args/config templates/docs.

## Key design decisions

- **Finality = `checkpoint` field presence.** The node sets `checkpoint` only once the
  transaction is included in a committee-certified checkpoint ("included and hence
  finalized"); Sui has no reorgs. Absent → transient `NotFinalized`, so the fan-out
  retries until checkpointed.
- **The signed payload carries the event's BCS bytes.** gRPC delivers them natively
  (`Event.contents`), protobuf-typed — no base64/base58 re-encoding layer at all, and no
  reliance on `parsedJson`, whose rendering is node-version-dependent. Verified
  byte-identical to the JSON-RPC `bcs` field on live mainnet, so payloads are
  transport-independent.
- **Type tags normalized to canonical long form** (`0x` + 64 lowercase hex). gRPC
  already emits the long form (verified live for framework events), making the
  normalization a cross-version safety net rather than an active rewrite; it also keeps
  signed payloads identical to what a JSON-RPC-based extraction would produce.
- **Error classification via gRPC status codes** (fail-closed, per the Aptos review
  policy): `NotFound` → non-transient `TransactionNotFound` (a clean status code — no
  message sniffing, unlike JSON-RPC where not-found shared `-32602` with
  invalid-params); `DeadlineExceeded`/`Unavailable`/`ResourceExhausted`/`Internal`/
  `Unknown`/`Cancelled`/`Aborted` → transient; every other code (auth failures, bad
  requests, unimplemented method) → non-transient `RpcRequestRejected`. Note that
  "not found" also covers a provider that pruned the transaction.
- **Malicious-provider hardening**: digest-echo check (length-capped before base58
  decoding — `bs58`'s decode is superlinear), per-field malformed-response checks on
  the proto's optional fields, and a consistency check that the type name shipped
  inside the BCS message (when present) agrees with the event type being signed —
  compared after normalization, so differing address renderings don't reject honest
  responses.

## Trust model

Same as the existing inspectors (EVM/Bitcoin/Starknet/Aptos): trust the whitelisted RPC
providers plus the fan-out quorum (all providers must return the same extracted value,
else `InspectorResponseMismatch` blocks signing). No Sui light-client / checkpoint-
signature verification — out of scope, consistent with the other chains.

## Provider note

Mysten's fullnodes (mainnet, testnet, archive) serve gRPC unauthenticated today, and
testnet serves *only* gRPC. Commercial providers expose gRPC per Mysten's provider list;
some free JSON-RPC gateways do not proxy gRPC yet, so operators must pick gRPC-capable
endpoints. The config-tester (stacked in #3791) surfaces a non-gRPC provider immediately.

## Testing

- Inspector unit tests: status-code classification (not-found, transient hiccups,
  deterministic rejections), digest echo (mismatch / non-base58 / wrong length /
  oversized), type-tag normalization (framework short forms, nested generics,
  identifiers containing `0x`), address parsing.
- `cargo fmt`, `cargo clippy --all-targets -- -D warnings` pass workspace-wide.

The mocked-client integration tests (full extract flow), the config-tester unit tests,
and the live `#[ignore]`d archive test are in the stacked **#3791**.


crates/foreign-chain-inspector/src/sui/inspector.rs
```diff
@@ -0,0 +1,633 @@
...
+
+fn decode_event_bcs(event: &SuiEventResponse) -> Result<Vec<u8>, ForeignChainInspectionError> {
+    match event.bcs_encoding.as_deref() {
+        Some("base64") => base64::engine::general_purpose::STANDARD
+            .decode(&event.bcs)
+            .map_err(|e| {
+                ForeignChainInspectionError::MalformedRpcResponse(format!(
+                    "non-base64 event bcs: {e}"
+                ))
+            }),
```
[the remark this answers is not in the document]
frolvanya: Added restriction in 64cb875d42a55365135856102dcaa85a0f61e376

crates/foreign-chain-config-tester/src/golden.rs
```diff
@@ -134,6 +154,16 @@ pub fn felt32(felt: &str) -> anyhow::Result<[u8; 32]> {
     hex32(&format!("{stripped:0>64}"))
 }
 
+/// Decode a base58-encoded 32-byte digest (the form Sui APIs use).
+pub fn base58_32(digest: &str) -> anyhow::Result<[u8; 32]> {
+    let bytes = bs58::decode(digest)
+        .into_vec()
+        .with_context(|| format!("invalid base58: {digest}"))?;
```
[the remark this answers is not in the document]
frolvanya: input is a compile-time golden constant, not CLI/user input; malformed vectors are already rejected by the 32-byte check and `golden_sets__should_all_parse`

But anyway, it won't hurt to add simple check: 64cb875d42a55365135856102dcaa85a0f61e376

frolvanya:
@claude review

frolvanya:
@claude review

crates/foreign-chain-inspector/src/sui/inspector.rs
```diff
@@ -0,0 +1,484 @@
...
+    /// Included in a committee-certified checkpoint. Sui has no reorgs, so this is final.
+    Checkpointed,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum SuiExtractor {
+    Event { event_index: usize },
+}
+
+/// Unlike the JSON-RPC API, the gRPC [`Event`](foreign_chain_rpc_interfaces::sui::proto::Event)
```
anodar: nit: no need to mention JSON-RPC API here, it's unnecessary burden for the reader to know about how deprecated API differed.

Also `Cross-provider protection comes from the fan-out quorum` this is not specific for SUI so fine to drop as well.

crates/foreign-chain-inspector/src/sui/inspector.rs
```diff
@@ -0,0 +1,484 @@
...
+        | Code::Aborted => ForeignChainInspectionError::RpcRequestFailed(status.to_string()),
+        _ => ForeignChainInspectionError::RpcRequestRejected(status.to_string()),
+    }
+}
+
+/// Rejects a backend that returned a different transaction than queried. A non-base58 or
+/// wrong-length `returned` digest is a malformed response; a well-formed but different
+/// digest is a hard inconsistency.
+fn ensure_digest_matches(
+    requested: &[u8; 32],
```
anodar, line 126: These conversions are in a quite a few places. Let's introduce `SuiTransactionDigestBase58` type and implement TryFrom / TryInto from `SuiTransactionDigest`. That way we have strong types and also get rid of length checks and manual bs58::encode/decode in a few places.

crates/foreign-chain-inspector/src/sui/inspector.rs
```diff
@@ -0,0 +1,484 @@
...
+/// - `0x2::sui::SUI` → `0x00…02::sui::SUI` (64 hex digits)
+/// - `0xAB::m::S` → `0x00…ab::m::S`
+/// - `0x2::coin::Coin<0x2::sui::SUI>` → `0x00…02::coin::Coin<0x00…02::sui::SUI>`
+///
+/// Addresses appear at the start of the tag or of a generic type argument — that is, right
+/// after `<`, `,` or a space — and are always followed by `::`. Splitting on those delimiters
+/// (keeping them) yields pieces that each begin at a potential address position, so only a
+/// leading `0x<hex>::` of a piece is rewritten; anything else (identifiers that merely contain
+/// `0x` such as a module named `m0x01`, primitive type args, …) is copied verbatim.
+fn normalize_type_tag(tag: &str) -> String {
```
anodar: Can't we just do `sui_sdk_types::TypeTag::from_str(event_type)
    .map_err(|e| malformed...)?
    .to_string();
`  

here ?

crates/foreign-chain-inspector/src/sui/inspector.rs
```diff
@@ -0,0 +1,484 @@
...
+    if !is_address {
+        return Cow::Borrowed(piece);
+    }
+    let padded = format!("{:0>64}", hex.to_ascii_lowercase());
+    Cow::Owned(format!("0x{padded}::{rest}"))
+}
+
+/// Parse a Sui address string (0x-prefixed hex, possibly short) into [`SuiAddress`].
+/// Short addresses like "0x2" are zero-padded to 32 bytes.
+fn parse_sui_address(s: &str) -> Result<SuiAddress, String> {
```
anodar, line 242: We should be able to use `sui_sdk_types::Address::from(hex)` (we already have have that dependency transitively from sui-rpc), and implement from/into trait for our SUIAddress from contract interface. That way we avoid manually parsing/normalizing.

crates/foreign-chain-inspector/src/contract_interface_conversions.rs
```diff
@@ -363,6 +365,85 @@ impl From<AptosExtractedValue> for dtos::ExtractedValue {
...
+        match value {
+            dtos::SuiExtractor::Event { event_index } => Ok(SuiExtractor::Event {
+                event_index: usize::try_from(event_index).map_err(|_| {
+                    ConversionError::IntegerOverflow {
+                        context: "SuiExtractor::Event event_index exceeds platform usize",
+                    }
+                })?,
+            }),
+            _ => Err(ConversionError::UnsupportedVariant {
+                context: "SuiExtractor",
```
anodar, line 431: nit: would be nice to include value in context as well so that we know which variant we're missing if that happens (I know this follows existing convention in the file though).

crates/foreign-chain-rpc-interfaces/src/sui.rs
```diff
@@ -0,0 +1,219 @@
...
+    client: sui_rpc::Client,
+    timeout: Duration,
+}
+
+impl GrpcSuiClient {
+    /// `auth_header` is sent with every request (for `Header` auth). gRPC has no notion of
+    /// auth in the URL path or query, so header auth is the only supported scheme.
+    ///
+    /// Must be called within a tokio runtime: the (lazy) channel captures the runtime handle.
+    pub fn new(
```
anodar, line 83: We have tests that auth is substituted correctly since https://github.com/near/mpc/commit/e577de24752acea9ddd2615a99ef97ed6cceec1c, can you please also add test-case for this ?

anodar (dismissed):
LGTM.

I'd like to hear your opinion on using `sui_sdk_types` instead of manually normalizing it here. If possible we should switch to that instead.

crates/foreign-chain-rpc-interfaces/src/sui.rs
frolvanya: Added test in 287f9dafa295039ca5dbbac62f5e42b9e731dd4f

crates/foreign-chain-inspector/src/contract_interface_conversions.rs
frolvanya: Added extra field in 287f9dafa295039ca5dbbac62f5e42b9e731dd4f

crates/foreign-chain-inspector/src/sui/inspector.rs
frolvanya: Refactored in 287f9dafa295039ca5dbbac62f5e42b9e731dd4f
`sui_sdk_types::Address` already has `Into` for `[u8; 32]`

crates/foreign-chain-inspector/src/sui/inspector.rs
frolvanya: Refactored in 287f9dafa295039ca5dbbac62f5e42b9e731dd4f

crates/foreign-chain-inspector/src/sui/inspector.rs
frolvanya: Used `sui_sdk_types::Digest` instead in 287f9dafa295039ca5dbbac62f5e42b9e731dd4f

crates/foreign-chain-inspector/src/sui/inspector.rs
frolvanya: Changed comment in 287f9dafa295039ca5dbbac62f5e42b9e731dd4f

anodar (approved):
Thank you!

crates/node-config/src/foreign_chains.rs
```diff
@@ -75,8 +77,18 @@ impl ForeignChainProviderConfig {
...
+        auth::validate_auth_config(&self.auth, &self.rpc_url)?;
+
+        // Sui is reached over gRPC, which carries credentials in request metadata; a token
+        // substituted into the URL path or query would never be sent.
+        if chain == dtos::ForeignChain::Sui {
+            anyhow::ensure!(
+                matches!(self.auth, AuthConfig::None | AuthConfig::Header { .. }),
+                "path or query auth is not supported: gRPC providers support only header auth",
+            );
+        }
```
haiyuechen-nearone, line 90: nit: Should we consider also rejecting URLs with a path or query here? Tonic only uses scheme + authority from the endpoint URL, so path and query portions are dropped silently.

Currently `https://gateway.example.com/sui-mainnet` would validate but the actual connect is to the host root. Known Sui gRPC providers all use hostname + (maybe) port, so a path can be a misconfiguration.

haiyuechen-nearone:
Nit: The PR description looks stale after the split: it still lists the config-tester, localnet docs, and the mocked-client integration tests under "What's included" / "Testing", but those moved to the stacked #3791. Could you update it to reflect what's actually in this PR?

haiyuechen-nearone (approved):
General logic LGTM, added 2 nits that are not merge blocking.

PR #3761, opened 2026-07-09
fix: align allowed_docker_image_hashes consumers with new return shape

barakeinav1:
Fixups for #3752 to pull in — consumers of `allowed_docker_image_hashes` that the return-shape change misses (targets your branch so you can merge it into the PR).

- **Launcher localnet scripts** (`test-verify-and-upgrade.sh`, `test-hash-override.sh`) parse the view with `jq '.[] | select(. == $h)'` — that compares each element to a hash string, which never matches now that elements are `{ image_hash, expiry_timestamp_seconds }` objects, so they'd fail even when the hash is approved. Switched to `select(.image_hash == $h)`.
- **Docs return type**: `contract/README.md` and `securing-mpc-with-tee-design-doc.md` still showed `Vec<NodeImageHash>`.

Not touched (flagging, your call): softer prose mentions in operator/design docs (`running-an-mpc-node-in-tdx-external-guide.md`, `chain-gateway-design.md` receiver type, the attestation-cli README "look in contract state" hint) — they reference the method but not its return shape directly. attestation-cli code itself is unaffected (takes hashes as CLI args).


kevindeforth (approved):
Thank you!

PR #3766, opened 2026-07-10
refactor: alternative without yield resume

kevindeforth:
This is proposed change to #3642 that does not contain a yield resume logic.

https://github.com/near/mpc/pull/3714#discussion_r3556106773


crates/attestation/src/tcb_info.rs
```diff
@@ -60,6 +64,33 @@ pub struct EventLog {
...
+        ]);
+        borsh::schema::add_definition(
+            Self::declaration(),
+            borsh::schema::Definition::Struct { fields },
+            definitions,
+        );
+        <[u8; N] as borsh::BorshSchema>::add_definitions_recursively(definitions);
+    }
+}
+
```
kevindeforth, line 93: so this is purely vibe-coded and will need some inspection before merging. Everything else I had claude on a very tight leash.

pbeza:
@kevindeforth Thanks a lot for taking the time to simplify my PR. 🙏🏼 I really like it!

I thought there was no way around parking the transaction until it was resumed or timed out using `on_attestation_verified` / `env::promise_yield_resume`.

Perhaps we could also simplify the yield-resume logic in the other places where we use it:

https://github.com/near/mpc/blob/52b0c01ca82582c7263e0776c454b7af82ad1467/crates/contract/src/lib.rs#L153-L155

Although I see [you pointed out](https://github.com/near/mpc/pull/3714/changes#r3556106773) that it’s justified there.

Anyway, I think the only downside of switching to this approach is that we’ll no longer be able to prevent concurrent `verify_quote` calls for the same account. That shouldn’t be a problem as long as the node doesn’t retry submitting attestations too quickly. The call is also idempotent, so this all looks good to me.

crates/attestation/src/tcb_info.rs
pbeza: I did some research with Claude, and there doesn’t seem to be an easy way to make this much cleaner without introducing extra types and additional complexity, which would unnecessarily bloat the code, IMO. So I’d leave this manual `BorshSchema` implementation as is.

PR #3759, opened 2026-07-09
test: cover all foreign provider auth kinds in verify-foreign-tx e2e

haiyuechen-nearone:
Closes #2786

### Summary

Extends `verify_foreign_transaction__should_sign_all_supported_chains` so one chain exercises each credential-carrying `AuthConfig` kind: Bitcoin `path`, Base `header`, BNB `query` (the rest keep `none`). Each authenticated mock only answers requests carrying the exact configured credential values, so verification can only succeed if the node injected them — covering the path-auth API key substitution from #2637 (fixed in #2762 with unit tests only) end to end, through real node binaries and config loading.

### Why mocks answer bad auth with 401

Requests without the expected credentials fall through to a catch-all returning HTTP 401 with a JSON-RPC error body, mirroring the providers on the testnet on-chain whitelist: Alchemy returns exactly this body (probed live — including for a literal unsubstituted `{API_KEY}` path, the #2637 scenario; their error reference lists no 404), Geomi returns 401 for an invalid Bearer token, and QuickNode documents 401 `"error": "UNAUTHORIZED"`. A regression therefore produces failure logs matching a real credential misconfiguration.

### Testing

`verify_foreign_transaction__should_sign_all_supported_chains` passes locally in an isolated run (~21s against prebuilt binaries).

PR #3767, opened 2026-07-10
chore: bump crossbeam-epoch to 0.9.20 to fix RUSTSEC-2026-0204

### LINKED ISSUE #3768: Extra CI checks fail on RUSTSEC-2026-0204 (crossbeam-epoch 0.9.18)
### Background

`cargo make check-extra` (which runs `cargo deny check`) fails on [RUSTSEC-2026-0204](https://rustsec.org/advisories/RUSTSEC-2026-0204), published 2026-07-06: the `fmt::Pointer` impl for `Atomic`/`Shared` in `crossbeam-epoch 0.9.18` dereferences the underlying pointer, which is undefined behavior when the pointer is null/invalid.

The crate is a transitive dependency only: nearcore's `near-store` → `crossbeam` → `crossbeam-epoch`, and `rayon-core` → `crossbeam-deque` → `crossbeam-epoch`.

Every CI run that fetches a fresh advisory DB now fails the Extra CI checks job — first seen on #3741 (2026-07-07), also blocking #3759. Runs that passed on 2026-07-08 were using stale cached advisory DBs, which makes the failure look intermittent when it is deterministic.

Same failure mode as the quick-xml episode (#3731 / #3730), but unlike that case no `deny.toml` ignore is needed: the patched version is reachable with a semver-compatible lockfile update.

### User Story

As a developer I want CI on main and PRs to be green so that a red check reliably indicates a problem with the change under review.

### Acceptance Criteria

`cargo deny --all-features check` passes again, with `crossbeam-epoch` bumped to a patched version (>= 0.9.20).

- Upstream fix: crossbeam-rs/crossbeam#1276
- Fix PR: #3767 (lockfile-only bump to 0.9.20; full local test-suite evidence in the PR description)

haiyuechen-nearone:
Closes #3768

### Summary

`cargo deny check` (the Extra CI checks job) fails on [RUSTSEC-2026-0204](https://rustsec.org/advisories/RUSTSEC-2026-0204): `crossbeam-epoch 0.9.18` dereferences the underlying pointer in the `fmt::Pointer` impl for `Atomic`/`Shared`, which is UB when the pointer is null/invalid. The crate is transitive-only for us: nearcore's `near-store` → `crossbeam`, and `rayon-core` → `crossbeam-deque`.

The advisory published 2026-07-06, so every run with a fresh advisory DB now fails (first seen on #3741; also blocking #3759 — runs that passed in between used stale cached advisory DBs). Unlike the quick-xml case (#3730), no `deny.toml` ignore is needed: the patched version is a semver-compatible lockfile bump.

### What we tested / why this is safe

The 0.9.18 → 0.9.20 changelog contains only the fmt fix, ThreadSanitizer compatibility, wider epochs on 32-bit platforms (we target 64-bit), an additive `Guard::defer_destroy` relaxation, and one behavior fix (`Atomic::fetch_update` return value). That last one is the only plausible regression surface, so the full suite was run locally against the bumped lockfile:

- workspace tests: 1452 passed
- `mpc-node`, incl. the 4-node cluster tests — the code paths where nearcore storage and rayon (the actual crossbeam consumers) are exercised hardest: 271/271
- `mpc-contract`: 493/493
- e2e suite against a freshly built node binary: all tests pass (three multi-cluster tests flaked under parallel local load and pass single-threaded — known machine behavior, predates this change)
- `cargo make check-extra`: advisories ok, bans ok, licenses ok, sources ok (fails on `main`)


haiyuechen-nearone:
Crossbeam-epoch version already bumped in Main, closing this PR

PR #3771, opened 2026-07-10
chore(contract): cleanup migrations after 3.13.0 release

### LINKED ISSUE #3758: Bump contract migration baseline to 3.13.0 once it is deployed
## Background

The contract `migrate()` currently reads a single shadow, `v3_12_0_state::MpcContract` — i.e. it migrates **from 3.12.0**. That is correct today: per `crates/contract-history` (hash-checked against on-chain), `current_mainnet()`/`current_testnet()` are both `version_3_12_0()`, so 3.12.0 is the live baseline on both networks. 3.13.0 is tagged but not yet deployed, and there is no 3.13.0 wasm archived.

Once 3.13.0 is actually deployed, the next state-changing release must migrate **from 3.13.0**, not 3.12.0 — otherwise it will fail to deserialize the live state and brick the upgrade. This mirrors #3584 (which bumped the baseline 3.11.2 → 3.12.0).

## Acceptance Criteria

- [ ] 3.13.0 is confirmed deployed on the target network(s)
- [ ] Archive `signer-3_13_0.wasm` in `crates/contract-history`; point `current_mainnet()`/`current_testnet()` at `version_3_13_0()` and update the on-chain-hash assertion
- [ ] Add `crates/contract/src/v3_13_0_state.rs` shadowing the 3.13.0 layout; drop `v3_12_0_state.rs`
- [ ] Repoint `migrate()` (`pub mod` + `try_state_read`) to `v3_13_0_state`
- [ ] `upgrade_to_current_contract` sandbox test passes against the new baseline

## Resources

- Precedent: #3584
- Surfaced during review of #3564 (launcher-hash auto-expiry adds `last_used` + two `Config` fields, so the next migration must account for the 3.13.0 layout)

gilcu3:
Closes #3758


crates/contract/src/v3_12_0_state.rs
```diff
@@ -1,147 +0,0 @@
...
-    node_migrations: NodeMigrations,
-    metrics: Metrics,
-    foreign_chain_rpc_whitelist: ForeignChainRpcWhitelist,
-}
-
-impl From<MpcContract> for crate::MpcContract {
-    fn from(old: MpcContract) -> Self {
-        if let ProtocolContractState::Running(running) = &old.protocol_state {
-            validate_threshold_relation_on_migration(running);
-        }
```
gilcu3, line 105: just realized this was wrong, the contract must be in Running state before migrating

pbeza (approved):

anodar (approved):

haiyuechen-nearone (approved):

PR #3772, opened 2026-07-10
refactor: deduplicate test port allocators into a single allocator

haiyuechen-nearone:
Closes #2781

### Summary
Extract `PortSeed` (mpc-node) and `E2ePortAllocator` (e2e-tests) into `TestPorts` type and specialized traits in the `test-port-allocator` crate. The originally shared accessors are implemented directly on `TestPorts`. The specialized surface lives on per-callsite capability traits
- `NodeTestPorts` (`with_case`)
- `E2eTestPorts` (cluster + `near_*` ports) 

Exact port numbers and callsite behaviours are unchanged.

### Invariants
- The three ranges (`10000+` / `20000+` / `40000+`) are partitioned in `test-port-allocator`: each scheme's `space_end` is pinned to the next range's base, and `const fn TestPorts::new` rejects any `test_id` whose block would cross it — evaluated at compile time for the `port_seed` registry.
- Golden-value tests pin the exact port numbers.

### Cargo wiring
`test-port-allocator` stays an optional dependency of `mpc-node` behind the `test-utils` feature (+ dev-dependency), so the release node binary links nothing new; `e2e-tests` and `backup-cli` tests use it too.


anodar (changes requested):
This doesn't quite do what issue is describing. There's still a lot of duplication between `E2ePortAllocator` and `PortSeed` struct, while they both serve the same purpose: generate ports deterministically based on some parameters to avoid collision. 

E.g. `p2p_port`, `web_port`, `migration_web_port, `pprof_web_port` are all shared between the two. 

I'd suggest to factor out port allocator in some helper crate, then have 2 traits: one with methods that e2e tests need (from E2EPortAllocator, and another from PortSeed), and implement both traits for the same struct from that helper crate.

That way you have single struct, but each call-site only has interface that it needs and nothing more.

What do you think?

haiyuechen-nearone:
Makes sense, I will restructure the code.

crates/test-port-allocator/src/lib.rs
```diff
@@ -1,32 +1,224 @@
...
+        assert!(
+            offset < self.ports_per_node,
+            "node offset {offset} exceeds ports_per_node {}",
+            self.ports_per_node
+        );
+        self.test_base(test_id)
+            + self.cluster_ports
+            + node_index as u16 * self.ports_per_node
+            + offset
+    }
```
netrome: nit: Looks like you want to take the u16 as input and leave the conversion to the caller for `node_index`
```suggestion
    fn node_port(&self, test_id: u16, node_index: u16, offset: u16) -> u16 {
        assert!(
            (node_index) < self.max_nodes,
            "node_index {node_index} exceeds max_nodes {}",
            self.max_nodes
        );
        assert!(
            offset < self.ports_per_node,
            "node offset {offset} exceeds ports_per_node {}",
            self.ports_per_node
        );
        self.test_base(test_id)
            + self.cluster_ports
            + node_index * self.ports_per_node
            + offset
    }
```

crates/test-port-allocator/src/lib.rs
```diff
@@ -1,32 +1,224 @@
...
+/// drive several disjoint test cases via [`TestPorts::with_case`].
+pub trait MultiplexedPortSpace: PortSpace {
+    /// Consecutive offsets each case occupies within a node's block.
+    const PORTS_PER_CASE: u16;
+}
+
+/// A bundle of TCP ports for one `test_id`, laid out by space `S`. Space-specific
+/// surface (cluster ports, `with_case`) is gated to the space that defines it.
+#[derive(Copy, Clone, Debug)]
+pub struct TestPorts<S: PortSpace> {
```
netrome: Trait bounds on structs is typically code smell. See https://stackoverflow.com/questions/49229332/should-trait-bounds-be-duplicated-in-struct-and-impl for a more general elaboration on it.

What's the purpose of holding phantom data of a trait which holds a set of constants as opposed to just adding a concrete port allocation scheme to this struct?

netrome (changes requested):
I'm a bit confused as to why we need the phantom data + trait with an associated constant. It feels like this should be possible to simplify. I also left another nit. However, the only concrete blocker is that trait bounds should not be specified on structs (with very few exceptions, mostly relating to `send` or `sync` bounds iirc), but generally only in impl blocks.

If you remove the trait bounds on the `TestPorts` struct I'll approve.

netrome (dismissed):
Looks better now, thanks for updating 🙏  (shallow review)

crates/e2e-tests/README.md
```diff
@@ -216,19 +216,26 @@ impl MpcClusterConfig {
...
-### 5. `E2ePortAllocator` — deterministic port layout
+### 5. `TestPorts::e2e_tests` — deterministic port layout
 
-Each test declares a `port_seed: u16`. Ports are computed as:
+Each test declares a `port_seed: u16` (the allocator's `test_id`). Ports are
+computed as:
 
 ```
-BASE_PORT (20000) + test_id * PORTS_PER_TEST + offset
+e2e base (20000) + test_id * 82 + offset
```
anodar:
```suggestion
e2e base (20000) + test_id * ports_per_test() + offset
```

crates/e2e-tests/README.md
```diff
@@ -216,19 +216,26 @@ impl MpcClusterConfig {
...
 ```
 
-with `PORTS_PER_TEST = 2 + 10 * 8` (2 cluster ports + 8 per-node ports × 10
-maximum nodes). Cluster-level ports cover the sandbox RPC and network; per-node
+with 82 ports per test (2 cluster ports + 8 per-node ports × 10 maximum
```
anodar: This is 82 only for `E2E_TESTS_SCHEME`, not for `MPC_NODE_TESTS_SCHEME`. Something worth rewording.

anodar (dismissed):
Thank you!

netrome (approved):

anodar (approved):

PR #3777, opened 2026-07-10
docs: design smart contract voting with commit-then-reveal approach

### LINKED ISSUE #3482: Document the new smart contract voting mechanism
### Background

With @netrome and @kevindeforth  we discussed the next-gen voting strategy for voting on the smart contract including the voter's API, smart contract storage and smart contract logic.

This needs to be documented with a vision over the next potential PRs

### User Story

_No response_

### Acceptance Criteria

- Compress the knowledge collected in https://nearone.slack.com/archives/C07UW93JVQ8/p1780309095972559 into a doc.
- Describe the implementation strategy for the next PRs togo

_No response_

SimonRastikian:
Closes #3482

### DIFF
--- a/docs/design/unified-contract-voting.md
+++ b/docs/design/unified-contract-voting.md
@@ -5,7 +5,11 @@ backed by its own bespoke vote-storage struct that re-implements the same primit
 register a vote, replace a voter's previous vote, count votes for a proposal, prune voters
 who left the participant set, and clear votes after a decision. This document proposes
 consolidating the **governance / CRUD / configuration** flows onto a single
-**commit-reveal** voting interface built on the existing generic `Votes<V>` primitive.
+**on-chain commit-reveal** voting interface built on the existing generic `Votes<V>` primitive.
+Here voting stays on-chain — each participant submits a `vote(hash)` transaction — and only the
+proposal *payload* is shared off-chain. A more radical, *fully off-chain* voting alternative
+(the tally itself moves off-chain, with a single aggregated submission on-chain) is a distinct
+architecture, kept separate in §10.
 
 To try and keep the scope small and with little complexity, we took two decisions: we build on the primitive that already exists rather
 than inventing a new framework, and we explicitly **exclude** the protocol-driven key-event
@@ -16,25 +20,29 @@ votes, whose semantics do not fit a generic proposal interface (see §6).
 1. [Goals](#1-goals)
 2. [Current state analysis](#2-current-state-analysis)
 3. [Proposed design — commit-reveal unified interface](#3-proposed-design--commit-reveal-unified-interface)
-4. [Conclusion mechanics and conflicting proposals](#4-conclusion-mechanics-and-conflicting-proposals)
+4. [Execution mechanics and conflicting proposals](#4-execution-mechanics-and-conflicting-proposals)
 5. [Cleanup and TTL semantics](#5-cleanup-and-ttl-semantics)
 6. [Scope — what migrates and what stays bespoke](#6-scope--what-migrates-and-what-stays-bespoke)
 7. [Migration strategy](#7-migration-strategy)
-8. [Alternatives considered](#8-alternatives-considered)
-9. [Open questions](#9-open-questions)
+8. [Cost model](#8-cost-model)
+9. [Alternatives considered](#9-alternatives-considered)
+10. [Alternative architecture: fully off-chain aggregated voting](#10-alternative-architecture-fully-off-chain-aggregated-voting)
+11. [Open questions](#11-open-questions)
 
 ## 1. Goals
 
 - **Consolidate** the ad-hoc governance/CRUD/config voting flows onto one interface, removing
   duplicated vote/count/replace/cleanup logic.
 - **Reduce on-chain cost** by letting participants commit to a proposal with a 32-byte hash
-  and revealing the full payload only once, when the proposal is concluded.
+  and revealing the full payload only once, when the proposal is executed.
 - **Unify cleanup and TTL semantics** so that abandoned proposals cannot accumulate in
   contract storage indefinitely.
 
 Non-goals: folding the protocol-driven key-event votes into the governance interface (§6.1
 explains why they stay separate — though §6.2 notes they can be consolidated *among
-themselves*), or building the off-chain proposal-sharing service (§3, treated as optional UX).
+themselves*), or building a dedicated off-chain proposal-sharing *service* — that service is
+optional UX (§3.3); distributing proposals off-chain so voters can hash them is required, but
+ad-hoc gossip suffices.
 
 ## 2. Current state analysis
 
@@ -44,8 +52,9 @@ All entry points live in `crates/contract/src/lib.rs`. The **threshold** column 
 rule — how many votes a proposal needs to take effect. There are only three distinct rules
 plus one special case:
 
-- **`t-of-n`** — the configured signing threshold in `ThresholdParameters` (a subset
-  suffices).
+- **`t-of-n`** — the **governance threshold** in `ThresholdParameters` (a subset suffices).
+  The contract keeps this `>= max(reconstruction_threshold)` across domains, so a governance
+  action can never pass with fewer voters than are needed to reconstruct a key.
 - **`n-of-n (current)`** — every participant in the *current* set must vote (unanimity).
 - **`n-of-n (proposed)`** — every participant in the *incoming* set must vote; distinct from
   the above only because the participant set itself is being changed.
@@ -82,8 +91,8 @@ Backing structs:
 
 ### 2.2 The primitive already exists
 
-`crates/contract/src/primitives/votes.rs` defines a generic `Votes<V>` that is exactly the
-"hashing mechanism" the original sketch reached for:
+`crates/contract/src/primitives/votes.rs` defines a generic `Votes<V>` that provides exactly
+the hashing-based vote storage this design needs:
 
 ```rust
 pub struct Votes<V>
@@ -113,8 +122,8 @@ remaining flows each reinvent a subset of the same behavior over a plain `BTreeM
 
 Every `vote_*` method — including the two `Votes<V>` adopters — takes the **full proposal**
 at vote time and lets the contract compute its hash. Nothing votes on a bare hash and defers
-the payload. The "defer revealing content until late" idea in the original sketch is the one
-genuinely new capability this document proposes.
+the payload. Deferring the payload until execution is the one genuinely new capability this
+document proposes.
 
 ## 3. Proposed design — commit-reveal unified interface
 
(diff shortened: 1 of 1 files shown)


kevindeforth (changes requested):
I think the proposed approach is a bit complicated. It proposes two changes:
- one shared enum and potentially entry-point for all votes
- a commit-and reveal mechanism with a locking mechanism.

It's not entirely obvious to me why we _need_ the above properties. The locking-mechanism makes this a stateful solution which may be hard to maintain.

I had a simpler mechanism in mind:
- everyone sends the thing they are voting for (e.g. proposed contract update)
- contract hashes the thing that was submitted (e.g. the code) and compares the hash against the received votes for the update
- if successful, the thing is applied (e.g. the update).

This way, we no longer need to store the thing that we are voting for in the contract state.

@SimonRastikian it would be helpful if you could outline the motivation for the above two properties, so we get a better idea how this would be useful the network in production, and such that we can verify if the additional effort compared to the minimal solution is worth pursuing.

SimonRastikian:
@kevindeforth 
- The enum tries to unify the proposal function: one single entry point for all kinds of proposals making it much simpler for a caller as in theory the purpose is to have one single propose function
- The lock mechanism is quite important when you have concurrent proposals that get both enough votes. Without the locking, there could be cases where things would become chaotic especially when changing thresholds.

kevindeforth:
> The enum tries to unify the proposal function: one single entry point for all kinds of proposals making it much simpler for a caller as in theory the purpose is to have one single propose function

Yeah, I agree that would be nice! I initially thought it might be too cumbersome to do, but I think the refactor might be Ok. Need to check on any return types though.

>  Without the locking, there could be cases where things would become chaotic especially when changing thresholds.

This is not necessary if we rely on atomic receipt execution and allow a receipt to carry at most one vote - the transaction that brings the vote over the threshold will also be the transaction that issues the change. We won't need a lock.

SimonRastikian:
> This is not necessary if we rely on atomic receipt execution and allow a receipt to carry at most one vote - the transaction that brings the vote over the threshold will also be the transaction that issues the change. We won't need a lock.

I do not know how to do that. I am not sure I have the same mental model as you. Would be glad to discuss through

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+who left the participant set, and clear votes after a decision. This document proposes
+consolidating the **governance / CRUD / configuration** flows onto a single
+**on-chain commit-reveal** voting interface built on the existing generic `Votes<V>` primitive.
+Here voting stays on-chain — each participant submits a `vote(hash)` transaction — and only the
+proposal *payload* is shared off-chain. A more radical, *fully off-chain* voting alternative
+(the tally itself moves off-chain, with a single aggregated submission on-chain) is a distinct
+architecture, kept separate in §10.
+
+To try and keep the scope small and with little complexity, we took two decisions: we build on the primitive that already exists rather
+than inventing a new framework, and we explicitly **exclude** the protocol-driven key-event
```
barakeinav1, line 15: nit

```suggestion
than inventing a new framework, and we explicitly **exclude** the MPC protocol-driven key-event
```

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+
+## 2. Current state analysis
+
+### 2.1 The `vote_*` entry points
+
+All entry points live in `crates/contract/src/lib.rs`. The **threshold** column is the passing
+rule — how many votes a proposal needs to take effect. There are only three distinct rules
+plus one special case:
+
+- **`t-of-n`** — the **governance threshold** in `ThresholdParameters` (a subset suffices).
```
barakeinav1, line 55: what is T in this case?
do we already have a "governance threshold" today that is separated from each protocol threshold?

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+- A vote stores a `ProposalHash` — a 32-byte SHA-256 digest derived from any type that
+  implements `ProposalHashEncoding` (`votes.rs:148`), via borsh serialization.
+- `vote()` replaces a voter's prior vote automatically (idempotent re-votes).
+- `votes_by_proposal` is a reverse index, so counting votes for a proposal is cheap.
+- `VoterSet::count_for(predicate)` counts only the voters matching a predicate — used to
+  count only *current* participants.
+- Cleanup is first-class: `remove_votes_for_proposal()`, `retain_votes(predicate)`, and
+  `clear()`.
+
+**Two of the flows above already use it:** `TeeVerifierVotes`
```
barakeinav1, line 116: this is outdated, 
look at https://github.com/near/mpc/commit/676f5b693b0292181915aa2e8294eae0694d5d74 
and https://github.com/near/mpc/commit/96731032497911271da80caee625affb17620cd6

seems like @kevindeforth has ready start implementing :-)

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+(`tee/verifier_votes.rs:44`) wraps `Votes<AuthenticatedParticipantId>`, and `ProviderVotes`
+(`foreign_chain_rpc.rs:147`) wraps `Votes<(AuthenticatedParticipantId, ForeignChain)>`. The
+remaining flows each reinvent a subset of the same behavior over a plain `BTreeMap`.
+
+### 2.3 What is *not* implemented today
+
+Every `vote_*` method — including the two `Votes<V>` adopters — takes the **full proposal**
```
barakeinav1, line 123: So you want to reverse the direction, first we vote, and only then supply the payload at the end, right?

what is the main benefit(s) for this?
the main advantage I see is the storage usage (e.g for contract upgrades)

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+    AddLauncherHash(LauncherImageHash),
+    RemoveLauncherHash(LauncherImageHash),
+    AddOsMeasurement(ContractExpectedMeasurements),
+    RemoveOsMeasurement(ContractExpectedMeasurements),
+    CodeHash(NodeImageHash),
+    TeeVerifierChange(VerifierChangeProposal),
+    ForeignChainProviders(/* per-chain entries */),
+    Update(UpdateId),
+    AddDomains(Vec<DomainConfig>),
+    NewParameters(ProposedThresholdParameters),
```
barakeinav1: we should take the opportunity to change some names
e.g
CodeHash ->NodeManifestHash
NewParameters->some better name

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+}
+```
+
+Because the borsh encoding of an enum is prefixed with its variant discriminant, the derived
+`ProposalHash` is domain-separated across variants for free: a hash committed for
```
barakeinav1: I didn't underrated this sentence

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+to apply its effect and (b) its required threshold (`t-of-n` vs `n-of-n`, per the table in
+§2.1).
+
+The `ProposalHash` derived from the `Proposal` enum doubles as the proposal's identifier, so
+no separate id type is needed.
+
+### 3.2 The API
+
+```
+vote(proposal_hash: ProposalHash)   // commit
```
barakeinav1: related to what comment above. how do make the relation between the **vote** and **execute** commands.

vote(proposal_hash: ProposalHash) - should there be another parameter here?

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+execute(proposal: Proposal)        // reveal + execute
+```
+
+- `vote(class, proposal_hash)` records a 32-byte commitment for the caller in the `Votes<V>`
+  partition for the proposal's conflict class (§4.2), keyed by their
+  `AuthenticatedParticipantId`. Cheap and constant-size regardless of how large the proposal
+  is.
+- `execute(proposal)` recomputes `ProposalHash::from(&proposal)`, looks up its `VoterSet`, and
+  **verifies the hash has enough votes to meet the variant's required threshold** —
+  `count_for(is_current_participant)` against that threshold, which for the `t-of-n` governance
```
barakeinav1: >`t-of-n`

this should depend on what different voting governance scheme and can be n-of-n, etc..? right?

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+### 3.3 Off-chain payload sharing
+
+The only off-chain step is distributing the proposal so voters can hash it: `vote` carries just
+the hash. Gossip suffices — the author has every incentive to distribute it — so a dedicated
+sharing *service* is **optional** and out of scope.
+
+The preimage is no off-chain secret: `execute(proposal)` delivers it on-chain, and the contract
+checks `hash(proposal)` matches a threshold-reaching vote before applying (§4.1). The only
+liveness need is that one honest party still holds the proposal at execute time; §4.5 shows this
+falls out of normal voting.
```
barakeinav1: I didn't understand the last sentence

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+proposal acquires a per-class **execution lock**: while it is held, no other proposal in the
+same class may execute, so a conflicting sibling cannot also be applied. The lock is on
+*execution only* — it does **not** block vote gathering. Voting stays optimistic and
+concurrent: several proposals in a class may keep accumulating votes at once (as `ProviderVotes`
+already does per-chain, `foreign_chain_rpc.rs:192-196`), and the lock decides only which one gets
+to apply. Sibling votes are **preserved**, not wiped.
+
+The granularity of a conflict class is *not* uniformly "same proposal type":
+
+| Flow shape | Conflict key | Executing one blocks… |
```
barakeinav1: this table look ok, but need to think which concurrent flows we really want to supports and which not.

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+that supersedes the stuck one, re-opening the class under a new holder and expiry. This override
+is a governance action (a threshold of participants) and matters mainly on the governance side;
+the state-transition classes already have their own cancel votes and the state machine to fall
+back on. Crucially, the function that installs the ephemeral lock is **not itself gated by any
+execution lock** — otherwise a stuck lock could block its own recovery — so it can always make
+progress.
+
+### 4.4 Precondition re-validation (the safety guarantee)
+
+The execution lock stops a *decided* sibling from also applying, but it only serializes proposals
```
barakeinav1: I find this paragraph 4.4 very hard to read/understand

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+authored against a since-replaced participant set). Therefore `execute` also **re-validates the
+proposal's precondition against current state** and rejects it if no longer valid.
+
+The lock keeps execution serialized and storage tidy; the precondition check is the actual safety
+guarantee. Together they make execution order-independent. This mirrors existing prior art: the
+key-event votes bind to a `KeyEventId` and `verify_vote` (`state/key_event.rs:173`) rejects votes
+targeting a superseded attempt — the same slot-plus-staleness-check pairing the execution lock
+uses.
+
+### 4.5 Observability: how `execute` gets triggered
```
barakeinav1: how do node, observe those state changes today?
do the nodes watch voting? if not, they there is no node change needed for this.

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+
+### 4.6 Local hashing in the node's `vote` command
+
+The soft prevention above is a client-side convention. The *contract* method is unchanged —
+`vote(proposal_hash)` only ever sees a hash — but the node's vote helper takes the full
+`Proposal` and hashes it locally, submitting only the 32-byte hash:
+
+```rust
+// today: the full payload goes on-chain; the contract hashes it
+fn vote_code_hash(code_hash) { tx.call("vote_code_hash", code_hash); }
```
barakeinav1: this in not a good example, 'vote_code_hash' is not sent from the node but from the operator.

which vote command are needed by the node?

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+**Caveat — canonical encoding.** Every voter must hash the *identical* borsh bytes, or two votes
+for the "same" proposal produce different hashes and never converge on a threshold. The node's
+local hashing must therefore use exactly the `ProposalHashEncoding` the contract applies at
+`execute` time (§3.1), so the `Proposal` enum's serialization must be stable and shared.
+
+## 5. Cleanup and TTL semantics
+
+### 5.1 TTL
+
+Because the execution lock (§4.3) is non-destructive, votes are never wiped when a proposal is
```
barakeinav1: this is different than today, right? we do cleanup remaining votes after a successful vote and state change. why do we want to change this behavior?

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+
+Participant-set changes mid-vote are a subtlety the design must account for, and `Votes<V>`
+handles it: it is keyed by `AuthenticatedParticipantId`,
+counting always filters to the *current* set via `count_for`, and `retain_votes(predicate)`
+prunes voters who have left after a resharing. This is a primary reason to build on the
+existing primitive rather than a fresh `BTreeMap`.
+
+## 6. Scope — what migrates and what stays bespoke
+
+**In scope (commit-reveal candidates)** — the governance/CRUD/config flows: add/remove
```
barakeinav1: we have a table above, maybe add another column to it, (changes yes/no). to avoid duplication

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+  honest-but-diverging keys would simply produce two different hashes and reach no consensus
+  *silently*, losing the explicit abort signal.
+- **`execute` would become a god-method.** Executing these votes drives the core state
+  machine — advancing domains, transitioning between Running / Initializing / Resharing, and
+  firing the post-resharing cleanup promise fan-out (`lib.rs:1177`). That is not "apply a
+  proposal effect."
+- **No commit-reveal payoff.** The payloads are tiny and protocol-generated, and must hit the
+  chain regardless — there is no storage to defer and no need for off-chain gossip.
+
+### 6.2 Consolidation available *within* the key-event family
```
barakeinav1: I'm not familiar enough with this part, so skipping for now.

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+```rust
+enum Proposal {
+    AddLauncherHash(LauncherImageHash),
+    RemoveLauncherHash(LauncherImageHash),
+    AddOsMeasurement(ContractExpectedMeasurements),
+    RemoveOsMeasurement(ContractExpectedMeasurements),
+    CodeHash(NodeImageHash),
+    TeeVerifierChange(VerifierChangeProposal),
+    ForeignChainProviders(/* per-chain entries */),
+    Update(UpdateId),
```
barakeinav1: what does this one do?

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+  bitmap.
+- The contract reconstructs the aggregate public key from the participant set's registered BLS
+  keys, runs a single pairing check, counts the bitmap (filtered to current participants, §5.2)
+  against the variant's threshold, re-validates the precondition (§4.4), and applies the effect.
+
+### 10.2 Why it is attractive
+
+- **One transaction per decision, fleet-wide.** Where the on-chain design puts `n` `vote`
+  transactions on-chain (plus an `execute`), this puts down a single submission regardless of
+  participant count — the largest possible gas saving.
```
barakeinav1: >the largest possible gas saving.

I don't think this is real consideration.

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+- Once enough signatures are gathered, a **single** participant submits one transaction carrying
+  the proposal opening plus the collected signatures.
+- Ideally the signatures are combined into one constant-size **BLS aggregate signature** — all
+  voters sign the *same* proposal hash, the ideal case for aggregation — accompanied by a signer
+  bitmap.
+- The contract reconstructs the aggregate public key from the participant set's registered BLS
+  keys, runs a single pairing check, counts the bitmap (filtered to current participants, §5.2)
+  against the variant's threshold, re-validates the precondition (§4.4), and applies the effect.
+
+### 10.2 Why it is attractive
```
barakeinav1: Just thinking out loud. 
I wonder if having all signatures gathered off chain has some value regarding privacy, anonymity or non influence. e.g something like secret voting in a Parlement.

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+
+- **One transaction per decision, fleet-wide.** Where the on-chain design puts `n` `vote`
+  transactions on-chain (plus an `execute`), this puts down a single submission regardless of
+  participant count — the largest possible gas saving.
+- **No out-of-order hazard.** Voting and application collapse into one atomic transaction, so it
```
barakeinav1: I wonder if this complexly is really worth it

docs/design/unified-contract-voting.md
barakeinav1: can you add an example of a near CLI command line how of new voting commans  (maybe new v.s old)

docs/design/unified-contract-voting.md
barakeinav1: I think the main storage win would be contract updates, but with `Update(UpdateId)` the code is still stored via `propose_update`. What if we vote on the code hash and upload the wasm only once, at execute? Then we don't need `propose_update` or its deposit at all.

docs/design/unified-contract-voting.md
```diff
@@ -0,0 +1,654 @@
...
+`vote_new_parameters` / `vote_add_domains` transition state inline and discard the old state's
+tally. There is no "decided but not yet applied" window today, so this ordering problem is a
+genuine *cost introduced by* commit-reveal, not a pre-existing bug (see also §9).
+
+The design closes it with two complementary mechanisms: a per-class **execution lock** (§4.3)
+and **precondition re-validation** (§4.4).
+
+### 4.3 Conflict-class execution lock
+
+Each `Proposal` variant declares a **conflict class** via a `conflict_key()`. Executing a
```
barakeinav1: Not sure we need the lock. With one vote per voter per class and a threshold ≥60%, two proposals in the same class can't both have enough votes at the same time. So if we clear the votes on execute and validate at execute time, the first execute wins, which is the same as the first to cross the threshold today. Is there a concrete case where the lock is still needed?

docs/design/unified-contract-voting.md
barakeinav1: Thinking about it more, I think this is where the complexity comes from. Today we keep one vote per voter per flow and clear all of them once a proposal is applied. If we keep that rule, a sibling can't be executed later, so we don't need the lock, lock TTL, or lock recovery. What's the reason for keeping the sibling votes?

barakeinav1 (changes requested):
I added some comments, questions and suggestions.
I'm not sure if we should go with option B (Consolidate onto Votes<V>, keep reveal-at-vote)
or C (Full commit-reveal)

PR #3778, opened 2026-07-10
refactor: Reducing code redundancy

SimonRastikian:
Tiny cleanup. Creating a helper function and reducing code redundancy.

### DIFF
--- a/crates/e2e-tests/tests/request_during_resharing.rs
+++ b/crates/e2e-tests/tests/request_during_resharing.rs
@@ -7,18 +7,15 @@ use near_mpc_contract_interface::types::{
 };
 use rand::SeedableRng;
 
-/// Returns the single domain running `protocol_type`.
-///
-/// Each protocol appears at most once in this test's domain registry, so the
-/// protocol uniquely identifies its domain.
-fn create_domain(contract_state: &RunningContractState, protocol_type: Protocol) -> DomainConfig {
+/// Panics if no domain matches; each protocol appears at most once in this test's domain set.
+fn find_domain_id(contract_state: &RunningContractState, protocol_type: Protocol) -> DomainId {
     contract_state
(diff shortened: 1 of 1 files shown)


haiyuechen-nearone (approved):

crates/e2e-tests/tests/request_during_resharing.rs
```diff
@@ -1,11 +1,23 @@
...
 use rand::SeedableRng;
 
+/// Panics if no domain matches; each protocol appears at most once in this test's domain set.
+fn find_domain_id(contract_state: &RunningContractState, protocol_type: Protocol) -> DomainId {
+    contract_state
+        .domains
+        .domains
+        .iter()
+        .find(|d| d.protocol == protocol_type)
+        .unwrap_or_else(|| panic!("no domain with protocol {protocol_type:?}"))
```
barakeinav1, line 17: nit (non-blocking): `.find()` returns the first match and this only panics on *zero* matches — not on multiple. The doc says "each protocol appears at most once in this test's set," but nothing enforces it, and production actually has two `CaitSith` domains (Sign + ForeignTx). If `default_for_test` ever grows a second same-protocol domain, this would silently pick whichever comes first (the old `purpose == Sign` filter guarded against that). Cheap to make it fail loudly instead:

```rust
let mut matches = contract_state.domains.domains.iter().filter(|d| d.protocol == protocol_type);
let domain = matches.next().unwrap_or_else(|| panic!("no domain with protocol {protocol_type:?}"));
assert!(matches.next().is_none(), "multiple domains with protocol {protocol_type:?}");
domain.id
```

barakeinav1 (approved):
Thanks, approved — pure test refactor, no logic change. One non-blocking nit on the helper's uniqueness assumption.

PR #3781, opened 2026-07-13
feat: small defense in depth check for the evm inspector

### LINKED ISSUE #3681: Bind EVM inspector receipt and log to the requested transaction
Note: this is defense in depth, not a cryptographic fix. The proposed checks only compare provider-returned fields against each other and the request — a fully Byzantine provider returning internally-consistent forged data still passes. Genuine protection requires verifying the receipt against the block (e.g. receiptsRoot Merkle proof), which is out of scope here.

### Background

The EVM foreign-chain inspector never binds the RPC responses it receives to what was actually queried. Two gaps:

1. **Receipt not bound to the requested tx hash.** `GetTransactionReceiptResponse` (`crates/foreign-chain-rpc-interfaces/src/evm.rs`) carries no `transactionHash`, and `EvmInspector::extract` (`crates/foreign-chain-inspector/src/evm/inspector.rs:50-85`) validates only finality, canonical block, and `status == 1` — never that the receipt corresponds to the requested tx.

2. **Log not bound to its receipt/block.** `EvmExtractor::Log` selects a log purely by index:

```rust
EvmExtractor::Log { log_index } => {
    let log = rpc_response.logs.iter()
        .find(|log| log.log_index == target_index)
        ...
}
```

The log's own `transaction_hash` / `block_hash` / `block_number` are never checked.

~~The extracted values flow into the threshold-signed `ForeignTxSignPayloadV1`, so a malicious or MITM'd RPC provider can influence the attested data. The Bitcoin inspector already echoes back and checks the queried hash; the EVM path (and similarly Starknet) does not.~~
EDIT: the text above was misleading

### Acceptance Criteria

- `GetTransactionReceiptResponse` includes the transaction hash, and `extract` rejects a receipt whose hash differs from the requested tx hash.
- After selecting a log by index, `extract` asserts `log.block_hash == receipt.block_hash`, `log.block_number == receipt.block_number`, and `log.transaction_hash == requested_tx_hash`.
- A clear error variant is returned on mismatch (not a panic).
- Tests cover each mismatch path.

gilcu3:
Closes #3681

### DIFF
--- a/crates/foreign-chain-inspector/tests/evm_inspector.rs
+++ b/crates/foreign-chain-inspector/tests/evm_inspector.rs
@@ -566,10 +566,12 @@ macro_rules! evm_inspector_tests {
             async fn extract__should_reject_receipt_whose_transaction_hash_differs_from_request()
             {
                 // Given
-                let tx_id = TxHash::from([1; 32]);
+                let requested_tx_bytes: [u8; 32] = [1; 32];
+                let returned_tx_bytes: [u8; 32] = [0xdd; 32];
+                let tx_id = TxHash::from(requested_tx_bytes);
 
                 let tx_response = GetTransactionReceiptResponse {
-                    transaction_hash: H256::from([0xdd; 32]),
+                    transaction_hash: H256(returned_tx_bytes),
                     block_hash: H256::from([2; 32]),
                     block_number: U64::from(90),
                     status: U64::one(),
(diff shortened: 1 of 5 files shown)


haiyuechen-nearone:
Should we add this guard for Starknet as well?

gilcu3:
> Should we add this guard for Starknet as well?

I thought about it but it seems not simple for Starknet, as the receipts have a different structure (but I do not understand the details very well, as this is Starknet specific stuff)

haiyuechen-nearone:
I think we have transaction hashes for the Starknet providers as well:

https://github.com/starkware-libs/starknet-specs/blob/c2765866e1d5f2e62c5649f7baf5ca53d0e84872/api/starknet_api_openrpc.json#L2975-L2979

https://www.alchemy.com/docs/chains/starknet/starknet-api-endpoints/starknet-get-transaction-receipt

https://www.quicknode.com/docs/starknet/starknet_getTransactionReceipt

@anodar please keep me honest on this

haiyuechen-nearone (dismissed):
Starknet tx hash check can be added as a follow up, I think this is good to go.

crates/foreign-chain-inspector/tests/evm_inspector.rs
```diff
@@ -550,6 +561,164 @@ macro_rules! evm_inspector_tests {
...
+                assert_matches!(
+                    response,
+                    Err(ForeignChainInspectionError::LogNotBoundToReceipt {
+                        log_index: 1,
+                        log_block_number: 91,
+                        receipt_block_number: 90,
+                        ..
+                    })
+                );
+            }
```
anodar, line 654: nit: consider deduplicating this 3 tests into single test with table driven tests.

crates/foreign-chain-inspector/tests/evm_inspector.rs
```diff
@@ -550,6 +561,164 @@ macro_rules! evm_inspector_tests {
...
+            async fn extract_log_from_receipt_with(
+                tx_id: TxHash,
+                log: Log,
+            ) -> Result<Vec<ExtractedValue>, ForeignChainInspectionError> {
+                let finality_block_response = GetBlockByNumberResponse {
+                    number: U64::from(100),
+                    hash: H256::from([0xaa; 32]),
+                };
+                let tx_response = GetTransactionReceiptResponse {
+                    transaction_hash: H256::from([3; 32]),
```
anodar: nit: why not use tx_id here instead of hardcoding?

anodar (dismissed):
LGTM. Dropped some optional nits.

crates/foreign-chain-inspector/tests/evm_inspector.rs
gilcu3: no reason really, but there were a lot of hardcoded values around here, so I didn't see a problem in following the pattern. Will check if this could be just using `tx_id` instead

anodar (approved):
Thank you!

haiyuechen-nearone (approved):

PR #3784, opened 2026-07-13
ci: run doctests in CI (and fix contract README doctest)

### LINKED ISSUE #3783: ci: doctests are never run in CI
### Background

CI runs the test suite via `cargo nextest`, which does **not** execute doctests, and `cargo make check-docs` only runs `cargo doc` (intra-doc link checking). So **no CI job compiles or runs doctests at all**.

This surfaced when a `rust` block in `crates/contract/README.md` — pulled in as a doctest via `#![doc = include_str!("../README.md")]` — broke: it references `Quote`, `Collateral`, `TcbInfo`, and `ExpectedMeasurements` with no imports (regression from #3729). It only fails under `cargo test --doc --all-features`, which nothing in CI runs, so it was found by hand locally.

### User Story

As a contributor, I want doctests exercised in CI so doctest regressions are caught on the PR that introduces them, not discovered by hand.

### Acceptance Criteria

- [ ] CI runs `cargo test --doc` (workspace, `--all-features`).
- [ ] Doctests are runnable locally via a `cargo make` task.
- [ ] The existing `mpc-contract` README doctest passes.

- Regression that exposed the gap: #3729
- Broken block: https://github.com/near/mpc/blob/main/crates/contract/README.md#L362
- Fixed by #3784

barakeinav1:
Closes #3783

## Problem

`cargo nextest` — what CI runs for the test suite — does not execute doctests at all (https://nexte.st/docs/running/), and `cargo make check-docs` only runs `cargo doc` (intra-doc link checking). So **doctests were never exercised in CI**, which is why the broken contract README doctest (a regression from #3729) slipped through. Running `cargo test --doc --workspace --all-features` locally is the only thing that caught it.

## Changes

- **Fix the broken doctest**: mark the `DstackAttestation` example in `crates/contract/README.md` (pulled in as a doctest via `#![doc = include_str!("../README.md")]`) as `rust,ignore` — it references `Quote`, `Collateral`, `TcbInfo`, `ExpectedMeasurements` without imports and only illustrates the struct shape.
- **Run doctests in CI**: add a `Run doctests` step to the `mpc-unittests` **node** leg. That leg already compiles `mpc-node` and most of the workspace under the `test-release` profile, so the doctest run reuses those artifacts rather than paying for a fresh build or a new job/cache.
- **Local parity**: add a `doctests` cargo-make task and wire it into `check-all`, so `cargo make check-all` catches doctest breakage locally too.

A full `cargo test --doc --workspace --all-features` run confirms the contract block was the only broken doctest in the repo; everything else is green.

## Note / possible follow-up

A PR that changes *only* `.md` files is classified docs-only by `changes.yml` and skips all test jobs, including this doctest step. Since the two crate READMEs are doctest *sources*, a doctest broken by a README-only edit would still pass PR CI (it would fail on the subsequent `push` to `main`, where `code_changed` defaults to `true`). Left as-is to keep this PR focused; happy to tighten `changes.yml` in a follow-up if wanted.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,6 +160,14 @@ jobs:
         if: needs.changes.outputs.code_changed != 'false'
         run: nix develop --command cargo nextest run --cargo-profile=test-release --all-features --locked ${{ matrix.nextest-args }}
 
+      # nextest does not run doctests, so they are never exercised by the step
+      # above. Run them once (workspace-wide) on the node leg, whose test build
+      # already compiled mpc-node and most of its dependency tree with the same
+      # `test-release` profile, so this reuses those artifacts.
+      - name: Run doctests
+        if: needs.changes.outputs.code_changed != 'false' && matrix.group == 'node'
+        run: nix develop --command cargo test --doc --workspace --all-features --locked --profile test-release
+
   mpc-contract-reproducible-build:
     name: "MPC contract reproducible build"
     needs: changes
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -22,7 +22,7 @@ dependencies = [
 
 [tasks.check-all]
 description = "Run all checks"
-dependencies = ["check-all-fast", "nextest"]
+dependencies = ["check-all-fast", "nextest", "doctests"]
 
 [tasks.check-extra]
 description = "Run extra checks"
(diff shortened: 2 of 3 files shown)


gilcu3:
@barakeinav1 this one seems ready, when you are back please share it in the PR channel so we can get it done

PR #3785, opened 2026-07-13
fix(contract): make mock attestations cleanable via expiry

### LINKED ISSUE #3293: MockAttestation::Valid entries cannot be cleaned from contract storage
### Background

After PR #2994 narrowed `clean_tee_status` to votes-only and added `clean_invalid_attestations` (which removes entries only when `re_verify` returns `Invalid`), `MockAttestation::Valid` entries become uncleanable:

- `MockAttestation::Valid` has no image hash, no cert, no expiry — it passes `re_verify` unconditionally.
- `clean_invalid_attestations` (`crates/contract/src/lib.rs:1666`) only removes entries whose `re_verify` is `Invalid`.
- `clean_tee_status` (post-#2994) only removes votes.

Result: when an operator migrates from a Mock TLS key to a real Dstack key (or to another Dstack key), the Mock entry persists forever in `stored_attestations` and shows up in `get_tee_accounts`.

Observed on `v1.signer-prod.testnet` at epoch 18: `n1-multichain.testnet` has 3 stored entries — one `Mock::Valid` and two `Dstack` — only one of which is the current participant's `sign_pk`. The Mock entry has survived multiple resharings.

### User Story

As an operator / contract consumer, I want `get_tee_accounts` to reflect actual attestations in use, not stale Mock entries that can never be removed.

### Acceptance Criteria

- [ ] A cleanup path exists that can remove stale `MockAttestation::Valid` entries (e.g. give Mocks an expiry, treat any `Mock` entry as "invalid for cleanup purposes" once it's no longer in the participant set, or expose an admin-only `vote_remove_attestation(tls_public_key)`).
- [ ] Re-running the cleanup on a contract in the state above removes the stale Mock entry.

### Resources

- Related: [#1399](https://github.com/near/mpc/issues/1399) — adding timestamps to mock attestations would unblock one fix.
- Related: [#1711](https://github.com/near/mpc/issues/1711) — disallowing mocks for the TEE hard launch reduces but doesn't eliminate the leak (existing entries still need cleanup).
- Cleanup change: PR [#2994](https://github.com/near/mpc/pull/2994).
- Code: `crates/contract/src/lib.rs:1666` (`clean_invalid_attestations`), `crates/contract/src/tee/tee_state.rs` (`reverify_participants`), `crates/mpc-attestation/src/attestation.rs:47` (`MockAttestation::Valid`).

barakeinav1:
Closes #3293

Implements the expiry-based approach from the issue thread: give mock attestations the same expiry window real (Dstack) ones already get, so the existing `clean_invalid_attestations` flow can evict them.

### Contract
- `AcceptedAttestation::mock` now stamps `DEFAULT_EXPIRATION_DURATION_SECONDS` on every accepted mock, converting a bare `MockAttestation::Valid` into an otherwise-unconstrained expiring `WithConstraints` (an explicit caller-set expiry is preserved). No borsh-layout change — `WithConstraints` already exists.
- Existing stored mock entries are migrated during upgrade (`stamp_expiry_on_legacy_mocks`), so the stale `Mock::Valid` entries already on-chain also become cleanable.

### Node (upgrade-flow compatibility)
Assumes the node is upgraded **before** the contract, so the upgraded node must work against **both** the old and new contract.

The node confirmed mock submissions landed by identity (`stored == submitted`). Once the contract stores a submitted `Mock::Valid` as `WithConstraints{expiry}`, identity never matches, so a node on the new contract would treat every mock submit as not-landed and retry until it errored (and my expiry change also makes mocks expire, which triggers resubmission). Fixed:
- `submitted_attestation_landed`: confirm an expiry-carrying mock via the same expiry-changed heuristic already used for Dstack; fall back to identity for the bare `Valid` form stored by older contracts.
- `read_stored_attestation_expiry` (renamed from `read_stored_dstack_expiry`): also return the mock expiry so the pre-submit baseline is accurate.
- Added `VerifiedAttestation`/`MockAttestation::expiry_timestamp_seconds` helpers on the shared dtos.

Compatibility matrix (new node):
- **old contract** → stored `Mock::Valid` (no expiry) → identity match → landed ✓
- **new contract** → stored `WithConstraints{expiry}` → expiry-changed heuristic → landed ✓

### Tests
- mpc-attestation: `with_expiry__should_*` (Valid→expiring constraints, fill missing expiry + keep other constraints, preserve explicit expiry, leave Invalid unchanged); updated `valid_mock_attestation_succeeds_verification` for the new stored shape
- contract: `clean_invalid_attestations__should_remove_accepted_mock_valid_after_expiry`, `stamp_expiry_on_legacy_mocks__should_make_valid_mock_cleanable`
- node: `submitted_attestation_landed__should_confirm_mock_with_changed_expiry`, `..._should_reject_mock_with_unchanged_expiry` (old-contract identity path still covered by the existing mock tests)

### Out of scope / open question
The genesis sentinel path (`with_mocked_participant_attestations`, used by `init`/`init_running`) still stores non-expiring `Valid` mocks. The migration converts existing ones, but fresh inits create new non-expiring entries. Stamping those too would change the #1087 placeholder semantics (participants would auto-expire if they never submit a real attestation), so I left it out of this PR.


crates/node/src/indexer/tx_sender.rs
```diff
@@ -207,7 +209,15 @@ fn submitted_attestation_landed(
         (VerifiedAttestation::Dstack(stored), Attestation::Dstack(_)) => {
             attestation_expiry_changed(pre_submit_expiry, stored.expiry_timestamp_seconds)
         }
-        (VerifiedAttestation::Mock(stored), Attestation::Mock(submitted)) => stored == submitted,
+        (VerifiedAttestation::Mock(stored), Attestation::Mock(submitted)) => {
+            match stored.expiry_timestamp_seconds() {
+                Some(expiry) => attestation_expiry_changed(pre_submit_expiry, expiry),
```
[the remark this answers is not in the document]
barakeinav1: Valid, but pre-existing and out of scope here: the `None ⇒ landed` behavior and the read-error→`None` mapping predate this PR and apply equally to Dstack — this PR just routes mocks through the same path. Low severity (self-heals on the next check). Better addressed as a separate hardening (the read-error-vs-no-baseline enum) rather than in this PR.

crates/near-mpc-contract-interface/src/types/attestation.rs
```diff
@@ -75,6 +75,18 @@ pub enum VerifiedAttestation {
     Mock(MockAttestation),
 }
 
+impl VerifiedAttestation {
+    /// The stored expiry timestamp, if the attestation carries one. `Dstack`
+    /// entries always do; a `Mock` entry does only when it was stamped with an
+    /// expiry — contracts predating #3293 store `Mock::Valid` without one.
```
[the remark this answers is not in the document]
barakeinav1: Good catch — fixed in a8749941. Reworded so `None` means "unstamped mock (older contract **or** genesis sentinel)", not "old contract only".

crates/node/src/indexer.rs
```diff
@@ -444,9 +444,9 @@ impl ReadSupportedForeignChain for RealForeignChainPolicyReader {
 }
 
 pub(crate) trait ReadAttestationExpiry: Send + Sync {
-    /// The Dstack attestation expiry currently stored for `tls_public_key`, or `None` if none is
-    /// stored.
-    fn read_stored_dstack_expiry<'a>(
+    /// The attestation expiry currently stored for `tls_public_key`, or `None` if none is stored or
+    /// the stored attestation carries no expiry (a mock stored by a contract predating #3293).
```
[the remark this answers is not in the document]
barakeinav1: Fixed in a8749941 — same clarification on the trait doc (unstamped mock can be an older contract or a genesis sentinel).

barakeinav1:
Thanks — dispositions per finding (addressed in a8749941):

- **`with_expiry` preserves caller expiry → resubmit loop** — now stale: `with_expiry` was changed to **cap** (not preserve) the expiry, so the contract owns the max lifetime. The residual corner only affects a caller resubmitting a *fixed* `WithConstraints` expiry (caller misuse); prod submits `Mock::Valid`, whose stamped expiry advances each resubmit and always confirms. No change.
- **raw `+` → `checked_add`** — declined: matches the sibling `AcceptedAttestation::dstack`, overflow is unreachable (timestamp + 1 day), and it's engineering-standards exception #3. Changing only these would diverge from `dstack`.
- **redundant `if let Mock` → `expect(...)`** — declined: the defensive `if let` is exactly what "Maintain Local Reasonability" prescribes; `expect()` would add a panic that depends on the distant `.filter()`, which that section warns against.
- **migration doc / genesis sentinels** — fixed: doc now says it stamps *all* stored mocks, incl. genesis sentinels.
- **double `set_block_timestamp` in the legacy-mock test** — fixed: dropped the redundant second call, comment corrected.
- **`fn mock` docstring intra-doc link** — fixed: `Valid` is now `[`MockAttestation::Valid`]`.
- **"collect keys" comment** — kept: it's a genuine *why* (can't mutate `IterableMap` while iterating).
- **`TODO(#3786)` wording** — kept as-is (verified correct).

barakeinav1:
Keeping `fix`: this resolves a reported defect (#3293 — stored mock entries could never be cleaned). `feat` is defensible, but `fix` matches the bug it closes.

crates/node/src/indexer/tx_sender.rs
```diff
@@ -192,11 +192,13 @@ fn attestation_expiry_changed(pre_submit_expiry: Option<u64>, stored_expiry: u64
 
 /// Whether the attestation we submitted is now the one stored on chain.
 ///
-/// Mock is matched by identity. A Dstack entry keeps no stored per-submission identity, so it is
-/// confirmed via [`attestation_expiry_changed`]: an accepted submit re-stamps the entry's expiry
-/// (to the submit block time plus
+/// An entry that carries an expiry keeps no stored per-submission identity, so it is confirmed via
```
haiyuechen-nearone: nit: Is this comment 100% correct? My understanding now is that:
- Real `Dstack` attestations have identity (PublicKey) + expiry
- ?? New `Mock::Valid` attestations that include an expiry also don't have identity ??
- Old `Mock::Valid` and `Mock::Invalid` attestations don't have identity

I would love to get some input from others that have more knowledge around this

haiyuechen-nearone (dismissed):
I think the general flow LGTM, left a question about the comments in `crates/node/src/indexer/tx_sender.rs`

haiyuechen-nearone (dismissed):

crates/node/src/indexer/tx_sender.rs
netrome: I'm not completely following this, what does "identity" mean here? From the previous comment I percieved this as the "identity equality", i.e. we just check that the stored attestation equals the submitted one for mock attestations - but for Dstack attestations, the identity check is not possible since the verified and stored attestations have different layouts so we just check that the expiry changed to see if the submission had any effect.

So from this perspective, the comment does seem invalid.

netrome: I think we could just simplify this whole docstring to explaining that a changed expiry is sufficient to observe to conclude that the submit attestation call landed. The legacy fallback TODO comment is sufficient to explain the legacy identity check.

crates/contract/src/tee/tee_state.rs
```diff
@@ -473,6 +473,46 @@ impl TeeState {
...
+    /// Legacy [`MockAttestation::Valid`] entries pass re-verification forever and
+    /// can therefore never be evicted by [`TeeState::clean_invalid_attestations`]. [`MockAttestation::with_expiry`]
+    /// rewrites them as expiring [`MockAttestation::WithConstraints`] mocks so the
+    /// normal cleanup flow can remove stale entries once the window elapses. An
+    /// entry whose expiry is longer than (or missing) the default window is capped
+    /// at it; a shorter existing expiry is left as-is.
+    ///
+    // TODO(#3978): transitional — remove this and its migration call site once the
+    // pre-expiry state migration is retired.
+    pub(crate) fn stamp_expiry_on_legacy_mocks(&mut self, current_timestamp_seconds: u64) {
```
netrome: This is a migration helper function, so it should live in the migration module `v3_13_0_state` imo.

crates/mpc-attestation/src/attestation.rs
```diff
@@ -84,10 +84,19 @@ impl AcceptedAttestation {
...
         Self {
-            attestation: VerifiedAttestation::Mock(mock_attestation.clone()),
+            attestation: VerifiedAttestation::Mock(
+                mock_attestation
+                    .clone()
+                    .with_expiry(expiry_timestamp_seconds),
+            ),
             advisory_ids: Vec::new(),
         }
     }
```
netrome, line 102: Hmm definitely not a fan of overwriting the inner expiry timestamp on the `verify` call. This feels like it opens the door to bugs and risks, but I see you raised https://github.com/near/mpc/issues/4005 to tackle this so I won't consider it a hard blocker.

netrome (changes requested):
The change looks correct, but it feels hacky to overwrite expiry times on mock attestations on the `verify` call. That alone is not a hard blocker though, but the migration helper should be moved and the comment on `submitted_attestation_landed` should be simplified so it's not confusing people navigating the code base.

crates/node/src/indexer/tx_sender.rs
barakeinav1: Simplified the docstring in d9943892 — dropped the confusing "identity" framing. It now just says: an accepted submit re-stamps the expiry (only the owner can rewrite it), so a **changed stored expiry** is enough to conclude the submit landed (Dstack + expiry-carrying mocks); the legacy no-expiry mock falls back to an equality check, covered by the inline `TODO(#3786)`. (To your question, @haiyuechen-nearone: there's no per-submission "identity" field for Dstack or stamped mocks — that phrasing was the source of confusion, now gone.)

crates/contract/src/tee/tee_state.rs
barakeinav1: Agreed — moved `stamp_expiry_on_legacy_mocks` (and its unit test) into the `v3_13_0_state` migration module in d9943892. It now lives with the version-specific migration code and is removed together with the module when the migration is retired.

crates/mpc-attestation/src/attestation.rs
barakeinav1: I don't think the location of this is wrong.
After verifying an attestation, we store it. In this case we cap the expiry at our default `now + DEFAULT_EXPIRATION_DURATION_SECONDS` (if the submitted one was larger). This is deliberate: it stops a submitter from storing an arbitrarily long, uncleanable expiry (the goal of this PR).
For dstack we always set `now + DEFAULT_EXPIRATION_DURATION_SECONDS`, since extracting the actual expiry from the cert chain was hard (#1639).

#4005 is meant to align both dstack/mock to the same logic.

netrome: Ah right I see the `.with_expiry` method performs a min between the existing expiry and the cap. That's a bit confusing though. I'd expect `.with_expiry` to be a plain setter. Perhaps worth renaming it to `.cap_expiry()` or something similar?

netrome (dismissed):
Thanks for updating 🙏

crates/mpc-attestation/src/attestation.rs
barakeinav1: Good call — renamed `with_expiry` → `with_expiry_capped_at` in 1e4b6638 (method + both call sites + tests), so it no longer reads as a pure setter. If #4005 later makes it a true override (contract owns the expiry outright, Dstack-style), it'd go back to a plain `with_expiry`.

haiyuechen-nearone (approved):

netrome (approved):
Thanks for renaming 🙏

Issue #3720, opened 2026-07-01
Trim `VerificationContext` to carry only `TcbInfo` (drop `quote`/`collateral`)


pbeza:
`VerificationContext` (`crates/contract/src/tee/verification_context.rs`) carries the full `DstackAttestation` — including `quote` and `collateral` (~2-8 KiB combined) — through the `resolve_verification` callback arg, but only `tcb_info` is read post-DCAP (`DstackAttestation::verify_with_report` plus the verifier-returned `VerifiedReport`). `quote`/`collateral` are sent once to the verifier's `verify_quote` call and never read again on-chain, so they inflate the serialized callback args (gas / receipt size) of every async submission for no benefit.

Trim `VerificationContext.attestation` to just `tcb_info: TcbInfo`.

### Approach
- Replace `VerificationContext.attestation: DstackAttestation` with `tcb_info: TcbInfo`.
- Re-key the post-DCAP path on `&TcbInfo`: the helpers in `crates/attestation/src/attestation.rs` already take `tcb_info`/`report` as params, so `verify_with_report` can accept `&TcbInfo` with the `&self` wrapper kept for `local-verify` callers.
- Route `resolve_verification` (lib.rs) to build/consume `tcb_info` instead of the full attestation.

### Notes
- Raised by a review comment on #3714 (Copilot). Re-scoped from the original yield/`PendingAttestation` design, which #3766 replaced with the promise-chain + `VerificationContext` callback arg.

PR #3788, opened 2026-07-13
chore: try smaller struct in callback

gilcu3:


gilcu3:
tracked already in #3720

Issue #3789, opened 2026-07-13
Make 3.13.2 patch release — nearcore 2.13.1 (both networks)


barakeinav1:
### Background

Per nearcore (Anton), **wait for nearcore `2.13.1`** (CODE_YELLOW, releasing later today) rather than `2.13.0-rc.3`. 2.13.1 includes the rc.3 state-sync change, so it's compatible with **both testnet and mainnet** (confirmed).

Plan:
- Ship a **3.13.2 patch release** off `release/v3.13`, bumping nearcore `2.13.0-rc.2 → 2.13.1`, rolled out to **both networks**.
- Also bump **`main`** to `2.13.1`.

nearcore 2.13 rejects the deprecated centralized `ExternalStorage` state sync (`neard` exits on startup), so the release must also carry the `ExternalStorage → Peers` migration (#3795 / #3796), cherry-picked to the release branch in #3807.

> ⏳ **Blocked:** nearcore `2.13.1` is not published yet. As of now only `2.13.0` / `-rc.1..3` exist. The dependency bump can't land until the tag is available.

### Scope

- Bump nearcore `2.13.0-rc.2 → 2.13.1` on **`release/v3.13`** (for 3.13.2) **and** on **`main`**.
- Include the state-sync migration (#3796 → cherry-pick #3807) on `release/v3.13`.
- Confirm node-only (no contract change).

### Timeline / rollout

- **Tight timeline — release and roll out ASAP once 2.13.1 is published.**
- Roll out to **both testnet and mainnet**; ask operators to update **both** nodes in one message.
- Dev-mainnet cluster dry-run (seed an `ExternalStorage` config so the migration is exercised — mpc-private#475), then `n1-multichain` canary, then fleet one-at-a-time.

### Acceptance Criteria

- [ ] nearcore `2.13.1` published (blocking).
- [ ] nearcore bumped to `2.13.1` on `release/v3.13`; 3.13.2 built and reproducible.
- [ ] nearcore bumped to `2.13.1` on `main`.
- [ ] State-sync migration present on `release/v3.13` (#3807).
- [ ] Diff confirmed node-only (no contract change).
- [ ] 3.13.2 node + launcher hashes voted in.
- [ ] Rolled out to **testnet and mainnet**.

### Resources

- Migration: #3795 / #3796 · release cherry-pick: #3807 · main bump: #3808 (#3817).
- Runbook + fleet audit: mpc-private#475. Release process: `RELEASES.md`.

[...]

PR #3796, opened 2026-07-13
fix(node): migrate state_sync ExternalStorage → Peers on start

### LINKED ISSUE #3795: Migrate all nodes' state_sync ExternalStorage → Peers before nearcore 2.13.0 (blocks #3789)
### Background

Blocks #3789 (nearcore 2.13.0 / 3.13.2). That issue's third acceptance criterion assumes `state_sync.sync = Peers` is "likely a no-op; verify". **It is not a no-op.** nearcore 2.13.0 (`fix(state-sync): reject external-storage state sync config`, nearcore#16025) rejects the deprecated centralized `ExternalStorage` state sync and `neard` **exits on startup** if a node's `config.json` still has it. We currently have two divergent config-writing paths, and only one of four cases actually writes `Peers`.

### The two init paths

- **Non-TDX** — `Dockerfile-node` / `Dockerfile-node-gcp`, `CMD start.sh` → `mpc-node start` (`near_init: None`). `start.sh` owns the near config and its `update_near_node_config` runs on **every boot**, but its non-localnet branch hardcodes **`ExternalStorage`** (`deployment/start.sh:46`) — it was never updated when DSS became the default.
- **TDX** — `Dockerfile-rust-launcher` → `tee-launcher` → `mpc-node start-with-config-file` → `ensure_near_initialized()` → `patch_near_config` writes **`Peers`** (`crates/node/src/config/start.rs:197`), but **only when `config.json` does not yet exist** (`start.rs:56-59` early-returns otherwise). #3533 fixed only this path.

### The 2×2 — actual `state_sync.sync` today

| | Non-TDX (`start.sh`) | TDX (`start-with-config-file`) |
|---|---|---|
| **New config** (fresh init) | `ExternalStorage` ❌ | `Peers` ✅ |
| **Old config** (pre-initialized) | `ExternalStorage` ❌ | `ExternalStorage` ❌ (patch skipped) |

Only TDX + fresh init is on DSS. Any other node upgraded to 3.13.2 (nearcore 2.13.0) would fail to start.

### Fix

- **A — `start.sh` (non-TDX):** set `state_sync.sync = 'Peers'` (shape-agnostic overwrite; idempotent, can't `KeyError` on a config whose default already changed). Since `update_near_node_config` re-runs every boot, both fresh and existing non-TDX nodes converge on restart. Also plumb an optional `tier3_public_addr` (advertised `IP:24567`) — DSS stalls silently without a reachable advertised address.
- **B — `ensure_near_initialized` (TDX / Rust):** on the `config.json`-exists path, migrate `state_sync.sync` → `Peers` in place, leaving the rest of the operator's config untouched. Covers TDX nodes initialized before 3.12.

### Sequencing (single release, per release-prep preference)

A+B ship **in 3.13.2 itself**, ahead of / together with the nearcore bump — no separate release. This is safe because in both paths the config is rewritten **before** nearcore parses it on the same boot (`start.sh` runs `update_near_node_config` before `exec mpc-node start`; `cli.rs` calls `ensure_near_initialized()` before `run_mpc_node()`), so a node upgrading to 3.13.2 self-heals on first start. Trade-off vs. a pre-release: no fleet-wide verification window before `ExternalStorage` support is gone, so the migration + its failure modes (TDX `config.json` writability; `tier3` presence) must be validated on the dev-mainnet cluster and the `n1-multichain` canary (already in #3789's plan).

### Acceptance Criteria

- [ ] `start.sh` writes `state_sync.sync = Peers` for non-localnet (A).
- [ ] `ensure_near_initialized` migrates an existing `ExternalStorage` config to `Peers` on startup (B), scoped to that field only.
- [ ] Unit test covering the `ExternalStorage → Peers` rewrite (pure JSON half).
- [ ] Verified on the dev-mainnet cluster that a node with a pre-existing `ExternalStorage` config starts cleanly on 3.13.2.
- [ ] Confirmed `config.json` is writable at startup on TDX/CVM (sealed-disk constraint applies to app-compose, not this file — verify).

### Follow-up (not blocking)

- `TODO(#2334)`: collapse non-TDX onto `start-with-config-file` so config patching has a single source of truth (root cause of this drift).
- `tier3_public_addr` for old TDX nodes migrated to `Peers` (DSS quality, not a startup blocker).

barakeinav1:
Closes #3795. Blocks #3789 (nearcore 2.13.0 / 3.13.2).

## Why

nearcore 2.13.0 ([near/nearcore#16025](https://github.com/near/nearcore/pull/16025)) rejects the deprecated centralized `ExternalStorage` state sync — `neard` **exits on startup** if `config.json` still has it. #3789 assumed `Peers` was "likely a no-op"; a fleet audit showed otherwise — **all 10 mainnet nodes are on `ExternalStorage`** today (9 non-TDX + 1 TDX). Two divergent config-writing paths keep producing it, and #3533 (which made `Peers` the default) only fixed one of them.

| | Non-TDX (`start.sh`) | TDX (`start-with-config-file`) |
|---|---|---|
| New config (fresh init) | `ExternalStorage` ❌ | `Peers` ✅ |
| Old config (pre-initialized) | `ExternalStorage` ❌ | `ExternalStorage` ❌ (patch skipped) |

## Changes

- **A — `deployment/start.sh` (non-TDX, most of mainnet):** write `state_sync.sync = Peers` for non-localnet, shape-agnostically (handles an `ExternalStorage` block, absent/null `state_sync`, or already-`Peers`; preserves siblings like `parts_compression_lvl`). It runs every boot, so existing nodes converge on restart. Optional `MPC_TIER3_PUBLIC_ADDR` → `network.experimental.tier3_public_addr`, with a non-fatal `INFO` line when unset.
- **B — `crates/node/src/config/start.rs` (TDX):** `ensure_near_initialized` migrates an existing config's `state_sync.sync → Peers` in place and (re)applies `tier3_public_addr`, scoped to those fields (rest of the operator's config untouched). Fresh inits still go through `apply_near_config_patches`, refactored to share the `set_decentralized_state_sync` / `set_tier3_public_addr` helpers.

Both paths rewrite the config **before** nearcore parses it (`start.sh` before `exec mpc-node start`; `ensure_near_initialized()` before `run_mpc_node()`), so a node upgrading to 3.13.2 self-heals on first start — shipped in 3.13.2 itself, no separate release.

## Testing

- **Rust unit tests** over the real captured `ExternalStorage` config: `sync → Peers`, sibling preserved, plus null/absent/already-`Peers`; `tier3` set/unset. Runs in the normal `mpc-unittests` job.
- **`start.sh` transform** against every real config shape; `MPC_TIER3_PUBLIC_ADDR` verified **set** (populates, preserves existing `network` fields) and **unset** (omitted, `INFO` logged).
- **Real nearcore 2.13.0 binary:** `ExternalStorage` → crashes at config validation with the exact reject error; migrated `Peers` → passes into chain init.
- **Container upgrade (operator flow):** new image over a persistent `ExternalStorage` volume → `start.sh` rewrote it to `Peers`. This end-to-end run also caught + fixed a bash-heredoc backtick bug in `start.sh`.

Still missing / follow-up:
- In-container `neard` on a glibc-matched **reproducible 3.13.2 image** (the local test binary was glibc 2.38 vs the `bookworm-slim` base 2.36, so `neard` was exercised on the host instead).
- Path B on a **live TDX node** — the `n1-multichain` canary at rollout.
- **Seeded dev-cluster dry-run** and **`MPC_TIER3_PUBLIC_ADDR` wiring** for the 9 non-TDX mainnet nodes (ops).
- The nearcore 2.13.0 bump itself is #3789; the tests above used a local bump.

## Docs

Operator runbook + the pre-release cluster-status audit live in **near/mpc-private#475** (they reference internal fleet details — node IPs, cluster status, deployment env).


deployment/start.sh
```diff
@@ -43,7 +43,25 @@ config['store']['load_mem_tries_for_tracked_shards'] = True
 if "$MPC_ENV" == "mpc-localnet":
     config['state_sync_enabled'] = False
 else:
-    config['state_sync']['sync']['ExternalStorage']['external_storage_fallback_threshold'] = 0
+    # Decentralized (peer-to-peer) state sync. Centralized (ExternalStorage)
```
barakeinav1, line 46: I thought about extracting this to python file, so it could be unit tested.
but in the end decided not to ( bigger refactor,touches start.sh + both Dockerfiles + python3 in the nix shell), and did a local  node/container run instead

crates/node/src/config/start.rs
```diff
@@ -170,6 +176,73 @@ fn patch_near_config(
...
+    if !(sync_changed || tier3_changed) {
+        return Ok(());
+    }
+
+    let patched =
+        serde_json::to_string_pretty(&config).context("failed to re-serialize NEAR config.json")?;
+    std::fs::write(config_path, patched)
+        .with_context(|| format!("failed to write {}", config_path.display()))?;
+
+    tracing::info!("migrated NEAR config.json to decentralized (peer-to-peer) state sync");
```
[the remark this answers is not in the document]
barakeinav1: Good catch — fixed in 9b5e47cd. The write happens when *either* `state_sync.sync` or `tier3_public_addr` changed, so the single unconditional log was misleading. Now each field logs independently only when it actually changed: "migrated … state_sync to decentralized (Peers)" and/or "set … network.experimental.tier3_public_addr".

barakeinav1:
Thanks — went through all six non-blocking findings. Addressed 1, 2, 4, 6 in 7e76a490; declined 3 and deferred 5 with rationale.

> **non-atomic write (`212-213`)** … temp-file-plus-rename

**Done.** Added a shared `write_near_config_atomic` helper (temp file + `rename`) used by *both* the migration and `patch_near_config`, so neither can leave a truncated `config.json`. Extra worthwhile since the migration runs fleet-wide on the 3.13.2 upgrade.

> **existing node migrated to Peers without tier3 … not warned … consider a `tracing::warn!`**

**Done** (as `INFO`, not `warn`): logged at the migration site when non-localnet + `tier3` unset, symmetric to `start.sh`'s line. Kept it `INFO` to match `start.sh`'s deliberately non-alarmist choice — some nodes are publicly reachable and legitimately don't need it.

> **no end-to-end test of `migrate_near_config_state_sync`**

**Done.** Added tempfile tests: on-disk `ExternalStorage → Peers` with siblings preserved + tier3 applied, idempotent second call, and localnet left untouched.

> **docstring quite long**

**Done.** Trimmed to the load-bearing bits (why it runs every start, the `apply_near_config_patches` reference, no-op-when-migrated).

> **panic if `state_sync` is a scalar (`222-243`)**

**Declined.** It guards a shape nearcore never produces (`serde_json` already turns `null → object`); adding a branch for an impossible input is net noise, and it's within the top-level-init panic carve-out.

> **heredoc env interpolation (`start.sh:58`)**

**Deferred.** It's the pre-existing pattern for *every* `"$VAR"` in `start.sh`, the value is an op-controlled `IP:port`, and fixing only tier3 would be inconsistent. This is exactly what the (already-deferred) `start.sh` → standalone-`.py` extraction would fix wholesale — folding it there.

fmt/clippy clean; 22/22 config tests pass.

crates/node/src/config/start.rs
```diff
@@ -55,6 +55,12 @@ impl StartConfigExt for StartConfig {
         let near_config_path = near_config_file(&self.home_dir);
         if near_config_path.exists() {
             tracing::info!("NEAR node already initialized, skipping init");
+            // Nodes initialized before decentralized state sync became the
+            // default still carry the deprecated centralized `ExternalStorage`
+            // block, which nearcore 2.13 rejects (it exits on startup). Migrate
+            // that one field in place, leaving the rest of the operator's config
+            // untouched. Fresh inits get `Peers` via `apply_near_config_patches`.
+            migrate_near_config_state_sync(&near_config_path, near_init)?;
```
gilcu3: instead of the big comment I would prefer to open an issue so that this code is removed after the next release, then explain the situation on the issue instead. Here we would leave a TODO with a short comment.

gilcu3 (dismissed):
Thank you!

crates/node/src/config/start.rs
barakeinav1: Done in 32281fca. Opened #3801 to track removing the migration once the fleet is off ExternalStorage (post-3.13.2) — it explains the situation and lists what to delete. Replaced the block here with a short `TODO(#3801)`.

gilcu3 (dismissed):

crates/node/src/config/start.rs
```diff
@@ -161,15 +164,103 @@ fn patch_near_config(
...
+    std::fs::rename(&tmp_path, config_path)
+        .with_context(|| format!("failed to replace {}", config_path.display()))?;
+    Ok(())
+}
+
+/// Sets `state_sync.sync` to decentralized (`Peers`), replacing any inherited
+/// value. Returns whether the config was changed.
+fn set_decentralized_state_sync(config: &mut serde_json::Value) -> bool {
+    let peers = serde_json::json!("Peers");
+    if config["state_sync"]["sync"] == peers {
```
netrome: Iiuc this could panic if the config object doesn't have the expected format. I think we should return an error instead.

crates/node/src/config/start.rs
```diff
@@ -161,15 +164,103 @@ fn patch_near_config(
...
+
+/// Sets the advertised `tier3_public_addr` when `near_init` provides one (it is
+/// required for decentralized state sync to be reachable). No-op when unset.
+/// Returns whether the config was changed.
+fn set_tier3_public_addr(config: &mut serde_json::Value, near_init: &NearInitConfig) -> bool {
+    let Some(tier3) = &near_init.tier3_public_addr else {
+        return false;
+    };
+    let tier3 = serde_json::Value::String(tier3.to_string());
+    if config["network"]["experimental"]["tier3_public_addr"] == tier3 {
```
netrome: Same here

netrome (changes requested):
Overall good stuff, but I don't think we should panic and take down the node if the config has an unexpected format.

crates/node/src/config/start.rs
barakeinav1: Fixed in 9f04865b. These now go through a new `object_at` helper that returns an `anyhow` error instead of `[]`-indexing, so an unexpected config shape yields a clean startup error rather than a panic. Applied across the whole method (store/rpc/network had the same `[]` pattern), and `apply_near_config_patches` now returns `Result`. Added tests asserting a non-object field / non-object root errors instead of panicking.

crates/node/src/config/start.rs
barakeinav1: Done — same fix (9f04865b): `set_tier3_public_addr` now navigates via `object_at` and returns an error on an unexpected type instead of panicking on `[]`.

gilcu3 (approved):

netrome (approved):
Thanks for updating 🙏

Issue #3764, opened 2026-07-09
Foreign chain RPC provider health checks: validate config without curated reference data


anodar:
### Background

Foreign chain provider config is only exercised when the first real verify request arrives, so a URL typo or an API key not enabled for that chain becomes a production failure (we have hit both). The one existing check is the offline CLI, and it probes pinned golden transactions, which age out as providers prune history, go stale because they are curated by hand, and cannot cover local networks (#3860).

- [ ] Phase 1: node startup probe, identity only (#4003)
- [ ] Phase 2: CLI reuses the node's probe, golden route deleted (#3969)
- [ ] Phase 3: dynamic transaction discovery in the shared probe (#4004)

### User Story

As a node operator I want logs and metrics on whether each configured RPC provider is reachable, authenticated and on the expected network, at startup and from the CLI, without either depending on curated reference data.

### Acceptance Criteria

- A misconfigured or unusable provider is visible from the node's logs, metrics and debug endpoint shortly after startup, without waiting for a real verify request.
- The offline CLI and the node reach the same verdict for the same config.
- Neither one relies on curated reference data, so both keep working as providers prune history and on networks we ship nothing for.

haiyuechen-nearone:
@anodar Does this issue also include periodic checks against each provider when the node is running? A provider might be healthy at startup but goes down some time later.

PR #3797, opened 2026-07-14
refactor: extract foreign-chain-health-check library from config-tester

haiyuechen-nearone:
Part of #3764. The plan (stacked, merge bottom-up):

1. **#3797 (this PR)** — extract a shared `foreign-chain-health-check` library from the config-tester CLI
2. #3848 — probe every configured provider on node startup + per-provider logs
3. #3849 — per-chain `configured` / `healthy` Prometheus counts
4. #3850 — healthy counts on `/debug/node_config`

Moves the golden vectors, per-provider checks, and `check_all_providers` out of the CLI into a new production crate, so the node can reuse them without depending on the CLI (which pulls in clap/serde_yaml/toml). The extraction itself is behavior-neutral.

Notes for reviewers:
- The CLI keeps arg parsing, config-shape detection, and table rendering.
- To keep `clap` out of the library, `Network` is parsed via a `FromStr` impl (dedicated `ParseNetworkError`) instead of `clap::ValueEnum`, and the CLI takes `--network` as `Option<Network>` directly. Accepted values are unchanged (`mainnet`/`testnet`); the only user-visible difference is the invalid-value error text (now `expected one of: mainnet, testnet`).


crates/foreign-chain-config-tester/src/main.rs
```diff
@@ -42,9 +22,9 @@ struct Args {
     #[arg(long)]
     config: PathBuf,
 
-    /// Network the reference transactions belong to. Auto-detected from the
-    /// config (`chain_id` / `mpc_contract_id`) when omitted.
-    #[arg(long, value_enum)]
+    /// Network the reference transactions belong to. Auto-detected from
+    /// the config (`chain_id` / `mpc_contract_id`) when omitted.
+    #[arg(long, value_name = "NETWORK")]
```
anodar: Why drop value_enum? Auto-detection of contract doesn't always work (e.g. dev clusters), and with value_enum it would probably show possible values [mainnet | testnet].

crates/foreign-chain-health-check/src/lib.rs
```diff
@@ -0,0 +1,537 @@
+//! Foreign-chain RPC provider health checks: probe every configured provider
+//! with a fixed golden request and report a per-provider result.
```
anodar: > with a fixed golden request

Not for SUI though.

anodar (dismissed):
Please have a look at Blocking findings from claude and either address or explain why it doesn't apply.

In this particular case, I believe it's false positive:
```
The new check_all_providers keeps that branch only for sui. Result: the CLI-rendered table previously included a - not configured row for every absent chain; now it silently omits every unconfigured chain except sui.
```
clearly other chains have same exact branch for emitting unconfigured chain log.

crates/foreign-chain-config-tester/src/main.rs
haiyuechen-nearone: I wanted to keep the new `foreign-chain-health-check` crate completely free of `clap`. Since `Network` is moved into the new crate, deriving `value_enum` would pull `clap` into the new crate as well. `CLI` crate is also not allowed to implement `value_enum` from the outside either in the current dependency setup.

I made sure that the custom error string still lists the valid values, and its generated based on the code values so it won't drift.

I do see that this is quite a lot of code to achieve a worse version of what `clap` already offers. Should I instead pull in `clap` as optional dependency in `foreign-chain-health-check`?

haiyuechen-nearone:
> Please have a look at Blocking findings from claude and either address or explain why it doesn't apply.

That review did apply to an earlier version, and the blockers were addressed in [c689402](https://github.com/near/mpc/pull/3797/commits/c6894029911fc788d0b60921e656587364886a2f)
image

Do we only trigger Claude review on the first publish?

anodar:
> Do we only trigger Claude review on the first publish?

you can re-trigger it using @claude review.

crates/foreign-chain-config-tester/src/main.rs
anodar: You can avoid pulling clap in that library crate by having separate CLI enum in main.rs e.g. NetworkArg and then implement from trait for Network (type from foreign chain health check lib) to convert CLI Network Arg type into it.

crates/foreign-chain-health-check/src/lib.rs
gilcu3: is it not possible for SUI, or just not yet implemented?

gilcu3 (dismissed):
Thank you!

crates/foreign-chain-health-check/src/lib.rs
anodar: Not possible for SUI. 

https://github.com/near/mpc/blob/2b539b7c819fe46931890f567453cdd8cc771043/crates/foreign-chain-config-tester/README.md?plain=1#L10

```
Sui is the exception: its providers prune
transactions after a few weeks, so the check instead verifies the provider's
chain identity and inspects a transaction from its latest checkpoint. 
```

haiyuechen-nearone: Added comments about SUI being an exception

crates/foreign-chain-config-tester/src/main.rs
haiyuechen-nearone: I think pulling in optional clap dependency is better here and implemented that [58380d5](https://github.com/near/mpc/pull/3797/commits/58380d580d2f769266bcffae72c897193730aebf). I feel having a separate CLI enum does allow using clap's CLI features, but additionally need to introduce more code to make sure the CLI enum and the lib enum are in sync.

gilcu3 (approved):

anodar (approved):

PR #3680, opened 2026-06-26
feat: Allow changing URLs of nodes

netrome:
Purely Claude-generated - just an example implementation of how we could resolve https://github.com/near/mpc/issues/3677

Issue #1163, opened 2025-09-22
check if deposit is required for migration functions


kevindeforth:
### Background

Check if a deposit is requred for tome of the methods introduced in #1162 

### Acceptance Criteria

add deposit or not

netrome:
Economic security is important. Let's get this one figured out this week.

PR #3806, opened 2026-07-14
feat: support operators changing node urls

### LINKED ISSUE #3677: Support easily changing the node URL
### Background

Currently, there's no convenient way for a node to change their URL. This is annoying if they have a typo or want to change to a new domain for whatever reason. We should make it easy for nodes to update their URLs registered in the contract.

### User Story

As a node operator I'd like to be able to update the registered URL of my node.

### Acceptance Criteria

Node operators can update their URLs easily.

_No response_

gilcu3:
Closes #3677 , supersedes #3680

A small refactor was done in a89dc0f921d57f8dc36477535539e006588ec674 to avoid introducing code duplication in the in-process tests here


crates/contract/src/lib.rs
```diff
@@ -2522,6 +2523,37 @@ impl MpcContract {
...
+        let account_id = Self::assert_caller_is_signer();
+        log!(
+            "update_participant_url: signer={:?}, url={:?}",
+            account_id,
+            url
+        );
+
+        let ProtocolContractState::Running(running_state) = &mut self.protocol_state else {
+            return Err(errors::InvalidState::ProtocolStateNotRunning.into());
+        };
```
gilcu3, line 2539: just to be on the safe side. I don't think this is a hard requirement

crates/node/src/p2p.rs
```diff
@@ -474,13 +479,34 @@ impl SenderConnectionId for IncomingConnection {
...
+/// [`new_tls_mesh_network_with_address_updates`] to let peers' addresses be updated live.
 pub async fn new_tls_mesh_network(
     config: &MpcConfig,
     p2p_private_key: &ed25519_dalek::SigningKey,
 ) -> anyhow::Result<(
     impl MeshNetworkTransportSender + use<>,
     impl MeshNetworkTransportReceiver + use<>,
 )> {
+    new_tls_mesh_network_with_address_updates(config, p2p_private_key, |_| None).await
+}
```
gilcu3, line 498: this method is now used only during initialization and tests, so I am a bit torn about leaving it here or removing, but ended up leaving it

crates/node/src/p2p.rs
```diff
@@ -474,13 +479,34 @@ impl SenderConnectionId for IncomingConnection {
...
+    config: &MpcConfig,
+    p2p_private_key: &ed25519_dalek::SigningKey,
+    resolve_peer_address: F,
+) -> anyhow::Result<(
+    impl MeshNetworkTransportSender + use<F>,
+    impl MeshNetworkTransportReceiver + use<F>,
+)>
+where
+    F: Fn(ParticipantId) -> Option<String> + Clone + Send + Sync + 'static,
+{
```
gilcu3, line 514: I changed from the original draft #3680 that was using a channel here as this way seems fitter to the purpose and allows a cleaner separation of concerns

crates/node/src/coordinator.rs
```diff
@@ -836,14 +845,74 @@ impl Drop for ReportCurrentJobGuard {
...
+/// address/port change is hot-swapped ([`peer_address_from_state`]); only a change to threshold,
+/// identity, or our *own* listening port (which re-binds the listener) needs a restart.
+fn participants_change_requires_restart(
+    old: &ParticipantsConfig,
+    new: &ParticipantsConfig,
+    my_near_account_id: &AccountId,
+) -> bool {
+    if old.threshold != new.threshold {
+        return true;
+    }
```
gilcu3, line 867: this is the status quo, but I believe that changing this, which is the governance threshold, does not require any restart. If reviewers agree we should add an issue for it to be handled later

gilcu3: well of course, this is assuming the per domain threshold lands, so currently it is still correct

crates/node/src/coordinator.rs
```diff
@@ -836,14 +845,74 @@ impl Drop for ReportCurrentJobGuard {
...
+    };
+    if identities(old) != identities(new) {
+        return true;
+    }
+
+    let my_port = |cfg: &ParticipantsConfig| {
+        cfg.get_info_by_account_id(my_near_account_id)
+            .map(|p| p.port)
+    };
+    my_port(old) != my_port(new)
```
gilcu3, line 896: this is important in the case we don't have a port override. Afaik we do, but we still need this in case we didn't

crates/contract/src/lib.rs
```diff
@@ -2522,6 +2523,32 @@ impl MpcContract {
         Ok(())
     }
 
+    /// Updates the calling participant's registered URL, keeping the TLS key and participant ID.
+    #[handle_result]
+    pub fn update_participant_url(&mut self, url: String) -> Result<(), Error> {
+        // TODO(#1163): require a deposit
```
anodar, line 2529: Please reword the issue to include this endpoint as well. Currently it only refers to migration functions.

crates/contract/tests/inprocess/common.rs
```diff
@@ -0,0 +1,105 @@
...
+    VMContextBuilder::new()
+        .signer_account_id(account_id.clone())
+        .predecessor_account_id(account_id.clone())
+        .block_timestamp(near_sdk::env::block_timestamp())
+        .build()
+}
+
+/// Builds a contract already in `Running` with a single Sign domain and `participant_count`
+/// participants.
+pub fn build_running_contract(
```
anodar: nit: this could just be named ~new()~ init_contract (see [comment](https://github.com/near/mpc/pull/3806#discussion_r3595633477)).

crates/contract/src/lib.rs
netrome: Hmm yeah on one hand if we'd ever get stuck in resharing because of this it would be nice to be able to update URLs. For example, if a node migrates to an invalid but correct URL I'm pretty sure we'd get stuck in resharing. Being able to change the URL rather than having to abort the resharing would be nice. But it's not strictly necessary.

crates/contract/src/lib.rs
```diff
@@ -2522,6 +2523,32 @@ impl MpcContract {
...
+        let Some(existing_info) = running_state.parameters.participants().info(&account_id) else {
+            return Err(errors::InvalidState::NotParticipant { account_id }.into());
+        };
+
+        let new_info = ParticipantInfo {
+            url,
+            tls_public_key: existing_info.tls_public_key.clone(),
+        };
+        running_state.parameters.update_info(account_id, new_info)
+    }
```
netrome, line 2550: One consideration here, if we have any pending votes wouldn't we like to update those in-place as well?

crates/node/src/coordinator.rs
```diff
@@ -963,3 +1038,296 @@ fn make_initializing_stop_fn(
...
+            "carol.near",
+            "carol.example.com",
+            7000,
+            3,
+        ));
+        let state = resharing(base_config(), new_participants.clone(), 5);
+
+        // When / Then
+        assert_eq!(state.mesh_participants(), Some(&new_participants));
+    }
```
netrome: Shouldn't this test live in `indexer/participants.rs`?

crates/node/src/coordinator.rs
netrome: Yeah just changing the governance threshold should not require restart once per-domain thresholds land.

netrome: So this is effectively conflicting with https://github.com/near/mpc/pull/3640? How can we make sure we don't forget to update this? Add a TODO and tracking issue or could we formulate a test case that fails with per-domain thresholds here? The latter sounds non-trivial.

crates/node/src/p2p.rs
netrome: Yeah I was also thinking about the same thing. Feels ok to leave it.

crates/node/src/p2p.rs
netrome: Nice!

netrome (dismissed):
Nice stuff! My main consideration is if we should overwrite votes or now. I'm not sure though, and if we do, it's probably nice to leave as a follow-up.

crates/node/src/coordinator.rs
gilcu3: I wouldn't call it a conflict because this is the status quo and #3640 was certainly not going to touch this, but the sensitive thing if we do want to change it is to add a follow up issue and TODO, will do.

crates/contract/src/lib.rs
gilcu3: I don't like too much the idea of changing votes because another operator decided to change its url. As we discussed in slack, seems a bit better to avoid having the URLs in the votes, but that is a bigger change.

crates/node/src/coordinator.rs
netrome: Yeah and I'd love to get rid of that port override.

crates/contract/src/lib.rs
netrome: Yeah agree, let's leave it for now. The model is still consistent. The only risk is that a vote may revert back your URL, but if there's a pending vote you can always update your URL in your vote and ask everyone to vote again.

crates/contract/tests/inprocess/update_participant_url.rs
```diff
@@ -0,0 +1,89 @@
...
+use near_account_id::AccountId;
+use near_mpc_contract_interface::types::{
+    ParticipantInfo as DtoParticipantInfo, ProtocolContractState,
+};
+use near_sdk::testing_env;
+use std::str::FromStr;
+
+use assert_matches::assert_matches;
+
+fn participant_info(rc: &RunningContract, account_id: &AccountId) -> DtoParticipantInfo {
```
anodar: nit: you only use contract from this (RunningContract) input, why not just accept contract or contract state ?

netrome (dismissed):
Thanks! 🙏

crates/node/src/coordinator.rs
gilcu3: yeah, it is certainly weird

crates/contract/tests/inprocess/attestation_submission.rs
```diff
@@ -79,99 +77,31 @@ impl TestSetupBuilder {
...
-            keyset,
-            parameters.clone().into(),
-            init_config.clone(),
-        )
-        .unwrap();
+        let common::RunningContract {
+            contract,
+            participants: participants_list,
+            parameters,
+        } = common::build_running_contract(participant_count, threshold, self.init_config);
```
anodar: nit: I like that you factored out code in helper methods but `RunningContract` itself is never actually used, rather destructured immediately. It's not needed outside this file either (used in update_participant_url.rs‎ but not needed).

I think it would be cleaner if you just do:
```
 let participants = gen_participants(participant_count);
 let parameters = ThresholdParameters::new(participants, Threshold::new(threshold))
let contract = init_contract(participants, parameters)
```

anodar (dismissed):
LGTM. 


One question though: why not propagate participants instead of whole contract state in channel ?

crates/contract/src/lib.rs
gilcu3: ahh right, you mean #1163, yes I can add a comment there

gilcu3: done, although I am not afraid of dangling TODOs because our CI would catch it :)

crates/contract/tests/inprocess/attestation_submission.rs
gilcu3: let me give that a try

gilcu3: addressed in [0b1b97f](https://github.com/near/mpc/pull/3806/commits/0b1b97f80d168168795bde979c9d1ab8ce020078)

crates/contract/tests/inprocess/common.rs
gilcu3: addressed in [0b1b97f](https://github.com/near/mpc/pull/3806/commits/0b1b97f80d168168795bde979c9d1ab8ce020078)

crates/contract/tests/inprocess/update_participant_url.rs
gilcu3: addressed in [0b1b97f](https://github.com/near/mpc/pull/3806/commits/0b1b97f80d168168795bde979c9d1ab8ce020078)

anodar (approved):

netrome (approved):

PR #3815, opened 2026-07-15
test: add test vectors for CKD

### LINKED ISSUE #3351: Add test vectors for CKD
### Background

CKD is a new cryptographic functionality that others might want to reproduce and verify, potentially in other languages. For this reason, it would be nice if our implementation provided some test vectors.

### User Story

_No response_

### Acceptance Criteria

We provide test vectors for CKD, for both variants implemented (public and private verifiability).

[thread](https://nearone.slack.com/archives/C0B12RKBSAV/p1779619937654779)

gilcu3:
Closes #3351


crates/threshold-signatures/src/confidential_key_derivation.rs
```diff
@@ -4,7 +4,8 @@
 //! which allows a client to derive a unique key for a specific application without revealing
 //! the application identifier to the key derivation service.
 //!
-//! The protocol is based on a combination of Oblivious Transfer (OT) and Diffie-Hellman key exchange.
+//! The protocol is a threshold BLS signature over BLS12-381, blinded with EC-ElGamal
+//! encryption so no single node ever sees the derived key.
```
gilcu3, line 8: this is a broken window fix. We had already fixed it elsewhere, but missed it here

pbeza (dismissed):

crates/threshold-signatures/src/confidential_key_derivation.rs
SimonRastikian: Ouuupsss

crates/threshold-signatures/tests/ckd_test_vectors.rs
```diff
@@ -0,0 +1,385 @@
+//! Test vectors for Confidential Key Derivation (CKD).
+//!
+//! The committed vectors live in `tests/vectors/ckd_test_vectors.json` (see its
+//! `_encoding` block for the serialization conventions). Each fixes the scalars
+//! `msk`, `a`, `y` and an `app_id`, and records every intermediate and final
+//! value so an independent implementation can reproduce and check the scheme.
+//!
+//! `generate_ckd_test_vectors` (ignored) regenerates the JSON; the other tests
+//! verify the committed values against the public crate API.
```
SimonRastikian, line 9: Nit: I think most of this could be deleted

crates/threshold-signatures/tests/ckd_test_vectors.rs
```diff
@@ -0,0 +1,385 @@
...
+
+fn g2_from_hex(hex_compressed: &str) -> G2Projective {
+    let bytes: [u8; 96] = hex::decode(hex_compressed)
+        .expect("g2 hex")
+        .try_into()
+        .expect("g2 point is 96 bytes");
+    G2Projective::from_compressed(&bytes)
+        .into_option()
+        .expect("valid compressed G2 point")
+}
```
SimonRastikian, line 104: Was wondering whether these two could be unified under the same macro rule:
```rust
macro_rules! impl_from_hex {
    ($name:ident, $ty:ty, $len:expr) => {
        fn $name(hex_compressed: &str) -> $ty {
            let bytes: [u8; $len] = hex::decode(hex_compressed)
                .expect("invalid hex")
                .try_into()
                .expect("incorrect compressed point length");

            <$ty>::from_compressed(&bytes)
                .into_option()
                .expect("invalid compressed point")
        }
    };
}
```

Which then can be used as:
impl_from_hex!(g1_from_hex, G1Projective, 48);
impl_from_hex!(g2_from_hex, G2Projective, 96);


Or is it too complicated?

crates/threshold-signatures/tests/ckd_test_vectors.rs
```diff
@@ -0,0 +1,385 @@
...
+        .final_exponentiation()
+        .is_identity()
+        .into()
+}
+
+#[test]
+#[expect(non_snake_case)]
+fn ckd_test_vectors__should_derive_expected_public_values() {
+    // Given
+    let vectors = load_vectors();
```
SimonRastikian, line 196: Are the vectors generated using a real CKD instance?

crates/threshold-signatures/tests/ckd_test_vectors.rs
```diff
@@ -0,0 +1,385 @@
...
+enum ScalarSpec {
+    Label(&'static str),
+    Hex(&'static str),
+}
+
+fn resolve_scalar(spec: &ScalarSpec) -> ([u8; 32], Scalar) {
+    let be: [u8; 32] = match spec {
+        ScalarSpec::Label(label) => {
+            let mut be: [u8; 32] = Sha256::digest(label.as_bytes()).into();
+            be[0] &= 0x3f;
```
SimonRastikian, line 312: Why is this operation necessary?

crates/threshold-signatures/tests/ckd_test_vectors.rs
```diff
@@ -0,0 +1,385 @@
...
+        ),
+        // High-bit scalars: exercises little-endian encoding and canonical
+        // reduction (msk = q - 1, the largest valid scalar).
+        (
+            "ckd-vector-4/app_id",
+            ScalarSpec::Hex(MAX_SCALAR),
+            ScalarSpec::Hex("6fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"),
+            ScalarSpec::Hex("5abbccddeeff00112233445566778899aabbccddeeff00112233445566778899"),
+        ),
+    ];
```
SimonRastikian, line 413: It's not clear to me what's happening here. Can you enlighten me?

crates/threshold-signatures/tests/ckd_test_vectors.rs
gilcu3: For someone wanting to use the test vectors this comment is helpful. Therefore, I am inclined to leave it.

crates/threshold-signatures/tests/ckd_test_vectors.rs
gilcu3: Not too complicated, but I prefer to stay away from macros, they make code harder to understand and compiling slower. For two such short functions it is an overkill

crates/threshold-signatures/tests/ckd_test_vectors.rs
gilcu3: ~~yes, there is another method to generate them~~

EDIT: reimplemented in [2825b47](https://github.com/near/mpc/pull/3815/commits/2825b472c9f3b4f4c8c397417b4d52443cb2b680)

crates/threshold-signatures/tests/ckd_test_vectors.rs
gilcu3: this are the inputs for the test vectors generators. Some inputs are fixed to cover corner cases I thought were relevant

crates/threshold-signatures/tests/ckd_test_vectors.rs
gilcu3: to guarantee that the generated scalar is less than the group order. It basically clears the highest bits to achieve it

crates/threshold-signatures/src/test_utils/dkg.rs
```diff
@@ -129,13 +142,22 @@ pub fn assert_public_key_invariant<C: Ciphersuite>(
...
+}
+
+/// Like [`generate_test_keys`] but pins the polynomial's constant term to
+/// `secret`, so the reconstructed master secret is known in advance.
+pub fn generate_test_keys_with_secret<C: Ciphersuite>(
+    secret: Scalar<C>,
+    degree: usize,
+    rng: &mut impl CryptoRngCore,
+) -> (Polynomial<C>, VerifyingKey<C>) {
+    let f = Polynomial::<C>::generate_polynomial(Some(secret), degree, rng).unwrap();
```
SimonRastikian, line 160: Followup: Those two functions could most probably be used in other tests btw.

SimonRastikian (approved):
Cool, much better thanks!

pbeza (approved):

PR #3816, opened 2026-07-15
chore: bump cargo near and use its new check feature

### LINKED ISSUE #3814: Bump cargo-near to 0.22.0 / cargo-near-build 0.11.5 and adopt new features
### Background

We pin `cargo-near-build = "0.11.3"` (`Cargo.toml:98`) and the `sourcescan/cargo-near:0.21.1-rust-1.93.0` image (`crates/contract/Cargo.toml:21`, `crates/tee-verifier/Cargo.toml:11`, `crates/chain-gateway-test-contract/Cargo.toml:12`). Latest is `cargo-near-build` 0.11.5 / `cargo-near` 0.22.0.

Relevant changes:
- **#447** isolates ABI generation into its own `CARGO_TARGET_DIR` — the artifact collision our harness works around by hand-assigning `out_dir`s (`crates/test-utils/src/contract_build.rs`, `crates/contract/tests/sandbox/utils/contract_build.rs`).
- **#442** adds `cargo near check`, which type-checks/lints the contract under the real build env (`--cfg near`, wasm32, matching features/profile). Our current `cargo clippy --all-targets` (`Makefile.toml:49`) sets neither `--cfg near` nor the wasm target, so the contract is linted in the wrong configuration (near-sdk 5.28 requires `--cfg near`).
- **#441** prints built contract size; **#448** clears fresh RustSec advisories.

The Docker-image bump changes the reproducible build environment, so the ABI snapshot (`test_abi_has_not_changed`) and embedded `build.builder` metadata will shift and need re-accepting.

### Acceptance Criteria

- `cargo-near-build` bumped to 0.11.5 in `Cargo.toml`.
- `sourcescan/cargo-near` image bumped to the 0.22.0 tag in all three `Cargo.toml`s.
- ABI snapshot (`test_abi_has_not_changed`) updated for the new build environment.
- A `cargo near check --clippy` step added for the contract crates so they are linted in the correct build config.

gilcu3:
Closes #3814 

~~Notice we are not bumping the images from sourcescan we use for reproducible builds, because they do not support rust 1.93 anymore for their cargo near 0.22 versions. This has not been a problem in the past, but if possible we should ask them to publish that image as well to keep everything in sync.~~

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -18,6 +18,7 @@ dependencies = [
   "check-docs",
   "clippy",
   "clippy-all-features",
+  "contract-check",
 ]
 
 [tasks.check-all]
@@ -47,6 +48,23 @@ args = [
 [tasks.clippy]
 command = "cargo"
 args = ["clippy", "--all-targets", "--locked", "--", "-D", "warnings"]
+
+# The wasm32 + `--cfg near` counterpart of the `clippy` task: it lints the contract
+# crates in the build config they actually deploy under, which plain `cargo clippy`
+# (host target, no `--cfg near`) can't reach. `RUSTFLAGS=-Dwarnings` makes it a gate;
+# Cargo caps lints on non-workspace deps, so only our crates can fail.
+[tasks.contract-check]
+description = "Lint contract crates under the real build config via `cargo near check`"
+script = '''
+#!/usr/bin/env bash
+set -euo pipefail
+# Contract crates are exactly the workspace cdylibs; ask Cargo so the list self-maintains.
+cargo metadata --no-deps --format-version 1 \
+  | jq -r '.packages[] | select(any(.targets[].crate_types[]; . == "cdylib")) | .manifest_path' \
+  | while read -r manifest; do
+      cargo near check --clippy --locked --env 'RUSTFLAGS=-Dwarnings' --manifest-path "$manifest"
+    done
+'''
 
 [tasks.fmt]
 command = "cargo"
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -49,22 +49,10 @@ args = [
 command = "cargo"
 args = ["clippy", "--all-targets", "--locked", "--", "-D", "warnings"]
 
-# The wasm32 + `--cfg near` counterpart of the `clippy` task: it lints the contract
-# crates in the build config they actually deploy under, which plain `cargo clippy`
-# (host target, no `--cfg near`) can't reach. `RUSTFLAGS=-Dwarnings` makes it a gate;
-# Cargo caps lints on non-workspace deps, so only our crates can fail.
 [tasks.contract-check]
 description = "Lint contract crates under the real build config via `cargo near check`"
-script = '''
-#!/usr/bin/env bash
-set -euo pipefail
-# Contract crates are exactly the workspace cdylibs; ask Cargo so the list self-maintains.
-cargo metadata --no-deps --format-version 1 \
-  | jq -r '.packages[] | select(any(.targets[].crate_types[]; . == "cdylib")) | .manifest_path' \
-  | while read -r manifest; do
-      cargo near check --clippy --locked --env 'RUSTFLAGS=-Dwarnings' --manifest-path "$manifest"
-    done
-'''
+command = "bash"
+args = ["scripts/check-contract-clippy.sh"]
 
 [tasks.fmt]
 command = "cargo"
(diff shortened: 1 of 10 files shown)


crates/contract/src/crypto_shared/kdf.rs
```diff
@@ -1,14 +1,10 @@
...
 use curve25519_dalek::constants::ED25519_BASEPOINT_POINT;
-#[cfg(target_arch = "wasm32")]
-use k256::EncodedPoint;
 use k256::{
     Secp256k1,
     elliptic_curve::{CurveArithmetic, PrimeField, point::AffineCoordinates},
 };
 use near_mpc_contract_interface::types::Tweak;
-#[cfg(target_arch = "wasm32")]
-use near_sdk::env;
```
gilcu3, line 11: this is what the new check captures that was not captured before

Makefile.toml
```diff
@@ -48,6 +49,23 @@ args = [
...
+script = '''
+#!/usr/bin/env bash
+set -euo pipefail
+# Contract crates are exactly the workspace cdylibs; ask Cargo so the list self-maintains.
+cargo metadata --no-deps --format-version 1 \
+  | jq -r '.packages[] | select(any(.targets[].crate_types[]; . == "cdylib")) | .manifest_path' \
+  | while read -r manifest; do
+      cargo near check --clippy --locked --env 'RUSTFLAGS=-Dwarnings' --manifest-path "$manifest"
+    done
+'''
```
gilcu3: this is unfortunately the best way I found to do this for all our contracts...

pbeza: Should we extract this into a separate file? I think we generally prefer not to inline Bash scripts.

gilcu3: makes sense, I forgot about that :)

gilcu3: addressed

pbeza (approved):

crates/contract/Cargo.toml
```diff
@@ -18,9 +18,9 @@ repository = "https://github.com/near/mpc"
 [package.metadata.near.reproducible_build]
 # docker image, descriptor of build environment; pinned to match
 # `rust-toolchain.toml` (1.97.0)
```
barakeinav1, line 20: This comment says the image is "pinned to match `rust-toolchain.toml` (1.97.0)", but `rust-toolchain.toml` is actually on `1.93.0`. So `cargo near check` ends up linting the contract under 1.93 while the reproducible build ships it under 1.97 — which slightly undercuts the "lint in the real config" goal we're adding here. Is the plan to bump the toolchain to 1.97 (which would also make this comment true again), or is the 1.93/1.97 split intentional and we should just fix the comment? Not blocking, just want to make sure we're not linting under the wrong rustc.

scripts/check-contract-clippy.sh
```diff
@@ -0,0 +1,16 @@
...
+# Lint contract crates under the real build config via `cargo near check`.
+#
+# The wasm32 + `--cfg near` counterpart of plain `cargo clippy`: it lints the
+# contract crates in the build config they actually deploy under, which the host
+# target (no `--cfg near`) can't reach. `RUSTFLAGS=-Dwarnings` makes it a gate;
+# Cargo caps lints on non-workspace deps, so only our crates can fail.
+#
+# Contract crates are exactly the workspace cdylibs; ask Cargo so the list self-maintains.
+cargo metadata --no-deps --format-version 1 \
+  | jq -r '.packages[] | select(any(.targets[].crate_types[]; . == "cdylib")) | .manifest_path' \
```
barakeinav1, line 13: nit: if this cdylib filter ever comes back empty (metadata change, filter typo), the loop just runs nothing and the task goes green having linted zero contracts — a silent no-op. Might be worth a quick guard that we found at least one manifest, so a future break in the detection fails loudly instead.

barakeinav1 (approved):
Approved — one nit and a question, both non-blocking. Nice to finally lint the contract under the real wasm32 + `--cfg near` config.

scripts/check-contract-clippy.sh
gilcu3: near contracts recommend to have `cdylib`, so I think it is fine. Anyway this is just an extra lint

crates/contract/Cargo.toml
gilcu3: We already bumped the toolchain to 1.97 in https://github.com/near/mpc/commit/f29f196ecd8aabc1e02c62ecd1d3d2aae1cae038

PR #3821, opened 2026-07-15
fix: No silent failing using join_all

SimonRastikian:
Following [comments](https://github.com/near/mpc/pull/3640#discussion_r3578265609) from @kevindeforth and @gilcu3 I decided to make this as a small follup PR.
 
This introduces select_all + bail! fast-fail handling in the ECDSA and robust-ECDSA background task spawners.
This tries to fix the copilot blocker https://github.com/near/mpc/pull/3640#issuecomment-4901034077.


gilcu3:
Marking it as draft as this is not ready to review in its current state.

PR #3829, opened 2026-07-15
feat(node): register foreign-chain configs via the new contract API

anodar:
- Use new API for registering (along with legacy one until that's dropped in stacked PR's) 
- Adapt Fake contract to be like real contract: hold TLS-keyed (per node) config and mirror available set calculation.

Work towards: https://github.com/near/mpc/issues/3569

Note: this can be merged in isolation, doesn't require stacked PR's to be merged.


anodar:
@claude review

crates/node/src/indexer/fake.rs
```diff
@@ -98,85 +104,68 @@ impl FakeMpcContractState {
...
+    pub fn foreign_chains_configs(&self) -> &dtos::ForeignChainsConfigs {
+        &self.foreign_chains_configs
     }
 
-    #[expect(deprecated)]
-    pub fn register_foreign_chain_config(
+    /// Deviation from the real contract: registrations are accepted only in
+    /// `Running`, while the real endpoint also accepts existing or prospective
+    /// participants during `Initializing` and `Resharing`.
+    pub fn register_foreign_chains_config(
```
kevindeforth: It's not ideal if the test framework is more restrictive than the production framework., as we can't test production behavior. Lets be more permissive here.

crates/node/src/indexer/fake.rs
```diff
@@ -98,85 +104,68 @@ impl FakeMpcContractState {
...
-            for chain in chains.iter().copied() {
-                chain_to_supporters
-                    .entry(chain)
-                    .or_default()
-                    .insert(voter_id.clone());
+    /// Unlike the real contract, no whitelisting is required and the
+    /// recomputation runs even without a ForeignTx domain.
+    /// TODO(#3640): use the max reconstruction threshold across ForeignTx
+    /// domains instead of the governance threshold.
+    fn recompute_available_foreign_chains(&mut self) {
```
kevindeforth: I think we can already improve this behavior. IIRC, the domain-specific thresholds is already implemented in the contract, and the fake indexer depends on the contract state, so we have access to the domain thresholds.

crates/node/src/indexer/fake.rs
```diff
@@ -74,12 +74,18 @@ impl ReadAttestationExpiry for FakeAttestationExpiryReader {
 
 impl ReadSupportedForeignChain for FakeReadSupportedForeignChain {
     async fn get_supported_chains(&self) -> anyhow::Result<dtos::SupportedForeignChains> {
+        // Shim while the node's read path still asks for "supported" chains:
+        // the fake models the new API only, so serve the available set.
+        // TODO(#3630): removed with the trait once the node reads available chains.
         Ok(self
```
kevindeforth: This is somewhat risky and will be hard to track across the PRs - in production, we are still relying on the old contract logic, but in testing, we already test the new logic.

I would suggest to add the new logic here and remove the old one once the node depends on it.

kevindeforth (commented):
Two blockers:
- test framework shouldn't be less permissive than production
- the integration testsand production logic have diverged, we should keep them close. We never know how long it may take to get the rest of the feature changes merged.

crates/node/src/indexer/fake.rs
anodar: Addressed in [f29454b](https://github.com/near/mpc/pull/3829/commits/f29454bbcbeb35f350ccfacb86ce73e4dbddcc6c)

crates/node/src/indexer/fake.rs
anodar: Addressed in [f29454b](https://github.com/near/mpc/pull/3829/commits/f29454bbcbeb35f350ccfacb86ce73e4dbddcc6c)

crates/node/src/indexer/fake.rs
anodar: Addressed in [f29454b](https://github.com/near/mpc/pull/3829/commits/f29454bbcbeb35f350ccfacb86ce73e4dbddcc6c)

anodar:
@claude review, verify that @kevindeforth's comments are addressed.

crates/node/src/indexer/fake.rs
```diff
@@ -182,6 +191,86 @@ impl FakeMpcContractState {
...
+        let Some(threshold) = self.state.domain_registry().ok().and_then(|registry| {
+            registry
+                .domains()
+                .iter()
+                .filter(|domain| domain.purpose == dtos::DomainPurpose::ForeignTx)
+                .map(|domain| domain.reconstruction_threshold.inner())
+                .max()
+        }) else {
+            return;
+        };
```
kevindeforth, line 256: Note (separate issue): this is probably how we do it in the contract as well.
Practically though, we need reconstruction_threshold() participants from the intersection of the the current and the prospective epoch (because the nodes don't connect with previous participants that are not in the prospective epoch during resharing).

It's out of scope of this PR, because that will require changes to the contract and should probably be noted in the design doc (it's something we missed previously).

crates/node/src/tests/foreign_chain_configuration.rs
```diff
@@ -76,11 +87,11 @@ async fn foreign_chain_configuration_auto_registered_to_contract_on_startup__sho
         .collect::<Vec<_>>();
 
     // When
-    let wait_result = tokio::time::timeout(Duration::from_secs(10), async {
+    let wait_result = tokio::time::timeout(Duration::from_secs(30), async {
```
kevindeforth: note: this timeout needs to be increased, because we now have a key generation in this test. The `contract.add_domains` above sets the contract into initializing state.

Here, better to rely on `DEFAULT_MAX_PROTOCOL_WAIT_TIME` from `node/src/tests.rs`, as it's what we use in all the other tests as well.

kevindeforth (dismissed):
Thank you, no blockers, one suggestion for test org and one general remark about how we calculate available chains that is a separate issue.

crates/node/src/tests/foreign_chain_configuration.rs
anodar: >note: this timeout needs to be increased, because we now have a key generation in this test. The contract.add_domains above sets the contract into initializing state.

Yeah, I meant to comment in PR itself explaining exactly this was increased (as that comment doesn't belong in the code itself) but forgot to. 

Changed to using constant in [d410173](https://github.com/near/mpc/pull/3829/commits/d41017342d983d6bf85407fdcd8584dbf8bdc65a)

crates/node/src/indexer/fake.rs
anodar: Good point! Created bug for it: https://github.com/near/mpc/issues/3833.

Set priority to P2 as it's not very urgent because all participants should support all chains during normal operations so this discrepency (current vs current & prospective) should not make difference in practice (during normal operations).

kevindeforth (dismissed):

haiyuechen-nearone (dismissed):
This PR LGTM

Proposal for PRs that is in a series: I would like the description to include the overall plan for what the series is trying to solve, and where the given PR fits in that series, so reviewers can easily load in the context. Wdyt?

anodar:
> Proposal for PRs that is in a series: I would like the description to include the overall plan for what the series is trying to solve, and where the given PR fits in that series, so reviewers can easily load in the context. Wdyt?


Replied in the slack thread. Will update main issue with plan breakdown.

haiyuechen-nearone (approved):

kevindeforth (approved):
Thank you!

PR #3832, opened 2026-07-16
refactor(contract): share NodeId/auth test helpers via `primitives::test_utils`

pbeza:
Closes #3831

PR #3846, opened 2026-07-16
feat(e2e): extend e2e cluster with new API from available chains

### LINKED ISSUE #3842: Add available chains API in e2e tests
### Background

Part of https://github.com/near/mpc/issues/3569

We need to add new available chains API in e2e testing framework and in its tests.

### Acceptance Criteria

New API is integrated in e2e tests.

anodar:
We've made changes in the contract (in https://github.com/near/mpc/pull/3511), this extends e2e framework with that new API.

Closes: https://github.com/near/mpc/issues/3842


crates/e2e-tests/src/cluster.rs
```diff
@@ -93,6 +96,7 @@ pub struct MpcClusterConfig {
     pub migration_targets: Vec<usize>,
     /// Wire format used when calling `init`. See [`ContractInitFormat`].
     pub init_format: ContractInitFormat,
+    pub whitelisted_chains: BTreeSet<ForeignChain>,
```
pbeza: I think it would make sense to add a struct inside `MpcClusterConfig` that groups all foreign-chain-related config. Just looking at `whitelisted_chains`, it’s not obvious that it belongs to the foreign chain feature.

crates/e2e-tests/src/cluster.rs
```diff
@@ -889,6 +894,79 @@ impl MpcCluster {
...
+            auth_scheme: AuthScheme::None,
+            chain_routing: ChainRouting::Embedded,
+        };
+        let mut providers =
+            NonEmptyBTreeMap::new(ProviderId("alchemy".to_string()), provider("example1.com"));
+        providers.insert(
+            ProviderId("quicknode".to_string()),
+            provider("example2.com"),
+        );
+        providers.insert(ProviderId("public".to_string()), provider("example3.com"));
```
pbeza: Just from the function name (`whitelist_foreign_chains`), I wouldn’t expect it to use hardcoded providers. Perhaps it would be better to pass them in as arguments or rename the function to make that behavior clearer.

crates/e2e-tests/tests/foreign_chain_configuration.rs
```diff
@@ -130,3 +133,104 @@ async fn supported_foreign_chains__should_require_all_participants_to_register()
...
+        let available = cluster
+            .view_available_foreign_chains()
+            .await
+            .expect("failed to view available chains");
+
+        anyhow::ensure!(
+            registrations.len() == 3,
+            "expected exactly 3 registrations, got {}",
+            registrations.len()
+        );
```
pbeza: Any specific reason for using `anyhow::ensure!` instead of `assert_eq!` here? Same for a few other places in the code.

crates/e2e-tests/tests/foreign_chain_configuration.rs
```diff
@@ -130,3 +133,104 @@ async fn supported_foreign_chains__should_require_all_participants_to_register()
...
+        let empty_registrations = registrations.values().filter(|c| c.is_empty()).count();
+        anyhow::ensure!(
+            empty_registrations == 1,
+            "expected exactly 1 empty registration (node 2), got {empty_registrations}"
+        );
+        anyhow::ensure!(
+            !available.contains(&ForeignChain::Solana),
```
pbeza: I prefer strict comparisons for `available` rather than just `.contains()`, since they make it clear exactly what it contains. But I get that you mainly wanted to highlight that Solana is missing here.

crates/e2e-tests/tests/foreign_chain_configuration.rs
```diff
@@ -130,3 +133,104 @@ async fn supported_foreign_chains__should_require_all_participants_to_register()
...
+            "Solana must not be available before it is whitelisted"
+        );
+        Ok(())
+    })
+    .retry(
+        ConstantBuilder::default()
+            .with_delay(common::POLL_INTERVAL)
+            .with_max_times(
+                (CLUSTER_WAIT_TIMEOUT.as_millis() / common::POLL_INTERVAL.as_millis()) as usize,
+            ),
```
pbeza: It’s copy-pasted, so we could probably extract it somehow.

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
```diff
@@ -260,33 +272,35 @@ async fn setup_foreign_tx_cluster() -> anyhow::Result<ForeignTxTestEnv> {
...
-    .into_iter()
-    .collect();
-
     (|| async {
+        let available = cluster
+            .view_available_foreign_chains()
+            .await
+            .context("failed to view available chains")?;
+        let available_set: std::collections::BTreeSet<ForeignChain> =
+            available.iter().copied().collect();
```
pbeza:
```suggestion
        let available_set: BTreeSet<ForeignChain> =
            available.iter().copied().collect();
```

crates/e2e-tests/tests/foreign_chain_configuration.rs
```diff
@@ -130,3 +133,104 @@ async fn supported_foreign_chains__should_require_all_participants_to_register()
...
+    // configuration) without Solana becoming available.
+    (|| async {
+        let registrations = cluster
+            .view_foreign_chains_configs()
+            .await
+            .expect("failed to view registered configs");
+        let available = cluster
+            .view_available_foreign_chains()
+            .await
+            .expect("failed to view available chains");
```
pbeza: Personally, I’d probably just `assert_eq!` these values instead of doing all the assertions and conversions below, if it’s feasible and still readable. But I guess it’s mostly a matter of taste, so I’m not pushing for a change.

pbeza (dismissed):
Left a couple of debatable nits. None of them are blockers, so feel free to address the ones that make the most sense to you.

crates/e2e-tests/src/cluster.rs
anodar: Addressed in: [1d68731](https://github.com/near/mpc/pull/3846/commits/1d687311771993a97387f665b13c75bf8c09ac76)

crates/e2e-tests/tests/foreign_chain_configuration.rs
anodar: Yes, we want to propagate error not panic. It's inside closure with `.retry()`

crates/e2e-tests/tests/foreign_chain_configuration.rs
anodar: Yeah contains was attempt to express intent more clearly. Changed in [1b52dfb](https://github.com/near/mpc/pull/3846/commits/1b52dfbb505985f5d489cc6afd326cb8e417ce9a)

crates/e2e-tests/tests/foreign_chain_configuration.rs
anodar: Done in [1b52dfb](https://github.com/near/mpc/pull/3846/commits/1b52dfbb505985f5d489cc6afd326cb8e417ce9a)

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
anodar: Done in [1b52dfb](https://github.com/near/mpc/pull/3846/commits/1b52dfbb505985f5d489cc6afd326cb8e417ce9a)

crates/e2e-tests/tests/foreign_chain_configuration.rs
anodar: Expanded assertions like this, rather than single assert_eq! simplifies debugging when test fails because it points much narrower point of failure rather than "result wasn't what we expected".

Changed in [1b52dfb](https://github.com/near/mpc/pull/3846/commits/1b52dfbb505985f5d489cc6afd326cb8e417ce9a)

crates/e2e-tests/tests/common.rs
```diff
@@ -14,6 +14,13 @@ use near_mpc_crypto_types::Bls12381G1PublicKey;
 use serde_json::json;
 
 pub const POLL_INTERVAL: Duration = Duration::from_millis(500);
+
+/// Constant-interval retry policy: poll every `POLL_INTERVAL` until `CLUSTER_WAIT_TIMEOUT`.
```
pbeza: Ditto: https://github.com/near/mpc/pull/3832#discussion_r3594080263. TL;DR: @kevindeforth won’t approve it. :D

pbeza (dismissed):

crates/e2e-tests/src/cluster.rs
```diff
@@ -93,6 +93,38 @@ pub struct MpcClusterConfig {
     pub migration_targets: Vec<usize>,
     /// Wire format used when calling `init`. See [`ContractInitFormat`].
     pub init_format: ContractInitFormat,
+    pub foreign_chains: ForeignChainsClusterConfig,
+}
+
+/// Foreign-chain feature configuration for the cluster.
+#[derive(Default)]
+pub struct ForeignChainsClusterConfig {
```
kevindeforth: nit: lets try not to paraphrase struct names in titles please. 

crates/e2e-tests/src/cluster.rs
```diff
@@ -93,6 +93,38 @@ pub struct MpcClusterConfig {
...
+    /// Per-node configs. If empty, all nodes get the default (empty) config;
+    /// if non-empty, must have exactly `num_nodes` entries.
+    pub node_configs: Vec<mpc_node_config::ForeignChainsConfig>,
+    /// Chains whitelisted during cluster setup via a threshold vote.
+    pub whitelisted_chains: BTreeSet<ForeignChain>,
```
kevindeforth: How can this struct guarantee that it requires a threshold vote to whitelist a chain?

crates/e2e-tests/src/cluster.rs
```diff
@@ -93,6 +93,38 @@ pub struct MpcClusterConfig {
...
+}
+
+/// Whitelist entry for tests where only whitelist membership matters; the
+/// provider content is never contacted.
+pub fn placeholder_chain_entry() -> ChainEntry {
```
kevindeforth: How does this method guarantee that its output will get whitelisted?

crates/e2e-tests/src/cluster.rs
```diff
@@ -889,6 +921,64 @@ impl MpcCluster {
             .await
     }
 
+    /// View the available foreign chains (threshold-covered + whitelisted).
+    pub async fn view_available_foreign_chains(
```
kevindeforth: https://github.com/near/mpc/blob/main/docs/engineering-standards.md#write-helpful-code-comments

crates/e2e-tests/src/cluster.rs
```diff
@@ -889,6 +921,64 @@ impl MpcCluster {
...
+        &self,
+    ) -> anyhow::Result<near_mpc_contract_interface::types::AvailableForeignChains> {
+        self.contract
+            .view(method_names::GET_AVAILABLE_FOREIGN_CHAINS)
+            .await
+    }
+
+    /// View the per-node foreign chain configs keyed by TLS public key.
+    pub async fn view_foreign_chains_configs(
```
kevindeforth: https://github.com/near/mpc/blob/main/docs/engineering-standards.md#write-helpful-code-comments

crates/e2e-tests/src/cluster.rs
```diff
@@ -889,6 +921,64 @@ impl MpcCluster {
...
+            .iter()
+            .map(|&chain| (chain, entry.clone()))
+            .collect::<BTreeMap<_, _>>()
+            .try_into()
+            .expect("non-empty: checked above");
+
+        // Deliberately over-votes: the proposal applies at the protocol threshold;
+        // voting from every index stays correct for any threshold (e.g. after
+        // resharing) at the cost of surplus votes leaving pending rows behind.
+        for &idx in participant_indices {
```
kevindeforth: Do we test somewhere else that the excess votes actually get recoreded?
Unfortunately, the comment doesn't explain _why_ we choose to over-vote. I would have assumed that for testing, we would prefer to submit the _minimum_ required number of votes, to ensure that the expected minimum is always sufficient to change the foreign chain providers list.
So,over-voting feels wrong, from a test-correctness perspective.

I don't feel comfortable approving the PR until I understand the rationale behind it.

crates/e2e-tests/src/cluster.rs
kevindeforth: For this method, I would suggest to take as argument the chain-id it is supposed to be the entry of (could be the entry itself), and using that id as a seed for some randomness. Otherwise, all entries look exactly the same and tests lose some meaning.

We could probably move `build_providers_from_urls` from the test here.

_Reasoning:
Assume you expect `[(chain_1, chain_entry_1), (chain_2, entry_for_chain_2)]`. If both entries are identical, we don't really know if the insertion works as expected. Stupid example, but assume the contract, for some reason, overrode every entry with the latest `chain_entry`, we wouldn't catch that bug here, but only in production (or unit-test, if we have, but my point is that it's easy to improve the e2e-test to be more meaningful)._

crates/e2e-tests/src/cluster.rs
```diff
@@ -889,6 +921,64 @@ impl MpcCluster {
...
+        if chains.is_empty() {
+            return Ok(());
+        }
+
+        let batch: NonEmptyBTreeMap<ForeignChain, ChainEntry> = chains
+            .iter()
+            .map(|&chain| (chain, entry.clone()))
+            .collect::<BTreeMap<_, _>>()
+            .try_into()
+            .expect("non-empty: checked above");
```
kevindeforth: This looks less-than ideal to me, for the same reason. I would suggest to require one entry per chain.

crates/e2e-tests/tests/foreign_chain_configuration.rs
```diff
@@ -120,13 +116,106 @@ async fn supported_foreign_chains__should_require_all_participants_to_register()
...
+///
+/// 3-node cluster (ForeignTx domain threshold 2): nodes 0 and 1 are configured
+/// with the Solana foreign chain, node 2 has no foreign chain config.
+///
+/// 1. All three nodes auto-register on startup — nodes 0 and 1 with Solana,
+///    node 2 with an empty configuration. Solana must NOT be available while
+///    it is not whitelisted.
+/// 2. After the participants whitelist Solana, two registrations reach the
+///    threshold and the chain becomes available — node 2's foreign-chain-config
+///    registration is not required (it still votes for the whitelist).
```
kevindeforth: This comment can be improved by explaining _why_ we have node 2 submit a vote, despite it not being required.

crates/e2e-tests/tests/common.rs
anodar: Addressed in [2723b65](https://github.com/near/mpc/pull/3846/commits/2723b658fdeea9f158839ac4e80c4bee7ed4cdf9).

crates/e2e-tests/tests/foreign_chain_configuration.rs
```diff
@@ -120,13 +116,106 @@ async fn supported_foreign_chains__should_require_all_participants_to_register()
...
+#[tokio::test]
+#[expect(non_snake_case)]
+async fn available_foreign_chains__should_require_whitelist_and_threshold_of_registrations() {
+    // given — 3-node cluster with a ForeignTx domain; Solana configured on
+    // nodes 0 and 1 only, and no chain whitelisted yet.
```
haiyuechen-nearone, line 140: Do we need this comment given the block above?

crates/e2e-tests/tests/foreign_chain_configuration.rs
```diff
@@ -120,13 +116,106 @@ async fn supported_foreign_chains__should_require_all_participants_to_register()
...
+        );
+        anyhow::ensure!(
+            available.is_empty(),
+            "no chain must be available before whitelisting, got {available:?}"
+        );
+        Ok(())
+    })
+    .retry(common::cluster_poll_retry())
+    .await
+    .expect("timed out waiting for all three registrations with one empty");
```
kevindeforth: This would look good as a helper method on the cluster:
```rust
cluster.wait_for_foreign_chains_registrations(expected_registrations);
cluster.must_have_available_foreign_chains(BTreeMap::new());
```

crates/e2e-tests/tests/foreign_chain_configuration.rs
anodar: Block above explains the entire test. Given/when/then break it down in steps. It's repetitive but follows existing patter in many tests already.

Happy to change if there are strong opinions on it.

crates/e2e-tests/tests/foreign_chain_configuration.rs
```diff
@@ -120,13 +116,106 @@ async fn supported_foreign_chains__should_require_all_participants_to_register()
...
+
+        anyhow::ensure!(
+            *available == BTreeSet::from([ForeignChain::Solana]),
+            "expected exactly [Solana] to be available, got {available:?}"
+        );
+        Ok(())
+    })
+    .retry(common::cluster_poll_retry())
+    .await
+    .expect("timed out waiting for Solana to become available");
```
kevindeforth: and here:
```rust
cluster.wait_for_available_foreign_chains(...)
```

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
```diff
@@ -54,6 +55,25 @@ struct MockServerUrls {
...
+            let cfg = ForeignChainProviderConfig {
+                rpc_url: url.clone(),
+                auth: Default::default(),
+            };
+            (format!("mock-{chain_name}-{i}").into(), cfg)
+        })
+        .collect();
+    map.try_into()
+        .unwrap_or_else(|_| panic!("at least one {chain_name} provider must be configured"))
+}
```
kevindeforth: this one looks like we could use it in the `common.rs`

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
```diff
@@ -259,47 +270,35 @@ async fn setup_foreign_tx_cluster() -> anyhow::Result<ForeignTxTestEnv> {
...
         Ok(())
     })
-    .retry(
-        ConstantBuilder::default()
-            .with_delay(common::POLL_INTERVAL)
-            .with_max_times(
-                (CLUSTER_WAIT_TIMEOUT.as_millis() / common::POLL_INTERVAL.as_millis()) as usize,
-            ),
-    )
+    .retry(common::cluster_poll_retry())
```
kevindeforth: this would be shorter with the helper method

crates/e2e-tests/src/cluster.rs
```diff
@@ -82,9 +85,6 @@ pub struct MpcClusterConfig {
     /// An empty vec means all nodes are participants. Set to a subset to start
     /// extra non-participant nodes (useful for resharing and attestation tests).
     pub initial_participant_indices: Vec<usize>,
-    /// Per-node foreign chains configuration. If empty, all nodes get the default
-    /// (empty) config. If non-empty, must have exactly `num_nodes` entries.
-    pub node_foreign_chains_configs: Vec<mpc_node_config::ForeignChainsConfig>,
```
haiyuechen-nearone, line 87: I think this change is not reflected in the README for this crate, could you ask Claude to run a quick pass to generate the docs 

Image
https://github.com/near/mpc/blob/1b52dfbb505985f5d489cc6afd326cb8e417ce9a/crates/e2e-tests/README.md#L210

kevindeforth (commented):
Thank you!

Left some comments, I think the following can be improved:

- some of the comments paraphrase struct names, which goes against our [coding standards](https://github.com/near/mpc/blob/main/docs/engineering-standards.md#write-helpful-code-comments)
- The e2e test choses to submit extra votes, which is also explicitly mentioned in the code comments. It's not clear to me why we want extra-votes if we never check for them.
- Some of the helper methods in the tests can be improved to reduce code surface and improve readability of the tests.

crates/e2e-tests/src/cluster.rs
anodar: Done in: [4f0ccaf](https://github.com/near/mpc/pull/3846/commits/4f0ccafec8b4233032f623936ac3077c4e00990c)

crates/e2e-tests/src/cluster.rs
anodar: Got rid of comment.

crates/e2e-tests/src/cluster.rs
anodar: Got rid of comment.

crates/e2e-tests/src/cluster.rs
anodar: Got rid of comment.

crates/e2e-tests/src/cluster.rs
anodar: Got rid of comment.

crates/e2e-tests/src/cluster.rs
anodar: Dropped over-voting in: [e808df1](https://github.com/near/mpc/pull/3846/commits/e808df1e7ce71f8f346b202b3dae55cb33cb5ed7)

crates/e2e-tests/src/cluster.rs
anodar: Updated in: [14a25cb](https://github.com/near/mpc/pull/3846/commits/14a25cb118d395d1b8ef4575dd52596bcbaa2f3f)

crates/e2e-tests/src/cluster.rs
pbeza: > (...) Otherwise, all entries look exactly the same and tests lose some meaning.

Yeah, I had the same feeling. I left a similar comment here: https://github.com/near/mpc/pull/3641#discussion_r3543722204:

> Nit: Isn’t it a bit confusing that all chains get the same set of provider names and URLs? I think it’s fine for the test fixture, but it took me a while to realize we mock all chains the same way.

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
anodar: Done.

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
anodar: Done.

crates/e2e-tests/tests/foreign_chain_configuration.rs
anodar: Done.

crates/e2e-tests/tests/foreign_chain_configuration.rs
anodar: Done.

crates/e2e-tests/tests/foreign_chain_configuration.rs
anodar: Switched to voting only from 2/3 nodes.

crates/e2e-tests/src/cluster.rs

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
```diff
@@ -259,47 +250,38 @@ async fn setup_foreign_tx_cluster() -> anyhow::Result<ForeignTxTestEnv> {
...
-        ConstantBuilder::default()
-            .with_delay(common::POLL_INTERVAL)
-            .with_max_times(
-                (CLUSTER_WAIT_TIMEOUT.as_millis() / common::POLL_INTERVAL.as_millis()) as usize,
-            ),
-    )
-    .await
-    .context("timed out waiting for every participant to register its foreign chains")?;
+    // The nodes dual-write, so the legacy view must stay in lockstep until
+    // the deprecated API is dropped.
```
kevindeforth: nit: This reads like it should be a `TODO`

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
```diff
@@ -259,47 +250,38 @@ async fn setup_foreign_tx_cluster() -> anyhow::Result<ForeignTxTestEnv> {
...
+        *supported == expected_chains,
+        "expected supported chains {expected_chains:?}, got {supported:?}"
+    );
+    let allowed = cluster
+        .view_allowed_foreign_chain_providers()
+        .await
+        .context("failed to view allowed foreign chain providers")?;
+    let expected_allowed: BTreeMap<_, _> = expected_chains
+        .iter()
+        .map(|&chain| (chain, placeholder_chain_entry(chain)))
```
kevindeforth: This doesn't seem ideal - the test needs to know what helper-method the cluster setup invokes. If we change the logic in the setup, we need to change the test logic as well. Would be better to fetch the value of `expected_allowed` from the cluster itself, or to set it explicitly during setup in case the test cares about it.

It's not a blocker, because violating the invariant  would result a failing test, but it's something I would like to get @anodar's opinion on.

kevindeforth (dismissed):
Thank you! Just one question / discussion, but not a blocker.

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
anodar: Fair point. Got rid of placeholder_chain_entry from setup in [f8f5d3d](https://github.com/near/mpc/pull/3846/commits/f8f5d3def03d0b0f8a093b54fc1f523403ac40bc)

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
anodar: Changed to TODO: [f8f5d3d](https://github.com/near/mpc/pull/3846/commits/f8f5d3def03d0b0f8a093b54fc1f523403ac40bc)

kevindeforth (approved):

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
anodar: To be clear, I got rid of it from must_setup_cluster, but added it to setup_foreign_tx_cluster. I think that's a fair place to be as test directly calls that setup helper (unlike cluster setup which is called transitively).

pbeza (approved):

PR #3848, opened 2026-07-17
feat(node): probe foreign-chain RPC providers on startup

haiyuechen-nearone:
Part of #3764. Wires the identity-based check into the node. Merge the identity-probe stack (#3932 → #3962 → #3963 → #3964) first, then this trio bottom-up:

1. **#3848 (this PR)** — probe every configured provider on startup + per-provider logs
2. #3849 — per-chain `configured` / `healthy` Prometheus gauges
3. #3850 — healthy counts on `/debug/node_config`

Spawns a detached startup task that concurrently probes every configured foreign-chain RPC provider once — verifying its chain identity and inspecting a recent transaction, the same inspector and auth the real verification path uses — and logs a per-provider result plus an `x/y providers healthy` summary. The probe never blocks startup, and failure reasons aren't logged (they can carry secrets).

Expected identities come from `foreign_chain_health_check.identities` (no built-ins, so any network — including local — is checkable; a configured chain with no identity fails). `ExpectedIdentities` moves to node-config so the node and config-tester share one type.


haiyuechen-nearone:
@claude

crates/foreign-chain-health-check/src/network.rs
```diff
@@ -17,3 +17,216 @@ impl Network {
...
+/// Classify mainnet/testnet from a contract-id suffix.
+pub fn network_from_contract_id(contract_id: &str) -> Option<Network> {
+    if contract_id.ends_with(".testnet") {
+        Some(Network::Testnet)
+    } else if contract_id.ends_with(".near") || contract_id == "v1.signer" {
+        Some(Network::Mainnet)
+    } else {
+        None
+    }
+}
```
haiyuechen-nearone: To reviewers: please double check that I have the correct understanding of our network and contract naming conventions

crates/node/src/foreign_chain_health.rs
```diff
@@ -0,0 +1,359 @@
...
+/// Startup healthcheck entrypoint: builds a [`HealthCheckRoute`] from the config
+/// and runs it, warning if a golden config is set on a public network. A probe
+/// panic is caught and logged, never propagated (diagnostic-only).
+pub async fn run_startup_health_check(
+    foreign_chains: ForeignChainsConfig,
+    network: NetworkKind,
+    golden: Option<HealthCheckGoldenConfig>,
+) {
+    // A config-supplied golden is only meaningful on a local chain; on
+    // mainnet/testnet the built-in set is always used.
```
netrome: Why don't we provide golden configurations on mainnet/testnet? Hard-coding values feels like a time bomb, as many RPCs might not serve requests for old transactions.

crates/node/src/foreign_chain_health.rs
```diff
@@ -0,0 +1,359 @@
...
+    if let NetworkKind::Public(network) = network
+        && golden.is_some()
+    {
+        warn!(
+            network = network.label(),
+            "foreign_chain_health_check_golden is set in config but ignored on \
+             mainnet/testnet; the built-in golden set is always used — remove it \
+             from config.yaml"
```
netrome: Wouldn't it make sense to override the golden set if provided manually? The hard-coded values could exist as a convenience so we don't have to ask all operators to manually add these but retaining the ability to always override sounds like it would simplify our code. I'd prefer if we have as few network-specific code paths as possible. Ideally none.

crates/node-config/src/foreign_chains.rs
```diff
@@ -56,6 +56,56 @@ pub struct ForeignChainProviderConfig {
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub sui: Option<SuiGolden>,
+}
+
+/// Hashes are hex, with or without a `0x` prefix.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct BlockHashGolden {
+    pub tx: String,
+    pub block_hash: String,
+}
```
netrome: I think this is a dangerous construction. Many RPCs won't serve transactions forever. It's better to ask for genesis hash or chain ID, like we do for Sui.

netrome (changes requested):
Started looking at this now, some initial thoughts:

1. We should not expect RPCs to retain individual TX information forever (blocking)
2. I'd prefer to minimize network-specific code paths. Ideally we should not have any network-specific code paths, but I can see the point of providing defaults for mainnet and testnet for convenience.

crates/node/src/foreign_chain_health.rs
haiyuechen-nearone: My understanding was that we want to always check for the same transaction, which is baked in to the binary, maybe I misunderstood. 

@anodar Judging from the foreign chain tester CLI code I assumed that we would like to continue using hard coded golden transactions for health check, maybe you had something else in mind?

haiyuechen-nearone:
Superseded by https://github.com/near/mpc/pull/4013

PR #3849, opened 2026-07-17
feat(node): expose foreign-chain RPC provider health metric

haiyuechen-nearone:
Part of #3764. Stacked on #3848, merge bottom-up:

1. #3848 — startup probe + per-provider logs
2. **#3849 (this PR)** — per-chain `configured` / `healthy` Prometheus gauges
3. #3850 — healthy counts on `/debug/node_config`

Publishes two per-chain gauges from the startup probe: `mpc_foreign_chain_rpc_providers_configured` (set up front for every configured chain) and `mpc_foreign_chain_rpc_providers_healthy` (providers that passed the identity probe). A configured chain with no identity reads as `0/N`.

Adds `ForeignChain::label()` as the single source of the snake_case chain label shared by config keys, metric labels, and health-check result keys.


haiyuechen-nearone:
@claude

Issue #3164, opened 2026-05-07
Node consumes per-domain ReconstructionThreshold from DomainConfig


gilcu3:
### Background

#3143 added `reconstruction_threshold` to `DomainConfig` and validates it on `vote_add_domains` (universal `2 ≤ t ≤ n`, plus DamgardEtAl `2t - 1 ≤ n`). The contract is the source of truth, but the node still derives crypto thresholds from `mpc_config.participants.threshold` (the network's governance threshold). For the field to actually matter at signing time, the node has to read the per-domain value and use it for keygen, presign, sign, and resharing.

This work corresponds to "PR 7" in `docs/design/domain-separation.md` §4.2. A first pass was started inside the #3143 PR and reverted because the diff got too large to review alongside the contract-side changes; this issue tracks the focused follow-up.

### Acceptance Criteria

- Each provider's `PerDomainData` (in `ecdsa.rs`, `eddsa.rs`, `ckd.rs`, `robust_ecdsa.rs`) carries the per-domain `ReconstructionThreshold`.
- `vote_add_domains`-style sign/presign sites read from `domain_data.reconstruction_threshold` instead of `mpc_config.participants.threshold`.
- Keygen (`keygen_computation_inner`) derives the threshold from `domain.reconstruction_threshold`. The threshold parameter is dropped from `keygen_leader` / `keygen_follower` / `keygen_computation`.
- `ResharingArgs` carries the previous epoch's per-domain thresholds, and `resharing_computation_inner` patches `old_participants.threshold` per-key so the underlying CGGMP resharing sees the right `t` on each side of the transition.
- DamgardEtAl semantics: replace `translate_threshold` with the design's direct `MaxMalicious::from(t - 1)`. Drop `translate_threshold` and `get_number_of_signers` from `crates/node/src/providers/robust_ecdsa.rs` along with their tests.
- ECDSA triples must be computed **per domain**. Handled in #3292

gilcu3:
Setting as OnHold until #3299 is merged

gilcu3:
@netrome I am removing my assignment from this one to focus on https://github.com/near/mpc/issues/1175 as we mentioned during the meeting.

@SimonRastikian if you take this one, feel free to reuse the code in this draft https://github.com/near/mpc/pull/3170 (now closed) or start fresh

netrome:
Head's up, since we performed a resharing in testnet before https://github.com/near/mpc/issues/3169 we need to:

[...]

Issue #1680, opened 2025-12-17
Simplify PresignatureStorage::new parameters


gilcu3:
### Background

Currently we take `alive_participant_ids_query: Arc<dyn Fn() -> Vec<ParticipantId> + Send + Sync>,` as a parameter for `PresignatureStorage::new`. As mentioned in https://github.com/near/mpc/pull/1645#discussion_r2620906941 we could instead take a reference or clone of the client itself

### Acceptance Criteria

The function is simplified and fixed for all crypto providers

netrome:
Would be nice to fix, but low prio. Setting as P3.

netrome:
This could either be an easy refactor or could lead to some nasty borrow-checker battles. Could be worth having a stab at it, but if it explodes let's prioritize something else @AmHelm

SimonRastikian:
Sorry @AmHelm I am fixing this quick in my PR where I already touch this thing

SimonRastikian:
Btw this issue should be extended to simplifying new in triple.rs

PR #3851, opened 2026-07-17
refactor: simplifying storage by giving it to the client directly

SimonRastikian:
Simplifies the triple generation part following https://github.com/near/mpc/issues/3164. Concretely, it refactors TripleStorage::new so the storage takes the network client directly, instead of being handed the two pieces of state it used to derive from that client.

Net effect: TripleStorage depends on one cohesive object (the mesh client) rather than being passed several client-derived values separately.

Reference https://github.com/near/mpc/issues/1680#issuecomment-4852112094


gilcu3:
This has diverged from main, so marking as draft for now. IIRC the change was relatively small, so we could get this done quickly once we are back at it.

Issue #1956, opened 2026-02-04
IndexerViewClient code duplication


gilcu3:
### Background

In the node, the `IndexerViewClient` struct contains quite a lot of duplicated code. It is probably possible to reduce is considerably by using generics

### Acceptance Criteria

The `IndexerViewClient` is simplified such that the amount of duplicated logic is minimal

netrome:
Considering the chain gateway migration, this feels irrelevant since we'll remove the `IndexerViewClient` in favor of the new chain gateway.

PR #3853, opened 2026-07-17
feat(node): indexer publishes foreign-chain supporters via a watch channel

### LINKED ISSUE #3852: Publish foreign-chain supporters from the indexer via a watch channel
### Background

 The node still reads foreign-chain support through the legacy path: every verify-foreign-tx request triggers a view call to the deprecated `get_supported_foreign_chains`.

The contract already exposes the new views `get_available_foreign_chains` and `get_foreign_chains_configs` (https://github.com/near/mpc/issues/3568), and the indexer already tracks other contract state through monitor tasks publishing to watch channels (contract state, allowed docker images).

 Part of https://github.com/near/mpc/issues/3569: before the node can switch availability checks and leader election to the new API, the indexer needs to publish that data.

 ### Acceptance Criteria

Indexer monitor task periodically reads `get_available_foreign_chains` and `get_foreign_chains_configs` and publishes a single watch channel mapping each available foreign chain to the TLS public keys of the nodes registered as supporting it. On read errors the previously published value stays in effect. The fake indexer publishes the same channel for integration tests. No consumers yet — the request execution path is unchanged until the switch-over under https://github.com/near/mpc/issues/3569.

anodar:
Closes https://github.com/near/mpc/issues/3852


crates/node/src/indexer/foreign_chain.rs
```diff
@@ -0,0 +1,214 @@
...
+        .with_max_delay(MAX_BACKOFF_DURATION)
+        .without_max_times()
+        .with_jitter()
+        .build();
+
+    loop {
+        match fetch().await {
+            Ok(value) => return value,
+            Err(e) => {
+                tracing::error!(target: "mpc", "{error_context}: {:?}", e);
```
gilcu3: we have `info` in the other loops (I think Claude already found it), so we should use the same here

crates/node/src/indexer/foreign_chain.rs
```diff
@@ -0,0 +1,214 @@
...
+            "error reading foreign-chain supporters from chain",
+        )
+        .await;
+        publish_if_changed(&sender, supporters);
+        tokio::time::sleep(FOREIGN_CHAIN_SUPPORTERS_REFRESH_INTERVAL).await;
+    }
+}
+
+/// TODO(#1956): share the retry loop with the monitors in `indexer::tee`.
+async fn fetch_with_retry<T, Fetch, FetchFuture>(fetch: Fetch, error_context: &str) -> T
```
gilcu3: I believe this TODO might be wrong, as the issue seems to talk about something else and is closed #1956

crates/node/src/indexer/foreign_chain.rs
```diff
@@ -0,0 +1,214 @@
...
+fn publish_if_changed<T: PartialEq>(sender: &watch::Sender<T>, value: T) -> bool {
+    sender.send_if_modified(|previous| {
+        if *previous == value {
+            false
+        } else {
+            *previous = value;
+            true
+        }
+    })
+}
```
gilcu3: there seems to be some duplication regarding this function. Not sure if it can be unified though

crates/node/src/indexer/foreign_chain.rs
```diff
@@ -0,0 +1,214 @@
...
+#[expect(non_snake_case)]
+mod tests {
+    use super::*;
+    use ed25519_dalek::SigningKey;
+    use std::cell::Cell;
+
+    fn tls_key(seed: u8) -> dtos::Ed25519PublicKey {
+        let signing_key = SigningKey::from_bytes(&[seed; 32]);
+        dtos::Ed25519PublicKey::from(&signing_key.verifying_key())
+    }
```
gilcu3, line 86: maybe we have some utility function for this already

crates/node/src/indexer/fake.rs
gilcu3, line 213: we currently only call `recompute_available_foreign_chains` here, so if a test in the future mixes some resharing and some foreign chain stuff, it might be using a stale foreign chain config. Not sure if we should try to fix it or just take it as a short-coming of this fake indexer. Just mentioning in case it was not known

gilcu3 (dismissed):
Thank you!

Left some minor comments

crates/node/src/indexer/foreign_chain.rs
kevindeforth: With "other loops", do you refer to the tee-getters? If so, then the only reason those emit info and not error was because, when they were introduced, we updated the node first and it was expected for the call to fail until we also updated the contract.
 
Here, the contract already has all the required API, so emitting an `error` seems correct - we don't expect this call to fail.

We also don't really need the exponential backoff logic that claude is asking for, it's LLM-defensive programming.

kevindeforth (commented):
Thank you, this looks good, with the exception of the refactor. I don't think we need the exponential backoff for these new endpoints.

crates/node/src/indexer/fake.rs
anodar: Added it in other places to align with real indexer.

crates/node/src/indexer/foreign_chain.rs
anodar: Replaced with `dtos::Ed25519PublicKey::from([seed; 32])`

crates/node/src/indexer/foreign_chain.rs
anodar: Done.

crates/node/src/indexer/foreign_chain.rs
anodar: Good catch. Refactored it.

crates/node/src/indexer/foreign_chain.rs
gilcu3: ahh I did remember we moved from `error` to `info` at some point, but did not remember why. If this is the case then yes, having all of them as info seems correct.

> We also don't really need the exponential backoff logic that claude is asking for, it's LLM-defensive programming.

I think we might have added that in part to avoid flooding the logs? Would need to dig it though.

crates/node/src/indexer/foreign_chain.rs
gilcu3: well if we really don't care about that public key that works yeah :)

gilcu3 (dismissed):

crates/node/src/indexer/foreign_chain.rs
kevindeforth: > I think we might have added that in part to avoid flooding the logs?

Yeah, that's true, but that also only applies for when we expect to get an error. Here, we always expect a result, never an error, so not having exponential backoff is fine and saves us code complexity.

kevindeforth:
@claude review

crates/node/src/indexer/fake.rs
```diff
@@ -56,6 +57,8 @@ pub struct FakeMpcContractState {
     supported_foreign_chains_by_node: dtos::ForeignChainSupportByNode,
     available_foreign_chains: dtos::AvailableForeignChains,
     foreign_chains_configs: dtos::ForeignChainsConfigs,
+    /// Mirrors the real indexer's foreign-chain supporters watch channel.
+    foreign_chain_supporters_sender: watch::Sender<ForeignChainSupporters>,
```
kevindeforth: This is a bit odd. We store the `FakeMpcContractState` behind an `Arc::Mutex`, I don't think we need a channel here.

I understand that in this PR, we are inserting some monitoring logic in the `Fake MpcContractState`, but `ForeignChainSupporters` is not really part of the `MpcContractState`. Rather, it is something that is inferred from the contract state.

The problem with putting this logic here is that we now need to keep track of everything that might change this state (c.f. https://github.com/near/mpc/pull/3853#discussion_r3622622407).

Usually, we add these monitoring tasks to the loop inside `FakeIndexerCore::run` and avoid a lot of these head-aches. Was there a reason to deviate from the existing pattern?


https://github.com/near/mpc/blob/b871f5d6cdacc88e9d7baa2c336150079a6a8eff/crates/node/src/indexer/fake.rs#L594-L610

crates/node/src/indexer/foreign_chain.rs
```diff
@@ -0,0 +1,129 @@
...
+                tracing::error!(target: "mpc", "error reading foreign-chain supporters from chain: {:?}", e)
+            }
+        }
+        tokio::time::sleep(FOREIGN_CHAIN_SUPPORTERS_REFRESH_INTERVAL).await;
+    }
+}
+
+/// The two view calls are not atomic: a change finalized between them yields
+/// a transiently inconsistent snapshot, corrected on the next poll.
+async fn read_supporters(indexer_state: &IndexerState) -> anyhow::Result<ForeignChainSupporters> {
```
kevindeforth, line 47: What's the impact such an inconsistent state would have on the consumer?

crates/node/src/indexer/foreign_chain.rs
```diff
@@ -0,0 +1,129 @@
...
+use crate::indexer::IndexerState;
+
+const FOREIGN_CHAIN_SUPPORTERS_REFRESH_INTERVAL: Duration = Duration::from_secs(1);
+
+/// TLS keys of the nodes whose registered config supports each available chain.
+pub type ForeignChainSupporters = BTreeMap<dtos::ForeignChain, BTreeSet<dtos::Ed25519PublicKey>>;
+
+/// Publishes the contract's available chains mapped to their registered
+/// supporters. The previously published value stays in effect until a read succeeds.
+pub async fn monitor_foreign_chain_supporters(
```
kevindeforth: This comment is not ideal. A reader needs to look at the function to understand what "read" this comment is referring to. It would also be helpful if the comment was more specific that with "publishing" it means updating the value in the watch-channel.

kevindeforth (changes requested):
Thank you, this is good, but one blocker remains: that is the wiring inside the fake indexer leads to a tight coupling between contract-logic and node-integration test logic. Following existing patterns would be preferred here _(I missed this in my first review)_.

crates/node/src/indexer/foreign_chain.rs
```diff
@@ -0,0 +1,129 @@
...
+/// a transiently inconsistent snapshot, corrected on the next poll.
+async fn read_supporters(indexer_state: &IndexerState) -> anyhow::Result<ForeignChainSupporters> {
+    let (_height, available_chains) = indexer_state
+        .view_client
+        .get_available_chains(&indexer_state.mpc_contract_id)
+        .await?;
+    let (_height, configs) = indexer_state
+        .view_client
+        .get_foreign_chains_configs(&indexer_state.mpc_contract_id)
+        .await?;
```
kevindeforth: I saw that Claude had a good suggestion about this one, we can query them in parallel instead of sequentially.

crates/node/src/indexer/fake.rs
gilcu3: Right, I missed we had an existing pattern for this. It did feel strange to have to track down stuff in each call

anodar: Moved it into fake indexer. I think that's arguable, but I see your point.

> The problem with putting this logic here is that we now need to keep track of everything that might change this state

We already have to do that anyway, if real contract changes when it recomputes available chains, we have to change that in fakeContractState as well. Only place that indicates there's change in available configs is that `recompute_available_foreign_chains`.  If find it easier to have that model to propagate that state change from there (from `recompute_available_foreign_chains`) rather than introducing monitor unnecessarily.

Yes it deviates bit more from real setup, but we're not really testing real contract here anyway (we do that in e2e tests), we just duplicate whole contract logic into this fake. So we do have to keep track of most of the changes anyway.

crates/node/src/indexer/foreign_chain.rs
anodar: If the state changed, that means either some node added or removed foreign chain support that tipped the scale with threshold so that now network either supports new chain or disabled support for it.

In case of disabling support node might think chain is supported and continue processing that validation request which will fail. In case of adding support it will fail early transiently.

So impact is transient error or late failure.

crates/node/src/indexer/foreign_chain.rs
anodar: Reworded.

crates/node/src/indexer/foreign_chain.rs
anodar: Addressed.

kevindeforth (approved):
Thank you!

crates/node/src/indexer/foreign_chain.rs
kevindeforth: I see, so it's because the contract infers availability from whitelist and configs, and here, we poll availability and configs. Thanks!

gilcu3 (approved):

crates/node/src/indexer/fake.rs
kevindeforth: > > The problem with putting this logic here is that we now need to keep track of everything that might change this state

> We already have to do that anyway, if real contract changes when it recomputes available chains, we have to change that in fakeContractState as well. Only place that indicates there's change in available configs is that `recompute_available_foreign_chains`.


Yeah, I see we still have this one: 
image

I think it would be best to call the recompute function every time we call `FakeMpcContractState::available_foreign_chains`, to ensure we have a current view and can get rid of these invarinats. Wdyt?


> If find it easier to have that model to propagate that state change from there (from recompute_available_foreign_chains) rather than introducing monitor unnecessarily.

The same reasoning would apply to many of the monitoring functions we have in the real indexer, but we just have to acknowledge that this code-base at some point made a choice, for better or worse, and we now need to live with that choice.
We should avoid introducing new patterns if the existing ones gets the job done just fine, otherwise it will become really difficult to maintain and refactor the code.

anodar: > I think it would be best to call the recompute function every time we call FakeMpcContractState::available_foreign_chains, to ensure we have a current view and can get rid of these invarinats. Wdyt?

That does make fake implementation nicer but deviates from real one. Since we're factoring out those monitoring tasks out of this I assume we do want to keep this as close as real one, otherwise would make sense.

> We should avoid introducing new patterns if the existing ones gets the job done just fine, otherwise it will become really difficult to maintain and refactor the code.

I agree.

PR #3855, opened 2026-07-18
test(localnet): deploy and vote in the TEE verifier contract

### LINKED ISSUE #3854: Deploy the TEE verifier contract in `launch-localnet.sh`
## Background

`scripts/launch-localnet.sh` stands up the full local MPC cluster but never deploys the TEE verifier contract. As a result the mpc-contract's `tee_verifier_account_id` stays `None` on localnet, and any Dstack `submit_participant_info` fails with `VerifierNotConfigured`. There is no other tooling that deploys the verifier to localnet.

## Acceptance Criteria

- [ ] `launch-localnet.sh` optionally deploys the stateless `tee-verifier` contract to `tee-verifier.test.near`
- [ ] Every node votes it in via `vote_tee_verifier_change`, so it is adopted as the trusted verifier
- [ ] The step is gated on the verifier WASM being present, so runs that never built it are unaffected

Verifier crate: `crates/tee-verifier` (stateless, single `verify_quote`, no init). Wiring reference: the sandbox test `crates/contract/tests/sandbox/tee_verifier.rs` (`deploy_and_trust_stub`).

pbeza:
Closes #3854

Extend `launch-localnet.sh` to optionally deploy the stateless `tee-verifier` contract to `tee-verifier.test.near` and vote it in as the trusted verifier via `vote_tee_verifier_change` from every node. This mirrors the existing mpc-contract deploy and the `vote_add_domains` loop, so it automatically honors `--nodes`/`--threshold`.

Gated on `--tee-verifier-path` (default `target/near/tee-verifier/tee_verifier.wasm`); when the WASM is absent the verifier steps are skipped, so existing runs are unaffected. The real verifier is stateless, so it is deployed `without-init-call`. `expected_code_hash` is the WASM sha256; voters only need to agree on the same value, which the block computes once and reuses.

Build the verifier WASM before running:

```
cargo near build non-reproducible-wasm --locked --profile=release-contract --features abi --manifest-path crates/tee-verifier/Cargo.toml
```


docs/localnet/tee-verifier-proof.md
gilcu3, line 1: We already have a guide in localnet in  https://github.com/near/mpc/blob/main/docs/localnet/localnet.md and I think this fits as a new section (keeping it short) there, as this is a variant of what already exists that shares many parts

docs/localnet/tee-verifier-proof.md
```diff
@@ -0,0 +1,236 @@
...
+# mpc-node, from this workspace
+cargo install --path crates/node --locked
+```
+
+```console
+$ neard --version
+neard (release 2.13.1) (commit 9d05464c13dca4794c1802e11f45f163ed9936c2) (rustc 1.97.0) (protocol 86) (db 49)
+$ mpc-node --version
+mpc-node 3.13.0
+$ nix develop --command near --version
```
gilcu3: we are mentioning `nix develop` too many times. A small comment at the top saying you can run everything inside nix (using `nix develop`) should be enough. I for example prefer usually the host, so having to copy commands and then remove the nix prefix would be really annoying, and makes the commands longer without good reason.

docs/localnet/tee-verifier-proof.md
```diff
@@ -0,0 +1,236 @@
...
+> Everything runs inside `nix develop`. (A `connection refused (localhost:3030)`
+> error means the chain isn't up — either the script already exited, or you used
+> the host near-cli instead of the nix one.)
+
+## Phase 1 — build and launch (Terminal 1)
+
+Build both contracts. `--no-abi` is required for the verifier: its `verify_quote`
+types don't derive `BorshSchema`, so the default ABI-embedding build fails; the
+plain wasm is all localnet needs.
+
```
gilcu3: in this case we could add a command for building that contract to the cargo-make config, I think we have it for the mpc contract already

docs/localnet/tee-verifier-proof.md
```diff
@@ -0,0 +1,236 @@
...
+The deployed contract ran its real DCAP path (`dcap_qvl::verify` parsed the quote
+and collateral and reached the TCB-expiry check) and correctly rejected: the
+fixture collateral's `nextUpdate` (2026-04-29) is past the live block time. The
+contract works; only the fixture is time-expired on a live clock.
+
+## Phase 4 — tear down (Terminal 1)
+
+Press Enter at the launch script's prompt. It stops the nodes and validator and
+removes `~/.near`.
+
```
gilcu3: We usually did not have docs explaining the script as the canonical guide is https://github.com/near/mpc/blob/main/docs/localnet/localnet.md and the script is just a helper over that

scripts/launch-localnet.sh
```diff
@@ -236,6 +245,37 @@ EOF
...
+    done
+
+    for pid in "${pids_verifier_votes[@]}"; do
+      wait "$pid"
+    done
+  else
+    echo "Warning: TEE verifier WASM not found at ${TEE_VERIFIER_PATH}; skipping verifier deploy and vote." >&2
+    echo "Warning: to include it, build the verifier and pass --tee-verifier-path:" >&2
+    echo "cargo near build non-reproducible-wasm --no-abi --profile=release-contract --manifest-path crates/tee-verifier/Cargo.toml --locked" >&2
+  fi
```
gilcu3: I am not sure about a warning. If we expect this to be a required step, we could just fail if not provided.

.ecrc
```diff
@@ -0,0 +1,3 @@
+{
+  "Exclude": ["\\.borsh$"]
+}
```
gilcu3: why do we need this?

gilcu3 (changes requested):
Looks good overall, but I think we should modify the canonical localnet doc in https://github.com/near/mpc/blob/main/docs/localnet/localnet.md with the necessary extra step instead of creating a new doc

docs/localnet/tee-verifier-proof.md
barakeinav1: But when actually reading this doc, it is too big to a section to https://github.com/near/mpc/blob/main/docs/localnet/localnet.md

and the scope is different, but still worth checking what can be consolidated.

we also have https://github.com/near/mpc/blob/main/docs/localnet/tee-localnet.md which builds upon the localnet.md.

gilcu3: > it is too big to a section to

if added as is it is big yes, but if simplified it could be much smaller IMO

> we also have https://github.com/near/mpc/blob/main/docs/localnet/tee-localnet.md which builds upon the localnet.md.

yeah that one could be consolidated as well. Running in TEE should require some extra step, but in general be pretty much in line with the canonical guide

docs/localnet/tee-verifier-proof.md
```diff
@@ -0,0 +1,236 @@
+# Proving the TEE verifier on localnet
```
barakeinav1: make sure I fully understand the testing scope of this PR
1. script - deploy the verifier contract, and vote for it in the mpc contract.
2. manually sending attestation verification request to the verifier contract.

scripts/launch-localnet.sh
pbeza: https://github.com/near/mpc/pull/3855/commits/534402a2e5ba6e739c414d8d69e049744d4852a7 ✅

https://github.com/near/mpc/blob/534402a2e5ba6e739c414d8d69e049744d4852a7/scripts/launch-localnet.sh#L88-L93

.ecrc
pbeza: If you remove it, [`editorconfig-checker`](https://github.com/editorconfig-checker/editorconfig-checker) will fail:

```
$ cargo make check-all-fast
(...)
[cargo-make] INFO - Execute Command: "editorconfig-checker"
crates/tee-verifier/tests/fixtures/verify_quote_args.borsh:
        Wrong line endings or no final newline
        Not all lines have the correct end of line character
        Wrong character encoding ("MacRoman" instead of "utf-8")
        6: Trailing whitespace

4 errors found
Error while executing command, exit code: 1
```

docs/localnet/tee-verifier-proof.md
pbeza: Rephrased: https://github.com/near/mpc/pull/3855/commits/534402a2e5ba6e739c414d8d69e049744d4852a7. ✅

docs/localnet/tee-verifier-proof.md
pbeza: https://github.com/near/mpc/pull/3855/commits/534402a2e5ba6e739c414d8d69e049744d4852a7 ✅

https://github.com/near/mpc/blob/ada8871050c5d1041fdae9f70976150bf15e6230/Makefile.toml#L185-L197

.ecrc
gilcu3: interesting. Do we have some centralized place for such exceptions? Could be slightly more friendly given that these are non-standard files

docs/localnet/tee-verifier-proof.md
pbeza: https://github.com/near/mpc/pull/3855/commits/534402a2e5ba6e739c414d8d69e049744d4852a7 + https://github.com/near/mpc/pull/3855/commits/ade20d91a7aebe4fd020cd30a1f96449a65f794b ✅

https://github.com/near/mpc/blob/ade20d91a7aebe4fd020cd30a1f96449a65f794b/docs/localnet/localnet.md?plain=1#L15

docs/localnet/tee-verifier-proof.md
pbeza: Consolidated and rephrased here: https://github.com/near/mpc/pull/3855/commits/534402a2e5ba6e739c414d8d69e049744d4852a7. ✅

docs/localnet/tee-verifier-proof.md
pbeza: Yes, correct, that’s the scope. What’s [missing](https://nearone.slack.com/archives/C0912BTG51T/p1784628617285429?thread_ts=1784619753.611879&cid=C0912BTG51T) here is automating localnet with real TEE attestation.

.ecrc
pbeza: I can probably move it to `https://github.com/near/mpc/blob/main/.editorconfig`... \<WIP\>

pbeza: > I can probably move it to `https://github.com/near/mpc/blob/main/.editorconfig`... <WIP>

Done: https://github.com/near/mpc/pull/3855/commits/f253f63f20d845169186d444eb5587fde9e4fcb6. ✅

https://github.com/near/mpc/blob/f253f63f20d845169186d444eb5587fde9e4fcb6/.editorconfig#L28-L34

gilcu3 (dismissed):
Thank you!

docs/localnet/localnet.md
```diff
@@ -293,7 +295,65 @@ near contract call-function as-transaction mpc-contract.test.near vote_add_domai
 near contract call-function as-transaction mpc-contract.test.near vote_add_domains file-args docs/localnet/args/add_domain.json prepaid-gas '300.0 Tgas' attached-deposit '0 NEAR' sign-as sam.test.near network-config mpc-localnet sign-with-keychain send
 ```
 
-## 6. Send a sign request to the network
+## 6. (Optional) Deploy and vote in the TEE verifier
```
barakeinav1, line 298: is this really "optional"? or only before we merge the verifier contract lats PR?

barakeinav1: answering my own question, optional since for mock attestation we don't need the verifier contract

crates/tee-verifier/tests/fixtures/verify_quote_args.borsh
barakeinav1, line 1: are you hiding a backdoor here :-) 

https://www.akamai.com/blog/security-research/critical-linux-backdoor-xz-utils-discovered-what-to-know

scripts/launch-localnet.sh
```diff
@@ -77,6 +85,13 @@ main() {
     exit 1
   fi
 
+  if [[ ! -f "$TEE_VERIFIER_PATH" ]]; then
```
barakeinav1, line 88: I think this makes the verifier mandatory — it `exit 1`s when the WASM isn't built, and the deploy/vote below runs unconditionally. But the docs call it "(Optional)" and the PR says absent = skipped, so this'll break existing localnet runs for anyone who hasn't built it. And it genuinely is optional: localnet nodes use Mock attestations (verified in-WASM), so nothing here calls `verify_quote` unless you do it by hand.

barakeinav1:
nit (PR description): the "Build the verifier WASM before running" command in the body uses `--features abi`, but the tee-verifier can't build with ABI (no `BorshSchema`) — the docs and the new `build-tee-verifier-optimized` task correctly use `--no-abi`. Copying the body command as-is fails; worth updating it to `--no-abi` (or just pointing at `cargo make build-tee-verifier-optimized`).

crates/tee-verifier/tests/fixtures/verify_quote_args.borsh
anodar: Looks like this is expected to be same as `quote.json` but borsch encoded `https://github.com/near/mpc/blame/deb99b3c6416e2be91b2778f4c89bf203e317e0b/crates/test-utils/assets/quote.json (commit from you)

Are you in this ("hiding a backdoor") together with Patrick ? :D 

Jokes aside, would be nice to also commit a script that was used to generate either this borsh encoding or quote.json bytes.

docs/localnet/localnet.md
```diff
@@ -293,7 +295,65 @@ near contract call-function as-transaction mpc-contract.test.near vote_add_domai
...
+near contract call-function as-transaction mpc-contract.test.near vote_tee_verifier_change json-args '{"candidate_account_id":"tee-verifier.test.near","expected_code_hash":"'"$TEE_VERIFIER_HASH"'"}' prepaid-gas '300.0 Tgas' attached-deposit '0 NEAR' sign-as frodo.test.near network-config mpc-localnet sign-with-keychain send
+
+near contract call-function as-transaction mpc-contract.test.near vote_tee_verifier_change json-args '{"candidate_account_id":"tee-verifier.test.near","expected_code_hash":"'"$TEE_VERIFIER_HASH"'"}' prepaid-gas '300.0 Tgas' attached-deposit '0 NEAR' sign-as sam.test.near network-config mpc-localnet sign-with-keychain send
+```
+
+Once both votes agree, the change is applied and the MPC contract has no pending
+verifier votes:
+
+```shell
+near contract call-function as-read-only mpc-contract.test.near tee_verifier_votes json-args {} network-config mpc-localnet now
```
barakeinav1, line 335: I just noticed that we don't have any view method to check what verifier the contract is using.
e.g something like get_attastation_verifier_id view method.

@pbeza - we should create a follow up issue for this

anodar (dismissed):

crates/tee-verifier/tests/fixtures/verify_quote_args.borsh
gilcu3: there is another test in the PR doing the borsh match. For the json that's harder as it comes from a TDX machine

scripts/launch-localnet.sh
```diff
@@ -236,6 +251,29 @@ EOF
...
+  pids_verifier_votes=()
+  for ((i = 1; i <= N; i++)); do
+    node_name="mpc-node-$i.test.near"
+    run_quiet_on_success "near contract call-function as-transaction mpc-contract.test.near vote_tee_verifier_change json-args '{\"candidate_account_id\":\"tee-verifier.test.near\",\"expected_code_hash\":\"$TEE_VERIFIER_HASH\"}' prepaid-gas '300.0 Tgas' attached-deposit '0 NEAR' sign-as ${node_name} network-config mpc-localnet sign-with-keychain send" &
+    pids_verifier_votes+=($!)
+  done
+
+  for pid in "${pids_verifier_votes[@]}"; do
+    wait "$pid"
+  done
```
barakeinav1, line 275: nit: after waiting on the votes, could we confirm the change actually landed by polling `tee_verifier_votes` until it's empty (the docs already show this check)? Right now the script fires the votes and moves on — a self-validate step here would match how it `wait_for_success`es on the contract reaching Running, and catch a vote that silently didn't go through.

barakeinav1 (commented):
added a few comments, but nothing that is blocking.
but I didn't try to run the script or the tests

crates/tee-verifier/tests/fixtures/verify_quote_args.borsh
pbeza: > Jokes aside, would be nice to also commit a script that was used to generate either this borsh encoding or quote.json bytes.

@anodar, the generator already exists:

https://github.com/near/mpc/blob/f253f63f20d845169186d444eb5587fde9e4fcb6/crates/tee-verifier/tests/verify_quote.rs#L181-L207

You can regenerate it yourself:
```
UPDATE_FIXTURES=1 cargo test -p tee-verifier --test verify_quote verify_quote_args_fixture
```

docs/localnet/localnet.md
pbeza: Good point. Done: https://github.com/near/mpc/issues/3889. ✅

scripts/launch-localnet.sh
pbeza: https://github.com/near/mpc/pull/3855/commits/bf84f1ef47e97b12f117dcfc190386579fb7495c ✅

scripts/launch-localnet.sh
pbeza: > I think this makes the verifier mandatory (...)

It’s intentional. See: https://github.com/near/mpc/pull/3855#discussion_r3619891390.

gilcu3 (approved):

barakeinav1 (approved):
thanks

PR #3869, opened 2026-07-20
ci: add semgrep job in the ci workflow

### LINKED ISSUE #594: Add semgrep to CI checks
To be refined

anodar:
closes: https://github.com/near/mpc/issues/594

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -259,16 +259,14 @@ jobs:
   # Static analysis.
   # silence false positives with a `nosemgrep: <rule-id>` comment
   # or a path entry in .semgrepignore.
-  semgrep:
-    name: "Semgrep"
+  opengrep:
+    name: "Opengrep"
     needs: changes
     if: ${{ always() && needs.changes.outputs.code_changed != 'false' }}
     runs-on: warp-ubuntu-2404-x64-2x
     timeout-minutes: 15
     permissions:
       contents: read
-    container:
-      image: semgrep/semgrep:1.170.0@sha256:c98f8829eea377274ee4b10656458b078b88232469b2ff913f091c2317347c9d
 
     steps:
       - name: Checkout repository
@@ -276,19 +274,14 @@ jobs:
         with:
           persist-credentials: false
 
-      # Only WARNING and ERROR findings block.
-      # panic-in-function-returning-result is excluded: docs/engineering-standards.md
-      # allows panics only in narrow cases, but we have a lot of such narrow cases.
-      - name: Run semgrep
-        run: >
-          semgrep scan
-          --config p/trailofbits
-          --config p/rust
-          --config p/secrets
-          --exclude-rule trailofbits.rs.panic-in-function-returning-result.panic-in-function-returning-result
-          --severity WARNING
-          --severity ERROR
-          --error
+      - name: Install opengrep
+        run: |
+          curl -fsSL -o opengrep 'https://github.com/opengrep/opengrep/releases/download/v1.25.0/opengrep_manylinux_x86'
+          echo '9ac4aebb47ba3f7b0d8fc641ac8749cb6c2f253f616131a67d9631e00d4bea33  opengrep' | sha256sum -c -
+          sudo install -m755 opengrep -t /usr/bin
+
+      - name: Run opengrep
+        run: bash scripts/run-opengrep.sh
 
   mpc-e2e-tests:
     name: "MPC E2E tests"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -274,14 +274,14 @@ jobs:
         with:
           persist-credentials: false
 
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -256,6 +256,40 @@ jobs:
...
+      # Only WARNING and ERROR findings block.
+      # panic-in-function-returning-result is excluded because intentional
+      # panics on invariant violations are allowed by docs/engineering-standards.md.
+      - name: Run semgrep
+        run: >
+          semgrep scan
+          --config p/trailofbits
+          --config p/rust
+          --config p/secrets
+          --exclude-rule trailofbits.rs.panic-in-function-returning-result.panic-in-function-returning-result
```
anodar: We have quite a bit of places where we do this, so decided to exclude it instead of adding annotation to over 20 places.

haiyuechen-nearone: Should we maybe annotate those call sites properly as a follow up?

anodar: Created https://github.com/orgs/near/projects/179/views/1?pane=issue&itemId=216208491&issue=near%7Cmpc%7C3877

.github/workflows/ci.yml
```diff
@@ -256,6 +256,33 @@ jobs:
...
+          persist-credentials: false
+
+      - name: Install Nix
+        uses: DeterminateSystems/nix-installer-action@ef8a148080ab6020fd15196c2084a2eea5ff2d25 # v22
+
+      - name: Restore /nix from cache
+        uses: ./.github/actions/restore-nix-cache
+
+      - name: Run opengrep
+        run: nix shell .#opengrep -c bash scripts/run-opengrep.sh
```
gilcu3, line 284: nit: we could use cargo-make here, but this is also fine ofc

nix/opengrep.nix
```diff
@@ -0,0 +1,50 @@
...
+    };
+    aarch64-linux = {
+      name = "opengrep_manylinux_aarch64";
+      hash = "sha256-/UASQnLQBggqVZSxmuzuB7Ad1Qkz2K3XpP1cVX0r5fY=";
+    };
+    aarch64-darwin = {
+      name = "opengrep_osx_arm64";
+      hash = "sha256-NUP8q66dsq5byXSjt1QmNT8KPjaRgbIVfvJ/RoZ5lsg=";
+    };
+  };
```
gilcu3, line 24: are we fine with downloading binaries for this one? IIRC we were leaning towards the hard core build everything in nix stuff :)

gilcu3 (approved):
Thank you!

haiyuechen-nearone (approved):

.github/workflows/ci.yml
anodar: Leaving as is as it's marked nit.

Also, I assume you mean `nix develop --command cargo make opengrep` but that will need to build everything (in case of cache miss, if I understand it correctly) hereas this will be lot more isolated.

nix/opengrep.nix
anodar: I've looked a bit into it now, the ones that we build is different because it's rust tools published in crates.io and we can pin hash and build with locked versions. I don't see any non-rust tool that we build here.

On the other hand, while opengrep publishes their flake, they use `opam` package manager for OCalm (which is apparently language opengrep is written on). [To nixify their opam deps they use opam-nix](https://github.com/opengrep/opengrep/blob/d77978ddcc4c3bf0924fe1230026efa624aa70e8/flake.nix#L218) but somewhere down the dependency graph something's not pinned and it can't be built without `--impure`:

```
nix build  "github:opengrep/opengrep/v1.25.0?submodules=1#opengrep"
error:
       … while calling a functor (an attribute set with a '__functor' attribute)
         at «github:opengrep/opengrep/27b5fba1b315fbc63d0fd474d01f5d8235cc86e2»/flake.nix:413:11:
          412|         pyopengrep = with pythonPackages;
          413|           buildPythonApplication {
             |           ^
          414|             # thanks to @06kellyjac

       … while evaluating a branch condition
         at «github:nixos/nixpkgs/4bd9165a9165d7b5e33ae57f3eecbcb28fb231c9?narHash=sha256-l/iNYDZ4bGOAFQY2q8y5OAfBBtrDAaPuRQqWaFHVRXM%3D»/pkgs/development/interpreters/python/python-packages-base.nix:32:7:
           31|       in
           32|       if lib.isAttrs result then
             |       ^
           33|         result

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: [opam-nix] a git dependency without an explicit sha1 is not supported in pure evaluation mode; try with --impure
```

Maybe we can work around this but not sure it's worth going down the rabbit hole.

gilcu3: yeah, ocaml makes it much harder here, thank you for taking a look

PR #3870, opened 2026-07-20
refactor(chain-gateway): unify view calls behind a single ViewContract trait

### LINKED ISSUE #3865: refactor(chain-gateway): unify view calls behind a single `ViewContract` trait
### Background

chain-gateway has several ways to express a contract view call:
- `QueryViewFunction` takes `(method_name: &str, args: &[u8])`
- `ViewMethod` JSON-serializes a typed `Arg`
- `subscribe_to_contract_method` hardcodes `b"{}"` (restricting subscriptions to zero-argument views).

We want to unify these interfaces to have a single view-request argument type.
This is part of #3693 

### User Story

As a developer using chain-gateway, I want one type that expresses a view request, so that one-shot reads and subscriptions share the same vocabulary and subscriptions are no longer limited to zero-argument views.

### Acceptance Criteria

- [ ] `ViewArgs { method_name, args }` with `new`/`no_args` constructors is the input to `view_raw`, `view_method`, and `subscribe_to_contract_method`.
- [ ] `QueryViewFunction` is renamed to `ViewContract` (owned `ViewArgs` parameter); exactly one view impl per backend (`ChainGateway`, `NearViewClientActorHandle`, `MockChainState`).
- [ ] `NoArgs` and the now-unused `ChainGatewayError::Serialization` variant are removed.
- [ ] chain-gateway unit, doc, and integration tests pass.

`view_method` loses its `Arg` generic (callers encode args; the typed-encode convenience returns with the planned view-descriptor layer). Mock builder renamed to `with_view_response` to match `set_view_response`.

This is part of #3693

kevindeforth:
resolves #3865, part of https://github.com/near/mpc/issues/3693

- [x] `ViewArgs { method_name, args }` with `new`/`no_args` constructors is the input to `view_raw`, `view_method`, and `subscribe_to_contract_method`.
- [x] `QueryViewFunction` is renamed to `ViewContract` (owned `ViewArgs` parameter); exactly one view impl per backend (`ChainGateway`, `NearViewClientActorHandle`, `MockChainState`).
- [x] `NoArgs` and the now-unused `ChainGatewayError::Serialization` variant are removed.
- [x] chain-gateway unit, doc, and integration tests pass.

### DIFF
--- a/crates/chain-gateway/src/state_viewer/traits.rs
+++ b/crates/chain-gateway/src/state_viewer/traits.rs
@@ -1,10 +1,12 @@
 use std::future::Future;
 
 use crate::errors::{ChainGatewayError, ChainGatewayOp};
-use crate::primitives::{IsSyncing, QueryViewFunction};
+use crate::primitives::IsSyncing;
+use crate::primitives::ViewContract;
 use crate::types::ObservedState;
+use crate::types::ViewArgs;
 use near_account_id::AccountId;
-use serde::{Serialize, de::DeserializeOwned};
+use serde::de::DeserializeOwned;
 
 use super::subscription::ContractMethodSubscription;
 
@@ -15,6 +17,7 @@ use super::subscription::ContractMethodSubscription;
 /// # Example
 ///
 /// ```
+/// use chain_gateway::ViewArgs;
 /// use chain_gateway::mock::{MockChainState, Call};
 /// use chain_gateway::state_viewer::{WatchContractState, SubscribeToContractMethod};
 /// use chain_gateway::types::ObservedState;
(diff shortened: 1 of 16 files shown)


crates/chain-gateway/src/state_viewer/traits.rs
```diff
@@ -84,25 +94,22 @@ pub trait SubscribeToContractMethod {
 /// }
 /// ```
 pub trait ViewMethod {
-    fn view_method<Arg, Res>(
+    fn view_method<Res>(
         &self,
         contract_id: AccountId,
-        method_name: &str,
-        args: &Arg,
```
kevindeforth, line 91: This is the only functional change in the PR: `view_method` no longer accepts generic `Args`.

We initially wanted this to improve ergonomics, but it seems somewhat counter-productive in achieving that goal.

crates/chain-gateway/src/types.rs
```diff
@@ -12,9 +12,26 @@ pub(crate) fn to_action_gas(gas: NearGas) -> Gas {
...
 }
 
-/// An empty argument struct for contract view calls that take no arguments.
-#[derive(Serialize)]
-pub struct NoArgs {}
+/// A NEAR view-function request: method name and encoded args.
+#[derive(Debug, Clone)]
+pub struct ViewArgs {
+    pub method_name: String,
+    pub args: Vec<u8>,
```
haiyuechen-nearone, line 19: QQ: Is it the correct understanding that serialization of args is the responsibility of the individual call sites going forward?

haiyuechen-nearone (approved):
LGTM as a refactor. Left a question about the plan going forward

crates/chain-gateway/src/types.rs
kevindeforth: Yes, that's correct. We don't really need typed args here once we address #3693.
Allowing typed args opens up multiple ways to implement the same thing, which makes it harder for developers to identify the "correct" way of doing it.

anodar (approved):
Nice!

PR #3871, opened 2026-07-20
refactor: near-contract-transport crate for NEAR call/view vocabulary

kevindeforth:
resolves #3866, part of #3693 

- [x] Crate renamed `mpc-call-args` → `near-contract-transport`; `ViewArgs`, `ViewContract`, `BlockHeight`, `ObservedState` move in.
- [x] `ObservedState<T, H = BlockHeight>` is generic over the height witness and `ViewContract` gains `type ObservedAt` (`BlockHeight` for chain-gateway backends; `()` later for backends that cannot report a height) — no `Option`, no placeholder heights.
- [x] chain-gateway re-exports none of the moved types; consumers (mpc-node, tee-context, chain-gateway integration tests) depend on `near-contract-transport` directly.

PR #3872, opened 2026-07-20
feat(near-contract-transport): Introducing `CallContract` trait enabling different back-ends to share the same contract interface.

### LINKED ISSUE #3867: feat(near-contract-transport): Introducing CallContract trait enabling different back-ends to share the same contract interface.
### Background

Submitting a contract call through chain-gateway requires picking a `TransactionSigner` and calling `submit_function_call_tx` manually at every site. Introduce the `CallContract` trait (in `near-contract-transport`) and `AccountCaller` in chain-gateway.

Depends on #3866.

### User Story

As a component that submits contract transactions (tee-context today, the node after the gateway migration), I want a caller, so that signer selection and nonce handling are encapsulated behind one seam.

### Acceptance Criteria

- [ ] `CallContract { type Output; type Error; }` in `near-contract-transport` (behind the `traits` feature), with a `&T` blanket impl.
- [ ] `AccountCaller<T: SubmitFunctionCall>` in chain-gateway over an owned `NonEmptyVec<TransactionSigner>` (signers are `!Clone` — one nonce counter per key by construction).
- [ ] Unit tests: round-robin across the pool; backend error propagation.

`TransactionSigner` ownership is deliberate: sharing a key between two callers would fork its nonce counter.

kevindeforth:
resolves #3867, part of #3693 

Introduces the `CallContract` trait in the `near-contract-transport` crate and implements the trait for the chain-gateway crate. Implementations for our other back-ends is to follow (e2e test, devnet, sandbox, chain-gateway, node, tee-context,...).

We will then implement the contract-interface over `CallContract`. This has two benefits:
- it will allow us to finally achieve #3693;
- integrating the chain-gateway with the node will become much easier, as they will share the same interface.

Changes:
- [x] `CallContract { type Output; type Error; }` in `near-contract-transport` (behind the `traits` feature), with a `&T` blanket impl
- [x] `AccountCaller<T: SubmitFunctionCall>` in chain-gateway over an owned `NonEmptyVec<TransactionSigner>` (signers are `!Clone` — one nonce counter per key by construction)
- [x] Unit tests: round-robin across the pool; backend error propagation.

### DIFF
--- a/crates/chain-gateway/Cargo.toml
+++ b/crates/chain-gateway/Cargo.toml
@@ -18,6 +18,7 @@ near-contract-transport = { workspace = true, features = ["traits"] }
 near-crypto = { workspace = true }
 near-indexer = { workspace = true }
 near-indexer-primitives = { workspace = true }
+near-mpc-bounded-collections = { workspace = true }
 near-o11y = { workspace = true }
 nearcore = { workspace = true }
 prometheus = { workspace = true }
--- a/crates/chain-gateway/src/transaction_sender/caller.rs
+++ b/crates/chain-gateway/src/transaction_sender/caller.rs
@@ -1,0 +1,140 @@
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+use near_account_id::AccountId;
+use near_contract_transport::{CallContract, FunctionCallArgs};
+use near_indexer_primitives::CryptoHash;
+use near_mpc_bounded_collections::NonEmptyVec;
+
+use crate::errors::ChainGatewayError;
+use crate::transaction_sender::{SubmitFunctionCall, TransactionSigner};
+
+pub struct AccountCaller<T> {
+    submitter: T,
+    signers: NonEmptyVec<TransactionSigner>,
+    next: AtomicUsize,
+}
+
+impl<T> AccountCaller<T> {
+    pub fn new(submitter: T, signers: NonEmptyVec<TransactionSigner>) -> Self {
+        Self {
+            submitter,
+            signers,
+            next: AtomicUsize::new(0),
+        }
+    }
+}
+
+impl<T: SubmitFunctionCall + Sync> CallContract for AccountCaller<T> {
+    type Output = CryptoHash;
+    type Error = ChainGatewayError;
+
+    async fn call_contract(
+        &self,
+        contract_id: &AccountId,
+        call_args: FunctionCallArgs,
+    ) -> Result<CryptoHash, ChainGatewayError> {
+        let signer = &self.signers[self.next.fetch_add(1, Ordering::Relaxed) % self.signers.len()];
+        self.submitter
+            .submit_function_call_tx(signer, contract_id.clone(), call_args)
+            .await
+    }
+}
+
+#[cfg(test)]
+#[expect(non_snake_case)]
+mod tests {
+    use super::AccountCaller;
+    use crate::errors::{ChainGatewayError, ChainGatewayOp};
+    use crate::mock::{MockChainStateBuilder, MockError};
+    use crate::transaction_sender::TransactionSigner;
+    use crate::types::LatestFinalBlockInfo;
+    use ed25519_dalek::{SigningKey, Verifier, VerifyingKey};
+    use near_account_id::AccountId;
+    use near_contract_transport::{CallContract, FunctionCallArgs, NearGas, NearToken};
+    use near_indexer::near_primitives::transaction::SignedTransaction;
+    use near_indexer_primitives::CryptoHash;
+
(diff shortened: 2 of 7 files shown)


crates/near-contract-transport/src/traits.rs
```diff
@@ -2,9 +2,35 @@ use std::future::Future;
...
+    /// Backend-specific successful call outcome.
+    type Output;
+    type Error;
+
+    fn call_contract(
+        &self,
+        contract_id: &AccountId,
+        call_args: FunctionCallArgs,
+    ) -> impl Future<Output = Result<Self::Output, Self::Error>> + Send;
+}
```
kevindeforth, line 19: reviewers are invited to take a look at https://github.com/near/mpc/pull/3873 for an example of how this will be used.

crates/chain-gateway/src/transaction_sender/caller.rs
```diff
@@ -0,0 +1,140 @@
...
+impl<T: SubmitFunctionCall + Sync> CallContract for AccountCaller<T> {
+    type Output = CryptoHash;
+    type Error = ChainGatewayError;
+
+    async fn call_contract(
+        &self,
+        contract_id: &AccountId,
+        call_args: FunctionCallArgs,
+    ) -> Result<CryptoHash, ChainGatewayError> {
+        let signer = &self.signers[self.next.fetch_add(1, Ordering::Relaxed) % self.signers.len()];
```
netrome: Nit: I'm always a bit hesitant to use the slicing operator since it's an implicit panic. This is obviously correct due to the `% self.signers.len()`, but I would prefer a more explicit `.get(...).expect("cannot fail because of <reason>")`

netrome (dismissed):
Just one nit, otherwise looks great!

netrome (approved):

crates/chain-gateway/src/transaction_sender/caller.rs
kevindeforth: resolved in https://github.com/near/mpc/pull/3872/commits/fac9e791894af00fdfb6639f55028b9cde7e5679

crates/chain-gateway/Cargo.toml
```diff
@@ -18,6 +18,7 @@ near-contract-transport = { workspace = true, features = ["traits"] }
 near-crypto = { workspace = true }
 near-indexer = { workspace = true }
 near-indexer-primitives = { workspace = true }
+near-mpc-bounded-collections = { workspace = true }
```
SimonRastikian, line 21: Independent of this PR. I always wondered why we sometimes have crates that start with near-<something> e.g. near-contract-transport and other don't e.g. contract.

crates/tee-context/Cargo.toml
```diff
@@ -7,9 +7,10 @@ license = { workspace = true }
 [dependencies]
 chain-gateway = { workspace = true }
 mpc-primitives = { workspace = true }
-near-account-id = { workspace = true }
 near-contract-transport = { workspace = true }
 near-mpc-contract-interface = { workspace = true, features = ["call-args"] }
+
+near-account-id = { workspace = true }
```
SimonRastikian, line 13: Why was this changed?

crates/chain-gateway/src/transaction_sender/caller.rs
```diff
@@ -0,0 +1,144 @@
...
+            args: b"{}".to_vec(),
+            gas: NearGas::from_gas(1),
+            deposit: NearToken::from_yoctonear(0),
+        }
+    }
+
+    fn signed_by(tx: &SignedTransaction, verifying_key: &VerifyingKey) -> bool {
+        match &tx.signature {
+            near_crypto::Signature::ED25519(sig) => {
+                verifying_key.verify(&tx.get_hash().0, sig).is_ok()
```
SimonRastikian, line 73: Why is this oriented to EdDSA only?

SimonRastikian (approved):

crates/chain-gateway/Cargo.toml
gilcu3: all crates that start with near-mpc are our crates that at some point we published in crates.io
image

but they are stale by now so not sure if anyone is actually using them

crates/tee-context/Cargo.toml
kevindeforth: it's a carry-over from the last PR: https://github.com/near/mpc/pull/3871#discussion_r3622462312

crates/chain-gateway/src/transaction_sender/caller.rs
kevindeforth: I don't think we are using anything else at the moment for signing near-transactions.

Issue #903, opened 2025-08-21
[Docs-Missing] Document cost in NEAR tokens for calling submit_participant_info


barakeinav1:
### Background

Flow up of task #622 - TEE node operator guide.  
The guide at [docs/running_an_mpc_node_in_tdx_external_guide.md](https://github.com/near/mpc/blob/main/docs/running_an_mpc_node_in_tdx_external_guide.md), under the **“Joining the MPC Cluster → Submitting participant info”** section, contains the following placeholder:  

doc snippet:  
```text
Note - Calling this method will cost TBD Near to none participants.

```

This indicates missing documentation about the expected NEAR token cost for calling `submit_participant_info`. Without this information, operators cannot plan account funding properly.

### User Story

As an operator, I would like the documentation to be complete and readable, so that I can follow the guide without encountering missing information.

### Acceptance Criteria

- Determine the required NEAR token cost for calling `submit_participant_info`.  
- Document the exact (or estimated) amount needed.  
- Clarify whether the cost differs between testnet and mainnet.  
- Review updated guide for accuracy and completeness.  

### Resources & Additional Notes

Operator user guide section: [docs/running_an_mpc_node_in_tdx_external_guide.md](https://github.com/near/mpc/blob/main/docs/running_an_mpc_node_in_tdx_external_guide.md) → *Joining the MPC Cluster → Submitting participant info*

netrome:
Still relevant

netrome:
Downgrading to P4, unclear if we want to tackle this.

barakeinav1:
Closing via #3940, which makes the answer "nothing".

`submit_participant_info` is no longer `#[payable]`: it takes no deposit and rejects one if attached. The contract funds the stored attestation entry from its own balance — 604 bytes, ~0.006 NEAR, pinned by `stored_attestation_entry__should_have_the_pinned_size`. There is no caller-facing amount left to document.

The runbook bullet that sent people here — *"first-time joiners must attach enough yoctoNEAR for storage; call `submit_participant_info` manually with `--deposit` once. Exact amount tracked in #903"* — has been removed, since following it now fails with `Method submit_participant_info doesn't accept deposit`.

Should the funding model change later, #3972 tracks moving the storage cost onto whoever onboards a node. That would introduce a *new* deposit with its own amount to document, so it belongs in that issue rather than reopening this one.

barakeinav1:
Reopening the substance of this briefly, since the resolution no longer holds.

[...]

Issue #166, opened 2025-01-30
Decide attached gas based on transaction type


saketh-are:
Currently we attach the maximum of 300 Tgas to every outgoing transaction:

https://github.com/Near-One/mpc/blob/2ff7716f9171d6c14f6a08e611dbff4e618c87a4/node/src/indexer/transaction.rs#L40-L45

As a result, we are able to send fewer `respond` txs before running into congestion limits. Based on testing `respond` calls consume only 7.62 Tgas, so we should lower the attached gas to 10 Tgas.

We have introduced other types of outgoing transactions as well, which may consume more or less gas:

https://github.com/Near-One/mpc/blob/2ff7716f9171d6c14f6a08e611dbff4e618c87a4/node/src/indexer/response.rs#L124-L133

The attached gas should be decided based on the transaction type.

netrome:
Still relevant but low prio.

netrome:
Setting as P4

PR #3873, opened 2026-07-20
feat: introduce MpcContractHandle as shared contract interface

### LINKED ISSUE #3868: feat: introduce MpcContractHandle as shared contract interface
### Background

We have many different back-ends calling the mpc-contract in different ways. `TeeContext` builds `FunctionCallArgs` by hand (method-name constants, its own gas constants) and takes a `&TransactionSigner` on every call. Introduce the typed client in `near-mpc-contract-interface` — `MpcContractHandle<C: CallContract>`, the single source of each method's wire format — with its first two methods (`submit_participant_info`, `verify_tee`), and make `TeeContext` hold `MpcContractHandle<AccountCaller<S>>` with the signer absorbed at construction.

In future PRs, we will extend other back-ends to implement `CallContract` and start to rely exclusively on the `MpcContractHandle` for making calls to our contract (c.f. #3867).

### User Story

As the maintainer of tee-context, I want contract calls to go through a typed handle whose wire format is defined once, so that call sites cannot drift from the contract interface and future methods land with one definition.

### Acceptance Criteria

- [ ] `near-mpc-contract-interface` gains a `client` feature and module: `MpcContractHandle`, `MpcContractHandleError<E>`, wire-format snapshot tests (one catalog snapshot for all methods).
- [ ] `TeeContext::new` takes a `TransactionSigner`; `submit_attestation`/`verify_tee` lose the per-call signer parameter.
- [ ] `TeeContextError` collapses to `HashWatcherClosed` + `ContractCall`; propagation tests assert the nested error kinds.
- [ ] The `client` feature stays out of the contract's wasm build.

The snapshot catalog makes each method's method name, gas, deposit, and JSON body a reviewable artifact; drift fails as a snapshot diff.

kevindeforth:
resolves #3868, part of #3693 

This PR introduces the `MpcContractHandle`, which will serve as a single source of truth for making contract calls in all our different back-ends (e2e test,  devnet, node, sandbox,...).

Changes:

- [x] `near-mpc-contract-interface` gains a `client` feature and module: `MpcContractHandle`, `MpcContractHandleError<E>`, wire-format snapshot tests (one catalog snapshot for all methods).
- [x] `TeeContext::new` takes a `TransactionSigner`; `submit_attestation`/`verify_tee` lose the per-call signer parameter.
- [x] `TeeContextError` collapses to `HashWatcherClosed` + `ContractCall`; propagation tests assert the nested error kinds.
- [x] The `client` feature stays out of the contract's wasm build.

### DIFF
--- a/crates/near-mpc-contract-interface/src/client.rs
+++ b/crates/near-mpc-contract-interface/src/client.rs
@@ -78,8 +78,6 @@ impl<C: CallContract> MpcContractHandle<C> {
 pub enum MpcContractHandleError<E> {
     #[error("failed to serialize call arguments: {0}")]
     Serialize(#[from] serde_json::Error),
-    #[error("failed to borsh-encode call arguments: {0}")]
-    Encode(#[from] std::io::Error),
     #[error("contract call failed: {0}")]
     Call(E),
 }
--- a/crates/near-mpc-contract-interface/src/client.rs
+++ b/crates/near-mpc-contract-interface/src/client.rs
@@ -1,0 +1,160 @@
+//! Typed client for the NEAR MPC signer contract.
+//!
+//! [`MpcContractHandle`] is the single source of each method's wire format
+//! (method name, argument struct, gas, deposit), generic over a transport
+//! backend implementing [`CallContract`].
+
+use near_contract_transport::{CallContract, FunctionCallArgs, NearGas, NearToken};
+
+use crate::call_args::SubmitParticipantInfoArgs;
+use crate::method_names::{SUBMIT_PARTICIPANT_INFO, VERIFY_TEE};
+use crate::types::{AccountId, Attestation, Ed25519PublicKey};
+
+/// Gas attached to every handle-issued call.
+// TODO(#166): 300 Tgas is the protocol maximum and higher than most methods
+// need; benchmark per method and reduce.
+pub const MAX_GAS: NearGas = NearGas::from_gas(300_000_000_000_000);
+
+pub const SUBMIT_PARTICIPANT_INFO_DEPOSIT_MILLINEAR: u128 = 100;
+
+/// Typed interface to the MPC signer contract at a fixed account, generic over
+/// the transport backend `C`.
+#[derive(Clone)]
+pub struct MpcContractHandle<C> {
+    caller: C,
+    contract_id: AccountId,
+}
+
+impl<C> MpcContractHandle<C> {
+    pub fn new(caller: C, contract_id: AccountId) -> Self {
+        Self {
+            caller,
+            contract_id,
+        }
+    }
+}
+
+impl<C: CallContract> MpcContractHandle<C> {
+    pub async fn submit_participant_info(
+        &self,
+        proposed_participant_attestation: Attestation,
+        tls_public_key: Ed25519PublicKey,
+    ) -> Result<C::Output, MpcContractHandleError<C::Error>> {
+        let args = serde_json::to_vec(&SubmitParticipantInfoArgs::new(
+            proposed_participant_attestation,
+            tls_public_key,
+        ))?;
+        self.caller
+            .call_contract(
+                &self.contract_id,
+                FunctionCallArgs {
+                    method_name: SUBMIT_PARTICIPANT_INFO.to_string(),
+                    args,
+                    gas: MAX_GAS,
+                    deposit: NearToken::from_millinear(SUBMIT_PARTICIPANT_INFO_DEPOSIT_MILLINEAR),
+                },
+            )
+            .await
+            .map_err(MpcContractHandleError::Call)
+    }
+
+    pub async fn verify_tee(&self) -> Result<C::Output, MpcContractHandleError<C::Error>> {
+        self.caller
+            .call_contract(
+                &self.contract_id,
+                FunctionCallArgs {
+                    method_name: VERIFY_TEE.to_string(),
+                    args: b"{}".to_vec(),
+                    gas: MAX_GAS,
+                    deposit: NearToken::from_yoctonear(0),
+                },
+            )
+            .await
+            .map_err(MpcContractHandleError::Call)
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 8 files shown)


crates/tee-context/src/lib.rs
```diff
@@ -6,14 +6,13 @@ pub use types::{AllowedTeeHashes, TeeNodeIdentity};
 
 use chain_gateway::{
     state_viewer::{SubscribeToContractMethod, WatchContractState},
-    transaction_sender::{SubmitFunctionCall, TransactionSigner},
+    transaction_sender::{AccountCaller, SubmitFunctionCall, TransactionSigner},
 };
```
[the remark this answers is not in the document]
kevindeforth: The import is used via method resolution, not by name

crates/near-mpc-contract-interface/src/client.rs
```diff
@@ -0,0 +1,158 @@
...
+use crate::call_args::SubmitParticipantInfoArgs;
+use crate::method_names::{SUBMIT_PARTICIPANT_INFO, VERIFY_TEE};
+use crate::types::{AccountId, Attestation, Ed25519PublicKey};
+
+/// Gas attached to every handle-issued call.
+// TODO(#166): 300 Tgas is the protocol maximum and higher than most methods
+// need; benchmark per method and reduce.
+pub const MAX_GAS: NearGas = NearGas::from_gas(300_000_000_000_000);
+
+pub const SUBMIT_PARTICIPANT_INFO_DEPOSIT_MILLINEAR: u128 = 100;
```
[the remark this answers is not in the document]
kevindeforth: The guide (…:2073: "the node attaches 0, so call submit_participant_info manually with --deposit once. Exact amount tracked in #903") is still accurate after this PR: the node's live submission path   (ChainSendTransactionRequest::SubmitParticipantInfo → tx_signer, deposit 0) is untouched. The 0.1 NEAR
  deposit lives in MpcContractHandle, whose only consumer is TeeContext — which nothing in the node constructs
  yet. The doc goes stale at the future node↔tee-context integration PR, and that PR owns the guide update.

kevindeforth:
> crates/near-mpc-contract-interface/src/client.rs:54 — Attaching a fixed SUBMIT_PARTICIPANT_INFO_DEPOSIT_MILLINEAR = 100 (0.1 NEAR) to every submit_participant_info call is a semantic change over the previous NearToken::from_yoctonear(0) and creates a fund-leak path. In crates/contract/src/lib.rs:824-850, the refund block only runs when attestation_storage_must_be_paid_by_caller (new attestation OR caller is not a participant). For an existing participant re-submitting the same/shrinking attestation neither the charge nor the refund path is taken, so the attached 0.1 NEAR is retained by the contract on every re-submission. Once TeeContext is wired into the node's re-submission loop, this will slowly drain each participant's balance. Options: (a) fix the contract to refund the full deposit whenever the storage check is skipped, (b) parameterize the deposit per call-site so re-submissions send zero, or (c) require callers to pass the deposit and default the constant to zero. This is worth resolving (or at least tracked with an issue and TODO on the constant) before the handle becomes the shared source of truth.

Not a blocker for this PR, this is referring to other, known & existing issues in the codebase, some of which will be addressed in the verifier-contract PRs by @pbeza

netrome (approved):
Nice stuff!

haiyuechen-nearone (approved):
:rocket:

crates/tee-context/src/lib.rs
```diff
@@ -341,12 +328,27 @@ mod tests {
...
+    macro_rules! assert_call_error {
+        ($result:expr, $pattern:pat) => {
+            assert_matches!(
+                $result,
+                Err(TeeContextError::ContractCall(MpcContractHandleError::Call(
+                    $pattern
+                )))
+            )
+        };
+    }
```
gilcu3, line 350: oh we are in macro land now :D

crates/near-mpc-contract-interface/src/client.rs
```diff
@@ -0,0 +1,158 @@
...
+//! backend implementing [`CallContract`].
+
+use near_contract_transport::{CallContract, FunctionCallArgs, NearGas, NearToken};
+
+use crate::call_args::SubmitParticipantInfoArgs;
+use crate::method_names::{SUBMIT_PARTICIPANT_INFO, VERIFY_TEE};
+use crate::types::{AccountId, Attestation, Ed25519PublicKey};
+
+/// Gas attached to every handle-issued call.
+// TODO(#166): 300 Tgas is the protocol maximum and higher than most methods
```
gilcu3, line 14: we could already correct this, the maximum is 1000 Tgas now

crates/near-mpc-contract-interface/src/client.rs
```diff
@@ -0,0 +1,158 @@
...
+
+        // Then
+        let calls = caller.calls.lock().unwrap();
+        assert_eq!(calls.len(), 2);
+        let catalog = calls
+            .iter()
+            .map(|(contract_id, call)| render(contract_id, call))
+            .collect::<Vec<_>>()
+            .join("\n\n");
+        insta::assert_snapshot!(catalog);
```
gilcu3, line 156: what exactly are we capturing here? Using `insta` just to save some code?

gilcu3 (approved):
Thank you!

crates/near-mpc-contract-interface/src/client.rs
kevindeforth: yes, and it will make reviewing changes much easier.

crates/near-mpc-contract-interface/src/client.rs
kevindeforth: We don't need more than 300 Tgas for now, so I think it's fine to keep this. Issue #166 is about reducing the gas, not increasing it.

gilcu3: yeah I was only speaking about the comment that is no longer correct, leaving 300 in the actual constant is fine

PR #3876, opened 2026-07-21
feat: allow smaller than 32 bytes messages for edDSA

### LINKED ISSUE #3863: Remove lower message size limit for EdDSA signatures
### Background

Currently we have both an upper and a lower limit for the messages signed with EdDSA. They were introduced in #311 and while the upper bound stems from the maximum transaction size of a Solana transaction (see https://github.com/near/mpc/pull/2202#discussion_r2840587045) the lower bound has never been explained or documented. Additionally, protecting against arbitrarily large payloads feels reasonable and defensive, but there's no good reason (that I know of at the time of writing) to impose a lower size limit here.

Additionally, now we have builders who want to be permissive in terms of payload sizes so we should get rid of this bound asap.

### User Story

As an MPC user I'd like to be able to sign messages shorter than 32 bytes with EdDSA.

### Acceptance Criteria

The contract and node have no lower bound on the size of EdDSA payloads.

_No response_

gilcu3:
Closes #3863


crates/devnet/src/contracts.rs
```diff
@@ -3,13 +3,12 @@ use std::collections::BTreeMap;
 use blstrs::G1Projective;
 use group::Group;
 use near_account_id::AccountId;
-use near_mpc_bounded_collections::BoundedVec;
+use near_mpc_bounded_collections::EmptyBoundedVec;
```
netrome: I know this alias pre-dates this PR `EmptyBoundedVec` is a misleading name. This vec doesn't need to be empty and in most cases it's not. This should be renamed to `UpperBoundedVec` or something similar imo. It's a vec with an upper bound, but not a lower bound. It's not an empty vec.

crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
```diff
@@ -4100,7 +4100,7 @@ expression: abi
               ],
               "properties": {
                 "Eddsa": {
-                  "$ref": "#/definitions/HexString_Min32_Max1232"
+                  "$ref": "#/definitions/HexString_Min0_Max1232"
```
netrome, line 4103: Iiuc, this tag change has no impact on the bytes sent over the wire so this should be fine.

crates/near-mpc-bounded-collections/src/bounded_vec.rs
```diff
@@ -1137,6 +1140,32 @@ mod hex_serde_tests {
         data: BoundedVec<u8, 2, 4>,
     }
 
+    #[derive(serde::Serialize, serde::Deserialize, Debug, PartialEq)]
+    struct EmptyWrapper {
+        #[serde(with = "hex_serde")]
+        data: EmptyBoundedVec<u8, 4>,
+    }
```
netrome, line 1177: What's the intention of this wrapper? I see it's used in the serde roundtrip test. Perhaps we should rename this struct to convey the intention? Something like `HexSerializationTestData`? wdyt?

crates/contract/tests/sandbox/utils/sign_utils.rs
```diff
@@ -565,8 +564,7 @@ pub fn create_response_ed25519(
         .try_into()
         .unwrap();
 
-    let bytes = BoundedVec::from(payload);
-    let payload = Payload::Eddsa(bytes);
```
netrome, line 569: It's a bit of a shame we're loosing the From implementation, since the current one is only defined at the bounds. We should be able to generalize it to cover all arrays. I got this suggestion from Claude, but it only seems to concern the lower bound:

```rust
impl<T, const N: usize, const U: usize> From<[T; N]>
    for BoundedVec<T, 0, U, witnesses::PossiblyEmpty<U>>
{
    fn from(arr: [T; N]) -> Self {
        const { assert!(N <= U, "array length exceeds upper bound") };
        BoundedVec {
            inner: arr.into(),
            witness: witnesses::possibly_empty::<U>(),
        }
    }
}
```

However, this is obviously quite complex so worth tackling in a separate follow-up imo.

netrome (dismissed):
Thanks for fixing this. No blockers from my side.

crates/contract/tests/sandbox/utils/sign_utils.rs
gilcu3: I will give it a try here

crates/near-mpc-bounded-collections/src/bounded_vec.rs
```diff
@@ -482,6 +482,24 @@ impl<T, const L: usize, const U: usize> From<[T; L]>
...
+/// oversized arrays are rejected at compile time.
+impl<T, const N: usize, const U: usize> From<[T; N]>
+    for BoundedVec<T, 0, U, witnesses::PossiblyEmpty<U>>
+{
+    fn from(arr: [T; N]) -> Self {
+        const {
+            if N > U {
+                panic!("array length exceeds upper bound")
+            }
+        }
```
netrome, line 495: It feels like we should check the lower bound here as well, now we just fixed this for the upper bounded vec effectively. Or what do you think?

netrome (approved):
Thanks for the updates 🙏

crates/near-mpc-bounded-collections/src/bounded_vec.rs
gilcu3: which lower bound do you mean? For this conversion the lower bound is just 0

SimonRastikian (commented):
I wonder whether we should reject empty strings and accept starting from 1 byte. I am unsure about the behavior of inputting empty vector

anodar (approved):

gilcu3:
> I wonder whether we should reject empty strings and accept starting from 1 byte. I am unsure about the behavior of inputting empty vector

As we discussed on slack having signatures over empty strings is fine. We are also discussing other stuff, like other upper bounds, but that's out of scope here so I am merging this now

SimonRastikian (commented):
Approved -- late

PR #3879, opened 2026-07-21
ci: make workflow runs cancellable automatically

### LINKED ISSUE #3878: New pushes to a PR queue behind the old CI run instead of cancelling it
Pushing a new commit to a PR branch used to cancel the previous in-progress CI run so the new commit started right away. Now the new run sits in `queued` and does not start until the old run finishes — the two are serialized, roughly doubling time-to-signal on a PR unless the previous run is cancelled by hand.

### Background

The cause is the `if: always()` gate on most jobs in `ci.yml` (e.g. `if: ${{ always() && needs.changes.outputs.code_changed != 'false' }}`). GitHub evaluates `always()` as true even while a run is being cancelled by `cancel-in-progress`, so those jobs keep running to completion instead of stopping. They hold the `CI-${{ github.ref }}` concurrency slot until they finish, so the superseding run stays queued behind them.

Observed on a recent PR branch: an older run superseded at 12:33:38 kept its `always()` jobs running (E2E, TEE build, nix build) until 12:34:43; the newer run, created at 12:33:38, did not start its first job until 12:35:22 — only after the old run drained. Jobs without `always()` (e.g. "Fast CI checks") were cancelled immediately, confirming the gate as the differentiator.

Replacing `always()` with `!cancelled()` in these gates fixes it: `!cancelled()` still overrides the implicit `success()` on `needs` (so a failed or skipped `changes` detector does not silently skip its dependents — the reason the gate exists, per the comment at `ci.yml:22-26`), but it stops evaluating to true once the run is cancelled, letting the jobs stop and free the concurrency slot.

### Acceptance Criteria

- Pushing a new commit to a PR promptly stops the previous in-progress CI run and starts the new one, without waiting for the old run's jobs to finish and without manual cancellation.
- A failed or skipped `changes` detector still runs its dependent jobs (no silent skip that could satisfy required checks with zero CI signal).

gilcu3:
Closes #3878 

CI getting cancelled just fine again!

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,11 +19,13 @@ on:
     - cron: "5 0 * * 0"
 
 jobs:
-  # Downstream gates use `if: ${{ always() && needs.changes.outputs.code_changed != 'false' }}`.
-  # The `always()` overrides GitHub's implicit `success()` on `needs`, so a
+  # Downstream gates use `if: ${{ !cancelled() && needs.changes.outputs.code_changed != 'false' }}`.
+  # `!cancelled()` overrides GitHub's implicit `success()` on `needs`, so a
   # failed detector doesn't silently skip dependents (skipped checks satisfy
   # required-check rules and could let a PR merge with zero CI signal). Only
-  # an explicit `'false'` (docs-only PR) skips the job.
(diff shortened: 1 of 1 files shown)


netrome (approved):

anodar (approved):
Nice, thanks!

PR #3885, opened 2026-07-21
fix(tee-verifier): inherit repository for reproducible builds

### LINKED ISSUE #3884: tee-verifier reproducible build fails: missing [package.repository]
## Background

`cargo near build reproducible-wasm --manifest-path crates/tee-verifier/Cargo.toml` fails with `Malformed NEP330 metadata in Cargo.toml: [package.repository]: should not be empty`.

Why the line is needed: the reproducible (docker) build embeds NEP-330 provenance into the WASM, and `cargo-near` validates that `[package.repository]` is set and non-empty (enforced in `cargo-near-build`'s `docker_build/metadata/validate.rs::validate_repository`). The `tee-verifier` crate inherits `version`/`license`/`edition` from the workspace but not `repository`, so that field is empty and the build is rejected. Adding `repository = { workspace = true }` inherits the workspace value (`https://github.com/near/mpc`), which is how `mpc-contract` already satisfies the same check.

Without it, no reproducible (operator-auditable) verifier WASM can be produced.

## Acceptance Criteria

- [ ] `cargo near build reproducible-wasm --manifest-path crates/tee-verifier/Cargo.toml` succeeds

Required only for the reproducible (docker) build; `non-reproducible-wasm` and normal `cargo build` do not consult `[package.repository]` and are unaffected.

pbeza:
Fixes #3884.

`cargo near build reproducible-wasm` embeds NEP-330 provenance and requires a non-empty `[package.repository]`.

### DIFF
--- a/crates/tee-verifier/Cargo.toml
+++ b/crates/tee-verifier/Cargo.toml
@@ -3,6 +3,7 @@ name = "tee-verifier"
 version = { workspace = true }
 license = { workspace = true }
 edition = { workspace = true }
+repository = { workspace = true }
 
 [package.metadata.cargo-shear]
 ignored = ["borsh"]


gilcu3 (approved):
:)

barakeinav1 (approved):
thank you!

PR #3888, opened 2026-07-21
test(localnet): deploy + vote the tee-verifier in the cluster script

### LINKED ISSUE #3887: update Localnet TEE cluster scripts to cover the tee-verifier contract
## Background

DCAP attestation verification is moving out of `mpc-contract` into a standalone `tee-verifier` contract, with `submit_participant_info` becoming asynchronous (cross-contract `verify_quote` + yield/resume) — see #3642, #3714 (core), #3715 (tests).

The localnet TDX deploy tooling (`localnet/tee/scripts/rust-launcher/deploy-tee-cluster.sh`) has no step to deploy the `tee-verifier` contract or vote it in, so the verifier path can't be exercised on a real-TDX localnet cluster. Running the async path against real quotes on such a cluster also surfaced that the contract's `DEFAULT_VERIFIER_TERA_GAS = 100` is too low: real `dcap_qvl::verify` on a valid quote burns **~173 Tgas**, so the cross-call OOGs (`Exceeded the prepaid gas`) and attestation never completes. The sandbox tests don't catch this because they only feed a malformed quote that's rejected at parse (≈no gas).

## User Story

As an MPC developer, I want the localnet TEE deploy to deploy + vote in the `tee-verifier` and let me set the `verify_quote` gas budget, so I can validate the full async attestation → verifier → keygen → sign path on real hardware before it ships.

## Acceptance Criteria

- [ ] `deploy-tee-cluster.sh` builds + deploys `tee-verifier.<root>` and votes it in from threshold participants (`vote_tee_verifier_change`).
- [ ] Optional `verifier_tera_gas` override at init (opt-in; harmless on `main`, needed once #3714 lands).
- [ ] A how-to doc for the verifier e2e (deploy → vote → attest → keygen → sign), including the gas note.
- [ ] (follow-up) A sandbox/e2e test that runs `verify_quote` against the real verifier at production gas with a *valid* quote and asserts a stored Dstack attestation — so the gas budget is covered, not just the happy path (see #3715).

## Resources

- #3642 (tracking), #3714 (async core), #3715 (tests)
- e2e finding: real `verify_quote` ≈173 Tgas vs the 100 default; `verifier_tera_gas=200` fixes it (verify + resolve[60] + callback[10] fits under the 300 Tgas tx cap). Verified end-to-end on a localnet TDX cluster: both nodes attested as **Dstack**, contract `Running`, ECDSA `sign` succeeded.

barakeinav1:
Closes #3887 (script + doc portion).

Adds TEE verifier support to the localnet TDX deploy so the verifier path can be exercised on a real cluster (the account voting `vote_tee_verifier_change` is already on `main`; the async `verify_quote` wiring lands with #3714):

- **`deploy_verifier`** (in the `near_contract` phase): builds `crates/tee-verifier` (or reuses `TEE_VERIFIER_PATH`), creates `tee-verifier.<root>`, deploys the wasm (stateless, no init).
- **`vote_tee_verifier_threshold`** (right after `init`, while `Initializing`): each node account votes `vote_tee_verifier_change` committing to the wasm sha256, crossing threshold so the verifier is trusted before keygen finishes.
- **`VERIFIER_TERA_GAS_OVERRIDE`** (opt-in): sets `init_config.verifier_tera_gas`. Off by default so this stays valid on `main` (no such `InitConfig` field there yet); needed once #3714 lands, since real `dcap_qvl::verify` burns ~173 Tgas and the contract's `100` default OOGs the cross-call.
- A how-to doc: `localnet/tee/scripts/rust-launcher/how-to-test-tee-verifier-e2e.md`.

Verified end-to-end on a localnet TDX cluster (stock `release-v3.13` node + a #3714 contract): both nodes attested as **Dstack** (`get_attestation` → `Dstack`), contract reached `Running`, and an ECDSA `sign` succeeded. Full context in #3887.

Follow-up (tracked in #3887): a sandbox/e2e test that runs `verify_quote` at production gas against a valid quote (see #3715).

### DIFF
--- a/localnet/tee/scripts/rust-launcher/deploy-tee-cluster.sh
+++ b/localnet/tee/scripts/rust-launcher/deploy-tee-cluster.sh
@@ -223,6 +223,13 @@ ROOT_ACCOUNT="${MPC_NETWORK_NAME}${ACCOUNT_SUFFIX}"
 MPC_CONTRACT_ACCOUNT="${MPC_CONTRACT_ACCOUNT:-mpc.${ROOT_ACCOUNT}}"
 node_account_for_i() { echo "node$1.${ROOT_ACCOUNT}"; }
 
+# TEE verifier contract account (DCAP verification, near/mpc#3642). Deployed and
+# voted in as the trusted verifier. Set TEE_VERIFIER_PATH to a prebuilt wasm to
+# skip the in-script build. Optional: VERIFIER_TERA_GAS_OVERRIDE sets
+# init_config.verifier_tera_gas (only meaningful once the async-verify PR #3714
+# lands; leave unset on main).
+TEE_VERIFIER_ACCOUNT="${TEE_VERIFIER_ACCOUNT:-tee-verifier.${ROOT_ACCOUNT}}"
+
 # (Resolved-naming echo lives further down, after ip_for_i() is defined.)
 
 NODE_RANGE_START="${NODE_RANGE_START:-0}"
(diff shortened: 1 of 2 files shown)


barakeinav1:
Superseded by the move to mpc-private instead of adding more script logic to the public repo. The verifier deploy+vote steps + opt-in gas override are now in near/mpc-private#495 (tracked by #3894); this PR's mpc-side removal is #3895. Closing.

PR #3902, opened 2026-07-22
test(contract): implement CallContract for sandbox tests

### LINKED ISSUE #3901: Implement `CallContract` for sandbox back-end
### Background

We want to migrate the sandobx contract tests to the new MpcContractHandle interface. For this, we need to implement the CallContract trait for the sandbox backend.

### User Story

As a dev, I like to reduce code duplication

### Acceptance Criteria

Implement `CallContract` for the sandbox backend

part of https://github.com/near/mpc/issues/3693

kevindeforth:
resolves #3901, part of https://github.com/near/mpc/issues/3693

### DIFF
--- a/crates/contract/tests/sandbox/utils/mpc_contract.rs
+++ b/crates/contract/tests/sandbox/utils/mpc_contract.rs
@@ -59,10 +59,10 @@ pub async fn submit_participant_info(
 ) -> anyhow::Result<ExecutionFinalResult> {
     // TODO(#3906): check if inlining is nicer once we ported the entire contract interface.
     let contract_handle = MpcContractHandle::new(SandboxCaller(account), contract.id().clone());
-    let result = contract_handle
+    contract_handle
         .submit_participant_info(attestation.clone(), tls_key.clone())
-        .await?;
-    Ok(result)
+        .await
+        .map_err(Into::into)
 }
 
 pub async fn vote_tee_verifier_change(
--- a/crates/contract/tests/sandbox/utils/mpc_contract.rs
+++ b/crates/contract/tests/sandbox/utils/mpc_contract.rs
@@ -1,10 +1,10 @@
 use std::collections::BTreeSet;
 
-use super::consts::SUBMIT_PARTICIPANT_INFO_DEPOSIT;
-use super::transactions::all_receipts_successful;
+use super::transactions::{SandboxCaller, all_receipts_successful};
 use mpc_contract::tee::tee_state::NodeId;
 use mpc_primitives::hash::{LauncherImageHash, NodeImageHash, TeeVerifierCodeHash};
 use near_mpc_contract_interface::{
+    client::MpcContractHandle,
     method_names,
     types::{Attestation, Ed25519PublicKey, Participants, ProtocolContractState, Threshold},
 };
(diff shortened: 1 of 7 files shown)


kevindeforth:
> - `crates/contract/tests/sandbox/utils/mpc_contract.rs:54-65` — Two now-divergent serialization paths for the same on-chain method live side by side in this file: the refactored `submit_participant_info` serializes `SubmitParticipantInfoArgs` (JSON object with named fields), while `submit_participant_info_with_deposit` (line 74-80) keeps `args_json((attestation, tls_key))` (JSON array of positional args). Both currently work because near-sdk accepts either, but the divergence invites confusion the next time someone touches these helpers. Consider extending `MpcContractHandle::submit_participant_info` to accept an override deposit (or adding a sibling method), then routing `_with_deposit` through the client too and deleting the last hand-rolled call — that finishes the migration this PR starts and keeps the wire format in one place.

good catch, addressed in https://github.com/near/mpc/pull/3902/commits/96e105923c0ee425b954af89404573afb391eede


> - `crates/contract/tests/sandbox/utils/mpc_contract.rs:60-62` — Three fresh clones (`contract.id().clone()`, `attestation.clone()`, `tls_key.clone()`) where the previous implementation was borrow-only. Fine for test code, but `MpcContractHandle::submit_participant_info` could take `&Attestation`/`&Ed25519PublicKey` (they only need to be serialized) and `MpcContractHandle::new` could accept `&AccountId`; that's a client-side change and out of scope here, worth a follow-up on the client crate.

this is the second or third time claude brings this up in a review, it's wrong, the clone happens just later, due to our DTO.

> - `crates/contract/tests/sandbox/utils/mpc_contract.rs:64` — `Ok(result)` on the previous line is equivalent to `contract_handle.submit_participant_info(...).await.map_err(Into::into)`; the `let result = ...?; Ok(result)` shape is a tiny bit noisier than needed. Purely stylistic.

addressed in https://github.com/near/mpc/pull/3902/commits/211049450ccccbe3d6409fcd1d1074d5493a5bef

> - `crates/contract/tests/sandbox/utils/transactions.rs:6-7` — Doc comment leans toward paraphrasing the impl ("transacts as the wrapped `Account` via near-workspaces") which is already visible from the field type and the `impl` block. Per engineering-standards.md §Write helpful code comments the *why* is the useful part — something like "Adapter used only by sandbox integration tests" is enough, or drop the comment. Not blocking.

addressed in https://github.com/near/mpc/pull/3902/commits/211049450ccccbe3d6409fcd1d1074d5493a5bef

crates/contract/tests/sandbox/utils/mpc_contract.rs
```diff
@@ -57,30 +57,12 @@ pub async fn submit_participant_info(
...
-    deposit: NearToken,
-) -> anyhow::Result<ExecutionFinalResult> {
-    Ok(account
-        .call(contract.id(), method_names::SUBMIT_PARTICIPANT_INFO)
-        .args_json((attestation, tls_key))
-        .deposit(deposit)
-        .max_gas()
-        .transact()
-        .await?)
+    // TODO(#3906): check if inlining is nicer once we ported the entire contract interface.
```
kevindeforth, line 60: deferring this to later, as at this point, it wouldn't be an improvement.

crates/contract/tests/sandbox/utils/transactions.rs
```diff
@@ -1,7 +1,30 @@
+use near_contract_transport::{CallContract, FunctionCallArgs};
 use near_sdk::Gas;
-use near_workspaces::{Account, Contract, result::ExecutionFinalResult};
+use near_workspaces::{Account, AccountId, Contract, result::ExecutionFinalResult};
 use serde::Serialize;
 
+/// [`CallContract`] adapter used only by the sandbox integration tests.
+pub struct SandboxCaller<'a>(pub &'a Account);
```
SimonRastikian, line 7: Why is this struct needed? Can we not skip it?

SimonRastikian (approved):
Nice!

crates/contract/tests/sandbox/utils/transactions.rs
kevindeforth: Rust's orphan rule says you can implement a trait for a type only if at least one of the two (the trait or the type) is defined in the current crate. Here, `Account` comes from near-workspaces and `CallContract` comes from the contract-transport crate, so we need to introduce a helper struct over which we can then define `SandboxCaller`.

pbeza (approved):

PR #3912, opened 2026-07-22
feat(node): gate verify-foreign-tx on available chains via the watch channel

### LINKED ISSUE #3910: Gate verify-foreign-tx on available chains via the watch channel, consume available chains from channel
### Background

https://github.com/near/mpc/issues/3852 implements watch channel that writes available chains in a channel.

The node still gates verify-foreign-tx requests through the legacy read path,  on every request it issues a `get_supported_foreign_chains` view call against the contract, which implements the old intersection model.

https://github.com/near/mpc/issues/3852 implemented producer side for that channel but nothing consumes it yet. We should read that channel and use that for decision making instead of issuing view method on every request.

### User Story

_No response_

### Acceptance Criteria

The verify-foreign-tx provider checks chain availability against the channel snapshot on both the leader and follower paths; no per-request contract view call remains.
- The legacy read path (`ReadSupportedForeignChain`, the `get_supported_foreign_chains` view call, and the `ForeignChainPolicyReader` type parameter on `IndexerAPI`) is deleted from the node.

_No response_

anodar:
Closes https://github.com/near/mpc/issues/3910

Followup on https://github.com/near/mpc/issues/3852

NOTE: this should not be merged until predecessor is not released and deployed (one that registers config to the new endpoint), currently available chains is an empty list.


anodar:
@claude review

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -373,42 +382,48 @@ where
 }
 
 #[derive(Debug, thiserror::Error)]
-enum ForeignChainSupportError {
-    #[error("failed to fetch supported chains on the contract")]
-    FailedToReadContract(#[source] anyhow::Error),
+enum ChainAvailabilityError {
+    #[error("the foreign-chain supporters snapshot has not been received from the contract yet")]
+    SupportersSnapshotNotReady,
```
gilcu3: This name does not seem natural to me. Why is this a snapshot? Is this error reachable in our current architecture? As I understand it it could happen when the node has started and we have not yet read this info from the contract. But I don't really see why instead of having this error we cannot just wait to have this (as we wait to have the contract state) before starting the node. It seems impossible to have one and not the other, as both come from the same place (even if we use different watch channels for them)

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -48,16 +48,22 @@ fn build_signature_request(
...
         let foreign_tx_request = self.verify_foreign_tx_request_store.get(id).await?;
 
+        // Also checked in `execute_foreign_chain_request`; checked early here
+        // because `take_owned` below irreversibly consumes a presignature. An
+        // availability flip between the two checks still costs one presignature.
+        ensure_chain_is_available(
+            &self.supporters_by_foreign_chain.borrow(),
+            &foreign_tx_request.request,
+        )
+        .inspect_err(|_| metrics::MPC_NUM_VERIFY_FOREIGN_TX_UNAVAILABLE_CHAIN_REJECTIONS.inc())?;
```
gilcu3, line 65: I think I never considered this case before, in which something is supported and the not supported. It opens the possibility for race conditions. We probably need to have some robust way of handling this.

crates/node/src/indexer.rs
```diff
@@ -620,15 +583,11 @@ pub struct IndexerAPI<TransactionSender, ForeignChainPolicyReader> {
...
-    /// registered supporters (by TLS key).
-    #[expect(
-        dead_code,
-        reason = "consumed once the verify-foreign-tx provider switches to the new API"
-    )]
-    pub foreign_chain_supporters_receiver: watch::Receiver<foreign_chain::ForeignChainSupporters>,
+    /// registered supporters (by TLS key). Holds `None` until the first
+    /// successful read after the indexer syncs.
+    pub foreign_chain_supporters_receiver:
+        watch::Receiver<Option<foreign_chain::ForeignChainSupporters>>,
```
gilcu3: Ahh this is the origin of the `None` that to me we could simply avoid from the start.

crates/node/src/metrics.rs
```diff
@@ -151,6 +151,18 @@ pub static MPC_NUM_VERIFY_FOREIGN_TX_REQUESTS_INDEXED: LazyLock<prometheus::IntC
...
+    prometheus::IntCounter,
+> = LazyLock::new(|| {
+    prometheus::register_int_counter!(
+        "mpc_num_verify_foreign_tx_unavailable_chain_rejections",
+        "Number of gate rejections of verify foreign tx attempts, at most one per node per \
+         attempt: the requested chain is not available or the supporters snapshot has not \
+         been received yet"
+    )
+    .unwrap()
+});
```
gilcu3, line 163: this is conflating two very different problems. We should probably have 2 metrics (although I think the second one could be avoided by an invariant)

crates/node/src/coordinator.rs
gilcu3, line 119: I think that if we can guarantee that at this point, the `self.indexer.foreign_chain_supporters_receiver` already returns a value, then the `Option` in that receiver would no longer be necessary. But it might be that this happens higher in the call stack

crates/node/src/tests/verify_foreign_tx_gating.rs
```diff
@@ -0,0 +1,207 @@
...
+            .as_ref()
+            .is_some_and(|map| map.is_empty())
+        {
+            supporters.changed().await.unwrap();
+        }
+    })
+    .await
+    .expect("timed out waiting for the empty supporters snapshot to publish");
+    tokio::time::sleep(Duration::from_secs(1)).await;
+
```
gilcu3, line 186: hmm this is something that we should not have in tests. We tried hard to remove this kind of code in the past, and apparently it is also caused by the use of the `Option` in the node

crates/node/src/tests/verify_foreign_tx_gating.rs
gilcu3, line 1: I am a bit surprised seeing this totally new test here. Is it covering something we were not covering before? As this feature is just substituting an existing one, I expected a similar test to be erased.

crates/node/src/foreign_chain_policy.rs
```diff
@@ -0,0 +1,414 @@
...
+        }
+    });
+    (receiver, task)
+}
+
+async fn await_updated_supporters(
+    upstream: &mut watch::Receiver<Option<ForeignChainSupporters>>,
+    participants_config: &ParticipantsConfig,
+    foreign_tx_reconstruction_threshold: Option<u64>,
+) -> anyhow::Result<Option<SupportersByForeignChain>> {
```
gilcu3: it seems we are using this `Option` to encode that there is no Domain for ForeignChainTx. I think there should be a more explicit way of doing so. For example, if there is no Domain, why do we need this task to be run?

crates/node/src/foreign_chain_policy.rs
```diff
@@ -0,0 +1,414 @@
...
+
+/// Mirrors the contract's availability rule: the max reconstruction threshold
+/// across ForeignTx domains, `None` when no such domain exists.
+pub(crate) fn foreign_tx_reconstruction_threshold(domains: &[dtos::DomainConfig]) -> Option<u64> {
+    domains
+        .iter()
+        .filter(|domain| domain.purpose == dtos::DomainPurpose::ForeignTx)
+        .map(|domain| domain.reconstruction_threshold.inner())
+        .max()
+}
```
gilcu3, line 94: This took me some time to understand. Afaik the simplifying assumption we have now is that we will have a single Foreign Chain TX domain, therefore computing the max (or min) is fine. Do we have issues for this, or any way to keep that invariant in code? (I might be suffering from lack of memory here, so forgive me if we discussed it already)

gilcu3 (commented):
Left some comments. Not requesting changes because we probably need some discussion around this, and anyway we are not merging this soon.

My main concern is the use of `Option` in two places.
The first one if for the supported chains. As far I can see, this is a case that cannot happen in practice, once the node is parked on startup until it gets the first set, which it should anyway have as it already needs the contract state to do anything, and these two things are tied together (both as view methods in the contract). I drafted this alternative in #3966 and it seems to work. Happy to hear if there is a downside.

The other place where I think `Option` might be avoided is in the thresholds in `crates/node/src/foreign_chain_policy.rs`. The None branch seems to refer to the case where there are no foreign chain domains, but I think we should handle that explicitly, for example by not running the tasks in that case (or some more explicit method).

crates/node/src/foreign_chain_policy.rs
anodar: Good point. We [decided]( https://nearone.slack.com/archives/C0912BTG51T/p1781466690642139?thread_ts=1781047607.002969&cid=C0912BTG51T) we'll use max for now till we have single domain and revisit it later. So added todo: [37dd1c0](https://github.com/near/mpc/pull/3912/commits/37dd1c01de433ac2547a73462d911f9894d2ff2b)

crates/node/src/foreign_chain_policy.rs
anodar: Got rid of option in [a87bc1f](https://github.com/near/mpc/pull/3912/commits/a87bc1f692c58f69b87ae65aa3e7a8a79058f8b6)

crates/node/src/tests/verify_foreign_tx_gating.rs
anodar: We didn't integration test for what this covers. We had test for `chain_is_supported` but no integration test for chain availability. 

We do have e2e tests that remain the same (merged in parent pr).

crates/node/src/tests/verify_foreign_tx_gating.rs
anodar: Got rid of it in [a87bc1f](https://github.com/near/mpc/pull/3912/commits/a87bc1f692c58f69b87ae65aa3e7a8a79058f8b6)

crates/node/src/metrics.rs
anodar: No longer applicable since we now wait for the snapshot.

crates/node/src/indexer.rs
anodar: Got rid of option in [a87bc1f](https://github.com/near/mpc/pull/3912/commits/a87bc1f692c58f69b87ae65aa3e7a8a79058f8b6)

crates/node/src/providers/verify_foreign_tx/sign.rs
anodar: Got rid of this error (along with option) in [a87bc1f](https://github.com/near/mpc/pull/3912/commits/a87bc1f692c58f69b87ae65aa3e7a8a79058f8b6) and waiting instead as you suggested.

crates/node/src/coordinator.rs
anodar: Got rid of option in [a87bc1f](https://github.com/near/mpc/pull/3912/commits/a87bc1f692c58f69b87ae65aa3e7a8a79058f8b6)

crates/node/src/providers/verify_foreign_tx/sign.rs
anodar: We had similar discussion in https://github.com/near/mpc/pull/3853#discussion_r3628628211. 

Worst case it either times out (if node dropped support for chain that we thought it supported) or we think it's not supported (if node enables support at the "right" time) and fail early, in which case retry will succeed.

We just also check availability (`ensure_chain_is_available`) here as a way to reduce that race condition window (although not by much), but not trying to eliminate it.  Since all nodes are expected to support all the chains, this is similar to node going down/up race condition.

anodar:
@claude review, check that @gilcu3's comments were addressed.

crates/node/src/indexer/foreign_chain.rs
```diff
@@ -12,34 +12,46 @@ const FOREIGN_CHAIN_SUPPORTERS_REFRESH_INTERVAL: Duration = Duration::from_secs(
...
-                        *previous = supporters;
-                        true
-                    }
-                });
-            }
+            Ok(supporters) => break supporters,
             Err(e) => {
-                tracing::error!(target: "mpc", "error reading foreign-chain supporters from chain: {:?}", e)
+                tracing::error!(target: "mpc", "error reading foreign-chain supporters from chain: {:?}", e);
+                tokio::time::sleep(FOREIGN_CHAIN_SUPPORTERS_REFRESH_INTERVAL).await;
```
netrome, line 28: I know this predates this PR, but shouldn't we do exponential backoff here? It makes logs cleaner if the node gets stuck on this, and also avoids spamming if we'd get rate limited anywhere.

crates/node/src/indexer/foreign_chain.rs
```diff
@@ -12,34 +12,46 @@ const FOREIGN_CHAIN_SUPPORTERS_REFRESH_INTERVAL: Duration = Duration::from_secs(
...
             }
         }
-        tokio::time::sleep(FOREIGN_CHAIN_SUPPORTERS_REFRESH_INTERVAL).await;
-    }
+    };
+
+    let (sender, receiver) = watch::channel(initial);
+    tokio::spawn(async move {
+        loop {
+            tokio::time::sleep(FOREIGN_CHAIN_SUPPORTERS_REFRESH_INTERVAL).await;
```
netrome, line 36: Every time we add one of these direct sleeps without properly injecting time, a kitten dies.

Here it would over-complicate the code to try to shoehorn in a proper time abstraction so that's a separate effort. But anyway.

crates/node/src/providers/verify_foreign_tx/sign.rs
netrome: Speaking of this, it's unfortunate that there's no way to batch multiple view calls into a single atomic request. Feels like it would be a useful feature. I wonder how complex this would be in nearcore to support.

netrome (dismissed):
Somewhat shallow review, but overall this looks good to me. Thanks!

crates/node/src/indexer/foreign_chain.rs
anodar: I had this as config parameter (arguably could have been placed to better place) so that I could set different values in production vs testing code, but changed to hardcoded value as per reviewers suggestion.

Some context in: https://github.com/near/mpc/pull/3641#discussion_r3556544126, but more was discussed offline on this.

crates/node/src/indexer/foreign_chain.rs
anodar: I did have exponential backoff initially on that, but changed based on reviewers feedback.

https://github.com/near/mpc/pull/3853#discussion_r3623021101

netrome: Hmm I'm not following the reasoning in that thread.

> Here, the contract already has all the required API, so emitting an error seems correct - we don't expect this call to fail.

This reasoning seems to assume the contract lacking the API being the only reason for the call to fail. I see several levels of type-erased errors in the view client, so I wouldn't directly trust this to always hold. Are we sure about this? What about network errors? Or other issues in nearcore? How can we be so confident this won't fail? (I know Kevin raised the point, but I assume you had some understanding of this argument when you changed the code)

crates/node/src/indexer/foreign_chain.rs
netrome: I don't mind the config parameter, I'm just ranting (off topic) about the lack of dependency injection. Time is I/O and I/O should be separated from business logic. When you don't, any code that touches time interactions becomes incredibly hard to test.

For example, if I have:

```rust
fn delayed_add(delay: Duration, a: u64, b: u64) -> u64 {
    std::time::sleep(delay);
    a.checked_add(b).expect("overflows aren't the point of this example")
}
```

This is a very simple function, but how do we write unit tests for this? Yes, we can ignore the delay, but there's no way to control the delay or test it. In more complex systems this becomes a much more real problem since you can't control the timing between components which leads to huge gaps in what you can and cannot test, and buries interesting code behind race conditions and whatnot.

If I instead simply did:
```rust
fn delayed_add(sleep: impl Sleep,delay: Duration, a: u64, b: u64) -> u64 {
    sleep.sleep(delay);
    a.checked_add(b).expect("overflows aren't the point of this example")
}

trait Sleep {
    fn sleep(duration: Duration);
}

struct RealTime {}

impl Sleep for RealTime {
    fn sleep(duration: Duration){ std::time::sleep(duration) }
}
```

I could easily inject a fake sleep in tests that records how long the function actually intended to sleep. Additionally I can introduce wait point to fine grained control over which sleep points are passed and when. This makes time testable, and also make unit tests that don't depend on time instant. You should never have to wait for a timeout in tests.

crates/node/src/tests/verify_foreign_tx_gating.rs
```diff
@@ -0,0 +1,205 @@
...
+async fn verify_foreign_tx__should_only_be_served_while_chain_is_available() {
+    const NUM_PARTICIPANTS: usize = 4;
+    const THRESHOLD: usize = 3;
+    const TXN_DELAY_BLOCKS: u64 = 1;
+    const SUPPORTERS_PUBLISH_WAIT: Duration = Duration::from_secs(10);
+    const UNAVAILABLE_RESPONSE_WAIT: Duration = Duration::from_secs(15);
+
+    // Given: four nodes with a mocked Bitcoin RPC (so inspection would
+    // succeed) and a ForeignTx domain whose reconstruction threshold is met
+    // by all four auto-registrations.
```
gilcu3: nit: we are trying to get all "given,when,then" without further explanatory comments on them, as that should be readable from the code.

crates/node/src/foreign_chain_policy.rs
```diff
@@ -0,0 +1,353 @@
...
+/// Recomputed node-side even though the contract already threshold-gates
+/// availability: the upstream snapshot keys supporters by TLS key over all
+/// registrations (prospective or stale ones included) and may come from a
+/// different block than `participants_config`. Only supporters resolving to
+/// `participants_config` — the participants this node can sign with — count
+/// towards the quorum. An empty map means no chain reaches the quorum.
+fn resolve_supporters_by_foreign_chain(
+    supporters_by_tls_key: &ForeignChainSupporters,
+    participants_config: &ParticipantsConfig,
+    foreign_tx_reconstruction_threshold: u64,
```
gilcu3: nit: I am not sure if this is best, but probably we should use the correct type here `ReconstructionThreshold`

crates/node/src/foreign_chain_policy.rs
```diff
@@ -0,0 +1,353 @@
...
+) -> SupportersByForeignChain {
+    supporters_by_tls_key
+        .iter()
+        .filter_map(|(chain, tls_keys)| {
+            let ids = resolve_participant_ids(tls_keys, participants_config);
+            (ids.len() as u64 >= foreign_tx_reconstruction_threshold).then_some((*chain, ids))
```
gilcu3: nit: is this as conversion unavoidable? @netrome did you miss this one or left on purpose?

gilcu3 (dismissed):
Thank you for all the fixes!

crates/node/src/indexer/foreign_chain.rs
anodar: > How can we be so confident this won't fail?

We can't, and all of the cases you listed will start spamming with error logs. 

>  but I assume you had some understanding of this argument when you changed the code

Only argument against exponential backoff was to save us code complexity. While I don't necessarily agree on it, worst case this will spam 1 log per second so it's not the end of the world. If it keeps spamming long enough (can't read contract state for some reason) we likely have bigger issue (but again not a good reason to avoid backoff).

gilcu3: > worst case this will spam 1 log per second so it's not the end of the world.

I was not aware of this. As an operator, 1 log per second is not the end of the world, but not too far either as it basically makes the log unreadable after a while

crates/node/src/indexer/foreign_chain.rs
anodar: Fully agree. 

In other languages (e.g. Go), I'd stub out `foreign_chain_supporters_refresh_interval` as a function type variable (maybe even at the package level) and set to different functions depending on production vs testing (pretty close to what you're suggesting). Having time configurable was my way to achieve similar thing in Rust without changing function signature.


I'd rather not refactor this now though as this is existing pattern in the codebase and probably worth something agreeing with others.

crates/node/src/tests/verify_foreign_tx_gating.rs
anodar: Got rid of the comments in these sections but reworded and added it on top of the test. I think it's useful to have some kind of comments explaining what the test is supposed to be testing. 

[2434973](https://github.com/near/mpc/pull/3912/commits/243497372d8a917cf51bbd13356ee8f380044d11)

crates/node/src/foreign_chain_policy.rs
anodar: Addressed in [2434973](https://github.com/near/mpc/pull/3912/commits/243497372d8a917cf51bbd13356ee8f380044d11)

crates/node/src/foreign_chain_policy.rs
anodar: Addressed in [2434973](https://github.com/near/mpc/pull/3912/commits/243497372d8a917cf51bbd13356ee8f380044d11)

gilcu3 (dismissed):

crates/node/src/indexer/foreign_chain.rs
netrome: Yeah I'm a big fan of functional patterns, although it can sometimes be tricky in Rust if you want to take it too far due to lack of higher order data types etc.

> I'd rather not refactor this now though as this is existing pattern in the codebase and probably worth something agreeing with others.

100% agree. My main purpose of the rant is just to create alignment of desired code patterns so we don't leak sub optimal patterns into new modules.

crates/node/src/indexer/foreign_chain.rs
netrome: Yeah these kinds of spammy error logs are very annoying, especially considering how clunky the log UX is through dstack, so imo the extra code complexity is worth it. But I wouldn't block the PR on it.

Also, exponential backoff is a well-established pattern in the code base, so I wouldn't consider reusing it in more places adding much complexity. Especially since we still have a loop here. But anyway, do as you please for now. If this happens in production and annoys us we'll revisit and fix it later.

gilcu3 (approved):

haiyuechen-nearone (approved):

PR #3920, opened 2026-07-23
refactor: typed propose_update and vote_update on MpcContractHandle

### LINKED ISSUE #3919: feat: typed `propose_update` and `vote_update` on `MpcContractHandle`
### Background

`propose_update` is borsh-encoded and carries a 17 NEAR deposit; `vote_update` already has a typed `VoteUpdateArgs` but its call site still hand-rolls `json!({id})`. The handle needs its first borsh method, which also reintroduces the `Encode` error variant removed in #3873 for lack of a consumer.

### User Story

As a consumer of the MPC contract interface, I want typed `propose_update` and `vote_update` methods, so that contract upgrades don't require knowing the borsh encoding or the deposit amount.

### Acceptance Criteria

- [ ] `MpcContractHandle::propose_update(ProposeUpdateArgs)` (borsh; `MpcContractHandleError::Encode`; 17 NEAR via `deposits::PROPOSE_UPDATE_DEPOSIT_MILLINEAR`) and `vote_update(id: u64)`.
- [ ] The snapshot catalog renders non-printable (borsh) args as hex.
- [ ] e2e `propose_and_vote_contract_update` routes through the handle; `CONTRACT_UPDATE_GAS`/`CONTRACT_UPDATE_DEPOSIT` leave e2e.

kevindeforth:
Resolves #3919, part of #3693

This turned out more complex than anticipated, because I wanted the contract and interface to share the logic calculating the required deposit for update proposals.
Since we lacked a test for assessing whether the contract charges the correct deposits for updates, this PR adds that as well.

In the course of this, I discovered that we never use `UpdateEntry::bytes_used`, but deferred deleting that field to later, as it would require migration logic (c.f. #3965).

### DIFF
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -1333,10 +1333,19 @@ impl MpcContract {
     ) -> Result<UpdateId, Error> {
         // Only voters can propose updates:
         let proposer = self.voter_or_panic();
+        let payload_bytes =
+            args.payload_bytes()
+                .map_err(|err| InvalidParameters::MalformedPayload {
+                    reason: err.to_string(),
+                })?;
         let update: Update = args.try_into()?;
 
         let attached = env::attached_deposit();
-        let required = ProposedUpdates::required_deposit(&update);
+        let required = ProposedUpdates::required_deposit(payload_bytes).map_err(|err| {
+            InvalidParameters::MalformedPayload {
+                reason: err.to_string(),
+            }
+        })?;
         if attached < required {
             return Err(InvalidParameters::InsufficientDeposit {
                 attached: attached.as_yoctonear(),
@@ -2850,6 +2859,7 @@ mod tests {
     use crate::state::resharing::ResharingContractState;
     use crate::state::test_utils::{
         gen_initializing_state, gen_resharing_state, gen_running_state,
+        gen_running_state_with_params,
     };
     use crate::tee::measurements::{
         KeyProviderEventDigest, MrtdHash, Rtmr0Hash, Rtmr1Hash, Rtmr2Hash,
@@ -5809,6 +5819,53 @@ mod tests {
         );
         // then: threshold met (have 2 valid votes, need 2)
         assert!(contract.vote_update(update_id).unwrap());
+    }
+
+    #[test]
+    #[expect(non_snake_case)]
+    fn propose_update__should_charge_a_deposit_covering_the_worst_case_storage_cost() {
+        // Given
+        let running_state = gen_running_state_with_params(1, 129, 129);
+        let voters: Vec<AccountId> = running_state
+            .parameters
+            .participants()
+            .participants()
+            .iter()
+            .map(|(account_id, _, _)| account_id.clone())
+            .collect();
+        let mut contract =
+            MpcContract::new_from_protocol_state(ProtocolContractState::Running(running_state));
+        let code = vec![0xff; 4096];
+        let required_deposit = ProposedUpdates::required_deposit(
+            u128::try_from(code.len()).expect("usize always fits in u128"),
+        )
+        .expect("the deposit for a 4 KiB payload fits in u128");
+
+        // When
+        let mut environment = Environment::new(None, Some(voters[0].clone()), None);
+        environment.set_deposit(required_deposit);
+        let storage_before = env::storage_usage();
+        let id = contract
+            .propose_update(ProposeUpdateArgs {
+                code: Some(code),
+                config: None,
+            })
+            .unwrap();
+        for voter in &voters[1..] {
+            Environment::new(None, Some(voter.clone()), None);
+            assert!(!contract.vote_update(id).unwrap());
+        }
+        // near-sdk `store` collections write their cached changes to storage
+        // in `Drop`; dropping the contract persists the entry and the votes.
+        drop(contract);
+        let bytes_grown = u128::from(env::storage_usage() - storage_before);
+        let storage_cost = env::storage_byte_cost().saturating_mul(bytes_grown);
+
+        // Then
+        assert!(
+            required_deposit >= storage_cost,
+            "deposit {required_deposit} does not cover {storage_cost} of storage"
+        );
     }
 
     /// Callers authorized to drive `remove_non_participant_update_votes`.
--- a/crates/contract/src/update.rs
+++ b/crates/contract/src/update.rs
@@ -117,7 +117,7 @@ impl TryFrom<ProposeUpdateArgs> for Update {
 pub(crate) struct UpdateEntry {
     pub(super) update: Update,
     // TODO(#3965): remove this field, it's not read anywhere.
-    pub bytes_used: u128,
+    bytes_used: u128,
 }
 
 #[derive(Clone, Debug, PartialEq)]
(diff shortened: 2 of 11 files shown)


crates/near-mpc-contract-interface/src/deposits.rs
```diff
@@ -6,3 +6,5 @@
 pub const SUBMIT_PARTICIPANT_INFO_DEPOSIT_MILLINEAR: u128 = 100;
 
 pub const SIGN_DEPOSIT_YOCTONEAR: u128 = 1;
+
+pub const PROPOSE_UPDATE_DEPOSIT_MILLINEAR: u128 = 17_000;
```
kevindeforth: The required deposit depends on the binary size of the proposed update, which is why I am unsure if we want to hard-code it, or make it a parameter to be passed with the contract handle (or to compute it from the args).

_edit: I will stack a PR that computes the deposit based on the args, and make sure contract and interface use the same method for computing storage costs._

kevindeforth: proposal: https://github.com/near/mpc/pull/3933

crates/e2e-tests/src/cluster.rs
```diff
@@ -1461,11 +1452,6 @@ async fn create_user_accounts(
     Ok(map)
 }
 
-/// JSON mirror of the contract's `UpdateId`. `propose_update` returns it and
-/// `vote_update` consumes it via the `id` field; both wire it as a bare u64.
-#[derive(Debug, Clone, Copy, serde::Serialize, serde::Deserialize)]
-struct UpdateId(u64);
```
kevindeforth, line 1467: We didn't use this in a meaningful way, so might as well delete.

crates/contract/src/lib.rs
```diff
@@ -5811,6 +5826,55 @@ mod tests {
         assert!(contract.vote_update(update_id).unwrap());
     }
 
+    #[test]
+    #[expect(non_snake_case)]
+    fn propose_update__should_charge_a_deposit_covering_the_worst_case_storage_cost() {
```
kevindeforth, line 5826: This is testing the worst-case and checking that the required deposit covers the actual storage costs.

crates/near-mpc-contract-interface/Cargo.toml
```diff
@@ -39,11 +33,11 @@ near-contract-transport = { workspace = true, optional = true }
 near-mpc-bounded-collections = { workspace = true }
 near-mpc-crypto-types = { workspace = true }
 serde = { workspace = true }
-serde_json = { workspace = true, optional = true }
+serde_json = { workspace = true }
 serde_repr = { workspace = true }
 serde_with = { workspace = true }
 sha2 = { workspace = true }
-thiserror = { workspace = true, optional = true }
+thiserror = { workspace = true }
```
kevindeforth, line 40: note that the contract already has these dependencies, so adding them here doesn't lead to additional bloat in the binary size.

crates/contract/src/update.rs
```diff
@@ -113,7 +116,8 @@ impl TryFrom<ProposeUpdateArgs> for Update {
 )]
 pub(crate) struct UpdateEntry {
     pub(super) update: Update,
-    pub(super) bytes_used: u128,
+    // TODO(#3965): remove this field, it's not read anywhere.
+    pub bytes_used: u128,
```
kevindeforth: to my surprise...

crates/near-mpc-contract-interface/src/client.rs
```diff
@@ -195,12 +248,20 @@ mod tests {
...
+        //    In this case, we print the hex-encoding.
+        let args = match serde_json::from_slice::<serde_json::Value>(&call.args) {
+            Ok(_) => String::from_utf8(call.args.clone()).expect("JSON args are UTF-8"),
+            Err(_) => format!("0x{}", hex::encode(&call.args)),
+        };
         format!(
-            "contract: {contract_id}\nmethod:   {}\ngas:      {}\ndeposit:  {}\nargs:     {}",
+            "contract: {contract_id}\nmethod:   {}\ngas:      {}\ndeposit:  {}\nargs:     {args}",
             call.method_name,
             call.gas,
```
kevindeforth, line 265: So, claude complained about this comment, saying it was paraphrasing, but I actually deliberately wrote this, because I felt the need to explain _why_ we are doing this. Lmk if you prefer I delete it.

crates/near-mpc-contract-interface/src/deposits.rs
```diff
@@ -6,3 +6,51 @@
 pub const SUBMIT_PARTICIPANT_INFO_DEPOSIT_MILLINEAR: u128 = 100;
 
 pub const SIGN_DEPOSIT_YOCTONEAR: u128 = 1;
+
+pub const STORAGE_BYTE_COST_YOCTONEAR: u128 = 10_000_000_000_000_000_000;
```
anodar, line 10: Why not just use `env::storage_byte_cost()` ?

anodar (approved):

crates/near-mpc-contract-interface/src/deposits.rs
kevindeforth: Excellent question, I probably should have mentioned that in the PR.
We would need to bring in the near-sdk dependency and every user of this interface crate would also get it. It's a bit much, so I preferred to have a simple unit-test.

anodar: Makes sense, thanks for the explanation.

crates/contract/src/lib.rs
```diff
@@ -5811,6 +5821,53 @@ mod tests {
...
+        }
+        // near-sdk `store` collections write their cached changes to storage
+        // in `Drop`; dropping the contract persists the entry and the votes.
+        drop(contract);
+        let bytes_grown = u128::from(env::storage_usage() - storage_before);
+        let storage_cost = env::storage_byte_cost().saturating_mul(bytes_grown);
+
+        // Then
+        assert!(
+            required_deposit >= storage_cost,
```
netrome, line 5866: Would be cool if we could have gotten exact equality here but I guess that's infeasible 😅

crates/near-mpc-contract-interface/src/client.rs
```diff
@@ -107,6 +112,52 @@ impl<C: CallContract> MpcContractHandle<C> {
...
     }
 
+    pub async fn propose_update(
+        &self,
+        args: ProposeUpdateArgs,
+    ) -> Result<C::Output, MpcContractHandleError<C::Error>> {
+        let payload_bytes = args.payload_bytes().map_err(|err| match err {
+            PayloadBytesError::Serialize(err) => MpcContractHandleError::Serialize(err),
+            PayloadBytesError::Overflow => MpcContractHandleError::Deposit(DepositOverflowError),
+        })?;
```
netrome, line 122: Did you consider implementing `From<PayloadBytesError>` for `MpcContractHandleError` to skip this conversion here?

crates/near-mpc-contract-interface/Cargo.toml
netrome: Thanks for highlighting!

netrome (approved):
Nice stuff!

crates/near-mpc-contract-interface/src/client.rs
kevindeforth: Hmm, I didn't, that's a good suggestion. Will include this in a follow-up PR.

PR #3932, opened 2026-07-23
feat: probe Starknet by chain id and a recent transaction

haiyuechen-nearone:
Part of #3764; rationale: #3969. Stacked, merge bottom-up:

1. **#3932 (this PR)** — Starknet: chain id + a recent L1-accepted tx
2. #3962 — 6 EVM chains: `eth_chainId` + a recent finalized tx
3. #3963 — Bitcoin: genesis hash + a recent tx
4. #3964 — Aptos: ledger chain id + a recent tx

Replaces Starknet's pinned golden tx with the Sui-style probe: verify `starknet_chainId`, then inspect a real tx from a recent `l1_accepted` block at `AcceptedOnL1` (walks back past empty/not-yet-final blocks; needs provider JSON-RPC v0.9+).

Identities come from config (`foreign_chain_health_check.identities`), no built-ins — so any network is checkable, and a configured chain fails until its identity is set. Also lands the shared scaffolding the rest of the stack reuses.

### DIFF
--- a/crates/foreign-chain-config-tester/src/config.rs
+++ b/crates/foreign-chain-config-tester/src/config.rs
@@ -12,7 +12,7 @@ use serde::Deserialize;
 use serde::de::IntoDeserializer;
 use serde::de::value::{Error as ValueError, StrDeserializer};
 
-use foreign_chain_health_check::Network;
+use foreign_chain_health_check::{ExpectedIdentities, Network};
 
 /// Paths where `foreign_chains` may live, most-nested first so a wrapped config
 /// matches before a barer one.
@@ -31,6 +31,19 @@ const CONTRACT_ID_PATHS: &[&[&str]] = &[
     &["mpc_node_config", "node", "indexer", "mpc_contract_id"],
     &["node", "indexer", "mpc_contract_id"],
     &["indexer", "mpc_contract_id"],
+];
+
+/// Where the per-chain expected identities live: an `identities` map (chain label ->
+/// expected identity) under a sibling of `foreign_chains`.
+const EXPECTED_IDENTITY_PATHS: &[&[&str]] = &[
+    &[
+        "mpc_node_config",
+        "node",
+        "foreign_chain_health_check",
+        "identities",
+    ],
+    &["node", "foreign_chain_health_check", "identities"],
+    &["foreign_chain_health_check", "identities"],
 ];
 
 fn classify_network(chain_id: Option<&str>, contract_id: Option<&str>) -> Option<Network> {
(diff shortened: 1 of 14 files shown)


crates/foreign-chain-config-tester/src/config.rs
```diff
@@ -100,6 +127,20 @@ fn yaml_str<'a>(root: &'a serde_yaml::Value, path: &[&str]) -> Option<&'a str> {
     path.iter().try_fold(root, |v, k| v.get(k))?.as_str()
 }
 
+/// Per-chain expected identities from config. Absent chains stay `None` (their check then
+/// fails until configured); an unknown chain key or non-string value is a hard error.
+pub fn detect_expected_identities(
```
netrome: Nit: What do you think about renaming this to `parse_expected_identities` to maintain parity with `parse_foreign_chains`?

crates/foreign-chain-health-check/src/lib.rs
```diff
@@ -324,6 +327,24 @@ fn prepare_sui(
...
 
+/// A configured identity-probed chain without an expected identity is a config error:
+/// every provider row fails until `foreign_chain_health_check.identities.<chain>` is set.
+fn mark_missing_identity(
+    chain: &'static str,
+    cfg: &ForeignChainConfig,
+    out: &mut Vec<ProviderResult>,
+) {
+    for name in cfg.providers.keys() {
+        out.push(ProviderResult {
```
netrome: Hmm not a fan of pushing to a mutable vector here. I see this was the established pattern, but we should consider refactoring this for a cleaner and more maintainable implementation now when this becomes part of our node code.

deployment/cvm-deployment/user-config.toml
```diff
@@ -148,3 +148,16 @@ max_retries = 3
...
+max_retries = 3
+
+[mpc_node_config.node.foreign_chains.sui.providers.fullnode]
+rpc_url = "https://fullnode.testnet.sui.io:443"
+
+# Expected identity per chain for the provider health check — see
+# crates/foreign-chain-config-tester/README.md for the well-known values.
+[mpc_node_config.node.foreign_chain_health_check.identities]
+starknet = "0x534e5f5345504f4c4941" # mainnet: "0x534e5f4d41494e"
+sui = "69WiPg3DAQiwdxfncX6wYQ2siKwAe6L9BZthQea3JNMD" # mainnet: "4btiuiMPvEENsttpZC7CZ53DruC3MAgfznDbASZ7DR6S"
```
netrome: This looks really nice.

netrome (changes requested):
So we need to discuss this further. Now this PR modifies a separate health check CLI. For this CLI, I'm not as convinced that querying transactions is wrong. I can see the point of doing that as long as it's stand-alone. Additionally, I am not too fond of the idea of reusing too much of the `foreign-chain-health-check` crate in our node code as-is, as it seems to me like a vibe-coded CLI with questionable code-quality that only passed shallow reviews in: https://github.com/near/mpc/pull/3716

There's a lot of code smell (large functions, helpers mutating shared collections instead of being pure functions that could be executed in parallel and tested more cleanly) that I'd be very hesitant to bring into node code.

Finally, I don't think it should be too complicated to just extend our inspector without touching any of the `foreign-chain-config-tester` code. I'd picture us adding some kind of trait to get the chain identity of any inspector and checking these against configured values on startup. I think this could greatly simplify things.

I'll schedule a call so we can chat about this tomorrow.

haiyuechen-nearone:
> Now this PR modifies a separate health check CLI.

My intention was that the CLI and the healthcheck should share the core probing logic, so that the same code flow can be tested against real providers easily through the CLI (which is what I have tested this PR with as well). And then by extension the CLI needed the changes because of the config changes.

crates/foreign-chain-health-check/src/lib.rs
haiyuechen-nearone: Something like this maybe?
```
enum HealthCheckStatus {
    Healthy
    Unhealthy
}

struct ForeignChainProviderState {
    provider_name: String
    health_check_status: HealthCheckStatus
}

struct ForeignChainSnapshot {
    chain: ForeignChain
    configured: bool
    providers_state: Vec<ForeignChainProviderState>
}

struct ForeignChainProvidersSnapshot {
    snapshots: Vec<ForeignChainSnapshot>
}
```

My habit is to introduce minimal changes for each PR, and leave any refactors for a separate PR where there is no behavior change. I can try to put the refactors upfront if that is the preferred work flow.

netrome:
> My intention was that the CLI and the healthcheck should share the core probing logic, so that the same code flow can be tested against real providers easily through the CLI (which is what I have tested this PR with as well). And then by extension the CLI needed the changes because of the config changes.

Fair point, I'm happy to share the probing logic between the two. However, the shared surface should uphold production standards that I'm not convinced that the CLI does. To achieve this shared surface, I believe the simplest approach would be to introduce the chain ID probes in the node first, and then reuse those probes in the CLI.

I'm happy to do it the other way around, but this front-loads much more code cleanups in the health-check and it's harder to picture how the probes will be integrated in the node requiring more context in reviews.

gilcu3:
Superseded by #4013

PR #3933, opened 2026-07-23
refactor(interface): share the propose_update deposit calculation

kevindeforth:
This is part of https://github.com/near/mpc/pull/3920#discussion_r3636690888


kevindeforth:
@claude These are all non-issues. In case of divergence, the tests will fail and we will catch it.

Edit: though, maybe a vote might fail? Seems unlikely, but well.

Issue #3903, opened 2026-07-22
Get completely rid of pure "threshold" terminology


SimonRastikian:
### Background

The terminology "threshold" still exists in the node and in the smart contract and should be completely get rid of as it is very ambiguous and replaced either by reconstruction_threshold or by governance_threshold.

### Acceptance Criteria

Cover the changes in the entire code base including in, structures, variable names, and function names.

PR #3935, opened 2026-07-23
refactor: Contract internal-only renames from threshold to reconstruction threshold

SimonRastikian:
Fixes part of #3903. Namely the contract internals. This does not create migration "problems".

Upcoming PR: Cross-crate TYPE renames

### DIFF
--- a/crates/contract/src/primitives/thresholds.rs
+++ b/crates/contract/src/primitives/thresholds.rs
@@ -357,7 +357,7 @@ mod tests {
     }
 
     #[test]
-    fn validate_threshold__should_not_produce_empty_window_for_small_n() {
+    fn validate_governance_threshold__should_not_produce_empty_window_for_small_n() {
         // The relative upper cap is clamped up to the ceil(0.6n) lower bound, so the
         // feasible window must always hold at least one valid threshold.
         for n in 2..=12u64 {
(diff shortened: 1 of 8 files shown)


crates/contract/src/primitives/thresholds.rs
```diff
@@ -38,7 +38,7 @@ impl ThresholdParameters {
     /// Constructs Threshold parameters from `participants` and `threshold` if the
     /// threshold meets the absolute and relative validation criteria.
     pub fn new(participants: Participants, threshold: Threshold) -> Result<Self, Error> {
-        match Self::validate_threshold(participants.len() as u64, threshold) {
+        match Self::validate_governance_threshold(participants.len() as u64, threshold) {
```
pbeza, line 41: Nit: Isn’t it a bit odd to have a `validate_governance_threshold` function on a struct named `ThresholdParameters`, which sounds more generic? Should we rename the struct to `GovernanceThresholdParameters` and shorten the function name back to `validate_threshold`, since the struct’s scope already makes it clear what the threshold is for?

SimonRastikian: This will be done in the next PR as it kind of affect abi generation

pbeza (dismissed):

pbeza (approved):

haiyuechen-nearone (approved):
🏎️

PR #3564, opened 2026-06-15
feat(contract): auto-expire unused launcher image hashes

barakeinav1:
Implements the approved design ([docs/design/auto-remove-launcher-hashes-design.md](https://github.com/near/mpc/blob/main/docs/design/auto-remove-launcher-hashes-design.md), PR #3488).

Usage-based expiry for `allowed_launcher_image_hashes`, replacing unanimous-vote-only removal for routine cleanup.

**Model.** Each `AllowedLauncherImage` stores an `expires_at: Timestamp`, stamped `now + TTL` at write time (vote-in, re-vote, and on a successful attestation). An entry is expired when `expires_at < now`. Because expiry is fixed at write time — mirroring how attestations store their own expiry — reads are a plain comparison and the TTL only touches the write sites; it isn't threaded through the ~20 read/verify paths.

- **Refresh on use** — a successful attestation restamps `expires_at = now + TTL`, but **only for a current participant** (enforced by requiring an `AuthenticatedParticipantId` capability token), so a prospective/non-participant node cannot keep a launcher alive. Applies to both attestation paths: the mock path (`MockAttestation::WithConstraints` may reference a launcher) and the Dstack path (refresh runs in the async `resolve_verification` callback; the signer is preserved across the verifier promise). No node-side changes.
- **Read-time filtering** — all reads of the allowed set skip expired entries; newest-entry fallback so the set is never empty.
- **Inline eviction** — `reverify_and_cleanup_participants` (the body of `verify_tee`) evicts expired entries inline, right after the analogous cleanup of the MPC docker-image hashes; the allowed set is a small in-memory `Vec` (`retain`, no separate receipt) and never removes the last entry. No `#[private]` method, no extra config field.
- **Re-vote recovery** — `vote_add_launcher_hash` on an already-present hash restamps its `expires_at` (threshold vote), recovering a never-adopted hash.
- **Config** — new `launcher_hash_unused_ttl_seconds` (default 14d), validated `>= DEFAULT_EXPIRATION_DURATION_SECONDS` (the attestation validity window). Validation is folded into the DTO→`Config` `TryFrom`, so `init` / `init_running` / `update_config` can't skip it.

## Migration

[...]

PR #3936, opened 2026-07-23
test(contract): gas-budget guard for clean_expired_launcher_hashes

barakeinav1:
Adds a sandbox **gas-budget guard** for the detached `clean_expired_launcher_hashes` self-call.

`verify_tee` spawns that `#[private]` call with a fixed `clean_expired_launcher_hashes_tera_gas` budget (default 5 TGas). The test builds a `Running` contract with a small launcher allowlist, invokes the method, and asserts the burned gas stays under the budget — so a future change that bloats the method can't silently exceed the attached gas.

Measured on the sandbox: **~3.0 TGas** (function-call receipt) vs the 5 TGas budget. The cost is dominated by contract state load/store + the base function-call charge, not the launcher count.

> [!NOTE]
> The launcher-expiry **correctness** tests (expiry rejection + participant-gated refresh) landed directly in #3564. This PR is now just the gas guard.
>
> Stacked on #3564 (base branch `feat/auto-remove-launcher-hashes`); GitHub retargets to `main` once #3564 merges.


barakeinav1:
Obsolete — superseded by #3564.

The expired-launcher eviction no longer runs in a detached `#[private] clean_expired_launcher_hashes` self-call. Per review, it's now an inline `Vec::retain` inside `reverify_and_cleanup_participants` (right after the MPC docker-hash cleanup), so there is no separate promise to reserve a gas budget for — the method, the config field, and the guard this PR added are all gone (near/mpc@a6571105).

This PR shows as merged, but that merge was dropped when `feat/auto-remove-launcher-hashes` was later rewritten, so its changes aren't in the current tip and nothing needs reverting. The `bench/clean-expired-launcher-gas` branch can be deleted.

PR #3938, opened 2026-07-23
refactor: cross-crate type renames

SimonRastikian:
Closes part of #3903 namely adding Governance prefix to struct names.
It also regen abi.

### DIFF
--- a/crates/contract/src/state/key_event.rs
+++ b/crates/contract/src/state/key_event.rs
@@ -4,7 +4,7 @@ use crate::errors::KeyEventError;
 use crate::errors::VoteError;
 use crate::primitives::key_state::KeyEventId;
 use crate::primitives::key_state::{AttemptId, EpochId};
-use crate::primitives::thresholds::ThresholdParameters;
+use crate::primitives::thresholds::GovernanceThresholdParameters;
 use crate::state::AuthenticatedParticipantId;
 use near_mpc_contract_interface::types::{DomainConfig, DomainId};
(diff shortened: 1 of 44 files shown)


crates/contract/src/state/key_event.rs
```diff
@@ -20,7 +20,7 @@ pub struct KeyEvent {
     /// The domain that we're generating or resharing the key for.
     domain: DomainConfig,
     /// The participants and threshold that shall participate in the key event.
-    parameters: ThresholdParameters,
+    parameters: GovernanceThresholdParameters,
```
gilcu3, line 23: What about saying `NetworkParameters` instead of `GovernanceThresholdParameters`? The later seems a bit misleading, as it includes everything, not only the governance stuff

gilcu3 (dismissed):
LGTM

crates/contract/src/state/key_event.rs
SimonRastikian: I will keep it as is right now as the change you are proposing is out of scope for this PR. This PR is supposed to not let hanging the term threshold alone

gilcu3: Yeah not a problem. But do you think that name would make more sense?

gilcu3 (dismissed):

pbeza (dismissed):

pbeza (approved):

gilcu3 (approved):

PR #3939, opened 2026-07-23
refactor(sandbox): use mpc contract interface

### LINKED ISSUE #4120: use `CallContract` interface in sandbox
### Background

With https://github.com/near/mpc/issues/3693, we started exposing the MpcContract interface for different back-ends. We should use it in the sandbox and wire up the methods we already have.

### User Story

As a dev, I like smaller scoped issues.

### Acceptance Criteria

Admin and signature request logic should go through the Mpc contract interface.

_No response_

kevindeforth:
resolves #4120, part of https://github.com/near/mpc/issues/3693

### DIFF
--- a/crates/contract/tests/sandbox/common.rs
+++ b/crates/contract/tests/sandbox/common.rs
@@ -25,7 +25,6 @@ use mpc_contract::{
         },
     },
     tee::tee_state::NodeId,
-    update::UpdateId,
 };
 use near_account_id::AccountId;
 use near_mpc_contract_interface::types::{
@@ -368,7 +367,7 @@ pub async fn propose_and_vote_contract_binary(
         "propose update call failed"
     );
 
-    let proposal_id: UpdateId = propose_update_execution.json().unwrap();
+    let proposal_id: u64 = propose_update_execution.json().unwrap();
 
     // Try calling into state and see if it works.
     let state_request_execution = accounts[0]
@@ -381,7 +380,7 @@ pub async fn propose_and_vote_contract_binary(
         .json()
         .expect("state is deserializable.");
 
-    vote_update_till_completion(contract, accounts, &proposal_id).await;
+    vote_update_till_completion(contract, accounts, proposal_id).await;
 
     let contract_binary_post_upgrade = contract.view_code().await.unwrap();
     assert_eq!(
@@ -394,12 +393,12 @@ pub async fn propose_and_vote_contract_binary(
 pub async fn vote_update_till_completion(
     contract: &Contract,
     accounts: &[Account],
-    proposal_id: &UpdateId,
+    proposal_id: u64,
 ) {
     for voter in accounts {
         let execution = voter
             .call_mpc(contract.id())
-            .vote_update(**proposal_id)
+            .vote_update(proposal_id)
             .await
             .unwrap();
 
--- a/crates/contract/tests/sandbox/common.rs
+++ b/crates/contract/tests/sandbox/common.rs
@@ -1,5 +1,5 @@
 use crate::sandbox::utils::{
-    consts::{CURRENT_CONTRACT_DEPLOY_DEPOSIT, GAS_FOR_INIT, GAS_FOR_VOTE_UPDATE, PARTICIPANT_LEN},
+    consts::{GAS_FOR_INIT, PARTICIPANT_LEN},
     contract_build::current_contract,
     initializing_utils::{start_keygen_instance, vote_add_domains, vote_public_key},
     mpc_contract::{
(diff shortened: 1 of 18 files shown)


kevindeforth:
> Sandbox-only gas tripwires are promoted to the production wire format

This is intentional and absolutely irrelevant for this PR. The node still attaches 300 Tgas to every call. The five changed methods are operator-cast governance votes.

Also, please note that we have the intention to lower the gas constants even for the node: https://github.com/near/mpc/issues/166.


> crates/contract/tests/sandbox/utils/consts.rs:16 — the surviving doc comment is now stale in two ways: it formally documents only GAS_FOR_VOTE_RESHARED (leaving GAS_FOR_VOTE_PK / GAS_FOR_INIT orphaned under a comment written for the whole removed group), and "nodes usually attach the maximum available gas" stops being true for the vote_* family once this PR lands. Worth rewording alongside whichever option above you pick.

Yeah, this is the problem with low-quality LLM-generated comments. But the orphaning predates this PR. That /// was already attached to GAS_FOR_VOTE_RESHARED alone. Also, "nodes usually attach the maximum available gas" is still accurate (even though it is a bad, LLM-generated comment. Thank your peers for that).

> The two call sites it enables (crates/e2e-tests/src/cluster.rs:824, client.rs:406) trade named struct-literal fields for three positional args, which reads worse than what it replaced.

That's debatable.  Generally, we prefer constructors because refactors are easier to handle.

>  the ```ignore example never compiles (rustdoc does not document tests/ crates) and references an undefined `id`, so it can only rot. The prose line above it already carries the non-obvious part — that every transaction is in flight before any is awaited. Suggest deleting the code block.

For now, I would prefer to leave it in, as it illustrates how to use the method. Will remove if humans object.

>  the deleted doc on SandboxCaller leaves both adapters undocumented, and the distinction that matters (transact() → ExecutionFinalResult vs transact_async() → TransactionStatus, i.e. which one lets you batch) is not recoverable from the names. One line on AsyncSandboxCaller would carry it.

What do you mean _is not recoverable from the names_? One is explicitly called _async_. You are asking to paraphrase the code, which is not advisable.

_edit: maybe, it migh make sense to add a comment explaining why both exists. I can do that._

>  the largest sign-test file still submits via the raw .call(...).args_json(req.request_json_args()).max_gas() path, so sign requests now run at 300 Tgas there and 15 Tgas everywhere else in the same suite. [...] worth either finishing here or noting in the tracking issue.

Did you use a context window of 1? There are two tests in which we use 300 Tgas, one is for V1 compatibility, the other is because we don't attach a deposit and expect the call to fail. Nothing to finish or track here.

crates/contract/tests/sandbox/vote.rs
```diff
@@ -809,22 +802,19 @@ async fn vote_new_parameters_errors_if_new_participant_is_missing_valid_attestat
...
-                    threshold_parameters.clone(),
-                    BTreeMap::new(),
-                ),
-            }))
-            .transact()
+        let call_result = account.call_mpc(contract.id())
+            .vote_new_parameters(
+                prospective_epoch_id,
+                proposed_parameters.clone()
+            )
```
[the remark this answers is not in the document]
kevindeforth: yeah, I was very surprised this passed. But I think it also passes with the newline, so I went with that: https://github.com/near/mpc/pull/3939/commits/c1f7cf712da2699ab5791302ef64b7651063aad1 + https://github.com/near/mpc/pull/3939/commits/2105e2286d4c088e0640ae3dc330a6b058901e38

edit:
oh this seems to be passing rustfm because the expression is too long. There seems to be a nightly option, so I would expect at some point, we will have this enabled:

>   - Make it loud: error_on_line_overflow = true in rustfmt.toml turns the silent bail into an error, so this can't hide again. It's a nightly-only option though, so it would only bite on a nightly cargo fmt — worth knowing, probably not worth adopting.

haiyuechen-nearone (dismissed):
LGTM

crates/contract/tests/sandbox/common.rs
```diff
@@ -400,12 +398,8 @@ pub async fn vote_update_till_completion(
 ) {
     for voter in accounts {
         let execution = voter
-            .call(contract.id(), method_names::VOTE_UPDATE)
-            .args_json(serde_json::json!({
-                "id": proposal_id,
```
haiyuechen-nearone, line 405: nit: why don't we take `proposal_id` in as `UpdateId` instead of `&UpdateID`? We are derefing and copying in the loop anyways.

haiyuechen-nearone: Github rendering really sucks, I meant the location here
image

kevindeforth: Yeah, that's eye cancer. The issue is that we were using the `UpdateId` from the contract, which we shouldn't use in the interface crate (as it's a contract-internal type).
I modified the code to use plain u64 instead of `UpdateId` inside the sandbox: https://github.com/near/mpc/pull/3939/commits/64da134bd282ce4fd5324d9e362c365803b2e893. This makes it consistent with the e2e tests and can be kept like this until we decide to make `UpdateId` part of the contract interface crate.

haiyuechen-nearone (approved):

pbeza (approved):

Issue #3825, opened 2026-07-15
docs: align TEE attestation docs with the no-yield `submit_participant_info` design


pbeza:
## Background

The async-attestation work moved the Dstack `submit_participant_info` flow from a yield-resume design to a no-yield promise-chain design. The TEE attestation docs still describe the old design and show stale on-chain error strings, so they need to be brought in line with the shipped behavior. These doc updates were split out of the async-attestation tests PR to keep that PR test-only.

## Acceptance Criteria

- [ ] `docs/design/attestation-verifier-contract.md` describes the no-yield promise-chain flow (`submit_dstack_attestation` + `resolve_verification`, deposit forwarded on the callback, `fail_attestation_submission` on failure) instead of yield-resume / `pending_attestations` / timeout
- [ ] Updated `re_verify` source reference points to the current location
- [ ] Stale on-chain error strings corrected in `docs/localnet/tee-localnet.md` and `docs/running-an-mpc-node-in-tdx-external-guide.md`

## Resources & Additional Notes

Delivered as a stacked PR on top of the async-attestation tests PR.

barakeinav1:
The first acceptance criterion needs updating: #3940 removes the deposit from `submit_participant_info` entirely.

> - [ ] `docs/design/attestation-verifier-contract.md` describes the no-yield promise-chain flow (`submit_dstack_attestation` + `resolve_verification`, **deposit forwarded on the callback**, `fail_attestation_submission` on failure) instead of yield-resume / `pending_attestations` / timeout

There is no deposit to forward. Neither method is `#[payable]` any more, nothing rides forward via `.with_attached_deposit`, and there is no charge or refund — attestation storage is funded from the contract's own balance, because the node signs with a function-call access key that cannot attach a deposit (#3925). So that clause should read something like "no deposit: attestation storage is contract-funded" instead.

The rest of the criteria still hold. Worth noting the ordering: #3826 already implements this issue, but was written before #3940 and documents the deposit machinery throughout — details and the affected passages are in https://github.com/near/mpc/pull/3826#issuecomment-5102147783. #3826 needs rebasing onto #3940 rather than merging independently, or the doc swaps yield drift for deposit drift.

I've added a `TODO(#3825)` marker over the two stale sections on #3940 so `main` doesn't silently carry the yield-resume description in the meantime.

PR #3940, opened 2026-07-23
fix(contract): drop the deposit for `submit_participant_info`

### LINKED ISSUE #903: [Docs-Missing] Document cost in NEAR tokens for calling submit_participant_info

### LINKED ISSUE #3857: Refine attestation storage deposit: charge by actual storage delta
## Background

The flat attestation storage fee introduced in #3714 keeps the whole 0.1 NEAR deposit on a successful `submit_participant_info` and refunds it on failure. Two known shortcomings (from #3714 review):

- Re-submissions overpay: they change little to no stored storage yet pay the full deposit each time.
- Changes to the attestation structure can cause over- or under-payment against the flat fee.

Acceptable short/medium-term and out of scope for #3714, but worth addressing soon.

## Acceptance Criteria

- [ ] The deposit charged reflects the actual storage delta of the submission rather than a flat fee, so re-submissions with no storage change are not overcharged.
- [ ] The charge stays correct when the stored attestation structure changes size (no over- or under-payment).

Follow-up from #3714 review: https://github.com/near/mpc/pull/3714#pullrequestreview-4733749423

### LINKED ISSUE #3925: main: node with a function-call signer key cannot submit_participant_info (deposit → DepositWithFunctionCall)
## Background

`submit_participant_info` on `main` now attaches a **0.1 NEAR deposit**, which a node's **function-call access key cannot pay** — so a node whose signer is a function-call key (the standard, deliberately-not-full-access setup) can no longer submit its attestation. It fails at transaction validation with `InvalidAccessKeyError(DepositWithFunctionCall)`.

we use access-key for the node
see here:
[Add the Node Account Key to Your Account](https://github.com/near/mpc/blob/main/docs/running-an-mpc-node-in-tdx-external-guide.md#add-the-node-account-key-to-your-account)

**Facts:**
- #3714 (merged to `main`) introduced the deposit: new `crates/near-mpc-contract-interface/src/deposits.rs` → `SUBMIT_PARTICIPANT_INFO_DEPOSIT_MILLINEAR = 100`, attached in `crates/near-mpc-contract-interface/src/client.rs` (`deposit: NearToken::from_millinear(SUBMIT_PARTICIPANT_INFO_DEPOSIT_MILLINEAR)`). `release/v3.13` attached **no** deposit on this call.
- NEAR protocol rule (nearcore `runtime/runtime/src/verifier.rs`): a function-call access key cannot attach **any** deposit:
  ```rust
  if function_call.deposit > Balance::ZERO {
      return Err(InvalidTxError::InvalidAccessKeyError(InvalidAccessKeyError::DepositWithFunctionCall));
  }
  ```
  side note: This check is independent of the key's `method_names` — it runs before the receiver/method checks, so an fc key allowed to call *any* method on the contract still can't carry a deposit.
- Observed on a testnet TEE cluster after upgrading nodes to the `main` image (contract still 3.13.2; the check is client/protocol-side, so the contract version is irrelevant): every attestation submit fails:
  ```
  ERROR tx_sender: Failed to forward transaction SubmitParticipantInfo ... err=unexpected ProcessTxResponse: InvalidTx(InvalidAccessKeyError(DepositWithFunctionCall))
  ERROR periodic_attestation_submission: failed to submit attestation cause=attestation submission was not executed
  ```
  The node's signer is a function-call key (`near_signer_public_key` maps to a `FunctionCall` access key restricted to the mpc contract). The same key attested fine on `release/v3.13` (no deposit) — the previously stored attestation is still on-chain.

## User Story

As an operator running an MPC node with a **non-full-access (function-call) signer key** — the secure default so the always-online key can't move funds or add keys — I need my node to be able to submit its TEE attestation, so it can (re)attest and participate after upgrading to `main`.

## Theory (unverified — to confirm with the PR authors)

Whether the intended node-signer model is a function-call key (in which case attaching a deposit to `submit_participant_info` is a regression) or full-access (in which case this is an intentional tradeoff that should be documented, with its security implications). The function-call signer is the standard deploy output, so this reads as a regression — but I'm flagging it as a question, not asserting intent.

## Acceptance Criteria

- [ ] A node using a **function-call** signer key can submit `submit_participant_info` on `main` and become attested. 

Options to consider: 

don't attach a caller deposit (cover the async `pending_attestation` storage another way — e.g., contract-funded, or drawn from the participant's existing on-contract balance), 

**or** explicitly require full-access node signer keys and document that (calling out the security tradeoff).  - this is a problematic option, since this violets separation of duty and least privilege principles.  (note - we already compromise on this a bit in 03530d6d904eb5739e52107e445f9bcee337f01b 

- [ ]  make sure we have RUST integration tests to cover this

- #3714 (async TEE attestation; introduced the deposit)
- `deposits.rs` / `client.rs` (deposit constant + attach site) on `main` (HEAD `ebc7f0d5`)
- nearcore `runtime/runtime/src/verifier.rs` (the `DepositWithFunctionCall` rule)
- Related: #3909 (separate stale-node epoch-sync issue on the same cluster)

pbeza:
Closes #3925.
Closes #3857.
Closes #903.

`submit_participant_info` is no longer `#[payable]` — it takes no deposit and **rejects** one if attached, and the contract funds attestation storage from its own balance. A node self-submits with its function-call access key for its first attestation and re-attestations alike.

This fixes onboarding: #3714 introduced a deposit, which a function-call key cannot attach (`DepositWithFunctionCall`), so nodes could no longer attest.

## Why removing the deposit is safe

Pre-#3714 the contract had "charge the caller for a new entry" logic that never collected: it read `env::storage_usage()` immediately after inserting into the `IterableMap`, before the deferred write was flushed, so the delta read as 0 and nothing was charged. New entries were **already contract-funded, by accident**. This makes that intentional.

The cost is bounded by construction — every field of the stored entry is fixed-width except `account_id`, which NEAR caps at 64 bytes. Its exact size is pinned at 604 bytes (~0.006 NEAR) by `submit_participant_info__should_store_exactly_the_pinned_entry_size`, under a 0.01 NEAR ceiling, so a schema change that bloats it fails CI.

`Dstack` entries expire and are reclaimed by `clean_invalid_attestations`.
 `Mock::Valid` never expires and is never reclaimed — pre-existing (but will be fix in https://github.com/near/mpc/pull/3785 -) , and the same exposure `release/v3.13` shipped with; see #3972. 

## Behavior after this change

- any submit (first attestation or re-attestation) -> no deposit; contract funds the storage
- async (dstack) failure -> submission fails, nothing stored

## Follow-ups

Referenced, not closed by this PR:

- **#3972** — moving the storage cost back onto whoever onboards a node. Its pay-on-behalf criterion needs rework: an operator cannot submit for a node, because `report_data` binds the quote to `env::signer_account_pk()` ([details](https://github.com/near/mpc/issues/3972#issuecomment-5102251201)).
- **#3991** — `fail_attestation_submission` only existed so a failing `resolve_verification` could commit its refund first. With the deposit gone nothing needs to survive the error branch, so the method and its `Config` gas knob can go — a state-schema change, hence not folded in here. Sequence it after #3972.
- **#3825** — `docs/design/attestation-verifier-contract.md` still describes the superseded yield-resume design. Out of scope here, so the affected sections carry a `TODO(#3825)`. #3826 is the in-flight fix and needs rebasing onto this PR, since it documents the deposit machinery removed here.


crates/contract/tests/sandbox/utils/mpc_contract.rs
```diff
@@ -65,6 +65,22 @@ pub async fn submit_participant_info(
...
+    attestation: &Attestation,
+    tls_key: &Ed25519PublicKey,
+) -> anyhow::Result<NearToken> {
+    let balance_before = account.view_account().await?.balance;
+    let result = submit_participant_info(account, contract, attestation, tls_key).await?;
+    assert!(result.is_success(), "submission should succeed: {result:?}");
+    let balance_after = account.view_account().await?.balance;
+    let net_spent = balance_before.saturating_sub(balance_after);
+    Ok(net_spent.saturating_sub(total_gas_fee(&result)))
+}
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/commit/ba694a913

crates/contract/src/lib.rs
```diff
@@ -160,6 +159,19 @@ fn refund_to(account_id: &AccountId, amount: NearToken) {
...
+/// [`TeeState::store_verified_attestation`]); the [`MINIMUM_ATTESTATION_STORAGE_DEPOSIT`]
+/// floor guarantees the refund never underflows.
+fn keep_storage_delta_and_refund_rest(account_id: &AccountId, initial_storage: u64) {
+    // saturating_sub: a shrink charges nothing rather than underflowing.
+    let bytes_grown = env::storage_usage().saturating_sub(initial_storage);
+    let cost = env::storage_byte_cost().saturating_mul(u128::from(bytes_grown));
+    if let Some(refund) = env::attached_deposit().checked_sub(cost) {
+        refund_to(account_id, refund);
+    }
+}
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/commit/c0886f071

docs/design/attestation-verifier-contract.md
```diff
@@ -139,7 +139,7 @@ Each [`PendingAttestation`](#mpc-contractsubmit_participant_info) holds:
 
 - **The submitter's `Attestation::Dstack` payload** — the RTMR3 event log, app-compose, and report-data the post-DCAP checks consume.
 - **The submitter's TLS public key** — the callback hashes it with the submitter's account public key and compares to the quote's `report_data` field, proving the enclave produced the quote for this specific submitter.
-- **The attached deposit** — covers storage staking on success, refunded to the signer of the original `submit_participant_info` transaction on failure. `env::attached_deposit()` is not visible from the callback receipt, so the value is stashed at submit time and the recipient `AccountId` is the same one used to key the entry (set by `Self::assert_caller_is_signer()`).
+- **The attached deposit** — on success the contract keeps only the actual storage delta of the stored entry and refunds the excess; on failure the full deposit is refunded to the signer of the original `submit_participant_info` transaction. `env::attached_deposit()` is not visible from the callback receipt, so the value is stashed at submit time and the recipient `AccountId` is the same one used to key the entry (set by `Self::assert_caller_is_signer()`).
```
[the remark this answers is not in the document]
pbeza: https://github.com/near/mpc/commit/c0886f071

crates/contract/src/lib.rs
```diff
@@ -784,9 +778,8 @@ impl MpcContract {
     ///   `verify_quote` call, with [`Self::resolve_verification`] chained as its
     ///   callback to run the post-DCAP checks and store the attestation.
     ///
-    /// The caller must attach a flat 0.1 NEAR fee for the stored entry; the whole
-    /// fee is kept on success and refunded if the attestation is not accepted.
-    #[payable]
+    /// Storage is funded by the contract's own balance, so a node submits with no deposit via its
+    /// function-call access key, for its first attestation and re-attestations alike.
     #[handle_result]
```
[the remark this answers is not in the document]
pbeza: It’s effectively the same as the pre-#3714 version of the contract. IIUC, v3.13.0 charged the measured delta, but `add_participant` never flushed the storage, so the delta was always read as `0` and new `Mock` entries were effectively free. The drain existed there too.

I’m not quite sure what a better funding model for the new attestations would be, given that the nodes’ function-call access keys can’t pay for storage.

crates/contract/tests/sandbox/tee.rs
```diff
@@ -279,6 +279,54 @@ async fn test_submit_participant_info_succeeds_with_mock_attestation() -> Result
     Ok(())
 }
 
+/// Regression test for #3925: a participant re-attesting through a function-call access key (which
+/// cannot attach a deposit) succeeds with zero deposit (storage is contract-funded).
+#[tokio::test]
+async fn submit_participant_info__should_accept_zero_deposit_via_function_call_key() -> Result<()> {
```
gilcu3: nit: I think we can remove the comment, it is very clear what the function aims to achieve

crates/node/src/indexer/types.rs
```diff
@@ -129,12 +128,8 @@ impl ChainSendTransactionRequest {
...
     pub fn deposit_required(&self) -> Balance {
-        match self {
-            Self::SubmitParticipantInfo { .. } => {
-                Balance::from_millinear(SUBMIT_PARTICIPANT_INFO_DEPOSIT_MILLINEAR)
-            }
-            _ => Balance::from_near(0),
-        }
+        // The node signs with a function-call access key, which cannot attach a deposit.
+        Balance::from_near(0)
     }
```
gilcu3: this probably grants a bigger change. The node will never be able to add a deposit

gilcu3 (dismissed):
Thank you!

crates/node/src/indexer/types.rs
pbeza: https://github.com/near/mpc/pull/3940/commits/afacfef96574556a35aea8dd272a7cae101954fd ✅

crates/contract/tests/sandbox/tee.rs
pbeza: https://github.com/near/mpc/pull/3940/commits/afacfef96574556a35aea8dd272a7cae101954fd ✅

gilcu3 (dismissed):

pbeza:
Ready for review @gilcu3 @netrome @kevindeforth. See also the related discussion [here](https://nearone.slack.com/archives/C0912BTG51T/p1784896963691649?thread_ts=1784896517.987259&cid=C0912BTG51T) for context.

crates/contract/src/tee/tee_state.rs
```diff
@@ -228,13 +231,32 @@ impl TeeState {
...
+        let entry = NodeAttestation {
+            node_id,
+            verified_attestation,
+        };
+
+        let before = env::storage_usage();
+        let previous = self.stored_attestations.insert(tls_pk.clone(), entry);
+        self.stored_attestations.flush();
+        let cost =
+            env::storage_byte_cost().saturating_mul(u128::from(env::storage_usage() - before));
```
barakeinav1: **`env::storage_usage() - before` can underflow — the delta is signed.**

`insert` replaces the old value, so if the new entry is smaller than the one it replaces, `storage_usage() < before`.

On an update, three of the four size contributors are pinned: `account_id` (the check just above forces equality), `tls_public_key` (it's the map key), and `account_public_key` (fixed 32 bytes). The `VerifiedAttestation` variant is the one thing that can change size — and the caller picks it freely on every submission. Borsh sizes of the stored value:

| variant | bytes |
|---|---|
| `Mock::Valid` / `Mock::Invalid` | 134 |
| `Mock::WithConstraints{None × 4}` | 138 |
| `Dstack` | 445 |
| `Mock::WithConstraints{all Some}` | 450 |

Reachable shrinks:

- `WithConstraints{None,None,None,None}` → `Valid`: **−4 bytes**, and both verify — all-`None` hits no constraint branch in `MockAttestation::verify_constraints`, and `Valid` returns `Ok(())` unconditionally. Two ordinary transactions from one account, no TEE needed.
- `Dstack` → `Mock::Valid`: **−311 bytes**. Nothing stops an account that attested for real from then submitting a mock.
- `WithConstraints{all Some}` → `Dstack`: **−5 bytes**, and this one lands inside the `resolve_verification` callback. That's the "testnet node running on a mock attestation switches to a real quote" path.

The contract is always built with `release-contract`, which inherits `release` and does not set `overflow-checks` (the `[profile.reproducible]` that does set it is the node binary — `nix/mpc-node.nix:154`). So this wraps rather than panics:

```
storage_usage() - before   =  u64::MAX - 3   ≈ 1.8e19
× storage_byte_cost (1e19)                   ≈ 1.8e38 yocto ≈ 1.8e14 NEAR
```

That's still a valid `u128` (max ≈ 3.4e38), so `saturating_mul` doesn't even clamp — the check just sees a nonsense value, rejects, and restores the previous entry. Net effect: **an account whose entry would shrink can never update it again**, and the error it gets is `stored attestation entry costs 184467440737095.516161 NEAR which exceeds the maximum 0.1 NEAR`, which gives no hint at the real cause.

A well-behaved mainnet node never hits this — it always submits `Dstack`, entries are constant-size, so its steady-state delta is exactly 0. Which is why CI and e2e are green.

```suggestion
        let cost = env::storage_byte_cost()
            .saturating_mul(u128::from(env::storage_usage().saturating_sub(before)));
```

A shrink then reads as cost 0, which is the correct answer — the contract's storage stake went *down*, so there is nothing to cap.

crates/contract/src/lib.rs
```diff
@@ -117,11 +117,9 @@ const MINIMUM_CKD_REQUEST_DEPOSIT: NearToken = NearToken::from_yoctonear(1);
...
 pub const MINIMUM_NODE_MANAGEMENT_DEPOSIT: NearToken = NearToken::from_yoctonear(1);
 
-/// Flat fee a node attaches to [`MpcContract::submit_participant_info`] for its
-/// stored attestation entry. The entry is bounded, so the fee is fixed and
-/// nothing is refunded; its margin over the true cost absorbs storage-price and
-/// layout changes. A unit test asserts it covers the worst-case entry.
-const MINIMUM_ATTESTATION_STORAGE_DEPOSIT: NearToken = NearToken::from_millinear(100);
+/// Caps the contract-funded storage cost of a single stored attestation entry, guarding against a
+/// future schema/attestation-size change bloating an entry. Does not bound total attestation storage.
+pub(crate) const MAX_ATTESTATION_ENTRY_STORAGE_COST: NearToken = NearToken::from_millinear(100);
```
barakeinav1: **This cap is ~16× looser than the thing it guards, and it bounds growth rather than size.**

0.1 NEAR permits a 10,000-byte entry. The real worst case is **604 bytes / 0.006 NEAR** — measured by instrumenting `submit_participant_info__should_bound_worst_case_entry_cost` in this PR. An entry would have to bloat 16× before either the runtime check or that test noticed, so neither really delivers the "guards against a future schema/attestation-size change" claim. The test asserts against this same constant, so it inherited the blind spot from the old `MINIMUM_ATTESTATION_STORAGE_DEPOSIT`.

Separately, on an update the measured delta is `new − old`, so the check constrains the *increment*, not the entry — which is not what this doc comment says. That matters for #3972, whose AC is "the flat deposit is derived from the per-entry storage cap": a deposit can't be sized off a delta-bound.

@gilcu3's point in Slack holds — the type is fixed-size, so a static bound is the natural tool here. Either:

- keep the runtime check, tighten to ~0.02 NEAR (3× headroom), and measure absolute entry cost rather than delta; or
- drop the runtime check and keep only the unit test with a tight bound — no gas cost, and it can't be evaded, since the type is bounded by construction.

One trap either way: **the maximum is a mock entry, not a real one.** `Mock::WithConstraints{all Some}` is 450 bytes vs `Dstack`'s 445. Any tightened bound has to be sized off the mock branch as long as `Attestation::Mock` is ungated in production.

crates/contract/src/tee/tee_state.rs
```diff
@@ -228,13 +231,32 @@ impl TeeState {
...
+                }
+                None => {
+                    self.stored_attestations.remove(&tls_pk);
+                }
+            }
+            self.stored_attestations.flush();
+            return Err(AttestationSubmissionError::EntryStorageCostExceeded {
+                cost,
+                max: MAX_ATTESTATION_ENTRY_STORAGE_COST,
+            });
```
barakeinav1: **Nothing exercises this branch.**

It's the only thing stopping the dstack callback from committing an oversized entry — `resolve_verification` deliberately doesn't panic on error (it defers to `fail_attestation_submission` in a separate receipt precisely so prior state commits), so if this rollback is ever wrong, the bad entry persists. The added test `store_verified_attestation__should_accept_entry_within_cap` only covers the accept path, and with the cap at 0.1 NEAR vs a 604-byte worst case there's no input that can reach here.

Easiest fix is to make the cap injectable rather than a module const — the same shape as `tee_upgrade_deadline_duration`, which is already threaded through `verify_and_store_mock`/`verify_and_store_dstack` as a parameter:

```rust
fn store_verified_attestation(
    &mut self,
    node_id: NodeId,
    verified_attestation: VerifiedAttestation,
    max_entry_cost: NearToken,
) -> Result<ParticipantInsertion, AttestationSubmissionError>
```

Production passes the const; a unit test passes something tiny and asserts both halves that matter: the `EntryStorageCostExceeded` error, and that state is restored intact — the `Some(prev)` arm puts the *original* entry back (not the rejected one), and the `None` arm leaves no entry behind.

Worth doing whichever way the cap-sizing discussion above lands, as long as the runtime check stays.

crates/near-mpc-contract-interface/src/client.rs
```diff
@@ -123,7 +123,8 @@ impl<C: CallContract> MpcContractHandle<C> {
                     method_name: SUBMIT_PARTICIPANT_INFO.to_string(),
                     args,
                     gas: MAX_GAS,
-                    deposit: NearToken::from_millinear(SUBMIT_PARTICIPANT_INFO_DEPOSIT_MILLINEAR),
+                    // Zero: the node's function-call key cannot attach a deposit; the contract only charges new entries.
+                    deposit: NearToken::from_yoctonear(0),
```
barakeinav1: Stale as of this PR — the contract no longer "charges new entries"; it funds them from its own balance. Copilot flagged this one too.

```suggestion
                    // The node's function-call key cannot attach a deposit; attestation storage is
                    // funded by the contract's own balance.
                    deposit: NearToken::from_yoctonear(0),
```

docs/design/attestation-verifier-contract.md
```diff
@@ -139,7 +139,7 @@ Each [`PendingAttestation`](#mpc-contractsubmit_participant_info) holds:
 
 - **The submitter's `Attestation::Dstack` payload** — the RTMR3 event log, app-compose, and report-data the post-DCAP checks consume.
 - **The submitter's TLS public key** — the callback hashes it with the submitter's account public key and compares to the quote's `report_data` field, proving the enclave produced the quote for this specific submitter.
-- **The attached deposit** — covers storage staking on success, refunded to the signer of the original `submit_participant_info` transaction on failure. `env::attached_deposit()` is not visible from the callback receipt, so the value is stashed at submit time and the recipient `AccountId` is the same one used to key the entry (set by `Self::assert_caller_is_signer()`).
+- No deposit is stashed: attestation storage is funded by the contract's own balance, so `submit_participant_info` attaches nothing and there is no deposit to forward to the callback or refund.
```
barakeinav1, line 144: Three things in this file, in increasing order of scope.

**1. This bullet (line 142).** It's a negation dropped into a list of things `PendingAttestation` *holds*, and the only unbolded entry in it. Reads better as prose right after the list than as a list item.

**2. Line 108 is still stale.** It reads "...and the deposit stays locked because the refund is part of the cleanup the contract never got around to." Outside this PR's diff so I can't suggest inline, but it needs the same treatment you gave :142.

**3. Pre-existing, so your call — this whole section describes a design that never shipped.** `PendingAttestation` doesn't exist in the code:

```
$ grep -rn "PendingAttestation\|on_attestation_verified\|pending_attestation" crates/ --include=*.rs
(no matches)
```

`submit_participant_info` passes `VerificationContext` directly into a plain `.then(resolve_verification)` chain — no pending map, no `data_id`, no `promise_yield_create` (the one at `lib.rs:341` is the generic `enqueue_yield_request` helper used by sign/CKD/foreign-tx, not this flow). So lines ~136-146 and the "Handling failures" discussion above describe a superseded mechanism. Copilot noticed this too.

Not introduced by this PR, and I wouldn't ask you to rewrite the section here. But AGENTS.md says a design doc describing a superseded design has to be updated, removed, or prominently marked rather than left silently stale — so the cheap out is a `Status: superseded by #NNNN` banner plus a follow-up issue, rather than editing individual bullets inside a section that no longer matches the code.

crates/contract/src/lib.rs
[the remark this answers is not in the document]
barakeinav1: Agreed this isn't a regression and shouldn't block. One correction to the framing though, since #3972 is being written against it: **this needs no TEE at all.**

`MockAttestation::Valid` returns `Ok(())` unconditionally from `verify_constraints`, `submit_participant_info` has no participant gate (just `assert_caller_is_signer`, which only rules out cross-contract calls), and entries are keyed by `tls_public_key` with no per-account limit — so a single ordinary account can mint unbounded entries with no attestation hardware, for gas only.

They're also permanent: `clean_invalid_attestations` removes only entries that fail re-verification, and `Mock::Valid` re-verifies as valid forever with no expiry field to age out. So the "bounded and self-healing, reclaimed by `clean_invalid_attestations`" argument holds for `Dstack` (which does expire and does get swept) but not for the cheap path.

So the Slack framing of "an arbitrary entity with a single TEE can slowly drain" is understating it. Worth #3972 pricing out the mock path explicitly — and gating `Attestation::Mock` behind a feature flag would remove the cheap variant independently of the deposit work.

barakeinav1 (changes requested):
Direction is right — contract-funded storage is the only model that works with fc keys, and deferring the rework to #3972 makes sense. Requesting changes on one thing.

**Blocker:** underflow in the new entry-size cap — https://github.com/near/mpc/pull/3940#discussion_r3656710534. `storage_usage() - before` is a signed delta read as unsigned, so any submission that *shrinks* the entry wraps and permanently pins that account to its larger entry. Two ordinary txs, no TEE. `release-contract` sets no `overflow-checks`, so it wraps silently rather than panicking.

**More generally, I think a static check is enough here** — and better, given the bug. `NodeAttestation` is bounded by construction (only `account_id` varies, capped at 64B), so nothing can vary at runtime. The runtime check costs gas on every submit, adds an untested rollback path ([r3656731057](https://github.com/near/mpc/pull/3940#discussion_r3656731057)), and at 0.1 NEAR against a measured 604-byte worst case wouldn't catch bloat anyway ([r3656725797](https://github.com/near/mpc/pull/3940#discussion_r3656725797)).

Suggestion: revert e845e904, keep the unit test, tighten it to ~0.02 NEAR. Closes all three in one move.

@kevindeforth your call, since it counters your runtime-check preference. If it stays, it needs all three: underflow fix, tighter constant, rejection-path coverage.

Also, doc-alignment rule: stale comment at `client.rs:126`, stale deposit sentence at `attestation-verifier-contract.md:108`.

barakeinav1 (dismissed):
Approving — the blocker is resolved.

- **Runtime entry-size cap removed in 158e8e4a.** `NodeAttestation` is bounded by construction, so there was no runtime-variable input for it to catch; it only guarded against a future schema change, and it carried an underflow (`storage_usage() - before` reading a signed delta as unsigned). The revert lands byte-identical to afacfef9, the version @gilcu3 approved.
- **Replaced with stricter build-time checks in 74ea6a0e.** The exact stored size of each attestation variant is now pinned — 599 bytes Dstack, 604 bytes Mock — so a layout change fails CI instead of surfacing in production by rejecting live submissions. `WORST_CASE_ENTRY_COST_CEILING` tightened from 0.1 to 0.01 NEAR; the old value was inherited from the flat deposit, where wide margin was wanted, and left 16x slack as a regression ceiling.

Net effect is a strictly tighter guard than the runtime check it replaces, caught earlier and at no gas cost.

Two doc items are still open and I'm happy to push them myself: the stale comment at `client.rs:126` and the stale deposit sentence at `docs/design/attestation-verifier-contract.md:108`. Not blocking.

docs/securing-mpc-with-tee-design-doc.md
```diff
@@ -386,7 +386,7 @@ pub struct Contract {
 }
 ```
 
-_Note_: submit_participant_info - can be called either by the node or by the operator.
+_Note_: submit_participant_info - can be called either by the node or by the operator. Attestation storage is funded by the contract's own balance, so submissions attach no deposit and the node's function-call access key works for both a first-time (join) submission and re-attestations.
```
barakeinav1: our current flow is for node submission only.
IIRC we check the reportdata that contains the hash of public key hash, against who submitted 'submit_participant_info'

docs/design/attestation-verifier-contract.md
barakeinav1, line 144: add follow-up issue here

gilcu3:
@claude review

barakeinav1:
Went through all five. Three fixed, one rejected with reasoning, one moot.

> `docs/running-an-mpc-node-in-tdx-external-guide.md:2074` — the operator runbook still tells first-time joiners to *"call `submit_participant_info` manually with `--deposit` once"*

Fixed in aae164f9 — removed that bullet, which took the stale #903 pointer with it, plus the "the contract's deposit logic" clause at :2070. #903 is now `Closes`-linked here, since the answer to "what does this cost the caller" is "nothing".

> the `TODO(#3825)` banner is scoped to *"This subsection and §Contract state changes below"*, but it sits inside `#### Handling failures` ... Two of ~15 stale deposit/refund references in this file were fixed; the rest read as current design.

Correct, and my mistake — I grepped with too narrow a pattern and only inspected the section I was editing. A plain `deposit|refund` grep finds references at :62, :86, :91, :97, :121, :147, :289, :292 and throughout :418–:613, most of them above or outside the marker.

Fixed in aae164f9 by replacing the subsection TODO with a file-level `**Status:**` line, matching the convention in `node-operator-metrics.md`. It names what is superseded (the `mpc-contract`-side flow, state, and API proposal) and what is not (the verifier split, governance, crate layout, the verifier's own API), and keeps `TODO(#3825)` inside so the CI TODO check still links it.

> dropping `#[payable]` does more than make the deposit optional ... keeping `#[payable]` plus `refund_to(&account_id, env::attached_deposit())` gives the same fc-key fix without breaking those callers

Keeping non-payable, deliberately. Restoring `#[payable]` means restoring a refund path on every error branch — exactly the machinery this PR deletes — to serve a caller that should not exist. A loud `doesn't accept deposit` is better operator feedback than a silent refund that makes `--deposit` look meaningful.

Checked what actually breaks first. Pre-#3714 nodes attached **zero**:

- `crates/node/src/indexer/tx_signer.rs` at `ebc7f0d5^`: `deposit: Balance::from_near(0)`, hardcoded
- `ChainSendTransactionRequest::deposit_required` did not exist
- the node never used `MpcContractHandle` (0 matches in `crates/node/src/`)

So this restores the pre-#3714 node path line-for-line. The 0.1 NEAR in `client.rs` at `ebc7f0d5^` came from #3873 and only reaches sandbox tests, e2e, and `tee-context` — and no crate depends on `tee-context`. After this PR every in-repo caller attaches 0.

> nothing currently pins it — the old below-fee sandbox test was deleted and no test asserts that a deposit-attaching call is rejected

Fair. Added `submit_participant_info__should_reject_an_attached_deposit` in 26c343bd: attaches 1 yoctoNEAR, asserts the transaction fails with `doesn't accept deposit`, and asserts nothing is stored.

> `resolve_verification` also loses `#[payable]`. If the deposit-requiring contract has been deployed anywhere, an upgrade landing between `verify_quote` and its callback will hit the new non-payable `resolve_verification`

Moot on both conditions. No contract carrying #3714 is deployed — mainnet and the testnet cluster in #3925 are both on 3.13.2. And the verifier is not a party to the deposit in any case: `submit_dstack_attestation` attaches `NearToken::from_near(0)` to `verify_quote`, and `crates/tee-verifier/src/` contains no `payable` or `attached_deposit` at all. The deposit only ever travelled from `mpc-contract` to its own callback.

> the tests never call `submit_participant_info`; they insert straight into `tee_state.stored_attestations` ... The doc comment on the first one also says it pins *"every attestation variant"* while only the two worst-case variants are cased.

Both fixed in d8c35796 — renamed to `stored_attestation_entry__should_have_the_pinned_size` and `stored_attestation_entry__should_stay_under_the_cost_ceiling`, and the doc comment now reads "the largest variant of each attestation kind".

crates/contract/src/lib.rs
```diff
@@ -8106,9 +8174,51 @@ mod tests {
...
+    }
+
+    /// Pins the exact stored size of the largest variant of each attestation kind.
+    ///
+    /// Do not update these numbers just to make a failing case pass. The contract funds every
+    /// entry from its own balance, so a size change alters what the contract pays per node, and
+    /// everything derived from it must be revisited in the same change — today
+    /// [`WORST_CASE_ENTRY_BYTES`] and [`WORST_CASE_ENTRY_COST_CEILING`].
+    ///
+    /// TODO(#3972): the flat onboarding deposit will be derived from these sizes too.
```
gilcu3, line 8188: nit: this seems a bit LLM verbose, but I don't mind

docs/design/attestation-verifier-contract.md
```diff
@@ -1,5 +1,7 @@
 # Attestation Verifier Contract Breakout
 
+**Status:** Partially superseded — TODO(#3825). The `mpc-contract` side shipped as a no-yield promise chain, so every mention below of yield-resume, `pending_attestations` / `PendingAttestation`, `data_id`, the "already pending" guard, `on_attestation_verified`, and the attached deposit and its refunds describes a design that was not built. The verifier-contract split, governance, crate layout, and the verifier's own API are accurate.
+
```
gilcu3: what do you mean a no-yield promise chain? Afaik we still have the yields, else how are we handling the cross-contract return call?

docs/securing-mpc-with-tee-design-doc.md
```diff
@@ -386,7 +386,7 @@ pub struct Contract {
 }
 ```
 
-_Note_: submit_participant_info - can be called either by the node or by the operator. Attestation storage is funded by the contract's own balance, so submissions attach no deposit and the node's function-call access key works for both a first-time (join) submission and re-attestations.
+_Note_: submit_participant_info must be called by the node itself - the quote's report_data binds hash(tls_public_key, account_public_key) to env::signer_account_pk(), so a submission signed by any other key fails verification. Attestation storage is funded by the contract's own balance, so submissions attach no deposit and the node's function-call access key works for both a first-time (join) submission and re-attestations.
```
gilcu3, line 389: nice fix :)

gilcu3 (approved):
LGTM

One thing to flag, apparently we no longer need the `fail_attestation_submission` method, or its usage, because we are no longer accepting deposits that would need roll-back. If you agree, feel free to clean that up here, or open an issue for a follow-up (if none exists)

docs/design/attestation-verifier-contract.md
barakeinav1: not sure about this, but this is not strictly related to this PR, and we do have (#3825) by @pbeza . so I just put the todo and didn't dig into this.

barakeinav1 (approved):

PR #3626, opened 2026-06-18
feat(contract): TEE verifier contract-account voting

pbeza:
Closes #3625. Next slice of #3540, stacked on #3590; each slice leaves `main` runnable.

## Why

Moving DCAP verification off the [`mpc-contract`](https://github.com/near/mpc/tree/main/crates/contract) WASM means offloading it to a standalone [`tee-verifier`](https://github.com/near/mpc/tree/main/crates/tee-verifier) contract over a [cross-contract call](https://github.com/near/mpc/blob/main/docs/design/attestation-verifier-contract.md#submission-flow) — which needs a *trusted verifier contract account* chosen by participants, not hardcoded. This PR adds that [governance layer](https://github.com/near/mpc/blob/main/docs/design/attestation-verifier-contract.md#governance-and-upgrades) (the voting that picks the verifier contract account).

## What changed

- Verifier-account voting was added (new [`tee::verifier_votes`](https://github.com/near/mpc/blob/46d04a0302008831de87239efa7e33315da44b38/crates/contract/src/tee/verifier_votes.rs), built on the existing [`Votes`](https://github.com/near/mpc/blob/e46c70d141216c32eb1486e3f6fb68780a7b46d4/crates/contract/src/primitives/votes.rs#L11-L20) primitive):
   - `vote_tee_verifier_change(candidate, expected_code_hash)`
      https://github.com/near/mpc/blob/46d04a0302008831de87239efa7e33315da44b38/crates/contract/src/lib.rs#L1688-L1736

      `expected_code_hash` makes each voter commit to the code they [audited](https://github.com/near/mpc/blob/main/docs/design/attestation-verifier-contract.md#voting-on-the-trusted-verifier-in-mpc-contract) — same account, different hash ⇒ different buckets, so neither reaches threshold alone. Crossing threshold updates `tee_verifier_account_id` and clears pending votes; stale votes are swept after resharing.

   - `withdraw_tee_verifier_vote()`
      https://github.com/near/mpc/blob/46d04a0302008831de87239efa7e33315da44b38/crates/contract/src/lib.rs#L1738-L1756

   - `MpcContract` gains `tee_verifier_account_id` and `tee_verifier_votes`.
      https://github.com/near/mpc/blob/46d04a0302008831de87239efa7e33315da44b38/crates/contract/src/lib.rs#L185-L190

[...]

PR #3948, opened 2026-07-23
fix(attestation): revert DEFAULT_EXPIRATION_DURATION_SECONDS to 7 days

### LINKED ISSUE #3947: fix(attestation): revert DEFAULT_EXPIRATION_DURATION_SECONDS from 1 day back to 7 days
### Background

`DEFAULT_EXPIRATION_DURATION_SECONDS` (`crates/mpc-attestation/src/attestation.rs:28`) was reduced from **7 days to 1 day** in #3626. The contract stamps each accepted attestation with `expiry_timestamp_seconds = block_timestamp + DEFAULT_EXPIRATION_DURATION_SECONDS` (`TeeState::add_participant` → `AcceptedAttestation::dstack`), so this constant sets how long a node's on-chain attestation stays trusted before it must re-attest.

A 1-day window leaves almost no margin: if a node can't refresh its attestation for ~24h it gets dropped. This just bit us during an Intel PCS incident — the PCK CRL went stale (Intel didn't re-sign it for 7+ days) and the node-side collateral freshness check began rejecting fresh quote generation. With a 1-day expiry, nodes are ~24h from being kicked; with the previous 7-day expiry they'd have had a week of runway to ride out the upstream issue. 7 days is also the value the node's hourly re-attestation cadence was designed around.

### What

Revert the constant to 7 days and fix the now-stale doc comment.

### Acceptance Criteria

- [ ] `DEFAULT_EXPIRATION_DURATION_SECONDS = 60 * 60 * 24 * 7` (7 days)
- [ ] Doc comment updated to describe the 7-day window (currently says "1 day")
- [ ] Note in PR: takes effect for **new** submissions after contract redeploy; it does **not** retroactively extend attestations already stored on-chain

### Safety — no node/contract version-skew hazard

Made safe by #3736 (`fix(node): confirm submit_participant_info via advancing attestation expiry`, already on `main`). The node used to confirm its submission by reconstructing the attestation's creation time as `stored_expiry − DEFAULT_EXPIRATION_DURATION_SECONDS`; under skew (contract stamping 7d while a node still assumes 1d) the reconstructed age (~6d) exceeds the 120s freshness bound, so every submission is treated as `NotExecuted` and retried indefinitely. #3736 replaced that with an *expiry-advanced* check that doesn't depend on the constant's value, so bumping the contract expiry 1d→7d cannot trigger that retry loop. (Holds provided nodes are on the #3736 build.)

- Reduced to 1 day in #3626 (`feat(contract): TEE verifier contract-account voting`)
- Node-side confirmation is independent of this constant since #3736 (expiry-advanced check), so contract and node no longer need to agree on its value — the only remaining node reference is a doc comment.
- The node-side collateral freshness fix (relaxing `MAX_COLLATERAL_AGE` / per-field PCK CRL handling) is tracked separately — those are the node changes; this issue is only the expiry constant.

barakeinav1:
## What

Reverts `DEFAULT_EXPIRATION_DURATION_SECONDS` (`crates/mpc-attestation/src/attestation.rs`) from **1 day back to 7 days**, and fixes the stale `// 1 day` inline comment.

## Why

The contract stamps each accepted attestation with `expiry = block_timestamp + DEFAULT_EXPIRATION_DURATION_SECONDS` (`TeeState::add_participant`). The value was dropped from 7d to 1d in #3626, which leaves almost no margin: if a node can't refresh its attestation for ~24h it gets dropped. This bit us during an Intel PCS incident where the PCK CRL went stale (>7 days without re-sign) and node-side collateral freshness rejected fresh quote generation — with a 1-day expiry, nodes are ~24h from being kicked; the previous 7-day window would have given a week of runway. Nodes re-attest hourly, well within 7 days.

## Safety — no node/contract version-skew hazard

Made safe by #3736 (`fix(node): confirm submit_participant_info via advancing attestation expiry`, already on `main`). The node used to confirm its submission by reconstructing the attestation's creation time as `stored_expiry − DEFAULT_EXPIRATION_DURATION_SECONDS`; under skew (contract stamping 7d while a node still assumes 1d) the reconstructed age (~6d) exceeds the 120s freshness bound, so every submission is treated as `NotExecuted` and retried indefinitely. #3736 replaced that with an *expiry-advanced* check that doesn't depend on the constant's value, so this 1d→7d change cannot trigger that retry loop. (Holds provided nodes are on the #3736 build.)

## Notes

- Takes effect for **new** submissions after redeploy; does **not** retroactively extend attestations already stored on-chain.
- Node-side confirmation is independent of this constant since #3736 (it now checks the expiry *advanced* rather than reconstructing from the value), so contract and node no longer need to agree on it. The only remaining reference in the node is a doc comment.
- The node-side collateral freshness fix (relaxing `MAX_COLLATERAL_AGE` / per-field PCK CRL handling) is a separate PR.

Closes #3947

### DIFF
--- a/crates/mpc-attestation/src/attestation.rs
+++ b/crates/mpc-attestation/src/attestation.rs
@@ -25,7 +25,7 @@ use crate::alloc::string::{String, ToString};
 /// re-verified via [`VerifiedAttestation::re_verify`]. Nodes resubmit hourly,
 /// well within this window, so valid attestations refresh in time.
 // TODO(#1639): extract timestamp from certificate itself
-pub const DEFAULT_EXPIRATION_DURATION_SECONDS: u64 = 60 * 60 * 24; // 1 day
+pub const DEFAULT_EXPIRATION_DURATION_SECONDS: u64 = 60 * 60 * 24 * 7; // 7 days
 
 // `large_enum_variant` fires only where `usize` is 64-bit; under the contract's
 // wasm32 build the variants are close enough in size that it doesn't, so gate the


gilcu3 (approved):
LGTM

netrome (approved):

PR #3956, opened 2026-07-24
chore: use trait to avoid free-form strings in tracking

### LINKED ISSUE #2516: Use a trait with associated const for protocol names in run_protocol
### Background

From [PR #2495 review comment](https://github.com/near/mpc/pull/2495#discussion_r2958722491) by @netrome — `run_protocol` takes a free-form `name: &'static str` at each call site (~14 in `crates/node/src/providers/`). A trait with an associated const would make this consistent and compile-time checked.

### User Story

As a developer, I want protocol names defined as associated constants on a trait so naming is consistent and discoverable.

### Acceptance Criteria

- New trait (e.g. `NamedProtocol`) with `const NAME` and a default `run` method wrapping `run_protocol`
- All call sites in `crates/node/src/providers/` migrated to use it

gilcu3:
Closes #2516

### DIFF
--- a/crates/node/src/protocol.rs
+++ b/crates/node/src/protocol.rs
@@ -3,9 +3,23 @@ use crate::tracking;
 use crate::{network::NetworkTaskChannel, tracking::TaskHandle};
 use futures::TryFutureExt;
 use std::collections::{BTreeMap, HashMap};
+use std::future::Future;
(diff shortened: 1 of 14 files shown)


netrome (approved):
Nice, thanks!

haiyuechen-nearone (approved):

PR #3958, opened 2026-07-24
perf(tests): improve newly introduced test performance by reducing timout

### LINKED ISSUE #3957: Speed up reconstruction_threshold node tests by lowering the shared sign-wait budget
### Background

The two `reconstruction_thresholds` tests are the slowest in the node suite and gate the whole "Cargo test: node" step wall-clock (~298s):

```
PASS [ 297.564s] tests::reconstruction_thresholds::per_domain_reconstruction_threshold__should_gate_signing_availability_when_nodes_go_offline
PASS [ 239.086s] tests::reconstruction_thresholds::resharing__should_apply_updated_thresholds_while_preserving_unchanged_ones
Summary [ 298.181s] 318 tests run
```

Both run on `Clock::real()`. The dominant cost is `assert_cannot_sign`: it returns only when `request_and_await_response` hits its `timeout` (`crates/node/src/tests.rs:357`), currently `DEFAULT_MAX_SIGNATURE_WAIT_TIME = 60s`. Since the online set is below threshold no response can ever arrive, so each negative assertion blocks the full 60s — 4 such calls in the first test (~240s) and 3 in the second (~180s), which accounts for nearly all the runtime.

The lever is to lower this timeout, but `assert_can_sign` and `assert_cannot_sign` must share the same budget: the negative window has to be long enough that a domain which *could* sign already would have, or `assert_cannot_sign` could pass merely because signing was slow. Positive signs complete in a few seconds, so a shared 10s budget (vs the current 60s) should be safe and cut ~50s off each of the 7 negative waits.

### Acceptance Criteria

- `assert_can_sign` and `assert_cannot_sign` share a single 10s sign-wait budget.
- The two `reconstruction_thresholds` tests are significantly faster and remain non-flaky.

gilcu3:
As explained in the issue, the new reconstruction threshold tests became the bottleneck of the CI. The reason was the use of the full timeout (60 seconds) for the negative case, when a much smaller one should suffice. A better solution would be to make the negative test time independent, but that would be a much bigger endeavor that we probably don't need for now.

Here due to assets taking a long time to generate the solution is to have a higher timeout for a warm-up signature, then use the much lower budget for the rest. Hopefully that keeps CI fast and does not produce flakiness

After this PR these tests are no longer the bottleneck, going from 240s+ to ~100s. Re-ran a few times to ensure it is not flaky

### DIFF
--- a/crates/node/src/tests/reconstruction_thresholds.rs
+++ b/crates/node/src/tests/reconstruction_thresholds.rs
@@ -19,7 +19,7 @@ use std::collections::BTreeMap;
 
 /// Shared budget for both assertions: the negative window must exceed the worst-case positive
 /// latency, else [`assert_cannot_sign`] could pass merely because a capable domain was slow.
-/// [`warm_up`] keeps that latency low, so 10s is safe.
+/// [`warm_up`] keeps that latency low.
 const REQUEST_WAIT_BUDGET: std::time::Duration = std::time::Duration::from_secs(10);
 
 /// Generous budget for [`warm_up`], absorbing the one-time cold-start after each online-set change.
@@ -40,8 +40,8 @@ async fn request_and_await_response(
     }
 }
 
-/// Primes each domain's presignatures for the current online set with an untimed sign, whose
-/// cold-start can exceed [`REQUEST_WAIT_BUDGET`]. Only CaitSith and DamgardEtAl consume
+/// Primes each domain's presignatures for the current online set with a generously-budgeted sign,
+/// whose cold-start can exceed [`REQUEST_WAIT_BUDGET`]. Only CaitSith and DamgardEtAl consume
 /// pre-generated presignatures; Frost and CKD sign directly, so pass only the block's signable
 /// CaitSith/DamgardEtAl domains after every online-set change.
 async fn warm_up(indexer: &mut FakeIndexerManager, domains: &[&DomainConfig]) {
--- a/crates/node/src/tests/reconstruction_thresholds.rs
+++ b/crates/node/src/tests/reconstruction_thresholds.rs
@@ -8,8 +8,8 @@ use crate::indexer::participants::ContractState;
 use crate::p2p::testing::port_seed;
 use crate::tests::common::{ckd_domain, sign_domain};
 use crate::tests::{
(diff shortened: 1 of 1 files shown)


gilcu3:
@claude review

Responding to your comments above. We need to keep the positive and the negative tests in this file with the same budget, else we risk the negative tests passing because CI is too slow. 10 seconds should be well above what is needed here anyway, the global constant of 60 seconds is only used because that is the value in production, with many more nodes.

crates/node/src/tests/reconstruction_thresholds.rs
```diff
@@ -8,41 +8,51 @@ use crate::indexer::participants::ContractState;
...
 };
 use crate::tracking::AutoAbortTask;
 use mpc_primitives::domain::Curve;
 use near_mpc_contract_interface::types::{DomainConfig, Protocol, ReconstructionThreshold};
 use near_time::Clock;
 use std::collections::BTreeMap;
 
+/// Shared budget for both assertions: the negative window must exceed the worst-case positive
+/// latency, else [`assert_cannot_sign`] could pass merely because a capable domain was slow.
+/// [`warm_up`] keeps that latency low, so 10s is safe.
```
pbeza: I’d avoid putting specific numbers in the comment, since it’s easy to forget to update them when the constant changes.
 
```suggestion
/// [`warm_up`] keeps that latency low.
```

pbeza (dismissed):

crates/node/src/tests/reconstruction_thresholds.rs
```diff
@@ -8,41 +8,51 @@ use crate::indexer::participants::ContractState;
...
+const REQUEST_WAIT_BUDGET: std::time::Duration = std::time::Duration::from_secs(10);
```
haiyuechen-nearone, line 23: nit: my personal preference is to keep the actual numbers as `const,` so other places can do arithmetic on top again, if needed.

might not be applicable here tho

haiyuechen-nearone (dismissed):

crates/node/src/tests/reconstruction_thresholds.rs
gilcu3: I think in rust it is customary to have strong types, even for constants. I used to prefer actual numbers before as well.

pbeza (approved):

haiyuechen-nearone (approved):
🚢

PR #3963, opened 2026-07-24
feat(health-check): probe Bitcoin by genesis hash and a recent transaction

haiyuechen-nearone:
Part of #3764; rationale: #3969. Stacked on #3962, merge bottom-up:

1. #3932 — Starknet: chain id + a recent L1-accepted tx
2. #3962 — 6 EVM chains: `eth_chainId` + a recent finalized tx
3. **#3963 (this PR)** — Bitcoin: genesis hash + a recent tx
4. #3964 — Aptos: ledger chain id + a recent tx

Replaces Bitcoin's pinned golden tx with the identity probe: verify the genesis block hash (`getblockhash 0`, never pruned), then inspect the coinbase of a block 10 below the tip. Tip height comes from `getbestblockhash` + `getblock` — some edges serve `getblockcount` as a JSON-RPC 1.0 response the 2.0 transport rejects.

Expected genesis comes from config, no built-ins.

PR #3962, opened 2026-07-24
feat(health-check): probe EVM chains by chain id and a recent finalized transaction

haiyuechen-nearone:
Part of #3764; rationale: #3969. Stacked on #3932, merge bottom-up:

1. #3932 — Starknet: chain id + a recent L1-accepted tx
2. **#3962 (this PR)** — 6 EVM chains: `eth_chainId` + a recent finalized tx
3. #3963 — Bitcoin: genesis hash + a recent tx
4. #3964 — Aptos: ledger chain id + a recent tx

Replaces the six EVM chains' pinned golden txs with the identity probe: verify `eth_chainId`, then inspect a real tx from a block 10 below the finalized head (walks back past empty blocks).

Expected chain ids come from config (`foreign_chain_health_check.identities`), no built-ins — well-known values are in the config-tester README.

PR #3964, opened 2026-07-24
feat(health-check): probe Aptos by ledger chain id and a recent transaction

### LINKED ISSUE #3764: Foreign chain RPC provider health checks: validate config without curated reference data

haiyuechen-nearone:
Part of #3764; rationale: #3969. Stacked on #3963, merge bottom-up:

1. #3932 — Starknet: chain id + a recent L1-accepted tx
2. #3962 — 6 EVM chains: `eth_chainId` + a recent finalized tx
3. #3963 — Bitcoin: genesis hash + a recent tx
4. **#3964 (this PR)** — Aptos: ledger chain id + a recent tx

Replaces Aptos's pinned golden tx with the identity probe: verify the ledger `chain_id` (`GET /v1`), then inspect a committed tx 100 versions behind the tip.

The last golden chain: with every chain now identity-based, this deletes the `Network` enum, the `--network` flag, and network auto-detection (all only picked between built-in reference sets), and turns the golden module into a small `parse` module of identity decoders.

The node-startup stack (#3848 → #3849 → #3850) builds on this and closes #3764.


haiyuechen-nearone:
Superseded by the #4003 stack, which reimplements chain-identity probing config-first: #4017 (config field) and #4013 (probe machinery, Starknet-first). Reopen if the per-chain hybrid approach is revived.

PR #3967, opened 2026-07-24
feat: add backup-cli service and node side metrics

### LINKED ISSUE #3864: Automatic backups
### Background

Currently we have the `backup-cli` which allow node operators to back up their keyshares of running nodes. Long term we want to turn this in to a fully fledged TEE attested service. However, that is quite an extensive effort and in the meantime we cannot rely on node operators remembering to perform manual backups after every resharing.

Therefore, we should extend the backup CLI to allow starting a running process that automatically performs backups after any resharing has concluded.

### User Story

As an MPC node operator I'd like to have an automatic service that captures keyshare backups when the contract goes through resharing.

### Acceptance Criteria

We provide a mechanism to support automatic backups.

We had a brainstorming session today, and there are some ideas floating. The most obvious solution would be to take the existing backup CLI, extend it into a service that poll a third-party RPC provider for contract changes, and performs backups on every contract change.

However, variants where the nodes are in charge of pushing backups could also be viable. We also considered having the backup CLI fetch the contract state directly from the MPC nodes themselves. All of these would suffice to fill our immediate needs.

gilcu3:
Closes #3864

WIP

### DIFF
--- a/crates/backup-cli/src/adapters/contract_state_rpc.rs
+++ b/crates/backup-cli/src/adapters/contract_state_rpc.rs
@@ -1,39 +1,24 @@
-use anyhow::Context as _;
 use near_account_id::AccountId;
-use near_jsonrpc_client::errors::{JsonRpcError, JsonRpcServerError};
-use near_jsonrpc_client::methods::query::{RpcQueryError, RpcQueryRequest};
-use near_jsonrpc_client::{JsonRpcClient, auth};
-use near_jsonrpc_primitives::types::query::{QueryResponseKind, RpcQueryResponse};
+use near_kit::Near;
 use near_mpc_contract_interface::method_names;
 use near_mpc_contract_interface::types::ProtocolContractState;
-use near_primitives::types::{BlockReference, Finality, FunctionArgs};
-use near_primitives::views::QueryRequest;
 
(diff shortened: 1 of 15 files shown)


crates/backup-cli/src/adapters/contract_state_rpc.rs
```diff
@@ -0,0 +1,34 @@
...
+
+impl ContractStateReader for RpcContractStateReader {
+    type Error = anyhow::Error;
+
+    async fn get_contract_state(&self) -> Result<ProtocolContractState, Self::Error> {
+        self.client
+            .view::<ProtocolContractState>(&self.contract_id, method_names::STATE)
+            .await
+            .map_err(|err| anyhow::anyhow!("state view call failed: {err}"))
+    }
```
kevindeforth, line 33: NearKit is a back-end compatible with our `ViewContract` trait, which would give it access to our MPC contract for free, c.f. https://github.com/near/mpc/blob/main/crates/near-mpc-contract-interface/src/client.rs

(haven't gotten to implement the `ViewContract` traits yet, but the `state` method will be one of them.)

gilcu3:
closed in favor of #4025

Issue #3909, opened 2026-07-22
Stale node (offline ~days) wedges in neard epoch-sync loop; never reaches head


barakeinav1:
## Background

A node that had been offline for a while, on restart, gets stuck in neard's epoch-sync and never reaches head. Observed on a testnet TEE cluster while trying to wake previously-stopped nodes.

**Observed (facts):**
- Image: `nearone/mpc-node:release-v3.13-389dd4e` (3.13.2, nearcore 2.13.1).
- On start the node resumed from existing data: `mpc_node::config::start: NEAR node already initialized, skipping init`.
- `state_sync = {"sync":"Peers"}` (via `/debug/nearcore_config`); `near_sync_status = 2` (EpochSync).
- `near_block_height_head` frozen at `260267913` for ~1h (14:23→15:24 UTC), ~163k blocks (~2 days) behind tip; not advancing across repeated samples.
- The log cycles indefinitely (~every 60s), rotating source peers:
  ```
  WARN sync: epoch sync from peer timed out, retrying source_peer_id=…
  INFO sync: bootstrapping node via epoch sync peer_id=…
  INFO sync: stale node validated epoch sync proof, requesting data reset
  ```
- All 3 nodes (`51.68.219.1/.2/.3`) behave identically.

**Explicitly NOT #3713** (`fix: indexer wedging on startup`): that fixes the mpc-node *indexer* pinning at `genesis_height`, is main-only (not in this image), and is a different layer — here neard never reaches head, so the indexer never runs.

## Theories (unverified — not fact)

- **Root cause unknown.** The node repeatedly *validates an epoch-sync proof then requests a data reset* without ever committing forward progress. Suspicion (unverified): epoch-sync doesn't converge for a node this stale, or the validate→data-reset path loops. Needs someone familiar with nearcore epoch-sync to confirm.
- **Layer/attribution (theory):** the loop is entirely in nearcore's `sync:` module, so this is likely nearcore epoch-sync behavior surfaced via the node's neard config — not confirmed to be an mpc-node bug.
- **Suspected workaround (not yet confirmed on this cluster):** wiping near-data to force a fresh header+state sync. Based on prior similar cases; will confirm and update.

## Acceptance Criteria

- [ ] Determine why a stale node loops in epoch-sync instead of recovering (root cause, or escalate to nearcore).
- [ ] A supported recovery path for a node offline for days that doesn't require manually wiping data — or document the wipe as the intended path.

## Resources & Additional Notes

[...]

PR #3975, opened 2026-07-25
fix(chain-gateway): recover from stale node epoch-sync wedge

metalurgical:
This PR is related to #3909 and implements the recovery mechanism for the `EpochSyncDataReset` shutdown reason in chain-gateway that is otherwise only present in the neard CLI.

It swaps `start_with_config()` for `start_with_config_and_synchronization()`.. Then replicates the data reset marker on shutdown and the marker processing on startup. The node is restarted in-process without affecting the rest of the system when the above reason is received. It is ensured that chain-gateway still functions transparently across the restart and reconnects to the node automatically.

This PR requires testing to confirm that the wedging problem is solved and the code doesn't need further adjustments.


metalurgical:
@barakeinav1 Could you assist with testing here? Feedback is appreciated.

metalurgical:
Superceded by #4346

Issue #3648, opened 2026-06-23
Implement script for verifying borsh encoded arguments


anodar:
### Background

We need to sometimes call functions that only accept borsch encoding. It's not very straightforward to verify those arguments and we need to come up with a way for it.

E.g. 

### Acceptance Criteria

We have a straightforward way to verify borsh encoded args given the type that was used to serialize it.

PR #3993, opened 2026-07-28
feat: implement borsch-args-cli for deserializing and decoding contract arguments

metalurgical:
Currently processes:
- `propose-update-args`
- `verify-quote-args`
- `foreign-chain-provider-votes`

Related to #3648, after looking at #3646


metalurgical:
@anodar What are your thoughts on this?

PR #3995, opened 2026-07-29
fix: make online assets metric more informative

### LINKED ISSUE #2123: nodes reporting 0 online presignatures while leading signatures
### Background

Currently a couple of nodes report 0 presignatures online:

```
# HELP mpc_owned_num_presignatures_available Number of presignatures generated that we own, and not yet used
# TYPE mpc_owned_num_presignatures_available gauge
mpc_owned_num_presignatures_available 10801
# HELP mpc_owned_num_presignatures_online Number of presignatures generated that we own, and not yet used,\n                for which the participant set is confirmed alive
# TYPE mpc_owned_num_presignatures_online gauge
mpc_owned_num_presignatures_online 0
```

```
# HELP mpc_owned_num_presignatures_available Number of presignatures generated that we own, and not yet used
# TYPE mpc_owned_num_presignatures_available gauge
mpc_owned_num_presignatures_available 8389
# HELP mpc_owned_num_presignatures_online Number of presignatures generated that we own, and not yet used,\n                for which the participant set is confirmed alive
# TYPE mpc_owned_num_presignatures_online gauge
mpc_owned_num_presignatures_online 0
```

These same nodes were participating in signature generation as leaders just fine. Therefore we believe the metric itself might be computed incorrectly. Surprisingly, both nodes also report a total number of presignatures above the required value 8192, therefore there seems to be a correlation between these facts.

### User Story

_No response_

### Acceptance Criteria

The nodes report the correct amount of online presignatures

_No response_

gilcu3:
Closes #2123

### DIFF
--- a/crates/node/src/providers/ecdsa/presign.rs
+++ b/crates/node/src/providers/ecdsa/presign.rs
@@ -138,7 +138,7 @@ impl EcdsaSignatureProvider {
             }
 
             // If the store is full, try to discard some presignatures which cannot be used right now
-            if my_presignatures_count == config.desired_presignatures_to_buffer {
+            if my_presignatures_count >= config.desired_presignatures_to_buffer {
                 presignature_store.maybe_discard_owned(1).await;
             }
 
--- a/crates/node/src/providers/ecdsa/triple.rs
+++ b/crates/node/src/providers/ecdsa/triple.rs
@@ -220,7 +220,7 @@ impl EcdsaSignatureProvider {
             }
 
             // If the store is full, try to discard some triples which cannot be used right now
-            if my_triples_count == config.desired_triples_to_buffer {
+            if my_triples_count >= config.desired_triples_to_buffer {
                 triple_store.maybe_discard_owned(32).await;
             }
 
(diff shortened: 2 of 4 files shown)


gilcu3:
> * `crates/node/src/providers/ecdsa/presign.rs:141`, `crates/node/src/providers/ecdsa/triple.rs:223`, `crates/node/src/providers/robust_ecdsa/presign.rs:146` — **No test accompanies the behaviour change.** `docs/engineering-standards.md:185` is explicit: "Any change to our system should come with tests such that if the change was reverted, these tests would fail." Reverting all three edits to `==` breaks nothing in the suite today. The guard is also copy-pasted three times, so nothing prevents one site drifting back. Suggest hoisting it onto the store so it is both de-duplicated and unit-testable:

Answering the blocking point: I would rather not change the store behavior for this, and for this reason the change is not really unit testable right now.

netrome (approved):

crates/node/src/providers/ecdsa/presign.rs
haiyuechen-nearone, line 145: QQ: not related to the change, but why do we sleep here?

haiyuechen-nearone (approved):
LGTM, left a question for my own understanding

crates/node/src/providers/ecdsa/presign.rs
gilcu3: Notice this is an infinite generation loop, without a sleep right now it would become a hot loop

PR #4000, opened 2026-07-29
refactor: governance dto compat

SimonRastikian:
closes partly #3903.
This PR introduces the governance threshold terminology in the smart contract. This implied changing the smart contract which could affect migration. Thus we introduced Compat structs


SimonRastikian:
The TODO(XXXX) will be added once the PR is validated

PR #4002, opened 2026-07-29
ci: fix nightly build sed branch OR tags

andrei-near:
because Cargo.toml uses tag instead of branch, this CI does not properly replace the branch on nightly runs


gilcu3 (approved):
LGTM

barakeinav1 (approved):

PR #4008, opened 2026-07-29
feat: Add optional payload hash to foreign transaction requests

### LINKED ISSUE #3997: [Audit] `respond_verify_foreign_tx` doesn't check that `payload_hash` matches the response
### Background

Finding from ToB audit. Currently, the `respond_verify_foreign_tx` method only verifies that it receives a signed response, but it doesn't bind the response to the request in any way. This allows old responses to be submitted to the contract to DoS foreign tx validation requests.

### User Story

As an MPC user I'd like to get responses for the requests I've sent.

### Acceptance Criteria

The MPC contract can ensure that responses submitted `respond_verify_foreign_tx` actually matches the request.

Note that callers are expected to validate received payloads and the signed payload hash with their expected values, so this should not pose a risk of invalid transfers for callers.

We've discussed alternatives here, such as allowing an optional payload hash to be submitted alongside the request. We just need to be careful so that users can't drain needless resources from the network by submitting bogus payload hashes.

netrome:
closes #3997


crates/contract/src/lib.rs
```diff
@@ -730,6 +730,12 @@ impl MpcContract {
             return Err(TeeError::TeeValidationFailed.into());
         }
 
+        if let Some(expected_payload_hash) = &request.expected_payload_hash
+            && &response.payload_hash != expected_payload_hash
+        {
+            return Err(RespondError::UnexpectedPayloadHash.into());
+        }
+
```
netrome: Note that this effectively means a request with an invalid payload hash gets timed out with no informative response. This is a bit unfortunate, but aligned with how we're currently not responding if values are incorrect. Longer term we should introduce proper signed responses to return error messages but it's too complicated to shoehorn into this PR imo.

crates/contract/src/v3_13_0_state.rs
```diff
@@ -113,7 +113,12 @@ impl From<MpcContract> for crate::MpcContract {
             protocol_state: old.protocol_state,
             pending_signature_requests: old.pending_signature_requests,
             pending_ckd_requests: old.pending_ckd_requests,
-            pending_verify_foreign_tx_requests: old.pending_verify_foreign_tx_requests,
+            // `VerifyForeignTransactionRequest` gained `expected_payload_hash`, changing the
+            // borsh key encoding, so entries pending at upgrade time are abandoned; their
+            // yields time out on chain as if never responded to.
+            pending_verify_foreign_tx_requests: LookupMap::new(
+                crate::storage_keys::StorageKey::PendingVerifyForeignTxRequestsV3,
+            ),
```
netrome: Note that this means we may have a few request failures during the upgrade, which I think is acceptable although a bit unfortunate. Migrating requests would becomplex since we cannot easily iterate over requests since they are stored in a LookupMap.

crates/near-mpc-sdk/src/foreign_chain.rs
```diff
@@ -114,12 +114,14 @@ impl<Request: Into<ForeignChainRpcRequestWithExpectations>>
 impl<Request: Into<ForeignChainRpcRequestWithExpectations>>
     ForeignChainRequestBuilder<Request, DomainId>
 {
+    /// Errors if borsh-serializing the expected payload for hashing fails; this is not
+    /// expected to happen for any current payload type.
     pub fn build(
         self,
-    ) -> (
+    ) -> std::io::Result<(
```
netrome: The current msg hash computation currently should not be able to fail, so we could use an .expect() there instead of adding the result here. However, proper error handling felt better here since I cannot guarantee that this property holds for future borsh crate versons etc.

netrome (commented):
Sharing some of my thoughts. I'm happy with the current code. This is ready for LLM review now.

crates/contract/src/v3_13_0_state.rs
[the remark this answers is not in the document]
netrome: Yes, I'm aware of this and chose to accept that we'll lose some in-flight requests. We've done similar upgrades in the past. It's not ideal, but not a catastrophe either.

PR #4009, opened 2026-07-30
ci: harden claude reviews

gilcu3:
Closes https://github.com/near/mpc-private/issues/506

### DIFF
--- a/.github/workflows/claude-pr-review.yml
+++ b/.github/workflows/claude-pr-review.yml
@@ -8,11 +8,14 @@ on:
 
 jobs:
   claude-review:
-    # Run if: (PR opened/ready AND not draft) OR @claude review in PR comment
+    # Run if: (PR opened/ready AND not draft) OR (@claude review in PR comment AND
+    # commenter is an owner/member, so non-trusted commenters can't trigger the
+    # privileged job)
     if: |
       (github.event_name == 'pull_request' && !github.event.pull_request.draft) ||
       (github.event_name == 'issue_comment' &&
        github.event.issue.pull_request &&
+       contains(fromJSON('["OWNER", "MEMBER"]'), github.event.comment.author_association) &&
        (contains(github.event.comment.body, '@claude review') ||
         contains(github.event.comment.body, '@claude code review')))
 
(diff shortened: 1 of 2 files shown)


.github/workflows/claude-pr-review.yml
```diff
@@ -8,11 +8,14 @@ on:
...
   claude-review:
-    # Run if: (PR opened/ready AND not draft) OR @claude review in PR comment
+    # Run if: (PR opened/ready AND not draft) OR (@claude review in PR comment AND
+    # commenter is an owner/member, so non-trusted commenters can't trigger the
+    # privileged job)
     if: |
       (github.event_name == 'pull_request' && !github.event.pull_request.draft) ||
       (github.event_name == 'issue_comment' &&
        github.event.issue.pull_request &&
+       contains(fromJSON('["OWNER", "MEMBER"]'), github.event.comment.author_association) &&
```
netrome, line 18: Just sanity checking, are these all variants? I would half expect "ADMIN" to be an alternative here as well.

netrome (approved):
Thanks!

netrome:
@claude review if the author_association check makes sense

gilcu3:
> github.event.comment.author_association

I am actually surprised this ran, I thought we could not run reviews when the workflow itself was modified :| but maybe it used the original workflow itself

About the missing `COLLABORATOR` I think we don't have those, but I don't mind adding it (I had removed it to make it more strict)

.github/workflows/claude-pr-review.yml
gilcu3: this was not easy to find at all:

https://docs.github.com/en/graphql/reference/issues#enum-commentauthorassociation

haiyuechen-nearone: I guess ADMIN has to be a MEMBER

haiyuechen-nearone (approved):
🦺

haiyuechen-nearone:
> About the missing COLLABORATOR I think we don't have those, but I don't mind adding it (I had removed it to make it more strict)

I think its good to NOT allow COLLABORATORS, since we need 2 members to approve PR to be merged anyways

.github/workflows/claude-pr-review.yml
gilcu3: ADMIN is not an allowed value there :upside_down_face:

Issue #3991, opened 2026-07-28
refactor(contract): drop fail_attestation_submission now that the deposit is gone


barakeinav1:
### Background

`resolve_verification` reports a failed Dstack submission by scheduling a separate `fail_attestation_submission` receipt whose panic fails the submitter's transaction, instead of panicking directly. That indirection existed so the refund and any partial store committed before the failure — a panic in the same receipt would have rolled them back.

#3940 removes the deposit, and with the runtime entry-size cap reverted `store_verified_attestation` writes nothing on any error path, so no state needs to survive the `Err` branch any more. `resolve_verification` can panic directly, and the method plus its gas knob can go.

### Acceptance Criteria

- [ ] `resolve_verification` fails the submitter's transaction directly; the submitter still sees the underlying error (`QuoteRejected`, `VerifierUnavailable`, post-DCAP failure).
- [ ] `fail_attestation_submission` and `method_names::FAIL_ATTESTATION_SUBMISSION` are removed.
- [ ] `fail_attestation_submission_tera_gas` is removed from `Config` (`crates/contract/src/config.rs:67`) and its DTO plumbing (`crates/contract/src/dto_mapping.rs`, `crates/near-mpc-contract-interface/src/types/config.rs`), with state migration handled the way `crates/contract/src/v3_13_0_state.rs` handles prior `Config` changes.
- [ ] ABI and Borsh-schema snapshots regenerated.
- [ ] Sandbox coverage that a rejected quote and an unreachable verifier still fail the submitter's transaction.

### Resources & Additional Notes

- Raised by @gilcu3 reviewing #3940: https://github.com/near/mpc/pull/3940#pullrequestreview-4796451525
- Predecessor: #3940 (removes the deposit, which is what makes this dead weight).
- **Sequence after #3972.** #3972 reintroduces payment for attestation storage. If the charge lands inside the callback it revives the same "this must commit even though we're failing" problem this method was built for, so removing it before that design settles risks rebuilding it. Decide #3972's shape first.
- Removing a `Config` field is a state-schema change, not a cleanup — that is why this was not folded into #3940.

PR #4011, opened 2026-07-30
docs(design): operator-prepaid attestation storage

### LINKED ISSUE #903: [Docs-Missing] Document cost in NEAR tokens for calling submit_participant_info

### LINKED ISSUE #3972: design: funding model for attestation storage
### Background

`submit_participant_info` takes no deposit and the contract funds attestation storage from its own balance — true of every deployed version, not just since #3940. Any account can create unlimited entries at roughly 1/7th of what each costs the contract, and once the balance no longer covers the contract's own storage it cannot write state at all, which halts signing. Measurements are in the [drain analysis](https://github.com/near/mpc/issues/3972#issuecomment-5119758184) below.

This issue covers **agreeing the funding model**. Implementation is tracked separately in #4015.

The design has to work around two protocol constraints: a function-call access key cannot attach a deposit (not even inside a meta-transaction), and `report_data` binds the quote to `env::signer_account_pk()`, so an operator cannot submit on a node's behalf. Payment and submission therefore have to be separate transactions.

### Acceptance Criteria

- [ ] A funding model is agreed that keeps the node self-submitting with its function-call key and moves the storage cost to whoever onboards it.
- [ ] The decision record names what was chosen, what was rejected, and why — including the alternatives that cannot work (full-access node keys, meta-transactions, operator-submits-on-behalf) and those that were judged not worth it (per-account entry caps, refundable bonds, an owner-callable release method, gating `Attestation::Mock`).
- [ ] The bound is stated: what the contract can still be made to fund, and what is accepted as residual.
- [ ] The operator-facing impact is described, including how many steps change.
- [ ] Design doc merged under `docs/design/`, reviewed by the team.

### Outcome

Settled on a **grant counter**: one prepayment buys one grant, a grant permits one stored attestation entry, and a grant returns when its entry is reclaimed. The contract stores counts rather than amounts, nothing is refunded, and multi-node or migrating operators simply prepay again — so there is no cap constant, no participant-status check, and no assumption about how many nodes share an account.

Doc: [`docs/design/operator-prepaid-attestation-storage.md`](https://github.com/near/mpc/pull/4011) (#4011).

- Implementation: #4015.
- Follows #3940, which made contract-funded storage intentional rather than accidental.
- Related: #3991 (`fail_attestation_submission` becomes removable — the chosen model needs no charge to survive a failing callback, so it is not blocked on this), #3785 (mock attestation expiry, merged, which makes abandoned entries sweepable).
- Slack: https://nearone.slack.com/archives/C0912BTG51T/p1785402605441239?thread_ts=1784896517.987259&cid=C0912BTG51T

barakeinav1:
Closes #3972.

Design doc only — no code. Implementation is tracked in #4015.

- **One prepayment buys one grant** — permission for a node account to store one attestation entry. A grant returns when the entry it paid for is reclaimed, so it is a slot the operator keeps rather than a per-attestation charge. An operator prepays for the nodes they run plus a spare, typically two: the live node and the one they are migrating to. No NEAR is ever refunded.
- **A counter, not a balance.** The contract stores how many unused grants an account has, never how much anyone paid. That removes the arithmetic, and because grants are denominated in *entries* a fee change cannot strand an operator whose deposit is suddenly worth less than an entry.
- **Payment and submission must be separate transactions.** A function-call key cannot attach a deposit — not even inside a meta-transaction, since `validate_delegate_action_key` applies the same check to inner actions — and `report_data` binds the quote to `env::signer_account_pk()`, so an operator cannot submit on the node's behalf.
- **Check early, consume late.** The grant is checked before any verification, so an ungranted call never reaches the `Mock` checks or a `verify_quote` round trip, and consumed at insert, so a failed attestation consumes nothing. That also keeps #3991 unblocked: no charge has to survive a failing callback.
- **Multi-node and migration need no special rule** — prepay again. There is no cap constant, no participant-status check, and no assumption about how many nodes share an account.
- Records the rejected alternatives with their evidence: full-access node keys, meta-transactions, operator-submits-on-behalf, gating `Attestation::Mock`, per-account entry caps, the governance gate, refundable bonds, and an owner-callable release method.

Fee is 0.02 NEAR, ~2.7× the ~0.0073 floor (604-byte worst-case entry plus its ~130-byte grants-map row).

Implementation notes captured for whoever picks this up: it needs a state migration for both the grants map and the `Config` fee field, and rule 3's per-entry write means `clean_invalid_attestations`'s gas budget has to be re-validated.

Separately noted in the doc: `docs/running-an-mpc-node-in-tdx-external-guide.md` still tells operators the call "will incur a cost (TBD, XXX NEAR)" and points at the now-closed #903. Stale on `main` today, independent of this design.


netrome (dismissed):
The doc feels a bit verbose, and I think it could be improved if we slimmed it down and put more emphasis on the core model earlier on in the doc, combined with some sequence diagrams.

But since this is transient, and we have alignment on the core idea I don't think it's critical to polish this doc.

barakeinav1:
Thanks — went through all ten. Nine addressed, one accepted as a bounded residual. Also slimmed the doc from 259 to 162 lines, per @netrome and @gilcu3 in Slack.

> **Rule 3 mints grants for entries that never consumed one.** … every pre-upgrade entry becomes a free grant the moment it is swept.

Accepted deliberately rather than fixed, but your analysis corrected our reasoning for accepting it, so the doc now says something different from what we decided in the meeting.

We had reasoned "it's a small bounded gift to the operators already running nodes". That was wrong, and your note is why: a live node re-attests under rule 1, so its entry never fails re-verification, is never swept, and yields **no** grant. Live operators get nothing — and need nothing. Only *abandoned* entries convert, which is the opposite of the population we thought we were rewarding.

The bound still holds: 14 entries on mainnet and 31 on testnet, nearly all live, so the abandoned set is ~0 today. What changed is that the rationale is now "abandoned entries are rare", not "this rewards our operators" — and because that is only true *at deploy time*, the doc makes it a release-checklist item: check the stored-entry count before shipping, and purge rather than grant if it is in the thousands, which would mean the still-open drain was exploited first. Adding a grant-backed marker would cost ~110 bytes per entry on the fee and a second map, which we judged not worth it at this scale.

> **"Amplification drops below 1" is conditional on `fee >= actual entry cost` holding for the life of every outstanding grant, and grants never expire.**

Agreed and now stated rather than implied. The security section records it as an accepted residual: exposure is bounded by outstanding grants, the contract stores counts rather than purchase prices so it cannot even identify which grants are underwater, governance can re-price only future grants, and a future update can address it if the gap becomes material. We took the operator-certainty side of that trade knowingly.

> **"Re-provisioned … indefinitely without another prepayment" does not hold, because a grant returns only once the old entry becomes *invalid*.**

Correct, and the claim is dropped. A fresh CVM means a new TLS key, so the new submission is rule 2 while the old entry stays valid for up to `DEFAULT_EXPIRATION_DURATION_SECONDS`. The doc now says an operator prepays for the nodes they run **plus a spare** — typically two, the live node and its migration target — and a voluntary-release method is recorded under Alternatives as declined for now, since one spare grant is cheaper than another entry point that has to refuse to evict a live participant. Worth revisiting if the 7-day wait proves painful.

> **The fee-sizing guidance contradicts the fee table.** … 450 bytes is 0.0045 NEAR — below the 0.00604 floor for the entry alone

Fixed. All figures are now explicitly charged bytes, and the mock-is-larger point is restated as 604 vs 599 with the borsh pair removed from the sizing guidance entirely.

> **Rollout step 1 is already done.** PR #3785 is merged and present on `main`

Fixed — this one the earlier summary I was working from had omitted, so thanks for the specificity. Step 1 now records that mock entries are already sweepable and that reclaimability is still not universal: `with_mocked_participant_attestations` stores bare non-expiring `Mock::Valid` sentinels at init, which never fail re-verification and so are never swept or granted.

### Non-blocking

> `WORST_CASE_ENTRY_BYTES` lives inside `#[cfg(test)] mod tests` … both mentions read as if it were a production constant

Fixed. The constant is no longer cited; the doc names the test that pins 604 and notes the fee is a `Config` value, so nothing in the contract reads it. Alternative (a) would have needed the promotion you describe, which is one more reason we did not take it.

> `available_attestation_grants` … in steady state it returns 0 … a consumed prepayment is indistinguishable from none

Fixed: the doc now says `0` is ambiguous and to disambiguate with `get_tee_accounts` — no entry means prepay, entry present is the normal steady state.

> rule 3 adds a map write per removed entry — and a row *insert* when the counter row was deleted at zero — inside a gas-bounded sweep

Good catch, this was missed entirely. Now an implementation note: re-validate `clean_invalid_attestations_tera_gas` and `RESHARE_CLEAN_INVALID_ATTESTATIONS_MAX_SCAN` against the heavier per-entry cost, with a guard test of the kind #3936 adds, plus a worst-case test (every scanned entry removed, every owner's row absent so each write is an insert).

> the early check needs the owning-account comparison, not just key presence

Fixed. Rule 1 now says explicitly that it keys on ownership rather than presence, and why: a presence-only precondition would classify someone else's entry as "existing, no grant needed" and still reach `verify_quote`.

> "Any remainder is kept" deviates from the contract's established deposit convention

Fixed — one sentence on why: a grant is a discrete unit, so the leftover is at most one fee short of the next grant, and a transfer path for sub-0.02 NEAR dust is not worth it.

> rollout step 2 mentions only state migration, but making the fee votable also needs the `ConfigExt` DTO plumbing … and an updated borsh-schema snapshot

Fixed, both named.

Also updated #3972's acceptance criteria, which had drifted from the agreed model — it still required refunding excess and still claimed an operator could pay the first submission on the node's behalf, which the `report_data` binding makes impossible.

docs/design/operator-prepaid-attestation-storage.md
```diff
@@ -0,0 +1,240 @@
...
+```text
+entries an account currently holds  ≤  grants that account has bought
+```
+
+Recycling the grant on reclamation is what makes that work, and it earns its keep twice over:
+
+- **Fairness** — the fee bought storage. Once the contract has that storage back, charging again for the next entry would be charging twice for the same thing.
+- **Operator experience** — an operator prepays for the capacity they actually run (three grants for three nodes, say) and is then done with it. Nodes can be torn down, re-provisioned, rotated and migrated indefinitely without another prepayment, as long as no more than three entries are held at once. Without recycling, every re-provisioned node is a fresh purchase and a fresh thing to remember.
+
+**No NEAR is ever returned.** Recycling gives back capacity, never money: a deposit is consumed permanently the moment it is made, and the only thing that ever comes back is the right to store another entry. There is no withdrawal path and none is planned.
```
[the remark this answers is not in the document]
barakeinav1: > Consider tightening the wording to distinguish "no withdrawals / redeem" from "refund any overpayment in the prepay call", so the document is internally consistent and aligned with #3972.

Done — the two were conflated. The doc now states that "No NEAR is ever returned" covers both cases: no withdrawal method, and no refund of an overpayment. It also says why keeping the remainder deviates from the contract's usual `require_deposit` + `refund_to` pattern — a grant is a discrete unit, so the leftover is at most one fee short of the next grant, and a transfer path for sub-0.02 NEAR dust is not worth it.

On the #3972 alignment: the issue was the stale side, not the doc. Its first acceptance criterion still required refunding excess, which the design meeting overrode — that AC is now updated, along with three others that had drifted from the agreed model.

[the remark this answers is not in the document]
barakeinav1: Correction to my reply above: the design changed rather than the wording.

`prepay_attestation_storage` now takes an explicit grant count and requires an attached deposit of exactly `fee × grants`, rejecting anything else:

```
prepay_attestation_storage(account_id, grants)   #[payable]
```

So there is no remainder to keep, and nothing to reconcile against the `require_deposit` + `refund_to` convention — the case your finding was about no longer exists. My earlier reply justified keeping the remainder; disregard that reasoning. "No NEAR is ever returned" now means only that there is no withdrawal method.

docs/design/operator-prepaid-attestation-storage.md
```diff
@@ -0,0 +1,140 @@
...
+2. **New entry** — consume one grant; reject if the account has none.
+3. **Entry reclaimed** by `clean_invalid_attestations` — return one grant to its owner. There is exactly one removal site, so the counter cannot drift. This adds a grants-map write per removed entry — a row *insert*, not an update, whenever the owner's row was deleted at zero — inside a gas-bounded sweep, so `clean_invalid_attestations_tera_gas` and `RESHARE_CLEAN_INVALID_ATTESTATIONS_MAX_SCAN` need re-validating against the heavier per-entry cost.
+
+Consuming at insert means a failed attestation consumes nothing — nothing was stored, so nothing is owed — and keeps [#3991](https://github.com/near/mpc/issues/3991) unblocked, since no charge has to survive a failing callback.
+
+Migration and multi-node operators need no special rule: prepay again. Hence no cap constant, no participant-status check, and no assumption about how many nodes share an account.
+
+### Fee
+
+0.02 NEAR, about 2.7× the floor. Figures are **charged** bytes — key and record overhead included — not borsh sizes:
```
netrome: Implementation note: I think this should be configurable on the contract and not hard-coded.

netrome (dismissed):
Thanks for updating

docs/design/operator-prepaid-attestation-storage.md
```diff
@@ -0,0 +1,140 @@
...
+
+The counter is **available** grants, not lifetime total, so `0` means either "never prepaid" or "prepaid, now backing an entry"; disambiguate with `get_tee_accounts`. Reclamation waits for the entry to expire (7 days), so an operator prepays for the nodes they run **plus a spare** — typically two: the live node and its migration target.
+
+### API
+
+| Method | Kind | Purpose |
+|---|---|---|
+| `prepay_attestation_storage(account_id, grants)` | `#[payable]` | Adds `grants` to `account_id`. Requires an attached deposit of exactly `fee × grants` and rejects anything else, so there is no remainder to keep or refund. Permissionless — anyone may prepay for any account. |
+| `available_attestation_grants(account_id) -> u32` | view | Grants available. |
+| `attestation_storage_fee() -> NearToken` | view | Current fee. |
```
gilcu3: we might be fine skipping this one. If this is anyway a manual operation by operators, they should read the doc in the `prepay_attestation_storage` to know exactly how much to pay anyway

docs/design/operator-prepaid-attestation-storage.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Fee
+
+0.02 NEAR, about 2.7× the floor. Figures are **charged** bytes — key and record overhead included — not borsh sizes:
+
+| Component | Charged bytes | Cost |
+|---|---|---|
+| Worst-case entry: a `Mock` one at 604 (`Dstack` is 599) | 604 | 0.00604 NEAR |
+| Grants-map row | ~130 | ~0.0013 NEAR |
+| **Floor** | **~734** | **~0.0073 NEAR** |
```
gilcu3, line 62: this is not exactly true in the sense that when we insert elements in a datastructure, such as a HashMap, more bytes are probably needed. The item's size is just the best possible case. It would be good to backup the values we end up using with some benchmark using many attestations.

docs/design/operator-prepaid-attestation-storage.md
```diff
@@ -0,0 +1,140 @@
...
+
+    OP ->> BC: prepay_attestation_storage(node_account, 1) + fee
+    BC ->> BC: grants[node_account] += 1
+
+    NODE ->> BC: submit_participant_info(attestation, tls_pubkey)
+    BC ->> BC: precondition: new entry, owned by caller, grant available?<br/>rejects here if not, before any verification
+    BC ->> BC: verify quote, report_data == hash(tls_pk, signer_pk)
+    BC ->> BC: consume one grant, insert entry
+    BC -->> NODE: ok
+```
```
gilcu3, line 93: it would be more useful to explain that before setting up the node, or starting it, the operator needs to submit the deposit. That is not clear from this graph

docs/design/operator-prepaid-attestation-storage.md
```diff
@@ -0,0 +1,140 @@
...
+
+Without the prepayment the submission fails immediately and the node retries in a loop until granted.
+
+## Operator UX
+
+One new step in the [operator guide](../running-an-mpc-node-in-tdx-external-guide.md), after [Add the Node Account Key](../running-an-mpc-node-in-tdx-external-guide.md#add-the-node-account-key-to-your-account) and before [Submitting Participant Info](../running-an-mpc-node-in-tdx-external-guide.md#submitting-participant-info). Every other step is unchanged, including the off-chain `attestation-cli` check, which this design does not touch.
```
gilcu3: by the time you can get the node account, that node would already be sending attestations that would fail. So this new step should be done before starting the node.

docs/design/operator-prepaid-attestation-storage.md
```diff
@@ -0,0 +1,140 @@
...
+
+Ask for as many grants as the operator runs nodes, plus the spare — the example buys two. Existing attested nodes need no action; they re-attest under rule 1.
+
+## Decisions
+
+| Decision | Chosen | Instead of | Why |
+|---|---|---|---|
+| Accounting | Grant counter | Per-account NEAR balance, NEP-145 shaped | No arithmetic, no amounts stored, and a fee change cannot strand a sold grant. |
+| Multi-node / migration | Prepay per node | Free extra entry gated on `ongoing_migrations`; or a hard cap of N | Both needed a participant check or a magic number; repeating the prepayment needs neither. |
+| Grant semantics | Capacity — reclaiming returns it | Consumable ticket | The contract got the storage back, so charging again charges twice. Also lets an operator re-provision on the grants they hold. |
```
gilcu3: this is a bit confusing to me
```suggestion
| Grant semantics | Capacity — removing/expiring attestation returns it | Consumable ticket | The contract got the storage back, so charging again charges twice. Also lets an operator re-provision on the grants they hold. |
```

docs/design/operator-prepaid-attestation-storage.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Security
+
+Every new entry consumes a paid grant, so the contract funds no new attestation storage and amplification drops below 1. Recycling does not weaken that — a grant returns only once its entry is gone, so `entries held ≤ grants bought` holds throughout. It does permit unlimited churn on a fixed number of grants, which is harmless because each cycle requires the previous entry to be reclaimed first.
+
+Two accepted residuals:
+
+- **Pre-upgrade entries yield a grant when swept**, since rule 3 cannot distinguish them from paid ones. Accepted: there are 14 such entries on mainnet and 31 on testnet, so the free capacity is negligible (and init-time mock sentinels never expire, so they are never swept and never yield one). Check the count at deploy though — a count in the thousands would mean the still-open drain was exploited first, and those entries should be purged rather than granted.
```
gilcu3: why is this an accepted residual? we can just do a contract migration to cover existing ones no? Also we are marking all attestations with an expire date soon, so at that time this is no longer true

docs/design/operator-prepaid-attestation-storage.md
```diff
@@ -0,0 +1,140 @@
...
+- **The contract carries the price risk on sold grants**, bounded by the number outstanding. It stores counts, not purchase prices, so it cannot retro-price them; a future update can address this if the gap becomes material.
```
gilcu3: I know what this means, but reads really cumbersome. Just saying that if storage prices change in the future we might need to adapt.

gilcu3 (dismissed):
Looks fine to me, although there are quite a few things that could be made more readable/accurate to ensure we don't have issues later with the implementation. Would be good if you could change those, although ultimately only the final implementation matters here.

docs/design/operator-prepaid-attestation-storage.md
barakeinav1: do you mean skip this attestation_storage_fee? I think is better to have this in the contract. (docs can get stale)

docs/design/operator-prepaid-attestation-storage.md
barakeinav1: this is why I put 0.02 near as the suggested value- it give us a lot of buffer.

docs/design/operator-prepaid-attestation-storage.md
barakeinav1: in my mind I thought the step would be near the add key to your contract, but actually you are right, this step be done before.

docs/design/operator-prepaid-attestation-storage.md
barakeinav1: not exactly, the node account/key - are exposed in /public_data very soon after node start-up but attestation is send only after the node is sync (at this is what I recall). - but as I said in the comment above, this step can be done sooner.

docs/design/operator-prepaid-attestation-storage.md
barakeinav1: Maybe I didn't explain this right,
existing entries -> are just kept, nothing that we need to do. (it is like we just give them the grant for free, and this is totally fine). I explicitly motioned since one of the bot flag it before.
but we should check before the migration that the number of entries matches the expected number of entries. (in case there is some spamming until we release this fix.

docs/design/operator-prepaid-attestation-storage.md
barakeinav1: Agreed, and that's the design already — it's a governance-votable `Config` field, not a constant. The doc said so eight lines below your comment (and again in the Decisions table), which is too far from the number to count.

Moved it onto the number itself: "0.02 NEAR — a governance-votable `Config` field, not a constant — about 2.7× the floor."

docs/design/operator-prepaid-attestation-storage.md
barakeinav1: Keeping it, but with a better reason than my "docs get stale" — the doc here is *guaranteed* to go stale. The fee is votable, so it changes without a release, and `prepay_attestation_storage` requires exactly `fee × grants` and rejects anything else. A number published in a doc therefore produces failed transactions, not overpayments.

Your point does land in one way: `config()` already exposes the field, so this view isn't the only on-chain source. The doc now says that, and justifies the view as convenience — an operator scripting `attached-deposit` shouldn't have to parse the whole config blob.

docs/design/operator-prepaid-attestation-storage.md
barakeinav1: Correcting my earlier reply — the buffer isn't the argument, and the numbers aren't best-case. 604 is not a computed borsh size: `measure_stored_entry_bytes` (`crates/contract/src/lib.rs`) inserts into the real `stored_attestations`, calls `.flush()`, and returns the `env::storage_usage()` delta. That's the charged cost with the `IterableMap` record and key overhead included — the benchmark you're asking for, already run, and pinned by a test at 604 mock / 599 dstack whose doc comment forbids updating the number to make a failure pass.

The doc asserted "charged bytes" without saying where they came from. It now cites the measurement. The grants-map row (~130) is genuinely an estimate, and the doc now says so; the headroom covers it.

docs/design/operator-prepaid-attestation-storage.md
barakeinav1: Fixed — the diagram now puts the prepayment before the CVM exists, with notes marking the two phases:

```
Note over OP,BC: Before the CVM is started — the prepayment needs
                 only the account id, not the node's key
OP ->> BC: prepay_attestation_storage(node_account, 1) + fee
...
Note over NODE,BC: After start-up and indexer sync
```

docs/design/operator-prepaid-attestation-storage.md
barakeinav1: the doc now anchors the new step to node-account creation

docs/design/operator-prepaid-attestation-storage.md
barakeinav1: Taken, with one correction — "expiring" on its own doesn't return the grant. Expiry only makes the entry sweepable; the grant comes back when `clean_invalid_attestations` actually removes it. That gap is exactly why the design tells operators to buy a spare, so the row shouldn't blur it:

| Grant semantics | Capacity — the grant returns when the entry is swept | Consumable ticket | The contract got the storage back, so charging again charges twice. Expiry alone does not return it, which is why an operator keeps a spare. |

You were right that "reclaiming" was doing unexplained work in that cell.

docs/design/operator-prepaid-attestation-storage.md
barakeinav1: Adopted your migration suggestion — it's better than what I defended, and cheaper than the alternative I rejected.

The migration now credits one grant to each account already holding an entry. Rule 3 then returns a grant that was genuinely issued, so the invariant holds for old and new entries alike, with no per-entry "was this paid for?" marker and no second map (the ~110 bytes/entry option I'd turned down on cost). One residual drops off the list.

Two things it doesn't change, so the doc still says them: the free capacity is the same 14 mainnet / 31 testnet either way — the migration just issues it explicitly and auditably up front instead of drip-feeding it at sweep time — and the deploy-time count check stays, since that's what catches a pre-upgrade spam flood.

On the expiry-date point: that already landed, and it doesn't quite close the case. `stamp_expiry_on_legacy_mocks` is on `main` (`crates/contract/src/v3_13_0_state.rs:122`), but `with_mocked_participant_attestations` still inserts non-expiring `MockAttestation::Valid` sentinels at init (`crates/contract/src/tee/tee_state.rs:136-142`), so those are never swept. Grandfathering handles them cleanly: the sentinel's account gets its grant, and a real submission for the same TLS key is rule 1, which consumes nothing.

docs/design/operator-prepaid-attestation-storage.md
barakeinav1: Fair — you got it and still had to work for it. Rewritten:

> **If storage gets more expensive, grants already sold may not cover their entry** — the contract absorbs the difference, bounded by the number of grants outstanding. It stores counts, not purchase prices, so sold grants cannot be re-priced; revisit if the gap becomes material.

barakeinav1:
Addressed all open review comments in db9badf1 — replies are on each thread. Two are design changes, not wording:

**Prepay moves to node-account creation** (@gilcu3, flow + operator UX). We were both anchoring it too late. `prepay_attestation_storage` needs only the account id, and the operator holds that account's full-access key at [Create a NEAR Account for Your Node](https://github.com/near/mpc/blob/main/docs/running-an-mpc-node-in-tdx-external-guide.md#create-a-near-account-for-your-node) — before the CVM is started, before the node key is retrieved or added. So the node never starts into an ungranted state, and the sync-timing question is moot.

**Existing entries are grandfathered by the migration** (@gilcu3 on the residual). One grant credited per account already holding an entry, so rule 3 returns grants that were genuinely issued. Removes an accepted residual and the rule-3 asymmetry, without the per-entry marker I'd rejected on cost. Same 14/31 free slots, issued explicitly up front; the deploy-time count check stays.

Rest: fee is now labelled a votable `Config` field where the number appears (@netrome — it already was, just stated too far down); the 604/599 figures now cite the `env::storage_usage()` measurement behind them rather than asserting them (@gilcu3 — and my "0.02 gives buffer" reply was the wrong defence); plainer wording for the grant-semantics row and the price-risk residual.

crates/contract/src/lib.rs
gilcu3, line 8197: Thanks for the reference, but this does not measure real deposit costs, as it only inserts one attestation into an empty state

docs/design/operator-prepaid-attestation-storage.md
gilcu3: > but with_mocked_participant_attestations still inserts non-expiring MockAttestation::Valid sentinels at init (crates/contract/src/tee/tee_state.rs:136-142), so those are never swept. Grandfathering handles them cleanly: the sentinel's account gets its grant, and a real submission for the same TLS key is rule 1, which consumes nothing.

oh, if that is true, why didn't we close that gap in https://github.com/near/mpc/pull/3785 ?

docs/design/operator-prepaid-attestation-storage.md
gilcu3: This benchmark is still not so accurate for what we need. The underlying data structure is a hash map, so its total size "per entry" is not as easy to compute. To really figure this out we would need to insert many attestations and then get the average or maximum of the cost per attestation. Therefore, we still don't have proper numbers, but we could do it as part of the PR following this design. No need to mention an amount here, except as a rough estimate

docs/design/operator-prepaid-attestation-storage.md
gilcu3: >  The fee is votable, so it changes without a release

Right, we are making it votable :) Probably we didn't need that though, I don't see operators ever changing that, but that's fine. But again, if the operator can already get the value from `config()` then we don't need this convenience. Extracting the value from the config blob is trivial, and we do not expect any operator to do this automatically anyway, as this is meant to be a manual step after all, specially for a value that would almost never change.

gilcu3 (dismissed):
Thanks for the updates

Replied to some comments. We still don't fully agree on some of them, but this is approved either way as none of them are blockers to me

barakeinav1:
> The underlying data structure is a hash map, so its total size "per entry" is not as easy to compute. To really figure this out we would need to insert many attestations and then get the average or maximum of the cost per attestation.

Ran that experiment — 200 sequential inserts into `stored_attestations`, taking the `env::storage_usage()` delta of each one individually:

```
first=232  2nd=232  50th=233  100th=233  200th=234
distinct deltas across all 200 = [232, 233, 234]
```

**Per-entry cost is constant in map size.** The 2-byte drift is my probe's own fault: I named the accounts `op0.near` … `op199.near`, so the account id grew by two characters between the first insert and the last. Nothing to do with occupancy.

The reason is that `near_sdk::store::IterableMap` is not a bucketed hash map. It is a `LookupMap` (one trie record per entry, keyed by prefix + serialised key) plus a `Vector` of keys (one record per element). NEAR's storage accounting charges per *record* — key bytes + value bytes + a fixed 40-byte per-record overhead — and never for trie internal nodes. So the marginal cost of the Nth insert does not depend on N, and averaging over many inserts returns the same number as measuring one.

That means `measure_stored_entry_bytes` inserting into empty state is a valid measurement rather than a best case, and 604 mock / 599 dstack stand. Happy to add this probe as a permanent test in the implementation PR if you would rather it were pinned than argued — it is ~20 lines.

Two caveats I do want to keep, and the doc says both:

- The **~130 bytes for the grants-map row is an estimate**, not a measurement. It is a much smaller term than the entry and the fee's headroom covers it, but it has not been measured the way 604 has.
- The fee is deliberately ~2.7× the ~734-byte floor precisely so that neither figure has to be exact.

crates/contract/src/lib.rs
barakeinav1: see here some measurement results https://github.com/near/mpc/pull/4011#issuecomment-5157838922

docs/design/operator-prepaid-attestation-storage.md
barakeinav1: Please disregard my previous reply on this thread — it was drafted by an agent and I posted it without reviewing it first.

Two things:

**1. The migration in this PR is deliberately simple.** Existing attestations stay in place and get no special handling: an entry already holds a slot that no grant was ever bought for, so we treat it as already paid for. No crediting step, no per-entry marker, no second map. Re-attestation under the same TLS key is rule 1, which consumes nothing, so those operators need no grant and no action from us.

**2. `with_mocked_participant_attestations` runs only at init, so no real network is affected.** Its only non-test callers are `init` and `init_running`, both `#[init]` without `ignore_state`, so they panic if state already exists — mainnet and testnet are long past that point and nothing re-runs them. The non-expiring sentinels therefore only exist in freshly initialised state: localnet, e2e and the sandbox tests. They are also overwritten as soon as the real node submits for the same TLS key.

That is why I left the init path alone in #3785 — I wanted that PR limited to state that actually exists on a live network. #1087 already tracks the init-time attestation question more broadly.

docs/design/operator-prepaid-attestation-storage.md
barakeinav1: Nothing changed yet — I flagged this rather than acting on it, since it is permanent public API.

I am fine with removing it. Your argument holds: `config()` already returns the field, the operator reads it once by hand as part of a manual step, and the value will almost never change. That does not pay for a dedicated public method, an ABI entry and a thing to keep working forever.

@netrome — a second opinion before it goes? You were the one who pushed for the fee to be configurable rather than hard-coded, so you may have a view on whether a dedicated read is worth keeping alongside `config()`.

I will start the implementation with the view in place, since dropping it later is a one-line removal plus an ABI snapshot regeneration. If we settle on removing it, that happens before the implementation PR merges rather than after.

gilcu3 (approved):

docs/design/operator-prepaid-attestation-storage.md
netrome: Yeah if it's in the config, there's no need for a dedicated view function.

docs/design/operator-prepaid-attestation-storage.md
netrome: Yeah I think this is also why it's worth keeping this configurable.

docs/design/operator-prepaid-attestation-storage.md
```diff
@@ -0,0 +1,156 @@
...
+| `available_attestation_grants(account_id) -> u32` | view | Grants available. |
+| `attestation_storage_fee() -> NearToken` | view | **Optional — open question.** The fee is already returned by `config()`, and reading it is a manual step on a value that almost never changes, so a dedicated method may not earn its permanent place in the API. Left out of the implementation until decided; adding it later is cheaper than removing it. |
+
+An operator reads the current fee from `config()` and attaches the exact multiple — the fee is votable, so a number published in a doc instead would eventually produce failed transactions, not overpayments. "No NEAR is ever returned" therefore means simply: no withdrawal method.
+
+### Charging rules
+
+Evaluated read-only at the top of `submit_participant_info` — before any verification, so an ungranted or unauthorised call never reaches the `Mock` checks or a `verify_quote` round trip — and applied at insert.
```
netrome, line 44: Nit: Don't see the point in distinguishing the different verifications we have here

```suggestion
Evaluated read-only at the top of `submit_participant_info` — before any verification, so an ungranted or unauthorised call never reaches attestation verification checks.
```

docs/design/operator-prepaid-attestation-storage.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Existing nodes
+
+Already-attested nodes need no grant and their operators need to do nothing. Their entry is already stored, and re-attesting under the same key updates it in place for free (rule 1) — so an existing entry is a grant already spent: it holds a slot, and none was ever bought for it.
+
+Nothing in the contract treats these entries specially: no migration step, no per-entry marker, no second map. That is the point — for existing nodes this change needs no code at all.
+
+Accepted deliberately: if such an entry is later swept, rule 3 hands its owner a grant they never bought. Negligible at 14 entries on mainnet and 31 on testnet, but check the count before deploying — thousands would mean the drain was exploited first, and those entries should be purged rather than left to become grants.
+
+## Operator UX
```
netrome, line 105: This feels a bit strange but sure, it's worth keeping as a reminder to update the operator guide - although we should always remember to do this for operator-facing changes.

netrome (approved):
Thanks!

PR #4013, opened 2026-07-30
feat: Starknet RPC healthcheck probing logic

### LINKED ISSUE #4095: Probe Starknet for its chain id
### Background

Part of #4003. First slice: introduces the probe machinery in `foreign-chain-health-check` (`probe_all_providers`, per provider status) with Starknet as the first covered chain, via `starknet_chainId`.

### Acceptance Criteria

- A Starknet provider answering the expected chain id is reported healthy; a different chain id is reported unhealthy.
- Every configured provider gets its own status row; a chain without probing logic reports `ProbeNotImplemented`.
- No report row renders upstream error text, so credentials spliced into URLs cannot leak.

haiyuechen-nearone:
Closes #4095. First slice of the identity probing series, continued by #4049 and the stack of #4067, #4069, #4070 and #4071.

Probe machinery, Starknet only. Introduces a `ChainIdentity` value, the `ChainIdentityInspector` trait, and `probe_all_providers`. Nothing calls it at startup yet. Node wiring, metrics and the remaining chains in follow-up PRs.

### Notes for review

- **`FanOut` is re-keyed by `ProviderId`.** The names were already available at construction and discarded, so fan-out failures logged `inspector_index = 2` rather than naming the provider. The only behavioural change here on the signing path; `extract`'s consensus logic is untouched.

- **`chain_identities()` is deliberately not `extract()`'s quorum.** `extract` collapses dissent into `InspectorResponseMismatch`: right for signing, wrong for diagnostics, where each provider's failure must stay visible. Hence a separate method, and a separate trait rather than a method on `ForeignChainInspector`.

- **`max_retries` is now honoured** on the probing path. The probing logic retries transient failures. Currently still a TODO for the signing path.

- **`ForeignChainInspectionError::provider_failure() -> Option<ProviderFailure>`** is new public API on the inspector crate, so the classification lives next to the variants it matches. `None` covers the variants that report the transaction's own state: a reverted or non-final transaction is an answer, not a provider fault, and only the probe treats it as anomalous.

- `ProviderStatus` carries no rendered error text. `Path` auth splices the API key into the URL and upstream errors interpolate that into their `Display`; a guard test asserts neither reaches the report.

- Chains with no identity impl report `ProbeNotImplemented` for the time being. This variant goes away once every chain has probing logic implemented.


haiyuechen-nearone:
@claude review

crates/foreign-chain-health-check/src/lib.rs
```diff
@@ -1,10 +1,13 @@
...
+//!
+//! * [`check_all_providers`] replays a fixed golden transaction per chain.
+//! * [`probe::probe_all_providers`] asks each provider for the network it serves and compares that
+//!   against the operator's configured expectation.
 
 mod checks;
 mod golden;
 mod network;
+pub mod probe;
 mod results;
```
netrome, line 12: Nit: I'd separate public and private module declarations

```suggestion
pub mod probe;

mod checks;
mod golden;
mod network;
mod results;
```

crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -0,0 +1,780 @@
...
+    AuthTokenUnresolved,
+    /// The RPC client could not be built from the provider's URL and auth.
+    ClientSetupFailed,
+    /// The chain is configured without an `expected_network_fingerprint`, so its providers cannot
+    /// be checked. Reported rather than skipped: silence would read as healthy.
+    MissingExpectedFingerprint,
+    /// The chain has no fingerprint probe yet, either because none is written for it or because the
+    /// node cannot inspect it at all.
+    ProbeNotImplemented,
+}
```
netrome, line 37: I'd trim down most of these docstrings as they add more noise than signal. "Carries no rendered RPC error" is a weird statement. This feels very LLM generated.

```suggestion
/// One provider's verdict. Anything other than [`ProviderStatus::Healthy`] is unhealthy.
#[derive(Debug, Clone, PartialEq, Eq)]
pub enum ProviderStatus {
    Healthy,
    WrongNetwork {
        expected: NetworkFingerprint,
        observed: NetworkFingerprint,
    },
    /// DNS, TLS, connection refused, 5xx, or rate limiting.
    Unreachable,
    /// The provider answered and refused: credentials invalid, or not enabled for this chain.
    RequestRejected,
    MalformedResponse,
    TimedOut,
    AuthTokenUnresolved,
    ClientSetupFailed,
    /// The chain is configured without an `expected_network_fingerprint`, so its providers cannot
    /// be checked.
    MissingExpectedFingerprint,
    ProbeNotImplemented,
}
```

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -35,6 +41,25 @@ pub trait ForeignChainInspector {
...
 
+/// The network a provider serves, as the chain itself reports it: a chain id or a genesis hash, in
+/// one canonical text form per chain.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Display, From)]
+pub struct NetworkFingerprint(String);
+
+/// Reports the [`NetworkFingerprint`] of the provider an inspector talks to.
+///
+/// Fingerprints are compared verbatim, so both the reported and the expected one go through the
+/// impl's canonical form. Fetches a chain-wide constant providers never prune.
```
netrome: What does "impl's canonical form" mean here? I think this part of the comment adds more confusion than help.

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -35,6 +41,25 @@ pub trait ForeignChainInspector {
...
+pub trait NetworkFingerprintInspector {
+    fn network_fingerprint(
+        &self,
+    ) -> impl Future<Output = Result<NetworkFingerprint, ForeignChainInspectionError>> + Send;
+
+    /// Puts an operator-supplied fingerprint into the form [`Self::network_fingerprint`] returns,
+    /// so that a spec-legal spelling of the right network does not read as the wrong network.
+    fn canonical_fingerprint(expected: &str) -> NetworkFingerprint;
```
netrome: This also feels confusing. Is this canonicalization only valid for the operator flow? Otherwise I'd have the docstring focus on the expected behavior rather than providing some ad-hoc context on our current usage.

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -168,6 +198,50 @@ where
...
+                        break;
+                    }
+                    tokio::time::sleep(RETRY_BACKOFF).await;
+                    outcome = ask().await;
+                }
+                (provider, outcome)
+            });
+        }
+        join_set.join_all().await
+    }
```
netrome, line 240: It's a bit unfortunately that we're not injecting time properly here, but we haven't established patterns for doing so in this repo yet so for now this is okay.

crates/foreign-chain-inspector/tests/rpc_error_classification.rs
```diff
@@ -0,0 +1,207 @@
...
+    // When
+    let classified = ForeignChainInspectionError::classify_rpc_client_error(error);
+
+    // Then
+    assert_matches!(
+        &classified,
+        ForeignChainInspectionError::RpcRequestRejected(_)
+    );
+    assert!(!classified.is_transient());
+}
```
netrome: Why are these tests written in an integration test module? They look like unit tests to me

crates/foreign-chain-inspector/tests/starknet_inspector.rs
```diff
@@ -507,3 +514,145 @@ async fn extract__should_return_event_log_for_specific_index_via_http_rpc_client
         extracted_values,
     );
 }
+
+/// Starknet mainnet's chain id, `SN_MAIN` in ASCII.
+const MAINNET_CHAIN_ID: &str = "0x534e5f4d41494e";
+
+#[tokio::test]
+async fn network_fingerprint__should_return_the_canonical_chain_id() {
+    // Given: the chain id padded and uppercased, as a provider is free to send it.
```
netrome: The given comment should not need further details.

```suggestion
    // Given
```

crates/foreign-chain-inspector/tests/starknet_inspector.rs
```diff
@@ -507,3 +514,145 @@ async fn extract__should_return_event_log_for_specific_index_via_http_rpc_client
...
+    assert_eq!(calls.load(Ordering::SeqCst), 3);
+    assert_matches!(
+        results[0].1,
+        Err(ForeignChainInspectionError::RpcRequestFailed(_))
+    );
+}
+
+#[tokio::test]
+async fn network_fingerprints__should_not_retry_a_provider_that_refused_the_request() {
+    // Given a provider refusing with a JSON-RPC error object, as one does for a bad API key
```
netrome:
```suggestion
    // Given
```

crates/foreign-chain-inspector/tests/starknet_inspector.rs
```diff
@@ -507,3 +514,145 @@ async fn extract__should_return_event_log_for_specific_index_via_http_rpc_client
...
+            None::<()>,
+        )))
+    });
+
+    // When
+    let results = fan_out
+        .network_fingerprints(Duration::from_secs(1), NonZeroU64::new(3).unwrap())
+        .await;
+
+    // Then the refusal is reported as one, and the remaining attempts are not spent on it
```
netrome:
```suggestion
    // Then
```

crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -0,0 +1,780 @@
...
+        mock.assert_async().await;
+        assert_eq!(
+            must_status_of(&report, ForeignChain::Starknet, "publicnode"),
+            ProviderStatus::Healthy
+        );
+    }
+
+    #[tokio::test]
+    async fn probe_all_providers__should_report_a_provider_on_another_network_as_wrong_network() {
+        // Given a provider serving Sepolia while the operator configured mainnet
```
netrome: I know Claude loves to add these comments, but they just add noise imo - and the risk is that they grow stale over time.

```suggestion
        // Given
```

Please fix this for the rest of the functions as well

crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -0,0 +1,780 @@
...
+
+        // Then
+        assert_eq!(
+            must_status_of(&report, ForeignChain::Starknet, "publicnode"),
+            ProviderStatus::MalformedResponse
+        );
+    }
+
+    #[tokio::test]
+    async fn probe_all_providers__should_report_a_provider_that_does_not_answer_in_time() {
```
netrome, line 517: This test runs in around a second right now, which is horribly slow imo. With better dependency injection (e.g. by controlling how time and the http client) is injected in the code, it should be possible to make this test effectively instant.

We don't need to do this now, since the state of the rest of the existing code base is in a pretty bad shape in this regard. But it feels like a broken window situation when we start letting these things slip in new code just because the existing code doesn't handle it properly.

A nice engineering quality follow-up would be to refactor the `probe_all_providers` method and the `network_fingerprints` method to properly inject these dependencies but I won't consider it a hard blocker for this PR

netrome (dismissed):
I think there's some room for improvement here, but nothing strictly blocking.

haiyuechen-nearone (commented):
> Why are these tests written in an integration test module? They look like unit tests to me 

I felt the file was getting too big so moved them to a dedicated file, moving them back into  crates/foreign-chain-inspector/src/lib.rs

crates/foreign-chain-health-check/src/lib.rs
haiyuechen-nearone: Will fix and also raised https://github.com/near/mpc/pull/4036 to address this across the whole repo

crates/foreign-chain-inspector/src/lib.rs
haiyuechen-nearone: Raised https://github.com/near/mpc/issues/4043 to address this in a follow up

crates/foreign-chain-health-check/src/probe.rs
haiyuechen-nearone: Raised https://github.com/near/mpc/issues/4043 to address this in a follow up

crates/foreign-chain-health-check/src/lib.rs
netrome: Thanks, much appreciated :pray:

crates/foreign-chain-inspector/src/lib.rs
netrome: Thanks for raising :pray:

crates/foreign-chain-health-check/src/lib.rs
```diff
@@ -1,6 +1,10 @@
-//! Foreign-chain RPC provider health checks: probe every configured provider
-//! with a fixed golden request and report a per-provider result. Sui is the
-//! exception — see `run_sui`.
+//! Foreign-chain RPC provider health checks, in two independent routes:
+//!
+//! * [`check_all_providers`] replays a fixed golden transaction per chain.
+//! * [`probe::probe_all_providers`] asks each provider for the network it serves and compares that
+//!   against the operator's configured expectation.
+
+pub mod probe;
```
netrome, line 7: Made me think of https://linux.die.net/man/8/modprobe 😅

netrome (approved):
Thanks for updating 🙏 only did a shallow sweep now, but seems like most of my nits have been addressed.

crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -0,0 +1,793 @@
...
+pub struct ProbeReport {
+    rows: Vec<ProviderHealth>,
+}
+
+impl ProbeReport {
+    pub fn rows(&self) -> &[ProviderHealth] {
+        &self.rows
+    }
+
+    /// Only configured chains appear, never reports on a chain the operator did not configure.
```
anodar, line 68: nit: it's clear from the function name, comment doesn't seem to add much value. Consider dropping it.

crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -0,0 +1,793 @@
...
+        }
+        counts
+    }
+}
+
+/// Probe every configured provider concurrently.
+///
+/// Each provider is tried up to `max_retries` times, `timeout_sec` per try, and only for as long as
+/// the failures stay transient. This returns within the largest configured `timeout_sec *
+/// max_retries`, plus the [`foreign_chain_inspector::RETRY_BACKOFF`] between tries.
```
anodar, line 86: Not part of your change and non-actionable, just a a thought: I think we don't really need max_retries, timeout_sec to be configurable per chain. 

I don't see why would we retry one different times or with different timeouts than the other.  I think ideally we should just factor it out into retry parameter constants (of `backon::backoff::constant::ConstantBuilder` type) and pass it to retry.

crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -0,0 +1,793 @@
...
+    let mut inspectors = Vec::new();
+    let mut rows = Vec::new();
+    for (name, provider) in config.providers.iter() {
+        let provider_id = ProviderId(name.as_str().to_owned());
+        match new_inspector(provider) {
+            Ok(inspector) => inspectors.push((provider_id, inspector)),
+            Err(error) => rows.push(ProviderHealth {
+                chain,
+                provider: provider_id,
+                status: setup_failure(&error),
```
anodar, line 129: We're swallowing actual error here and only surfacing AuthTokenUnresolved/ClientSetupFailed, including the error we get from new_inspector would be lot more helpful.

crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -0,0 +1,793 @@
...
+            chain,
+            provider: ProviderId(name.as_str().to_owned()),
+            status: status.clone(),
+        })
+        .collect()
+}
+
+/// A provider answers what it likes and the report reaches logs and metric labels, so the length is
+/// capped well clear of the longest real fingerprint: Bitcoin's genesis hash, at 66 characters.
+fn bounded(observed: NetworkFingerprint) -> NetworkFingerprint {
```
anodar, line 180: Why don't we crop it at NetworkFingerprint object creation and drop this method entirely?

That is, drop derived From and add `NetworkFingerprint::new(impl Into<String>)`

anodar (approved):
Didn't have a deep look but left few non-blocking comments. 

Once we wire this in the node startup, we should ideally also add e2e test for it.

crates/foreign-chain-health-check/src/probe.rs
haiyuechen-nearone: https://github.com/near/mpc/issues/4043 Folded this improvement here as well since I will be refactoring with proper dependency injection

crates/foreign-chain-health-check/src/probe.rs
haiyuechen-nearone: I think max_retries yes, but timeout_sec maybe not. I can imagine different providers having different rate limiting setups.

PR #4014, opened 2026-07-30
feat: add metrics for keyshare backups

### LINKED ISSUE #4010: Node metrics for keyshare backups served to a backup service
### Background

A node can serve its keyshares to a backup client over the migration service, but nothing on the node side records that it did. Operators have no way to tell a node whose keyshares are backed up from one whose newest keyset only exists on the node itself, and no signal to alert on when backups silently stop happening. The node is the right place to emit this: it knows which keyset it handed out and when, regardless of how or by whom backups are taken.

### Acceptance Criteria

The node exposes `mpc_last_backup_served_epoch` (epoch id of the last keyset served) and `mpc_last_backup_served_timestamp_seconds` (Unix time of that response).

Both are only advanced when the migration-service web server actually returns keyshares; a request for a keyset the node does not hold leaves them untouched.

`docs/design/node-operator-metrics.md` documents both metrics, the recommended staleness alert, and the caveat that they confirm the node served the keyshares, not that the backup client persisted them.

gilcu3:
Closes #4010

### DIFF
--- a/crates/node/src/coordinator.rs
+++ b/crates/node/src/coordinator.rs
@@ -110,7 +110,7 @@ where
     pub async fn run(mut self) -> anyhow::Result<()> {
         loop {
             let state = self.indexer.contract_state_receiver.borrow().clone();
-            if let Some(epoch_id) = current_epoch_id_gauge_value(&state) {
+            if let Some(epoch_id) = current_epoch_id(&state) {
                 metrics::MPC_CURRENT_EPOCH_ID.set(epoch_id);
             }
             let mut job: MpcJob = match state {
@@ -856,11 +856,10 @@ impl Drop for ReportCurrentJobGuard {
     }
 }
 
-fn current_epoch_id_gauge_value(state: &ContractState) -> Option<i64> {
-    match state {
-        ContractState::Running(running) => i64::try_from(running.keyset.epoch_id.get()).ok(),
-        ContractState::Invalid | ContractState::Initializing(_) => None,
-    }
+fn current_epoch_id(state: &ContractState) -> Option<i64> {
+    state
+        .epoch_id()
+        .and_then(|epoch_id| i64::try_from(epoch_id.get()).ok())
 }
 
 /// The `host:port` a peer is currently reachable at in live contract state, re-read on every
@@ -1090,8 +1089,8 @@ fn make_initializing_stop_fn(
 #[expect(non_snake_case)]
 mod tests {
     use super::{
-        current_epoch_id_gauge_value, participants_change_requires_restart,
-        peer_address_from_state, register_foreign_chains, stop_running,
+        current_epoch_id, participants_change_requires_restart, peer_address_from_state,
+        register_foreign_chains, stop_running,
     };
     use crate::indexer::participants::ContractState;
     use crate::indexer::participants::test_utils::{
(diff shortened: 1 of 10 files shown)


crates/node/src/migration_service/web/server.rs
```diff
@@ -118,6 +124,40 @@ async fn spawn_expected_peer_info_monitoring(
...
+    );
+    response
+}
+
+#[derive(Debug, thiserror::Error)]
+enum RecordBackupServedError {
+    #[error("epoch id {0} does not fit in an i64 gauge")]
+    EpochOutOfRange(u64),
+    #[error("unix timestamp {0} does not fit in an i64 gauge")]
+    TimestampOutOfRange(u64),
```
gilcu3: this reads a bit strange, but it is for u64 -> i64 conversion

gilcu3:
> * `docs/design/node-operator-metrics.md:51` — the recommended alert can never fire for a node that has **never** served a backup, which is exactly the population the linked issue wants to surface ("a node whose newest keyset only exists on the node itself"). Both gauges live behind `LazyLock`, and `register_int_gauge!` runs _inside_ the `LazyLock` closure, so the collector isn't in `default_registry()` until the first `.set()` — i.e. until the first successful serve. `metrics()` (`crates/node/src/web.rs:49-53`) only force-initializes the build-info metric, so `/metrics` omits `mpc_last_backup_served_timestamp_seconds` entirely, and `time() - <absent series>` evaluates to an empty vector: no alert, ever. Fix either side:

our alerts do trigger if there is no data, so this is not a problem. Addressed most other comments

@claude review

crates/node/src/migration_service/web/server.rs
netrome: Perhaps `TimestampTooLargeForI64` or `TimestampConversion` could be more descriptive names? The current one also works. Wouldn't have reacted if it wasn't for your comment.

netrome (dismissed):
Nice stuff!

netrome:
While technically out of scope for the issue it would be nice to also add metrics for the current epoch, so that we could add alerts if the latest epoch isn't backed up within X hours.

crates/node/src/coordinator.rs
```diff
@@ -110,6 +110,9 @@ where
     pub async fn run(mut self) -> anyhow::Result<()> {
         loop {
             let state = self.indexer.contract_state_receiver.borrow().clone();
+            if let Some(epoch_id) = current_epoch_id_gauge_value(&state) {
```
netrome: This looks strange to me, it sounds like we're getting the current gauge value and setting the same value back to the gauge. I see the method extracts the epoch ID from the state, so I'd suggest dropping the `_gauge_value` suffix from the name.

```suggestion
            if let Some(epoch_id) = current_epoch_id(&state) {
```

crates/node/src/coordinator.rs
```diff
@@ -853,6 +856,13 @@ impl Drop for ReportCurrentJobGuard {
     }
 }
 
+fn current_epoch_id_gauge_value(state: &ContractState) -> Option<i64> {
```
netrome: Suggest renaming this. Moreover, did you consider making this a setter on the contract state? I think it would read even better as `state.epoch_id()`.

```suggestion
fn current_epoch_id(state: &ContractState) -> Option<i64> {
```

netrome (dismissed):
Thanks for updating, some improvement suggestions but not hard blockers.

crates/node/src/coordinator.rs
gilcu3: Yeah I can add the getter and drop the function here

gilcu3:
Extra CI check still fail (`ruint`), but I am leaving those as we already have a PR for that fix

haiyuechen-nearone (approved):
🐐

netrome (approved):
Thanks!

PR #4016, opened 2026-07-30
feat(contract): operator-prepaid attestation storage (grant counter)

### LINKED ISSUE #4015: feat(contract): operator-prepaid attestation storage (grant counter)
### Background

Implements the funding model designed in #3972 and specified in [`docs/design/operator-prepaid-attestation-storage.md`](https://github.com/near/mpc/pull/4011). An operator prepays for attestation-entry storage in a separate transaction; the node keeps self-submitting with its deposit-less function-call key. One prepayment buys one **grant** — permission to hold one attestation entry — and a grant returns when its entry is reclaimed. Fee is 0.02 NEAR. Nothing is refunded.

This closes the drain that is live today: `submit_participant_info` takes no deposit on every deployed version, so any account can create unlimited entries at roughly 1/7th of what each costs the contract, and an exhausted balance stops the contract writing state at all.

### Acceptance Criteria

- [ ] `prepay_attestation_storage(account_id, grants)` — `#[payable]`, permissionless, requires an attached deposit of exactly `fee × grants` and rejects anything else.
- [ ] `available_attestation_grants(account_id) -> u32` view. No dedicated fee view: `config()` already returns `attestation_storage_fee_millinear`, and an operator reads it once by hand ([#4011 discussion](https://github.com/near/mpc/pull/4011#discussion_r3688452375)).
- [ ] `available_attestation_grants: LookupMap<AccountId, u32>` holds *available* grants; prepay `+1`, new entry `−1`, reclaimed entry `+1`, row deleted at zero.
- [ ] Charging rules: a re-attestation under a TLS key the caller already owns consumes nothing; a new entry consumes one grant and is rejected if none is available; `clean_invalid_attestations` returns one grant to the owner of each entry it removes.
- [ ] The precondition is evaluated read-only at the top of `submit_participant_info` — before any `Mock` verification or `verify_quote` round trip — and keys on *ownership*, not TLS-key presence. The grant is consumed at insert, so a failed attestation consumes nothing.
- [ ] `Config.attestation_storage_fee` is governance-votable, defaulting to 0.02 NEAR, with the `ConfigExt` DTO plumbing in `crates/contract/src/dto_mapping.rs`.
- [ ] State migration for the new map and `Config` field: frozen snapshot module plus a `migrate()` arm as `crates/contract/src/v3_13_0_state.rs` does, with regenerated borsh-schema and ABI snapshots.
- [ ] `clean_invalid_attestations_tera_gas` and `RESHARE_CLEAN_INVALID_ATTESTATIONS_MAX_SCAN` re-validated against the added per-entry write, which is a row *insert* whenever the owner's row was deleted at zero.
- [ ] Operator guide gains one step between [Add the Node Account Key](https://github.com/near/mpc/blob/main/docs/running-an-mpc-node-in-tdx-external-guide.md#add-the-node-account-key-to-your-account) and [Submitting Participant Info](https://github.com/near/mpc/blob/main/docs/running-an-mpc-node-in-tdx-external-guide.md#submitting-participant-info), and the stale "will incur a cost (TBD, XXX NEAR)" note citing the closed #903 is removed.
- [ ] Tests: prepay-then-submit; re-attestation consuming no grant; prepay-on-behalf by another account; rejection with no grant; a TLS key owned by another account rejected by the early check; a failed attestation consuming nothing on both sync and async paths; a reclaimed entry returning a spendable grant; the row deleted at zero; a kicked node keeping its grants; two prepayments allowing two entries; and a `clean_invalid_attestations` gas-budget guard for the worst case (every scanned entry removed, every owner's row absent).

https://github.com/near/mpc/issues/3991 - can potentially be done in same PR as well.

- Design: #3972, doc in #4011.
- **Check the stored-entry count before deploying.** Pre-upgrade entries yield a grant when swept, since the rule cannot distinguish them from paid ones. That is accepted at today's scale — 14 entries on mainnet, 31 on testnet — but a count in the thousands would mean the still-open drain was exploited first, and those entries should be purged rather than granted.
- Unblocks #3991: nothing is refunded and no charge has to survive a failing callback, so `fail_attestation_submission` stays removable independently of this work.
- Operators need one grant per node plus a spare, because a grant returns only once its entry expires (7 days). An owner-callable release method was considered and declined; see the design doc's Alternatives.

barakeinav1:
Closes #4015. Implements [`docs/design/operator-prepaid-attestation-storage.md`](https://github.com/near/mpc/blob/main/docs/design/operator-prepaid-attestation-storage.md), agreed in #3972 / #4011 — read that for the model, the alternatives considered and the fee derivation.

An operator prepays for attestation-entry storage in a separate transaction; the node keeps self-submitting with its deposit-less function-call key. One prepayment buys one **grant**, and the grant returns when the entry it paid for is reclaimed.

- `prepay_attestation_storage(account_id, grants)` — payable, permissionless, requires exactly `fee × grants`.
- `available_attestation_grants(account_id)` view. The fee is read from `config()`; there is deliberately no dedicated view ([why](https://github.com/near/mpc/pull/4011#discussion_r3688452375)).
- `Config.attestation_storage_fee_millinear`, default 20 (0.02 NEAR), votable via `ConfigExt`.
- Charging: re-attestation consumes nothing; a new entry consumes one grant; `clean_invalid_attestations` returns one grant per entry it removes.

## Worth a reviewer's attention

- The precondition is read-only and runs **before any verification**, and compares the *owning account* rather than just testing key presence — otherwise a submission for somebody else's key would be classified as needing no grant and would still reach `verify_quote`. It is re-checked inside `resolve_verification`, because that callback runs a receipt later, where the grant may since have been consumed.
- `TeeState::clean_invalid_attestations` now returns the **owners** of the entries it removed rather than a count, so the caller can credit them. `MpcContract::clean_invalid_attestations` still returns the count, so the external interface is unchanged.
- Entries predating the fee need no handling — they already hold a slot no grant was bought for, and re-attestation is free. Migration only initialises the map. See the doc's *Existing nodes*.
- All three test harnesses prepay only when a submission would actually consume a grant. The e2e harness prepays for **every** node, not just initial participants: a node joining by resharing attests from its own process with a key that cannot attach a deposit, so its grant must exist beforehand.

## Out of scope

Two follow-ups, neither affecting the contract:

- **#4082** — the e2e cluster prepays a grant for every node, including initial participants, whose sentinel entry from `init` means they never consume it. Wasteful, not harmful.
- **#4035** — sweep gas. This adds a per-removal write taking the marginal cost from 0.347 to 0.504 TGas, but `clean_invalid_attestations_tera_gas` was already short of `RESHARE_CLEAN_INVALID_ATTESTATIONS_MAX_SCAN` on main. Nothing to do here, though: those two constants govern only the promise `vote_reshared` schedules. `clean_invalid_attestations` is permissionless and takes `max_scan` as a parameter, so an external caller is bound by neither — one manual call at ~300 TGas clears roughly 550 entries. The worst this PR contributes is a manual invocation, not a stuck backlog.


barakeinav1:
@claude review

crates/contract/src/lib.rs
```diff
@@ -832,6 +902,68 @@ impl MpcContract {
         }
     }
 
+    /// Fee for one attestation-storage grant. Not exposed as its own view: `config()`
+    /// already returns `attestation_storage_fee_millinear`, and an operator reads it once
+    /// by hand. TODO(#4015): add a dedicated view if operators ask for one.
+    fn attestation_storage_fee_internal(&self) -> NearToken {
+        NearToken::from_millinear(u128::from(self.config.attestation_storage_fee_millinear))
+    }
```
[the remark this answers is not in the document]
barakeinav1: Right — the view was dropped on the design PR (near/mpc#4011 (comment)) because `config()` already returns `attestation_storage_fee_millinear` and an operator reads it once by hand, so a permanent public method and ABI entry earned nothing. The stale side was the stated contract, not the code: updated #4015's acceptance criteria and this PR's description to drop it.

crates/contract/src/lib.rs
```diff
@@ -832,6 +902,68 @@ impl MpcContract {
...
+                .insert(account_id.clone(), remaining);
+        }
+    }
+
+    /// Returns one grant to `account_id`, for an entry that has been reclaimed.
+    fn return_attestation_storage_grant(&mut self, account_id: &AccountId) {
+        let available = self.available_attestation_grants(account_id.clone());
+        self.attestation_grants
+            .insert(account_id.clone(), available.saturating_add(1));
+    }
```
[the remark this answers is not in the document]
barakeinav1: Fixed in 9375044b — `checked_add`, matching `prepay_attestation_storage`. At `u32::MAX` a saturating add would silently drop a grant that was genuinely returned; it now logs and leaves the counter alone.

crates/contract/tests/sandbox/utils/mpc_contract.rs
```diff
@@ -53,11 +53,76 @@ pub async fn get_tee_accounts(contract: &Contract) -> anyhow::Result<BTreeSet<No
...
+        {
+            Ok(result) => result.json::<u32>().unwrap_or(0),
+            Err(_) => 1,
+        };
+        if granted == 0 {
+            prepay_attestation_grants(account, contract, account.id(), 1).await?;
+        }
+    }
+    submit_participant_info_raw(account, contract, attestation, tls_key).await
+}
```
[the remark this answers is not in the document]
barakeinav1: Fixed in 0a2a56e2 — the inner copy is gone; the comment lives once, on the branch that needs it.

crates/contract/src/lib.rs
```diff
@@ -832,6 +902,68 @@ impl MpcContract {
...
+        }
+        Ok(())
+    }
+
+    /// Consumes one grant for `account_id`. Infallible by construction: every caller
+    /// has just established, in this same receipt, that a grant is available.
+    fn consume_attestation_storage_grant(&mut self, account_id: &AccountId) {
+        let available = self.available_attestation_grants(account_id.clone());
+        let remaining = available.saturating_sub(1);
+        if remaining == 0 {
```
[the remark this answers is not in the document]
barakeinav1: Keeping `saturating_sub` here, deliberately. Both call sites evaluate `assert_attestation_storage_grant_available` three lines earlier in the same receipt, so it cannot be reached without a grant. A `debug_assert!` would be the usual guard against a future third caller, but `crates/contract/src/` has none today and `release-contract` builds with debug assertions off, so it would only ever fire in tests — where the invariant already holds.

barakeinav1:
Went through all of these. Two were already stale, three fixed, one deferred with reasoning, and the non-blocking list mostly done. Everything is in the single squashed commit.

> **The e2e harness was not updated and the `MPC E2E tests` job is red on this PR.** … Fix by mirroring the operator step in the harness: after `init`, call `prepay_attestation_storage` for **every** node account, not just the participants

Fixed, and your diagnosis was exactly right including the "every node, not just participants" part — a node joining by resharing attests from its own process with a function-call key, so its grant has to exist beforehand. Took several rounds: the fee could not be read from `config()` there (the prepayment runs while the cluster is still starting, before the contract has state, so the view panics with "Calling default not allowed"), and the upgrade-compatibility tests run a production binary with no `prepay_attestation_storage`, so a method-not-found response now skips the prepayment instead of failing the cluster.

> **Doc drift on the funding model: the claim this PR inverts is still asserted in six places**

Fixed, all six. `submit_participant_info`, `resolve_verification` and `verify_post_dcap_and_store` now describe the grant; `tee_state::clean_invalid_attestations` says it returns the owners; `securing-mpc-with-tee-design-doc.md` and `attestation-verifier-contract.md` both updated. The ABI snapshot is re-accepted, since the first two strings are embedded in it. The README's User API also gained rows for both new methods.

> `submit_participant_info__should_store_new_entry_with_zero_deposit` and its doc … now claim something false; the test passes only because the shared helper silently prepays

Fixed. Renamed to `…should_store_a_new_entry_against_a_prepaid_grant`, with a separate account prepaying explicitly and an assertion that the grant reaches zero. Worth noting the "zero deposit" half was never wrong — it refers to what the *node* attaches, which is the point of the design; what was wrong was the doc claiming the storage was free.

You also caught a real flaw underneath this: both harnesses prepaid whenever an account had no grant, including for re-attestations that need none. Since participants already hold sentinel entries from init, every setup submission left a stray unused grant — which is why the function-call-key sandbox test was passing without ever prepaying. Both now prepay only when a submission would actually consume a grant.

> **Dangling reference: the comment points at `docs/design/operator-prepaid-attestation-storage.md`, which does not exist**

Stale — #4011 merged and the doc is on `main`, pulled in here by a merge from main.

> **The per-removal grant write pushes the reshare-time sweep below the current fleet size.**

Deferred to #4035 deliberately, and the severity is lower than we both had it. `clean_invalid_attestations` is permissionless and takes `max_scan` as a **parameter**, so an external caller is bound by neither `RESHARE_CLEAN_INVALID_ATTESTATIONS_MAX_SCAN` nor `clean_invalid_attestations_tera_gas` — those govern only the promise `vote_reshared` schedules. One manual call at ~300 TGas clears roughly 550 entries, so a backlog is never stuck and the worst this PR contributes is a manual invocation. Details and a correction to my own framing are on #4035.

### Non-blocking

- **Loose `Err(_)` assertion** — fixed, pins `TlsKeyOwnedByOtherAccount`. You were right that it proved nothing: the store rejects a foreign key too, so the test held even with the precondition removed.
- **Fee of `0`** — left permitted by decision: the value is governance's to choose, and zero is a legitimate (if unwise) choice.
- **"Rule 3:" naming** — fixed, states the behaviour instead.
- **Sentinels also mint grants** — correct, and sharper than the PR body was. Acknowledged, no change; still bounded by participant count.
- **Owned `AccountId` forcing clones** — fixed with a private `grants_for(&AccountId)` that the view delegates to; the view keeps an owned parameter because NEAR deserialises view arguments by value.
- **Intra-doc links** — converted the three references that name items. But note the rot detection does not actually apply here: near-sdk refuses to build this crate outside wasm32, so `cargo doc` skips it and `check-docs` never sees these links. Verified by inserting a link to a nonexistent method and watching `cargo doc --workspace` exit 0. The paths were checked by hand.
- **Typo in `account_id` burns the deposit** — said outright in the guide now, along with the current fee in NEAR (which also closes out the substance of #903).

barakeinav1:
@claude review

barakeinav1:
@copilot lite review

barakeinav1:
Addressed in e0b4e8c3 and 1eac9452. Replying point by point.

---

> `crates/contract/src/lib.rs:2511` — **The callback half of the grant logic has no test.** Both new behaviours in `verify_post_dcap_and_store` — the re-check guarding against the grant being spent between the submit receipt and the callback receipt, and the consume on `NewlyInsertedParticipant` — can be deleted without turning the suite red.

Agreed, and fixed with the two tests you sketched, on the existing `dstack_verification_setup()`:

- `resolve_verification__should_consume_one_grant_when_the_entry_is_new` — asserts the counter reaches 0.
- `resolve_verification__should_store_nothing_when_the_grant_was_spent_before_the_callback` — removes the grant from the fixture, asserts `PromiseOrValue::Promise(_)` and that nothing was stored.

Since your objection was specifically that the logic could be deleted without the suite going red, I checked that against each test rather than just watching them pass: deleting the `consume` call fails the first, deleting the re-check fails the second.

---

> `crates/e2e-tests/src/cluster.rs:1314` — "the prepayment happens while the cluster is still starting, before the contract has state, and a view call there panics with `Calling default not allowed`". The call site is `cluster.rs:1361`, immediately after `anyhow::ensure!(outcome.is_success(), "init failed: …")` in the same function, so the contract *does* have state […] The real constraint looks like the upgrade-compatibility clusters: they boot a production WASM whose `config()` carries no `attestation_storage_fee_millinear`. Either state that reason, or read `config()` and fall back to the constant when the field is absent, which also removes the silent-desync risk the rest of the comment warns about.

You were right that the comment was wrong, and right that the call site sits after a successful `init`. But the cause you proposed doesn't fit either: a missing field would surface as an error from the `.context(...)` on the deserialize, not the contract panic that was actually observed, and it wouldn't explain all 24 e2e tests failing rather than only the upgrade-compatibility ones.

Rather than swap one unverified *why* for another, I went and found it. It is a **finality mismatch**. `DeployedContract::call` uses near-kit's `.send()` with no `wait_until`, whose default is `ExecutedOptimistic` (`TransactionSend<W: WaitLevel = ExecutedOptimistic>`, near-kit 0.12.1 `src/client/transaction.rs:1483`), so `init` returns before its block is final. `DeployedContract::view` goes to `ViewCall`, whose `block_ref` is `BlockReference::default()` (`src/client/query.rs:347`) — and that default is `Finality::Final` (`src/types/block_reference.rs:32-36`). The view therefore executes against the last finalized block, still pre-`init`; `state_read()` returns `None`, near-sdk falls back to `MpcContract::default()`, and that is the `env::panic_str("Calling default not allowed.")` at `crates/contract/src/lib.rs:158`. Deterministic, hence all 24 tests.

Reproduced with a throwaway probe against a real sandbox and the real contract wasm:

```
[2] init is_success=true failure=None
[3] config() AFTER init, DEFAULT finality (Final) -> Err(… "Calling default not allowed.")
[4] config() AFTER init, Finality::Optimistic     -> Ok(config returned)
[5] config() 3s LATER,  DEFAULT finality (Final)  -> Ok(config returned)
```

Lines 3 and 4 isolate it to the block reference alone — same binary, same moment, only the finality differs. Line 5 shows the same default-finality read succeeding once finality catches up. Line 4 is also what rules out the production-WASM explanation: the identical deployment answers correctly at the same height.

So I took your second option rather than the first. Added `DeployedContract::view_optimistic` and the fee is now read from `config()` through it, which removes the duplicated constant and the desync risk entirely. It returns `None` when `config()` carries no fee field, and that is now how a binary predating grants is recognised and prepayment skipped. Verified on the real path: `view_optimistic("config")` immediately after `init` returns `attestation_storage_fee_millinear = 20`.

---

> `crates/contract/tests/sandbox/tee_verifier.rs:58` — "Prepays a grant, then submits, returning only the submission's result", on a `submit_dstack` whose body this PR changes to `submit_participant_info_raw` […] Describe what the function does now and leave the payer-separation rationale on the call sites, where it already is.

Agreed. The docstring now says it submits without prepaying and that callers prepay separately from a different account.

---

> `crates/contract/README.md:315` — the `submit_participant_info` row still reads "Submits the tee participant info for a potential candidate. c.f. TEE section", with no mention that a new entry now requires a prepaid grant […] Raised last round and not picked up.

Fixed. The row now states that storing a new entry consumes a grant and needs `prepay_attestation_storage`, and that re-submitting a TLS key the account already owns does not.

---

> `crates/contract/tests/inprocess/attestation_submission.rs:1088` — `assert!(removed >= 1)` […] `assert_eq!(removed, 1)`.

Done, with the reason in the message: `assert_eq!(removed, 1, "only the one stored entry should be swept")`.

---

B5 from the previous round (the sweep gas budget) remains deliberately out of scope here. It is tracked in #4035, whose title and body I have since corrected — "exhaustion is silent" was imprecise, since the failed receipt does appear in the block's execution outcomes and the method logs its invocation on entry; what is missing is anyone watching it.

crates/contract/src/lib.rs
```diff
@@ -771,15 +773,75 @@ impl MpcContract {
...
+    pub fn available_attestation_grants(&self, account_id: AccountId) -> u32 {
+        self.grants_for(&account_id)
+    }
+
+    fn grants_for(&self, account_id: &AccountId) -> u32 {
+        self.available_attestation_grants
+            .get(account_id)
+            .copied()
+            .unwrap_or(0)
+    }
```
netrome, line 787: Nit: I'm not sure it's worth breaking out the `grants_for` method here. Could be worth inlining it for now. But not a strong opinion. I think this reads pretty well as well.

crates/contract/src/lib.rs
```diff
@@ -771,15 +773,75 @@ impl MpcContract {
...
+            return Err(InvalidParameters::UnexpectedDeposit { attached, required }.into());
+        }
+
+        let available = self.grants_for(&account_id);
+        let credited =
+            available
+                .checked_add(grants)
+                .ok_or(InvalidParameters::MalformedPayload {
+                    reason: "grant counter would overflow".to_string(),
+                })?;
```
netrome, line 826: Bonus points for anyone who manages to trigger this error 😂

crates/contract/src/lib.rs
```diff
@@ -771,15 +773,75 @@ impl MpcContract {
...
+        self.available_attestation_grants
+            .insert(account_id.clone(), credited);
+
+        log!(
+            "prepay_attestation_storage: account_id={account_id}, grants={grants}, available={credited}"
+        );
```
netrome: I'm not sure if we need the logs here. I guess this mimics existing patterns, but is there any intention for these logs? Otherwise I'd default to not emitting transaction logs until someone explicitly asks for them.

crates/contract/src/lib.rs
```diff
@@ -854,6 +923,74 @@ impl MpcContract {
...
+    /// already returns `attestation_storage_fee_millinear`, and an operator reads it once
+    /// by hand. A dedicated view can be added later if operators ask for one.
+    fn attestation_storage_fee_internal(&self) -> NearToken {
+        NearToken::from_millinear(u128::from(self.config.attestation_storage_fee_millinear))
+    }
+
+    /// Whether storing `tls_public_key` for `account_id` would create a new entry and
+    /// therefore consume a grant. `Err` if the key belongs to somebody else, which the
+    /// store would reject anyway — checking it here keeps that failure ahead of
+    /// verification instead of after it.
```
netrome: Nit: I don't think this docstring adds anything that isn't obvious from the function name. I'd clear it out.

```suggestion
```

crates/contract/src/lib.rs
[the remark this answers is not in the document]
netrome: Okay but any reason not to go with the .expect then? In the event that a future call site would forget to do this check, I'd rather have a panic than silently allowing an unpaid grant.

netrome (dismissed):
A few notes from my side, but nothing strictly blocking.

I would however prefer the .expect over the saturating subtraction to be more defensive in the consume method.

crates/contract/src/lib.rs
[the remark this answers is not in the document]
barakeinav1: Agreed, done in fb7297ee — `checked_sub(1).expect(...)`.

For the record on reachability: no flow reaches it without a grant today. Both call sites check in the same receipt, and the check and the insert cannot disagree because both key on `tls_public_key` — `attestation_owner` returning `None` (needs a grant) is exactly the case where `previous.is_none()` gives `NewlyInsertedParticipant`, and a key the caller already owns yields `UpdatedExistingParticipant`, which never consumes. Two concurrent dstack submissions on one grant are handled by the callback re-check, not by this.

So this is a tripwire for a future call site rather than a live case, which is what makes it worth having: the panic reverts the receipt, whereas saturating at zero would store an entry no grant paid for. `engineering-standards.md` §Don't panic names the contract as the place where panicking is the correct choice for exactly that reason.

crates/contract/src/lib.rs
barakeinav1: Fair — removed in fb7297ee. The receipt already carries the predecessor, the args and the deposit; the only thing the log added was the resulting total, which `available_attestation_grants` returns on demand. No intention behind it beyond mirroring the existing pattern, so it goes.

crates/contract/src/lib.rs
barakeinav1: Removed in fb7297ee.

crates/contract/src/lib.rs
barakeinav1: Kept for now — it has five call sites (the public view, the precondition, consume, return, and prepay). Inlining would repeat `.get().copied().unwrap_or(0)` five times, and the alternative of having the private helpers call the public view needs an owned `AccountId` clone at each one, which is what an earlier revision of this PR did and I took out. Happy to fold it in if you feel differently.

crates/contract/src/lib.rs
barakeinav1: 😄

crates/contract/src/lib.rs
```diff
@@ -771,15 +773,75 @@ impl MpcContract {
...
+            .into());
+        }
+
+        let required = self
+            .attestation_storage_fee_internal()
+            .as_yoctonear()
+            .checked_mul(u128::from(grants))
+            .ok_or(InvalidParameters::MalformedPayload {
+                reason: "requested grants overflow the fee calculation".to_string(),
+            })?;
```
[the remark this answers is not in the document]
barakeinav1: Correct on the mechanics — a configured fee of 0 makes `required` 0, so grants could be minted with no deposit. Declining deliberately: the fee is a governance-votable `Config` field, and voting it to zero *is* a vote to stop charging for attestation storage. Guarding against it in the contract would override that choice rather than protect it. Participants should not set it to zero, but it is their call, not the contract's.

The two suppressed comments in the same review:

- **`tests/sandbox/utils/mpc_contract.rs:119`** — agreed and fixed in fb7297ee. `Err(_) => 1` did mask real failures: an uninitialised-contract panic there silently skipped prepayment and then surfaced as a confusing `NoAttestationStorageGrant` from the submission. It now only treats a missing method as a pre-grants binary and propagates everything else, matching what the e2e cluster helper already did.
- **`lib.rs:2029`, a sweep gas-regression guard test** — reasonable, but already acceptance criterion #2 on #4035 ("a gas-budget guard test pins the relationship, in the style of #3936"), so it belongs there rather than here. Worth noting the constants bound only the promise `vote_reshared` schedules: `clean_invalid_attestations` is permissionless and takes `max_scan` as a parameter, so an external caller is bound by neither.

crates/contract/src/lib.rs
netrome: Fair point. And using the separate private function rather than calling the public function probably makes sense there.

crates/contract/src/lib.rs
netrome: Thanks!

crates/contract/src/lib.rs
netrome: Thanks!

netrome (dismissed):
Thanks for updating 🙏

crates/contract/src/config.rs
```diff
@@ -23,6 +23,10 @@ const DEFAULT_RETURN_CK_AND_CLEAN_STATE_ON_SUCCESS_CALL_TERA_GAS: u64 = 7;
 const DEFAULT_FAIL_ON_TIMEOUT_TERA_GAS: u64 = 2;
 /// Prepaid gas for a `fail_attestation_submission` call
 const DEFAULT_FAIL_ATTESTATION_SUBMISSION_TERA_GAS: u64 = 2;
+/// Fee, in milliNEAR, for one attestation-storage grant. Covers the worst-case
+/// stored entry plus the grant-counter row it creates, with headroom for layout
+/// growth; see `docs/design/operator-prepaid-attestation-storage.md`.
+const DEFAULT_ATTESTATION_STORAGE_FEE_MILLINEAR: u64 = 20;
```
gilcu3: nit: the above header
> // --- Gas Defaults (in TeraGas) ---

 is now off, as there is a value that is not gas. We could remove it.

 Also this comment could be reduced to:
```
/// One attestation-storage grant in milliNEAR
```

crates/contract/src/storage_keys.rs
```diff
@@ -34,4 +34,5 @@ pub enum StorageKey {
     ForeignChainMetadata,
     TeeVerifierVotesByVoter,
     TeeVerifierVotesByProposal,
+    AttestationGrantsV1,
```
gilcu3: nit: we probably already have some inconsistency here, but I think it is better when we introduce a new key to avoid the `V1` suffix, as anyway we might never need `V2`. It was done this way for the last 4 values added here for example

crates/contract/tests/sandbox/utils/mpc_contract.rs
```diff
@@ -62,11 +62,81 @@ pub async fn get_tee_accounts(contract: &Contract) -> anyhow::Result<BTreeSet<No
...
 }
 
+/// Prepays `grants` attestation-storage grants for `beneficiary`, signed by `payer`.
+/// Permissionless, so any deposit-capable account may fund any other.
+pub async fn prepay_attestation_grants(
+    payer: &Account,
+    contract: &Contract,
+    beneficiary: &AccountId,
+    grants: u32,
+) -> anyhow::Result<ExecutionFinalResult> {
```
gilcu3: nit: this comment does not tell me anything that the signature does not say already

crates/contract/tests/sandbox/utils/mpc_contract.rs
```diff
@@ -62,11 +62,81 @@ pub async fn get_tee_accounts(contract: &Contract) -> anyhow::Result<BTreeSet<No
...
+    // Read the fee the way an operator does: from `config()`. There is deliberately no
+    // dedicated view for it.
+    let config: serde_json::Value = contract
+        .view("config")
+        .args_json(serde_json::json!({}))
+        .await?
+        .json()?;
+    let fee_millinear = config["attestation_storage_fee_millinear"]
+        .as_u64()
+        .expect("config must carry attestation_storage_fee_millinear");
```
gilcu3: nit: why using serde `Value` when we have a type for this?

crates/contract/tests/sandbox/tee_verifier.rs
gilcu3, line 63: if we need to change this function calls everywhere, I would rather use the prepay deposit function explicit, instead of making it implicit in `submit_participant_info` and then having to explain everywhere

crates/contract/tests/sandbox/tee_verifier.rs
```diff
@@ -54,8 +55,12 @@ async fn deploy_and_trust_verifier(
     trust_verifier(contract, participants, verifier.id()).await;
 }
 
+/// Submits without prepaying: callers that need a grant prepay separately, from a different
+/// account. Keeping the two apart keeps the balance assertions below about the submission
+/// alone -- the prepayment legitimately costs the fee, whereas a *failed* submission must
+/// cost nothing but gas.
```
gilcu3: nit: we could remove this comment, by default submit should not add a deposit, as that matches the node behavior

crates/contract/src/v3_13_0_state.rs
gilcu3, line 1: We cannot modify this file as we already upgraded to 3.14, so this is a blocker for me until #4080 land

crates/e2e-tests/src/blockchain.rs
```diff
@@ -234,6 +234,23 @@ impl DeployedContract {
...
+    pub async fn view_optimistic<T: DeserializeOwned + Send + 'static>(
+        &self,
+        method: &str,
+    ) -> anyhow::Result<T> {
+        self.client
+            .view::<T>(&self.contract_id, method)
+            .finality(near_kit::Finality::Optimistic)
+            .await
+            .map_err(|e| anyhow::anyhow!("contract view `{method}` (optimistic) failed: {e}"))
+    }
```
gilcu3: I am against introducing this here, I think it could bring back flaky tests. Waiting 1 or 2 blocks here is fine.

docs/design/operator-prepaid-attestation-storage.md
```diff
@@ -44,8 +44,8 @@ An operator reads the current fee from `config()` and attaches the exact multipl
...
 
 1. **Entry exists and this account owns it** — re-attestation. Updates in place, consumes nothing. Keys on *ownership*, not key presence: a presence-only test would classify someone else's key as "no grant needed".
-2. **New entry** — consume one grant; reject if the account has none.
-3. **Entry reclaimed** by `clean_invalid_attestations` — return one grant to its owner. There is exactly one removal site, so the counter cannot drift. This adds a grants-map write per removed entry — a row *insert*, not an update, whenever the owner's row was deleted at zero — inside a gas-bounded sweep, so `clean_invalid_attestations_tera_gas` and `RESHARE_CLEAN_INVALID_ATTESTATIONS_MAX_SCAN` need re-validating against the heavier per-entry cost.
+2. **New entry** — consume one grant; reject if the account has none. On the async `Dstack` path this is checked twice: once here, and again in `resolve_verification`, which runs a later receipt in which the grant may since have been consumed. Without the second check, two submissions from one account could both clear the first check and then store two entries against a single grant.
```
gilcu3, line 47: hmm this seems more complicated than expected then. Should be fine but I need to think a bit on it

docs/running-an-mpc-node-in-tdx-external-guide.md
```diff
@@ -682,6 +682,45 @@ near account create-account sponsor-by-faucet-service <ACCOUNT_NAME> use-manuall
...
+The node cannot pay for this itself. Its access key is a function-call key, and the NEAR protocol forbids those from attaching a deposit — so the operator prepays, using the full-access key of the account you just created.
+
+Do this now, before the CVM is started. `prepay_attestation_storage` needs only the account id, so nothing waits on the node's keys, and the node will not start up into a state where its attestation submissions are rejected.
+
+First read the current fee, `attestation_storage_fee_millinear`, in milliNEAR:
+
+```bash
+near contract call-function as-read-only \
+  v1.signer config json-args '{}' network-config mainnet now
+```
```
gilcu3, line 701: we could pipe this with `jq` so that the output is the single config the operator needs

crates/e2e-tests/src/cluster.rs
```diff
@@ -1268,6 +1268,68 @@ struct InitContractArgs {
...
+/// Fee for one attestation-storage grant, read from the contract the way an operator reads it.
+///
+/// `None` when `config()` carries no fee field, which means a contract binary older than
+/// grants: the upgrade-compatibility tests run one, and nothing there needs prepaying.
+///
+/// Kept behind a wrapper so the source of the value can change without touching callers.
+async fn attestation_storage_fee(
+    contract: &DeployedContract,
+) -> anyhow::Result<Option<near_kit::NearToken>> {
+    let config: serde_json::Value = contract.view_optimistic("config").await?;
```
gilcu3: why do we need optimistic here?

crates/e2e-tests/src/cluster.rs
```diff
@@ -1268,6 +1268,68 @@ struct InitContractArgs {
...
+) -> anyhow::Result<()> {
+    let Some(fee) = attestation_storage_fee(contract).await? else {
+        tracing::info!("contract config has no attestation storage fee; skipping prepayment");
+        return Ok(());
+    };
+
+    // Granting the initial participants too is harmless: their attestations update the
+    // sentinel entry written at init, so they consume nothing and the grant stays unspent.
+    for (i, near_key) in near_keys.iter().enumerate() {
+        let account = format!("node{i}.{SANDBOX_ROOT_ACCOUNT}");
```
gilcu3: this seems like it should come from somewhere, else it could diverge at any point. The accounts that are going to be part of the network should be defined in a single place.

crates/e2e-tests/src/cluster.rs
```diff
@@ -1268,6 +1268,68 @@ struct InitContractArgs {
...
+        let client = blockchain.client_for(&account, near_key).unwrap();
+        let outcome = contract
+            .call_from_with_deposit(
+                &client,
+                "prepay_attestation_storage",
```
gilcu3: We have constants for contract methods, should not be used inline

crates/e2e-tests/src/cluster.rs
```diff
@@ -1308,6 +1370,8 @@ async fn init_contract(
         outcome.failure_message()
     );
 
+    prepay_attestation_grants(blockchain, contract, &near_keys).await?;
+
```
gilcu3, line 1381: I think this is exactly the correct pattern. We should use the same in the sandbox tests.

crates/e2e-tests/src/cluster.rs
```diff
@@ -1268,6 +1268,68 @@ struct InitContractArgs {
...
+            // Fallback for a binary that predates grants: the fee read above already skips
+            // those, so reaching here means `config()` carried a fee the contract cannot
+            // charge. Nothing needs prepaying either way, so stop rather than fail the cluster.
+            anyhow::ensure!(
+                failure.contains("method not found"),
+                "prepay for node {i} failed: {failure}"
+            );
+            tracing::info!("contract has no prepay_attestation_storage; skipping grant prepayment");
+            break;
+        }
```
gilcu3, line 1321: which case is this one?

crates/contract/tests/inprocess/attestation_submission.rs
```diff
@@ -149,7 +149,35 @@ struct TestSetup {
...
+                && stored.account_id == node_id.account_id
+        });
+        if !already_owns_entry
+            && self
+                .contract
+                .available_attestation_grants(node_id.account_id.clone())
+                == 0
+        {
+            self.prepay_grants(&node_id.account_id, 1);
+        }
```
gilcu3: this logic should be removed, same as for the sandbox tests. We can submit deposits explicitly during initialization or inline in the tests that need it

crates/contract/tests/inprocess/attestation_submission.rs
```diff
@@ -265,13 +293,54 @@ fn submit_participant_info__should_reject_overwrite_from_other_account() {
...
+            .unwrap()
+            .is_none()
+    );
+}
+
+/// Once an operator has prepaid, the node self-submits with no deposit of its own and the
+/// grant is consumed.
 #[test]
-fn submit_participant_info__should_store_new_entry_with_zero_deposit() {
+fn submit_participant_info__should_store_a_new_entry_and_consume_one_grant() {
```
gilcu3: nit: comment seems redundant given test name

crates/contract/tests/inprocess/attestation_submission.rs
```diff
@@ -802,3 +946,167 @@ fn vote_code_hash_works_in_contract_protocol_states(#[case] state: ContractProto
...
 }
+
+/// The deposit must be exactly `fee × grants`. Anything else is rejected, which is what
+/// makes a remainder — and therefore a refund path — impossible.
+#[rstest]
+#[case::one_yocto_short(-1)]
+#[case::one_yocto_over(1)]
+fn prepay_attestation_storage__should_reject_a_deposit_that_is_not_an_exact_multiple(
+    #[case] offset: i128,
+) {
```
gilcu3: shouldn't something like this just be a unit test? or there is something that prevents that?

crates/contract/src/lib.rs
```diff
@@ -188,6 +188,8 @@ pub struct MpcContract {
     // non-optional via a migration that requires it be set.
     tee_verifier_account_id: Option<AccountId>,
     tee_verifier_votes: TeeVerifierVotes,
+    /// A row is removed at zero, so the map holds no entry for an account with none.
+    available_attestation_grants: LookupMap<AccountId, u32>,
```
gilcu3: it is not very clear to me when to put the new data in the higher level, or lower, for example inside `TeeState`. Did you decide to put it here for any particular reason?

crates/contract/src/lib.rs
gilcu3: I am not really following, why does it make sense to use the private function if using the public function is equivalent? I mean why a single:
```
pub fn available_attestation_grants(&self, account_id: AccountId) -> u32 {
        self.available_attestation_grants
            .get(account_id)
            .copied()
            .unwrap_or(0)
}
```

is not enough here? Both functions are new so I am not sure why the number of callers is important

crates/contract/src/lib.rs
```diff
@@ -815,18 +874,25 @@ impl MpcContract {
             account_public_key,
         };
 
+        // Fail fast: a submission that would create a new entry without a grant, or that
+        // targets a TLS key owned by another account, is rejected before any verification.
+        self.assert_attestation_storage_grant_available(
+            &node_id.account_id,
+            &node_id.tls_public_key,
+        )?;
```
gilcu3: I get it is good to fail fast, but also it creates a duplication for this check as you pointed out elsewhere. I am a bit torn about that, but cannot claim removing this is better.

gilcu3 (changes requested):
This looks mostly good, but I have a few blockers:

- The contract migration is being updated but we released 3.14 already, so we need to do the cleanup first (it is almost done in another PR)
- In the sandbox tests we are making paying the deposit part of `submit_participant_info`. I think this is not correct as that operation should be done independently by the operator anyway, so our tests that need it should do it explicitly, or as part of test setup. The decision taken here led to the same amount of code changes, but made the tests harder to read/create tests because they have implicit assumptions about this, now explained with verbose comments. 
(In this PR e2e tests modifications we already use the pattern I mentioned above BTW)

- The `view_optimistic` method was added in e2e tests to save some blocks in tests. I think this is not good as it could introduce flakiness. I don't see anything wrong with waiting for Finalized blocks.

- We are introducing new contract methods, but they are not being added in `crates/near-mpc-contract-interface/src/method_names.rs`

- Many tests added in `crates/contract/tests/inprocess/attestation_submission.rs` appear to be unit tests checking corner cases. I am not sure if there is any restriction that makes then not possible as unit tests, we should check that.

Overall there are many verbose comments, I think we should do some work to reduce them so that they become useful and not redundant

crates/contract/src/v3_13_0_state.rs
barakeinav1: Done. #4081 landed a few minutes after your review, so I merged main and moved the grants map init to `v3_14_0_state.rs`.

@gilcu3 please check I did this correctly.

crates/contract/src/config.rs
barakeinav1: Right, it was in the middle of the gas block. Moved it below and cut the comment to one line. Kept the header since it is correct for the gas constants above it.

crates/contract/src/storage_keys.rs
barakeinav1: Done, dropped the suffix.

crates/contract/tests/sandbox/utils/mpc_contract.rs
barakeinav1: Removed.

crates/contract/tests/sandbox/utils/mpc_contract.rs
barakeinav1: No good reason. Switched to the typed `Config`.

crates/contract/tests/sandbox/tee_verifier.rs
barakeinav1: Removed.

crates/contract/tests/sandbox/tee_verifier.rs
barakeinav1: Took a middle path here. The helper has 18 call sites, almost none of which care about grants, so prepaying explicitly at each one is more code, not the same. But you are right that hiding it in something called `submit_participant_info` is the problem, so I renamed it to `prepay_and_submit_participant_info` and deleted the comments that were explaining the hiding. `submit_participant_info_raw` stays for tests that assert on grants or on rejection.

barakeinav1:
All addressed in b6399387 (merge + config) and 6627a35d.

> The contract migration is being updated but we released 3.14 already, so we need to do the cleanup first (it is almost done in another PR)

#4081 landed a few minutes after your review, so I merged main and moved the grants map init to `v3_14_0_state.rs`. More on this below — it was a real blocker for a reason neither of us had spotted.

> In the sandbox tests we are making paying the deposit part of `submit_participant_info`. I think this is not correct as that operation should be done independently by the operator anyway, so our tests that need it should do it explicitly, or as part of test setup.

Middle path, see [the thread](https://github.com/near/mpc/pull/4016#discussion_r3727091927). The helper has 18 call sites and almost none care about grants, so explicit prepayment at each is more code, not the same amount. But you are right that hiding it inside something called `submit_participant_info` is the problem, so it is now `prepay_and_submit_participant_info` and the comments explaining the hiding are gone. Say the word and I will do the 18.

> The `view_optimistic` method was added in e2e tests to save some blocks in tests. I think this is not good as it could introduce flakiness. I don't see anything wrong with waiting for Finalized blocks.

Dropped. `init` now waits for a final block and the config read uses the normal `view`, so no optimistic read is left. Costs a block or two once per cluster start, which is nothing.

> We are introducing new contract methods, but they are not being added in `crates/near-mpc-contract-interface/src/method_names.rs`

Added both.

> Many tests added in `crates/contract/tests/inprocess/attestation_submission.rs` appear to be unit tests checking corner cases.

I would rather keep them there. That file already holds 23 tests for this same entry point, and the grant cases use the same `TestSetup` (participants, protocol state) as their neighbours, so moving only mine to `src/` splits related cases across two layers. Two of them (the exact-deposit rstest and zero grants) genuinely do not need the harness — happy to move just those.

> Overall there are many verbose comments, I think we should do some work to reduce them so that they become useful and not redundant

Fair, you are the third reviewer to say it. Did a pass over the whole diff rather than only the lines you flagged, dropping the ones that restated a function name or repeated a nearby comment.

---

One thing worth flagging from the migration work: the new shadow state uses the current `Config`, so adding a field broke deserialization of deployed state — `migrate()` failed with "could not deserialize contract state: Unexpected length of input". The soft launch upgrade test caught it. Fixed with an `OldConfig` shadow for 3.14.0, and the fee field is now last in `Config` so the shadow stays a clean prefix. Same bug would have hit against 3.13 if #4081 had not landed first.

crates/e2e-tests/src/blockchain.rs
barakeinav1: > I think it could bring back flaky tests
why?

anyway, If you don't think make the CI a bit slow is problem, I'll change to finalize. it make the code simpler.

docs/design/operator-prepaid-attestation-storage.md
barakeinav1: I want to fail fast in grant is missing, but then we have this race condition. 
we can also add some counter for incoming submission instead, but I think this is the simplest approach

crates/e2e-tests/src/cluster.rs
barakeinav1: We do not. Dropped it in 6627a35d — `init` now waits for a final block and this reads `config()` with the normal view.

Background, since it looks arbitrary: the read failed with "Calling default not allowed" and I traced it to `call` returning at near-kit's default `ExecutedOptimistic` while `view` resolves against the last final block. Waiting on init is the better fix.

crates/e2e-tests/src/cluster.rs
barakeinav1: Fixed in ce0420b9. I had just added the constant in the same round and then not used it here.

crates/e2e-tests/src/cluster.rs
barakeinav1: Right. Added `node_account(i)` in ce0420b9 and pointed all six copies of that `format!` at it.

crates/e2e-tests/src/cluster.rs
barakeinav1: The upgrade-compatibility tests, which boot a contract binary that predates grants and has no `prepay_attestation_storage`. Nothing there needs prepaying, so we stop instead of failing the cluster.

The comment used to say this; my comment trimming cut it. Restored in ce0420b9.

crates/contract/tests/inprocess/attestation_submission.rs
barakeinav1: Removed in ce0420b9.

docs/running-an-mpc-node-in-tdx-external-guide.md
barakeinav1: Good idea, done in ce0420b9 — piped through `jq .attestation_storage_fee_millinear`.

crates/contract/src/lib.rs
barakeinav1: Grants are payment state, not attestation state, so `TeeState` felt like the wrong home: it holds what a node proved about itself, this holds what someone paid for. `prepay_attestation_storage` has nothing to do with attestations, and putting the map inside would have it reach through `self.tee_state` to credit a counter.

The reason (suggested by claude) is migration cost. `migrate()` needs a description of the old on-chain layout: `v3_14_0_state.rs` spells out `MpcContract` field by field, but for `TeeState` it just reuses the live type, which only works while that layout is unchanged since 3.14.0. So a field on `MpcContract` is one line, because the struct is already hand-copied there. A field inside `TeeState` breaks the "unchanged" assumption and forces us to hand-copy `TeeState` too, plus a conversion — which is what `v3_13_0_state.rs` carried the last time `TeeState` changed.

One argument against my choice, for the record: if the map lived in `TeeState`, the sweep could remove the entry and credit the grant in one place, instead of returning owners and trusting the caller to credit them. That is less drift risk than what we have.

crates/contract/src/lib.rs
barakeinav1: Same feeling here. The duplication is real, but the two checks answer different questions: the first keeps an ungranted submission from paying for a `verify_quote` round trip, the second is the only thing standing between two concurrent submissions and two entries on one grant. Dropping either loses something.

The alternative is reserving the grant at submit time and releasing it if the callback fails, which trades this duplication for state that has to survive a failing callback. Worse, I think.

crates/contract/src/lib.rs
barakeinav1: Fair, the caller count is  not the exactly the point. The reason I had was avoiding an `account_id.clone()` in each caller: the public method has to take `AccountId` by value, since near-sdk deserializes entry point args from JSON, while the internal callers all hold a `&AccountId`.

see claude-bot comment

>crates/contract/src/lib.rs:777 — available_attestation_grants takes an owned AccountId because it is a view, forcing account_id.clone() at all five internal call sites; a private fn grants(&self, &AccountId) -> u32 that the view delegates to removes them.

crates/e2e-tests/src/cluster.rs
barakeinav1: Done, sandbox now follows it — bb0251ef. `prepay_and_submit_participant_info` always prepays, no state inspection, and re-attestations use `submit_participant_info_raw`. The setup loop got the same split, so there is now `prepay_and_submit_tee_attestations` next to `submit_tee_attestations`.

Worth knowing what this turned up: `init` seeds a sentinel entry for every initial participant, so a participant submitting its *own* TLS key is an update that consumes nothing. `submit_attestations` was in that case, while `submit_tee_attestations` mostly gets freshly generated accounts and keys, which do need a grant. Two loops that look the same needed opposite treatment, which is why the old helper guessed at runtime.

I classified every call site by asserting inside the prepaying helper and running the suite, rather than by eye. Both directions are covered: a missing grant fails the submission, and a temporary assert confirmed nothing prepays for an entry that already exists.

crates/contract/tests/inprocess/attestation_submission.rs
barakeinav1: Removed in 9e63d0cc, same shape: `prepay_and_submit_attestation_for_node` always prepays, `submit_attestation_for_node` never does, call sites choose.

Two tests were being charged twice, by the way — they prepay explicitly, which is the point of the test, and then the helper prepaid again. The conditional was hiding that.

crates/contract/tests/inprocess/attestation_submission.rs
barakeinav1: Nothing prevented it. Moved both this and `should_reject_zero_grants` to unit tests in `src/lib.rs` (73b93cd9) — they only need a contract and a deposit context.

I left the rest in the in-process file, since those exercise participants and protocol state through the shared setup.

crates/contract/src/lib.rs
gilcu3: The argument about removing it is just that for a failed attestation to wait a bit for the second receipt is not that bad, as attestations do not really have strict timing requirements, but I am fine leaving as is.

crates/contract/src/lib.rs
gilcu3: > prepay_attestation_storage has nothing to do with attestations

this reads a bit self-contradictory :smile:  

> The reason (suggested by claude) is migration cost. migrate() needs a description of the old on-chain layout: v3_14_0_state.rs spells out MpcContract field by field

Yeah and this is exactly the wrong reason. Migration is something we do once, and its cost is negligible. If this is the only strong reason, then to keep the contract more organized it would probably make sense to have it inside `TeeState`. I don't feel strongly about this, but I do see a future huge refactor happening here, as things appear to be piling up at the highest contract level.

crates/contract/tests/sandbox/tee_verifier.rs
gilcu3: That's fine, although I still wouldn't use `submit_participant_info_raw`, but instead `submit_participant_info`. Notice that there is nothing raw about it, it does exactly what it should do, and what the node does, submit attestations.

crates/e2e-tests/src/cluster.rs
gilcu3: Yeah that's why most of the time it is better to use Final instead of Optimistic, unless it is in some hot path for performance, that was not the case here

crates/contract/tests/inprocess/attestation_submission.rs
gilcu3: > Two tests were being charged twice, by the way — they prepay explicitly, which is the point of the test, and then the helper prepaid again. The conditional was hiding that.

nice that the change was able to capture that!

crates/contract/src/lib.rs
gilcu3: I see, that's an unexpected reason but indeed valid. I wouldn't worry too much about cloning account_id, it is just a 64 chars string at worst, but I am fine with keeping this.

crates/contract/src/v3_13_0_state.rs
gilcu3: yes it looks correct to me

crates/contract/tests/sandbox/utils/mpc_contract.rs
```diff
@@ -62,7 +63,45 @@ pub async fn get_tee_accounts(contract: &Contract) -> anyhow::Result<BTreeSet<No
...
+    attestation: &Attestation,
+    tls_key: &Ed25519PublicKey,
+) -> anyhow::Result<ExecutionFinalResult> {
+    let prepayment = prepay_attestation_grants(account, contract, account.id(), 1).await?;
+    anyhow::ensure!(prepayment.is_success(), "prepayment failed: {prepayment:?}");
+    submit_participant_info_raw(account, contract, attestation, tls_key).await
+}
+
+/// Submits without prepaying anything.
+pub async fn submit_participant_info_raw(
```
gilcu3: as I mentioned in another comment, here the `raw` no longer makes sense, as this function is just doing what its name says, submitting an attestation

crates/contract/tests/sandbox/tee.rs
```diff
@@ -1051,7 +1072,13 @@ async fn submit_participant_info__should_store_new_entry_with_zero_deposit() ->
...
 
     // When
-    let result = submit_participant_info(
+    // Prepay explicitly, from a different account, so the test asserts the grant flow rather
+    // than relying on the shared helper to fund it silently.
+    let payer = worker.dev_create_account().await?;
+    let prepayment = prepay_attestation_grants(&payer, &contract, outsider.id(), 1).await?;
+    assert!(prepayment.is_success(), "prepayment failed: {prepayment:?}");
+
+    let result = submit_participant_info_raw(
```
gilcu3: do we still need to use a different account here? Now we can use a shared helper that does not (pre-)pay deposits

crates/contract/tests/sandbox/tee.rs
```diff
@@ -1066,6 +1093,12 @@ async fn submit_participant_info__should_store_new_entry_with_zero_deposit() ->
     );
     let stored = get_participant_attestation(&contract, &fresh_tls_key).await?;
     assert!(stored.is_some(), "the entry should be stored on-chain");
+    let remaining: u32 = contract
+        .view("available_attestation_grants")
```
gilcu3: this is another constant that must be defined in the contract interface. It is only a minor blocker, luckily easy to do

crates/contract/tests/inprocess/attestation_submission.rs
gilcu3, line 1: I still believe some of these tests could still be unit tests, but we can leave that as a follow up, to not overcomplicate this one

crates/e2e-tests/src/blockchain.rs
```diff
@@ -155,6 +155,22 @@ impl DeployedContract {
...
+        args: serde_json::Value,
+    ) -> anyhow::Result<FinalExecutionOutcome> {
+        self.client
+            .call(&self.contract_id, method)
+            .args(args)
+            .gas(MAX_GAS)
+            .wait_until::<Final>()
+            .await
+            .map_err(|e| anyhow::anyhow!("contract call `{method}` failed: {e}"))
+    }
```
gilcu3, line 172: thought we could just reuse the `call` function above this one, but that one is not Final. It feels we could organize this a bit better to avoid so much duplicate code and to ensure the caller knows exactly what they are calling, but for now this is fine.

crates/e2e-tests/src/cluster.rs
```diff
@@ -1270,6 +1275,63 @@ struct InitContractArgs {
...
+            let failure = format!("{:?}", outcome.failure_message());
+            // A contract binary predating grants, as run by the upgrade-compatibility tests:
+            // it has no `prepay_attestation_storage` and needs no prepayment.
+            anyhow::ensure!(
+                failure.contains("method not found"),
+                "prepay for node {i} failed: {failure}"
+            );
+            tracing::info!("contract has no prepay_attestation_storage; skipping grant prepayment");
+            break;
+        }
```
gilcu3, line 1321: thanks for adding the explanatory comment. I would go one step ahead and also mention what contract version will not need this, such that we can remove the code in the future without fear of breaking something

crates/contract/src/lib.rs
```diff
@@ -856,6 +920,68 @@ impl MpcContract {
         }
     }
 
+    /// Read from `config()` by operators; deliberately not its own view.
+    fn attestation_storage_fee_internal(&self) -> NearToken {
+        NearToken::from_millinear(u128::from(self.config.attestation_storage_fee_millinear))
+    }
```
gilcu3: the internal suffix seems odd. The function is private so we already know it is `internal`

crates/contract/src/lib.rs
```diff
@@ -856,6 +920,68 @@ impl MpcContract {
...
+        }
+        Ok(())
+    }
+
+    /// Consumes one grant for `account_id`.
+    ///
+    /// Panics if none is available. Every caller has established, in this same receipt, that
+    /// a grant is there; panicking reverts the receipt, so an invariant broken by a future
+    /// call site costs a failed transaction rather than an entry nobody paid for.
+    fn consume_attestation_storage_grant(&mut self, account_id: &AccountId) {
```
gilcu3: nit: we could leave out this part of the comment. The expect line already explains it well

crates/contract/src/lib.rs
```diff
@@ -856,6 +920,68 @@ impl MpcContract {
...
+    fn return_attestation_storage_grant(&mut self, account_id: &AccountId) {
+        let available = self.grants_for(account_id);
+        // Checked, not saturating: at `u32::MAX` a returned grant would be dropped silently.
+        let Some(returned) = available.checked_add(1) else {
+            log!("grant counter for {account_id} is saturated; not returning a grant");
+            return;
+        };
+        self.available_attestation_grants
+            .insert(account_id.clone(), returned);
+    }
```
gilcu3: I see a small foot gun here related to the fact that at 0 the entry is deleted, then in the function below apparently that would cause that the grant is not returned. This is not the case because of the `unwrap_or(0)` inside `grants_for` but I wonder if there is a better way. I have no explicit suggestion though 😿

gilcu3 (dismissed):
Thanks for the fixes!

Left some nits (and a tiny blocker about a contract method name constant but I am not blocking for that one)

crates/contract/src/lib.rs
netrome: Yeah, it feels like we should spend more time thinking how to organize the contract code. I guess you're about to do this soon @gilcu3 so probably could move this in the same go.

netrome: Hmm should we consider using an IterableMap here? LookupMaps have been difficult to work with since we cannot iterate over the keys, if we ever want to clean them up or migrate the inner structure.

netrome (dismissed):
Quick sweep, looks alright but I'm questioning if we want to use a LookupMap vs an iterable map for the attestation grants. Don't need to block merging this PR, but could consider changing this as a follow-up.

crates/contract/src/lib.rs
gilcu3: well this is very far from what I am about to do. The current idea is to refactor the code to reduce the huge `lib.rs` file. What is needed here is to re-org the contract state, which would potentially require a lot of code changes and migration.

gilcu3: In this case the good thing is that we know the keys, so we should always be able to clean it. But it won't be as nice as with IterableMap

gilcu3: changed anyway in [7392953](https://github.com/near/mpc/pull/4016/commits/7392953f5f9668a8de6efa9197dd7f697bac5f0d)

gilcu3 (dismissed):

gilcu3 (approved):

netrome (approved):

PR #4017, opened 2026-07-30
feat: add expected_network_fingerprint to foreign chain config

### LINKED ISSUE #4003: health-check: probe foreign chain RPC providers periodically (chain identity only)
### Background

Phase 1 of #3764. The node probes chain identity only, meaning a constant that is never pruned: `eth_chainId`, `starknet_chainId`, `getblockhash 0`, the Aptos ledger `chain_id`, the Sui chain identifier. That catches the failures operators actually hit: an unreachable endpoint, credentials not valid for that chain, the wrong network. Expected values come from config rather than being baked into the binary, so mixed network and local setups stay checkable.

The CLI keeps its golden route until phase 2 (#3969).

### User Story

As a node operator I want logs and metrics on whether each configured RPC provider is reachable, authenticated and on the expected network, from startup onwards, so that a provider which breaks while the node is up surfaces before a real verify request depends on it.

### Acceptance Criteria

- Every configured provider is probed at startup and repeatedly while the node runs, and reported pass or fail on its own, named by chain and provider.
- A provider that is unreachable, rejects the configured credentials, or serves a different network than expected is reported unhealthy.
- The expected chain identity is supplied by operator config. Shipped config templates come with the expected identities as placeholders

The node must not depend on the CLI crate, which carries clap, serde_yaml and toml.

### Slices

- [x] Starknet (#4095)
- [x] EVM chains (#4096)
- [x] Bitcoin (#4091)
- [x] Aptos (#4092)
- [x] Sui (#4093)
- [ ] Node wiring (#4094)

haiyuechen-nearone:
Part of #4003.

Config only: adds the field, seeds the shipped templates, documents it. Reading it and probing providers comes in follow up PRs.

### Notes for review

- **`expected_network_fingerprint`, not `expected_chain_id` or `expected_chain_identity`.** The value is a chain id only for some chains; Bitcoin's is a genesis block hash, Sui's a genesis checkpoint digest. Fingerprint is the one word that covers all of them without colliding with the chains that do have a chain id.

- Optional, and not validated in `ForeignChainsConfig::validate()`.

- Always a quoted string, including the values that look numeric.

- The parsing tests now share one config fixture instead of fourteen copies of the same preamble. The fingerprint happy path rides along with `config_parsing__should_succeed_when_foreign_chains_are_set`; the absent case stays standalone, since it is what proves configs written before the field still parse.


haiyuechen-nearone:
@claude review

haiyuechen-nearone:
@claude review

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
```diff
@@ -54,106 +55,66 @@ struct MockServerUrls {
...
-                        token: TokenConfig::Val {
-                            val: QUERY_AUTH_TOKEN.to_string(),
-                        },
-                    },
+        bitcoin: Some(mock_chain(
+            &urls.bitcoin,
+            AuthConfig::Path {
+                placeholder: PATH_AUTH_PLACEHOLDER.to_string(),
+                token: TokenConfig::Val {
+                    val: PATH_AUTH_API_KEY.to_string(),
```
netrome, line 88: Nice to get rid of some code repetition here

netrome (dismissed):
I think @gilcu3 raised a good point on slack that "chain_identity" and "chain_id" meaning different things can be confusing. Perhaps could be worth finding a better name for this. Something like "chain_unique_value" perhaps? (Not super happy with that name, could be better options, but something conveying this being a unique value associated with a chain that's not necessarily the same as the chain ID if the chain already has that concept)

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
```diff
@@ -54,106 +55,66 @@ struct MockServerUrls {
...
-            providers: common::build_providers_from_urls(&urls.polygon, "polygon"),
-        }),
+            },
+        )),
+        arbitrum: Some(mock_chain(&urls.arbitrum, Default::default())),
+        hyper_evm: Some(mock_chain(&urls.hyper_evm, Default::default())),
+        polygon: Some(mock_chain_with_providers(
+            common::build_providers_from_urls(&urls.polygon, "polygon"),
+        )),
         ..Default::default()
```
gilcu3, line 118: not in this PR, but should we remove this line
>..Default::default()
 
so that we cannot miss adding this in the future for new chains?

crates/node-config/src/foreign_chains.rs
```diff
@@ -46,6 +46,10 @@ pub struct ForeignChainsConfig {
 pub struct ForeignChainConfig {
     pub timeout_sec: NonZeroU64,
     pub max_retries: NonZeroU64,
+    /// The network identity the operator expects every provider of this chain to report, in the
+    /// chain's canonical text form. Declarative: nothing reads it during config loading.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub expected_chain_identity: Option<String>,
```
gilcu3: I dont fully understand this one.

> #[serde(default, skip_serializing_if = "Option::is_none")]

What's the problem if we serialize it when None? Just worse operator UX?

> Declarative: nothing reads it during config loading.

How do we use the parameter if we do not read it during config load?

crates/node-config/src/foreign_chains.rs
```diff
@@ -180,6 +184,128 @@ impl ForeignChainsConfig {
 mod tests {
     use crate::ConfigFile;
 
+    #[test]
+    fn config_parsing__should_succeed_with_expected_chain_identity() {
```
gilcu3: nit: it feels natural to merge this test with the one below `config_parsing__should_succeed_when_foreign_chains_are_set` so that we avoid the combinatorial explosion of tests checking for each different parameter in the happy case.

docs/localnet/mpc-configs/config.yaml.template
```diff
@@ -32,6 +32,10 @@ foreign_chains:
   bitcoin:
     timeout_sec: 30
     max_retries: 3
+    # Genesis block hash, as block explorers render it. Distinguishes mainnet from
+    # testnet3 ("000000000933ea01ad0ee984209779baaec3ced90fa3f408719526f8d77f4943")
+    # and regtest ("0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206").
+    expected_chain_identity: "000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f"
```
gilcu3: ups this is a complicated chain indeed :)

docs/localnet/mpc-configs/config.yaml.template
```diff
@@ -40,14 +44,16 @@ foreign_chains:
...
     providers:
       public:
-        rpc_url: "https://api.testnet.abs.xyz"
+        rpc_url: "https://api.mainnet.abs.xyz"
         auth:
           kind: none
   starknet:
     timeout_sec: 30
     max_retries: 3
+    expected_chain_identity: "0x534e5f4d41494e" # SN_MAIN; Sepolia: "0x534e5f5345504f4c4941"
```
gilcu3: is sepolia = testnet for them?

docs/foreign-chain-transactions.md
```diff
@@ -523,9 +523,11 @@ Voting uses the protocol's existing signing threshold (`self.threshold()?.value(
...
+At startup, each resolved provider gets its self-identifying RPC called and the response is compared against that chain's `expected_chain_identity` from the operator's config. The probe is report-only: a provider serving the wrong network is logged, but is not dropped, because a boot-time network blip should not take a chain out of signing.
+
+Taking the expected value from operator config rather than a constant in the attested binary is a deliberate trade. It makes mixed-network and local deployments checkable at all, since a config may pair one chain's mainnet with another's testnet and no binary can ship a value for a devnet. The cost is that the check no longer binds an operator: they can set the wrong value, or omit the field and get no check at all, and either way they fool only their own node's diagnostics. The network-level defenses against a wrong URL are unchanged: threshold voter review of the whitelist, and the provider fan-out, which fails the individual request when a provider disagrees with its siblings.
```
gilcu3, line 530: Ah this was the explanation I was missing when I first saw, good that we have it here :)

gilcu3: Thinking about the downside here, we could also avoid it by providing a default value for each network, depending if our chain_id is mainnet or testnet, and overwriting it with the config value if it exists. What do you think? Depending on operators setting values will remain error prone :(

gilcu3 (dismissed):
Thank you!

Left mostly nit questions and one suggestion

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
haiyuechen-nearone: Will do in follow up PRs that adds concrete probing logic for each chain

docs/localnet/mpc-configs/config.yaml.template
haiyuechen-nearone: Yes https://docs.starknet.io/build/quickstart/sepolia
image

docs/foreign-chain-transactions.md
haiyuechen-nearone: I think we should keep config loading logic simple, and provide default values in the config templates instead, which is what I have done in this PR.

crates/node-config/src/foreign_chains.rs
haiyuechen-nearone: > What's the problem if we serialize it when None? Just worse operator UX?

I believe so yes.

> How do we use the parameter if we do not read it during config load?

For the scope of this PR, it is not used. It will be used in follow up PRs where I introduce the probing logic. My habit is that each PR is a state change for the code, so the comments should reflect the exact current behavior in time. Ideally not reference "the future" if its not a TODO or something.

docs/foreign-chain-transactions.md
gilcu3: From experience this might have some friction with operators, but lets see, I agree software wise this is better.

deployment/cvm-deployment/user-config.toml
```diff
@@ -126,25 +126,31 @@ timeout_sec = 60
 
 # Foreign-chain RPC URLs below target testnet networks (Bitcoin testnet,
 # Abstract testnet, Starknet Sepolia). For mainnet, swap each provider's
-# rpc_url for the corresponding mainnet endpoint.
+# rpc_url for the corresponding mainnet endpoint, and any
+# expected_network_fingerprint for the mainnet value shown beside it.
```
haiyuechen-nearone, line 130: @netrome The foreign chain sections for CVM setup is a bit confusing. Are we supposed to target main or test net in the template? There's both in this file before my edits.

Also we are missing many of the chains that we support now, should we added template for them?

netrome: Hmm yeah this needs to be improved. This is just an example config template, but I'm not sure if our guides cover our current RPC setup. It feels like there's a high risk new operators misconfigure these

gilcu3 (approved):

crates/node-config/src/foreign_chains.rs
```diff
@@ -218,70 +216,18 @@ signature:
...
-            .foreign_chains
-            .starknet
-            .expect("starknet section should be present");
-        assert_eq!(
-            starknet.expected_network_fingerprint.as_deref(),
-            Some("0x534e5f4d41494e")
-        );
+    /// A valid config whose `foreign_chains` section holds `chains`, indented two spaces.
+    fn config_with_chains(chains: &str) -> String {
+        format!("{CONFIG_WITHOUT_FOREIGN_CHAINS}foreign_chains:{chains}")
```
netrome, line 223: Nice

netrome (approved):
Thanks for updating!

PR #4018, opened 2026-07-30
chore: bump ruint to 1.20.0 for RUSTSEC-2026-0220

haiyuechen-nearone:
`cargo-deny` started failing across the repo on RUSTSEC-2026-0220 against `ruint 1.17.2`. The patched range is `>= 1.20.0`, so this bumps the lockfile to it. Lockfile-only, one commit.

### Where ruint comes from

```
ruint → alloy-primitives → alloy → dstack-sdk → tee-authority → mpc-node
                                             → tee-launcher
```

`dstack-sdk` uses `alloy` for Ethereum-style keys and signatures. Nothing in the repo depends on `ruint` or `alloy` directly.

### Risk

Low. `Uint::overflowing_shl`/`overflowing_shr` returned false-negative overflow flags, so `checked_*` yields `Some` instead of `None` and `saturating_*` wraps. Reaching it needs `alloy` or `dstack-sdk` shifting attacker-controlled `Uint`s during attestation. Supply-chain hygiene, not a live vulnerability.

### Notes

- **No `licenses.html` regeneration here.** An earlier revision of this PR also regenerated `third-party-licenses/licenses.html`; that is dropped. Notices are regenerated during release prep by `scripts/ops/prepare-release.sh`, not per dependency bump, so `cargo make license-file` reports drift on this branch exactly as it already does on an unmodified `main`. That task is in neither `check-all-fast` nor `check-extra`, so nothing gates it either way.
- **The lockfile diff is wider than the six `ark-*` entries.** Adding them to the graph re-picks `windows-sys` 0.59.0 → 0.52.0 for eleven packages, `nu-ansi-term` → 0.61.2, and `itertools` → 0.14.0 for `alloy-primitives` and both `prost-derive` versions. `cargo update --package ruint --precise 1.20.0` on a clean tree reproduces this exactly, so it is resolver unification rather than an over-broad command. All of it is Windows-only or proc-macro-only code, and `deny.toml:35` allows multiple versions with `windows-sys` skipped, so `bans` is unaffected.
- The six `ark-*` entries are lockfile-only: optional deps of `ruint 1.20` that no feature here enables. `cargo tree -i ark-ff@0.6.0` finds nothing in the build graph; the existing `ark-ff` 0.3.0 / 0.4.2 are untouched.

### Verification

`cargo make cargo-deny` → `advisories ok, bans ok, licenses ok, sources ok`. `cargo check --workspace --all-targets --locked` passes, and `cargo nextest run -p mpc-node` is 324 passed / 0 failed. `Cargo.lock` is byte-identical to the revision those ran against; re-checked with `cargo metadata --locked` after dropping the license commits.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -386,7 +386,7 @@ dependencies = [
  "alloy-rlp",
  "alloy-serde",
  "alloy-sol-types",
- "itertools 0.13.0",
+ "itertools 0.14.0",
  "serde",
  "serde_json",
  "serde_with",
@@ -681,6 +681,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "ark-ff"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f7a806ac6c8307b929df4645776290a50ee2aac754ad09d8bdf73391309e43af"
+dependencies = [
+ "ark-ff-asm 0.6.0",
+ "ark-ff-macros 0.6.0",
+ "ark-serialize 0.6.0",
+ "ark-std 0.6.0",
+ "digest 0.10.7",
+ "educe",
+ "num-bigint 0.4.6",
+ "num-traits",
+ "zeroize",
+]
+
+[[package]]
 name = "ark-ff-asm"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


haiyuechen-nearone:
@claude review

netrome (dismissed):
Thanks!

gilcu3 (dismissed):
Thanks!

As mentioned in slack I would rather not include the license changes (makes the PR big when it is small) but that's a nit

gilcu3 (approved):

netrome (approved):

PR #4022, opened 2026-07-31
ci: make nix cache independent of Cargo.lock to avoid invalidating rust caches

### LINKED ISSUE #4021: Rust CI cache is invalidated by every Cargo.lock change
### Background

`.github/actions/restore-nix-cache/action.yml:24` hashes `Cargo.lock` together with the base-system inputs, and that hash is folded into the Rust cache `prefix-key` (`.github/workflows/ci.yml:159`, `:256`, `:289`). `Swatinem/rust-cache` only generates fallback restore keys *within* a prefix, so any `Cargo.lock` change changes the prefix and nothing can match: the whole Rust cache is discarded and every affected job cold-builds.

Observed in the merge-queue run for #4014, whose only lockfile change was `spin 0.9.8 → 0.9.9` and `spin 0.10.0 → 0.10.1` (9 lines):

```
prefix-key: v1-rust-tests-node-week-2026-30-nix-b51701d9...
No cache found.
```

The same job on main at `f5c0f4a9` got `Restored from cache key "...-nix-d822d080..." full match: true`. Only the hash differs, because `Cargo.lock` differs. Since `save-if` requires `main`, PR and merge-queue runs also never repopulate it, so the PR cold-builds on every push until it lands. Combined with weekly dependency bumps this is a recurring, avoidable cost on the slowest jobs in CI.

`Cargo.lock` does not belong in that hash. The saved `/nix` artifact is GC'd down to the pinned `nix develop` closure before upload (`gc-max-store-size: 0`, action.yml:36-42), so crane's vendored dependencies are never in it; its contents are a function of the flake, `nix/**` and `rust-toolchain.toml` only. The `/nix` cache is also content-addressed, so a stale restore is a performance question, not a correctness one.

The `nix` hash must stay in the Rust `prefix-key`: it is the hard barrier that stops a Rust cache built on an older base system from being reused after a nix update (#3632). `rust-cache`'s own environment hash covers the rustc version but not a nixpkgs bump that changes glibc/gcc/clang with rustc unchanged. Dropping only `Cargo.lock` from the hash preserves that barrier while letting a lockfile-only change fall back to a partial restore.

### Acceptance Criteria

- `Cargo.lock` is removed from the `hashFiles(...)` call in `.github/actions/restore-nix-cache/action.yml`.
- A change to `flake.nix`, `flake.lock`, `nix/**` or `rust-toolchain.toml` still produces a different Rust cache `prefix-key`, i.e. a full miss with no fallback to a cache built on the previous toolchain.
- A `Cargo.lock`-only change keeps the same prefix and restores via `rust-cache`'s fallback key.
- The `cache-key-hash` output description no longer claims to cover lockfiles and states why `Cargo.lock` is excluded.

gilcu3:
Closes #4021

### DIFF
--- a/.github/actions/restore-nix-cache/action.yml
+++ b/.github/actions/restore-nix-cache/action.yml
@@ -10,10 +10,12 @@ inputs:
 outputs:
   cache-key-hash:
     description: >
-      Hash of the Nix inputs (flake, toolchain, lockfiles) that key the /nix
-      cache. Dependent caches (e.g. the Rust build cache) should fold this into
-      their key so they are invalidated whenever the Nix-provided toolchain
-      changes.
+      Hash of the Nix-provided toolchain and base system (flake, nix/,
+      rust-toolchain) that key the /nix cache. Dependent caches (e.g. the Rust
+      build cache) should fold this into their key so they are invalidated
+      whenever the Nix-provided toolchain changes. Cargo.lock is deliberately
+      excluded: it does not affect the saved /nix closure, and including it
+      would make every dependency bump discard those dependent caches too.
     value: ${{ steps.compute-key.outputs.hash }}
 runs:
   using: composite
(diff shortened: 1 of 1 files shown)


netrome (approved):

haiyuechen-nearone (approved):
🏎️

Issue #1246, opened 2025-10-07
convert contract state from near_sdk::PublicKey to dtos type


gilcu3:
### Background

After #1239 , the contract state still uses `near_sdk::PublicKey`. This implies we need a few conversions every time the state is read and used.

### User Story

Less operations are better

### Acceptance Criteria

The contract state uses the dto types. A migration implementation is added, so that the contract state on chain can be migrated to the new types.

PR #4024, opened 2026-07-31
build: bump near-sdk to 5.29.0

karim-en:
This pull request updates the `near-sdk` dependency in the `Cargo.toml` file to a newer version. This is a straightforward dependency version bump.

- Dependency updates:
  * Upgraded `near-sdk` from version `5.28.0` to `5.29.0` in `Cargo.toml` to ensure compatibility with the latest features and fixes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -191,7 +191,7 @@ near-jsonrpc-primitives = "0.36.0"
 near-kit = { version = "0.12.1", default-features = false }
 near-primitives = "0.36.0"
 near-sandbox = "0.3.11"
-near-sdk = { version = "5.28.0", features = [
+near-sdk = { version = "5.29.0", features = [
     "legacy",
     "unstable",
 ] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -184,12 +184,12 @@ near-account-id = "2.6.0"
 # the git nearcore crate below, which the node's indexer stack requires.
 # Must be bumped in lockstep with `near-primitives`: its types appear in
 # near-primitives' transaction API but are not re-exported there.
-near-crypto-public = { version = "0.36.0", package = "near-crypto" }
+near-crypto-public = { version = "0.37.1", package = "near-crypto" }
 near-gas = "0.3.4"
-near-jsonrpc-client = "0.21.1"
-near-jsonrpc-primitives = "0.36.0"
+near-jsonrpc-client = "0.22.0"
+near-jsonrpc-primitives = "0.37.1"
 near-kit = { version = "0.12.1", default-features = false }
-near-primitives = "0.36.0"
+near-primitives = "0.37.1"
 near-sandbox = "0.3.11"
 near-sdk = { version = "5.29.0", features = [
     "legacy",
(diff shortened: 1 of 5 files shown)


crates/near-mpc-crypto-types/src/conversions/near.rs
```diff
@@ -15,6 +15,9 @@ impl From<&near_sdk::PublicKey> for PublicKey {
                 bytes.copy_from_slice(&pk.as_bytes()[1..]);
                 PublicKey::Ed25519(Ed25519PublicKey::from(bytes))
             }
+            near_sdk::CurveType::MLDSA65 => {
+                unreachable!("MLDSA65 near_sdk::PublicKey has no mapping to dtos::PublicKey")
+            }
```
gilcu3: this is the only place where we probably need to change something. I am not sure we can just crash here

karim-en: Agree, fixed in c25dd857d7e6528a8fb8c5144748fceb67878683.
Some explanations:
1- Copied the from logic to `impl From<PublicKeyExtended> for dtos::PublicKey` instead of using tryFrom to avoid panic here.
2- The from has been replaced by tryFrom instead of removing it because the crate is exposed publicly, so it could be breaking change if we totally remove it.

let me know if you have better idea.

Cargo.toml
```diff
@@ -184,14 +184,14 @@ near-account-id = "2.6.0"
...
-near-crypto-public = { version = "0.36.0", package = "near-crypto" }
+near-crypto-public = { version = "0.37.1", package = "near-crypto" }
 near-gas = "0.3.4"
-near-jsonrpc-client = "0.21.1"
-near-jsonrpc-primitives = "0.36.0"
+near-jsonrpc-client = "0.22.0"
+near-jsonrpc-primitives = "0.37.1"
 near-kit = { version = "0.12.1", default-features = false }
-near-primitives = "0.36.0"
+near-primitives = "0.37.1"
```
gilcu3, line 192: updated these as well so that don't have several copies of the crates in the worktree.  That is what made the lockfile much smaller

crates/contract/src/crypto_shared/types.rs
```diff
@@ -77,12 +79,14 @@ impl From<PublicKeyExtended> for dtos::PublicKey {
...
+                dtos::PublicKey::try_from(&near_public_key)
+                    .expect("Secp256k1 variant always has a secp256k1 key")
             }
             PublicKeyExtended::Ed25519 {
                 near_public_key_compressed,
                 ..
-            } => dtos::PublicKey::from(&near_public_key_compressed),
+            } => dtos::PublicKey::try_from(&near_public_key_compressed)
+                .expect("Ed25519 variant always has an ed25519 key"),
             PublicKeyExtended::Bls12381 { public_key } => public_key,
```
gilcu3, line 90: this conversion is only used in the contract, and the conversion themselves are guaranteed to not fail. We plan to remove this anyway in #1246 so I think this is fine

gilcu3:
@claude review

netrome (approved):
Thanks!

gilcu3 (approved):

PR #4025, opened 2026-07-31
feat: add backup-cli service based on RPCs

### LINKED ISSUE #3864: Automatic backups

gilcu3:
Closes #3864 

The PR got big because it includes many tests and corner cases (LLM reviewers have become quite picky recently). Also because it includes an e2e test, in the second commit


crates/backup-cli/src/cli.rs
```diff
@@ -115,16 +118,50 @@ pub struct PutKeysharesArgs {
...
+    /// (`BACKUP_ENCRYPTION_KEY_HEX`) rather than on the command line.
+    #[arg(long, env, hide_env_values = true)]
     pub backup_encryption_key_hex: String,
+    /// How often to poll the contract state, in seconds.
+    #[arg(long, env, default_value_t = 60)]
+    pub poll_interval_seconds: u64,
+    /// Deadline for a single contract read or keyshare fetch, in seconds. Neither the RPC
+    /// endpoint nor the MPC node connection imposes one of its own.
+    #[arg(long, env, default_value_t = 30, value_parser = clap::value_parser!(u64).range(1..))]
+    pub request_timeout_seconds: u64,
```
gilcu3: this was the results of LLM reviewers being very strict (which is not bad)

crates/backup-cli/src/test_utils.rs
gilcu3, line 1: this should eventually be unified, but I did not find anything usable right now.

gilcu3:
@claude review

Responding to some of your concerns

>  crates/backup-cli/src/keyset.rs:17 — rejecting Initializing silently regresses the existing get-keyshares command. 

the initializing state is transient, and only really valid when it moves to Running. Therefore, for simplicity it is better to not backup keyshares there, mainly to avoid potential inconsistent states.

> docs/node-migration-guide.md:243 — the bullet promises the service logs "a backup that needs your attention at error". No such log exists. service.rs:184 discards every Skipped outcome silently, failures log at warn (service.rs:187),

a failure that logs at warn is exactly what we want the operators to see. Other failures just finish the service, for example at startup

> docs/migration-service.md:485-486 — the "Soft Launch-Specific Components" callout is now false and was left untouched: "No blockchain interaction: Does not query or submit transactions to the contract" and "No automatic monitoring: Operator manually triggers backup/restore operations via CLI commands". backup-cli run does query the contract and does monitor it unattended.

This was left unmodified on purpose, given that the new service is already mentioned in the doc, but not necessarily the final one for soft launch

crates/backup-cli/src/adapters/contract_state_rpc.rs
```diff
@@ -0,0 +1,101 @@
...
+    /// view calls.
+    pub fn new(rpc_url: &str, chain_id: &str, contract_id: AccountId) -> Self {
+        Self {
+            client: Near::custom(rpc_url, chain_id).build(),
+            contract_id,
+        }
+    }
+}
+
+impl ContractStateReader for RpcContractStateReader {
```
netrome: Unrelated to this PR, but this trait should be called `GetContractState` or `ReadContractState` and not `ContractStateReader` as traits should be verbs, not nouns.

crates/backup-cli/src/service.rs
```diff
@@ -0,0 +1,844 @@
...
+    tracing::info!(
+        ?backed_up,
+        poll_interval_seconds = poll_interval.as_secs(),
+        request_timeout_seconds = request_timeout.as_secs(),
+        "starting automatic backup service"
+    );
+
+    let mut poll_ticker = interval(poll_interval);
+    poll_ticker.set_missed_tick_behavior(MissedTickBehavior::Skip);
+
```
netrome: I'm not a fan of hard-coding the polling behavior into the service. Instead, this should be an implementation detail of an adapter. The service needs a way to subscribe to contract state changes. This should be conveyed as a port, and could be represented using a `tokio::watch` or similar.

This simplifies the service, since it would only need to watch for contract changes using the port and take appropriate action.

This also makes it more flexible, since if we want to change the implementation to use the ChainGateway in the future we don't want to impose this polling on top of that one.

netrome (dismissed):
Didn't do a complete review, but I think this code should change. I don't want the service to be polling based as that leaks a short term solution into the core domain logic of the service. The point of using ports and adapters here is that we can have a clean core focused on the central business logic and isolate ugly temporary polling solutions to a single adapter.

crates/backup-cli/src/service.rs
gilcu3: reworked the PR to take this into account

crates/backup-cli/src/adapters/contract_state_rpc.rs
gilcu3: renamed

crates/backup-cli/src/cli.rs
```diff
@@ -93,38 +96,52 @@ pub struct RegisterArgs {
     pub signer_account_id: AccountId,
 }
 
+/// How to reach an MPC node. Shared by every subcommand that talks to one.
 #[derive(clap::Args, Debug)]
-pub struct GetKeysharesArgs {
-    /// host address of the MPC node to retrieve keyshares from (`host:port`).
+pub struct NodeConnectionArgs {
+    /// host address of the MPC node to exchange keyshares with (`host:port`).
```
gilcu3, line 102: there was a small revamp here to avoid adding the same arguments several times. These arguments are needed by 3 existing commands: get-keyshares, put-keyshares and run

crates/backup-cli/src/ports.rs
```diff
@@ -41,7 +52,7 @@ pub trait P2PClient {
     ) -> impl Future<Output = Result<(), Self::Error>> + Send;
 }
 
-pub trait ContractStateReader {
+pub trait ReadContractState {
```
gilcu3, line 51: this is the only grammar change (trait as an action and not a noun). Left the rest in place to keep the change focused on the issue.

crates/backup-cli/src/backup.rs
```diff
@@ -43,75 +47,138 @@ pub async fn run_command(args: cli::Args) {
...
 
+/// Opens the two adapters every keyshare exchange needs: a client for the MPC node, and local
+/// keyshare storage
+async fn open_node_client_and_storage(
+    home_dir: &Path,
+    args: &cli::NodeConnectionArgs,
+) -> (
+    adapters::p2p_client::MpcP2PClient,
+    adapters::keyshare_storage::KeyshareStorageAdapter,
+) {
```
gilcu3, line 107: this is also a small revamp because we had some code repeated 3 times after this PR

crates/backup-cli/src/ports.rs
```diff
@@ -5,6 +5,17 @@ use near_mpc_contract_interface::types::{Keyset, ProtocolContractState};
...
+pub trait WatchContractState {
+    type Error: std::fmt::Debug;
+
+    /// The last state observed on chain. Marks that state as seen, so a following
+    /// [`Self::changed`] waits for the next one.
+    fn latest(&mut self) -> Result<ProtocolContractState, Self::Error>;
+
+    /// Waits until the observed state changes. `Err` means no further states will arrive.
+    fn changed(&mut self) -> impl Future<Output = Result<(), Self::Error>> + Send;
+}
```
netrome, line 13: Looks good to me. The docstrings feel redundant though.

crates/backup-cli/src/ports.rs
```diff
@@ -5,6 +5,17 @@ use near_mpc_contract_interface::types::{Keyset, ProtocolContractState};
...
 use crate::types;
 
+pub trait WatchContractState {
+    type Error: std::fmt::Debug;
+
+    /// The last state observed on chain. Marks that state as seen, so a following
+    /// [`Self::changed`] waits for the next one.
+    fn latest(&mut self) -> Result<ProtocolContractState, Self::Error>;
+
+    /// Waits until the observed state changes. `Err` means no further states will arrive.
```
netrome: If we already attribute a semantic meaning to the error, I don't think the error should be a generic type parameter. I'm not really following why errors shouldn't be allowed to be transient though.

crates/backup-cli/src/service.rs
```diff
@@ -0,0 +1,767 @@
...
+            }
+        }
+
+        tracing::info!("stopping automatic backup service");
+        Ok(())
+    }
+
+    /// Backs up keyshares unless local storage already covers `state`'s keyset. `backed_up` is
+    /// derived from what was stored, not from what was requested, so the two cannot drift apart.
+    async fn tick(
```
netrome: Tick feels like a weird name for this.

netrome (commented):
Started reviewing, this is more in line with what I pictured. Will continue reviewing tonight, but in the meantime it seems like there's room for polishing the PR still. I'm not fully sold on whether we want a generic error parameter in the watch interface or not, and the implicit "error means no more events" invariant feels like something that would be better enforced by the type system instead of a commented runtime invariant.

I'm curious at which point we'd like to stop the event stream though.

crates/backup-cli/src/service.rs
gilcu3: yeah, agree, will try to find a better one

crates/backup-cli/src/ports.rs
gilcu3: yeah the trait itself does not need to figure how the ports will implement it. An alternative if we want to be restrictive is to make it return an option. But for now I will just remove the semantic meaning

crates/backup-cli/src/adapters/contract_state_polling.rs
```diff
@@ -0,0 +1,300 @@
...
+
+async fn poll<Reader: ReadContractState + Sync>(
+    reader: Reader,
+    poll_interval: Duration,
+    read_timeout: Duration,
+    states: watch::Sender<Observed>,
+    cancel: CancellationToken,
+) {
+    let mut ticker = interval(poll_interval);
+    ticker.set_missed_tick_behavior(MissedTickBehavior::Skip);
```
netrome, line 95: Would have been bonus points to inject this ticker as a generic trait bound to make the polling logic independently testable, but this is in an adapter already so would probably be a bit overkill.

crates/backup-cli/src/test_utils.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+pub fn must_get_running_state_with_epoch(epoch_id: u64) -> ProtocolContractState {
+    let mut running = must_get_running_contract_state();
+    running.keyset.epoch_id = EpochId::new(epoch_id);
+    ProtocolContractState::Running(running)
+}
+
+/// What contract initialization leaves behind: `Running`, but without any key yet.
+pub fn must_get_running_state_without_domains() -> ProtocolContractState {
```
netrome: I feel this `must_get_XXX` convention has been misinterpreted since it was added to the engineering standards. This was intended for setup functions with side effects that could fail. This method is practically infallible, and only runs expects on state fixtures.

I think this is an engineering standards bug, and we should clarify this - because this reads weirdly in tests. I had the same problem in my PR https://github.com/near/mpc/pull/4008 - which I've not finished de-slopifying yet.

I think this would be more readable:
```suggestion
pub fn running_state_without_domains() -> ProtocolContractState {
```

crates/backup-cli/src/adapters/contract_state_polling.rs
netrome: However, thinking more on this - since we have quite a bit of tests, properly injecting time here would improve our ability to test the polling logic. Right now we rely on paused tokio time to ensure the tests in this module doesn't take forever.

Since the PR is large it's not worth doing now, but the backup-cli is greenfield code so I would like to establish better patterns here. I think this would be worth adding as a follow-up engineering quality issue. I'll create one.

netrome: Raised a low-priority tech-debt issue to follow this up https://github.com/near/mpc/issues/4059

crates/backup-cli/src/adapters/contract_state_polling.rs
```diff
@@ -0,0 +1,300 @@
...
+
+#[cfg(test)]
+#[expect(non_snake_case)]
+mod tests {
+    use std::collections::VecDeque;
+    use std::sync::Mutex;
+
+    use rstest::rstest;
+
+    use super::*;
```
netrome: Nit: I'm used to leading with the super import in test modules. Not sure if we're consistent here though.

```suggestion
mod tests {
    use super::*;
    
    use std::collections::VecDeque;
    use std::sync::Mutex;

    use rstest::rstest;
```

crates/backup-cli/src/service.rs
```diff
@@ -0,0 +1,781 @@
...
+
+            tokio::select! {
+                observed = self.contract_state.changed() => {
+                    if let Err(err) = observed {
+                        // Not a clean stop: the process should exit non-zero so a supervisor
+                        // restarts it.
+                        return Err(anyhow!("stopped observing the contract state: {err:?}"));
+                    }
+                }
+                () = sleep_or_pending(retry_in) => {}
```
netrome, line 96: I feel I'm a bit hung up on the time port today 😅 but yeah, here we also have another untestable impurity.

netrome (dismissed):
I think this is alright, done a relatively shallow review since the PR is big. In hindsight I think we could have broken this down further. But at the same time, this is a relatively urgent feature.

I got a bit hung up on time management (major theme for me today 😅 ) but considering the size of this PR I think that's better left as a follow-up.

crates/backup-cli/src/service.rs
```diff
@@ -0,0 +1,781 @@
...
+    keyshares: Keyshares,
+    storage: Storage,
+    contract_state: Contract,
+    retry_delay: Duration,
+    backed_up: Option<BackedUpKeyset>,
+}
+
+impl<Keyshares, Storage, Contract> Service<Keyshares, Storage, Contract>
+where
+    Keyshares: P2PClient,
```
pbeza, line 24: Isn’t it odd for `P2PClient` to also be `Keyshares`?

crates/backup-cli/src/service.rs
```diff
@@ -0,0 +1,781 @@
...
+    Storage: KeyShareRepository,
+    Contract: WatchContractState,
+{
+    /// Reads back what local storage already holds, so a restart neither re-fetches nor
+    /// overwrites an existing backup.
```
pbeza: Technically, `KeyShareRepository` doesn’t guarantee permanence (`PermanentKeyStorage` does), so this comment may not always be accurate.

crates/backup-cli/src/service.rs
```diff
@@ -0,0 +1,781 @@
...
+                    Ok(BackupOutcome::Skipped(_)) => retry_in = None,
+                    Ok(BackupOutcome::Cancelled) => break,
+                    Err(err) => {
+                        retry_in = Some(self.retry_delay);
+                        tracing::warn!(
+                            ?err,
+                            retry_in_seconds = self.retry_delay.as_secs(),
+                            "keyshare backup failed, retrying"
+                        );
+                    }
```
pbeza, line 84: What if the backup error is permanent for some reason? We’ll keep retrying forever. Maybe that’s the best we can do here, but I’m not sure.

crates/backup-cli/src/service.rs
```diff
@@ -0,0 +1,781 @@
...
+    /// Returns an error if the contract state stops being observed, since that is a failure
+    /// rather than a stop that was asked for.
+    pub async fn run(mut self, shutdown: CancellationToken) -> anyhow::Result<()> {
+        tracing::info!(
+            backed_up = ?self.backed_up,
+            "starting automatic backup service"
+        );
+
+        // Armed only after a failed backup: there is work left to re-attempt, and a contract
+        // whose state never changes again would otherwise never wake us to do it.
```
pbeza, line 60: Is it realistic for the MPC contract’s state to never change again? If not, this could be simplified, but I’m okay with leaving it as is just in case.

crates/backup-cli/src/service.rs
```diff
@@ -0,0 +1,781 @@
...
+                () = sleep_or_pending(retry_in) => {}
+                _ = shutdown.cancelled() => break,
+            }
+        }
+
+        tracing::info!("stopping automatic backup service");
+        Ok(())
+    }
+
+    /// Backs up keyshares unless local storage already covers `state`'s keyset. `backed_up` is
```
pbeza:
```suggestion
    /// Backs up keyshares unless local storage already covers `state`'s keyset. [`Service::backed_up`] is
```

crates/backup-cli/src/service.rs
```diff
@@ -0,0 +1,781 @@
...
+                            "keyshare backup failed, retrying"
+                        );
+                    }
+                }
+            }
+
+            tokio::select! {
+                observed = self.contract_state.changed() => {
+                    if let Err(err) = observed {
+                        // Not a clean stop: the process should exit non-zero so a supervisor
```
pbeza, line 91: Does “supervisor” mean the node operator, or is it something automated?

crates/backup-cli/src/service.rs
```diff
@@ -0,0 +1,781 @@
...
+/// What the keyshares in local storage cover.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BackedUpKeyset {
+    pub epoch_id: EpochId,
+    pub domain_ids: BTreeSet<DomainId>,
+}
+
+impl BackedUpKeyset {
+    /// `None` when there is nothing stored. All keyshares of a stored keyset share its epoch,
+    /// enforced by `PermanentKeyshareData`.
```
pbeza:
```suggestion
    /// enforced by [`PermanentKeyshareData`].
```

crates/backup-cli/src/service.rs
```diff
@@ -0,0 +1,781 @@
...
+
+/// What the keyshares in local storage cover.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BackedUpKeyset {
+    pub epoch_id: EpochId,
+    pub domain_ids: BTreeSet<DomainId>,
+}
+
+impl BackedUpKeyset {
+    /// `None` when there is nothing stored. All keyshares of a stored keyset share its epoch,
```
pbeza: This `None` handling belongs in the `Service` struct.

```suggestion
    /// All keyshares of a stored keyset share its epoch,
```

crates/backup-cli/src/service.rs
```diff
@@ -0,0 +1,781 @@
...
+#[expect(non_snake_case)]
+mod tests {
+    use std::sync::Arc;
+    use std::sync::atomic::{AtomicUsize, Ordering as AtomicOrdering};
+    use std::time::Duration;
+
+    use mpc_node::keyshare::Keyshare;
+    use mpc_node::keyshare::test_utils::generate_dummy_keyshare;
+    use rand::SeedableRng;
+    use rand::rngs::StdRng;
```
pbeza, line 217: Nit: I like reducing the code by combining imports, but I know Mårten prefers not to, so I’ll pass on this one.

crates/backup-cli/src/ports.rs
pbeza: Perhaps it’s worth adding this rule to `AGENTS.md` as a follow-up? We want to catch these automatically.

crates/e2e-tests/tests/migration_service.rs
```diff
@@ -138,6 +181,27 @@ impl BackupService {
...
+    contract_account_id: &'a str,
+    node_migration_address: &'a str,
+    node_p2p_key: &'a str,
+    backup_encryption_key_hex: &'a str,
+}
+
+/// Kills the service on drop so a failing assertion cannot leak the process.
+struct RunningBackupService {
+    child: std::process::Child,
+}
```
pbeza: We already have a similar struct here:

https://github.com/near/mpc/blob/86f0c3f6cbc27abde83872fbb87fb1a70a6647cc/crates/e2e-tests/src/mpc_node.rs#L155-L156

Perhaps we could reuse it?

crates/e2e-tests/tests/migration_service.rs
```diff
@@ -109,6 +111,47 @@ impl BackupService {
...
+        let keys_dir = self.home_dir.path().join("permanent_keys");
+        let Ok(read_dir) = std::fs::read_dir(&keys_dir) else {
+            return Vec::new();
+        };
+        let mut names: Vec<String> = read_dir
+            .filter_map(|entry| entry.ok()?.file_name().into_string().ok())
+            .collect();
+        names.sort();
+        names
+    }
```
pbeza, line 151: FWIW, by Claude: `snapshot_permanent_keys` (pre-existing, same file, `crates/e2e-tests/tests/migration_service.rs`) and the new `BackupService::stored_keyshare_files` both `read-dir` + `filter_map` + sort the `permanent_keys/` directory.

Extract a free function in `migration_service.rs`, e.g. `fn list_permanent_keys(home_dir: &Path) -> Vec<(String, u64)>`. `snapshot_permanent_keys` becomes a wrapper that resolves the node home dir and delegates; `stored_keyshare_files` delegates and maps pairs to names.

pbeza (dismissed):
LGTM. I only left a few optional nits.

Normally, I’d push to split a PR this big into smaller pieces. Around 1k LoC is usually the most I’m comfortable reviewing, and this one is closer to 2k. But since we’re planning to replace it with a TEE-based `backup-cli` in the long run anyway, I’m fine with merging it as is. 🚢

crates/backup-cli/src/service.rs
gilcu3: the problem is that if the cli fails to get the keyshares from the node, for example due to a transient problem, then unless we have this, we would need to wait for the next contract change (or service restart) to actually try again, which might be already too late

crates/backup-cli/src/service.rs
gilcu3: yeah if the node is down for example, the service cannot do much except retry and retry. We have some delay so that would be a bit of a noisy log, but not a flood. Given that this is a very small service, this should not hide other things so it should be fine

crates/backup-cli/src/service.rs
gilcu3: I am not sure I understand this one. The comment just refers to the fact that on startup the service tries to load the existing backup to avoid getting it again from the node

crates/backup-cli/src/service.rs
gilcu3: yes 🙃

crates/backup-cli/src/adapters/contract_state_polling.rs
gilcu3: thank you! I would be happy to deal with that one sooner or later

crates/backup-cli/src/service.rs
gilcu3: It does read weird. I think the problem comes from `P2PClient` being a noun, while it should be a "verb", as raised by @netrome in this PR before. We should fix that, but leaving as is for now to avoid adding more noise here.

crates/backup-cli/src/service.rs
gilcu3: supervisor here should be something automated, like a tmux session or a systemd service. I think the origin of the term is https://en.wikipedia.org/wiki/Supervisory_program

crates/backup-cli/src/service.rs
gilcu3: do you know if there is any way to enforce this?

crates/backup-cli/src/service.rs
netrome: Yeah this is the problem with naming traits after types (nouns) instead of focusing on the behavior.

```
pub trait P2PClient {
    type Error: std::fmt::Debug;

    fn get_keyshares(
        &self,
        keyset: &Keyset,
    ) -> impl Future<Output = Result<Vec<Keyshare>, Self::Error>> + Send;
    fn put_keyshares(
        &self,
        key_shares: &[Keyshare],
    ) -> impl Future<Output = Result<(), Self::Error>> + Send;
}
```

The problem is here. This should be two separate traits. One `GetKeyshares` trait and one `PutKeyshares` trait. We should not assume a P2PClient being the only way to get and put keyshares. Additionally, it's better to declare in the code only the capability you need and not mix both reads and writes in the same trait.

pbeza (approved):

netrome (approved):
Thanks for the updates 🙏

crates/backup-cli/src/service.rs
pbeza: Not sure. Probably the best we can do is add it to `AGENTS.md`.

crates/backup-cli/src/ports.rs
pbeza: https://github.com/near/mpc/pull/4074 ✅

crates/backup-cli/src/service.rs
pbeza: https://github.com/near/mpc/pull/4074 ✅

Issue #2006, opened 2026-02-06
Unchecked buffer lengths may lead to resource‬ ‭exhaustion


gilcu3:
### Background

‭Each MPC node runs a TLS server that continuously handles incoming TLS connections. The‬
‭server allocates a byte buffer whose length is determined by the first bytes of certain‬
i‭ncoming data-carrying packets. An attacker can set the packet length maliciously to‬
‭exhaust the resources of a target server. Furthermore, a malicious client may create‬
‭multiple connections to amplify resource exhaustion.

### Acceptance Criteria

‭Short term, validate the length of messages that are read from the TCP stream. Ensure the‬
‭message length does not exceed a predefined maximum message length.‬
‭
Long term, review the entire code base to ensure all potentially attacker-controlled inputs‬
‭are validated before usage.

netrome:
Still relevant

PR #4028, opened 2026-08-01
fix(p2p): cap concurrent incoming connections

metalurgical:
Caps how many connections come from a single participant, so one peer can't starve the node via a connection flood.

Related to #2006, case where a malicious participant can open a large number of connections.


gilcu3:
@claude review

metalurgical:
@gilcu3 I approached this from the line of thinking that even connections which could pass authentication currently could potentially be a bad actor or become a bad actor (for whatever reason). What are your thoughts here?

gilcu3:
> @gilcu3 I approached this from the line of thinking that even connections which could pass authentication currently could potentially be a bad actor or become a bad actor (for whatever reason). What are your thoughts here?

That's exactly right. The issue refers to potential malicious participants, and the principle we use is that a participant should not be able to make any other participant crash, nor exhaust its resources.

metalurgical:
> > @gilcu3 I approached this from the line of thinking that even connections which could pass authentication currently could potentially be a bad actor or become a bad actor (for whatever reason). What are your thoughts here?
> 
> That's exactly right. The issue refers to potential malicious participants, and the principle we use is that a participant should not be able to make any other participant crash, nor exhaust its resources.

Ok, will revise this PR.

Cargo.toml
```diff
@@ -199,6 +199,7 @@ near-token = "0.3.4"
 near-workspaces = { version = "0.23.0" }
 num_enum = { version = "0.7.6", features = ["complex-expressions"] }
 pairing = { version = "0.23.0" }
+parking_lot = { version = "0.12.5" }
```
gilcu3, line 202: I would rather avoid introducing a new dependency unless strictly needed, which does not seem the case here

crates/node/src/p2p.rs
```diff
@@ -68,6 +68,10 @@ const WRITE_OPERATION_TIMEOUT: std::time::Duration = std::time::Duration::from_s
 /// never sends a ClientHello would hang the accept task forever.
 const TLS_ACCEPT_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
 
+/// Caps how many incoming connections may be from the same participant at once.
+const MAX_CONCURRENT_CONNECTIONS_PER_PARTICIPANT: usize = 4;
+const CONCURRENT_CONNECTIONS_HEADROOM: usize = 16;
```
gilcu3, line 73: how were these numbers computed?

crates/node/src/p2p.rs
```diff
@@ -630,7 +685,16 @@ async fn incoming_connection_handler(
...
     my_id: ParticipantId,
+    connection_limiter: Arc<tokio::sync::Semaphore>,
+    participant_slots: Arc<PerParticipantConnectionSlots>,
 ) -> anyhow::Result<()> {
+    let Ok(_connection_permit) = connection_limiter.try_acquire_owned() else {
+        tracing::warn!(
+            "dropping incoming connection: at global concurrent incoming connection limit"
+        );
+        return Ok(());
+    };
```
gilcu3, line 696: why do we need this one? If we are capping connections per participant, we already have a de-facto global cap. Introducing this here opens the possibility of unauthenticated peers to block the node, at least for some short period.

gilcu3 (changes requested):
Thanks for the contribution! Requested a couple of changes, not including that CI is currently failing

Cargo.toml
metalurgical: parking_lot contains a non-poisoning optimized variant of the std::sync::Mutex. Just seemed more sensible to use here. Though since async with tokio is present, mutexes should mostly likely be the tokio::sync::Mutex everywhere. Thoughts?

crates/node/src/p2p.rs
metalurgical: Thought they would be sensible, do you see a problem with them?

crates/node/src/p2p.rs
metalurgical: Have been meaning to revise this PR, just haven't gotten to it yet.

Not looking at the code right now, though I don't remember seeing rejections of new peers when a participant set is filled. Perhaps I missed something. 

So the global cap would actually expand each time a new peer presents itself, potentially exhausting available sockets (which are just file descriptors in linux/unix and have a default OS limit) in a connection flood.

Cargo.toml
metalurgical: Have removed it. Just note it is a transient dependency in the codebase elsewhere, so not new.

crates/node/src/p2p.rs
metalurgical: This has been revised

crates/node/src/p2p.rs
metalurgical: Have revised the code in this PR.

gilcu3:
@claude review

crates/node/src/p2p.rs
```diff
@@ -83,6 +86,40 @@ pub struct TlsMeshReceiver {
...
+    fn try_reserve(
+        &self,
+        peer_id: ParticipantId,
+    ) -> anyhow::Result<Option<tokio::sync::OwnedSemaphorePermit>> {
+        let semaphore = self
+            .per_participant
+            .get(&peer_id)
+            .with_context(|| format!("no incoming connection limit for participant {peer_id}"))?;
+        Ok(semaphore.clone().try_acquire_owned().ok())
+    }
```
gilcu3: this looks a bit off. Could be worth changing to return `Option<tokio::sync::OwnedSemaphorePermit>` instead

crates/node/src/p2p.rs
```diff
@@ -1123,8 +1170,8 @@ mod tests {
 
             for _ in 0..100 {
                 // TODO: adjust test?
-                let domain_id = rand::thread_rng().r#gen();
-                let epoch_id = rand::thread_rng().r#gen();
+                let domain_id = random();
+                let epoch_id = random();
```
gilcu3: could you actually make this deterministic, as other tests in the codebase currently do?

gilcu3 (dismissed):
Thank you!

Left some minor comments, and please fix the blocking point by Claude above.

crates/node/src/p2p.rs
metalurgical: Done

crates/node/src/p2p.rs
metalurgical: Have done so. Seedable now, same initialization as elsewhere.

metalurgical:
> Thank you!
> 
> Left some minor comments, and please fix the blocking point by Claude above.

@gilcu3 

Thanks for the review. :)

Have addressed the blocking issue (additional test) and your comments.

Additionally also addressed the following:
* Reduced max connections per peer from 8 down to 4.
* Added a metric instead of the debug log, this will need to be wired into a dashboard separately.

Then I think that the buffer size mentioned should probably be reduced considerably, it is probably 10x-20x larger than needed? Something for a separate PR.

crates/node/src/p2p.rs
```diff
@@ -679,12 +677,10 @@ async fn incoming_connection_handler(
...
-    let Some(_connection_slot) = incoming_connection_limits.try_reserve(peer_id)? else {
-        tracing::debug!(
-            peer_id = %peer_id,
-            max = MAX_CONCURRENT_CONNECTIONS_PER_PARTICIPANT,
-            "dropping incoming connection: participant at concurrent connection limit"
-        );
+    let Some(_connection_slot) = incoming_connection_limits.try_reserve(peer_id) else {
+        INCOMING_CONNECTIONS_REJECTED
+            .with_label_values(&[peer_id.to_string().as_str()])
+            .inc();
```
gilcu3: nit: we could keep both log and metric, as the latter will not carry at what time did it happen exactly

crates/node/src/p2p.rs
```diff
@@ -1746,25 +1724,68 @@ mod tests {
...
-    fn incoming_connection_limits__should_error_for_unknown_participant() {
-        // Given
-        let limits = IncomingConnectionLimits::new([ParticipantId::from_raw(1)]);
+    #[tokio::test]
+    #[test_log::test]
+    async fn incoming_connection_handler__should_close_connection_when_peer_connection_limit_is_reached()
+     {
+        start_root_task_with_periodic_dump(async move {
+            // Given a known peer that already has its slots filled
+            let my_id = ParticipantId::from_raw(0);
```
gilcu3, line 1780: nit: for given/when/then we prefer bare comments in most cases

gilcu3 (dismissed):
Thank you!

crates/node/src/p2p.rs
```diff
@@ -1654,4 +1699,93 @@ mod tests {
...
+            .map(|_| limits.try_reserve(peer).unwrap())
+            .collect();
+        assert!(limits.try_reserve(peer).is_none());
+
+        // When
+        held_slots.pop();
+
+        // Then
+        assert!(limits.try_reserve(peer).is_some());
+    }
```
pbeza: IIUC, this test effectively tests `tokio::sync::Semaphore`, which I’m not sure we need to do here. I don’t have a strong opinion though, so if @gilcu3 disagrees, I’m fine with keeping it.

crates/node/src/p2p.rs
```diff
@@ -1654,4 +1699,93 @@ mod tests {
...
+        let _held_slots: Vec<_> = (0..MAX_CONCURRENT_CONNECTIONS_PER_PARTICIPANT)
+            .map(|_| limits.try_reserve(flooding_peer).unwrap())
+            .collect();
+
+        // When
+        let other_peer_slot = limits.try_reserve(other_peer);
+
+        // Then
+        assert!(other_peer_slot.is_some());
+    }
```
pbeza: [Ditto](https://github.com/near/mpc/pull/4028/changes#r4073051157).

crates/node/src/p2p.rs
```diff
@@ -1654,4 +1699,93 @@ mod tests {
...
+
+    #[tokio::test]
+    #[test_log::test]
+    async fn incoming_connection_handler__should_close_connection_when_peer_connection_limit_is_reached()
+     {
+        start_root_task_with_periodic_dump(async move {
```
pbeza, line 1778: Why `start_root_task_with_periodic_dump` instead of `start_root_task`? Do we actually need a loop here?

pbeza (dismissed):

crates/node/src/p2p.rs
metalurgical: Added it back in.

crates/node/src/p2p.rs
metalurgical: Have removed them

crates/node/src/p2p.rs
metalurgical: Used the existing tests as a baseline to build this one from. Think is could stay as it is though.

gilcu3 (approved):

crates/node/src/p2p.rs
```diff
@@ -649,6 +687,21 @@ async fn incoming_connection_handler(
...
+        tracing::debug!(
+            peer_id = %peer_id,
+            max = MAX_CONCURRENT_CONNECTIONS_PER_PARTICIPANT,
+            "dropping incoming connection, participant at connection limit"
+        );
+        INCOMING_CONNECTIONS_REJECTED
+            .with_label_values(&[peer_id.to_string().as_str()])
+            .inc();
+        if let Err(err) = tls_stream.shutdown().await {
+            error!(err = %err, "TLS shutdown failed");
```
kevindeforth: nit: we prefer `tracing::error`

kevindeforth (approved):
Thanks! 
Approving in lieu of @pbeza because I work on a conflicting branch. Lets merge this.

crates/node/src/p2p.rs
metalurgical: Could you explain why? By adding it to the imports section (use tracing::{info, error}), you don't need to use the longer path anymore, which is good practice?

Issue #3477, opened 2026-06-04
Foreign-tx sub-quorum verification: retry vs terminal policy


anodar:
### Background

A node accepts a foreign-tx verification result only when ≥ rpc_quorum(C) of its whitelisted providers return the same response. A sub-quorum (fewer agree) must be terminal for the leader — it must not be  re-attempted via the generic request queue's MAX_ATTEMPTS_PER_REQUEST_AS_LEADER retry.

### Acceptance Criteria

We don't retry foreign tx validation requests where only fewer than rpc_quorum(C) agrees (others return different values).

Transient errors (such as timeout) should be retried.

PR #4032, opened 2026-08-03
feat: Add optional payload hash to foreign transaction requests

### LINKED ISSUE #3997: [Audit] `respond_verify_foreign_tx` doesn't check that `payload_hash` matches the response

netrome:
closes #3997 

Supersedes https://github.com/near/mpc/pull/4008 which was closed during a hack

### DIFF
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -4088,26 +4088,6 @@ mod tests {
         }
     }
 
-    fn must_sign_foreign_tx_payload(
-        secret_key: &k256::Scalar,
-        payload: &ForeignTxSignPayload,
-    ) -> VerifyForeignTransactionResponse {
-        let payload_hash = payload.compute_msg_hash().unwrap();
-        let secret_key_ec: elliptic_curve::SecretKey<Secp256k1> =
-            elliptic_curve::SecretKey::from_bytes(&secret_key.to_bytes()).unwrap();
-        let secret_key = SigningKey::from_bytes(&secret_key_ec.to_bytes()).unwrap();
-        let (signature, recovery_id) = secret_key
-            .sign_prehash_recoverable(&payload_hash.0)
-            .unwrap();
-        let signature = dtos::SignatureResponse::Secp256k1(
-            dtos::K256Signature::from_ecdsa_recoverable(&signature, recovery_id),
-        );
-        VerifyForeignTransactionResponse {
-            payload_hash,
-            signature,
-        }
-    }
-
     #[test]
     fn respond_verify_foreign_tx__should_reject_response_with_unexpected_payload_hash() {
         // Given
@@ -4137,7 +4117,7 @@ mod tests {
                 BitcoinExtractedValue::BlockHash([42u8; 32].into()),
             )],
         });
-        let response = must_sign_foreign_tx_payload(&secret_key, &payload);
+        let response = sign_foreign_tx_payload(&secret_key, &payload);
         with_active_participant_and_attested_context(&contract);
 
         // When
@@ -4186,7 +4166,7 @@ mod tests {
         };
         let request = args_into_verify_foreign_tx_request(request_args.clone());
         contract.verify_foreign_transaction(request_args);
-        let response = must_sign_foreign_tx_payload(&secret_key, &payload);
+        let response = sign_foreign_tx_payload(&secret_key, &payload);
         with_active_participant_and_attested_context(&contract);
 
         // When
@@ -4197,6 +4177,26 @@ mod tests {
             result.is_ok(),
             "response matching the expected payload hash must be accepted: {result:?}",
         );
+    }
+
+    fn sign_foreign_tx_payload(
+        secret_key: &k256::Scalar,
+        payload: &ForeignTxSignPayload,
+    ) -> VerifyForeignTransactionResponse {
+        let payload_hash = payload.compute_msg_hash().unwrap();
+        let secret_key_ec: elliptic_curve::SecretKey<Secp256k1> =
+            elliptic_curve::SecretKey::from_bytes(&secret_key.to_bytes()).unwrap();
+        let secret_key = SigningKey::from_bytes(&secret_key_ec.to_bytes()).unwrap();
+        let (signature, recovery_id) = secret_key
+            .sign_prehash_recoverable(&payload_hash.0)
+            .unwrap();
+        let signature = dtos::SignatureResponse::Secp256k1(
+            dtos::K256Signature::from_ecdsa_recoverable(&signature, recovery_id),
+        );
+        VerifyForeignTransactionResponse {
+            payload_hash,
+            signature,
+        }
     }
 
     #[test]
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -4124,9 +4124,9 @@ mod tests {
         let result = contract.respond_verify_foreign_tx(request.clone(), response);
 
         // Then
-        assert!(
-            result.is_err(),
-            "response with a payload hash differing from the request's expected hash must be rejected"
+        assert_matches!(
+            result.unwrap_err(),
+            Error::Respond(RespondError::UnexpectedPayloadHash)
         );
         assert!(
             contract
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -4165,6 +4165,58 @@ mod tests {
         );
     }
 
+    #[test]
+    fn respond_verify_foreign_tx__should_reject_request_with_erased_expected_payload_hash() {
+        // Given
+        let mut rng = rand::rngs::StdRng::from_seed([42u8; 32]);
+        let (context, mut contract, secret_key) =
+            basic_setup_with_protocol(Protocol::CaitSith, DomainPurpose::ForeignTx, &mut rng);
+        register_supported_chains(&mut contract, [dtos::ForeignChain::Bitcoin]);
+        testing_env!(context.clone());
+        let SharedSecretKey::Secp256k1(secret_key) = secret_key else {
+            unreachable!();
+        };
+        let request_args = VerifyForeignTransactionRequestArgs {
+            domain_id: DomainId::default().0.into(),
+            payload_version: ForeignTxPayloadVersion::V1,
+            expected_payload_hash: Some(dtos::Hash256([1u8; 32])),
+            request: dtos::ForeignChainRpcRequest::Bitcoin(BitcoinRpcRequest {
+                tx_id: [7u8; 32].into(),
+                confirmations: 2.into(),
+                extractors: vec![BitcoinExtractor::BlockHash],
+            }),
+        };
+        let request = args_into_verify_foreign_tx_request(request_args.clone());
+        contract.verify_foreign_transaction(request_args);
+        let payload = ForeignTxSignPayload::V1(ForeignTxSignPayloadV1 {
+            request: request.request.clone(),
+            values: vec![ExtractedValue::BitcoinExtractedValue(
+                BitcoinExtractedValue::BlockHash([42u8; 32].into()),
+            )],
+        });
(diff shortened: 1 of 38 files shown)


crates/contract/src/v3_13_0_state.rs
```diff
@@ -163,7 +163,14 @@ impl From<MpcContract> for crate::MpcContract {
             protocol_state: old.protocol_state,
             pending_signature_requests: old.pending_signature_requests,
             pending_ckd_requests: old.pending_ckd_requests,
-            pending_verify_foreign_tx_requests: old.pending_verify_foreign_tx_requests,
+            // `VerifyForeignTransactionRequest` gained `expected_payload_hash`, changing the
+            // borsh key encoding, so entries pending at upgrade time are abandoned; their
+            // yields time out on chain as if never responded to. The abandoned V2 entries
+            // are no longer addressable, so their storage staking is never reclaimed
+            // (bounded by the number of requests in flight at upgrade time).
```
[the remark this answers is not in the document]
netrome: This is intentional although not ideal I agree. However, migrating keys is complex as we cannot iterate over a lookup map so we'll just accept the short hiccup during the upgrade.

crates/contract/src/v3_13_0_state.rs
```diff
@@ -163,7 +163,14 @@ impl From<MpcContract> for crate::MpcContract {
             protocol_state: old.protocol_state,
             pending_signature_requests: old.pending_signature_requests,
             pending_ckd_requests: old.pending_ckd_requests,
-            pending_verify_foreign_tx_requests: old.pending_verify_foreign_tx_requests,
+            // `VerifyForeignTransactionRequest` gained `expected_payload_hash`, changing the
+            // borsh key encoding, so entries pending at upgrade time are abandoned; their
+            // yields time out on chain as if never responded to. The abandoned V2 entries
```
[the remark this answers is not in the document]
netrome: Addressed in https://github.com/near/mpc/pull/4032/commits/328221d93c82768e2c590783c290e29dfe23ecd5

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -410,24 +417,42 @@ mod tests {
...
+    fn build_signature_request__should_reject_payload_not_matching_expected_hash() {
+        // Given
+        let request = verify_foreign_tx_request(Some(dtos::Hash256([1u8; 32])));
+        let payload = bitcoin_payload();
+
+        // When
+        let result = build_signature_request(&request, &payload);
+
+        // Then
+        result.unwrap_err();
```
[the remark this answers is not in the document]
netrome: Addressed in https://github.com/near/mpc/pull/4032/commits/328221d93c82768e2c590783c290e29dfe23ecd5

crates/contract/src/lib.rs
```diff
@@ -730,6 +730,12 @@ impl MpcContract {
             return Err(TeeError::TeeValidationFailed.into());
         }
 
+        if let Some(expected_payload_hash) = &request.expected_payload_hash
+            && &response.payload_hash != expected_payload_hash
+        {
+            return Err(RespondError::UnexpectedPayloadHash.into());
+        }
```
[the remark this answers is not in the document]
netrome: I can see your point, but in this case I'd prefer to do response validation which do not touch the contract state first to save some gas in the error case.

crates/near-mpc-sdk/src/foreign_chain.rs
```diff
@@ -45,13 +45,8 @@ impl ForeignChainSignatureVerifier {
...
-        let expected_payload = ForeignTxSignPayload::V1(ForeignTxSignPayloadV1 {
-            request: self.request,
-            values: self.expected_extracted_values,
-        });
-
-        let expected_payload_hash = expected_payload
-            .compute_msg_hash()
+        let expected_payload_hash = self
+            .expected_payload_hash()
             .map_err(|_| VerifyForeignChainError::FailedToComputeMsgHash)?;
```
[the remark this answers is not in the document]
netrome: I'd push back on this. I agree we could get rid of the clone, but the extra lines but adding a separate construction path beyond the existing `expected_payload_hash` function feels bad. More correct would be refactoring `expected_payload_hash` to not impose any clones, but that would also further bloat the PR. And the extra clones are negligible for now imo.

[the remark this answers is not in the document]
netrome: Edit, this was a tiny refactor - so added it in https://github.com/near/mpc/pull/4032/commits/d75b1765c5290e59bce0703a28b872ac56cca85c

netrome:
> crates/near-mpc-sdk/src/foreign_chain.rs:86 — ForeignChainSignatureVerifier is borsh-persistable but does not record the payload version; expected_payload_hash() re-reads the crate-global DEFAULT_PAYLOAD_VERSION at verify time while build() pins payload_version into the on-chain request. Once a V2 exists and the default moves, a verifier persisted across an SDK upgrade will recompute a V2 hash for a V1 request and reject a valid response. The new ForeignTxSignPayload::new(version, …) already takes a version — storing it in the verifier at build time closes this cheaply.

This sounds like a potential good point, but orthogonal to this PR and I don't have the capacity to distinguish this statement from slop/hallucinations atm so I will leave this for now.

netrome:
> > crates/near-mpc-sdk/src/foreign_chain.rs:86 — ForeignChainSignatureVerifier is borsh-persistable but does not record the payload version; expected_payload_hash() re-reads the crate-global DEFAULT_PAYLOAD_VERSION at verify time while build() pins payload_version into the on-chain request. Once a V2 exists and the default moves, a verifier persisted across an SDK upgrade will recompute a V2 hash for a V1 request and reject a valid response. The new ForeignTxSignPayload::new(version, …) already takes a version — storing it in the verifier at build time closes this cheaply.
> 
> This sounds like a potential good point, but orthogonal to this PR and I don't have the capacity to distinguish this statement from slop/hallucinations atm so I will leave this for now.

Looked deeper into this and created a follow-up as it may be a good thing to look into https://github.com/near/mpc/issues/4079

crates/node/src/tests.rs
```diff
@@ -501,6 +501,7 @@ pub async fn request_verify_foreign_tx_and_await_response(
             }),
             domain_id: domain.id.0.into(),
             payload_version: ForeignTxPayloadVersion::V1,
+            expected_payload_hash: None,
         },
     };
     tracing::info!(
```
haiyuechen-nearone, line 507: nit: not directly related to the changes in this PR.
I couldn't find any references to `request_verify_foreign_tx_and_await_response`. Is this dead or are we waiting to add some tests that will use this method?

netrome: Hmm interesting, yeah this looks like dead code. The function was used back in https://github.com/near/mpc/pull/1948#discussion_r2763662350 at which point I was already a bit confused by it. Now there's a dangling TODO - which shouldn't be allowed.

Yes, orthogonal to this PR, but feel free to clean this up. I think we should remove this, or add the missing test coverage the comment seems to hint at.

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -32,8 +32,15 @@ fn build_signature_request(
...
-    let payload_hash: [u8; ECDSA_PAYLOAD_SIZE_BYTES] =
-        foreign_tx_payload.compute_msg_hash()?.into();
+    let msg_hash = foreign_tx_payload.compute_msg_hash()?;
+    if let Some(expected_payload_hash) = &request.expected_payload_hash
+        && expected_payload_hash != &msg_hash
+    {
+        bail!(
+            "computed payload hash {msg_hash:?} does not match the request's expected payload hash {expected_payload_hash:?}"
+        );
+    }
```
haiyuechen-nearone, line 42: As part of this PR we are adding the payload hash check on the contract already, so I don't think we should put a check here again. 

Seeing from the MPC networks perspective, the node did nothing wrong. We can only reach this branch if every provider answered "tx exist and at expected location", but the hash of the payload is somehow different. If I understand it correctly, this indicates that the requester made a mistake or is malicious, and the node should not be retrying the computation for this request again.

Having this bail!() statement would actually cause the Node to exhaust all 10 retries for this request, each time consuming another presignature. The cache [here](https://github.com/near/mpc/blob/014b6a71aef65a15617831f7ab18c2de6955765c/crates/node/src/mpc_client.rs#L493-L497) is never set.

So to "correctly" handle such situations, I think we should treat this as a terminal state for the request and exit the round explicitly as part of [#3477](https://github.com/near/mpc/issues/3477). However since we don't have a clean setup for terminal states right now, I think its better to just let this request slide at the node and let the contract handle the gating.

netrome: Hmm you're right, I didn't consider the consumption of presignatures happening before this. I wanted to make sure to bail before any signature computation happened to not waste a signing round.

However, if we only check this in the contract I believe we'd still have the network retry 10 times which would cause 10 signing rounds + 10 presignatures. I'll double check this, but I think invalid requests should not consume any assets.

netrome: Reordering so the check happens before presignatures are consumed is trivial, so changed it in https://github.com/near/mpc/pull/4032/commits/cd824877fb3c6d5891b72bf46b58783718ee31e4

Unfortunately, this legacy code is pretty hard to test (loads of concrete types that should be put behind traits to allow dependency injection). A refactor of this is out of scope for the current PR, so for now this particular behavior (presignatures should not be popped on failing requests) remains untested. Which feels very unfortunate but because

1. This fix is needed before our audit fix review in September.
2. I'm going out on vacation next week.

I'd prefer to merge this as-is and leave this larger refactor for the future. I created https://github.com/near/mpc/issues/4083 as a follow-up for this test case instead.

haiyuechen-nearone (dismissed):
🚢

crates/node/src/providers/verify_foreign_tx/sign.rs
haiyuechen-nearone: Thanks for reordering the operations, approving as I am happy with the solution for the time being. I will see if I can pick up https://github.com/near/mpc/issues/4083 after I finish the probing logic

crates/contract/src/lib.rs
```diff
@@ -732,6 +732,12 @@ impl MpcContract {
             return Err(TeeError::TeeValidationFailed.into());
         }
 
+        if let Some(expected_payload_hash) = &request.expected_payload_hash
```
anodar, line 735: Consider adding a test where malicious elected leader erases this expected payload hash and ensure that `respond_verify_foreign_tx` fails.

anodar (dismissed):

crates/contract/src/lib.rs
netrome: Ah right, makes sense. Will add.

netrome: Added in https://github.com/near/mpc/pull/4032/commits/63e84dd10b272daaddeedce00783a55c3bc9b716

anodar (approved):
Thank you!

haiyuechen-nearone (approved):

PR #4036, opened 2026-08-03
chore: lint module declaration order by visibility and grouping

### LINKED ISSUE #4034: Lint module declaration order by visibility
### Background

`pub`, `pub(crate)` and private `mod` declarations are interleaved across the workspace (review nit on #4013), and in `crates/threshold-signatures/src/lib.rs` they are additionally split into two blocks either side of the `use` statements. rustfmt's `reorder_modules` sorts alphabetically and ignores visibility, and clippy has no equivalent lint, so this needs custom ast-grep rules like `lints/no-use-in-fn.yml`.

Two cases any ordering has to tolerate: `#[macro_use] mod macros;` at `crates/threshold-signatures/src/lib.rs:1` must stay textually first, and `#[cfg(test)]` modules conventionally sit at the bottom of a file.

### Acceptance Criteria

- Module declarations are grouped by visibility, widest first: `pub`, then `pub(crate)` and `pub(super)`, then private, with a blank line between groups.
- A scope's module declarations sit in one block rather than being split apart by other items.
- `#[macro_use]` modules keep their leading position and `#[cfg(test)]` modules stay at the bottom of their file.
- Violations fail CI rather than surfacing as a review comment.
- The whole workspace already satisfies both rules when the checks start running.

haiyuechen-nearone:
Closes #4034

Two rules: declarations ordered by visibility, and kept in one block per scope.

- `pub(super)` groups with `pub(crate)`.
- Contiguity only rejects a split block, not a block that sits below other items.

### What the two rules do

**`mod-declarations-contiguous`** — declarations of a scope stay in one block.

    for each `mod x;` / `pub mod x;`:
        above = nearest item above, skipping comments
        if above is a declaration, or an attribute:  ok      # in a block
        else if no declaration anywhere above:       ok      # first block
        else:                                        error   # second block

Blank lines are invisible to it, only the first line of a stray group is reported, and an
attributed declaration may stand alone, which keeps `#[cfg(test)] mod tests;` at the foot
of a file legal.

**`mod-declaration-visibility-order`** — that block runs widest visibility first.

    tier: pub = 2, pub(crate) / pub(super) = 1, private = 0

    for each `pub...` declaration:
        above = nearest declaration above, skipping comments and attributes
        if above carries #[macro_use]:  ok      # has to stay textually first
        if tier(above) < tier(self):    error

It only compares neighbours, so it speaks up after the first rule is satisfied. Keep a
blank line between the tiers, since rustfmt sorts within a run.

Accepted and rejected shapes for both are in `lints/tests/`.


haiyuechen-nearone:
@claude review

haiyuechen-nearone:
Fixed last Claude blocking finding in [a5a260c](https://github.com/near/mpc/pull/4036/commits/a5a260c474bac5600a75bba20361169f1771d15b)

Ready for human review

lints/rules/mod-declaration-visibility-order.yml
gilcu3, line 1: Oh I wont pretend I understand this logic. The `ast-grep` is quite general

lints/rules/mod-declaration-visibility-order.yml
```diff
@@ -0,0 +1,57 @@
+id: mod-declaration-visibility-order
+language: rust
+severity: error
+message: "`$VIS mod $M;` is declared directly after a module with narrower visibility"
+note: Declare modules widest visibility first, `pub`, then `pub(crate)`, then private.
+# Repeated per branch: in ast-grep 0.42 a `matches:` reference inside a relational
+# rule matches nothing instead of erroring.
```
gilcu3: do we need this comment? If that's something that changed later, could we just bump the crate we use?

lints/tests/__snapshots__/mod-declarations-contiguous-snapshot.yml
gilcu3, line 1: this one I cannot understand. What exactly does this guarantee?

lints/tests/__snapshots__/mod-declaration-visibility-order-snapshot.yml
gilcu3, line 1: also this one is hard. Are these all valid examples?

lints/tests/mod-declaration-visibility-order-test.yml
gilcu3, line 1: I am not a big fan of these tests because they feel like "tests for the tests". But in this case it works as well as an example for what is allowed and what isnt.

gilcu3 (dismissed):
Thank you!

Are you ok with the non-blocking problems in the last LLM review? The first point would be good, if it is not too complicated ofc.

netrome (dismissed):
Also not going to pretend I understand ast-grep but the module reorg looks very reasonable, and from what I can see, nothing super weird stands out in the yaml files although I'm not following what's going on there.

lints/rules/mod-declaration-visibility-order.yml
haiyuechen-nearone: Revamped the rule with Claude again and the comment is not needed anymore. However ast-grep version is pinned in Nix as well, can we even bump that easily?

gilcu3: yeah, bumping something there is exactly the only way to bump our tools right now

gilcu3 (dismissed):
Still LGTM

One thing now that we have an update command, I think we need to add some small section somewhere so that humans know exactly how this is used, specially why do we need to update snapshots should be explained somewhere. At least for me I thought the snapshots would never change for example

gilcu3 (approved):

lints/rules/mod-declaration-visibility-order.yml
haiyuechen-nearone: Please see the "pseudo code" section I added in the PR description

lints/tests/__snapshots__/mod-declaration-visibility-order-snapshot.yml
haiyuechen-nearone: Removed now, but no they were all invalid examples. Aka "samples where the lint rules should mark red" and on which positions should the output from `ast-grep` be.

lints/tests/__snapshots__/mod-declarations-contiguous-snapshot.yml
haiyuechen-nearone: For a lint rule that only shouts but not fix, not much, and therefore I removed these now. I initially thought they were required by `ast-grep` so added them. They essentially verify that the lint rule would colour the "right characters".

However if we later want to add a fixer in the linter definition then we should have a snapshot, which can verify that the "code after fix" is the stable, if the linter definition changes.

lints/tests/mod-declaration-visibility-order-test.yml
haiyuechen-nearone: I removed the snapshots now, so these are the only "tests". I think this is kind of "best we can do" tests, that demonstrates what is captured by the rule and what is allowed by the rule, in addition to what is being fired in CI / fixed in this PR

haiyuechen-nearone:
> One thing now that we have an update command, I think we need to add some small section somewhere so that humans know exactly how this is used, specially why do we need to update snapshots should be explained somewhere. At least for me I thought the snapshots would never change for example

Removed the snapshots in latest iteration. I misunderstood and thought `ast-grep` required snapshot tests. For these 2 rules currently I think we only need tests to verify what would trigger and what would not.

However in the future if we want to add a fixer to these lint rules then we should probably have snapshot tests for before/after fix. Right now I only defined rules to capture faults and left the fixing part to the ~~developer~~ Claude

netrome (approved):

PR #4038, opened 2026-08-03
ci: scan changed files for malicious code and disguised content

### LINKED ISSUE #4037: ci: scan changed files for malicious code and disguised content
### Background

Nothing in CI inspects the content of files a change adds. Scanners that select files by extension never open a payload named `.woff2`, and dependency scanners only read the dependency graph, so code vendored directly into the tree goes unexamined.

### Acceptance Criteria

- [ ] A required job scans changed files with YARA and Magika, neither selecting files by extension.
- [ ] A rule blocks only once measured at zero false positives on this repo; the rest annotate.
- [ ] The rule pack is fetched from a pinned, checksum-verified release rather than committed.
- [ ] Runs on all branches, not only pull requests.
- [ ] Missing inputs or a scanner crash fail the job rather than passing with nothing scanned.

https://github.com/near/mpc/commit/93c9ab53eba97011e1330accd9a5062ac6d4533d

Blocking/advisory split, measurement procedure and version-bump steps documented in `.github/yara/README.md`.

pbeza:
Closes #4037

### DIFF
--- a/.github/scripts/fetch-scanners.sh
+++ b/.github/scripts/fetch-scanners.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+#
+# Downloads the scanners the security scan needs, verifying each checksum before
+# use, and prints the directory they landed in.
+#
+#   eval "$(.github/scripts/fetch-scanners.sh <destination-directory>)"
+#
+# CI only, and x86_64 Linux only. Locally the same versions come from the nix dev
+# shell (nix/yara-x.nix, nix/magika.nix) - keep the versions here in step with
+# those files.
+#
+# This replaces apt + pipx, which cost ~16s of a 26s job: apt-get update 5.4s,
+# apt-get install yara 2.9s, pipx install magika 7.4s.
+
+set -euo pipefail
+
+# Release assets are maintainer-mutable - a tag can be deleted and re-uploaded -
+# so the digest is what protects this job, not the tag. Never use a floating tag
+# such as magika's `cli-latest`.
+readonly YARA_X_VERSION="1.19.0"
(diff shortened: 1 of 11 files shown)


pbeza:
Closing in favour of #4056, which takes the same threat in a substantially simpler direction.

Two findings while reviewing this one drove the rewrite:

**The GuardDog rule pack does not cover Rust.** All 54 rules carry `path_include`, and none names `.rs`. Coverage across the pack is `.py/.pyx/.pyi/.pth` (41 rules), `.js/.ts/.jsx/.tsx/.mjs/.cjs` (41), `.go` (18), `.rb/.gemspec` (10), `.sh` (2). Running them over every file regardless of type is off-label use, which is what forced 18 of 54 into the advisory list here, along with a 54-line allowlist that has to be re-measured on every pack bump. For the payload we actually received, the YARA hit was redundant anyway: magika independently reports "content is javascript, name claims a font".

**Native code scanning already provides the machinery this reimplements.** Emitting SARIF moves severity, false-positive triage, and merge blocking onto the platform: alert dismissal is per finding and persistent, so `blocking-rules.txt` disappears entirely, and a ruleset code scanning rule blocks a merge on a severity threshold *and* when the tool has not reported. That last part also resolves the open question at `.github/yara/README.md:115-124` about the skipped duplicate check satisfying a required-status-check.

#4056 keeps everything from this PR that was load-bearing: the content-vs-extension check, and every cannot-scan-is-failure guard (missing commits, unreadable files, the `-z` homoglyph handling, newline paths). Net difference is two fewer scripts, one fewer nix derivation, no rule allowlist, no measurement procedure, and 721 added lines instead of 960.

Also carried over into #4056 as a required admin follow-up: `require_last_push_approval` is currently `false` on the `main` ruleset, which is the vector both PRs are ultimately about.

PR #4046, opened 2026-08-04
chore: make rust doc lints cover all cases

### LINKED ISSUE #4045: check-docs does not check intra-doc links on private or feature-gated items
### Background

`check-docs` runs `cargo doc --workspace --no-deps` with `RUSTDOCFLAGS="-D warnings"` and is part of `check-all-fast`, so CI does gate on broken intra-doc links. But it misses two categories, verified by injecting a broken ``[`NoSuchType`]`` link and running the exact command:

| Case | Caught |
|---|---|
| Public item, default features | yes (exit 101) |
| Private item | no (exit 0) |
| `#[cfg(feature = "...")]` public item | no (exit 0) |
| Feature-gated item with `--all-features` | yes (exit 101) |

Without `--document-private-items`, rustdoc strips private items before the link-resolution pass, so links in private internals are never validated. Without `--all-features`, anything behind `test-utils` or `network-hardship-simulation` is unchecked, even though `clippy-all-features` already covers feature-gated code for lints.

Running with both flags surfaces 24 warnings across 5 crates: `threshold-signatures` (9), `mpc-node` (9), `foreign-chain-inspector` (3), `mpc-tls` (2), `tee-authority` (1). About half are markdown that needs escaping (citation brackets like `[GS21]`, index notation like `[i]`, an unclosed `<u8>` tag, bare URLs). The rest are genuinely stale references, for example ``[`IndexerApi`]`` in `crates/node/src/indexer/real.rs:54` and ``[`RobustEcdsaSignatureProvider`]`` in `crates/node/src/providers/ecdsa_common.rs:2`.

Task definition: https://github.com/near/mpc/blob/f13945f6fa00457920cdb7ed687d6deca24b7234/Makefile.toml#L135-L141

### Acceptance Criteria

- `check-docs` passes `--all-features` and `--document-private-items` to `cargo doc`.
- The 24 warnings this surfaces are fixed, either by correcting the link target or escaping the markdown.

gilcu3:
Closes #4045

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -141,6 +141,7 @@ args = [
     "--no-deps",
     "--all-features",
     "--document-private-items",
+    "--locked",
 ]
 
 [tasks.check-docs.env]
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -135,7 +135,13 @@ args = ["scripts/check-third-party-licenses.sh"]
 [tasks.check-docs]
 description = "Check rustdoc for broken intra-doc links"
 command = "cargo"
-args = ["doc", "--workspace", "--no-deps"]
+args = [
+    "doc",
+    "--workspace",
+    "--no-deps",
+    "--all-features",
+    "--document-private-items",
+]
 
 [tasks.check-docs.env]
 RUSTDOCFLAGS = "-D warnings"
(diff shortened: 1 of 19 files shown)


SimonRastikian (approved):
Thanks

haiyuechen-nearone (approved):
✅

Issue #4043, opened 2026-08-04
Inject the inspectors into the foreign chain startup probe


haiyuechen-nearone:
### Background

`probe_all_providers` in `crates/foreign-chain-health-check/src/probe.rs` builds its own inspectors through real HTTP clients, so its tests stand up HTTP servers and wait on the wall clock.

inject inspectors as dependencies at `probe_all_providers`

Reporting why a provider's setup failed is split out into #4229.

### Acceptance Criteria

- The startup probing path receives the inspectors as a dependency instead of building them itself.
- Retries, attempt caps and timeouts are tested without waiting on the real clock.
- A test chooses what a provider answers, and how slowly, without starting an HTTP server.

netrome:
Thanks for raising this. Implementation note: I'm not convinced about the `near_time::Clock` construct. It mixes the concepts of capturing wall time and execution waitpoints into the same type. Most components only need one of these, and fundamentally speaking they are separate concepts, and in my experience components that interact with one of these rarely interact with the other.

My default approach (should be adjusted for the use case) would be to define a port (e.g. trait) for the time interface any specific module or piece of logic needs, e.g. `trait Sleep { fn sleep(&self, duration) -> impl Future<()> }` for wait points and `trait ReadWallTime { fn now(&self) -> Timestamp` for wall time (which iirc we're not using anywhere).

But ofc, the right approach for the implementation is up to whoever picks up this issue and should be adjusted with deeper consideration of our specific needs.

netrome:
Cousin issue in the backup CLI https://github.com/near/mpc/issues/4059

PR #4049, opened 2026-08-04
feat: EVM chains RPC probing logic

### LINKED ISSUE #4096: Probe the EVM chains for their chain id
### Background

Part of #4003. `EvmInspector` serves all six EVM chains, so one `eth_chainId` probe covers Abstract, Arbitrum, Base, BNB, HyperEVM and Polygon.

### Acceptance Criteria

- An EVM provider answering the expected chain id is reported healthy; a different chain id is reported unhealthy.
- Chain ids compare in decimal, the form they are published in, regardless of the hex quantity the RPC answers.

haiyuechen-nearone:
Closes #4096. Second slice of the identity probing series, after #4013

`EvmInspector` already serves all six chains, so `eth_chainId` covers Abstract, Arbitrum, Base, BNB, HyperEVM and Polygon at once.

### Notes for review

- **Chain ids are compared in decimal**, the form they are published and configured in, while `eth_chainId` answers a `0x` hex quantity. Parsed as `U256`, since EIP-155 permits ids wider than a `u64`. Text that is neither reads back unchanged, so the report shows what the provider actually claimed.

- **The `NetworkFingerprint` length cap moved into `NetworkFingerprint::new`**. Replaces the probe's `bounded` helper function. An answer past the cap ends in `_TRUNCATED`.

- **The injection of time and the RPC client (#4043) is deliberately not folded in**: I plan to refactor after all chains are wired up for startup probing.

### DIFF
--- a/crates/foreign-chain-health-check/src/probe.rs
+++ b/crates/foreign-chain-health-check/src/probe.rs
@@ -249,7 +249,7 @@ mod tests {
             self.chain_id.to_string()
         }
 
-        /// The form `eth_chainId` answers.
+        /// The `0xXXX` hex quantity an RPC provider answers to an `eth_chainId` request.
         fn answered(&self) -> String {
             format!("{:#x}", self.chain_id)
         }
--- a/crates/foreign-chain-health-check/src/probe.rs
+++ b/crates/foreign-chain-health-check/src/probe.rs
@@ -831,7 +831,13 @@ mod tests {
         else {
             panic!("expected the flood to read as the wrong network");
         };
-        assert!(observed.to_string().chars().count() < 100);
+        let observed = observed.to_string();
+        assert!(observed.ends_with("_TRUNCATED"), "{observed}");
+        assert_eq!(
+            observed.chars().count(),
+            NetworkFingerprint::MAX_CHARS,
+            "{observed}"
+        );
     }
 
     #[test]
--- a/crates/foreign-chain-health-check/src/probe.rs
+++ b/crates/foreign-chain-health-check/src/probe.rs
@@ -4,6 +4,13 @@
 use std::collections::BTreeMap;
 use std::time::Duration;
 
+use foreign_chain_inspector::abstract_chain::inspector::Abstract;
+use foreign_chain_inspector::arbitrum::inspector::Arbitrum;
+use foreign_chain_inspector::base::inspector::Base;
+use foreign_chain_inspector::bnb::inspector::Bnb;
+use foreign_chain_inspector::evm::inspector::{EvmChain, EvmInspector};
+use foreign_chain_inspector::hyperevm::inspector::HyperEvm;
+use foreign_chain_inspector::polygon::inspector::Polygon;
 use foreign_chain_inspector::starknet::inspector::StarknetInspector;
 use foreign_chain_inspector::{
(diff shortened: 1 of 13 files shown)


crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -224,6 +235,19 @@ mod tests {
...
+    /// Every EVM chain the probe covers: what `eth_chainId` answers on mainnet, and the decimal
+    /// chain id an operator configures.
+    const EVM_MAINNETS: [(ForeignChain, &str, &str); 6] = [
+        (ForeignChain::Abstract, "0xab5", "2741"),
+        (ForeignChain::Arbitrum, "0xa4b1", "42161"),
+        (ForeignChain::Base, "0x2105", "8453"),
+        (ForeignChain::Bnb, "0x38", "56"),
+        (ForeignChain::HyperEvm, "0x3e7", "999"),
+        (ForeignChain::Polygon, "0x89", "137"),
+    ];
```
netrome: I find this hard to read. Could you add a struct instead of using nested tuples here? This would remove the need for the comment. I see how this is used in tests, but it would be nice with a struct like:

```rust
struct EvmMainnetFixture {
    chain: ForeignChain,
    expected: &'static str,
    answered: &'static str,
}
```

There are probably better field names here.

Also, the relation between "expected" and "answered" seems to be a hex to decimal conversion. Shouldn't this be handled by logic instead of duplicating the same information in the setup?

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -41,11 +41,39 @@ pub trait ForeignChainInspector {
...
+            NetworkFingerprint::MAX_CHARS - NetworkFingerprint::CUT_SHORT_MARKER.len();
+
+        let fingerprint = fingerprint.into();
+        let mut characters = fingerprint.chars();
+        let within_cap: String = characters.by_ref().take(Self::MAX_CHARS).collect();
+        match characters.next() {
+            None => Self(within_cap),
+            Some(_) => {
+                let kept: String = within_cap.chars().take(KEPT_CHARS).collect();
+                Self(kept + Self::CUT_SHORT_MARKER)
```
netrome: Adding strings with `+` is a bit unconventional. I'd find using `format!` more readable here.
```suggestion
                Self(format!{"{kept}{Self::CUT_SHORT_MARKER}"})
```

netrome: Also see https://rust-unofficial.github.io/patterns/idioms/concat-format.html

netrome (dismissed):
A few nits, nothing blocking - though I did a relatively quick review

anodar (dismissed):
Thank you!

crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -239,15 +238,49 @@ mod tests {
...
+        /// The form an operator configures.
+        fn expected(&self) -> String {
+            self.chain_id.to_string()
+        }
+
+        /// The `0xXXX` hex quantity an RPC provider answers to an `eth_chainId` request.
+        fn answered(&self) -> String {
+            format!("{:#x}", self.chain_id)
+        }
+    }
```
netrome, line 256: Nice!

netrome (approved):
Thank you for updating 🙏

anodar (approved):

Issue #3961, opened 2026-07-24
Narrow verify-foreign-tx leader selection to chain supporters


anodar:
### Background

With https://github.com/near/mpc/issues/3910 the verify-foreign-tx provider rejects requests for chains it does not support, but leader election still treats every alive participant as eligible: a non-supporter can be elected leader for a verify-foreign-tx request, immediately reject it, and burn the attempt — delaying the request until a supporter happens to be selected. Requests for chains without a signing quorum keep being attempted until they expire.

### Acceptance Criteria

- The request queue supports an optional eligible-leaders refiner, the verify-foreign-tx queue narrows the eligible-leader set to the requested chain's supporters
- Leader election for other request types is unchanged.

Issue #377, opened 2025-04-22
Implement robust recovery from lost assets


robin-near:
Design: https://docs.google.com/document/d/1aVu7PreP8jz6LUZyYvog4oP8Wp766fAedQOiLNaxrjY/edit?tab=t.0

Alternative design is to reuse current double queue but add the handling of lost assets.

netrome:
Robustness isn't happening anytime soon, so we'll need to revisit this as part of our DKG scalability work.

netrome:
Retaining as P2.

kevindeforth:
A solution to this problem should also take into account foreign chain transactions and their heterogeneous requirements on participants.

PR #4054, opened 2026-08-04
feat(node): chain-compatible presignature selection for verifying foreign transactions

### LINKED ISSUE #3959: Chain-compatible presignature selection for verify-foreign-tx
### Background

verify-foreign-tx provider gates requests on the available-chains watch channel, but the leader still acquires an arbitrary presignature via `take_owned`. 

Presignature's co-owners may not all support the requested chain: a non-supporting follower rejects the attempt against its supporters snapshot, the signing round fails, and the presignature is wasted. 

The leader must instead select a presignature co-owned exclusively by participants that support the request's chain.

### User Story

_No response_

### Acceptance Criteria

- The verify-foreign-tx leader only takes presignatures whose participants all support the requested chain (alive participants intersected with the supporters map from https://github.com/near/mpc/issues/3852).
- A request for a chain whose supporter set cannot cover any presignature does not consume presignatures.

_No response_

anodar:
Extends queue with `take_first_matching` and `ingest`

Unlike `take`, `take_first_matching` does full scan (accepted degradation risk when nodes drop support for given chain).

Closes: https://github.com/near/mpc/issues/3959


crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -55,19 +59,51 @@ impl VerifyForeignTxProvider {
...
+        let chain_supporters: HashSet<ParticipantId> = {
+            let snapshot = self.supporters_by_foreign_chain.borrow();
+            ensure_chain_is_available(&snapshot, &foreign_tx_request.request).inspect_err(
+                |_| metrics::MPC_NUM_VERIFY_FOREIGN_TX_UNAVAILABLE_CHAIN_REJECTIONS.inc(),
+            )?;
+            snapshot.get(&requested_chain).cloned().unwrap_or_default()
+        };
+
+        // Owned presignatures always include this node, so a non-supporting
+        // leader could never match one. Bail before waiting on the take.
```
haiyuechen-nearone: I found this comment a bit confusing. Would this be correct? I think it read a bit better

```suggestion
        // Election is not chain support aware, so a non-supporting node can be
        // assigned leader for a request. Owned presignatures always include this
        // node, so `take_owned_matching` below would never resolve.
        // TODO(#3961): narrow election to chain supporters.
```

crates/node/src/assets.rs
```diff
@@ -165,6 +168,35 @@ impl<T, CondVal: Default + Eq> ColdQueue<T, CondVal> {
...
+    }
+
+    /// Removes and returns the first element satisfying both the standing
+    /// condition and the caller-supplied `cond_val`, shifting the barriers
+    /// that lie past the removed position.
+    pub(self) fn take_first_matching(&mut self, cond_val: &CondVal) -> Option<(UniqueId, T)> {
+        self.update_condition_value_if_due();
+        let pos = self.cold_queue.iter().position(|(_, val)| {
+            (self.condition)(&self.last_condition_value, val) && (self.condition)(cond_val, val)
+        })?;
```
haiyuechen-nearone: Is scanning the full cold_queue required here? If I understand correcly we only need to scan up to `cold_available`, which can be less than the length of the queue if some node went offline.


```suggestion
    pub(self) fn take_first_matching(&mut self, cond_val: &CondVal) -> Option<(UniqueId, T)> {
        self.update_condition_value_if_due();
        let pos = self
            .cold_queue
            .iter()
            .take(self.cold_available)
            .position(|(_, val)| {
                (self.condition)(&self.last_condition_value, val)
                    && (self.condition)(cond_val, val)
            })?;
```

haiyuechen-nearone (changes requested):
Feel free to dismiss the suggestions if I misunderstood

crates/node/src/assets.rs
anodar: You're right, queue part that's confirmed not to support chain can be skipped. Addressed in [f671291](https://github.com/near/mpc/pull/4054/commits/f671291e3730db523c442f108b1ff39a83fee281)

crates/node/src/providers/verify_foreign_tx/sign.rs
anodar: Reworded in [f671291](https://github.com/near/mpc/pull/4054/commits/f671291e3730db523c442f108b1ff39a83fee281)

haiyuechen-nearone (dismissed):

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -55,19 +59,53 @@ impl VerifyForeignTxProvider {
...
+                Ok(taken) => taken,
+                Err(_) => {
+                    metrics::MPC_NUM_VERIFY_FOREIGN_TX_PRESIGNATURE_WAITS.inc();
+                    tracing::warn!(
+                        ?requested_chain,
+                        "no chain-compatible presignature available, waiting"
+                    );
+                    take.await
+                }
+            };
```
netrome: This is generic code for waiting for a timeout, did you consider breaking it out so we can keep the `make_verify_foreign_tx_leader` function cleaner?

crates/node/src/assets.rs
```diff
@@ -270,6 +306,31 @@ where
         }
     }
 
+    pub async fn take_owned_matching(&self, cond_val: CondVal) -> (UniqueId, T) {
+        loop {
+            {
+                let mut cold = self.cold_queue.lock().unwrap();
+                while let Some(Ok((id, value))) = self.hot_receiver.recv_async().now_or_never() {
```
netrome, line 312: `.recv_async().now_or_never()` feels weird. Wouldn't `.try_recv()` be sufficient?

crates/node/src/assets.rs
```diff
@@ -270,6 +306,31 @@ where
...
+                }
+            }
+            // If the cold queue is exhausted, wait for a new element.
+            tokio::select! {
+                _ = self.clock.sleep(near_time::Duration::seconds(1)) => {
+                    continue;
+                }
+                received = self.hot_receiver.recv_async() => {
+                    // can't fail, because self keeps a sender.
+                    let (id, val) = received.unwrap();
```
netrome: Nit: Would be better to use expect here to convey this broken invariant if this were ever to fail
```suggestion
                    let (id, val) = received.expect("should never fail because self keeps a sender");
```

crates/node/src/assets.rs
```diff
@@ -569,6 +630,20 @@ where
...
+    /// the supplied `eligible` set. Blocks indefinitely if none becomes
+    /// available.
+    /// Callers are expected to enforce their own timeout.
+    pub async fn take_owned_matching(&self, eligible: Vec<ParticipantId>) -> (UniqueId, T) {
+        let (id, val) = self.owned_queue.take_owned_matching(eligible).await;
+        let mut update = self.db.update();
+        update.delete(self.col, &self.make_key(id));
+        update
+            .commit()
+            .expect("Unrecoverable error writing to database");
```
netrome, line 643: How do we know this is an unrecoverable error? This error is an opaque string crossing an FFI boundary from RocksDB, are we sure all of these errors are unrecoverable?

I'd prefer to return the error here (wrapped in a more helpful error type so it can be explicitly matched) and let the caller decide how to handle it.

netrome (changes requested):
Thanks for the well-sliced PR. I find it difficult to navigate and understand the asset queue, so I'm happy I don't have to consider anything else here. I still don't understand everything going on here but the intention is clear, and most of the code looks reasonable.

One blocker though: The .expect() on the rocksdb commit error is giving me a bit of a pause - couldn't we handle this error instead.

crates/node/src/assets.rs
anodar: `.recv_async().now_or_never()` returns Option, whereas `.try_recv()` will return Result, and we don't really care for that error. 

We could do `.try_recv()` and ignore the error but that would look more awkward imo.

Edit: this follows pattern used in `maybe_discard_owned`.

crates/node/src/assets.rs
anodar: This follows same pattern used throughout this file, in: `take_owned`, `add_owned`, `maybe_discard_owned`, `add_unowned`, `take_unowned_inner`.

Created https://github.com/near/mpc/issues/4090 and added todo here.

crates/node/src/assets.rs
anodar: Change here and above (where this was copy/pasta from).

crates/node/src/providers/verify_foreign_tx/sign.rs
anodar: Done in: [739fe13](https://github.com/near/mpc/pull/4054/commits/739fe13b64954282404de50e28d253b56ea2a0fa)

crates/node/src/assets.rs
netrome: We're already swallowing an error. This would just turn the while let to `while let Ok(...)` instead of the nested `while let Some(Ok(...))` which would be cleaner imo.

> Edit: this follows pattern used in maybe_discard_owned.

Perhaps worth changing it there as well then. I haven't looked though, could be justified but I'm not buying the `Option` vs `Result` argument. A single call returning a `Result` that conveys whether the channel is empty or the sender disconnected sounds strictly better than calling two separate methods to achieve the same as a nested Option<Result<>>

crates/node/src/providers/verify_foreign_tx/sign.rs
netrome: Thanks! :pray:

crates/node/src/assets.rs
netrome: Thanks :pray:

crates/node/src/assets.rs
netrome: Fair enough. I'm skeptical about those places too. I would expect this error to be very unlikely, but this feels potentially brittle.

netrome (dismissed):
Thanks for the updates, and creating the follow-up 🙏

haiyuechen-nearone (dismissed):

crates/node/src/assets.rs
metalurgical: I think Result is preferable to Option in any case where an Option could be otherwise avoided.

Simplest way to explain this is using a bool:
1:
Result<bool> ->
Ok(true, false), Err(e), e is an error, e can then be handled separately. Get to keep the `?` operator.

2:
Option<bool> ->
Some(true, false), None: but is this now true, false, undefined, or error here? Can't use `?` operator, must essentially rethrow.

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -55,21 +59,51 @@ fn build_signature_request(
...
         id: SignatureId,
     ) -> anyhow::Result<((dtos::ForeignTxSignPayload, Signature), VerifyingKey)> {
         let foreign_tx_request = self.verify_foreign_tx_request_store.get(id).await?;
+        let requested_chain = foreign_tx_request.request.chain();
+
+        // Also checked in `execute_foreign_chain_request`; resolved early here
+        // because the supporter set scopes which presignature may be taken. An
+        // availability flip after the take still costs one presignature.
+        let chain_supporters: HashSet<ParticipantId> = {
+            let snapshot = self.supporters_by_foreign_chain.borrow();
```
kevindeforth: Is it necessary to prevent the sender from writing? If not, lets keep the borrow short and clone the value, it will make it harder to accidentally break something in the future.

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -55,21 +59,51 @@ fn build_signature_request(
...
     pub(crate) async fn make_verify_foreign_tx_leader(
         &self,
         id: SignatureId,
     ) -> anyhow::Result<((dtos::ForeignTxSignPayload, Signature), VerifyingKey)> {
         let foreign_tx_request = self.verify_foreign_tx_request_store.get(id).await?;
+        let requested_chain = foreign_tx_request.request.chain();
+
+        // Also checked in `execute_foreign_chain_request`; resolved early here
+        // because the supporter set scopes which presignature may be taken. An
+        // availability flip after the take still costs one presignature.
```
kevindeforth: if we reference methods or functions, let's link to them please, as per our engineering standards. 
https://github.com/near/mpc/blob/c8ae0e14c292eb474bc526bf40c7495f777f8b99/docs/engineering-standards.md?plain=1#L359-L372

I would also suggest to re-write this comment, to make its intent clearer. As a reader, I am not sure what it is trying to tell me and how it is relevant to this piece of code.

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -55,21 +59,51 @@ fn build_signature_request(
...
+        let chain_supporters: HashSet<ParticipantId> = {
+            let snapshot = self.supporters_by_foreign_chain.borrow();
+            ensure_chain_is_available(&snapshot, &foreign_tx_request.request).inspect_err(
```
kevindeforth: non-blocking: I notice now that `ensure_chain_is_available` only requires the `ForeignChain`, which we already extract on line 85. We cold probably just pass that instead of a reference to the the full `ForeignChainRpcRequest`.

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -55,21 +59,51 @@ fn build_signature_request(
...
-        // availability flip between the two checks still costs one presignature.
-        ensure_chain_is_available(
-            &self.supporters_by_foreign_chain.borrow(),
-            &foreign_tx_request.request,
-        )
-        .inspect_err(|_| metrics::MPC_NUM_VERIFY_FOREIGN_TX_UNAVAILABLE_CHAIN_REJECTIONS.inc())?;
+        // Election is not chain support aware, so a non-supporting node can be
+        // assigned leader for a request. Owned presignatures always include this
+        // node, so `take_owned_matching` below would never resolve.
+        // TODO(#3961): narrow election to chain supporters.
```
kevindeforth: There are some issues with this comment:
- it references a function without linking to it, risking the comment to get stale (c.f. our engineering standards linked above)
- It lacks precision: What "Election" is it referring to?
- what does it mean with "chain support aware"?

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -55,21 +59,51 @@ fn build_signature_request(
...
+        let my_participant_id = self.ecdsa_signature_provider.my_participant_id();
+        if !chain_supporters.contains(&my_participant_id) {
+            metrics::MPC_NUM_VERIFY_FOREIGN_TX_UNAVAILABLE_CHAIN_REJECTIONS.inc();
+            anyhow::bail!("this node does not support the requested chain {requested_chain:?}");
+        }
```
kevindeforth: I guess keeping #3961 open for a bit isn't a huge problem, as we currently expect all our nodes to support all the chains and thus, we shouldn't really end up in line 105?

I would still suggest to make it P0 once we merge this PR.

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -27,6 +30,7 @@ use near_mpc_contract_interface::types::{Payload, Tweak};
 use tokio::time::{Duration, timeout};
 
 const FOREIGN_CHAIN_INSPECTION_TIMEOUT: Duration = Duration::from_secs(5);
+const PRESIGNATURE_TAKE_GRACE_PERIOD: Duration = Duration::from_secs(1);
```
kevindeforth, line 35: These time-outs are a bit scary. 1 second seems potentially tight, yet, I suppose there exists another timeout after which other nodes start to attempt the computation. Do you know how long they wait for the leader to complete the computation?

kevindeforth (commented):
We had a little attestation expiration emergency which interrupted me mid-review. Will pick this up tomorrow.

crates/node/src/providers/verify_foreign_tx/sign.rs
anodar: It's not a timeout. It's equivalent of `take_owned_matching().await` except it also just logs warning whenever that takes longer than 1 sec. But we still await.

crates/node/src/providers/verify_foreign_tx/sign.rs
anodar: 3961 is immediate followup on this, making it P0 as I merge this sounds good.

crates/node/src/providers/verify_foreign_tx/sign.rs
anodar: That was the wording proposed in this review: https://github.com/near/mpc/pull/4054#discussion_r3728404363

My original (arguably unclear) wording was:
```
// Owned presignatures always include this node, so a non-supporting
// leader could never match one. Bail before waiting on the take.
```




- In the proposal `Election` refers to `Selection`.
- "chain support aware" - I interpreted wording proposal as "Leader selection does not take into account who supports which chain, so selected leader may not support the chain". Because `get_requests_to_attempt` compiles list of `eligible_leaders` based only on availability so far not per chain support.

Reverted back to my wording and elaborated it a bit more to (hopefully) make it more clear in [fe7b7c1](https://github.com/near/mpc/pull/4054/commits/fe7b7c17bc567b0949a8ca6d05f2ba86b636235c)

```
        // Owned presignatures always include current (leader) node, so it would
        // never be able to find a presignature from presignatures it owns
        // where all participants (including itself) support the chain.
        // TODO(#3961): narrow leader selection to only chain supporters.
```

crates/node/src/providers/verify_foreign_tx/sign.rs
anodar: Addressed in [fe7b7c1](https://github.com/near/mpc/pull/4054/commits/fe7b7c17bc567b0949a8ca6d05f2ba86b636235c)

crates/node/src/providers/verify_foreign_tx/sign.rs
anodar: Dropped the comment entirely as it introduced more confusion than brings benefit.

To clear it up, we check supporters twice:

- first at the beginning of make_verify_foreign_tx_leader when checking supporters to return from function if current participant isn't a supporter
- second in execute_foreign_chain_request. 

Technically we don't need to do `ensure_chain_is_available` in both places as `make_verify_foreign_tx_leader` calls `execute_foreign_chain_request`, but it doesn't hurt, shortly takes read lock on the supporters receiver. With negligible benefit: theoretically if the support changes in between this would prevent waste of presignature).

I chose to leave it for redundancy as overhead is minimal although I can be convinced otherwise as it doesn't really matter too much.

crates/node/src/providers/verify_foreign_tx/sign.rs
anodar: Good point, no reason to hold the lock. Changed to clone in [fe7b7c1](https://github.com/near/mpc/pull/4054/commits/fe7b7c17bc567b0949a8ca6d05f2ba86b636235c)

crates/node/src/providers/verify_foreign_tx/sign.rs
kevindeforth: Ah, I see, yeah, completely missed that!

crates/node/src/assets.rs
kevindeforth: Fwiw, I also prefer the `try_recv()` on the channel over the `now_or_never()` on the future.
I thought that maybe `now_or_never()` is preferred over `try_recv()` for performance reason, but I quizzed claude about it and it told me `try_recv()` is more efficent.

I would have loved to delve into the code, but this review has been sitting for too long, so I abstained from that. So, I feel like it's fine to share some unfiltered LLM output here and hope that in case Claude got it wrong, someone else will correct me:


| Per empty poll        | `try_recv()` | `recv_async().now_or_never()` |
|-----------------------|--------------|-------------------------------|
| Chan lock acquisitions| 1            | 2 (`recv` + `reset_hook` on drop) |
| Allocations           | 0            | 1 `Arc<Hook>` + waker clone   |
| `chan.waiting`        | untouched    | `push_back`, then `retain` removal |
| Work under lock       | O(1)         | **O(waiting)** ptr-compare scan |
| Spurious wake         | no           | possible (`try_wake_receiver_if_pending`) |
| Returns               | `Result<T, TryRecvError>` | `Option<Result<T, RecvError>>` |

note: claude says that for non-empty or dropped channels, they are equivalent.

crates/node/src/assets.rs
```diff
@@ -165,6 +168,39 @@ impl<T, CondVal: Default + Eq> ColdQueue<T, CondVal> {
...
+            self.cold_ready += 1;
+            self.cold_available += 1;
+        } else {
+            self.cold_queue.push_back((id, value));
+        }
+    }
+
+    /// Removes and returns the first element satisfying both the standing
+    /// condition and the caller-supplied `cond_val`, shifting the barriers
+    /// that lie past the removed position.
```
kevindeforth: > shifting the barriers that lie past the removed position.

I found this comment quite useful, but it requires context knowledge. It would be good to mention to the "barriers" that are getting shifted by name:

```
shifting the barriers  [`ColdQueue::cold_ready`] and [`ColdQueue::cold_available`]
```

crates/node/src/assets.rs
```diff
@@ -270,6 +305,30 @@ where
...
+            tokio::select! {
+                _ = self.clock.sleep(near_time::Duration::seconds(1)) => {
+                    continue;
+                }
+                received = self.hot_receiver.recv_async() => {
+                    let (id, value) = received.expect("should never fail because self keeps a sender");
+                    self.cold_queue.lock().unwrap().ingest(id, value);
+                }
+            }
+        }
```
kevindeforth, line 360: Hmm, I think this method surfaces some severe limitations of the existing design of our queue.

In this function:
1. We drain the `hot_receiver` channel and store its elements in the cold queue
2. We iterate over the elements in the cold queue until we find an element that satisfies our conditions.
3. If no such element exists, we wait for another element to arrive in the receiver channel and then add it to the queue. Once it arrives (or if we timeout), we re-start the loop.

There are two problems here:
1. We are double-handling elements coming from the `hot_receiver`. The last sub-sequence of elements not satisfying our conditions and arriving over the `hot_receiver` will be in the beginning of the cold queue. We re-iterate over that sub-sequence on line 315 in `take_first_matching`.
2. We are competing with other tasks waiting in `take_owned`. There, the task first inspects the cold-queue and if it doesn't find a suitable asset, it awaits on the receiver. It ~never returns inspecting the cold-queue~ (_sorry, it does after 1s timeout, it's a loop_). The condition for `take_owned` is a subset of the conditions of `take_owned_matching`, so a task waiting in `take_owned_matching` may move elements from the hot receiver to the cold queue that other tasks are awaiting on the hot receiver.


At the very least, we need a re-write of `take_owned`, but I suspect we need to improve our asset queue in general.

I am thinking that we need some sort of dispatcher / waiting queue for tasks that are in need of assets.


_Edit: after further inspection, I think it's less about `take_owned_matching` starving `take_owned`, and more about the reverse (though, `take_owned_matching` can force latency upon `take_owned` and I think we may see issues there with leaders timing out...)._

kevindeforth (dismissed):
Unfortunately, we are running into limitations of our existing asset queue implementation.
The currently proposed solution may lead to delayed signature responses in case the network does not have a uniform foreign chain tx support (https://github.com/near/mpc/pull/4054#discussion_r3819116342).

A proper solution will probably require a refactor of the asset queue, so it's a trade-off:
1. We ship it and gain more robust foregin-chain tx requests, at the risk of delayed signature responses and possibly at the cost of future refactors.
2. We take time to refactor the queue now and solve other existing problems with it (such as offline participants), at the cost of more stringent requirements wrt to foreign chain configs in the MPC network.

I think I have a preference for 2 here, as we now have much better tooling for the foreign chain configs and I think the asset queue has been a thorn in our eye for a long time.

But I don't have full visibility on the demands of the community / business.
In any case, I think his is @netrome's call, as it's essentially a choice between shipping a feature now at the cost of increased tech-debt, or addressing tech-debt at the cost of delaying the feature.


Remark: the required refactor might be smaller than it might seem here. I haven't thought too much about possible designs yet, but I can if that is desired.

crates/node/src/assets.rs
```diff
@@ -307,20 +314,32 @@ where
...
+            if let Some(taken) = taken {
+                return taken;
+            }
+
             // If the cold queue is exhausted, wait for a new element.
             tokio::select! {
                 _ = self.clock.sleep(near_time::Duration::seconds(1)) => {
                     continue;
                 }
+                _ = cold_queue_changed => {
```
anodar: This might seem no-op as we don't notify from `take_owned` but when processing request for foreign chain A we might move asset into cold queue and notify another task processing request for foreign chain B for which this asset may be suitable.

crates/node/src/assets.rs
```diff
@@ -233,6 +234,7 @@ where
     hot_receiver: flume::Receiver<(UniqueId, T)>,
     cold_queue: Arc<Mutex<ColdQueue<T, CondVal>>>,
     clock: Clock,
+    cold_queue_changed: tokio::sync::Notify,
```
anodar: Not adding anything indicating that something is `ready` here, intentionally keeping the name generic "changed" since even though values is added to the front of the queue, it may still not be suitable for the other foreign chain requests.

crates/node/src/assets.rs
anodar: Addressed in [92d15f5](https://github.com/near/mpc/pull/4054/commits/92d15f578c736d2f513c20028274076b615f485b)

anodar:
As per our offline discussion, adding notifies to mitigate front-running problem.

crates/node/src/assets.rs
```diff
@@ -270,6 +312,43 @@ where
...
     }
 
+    pub async fn take_owned_matching(&self, cond_val: CondVal) -> (UniqueId, T) {
+        loop {
+            let cold_queue_changed = self.cold_queue_changed.notified();
+            let (taken, ingested) = {
+                let mut cold = self.cold_queue.lock().unwrap();
+                let mut ingested = false;
+                while let Some(Ok((id, value))) = self.hot_receiver.recv_async().now_or_never() {
+                    cold.ingest(id, value);
```
kevindeforth, line 330: @anodar I think we touched on this, but I don't remember exactly. What was the reason for not checking if `value` can be used by the caller and early-return here, instead of ingesting and then re-inspecting every element in the cold-queue?

crates/node/src/assets.rs
kevindeforth, line 307: Would be good to add a comment explaining the invariant that allows us to not notify here.

crates/node/src/providers/verify_foreign_tx/sign.rs
kevindeforth: I think this can be simplified to: 
```
// If we don't support the foreign chain, we can't lead the computation.
// TODO(#3961): narrow leader selection to only chain supporters.
```

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -85,7 +117,25 @@ impl VerifyForeignTxProvider {
         let keyshare = self
             .ecdsa_signature_provider
             .keyshare(foreign_tx_request.domain_id)?;
-        let (presignature_id, presignature) = keyshare.presignature_store.take_owned().await;
+        // Aliveness is enforced inside the store: the take requires the
+        // presignature's participants to satisfy the (continuously refreshed)
+        // alive condition and be a subset of the chain's supporters. Since
+        // presignature generation is not chain-aware, a compatible one may
+        // not exist yet, count and log when the take has to wait.
+        let (presignature_id, presignature) = await_with_slow_hook(
```
kevindeforth, line 119: I think the wording "aliveness is enforced inside the store" is a bit misleading
Aliveness depends on `chain_supporters`, which is a snapshot subject to change, so the store can't give us that guarantee.

The only thing the store can do is to ensure the borrowers of the returned asset have been marked "alive" recently, but that's kind of expected from the store and maybe not worth a comment here.

> Since presignature generation is not chain-aware, a compatible one may not exist yet, count and log when the take has to wait.

I am not sure why we need to mention that presignature generation is not chain-aware. It's not something this code can guarantee. Besides, even if it was chain-aware, we may need to wait here, as presignature generation depends on other tasks.

crates/node/src/assets.rs
```diff
@@ -165,6 +168,40 @@ impl<T, CondVal: Default + Eq> ColdQueue<T, CondVal> {
         self.cold_queue.push_back((id, value));
         ColdQueueAddIfNotSatisfiedResult::Enqueued
     }
+
+    /// Adds an element to the cold queue unconditionally, never returned.
+    pub(self) fn ingest(&mut self, id: UniqueId, value: T) {
```
kevindeforth: What does the comment mean with "never returned"?

crates/node/src/assets.rs
kevindeforth: Maybe instead of `*changed`, we can call it `*new_elements`?
We don't notify in case we take elements, only in case we add new.

docs/asset-generation.md
```diff
@@ -85,7 +86,9 @@ operations:
 
 ### Properties of an asset taken from the queue
 
-When `take_owned()` returns an asset, the following holds:
+When `take_owned()` or `take_owned_matching(eligible)` returns an asset,
+the following holds (`take_owned_matching` additionally guarantees all
+borrowers are in `eligible`):
```
kevindeforth: Lets add this as an explicit third point below.

docs/asset-generation.md
```diff
@@ -163,6 +166,11 @@ A `VecDeque` divided into three logical regions by two barriers:
 - **Non-satisfying** (cold_available..len): offline assets. Skipped by
   `take()`, targeted by `discard()`.
 
+`take_first_matching()` (backing `take_owned_matching`) is the exception
+to front/back access: it removes the first online asset also matching the
+caller's set from *any* position before `cold_available`, shifting the
+barriers past the removal point down by one.
```
kevindeforth: this reeds odd, better to add  a `## take_first_matching` to stay consistent with the rest of the doc.

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -111,11 +123,15 @@ dropped a chain to zero availability.
...
-every chain.
+A node that's up but not covering a chain `C` shrinks the eligible presignature
+pool for `C`: presignatures it co-owns are excluded from selection for `C`'s
+requests (they stay usable for chains the node does cover), and the smaller
+the supporter set, the fewer generated presignatures qualify (see
+[Participant election](#participant-election)). Its **triples are
+not lost** — they're shared across domains and stay in use, so triples go
+offline only if the node is genuinely down. Mitigation is operational:
+alerting keeps coverage high and operators are expected to configure every
+node for every chain.
```
kevindeforth: I am not sure why we need to mention this here, it's not really a "trade-off", seems like an LLM thing.

docs/design/calculating-supported-foreign-chains.md
```diff
@@ -71,9 +71,21 @@ disagreement being terminal. Tracked in [#3477](https://github.com/near/mpc/issu
...
+chain-aware — a non-supporting leader rejects the attempt instead of serving
+it — tracked in [#3961](https://github.com/near/mpc/issues/3961).
+
+Residual limitation, accepted as-is: presignature generation remains
+liveness-driven, so participant sets are random `t`-subsets of the alive set.
+When not all alive participants support `C`, many presignatures are
+incompatible, and a request may wait until a compatible one is generated or
+the request times out. We do not plan chain-aware generation, the
+`mpc_num_verify_foreign_tx_presignature_waits` metric (plus a warning log)
+makes such waits observable.
```
kevindeforth: would be good to reference the existing issue on asset-queue and generation.

kevindeforth (commented):
Thanks!
A few remaining comments to polish and clarifying questions, the most significant about the `take_owned_matching` method.
But looks good otherwise. Please make sure to address the existing open threads.

docs/design/calculating-supported-foreign-chains.md
anodar: Added.

docs/design/calculating-supported-foreign-chains.md
anodar: What you highlighted isn't something introduced in this PR. It's word by word same [as before](https://github.com/near/mpc/commit/fb28b6414c2ee58e61396ab93edff83284ee2f88). The change only explains participant selection that now we also take into account chain support not just aliveness.

In the discussions, it came up that in the worst case it could be just threshold number of participants supporting the chain. In that case we say we support the chain but effectively it will be (almost) impossible to find asset exactly with that set of participants. 

Even though we don't talk about queue internals here, I'd say it's still worth calling out this here that eligible set of assets shrinks quite a bit when node doesn't support the chain and tradeoff is that we need to manually monitor and alert node operators when they stop support for some chain.

docs/asset-generation.md
anodar: Done.

docs/asset-generation.md
anodar: Done.

crates/node/src/assets.rs
anodar: That part was explaining this `ingest` function in contrast with `add_if_condition_not_satisfied` which says:

```
If the condition is satisfied, instead of adding, it is returned.
```

so I wanted to emphasize here that this is unconditional. Dropped "Never returned part"

crates/node/src/providers/verify_foreign_tx/sign.rs
anodar: Dropped the comment since it adds more confusion than being helpful.

crates/node/src/assets.rs
anodar: Added explanation.

crates/node/src/assets.rs
anodar: It was just a design choice to drain the hot receiver and move taking mainly to cold queue but can't argue that it's better. So returning early as you suggest.

crates/node/src/assets.rs
anodar: Done.

crates/node/src/assets.rs
anodar: Returning early now when element from hot_receiver meets the criteria. As for dispatcher/waiting queue for tasks as discussed offline will be addressed as part of https://github.com/near/mpc/issues/377 instead.

crates/node/src/providers/verify_foreign_tx/sign.rs
anodar: Done.

crates/node/src/assets.rs
anodar: > I think Result is preferable to Option in any case where an Option could be otherwise avoided.

I would argue the opposite. Option should be used whenever you don't care about the error and it's part of expected outcome. E.g. in this case, we don't care about logging the error that queue is empty, we just care whether there's an element to receive or not.

@kevindeforth that sounds convincing. Switched to `try_recv` instead in here and in existing `maybe_discard_owned`. Although didn't validate the table myself either.

crates/node/src/assets.rs
kevindeforth: Ah yeah, I think there was another reason that is now voided due to the notifier-logic. Returning early seems like the right approach here, as we want to avoid handling the value twice.

crates/node/src/assets.rs
```diff
@@ -270,6 +315,51 @@ where
...
+                    ingested = true;
+                }
+                (
+                    taken.or_else(|| cold.take_first_matching(&cond_val)),
+                    ingested,
+                )
+            };
+
+            if ingested {
+                self.cold_queue_new_elements.notify_waiters();
```
kevindeforth, line 340: Note that we are waking ourselves here. Not ideal, but it's okay.

kevindeforth (approved):
Thank you!

haiyuechen-nearone (approved):

PR #4056, opened 2026-08-04
ci: report disguised file content to code scanning

### LINKED ISSUE #4037: ci: scan changed files for malicious code and disguised content

pbeza:
Closes #4037. Supersedes #4038.

Classifies the **content** of every changed file with [magika](https://github.com/google/magika) and reports any file whose bytes are code or an executable while its extension claims an inert asset. That is the implant we actually got: JavaScript named `fa-solid-400.woff2`, padded with 1700 spaces to sit off-screen in the diff, force-pushed onto an open PR's head branch. Extension-filtering scanners never open such a file, which is why the disguise worked.

One Python script, no shell, no bespoke gating:

- `uv` installs magika from the script's own inline metadata, so the pinned version lives in one place
- findings go out as SARIF to code scanning, so severity, dismissal and merge blocking are ruleset config rather than code
- the script exits 0 on a finding and 2 only when it *could not scan* — a base or head commit missing from the clone, or a file magika cannot read. Each has a test, because each otherwise looks clean
- `git diff -z`, because `core.quotePath` quotes bytes above 0x80 and the quoted literal matches no file on disk, silently dropping a homoglyph-named payload

## Needs a repo admin

Without the first, this reports but blocks nothing.

- [ ] Add a `code_scanning` rule to the `default` ruleset on `main` (`repos/near/mpc/rulesets/2666105`): `tool: disguised-content`, `alerts_threshold: errors`, `security_alerts_threshold: none`. It also blocks while analysis runs or if the tool is missing, so a job that never reports cannot look clean — which is why it is used instead of a required status check, since the skipped same-repo-PR run publishes the same check name and GitHub counts a skip as satisfied.
- [ ] Set `require_last_push_approval: true` in the same ruleset. Currently `false`: stale reviews are already dismissed on push and 2 approvals are required, but whoever pushed the new head can still supply one of them, which is the vector here.

## Deviation from #4037

The issue asks for YARA and magika; this ships magika only, and I think the issue is what should change. All 54 rules in GuardDog's pack carry `path_include` and **none names `.rs`** — coverage is Python (41 rules), JS/TS (41), Go (18), Ruby (10), shell (2). Running them over everything anyway is off-label, which is what forced 18 of 54 to advisory in #4038 plus a 54-line allowlist re-measured on every bump. For this payload the YARA hit was redundant: magika alone reports "content is javascript, name claims a font".

Worth its own issue: this repo depends on 11 third-party Actions and checks none. GuardDog's `github-action` ecosystem covers that natively, as an addition rather than a replacement.


gilcu3:
@pbeza are you still working on this one?

pbeza:
> @pbeza are you still working on this one?

I deprioritized this one, so let's close it. Thanks for catching it!

PR #4062, opened 2026-08-05
chore: e2e test in cargo-make now compatible with nextest filters

### LINKED ISSUE #4027: cargo make e2e-tests should forward arguments and filters to cargo nextest
### Background

`crates/e2e-tests/README.md:319` and `CLAUDE.md:39` both document `cargo make e2e-tests-skip-build -- <name>` as a way to run a single E2E test, but it does nothing. The `_run-e2e-logic` task (`Makefile.toml:236-255`) hardcodes its nextest `args` and never references `${@}`, and cargo-make does not append trailing arguments automatically, so they are silently discarded and the full suite runs. Given that E2E runs take a long time, being unable to filter is painful during development, and the silent drop makes it look like the filter was honored.

The fix needs to handle cargo-make's argument quirk: `${@}` expands to the arguments *prefixed with a literal `--`*. nextest accepts that for plain substring filters, but the leading `--` makes it treat everything after it as test-binary arguments, so flags break:

    error: failed to parse test binary arguments `-E`: arguments are unsupported

`CARGO_MAKE_TASK_ARGS` (semicolon-separated, preserves spaces within a single argument) is the reliable source; the leading `--` element must be stripped before forwarding.

### Acceptance Criteria

- `cargo make e2e-tests -- <filter>` and `cargo make e2e-tests-skip-build -- <filter>` forward the arguments to `cargo nextest run`.
- Both substring filters (`request_lifecycle`) and nextest flags (`-E 'test(...)'`, `--no-capture`, `--retries=0`, `-j1`) reach nextest unmangled, including arguments containing spaces.
- Passing no arguments keeps today's behavior (whole suite, `--profile ci-e2e`).
- Arguments are not passed to the build dependency tasks.
- `crates/e2e-tests/README.md` and `CLAUDE.md` document the working invocation, including how to pass flags.

gilcu3:
Closes #4027

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -36,8 +36,10 @@ cargo insta accept                                          # Accept all pending
 ```bash
 cargo make e2e-tests                            # Build required binaries and run all E2E tests
 cargo make e2e-tests-skip-build                 # Reuse binaries from a previous run
-cargo make e2e-tests-skip-build -- <name>       # Run a single test (filter passed to nextest)
+cargo make e2e-tests-skip-build <name>          # Run tests matching a substring filter
+cargo make e2e-tests-skip-build -E 'test(request_lifecycle)' --no-capture  # Forward nextest flags
 ```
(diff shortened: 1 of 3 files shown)


gilcu3:
Responding to the review above, I tested commands as currently explain in the README and Agents and they work. The `--` was actually not needed, and that's why the PR diverges from the issue in that regard, but it actually also works.

anodar (approved):
Nice, thanks!

haiyuechen-nearone (approved):

PR #4067, opened 2026-08-05
feat: probe Bitcoin provider by genesis block hash

### LINKED ISSUE #4091: Probe Bitcoin for its genesis block hash
### Background

Part of #4003. Bitcoin has no chain id, so the identity probe fingerprints the network by the genesis block hash, asking `getblockhash` for height 0, the same method the inspector uses for its canonical chain check.

### Acceptance Criteria

- A Bitcoin provider answering the expected genesis hash is reported healthy. If the provider answers anything else or does not answer is reported unhealthy.
- The configured expectation matches regardless of hex case or a leading `0x`.

haiyuechen-nearone:
Closes #4091.

Bitcoin has no chain id, so the fingerprint is the genesis block hash and the probe is `getblockhash` at height 0, the method the inspector's canonical chain check already uses.

### Notes for review

- **Nothing is trimmed.** Leading zeros are digits of the hash and the genesis hash opens with ten of them, so normalization is lowercase with an optional `0x` prefix stripped. Text that is not 64 hex characters reads back unchanged, so the report shows what the provider actually claimed.

### DIFF
--- a/crates/foreign-chain-health-check/src/probe.rs
+++ b/crates/foreign-chain-health-check/src/probe.rs
@@ -348,6 +348,13 @@ mod tests {
         }
     }
 
+    fn bitcoin_only(config: ForeignChainConfig) -> ForeignChainsConfig {
+        ForeignChainsConfig {
+            bitcoin: Some(config),
+            ..Default::default()
+        }
+    }
+
     fn must_put_chain(
         chains: &mut ForeignChainsConfig,
         chain: ForeignChain,
@@ -840,13 +847,10 @@ mod tests {
         // Given
         let server = httpmock::MockServer::start_async().await;
         mock_fingerprint(&server, BITCOIN_MAINNET).await;
-        let config = ForeignChainsConfig {
-            bitcoin: Some(chain_config(
-                Some(BITCOIN_MAINNET),
-                one_provider("publicnode", &server.base_url()),
-            )),
-            ..Default::default()
-        };
+        let config = bitcoin_only(chain_config(
+            Some(BITCOIN_MAINNET),
+            one_provider("publicnode", &server.base_url()),
+        ));
 
         // When
         let report = probe_all_providers(&config).await;
@@ -863,13 +867,10 @@ mod tests {
         // Given
         let server = httpmock::MockServer::start_async().await;
         mock_fingerprint(&server, BITCOIN_TESTNET3).await;
-        let config = ForeignChainsConfig {
-            bitcoin: Some(chain_config(
-                Some(BITCOIN_MAINNET),
-                one_provider("publicnode", &server.base_url()),
-            )),
-            ..Default::default()
-        };
+        let config = bitcoin_only(chain_config(
+            Some(BITCOIN_MAINNET),
+            one_provider("publicnode", &server.base_url()),
+        ));
 
         // When
         let report = probe_all_providers(&config).await;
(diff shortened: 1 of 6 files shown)


crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -107,8 +108,14 @@ pub async fn probe_all_providers(config: &ForeignChainsConfig) -> ProbeReport {
...
                 ForeignChain::Polygon => probe_evm::<Polygon>(chain, chain_config).await,
-                // TODO(#4003): probe Bitcoin, Aptos and Sui. Ethereum, Solana and Ton have no
-                // inspector, so there is nothing to probe them with.
+                ForeignChain::Bitcoin => {
+                    probe_chain(chain, chain_config, |provider| {
+                        Ok(BitcoinInspector::new(prepare_jsonrpc(provider)?))
+                    })
+                    .await
+                }
+                // TODO(#4003): probe Aptos and Sui. Ethereum, Solana and Ton have no inspector, so
```
netrome, line 117: Reminds me, we need to pick up the Ton inspector. Daniel's PR is growing stale there 😅

netrome (dismissed):
Nice stuff, thanks!

crates/foreign-chain-rpc-interfaces/src/bitcoin.rs
```diff
@@ -96,3 +96,80 @@ impl Serialize for GetBlockHashArgs {
 impl ToRpcParams for &GetBlockHashArgs {
     to_rpc_params_impl!();
 }
+
+/// RPC response for `getblockhash`: a block hash in the byte order block explorers render.
+/// <https://developer.bitcoin.org/reference/rpc/getblockhash.html>
+///
+/// Kept as text rather than parsed into a [`TransportBitcoinBlockHash`].
```
kevindeforth: It's best to avoid paraphrasing the code (as per our [engineering standards](https://github.com/near/mpc/blob/main/docs/engineering-standards.md#write-helpful-code-comments)).
This comment would be more helpful if it explained _why_ we keep it as a string instead of a `TransportBitcoinBlockhash`.

crates/foreign-chain-inspector/tests/bitcoin_rpc_manual.rs
```diff
@@ -118,3 +118,28 @@ struct GetBlockchainInfoResponse {
 struct GetBlockVerbosityOneResponse {
     tx: Vec<String>,
 }
+
+/// Bitcoin mainnet's genesis block hash, as shipped in `expected_network_fingerprint`.
+const EXPECTED_NETWORK_FINGERPRINT: &str =
+    "000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f";
```
kevindeforth: > as shipped in `expected_network_fingerprint`

What is this referring to? I don't understand that second part of the comment.

kevindeforth (dismissed):
Thank you!

pbeza (dismissed):

crates/foreign-chain-rpc-interfaces/src/bitcoin.rs
pbeza: Curious why Claude didn’t catch that. I recently added it to `AGENTS.md` too:

https://github.com/near/mpc/blob/4659ad7166b79d30e6d816d89c720fc16e4b4db1/AGENTS.md?plain=1#L139

kevindeforth (approved):

pbeza (approved):

PR #4069, opened 2026-08-05
feat: probe Aptos providers by ledger chain id

### LINKED ISSUE #4092: Probe Aptos for its ledger chain id
### Background

Part of #4003. Aptos serves its ledger chain id at the REST API root, the first probed chain that speaks REST instead of JSON RPC.

### Acceptance Criteria

- An Aptos provider answering the expected chain id is reported healthy; a different chain id is reported unhealthy.
- A URL that serves no Aptos API is reported as a refusal rather than a missing transaction.

haiyuechen-nearone:
Closes #4092.

Aptos reports its chain id in the ledger info every node serves at the REST root, so `AptosRpcClient` gains a `get_ledger_info` call and the inspector reads `chain_id` from it.

### Notes for review

- **A 404 means different things depending on which method was called.**
  - `get_transaction_by_hash`: the transaction is absent.
  - `get_ledger_info`: the endpoint does not serve an Aptos API.
  - The meaning is modeled as traits on the response type.
  - The shared types live in the `foreign-chain-instpector` crate root because Sui needs the same distinction for gRPC `NOT_FOUND`.

- **The transport step and the decode step fail with their own types.**  Splitting the parsing into two steps to identify network errors from permanent faults.

### DIFF
--- a/crates/foreign-chain-health-check/src/probe.rs
+++ b/crates/foreign-chain-health-check/src/probe.rs
@@ -260,9 +260,11 @@ mod tests {
     const CLOSED_PORT_URL: &str = "http://127.0.0.1:9";
     /// For a chain with no probe: the value is never read, only whether it is set at all.
     const ANY_FINGERPRINT: &str = "any-fingerprint";
-    /// Aptos publishes its ledger chain id in decimal.
+    /// Aptos providers reports its chain id as a bare JSON number (`uint8`). The configured fingerprint is
+    /// the same number as text.
     const APTOS_MAINNET: u64 = 1;
     const APTOS_TESTNET: u64 = 2;
+    const PROVIDER_NAME: &str = "publicnode";
     /// Bitcoin's genesis block hash, which is what tells its networks apart.
     const BITCOIN_MAINNET: &str =
         "000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f";
@@ -924,8 +926,8 @@ mod tests {
         mock_ledger_info(&server, APTOS_MAINNET).await;
         let config = ForeignChainsConfig {
             aptos: Some(chain_config(
-                Some("1"),
-                one_provider("publicnode", &server.base_url()),
+                Some(&APTOS_MAINNET.to_string()),
+                one_provider(PROVIDER_NAME, &server.base_url()),
             )),
             ..Default::default()
         };
@@ -935,7 +937,7 @@ mod tests {
 
         // Then
         assert_eq!(
-            must_status_of(&report, ForeignChain::Aptos, "publicnode"),
+            must_status_of(&report, ForeignChain::Aptos, PROVIDER_NAME),
             ProviderStatus::Healthy
         );
     }
@@ -947,8 +949,8 @@ mod tests {
         mock_ledger_info(&server, APTOS_TESTNET).await;
         let config = ForeignChainsConfig {
             aptos: Some(chain_config(
-                Some("1"),
-                one_provider("publicnode", &server.base_url()),
+                Some(&APTOS_MAINNET.to_string()),
+                one_provider(PROVIDER_NAME, &server.base_url()),
             )),
             ..Default::default()
         };
@@ -958,10 +960,10 @@ mod tests {
 
         // Then
         assert_eq!(
-            must_status_of(&report, ForeignChain::Aptos, "publicnode"),
+            must_status_of(&report, ForeignChain::Aptos, PROVIDER_NAME),
             ProviderStatus::WrongNetwork {
-                expected: NetworkFingerprint::new("1"),
-                observed: NetworkFingerprint::new("2"),
+                expected: NetworkFingerprint::new(APTOS_MAINNET.to_string()),
+                observed: NetworkFingerprint::new(APTOS_TESTNET.to_string()),
             }
         );
     }
--- a/crates/foreign-chain-health-check/src/probe.rs
+++ b/crates/foreign-chain-health-check/src/probe.rs
@@ -5,6 +5,7 @@ use std::collections::BTreeMap;
 use std::time::Duration;
 
 use foreign_chain_inspector::abstract_chain::inspector::Abstract;
+use foreign_chain_inspector::aptos::inspector::AptosInspector;
 use foreign_chain_inspector::arbitrum::inspector::Arbitrum;
 use foreign_chain_inspector::base::inspector::Base;
 use foreign_chain_inspector::bitcoin::inspector::BitcoinInspector;
@@ -16,11 +17,12 @@ use foreign_chain_inspector::starknet::inspector::StarknetInspector;
 use foreign_chain_inspector::{
     FanOut, ForeignChainInspectionError, NetworkFingerprint, ProviderFailure,
 };
+use foreign_chain_rpc_interfaces::aptos::ReqwestAptosClient;
 use mpc_node_config::{ForeignChainConfig, ForeignChainProviderConfig, ForeignChainsConfig};
 use near_mpc_bounded_collections::NonEmptyVec;
 use near_mpc_contract_interface::types::{ForeignChain, ProviderId};
 
-use crate::prepare_jsonrpc;
+use crate::{prepare_aptos, prepare_jsonrpc};
 
 /// One provider's verdict. Anything other than [`ProviderStatus::Healthy`] is unhealthy.
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -114,8 +116,20 @@ pub async fn probe_all_providers(config: &ForeignChainsConfig) -> ProbeReport {
                     })
                     .await
                 }
-                // TODO(#4003): probe Aptos and Sui. Ethereum, Solana and Ton have no inspector, so
-                // there is nothing to probe them with.
+                ForeignChain::Aptos => {
+                    let timeout = Duration::from_secs(chain_config.timeout_sec.get());
+                    probe_chain(chain, chain_config, move |provider| {
+                        let (url, auth_header) = prepare_aptos(provider)?;
+                        Ok(AptosInspector::new(ReqwestAptosClient::new(
+                            url,
+                            auth_header,
+                            timeout,
+                        )))
+                    })
+                    .await
+                }
+                // TODO(#4003): probe Sui. Ethereum, Solana and Ton have no inspector, so there is
+                // nothing to probe them with.
                 _ => rows_of(chain, chain_config, ProviderStatus::ProbeNotImplemented),
             }
         });
@@ -244,6 +258,9 @@ mod tests {
     const CLOSED_PORT_URL: &str = "http://127.0.0.1:9";
     /// For a chain with no probe: the value is never read, only whether it is set at all.
     const ANY_FINGERPRINT: &str = "any-fingerprint";
+    /// Aptos publishes its ledger chain id in decimal.
+    const APTOS_MAINNET: u64 = 1;
+    const APTOS_TESTNET: u64 = 2;
     /// Bitcoin's genesis block hash, which is what tells its networks apart.
     const BITCOIN_MAINNET: &str =
         "000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f";
@@ -881,6 +898,68 @@ mod tests {
             ProviderStatus::WrongNetwork {
                 expected: NetworkFingerprint::new(BITCOIN_MAINNET),
                 observed: NetworkFingerprint::new(BITCOIN_TESTNET3),
+            }
+        );
+    }
+
+    async fn mock_ledger_info<'a>(
+        server: &'a httpmock::MockServer,
+        chain_id: u64,
+    ) -> httpmock::Mock<'a> {
+        let body = serde_json::json!({"chain_id": chain_id, "ledger_version": "1"});
+        server
+            .mock_async(|when, then| {
+                when.method(httpmock::Method::GET);
+                then.status(200).json_body(body);
+            })
+            .await
+    }
+
+    #[tokio::test]
+    async fn probe_all_providers__should_report_aptos_on_its_expected_chain_id_as_healthy() {
+        // Given
+        let server = httpmock::MockServer::start_async().await;
+        mock_ledger_info(&server, APTOS_MAINNET).await;
+        let config = ForeignChainsConfig {
+            aptos: Some(chain_config(
+                Some("1"),
+                one_provider("publicnode", &server.base_url()),
+            )),
+            ..Default::default()
+        };
+
+        // When
+        let report = probe_all_providers(&config).await;
+
+        // Then
+        assert_eq!(
+            must_status_of(&report, ForeignChain::Aptos, "publicnode"),
+            ProviderStatus::Healthy
+        );
+    }
+
+    #[tokio::test]
+    async fn probe_all_providers__should_report_aptos_on_another_network_as_wrong_network() {
+        // Given
+        let server = httpmock::MockServer::start_async().await;
+        mock_ledger_info(&server, APTOS_TESTNET).await;
+        let config = ForeignChainsConfig {
+            aptos: Some(chain_config(
+                Some("1"),
+                one_provider("publicnode", &server.base_url()),
+            )),
+            ..Default::default()
+        };
+
+        // When
+        let report = probe_all_providers(&config).await;
+
+        // Then
+        assert_eq!(
+            must_status_of(&report, ForeignChain::Aptos, "publicnode"),
+            ProviderStatus::WrongNetwork {
+                expected: NetworkFingerprint::new("1"),
+                observed: NetworkFingerprint::new("2"),
             }
         );
     }
--- a/crates/foreign-chain-health-check/src/probe.rs
+++ b/crates/foreign-chain-health-check/src/probe.rs
@@ -93,8 +93,8 @@ impl ProbeReport {
 /// Each provider is tried up to `max_retries` times, `timeout_sec` per try, and only for as long as
 /// the failures stay transient. This returns within the largest configured `timeout_sec *
 /// max_retries`, plus the [`foreign_chain_inspector::RETRY_BACKOFF`] between tries.
-// TODO(#4043): take the inspectors as a dependency instead, so that choosing a client per chain,
-// and the deadline it is built with, happens outside the probe.
+///
+/// TODO(#4043): take the inspectors as a dependency instead
 pub async fn probe_all_providers(config: &ForeignChainsConfig) -> ProbeReport {
     let probe_attempts = config
         .iter_chains()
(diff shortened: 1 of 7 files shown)


crates/foreign-chain-rpc-interfaces/src/aptos.rs
```diff
@@ -47,6 +47,25 @@ pub enum AptosRpcError {
     Http(#[from] reqwest::Error),
     #[error("Aptos API returned HTTP {status}: {body}")]
     ApiError { status: u16, body: String },
+    #[error("failed to decode the Aptos API response: {0}")]
+    MalformedBody(#[from] serde_json::Error),
+}
```
pbeza, line 52: Is this an exhaustive list of Aptos RPC errors? If not, it might be worth adding `#[non_exhaustive]`, although I’m not 100% sure that’s the best practice here.

crates/foreign-chain-rpc-interfaces/src/aptos.rs
```diff
@@ -85,6 +108,21 @@ impl ReqwestAptosClient {
...
+            return Err(AptosRpcError::ApiError {
+                status: status.as_u16(),
+                body,
+            });
+        }
+
+        let body = response.bytes().await?;
+        Ok(serde_json::from_slice(&body)?)
+    }
 }
```
pbeza, line 127: I think this could be shortened a bit:

```suggestion
async fn get_json<T: DeserializeOwned>(&self, url: Url) -> Result<T, AptosRpcError> {
    let response = self.client.get(url).send().await?;
    let status = response.status();
    if !status.is_success() {
        return Err(AptosRpcError::ApiError {
            status: status.as_u16(),
            body: response.text().await.unwrap_or_default(),
        });
    }
    Ok(response.json().await?)
}
```

crates/foreign-chain-rpc-interfaces/src/aptos.rs
```diff
@@ -309,4 +340,37 @@ mod tests {
             r#"{"outer_a":0,"outer_b":{"inner_a":1,"inner_z":9}}"#
         );
     }
+
+    #[rstest]
+    #[case::mainnet("1", "1")]
+    #[case::padded("0002", "2")]
```
pbeza: Optional nit: if we treat `0002` as `2`, which means testnet, then perhaps it should be:

```suggestion
    #[case::testnet("0002", "2")]
```

or

```suggestion
    #[case::padded_testnet("0002", "2")]
```

crates/foreign-chain-rpc-interfaces/src/aptos.rs
```diff
@@ -309,4 +340,37 @@ mod tests {
...
+            "ledger_version": "1234",
+            "node_role": "full_node",
+        });
+
+        // When
+        let parsed: LedgerInfoResponse = serde_json::from_value(json).unwrap();
+
+        // Then
+        assert_eq!(parsed.chain_id, 1);
+    }
```
pbeza: I think this tests `serde` rather than our code. IIUC, `serde` ignores unknown JSON fields by default (unless you add `#[serde(deny_unknown_fields)]`).

crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -885,6 +904,68 @@ mod tests {
...
+    }
+
+    #[tokio::test]
+    async fn probe_all_providers__should_report_aptos_on_another_network_as_wrong_network() {
+        // Given
+        let server = httpmock::MockServer::start_async().await;
+        mock_ledger_info(&server, APTOS_TESTNET).await;
+        let config = ForeignChainsConfig {
+            aptos: Some(chain_config(
+                Some("1"),
```
pbeza: Nit: we could reuse `"1"` below in the `// Then` section if we extract it into a variable.

crates/foreign-chain-inspector/src/aptos/inspector.rs
```diff
@@ -103,6 +100,59 @@ where
...
+            // Aptos answers 404 both for a transaction it lacks and for a path it does not route.
+            AptosRpcError::ApiError { status: 404, .. } => match T::ABSENCE {
+                AbsenceMeaning::TransactionIsAbsent => {
+                    ForeignChainInspectionError::TransactionNotFound
+                }
+                AbsenceMeaning::ApiIsNotServed => {
+                    ForeignChainInspectionError::RpcRequestRejected(message)
+                }
+            },
+            // Rate limits and server errors are provider hiccups → transient, so the
```
pbeza: Nit: It was a bit confusing to me to see the comment explaining that this error is transient (until I consulted Claude), because the transient/non-transient split isn't decided in this file at all. The mapping in `classified()` only picks a `ForeignChainInspectionError` variant, while each variant's transientness is defined centrally in `is_transient()` in `lib.rs`. Same goes for:

https://github.com/near/mpc/blob/30611349572c5c843cdf7443237be3a30f3746fa/crates/foreign-chain-inspector/src/aptos/inspector.rs#L148

crates/foreign-chain-inspector/src/aptos/inspector.rs
```diff
@@ -732,4 +813,82 @@ mod tests {
...
+        let inspector = AptosInspector::new(MockAptosClient::on_chain(TESTNET_CHAIN_ID));
+
+        // When
+        let fingerprint = inspector
+            .network_fingerprint()
+            .await
+            .expect("network_fingerprint should succeed");
+
+        // Then
+        assert_eq!(fingerprint.to_string(), "2");
```
pbeza: We could reuse `TESTNET_CHAIN_ID` here.

pbeza (dismissed):

crates/foreign-chain-inspector/src/aptos/inspector.rs
```diff
@@ -103,6 +100,59 @@ where
...
+                    ForeignChainInspectionError::TransactionNotFound
+                }
+                AbsenceMeaning::ApiIsNotServed => {
+                    ForeignChainInspectionError::RpcRequestRejected(message)
+                }
+            },
+            // Rate limits and server errors are provider hiccups → transient, so the
+            // affected provider is dropped from the quorum instead of blocking it.
+            AptosRpcError::ApiError {
+                status: 408 | 429, ..
```
pbeza: FWIW, by Claude:

> These two arms re-encode the policy that already exists as `is_retryable_status` in the crate root (`408 | 429`, or `>= 500`), which is reachable from this module. Collapsing them into one `AptosRpcError::ApiError { status, .. } if is_retryable_status(status)` arm keeps the two classifiers from drifting when one of them learns a new status.

crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -885,6 +904,68 @@ mod tests {
...
+            )),
+            ..Default::default()
+        };
+
+        // When
+        let report = probe_all_providers(&config).await;
+
+        // Then
+        assert_eq!(
+            must_status_of(&report, ForeignChain::Aptos, "publicnode"),
```
kevindeforth: nit: would be good to store "publicnode" string in a variable.

crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -244,6 +260,9 @@ mod tests {
     const CLOSED_PORT_URL: &str = "http://127.0.0.1:9";
     /// For a chain with no probe: the value is never read, only whether it is set at all.
     const ANY_FINGERPRINT: &str = "any-fingerprint";
+    /// Aptos publishes its ledger chain id in decimal.
+    const APTOS_MAINNET: u64 = 1;
```
kevindeforth: The comment confuses me - we are storing inside an unsigned integer, but the comment says that aptos uses a decimal.

crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -885,6 +904,68 @@ mod tests {
...
+
+        // When
+        let report = probe_all_providers(&config).await;
+
+        // Then
+        assert_eq!(
+            must_status_of(&report, ForeignChain::Aptos, "publicnode"),
+            ProviderStatus::WrongNetwork {
+                expected: NetworkFingerprint::new("1"),
+                observed: NetworkFingerprint::new("2"),
```
kevindeforth: Where does this "2" come from? Is it the `APTOS_TESTNET` variable?

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -470,6 +470,25 @@ fn is_retryable_status(status_code: u16) -> bool {
     matches!(status_code, REQUEST_TIMEOUT | TOO_MANY_REQUESTS) || status_code >= SERVER_ERROR
 }
 
+/// The meaning of a provider's "not found" answer
```
kevindeforth: please avoid paraphrasing the code in comments (c.f. [engineering standards](https://github.com/near/mpc/blob/main/docs/engineering-standards.md))

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -470,6 +470,25 @@ fn is_retryable_status(status_code: u16) -> bool {
...
+    const ABSENCE: AbsenceMeaning;
+}
+
+/// Reads a chain client's outcome as an inspection outcome. The absence meaning comes from the
+/// response type.
+pub(crate) trait ClassifyRpcOutcome {
+    type Response: HasAbsenceMeaning;
+
+    fn classified(self) -> Result<Self::Response, ForeignChainInspectionError>;
+}
```
kevindeforth: nit: I think this is also paraphrasing the code, but maybe less clear. Not blocking.

crates/foreign-chain-inspector/tests/aptos_rpc_manual.rs
```diff
@@ -60,3 +60,25 @@ fn parse_tx_hash(hash: &str) -> AptosTransactionHash {
         .expect("transaction hash should be 32 bytes");
     AptosTransactionHash::from(array)
 }
+
+/// Aptos mainnet's ledger chain id, as shipped in the node config file
+/// `foreign_chains.aptos.expected_network_fingerprint`.
+const EXPECTED_NETWORK_FINGERPRINT: &str = "1";
```
kevindeforth, line 66: How does this comment guarantee it won't get stale? Why is it relevant that this is set in the config? If so, do we have a DEFAULT_* parameter somewhere we could use from prod, instead of hard-coding this in the test?

crates/foreign-chain-inspector/tests/aptos_rpc_manual.rs
```diff
@@ -60,3 +60,25 @@ fn parse_tx_hash(hash: &str) -> AptosTransactionHash {
...
+
+#[tokio::test]
+#[ignore = "manual test to sanity check against live Aptos RPC provider"]
```
kevindeforth, line 69: This test might be more useful if we could activate it with a flag, together with all the other live-RPC tests. 
As it stands, I don't know how we will remember / know that it exists and periodically run it.

crates/foreign-chain-rpc-interfaces/src/aptos.rs
```diff
@@ -47,6 +47,25 @@ pub enum AptosRpcError {
...
+    #[error("failed to decode the Aptos API response: {0}")]
+    MalformedBody(#[from] serde_json::Error),
+}
+
+/// Partial response of the API root: the ledger info every Aptos node reports.
+/// <https://api.mainnet.aptoslabs.com/v1/spec#/operations/get_ledger_info>
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+pub struct LedgerInfoResponse {
+    /// One byte in the spec, one digit in practice: 1 is mainnet, 2 is testnet.
+    pub chain_id: u64,
```
kevindeforth: I am not fully following this comment. What spec is this comment referring to? A different comment mentions that chain_id is a decimal, how do these two comments relate?

crates/foreign-chain-rpc-interfaces/src/aptos.rs
```diff
@@ -47,6 +47,25 @@ pub enum AptosRpcError {
...
+}
+
+/// Decimal without leading zeros, the form Aptos publishes chain ids in. Text that is not a
+/// number is returned unchanged, so a nonsense answer can be reported as it was given.
+pub fn canonical_chain_id_text(chain_id: &str) -> String {
+    match chain_id.parse::<u64>() {
+        Ok(parsed) => parsed.to_string(),
+        Err(_) => chain_id.to_owned(),
+    }
 }
```
kevindeforth: Have you considered returning `Result<u64, String>` here?

kevindeforth (commented):
Thank you. Left a few comments.
I think some of the code-comments and tests can be improved for readability and clarity.

crates/foreign-chain-inspector/src/lib.rs
haiyuechen-nearone: Updated, PTAL

crates/foreign-chain-rpc-interfaces/src/aptos.rs
haiyuechen-nearone: Yes and I think we should keep returning a plain string. Although Aptos chain_id is a number, we are not expected to perform arithmetics with it. The only operation we want to do is to compare the configured value and the provider returned value as normalized strings.

I am using the parsing logic to strip any leading zeros or decimal numbers, not to actually parse an Int for doing math later on.

crates/foreign-chain-health-check/src/probe.rs
haiyuechen-nearone: Updated

crates/foreign-chain-rpc-interfaces/src/aptos.rs
haiyuechen-nearone: Updated, PTAL

crates/foreign-chain-inspector/src/lib.rs
haiyuechen-nearone: Reworded the comment on the enum this trait relies on, which makes the comment here unnecessary

crates/foreign-chain-rpc-interfaces/src/aptos.rs
haiyuechen-nearone: Inlined the error body now 👍

On `response.json()`, I kept it deliberately split in two steps, first `response.bytes()` and then `serde_json::fron_slice()`. This is because `response.bytes()` failing means there is some network issue that caused the response to not arrive fully. `serde_json::fron_slice()` failing means the response arrived fully, but we could not parse it. Those are two different errors, first case is transient, second is permanent failure.

`response.json()` would throw `reqwest::Error` for both cases. To be able to distinguish the transport error from the parsing error while still using `response.json()`, we would have something like this which I think is worse:
```
response.json::<T>().await.map_err(|error| {
    if error.is_decode() {
        AptosRpcError::MalformedBody(error)   // variant type must change too
    } else {
        AptosRpcError::Http(error)
    }
})?
```

crates/foreign-chain-rpc-interfaces/src/aptos.rs
haiyuechen-nearone: You are right, deleted it now.

crates/foreign-chain-inspector/tests/aptos_rpc_manual.rs
haiyuechen-nearone: No hard guarantee, but Aptos mainnet's chain id is 1 and is not supposed to ever change, so I don't expect this to get stale. "As shipped in the node config file" refers to the default config file we ship with values already filled in. The operators can freely change this value. They can for instance change this to `2` to target Aptos testnet, for a testnet deployment.

There is deliberately no DEFAULT_* in code. We wanted a clean separation between config and code, the node just runs with whatever the config file says. By that logic the test should read the config file as well, but that felt like overkill for a sanity check. Since the chain id is fixed, I think hard coding it here is fine.

crates/foreign-chain-inspector/tests/aptos_rpc_manual.rs
haiyuechen-nearone: The flag already exists as the crate convention: every *_rpc_manual.rs test is #[ignore = "manual …"], and they all run together with `cargo nextest run -p foreign-chain-inspector --run-ignored only`. 

Running them periodically is a fair point, but idealy I would like to retire these tests completely once the probing logic is landed and picked up in the Foreign chain config tester CLI. The periodic run should use that CLI tool to exercise the same code as used in the Tx validation path, instead of these half baked tests.

My understanding is that these are more of a "trust me bro, I did test the request code against the real provider for my change" kind of test atm.

crates/foreign-chain-rpc-interfaces/src/aptos.rs
haiyuechen-nearone: It is exhaustive over how the Aptos client itself can fail (transport, non 2xx answer, malformed body). Aptos' API error catalog stays inside ApiError { status, body }. We don't need a new variant when Aptos adds new error codes.

I would not add #[non_exhaustive] here. Adding it forces a wildcard arm on the match in `ClassifyRpcOutcome::classified`, which is not what we want. A new variant here should be a compile error until it gets an explicit transient or not decision in `ClassifyRpcOutcome::classified`.

pbeza:
Sorry, I [committed](https://nearone.slack.com/archives/C091512F0DP/p1787123172446659) to re-reviewing it today, but I won’t make it. There were more review comments and follow-up commits than I expected. I’ll try to get to it tomorrow. Sorry about that. 🙏🏼

pbeza (approved):
I skimmed through it and it LGTM. I didn’t do a deep re-review because, as [discussed](https://nearone.slack.com/archives/C0912BTG51T/p1787219813267329?thread_ts=1786132114.069929&cid=C0912BTG51T) offline, it looks like you [force-pushed](https://github.com/near/mpc/compare/30611349572c5c843cdf7443237be3a30f3746fa..46ae5c4d838ade61314c97d220679e0d3bf47357) some changes after the reviews, which makes it much harder to tell what changed since then. Not sure if that was caused by the automatic restack or something else.

Let’s try to avoid force-pushing once reviews are in when possible, since the diff against the commit I reviewed is pretty noisy, presumably because of the force-push: https://github.com/near/mpc/compare/30611349572c5c843cdf7443237be3a30f3746fa..a4e826bc0fd99eb930007a499f56e0050c14177f.

pbeza (approved):

crates/foreign-chain-inspector/tests/aptos_rpc_manual.rs
kevindeforth: > We wanted a clean separation between config and code, the node just runs with whatever the config file says

Thanks for the explanation. I feel like the comment "as shipped in the node config file is at odds with that goal, but since this is in a test-file, this is not a blocker.

crates/foreign-chain-inspector/tests/aptos_rpc_manual.rs
kevindeforth: > idealy I would like to retire these tests completely once the probing logic is landed and picked up in the Foreign chain config tester CLI.

that does sound nice. Good plan :+1:

crates/foreign-chain-rpc-interfaces/src/aptos.rs
kevindeforth: Ok, makes sense :+1:

kevindeforth (approved):
Thank you, this looks good!

PR #4070, opened 2026-08-05
feat: probe Sui providers by genesis checkpoint digest

### LINKED ISSUE #4093: Probe Sui for its genesis checkpoint digest
### Background

Part of #4003. Sui identifies its network by the genesis checkpoint digest, which `GetServiceInfo` already carries on the golden route.

### Acceptance Criteria

- A Sui provider answering the expected digest is reported healthy; a different digest is reported unhealthy.
- With Sui covered, every chain that has an inspector is probed; `ProbeNotImplemented` remains only for chains without one.

haiyuechen-nearone:
Closes #4093.

### Notes for review

- **Nothing is normalized.** The Sui genesis checkpoint digest is Base58, which is case sensitive and carries no prefix or padding.

- **`NotFound` reads off the response type**, through the `HasAbsenceMeaning` traits #4069 added.

- **`DeadlineExceeded` is mapped to `Timeout` instead of `RpcRequestFailed` by this PR.** Both variants are transient, so retries and the signing path fan out are unchanged.

- **Separate mock servers in tests.** The mock HTTP server used by previous tests cannot serve gRPC requests.

- **Review feedback widened the diff beyond Sui.** Every chain's network fingerprint constant is now defined once in its inspector module and referenced from golden values, probe tests, and manual tests.


haiyuechen-nearone:
@claude review

crates/foreign-chain-health-check/Cargo.toml
```diff
@@ -26,6 +26,7 @@ tokio = { workspace = true }
 assert_matches = { workspace = true }
 httpmock = { workspace = true }
 serde_json = { workspace = true }
+tonic = { workspace = true, features = ["router", "server"] }
```
gilcu3: why do we need tonic if we already have httpmock?

gilcu3:
@claude review

crates/foreign-chain-health-check/Cargo.toml
gilcu3: discussed on [slack](https://nearone.slack.com/archives/C0912BTG51T/p1787555908892769?thread_ts=1786446909.471079&cid=C0912BTG51T)

crates/foreign-chain-inspector/tests/sui_inspector.rs
```diff
@@ -454,3 +469,55 @@ async fn extract__should_return_empty_when_no_extractors_are_requested() {
     let expected: Vec<SuiExtractedValue> = vec![];
     assert_eq!(expected, extracted_values);
 }
+
+/// Sui mainnet's genesis checkpoint digest.
+const MAINNET_CHAIN_ID: &str = "4btiuiMPvEENsttpZC7CZ53DruC3MAgfznDbASZ7DR6S";
```
gilcu3: nit: could be good to unify this constant, I think we have it in several places already.

crates/foreign-chain-health-check/Cargo.toml
gilcu3: Apparently we are missing the `codegen` feature here, and only getting it by chance. Please check

crates/foreign-chain-inspector/src/sui/inspector.rs
```diff
@@ -28,6 +33,26 @@ pub enum SuiExtractor {
...
+                "service info is missing the chain id".to_string(),
+            ));
+        };
+        Ok(Self::canonical_fingerprint(&chain_id))
+    }
+
+    /// Base58 is case sensitive and does not permit prefix or padding.
+    fn canonical_fingerprint(fingerprint: &str) -> NetworkFingerprint {
+        NetworkFingerprint::new(fingerprint)
+    }
```
gilcu3, line 54: The comment seems a bit out of place. I guess it means the expected input parameter is Base58, so no need to do any processing

crates/foreign-chain-inspector/src/sui/inspector.rs
```diff
@@ -100,21 +121,42 @@ where
...
+        let message = status.to_string();
+        Err(match status.code() {
+            Code::NotFound => match T::ABSENCE {
+                AbsenceMeaning::TransactionIsAbsent => {
+                    ForeignChainInspectionError::TransactionNotFound
+                }
+                AbsenceMeaning::ApiIsNotServed => {
+                    ForeignChainInspectionError::RpcRequestRejected(message)
+                }
+            },
```
gilcu3, line 151: This pattern is a bit weird, but was not introduced here. I tried to find something slightly better, and ended up with something like taking a `AbsenceMeaning` value in this function so that this match, nor the trait, is needed. Also I would call it something different, as `NotFound` and `Absence` are similar but not the same, so the relation between the two gets a bit lost. Something more explicit would be `NotFoundMeaning` for example, but of course if the other chains use absence we are out of luck

gilcu3 (dismissed):
Thank you!

crates/foreign-chain-inspector/src/sui/inspector.rs
haiyuechen-nearone: I left this comment here because for all the other chains we normalize the input strings, but for Sui the format is Base58 so we don't need to do any normalization.

gilcu3: Yeah I think leaving the comment is fine. I just had a hard time knowing what it was about, maybe expanding it slightly makes sense, else it is a bit cryptic unless you look at the other chain implementations

crates/foreign-chain-inspector/src/sui/inspector.rs
haiyuechen-nearone: I did start out with a `classify(code, absenceMeaning)` implementation as well, but I felt the trait approach is more generalizable across all possible chains, and less prone to developer errors.

For transaction lookup, NotFound means the TX doesn't exist.
For Aptos `get_ledger_info` and Sui `GetServiceInfo`, NotFound means the endpoint we are trying to hit is not there, hence 'absent'.
And we might in the future request another path, and error X means something unconventional.

Since the meaning is closely tied to the path/response type, I think its the response object's responsibility to transform a response code to our internal `ForeignChainInspectionError`. This centralizes how an external error code is transformed, and lets us decouple the internal call site from the external concepts.

crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -123,7 +123,7 @@ pub async fn probe_all_providers(config: &ForeignChainsConfig) -> ProbeReport {
                     .await
                 }
                 ForeignChain::Aptos => {
-                    let timeout = Duration::from_secs(chain_config.timeout_sec.get());
+                    let timeout = timeout_of(chain_config);
```
pbeza, line 126: Nit: Wouldn't it be more natural to have `impl ForeignChainConfig` instead of `timeout_of`?

```
impl ForeignChainConfig {
    pub fn timeout(&self) -> Duration {
        Duration::from_secs(self.timeout_sec.get())
    }
}
```

crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -264,6 +275,9 @@ mod tests {
     const CLOSED_PORT_URL: &str = "http://127.0.0.1:9";
     /// For a chain with no probe: the value is never read, only whether it is set at all.
     const ANY_FINGERPRINT: &str = "any-fingerprint";
+    /// Sui's genesis checkpoint digest, base58.
+    const SUI_MAINNET: &str = "4btiuiMPvEENsttpZC7CZ53DruC3MAgfznDbASZ7DR6S";
```
pbeza: It feels a bit off to have this const already defined in four places: https://github.com/search?q=repo%3Anear%2Fmpc+4btiuiMPvEENsttpZC7CZ53DruC3MAgfznDbASZ7DR6S&type=code.

crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -264,6 +275,9 @@ mod tests {
...
+    const SUI_TESTNET: &str = "69WiPg3DAQiwdxfncX6wYQ2siKwAe6L9BZthQea3JNMD";
```
pbeza: [Ditto](https://github.com/near/mpc/pull/4070/changes#r3844634728): https://github.com/search?q=repo%3Anear%2Fmpc+69WiPg3DAQiwdxfncX6wYQ2siKwAe6L9BZthQea3JNMD&type=code.

crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -982,6 +996,171 @@ mod tests {
...
+
+        // When
+        let report = probe_all_providers(&config).await;
+
+        // Then
+        assert_eq!(
+            must_status_of(&report, ForeignChain::Sui, PROVIDER_NAME),
+            ProviderStatus::Unreachable
+        );
+    }
```
pbeza: Some of these tests look very similar. Perhaps there’s an idiomatic way to [shorten](https://nearone.slack.com/archives/C07UW93JVQ8/p1787324683858979) this with `rstest`? (Maybe not, which is fine too.)

pbeza (dismissed):

crates/foreign-chain-health-check/src/probe.rs
haiyuechen-nearone: Yes, but the bigger problem here is that the inspectors and clients are constructed in this scope. I am working on a separate PR to have a central place to initialize inspectors and inject them into this scope. I will fold the `ForeignChainConfig` change you are suggesting into that stack.

gilcu3 (approved):
Thanks, that was a big dedup!

pbeza (approved):
I love the dedup!

PR #4071, opened 2026-08-05
feat: periodically report foreign chain RPC provider health

### LINKED ISSUE #4094: Report foreign chain RPC provider health periodically
### Background

Part of #4003, phase 1 of #3764. The probe in `foreign-chain-health-check` needs a caller: the node should run it without gating or delaying anything. Operators watch provider health on a dashboard, so the verdict has to leave the process as more than log text. A verdict taken once at boot goes stale, since a provider can start serving another network, or go down, while the node is up.

### Acceptance Criteria

- At startup, and repeatedly while the node runs, the node logs one status line per configured provider, named by chain and provider, and a summary of how many are healthy.
- The configured and healthy provider counts are exposed per chain as metrics, carrying the verdicts of the round that ran last.
- A provider that no probe covers is left out of the healthy summary and of the metrics.
- A slow probe, a probe failure or a panic never delays or blocks startup.
- A run that verified nothing is reported as a warning rather than a clean run.
- A node with no foreign chain configured is warned once and does not keep probing.

haiyuechen-nearone:
Closes #4094.

The node runs the foreign chain probe on a detached hourly loop, starting at
startup. Each round logs a line per provider plus an `x/y healthy` summary and
exports latest status to metrics.

### Notes for review

- **Safe to log:** no `ProviderStatus` variant carries RPC error text, and the
  observed fingerprint is length capped.

- **Metric is per chain, not per provider:** provider names are operator chosen, so
  the label would be unbounded. The logs contains the full name of failing providers.

- **A chain configured without `expected_network_fingerprint` counts as unhealthy**.

- **An empty `foreign_chains` section warns once and the loop exits**

- **`Tick` moved out of `remote_attestation`** into `tick.rs` and is now shared by both attestation_submission and probing loops.


haiyuechen-nearone:
@claude review

haiyuechen-nearone:
> the log output is untested ...

Deliberately not adding tests grepping for log lines. Having test grepping for text is very prone to drift. Log output was verified by running the cluster locally and inspecting it.

All green case:
image

Case with failures:
image

haiyuechen-nearone:
> MissingExpectedFingerprint lands in the unhealthy arm and logs "foreign-chain RPC provider is unhealthy". The provider is fine; the operator omitted expected_network_fingerprint.

Since we expect the operator to correctly configure all fields in the config file, not adding expected network fingerprint is considered "incorrect" and causes us to not know if a provider is healthy or not/

crates/e2e-tests/tests/foreign_chain_probe.rs
```diff
@@ -0,0 +1,95 @@
...
+        CLUSTER_WAIT_TIMEOUT,
+    )
+    .await
+    .expect("the provider serving the expected network should be healthy");
+
+    common::wait_metric_on_nodes(
+        &cluster,
+        &[1],
+        metrics::FOREIGN_CHAIN_RPC_PROVIDERS_HEALTHY,
+        |value| value == 0,
```
anodar: This does not really test much, consider getting rid of it: value is initially 0, it could be increased after this check and we'd never know because condition holds immediately on initial state.

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
```diff
@@ -1820,6 +1843,19 @@ mod tests {
     use super::*;
     use rstest::rstest;
 
+    /// The two whose config key is not the variant name lowercased, which is where a label would
```
anodar: I get hyper_evm, but how is this not the case for abstract ? It's variant name lowercased no ? 

```
ForeignChain::Abstract, "abstract"
```

crates/e2e-tests/src/foreign_chain_mock.rs
```diff
@@ -75,6 +75,7 @@ impl MockServerExt {
     }
 }
 
+pub const MOCK_EVM_CHAIN_ID: u64 = 8453;
```
netrome: Is this an arbitrary magic number or is there any meaning to it? Perhaps worth a brief comment?

crates/node/src/foreign_chain_probe.rs
anodar, line 1: Probing is split a bit awkwardly between `foreign_chain_health_check` and here. Why not move this `run_periodic_probe` there?

crates/node/src/foreign_chain_probe.rs
```diff
@@ -0,0 +1,342 @@
...
+    for (chain, counts) in report.counts_per_chain() {
+        if !probed_chains.contains(&chain) {
+            continue;
+        }
+        metrics::FOREIGN_CHAIN_RPC_PROVIDERS_CONFIGURED
+            .with_label_values(&[chain.label()])
+            .set(counts.configured as i64);
+        metrics::FOREIGN_CHAIN_RPC_PROVIDERS_HEALTHY
+            .with_label_values(&[chain.label()])
+            .set(counts.healthy as i64);
```
netrome: Nit: I prefer explicit conversions over implicitly overflowing conversions. Since we should never be close to overflowing i64 here I think an expect could be okay.

```suggestion
            .set(i64::try_from(counts.configured).expect("we should never be close to overflowing this metric"));
        metrics::FOREIGN_CHAIN_RPC_PROVIDERS_HEALTHY
            .with_label_values(&[chain.label()])
            .set(i64::try_from(counts.healthy as i64).expect("we should never be close to overflowing this metric"));
```

crates/node/src/foreign_chain_probe.rs
```diff
@@ -0,0 +1,342 @@
...
+}
+
+fn summarize(rows: &[ProviderHealth]) -> Summary {
+    Summary {
+        probed: rows.iter().filter(|row| is_probed(row)).count(),
+        healthy: rows.iter().filter(|row| row.status.is_healthy()).count(),
+    }
+}
+
+fn log_report(report: &ProbeReport) {
```
anodar, line 59: I understand why not report ProviderHealth in metrics. But I'd say we should log `ProviderHealth` instead of `ProbeReport` so that if there's an issue we can detect by metrics, and ask node operators to look into the logs that will give more insight what's wrong.

crates/node/src/tick.rs
```diff
@@ -0,0 +1,56 @@
...
+
+#[cfg(test)]
+use std::sync::Arc;
+#[cfg(test)]
+use tokio::sync::Semaphore;
+
+/// Allows repeatedly awaiting for something, like a [`tokio::time::Interval`].
+pub trait Tick {
+    async fn tick(&mut self);
+}
```
netrome, line 11: Seeing this trait made me happy 😊

netrome (dismissed):
Nice stuff! Just one nit and a question from my side.

crates/e2e-tests/src/foreign_chain_mock.rs
haiyuechen-nearone: It is a real chain id for Base, but it doesn't really matter. It's just a mock

anodar (changes requested):
Overall looks good, but one blocker: 

Node will report healthy foreign chain config if all configured RPC providers are healthy. I think we should rather check what's the quorum for reach chain (in the contract, in whitelist) and consider it healthy if at least quorum number of configured RPC's are healthy. Which means also not just log warning when node has no config at all (`if foreign_chains.is_empty() { warn! ....`)

crates/node/src/foreign_chain_probe.rs
haiyuechen-nearone: The loop below logs 1 line per ProviderHealth, do you mean something else?

https://github.com/near/mpc/pull/4071/changes#diff-df67ee122204a68216240afee24f12ba93130b52d9e942da77a8a207f793e6b8R77

crates/node/src/foreign_chain_probe.rs
haiyuechen-nearone: I think `run_periodic_probe` and `probe_periodically` fits in node. My thought is `foreign_chain_health_check` holds only the logic to send probing requests  (`probe_all_providers`), which will be shared between the node and the CLI tool. Node needing to run it periodically, so the node should import and wrap it with extra periodic logic.  

`probe_periodically` needs to publish metrics after every run, which is also node specific.

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
haiyuechen-nearone: This test is not really necessary, deleted it now.

crates/e2e-tests/tests/foreign_chain_probe.rs
haiyuechen-nearone: Updated the test to configure 2 providers with 1 faulty, and asserting 1.

haiyuechen-nearone:
@anodar As discussed in slack, I would like to defer the whitelist and quorum calculation to later and created this issue https://github.com/near/mpc/issues/4273

crates/node/src/foreign_chain_probe.rs
anodar: I completely misread the code thinking this would just log "configured x / healthy y" but this does log what we need.

Sorry for the noise.

anodar (dismissed):
Thank you!

netrome (approved):

anodar (approved):

PR #4073, opened 2026-08-05
fix(contract): size max_scan and the sweep gas budget together

### LINKED ISSUE #4035: clean_invalid_attestations: gas budget cannot cover max_scan
### Problem

`clean_invalid_attestations` is the only mechanism that reclaims stored attestation entries. Its only automatic caller is a detached promise in `vote_reshared`, which asks for `max_scan: 100` but attaches only 10 TGas.

The two constants were never sized against each other. **Scanning an entry costs gas whether or not it is removed**, so a 100-entry scan needs ~25 TGas before a single removal. When the promise runs out, the receipt fails and **every removal rolls back** — silently, because `vote_reshared` still succeeds.

This is already happening: a manual sweep on `v1.signer-prod.testnet` burned 13.1 TGas. Mainnet is still inside budget at 15 stored entries.

```
cost = 3.5 TGas + (entries scanned × 0.215) + (entries removed × 0.5 more)
```

| | cost |
|---|---|
| `max_scan: 100`, 0 removed | ~25 TGas |
| `max_scan: 30`, 0 removed | ~9.9 TGas |
| `max_scan: 30`, 10 removed | ~14.9 TGas |
| `max_scan: 30`, 20 removed | ~20 TGas |

Severity is bounded: the method is permissionless, so anyone can call it manually with up to 300 TGas (~750 removals per call). This is a reclamation gap, not a liveness risk.

### Proposed solution

1. **`max_scan` 100 → 30, budget 10 → 15 TGas.** A 30-entry scan costs ~9.9 TGas, leaving room for ~10 removals.
2. **Reset the budget in the migration** rather than carrying the deployed value forward, so it applies on upgrade instead of needing a governance config vote.

Deliberately not solved: more than ~10 of the 30 being removable (~20 TGas at 20 removals) stays unfunded. The permissionless entry point is the recovery path.

### Acceptance Criteria

Addressed in #4073:

- [x] `RESHARE_CLEAN_INVALID_ATTESTATIONS_MAX_SCAN` and `clean_invalid_attestations_tera_gas` sized against each other.
- [x] New budget applies on upgrade, without a governance vote.

Guarding the two against future drift is split out into #4119, to keep this change reviewable.

Related: #3785, #1087.

barakeinav1:
Closes #4035

`max_scan: 100` costs ~25 TGas to scan before a single removal, against a 10 TGas budget. The promise is detached, so exceeding it rolls back every removal silently — a live sweep on testnet burned 13.1 TGas.

- `RESHARE_CLEAN_INVALID_ATTESTATIONS_MAX_SCAN` 100 → 30, `clean_invalid_attestations_tera_gas` 10 → 15. A 30-entry scan costs ~9.9 TGas, leaving room for ~10 removals.
- Migration resets the budget instead of carrying the deployed value forward, so it applies on upgrade without a governance vote. Unit tests cover both the reset and that every other field still carries over.
- `GAS_FOR_VOTE_RESHARED` 50 → 60 (test const) — the promise reservation grows 39 → 44 TGas, and below that promise creation fails and takes the deciding reshare vote with it.
- Fix a doc comment claiming `reverify_and_cleanup_participants` removes attestations; it never touches `stored_attestations`. Same fix in the verifier design doc.

The gas harness these numbers came from is split out into a follow-up so this stays reviewable: #4118

More than ~10 of the 30 being removable (~20 TGas at 20 removals) stays unfunded; the permissionless entry point is the recovery path.

Migration unit tests covering the reset and the carry-forward were written and passing, then removed as review noise.


barakeinav1:
Thanks — went through each. Eight applied, one rejected on measurement, one declined. Pushed in `28269b62`.

---

### Blocking

> `docs/design/operator-prepaid-attestation-storage.md:48` — Stale after this merge.

**Applied.** Verified — it did still say the constants "need re-validating … tracked in #4035". It now states the resolved sizing and what stays unfunded.

> `crates/contract/src/lib.rs:123-124` — The new comment claims `max_scan` "must stay within what `clean_invalid_attestations_tera_gas` funds", but by the PR description's own numbers it does not.

**Applied, and it caught a worse error than reported.** The comment did assert an invariant the values don't satisfy, so I've replaced it with what is actually guaranteed, including the recovery path:

```rust
/// [`crate::config::Config::clean_invalid_attestations_tera_gas`] funds a full scan of this
/// many entries plus ~10 removals. A sweep with substantially more removable than that
/// exceeds the budget and, because the promise is detached, reverts silently; recovery is a
/// direct permissionless call with a smaller `max_scan`.
```

The `~18 TGas` this quotes was my own figure and it was wrong. Recomputed against the production-measured constants (3.49 TGas fixed, 0.215 per entry scanned, 0.498 more per removal):

| 30 scanned | cost |
|---|---|
| 10 removed | 14.9 TGas |
| 20 removed — the realistic worst case, since participants never get removed | **19.9 TGas** |
| 30 removed | 24.9 TGas |

Corrected in the PR body, #4035 and the design doc.

On the preferred fix of bumping to 20 TGas: the arithmetic is right — 20 TGas funds 20.2 removals, i.e. exactly the realistic worst case, and it is free given unspent promise gas is refunded. Keeping 15 for now because that is what the team settled on in the design thread — the corrected numbers are worth putting back to them, since 15 was chosen partly on the understated figure above.

---

### Non-blocking

> Hoisting those three out of the scan loop would cut the dominant term and likely let `max_scan` stay near 100 rather than dropping 3.3×.

**Rejected — measured.** Reasonable hypothesis from reading the code, but it does not survive experiment. Hoisting all three allowlist rebuilds *and* dropping the redundant per-entry `stored_attestations.get()` moves the cost **0.362 → 0.355 TGas, about 2%**.

The cost is trie I/O, not the allocations: `IterableMap` charges two storage reads per entry scanned (the `keys` vector slot and the `values` map), ~0.11 TGas in base costs alone, and a removal adds `storage_remove` plus the `swap_remove` bookkeeping. That is the floor for this layout.

> Nothing enforces the pairing at runtime … a floor in `Config::validate()`.

**Agreed, tracked.** Confirmed `validate()` runs on every config proposal (`dto_mapping.rs:516,591`), so a floor is enforceable. Added as an optional AC on #4119 rather than here, since rejecting governance votes is a behavioural change that deserves its own review.

> the sweep has no cursor … If the first `max_scan` entries are all valid, the automatic sweep makes no progress at all, forever.

**Correct, and out of scope.** Confirmed: `IterableMap` iterates in first-insertion order, and re-submitting an attestation under the same TLS key replaces the value in place without moving it, so entries never age toward the back. This was raised in the design thread and the team's position is that the permissionless entry point covers it.

> The reset is correct for the 3.14.0 → next upgrade, but it is a one-shot living in a per-release migration module.

**Applied — best catch of the non-blocking set.** Added `TODO(#4121)` at the reset, and #4121 to drop it after the next release.

> Consider a `log!` when the reset fires.

**Declined.** `config()` is a view method, so an operator can read the value directly after upgrade; a log in a migration receipt is harder to find than the state itself.

> The replacement wording is itself imprecise … the trailing parenthetical also now dangles.

**Applied**, using your wording. For context, this exact parenthetical was flagged and then deliberately restored to stay close to upstream's phrasing; two independent reads landing on it settled the question.

> `crates/contract/tests/sandbox/utils/consts.rs:21-23` — The stated arithmetic does not justify the new value.

**Applied.** Right: 44 + ~5 = 49, which 50 already satisfied, so the comment justified keeping 50 rather than bumping it. The actual reason is that 50 leaves 0.8 TGas of margin, which would be flaky. Reworded to state the intended headroom instead of a figure that rots when any of the six cleanup budgets change.

> CLAUDE.md asks for rustdoc intra-doc links over plain backticks.

**Applied**, verified it resolves. Worth noting for future reviews that this only works when visibility allows it — linking a private item from a `pub` item's docs is a hard rustdoc error, which is why two other references in this PR stay plain code spans.

> `config_migration__should_carry_every_other_deployed_value_forward` asserts field-by-field, which will not catch a *new* field being added.

**Applied.** Now compares a full `Config` literal, so adding a field fails to compile here until the test says whether it carries forward or defaults. It compiled with all 19 fields listed, which confirms current coverage.

> **Correction (later round):** this overstated it. `..Default::default()` was still in the `From` impl, so a *removed* carry-forward silently defaulted and the test passed anyway — and two fixture values collided with their own defaults. Both fixed in `0780e891`; see [the follow-up reply](#issuecomment-5263113074).



---

*Edited: see [the follow-up reply](#issuecomment-5263113074) for the second review round, which corrected one claim above.*

barakeinav1:
@claude review

barakeinav1:
All seven verified and applied in `0780e891`, with one subsequently reverted in `e57b054c` — details below.

---

### Blocking

> `crates/contract/src/v3_14_0_state.rs:183` and `:186` — the new fixture violates the invariant its own doc comment claims, leaving open the test hole it was added to close.

**Applied — confirmed, and the failure scenario is exactly as described.** Checked every fixture value against its own default; precisely the two named collide:

```
return_ck_and_clean_state_on_success_call_tera_gas   fixture=7   default=7
clean_tee_status_tera_gas                            fixture=10  default=10
```

Fixture renumbered to `1001..1018`, clear of every default.

Also took the paired suggestion of dropping `..Default::default()` from the `From` impl and listing `attestation_storage_fee_millinear` explicitly, which turns out to matter more than the fixture fix. Running the exact mutation described — deleting `clean_tee_status_tera_gas: old.…` from the impl — now gives:

```
error[E0063]: missing field `clean_tee_status_tera_gas` in initializer of `config::Config`
```

So the guarantee moved from "a test notices" to "it does not build".

This also corrects something I claimed in my [previous reply](#issuecomment-5255376860): I said the whole-struct comparison meant a new field "fails to compile here until the test says whether it carries forward or defaults". That held for the test literal, but `..Default::default()` in the impl meant a *removed* carry-forward still passed silently. The coverage was weaker than I stated; it now holds at the migration site.

> `crates/contract/src/v3_14_0_state.rs:63` (duplicated at `:198`) — the stated justification for overriding an operator-votable field does not match this PR's own measurements.

**Applied — confirmed.** `3.49 + 30 × 0.215 = 9.94` TGas, inside the deployed 10, so "cannot fund the scan" was false once `max_scan` became 30. Reworded to the stronger and accurate argument: the scan consumes nearly all of it, so the *first* eviction overruns the budget and the detached sweep rolls every removal back.

---

### Non-blocking

> the reset is unconditional, so it also *lowers* a budget an operator raised deliberately … `old.clean_invalid_attestations_tera_gas.max(...)`.

**Applied, then reverted after discussion — worth recording why.** The principle is right: `.max()` narrows the migration's claim from "we decide this value" to "we guarantee a floor", which is easier to defend for a governance-owned field.

Against it: both deployments currently sit at **10**, the untouched default, so nothing has been voted and the protection was hypothetical. And `.max()` makes the post-upgrade value depend on prior state, where a plain reset lands every deployment on the one reviewed number — worth more for a field whose problem *was* mis-sizing. We took predictability. `TODO(#4121)` drops the whole thing next release either way.

> *"A sweep with substantially more removable than that exceeds the budget"*: the boundary is exact, not soft.

**Applied.** Confirmed: 10 removals = 14.92 TGas, 11 = 15.42 > 15. Now reads "the eleventh removal exceeds the budget".

> `crates/contract/tests/sandbox/tee_cleanup_after_resharing.rs:66` — pre-existing `// Note: this test fails if vote_reshared needs to clean up more than 3 attestations`.

**Applied.** Dropped the figure, kept the warning tied to what actually bounds it rather than a number that goes stale on this merge.

> `docs/design/operator-prepaid-attestation-storage.md:48` — the second clause … reads as though only the 30-of-30 case is unfunded, when 11 of 30 already exceeds.

**Applied**, clause cut.

> two spellings of "take the default" inside one struct literal are easy to misread.

**Resolved** by removing `..Default::default()` per the first finding.

---

On the `TODO(#4121)`: considered adding one over the migration tests too, and decided against. Simulating a next-release maintainer dropping the reset fails *both* tests loudly and points at itself, so a marker there would duplicate what a red test already says at the moment it matters. The impl line is the only place that compiles and passes when copied forward, which is why it carries the marker.

barakeinav1:
@claude review

barakeinav1:
All three applied in `09737d31`.

> `crates/contract/src/lib.rs:128` — *"recovery is a direct permissionless call with a smaller `max_scan`"* points at the wrong lever.

**Applied — the best of the three, since the old wording would have misled someone mid-incident.** Confirmed the arithmetic: a fully removable 30-entry scan is `3.49 + 30 × 0.215 + 30 × 0.498 = 24.9` TGas against the 300 an external caller can attach, so recovery is one ordinary call and shrinking the window would be busywork. Took your wording.

> `crates/contract/tests/sandbox/utils/consts.rs:24` — the two halves of this doc comment now state opposite intents.

**Applied, with a different fix.** The conflict is real, but the reconciling clause would have preserved a claim that was wrong: the margin has nothing to do with insulating the constant from future budget changes. The actual reason 60 beats 50 is that 50 left 0.8 TGas and would flake. Saying that removes the contradiction outright rather than explaining it away:

> `…plus its own ~5 TGas burn. Below that, promise creation fails and takes the deciding reshare vote with it, so the margin is wide enough that ordinary variance does not make the reshare suite flaky.`

> `crates/contract/src/v3_14_0_state.rs:198` and the other five occurrences — the new tests use lowercase `// given` / `// when` / `// then`.

**Applied.** Verified the counts independently — 161 vs 21 under `crates/contract/src/`, 843 vs 214 repo-wide — and both `CLAUDE.md:107` and `docs/engineering-standards.md:207` capitalize. Also fixed the three occurrences on the stacked harness branch (#4118), which had the same issue.

crates/contract/src/lib.rs
```diff
@@ -119,9 +119,14 @@ const MINIMUM_CKD_REQUEST_DEPOSIT: NearToken = NearToken::from_yoctonear(1);
...
-/// callers may pick a different value; this only governs the automatic invocation.
-const RESHARE_CLEAN_INVALID_ATTESTATIONS_MAX_SCAN: u32 = 100;
+/// Entries to scan in the post-reshare `clean_invalid_attestations` sweep. External callers
+/// may pick a different value; this only governs the automatic invocation.
+///
+/// [`crate::config::Config::clean_invalid_attestations_tera_gas`] funds a full scan of this
+/// many entries plus ~10 removals. The eleventh removal exceeds the budget and, because the
+/// promise is detached, reverts every one of them silently; recovery is a direct
+/// permissionless call, which attaches the caller's own gas rather than this budget.
+const RESHARE_CLEAN_INVALID_ATTESTATIONS_MAX_SCAN: u32 = 30;
```
gilcu3, line 128: I think the original comment is fine. the explanation, if needed, should go with the code in `clean_invalid_attestations`

crates/contract/src/v3_14_0_state.rs
```diff
@@ -73,7 +77,8 @@ impl From<OldConfig> for Config {
             fail_on_timeout_tera_gas: old.fail_on_timeout_tera_gas,
             fail_attestation_submission_tera_gas: old.fail_attestation_submission_tera_gas,
             clean_tee_status_tera_gas: old.clean_tee_status_tera_gas,
-            clean_invalid_attestations_tera_gas: old.clean_invalid_attestations_tera_gas,
+            clean_invalid_attestations_tera_gas: Config::default()
+                .clean_invalid_attestations_tera_gas,
```
gilcu3, line 79: seems odd to add this issue. As a general rule all our migrations are one shot, so we don't need the TODO or the issue IMO, as it would be done that way anyway

crates/contract/src/v3_14_0_state.rs
```diff
@@ -84,7 +89,8 @@ impl From<OldConfig> for Config {
             verifier_tera_gas: old.verifier_tera_gas,
             resolve_verification_tera_gas: old.resolve_verification_tera_gas,
             launcher_hash_unused_ttl_seconds: old.launcher_hash_unused_ttl_seconds,
-            ..Default::default()
+            // New in this release, so no deployed value exists to carry.
+            attestation_storage_fee_millinear: Config::default().attestation_storage_fee_millinear,
```
gilcu3: why are we changing the pattern? All values that are supposed to take the default could be dealt with by

```
..Default::default()
```

as before. A comment mentioning them could make that extra clear

crates/contract/src/v3_14_0_state.rs
```diff
@@ -153,3 +159,106 @@ struct Metrics {
     sign_with_v1_payload_count: u64,
     sign_with_v2_payload_count: u64,
 }
+
+#[cfg(test)]
+#[expect(non_snake_case)]
+mod tests {
+    use super::*;
```
gilcu3: These tests seem trivial, so IMO they could be removed (even if that goes against our engineering standards). In some cases skipping such tests would make reviews much faster, for example this one. Without them this would have been a tiny PR for example

crates/contract/tests/sandbox/utils/consts.rs
```diff
@@ -17,7 +17,12 @@ pub const ALL_PROTOCOLS: &[Protocol; 4] = &[
 /// gas attachment; in practice, nodes usually attach the maximum available gas. For testing,
 /// we use this constant to attach a fixed amount to each call and detect if gas usage
 /// increases unexpectedly in the future.
-pub const GAS_FOR_VOTE_RESHARED: Gas = Gas::from_tgas(50);
+///
+/// Sized over what the concluding `vote_reshared` needs: the sum of the cleanup-promise
+/// budgets it reserves out of its own prepaid gas, plus its own ~5 TGas burn. Below that,
+/// promise creation fails and takes the deciding reshare vote with it, so the margin is wide
+/// enough that ordinary variance does not make the reshare suite flaky.
```
gilcu3: I think we could remove this explanation, as it is not special about this constant. All gas constants should have the same features: extra Tgas, cleanups awareness, avoid flakiness

crates/contract/tests/sandbox/tee_cleanup_after_resharing.rs
```diff
@@ -63,7 +63,8 @@ async fn reshare__should_leave_valid_non_participant_attestations_in_storage() -
     assert_eq!(nodes_with_tees, expected_node_ids);
 
     // Add two prospective Participants
-    // Note: this test fails if `vote_reshared` needs to clean up more than 3 attestations
+    // Note: this test fails if `vote_reshared` needs to clean up more attestations than
+    // `clean_invalid_attestations_tera_gas` funds within its scan limit.
     let (mut env_non_participant_accounts, non_participants) = gen_accounts(&worker, 1).await;
```
gilcu3, line 66: this reads very strange, are we sure the new comment (or the old comment) is accurate?

crates/contract/tests/sandbox/utils/consts.rs
gilcu3, line 25: this vote is submitted by the node, which currently uses MAX_GAS = 300 for this. Would be good to make sure the relation between these two stays, by defining that constant in the interface. Could be done as a follow-up, as it probably involves several constants

gilcu3 (dismissed):
Thank you, LGTM!

Left some minor comments and suggestions

crates/contract/tests/sandbox/utils/consts.rs
barakeinav1: agree, let do this as a follow-up.

crates/contract/src/v3_14_0_state.rs
barakeinav1: if we remove the todo? how will we remember to clear it?
do we always just remove the old state anyway? (and use a new one?)

crates/contract/src/lib.rs
barakeinav1: Agreed, trimmed. Kept one sentence, because a reader changing this number is editing here rather than in `clean_invalid_attestations`, and the pairing is the thing that must not drift:

> Sized against [`crate::config::Config::clean_invalid_attestations_tera_gas`]: scanning an entry costs gas whether or not it is removed, so raising this without raising that overruns the budget.

The cost model and recovery path are gone.

crates/contract/src/v3_14_0_state.rs
barakeinav1: Removed, and closed the issue. Took your word on the convention — I had assumed these modules were seeded by copying the previous one, so a one-shot line could ride along unnoticed. If they are rewritten per release that risk is not real.

crates/contract/src/v3_14_0_state.rs
barakeinav1: Reverted to `..Default::default()`, with a comment naming the field that takes it.

Context for why it changed: an automated review pointed out that with `..Default::default()` present, deleting a carry-forward line still compiles and silently defaults. That was true, but the same round also had me renumber the test fixture off the real defaults — and that alone closes it, since the test then fails on the mismatch. The enumeration was belt-and-braces, so restoring your pattern costs nothing.

crates/contract/tests/sandbox/utils/consts.rs
barakeinav1: Removed — you are right that none of it is specific to this constant.

crates/contract/tests/sandbox/utils/consts.rs
barakeinav1: Added to #4119 as a follow-up rather than doing it here, since it likely touches several constants.

crates/contract/src/v3_14_0_state.rs
barakeinav1: Removed both. They were run and passing before deletion; noted in the PR description.

crates/contract/tests/sandbox/tee_cleanup_after_resharing.rs
barakeinav1: Good catch — neither version is accurate, so I deleted the note.

some feedback from claude (I didn't have time to dive deep into this)

Tracing it: `git log -S` puts it in #1559 (Dec 2025), where `vote_reshared` spawned three promises and **`clean_tee_status`** was the one that pruned attestations, unbounded. So "fails if it needs to clean up more than 3 attestations" was a real statement about `clean_tee_status_tera_gas` running out on a large enough map.

Pruning later moved into the separate, `max_scan`-bounded `clean_invalid_attestations`, and today `clean_tee_status` only drops votes. The note was stranded by that split.

It is not true of this test either way: every attestation here is `MockAttestation::Valid` (`common.rs:516` and the explicit one for `mpc_signer_accounts[0]`), `Valid` never fails re-verification, and the final assertion is that nothing was removed. So no sweep budget is stressed regardless of map size. My rewrite just restated an obsolete claim in newer words.

crates/contract/src/v3_14_0_state.rs
gilcu3: exactly, they are always rewritten after a release

gilcu3 (approved):

haiyuechen-nearone (approved):

PR #4087, opened 2026-08-07
fix: do not crash on wrong participants in Start message

### LINKED ISSUE #3896: Byzantine participant can crash a node with a single Start message
### Background

An incoming `MpcMessageKind::Start` carries an attacker-controlled `participants` list. `sender_for` maps it directly into `connectivity(*id)` with no membership check, and `connectivity` unwraps a lookup that is `Err` for unknown ids:

```rust
// crates/node/src/p2p.rs:920-922
self.connectivities.get(participant_id).unwrap().clone()
```

So a Start message with any id not in the local participant set panics the node — reachable from any Byzantine participant.

### Acceptance Criteria

- `sender_for` rejects a Start message whose `participants` aren't a subset of the known participant set, instead of panicking.
- A test drives an inbound Start with a bogus participant id and asserts the node errors rather than panics.

gilcu3:
Closes #3896

### DIFF
--- a/crates/node/src/network.rs
+++ b/crates/node/src/network.rs
@@ -476,7 +476,7 @@ async fn run_receive_messages_loop(
         )
         .await
         {
-            tracing::warn!("run_receive_message failed with error: {err}");
+            tracing::warn!("run_receive_message failed with error: {err:#}");
         }
     }
 }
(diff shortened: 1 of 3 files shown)


netrome (approved):
Thank you for fixing this 🙏

pbeza (approved):

PR #4107, opened 2026-08-08
Use exponential backoff for outbound peer reconnection retries

### LINKED ISSUE #4063: Exponential backoff when trying to reconnect to peers
### Background

We're currently debugging an RPC issue, but our logs are spammed with the following message since one peer is also down (independent of the third-party RPC failure)

```
2026-08-05T11:14:53.283101028Z 2026-08-05T11:14:53.282939Z  INFO mpc_node::p2p: could not connect, retrying my_id=8 target_participant_id=13 error=failed to establish tcp stream: Connection refused (os error 111)
```

We should use exponential backoff for outbound connection retries imo, so we don't constantly hammer peers and spam our logs when a node is down.

### User Story

As a node operator I want clean logs and no needless spamming of reconnect attempts.

### Acceptance Criteria

We use exponential backoff in our reconnection attempts for offline peers.

_No response_

ayushsingh82:
## Summary
- `PersistentConnection`'s dial loop in `crates/node/src/p2p.rs` retried on a fixed 1s delay, so a peer that's persistently offline gets dialed — and logged at `info` level — once a second forever, spamming logs (and the network) while unrelated to any other issue.
- Replaced the fixed delay with `backon::ExponentialBuilder` (1s min, 60s max, jittered, unbounded retries), matching the pattern already used in `crates/node/src/indexer/tee.rs`.
- Backoff resets back to the fast 1s retry once a connection succeeds, so a peer that flaps doesn't inherit a long delay from its previous outage.
- Added the resolved `retry_delay` to the existing "could not connect, retrying" log line for visibility into the current backoff state.

Closes #4063

## Verification
- `cargo check -p mpc-node --all-features`: clean
- `cargo clippy -p mpc-node --all-targets --all-features -- -D warnings`: clean (type-checks test code too)
- `cargo fmt -- --check`: clean
- Didn't complete a full `cargo test` link+run locally (same disk constraint as #4102 — linking `mpc-node`'s test binary needs more disk than available here). The two existing `PersistentConnection` tests (`persistent_connection__should_keep_retrying_when_dial_attempt_hangs`, `persistent_connection__should_dial_new_address_after_resolver_update`) use `#[tokio::test(start_paused = true)]`, so tokio's virtual clock auto-advances through `sleep`s regardless of duration — they only assert the loop keeps retrying/dialing, not on the fixed 1s timing, so the growing backoff delays shouldn't affect them. Happy to fix up anything CI's test run surfaces.

## Test plan
- [ ] CI clippy job passes
- [ ] CI test suite passes (including the two `PersistentConnection` retry tests)

### DIFF
--- a/crates/node/src/p2p.rs
+++ b/crates/node/src/p2p.rs
@@ -20,6 +20,7 @@ use crate::protocol_version::CURRENT_PROTOCOL_VERSION;
 use crate::tracking::{self, AutoAbortTask, AutoAbortTaskCollection};
 use anyhow::{Context, anyhow};
 use async_trait::async_trait;
+use backon::{BackoffBuilder, ExponentialBuilder};
 use borsh::{BorshDeserialize, BorshSerialize};
 use bytes::Bytes;
 use ed25519_dalek::VerifyingKey;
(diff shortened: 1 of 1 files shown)


ayushsingh82:
@netrome tagging you as the issue author — this is ready for review whenever you have time.

PR #4109, opened 2026-08-10
chore(test-utils): regenerate attestation fixtures with a retained signer key

### LINKED ISSUE #4111: Regenerate the attestation test fixtures with a retained signer key
### Background

The attestation fixtures in `crates/test-utils/assets/` ship the fixture node's public keys but not its NEAR signer secret key: it was never committed when they were captured, and the node it came from is long gone.

That blocks the cross-contract tests for the `tee-verifier` contract (#4084). Those tests have to call `submit_participant_info` as the fixture node, because the quote's `report_data` binds the node's account key and the contract reads that key from the transaction signer. Sandbox accounts get random keys, so the post-DCAP check fails before anything is stored and the Verified path cannot be exercised at all.

### User Story

As an MPC contract maintainer, I need fixtures whose signer key is available, so that sandbox tests can submit an attestation as the fixture node and exercise the Verified store path.

### Acceptance Criteria

- [ ] Fixtures regenerated with the signer key committed alongside them

Related Slack discussion: https://nearone.slack.com/archives/C07UW93JVQ8/p1786089665038559

Getting the key out is the hard part: it is generated in-enclave and written only inside the CVM, and the one way to run anything in there is an app-compose script, which attestation verification rejects as arbitrary root code. So the hook that exports the key also makes the fixture one that production policy refuses, and test builds have to accept that single field. Full write-up: https://github.com/near/mpc/issues/3787#issuecomment-5216835638

The fixture node is a throwaway localnet account (`frodo.test.near`), which exists on neither mainnet nor testnet, so its secret key can live in the repo.

pbeza:
Needed by #4110, stacked on top: its sandbox tests submit a real attestation and must sign as the fixture node, which is impossible without the node's signer secret key.

The attestation fixtures in `crates/test-utils/assets/` shipped the fixture node's public keys but not that key: it was never committed when they were captured, and the node it came from is long gone. Since the quote's `report_data` binds it, no test could sign as that node. The whole set is regenerated from a fresh capture, with `near_account_secret_key` committed alongside it and the verifier's expected values updated to match.

Exporting the key [needs](https://nearone.slack.com/archives/C07UW93JVQ8/p1786105311587949?thread_ts=1786089665.038559&cid=C07UW93JVQ8) a `pre_launch_script`, which production verification rejects, so `attestation/allow-pre-launch-script` relaxes that one field where test code is built. The hook is committed as `export-signer-key-prelaunch.sh` so the capture can be repeated.

The rest is collection tooling and docs.

Closes #4111


pbeza:
@claude review

pbeza:
@claude review

pbeza:
@claude review

localnet/tee/scripts/rust-launcher/export-signer-key-prelaunch.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+cat > /etc/fixture-exfil.sh <<'EOF'
+#!/bin/sh
+while :; do
+  for f in /var/lib/docker/volumes/*/_data/secrets.json; do
+    [ -f "$f" ] && cat "$f" > /dev/console && exit 0
```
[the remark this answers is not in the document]
pbeza: On purpose, and test-only: the hook only runs when we collect fixtures, and an `app-compose` carrying it is only accepted under `allow-pre-launch-script`, which nothing we ship enables.

Ideas welcome on how to triple-check the flag can't reach a released build.

barakeinav1:
Heads up: this PR drops `validate_app_compose_config__rejects_present_pre_launch_script`, and #4110 is what brings the assertion back as `scripts_absent_with__should_reject_pre_launch_script_when_disallowed`.

So if 4109 lands on its own, main has the relaxation shipped with nothing checking that production still rejects the field.

Could the `scripts_absent_with` refactor move down into this PR? Then 4109 stands on its own and the split stays clean.

barakeinav1:
`allow-pre-launch-script` is a public feature, so a `--all-features` contract build would enable it. The comment says no released WASM can reach it — worth making that enforced rather than trusted:

```rust
#[cfg(all(target_arch = "wasm32", feature = "allow-pre-launch-script"))]
compile_error!("never enable in a wasm build");
```

Same for `sandbox-test-hooks` in #4110.

crates/test-utils/assets/near_account_secret_key
```diff
@@ -0,0 +1 @@
+ed25519:3ih6YgrZivpBXHY86XKtKf5XFmDnLGDwbvPSzZGrwuwJetkDt9NhC1D5889vMnsLBrf82U4QHC5NatoX6umd3AmX
```
barakeinav1: this may trip scanners that look for secrets.
worth adding a comment here that this is only for tests. (otherwise you will need to answer this later over and over).

haiyuechen-nearone:
> allow-pre-launch-script is a public feature, so a --all-features contract build would enable it.

Vibed with Claude to come up with this POC to gate the relaxation differently: https://github.com/near/mpc/pull/4131
@pbeza PTAL and see if that would be better

crates/test-utils/assets/create-assets.sh
```diff
@@ -36,8 +36,18 @@ jq -j '.near_signer_public_key' "$INPUT_FILE" > "$OUTPUT_DIR/near_account_public
...
-# Extract collateral
-jq -r '.tee_participant_info.Dstack.collateral' "$INPUT_FILE" > "$OUTPUT_DIR/collateral.json"
+# The endpoint emits the DER/signature fields as byte arrays (serde_bytes) while the fixture parser
+# reads them as hex, and the PEM chains keep the NUL terminator from the quote's C strings. Fields
+# already in the target form pass through unchanged.
+jq -r 'def tohex:
+         if type == "array" then "0123456789abcdef" as $h
+           | map($h[(. / 16 | floor):(. / 16 | floor) + 1] + $h[(. % 16):(. % 16) + 1]) | join("")
+         else . end;
+       def strip_nul: if type == "string" then split("\u0000")[0] else . end;
```
gilcu3, line 46: I would try to avoid this `jq` scripting if possible. Can we do something about this?

crates/test-utils/assets/public_data.json
gilcu3, line 1: In this file we lost a couple of hex values and now have arrays. Should we try and get the opposite? Was this a change in dstack?

crates/test-utils/src/attestation.rs
```diff
@@ -148,6 +155,21 @@ mod tests {
...
+            .strip_prefix("ed25519:")
+            .expect("secret key is ed25519-prefixed");
+
+        // When
+        let decoded = bs58::decode(secret).into_vec().expect("base58 secret key");
+
+        // Then
+        assert_eq!(decoded.len(), 64);
+        assert_eq!(decoded[32..], account_key());
+    }
```
gilcu3, line 175: that's a very strange way of checking that a private key corresponds to a public key. Doing this should require some crypto operation as well, as the secret key might be malformed

.editorconfig
```diff
@@ -29,6 +29,6 @@ insert_final_newline = false
 # localnet verify_quote call; the text rules do not apply.
 [crates/tee-verifier/tests/fixtures/*.borsh]
 end_of_line = unset
-insert_final_newline = false
+insert_final_newline = unset
```
gilcu3, line 32: what is the difference?

localnet/tee/scripts/rust-launcher/export-signer-key-prelaunch.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
```
barakeinav1: This script prints a private key to the console and now has no comments at all. I think you should add the warning back: "Only ever use this on a test environment: it exports a private key."

gilcu3 (commented):
Added some minor comments and suggestions. Main concern, already expressed on slack, is that it is tricky to avoid including this escape hatch in prod images in the future. There is apparently an alternative by modifying the launcher docker compose that could achieve the same (extracting the near key) without having to change anything in the verifier. If that's correct, it makes sense to try that solution as it would avoid the foot gun here.

pbeza:
@claude review

localnet/tee/scripts/rust-launcher/export-signer-key-prelaunch.sh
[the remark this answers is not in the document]
pbeza: Outdated

localnet/tee/scripts/rust-launcher/export-signer-key-prelaunch.sh
pbeza: Outdated

crates/test-utils/src/attestation.rs
pbeza: 2f71c861

crates/test-utils/assets/near_account_secret_key
pbeza: 2f71c861

crates/test-utils/assets/public_data.json
pbeza: > (...) Was this a change in dstack?

Yeah, good question. I was trying to figure that out before asking for a review, but I couldn’t find any specific commit or PR where it changed. :(

.editorconfig
pbeza: If you revert it, you'll get this error:

```
rates/tee-verifier/tests/fixtures/verify_quote_args.borsh:
	No final newline expected
1 errors found
```

Explained by Claude:
> The reason is a one-byte **coincidence**: `main`'s borsh fixture happens to end with `00`, so `insert_final_newline = false` was satisfiable there. Today's regenerated fixture ends with `0a`, and the checker reads that trailing byte as a final newline. Since it's encoded payload, "fixing" it would corrupt the fixture.

crates/test-utils/assets/create-assets.sh
pbeza: I agree, but I don’t think it’s trivial. We should probably handle it in a follow-up issue since it’s a separate, debatable thing: https://github.com/near/mpc/issues/4140. I already addressed it in this stacked PR: https://github.com/near/mpc/pull/4143. Feel free to review.

pbeza:
@gilcu3 @haiyuechen-nearone @barakeinav1 could you please re-review? Thanks!

crates/test-utils/assets/README.md
```diff
@@ -71,14 +87,25 @@ All files will be written into the specified output directory.
...
+   ```
+
+   The second run fails on `verify_quote__should_return_verified_td10_report_for_valid_fixture`
+   and prints the values actually produced; copy them into
+   `crates/tee-verifier/tests/verify_quote.rs`.
+
 ## Tests that depend on these assets
 
-After updating assets, these tests should pass:
+After updating assets, run the crates that consume them:
```
gilcu3: nit: run the crates seems of, probably is run the tests in crates that consume them

localnet/tee/scripts/rust-launcher/single-node-readme.md
```diff
@@ -44,6 +44,9 @@ export OS_IMAGE=dstack-0.5.8
 # Custom NEAR accounts
 export NODE_ACCOUNT=frodo.test.near
 export CONTRACT_ACCOUNT=mpc-contract.test.near
+
+# Launcher compose to deploy. The collection variant exports the node's in-enclave signer key.
+export COMPOSE_TEMPLATE=$PWD/localnet/tee/scripts/rust-launcher/export-signer-key-compose.yaml.template
```
gilcu3: what do we mean by "The collection variant"? Would be simpler to say "This variant"

gilcu3 (dismissed):
Thank you!

localnet/tee/scripts/rust-launcher/single-node-readme.md
pbeza: Fixed: https://github.com/near/mpc/pull/4109/commits/3118b2a50b11ff1d7767c5b19f2611159f64657a. ✅

crates/test-utils/assets/README.md
pbeza: Fixed: https://github.com/near/mpc/pull/4109/commits/3118b2a50b11ff1d7767c5b19f2611159f64657a. ✅

crates/contract/src/tee/proposal.rs
```diff
@@ -487,6 +487,23 @@ impl AllowedLauncherImages {
             .find(|e| &e.launcher_hash == launcher_hash)
             .map(|e| e.expires_at.as_secs())
     }
+
+    /// Test-only: allows one more compose hash for an already-allowed launcher. The attestation
+    /// fixture is captured from a CVM whose launcher compose carries a key-export service, so
+    /// [`get_docker_compose_hash`] cannot derive its hash.
+    #[cfg(test)]
```
haiyuechen-nearone, line 494: Nice

haiyuechen-nearone (approved):

gilcu3 (approved):

PR #4110, opened 2026-08-10
test(contract): sandbox tests for the attestation flow against the real tee-verifier

### LINKED ISSUE #4084: Sandbox tests for the attestation flow against the real `tee-verifier` contract
### Background

#3714 moved DCAP verification into the `tee-verifier` contract, called from `submit_participant_info` via a cross-contract Promise. The sandbox tests from #3715 drive the real deployed verifier, but only for the failure verdicts; the Verified verdict and the store path are covered in-process only, because the fixture collateral is time-expired against the forward-only sandbox clock. A stub verifier was deliberately dropped in #3715, so the coverage must reuse the real contracts.

### User Story

As an MPC contract maintainer, I need the Verified attestation path exercised against the real deployed verifier, so that wire-format, gas-budget, and storage regressions are caught in CI instead of on a live TDX deployment.

### Acceptance Criteria

- [ ] A sandbox test drives `submit_participant_info` through the real `tee-verifier` WASM to a Verified verdict at the production `verifier_tera_gas` budget.
- [ ] No stub or duplicated test-only verifier contract; a cfg-gated test hook in the real crate is acceptable.
- [ ] Store-path tests that must sign as the fixture account land guarded until the fixture secret key asset exists (#3787).
- [ ] E2E clusters deploy and vote in the tee-verifier for topology parity with production.

Related: #3714, #3715, #3738, #3787, #3897. Testing section of `docs/design/attestation-verifier-contract.md`.

pbeza:
Closes #4084


crates/tee-verifier/src/lib.rs
```diff
@@ -62,3 +65,19 @@ impl TeeVerifier {
         }
     }
 }
+
+/// The timestamp quotes are verified against: block time, unless a sandbox test
+/// pinned one under [`tee_verifier_interface::SANDBOX_TEST_PINNED_NOW_STORAGE_KEY`]
+/// (spelled in full because the import is feature-gated). The pin exists
+/// because sandbox chain time is wall-clock and forward-only, so it can never
+/// fall inside the validity window of a checked-in collateral fixture.
+fn now_seconds() -> u64 {
```
barakeinav1, line 83: Production compiles this down to just the block timestamp, so the gating itself is right. But `sandbox-test-hooks` is a public feature, so a `--all-features` build turns it on — and here that means expired collateral verifies.

Worth making it unbuildable rather than just unused:

```rust
#[cfg(all(target_arch = "wasm32", feature = "sandbox-test-hooks"))]
compile_error!("never enable in a wasm build");
```

pbeza:
@claude review

pbeza:
Closing in favor of https://github.com/near/mpc/pull/4165.

PR #4115, opened 2026-08-11
chore(contract): correct the grants map in the design doc and measure its row cost

### LINKED ISSUE #4114: Correct the grants map type in the design doc and measure its row cost
### Background

#4016 switched `available_attestation_grants` from `LookupMap` to `IterableMap` late in review, but `docs/design/operator-prepaid-attestation-storage.md` still describes the old type and sizes the fee off it: it puts the grants-map row at ~130 bytes / ~0.0013 NEAR, where a measured `IterableMap` row is 194 bytes / 0.00194 NEAR for a worst-case 64-char account.

The fee is unaffected — the floor moves from ~0.0073 to ~0.008 NEAR, still ~2.5x under the 0.02 NEAR fee.

### User Story

As an engineer reading the design doc, I want it to match what shipped, so that I can size the fee or reason about the grants map without re-deriving it from the code.

### Acceptance Criteria

- The doc names `IterableMap` and the measured row cost, and the fee table's floor matches.
- The floor is measured rather than pinned as a constant, and the default fee is held to twice it, so growth is caught before it breaches.
- The full prepay + submit flow is checked too, so a write outside the two maps fails.

- Introduced by: #4016
- Follow-up: #4135 (the sandbox test cannot yet submit `expected_measurements`)

barakeinav1:
Closes #4114.

The design doc drifted from what #4016 shipped:

- It calls the grants map a `LookupMap`; the contract uses an `IterableMap`. Corrected.
- It estimates the grants-map row at ~130 bytes; measured, it is 194 (0.00194 NEAR). Fee table and floor updated — ~0.008 NEAR, against the unchanged 0.02 NEAR fee.

Plus two tests, so the numbers stay true. Neither pins a byte count; both measure.

- A unit test measures the floor and requires the fee to be at least twice it, so growth is caught before it breaches.
- A sandbox test runs the real prepay + submit flow, checks the prepayment reaches the contract balance, and holds the storage one grant buys to half a fee. This is the one that catches a write outside the two maps.

The sandbox test submits the largest mock reachable today (image hash, launcher compose hash, expiry). `expected_measurements` needs a vote helper that does not exist yet — #4135.

### DIFF
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -8846,8 +8846,8 @@ mod tests {
     /// [`WORST_CASE_ENTRY_BYTES`] for storage-price changes.
     const WORST_CASE_ENTRY_COST_CEILING: NearToken = NearToken::from_millinear(10);
 
-    /// One `available_attestation_grants` row, for the longest possible account id. The fee
-    /// covers this on top of the entry, since the first grant for an account creates the row.
+    /// Worst case is the longest possible account id. The fee covers this row on top of the
+    /// entry; an account's first grant creates it.
     const WORST_CASE_GRANT_ROW_BYTES: u64 = 194;
 
     fn worst_case_dstack_attestation() -> VerifiedAttestation {
@@ -8892,8 +8892,6 @@ mod tests {
         env::storage_usage() - before
     }
 
-    /// Charged bytes for one grants-map row, measured the same way as
-    /// [`measure_stored_entry_bytes`]: insert into the real map, flush, take the delta.
     fn measure_grant_row_bytes() -> u64 {
         testing_env!(VMContextBuilder::new().build());
         let account: AccountId = "a".repeat(64).parse().unwrap();
@@ -8937,8 +8935,7 @@ mod tests {
         );
     }
 
-    /// Pinned for the same reason as [`stored_attestation_entry__should_have_the_pinned_size`]:
-    /// the prepaid-storage fee is sized from this number plus the entry, so a layout change
+    /// The prepaid-storage fee is sized from this number plus the entry, so a layout change
     /// that grows the row must be seen and the fee revisited.
     #[test]
     fn grant_row__should_have_the_pinned_size() {
@@ -8956,10 +8953,8 @@ mod tests {
     /// row. Pinning the byte counts alone would not catch a lowered default fee, which breaks
     /// the same guarantee.
     ///
-    /// Two things it does not guard. The fee is votable, so a network can choose a value below
-    /// the floor. And `storage_byte_cost` is a near-sdk constant, not a protocol read, so a
-    /// real re-pricing moves the floor without failing here — the residual the design doc's
-    /// Security section accepts.
+    /// Does not guard a votable fee set below the floor, nor a real storage re-pricing —
+    /// `storage_byte_cost` is a near-sdk constant, not a protocol read.
     #[test]
     fn attestation_storage_fee__should_cover_the_entry_and_its_grant_row() {
         // Given
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -8892,8 +8892,6 @@ mod tests {
         env::storage_usage() - before
     }
 
-    /// Measured through the contract's own map, so that changing its declaration moves this
-    /// number rather than leaving the fee sized on a stale one.
     fn measure_grant_row_bytes() -> u64 {
         let (_, mut contract, _) = basic_setup(Curve::Edwards25519, &mut OsRng);
         testing_env!(VMContextBuilder::new().build());
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -8957,7 +8957,8 @@ mod tests {
     #[test]
     fn attestation_storage_fee__should_cover_the_entry_and_its_grant_row() {
         // Given
-        let floor_bytes = WORST_CASE_ENTRY_BYTES + WORST_CASE_GRANT_ROW_BYTES;
+        let floor_bytes =
+            measure_stored_entry_bytes(worst_case_mock_attestation()) + measure_grant_row_bytes();
 
         // When
         let floor = env::storage_byte_cost().saturating_mul(u128::from(floor_bytes));
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -8847,7 +8847,7 @@ mod tests {
     const WORST_CASE_ENTRY_COST_CEILING: NearToken = NearToken::from_millinear(10);
 
     /// One `available_attestation_grants` row, for the longest possible account id. The fee
-    /// covers this on top of the entry, since a grant creates the row.
+    /// covers this on top of the entry, since the first grant for an account creates the row.
     const WORST_CASE_GRANT_ROW_BYTES: u64 = 194;
 
     fn worst_case_dstack_attestation() -> VerifiedAttestation {
@@ -8953,11 +8953,13 @@ mod tests {
     }
 
     /// The fee has to cover what one grant actually buys: the worst-case entry plus the grants
-    /// row that holding it creates. Pinning the byte counts alone would not catch a storage
-    /// price rise or a lowered default fee, which break the same guarantee.
+    /// row. Pinning the byte counts alone would not catch a lowered default fee, which breaks
+    /// the same guarantee.
     ///
-    /// Guards the shipped default only — the fee is votable, so a network can still choose a
-    /// value below the floor.
+    /// Two things it does not guard. The fee is votable, so a network can choose a value below
+    /// the floor. And `storage_byte_cost` is a near-sdk constant, not a protocol read, so a
+    /// real re-pricing moves the floor without failing here — the residual the design doc's
+    /// Security section accepts.
     #[test]
     fn attestation_storage_fee__should_cover_the_entry_and_its_grant_row() {
         // Given
--- a/crates/contract/src/lib.rs
+++ b/crates/contract/src/lib.rs
@@ -8848,8 +8848,6 @@ mod tests {
 
     /// Worst case is the longest possible account id. The fee covers this row on top of the
     /// entry; an account's first grant creates it.
-    const WORST_CASE_GRANT_ROW_BYTES: u64 = 194;
-
     fn worst_case_dstack_attestation() -> VerifiedAttestation {
         VerifiedAttestation::Dstack(ValidatedDstackAttestation {
             mpc_image_hash: MAX_HASH.into(),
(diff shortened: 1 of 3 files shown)


barakeinav1:
All three addressed in 3a69082e.

**The 194 figure.** Re-measured, and it holds: 194 bytes for a 64-char account, stable across five consecutive inserts. The hand count of ~230 does not match what the runtime charges here. Fair challenge though — my first probe reused the same storage prefix across `testing_env!` calls and gave nonsense (36 bytes for a 20-char account, 68 for an 8-char one), so it was worth checking.

**The reproducibility point is the real one, and it is fixed.** `measure_grant_row_bytes` now mirrors `measure_stored_entry_bytes`, and `grant_row__should_have_the_pinned_size` pins 194, so the number comes from the repo by the same method as 604/599 and a layout change fails the test rather than silently invalidating the fee. That is what the section argues, so it should have been guarded the same way.

**Breadcrumb restored** — the paragraph names both helpers again.

Also added to the table that 194 is the worst case and a realistic 20-char account is 151, since the conservative number is the one the fee is sized on.

barakeinav1:
@claude review

barakeinav1:
All three in 8eed9f60.

**The misplaced doc comment** — correct, and a fair thing to catch in this PR of all PRs. My insert anchored on the `fn` line rather than the doc block, so `measure_stored_entry_bytes` lost its doc and the new helper inherited one about `NodeId` field widths. Moved below the closing brace.

**The 20-char figure** — dropped. Worth recording why the arithmetic looked wrong: 151 is right, the label was not. `node-operator-01.near` is 21 characters, not 20, and 64 − 21 = 43, exactly the measured delta. So the account id contributes once, not twice — which is also why the earlier ~230 hand count came out high. Either way it was an unpinned number in prose, which is what this PR exists to remove.

**Floor vs fee** — added here rather than deferred, since it is the property the doc paragraph actually claims:

```rust
let floor = env::storage_byte_cost().saturating_mul(u128::from(WORST_CASE_ENTRY_BYTES + WORST_CASE_GRANT_ROW_BYTES));
assert!(floor <= NearToken::from_millinear(Config::default().attestation_storage_fee_millinear.into()));
```

Checked it bites: lowering the default fee to 5 mNEAR fails it with `attestation storage fee (0.005 NEAR) must cover the floor (798 bytes, 0.008 NEAR)`. It guards the shipped default only — the fee is votable, so a network can still choose a value below the floor, and the test says so.

Unrelated: `Extra CI checks` is red on a `cargo-deny` advisory (`advisories FAILED, bans ok, licenses ok, sources ok`). This PR adds no dependencies.

barakeinav1:
@claude review

barakeinav1:
All three in 7051029a.

**The storage-price claim** — you are right, and I checked rather than took it: `env::storage_byte_cost()` is a hardcoded `1e19` in near-sdk 5.29 (`env.rs:2703`), which is why `deposits.rs` pins that exact number. So the test trips on a lowered default fee or an SDK constant bump, not on a protocol re-pricing. Reworded to say what it guards, and to name the re-pricing as the residual the design doc's Security section already accepts.

**"a grant creates the row"** — correct, it is the first grant for an account. `prepay_attestation_storage(account, 5)` writes one row and later grants rewrite a fixed-width `u32`. Reworded; the floor is unchanged and stays conservative.

**Floor rounding** — now `0.00798 NEAR`, so the 2.5x is re-derivable from the table.

crates/contract/src/lib.rs
```diff
@@ -8918,6 +8937,48 @@ mod tests {
...
+    }
+
+    /// The fee has to cover what one grant actually buys: the worst-case entry plus the grants
+    /// row. Pinning the byte counts alone would not catch a lowered default fee, which breaks
+    /// the same guarantee.
+    ///
+    /// Two things it does not guard. The fee is votable, so a network can choose a value below
+    /// the floor. And `storage_byte_cost` is a near-sdk constant, not a protocol read, so a
+    /// real re-pricing moves the floor without failing here — the residual the design doc's
+    /// Security section accepts.
```
pbeza: I'd prefer not to refer to the design docs in the comments.

crates/contract/src/lib.rs
```diff
@@ -8918,6 +8937,48 @@ mod tests {
...
+    #[test]
+    fn attestation_storage_fee__should_cover_the_entry_and_its_grant_row() {
+        // Given
+        let floor_bytes = WORST_CASE_ENTRY_BYTES + WORST_CASE_GRANT_ROW_BYTES;
```
pbeza: I’m a bit torn about having these kinds of tests because they tightly couple the tests to our `IterableMap` implementation, but I think that’s consistent with how we’ve written similar tests already, so I’m okay with it.

pbeza (dismissed):

crates/contract/src/lib.rs
barakeinav1: why?, I don't care about this specific comment, but in general why should we not refer to the design doc for WHY things are done the way they are.

crates/contract/src/lib.rs
barakeinav1: in this case we want to check storage cost, and this is tied directly to the implantation and usage of `IterableMap`

crates/contract/src/lib.rs
```diff
@@ -8888,6 +8892,21 @@ mod tests {
         env::storage_usage() - before
     }
 
+    /// Charged bytes for one grants-map row, measured the same way as
+    /// [`measure_stored_entry_bytes`]: insert into the real map, flush, take the delta.
+    fn measure_grant_row_bytes() -> u64 {
```
kevindeforth: Please avoid low-quality LLM comments, they bloat the codebase.
 
- Why does the reader of this commentneed to know about `measure_stored_entry_bytes`?
- What is `delta` referring to?
- What is "the real map" referring to?

I find the function name descriptive enough, no need to add the comment.

crates/contract/src/lib.rs
```diff
@@ -8918,6 +8937,48 @@ mod tests {
...
+    fn grant_row__should_have_the_pinned_size() {
+        // Given / When
+        let bytes_stored = measure_grant_row_bytes();
+
+        // Then
+        assert_eq!(
+            bytes_stored, WORST_CASE_GRANT_ROW_BYTES,
+            "grants row size changed; see this test's doc comment before updating the number"
+        );
+    }
```
kevindeforth: We are testing test-only methods and constants. Why is that?

kevindeforth (changes requested):
Thanks, two things:

- I stopped the detailed review after encountering a low-quality LLM-generated comment. Please do a pass to make sure the PR is free of low-quality LLM artifacts before requesting review. I commented on the one I encountered, but I believe there are a few others as well.
- I'm not sure what the purpose is of defining the constant and testing it only inside the test module. It feels like we should either use it in production code or drop it. What led you to write the test?

crates/contract/src/lib.rs
barakeinav1: Fair, removing it.

crates/contract/src/lib.rs
barakeinav1: Removed.

crates/contract/src/lib.rs
barakeinav1: I fixed the method in 2016cde2 to use the real contract implementation, so what's under test is production storage, not the helper. The constant is just the pinned expectation.

The pin keeps the other test honest: `attestation_storage_fee__should_cover_the_entry_and_its_grant_row` reads `WORST_CASE_GRANT_ROW_BYTES`, so without it that test rests on an unverified number. We pin the attestation entry's size the same way, just above.

barakeinav1: One more reason to keep the constant: the floor test only fails once the fee is actually breached — the row could grow from 194 to ~1400 bytes before that happens. The pin catches it the moment anything moves.

crates/contract/src/lib.rs
pbeza: > why?, I don't care about this specific comment, but in general why should we not refer to the design doc for WHY things are done the way they are.

The way I see it, design docs _inevitably_ become outdated, and it’s _hard_ to keep them properly cross-linked with the code. IMHO, once a design is implemented, we shouldn’t try to keep the design doc in sync or treat it as an ongoing source of truth. The code and the documentation around it should be the source of truth.

It’s already hard enough to keep code comments up to date, let alone external design docs, so I’d rather not overcomplicate things. Design docs serve their purpose during the design and implementation phase; after that, we can largely leave them as a historical record.

crates/contract/src/lib.rs
barakeinav1: refactored PR.

barakeinav1:
@claude review

crates/contract/src/lib.rs
```diff
@@ -8846,6 +8846,8 @@ mod tests {
     /// [`WORST_CASE_ENTRY_BYTES`] for storage-price changes.
     const WORST_CASE_ENTRY_COST_CEILING: NearToken = NearToken::from_millinear(10);
 
+    /// Worst case is the longest possible account id. The fee covers this row on top of the
+    /// entry; an account's first grant creates it.
```
haiyuechen-nearone: Is this comment correct...? This function creates the worst case attestation but has no relation to an account, right?

barakeinav1: no, I left it by mistake during the refactor.

barakeinav1: Right — that comment was written for `measure_grant_row_bytes` and ended up on the wrong function when I moved it. Removed in b920446a.

barakeinav1:
All fixed in b920446a.

> `crates/contract/src/lib.rs:8849` — the new doc comment is attached to `worst_case_dstack_attestation`, which builds a Dstack attestation and has nothing to do with account ids or grants

Correct, and the same one @haiyuechen-nearone caught. It was written for `measure_grant_row_bytes` and landed on the wrong function when I moved it. Removed.

> `docs/design/operator-prepaid-attestation-storage.md:64` — "the tests hold the fee to twice the floor **rather than pinning these numbers**" is not true of the entry figures in the same table.

Right. I dropped the grants-row constant only; the entry pin predates this PR. The doc now says which figure is pinned and which is measured.

> `crates/contract/src/lib.rs:8948` — the floor measures the mock variant only, on the standing assumption that mock (604) is larger than dstack (599).

Done — the floor is now `max(mock, dstack)`, so the ordering stops being load-bearing.

> `crates/contract/tests/sandbox/tee.rs:1059` — the name promises something different from what the assertion delivers

Renamed to `prepay_and_submit_a_constrained_mock__should_use_at_most_half_a_grant_fee_of_storage`.

> `crates/contract/tests/sandbox/tee.rs:1057` — "240 of the 304 bytes between this entry and the worst case" is an unpinned prose number

Dropped from the comment; #4135 carries the breakdown.

> `crates/contract/src/lib.rs:8939` — `storage_byte_cost` is a plain code span where the convention asks for an intra-doc link

Linked.

> Cross-reference: issue #4123 ... names `attestation_storage_fee__should_cover_the_entry_and_its_grant_row` — a name this PR replaces

Updated the issue body.

haiyuechen-nearone (dismissed):
🚢

crates/contract/src/lib.rs
```diff
@@ -8918,6 +8930,36 @@ mod tests {
...
+    /// A failure here is a prompt to re-price, not a broken test.
+    ///
+    /// Does not guard a real storage re-pricing: [`env::storage_byte_cost`] is a near-sdk
+    /// constant, not a protocol read.
+    ///
+    /// TODO(#4123): a fee voted below the floor is not caught here either.
+    #[test]
+    fn attestation_storage_fee__should_keep_double_the_floor() {
+        // Given
+        const BUFFER: u128 = 2;
```
kevindeforth, line 8944: instead of hard-coding the `2`, can we use the constant `DEFAULT_ATTESTATION_STORAGE_FEE_MILLINEAR` which is used in the contract to initialize config?

(Not a blocker, because we have the sandbox test).

crates/contract/tests/sandbox/tee.rs
```diff
@@ -1046,6 +1049,86 @@ async fn verify_tee__should_keep_participants_and_stop_signing_when_kickout_drop
...
+        .await?
+        .json()?;
+    assert_eq!(remaining, GRANTS - 1, "the row must outlive the submission");
+
+    let after = contract.as_account().view_account().await?;
+    let grown = after.storage_usage - before.storage_usage;
+    let charged = STORAGE_BYTE_COST_YOCTONEAR.saturating_mul(u128::from(grown));
+    let fee = NearToken::from_millinear(u128::from(config.attestation_storage_fee_millinear))
+        .as_yoctonear();
+    let deposited = NearToken::from_yoctonear(fee.saturating_mul(u128::from(GRANTS)));
```
kevindeforth: nit: Calling this `expected_deposit` would improve readability a bit.

kevindeforth (dismissed):
Thank you!

crates/contract/tests/sandbox/tee.rs
barakeinav1: agree, will update.

crates/contract/src/lib.rs
barakeinav1: not sure what you meant.
DEFAULT_ATTESTATION_STORAGE_FEE_MILLINEAR - is what we charge per grant
2  - this is a buffer to make sure the test triggers when the real cost exceeds 
2*DEFAULT_ATTESTATION_STORAGE_FEE_MILLINEAR

kevindeforth (approved):

crates/contract/src/lib.rs
kevindeforth: Oh apologies, I mis-read this test. Thanks for your patience!

haiyuechen-nearone (approved):

PR #4116, opened 2026-08-11
chore(deny): ignore the unmaintained advisories for `im` and its dependencies

pbeza:
`cargo deny check advisories` fails on every PR since RUSTSEC-2026-0247 (bitmaps), -0248 (im) and -0251 (sized-chunks) were published. All three come in through `im v15.1.0`, pulled by nearcore's `near-network`, so they cannot be dropped here, and unmaintained advisories carry no patched version.

Ignored by ID rather than downgrading `unmaintained` to a warning, so future advisories of that class still fail the build.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -33,7 +33,7 @@ ignore = [
   # has no patched release. Only reachable through near-vm-runner, which never
   # shares a Store across Engines.
   "RUSTSEC-2026-0222", # wasmtime stores can mix up type indices between engines
-  # `im` and its dependencies, reached only through nearcore's `near-network`.
+  # TODO(#4117): remove once nearcore moves `near-network` off `im`, whose maintained fork is `imbl`.
   "RUSTSEC-2026-0247", # bitmaps is unmaintained
   "RUSTSEC-2026-0248", # im is unmaintained
   "RUSTSEC-2026-0251", # sized-chunks is unmaintained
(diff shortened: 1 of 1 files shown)


deny.toml
```diff
@@ -33,6 +33,10 @@ ignore = [
   # has no patched release. Only reachable through near-vm-runner, which never
   # shares a Store across Engines.
   "RUSTSEC-2026-0222", # wasmtime stores can mix up type indices between engines
+  # `im` and its dependencies, reached only through nearcore's `near-network`.
```
barakeinav1: should we a add a todo here?
# Remove once nearcore migrates `near-network` from `im` to `imbl``? Non-blocking.

barakeinav1 (dismissed):
LGTM. thanks.

could we add the exit condition like the entries above, e.g. `# Remove once nearcore migrates `near-network` from `im` to `imbl``?

haiyuechen-nearone (dismissed):

deny.toml
pbeza: https://github.com/near/mpc/pull/4116/changes/83fe5265f719a8dd1f8f27e711217fc0ab38e6a6 ✅

haiyuechen-nearone (approved):

barakeinav1 (approved):

PR #4130, opened 2026-08-12
docs+build-images: document catatonit as a reproducible-build dependency

### LINKED ISSUE #4129: docs+build-images: catatonit missing from reproducible-build prerequisites
### Background

`repro-env` runs its reproducible build inside podman and needs `/usr/bin/catatonit` as the container entrypoint, but the operator-facing prerequisite lists never mention it — `docs/reproducible-builds.md:16-19` (also missing `skopeo`, which `build-images.sh:63` requires unconditionally) and `docs/running-an-mpc-node-in-tdx-external-guide.md:1852`/`:1958` (missing `catatonit` + `podman`). An external operator reproducing the node image to verify the **manifest digest they vote on** hits the opaque `statfs /usr/bin/catatonit` error — the same one the CI outage (#4126) surfaced.

Separately, `build-images.sh`'s repro-env preflight (`:65`) omits `catatonit` from `require_cmds` and gates on `USE_NODE || USE_RUST_LAUNCHER` — missing `--node-gcp`, which also runs `repro-env build` (`:158`), so it skips the check entirely.

### Acceptance Criteria

- [ ] Prerequisite lists include `catatonit` (and `skopeo`); the node heading also covers the launcher image
- [ ] `build-images.sh` preflight checks `catatonit` and its gate includes `--node-gcp`

- Follow-up to #4125 / #4128 / #4126.

barakeinav1:
Follow-up to #4128 — the docs/preflight side (no workflow files).

`repro-env` requires `catatonit` (its podman container entrypoint), but the reproducible-build prerequisite lists never mentioned it — nor `skopeo` (required unconditionally). So an operator reproducing the node image to verify a manifest digest before voting hits the opaque `statfs /usr/bin/catatonit` error the CI outage surfaced. Also hardens `build-images.sh`'s preflight to check `catatonit` and cover `--node-gcp` (which reaches `repro-env build` but was skipping the check).

Docs-only + one preflight guard; no behavior change to successful builds (a missing dep now fails fast with `Missing dependency: catatonit` instead of podman's opaque error).

Fixes #4129

### DIFF
--- a/deployment/build-images.sh
+++ b/deployment/build-images.sh
@@ -62,8 +62,8 @@ require_cmds() {
 
 require_cmds docker git find touch skopeo
 
-if $USE_NODE || $USE_RUST_LAUNCHER; then
-    require_cmds repro-env podman
+if $USE_NODE || $USE_NODE_GCP || $USE_RUST_LAUNCHER; then
+    require_cmds repro-env podman catatonit
 fi
 
 if ! docker buildx &>/dev/null; then
--- a/docs/reproducible-builds.md
+++ b/docs/reproducible-builds.md
@@ -12,11 +12,13 @@ security and verification purposes.
 - `docker` with buildx support
 - `jq`
 - `git`
+- `skopeo`
 
-**Additional requirements for building the node image**:
+**Additional requirements for building the node or launcher image**:
 
 - `repro-env` - Tool for reproducible build environments ([install here](https://github.com/kpcyrd/repro-env))
 - `podman`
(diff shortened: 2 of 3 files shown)


barakeinav1:
Superseded and closing. `main` refactored the reproducible build to run skopeo **inside podman** (`build-images.sh` now `podman run`s the pinned skopeo image) and moved CI to **Ubuntu 26.04** runners, so host `catatonit` is no longer required — it's no longer referenced anywhere on `main`. The doc/`build-images.sh` changes here now conflict with that approach and are moot.

PR #4132, opened 2026-08-12
fix: compare foreign chain whitelist header names case insensitively

### LINKED ISSUE #4122: Compare foreign chain whitelist header names case insensitively
### Problem

An operator running Aptos gets this warning on every whitelist refresh, and cannot make it go away:

```
WARN mpc_node::foreign_chain_whitelist_verifier: foreign-chain whitelist mismatch chain=Aptos provider="geomi" kind=AuthSchemeNameMismatch { variant: "Header", local_name: "authorization", contract_name: "Authorization" }
```

HTTP headers are case insensitive so a difference in casing should not produce a warning.

Where is the bug?
The local header name is parsed into an `http::HeaderName`, which normalizes to lowercase. The contract stores whatever string the vote carried. `crates/node/src/foreign_chain_whitelist_verifier.rs:291` compares the two byte for byte, so an entry voted in as `Authorization` will never match local value stored in the Node.

The warning is diagnostic only, so no Node operation is blocked by this bug. The effect is a warning log produced for a non-error every 5 minutes.

### Acceptance Criteria

- [ ] Header names that differ only in case are reported as matching.
- [ ] Header names that differ beyond case are still reported as mismatched.

Reproduced on a two node e2e test cluster: an Aptos entry voted in as `Authorization` plus a local config naming the same header produces the line above verbatim, including when the local config also spells it `Authorization`.

Note for whoever picks this up: the node refetches the whitelist every 300 seconds, so a changed verdict takes up to 5 minutes to appear, or a restart.

haiyuechen-nearone:
Closes #4122

### DIFF
--- a/crates/node/src/foreign_chain_whitelist_verifier.rs
+++ b/crates/node/src/foreign_chain_whitelist_verifier.rs
@@ -288,9 +288,7 @@ fn compare_auth(
                 scheme: contract_scheme,
             },
         ) => {
-            // Field names are case insensitive: `HeaderName` lowercases the local name,
-            // while the contract stores the string the vote carried verbatim.
-            if !local_h.as_str().eq_ignore_ascii_case(contract_h) {
+            if !header_name_matches(local_h, contract_h) {
                 out.push(Diagnostic {
                     chain,
                     provider: Some(name.clone()),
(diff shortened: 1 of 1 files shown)


kevindeforth (approved):

crates/node/src/foreign_chain_whitelist_verifier.rs
```diff
@@ -288,7 +288,7 @@ fn compare_auth(
                 scheme: contract_scheme,
             },
         ) => {
-            if local_h.as_str() != contract_h {
+            if !header_name_matches(local_h, contract_h) {
```
barakeinav1, line 291: nit: I think this is an overkill since this is used only here.

barakeinav1 (approved):
approved,
but I think  adding a wrapper+tests for eq_ignore_ascii_case is an overkill

PR #4134, opened 2026-08-13
ci: move runner images to ubuntu 26.04

### LINKED ISSUE #3433: skopeo version matters for reproducibility
### Background

During the 3.11.1 release I observed that in my local system the rust-launcher binary matched CI, but not the docker manifest. This happened right after I upgraded skopeo to `1.23.0`.

### User Story

_No response_

### Acceptance Criteria

We should pin the version of skopeo that we use to achieve reproducible builds.

_No response_

### LINKED ISSUE #4133: Move CI runners from Ubuntu 24.04 to 26.04
### Background

The 2026-08-12 outage (#4126) came from [actions/runner-images#14412](https://github.com/actions/runner-images/pull/14412), which on 24.04 swapped apt `podman` for a pinned `mgoltzsche/podman-static` v5.8.4 bundle because Ubuntu 24.04 only ships podman 4.9.3. That dropped `/usr/bin/catatonit` (an apt `Recommends` of `podman`), which `repro-env` bind-mounts as the container entrypoint. It reached us in image `ubuntu24/20260810.271`. The bundle has already produced two upstream follow-up fixes ([#14480](https://github.com/actions/runner-images/pull/14480), [#14517](https://github.com/actions/runner-images/pull/14517)) for ownership, crun selection and a non-setuid `fusermount3`, so more fallout is likely.

On 26.04 that branch does not apply: apt provides podman 5.x directly, so `catatonit` is back and the workarounds in #4125 / #4128 can go. All our jobs pin `warp-ubuntu-2404-*` labels.

### Acceptance Criteria

- CI and image-publishing workflows run on Ubuntu 26.04 runners.
- The `apt-get install -y catatonit` steps added in #4125 / #4128 are removed.
- Confirm whether the `kernel.apparmor_restrict_unprivileged_userns=0` sysctl is still needed on 26.04, and drop it if not.

gilcu3:
Closes #4133

Build reproduced in alice, my machine, and ci https://github.com/near/mpc/actions/runs/31679388608/job/94381155133:

```
commit hash: 1606c18ab590ac4e929b4de09a48dd0999917f0c
SOURCE_DATE_EPOCH used: 0
rust launcher binary hash: 78a94cb6ccc2138badb46cc1de71b813f31f75abaf669b41a3a9765f9cce0105
rust launcher manifest digest: sha256:5e48bda2d031a73a179e8274d63119340a6e94802315d9c6e8e3d8b3d68cfc28
```

The second commit fixes #3433 because it also appeared to affect us after the image bump. I think it is a small change, but could be also done in a separate PR (as a premise)


scripts/build-and-verify-rust-launcher-docker-image.sh
```diff
@@ -8,10 +8,3 @@
 set -euo pipefail
 
 ./deployment/build-images.sh --rust-launcher
-
-temp_dir=$(mktemp -d)
-trap 'rm -rf "$temp_dir"' EXIT
-echo "using $temp_dir"
-skopeo copy --all --dest-compress docker-daemon:mpc-rust-launcher:latest dir:"$temp_dir"
-built_hex="$(sha256sum "$temp_dir/manifest.json" | cut -d' ' -f1)"
-echo "Built launcher image hash: sha256:${built_hex}"
```
gilcu3, line 17: this was redundant

gilcu3:
@claude review

gilcu3:
> .github/workflows/docker_build_node.yml:41, .github/workflows/docker_build_node_gcp.yml:41, .github/workflows/docker_build_rust_launcher.yml:35, .github/workflows/release.yml:101, .github/workflows/docker_node_release.yml:56, .github/workflows/docker_launcher_release.yml:47 — the host skopeo installs are removed, but six jobs still need a host skopeo, and none of them run on pull requests.

afaik skopeo is now part of the base image https://github.com/actions/runner-images/blob/20f9f7b2d2dbcf53e5c5a7e133f4867e8a555c24/images/ubuntu/scripts/build/install-container-tools.sh#L85-L88

> .github/workflows/ci.yml:54-57 (and the equivalent step in the three docker_build_* workflows) — the repro-env version pin is silently dropped.

this was a deliberate choice, given that the tool is part of ubuntu already. `repro-env` uses lock files to guarantee reproducibility, so a minor version change should not affect the result.

pbeza (approved):

barakeinav1 (approved):

barakeinav1:
@claude review

gilcu3:
> docs/running-an-mpc-node-in-tdx-external-guide.md:1852 — the worked example is pinned to git checkout 828f816be36aed6f0d2438e0131b3e9d7d0931ad (line 1848) with the expected digest sha256:331cfec... (line 1861), and the prerequisite list now says podman.

yeah I was aware of this, and changing the example is technically correct, but it should work either way, so leaving as is

PR #4137, opened 2026-08-13
feat(foreign-tx): svm inspector

frolvanya:
Adds the Solana JSON-RPC types and an `SvmInspector`, shared by Solana and Fogo behind chain-marker types
(`SolanaInspector` / `FogoInspector`, mirroring `EvmInspector<Client, Chain>`)
so an inspector built for one chain cannot be wired into the other's slot.

What it does per request:
  * fetches the transaction, and checks the returned signature is the one asked
    for;
  * checks finality, then that the transaction succeeded;
  * runs the requested extractors.

Details worth reviewing:
  * For `Finalized`, the rooted slot is read BEFORE the transaction. Reading it
    afterwards would let a block that got orphaned in between pass the check.
  * `meta.err` is a required field, not an `Option`: serde maps an absent field
    to `None`, which would make a provider that omits it indistinguishable from
    one reporting success. Absence is rejected; explicit `null` means success.
  * `innerInstructions: null` means the node was not started with extended
    transaction metadata, i.e. it cannot answer — not that the instruction is
    absent. Reported as a transient provider failure so the fan-out falls
    through to a provider that does record it, instead of signing an absence.
  * Hostile-provider bounds: instruction indices are `u8` (the wire format),
    inner-instruction data is capped at the runtime's 10 KiB CPI limit, the
    account list at the runtime's 255-meta limit (checked before indices are
    resolved into 32-byte pubkeys), and error messages never echo unbounded
    provider strings.

Extracted values are normalised so independent providers produce byte-identical
output — account indices are resolved against the static account keys plus the
addresses loaded from lookup tables (so v0 transactions work), and instruction
data is decoded from base58.

Tested with mocked providers, including a params-recording mock that pins the
commitment levels and encodings actually sent, plus manual (#[ignore]d) tests
that were run against live Solana mainnet.


frolvanya:
@claude review

frolvanya:
@claude review

crates/foreign-chain-health-check/src/probe.rs
```diff
@@ -143,7 +143,8 @@ pub async fn probe_all_providers(config: &ForeignChainsConfig) -> ProbeReport {
                     })
                     .await
                 }
-                // Solana and Ton have no inspector to probe them with.
+                // TODO(#4003): probe Solana and Fogo. Ton has no inspector, so there is
```
anodar: https://github.com/near/mpc/issues/4003 doesn't list Solana. Not sure if it's planned as part of that task.

cc: @haiyuechen-nearone

crates/foreign-chain-inspector/src/svm/inspector.rs
```diff
@@ -0,0 +1,528 @@
...
+
+        let commitment = match finalized_slot {
+            None => Commitment::Confirmed,
+            Some(finalized_slot) => {
+                if tx.slot > finalized_slot {
+                    return Err(ForeignChainInspectionError::NotFinalized);
+                }
+                Commitment::Finalized
+            }
+        };
```
anodar: Not very familiar with Solana but after reading a bit I believe I understand your trick here, that if transaction is Confirmed but its' slot <= latest processed slot, then you consider it Finalized based on Solana's guarantees. 

Few questions about this:

1. Is finalization check like that common practice ? I couldn't find anything like that on 
or elsewhere https://solana.com/developers/cookbook/transactions/confirmation
2. You implicitly set commitment to Finalized here even if tx.slot> finalized (but let's assume getTransaction(Finalized) would return null if you were to call it). So after this point if extractor is `AccountState` you still explicitly query with `Finalized` commitment instead of doing the same pattern you did here (that is querying with Confirmed and comparing to latest slot). My question: is there a reason you treat this two cases (getTransaction vs getAccountInfo) differently ? 
3. Have you considered alternatively doing this: query it with commitment (not hardcoded confirmed, but for finalized - finalized as well), if for finalized it returns null, you query it with Confirmed and if that exists then you throw retriable error: it should be mostly equivalent of your solution except for happy path it makes single RPC call vs 2 rpc calls. 
I say mostly equivalent because in a very unlikely case where `getSlot` query hits machine under load balancer that has more recent state than `getTransaction` right after that, your proposal will implicitly consider that Finalized (when tx.slot <= finalized_slot), but I'm not sure that's worth tradeoff over simplifying this a bit.

crates/foreign-chain-inspector/src/svm/inspector.rs
```diff
@@ -0,0 +1,528 @@
...
+                                .await?;
+                            accounts.insert(*pubkey, account.clone());
+                            account
+                        }
+                    };
+                    SvmExtractedValue::AccountState(account)
+                }
+            };
+            extracted_values.push(value);
+        }
```
anodar: nit: consider factoring this out into separate helper to make extract() bit more readable. There's quite a bit going on in this function that is conceptually simple but a bit hard to follow just by reading the code.

anodar (changes requested):
Overall looks good, but requested change as I want to clarify the part about finalization.

Also I only managed partial review, but fill finish it tomorrow.

crates/foreign-chain-inspector/src/svm/inspector.rs
```diff
@@ -0,0 +1,528 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum SvmExtractor {
+    InnerInstruction {
+        instruction_index: usize,
+        inner_instruction_index: usize,
+    },
+    AccountState {
+        pubkey: [u8; 32],
```
anodar: Why not use https://docs.rs/solana-pubkey/latest/solana_pubkey/struct.Pubkey.html here ?

crates/foreign-chain-inspector/src/svm/inspector.rs
```diff
@@ -0,0 +1,528 @@
...
+    /// Optimistically confirmed: voted by a supermajority, but not yet rooted.
+    Confirmed,
+    Finalized,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum SvmExtractor {
+    InnerInstruction {
+        instruction_index: usize,
+        inner_instruction_index: usize,
```
anodar, line 84: nit: Looks like we have inconsistency, sometimes we use u64 types for indexes in extractors (e.g.: `message_index`, `log_index`)  sometimes usize. 
Would be nice to converge to one type.

crates/foreign-chain-inspector/src/svm/inspector.rs
```diff
@@ -0,0 +1,528 @@
...
+        })?;
+        if !meta.err.is_null() {
+            return Err(ForeignChainInspectionError::TransactionFailed);
+        }
+
+        let account_keys = account_key_list(&tx.transaction.message.account_keys, meta);
+
+        // One read per distinct account: repeated pubkeys in one request would otherwise cost a
+        // round trip each and could observe different states within a single payload.
+        let mut accounts: BTreeMap<[u8; 32], SvmAccount> = BTreeMap::new();
```
anodar: nit: Let's use solana's pubkey type here as well.

crates/foreign-chain-inspector/src/svm/inspector.rs
```diff
@@ -0,0 +1,528 @@
...
+            .client
+            .request(GET_GENESIS_HASH_METHOD, NO_PARAMS)
+            .await
+            .map_err(ForeignChainInspectionError::classify_rpc_client_error)?;
+        Ok(Self::canonical_fingerprint(&genesis_hash))
+    }
+
+    /// A valid genesis hash is re-encoded through its 32 bytes, collapsing spelling
+    /// differences base58 itself cannot produce (surrounding whitespace); anything
+    /// unparseable is reported as answered.
```
anodar: > anything unparseable is reported as answered.

It's not very clear to me what you mean by this. I'm guessing you mean that we pass through whatever received if it fails to decode but wording can be improved.

crates/foreign-chain-inspector/src/svm/inspector.rs
```diff
@@ -0,0 +1,528 @@
...
+    fn canonical_fingerprint(fingerprint: &str) -> NetworkFingerprint {
+        match decode_base58_32(fingerprint.trim()) {
+            Ok(bytes) => NetworkFingerprint::new(bs58::encode(bytes).into_string()),
+            Err(_) => NetworkFingerprint::new(fingerprint),
```
anodar: Is the error expected here? If not we should log it.

also optional nit: (possibly biased) This could be simplified as something like:

```
let mut f: String = fingerprint.into();
if let Ok(bytes) = decode_base58_32(fingerprint.trim()) {
   f = bs58::encode(bytes).into_string();
}
NetworkFingerprint::new(f)
```

crates/foreign-chain-inspector/src/svm/inspector.rs
```diff
@@ -0,0 +1,528 @@
...
+        }
+    }
+}
+
+/// Rejects a backend that returned a different transaction than queried. The transaction
+/// id is the first signature; a missing or undecodable one is a malformed response, a
+/// well-formed but different one is a hard inconsistency.
+fn ensure_signature_matches(
+    requested: &[u8; 64],
+    returned: Option<&String>,
```
anodar: nit: consider using &str over &String so that this doesn't force caller to construct string from &str (e.g. in should_reject_non_base58_signature_as_malformed), but would should work for both (&String -> &str coercion)

crates/foreign-chain-inspector/src/svm/inspector.rs
```diff
@@ -0,0 +1,528 @@
...
+        .collect()
+}
+
+fn extract_inner_instruction(
+    meta: &TransactionMeta,
+    account_keys: &[&str],
+    instruction_index: usize,
+    inner_instruction_index: usize,
+) -> Result<SvmExtractedValue, ForeignChainInspectionError> {
+    // An absent list means the node does not record CPI metadata — a provider gap, kept
```
anodar: I think these chain specific acronyms will be hard to follow as we expand support for more and more chains. Please consider writing it in full: "Cross-Program Invocation (CPI)".

crates/foreign-chain-health-check/src/probe.rs
frolvanya: Probing is implemented in #4138, this comment was just a template that was used previously

crates/foreign-chain-inspector/src/svm/inspector.rs
frolvanya: Replaced it everywhere across

crates/foreign-chain-inspector/src/svm/inspector.rs
frolvanya: Agree, changed to `&str`

crates/foreign-chain-inspector/src/svm/inspector.rs
frolvanya: Not quite, refactored this method a little bit

crates/foreign-chain-inspector/src/svm/inspector.rs
frolvanya: Removed this wording entirely and moved refactored version in the next method

crates/foreign-chain-inspector/src/svm/inspector.rs
frolvanya: We already have `SvmAddress`, used it instead

crates/foreign-chain-inspector/src/svm/inspector.rs
frolvanya: Used `SvmAddress`

crates/foreign-chain-inspector/src/svm/inspector.rs
frolvanya: I left `usize` because most inspectors are using it

crates/foreign-chain-inspector/src/svm/inspector.rs
frolvanya: Refactored across multiple methods

crates/foreign-chain-inspector/src/svm/inspector.rs
frolvanya: 1) No, not common for SVM, but I believe it's common for EVM
2) Somehow I already forget my reasoning 2 weeks ago, right now I agree that this asymmetry looks strange, maybe I just missed it during some refactors
3) I didn't consider, but I like your idea better and we can even remove `getSlot` completely since it guards 

Refactored this

crates/foreign-chain-health-check/src/probe.rs
frolvanya: This TODO is now gone entirely after merging with main branch

frolvanya:
@claude review

anodar (approved):

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -70,6 +71,12 @@ pub enum Verdict<V> {
     },
     #[display("a requested log index is out of bounds")]
     LogIndexOutOfBounds,
+    /// A verdict rather than a tolerated error, for the same reason as
+    /// [`Self::TransactionNotFound`]: an absent account is the chain's answer, not a provider
+    /// fault. Only as settled as the account read behind it — an extractor that reads mutable
+    /// state can have honest providers disagree.
```
haiyuechen-nearone: nit: 

```suggestion
    /// Similar to [`Self::TransactionNotFound`], an absent account is the chain's answer, not a provider fault.
    /// A mutable account may leave two honest providers with
    /// different answers.
```

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -429,7 +436,10 @@ impl ForeignChainInspectionError {
 }
 
 /// Some providers report throttling as a JSON-RPC error object over HTTP 200 rather than a 429, and
-/// it is the one refusal worth retrying. Alchemy and Infura send `-32005`, others `-32029`.
+/// it is the one refusal worth retrying. Alchemy and Infura send `-32005`, others `-32029`. SVM
+/// nodes reuse `-32005` for `NodeUnhealthy`, which is a different condition but equally worth a
+/// retry; codes that mean something else per chain are intercepted by that chain's inspector before
+/// they reach here.
```
haiyuechen-nearone, line 442: Nothing to fix here, but I see that the pattern I introduced earlier is not quite right, raised https://github.com/near/mpc/issues/4420 to address this later (most likely by me)

crates/foreign-chain-inspector/src/lib.rs
haiyuechen-nearone, line 77: nit: 

```suggestion
    /// Similar to [`Self::TransactionNotFound`], an absent account is the chain's answer, not a provider fault.
    /// A mutable account may leave two honest providers with
    /// different answers.
```

haiyuechen-nearone (approved):
LGTM 🚢

PR #4138, opened 2026-08-13
feat(foreign-tx): wire svm chains into node

frolvanya:
Connects config -> inspector -> signing, so a node actually serves Solana and
Fogo verification requests. Before this PR the node still rejects them.

  * node config: `fogo` section added (`solana` already existed);
  * verify_foreign_tx: one typed inspector fan-out per chain
    (`SolanaInspector` / `FogoInspector`, so a slot mix-up is a compile error)
    and dispatch arms that replace the old
    `bail!("ForeignChainRpcRequest::Solana is unsupported")`; tests pin that
    each chain's config feeds its own slot;
  * health check: golden check + network-fingerprint probe for both chains;
  * operator docs, localnet templates and user-config gain both chains.

Every genesis hash committed here was read from a live node:
  solana mainnet  5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d
  solana devnet   EtWTRABZaYq6iMfeYKouRu166VU2xqa1wcaWoxPkrZBG
  fogo   mainnet  CDLtwKnaCoK157uaHQDj4fHu72AyD2519Cphmpiq6hvT
  fogo   testnet  9GGSFo95raqzZxWqKM5tGYvJp5iv4Dm565S4r8h5PEu9

Verified with the config tester against live endpoints (mainnet and testnet both
pass), including a negative check: pointing the Solana entry at Fogo's URL fails
with "chain id mismatch", so the check is not passing vacuously.


frolvanya:
@claude review

frolvanya:
@claude review

haiyuechen-nearone (approved):
LGTM 🚢

crates/foreign-chain-health-check/src/golden.rs
```diff
@@ -30,6 +30,15 @@ pub struct SuiVector {
     pub chain_id: &'static str,
 }
 
+/// Like Sui, SVM chains are verified by chain identity rather than a pinned reference
+/// transaction, since providers prune historical transactions — see
+/// [`check_svm`](crate::checks::check_svm).
+#[derive(Clone, Copy)]
+pub struct SvmVector {
+    /// Base58 of the 32-byte genesis hash, exactly as `getGenesisHash` returns it.
+    pub genesis_hash: &'static str,
```
anodar, line 39: Not introduced in this PR, but I just noticed we use &str instead of proper hash types in structs throughout this file.

anodar (approved):

PR #4143, opened 2026-08-13
refactor(test-utils): read the collateral straight from the captured `public_data.json`

### LINKED ISSUE #4140: Derive the attestation fixtures from the captured `public_data.json` instead of reshaping them
## Background

`create-assets.sh` post-processes the collateral with `jq`: it hex-encodes four byte fields and strips a NUL terminator. That is only needed because the fixtures predate a type change — `attestation::Collateral` used to deserialize via `#[serde(try_from = "Value")]` through a hex parser, so a hex-shaped file was the only thing that would load. `Collateral` and `QuoteBytes` are now plain derived-serde types over `Vec<u8>`, so the endpoint's own shape deserializes directly and the reshaping is obsolete.

Three consumers still expect hex: `collateral_from_str` and the `HexVec` DTO parse in `crates/test-utils/src/attestation.rs`, and the parser tests in `crates/attestation/tests/collateral.rs`, which read the shared fixture rather than their own input.

## Acceptance Criteria

- `create-assets.sh` extracts the collateral without transforming it.
- The fixture-consuming tests pass with the collateral in the shape `/public_data` returns.

- <https://github.com/near/mpc/pull/4109#discussion_r3772848077>

pbeza:
Follow-up on @gilcu3's code review comment: https://github.com/near/mpc/pull/4109#discussion_r3772848077.

Closes #4140


pbeza:
@claude review

gilcu3 (approved):
LGTM, nice simplification!

pbeza:
@barakeinav1 could you review this when you’re back on Sunday (unless someone beats you to it)? Thanks!

kevindeforth (approved):
Thanks!

PR #4100, opened 2026-08-07
feat(scripts): add dev-cluster node upgrade tooling

SimonRastikian:
Deals partly with #3934 namely the nodes upgrade on the dev cluster

Issue #3934, opened 2026-07-23
Automate MPC ops/release workflows (release-prep → Slack → dev nodes → propose/vote → prod nodes)


SimonRastikian:
### Background

 Our release/ops runbooks are entirely manual: hand-typed borsh serialization, copy-pasted announcement templates, per-account votes, and scattered verify commands across four docs. This is toil-heavy and error-prone under release
  pressure. This issue tracks incrementally automating it, sequenced by blast radius + secret-handling + determinism so the safest, highest-value pieces land first.

Guiding principles
  - Deterministic, no-secret, read-only work first; live-fleet mutations last.
  - Each phase produces reusable pieces the next phase depends on.
  - Outward-facing / irreversible actions (Slack posts, on-chain txns, mainnet fleet) stay opt-in and human-gated.

The tooling live here in scripts/ and the credentials for Nomad/Slack/GH  will be manually provided.

The interface will look like a menu.sh which will provide the possibility to run any of the steps, and perhaps proceed with one that has already been started.

### Acceptance Criteria

Phase 0 — Release-prep helper (scripts/mpc-release.sh)
Deterministic, no secrets, nothing irreversible. Touched by every release path.

  - [ ] prepare <VERSION>: gh release download contract WASM → build serialized.bin (borsh \x01 + u32 len + wasm + \x00) → print sha256. Replaces the Python heredoc in how-to/contract-upgrade.md.
  - [ ] skopeo inspect nearone/mpc-node:<VERSION> → print both bare-hex (for vote_code_hash) and sha256:-prefixed (for mpc_hash_override) digests.
  - [ ] Emit filled-in propose_update / vote_update / vote_code_hash commands with <contract>/<net> substituted.
  - [ ] status <net> (read-only): query version, proposed_updates, code_hash_votes, allowed_docker_image_hashes, state via near-cli + curl :8080/metrics | grep mpc_node_build_info per node. One-glance "where does this upgrade stand?".

Prereqs: gh, skopeo, near CLI on PATH.

**Phase 1 — Slack announcement generation + posting**

Templates in how-to/contract-upgrade.md and how-to/production-node-upgrade.md are placeholder substitution.

  - [ ] 1a: announce contract|node <VERSION> <net> … fills the template and prints it, reusing Phase 0's computed UpdateId/hash for consistency.
  - [ ] 1b: --post flag posts to the operators channel. Slack bot token is supplied manually (SLACK_TOKEN env var or prompt); generate-only is the default, posting is opt-in.

[...]

PR #4146, opened 2026-08-14
refactor(scripts): extract release helpers into common.sh

SimonRastikian:
This is a first cleanup for #3934 it's a split from https://github.com/near/mpc/pull/4100.

It basically moves generic functions into a file common.rs to be used in the next PRs that build the scripts.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -68,6 +68,9 @@ git checkout -b release-prep/v3.11.1
 ./scripts/ops/prepare-github-release.sh 3.11.1
 ```
 
+Pre-release versions carrying a suffix (`3.11.0-rc.1`) are accepted too — the
+[Release workflow](.github/workflows/release.yml) validates the same shape.
+
 Push the working branch and open a PR against `main` (for minor releases)
 or `release/vX.Y` (for patches). Once the PR is reviewed and merged, the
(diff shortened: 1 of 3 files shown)


pbeza:
I'm a bit hesitant to accept it if this PR doesn't actually reduce the LoC, only follow-ups do.

bash

Also, it feels like a first step toward making the Bash scripts more complex by introducing a shared “library,” which I’d prefer to avoid. We should reduce our use of Bash rather than rely on it more.

SimonRastikian:
@pbeza please check the discussion https://nearone.slack.com/archives/C0912BTG51T/p1786635232478689

SimonRastikian:
@pbeza if you can find me a tangible solution/design I would be happier

pbeza:
> @pbeza if you can find me a tangible solution I would be happier

When it comes to this PR: as I mentioned, I’d be happy to accept a common Bash library if it reduced LoC in the repo. But it seems like this is mostly preparation for something bigger and Bash-heavy, which might not be accepted by the team given our preference to avoid large Bash tooling.

I think this deserves a broader async discussion, so we can continue in the Slack thread you posted. I see there’s already some discussion going on there.

Issue #1962, opened 2026-02-04
refactor: Break out feature-specific contract methods and tests to their own modules


netrome:
### Background

Our contract `lib.rs` file is getting pretty huge. We're over-due for a refactor of this. I'd propose something along the lines of isolating separate features to separate modules (e.g. `sign`, `ckd`, `foreign_chain`, `update`, etc.), or something similar.

### User Story

As a developer I want to keep related logic close in isolated modules.

### Acceptance Criteria

We've broken out contract methods and tests to separate smaller modules.

netrome:
Note: Make sure to coordinate this refactor so it doesn't conflict with any significant work in the contract.

netrome:
This could help us conceptually separate things in the contract better, to pave the way for a future multi-contract architecture.

netrome:
Not a priority for next week though.

gilcu3:
### Plan

`crates/contract/src/lib.rs` is being split into one module per contract feature under a new `src/api/` tree. Each module holds that feature's `#[near] impl MpcContract` block plus its tests. The state types those entrypoints drive stay where they are (`update.rs`, `tee/`, `node_migrations.rs`, `foreign_chains_metadata.rs`). `lib.rs` ends up at 96 lines: the `MpcContract` state struct, its `Default`, the module declarations, and the borsh-schema snapshot test.

First done as a single PR (#4144, +9422/-8878), now split into a stack of 12, each chained onto the previous:

| Branch | Contents | Diff |
|---|---|---|
| `1962-01-scaffolding` | `api.rs` + `api/test_utils.rs` (shared test fixtures) | +315/-282 |
| `1962-02-common` | `api/common.rs` | +351/-324 |
| `1962-03-keys-sign` | `api/keys.rs`, `api/sign.rs` | +721/-666 |
| `1962-04-ckd` | `api/ckd.rs` | +533/-487 |
| `1962-05-lifecycle` | `api/lifecycle.rs` | +279/-226 |
| `1962-06-governance-key-events` | `api/governance.rs`, `api/key_events.rs` | +770/-688 |
| `1962-07-update` | `api/update.rs` | +716/-661 |
| `1962-08-node-migration` | `api/node_migration.rs` | +942/-890 |
| `1962-09-tee-verifier` | `api/tee_verifier.rs` | +270/-234 |
| `1962-10-tee-measurements` | `api/tee_measurements.rs` | +1118/-1083 |
[...]

PR #4148, opened 2026-08-14
refactor(contract): move shared request and auth plumbing into api/common.rs

gilcu3:
Part 2/12 of #1962

### DIFF
--- a/crates/contract/src/api/common.rs
+++ b/crates/contract/src/api/common.rs
@@ -1,11 +1,15 @@
 //! Plumbing shared by more than one feature
 
-use crate::errors::{Error, InvalidParameters, RequestError, TeeError};
+use crate::errors::{InvalidParameters, RequestError, TeeError};
 use crate::{MpcContract, MpcContractExt};
 use near_mpc_contract_interface::types as dtos;
 use near_sdk::{AccountId, CryptoHash, Gas, GasWeight, NearToken, Promise, env, log, near};
 
 use dtos::{DomainConfig, DomainId, DomainPurpose};
+
+/// Register used to receive data id from `promise_await_data`.
+/// Note: This is an implementation constant, not a configurable policy value.
+const DATA_ID_REGISTER: u64 = 0;
 
 impl MpcContract {
     /// Common preconditions enforced on every user-facing request method (`sign`,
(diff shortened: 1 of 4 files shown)


pbeza (dismissed):
I won't lie, Claude helped me confirm this is a mechanical refactor. Only these changes aren't mechanical (caught by Claude), but they look good:

> - Visibility widened from private to `pub(crate)` on the moved items, which the move requires and which changes nothing externally
> - Doc-comment-only edits: intra-doc links rewritten to full paths (`crate::state::ProtocolContractState::...`) so they resolve from the new module, and one added clause on `MINIMUM_SIGN_REQUEST_DEPOSIT` ("also charged for foreign-transaction verification")
> - Import shuffling in `lib.rs` (drops now-unused `GasWeight` and `RequestError`, imports the moved items from `api::common`) and the new file's own imports, module doc line, impl-block/test-module scaffolding, and closing braces

crates/contract/src/api/common.rs
```diff
@@ -0,0 +1,347 @@
...
+        env::panic_str(&RequestError::Timeout.to_string());
+    }
+}
+
+/// Register used to receive data id from `promise_await_data`.
+/// Note: This is an implementation constant, not a configurable policy value.
+const DATA_ID_REGISTER: u64 = 0;
+
+/// Minimum deposit required for sign requests, also charged for foreign-transaction verification
+pub(crate) const MINIMUM_SIGN_REQUEST_DEPOSIT: NearToken = NearToken::from_yoctonear(1);
```
kevindeforth: Lets move this to the contract-interface crate.

crates/contract/src/api/common.rs
```diff
@@ -0,0 +1,347 @@
...
+    #[private]
+    pub fn fail_on_timeout() {
+        // To stay consistent with the old version of the timeout error
+        env::panic_str(&RequestError::Timeout.to_string());
+    }
+}
+
+/// Register used to receive data id from `promise_await_data`.
+/// Note: This is an implementation constant, not a configurable policy value.
+const DATA_ID_REGISTER: u64 = 0;
```
kevindeforth: I generally prefer when we define constant at the top of the file.
Also, I would suggest to move this to a separate file together with `enqueue_yield_request`.

crates/contract/src/api/common.rs
```diff
@@ -0,0 +1,347 @@
...
+
+        assert_eq!(
+            signer_id, predecessor_id,
+            "Caller must be the signer account (signer: {}, predecessor: {})",
+            signer_id, predecessor_id
+        );
+
+        signer_id
+    }
+}
```
kevindeforth: Have you considered moving these into a separate file concerned solely with authorization? Something like`auth.rs` or `authorization_helpers.rs` or `access_control.rs`?

crates/contract/src/api/common.rs
```diff
@@ -0,0 +1,347 @@
...
+
+/// Transfers `amount` to `account_id` via a detached promise; no-op when zero.
+pub(crate) fn refund_to(account_id: &AccountId, amount: NearToken) {
+    if amount > NearToken::from_near(0) {
+        log!("refund {amount} to {account_id}");
+        Promise::new(account_id.clone()).transfer(amount).detach();
+    }
+}
```
kevindeforth, line 153: Hmm, I think it wouldn't be bad to have maybe one file where we track all the promises that this contract spawns (e.g. `promises.rs`).

This method and `enqueue_yield_request` kind of belong together in that sense.

kevindeforth (commented):
I would suggest to split this into even more files.

kevindeforth (dismissed):
I pre-approved, because it's an improvement and I don't want to block, but I would still suggest to consider splitting this further.

crates/contract/src/api/common.rs
gilcu3: Unfortunately we have many more methods using promises than this two, so it would not be very uniform to have them being located there. Leaving here for now, we could think later if we want to do some promise unification

kevindeforth (approved):
Thanks!

pbeza (approved):

PR #4155, opened 2026-08-17
refactor(contract): move public-key views and signing into api/keys.rs and api/sign.rs

gilcu3:
Part 3/12 of #1962

### DIFF
--- a/crates/contract/src/api/keys.rs
+++ b/crates/contract/src/api/keys.rs
@@ -1,0 +1,89 @@
+//! Public-key views: the domain public key, per-account derived keys, and the
+//! latest key version.
+
+use crate::crypto_shared::{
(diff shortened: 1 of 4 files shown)


crates/contract/src/api/keys.rs
```diff
@@ -0,0 +1,89 @@
+//! Public-key views: the domain public key, per-account derived keys, and the
+//! latest key version.
+
```
kevindeforth, line 3: not too fond of this comment.

kevindeforth (approved):
lgtm

crates/contract/src/api/keys.rs
gilcu3: yeah I am not too fond either, but `keys.rs` does not really tell me what the file contains, so the comment seemed helpful

anodar (approved):

PR #4157, opened 2026-08-17
fix: Surfacing failure by replacing join_all with try_join

SimonRastikian:
crates/node/src/providers/ecdsa.rs: has `spawn_background_tasks` which uses `join_all` to generate triples.

With multiple threshold-specific triple generators, if one task dies but others keep running (normal infinite loops), `join_all` never returns, so the failure is never surfaced.

This PR addresses this issue by changing to JoinSet/select_all so any unexpected task exit is observed immediately.
It also reworks the `bail!` which output is discarded one level up in mpc_client.rs:239 where the `anyhow::Result<()>` carrying the failing message is discarded using `let _ = [...]`.

### DIFF
--- a/crates/node/src/mpc_client.rs
+++ b/crates/node/src/mpc_client.rs
@@ -156,12 +156,13 @@ where
         debug_receiver: tokio::sync::broadcast::Receiver<DebugRequest>,
     ) -> anyhow::Result<()> {
         let client = self.client.clone();
-        let metrics_emitter = tracking::spawn("periodically emits metrics", async move {
-            loop {
-                client.emit_metrics();
-                tokio::time::sleep(std::time::Duration::from_secs(5)).await;
-            }
-        });
+        let metrics_emitter: AutoAbortTask<anyhow::Result<()>> =
+            tracking::spawn("periodically emits metrics", async move {
+                loop {
+                    client.emit_metrics();
+                    tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+                }
+            });
 
         let monitor_passive_channels = {
             tracking::spawn(
@@ -234,7 +235,7 @@ where
         // `try_join!` surfaces the failure and aborts the rest on drop.
         futures::try_join!(
             async move { monitor_passive_channels.await? },
-            async move { metrics_emitter.await.map_err(anyhow::Error::from) },
+            async move { metrics_emitter.await? },
             async move { monitor_chain.await? },
             async move { robust_ecdsa_background_tasks.await? },
             async move { ecdsa_background_tasks.await? },
--- a/crates/node/src/mpc_client.rs
+++ b/crates/node/src/mpc_client.rs
@@ -235,8 +235,7 @@ where
             self.ckd_provider.clone().spawn_background_tasks(),
         );
 
-        // Awaiting in sequence would park a failure behind the tasks that never
-        // return; `try_join!` surfaces it and aborts the rest on drop.
+        // `try_join!` surfaces the failure and aborts the rest on drop.
         futures::try_join!(
             async move { monitor_passive_channels.await? },
             async move { metrics_emitter.await.map_err(anyhow::Error::from) },
--- a/crates/node/src/mpc_client.rs
+++ b/crates/node/src/mpc_client.rs
@@ -19,7 +19,7 @@ use crate::requests::queue::{
 use crate::storage::{
     CKDRequestStorage, SignRequestStorage, VerifyForeignTransactionRequestStorage,
 };
-use crate::tracking::{self, AutoAbortTaskCollection};
+use crate::tracking::{self, AutoAbortTask, AutoAbortTaskCollection};
 use crate::trait_extensions::convert_to_contract_dto::IntoContractInterfaceType;
 use crate::types::SignatureRequest;
 use crate::types::{CKDRequest, RequestsUpdate, VerifyForeignTxRequest};
@@ -182,7 +182,7 @@ where
             )
         };
 
-        let tee_verification_handle = {
+        let tee_verification_handle: AutoAbortTask<anyhow::Result<()>> = {
             let chain_txn_sender = chain_txn_sender.clone();
             tracking::spawn("tee_verification", async move {
                 loop {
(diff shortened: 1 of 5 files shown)


crates/node/src/mpc_client.rs
```diff
@@ -235,14 +235,18 @@ where
...
         );
 
-        let _ = monitor_passive_channels.await?;
-        metrics_emitter.await?;
-        monitor_chain.await?;
-        let _ = robust_ecdsa_background_tasks.await?;
-        let _ = ecdsa_background_tasks.await?;
-        let _ = eddsa_background_tasks.await?;
-        let _ = ckd_background_tasks.await?;
-        tee_verification_handle.await?;
```
SimonRastikian, line 245: Awaiting in sequence would park a failure behind the tasks that never return

crates/node/src/providers/robust_ecdsa.rs
```diff
@@ -201,13 +203,16 @@ impl SignatureProvider for RobustEcdsaSignatureProvider {
             })
             .collect::<Vec<_>>();
 
-        for Err(join_error) in futures::future::join_all(generate_presignatures).await {
-            tracing::error!(
-                "Damgard et al background presignature task ended unexpectedly: {join_error}"
-            );
+        // Generators never return, so any exit is a failure.
+        if generate_presignatures.is_empty() {
+            return Ok(());
```
SimonRastikian: `select_all` (down) surfaces the first failure instead of `join_all` blocking on the siblings' infinite loops.

crates/node/src/providers/ecdsa.rs
```diff
@@ -252,54 +252,52 @@ impl SignatureProvider for EcdsaSignatureProvider {
...
                     self.client.clone(),
                     self.mpc_config.clone(),
                     self.config.triple.clone().into(),
                     triple_store.clone(),
                     reconstruction_threshold_bound,
                 ),
-            ));
+            );
+            background_tasks.push((description, task));
         }
```
SimonRastikian: That's just a clearer rewriting of the code

crates/node/src/providers/ecdsa.rs
```diff
@@ -252,54 +252,52 @@ impl SignatureProvider for EcdsaSignatureProvider {
...
-        }
-
-        for Err(join_error) in futures::future::join_all(generate_triples).await {
-            tracing::error!(
-                "ecdsa background triple generation task ended unexpectedly: {join_error}"
             );
-        }
-        for Err(join_error) in futures::future::join_all(generate_presignatures).await {
-            tracing::error!("ecdsa background presignature task ended unexpectedly: {join_error}");
+            background_tasks.push((description, task));
```
SimonRastikian: Same here, used background_tasks as the same underlying vector.

SimonRastikian:
@claude review

SimonRastikian:
@claude  review

crates/node/src/providers/ecdsa.rs
```diff
@@ -291,15 +288,11 @@ impl SignatureProvider for EcdsaSignatureProvider {
...
-                "ecdsa background triple generation task ended unexpectedly: {join_error}"
-            );
-        }
-        for Err(join_error) in futures::future::join_all(generate_presignatures).await {
-            tracing::error!("ecdsa background presignature task ended unexpectedly: {join_error}");
-        }
-
-        Ok(())
+        let Some((description, outcome)) = tracking::first_task_exit(background_tasks).await else {
+            return Ok(());
```
haiyuechen-nearone: Im a bit confused. If we reach this branch then we didn't start any background tasks. Shouldn't we signify that somehow?

crates/node/src/providers/robust_ecdsa.rs
```diff
@@ -201,13 +201,14 @@ impl SignatureProvider for RobustEcdsaSignatureProvider {
...
 
-        for Err(join_error) in futures::future::join_all(generate_presignatures).await {
-            tracing::error!(
-                "Damgard et al background presignature task ended unexpectedly: {join_error}"
-            );
-        }
-
-        Ok(())
+        let Some((description, outcome)) = tracking::first_task_exit(background_tasks).await else {
+            return Ok(());
```
haiyuechen-nearone: Same as comment in `ecdsa.rs`. If we reach here then there is no generator for any domain. Isn't that a problem?

crates/node/src/providers/ecdsa.rs
SimonRastikian: This is not important because it is never reached. In this case `background_tasks` always has at least one entry. We cannot hit None.
I will replace this by bail! but without any consequence. (BTW I did not `expect`  or `unreachable` because the repro is not to panic)

SimonRastikian:
@haiyuechen-nearone  both of your comments are fair for the confusion but both of the branches are unreachable:
In those case background_tasks always has at least one entry. We cannot hit None.

I still thought of a fix by replacing OK with bail  (not with expect or unreachable because the repro is not to panic). However I thought that the real fix is not to have `first_task_exit` output Option in the first place. So here I reworked it [50b511a](https://github.com/near/mpc/pull/4157/commits/50b511a1dd061af889ee099bc98e1bb490c68636)

haiyuechen-nearone (approved):
🚢

pbeza (approved):

pbeza (approved):
Nit: I think the PR title is outdated now, since there’s no `select_all` anymore.

crates/node/src/tracking.rs
```diff
@@ -90,6 +91,31 @@ where
...
+pub async fn first_task_exit<R>(
+    tasks: Vec<(String, AutoAbortTask<R>)>,
+) -> (String, Result<R, JoinError>) {
+    let mut running = tasks
+        .into_iter()
+        .map(|(description, task)| async move { (description, task.await) })
+        .collect::<FuturesUnordered<_>>();
+    match running.next().await {
+        Some(exit) => exit,
+        None => std::future::pending().await,
```
kevindeforth: Did you consider returning `None` here (i.e.`Option<(String, Result<R, JoinError>)>`) and have the caller decide if they need to block?

Blocking indefinitely because the caller passed an empty vector seems like an odd choice.

crates/node/src/tracking.rs
```diff
@@ -90,6 +91,31 @@ where
...
+    }
+}
```
kevindeforth: nit: these methods seem pretty custom. Placing them inside the `tracking` framework makes it somewhat more difficult for a user to figure out which methods are expected to use.

PR #4159, opened 2026-08-17
refactor(devnet): hand out access keys as shared handles

### LINKED ISSUE #4158: refactor(devnet): remove `OwnedMutexGuard` access key footgun
### Background

Devnet stores the operators key inside `OperatingAccessKey`, together with the nonce.
Those `OperatingAccessKey`'s are stored behind a mutex inside `OperatingAccount`, ensuring the nonce is valid at the time the key is used.

Currently, `OperatingAccount` returns an `OwnedMutexGuard` instead of a `Mutex`. The issue is that from a callers perspective, it's not obvious that  calling `any_access_key()` acquires a lock.

This works with the current, relatively simple devnet setup, but may lead to issues in case we expand devnet to handle more complicated tasks, as it could result in a potential deadlock.

https://github.com/near/mpc/blob/198fe4beb5cbb7aef50f9e74c0b7c9edffec369d/crates/devnet/src/account.rs#L475-L483
OperatingAccount::any_access_key`/`all_access_keys` ([`crates/devnet/src/account.rs:475`](crates/devnet/src/account.rs#L475))

https://github.com/near/mpc/blob/198fe4beb5cbb7aef50f9e74c0b7c9edffec369d/crates/devnet/src/account.rs#L88-L94

### User Story

As a dev, I like the absence of footguns.

### Acceptance Criteria

rename `any_access_key` / `all_access_keys` to `any_access_key_handle` / `all_access_key_handles`and have them return an `Arc<Mutex<OperatingAccessKey>>` instead of an owned mutex guard.

_No response_

kevindeforth:
resolves https://github.com/near/mpc/issues/4158

### DIFF
--- a/crates/devnet/src/account.rs
+++ b/crates/devnet/src/account.rs
@@ -41,7 +41,7 @@ use rand::rngs::OsRng;
 use reqwest::StatusCode;
 use std::collections::HashMap;
 use std::sync::Arc;
-use tokio::sync::OwnedMutexGuard;
+use tokio::sync::Mutex;
 
 /// Current state of an account while the CLI is running.
 pub struct OperatingAccount {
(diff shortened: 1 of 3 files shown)


gilcu3 (approved):
Thank you!

anodar (approved):

PR #4161, opened 2026-08-17
refactor(contract): move init, migrate and top-level views into api/lifecycle.rs

gilcu3:
part 5/12 of #1962

### DIFF
--- a/crates/contract/src/api/lifecycle.rs
+++ b/crates/contract/src/api/lifecycle.rs
@@ -1,0 +1,271 @@
+//! Contract lifecycle: initialization, state migration, and the top-level
+//! state and config views.
+
+use crate::config::Config;
+use crate::errors::{DomainError, InvalidState};
+use crate::{
+    dto_mapping::{IntoInterfaceType, TryIntoContractType},
+    errors::Error,
+    foreign_chains_metadata::ForeignChainsMetadata,
+    primitives::domain::AddDomainsVotes,
+    storage_keys::StorageKey,
+    tee::tee_state::TeeState,
+    tee::verifier_votes::TeeVerifierVotes,
+    update::ProposedUpdates,
+};
+use near_mpc_contract_interface::types::{self as dtos};
+
+use crate::node_migrations::NodeMigrations;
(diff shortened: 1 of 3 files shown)


haiyuechen-nearone (approved):

crates/contract/src/api/lifecycle.rs
```diff
@@ -0,0 +1,271 @@
...
+#[cfg(test)]
+#[expect(non_snake_case)]
+mod tests {
+    use super::*;
+
+    use crate::primitives::thresholds::{GovernanceThreshold, GovernanceThresholdParameters};
+
+    use crate::primitives::test_utils::gen_participants;
+
+    use crate::*;
```
SimonRastikian, line 236: OCD: No need for the spaces

crates/contract/src/api/lifecycle.rs
```diff
@@ -0,0 +1,271 @@
...
+            node_migrations: NodeMigrations::default(),
+            node_foreign_chain_support: Default::default(),
+            foreign_chains: Lazy::new(
+                StorageKey::ForeignChainMetadata,
+                ForeignChainsMetadata::default(),
+            ),
+            tee_verifier_account_id: None,
+            tee_verifier_votes: TeeVerifierVotes::default(),
+            available_attestation_grants: IterableMap::new(StorageKey::AttestationGrants),
+        })
```
SimonRastikian, line 176: Independently of this PR purpose but this is a huge repetition with lines 65 on

SimonRastikian (approved):

crates/contract/src/api/lifecycle.rs
gilcu3: Will handle this one as part of the next refactor PR (there are still quite a few left)

crates/contract/src/api/lifecycle.rs
gilcu3: This being a constructor it doesn't feel right to modify it in this pure refactor PRs, we can certainly do it later

PR #4165, opened 2026-08-17
test(contract): sandbox tests for the attestation flow against the real tee-verifier

### LINKED ISSUE #4084: Sandbox tests for the attestation flow against the real `tee-verifier` contract

pbeza:
Closes #4084


pbeza:
@claude review

docs/localnet/localnet.md
gilcu3: why do we need this mention of tests here in the localnet guide? I know this is in part existing, but seems a bit out of place. We already have a README somewhere else for updating the fixtures

pbeza:
@gilcu3 @anodar @haiyuechen-nearone @kevindeforth this PR is ready for review. Any subsequent commits will only merge `origin/main` to keep it up to date with `main`. Would appreciate your reviews. Thanks!

docs/deploy-tee-verifier.md
```diff
@@ -102,11 +102,24 @@ its full-access key:
 near account list-keys "$VERIFIER_ACCOUNT" network-config "$NETWORK" now
 ```
 
+The verifier holds no state of its own, so its storage must be empty. Deploying does not
+clear storage, so anything written before this point survives every later deploy and
+outlives the lock:
+
+```shell
+near contract view-storage "$VERIFIER_ACCOUNT" all as-text network-config "$NETWORK" now
+```
```
gilcu3, line 111: would having some storage lead to some attack? I can see how this is related to the change in this PR, but it seems that the code itself should be malicious, independent of the storage

crates/e2e-tests/README.md
```diff
@@ -220,7 +224,7 @@ pub struct MpcClusterConfig {
 
 pub struct ForeignChainsClusterConfig {
     pub node_configs: Vec<ForeignChainsConfig>, // per-node; empty = default for all
-    pub whitelisted_chains: BTreeSet<ForeignChain>, // voted in during setup
+    pub whitelist: BTreeMap<ForeignChain, ChainEntry>, // voted in during setup
```
gilcu3, line 240: why was this modified? seems very orthogonal to the current issue

crates/e2e-tests/src/cluster.rs
```diff
@@ -226,6 +227,15 @@ impl MpcClusterConfig {
...
         }
+        // Startup indexes the key vectors by participant index, so an out-of-range
+        // entry here would otherwise surface as a panic mid-startup.
+        for (i, &participant_idx) in self.initial_participant_indices.iter().enumerate() {
+            anyhow::ensure!(
+                participant_idx < self.num_nodes,
+                "initial_participant_indices[{i}]: index {participant_idx} must be < num_nodes ({})",
+                self.num_nodes,
+            );
+        }
```
gilcu3: but this is test code, so panicking is fine

crates/e2e-tests/src/cluster.rs
```diff
@@ -234,6 +244,15 @@ fn default_mpc_binary_path() -> PathBuf {
...
 }
 
+fn must_load_tee_verifier_wasm() -> Vec<u8> {
+    test_utils::contract_build::must_load_wasm(
+        "MPC_TEE_VERIFIER_WASM",
+        "target/near/tee_verifier/tee_verifier.wasm",
+        test_utils::contract_build::ContractBuilder::new("crates/tee-verifier/Cargo.toml")
+            .out_dir("target/near/tee_verifier"),
+    )
+}
```
gilcu3: shouldn't this function live in common, along with the other `must_load_...`?

crates/e2e-tests/src/cluster.rs
```diff
@@ -1356,6 +1378,48 @@ async fn init_contract(
     .context("contract did not reach Running state after init")
 }
 
+/// Deploys the tee-verifier and votes it in from every participant, mirroring
+/// the production topology (cf. `scripts/launch-localnet.sh`). Nodes submit mock
```
gilcu3: that script is very far from production topology, better remove the reference there

crates/contract/tests/sandbox/tee_verifier.rs
```diff
@@ -35,24 +54,104 @@ async fn setup() -> SandboxTestSetup {
...
+        .worker
+        .patch_state(
+            verifier.id(),
+            SANDBOX_TEST_PINNED_NOW_STORAGE_KEY,
+            &VALID_ATTESTATION_TIMESTAMP.to_le_bytes(),
+        )
+        .await
+        .unwrap();
+    verifier
+}
```
gilcu3: this seems a bit hacky, and led to some code. If we only need this for a single test that verifies a single fixed attestation, shouldnt it be enough to define a constant with "current time" for that ?

crates/contract/tests/sandbox/tee_verifier.rs
```diff
@@ -35,24 +54,104 @@ async fn setup() -> SandboxTestSetup {
...
+        .unwrap()
+        .remove(b"STATE".as_slice())
+        .expect("the contract must have a STATE entry");
+    let patched =
+        whitelist_dstack_in_state(&state, image_digest(), launcher_image_hash(), compose_hash);
+    setup
+        .worker
+        .patch_state(setup.contract.id(), b"STATE", &patched)
+        .await
+        .unwrap();
```
gilcu3, line 96: is this equivalent to voting? Seems pretty hacky. Although it is certainly more efficient. If we are doing it only for efficiency in tests, it would be better to introduce it later a bit more uniformly.

gilcu3: Ahh I realize now that the only thing that we do need to patch is the compose_hash, right?

crates/contract/tests/sandbox/tee_verifier.rs
gilcu3, line 1: This has got quite complex. I am wondering if this is the best approach to test this, but the answer is not simple as we want to have tests with real gas for the happy path. If we had sandbox tests in tee-verifier, and a simple stub for the single accepted attestation we have here, that might be an alternative.

gilcu3 (commented):
LGTM

Left a few comments, it was more complicated than expected.

Not approving yet as there are some merge conflicts right now.

docs/localnet/localnet.md
pbeza: Yeah, Claude suggested it automatically and it sounded accurate, so I just accepted it. Reverted https://github.com/near/mpc/pull/4165/changes/a35690424d3d852a30bd9c985257dd13f2088555. ✅

docs/deploy-tee-verifier.md
pbeza: True, the code would also need to be malicious and actually make use of storage in the first place (the current contract doesn’t, since it’s stateless). Deploying the verifier is an absolutely critical security step when we establish long-term trust in that contract, so personally I don’t mind having an extra check there. If the verifier contract ever starts using storage in the future, I’d rather have this check than not.

crates/e2e-tests/README.md
pbeza: Yeah, it’s orthogonal indeed (sorry for that!). I asked Claude to make sure the docs were up to date with the code in this PR, and it found and fixed this outdated bit. Personally, I’d leave it as-is and accept it, since it fixes a minor outdated piece of documentation while also adding some more docs related to this PR.

gilcu3: Ahh so it was just a doc change. I was afraid it was a code change as well when I wrote the comment. In that case it is totally fine

crates/e2e-tests/src/cluster.rs
pbeza: Ah... of course. Thanks for catching that! Reverted: https://github.com/near/mpc/pull/4165/commits/72728077fdee4fccc40de16874ede8e012e2a2d6. ✅

crates/e2e-tests/src/cluster.rs
pbeza: Good catch. Fixed: https://github.com/near/mpc/pull/4165/commits/0adc159f1d315ad83d5fe935f0760b6a8d96b375. ✅

crates/e2e-tests/src/cluster.rs
pbeza: https://github.com/near/mpc/pull/4165/commits/cdeebd2a215deefe2291e396856ecbb8e8f0f949 ✅

crates/contract/tests/sandbox/tee_verifier.rs
pbeza: > (...) If we only need this for a single test that verifies a single fixed attestation, shouldnt it be enough to define a constant with "current time" for that ?

Love the idea! I (together with Claude, ofc) overcomplicated it. Fixed: https://github.com/near/mpc/pull/4165/commits/26f5d9b7fa3695d6965b0cd9ff8d531fb977e394. ✅

crates/contract/tests/sandbox/tee_verifier.rs
pbeza: > Ahh I realize now that the only thing that we do need to patch is the compose_hash, right?

Yeah, exactly. The compose hash is the one thing we can’t vote in. The contract derives the allowed compose hashes itself from the voted launcher + MPC image using the compiled-in template ([`get_docker_compose_hash`](https://github.com/near/mpc/blob/efa148a0b4e1369a3d6fa0bf418b307d7e3c2bff/crates/contract/src/tee/proposal.rs#L509-L531)). Our fixture came from a CVM whose compose includes the [key-export service](https://github.com/near/mpc/blob/b0b6b7167e89c015ad59bfb4b0bbc78fcc8432cc/crates/test-utils/assets/launcher_image_compose.yaml#L28-L41), so no vote could ever produce that hash.

The rest (image digest, launcher hash, measurements) could go through real votes. But since we need the patch anyway, I just set everything in one go rather than adding a bunch of vote transactions to every fixture test.

crates/contract/tests/sandbox/tee_verifier.rs
pbeza: It got a bit simpler after your [other comments](https://github.com/near/mpc/pull/4165#discussion_r3822614796): the pinned clock is now just a compile-time constant in the test build:

https://github.com/near/mpc/blob/26f5d9b7fa3695d6965b0cd9ff8d531fb977e394/crates/tee-verifier-interface/src/lib.rs#L193-L196

so the storage key + `patch_state` is gone, and the whole setup is basically deploy + one state patch.

If you'd still prefer to simplify it further, how about doing that as a follow-up?

pbeza:
@gilcu3 I just addressed your comments. Ready for re-review. 🫡

pbeza:
@gilcu3 I just addressed your comments. Ready for re-review. 🫡

I also need one more reviewer; cc @haiyuechen-nearone / @anodar / @barakeinav1.

crates/contract/tests/sandbox/tee_verifier.rs
gilcu3: This looks much simpler now, thank you!

crates/tee-verifier-interface/src/lib.rs
```diff
@@ -190,13 +190,10 @@ pub enum VerificationResult {
...
-/// verifies against, as u64 little-endian seconds since the Unix epoch.
-///
-/// Production builds never read it. Tests write it from outside the contract
-/// (state patching), which is why the key must be shared between the verifier
-/// and its callers' test suites rather than staying private.
-pub const SANDBOX_TEST_PINNED_NOW_STORAGE_KEY: &[u8] = b"sandbox_test_pinned_now_seconds";
+/// The instant (Unix seconds) at which the committed attestation fixtures verify, used as
+/// the verification time by pinned-clock test builds of the verifier. Update whenever the
+/// fixtures are regenerated.
+pub const SANDBOX_TEST_PINNED_NOW_SECONDS: u64 = 1786622400;
```
gilcu3: nit: would be good to mention this in the README about how to update the attestation fixtures

gilcu3 (dismissed):
Thank you!

crates/contract/src/tee/test_utils.rs
```diff
@@ -104,14 +106,33 @@ pub fn set_block_timestamp(timestamp_nanos: u64) {
     );
 }
 
+/// The compose hash bypasses voting because the attestation fixture's hash is not
+/// derivable from the compiled-in template, so no vote can allow it.
```
barakeinav1: nit: the comment lost its own *why* in the refactor.

Both deleted call sites carried `// The fixture's launcher compose carries the key-export service, so its hash is not derivable.` — the surviving comment keeps "not derivable" but drops the reason, which is the one concrete fact that makes it click.

Two smaller things: it sits above the whole function but documents only the fourth parameter, and "compiled-in template" is unlinked, so a rename of `get_docker_compose_hash` won't be caught by CI.

```suggestion
/// Fills the four allowlists a Dstack submission is checked against: MPC image,
/// launcher image, expected measurements, and optionally one compose hash.
///
/// `compose_hash` is injected directly because it is the one value no vote can
/// authorize. [`get_docker_compose_hash`] derives the accepted hashes from the voted
/// launcher and MPC images through a compiled-in template, so only composes matching
/// that template are reachable. The fixture's CVM ran a modified compose carrying a
/// key-export service — the only way to get the signer key out of the CVM, which the
/// quote's `report_data` binds and tests must sign with.
///
/// Pass `None` to exercise the rejection path.
```

barakeinav1:
We probably could avoid this if we went in a different direction in #4109.

The launcher-compose idea we settled on there didn't fully work out. We assumed the test could just vote the fixture's compose hash in. It can't: the contract only derives compose hashes from its own template ([`get_docker_compose_hash`](https://github.com/near/mpc/blob/264102a0d042cb30a6980ea0b06a314d0a2a4486/crates/contract/src/tee/proposal.rs#L509-L531)), and there's no way to add an arbitrary one. So the test patches the raw `STATE` blob instead, and needs a `#[cfg(test)]` helper widened into the production crate.

Alternative: export the key from a modified **node image**, not the compose. The template substitutes the launcher hash and the node image hash — both votable — so the hash derives correctly and the test votes normally. That would drop `whitelist_dstack_in_state`, the `patch_state` call, and the `allow_compose_hash` cfg widening, which is the only production-file change in this PR.

Cost is another fixture capture, and a published image that prints its signer key (today's compose hash provably can never be voted in on mainnet).

Not blocking — this can be done in a follow-up if there is an agreement on this.

crates/near-mpc-contract-interface/src/client.rs
```diff
@@ -207,6 +208,23 @@ impl<C: CallContract> MpcContractHandle<C> {
...
+            candidate_account_id,
+            expected_code_hash,
+        ))?;
+        self.call(FunctionCallArgs::no_deposit(
+            VOTE_TEE_VERIFIER_CHANGE,
+            args,
+            VOTE_TEE_VERIFIER_CHANGE_GAS,
+        ))
+        .await
+    }
```
barakeinav1, line 237: nit: adding public API to `near-mpc-contract-interface` (a shipped crate) would sit better in a dedicated PR — every caller here is a test, and `test(contract):` doesn't hint that production API changed.

Not blocking.

crates/e2e-tests/README.md
```diff
@@ -343,12 +347,13 @@ so any nextest filter or flag works (substring filters, `-E` expressions,
...
-`MPC_CONTRACT_WASM` and `MPC_PARALLEL_CONTRACT_WASM` environment variables
-read by `must_load_contract_wasm` / `must_load_parallel_contract_wasm` in
-`tests/common.rs`; if the env var is unset and no pre-built WASM is found,
+The task runner builds five things before tests run: the mpc-node binary with
+the `network-hardship-simulation` feature, the MPC contract WASM, the
+tee-verifier WASM, the test parallel contract WASM, and the backup CLI. WASM
+paths are passed to tests via the `MPC_CONTRACT_WASM`,
+`MPC_TEE_VERIFIER_WASM` and `MPC_PARALLEL_CONTRACT_WASM` environment
+variables, read by the `must_load_*` helpers in `tests/common.rs` and
+`src/cluster.rs`; if the env var is unset and no pre-built WASM is found,
```
barakeinav1: `must_load_tee_verifier_wasm` moved to `tests/common.rs` in 0adc159f, so `src/cluster.rs` no longer has one.

```suggestion
variables, read by the `must_load_*` helpers in `tests/common.rs`; if the
env var is unset and no pre-built WASM is found,
```

crates/e2e-tests/README.md
barakeinav1, line 1: The documented `MpcClusterConfig` is missing the field this PR adds, and `default_for_test` still shows the old signature. Both are just outside the diff, so no suggestion block — the two edits are:

L226, after `contract_wasm`:
```rust
    pub tee_verifier_wasm: Vec<u8>,             // pre-compiled by the test
```

L244:
```rust
    pub fn default_for_test(port_seed: u16, contract_wasm: Vec<u8>, tee_verifier_wasm: Vec<u8>) -> Self;
```

crates/test-utils/src/attestation.rs
```diff
@@ -22,9 +22,7 @@ pub const TEST_MPC_IMAGE_DIGEST_HEX: &str = include_str!("../assets/mpc_image_di
 pub const TEST_LAUNCHER_IMAGE_COMPOSE_STRING: &str =
     include_str!("../assets/launcher_image_compose.yaml");
 
-/// Unix time as of 2026/08/13, a date where the measurements stored in ../assets are valid. Update
-/// this whenever those measurements are regenerated.
-pub const VALID_ATTESTATION_TIMESTAMP: u64 = 1786622400;
+pub use tee_verifier_interface::SANDBOX_TEST_PINNED_NOW_SECONDS as VALID_ATTESTATION_TIMESTAMP;
```
barakeinav1: This re-export makes the fixture-regeneration steps stale — `crates/test-utils/assets/README.md:52` still says:

> 4. Update `VALID_ATTESTATION_TIMESTAMP` in `crates/test-utils/src/attestation.rs` ...

which is no longer editable there. Minimal fix, swapping the name and path:

> 4. Update `SANDBOX_TEST_PINNED_NOW_SECONDS` in `crates/tee-verifier-interface/src/lib.rs` ...

(This is @gilcu3's [earlier comment](https://github.com/near/mpc/pull/4165#discussion_r3822614796) about mentioning it in the fixture README.)

crates/tee-verifier-interface/src/lib.rs
```diff
@@ -190,6 +190,11 @@ pub enum VerificationResult {
     Rejected(VerifierError),
 }
 
+/// The instant (Unix seconds) at which the committed attestation fixtures verify, used as
+/// the verification time by pinned-clock test builds of the verifier. Update whenever the
+/// fixtures are regenerated.
+pub const SANDBOX_TEST_PINNED_NOW_SECONDS: u64 = 1786622400;
```
barakeinav1: nit: a `SANDBOX_TEST_*` constant in a shipped interface crate reads a bit odd — it's the only test-only item in here.

I get why it landed here (both `tee-verifier` and `test-utils` need it, and this is their shared dep) and I don't have a better suggestion. Just noting it.

crates/tee-verifier/src/lib.rs
```diff
@@ -62,3 +62,14 @@ impl TeeVerifier {
...
+/// The timestamp quotes are verified against: block time, except in test builds,
+/// which pin [`tee_verifier_interface::SANDBOX_TEST_PINNED_NOW_SECONDS`] to keep
+/// the checked-in collateral fixture inside its validity window.
+fn now_seconds() -> u64 {
+    if cfg!(feature = "sandbox-test-hooks") {
+        tee_verifier_interface::SANDBOX_TEST_PINNED_NOW_SECONDS
+    } else {
+        env::block_timestamp_ms() / 1000
+    }
+}
```
barakeinav1: The simplification here is right, but the safety net went out with it.

26f5d9b7 removed two things along with the storage-key approach: the second gate (`--cfg mpc_sandbox_wasm` plus the `compile_error!` that failed a WASM build carrying the feature without it), and `verify_quote__should_ignore_pinned_timestamp_on_production_build`. What's left preventing a pinned-clock verifier from being built and deployed is that no build command happens to pass `--features sandbox-test-hooks` — nothing mechanical.

The part that bothers me more: `--all-features` (the test command in AGENTS.md) enables the feature crate-wide, so in CI `now_seconds()` always takes the pinned branch and `env::block_timestamp_ms()` is executed by no test at all. `set_valid_timestamp_context()` in `tests/verify_quote.rs` silently becomes a no-op. Invert this `if`/`else` and everything still passes green — while a deployed verifier would think it is permanently 2026-08-13 and accept expired or revoked collateral forever.

Cheap enough to close:

```rust
#[cfg(not(feature = "sandbox-test-hooks"))]
#[test]
fn now_seconds__should_track_block_time() { ... }
```

`--all-features` skips it, a plain `cargo test -p tee-verifier` runs it, and an inverted condition gets caught.

The audit step you added in `deploy-tee-verifier.md` ("A verified report after that point is a failed audit") does cover this on-chain, which helps — but only once the fixture window has lapsed, and only if the operator runs it.

barakeinav1 (changes requested):
Thanks @pbeza, coverage is much better now — driving the Verified path through the real verifier at the production gas budget is a big step up from the in-process coverage, and the post-DCAP checks (compose hash, `report_data`, TLS-key ownership) are valuable.

Added a few nits, one design idea as a [follow-up](https://github.com/near/mpc/pull/4165#issuecomment-5385939976), and one blocker: [the pinned clock lost its guard](https://github.com/near/mpc/pull/4165#discussion_r3838541544). Requesting changes on that one since it has security implications if that build ever reaches a deployed verifier — a few lines should close it.

crates/tee-verifier/src/lib.rs
netrome: Discussed this in a meeting now. We want to be able to control time in near-sandbox which seems like a missing feature we should contribute upstream.

To unblock this PR, we can make this a bit more safe by flipping so that the production behavior is gated behind a feature rather than the test behavior. This way we cannot accidentally enable the test behavior (since features can be added implicitly but not removed).

We should also add a TODO comment and contribute upstream to the near sandbox repo to ensure we can control the time in the tests.

pbeza: Addressed in https://github.com/near/mpc/pull/4165/commits/4411760cb0203b850deece780e0f437e3a733554, also taking [this](https://nearone.slack.com/archives/C0912BTG51T/p1787568009824359?thread_ts=1786461736.481479&cid=C0912BTG51T) discussion into account. ✅

crates/contract/src/tee/test_utils.rs
pbeza: Fixed: 79c8ff800963dab8f28e25d33b8be061bb4842d6.

crates/e2e-tests/README.md
pbeza: Addressed in 79c8ff800963dab8f28e25d33b8be061bb4842d6. ✅

crates/e2e-tests/README.md
pbeza: Addressed in 79c8ff800963dab8f28e25d33b8be061bb4842d6. ✅

crates/tee-verifier-interface/src/lib.rs
pbeza: 4411760cb0203b850deece780e0f437e3a733554

crates/test-utils/src/attestation.rs
pbeza: Addressed in 4411760cb0203b850deece780e0f437e3a733554. ✅

crates/tee-verifier-interface/src/lib.rs
pbeza: Moved to `crates/test-utils/src/attestation.rs` in 4411760cb0203b850deece780e0f437e3a733554. ✅

crates/near-mpc-contract-interface/src/client.rs
pbeza: It's too much overhead IMHO. I'll leave it as-is unless more people complain.

crates/tee-verifier-interface/src/lib.rs
pbeza: Addressed in 4411760c: `VALID_ATTESTATION_TIMESTAMP` is now defined in `crates/test-utils/src/attestation.rs`, and the fixture README's step 4 (`crates/test-utils/assets/README.md`) already points there:

https://github.com/near/mpc/blob/4411760cb0203b850deece780e0f437e3a733554/crates/test-utils/assets/README.md?plain=1#L52

pbeza:
@barakeinav1 @gilcu3, it's [ready](https://github.com/near/mpc/pull/4165#discussion_r3843179923) for re-review. :saluting_face:

gilcu3 (approved):
Thank you!

barakeinav1 (approved):
Thanks for updating, approved.

PR #4172, opened 2026-08-18
chore: release v3.14.1

kevindeforth:
resolves https://github.com/near/mpc/issues/4174

actual logic is in commit 11d4235a09b33e9ccf61f8c8c73daea6f6d3887a 
release logic is in commit 37bcff1f41df428046caaa323569499dd2aad098


crates/contract/src/v3_14_0_state.rs
```diff
@@ -0,0 +1,86 @@
...
+impl From<MpcContract> for crate::MpcContract {
+    fn from(old: MpcContract) -> Self {
+        if !matches!(old.protocol_state, ProtocolContractState::Running(_)) {
+            env::panic_str("Contract must be in running state when migrating.");
+        }
+
+        let mut tee_state = old.tee_state;
+        tee_state.extend_dstack_attestation_expiries(Duration::from_secs(
+            mpc_attestation::attestation::DEFAULT_EXPIRATION_DURATION_SECONDS,
+        ));
```
pbeza, line 65: Nit: `DEFAULT_EXPIRATION_DURATION_SECONDS` is meant to represent the validity duration of an attestation, but `.extend_dstack_attestation_expiries()` adds it to whatever expiry we currently have, so the effective validity can grow beyond `DEFAULT_EXPIRATION_DURATION_SECONDS`, no? Probably not a problem anyway.

crates/contract/src/lib.rs
```diff
@@ -8513,4 +8513,77 @@ mod tests {
...
+        // Given persisted 3.14.0 state holding a dstack and a mock attestation.
+        const EXPIRY_SECONDS: u64 = 1_000_000;
+        let mock = MpcMockAttestation::WithConstraints {
+            mpc_docker_image_hash: None,
+            launcher_docker_compose_hash: None,
+            expiry_timestamp_seconds: Some(EXPIRY_SECONDS),
+            expected_measurements: None,
+        };
+        let (mut contract, _, _, _) = setup_running_contract_with_domain(3, 2, 2);
+        contract.config.launcher_hash_unused_ttl_seconds = 1_209_600; // deployed 3.14.0 value
```
pbeza: Nit: perhaps it’s worth adding that this is 14 days. Otherwise, the value might feel a bit random.

crates/contract/src/lib.rs
```diff
@@ -8513,4 +8513,77 @@ mod tests {
...
+        assert_matches!(
+            &stored.get(&mock_key).unwrap().verified_attestation,
+            VerifiedAttestation::Mock(stored) if *stored == mock
+        );
+
+        // And the launcher TTL was raised, keeping `Config::validate`'s invariant true.
+        assert_eq!(
+            reloaded.config.launcher_hash_unused_ttl_seconds,
+            config::DEFAULT_LAUNCHER_HASH_UNUSED_TTL_SECONDS
+        );
```
pbeza, line 8570: Perhaps it’s also worth explicitly asserting that 

```
contract.config.launcher_hash_unused_ttl_seconds < config::DEFAULT_LAUNCHER_HASH_UNUSED_TTL_SECONDS
```
or just:

```
contract.config.launcher_hash_unused_ttl_seconds != config::DEFAULT_LAUNCHER_HASH_UNUSED_TTL_SECONDS
```

Otherwise, I need to manually verify that this assertion isn’t trivially true, which would be the case if these two values were equal. It would also make this part of the comment more explicit: `And the launcher TTL was raised`.

pbeza (dismissed):

crates/contract/src/v3_14_0_state.rs
kevindeforth: Yeah, that's correct. I could have added only the difference between the two, but it would just lead to more complex code with the same / similar result.

_edit: ok, it's not that complicated. I will add that I think. Would give us 2 weeks from now._

crates/contract/src/v3_14_0_state.rs
```diff
@@ -0,0 +1,89 @@
...
+        let mut tee_state = old.tee_state;
+        tee_state.extend_dstack_attestation_expiries(Duration::from_secs(
+            mpc_attestation::attestation::DEFAULT_EXPIRATION_DURATION_SECONDS
+                - PRE_3_14_1_EXPIRATION_DURATION_SECONDS,
+        ));
+
+        let mut config = old.config;
+        config.launcher_hash_unused_ttl_seconds = config
+            .launcher_hash_unused_ttl_seconds
+            .max(crate::config::DEFAULT_LAUNCHER_HASH_UNUSED_TTL_SECONDS);
```
gilcu3: is this because the launcher itself could be unused? That cannot happen right now because we have some nodes with good attestations in both networks

crates/contract/tests/sandbox/upgrade_from_current_contract.rs
```diff
@@ -126,7 +126,7 @@ async fn test_propose_update_config() {
         verifier_tera_gas: 15,
         resolve_verification_tera_gas: 16,
         // Must satisfy `Config::validate` (>= DEFAULT_EXPIRATION_DURATION_SECONDS).
-        launcher_hash_unused_ttl_seconds: 14 * 24 * 60 * 60,
+        launcher_hash_unused_ttl_seconds: 21 * 24 * 60 * 60,
```
gilcu3: Ahh so that is the reason, there is a validation step between launcher ttl and the attestation ttl, that's a bit not intuitive to me

CHANGELOG.md
```diff
@@ -5,6 +5,25 @@ All notable changes to this project will be documented in this file.
...
+
+- (@dependabot[bot]): Bump the rust-minor-and-patch group with 4 updates (#4113)
+
+- (@barakeinav1): Install catatonit for repro-env podman builds (#4125)
+
+- (@barakeinav1): Install catatonit in the image-publishing workflows (#4128)
+
+- (@gilcu3): Move runner images to ubuntu 26.04 (#4134)
+
+- (@kevindeforth): *(contract)* Extend attestation expiry time during migration
```
gilcu3, line 24: this is where some links are missing, for example below in the file:

> - [#3667](https://github.com/near/mpc/pull/3667)(@kevindeforth): Add metrics to detect starvation and resolve CI flakyness (#3667)

deny.toml
```diff
@@ -33,6 +33,10 @@ ignore = [
   # has no patched release. Only reachable through near-vm-runner, which never
   # shares a Store across Engines.
   "RUSTSEC-2026-0222", # wasmtime stores can mix up type indices between engines
+  # TODO(#4117): remove once nearcore moves `near-network` off `im`, whose maintained fork is `imbl`.
+  "RUSTSEC-2026-0247", # bitmaps is unmaintained
+  "RUSTSEC-2026-0248", # im is unmaintained
+  "RUSTSEC-2026-0251", # sized-chunks is unmaintained
```
gilcu3, line 39: nit: would be better to avoid the commit that included this and the dep bumps in this release. We already have all the ci commits that unfortunately cannot be completely avoided, although they could be squashed so that a single line in the changelog explains them

gilcu3 (approved):
Thanks!

Left a couple of minor comments

pbeza (approved):

deny.toml
kevindeforth: hmm, I was worried about ci failing, hence I added most of these extra stuff.
Why would it be better to exclude them?

CHANGELOG.md
kevindeforth: Oh boy, I need to check why those links don't show when I generate the changelog...

deny.toml
gilcu3: Only so that there are the least amount of changes in this patch release, CI should not fail without any of these, as they are about advisories

CHANGELOG.md
gilcu3: Probably easier to generate them manually as in any case the commits would be squashed being this a single PR adding all of them

kevindeforth:
closing this, it has been replaced with a series of PRs:

- making CI pass: https://github.com/near/mpc/pull/4177
- update contract migrations: https://github.com/near/mpc/pull/4179
- extend attestation expiry date: https://github.com/near/mpc/pull/4178
- actual release: https://github.com/near/mpc/pull/4187

Issue #4174, opened 2026-08-18
testnet contract release 3.14.1


kevindeforth:
### Background

https://nearone.slack.com/archives/C0APULZJVHN/p1787037314757519

### User Story

as a node operator, it's easier to vote in a new contract than to do an emergency migration

### Acceptance Criteria

release 3.14.1

kevindeforth:
this was resolved with #4187

PR #4177, opened 2026-08-19
ci: move runner images to ubuntu 26.04

kevindeforth:
part of https://github.com/near/mpc/issues/4174


gilcu3 (approved):
Thank you

one nit before merging, the PR title should be something like
```
ci: move runner images to ubuntu 26.04
```

pbeza (approved):

PR #4189, opened 2026-08-19
refactor(devnet): typed mpc-contract calls

### LINKED ISSUE #4188: refactor: integrate new mpc interface in devnet
### Background

Part of https://github.com/near/mpc/issues/3693, we can now route the governance in devnet calls via the new mpc contract caller.

### User Story

As a developer, I like when things are simple

### Acceptance Criteria

route existing governance calls in devnet via the mpc contract interface

_No response_

kevindeforth:
resolves #4188 part of https://github.com/near/mpc/issues/3693


crates/devnet/src/caller.rs
```diff
@@ -0,0 +1,75 @@
...
+
+pub(crate) trait CallMpcContract {
+    fn call_mpc(&self, contract_id: &AccountId) -> MpcContractHandle<DevnetCaller>;
+}
+
+impl CallMpcContract for OperatingAccount {
+    /// The returned handle logs each call and waits for transaction to be final
+    fn call_mpc(&self, contract_id: &AccountId) -> MpcContractHandle<DevnetCaller> {
+        MpcContractHandle::new(
+            DevnetCaller::new(self.any_access_key_handle(), TxExecutionStatus::Final, true),
```
anodar: nit: that last parameter there is awkward, from callsite not clear what is it exactly either. You are also introducing another constructor `quiet()` anyway which is the only place that sets verbose false there, and nothing is actually using new with verbose=false.


Consider renaming renaming constructors to `new_verbose` and `new` (or `new_quiet`) and drop that bool parameter from the constructor.

anodar (dismissed):

crates/devnet/src/caller.rs
```diff
@@ -0,0 +1,75 @@
...
+        verbose: bool,
+    ) -> Self {
+        Self {
+            key,
+            wait_until,
+            verbose,
+        }
+    }
+
+    pub fn quiet(self) -> Self {
```
gilcu3: nit: maybe call it `new_quiet`? the name now does not seems a constructor at all

crates/devnet/src/caller.rs
gilcu3: Agree that there is something that feels overengineered here. Not sure if the new `map_caller` function in `crates/near-mpc-contract-interface/src/client.rs` will also be used elsewhere

gilcu3 (dismissed):
Nice!

crates/devnet/src/caller.rs
kevindeforth: Hmm, so there are three different topics in this thread:
1. _Do we need that boolean in the constructor?_
Right now, we do not. But omitting this field from the constructor seems wrong to me - if we want to add it in the future, we would need to refactor the call-sites, which would be annoying.
2. _Do we need `quiet` if we have the constructor?_
`quiet()` isn't a constructor, it's a setter, taking ownership of `self`, setting a property and then returning itself back (this follows the builder pattern). So, it doesn't quite compare to `DevnetCaller::new(...)`, they have different purposes.
3. _Will `map_caller` be used elsewhere?_
Yes, in fact, I required it in https://github.com/near/mpc/pull/4108.

With the `map_caller`, we have{
```rust
 let result = proposer
            .call_mpc(&contract)
            .map_caller(DevnetCaller::quiet)
            .propose_update(ProposeUpdateArgs {
                code: Some(contract_code),
                config: None,
            })
```

without it, we would have:
```rust
 let result = proposer
            .call_mpc_quiet(&contract)
            .propose_update(ProposeUpdateArgs {
                code: Some(contract_code),
                config: None,
            })
```
or 
```rust
 let result = proposer
            .call_mpc(&contract, false)
            .propose_update(ProposeUpdateArgs {
                code: Some(contract_code),
                config: None,
            })
```
but then, we would need to pack all possible options into `call_mpc`, or build specific builders for that, which would be quite annoying and doesn't scale too well.

crates/devnet/src/caller.rs
kevindeforth: It's not a constructor. Lets continue in https://github.com/near/mpc/pull/4189#discussion_r3814105756, as it came up there as well.

crates/devnet/src/caller.rs
kevindeforth: Hmm, that reminds me, I could make a trait and then it would just read:

```rust
 let result = proposer
            .call_mpc(&contract)
            .quiet()
            .propose_update(ProposeUpdateArgs {
                code: Some(contract_code),
                config: None,
            })
```

gilcu3: yeah that looks a bit nicer :)

kevindeforth: changed in https://github.com/near/mpc/pull/4189/commits/bec1843691bd968a02172d9e41c0f46077f36de9

anodar (approved):

gilcu3 (approved):

PR #4197, opened 2026-08-20
refactor(contract): move TEE attestation lifecycle into api/attestation.rs

gilcu3:
part 11/12 of #1962


pbeza (dismissed):
NGL, I rely on Claude to make sure it’s just a mechanical cut-and-paste (it compared it programmatically using some ad hoc `diff`-based script). The only optional thing perhaps worth addressing is the non-blocker Claude raised about the ``[`IterableMap`]`` intra-doc link in the `WORST_CASE_ENTRY_BYTES` doc comment no longer resolving.

crates/contract/src/api/attestation.rs
```diff
@@ -0,0 +1,1338 @@
...
+
+        let (participant_id, _, participant_info) = participants
+            .participants()
+            .first()
+            .expect("at least one participant")
+            .clone();
+
+        let valid_attestation = Attestation::Mock(MockAttestation::Valid);
+
+        // ❌ Case: signer != predecessor — should panic
```
haiyuechen-nearone: Not added by this change, I'm just curious. Do we "like" emojis in the code? My experience in previous codebases it was a big no no.

haiyuechen-nearone (dismissed):

crates/contract/src/api/attestation.rs
gilcu3: Big no for me, will take the change to remove that one :)

pbeza (approved):

haiyuechen-nearone (approved):

PR #4201, opened 2026-08-20
fix(node): keep attestation tasks alive after submission retry timeout

### LINKED ISSUE #2287: Attestation tasks permanently die after 12h retry timeout
## Background

Both [`periodic_attestation_submission`](https://github.com/near/mpc/blob/7b27b96b5b9b17f3ae6bc8b83570b96946e7bbc5/crates/node/src/tee/remote_attestation.rs#L140) and [`monitor_attestation_removal`](https://github.com/near/mpc/blob/7b27b96b5b9b17f3ae6bc8b83570b96946e7bbc5/crates/node/src/tee/remote_attestation.rs#L187) tasks will eventually terminate if the node keeps trying to post an invalid attestation (for example, if it's running outdated firmware). `periodic_attestation_submission` exits after [~12 hours](https://github.com/near/mpc/blob/7b27b96b5b9b17f3ae6bc8b83570b96946e7bbc5/crates/node/src/tee/remote_attestation.rs#L89) of failed retries; `monitor_attestation_removal` follows later — once the old attestation expires and is cleaned up — and exits after its own 12-hour retry window. Both just log an error.

Is this behavior intentional, or should we avoid terminating them permanently?

## Acceptance Criteria

- [ ] Decide whether attestation tasks should be allowed to terminate permanently
- [ ] If they terminate: consider propagating the failure to trigger a node shutdown, since the MPC node currently continues running

- `periodic_attestation_submission`: [`remote_attestation.rs:140`](https://github.com/near/mpc/blob/7b27b96b5b9b17f3ae6bc8b83570b96946e7bbc5/crates/node/src/tee/remote_attestation.rs#L140)
- `monitor_attestation_removal`: [`remote_attestation.rs:187`](https://github.com/near/mpc/blob/7b27b96b5b9b17f3ae6bc8b83570b96946e7bbc5/crates/node/src/tee/remote_attestation.rs#L187)
- 12h retry timeout (`MAX_RETRY_DURATION`): [`remote_attestation.rs:29`](https://github.com/near/mpc/blob/7b27b96b5b9b17f3ae6bc8b83570b96946e7bbc5/crates/node/src/tee/remote_attestation.rs#L29)
- Both tasks spawned as fire-and-forget `tokio::spawn`: [`cli.rs:449-494`](https://github.com/near/mpc/blob/7b27b96b5b9b17f3ae6bc8b83570b96946e7bbc5/crates/node/src/cli.rs#L449-L494)
- Related: #2282 (TEE Context refinement, shutdown signal propagation)

pbeza:
Closes #2287


crates/node/src/tee/remote_attestation.rs
```diff
@@ -276,8 +266,6 @@ pub async fn monitor_attestation_removal<T: TransactionSender + Clone>(
     );
 
     let mut was_available = initially_available;
-    let report_data: ReportData =
-        ReportDataV1::new(*tls_public_key.as_bytes(), *account_public_key.as_bytes()).into();
 
     while tee_accounts_receiver.changed().await.is_ok() {
```
pbeza:
```suggestion
    loop {
```

crates/node/src/tee/remote_attestation.rs
```diff
@@ -132,103 +133,100 @@ fn validate_remote_attestation(
...
+        crate::metrics::MPC_TEE_ATTESTATION_ATTEMPTS_TOTAL
+            .with_label_values(&[outcome])
+            .inc();
+        let attestation = match result {
+            Ok(attestation) => attestation,
+            Err(error @ AttestationError::CollateralFetch(_)) => {
+                tracing::warn!(
+                    %error,
+                    "TEE attestation generation failed, will retry on the next trigger"
+                );
```
pbeza:
```suggestion
                tracing::warn!(%error, "TEE attestation generation failed");
```

crates/node/src/tee/remote_attestation.rs
```diff
@@ -294,72 +282,16 @@ pub async fn monitor_attestation_removal<T: TransactionSender + Clone>(
...
-                pre_submit_expiry,
-            )
-            .await?;
+            if !submitter.generate_and_submit().await {
+                // Stay armed so the next TEE-accounts update retries while the node is absent
+                continue;
+            }
         }
 
         was_available = is_available;
```
pbeza:
```suggestion
        was_available = is_available;
        if tee_accounts_receiver.changed().await.is_err() {
            break;
        }
```

crates/node/src/tee/remote_attestation.rs
```diff
@@ -491,233 +405,233 @@ mod tests {
...
+    async fn monitor_attestation_removal__should_survive_submission_retry_timeout() {
+        // Given: the node is removed and the resubmission burns its whole retry window
+        let setup = test_setup();
+        setup.sender().set_failing(true);
+        let handle = setup.spawn_monitor();
+        tokio::task::yield_now().await;
+        setup.remove_node_from_tee_accounts();
+        tokio::time::sleep(MAX_RETRY_DURATION + Duration::from_secs(1)).await;
+
+        // When: the next TEE-accounts update arrives while the node is still absent
```
pbeza:
```suggestion
        // When: a TEE-accounts update arrives while the node is still absent
```

crates/node/src/tee/remote_attestation.rs
```diff
@@ -294,72 +282,16 @@ pub async fn monitor_attestation_removal<T: TransactionSender + Clone>(
...
+                // Stay armed so the next TEE-accounts update retries while the node is absent
+                continue;
+            }
         }
 
         was_available = is_available;
     }
 
-    Ok(())
+    tracing::warn!("TEE accounts watch channel closed; stopping attestation removal monitoring");
```
pbeza:
```suggestion
    tracing::info!("TEE accounts watch channel closed; stopping attestation removal monitoring");
```

crates/node/src/tee/remote_attestation.rs
```diff
@@ -294,72 +282,16 @@ pub async fn monitor_attestation_removal<T: TransactionSender + Clone>(
...
-                fresh_attestation.clone(),
-                tls_public_key.clone(),
-                account_public_key.clone(),
-                &allowed_image_hashes_in_contract,
-                &allowed_launcher_compose_hashes_in_contract,
-                pre_submit_expiry,
-            )
-            .await?;
+            if !submitter.generate_and_submit().await {
+                // Stay armed so the next TEE-accounts update retries while the node is absent
```
pbeza:
```suggestion
                // Treat the removal as unhandled and retry after a delay: the watch fires
                // only on actual changes to the attested-nodes list, which a quiet network
                // may never produce again
                if let Ok(Err(_)) = tee_accounts_receiver
                    .changed()
                    .timeout(RESUBMISSION_RETRY_DELAY)
                    .await
                {
                    break;
                }
```

crates/node/src/tee/remote_attestation.rs
```diff
@@ -132,103 +133,100 @@ fn validate_remote_attestation(
...
+            self.tls_public_key.clone(),
             pre_submit_expiry,
         )
-        .await?;
+        .await
+        {
+            tracing::error!(
+                ?error,
+                "attestation submission failed; will retry with a fresh attestation on the next trigger"
+            );
```
pbeza:
```suggestion
            tracing::error!(?error, "attestation submission failed");
```

crates/node/src/tee/remote_attestation.rs
```diff
@@ -491,233 +405,233 @@ mod tests {
...
-        ));
+    #[tokio::test]
+    #[expect(non_snake_case)]
+    async fn monitor_attestation_removal__should_resubmit_when_attestation_removed() {
+        // Given
+        let setup = test_setup();
+        let monitoring_task = setup.spawn_monitor();
 
         // Yield control to allow the monitoring task to start and process initial state.
         // This is preferred over sleep() as it doesn't introduce arbitrary timing delays
```
pbeza:
```suggestion

```

crates/node/src/tee/remote_attestation.rs
```diff
@@ -132,103 +133,100 @@ fn validate_remote_attestation(
...
-            &allowed_launcher_compose_hashes_in_contract,
+        if let Err(error) = validate_remote_attestation(
+            &attestation,
+            self.tls_public_key.clone(),
+            self.account_public_key.clone(),
+            &allowed_image_hashes,
+            &allowed_launcher_compose_hashes,
+        ) {
+            // Non-blocking pre-flight check; its exact output is documented in the external
+            // node-operator guide, so keep the message format stable.
```
pbeza:
```suggestion
            // Submit anyway: the contract runs the authoritative verification, and this local
            // check may fail spuriously on a stale view of the allowed hashes. Its exact
            // output is documented in the external node-operator guide, so keep the message
            // format stable.
```

crates/node/src/tee/remote_attestation.rs
```diff
@@ -491,233 +405,233 @@ mod tests {
...
+        tokio::time::sleep(MAX_RETRY_DURATION + Duration::from_secs(1)).await;
+
+        // When: the next TEE-accounts update arrives while the node is still absent
+        setup.sender().set_failing(false);
+        setup.remove_node_from_tee_accounts();
+
+        // Then
+        tokio::time::timeout(MAX_RETRY_DURATION, setup.sender().wait_for_submission())
+            .await
+            .expect("expected a resubmission on the next update after the retry window timed out");
```
pbeza:
```suggestion
            .expect("expected a resubmission on the update after a failed one");
```

pbeza:
A few points worth highlighting for reviewers before they start reviewing:

1. I did a small refactor and introduced a shared `generate_and_submit` function used by both `periodic_attestation_submission` and `monitor_attestation_removal`. There was quite a bit of copy-paste between them:
   https://github.com/near/mpc/blob/9ce222ed996b94d49b5f26d2d129d3df98f78c1c/crates/node/src/tee/remote_attestation.rs#L298-L356

   https://github.com/near/mpc/blob/9ce222ed996b94d49b5f26d2d129d3df98f78c1c/crates/node/src/tee/remote_attestation.rs#L179-L231

2. I added:

   https://github.com/near/mpc/blob/38e27a20101e322ca5d183f0e269cdfbb7952fc0/crates/node/src/run.rs#L371-L373

   to make sure we don't burst with attestation resubmission retries (`tokio::time::interval` defaults to `MissedTickBehavior::Burst`).

3. I added a new metric:

   https://github.com/near/mpc/blob/38e27a20101e322ca5d183f0e269cdfbb7952fc0/crates/node/src/metrics.rs#L480-L490

   (We should probably set up an alert for that metric.)

   The existing `mpc_tee_attestation_attempts_total` only counts attestation generation. Since a permanently failing submission is now a quiet infinite retry loop rather than causing the task to die, operators need something they can alert on.

4. I reworked the tests: renamed to the repo's `<unit>__should_<behavior>` convention, and split `test_tee_attestation_removal_detection` into focused cases. The per-test setup (keys, watch channels, submitter) is now shared:
    https://github.com/near/mpc/blob/38e27a20101e322ca5d183f0e269cdfbb7952fc0/crates/node/src/tee/remote_attestation.rs#L430-L492

   `MockSender` also gained a `set_failing` switch. Nothing could simulate a failed submission before.

   I dropped two assertions. They were not separate tests, they were the tail of `test_tee_attestation_removal_detection`: `Expected submission count to remain stable after stopping monitoring service` and `Expected no resubmission when monitoring service is stopped`. Both run after `monitoring_task.abort()`, so they assert that an aborted task does nothing, which is a property of `JoinHandle::abort` rather than of our code:
   https://github.com/near/mpc/blob/9ce222ed996b94d49b5f26d2d129d3df98f78c1c/crates/node/src/tee/remote_attestation.rs#L566-L668

   They are replaced by `monitor_attestation_removal__should_stop_when_watch_channel_closes` (the real termination path) and `monitor_attestation_removal__should_resubmit_on_each_removal`. The five new behaviors: surviving the 12h submission retry timeout, retrying a failed resubmission on the fixed delay, retrying it on the next TEE-accounts update, resubmitting every time the node goes from present to absent rather than only the first time, and stopping when the watch channel closes.

Despite all these changes, the PR is still relatively small, so I think it’s fine not to split the refactor into a separate PR.
The extracted code also helped keep the new tests smaller.

crates/node/src/metrics.rs
```diff
@@ -477,5 +477,17 @@ pub static MPC_TEE_ATTESTATION_ATTEMPTS_TOTAL: LazyLock<prometheus::IntCounterVe
         .unwrap()
     });
 
+/// A node that cannot get its attestation onto the contract retries forever rather than giving
+/// up, so this counter is the only signal an operator has that the node is failing to attest.
+pub static MPC_TEE_ATTESTATION_SUBMISSIONS_TOTAL: LazyLock<prometheus::IntCounterVec> =
```
gilcu3: technically the operator can know the same signal if they look at any chain explorer and filter their last attestation submissions. I think we could skip this comment, what the metric is is clear already, the why it was added does not need to be mentioned here IMO

docs/backup-service-attestation-data.md
```diff
@@ -107,7 +107,7 @@ The following walkthroughs show how the attestation data defined above is used i
 2. Backup service generates an Ed25519 TLS keypair and a NEAR account keypair inside the TEE.
 3. Operator retrieves the account public key from the backup service's HTTP endpoint and adds it as an access key to their NEAR account (same as for MPC nodes — see [securing-mpc-with-tee §Bootstrapping](./securing-mpc-with-tee-design-doc.md#bootstrapping)).
 
-**Attestation registration (periodic — every 7 days, starting on first boot):**
+**Attestation registration (periodic — every hour, starting on first boot):**
```
gilcu3: why are we changing this file? we still don't know how the backup service will behave

docs/hot-tee-signing-design.md
```diff
@@ -466,7 +466,7 @@ sequenceDiagram
         Note over OP: Operator restarts CVM with correct image
     end
 
-    loop Every 7 days
+    loop Every hour
```
gilcu3, line 469: same, I don't think we need to update this old doc, we could actually remove this one to reduce the noise

docs/securing-mpc-with-tee-design-doc.md
```diff
@@ -395,8 +395,8 @@ _Note_: submit_participant_info must be called by the node itself - the quote's
 - Export remote attestation information via the same endpoint.
 - Send remote attestation to the contract (via submit_participant_info)
   - On each boot (after state sync is completed)
-  - Every 7 days.
-- Call verify_tee every 7 days - in order to trigger a re-validation of attestation information on the contract.
+  - Every hour.
+- Call verify_tee every 2 days - in order to trigger a re-validation of attestation information on the contract.
 - Monitor contract state - and if a new MPC docker image hash was voted successfully, pull that image and store on disk (so that the Launcher will use it for next boot)
```
gilcu3, line 400: I am in general opposed to keep design docs in line with the code. IMO we should have design docs, that are kept for historical reasons, and system docs, that need to be up to date. But we can discuss this in the retro today, if there is time

docs/tee-context-design.md
gilcu3, line 1: I don't see the need to change this file. A design does not need to strictly match implementation IMO, some details might differ, as the ones being "fixed" here. I take code in a design as pseudocode, even if it is written in rust

docs/tee-lifecycle.md
gilcu3, line 1: This one is a nice fix! Afaik this is one we want to keep up to date, and useful for external and internal readers

crates/node/src/run.rs
```diff
@@ -394,16 +367,19 @@ where
...
-        }
-    });
+    let mut attestation_interval = tokio::time::interval(ATTESTATION_RESUBMISSION_INTERVAL);
+    // A failed submission can retry internally for longer than the interval; skip the missed
+    // ticks instead of bursting stale submissions afterwards
+    attestation_interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+    tokio::spawn(periodic_attestation_submission(
+        submitter.clone(),
+        attestation_interval,
+    ));
```
gilcu3: It would look a bit cleaner IMO to define the attestation interval in a function `crates/node/src/tee/remote_attestation.rs`. Does not need to be `periodic_attestation_submission`, but something that calls that, so that we keep the same testability.

crates/node/src/tee/remote_attestation.rs
```diff
@@ -406,54 +354,37 @@ mod tests {
...
+            })
         }
     }
 
-    /// Simulates contract behavior by automatically adding the node back to TEE accounts
-    /// when an attestation submission occurs, mimicking real contract response to successful submissions.
-    struct ContractSimulator {
-        sender: watch::Sender<Vec<NodeId>>,
-        node_id: NodeId,
-    }
```
gilcu3, line 440: I am not sure we should remove such simulator. Could you share your thinking on why this was not useful and can be switched by a boolean?

crates/node/src/tee/remote_attestation.rs
```diff
@@ -406,54 +354,37 @@ mod tests {
...
-            Box::pin(async { Ok(None) })
-        }
+    struct StubAttestationExpiryReader {
+        fail: bool,
     }
 
-    struct FailingAttestationExpiryReader;
-
-    impl crate::indexer::ReadAttestationExpiry for FailingAttestationExpiryReader {
+    impl crate::indexer::ReadAttestationExpiry for StubAttestationExpiryReader {
```
gilcu3: nit: we could do the usual
```
use crate::indexer::ReadAttestationExpiry;
```

crates/node/src/tee/remote_attestation.rs
```diff
@@ -132,103 +134,101 @@ fn validate_remote_attestation(
...
+            *self.account_public_key.as_bytes(),
+        )
+        .into();
+        let result = self.tee_authority.generate_attestation(report_data).await;
+        crate::metrics::MPC_TEE_ATTESTATION_ATTEMPTS_TOTAL
+            .with_label_values(&[outcome_label(result.is_ok())])
+            .inc();
+        let attestation = match result {
+            Ok(attestation) => attestation,
+            Err(error @ AttestationError::CollateralFetch(_)) => {
```
gilcu3, line 159: I didn't know about that syntax, or at least had forgotten about it

crates/node/src/tee/remote_attestation.rs
```diff
@@ -132,103 +134,101 @@ fn validate_remote_attestation(
...
+            .read_stored_attestation_expiry(&self.tls_public_key)
             .await
         {
             Ok(baseline) => baseline, // None just means nothing stored yet (e.g. first submit)
             // Submit anyway on a read error: refreshing the attestation is the priority, and a
             // broken read must not block submission (the confirmation just can't use a baseline).
             Err(error) => {
-                tracing::warn!(%error, "could not read pre-submit attestation baseline; submitting without it");
+                tracing::warn!(
+                    ?error,
```
gilcu3, line 185: why changing from `%error` to `?error`, was the previous worse?

gilcu3 (commented):
Looks good for what it was trying to do. Left some minor comments, but will come back after some slack discussion, as I am not sure about the approach

gilcu3 (dismissed):
Thank you! I think all my comments were already there :)

crates/node/src/tee/remote_attestation.rs
pbeza: In some code paths the error is apparently more detailed with `?`. Claude fixed it while deduping the two copies of this code. Details (by Claude):
>  [the actor-send failure](https://github.com/near/mpc/blob/38e27a20101e322ca5d183f0e269cdfbb7952fc0/crates/node/src/indexer.rs#L241-L245):
> https://github.com/near/mpc/blob/38e27a20101e322ca5d183f0e269cdfbb7952fc0/crates/node/src/indexer.rs#L241-L245
> loses `Closed`/`Timeout`/`Dropped` under `%`, and [the deserialize failure](https://github.com/near/mpc/blob/38e27a20101e322ca5d183f0e269cdfbb7952fc0/crates/node/src/indexer.rs#L247-L251)
> https://github.com/near/mpc/blob/38e27a20101e322ca5d183f0e269cdfbb7952fc0/crates/node/src/indexer.rs#L247-L251
> loses the `serde` detail, which is the whole diagnostic there. On the other paths (a bare `QueryError`, which carries no `#[source]`, and the `bail!`) both render identically.

Relevant call trace:
```
generate_and_submit                                        remote_attestation.rs:169
 └─ attestation_reader: Arc<dyn ReadAttestationExpiry>
     └─ RealAttestationExpiryReader::read_stored_attestation_expiry     indexer.rs:412
         └─ view_client.get_participant_attestation(contract_id, tls_pk)  indexer.rs:218
              • QueryRequest::CallFunction { GET_ATTESTATION } @ Finality::Final
              • deserialize Option<VerifiedAttestation>
         └─ .and_then(|a| a.expiry_timestamp_seconds())
```

gilcu3: oh that's complicated

crates/node/src/tee/remote_attestation.rs
pbeza: Yeah, I (or my AI agents) wasn’t using it much either until I started insisting on keeping the code nice, clean, clear, idiomatic, self-explanatory, and as concise as possible. 😄

crates/node/src/tee/remote_attestation.rs
pbeza: https://github.com/near/mpc/pull/4201/changes/827ac94447224984615dcd5cb08b0a0d627ec8be ✅

docs/securing-mpc-with-tee-design-doc.md
pbeza: Yeah, I agree there’s no need to keep updating the design docs. This one, as you might guess, was just refactored automatically by Claude while updating the docs, so I’d leave it since it doesn’t hurt to have it updated. But going forward, I’ll try to make sure not to update the design docs. 🫡

docs/hot-tee-signing-design.md
pbeza: Ditto: https://github.com/near/mpc/pull/4201#discussion_r3873378656.

crates/node/src/metrics.rs
pbeza: Agree. I already trimmed a lot of redundant comments produced by Claude in this PR, but I missed this one. Perhaps I should just disallow Claude from adding comments and only add them retrospectively. 🤔 Anyway, fixed: https://github.com/near/mpc/pull/4201/changes/e64d94550531ba20be5f5c1f7f7e922528274d97. ✅

docs/backup-service-attestation-data.md
pbeza: Oh, I missed that it was a doc change for the backup service. Thanks for catching it! Reverted: https://github.com/near/mpc/pull/4201/commits/71e48340033a70be897e5d516403f00cc14dea25. ✅

docs/tee-context-design.md
pbeza: Ditto: https://github.com/near/mpc/pull/4201#discussion_r3873378656.

crates/node/src/run.rs
pbeza: Love the idea. Done: https://github.com/near/mpc/pull/4201/commits/c82903f50cff7084cbbd1bf73a61a8010202ac98. ✅

barakeinav1 (approved):
I did only a shallow review. but LGTM

crates/node/src/tee/remote_attestation.rs
pbeza: I remember I [originally introduced](https://github.com/near/mpc/blame/8a5725a41f7319a70ab9f746b5d521dc25db95c8/crates/node/src/tee/remote_attestation.rs#L475-L477) `ContractSimulator` in https://github.com/near/mpc/pull/1223 to simulate putting the node back into the TEE accounts list after a submission:

https://github.com/near/mpc/blob/8a5725a41f7319a70ab9f746b5d521dc25db95c8/crates/node/src/tee/remote_attestation.rs#L475-L477

After the refactor, we get the same thing from:

https://github.com/near/mpc/blob/c82903f50cff7084cbbd1bf73a61a8010202ac98/crates/node/src/tee/remote_attestation.rs#L502-L505

The `bool` is not really a replacement for it. `MockSender` had no way to fail before: [`send`](https://github.com/near/mpc/blob/8a5725a41f7319a70ab9f746b5d521dc25db95c8/crates/node/src/tee/remote_attestation.rs#L469-L483) always returned `Ok`, and [`send_and_wait`](https://github.com/near/mpc/blob/8a5725a41f7319a70ab9f746b5d521dc25db95c8/crates/node/src/tee/remote_attestation.rs#L485-L491) always returned `Executed`, so [`submit_remote_attestation`](https://github.com/near/mpc/blob/c82903f50cff7084cbbd1bf73a61a8010202ac98/crates/node/src/tee/remote_attestation.rs#L58-L117) could never reach its [retry timeout](https://github.com/near/mpc/blob/c82903f50cff7084cbbd1bf73a61a8010202ac98/crates/node/src/tee/remote_attestation.rs#L114). This timeout is what the three new tests need ([`should_survive_submission_retry_timeout`](https://github.com/near/mpc/blob/c82903f50cff7084cbbd1bf73a61a8010202ac98/crates/node/src/tee/remote_attestation.rs#L561-L579) and the two [`should_retry_failed_resubmission_*`](https://github.com/near/mpc/blob/c82903f50cff7084cbbd1bf73a61a8010202ac98/crates/node/src/tee/remote_attestation.rs#L581-L628)), so we needed some new way to simulate a failed submission.

We could probably put this `bool` into `ContractSimulator` and keep it.

gilcu3 (approved):

PR #4203, opened 2026-08-20
feat(node): narrow verify-foreign-tx leader selection to chain supporters

### LINKED ISSUE #3961: Narrow verify-foreign-tx leader selection to chain supporters

anodar:
closes: https://github.com/near/mpc/issues/3961


anodar:
@claude code review

crates/node/src/foreign_chain_policy.rs
```diff
@@ -129,6 +131,37 @@ fn resolve_participant_ids(
...
+        ForeignChainLeadersRefiner {
+            supporters_receiver,
+        }
+    }
+}
+
+impl RefineEligibleLeaders<VerifyForeignTxRequest> for ForeignChainLeadersRefiner {
+    fn refine(
+        &self,
+        request: &VerifyForeignTxRequest,
```
anodar, line 161: Note, while we only use chain from the request here, we define trait itself with request parameter anyway. Because 

```
pub(super) refine_eligible_leaders: Option<Box<dyn RefineEligibleLeaders<RequestType>>>
```

is defined in generic `PendingRequests` type. I found this approach cleanest.

haiyuechen-nearone (commented):
I think we should also update the docs about leader selection is chain support aware right?

crates/node/src/requests/metrics.rs
```diff
@@ -238,3 +238,15 @@ pub static MPC_CLUSTER_FAILED_SIGNATURES_COUNT: LazyLock<prometheus::IntCounterV
...
         .unwrap()
     });
+
+pub static MPC_NUM_REQUESTS_WITHOUT_REFINED_LEADER_TOTAL: LazyLock<prometheus::IntCounterVec> =
+    LazyLock::new(|| {
+        prometheus::register_int_counter_vec!(
+            "mpc_num_requests_without_refined_leader_total",
+            "Number of queue passes where a request had eligible leaders, but none was allowed \
+             by the eligible-leaders refiner (e.g. no participant supports the request's \
+             foreign chain), so the request stays parked",
```
haiyuechen-nearone, line 248:
```suggestion
            "Number of requests left "parked" because no eligible leader supports the requested foreign chain. Counted once per request per queue pass.",
```

Would this be better?

crates/node/src/foreign_chain_policy.rs
```diff
@@ -129,6 +131,37 @@ fn resolve_participant_ids(
...
+        eligible: &HashSet<ParticipantId>,
+    ) -> HashSet<ParticipantId> {
+        match self
+            .supporters_receiver
+            .borrow()
+            .get(&request.request.chain())
+        {
+            Some(supporters) => supporters & eligible,
```
haiyuechen-nearone, line 169: Should we select a leader even if we don't have enough online participants to satisfy the threshold? Here a node being eligible doesn't necessary mean that the node supports the chain requested right?

If I understand correctly, we will elect a leader and start attempts on a request, even if we only have a single node online that is registered as a supporter of the chain requested.

haiyuechen-nearone (commented):
Left a couple of comments, Im on the fence on whether they are blocking, @anodar PTAL

Also I think we should also update the docs about leader selection is chain support aware right?

crates/node/src/requests/metrics.rs
```diff
@@ -238,3 +238,15 @@ pub static MPC_CLUSTER_FAILED_SIGNATURES_COUNT: LazyLock<prometheus::IntCounterV
         )
         .unwrap()
     });
+
+pub static MPC_NUM_REQUESTS_WITHOUT_REFINED_LEADER_TOTAL: LazyLock<prometheus::IntCounterVec> =
```
haiyuechen-nearone, line 242: Can we key this metric with which chain and request type? We can easily see that the network is struggling without needing to dig into logs that way.

crates/node/src/requests/metrics.rs
anodar: Refiner is generic that's why I avoided taking concrete case of it "supports the requested foreign chain".  It will refine given the list, it just happens that right now we only use it for supported chains.

I followed existing generic wording convention we have for aliveness filtering.

crates/node/src/foreign_chain_policy.rs
anodar: This followed signature and ckd requests, we don't check whether we have quorum number of alive participants when selecting the leader either. But I guess makes sense to diverge from that here as we require additional condition from participants, added it.

crates/node/src/requests/metrics.rs
anodar: Not sure what do you mean by keying the metric with request type but we do label this metric by request type already.  Also this metric only applies to foreign chain request validation requests anyway (as that's the only one that has eligible leaders refiner). 

Issue with adding the chain is that this is used in `get_requests_to_attempt` which uses generic `RequestType` and makes it pretty awkward to extract the chain. We can do it, but I'm not sure we gain whole lot by it.

crates/node/src/providers/verify_foreign_tx/sign.rs
```diff
@@ -94,8 +94,9 @@ impl VerifyForeignTxProvider {
             snapshot.get(&requested_chain).cloned().unwrap_or_default()
         };
 
-        // If we don't support the foreign chain, we can't lead the computation.
-        // TODO(#3961): narrow leader selection to only chain supporters.
+        // Leader selection already narrows to chain supporters. Re-check as
+        // defense-in-depth, since the supporters snapshot may have changed
+        // between leader selection and this attempt.
         let my_participant_id = self.ecdsa_signature_provider.my_participant_id();
```
gilcu3, line 100: comment talking about code history is not very useful. We should reduce to explain the remaining code

crates/node/src/providers/verify_foreign_tx/sign.rs
gilcu3: related to the comment above. If this is just defense in depth, it would be nice to make the message inline with that, so that if we ever find it in some logs we look deeper on how could that happen

crates/node/src/requests/metrics.rs
gilcu3: If this metric is only about foreign chains, can we name it appropriately? We have no idea if we will need more "leader refiners", so keeping it to the point now seems an advantage, that way we avoid introducing generalizations before they are needed.

crates/node/src/requests.rs
```diff
@@ -1,4 +1,5 @@
 pub mod queue;
 
+pub(crate) mod metrics;
+
 mod debug;
-mod metrics;
```
gilcu3, line 4: I think this is not correct, it means we are now leaking data from the module to the rest of the crate. Why did we need this?

crates/node/src/requests/debug.rs
```diff
@@ -189,8 +189,16 @@ impl<RequestType: Request + Clone, ChainRespondArgsType: ChainRespondArgs> Debug
         let indexer_heights = self.network_api.indexer_heights();
 
         for request in self.requests.values() {
-            let debug_line =
-                request.debug_print(&self.clock, self.my_participant_id, &eligible_leaders);
+            let request_eligible_leaders = super::queue::eligible_leaders_for(
```
gilcu3: nit: better not use long inline statements such as `super::queue::...`

crates/node/src/requests/debug.rs
```diff
@@ -189,8 +189,16 @@ impl<RequestType: Request + Clone, ChainRespondArgsType: ChainRespondArgs> Debug
...
+                self.refine_eligible_leaders.as_deref(),
+                &request.request,
+                &eligible_leaders,
+            );
```
gilcu3: this seems a bit wasteful, but give we don't have that many requests or participants could be fine

crates/node/src/requests/queue.rs
```diff
@@ -71,6 +95,9 @@ pub struct PendingRequests<RequestType: Request, ChainRespondArgsType: ChainResp
 
     /// Recently completed requests, for debugging purposes only.
     pub(super) recently_completed_requests: CompletedRequests<RequestType, ChainRespondArgsType>,
+
+    /// See [`RefineEligibleLeaders`].
+    pub(super) refine_eligible_leaders: Option<Box<dyn RefineEligibleLeaders<RequestType>>>,
```
gilcu3: do we need this boxed dynamic dispatch here? IIRC @netrome usually complains about those, see this [comment](https://github.com/near/mpc/pull/1837#pullrequestreview-3697343613) for example

crates/node/src/providers/verify_foreign_tx.rs
```diff
@@ -161,6 +162,9 @@ pub struct VerifyForeignTxProvider {
     config: Arc<ConfigFile>,
     inspectors: ForeignChainInspectors<HttpClient>,
     supporters_by_foreign_chain: watch::Receiver<SupportersByForeignChain>,
+    /// [`foreign_tx_reconstruction_threshold`](crate::foreign_chain_policy::foreign_tx_reconstruction_threshold)
+    /// of the running domains; `None` when there is no ForeignTx domain.
+    foreign_tx_quorum: Option<ReconstructionThreshold>,
```
gilcu3: I am a bit puzzled about the name. Why not calling it reconstruction threshold? Quorum is a term we use to enforce the minimum number of RPCs, so this is confusing

crates/node/src/requests/queue.rs
```diff
@@ -37,6 +38,29 @@ const MAX_LATENCY_BEFORE_EXPECTING_TRANSACTION_TO_FINALIZE: Duration = Duration:
...
+pub(super) fn eligible_leaders_for<'a, RequestType>(
+    refiner: Option<&dyn RefineEligibleLeaders<RequestType>>,
+    request: &RequestType,
+    eligible: &'a HashSet<ParticipantId>,
+) -> Cow<'a, HashSet<ParticipantId>> {
+    match refiner {
+        Some(refiner) => Cow::Owned(refiner.refine(request, eligible)),
+        None => Cow::Borrowed(eligible),
+    }
+}
```
gilcu3: I generally prefer to avoid Cow unless a need for performance is granted. Apparently we wouldn't gain much here as an extra clone in the eligible set is a relatively fast operation compared to the rest. I added this [commit](https://github.com/near/mpc/commit/c1fce150576f7ba1558d9efabf41db5aaf768eea) with how it would look like without it, and seems a bit cleaner. Feel free to use it if you agree

crates/node/src/providers/verify_foreign_tx.rs
```diff
@@ -191,11 +196,20 @@ impl VerifyForeignTxProvider {
...
         })
     }
 
+    pub(crate) fn supporters_by_foreign_chain(&self) -> watch::Receiver<SupportersByForeignChain> {
+        self.supporters_by_foreign_chain.clone()
+    }
+
+    pub(crate) fn foreign_tx_quorum(&self) -> Option<ReconstructionThreshold> {
+        self.foreign_tx_quorum
+    }
```
gilcu3: It seems very strange to add these here, but then use them only in `mpc_client` to construct the refiner
```
        let eligible_leaders_refiner = ForeignChainLeadersRefiner::new(
            self.verify_foreign_tx_provider
                .supporters_by_foreign_chain(),
            self.verify_foreign_tx_provider.foreign_tx_quorum(),
        );
```

Maybe could be solved by instead letting the provider construct the refiner directly, if that's the aim?

crates/node/src/coordinator.rs
```diff
@@ -703,8 +703,9 @@ where
                 // and remain after the reshare supports it. With no ForeignTx
                 // domain nothing can be available, so the resolver isn't
                 // spawned and the provider sees a constant empty map.
+                let foreign_tx_quorum = foreign_tx_reconstruction_threshold(&running_state.domains);
                 let (supporters_by_foreign_chain, _supporters_resolver_task) =
-                    match foreign_tx_reconstruction_threshold(&running_state.domains) {
+                    match foreign_tx_quorum {
```
gilcu3: here it also seems off to call this quorum

crates/node/src/tests/verify_foreign_tx_gating.rs
```diff
@@ -198,11 +200,15 @@ async fn verify_foreign_tx__should_only_be_served_while_chain_is_available() {
...
     assert!(
-        crate::metrics::MPC_NUM_VERIFY_FOREIGN_TX_UNAVAILABLE_CHAIN_REJECTIONS.get()
-            > rejections_before,
-        "expected the availability gate to reject at least one attempt"
+        crate::requests::metrics::MPC_NUM_REQUESTS_WITHOUT_REFINED_LEADER_TOTAL
+            .with_label_values(&["verify_foreign_tx"])
+            .get()
+            > parked_before,
+        "expected the request to park with no refined eligible leader"
     );
```
gilcu3, line 213: are we losing some test coverage here, or these metrics are now a bit redundant?

gilcu3 (dismissed):
This LGTM, although a bit borderline, but I couldn't find any blocker (IMO). I think there are several patterns that could be improved, left some comments in that direction.

crates/node/src/providers/verify_foreign_tx/sign.rs
anodar: The comment is not about code history. It's about call flow:

Leader selection in `get_requests_to_attempt` (`monitor_block_updates` -> `pending_verify_foreign_txs.get_requests_to_attempt`) we already take request that this node can lead (supports the chain from this request). 

But we still check it here in `make_verify_foreign_tx_leader` (`monitor_block_updates` ->`compute_verify_foreign_tx_response` ->  `make_verify_foreign_tx_leader`) as defense-in-depth as supporters snapshot could have already changed.

gilcu3: Ahh I see, yeah that's fine then, I misinterpreted the comment (the use of narrow :sweat_smile: )

crates/node/src/requests/metrics.rs
anodar: It's not metric specific for foreign chains, I just noted that so far it only applies to foreign chain requests. Function that it's used in is generic to request types and metric is also generic. 

If we change the name to be specific to foreign chains, we either 
- leave the site where it's used as is and then it's not clear at that it's only used for foreign chains (not ideal), or
- somehow make it clear when doing metric::inc that it's specifically for request type VerifyForeignTx (tricky to make it clean as the refining itself is generic and not specific to VerifyForeignTx)

Neither of above approaches seem better than existing proposal and we have label on the metric which I believe (at least partially) addresses @haiyuechen-nearone's concern.

gilcu3: yeah the current metric is about the queue (I didn't realize that when I wrote the comment), which is kind of oblivious to the foreign chains, so changing the name as I suggested would be wrong in that regard :(

crates/node/src/requests.rs
anodar: We need it because we are using it in `crates/node/src/tests/verify_foreign_tx_gating.rs` tests. 

Before this change that test used:

```
crate::metrics::MPC_NUM_VERIFY_FOREIGN_TX_UNAVAILABLE_CHAIN_REJECTIONS.get();
```

which works fine because it leaves in `crates/node/src/metrics.rs` whereas new metric leaves in `crates/node/src/requests/metrics.rs` as it should belong in requsets not in root crate.

I'm not sure I follow what is your concern about making request metrics public in the rest of the crate. Can you please elaborate?

crates/node/src/requests/debug.rs
anodar: Yeah I figured it might be useful to have this for debugging as it shouldn't be a performance concern anytime soon.

crates/node/src/requests.rs
gilcu3: It was just about not making something more open than it should be, as I couldn't figure where the metric was being used. So that test currently classifies as "unit test" (because it is inside the crate), but cannot be inside some module because it accesses several ones? Might be ok then, sorry for the noise

crates/node/src/coordinator.rs
anodar: Right, addressed that too in [cbf8c86](https://github.com/near/mpc/pull/4203/commits/cbf8c86fd95f25dffa47ef72f87cab50d0ec7729)

crates/node/src/providers/verify_foreign_tx.rs
anodar: Renamed to threshold in [cbf8c86](https://github.com/near/mpc/pull/4203/commits/cbf8c86fd95f25dffa47ef72f87cab50d0ec7729)

crates/node/src/requests/debug.rs
anodar: Addressed.

crates/node/src/providers/verify_foreign_tx/sign.rs
anodar: Reworded the bail message.

crates/node/src/tests/verify_foreign_tx_gating.rs
gilcu3: I mean `MPC_NUM_VERIFY_FOREIGN_TX_UNAVAILABLE_CHAIN_REJECTIONS` when I said "these metrics"

crates/node/src/requests/queue.rs
anodar: I could have sworn that at some point in the huge pr I've sent out @kevindeforth pointed out we should use Cow instead of clone but I can't seem to find it anymore :/  Maybe I'm hallucinating.

Addressed in: [2e75931](https://github.com/near/mpc/pull/4203/commits/2e75931fea13bf1cba15391d6347afdf97293b70)

crates/node/src/requests.rs
anodar: Yes, it would be pretty tricky to move it inside requests crate. Not a noice, thanks for bringing it up.

crates/node/src/providers/verify_foreign_tx.rs
anodar: Goal was to construct them in mpc_client rather than in foreign chain tx provider. Semantically doesn't seem to belong there imo. But it does make code bit cleaner if we do it there. So changed to your proposal in [c93ee01](https://github.com/near/mpc/pull/4203/commits/c93ee01b55b7e5729cb19b81424e6ca2c1fd5123)

crates/node/src/tests/verify_foreign_tx_gating.rs
anodar: We're not losing the coverage: before this PR leader could be selected that would never be able to succeed and hit `MPC_NUM_VERIFY_FOREIGN_TX_UNAVAILABLE_CHAIN_REJECTIONS.` metric.

Now we're gating it with check that selected leader supports the chain but it could still (theoretically) happen that after that check supporters list changes and leader drops support (due to some incredible timing) and then that `MPC_NUM_VERIFY_FOREIGN_TX_UNAVAILABLE_CHAIN_REJECTIONS` metric would be hit.

It's pretty tricky and arguably not worth the effort to test that MPC_NUM_VERIFY_FOREIGN_TX_UNAVAILABLE_CHAIN_REJECTIONS metric.

crates/node/src/requests/metrics.rs
haiyuechen-nearone: Sorry I didn't realize that this was labeled by request type already, then I guess we are good on that front.

On label this per chain, I do think that we are getting a lot more value from this metric if we do it. My thinking is that with just a single count we cannot tell from metrics which chain has degraded performance due to refiner issues. Labeling it tells us directly that "chain x is having issues right now", or even tell apart when multiple chains are suffering.

crates/node/src/requests/queue.rs
anodar: We need that because this defined in generic PendingRequests struct. 

Cleanest way I see that we could avoid that box dyn is to use associated type on Request, and set it to no op (NoRefinement) type for other request types besides Foreign chain tx request. Sketch for illustration: https://github.com/near/mpc/pull/4237

Although I'm not sure that's better, particularly refiner in Request trait seems out of place. Let me know what you think.

gilcu3: Added some comments in https://github.com/near/mpc/pull/4237, but overall it does look more intuitive, as now it is clear that the refinement concept is a queue concept, therefore needed for all requests using the queue (the `NoRefinement` struct shows this perfectly)

gilcu3: It seems another alternative to #4237 is to do something like:
```rust
pub struct PendingRequests<RequestType, ChainRespondArgsType, Refiner>
  where
      RequestType: Request,
      ChainRespondArgsType: ChainRespondArgs,
      Refiner: RefineEligibleLeaders<RequestType>,
```

netrome: Yeah looking at the code a bit, I don't see any need for dynamic dispatch here. I could be missing something, but it seems to me that just using pure generics should be sufficient here. Will prepare a small suggestion and see if I hit any walls.

netrome: Trad-coded a suggestion of just using generics: https://github.com/near/mpc/pull/4249 this should be conceptually closer to the existing code - although I don't understand the change enough to judge whether an associated type or plain generics is the best way to go.

However, this is sufficient to convince me that dynamic dispatch is redundant here - so whatever you do I suggest avoiding it.

netrome: While we're on the topic of nits and typical things that I note in reviews - this struct has some trait bounds. We should get rid of them if possible (would be nice as a broken-window fix). Structs should not have trait bounds.

See https://stackoverflow.com/questions/49229332/should-trait-bounds-be-duplicated-in-struct-and-impl for more info on this.

anodar: > pub struct PendingRequests<RequestType, ChainRespondArgsType, Refiner>

That is an alternative I considered but did not want to change type (add another generic) of PendingRequests. IMO having too many generics makes code less readable. But if we're set on avoiding box dyn whenever possible, I'd still go this way, makes rest of the code cleaner on the expense of additional generic.

Changed to your alternative proposal rather then my sketched https://github.com/near/mpc/pull/4237 in [acf6433](https://github.com/near/mpc/pull/4203/commits/acf64339b445900f82e878d76afc2bc58ad598cb)

gilcu3 (approved):

haiyuechen-nearone (approved):

crates/node/src/requests/metrics.rs
anodar: Opened an issue here for it: https://github.com/near/mpc/issues/4262 as neither possible solutions we discussed seem ideal so far and this probably requires bit more thought to be done cleanly.

PR #4205, opened 2026-08-21
feat: expose finality status for e2e test framework and unify syntax with codebase

### LINKED ISSUE #4204: feat: enable finality status e2e test frameworks account caller and unify with sandbox and devnet
### Background

In the e2e test framework, we currently default to `ExecutedOptimistic` as expected transaction status sent over the near-kit back-end.

If we need anything other than that, our current approach is to hard-code the finality level in a helper function:
https://github.com/near/mpc/blob/236c27e7883adfba7b681c7570f0d4674f4cc494/crates/e2e-tests/src/blockchain.rs#L87-L102

With the current design, we would need to write one helper-method for each desired finality level.

Fortunately, there is a simpler way. Near-kit has a trait `WaitLevel` and types for each finality. This allows us to implement the `CallContract` trait over the generic finality level:

```rust
impl<T> CallContract for NearKitCaller<T>
where
    T: WaitLevel,
{
    type Output = T::Response;
    type Error = near_kit::Error;

    async fn call_contract(
        &self,
        contract_id: &near_kit::AccountId,
        call_args: FunctionCallArgs,
    ) -> Result<Self::Output, Self::Error> {
        self.inner
            .call(contract_id, &call_args.method_name)
            .args_raw(call_args.args)
            .gas(call_args.gas)
            .deposit(call_args.deposit)
            .wait_until::<T>()
            .await
    }
}
```

Additionally, we should use the opportunity to unify the e2e syntax with what we currently have in the sandbox and devnet:

Instead of:

```rust
self.contract
            .handle_for(self.operator_client_for(node_index)?)
            .start_node_migration(destination_node_info)
            .await
            .context("failed to start node migration")
```

we would prefer to write:

```rust
 self.operator_client_for(node_index)?
            .call_mpc(self.contract_id())
            .start_node_migration(destination_node_info)
            .await
            .context("failed to start node migration")
``` 

This keeps our call-syntax consistent across the code-base.

We will be able to chose transaction-finality status at the call-site:
```
 self.operator_client_for(node_index)?
            .call_mpc(self.contract_id())
            .start_node_migration(destination_node_info)
            .with_finality(Final)
            .await
            .context("failed to start node migration")
```

### User Story

As a developer, writing and reviewing code is much easier if syntax is consistent across the code-base.

### Acceptance Criteria

- implement a `NearKitCaller<T>` for generic finality status `T` in the e2e test-framework
- refactor the syntax to match what we currently have in the sandbox and devnet.

_No response_

kevindeforth:
resolves https://github.com/near/mpc/issues/4204

### DIFF
--- a/crates/e2e-tests/src/caller.rs
+++ b/crates/e2e-tests/src/caller.rs
@@ -11,12 +11,12 @@ pub struct NearKitCaller<T> {
 }
 
 pub trait CallMpc: Sized {
-    fn call_mpc(self, contract_id: near_account_id::AccountId) -> MpcContractHandle<Self>;
+    fn call_mpc(self, contract_id: &near_account_id::AccountId) -> MpcContractHandle<Self>;
 }
 
 impl<T> CallMpc for NearKitCaller<T> {
-    fn call_mpc(self, contract_id: near_account_id::AccountId) -> MpcContractHandle<Self> {
-        MpcContractHandle::new(self, contract_id)
+    fn call_mpc(self, contract_id: &near_account_id::AccountId) -> MpcContractHandle<Self> {
+        MpcContractHandle::new(self, contract_id.clone())
     }
 }
 
--- a/crates/e2e-tests/src/caller.rs
+++ b/crates/e2e-tests/src/caller.rs
@@ -1,0 +1,71 @@
+use std::marker::PhantomData;
+
+use near_contract_transport::{CallContract, FunctionCallArgs};
+use near_kit::WaitLevel;
+use near_mpc_contract_interface::client::MpcContractHandle;
+
+/// A [`near_kit::Near`] client bound to a specific account
+pub struct NearKitCaller<T> {
+    pub(crate) inner: near_kit::Near,
+    pub(crate) _wait_level: PhantomData<fn() -> T>,
+}
+
+pub trait CallMpc<T> {
+    fn call_mpc(
+        self,
+        contract_id: near_account_id::AccountId,
+    ) -> MpcContractHandle<NearKitCaller<T>>;
+}
+
+impl<T> CallMpc<T> for NearKitCaller<T> {
+    fn call_mpc(
+        self,
+        contract_id: near_account_id::AccountId,
+    ) -> MpcContractHandle<NearKitCaller<T>> {
+        MpcContractHandle::new(self, contract_id)
+    }
+}
+
+impl<T> CallContract for NearKitCaller<T>
+where
+    T: WaitLevel,
+{
+    type Output = T::Response;
+    type Error = near_kit::Error;
+
+    async fn call_contract(
+        &self,
+        contract_id: &near_kit::AccountId,
+        call_args: FunctionCallArgs,
+    ) -> Result<Self::Output, Self::Error> {
+        self.inner
+            .call(contract_id, &call_args.method_name)
+            .args_raw(call_args.args)
+            .gas(call_args.gas)
+            .deposit(call_args.deposit)
+            .wait_until::<T>()
+            .await
+    }
+}
+
+impl<T> NearKitCaller<T> {
+    pub(crate) fn with_wait_level<U>(self) -> NearKitCaller<U> {
+        NearKitCaller {
+            inner: self.inner,
+            _wait_level: PhantomData,
+        }
+    }
+}
+
+/// Allows to change the finality level the transaction is required to achieve before
+/// a call returns.
+/// Note that return type of the contract handle may change depending on the wait level.
+pub trait WithWaitLevel {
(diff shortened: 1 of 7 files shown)


kevindeforth:
@claude addressed your comment, except for the "unused structs / traits / generics" ones, as we will have a stacked PR shortly for those. Please re-view.

gilcu3:
> @claude addressed your comment, except for the "unused structs / traits / generics" ones, as we will have a stacked PR shortly for those. Please re-view.

@claude review, taking into account the comment above

crates/e2e-tests/tests/parallel_sign_calls.rs
```diff
@@ -83,7 +83,7 @@ async fn mpc_cluster_should_successfully_process_parallel_requests() {
         .call(
             "make_parallel_sign_calls",
             json!({
-                "target_contract": cluster.contract.contract_id(),
+                "target_contract": cluster.contract.account_id().as_str(),
```
gilcu3, line 86: is this change needed?

crates/e2e-tests/src/cluster.rs
```diff
@@ -801,9 +805,13 @@ impl MpcCluster {
         self.blockchain.client_for(account_id.as_ref(), key)
     }
 
-    pub fn contract_handle(&self, account_id: &AccountId) -> MpcContractHandle<NearKitCaller> {
-        self.contract
-            .handle_for(self.client_for(account_id).unwrap())
+    pub fn contract_handle(
+        &self,
+        account_id: &AccountId,
+    ) -> MpcContractHandle<NearKitCaller<ExecutedOptimistic>> {
```
gilcu3, line 811: I know the type was the same, just that now it is evident. But why do we use `ExecutedOptimistic` in most places, instead of `Final`?

crates/e2e-tests/src/caller.rs
```diff
@@ -0,0 +1,65 @@
...
+
+pub trait CallMpc: Sized {
+    fn call_mpc(self, contract_id: near_account_id::AccountId) -> MpcContractHandle<Self>;
+}
+
+impl<T> CallMpc for NearKitCaller<T> {
+    fn call_mpc(self, contract_id: near_account_id::AccountId) -> MpcContractHandle<Self> {
+        MpcContractHandle::new(self, contract_id)
+    }
+}
```
gilcu3, line 21: We have a lot of `.clone()` in the `call_mpc` calls in the PR. Wouldn't it be worth taking a ref in `call_mpc` then doing the clone internally, if needed? Not sure it is an anti-pattern, but there should be an ergonomic way of achieving this

gilcu3: Ahh apparently in devnet and sandbox we do take a ref there, so just to make it homogeneous it might be good to do it here as well

gilcu3 (dismissed):
Thank you!

crates/e2e-tests/tests/parallel_sign_calls.rs
kevindeforth: Do you mean renaming to `account_id()`? I thought that was consistent.
If you refer to returning a `AccountId` type instead of a string, yeah, I need the account-id in other places and this does make it more consistent with other places.
I didn't want to keep both `contract_id() -> String` and `account_id() -> AccountId`, as it felt odd. We can use one.

crates/e2e-tests/src/cluster.rs
kevindeforth: It's faster I think. `Final` is required if you want to make a contract call and expect the next view-call to return the updated result. We don't need that in most places.

crates/e2e-tests/src/caller.rs
kevindeforth: good catch, I thought that was odd :man_facepalming: . Changed in: https://github.com/near/mpc/pull/4205/commits/3a65ff1b4d21938e68e39b6ff0577798602e5363

crates/e2e-tests/tests/parallel_sign_calls.rs
gilcu3: Ahh now I see it was a bit ambiguous when reading just this comment. I just meant the `as_str` part, as maybe the type already converts automagically in this case

gilcu3 (approved):

haiyuechen-nearone (approved):

crates/e2e-tests/tests/parallel_sign_calls.rs
kevindeforth: hmm, I remember trying that and having rust-analyzer shout at me, so I did the `.as_str()`.

PR #4206, opened 2026-08-21
refactor(contract): move foreign-chain support and provider policy

### LINKED ISSUE #1962: refactor: Break out feature-specific contract methods and tests to their own modules

gilcu3:
part 12b/12 Closes #1962

only 96 lines left in lib.rs :smiley:


crates/contract/src/api.rs
```diff
@@ -5,6 +5,7 @@ pub mod authorization;
 pub mod ckd;
 pub mod common;
 pub mod foreign_chain;
+pub mod foreign_chain_support;
```
gilcu3, line 8: originally I had merged these two, but the file was already huge. So now it is separated. Mod naming could be improved. If reviewers have any good suggestion I am happy to take it

haiyuechen-nearone: maybe `foreign_chain_coverage`?

haiyuechen-nearone (approved):

crates/contract/src/api.rs
gilcu3: The problem is that the first mod now appears to be more general than the second. So some other organization/naming between the two would be needed. `foreign_chain_coverage` works the same as `foreign_chain_support` for me.

kevindeforth (approved):

PR #4207, opened 2026-08-21
refactor: add typed init call to mpc contract interface

### LINKED ISSUE #3949: refactor: typed `init` on `MpcContractHandle`
### Background

Lets move the `init` method to the `MpcContractHandle`

### User Story

As a dev on our code-base, I like when things are easy to integrate.

### Acceptance Criteria

- [ ] sandbox, devnet and our e2e tests use the typed init call instead of assembling the call themselves

kevindeforth:
resolves https://github.com/near/mpc/issues/3949


crates/e2e-tests/src/cluster.rs
```diff
@@ -1330,16 +1318,22 @@ async fn init_contract(
...
-        .call_final(
-            method_names::INIT,
-            json!({ "parameters": parameters_json, "init_config": init_config }),
-        )
-        .await?;
+    let outcome = match init_format {
+        ContractInitFormat::Current => {
+            contract
+                .client()
+                .call_mpc(contract.account_id())
```
kevindeforth, line 1343: claude pointed this out:
```
the e2e bootstrap now puts the complete current InitConfig (19 explicitly null fields) on the wire, where it previously sent only {"key_event_timeout_blocks": N}. Safe today, since the contract config DTO does not use deny_unknown_fields and the archived production WASMs exercised by contract_upgrade_compatibility ignore fields they do not know - but it is a new implicit dependency on that property when talking to older binaries. Worth a note next to the ContractInitFormat scaffold doc, or a skip_serializing_if = "Option::is_none" serde attribute on the InitConfig fields to keep the payload minimal.
```

It's worth mentioning, but not really relevant - if we have an older contract that would break here, we can add it as a variant to the `ContractInitFormat`.

haiyuechen-nearone:
https://github.com/near/mpc/blob/5b1caf817384ce0bbfe74919a059f334c7bde2e6/crates/contract/tests/sandbox/sign.rs#L318-L324

Should we refactor this as well?

pbeza:
> https://github.com/near/mpc/blob/5b1caf817384ce0bbfe74919a059f334c7bde2e6/crates/contract/tests/sandbox/sign.rs#L318-L324
> 
> Should we refactor this as well?

There are [two more](https://github.com/near/mpc/blob/5b1caf817384ce0bbfe74919a059f334c7bde2e6/crates/contract/tests/sandbox/sign.rs#L304-L338) in that file. And perhaps here too?

https://github.com/near/mpc/blob/b0b6b7167e89c015ad59bfb4b0bbc78fcc8432cc/crates/contract/tests/sandbox/upgrade_to_current_contract.rs#L62-L69

crates/near-mpc-contract-interface/src/client.rs
```diff
@@ -410,6 +421,26 @@ mod tests {
...
+                        participants: vec![(
+                            "alice.near".parse().unwrap(),
+                            ParticipantId(0),
+                            ParticipantInfo {
+                                url: "http://localhost:7".to_string(),
+                                tls_public_key: Ed25519PublicKey::from([7u8; 32]),
+                            },
+                        )],
+                    },
+                },
```
pbeza: Copy-paste that we could perhaps extract:

https://github.com/near/mpc/blob/5b1caf817384ce0bbfe74919a059f334c7bde2e6/crates/near-mpc-contract-interface/src/client.rs#L520-L533

BTW, see also [this](https://nearone.slack.com/archives/C07UW93JVQ8/p1787324683858979), which probably would’ve helped catch some/all of the issues raised in the code review comments.

kevindeforth: thanks, good find! https://github.com/near/mpc/pull/4207/commits/2d33fb3b0c535a83891b1d442e56b911aa0a412f

pbeza (dismissed):

haiyuechen-nearone (dismissed):
🚢

haiyuechen-nearone (dismissed):

haiyuechen-nearone (dismissed):

pbeza (approved):

haiyuechen-nearone (approved):

PR #4214, opened 2026-08-24
test: keep e2e tests artifacts on failure

### LINKED ISSUE #2874: e2e test temp dirs are cleaned up on failure, hiding node logs

gilcu3:
Closes #2874

### DIFF
--- a/crates/e2e-tests/tests/foreign_chain_tx_validation.rs
+++ b/crates/e2e-tests/tests/foreign_chain_tx_validation.rs
@@ -3,7 +3,7 @@ use std::num::NonZeroU64;
 
 use crate::common;
 
-use anyhow::{Context, bail};
+use anyhow::Context;
 use e2e_tests::cluster::placeholder_chain_entry;
 use e2e_tests::foreign_chain_mock::{
     MOCK_BLOCK_HASH, MockAuthExpectation, MockServerExt, setup_bitcoin_mock, setup_evm_mock,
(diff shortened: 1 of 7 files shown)


crates/e2e-tests/tests/foreign_chain_tx_validation.rs
gilcu3, line 1: Changes here came from Claude's review, the setup function was not homogeneous wrt to the rest of the crate, and would create a case where test dirs would be deleted on failure still.

haiyuechen-nearone (approved):
✅ Approved. I think we can improve the general UX around e2e test logs further so created https://github.com/near/mpc/issues/4225

anodar (approved):
Thank you!

PR #4227, opened 2026-08-24
refactor: simplifying storage by giving it to the client directly - #3851

SimonRastikian:
Simplifies the triple generation part following https://github.com/near/mpc/issues/3164. Concretely, it refactors TripleStorage::new so the storage takes the network client directly, instead of being handed the two pieces of state it used to derive from that client.

Net effect: TripleStorage depends on one cohesive object (the mesh client) rather than being passed several client-derived values separately.

Reference https://github.com/near/mpc/issues/1680#issuecomment-4852112094

### DIFF
--- a/crates/node/src/providers/ecdsa/triple.rs
+++ b/crates/node/src/providers/ecdsa/triple.rs
@@ -10,6 +10,7 @@ use crate::primitives::{ParticipantId, UniqueId};
 use crate::protocol::NamedProtocol;
 use crate::providers::HasParticipants;
 use crate::providers::ecdsa::{EcdsaSignatureProvider, EcdsaTaskId};
+use crate::providers::ecdsa_common::active_participants_query;
 use crate::tracking::AutoAbortTaskCollection;
 use mpc_node_config::TripleConfig;
 use mpc_primitives::ReconstructionThreshold;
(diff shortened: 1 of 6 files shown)


gilcu3 (dismissed):
Thank you!

crates/node/src/providers/ecdsa/triple.rs
```diff
@@ -55,18 +56,17 @@ impl TripleStorage {
     pub fn new(
         clock: Clock,
         db: Arc<SecretDB>,
-        my_participant_id: ParticipantId,
-        alive_participant_ids_query: Arc<dyn Fn() -> Vec<ParticipantId> + Send + Sync>,
```
kevindeforth, line 59: no, we shouldn't do this if it's not needed. There is no reason for the TripleStore to take the entire `MeshNetworkClient` as input.

kevindeforth (dismissed):

crates/node/src/providers/ecdsa/triple.rs
```diff
@@ -55,18 +56,17 @@ impl TripleStorage {
...
+        client: &Arc<MeshNetworkClient>,
```
netrome: A reference to an arc looks strange. Why not just take a reference directly?

SimonRastikian: I did it for two reasons:
(1) to stay in sync with what was implemented originally, mpc/crates/node/src/providers/ecdsa_common.rs:51 and mpc/crates/node/src/providers/ecdsa_common.rs:97
(2) to not have to client.clone() in extra places.

But thinking about it you are right, it's more idiomatic not to have &Arc<MeshNetworkClient>. Will fix that

SimonRastikian: [e301eed](https://github.com/near/mpc/pull/4227/commits/e301eedd314cd37f9f9b72bf48e2952c1e684d55)

gilcu3 (approved):

crates/node/src/providers/ecdsa/triple.rs
netrome: Thanks :pray:

netrome (approved):

PR #4228, opened 2026-08-24
feat: add tcb-status support for attestation-cli

### LINKED ISSUE #4226: attestation-cli tcb-status subcommand
### Background

Intel raises the TCB bar on its own schedule, and only `UpToDate` is accepted, so a node can start being rejected with nothing changed operator-side. That is what happened on 2026-08-13/14, when TCB-R 19 was promoted to 20 and 7 of our 22 TDX nodes stopped having attestations accepted for days before anyone noticed.

`attestation-cli verify` cannot surface this: it uses the collateral embedded in the node's boot-time payload, so it passes while live submissions fail. Nothing compares a platform against Intel's `early` set either, and nothing says whether it is the TDX
module or BIOS/microcode that is behind.

This is needed for #4217, so that operators can check their servers

### User Story

As a node operator, I want to know whether my platform clears Intel's TCB bar today, whether it will clear the next one, and which SVNs are short, so that I can schedule the firmware update before the contract stops accepting my attestation.

### Acceptance Criteria

- [ ] `attestation-cli tcb-status`, same `--url` / `--file` source as `verify`.
- [ ] Reports a verdict against the collateral the node served.
- [ ] Prints the platform's SVNs and, when not `UpToDate`, which ones are short and which update raises them.

gilcu3:
Closes #4226

This is how the output looks:

```
❯ cargo run -p attestation-cli -- tcb-status --url http://91.134.92.20:8080/public_data
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.18s
     Running `target/debug/attestation-cli tcb-status --url 'http://91.134.92.20:8080/public_data'`
=== MPC Node Platform TCB Status ===

--- Platform, as the quote reports it ---
FMSPC:                  B0C06F000000
tee_tcb_svn:            07010300000000000000000000000000
TDX module:             TDX_01 at ISV SVN 7
TDX TCB components:     [7, 1, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
SGX TCB components:     [3, 3, 2, 2, 4, 1, 0, 5, 0, 0, 0, 0, 0, 0, 0, 0]
PCESVN:                 11

--- served by the node: TCB recovery set 19, issued 2026-08-07T00:44:20Z ---
Status:                 UpToDate

--- Intel `standard`: TCB recovery set 20, issued 2026-08-24T14:06:46Z ---
Status:                 OutOfDate
Advisory IDs:           INTEL-SA-01192, INTEL-SA-01245, INTEL-SA-01312, INTEL-SA-01313
  TDX module TDX_01 ISV SVN is 7, needs 11 -> update the TDX module (SEAM loader)
  TDX TCB component 2 is 3, needs 4 -> update BIOS/microcode
  SGX TCB component 0 is 3, needs 4 -> update BIOS/microcode
  SGX TCB component 1 is 3, needs 4 -> update BIOS/microcode
Error: platform TCB check failed


❯ cargo run -p attestation-cli -- tcb-status --url http://multichain-testnet-01.staking4all.org:8080/public_data
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.18s
     Running `target/debug/attestation-cli tcb-status --url 'http://multichain-testnet-01.staking4all.org:8080/public_data'`
=== MPC Node Platform TCB Status ===

--- Platform, as the quote reports it ---
FMSPC:                  00A06D080000
tee_tcb_svn:            07030600000000000000000000000000
TDX module:             TDX_03 at ISV SVN 7
TDX TCB components:     [7, 3, 6, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
SGX TCB components:     [4, 4, 2, 2, 4, 1, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0]
PCESVN:                 13

--- served by the node: TCB recovery set 20, issued 2026-08-18T00:44:59Z ---
Status:                 UpToDate

--- Intel `standard`: TCB recovery set 20, issued 2026-08-24T13:49:58Z ---
Status:                 UpToDate
```

### DIFF
--- a/crates/attestation-cli/README.md
+++ b/crates/attestation-cli/README.md
@@ -190,7 +190,7 @@ The node's quote is evaluated against two sets, and both are reported:
 | Set | What it tells you |
 |-----|-------------------|
 | served by the node | What the node's own boot-time collateral says. `/public_data` serves the attestation built at CVM boot, so this can lag reality by days. |
-| Intel `standard` | What the contract decides today. Collateral is fetched with no `update` parameter, exactly as the node does. |
+| Intel `standard` | What the contract will decide the next time the node re-attests. Fetched with no `update` parameter, like the node, though straight from Intel rather than through the node's PCCS, so the two can differ in cache freshness. |
 
 Both verdicts come from `dcap-qvl`, the same verification the contract runs. Contacting Intel is anonymous: no PCS subscription key and no node credentials.
 
@@ -226,8 +226,6 @@ Advisory IDs:           INTEL-SA-01192, INTEL-SA-01245, INTEL-SA-01312, INTEL-SA
 The node above still serves an `UpToDate` set 19 while Intel has moved to 20, which is the lag that makes the served row alone unreliable.
 
 A quote is judged in two independent halves, and either can demote a platform. The TDX module half is `tee_tcb_svn[0]`, the module's ISV SVN, matched against the identity `tee_tcb_svn[1]` selects; a newer Intel TDX module (SEAM) raises it, usually shipped inside a vendor BIOS. The platform half is the rest of `tee_tcb_svn` plus the PCK certificate's SGX components and PCESVN; BIOS and CPU microcode raise those. The shortfall lines say which half is short, so you know which update you need.
-
-Exit code is 0 when Intel rates the platform `UpToDate` and 1 when it does not.
 
 ## Troubleshooting
 
--- a/crates/attestation-cli/README.md
+++ b/crates/attestation-cli/README.md
@@ -2,6 +2,8 @@
 
 Standalone verification tool for MPC node TEE attestations. It performs the same Intel TDX (DCAP) attestation verification that the NEAR contract and MPC nodes use, allowing external auditors, operators, and developers to independently validate that an MPC node is running trusted code inside genuine hardware.
 
+Two subcommands: `verify` answers "is this node's attestation acceptable?", and `tcb-status` answers "does this node's platform still clear Intel's TCB bar?".
+
 ## Building
 
 From the repository root:
@@ -37,7 +39,7 @@ Before running verification you need:
 ## Usage
 
 ```
-attestation-cli [OPTIONS]
+attestation-cli verify [OPTIONS]
 ```
 
 ### Required flags
@@ -65,7 +67,7 @@ attestation-cli [OPTIONS]
 ### Verify a live node
 
 ```bash
-attestation-cli \
+attestation-cli verify \
   --url http://<node-host>:8080/public_data \
   --allowed-image-hash abc123...def \
   --launcher-compose-file launcher-compose.yaml
@@ -82,7 +84,7 @@ curl -o public_data.json http://<node-host>:8080/public_data
 Then verify offline:
 
 ```bash
-attestation-cli \
+attestation-cli verify \
   --file public_data.json \
   --allowed-image-hash abc123...def \
   --launcher-compose-file launcher-compose.yaml
@@ -91,7 +93,7 @@ attestation-cli \
 ### Multiple allowed image hashes
 
 ```bash
-attestation-cli \
+attestation-cli verify \
   --url http://<node-host>:8080/public_data \
   --allowed-image-hash abc123...def \
   --allowed-image-hash 789012...345 \
@@ -101,7 +103,7 @@ attestation-cli \
 ### Custom expected measurements
 
 ```bash
-attestation-cli \
+attestation-cli verify \
   --file public_data.json \
   --allowed-image-hash abc123...def \
   --launcher-compose-file launcher-compose.yaml \
@@ -173,7 +175,59 @@ On failure the output includes the error details and ends with `Verdict: FAIL`.
 
 Attestation verification requires DCAP collateral (certificates, CRLs, TCB info) to validate the Intel TDX quote. The CLI uses the collateral **embedded in the node's attestation payload** (the same collateral that was fetched when the node generated its attestation). This is the same approach used by the MPC contract and node.
 
-The CLI does not currently support fetching fresh collateral from Intel's Provisioning Certification Service (PCS) or overriding CRLs. If the embedded collateral is stale, the verification may fail with a DCAP-related error. See [#2320](https://github.com/near/mpc/issues/2320) and [#2286](https://github.com/near/mpc/issues/2286) for planned improvements.
+`verify` does not currently support fetching fresh collateral from Intel's Provisioning Certification Service (PCS) or overriding CRLs. If the embedded collateral is stale, the verification may fail with a DCAP-related error. See [#2320](https://github.com/near/mpc/issues/2320) and [#2286](https://github.com/near/mpc/issues/2286) for planned improvements. `tcb-status` does fetch fresh collateral, so it is the way to tell a stale snapshot apart from a genuinely rejected platform.
+
+## `tcb-status`
+
+```
+attestation-cli tcb-status [OPTIONS]
+```
+
+Reports where a node's platform stands against Intel's TCB requirements. The contract accepts an attestation only when DCAP returns `UpToDate`, and that verdict depends on Intel's TCB info, which Intel revises on its own schedule: a node accepted yesterday can be rejected today with nothing changed on the operator's side.
+
+The node's quote is evaluated against two sets, and both are reported:
+
+| Set | What it tells you |
+|-----|-------------------|
+| served by the node | What the node's own boot-time collateral says. `/public_data` serves the attestation built at CVM boot, so this can lag reality by days. |
+| Intel `standard` | What the contract decides today. Collateral is fetched with no `update` parameter, exactly as the node does. |
+
+Both verdicts come from `dcap-qvl`, the same verification the contract runs. Contacting Intel is anonymous: no PCS subscription key and no node credentials.
+
+It takes the same `--url` / `--file` data source as `verify`, and nothing else.
+
+```bash
+attestation-cli tcb-status --url http://<node-host>:8080/public_data
+```
+
+```
+=== MPC Node Platform TCB Status ===
+
+--- Platform, as the quote reports it ---
+FMSPC:                  90C06F000000
+tee_tcb_svn:            06010300000000000000000000000000
+TDX module:             TDX_01 at ISV SVN 6
+TDX TCB components:     [6, 1, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
+SGX TCB components:     [3, 3, 2, 2, 4, 1, 0, 5, 0, 0, 0, 0, 0, 0, 0, 0]
+PCESVN:                 13
+
+--- served by the node: TCB recovery set 19, issued 2026-08-10T23:45:04Z ---
+Status:                 UpToDate
+
+--- Intel `standard`: TCB recovery set 20, issued 2026-08-24T11:12:32Z ---
+Status:                 OutOfDate
+Advisory IDs:           INTEL-SA-01192, INTEL-SA-01245, INTEL-SA-01312, INTEL-SA-01313
+  TDX module TDX_01 ISV SVN is 6, needs 11 -> update the TDX module (SEAM loader)
+  TDX TCB component 2 is 3, needs 4 -> update BIOS/microcode
+  SGX TCB component 0 is 3, needs 4 -> update BIOS/microcode
+  SGX TCB component 1 is 3, needs 4 -> update BIOS/microcode
+```
+
+The node above still serves an `UpToDate` set 19 while Intel has moved to 20, which is the lag that makes the served row alone unreliable.
+
+A quote is judged in two independent halves, and either can demote a platform. The TDX module half is `tee_tcb_svn[0]`, the module's ISV SVN, matched against the identity `tee_tcb_svn[1]` selects; a newer Intel TDX module (SEAM) raises it, usually shipped inside a vendor BIOS. The platform half is the rest of `tee_tcb_svn` plus the PCK certificate's SGX components and PCESVN; BIOS and CPU microcode raise those. The shortfall lines say which half is short, so you know which update you need.
+
+Exit code is 0 when Intel rates the platform `UpToDate` and 1 when it does not.
 
 ## Troubleshooting
 
@@ -191,6 +245,9 @@ The SHA256 of the `--launcher-compose-file` you provided does not match the comp
 
 **"failed to load expected measurements"**
 The `--expected-measurements` file could not be read or parsed. Ensure it is valid JSON in the `tcb_info.json` format.
+
+**`tcb-status` reports `Rejected` for "served by the node"**
+The collateral in the node's boot-time snapshot has expired. The Intel row is evaluated against freshly fetched collateral, so its verdict still stands; restart the CVM to make the served row meaningful again.
 
 **DCAP verification errors (quote validation, certificate chain, etc.)**
 These indicate the TDX attestation quote failed cryptographic verification. This could mean the attestation is invalid, expired, or the measurements do not match.
--- a/crates/attestation-cli/README.md
+++ b/crates/attestation-cli/README.md
@@ -187,14 +187,16 @@ Reports where a node's platform stands against Intel's TCB requirements. The con
 
 The node's quote is evaluated against two sets, and both are reported:
 
-| Set | What it tells you |
-|-----|-------------------|
+| Collateral source | What it tells you |
+|-------------------|-------------------|
 | served by the node | What the node's own boot-time collateral says. `/public_data` serves the attestation built at CVM boot, so this can lag reality by days. |
-| Intel `standard` | What the contract will decide the next time the node re-attests. Fetched with no `update` parameter, like the node, though straight from Intel rather than through the node's PCCS, so the two can differ in cache freshness. |
+| Intel `standard` | What the contract will decide the next time the node re-attests. Omitting the `update` parameter is what selects the `standard` tier, and it is what the node does too. The fetch goes straight to Intel rather than through the node's PCCS, so the two can differ in cache freshness. |
 
 Both verdicts come from `dcap-qvl`, the same verification the contract runs. Contacting Intel is anonymous: no PCS subscription key and no node credentials.
 
-It takes the same `--url` / `--file` data source as `verify`, and nothing else.
+It takes the same `--url` / `--file` data source as `verify`. Unlike `verify`, network access is always required: `--file` chooses where the quote comes from, not whether Intel is contacted.
+
+`--as-of <unix-timestamp>` evaluates collateral validity at a past instant instead of now. Intel serves only current collateral, so this is for reading a saved quote whose boot-time snapshot has since expired. The Intel row and the exit code stay present-day verdicts, so neither is meaningful with it set.
 
 ```bash
 attestation-cli tcb-status --url http://<node-host>:8080/public_data
@@ -223,9 +225,14 @@ Advisory IDs:           INTEL-SA-01192, INTEL-SA-01245, INTEL-SA-01312, INTEL-SA
   SGX TCB component 1 is 3, needs 4 -> update BIOS/microcode
 ```
 
-The node above still serves an `UpToDate` set 19 while Intel has moved to 20, which is the lag that makes the served row alone unreliable.
+The node above still serves an `UpToDate` set 19 while Intel has moved to 20. The node itself is not stuck: it re-attests to the contract every hour, each time with freshly fetched collateral. What is frozen is `/public_data`, which keeps returning the snapshot built at CVM boot. So a stale served row does not mean the node is submitting stale attestations, and the Intel row is the one that predicts what the next submission will be judged against.
 
-A quote is judged in two independent halves, and either can demote a platform. The TDX module half is `tee_tcb_svn[0]`, the module's ISV SVN, matched against the identity `tee_tcb_svn[1]` selects; a newer Intel TDX module (SEAM) raises it, usually shipped inside a vendor BIOS. The platform half is the rest of `tee_tcb_svn` plus the PCK certificate's SGX components and PCESVN; BIOS and CPU microcode raise those. The shortfall lines say which half is short, so you know which update you need.
+A quote is judged in two independent halves, and either one can demote the platform.
+
+- **TDX module half.** `tee_tcb_svn[0]`, the module's ISV SVN, matched against the identity that `tee_tcb_svn[1]` selects. A newer Intel TDX module (SEAM) raises it, usually shipped inside a vendor BIOS.
+- **Platform half.** The rest of `tee_tcb_svn`, plus the PCK certificate's SGX components and PCESVN. BIOS and CPU microcode raise those.
+
+The shortfall lines say which half is short, so you know which update you need.
 
 ## Troubleshooting
 
--- a/crates/attestation-cli/src/output.rs
+++ b/crates/attestation-cli/src/output.rs
@@ -2,7 +2,7 @@ use attestation::attestation::VerificationError;
 use node_types::http_server::StaticWebData;
 use time::OffsetDateTime;
 
-use dcap_qvl::{policy::PckIdentity, quote::TDReport10};
+use dcap_qvl::{TcbStatus, policy::PckIdentity, quote::TDReport10};
 
 use crate::tcb_status::{self, Report, TcbVerdict};
 use crate::verify::VerificationResult;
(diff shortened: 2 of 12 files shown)


crates/attestation-cli/src/tcb_status.rs
```diff
@@ -0,0 +1,376 @@
...
+///
+/// Mirrors the private `match_tdx_module_identity` in `dcap-qvl`'s
+/// `src/verify.rs`, which surfaces only the converged status, not these.
+pub fn tdx_module(tee_tcb_svn: &[u8; 16]) -> (Option<String>, u8) {
+    let [isv_svn, selector, ..] = *tee_tcb_svn;
+    (
+        (selector != 0).then(|| format!("TDX_{selector:02X}")),
+        isv_svn,
+    )
+}
```
gilcu3: this part is cryptic to me, but apparently that's what dcap does and we cannot reuse it

crates/attestation-cli/README.md
```diff
@@ -173,7 +175,57 @@ On failure the output includes the error details and ends with `Verdict: FAIL`.
...
+```
+
+Reports where a node's platform stands against Intel's TCB requirements. The contract accepts an attestation only when DCAP returns `UpToDate`, and that verdict depends on Intel's TCB info, which Intel revises on its own schedule: a node accepted yesterday can be rejected today with nothing changed on the operator's side.
+
+The node's quote is evaluated against two sets, and both are reported:
+
+| Set | What it tells you |
+|-----|-------------------|
+| served by the node | What the node's own boot-time collateral says. `/public_data` serves the attestation built at CVM boot, so this can lag reality by days. |
+| Intel `standard` | What the contract will decide the next time the node re-attests. Fetched with no `update` parameter, like the node, though straight from Intel rather than through the node's PCCS, so the two can differ in cache freshness. |
```
pbeza: It’s a bit confusing to mention `standard` while also saying it’s fetched without the `update` parameter. Might be worth explicitly saying that `standard` is the default when `update` is omitted.

BTW, it’d be nice to check against upcoming TCB-Rs as well, but that can probably be a follow-up.

crates/attestation-cli/README.md
```diff
@@ -173,7 +175,57 @@ On failure the output includes the error details and ends with `Verdict: FAIL`.
...
+
+```
+attestation-cli tcb-status [OPTIONS]
+```
+
+Reports where a node's platform stands against Intel's TCB requirements. The contract accepts an attestation only when DCAP returns `UpToDate`, and that verdict depends on Intel's TCB info, which Intel revises on its own schedule: a node accepted yesterday can be rejected today with nothing changed on the operator's side.
+
+The node's quote is evaluated against two sets, and both are reported:
+
+| Set | What it tells you |
```
pbeza:
```suggestion
| TCB Recovery Set | What it tells you |
```

crates/attestation-cli/src/tcb_status.rs
```diff
@@ -0,0 +1,469 @@
...
+pub struct Shortfall {
+    pub component: String,
+    pub have: u16,
+    pub needs: u16,
+    pub remedy: &'static str,
+}
+
+#[expect(clippy::large_enum_variant)]
+#[derive(Debug)]
+pub enum Outcome {
```
pbeza: `Outcome` feels a bit too generic name.

crates/attestation-cli/src/output.rs
```diff
@@ -98,3 +101,78 @@ fn format_timestamp(unix_secs: u64) -> String {
...
+pub fn print_tcb_status(report: &Report) {
+    println!("=== MPC Node Platform TCB Status ===");
+
+    // Both rows judge the same quote, so its platform numbers are the same in
+    // each; whichever verified first will do. When neither did, the rejection
+    // reasons below say why.
+    match (&report.served, &report.standard) {
+        (Outcome::Verified { claims, .. }, _) | (_, Outcome::Verified { claims, .. }) => {
+            print_platform(claims);
+        }
```
pbeza, line 115: Does that mean that if we get `Outcome::Verified` for both `served` and `standard`, then the both `claims` should be equal? If so, perhaps we should `assert_eq` that to double-check our thinking.

crates/attestation-cli/src/output.rs
```diff
@@ -98,3 +101,78 @@ fn format_timestamp(unix_secs: u64) -> String {
...
+    );
+    println!(
+        "TDX module:             {} at ISV SVN {}",
+        module.as_deref().unwrap_or("unnamed"),
+        module_svn
+    );
+    println!("TDX TCB components:     {:?}", report.tee_tcb_svn);
+    println!("SGX TCB components:     {:?}", pck.cpu_svn);
+    println!("PCESVN:                 {}", pck.pce_svn);
+}
```
pbeza: Doesn't `dcap-qvl` expose some nice printers for these? I see a lot of custom printing being built up around reports, collateral, and other attestation-related stuff. If not, that's fine; let's just make sure we're not duplicating printers implemented elsewhere in the repo.

crates/attestation-cli/src/tcb_status.rs
```diff
@@ -0,0 +1,469 @@
...
+use dcap_qvl::{
+    QuoteCollateralV3, TcbLevel, TcbStatus,
+    collateral::{CollateralClient, INTEL_PCS_URL},
+    policy::{PckIdentity, QuoteClaims, QuotePolicy},
+    quote::{Quote, TDReport10},
+    tcb_info::TcbInfo,
+    verify::QuoteVerifier,
+};
+use mpc_attestation::attestation::Attestation;
+use mpc_attestation::dcap_conversions::collateral_into_dcap;
```
pbeza: Why not merge the imports, like we did above?

crates/attestation-cli/src/tcb_status.rs
```diff
@@ -0,0 +1,469 @@
...
+const UPDATE_BIOS: &str = "update BIOS/microcode";
+
+/// One security version number that Intel's accepted TCB level puts out of reach.
+#[derive(Debug)]
+pub struct Shortfall {
+    pub component: String,
+    pub have: u16,
+    pub needs: u16,
+    pub remedy: &'static str,
+}
```
pbeza, line 37: Isn't it already defined somewhere in `dcap-qvl`?

crates/attestation-cli/src/tcb_status.rs
```diff
@@ -0,0 +1,469 @@
...
+    /// No status at all. Usually expired collateral in a long-running node's
+    /// boot-time snapshot.
+    Rejected(String),
+}
+
+pub struct Report {
+    /// What the node's own boot-time collateral says, which can lag by days.
+    pub served: Outcome,
+    /// What the contract decides today, against collateral fetched just now.
+    pub standard: Outcome,
```
pbeza: Why just `standard`? Why not also query `early`, which gives access to TCB-R 22? (can be a follow-up)

crates/attestation-cli/README.md
```diff
@@ -173,7 +175,57 @@ On failure the output includes the error details and ends with `Verdict: FAIL`.
...
+|-----|-------------------|
+| served by the node | What the node's own boot-time collateral says. `/public_data` serves the attestation built at CVM boot, so this can lag reality by days. |
```
barakeinav1, line 192: "/public_data` serves the attestation built at CVM boot, so this can lag reality by days."
is this statement true?

crates/attestation-cli/src/tcb_status.rs
```diff
@@ -0,0 +1,469 @@
...
+    tcb_info::TcbInfo,
+    verify::QuoteVerifier,
+};
+use mpc_attestation::attestation::Attestation;
+use mpc_attestation::dcap_conversions::collateral_into_dcap;
+use node_types::http_server::StaticWebData;
+
+/// Whole-fetch bound on the Intel PCS call. `with_default_http` builds a reqwest
+/// client with a 180s timeout of its own, far too long to leave an interactive
+/// command hanging on, and `dcap-qvl` keeps that client out of its public API.
```
pbeza, line 23: Are you sure? I think I saw it somewhere in their repo, but perhaps it wasn’t `pub` 🤔

crates/attestation-cli/src/tcb_status.rs
```diff
@@ -0,0 +1,469 @@
...
+    // failing one of them.
+    let report = *Quote::parse(quote)
+        .context("parsing the node's quote")?
+        .report
+        .as_td10()
+        .context("the node's quote does not carry a TDX report")?;
+    let now = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .context("system clock is before the UNIX epoch")?
+        .as_secs();
```
pbeza: Optional nit: This could be an optional parameter of the CLI tool, defaulting to `now`. I can imagine some scenarios where it could be useful when retrospectively inspecting quotes.

crates/attestation-cli/src/output.rs
```diff
@@ -98,3 +101,78 @@ fn format_timestamp(unix_secs: u64) -> String {
...
+    // reasons below say why.
+    match (&report.served, &report.standard) {
+        (Outcome::Verified { claims, .. }, _) | (_, Outcome::Verified { claims, .. }) => {
+            print_platform(claims);
+        }
+        _ => println!("\nThe platform's own numbers are unavailable: neither evaluation verified."),
+    }
+
+    print_outcome("served by the node", &report.served);
+    print_outcome("Intel `standard`", &report.standard);
```
pbeza: I think in most cases the `served` and `standard` reports will be the same, so perhaps in that case we could print them just once.

crates/attestation-cli/README.md
```diff
@@ -173,7 +175,57 @@ On failure the output includes the error details and ends with `Verdict: FAIL`.
...
+--- Intel `standard`: TCB recovery set 20, issued 2026-08-24T11:12:32Z ---
+Status:                 OutOfDate
+Advisory IDs:           INTEL-SA-01192, INTEL-SA-01245, INTEL-SA-01312, INTEL-SA-01313
+  TDX module TDX_01 ISV SVN is 6, needs 11 -> update the TDX module (SEAM loader)
+  TDX TCB component 2 is 3, needs 4 -> update BIOS/microcode
+  SGX TCB component 0 is 3, needs 4 -> update BIOS/microcode
+  SGX TCB component 1 is 3, needs 4 -> update BIOS/microcode
+```
+
+The node above still serves an `UpToDate` set 19 while Intel has moved to 20, which is the lag that makes the served row alone unreliable.
```
barakeinav1: I didn't fully underhand the meaning of this sentence

crates/attestation-cli/README.md
```diff
@@ -173,7 +175,57 @@ On failure the output includes the error details and ends with `Verdict: FAIL`.
...
+
+A quote is judged in two independent halves, and either can demote a platform. The TDX module half is `tee_tcb_svn[0]`, the module's ISV SVN, matched against the identity `tee_tcb_svn[1]` selects; a newer Intel TDX module (SEAM) raises it, usually shipped inside a vendor BIOS. The platform half is the rest of `tee_tcb_svn` plus the PCK certificate's SGX components and PCESVN; BIOS and CPU microcode raise those. The shortfall lines say which half is short, so you know which update you need.
```
barakeinav1: kind of hard for me to read and understand this sentence

crates/attestation-cli/src/tcb_status.rs
```diff
@@ -0,0 +1,469 @@
...
+        .min_by_key(|short| {
+            (
+                short.len(),
+                short
+                    .iter()
+                    .map(|s| u32::from(s.needs - s.have))
+                    .sum::<u32>(),
+            )
+        })
+        .unwrap_or_default();
```
pbeza, line 187: I didn’t really spend enough time on it to fully understand it, but Claude claims there might be a problem:

> Platform-level selection diverges from dcap-qvl. `tcb_status.rs:152-166` filters levels to `tcb_status == UpToDate` and picks the nearest by (count, summed deficit). `dcap-qvl`'s `match_platform_tcb` (`verify.rs:981-1042`) walks the canonicalized list — sorted by SGX components desc, then PCE SVN, then TDX components, with status not part of the ordering — and returns the first level cleared regardless of its status. Two reachable consequences: when no level is `UpToDate` at all (real Intel case — newest level `SWHardeningNeeded`), the filter yields nothing and shortfalls is silently empty while the contract rejects; and because SGX and TDX components advance independently for TDX FMSPCs, `dcap-qvl` can match a higher-sorted non-`UpToDate` level while the printed "needs N" numbers describe a different level that did not decide the status.

crates/attestation-cli/src/tcb_status.rs
```diff
@@ -0,0 +1,469 @@
...
+        .zip(pck.cpu_svn)
+        .enumerate()
+        .filter_map(|(index, (needs, have))| {
+            short(
+                format!("SGX TCB component {index}"),
+                have,
+                needs.svn,
+                UPDATE_BIOS,
+            )
+        });
```
pbeza, line 240: This feels complex. Could we double-check that this isn’t already implemented somewhere in `dcap-qvl` or our repo? It’d be nice to keep this complexity in one place. Same comment applies to: https://github.com/near/mpc/pull/4228/changes#r3852028924.

pbeza (commented):
Overall LGTM, but I left a couple of optional nits.

crates/attestation-cli/README.md
gilcu3: yes, it is

crates/attestation-cli/README.md
gilcu3: This is why we need to also request the latest collateral from intel. The node submits every our, but does not update its public_data. Anyway seems good to check collateral from intel, just in case they differ for some reason

crates/attestation-cli/README.md
gilcu3: Which sentence do you mean exactly? Above I see a full paragraph

crates/attestation-cli/README.md
gilcu3: > BTW, it’d be nice to check against upcoming TCB-Rs as well, but that can probably be a follow-up.

I tried, code exploded, so I left it to you in the next PR :smile:

crates/attestation-cli/src/output.rs
gilcu3: They don't need to be equal afaik, because the server one might use outdated collateral. This is related to this [comment ](https://github.com/near/mpc/pull/4228#discussion_r3851915034) as well

crates/attestation-cli/README.md
```diff
@@ -173,7 +175,57 @@ On failure the output includes the error details and ends with `Verdict: FAIL`.
...
+The node's quote is evaluated against two sets, and both are reported:
+
+| Set | What it tells you |
+|-----|-------------------|
+| served by the node | What the node's own boot-time collateral says. `/public_data` serves the attestation built at CVM boot, so this can lag reality by days. |
+| Intel `standard` | What the contract will decide the next time the node re-attests. Fetched with no `update` parameter, like the node, though straight from Intel rather than through the node's PCCS, so the two can differ in cache freshness. |
+
+Both verdicts come from `dcap-qvl`, the same verification the contract runs. Contacting Intel is anonymous: no PCS subscription key and no node credentials.
+
+It takes the same `--url` / `--file` data source as `verify`, and nothing else.
```
barakeinav1: maybe worth adding a note that even if using --file, this still need network connectivity.

crates/attestation-cli/src/output.rs
gilcu3: unfortunately no, they could defer. Not only because the public_data collateral is only computed on startup, but also because there might be a problem with the node's PCCS service configured. So at least for now, having both is strictly needed.

crates/attestation-cli/README.md
barakeinav1: this got me confused.
the problem is that /public_data serves an attention that was generate once,  but the node actually does submit a new one every time.

if this confused me, it will probably confuse operators as well.

crates/attestation-cli/src/tcb_status.rs
gilcu3: I can check again, but that's what I got. When doing the PR I tried very hard (with Claude) to avoid reimplementing anything, and this came up after a review

crates/attestation-cli/src/tcb_status.rs
gilcu3: I could not find anything like this. dcap-qvl unfortunately did not have the explanation part exposed as far I could see. Will check again just in case

crates/attestation-cli/README.md
barakeinav1: hmm,
I think we should fix this so public_data serve that last data the node used to send to the contract. 
what do you think? (not in this PR)

crates/attestation-cli/src/tcb_status.rs
gilcu3: could be done indeed

crates/attestation-cli/src/output.rs
gilcu3: not really. Although notice that the code is mostly about how we present the data, luckily I did not need to implement any Display for any actual struct, so unless they had exactly what we needed, we would still need presentation code like this.

crates/attestation-cli/README.md
barakeinav1: "A quote is judged in two independent halves, and either can demote a platform. The TDX module half is `tee_tcb_svn[0]`, the module's ISV SVN, matched against the identity `tee_tcb_svn[1]` selects; a newer Intel TDX module (SEAM) raises it, usually shipped inside a vendor BIOS. The platform half is the rest of `tee_tcb_svn` plus the PCK certificate's SGX components and PCESVN; BIOS and CPU microcode raise those."

maybe break it down struct format? or bullets?

crates/attestation-cli/README.md
```diff
@@ -173,7 +175,57 @@ On failure the output includes the error details and ends with `Verdict: FAIL`.
...
+PCESVN:                 13
+
+--- served by the node: TCB recovery set 19, issued 2026-08-10T23:45:04Z ---
+Status:                 UpToDate
+
+--- Intel `standard`: TCB recovery set 20, issued 2026-08-24T11:12:32Z ---
+Status:                 OutOfDate
+Advisory IDs:           INTEL-SA-01192, INTEL-SA-01245, INTEL-SA-01312, INTEL-SA-01313
+  TDX module TDX_01 ISV SVN is 6, needs 11 -> update the TDX module (SEAM loader)
+  TDX TCB component 2 is 3, needs 4 -> update BIOS/microcode
```
barakeinav1, line 223: nice, this is very helpful

crates/attestation-cli/README.md
gilcu3: could be added. I don't think we should aim to make this a fully documented tool, as that would be wasteful at this point. So I will check if we already mention it will call intel or not, and add that. At some point I did have an `--offline` option, but removed to simplify this as much as possible.

crates/attestation-cli/README.md
gilcu3: we could add that explanation to the readme

crates/attestation-cli/src/tcb_status.rs
gilcu3: yeah I tried that, but `dcap-qvl` has several things not `pub`, as they didn't think of this use case.

crates/attestation-cli/src/tcb_status.rs
gilcu3: I will try to align that code as much as possible, not sure if they currently diverge in a case we can get, but need to check

crates/attestation-cli/src/tcb_status.rs
gilcu3: left as follow-up because it requires quite a few new code, apparently another shortcoming in `dcap-qvl`

barakeinav1:
This PR has a small breaking change in how you run the CLI (`attestation-cli --url ...` is now `attestation-cli verify --url ...`), should we make the title `feat!`? Otherwise the changelog won't flag it as breaking.

crates/attestation-cli/src/tcb_status.rs
```diff
@@ -0,0 +1,469 @@
...
+    // Fetched with no `update` parameter, like the node, so this is Intel's
+    // `standard` set: what the contract will decide at the next re-attestation.
+    // The node reaches it through its PCCS, so the two can differ in cache
+    // freshness.
+    let client = CollateralClient::with_default_http(INTEL_PCS_URL)
+        .context("building the Intel PCS client")?;
+    let standard = tokio::time::timeout(PCS_TIMEOUT, client.fetch(quote))
+        .await
+        .with_context(|| format!("Intel's PCS did not respond within {PCS_TIMEOUT:?}"))?
+        .context("fetching collateral from Intel's PCS")?;
```
barakeinav1: nit: If Intel is unreachable we bail here and print nothing - I think it is better to print some error.

crates/attestation-cli/src/output.rs
```diff
@@ -98,3 +101,78 @@ fn format_timestamp(unix_secs: u64) -> String {
...
+            println!("Rejected:               {reason}");
+        }
+    }
+}
+
+fn print_platform(claims: &QuoteClaims) {
+    // `evaluate` already established the report is TD10, so this never misses.
+    let Some(report) = claims.report.as_td10() else {
+        return;
+    };
```
barakeinav1: nit (found by Claude): `run()` already parsed this report and threw it away (`tcb_status.rs:73-77`). If we kept it on `Report` we could pass it in directly and drop this "can't happen" branch - and then the TDX half of the platform section could still print when both rows are rejected.

gilcu3:
> This PR has a small breaking change in how you run the CLI (`attestation-cli --url ...` is now `attestation-cli verify --url ...`), should we make the title `feat!`? Otherwise the changelog won't flag it as breaking.

attestation-cli is not really part of our public API, it is just an internal tool. So I don't think we should consider this breaking, although it is just a technicality

crates/attestation-cli/src/tcb_status.rs
gilcu3: hmm I thought we did print something, will check

crates/attestation-cli/src/output.rs
gilcu3: will check this, I remember some dance around where to put that, with some tradeoffs, but it might be right.

crates/attestation-cli/README.md
gilcu3: yeah sounds reasonable, but yeah could require some refactor

crates/attestation-cli/src/tcb_status.rs
barakeinav1: [`match_platform_tcb`](https://docs.rs/dcap-qvl/0.6.1/src/dcap_qvl/verify.rs.html#950) and [`match_tdx_module_identity`](https://docs.rs/dcap-qvl/0.6.1/src/dcap_qvl/verify.rs.html#1047) are private, so we can't reuse them today.

@pbeza you did some upstream contributions before - should we try to make them public upstream? then we can drop our copy.

barakeinav1 (dismissed):
thanks.
left some comments, but nothing blocking.

crates/attestation-cli/src/tcb_status.rs
gilcu3: yeah I think based on this, we could try to upstream several bits to make our lives easier, probably behind a feature flag. Unfortunately we cannot wait for that before having this (as we want operators to be able to use this ASAP), else it would be the best choice

crates/attestation-cli/src/tcb_status.rs
gilcu3: This one was the only hard one. Change in this [commit](https://github.com/near/mpc/pull/4228/commits/b7869e4de3c7919e50689f45ec718b32df838554) Basically it seems to be a corner case, and the fix was to make that explicit in the output

pbeza (approved):

crates/attestation-cli/src/tcb_status.rs
```diff
@@ -0,0 +1,492 @@
...
+            .as_secs(),
+    };
+
+    let served = collateral_into_dcap(dstack.collateral.clone());
+    // An unreachable PCS must not cost the operator the served row, which needs
+    // no network at all.
+    let standard = match fetch_standard(quote).await {
+        Ok(collateral) => evaluate(quote, &report, &collateral, now),
+        Err(err) => TcbVerdict::Rejected(format!("{err:#}")),
+    };
```
barakeinav1, line 102: `--as-of` moves the clock for the Intel row too, and dcap-qvl rejects collateral issued after `now` ([verify.rs:676](https://docs.rs/dcap-qvl/0.6.1/src/dcap_qvl/verify.rs.html#676)). So today's collateral is always "not yet valid" at a past timestamp - the Intel row is always Rejected and the exit code is always 1.

Should we skip the Intel fetch when `--as-of` is set?

gilcu3: I would keep as is, as explained in [slack](https://nearone.slack.com/archives/C0912BTG51T/p1787663480444149?thread_ts=1787586816.065839&cid=C0912BTG51T)

barakeinav1 (approved):

PR #4232, opened 2026-08-25
ci: add cache layer for repro-env

### LINKED ISSUE #4191: Cache repro-env package downloads in CI
### Background

CI jobs that run `repro-env` download all 66 pinned `.deb`s from `snapshot.debian.org` on every run, sequentially and with no retry. `snapshot.debian.org` throttles, so jobs fail on a connect timeout, e.g. https://github.com/near/mpc/actions/runs/32255242317/job/96075348994:

```
Error: Failed to download package from url: ".../libgc1_8.2.2-3_amd64.deb"
  2: client error (Connect)
  3: operation timed out
```

The timeouts are hardcoded in upstream `repro-env` (connect 15s, read 240s) with no flag or env override, so they cannot be raised from our side.

`repro-env` keeps a content-addressed store at `${REPRO_ENV_CACHE:-$XDG_CACHE_HOME}/repro-env/pkgs/<sha256[0:2]>/<rest>` and skips the HTTP request entirely when the sha256 path already exists. Caching that directory keyed on the hash of `repro-env.lock` removes `snapshot.debian.org` from steady-state CI: only a lockfile bump needs the network.

The cache should follow the same policy as the Rust cache in `ci.yml`: restored on every run, saved only on `main` (`save-if: ${{ github.ref == 'refs/heads/main' }}` there), on the same cache backend (`cache-provider: warpbuild`). Since `actions/cache` saves in a `post-if: success()` step, the equivalent is `actions/cache/restore` on every run plus a save step gated on `github.ref == 'refs/heads/main'`.

All four `repro-env` job sites share the single `repro-env.lock`, so one key serves all of them, and a shared composite action under `.github/actions/` (as with `restore-nix-cache`) avoids repeating the pair of steps.

### Acceptance Criteria

- The `repro-env` package cache directory is restored in every `repro-env` job, keyed on `hashFiles('repro-env.lock')`, so a cache hit performs no `snapshot.debian.org` downloads.
- The cache is saved only on `main`; PR runs restore but never save.
- Applied to all four `repro-env` job sites: `ci.yml` (node build, rust-launcher build), `docker_build_node.yml`, `docker_build_node_gcp.yml`, `docker_build_rust_launcher.yml`.

gilcu3:
Closes #4191

### DIFF
--- a/.github/actions/cache-repro-env-pkgs/action.yml
+++ b/.github/actions/cache-repro-env-pkgs/action.yml
@@ -19,14 +19,14 @@ runs:
 
     - name: Restore and save repro-env packages
       if: inputs.save == 'true'
-      uses: WarpBuilds/cache@dad44e2487e9dc889ad6e6bd198e77bb2d828eee # v2.0.0
+      uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
       with:
         path: ~/.cache/repro-env/pkgs
         key: ${{ steps.compute-key.outputs.key }}
 
     - name: Restore repro-env packages
       if: inputs.save != 'true'
-      uses: WarpBuilds/cache/restore@dad44e2487e9dc889ad6e6bd198e77bb2d828eee # v2.0.0
+      uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
       with:
         path: ~/.cache/repro-env/pkgs
         key: ${{ steps.compute-key.outputs.key }}
--- a/.github/actions/cache-repro-env-pkgs/action.yml
+++ b/.github/actions/cache-repro-env-pkgs/action.yml
@@ -1,0 +1,32 @@
+name: Cache repro-env packages
+description: |
+  Restore repro-env's content-addressed package store, so `repro-env build`
+  finds every pinned `.deb` on disk and never contacts snapshot.debian.org,
+  which throttles and times out under the connect/read timeouts hardcoded in
+  repro-env.
+inputs:
+  save:
+    description: Whether this job should save the cache after running.
+    required: false
+    default: "false"
+runs:
+  using: composite
+  steps:
+    - name: Compute repro-env cache key
+      id: compute-key
+      shell: bash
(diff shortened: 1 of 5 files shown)


pbeza (approved):

haiyuechen-nearone (approved):
I see in the issue that this caching should be applied to all jobs, but Claude flagged that this is currently not applied to the release jobs. Is there a reason for the exclusion?

gilcu3:
> I see in the issue that this caching should be applied to all jobs, but Claude flagged that this is currently not applied to the release jobs. Is there a reason for the exclusion?

yes, release jobs should not use cache, to prevent cache poisoning attacks. For that reason zizmor complains about those by default

.github/actions/cache-repro-env-pkgs/action.yml
```diff
@@ -0,0 +1,32 @@
...
+    description: Whether this job should save the cache after running.
+    required: false
+    default: "false"
+runs:
+  using: composite
+  steps:
+    - name: Compute repro-env cache key
+      id: compute-key
+      shell: bash
+      run: echo "key=repro-env-pkgs-${{ runner.os }}-${{ hashFiles('repro-env.lock') }}" >> "$GITHUB_OUTPUT"
```
pbeza, line 18: Optional nit from Claude: we can make sure the PR saves to the same folder `repro-env` uses. Setting [`REPRO_ENV_CACHE`](https://github.com/kpcyrd/repro-env/blob/04df19d47abefb2e415b9f24e578ce661b8203a0/src/paths.rs#L19-L31) prevents the assumed `~/.cache/repro-env/pkgs` path from drifting.

```suggestion
      run: |
        # Pin the store root so it cannot drift from the cached path if the
        # runner image ever sets XDG_CACHE_HOME
        echo "REPRO_ENV_CACHE=$HOME/.cache/repro-env" >> "$GITHUB_ENV"
        echo "key=repro-env-pkgs-${{ runner.os }}-${{ hashFiles('repro-env.lock') }}" >> "$GITHUB_OUTPUT"
```

_(edit) I think I’d ignore it._

gilcu3: yeah I left this one out just for simplicity, it doesnt seem likely we will use that env var

pbeza (approved):

PR #4234, opened 2026-08-25
feat: Accept foreign chain requests based on available chain

### LINKED ISSUE #4221: Switch contract to accept foreign chain tx validation requests based on new API
### Background 
In https://github.com/near/mpc/issues/3569 we switched to new API for calculating available chains. Contract still uses the old API to get supported chains (as opposed to available chains), one calculated by intersection of all supported chains and the other based on threshold. 

We should switch to the new API.

### Acceptance criteria

Contract accepts requests based on the new API rather than the old one.

anodar:
Note: this change does not need to be deployed on the contract before node gets updated. All nodes support all the chains so foreign tx requests will be still accepted during normal operations. After this change is deployed on the contract it will just allow more flexibility.

We do need to vote for new whitelist that contains SUI though.

Closes: https://github.com/near/mpc/issues/4221

### DIFF
--- a/crates/contract/src/api/foreign_chain.rs
+++ b/crates/contract/src/api/foreign_chain.rs
@@ -649,6 +649,27 @@ mod tests {
     }
 
     #[test]
+    #[should_panic(expected = "Requested foreign chain, Bitcoin, is not available.")]
+    fn verify_foreign_tx__should_reject_whitelisted_chain_covered_by_fewer_than_threshold() {
+        // Given: 4 participants, ForeignTx reconstruction threshold 3; only 2 cover Bitcoin.
+        let mut rng = rand::rngs::StdRng::from_seed([42u8; 32]);
+        let (context, mut contract, _sk) =
+            basic_setup_with_protocol(Protocol::CaitSith, DomainPurpose::ForeignTx, &mut rng);
+        whitelist_chain(&mut contract, dtos::ForeignChain::Bitcoin);
+        for account_id in participant_account_ids(&contract).iter().take(2) {
+            register_foreign_chains_config_for(
+                &mut contract,
+                account_id,
+                [dtos::ForeignChain::Bitcoin],
+            );
+        }
+        testing_env!(context.clone());
+
+        // When
+        contract.verify_foreign_transaction(bitcoin_request_args());
+    }
+
+    #[test]
     fn verify_foreign_tx__should_accept_chain_covered_by_threshold_of_participants() {
         // Given: 4 participants, ForeignTx reconstruction threshold 3; only 3 cover Bitcoin.
         let mut rng = rand::rngs::StdRng::from_seed([42u8; 32]);
--- a/crates/contract/src/api/foreign_chain.rs
+++ b/crates/contract/src/api/foreign_chain.rs
@@ -625,7 +625,61 @@ mod tests {
         testing_env!(context.clone());
 
         // When - requesting Bitcoin which is not in the policy
-        contract.verify_foreign_transaction(VerifyForeignTransactionRequestArgs {
+        contract.verify_foreign_transaction(bitcoin_request_args());
+    }
+
+    #[test]
+    #[should_panic(expected = "Requested foreign chain, Bitcoin, is not available.")]
+    fn verify_foreign_tx__should_reject_chain_covered_by_all_participants_but_not_whitelisted() {
+        // Given
+        let mut rng = rand::rngs::StdRng::from_seed([42u8; 32]);
+        let (context, mut contract, _sk) =
+            basic_setup_with_protocol(Protocol::CaitSith, DomainPurpose::ForeignTx, &mut rng);
+        for account_id in participant_account_ids(&contract) {
+            register_foreign_chains_config_for(
+                &mut contract,
+                &account_id,
+                [dtos::ForeignChain::Bitcoin],
+            );
+        }
+        testing_env!(context.clone());
+
+        // When
+        contract.verify_foreign_transaction(bitcoin_request_args());
+    }
+
+    #[test]
+    fn verify_foreign_tx__should_accept_chain_covered_by_threshold_of_participants() {
+        // Given: 4 participants, ForeignTx reconstruction threshold 3; only 3 cover Bitcoin.
+        let mut rng = rand::rngs::StdRng::from_seed([42u8; 32]);
+        let (context, mut contract, _sk) =
+            basic_setup_with_protocol(Protocol::CaitSith, DomainPurpose::ForeignTx, &mut rng);
+        whitelist_chain(&mut contract, dtos::ForeignChain::Bitcoin);
+        for account_id in participant_account_ids(&contract).iter().take(3) {
+            register_foreign_chains_config_for(
+                &mut contract,
+                account_id,
+                [dtos::ForeignChain::Bitcoin],
+            );
+        }
+        testing_env!(context.clone());
+        let request_args = bitcoin_request_args();
+        let request = args_into_verify_foreign_tx_request(request_args.clone());
+
+        // When
+        contract.verify_foreign_transaction(request_args);
+
+        // Then
+        assert!(
+            contract
+                .pending_verify_foreign_tx_requests
+                .get(&request)
+                .is_some()
+        );
+    }
+
+    fn bitcoin_request_args() -> VerifyForeignTransactionRequestArgs {
+        VerifyForeignTransactionRequestArgs {
             domain_id: DomainId::default().0.into(),
             payload_version: ForeignTxPayloadVersion::V1,
             expected_payload_hash: None,
@@ -634,6 +688,6 @@ mod tests {
                 confirmations: 2.into(),
                 extractors: vec![BitcoinExtractor::BlockHash],
             }),
-        });
+        }
     }
 }
--- a/crates/contract/src/api/foreign_chain.rs
+++ b/crates/contract/src/api/foreign_chain.rs
@@ -49,10 +49,10 @@ impl MpcContract {
         );
 
         let requested_chain = request.request.chain();
-        let supported_chains = self.get_supported_foreign_chains();
-        if !supported_chains.contains(&requested_chain) {
+        let available_chains = self.get_available_foreign_chains();
+        if !available_chains.contains(&requested_chain) {
             env::panic_str(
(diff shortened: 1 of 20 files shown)


anodar:
@claude code review 

> crates/contract/src/api/foreign_chain.rs:52 — The rollout note in the PR description is not accurate in the direction that matters.  ...

Reworded the description.

> docs/tee-lifecycle.md:235, docs/hot-tee-signing-design.md:409 — Both method tables still list get_supported_foreign_chains() as the view to "Query active foreign chains".  ....

Fixed.

> crates/contract/src/api/key_events.rs:65-67 — vote_pk transitions Initializing → Running without calling recompute_available_foreign_chains() ...

This is intentional. We want to keep `vote_pk` is a critical path that we want to keep as slim as possible with least failure points.

> crates/contract/src/api/foreign_chain.rs:652 — The new tests cover "covered but not whitelisted → reject" ...

covered in `verify_foreign_tx__should_reject_whitelisted_chain_covered_by_fewer_than_threshold` now.

crates/contract/src/api/foreign_chain.rs
```diff
@@ -49,10 +49,10 @@ impl MpcContract {
         );
 
         let requested_chain = request.request.chain();
-        let supported_chains = self.get_supported_foreign_chains();
-        if !supported_chains.contains(&requested_chain) {
+        let available_chains = self.get_available_foreign_chains();
+        if !available_chains.contains(&requested_chain) {
             env::panic_str(
-                &InvalidParameters::ForeignChainNotSupported {
+                &InvalidParameters::ForeignChainNotAvailable {
```
anodar, line 55: This is pretty much the main change in this PR.

netrome: Ah this will be so nice to have!

netrome: Ah it really basically is only this, and the rest of the PR is test utilities and updates 😅

crates/contract/src/api/foreign_chain.rs
```diff
@@ -49,10 +49,10 @@ impl MpcContract {
         );
 
         let requested_chain = request.request.chain();
-        let supported_chains = self.get_supported_foreign_chains();
```
netrome, line 52: Are we still using the legacy API anywhere? It feels like we should be able to clean up these methods in this PR.

netrome (dismissed):
Looks good to me, but I wonder if we should clean up the legacy API methods from the contract while we're at it. They should not be used once we roll this out right?

crates/contract/src/api/foreign_chain.rs
gilcu3: Ok if this is true maybe I can review in the few minutes left, on it

crates/contract/src/api/foreign_chain_support.rs
```diff
@@ -21,6 +21,9 @@ impl MpcContract {
     /// set); panics in [`NotInitialized`](ProtocolContractState::NotInitialized) or when the caller is not a participant. Entries for
     /// accounts that are no longer participants are pruned after resharing by
     /// [`Self::clean_foreign_chain_data`].
+    #[deprecated(
+        note = "superseded by register_foreign_chains_config; feeds only the legacy get_supported_foreign_chains view (https://github.com/near/mpc/issues/3434)"
```
gilcu3: if what we want is a TODO, we could write it in the format. Else mentioning the issue seems a bit odd

crates/contract/src/api/foreign_chain_support.rs
```diff
@@ -282,6 +289,9 @@ impl MpcContract {
             .into()
     }
 
+    #[deprecated(
+        note = "superseded by get_foreign_chains_configs (https://github.com/near/mpc/issues/3434)"
+    )]
```
gilcu3: same for these two ofc

crates/contract/src/api/test_utils.rs
```diff
@@ -261,6 +264,53 @@ pub(crate) fn forwarded_participant_call_contract() -> MpcContract {
...
+        .info(account_id)
+        .expect("account must be a participant")
+        .tls_public_key
+        .clone();
+    let mut env = Environment::new(None, Some(account_id.clone()), None);
+    env.set_pk(near_sdk::PublicKey::from(tls_key));
+    contract
+        .register_foreign_chains_config(foreign_chains_config)
+        .expect("register should succeed");
+}
```
gilcu3, line 312: the need to moves these two here relates to the doubts I had when doing the contract refactor. I believe we should probably have a module for foreign chains including the two existing files, instead of moving these functions to this level. But that can be done later ofc if ever, this is just a nit

crates/contract/tests/sandbox/common.rs
```diff
@@ -662,26 +663,51 @@ fn hash(code: &[u8]) -> [u8; 32] {
...
+            let result = account
+                .call_mpc(contract.id())
+                .vote_update_foreign_chain_providers(batch)
+                .await
+                .unwrap()
+                .into_result();
+            assert!(result.is_ok(), "whitelist vote should succeed: {result:?}");
+        }
+    });
+    futures::future::join_all(votes).await;
```
gilcu3: usually we have a better way to do this, as our testing code supports submitting transactions async, which would make the futures join_all not needed. Maybe it should be used here, the pattern exists already in the sandbox

crates/contract/tests/sandbox/common.rs
```diff
@@ -662,26 +663,51 @@ fn hash(code: &[u8]) -> [u8; 32] {
...
+                .unwrap()
+                .into_result();
+            assert!(
+                result.is_ok(),
+                "{} should succeed: {result:?}",
+                method_names::REGISTER_FOREIGN_CHAINS_CONFIG
+            );
+        }
+    });
+    futures::future::join_all(registrations).await;
```
gilcu3: same here

crates/node/src/network/conn.rs
```diff
@@ -37,7 +37,7 @@ impl<T: Send + Sync + 'static> Clone for ConnectionWithVersion<T> {
     }
 }
 
-impl<T: Send + Sync + 'static> ConnectionWithVersion<T> {
+impl<T> ConnectionWithVersion<T> {
```
gilcu3, line 40: broken window fix?

docs/hot-tee-signing-design.md
gilcu3, line 1: hmm do we need to modify this one?

docs/design/calculating-supported-foreign-chains.md
gilcu3, line 1: this one I would not modify, as it is a design doc long implemented. But it is borderline in this PR

gilcu3 (dismissed):
Thank you!

crates/contract/src/api/foreign_chain.rs
anodar: Discussed offline, there's a bit more needed (such as migration logic) for this and it will be addressed as part of https://github.com/near/mpc/issues/3630

docs/design/calculating-supported-foreign-chains.md
anodar: I agree in general design docs ideally shouldn't be modified long after it's implemented. But this design doc was drafted (2 months ago) as part of this current robustness work, it just took longer than we expected to get to this point. 

This doc already describes this new method of calculating available chains based on quorum, it didn't ever describe existing supported chains list (interstection).

So I'd say we leave as is unless you object.

docs/hot-tee-signing-design.md
anodar: I don't know why, but this doc lists contract methods and one of them is the one we're deprecating and switching to the new one (`get_supported_foreign_chains` -> `get_available_foreign_chains`). Hence the update here.

crates/node/src/network/conn.rs
anodar: Yeah carry over from dropping trait bounds from structs, but this time on impls 👀  They don't seem to be needed.

crates/contract/tests/sandbox/common.rs
anodar: Nice, I assume it's `execute_async_handle_calls` you are referring to. Indeed makes it nicer, thank you! 

Addressed in [ab2c1c6](https://github.com/near/mpc/pull/4234/commits/ab2c1c653001a17a8cfaae8171326a86599b2647)

crates/contract/tests/sandbox/common.rs
anodar: Addressed in [ab2c1c6](https://github.com/near/mpc/pull/4234/commits/ab2c1c653001a17a8cfaae8171326a86599b2647)

crates/contract/src/api/foreign_chain_support.rs
anodar: You're right, this should be a TODO instead. Addressed in [ab2c1c6](https://github.com/near/mpc/pull/4234/commits/ab2c1c653001a17a8cfaae8171326a86599b2647)

crates/contract/src/api/foreign_chain_support.rs
anodar: Same here.

crates/contract/src/api/test_utils.rs
anodar: Factoring out `foreign_chain.rs`, `foreign_chain_support.rs` into separate module makes sense. Will follow up on that.

docs/hot-tee-signing-design.md
gilcu3: Ahh this relates to the conversation over whether design docs should be kept in sync. We didn't have time in the retro so probably we will have it another time. The idea I had at least is that once a design doc fulfills its purpose, such as when it gets implemented, changing it in the future actually damage its utility. Not all design docs become system docs thereafter.

docs/design/calculating-supported-foreign-chains.md
gilcu3: Yeah this one I am fine keeping, hopefully we get it done soon though :)

gilcu3 (approved):

crates/contract/src/api/test_utils.rs
anodar: Followup in: https://github.com/near/mpc/pull/4274

netrome (approved):

PR #4235, opened 2026-08-25
feat(foreign-tx): ethereum support

frolvanya:
Adds Ethereum mainnet to foreign transaction verification. With this, every
EVM chain supported by omni-bridge (Eth, Arb, Base, Bnb, Pol, HlEvm, Abs) is
covered. Like Avalanche and ADI (#4145), this is mechanical wiring — Ethereum
reuses the generic `EvmInspector` the same way the other EVM chains do.

## The chain

* **Ethereum** — chain id 1 (Sepolia 11155111). The one EVM chain with real
  `safe`/`finalized` semantics: `finalized` lags `latest` by ~13 minutes
  (two epochs), `safe` by one. The inspector's existing finality handling
  covers all three tags.

## Changes

Ethereum has been a `ForeignChain`/`ForeignChainRpcRequest` variant since the
beginning (discriminant 2), and already had a node config slot, web-UI
provider counts, and contract sandbox test rows — so unlike #4145 there are
**no contract-interface changes and no ABI/borsh snapshot churn**. What was
missing was everything behind the enum:

* inspector marker module (`foreign_chain_inspector::ethereum`) + the
  `eth_chainId` fingerprint constant;
* node dispatch: an `EthereumInspector` fan-out slot and a real
  `ForeignChainRpcRequest::Ethereum` arm in `sign.rs`, replacing the
  `bail!("… is unsupported")`;
* health check: golden vector + `run_evm::<Ethereum>` (was `mark_skipped`),
  fingerprint probe arm, and the matching test rows;
* SDK builder (`new_ethereum()`) and tests;
* e2e coverage: Ethereum joins the mock cluster in
  `verify_foreign_transaction__should_sign_all_supported_chains`; the
  negative "chain not in config" branch now uses Solana, the remaining
  requestable-but-unsupported chain;
* docs tables (`foreign-chain-transactions.md`): ethereum moves from the
  "no inspector" carve-outs into the `eth_chainId` probe row and the
  fingerprint table.

The pinned golden transaction is
`0x7f1c6a58dc880438236d0b0a4ae166e9e9a038dbea8ec074149bd8b176332cac`
(block 4,476,790, hash `0x34e5a6cf…76c9`): post-Byzantium (so the receipt
carries `status: 1`, which the inspector requires) with one log at
block-wide index 0x10. Values cross-checked against two independent
generations of ethers.js provider test fixtures (v5.7 and v6), which are
CI-validated against live mainnet.

## Testing

* clippy / fmt / touched crates' tests pass.
* The new `#[ignore]`d manual tests mirror the Avalanche/ADI ones
  (block-hash + log extraction on the pinned tx, `eth_chainId` fingerprint 1)
  — run them against a live RPC before merging; the sandbox this change was
  authored in has no egress to RPC providers.

### DIFF
--- a/crates/e2e-tests/tests/foreign_chain_tx_validation.rs
+++ b/crates/e2e-tests/tests/foreign_chain_tx_validation.rs
@@ -19,8 +19,9 @@ use near_mpc_contract_interface::types::{
     BitcoinExtractor, BitcoinRpcRequest, BitcoinTxId, BlockConfirmations, DomainConfig, DomainId,
     DomainPurpose, EvmExtractor, EvmFinality, EvmRpcRequest, EvmTxId, ForeignChain,
     ForeignChainRpcRequest, ForeignTxPayloadVersion, Protocol, ReconstructionThreshold,
-    StarknetExtractor, StarknetFelt, StarknetFinality, StarknetRpcRequest, StarknetTxId,
-    VerifyForeignTransactionRequestArgs, VerifyForeignTransactionResponse,
+    SolanaFinality, SolanaRpcRequest, SolanaTxId, StarknetExtractor, StarknetFelt,
+    StarknetFinality, StarknetRpcRequest, StarknetTxId, VerifyForeignTransactionRequestArgs,
+    VerifyForeignTransactionResponse,
 };
 use near_mpc_sdk::foreign_chain::ForeignChainRequestBuilder;
 
@@ -56,6 +57,7 @@ struct MockServerUrls {
     hyper_evm: String,
     avalanche: String,
     adi: String,
+    ethereum: String,
     polygon: Vec<String>,
 }
 
@@ -118,6 +120,7 @@ fn build_foreign_chains_config(urls: &MockServerUrls) -> ForeignChainsConfig {
         hyper_evm: Some(mock_chain(&urls.hyper_evm, Default::default())),
         avalanche: Some(mock_chain(&urls.avalanche, Default::default())),
         adi: Some(mock_chain(&urls.adi, Default::default())),
+        ethereum: Some(mock_chain(&urls.ethereum, Default::default())),
         polygon: Some(mock_chain_with_providers(
             common::build_providers_from_urls(&urls.polygon, "polygon"),
         )),
@@ -135,6 +138,7 @@ async fn setup_foreign_tx_cluster() -> anyhow::Result<ForeignTxTestEnv> {
     let hyper_evm_server = MockServer::start();
     let avalanche_server = MockServer::start();
     let adi_server = MockServer::start();
+    let ethereum_server = MockServer::start();
 
     let bitcoin_mock_id = setup_bitcoin_mock(
         &bitcoin_server,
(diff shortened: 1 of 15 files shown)


frolvanya:
@claude review

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
```diff
@@ -532,6 +540,25 @@ async fn verify_adi(env: &ForeignTxTestEnv) -> anyhow::Result<()> {
...
+async fn verify_ethereum(env: &ForeignTxTestEnv) -> anyhow::Result<()> {
+    let request = VerifyForeignTransactionRequestArgs {
+        request: ForeignChainRpcRequest::Ethereum(EvmRpcRequest {
+            tx_id: EvmTxId([0xbb; 32]),
+            extractors: vec![EvmExtractor::BlockHash, EvmExtractor::Log { log_index: 0 }],
+            finality: EvmFinality::Finalized,
+        }),
+        domain_id: env.foreign_tx_domain_id,
+        payload_version: ForeignTxPayloadVersion::V1,
+        expected_payload_hash: None,
```
anodar, line 552: nit: let's include expected hash to verify that we check against it.

crates/foreign-chain-inspector/tests/ethereum_rpc_manual.rs
```diff
@@ -0,0 +1,83 @@
...
+    ethereum::{
+        EthereumBlockHash, EthereumTransactionHash, MAINNET_CHAIN_ID,
+        inspector::{EthereumExtractedValue, EthereumExtractor, EthereumInspector},
+    },
+};
+
+const ETHEREUM_RPC_URL: &str = "https://ethereum-rpc.publicnode.com";
+
+#[tokio::test]
+#[ignore = "manual test to sanity check against live Ethereum RPC provider"]
```
anodar, line 13: [no action required] just some thoughts, we should do something differently with these manual tests. It's nice for debugging while one is developing this but then I doubt anyone ever runs this. That's why we silently get outdated transactions in the config checker as well.

On the other hand not a good practice to rely on that external RPC in tests. 

I'd still lean towards marking this test not manual but have it as non-required check for merging.

anodar (approved):

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
frolvanya: Hmm, it's `None` for all other `verify_*` methods. Should I add expected hash only for ethereum or for all chains?

P.S: This would add quite a lot of code to setup expected log + hash and response validation

crates/foreign-chain-inspector/tests/ethereum_rpc_manual.rs
frolvanya: > On the other hand not a good practice to rely on that external RPC in tests.

I agree, it would cause flaky ci. Even if it's not required, it's not really convenient to check every time if this specific test failed or we have a new failure

crates/e2e-tests/tests/foreign_chain_tx_validation.rs
anodar: I meant only for ethereum (since your code only touches that), similar to`verify_bitcoin_bound_to_expected_payload_hash` from `verify_foreign_transaction__should_sign_all_supported_chains`  test. 

But again, it's a nit (optional).

crates/foreign-chain-inspector/tests/ethereum_rpc_manual.rs
netrome: I've found them helpful when I want to quickly fiddle or play around with the inspectors. They also provide a nice illustrative example of how to run the inspector. Typically it's just a matter of updating them to a recent transaction hash to make it work.

However I'd be happy to torch them as they are effectively dead code.

netrome (approved):
Nice stuff, thanks! 🙏

Issue #3786, opened 2026-07-13
Follow-up (#3293): stamp expiry on genesis mock sentinels and drop node mock-identity fallback


barakeinav1:
### Background

Follow-up to #3293 / #3785. Two loose ends remain after making mock attestations cleanable:

1. The genesis sentinel path `TeeState::with_mocked_participant_attestations` (used by `init` / `init_running`, `crates/contract/src/tee/tee_state.rs`) still stores non-expiring `MockAttestation::Valid`. #3785 stamps an expiry on submitted mocks and migrates existing stored entries, but a fresh init can still create a mock that never expires and can never be cleaned up. Stamping these too changes the #1087 placeholder semantics (a participant that never submits a real attestation would auto-expire), which is why it was deferred.
2. Because a no-expiry mock can still exist, the node keeps a backward-compat fallback in `submitted_attestation_landed` (`crates/node/src/indexer/tx_sender.rs`): a mock with no stored expiry is matched by identity instead of the expiry-changed heuristic. Once every stored mock carries an expiry, that `None` arm is dead code and the mock case collapses into the same heuristic Dstack uses.

### Acceptance Criteria

- [ ] `with_mocked_participant_attestations` stamps an expiry on the mock sentinels it creates (decide on the #1087 auto-expiry behavior first).
- [ ] The node's `submitted_attestation_landed` mock-identity fallback (`None` arm) is removed once all stored mocks are guaranteed to carry an expiry.

### Resources & Additional Notes

- Depends on #3785 being deployed (contract upgraded + migration run).
- Related: #1087 (thread real attestations through initialization so the sentinel can go away), and the `TODO(#1639)` expiry heuristic which would eventually replace this landing check entirely.

barakeinav1:
Checked the #3785 deployment precondition while working on #3951.

**#3785 is deployed on both networks:** `version()` returns 3.14.0 on `v1.signer` and 3.14.1 on `v1.signer-prod.testnet`. The stamping commit (b4bea097) is contained in 3.14.0, so both carry it.

**No stored entry lacks an expiry today.** Swept every `get_tee_accounts` entry, not just participants — 15 on mainnet (9 `Mock::WithConstraints`, 6 `Dstack`), 20 on testnet (18 `Dstack`, 2 `Mock::WithConstraints`, including 3 stale non-participant rows). All 35 carry an expiry, so the #3785 migration plus subsequent re-stamping has cleared the legacy rows.

[...]

PR #4236, opened 2026-08-25
feat(node): metrics for attestation freshness

### LINKED ISSUE #3951: feat(node): expose attestation-freshness metrics on /metrics
### Background

In the 3.13.x incident, collateral fetches failed for hours and every TEE node was ~24h from being silently dropped from the participant set. Nothing alerted — we found it by hand-querying `get_attestation`. The node publishes no signal for how stale its attestation is.

Node side only; the Grafana/Prometheus half is https://github.com/near/mpc-private/issues/559, which needs upgraded nodes first.

**Proposed design.** Two gauges, written where the node already confirms a submission on chain (#3736):

- the expiry the contract stores for our TLS key — how long until eviction;
- when a submission last landed — how long re-attestation has been failing.

Both as absolute Unix timestamps rather than remaining durations, so a gauge frozen by a node that stopped re-attesting keeps decaying towards now and still trips its threshold. A single failed attempt is absorbed by the next hourly retry, so only sustained failure moves the signal. Documented thresholds stay relative to the configured expiry window rather than hard-coded, so they survive changes to it (cf. #3947).

### Acceptance Criteria

- [ ] The node exposes attestation freshness on `/metrics`.
- [ ] The signal distinguishes a sporadic failure from a sustained one.
- [ ] Metrics and suggested alert thresholds are documented for operators.

- PR: #4236
- Related: #3049, #3736, #3947, and #3752 whose gauge/sentinel convention this follows.
- DCAP freshness design doc, Gap 4 (no observability): `near/mpc-private` `research/dcap-collateral-freshness`.
- `mpc_tee_attestation_attempts_total` (#2727) and `mpc_own_image_hash_*` (#3752) already exist — the freshness family is what's missing.

barakeinav1:
Closes #3951

Two gauges written where the node already confirms a submission on chain (#3736), so no new RPC:

- `mpc_attestation_last_landed_timestamp_seconds` — advances only on a confirmed landing, so the gap to now is how long re-attestation has been failing. One failed attempt is absorbed by the next hourly tick; only sustained failure grows it.
- `mpc_attestation_expiry_timestamp_seconds` — the expiry the contract stores for our TLS key, so time-to-eviction is directly visible. Sentinels follow #3752: `0` nothing stored, `-1` stored without an expiry (defensive — the contract stamps every attestation it accepts).

Absolute timestamps rather than remaining durations: a node that stops re-attesting also stops updating the gauges, and only the absolute form keeps decaying towards now, so the alert still fires on a frozen gauge. A frozen "seconds remaining" would sit at a healthy value forever, which is the silent failure this fixes.

Metric interpretation, alert expressions, and the query separating one bad node from a fleet-wide cause are in `docs/design/node-operator-metrics.md`.

Alerting on these lives in https://github.com/near/mpc-private/issues/559 and needs deployed nodes first. Not covered: a node that is fully down publishes nothing, and we don't scrape third-party operators.

### DIFF
--- a/crates/node/src/metrics.rs
+++ b/crates/node/src/metrics.rs
@@ -484,8 +484,9 @@ pub static MPC_ATTESTATION_EXPIRY_TIMESTAMP_SECONDS: LazyLock<prometheus::IntGau
     LazyLock::new(|| {
         prometheus::register_int_gauge!(
             "mpc_attestation_expiry_timestamp_seconds",
-            "Unix time at which the attestation stored on chain for this node's TLS key expires \
-             (-1 if the stored attestation carries no expiry, 0 if none is stored)"
+            "Unix time at which the attestation stored on chain for this node's TLS key expires. \
+             -1 if the stored attestation carries no expiry; 0 if none is stored. Compare against \
+             mpc_indexer_latest_block_timestamp_seconds for the remaining time"
         )
         .unwrap()
     });
@@ -498,3 +499,14 @@ pub static MPC_ATTESTATION_LAST_LANDED_TIMESTAMP_SECONDS: LazyLock<prometheus::I
         )
         .unwrap()
     });
+
+/// Registers the attestation-freshness gauges, which are otherwise only touched once a submission
+/// reaches the chain.
+///
+/// A metric registers on first dereference, so without this a node that never gets that far
+/// exports no series at all — and an alert on an absent series never fires, in exactly the failure
+/// mode these gauges exist to catch.
+pub fn init_attestation_freshness_metrics() {
+    LazyLock::force(&MPC_ATTESTATION_EXPIRY_TIMESTAMP_SECONDS);
+    LazyLock::force(&MPC_ATTESTATION_LAST_LANDED_TIMESTAMP_SECONDS);
+}
--- a/crates/node/src/metrics.rs
+++ b/crates/node/src/metrics.rs
@@ -484,9 +484,10 @@ pub static MPC_ATTESTATION_EXPIRY_TIMESTAMP_SECONDS: LazyLock<prometheus::IntGau
     LazyLock::new(|| {
         prometheus::register_int_gauge!(
             "mpc_attestation_expiry_timestamp_seconds",
-            "Unix time at which the attestation stored on chain for this node's TLS key expires. \
-             -1 if the stored attestation carries no expiry; 0 if none is stored. Compare against \
-             mpc_indexer_latest_block_timestamp_seconds for the remaining time"
+            "NEAR block time at which the attestation stored on chain for this node's TLS key \
+             expires. -1 if the stored attestation carries no expiry; 0 if none is stored. \
+             Subtract mpc_indexer_latest_block_timestamp_seconds, not wall clock, for the \
+             remaining time"
         )
         .unwrap()
     });
@@ -495,17 +496,13 @@ pub static MPC_ATTESTATION_LAST_LANDED_TIMESTAMP_SECONDS: LazyLock<prometheus::I
     LazyLock::new(|| {
         prometheus::register_int_gauge!(
             "mpc_attestation_last_landed_timestamp_seconds",
-            "Unix time at which this node last confirmed an attestation submission landed on chain"
+            "Unix time, by this node's own clock, at which it last confirmed an attestation \
+             submission landed on chain"
         )
         .unwrap()
     });
 
-/// Registers the attestation-freshness gauges, which are otherwise only touched once a submission
-/// reaches the chain.
-///
-/// A metric registers on first dereference, so without this a node that never gets that far
-/// exports no series at all — and an alert on an absent series never fires, in exactly the failure
-/// mode these gauges exist to catch.
+/// An alert cannot fire on a series that does not exist yet.
 pub fn init_attestation_freshness_metrics() {
     LazyLock::force(&MPC_ATTESTATION_EXPIRY_TIMESTAMP_SECONDS);
     LazyLock::force(&MPC_ATTESTATION_LAST_LANDED_TIMESTAMP_SECONDS);
--- a/crates/node/src/metrics.rs
+++ b/crates/node/src/metrics.rs
@@ -479,3 +479,22 @@ pub static MPC_TEE_ATTESTATION_ATTEMPTS_TOTAL: LazyLock<prometheus::IntCounterVe
 
 pub const MPC_TEE_ATTESTATION_OUTCOME_SUCCESS: &str = "success";
 pub const MPC_TEE_ATTESTATION_OUTCOME_FAILURE: &str = "failure";
+
+pub static MPC_ATTESTATION_EXPIRY_TIMESTAMP_SECONDS: LazyLock<prometheus::IntGauge> =
+    LazyLock::new(|| {
+        prometheus::register_int_gauge!(
+            "mpc_attestation_expiry_timestamp_seconds",
+            "Unix time at which the attestation stored on chain for this node's TLS key expires \
+             (-1 if the stored attestation carries no expiry, 0 if none is stored)"
+        )
+        .unwrap()
+    });
+
+pub static MPC_ATTESTATION_LAST_LANDED_TIMESTAMP_SECONDS: LazyLock<prometheus::IntGauge> =
+    LazyLock::new(|| {
+        prometheus::register_int_gauge!(
+            "mpc_attestation_last_landed_timestamp_seconds",
+            "Unix time at which this node last confirmed an attestation submission landed on chain"
+        )
+        .unwrap()
+    });
--- a/crates/node/src/tee/attestation_freshness_metrics.rs
+++ b/crates/node/src/tee/attestation_freshness_metrics.rs
@@ -1,0 +1,68 @@
+//! Attestation-freshness gauges.
+//!
+//! Both are absolute timestamps rather than remaining durations: a node that stops re-attesting
+//! also stops updating them, and only the absolute form keeps decaying towards now, so a staleness
+//! alert still fires on a frozen gauge. Each landed submission moves them, so at rest they
+//! advance hourly.
+
+use near_mpc_contract_interface::types::VerifiedAttestation;
+use near_time::Clock;
+
+use crate::metrics::{
+    MPC_ATTESTATION_EXPIRY_TIMESTAMP_SECONDS, MPC_ATTESTATION_LAST_LANDED_TIMESTAMP_SECONDS,
+};
+
+const NO_ATTESTATION_STORED: i64 = 0;
+const NO_EXPIRY: i64 = -1;
+
+pub(crate) fn record_stored_attestation_expiry(stored: Option<&VerifiedAttestation>) {
+    MPC_ATTESTATION_EXPIRY_TIMESTAMP_SECONDS.set(expiry_gauge_value(stored));
+}
+
+pub(crate) fn record_attestation_landed() {
+    MPC_ATTESTATION_LAST_LANDED_TIMESTAMP_SECONDS.set(Clock::real().now_utc().unix_timestamp());
+}
+
+fn expiry_gauge_value(stored: Option<&VerifiedAttestation>) -> i64 {
+    match stored.map(VerifiedAttestation::expiry_timestamp_seconds) {
+        None => NO_ATTESTATION_STORED,
+        Some(None) => NO_EXPIRY,
+        Some(Some(expiry)) => i64::try_from(expiry).unwrap_or(i64::MAX),
+    }
+}
+
+#[cfg(test)]
+#[expect(non_snake_case)]
+mod tests {
+    use super::*;
+    use near_mpc_contract_interface::types::MockAttestation;
+    use rstest::rstest;
+
+    const EXPIRES_AT: u64 = 1_754_000_000;
+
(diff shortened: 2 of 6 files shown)


barakeinav1:
Thanks — checked each finding against the code. Fixed in e8c28c42.

> Both attestation alerts are unfirable in exactly the failure mode they are meant to catch

Correct, and worse than stated: gauges are per-process, so this recurs on every restart. `init_attestation_freshness_metrics()` now forces both next to `init_build_info_metric()` in `web::metrics()`, so the series exist from the first scrape. Fixed.

> `record_attestation_landed` has no test

Fixed — takes `&Clock`, asserted from a `FakeClock`. Note the fix does not address the "reverting either call site would not fail any test" part: that needs a test of `observe_tx_result` with a fake indexer, which is disproportionate here.

> A `-1` gauge (stored without expiry) satisfies neither `> 0` nor `== 0`

Right for the two expiry alerts, but not "all three" — the staleness alert is on `last_landed` and still covers such a node. Documented that; no new alert.

> The sibling `MPC_OWN_IMAGE_HASH_EXPIRY_TIMESTAMP_SECONDS` description tells the operator to compare against `mpc_indexer_latest_block_timestamp_seconds`

Fixed, HELP text now matches the sibling.

> It also says "Each landed submission moves them", which reads as if both gauges only move on a landing

Fair, reworded. Keeping the absolute-vs-relative rationale: it is the invariant a future "just export seconds remaining" change would silently break, which `CLAUDE.md` treats as worth a comment.

> `const NO_EXPIRY: i64 = -1;` duplicates `image_expiry_metrics.rs:9`

Declining. Independent metric families sharing a convention; hoisting couples them and widens the diff, and each value is pinned by its own HELP text and alerts.

> `RealAttestationExpiryReader::read_stored_attestation_expiry` ... runs on every hourly tick, *before* attestation generation can fail

It runs *after*. `generate_attestation` is at `remote_attestation.rs:180` and `continue`s on failure at `:194`; the read is at `:211`. On the collateral-failure path it is never reached, so the proposed benefit does not exist. Declining.

crates/node/src/metrics.rs
```diff
@@ -479,3 +479,34 @@ pub static MPC_TEE_ATTESTATION_ATTEMPTS_TOTAL: LazyLock<prometheus::IntCounterVe
...
+        .unwrap()
+    });
+
+/// Registers the attestation-freshness gauges, which are otherwise only touched once a submission
+/// reaches the chain.
+///
+/// A metric registers on first dereference, so without this a node that never gets that far
+/// exports no series at all — and an alert on an absent series never fires, in exactly the failure
+/// mode these gauges exist to catch.
+pub fn init_attestation_freshness_metrics() {
```
kevindeforth, line 506: Please refine this comment to be in line with our engineering standards. Every sentence is violating one of our rules.

crates/node/src/metrics.rs
```diff
@@ -479,3 +479,34 @@ pub static MPC_TEE_ATTESTATION_ATTEMPTS_TOTAL: LazyLock<prometheus::IntCounterVe
 
 pub const MPC_TEE_ATTESTATION_OUTCOME_SUCCESS: &str = "success";
 pub const MPC_TEE_ATTESTATION_OUTCOME_FAILURE: &str = "failure";
+
+pub static MPC_ATTESTATION_EXPIRY_TIMESTAMP_SECONDS: LazyLock<prometheus::IntGauge> =
+    LazyLock::new(|| {
+        prometheus::register_int_gauge!(
+            "mpc_attestation_expiry_timestamp_seconds",
+            "Unix time at which the attestation stored on chain for this node's TLS key expires. \
```
kevindeforth: Is this timestamp coming from the NEAR blockchain? If so, then I think we should mention that, because the two may not align perfectly (I think we can have up to a few minutes of divergence).

docs/design/node-operator-metrics.md
```diff
@@ -30,6 +30,15 @@ serves keyshares over the migration service to the backup service registered for
...
 | [`mpc_current_epoch_id`](../../crates/node/src/metrics.rs) | epoch id of the keyset the contract currently holds | the reference point for `mpc_last_backup_served_epoch`. Increments on every resharing; unset until the first keyset exists. During a resharing it stays at the old epoch, which is the one still available to back up. |
 
+Attestation freshness, in [`metrics.rs`](../../crates/node/src/metrics.rs). Set from the
+on-chain confirmation of an attestation submission, in
+[`attestation_freshness_metrics.rs`](../../crates/node/src/tee/attestation_freshness_metrics.rs):
+
+| Metric | Measures | How to interpret |
+| --- | --- | --- |
+| [`mpc_attestation_last_landed_timestamp_seconds`](../../crates/node/src/metrics.rs) | Unix time of the last attestation submission this node confirmed on chain | should be under an `ATTESTATION_RESUBMISSION_INTERVAL` (1h) old. Only a confirmed landing advances it, so the gap to now is how long re-attestation has been failing, wherever it broke. One failed attempt is absorbed by the next tick; a growing gap ends in eviction from the participant set. |
```
kevindeforth:
```suggestion
| [`mpc_attestation_last_landed_timestamp_seconds`](../../crates/node/src/metrics.rs) | Unix time of the last attestation submission this node confirmed on chain | should be under an `ATTESTATION_RESUBMISSION_INTERVAL` (1h) old. |
```

The last two sentences are discussing implementation details that are not relevant to the average node operator.

crates/node/src/metrics.rs
```diff
@@ -479,3 +479,34 @@ pub static MPC_TEE_ATTESTATION_ATTEMPTS_TOTAL: LazyLock<prometheus::IntCounterVe
...
+             -1 if the stored attestation carries no expiry; 0 if none is stored. Compare against \
```
kevindeforth: Hmm, that begs the question if `VerifiedAttestation` must still support this or not. I assume the contract updated a long time ago and we had resharings in the meantime, which should, theoretically, remove any non-participant.
So, we should only have attestations with timestamps now?

It's something to investigate and not a blocker, as we can remove the Optional from the metric quite easily.

docs/design/node-operator-metrics.md
```diff
@@ -30,6 +30,15 @@ serves keyshares over the migration service to the backup service registered for
...
+| [`mpc_attestation_expiry_timestamp_seconds`](../../crates/node/src/metrics.rs) | Unix time at which the attestation the contract stores for this node's TLS key expires | should sit a full expiry window ahead of chain time and step forward hourly. `0` = nothing stored (evicted, or never landed one), `-1` = stored without an expiry (legacy entries only; the contract stamps every attestation it accepts). Compare against `mpc_indexer_latest_block_timestamp_seconds`, the clock the contract expires entries against. |
```
kevindeforth: Here, we should again check if we are talking near-blockchain time or unix time.

> should sit a full expiry window ahead of chain time

I think we are leaking implementation details here. If operators want to write an alert based on this sentence, they would need to know the exact expiry window and where to look it up.

> and step forward hourly

I think this is more useful, but may also be wrong if we change the contract and expiration implementation in the future. Maybe, we can give some concise actionable advise like:  _Should be at least X days from now._  Wdyt?

kevindeforth (commented):
Thank you very much! There is one code-comment that needs cleanup and I have some suggestions for the metric / alert-doc, but the implementation itself looks good!

crates/node/src/tee/attestation_freshness_metrics.rs
```diff
@@ -0,0 +1,84 @@
...
+
+const NO_ATTESTATION_STORED: i64 = 0;
+const NO_EXPIRY: i64 = -1;
+
+pub(crate) fn record_stored_attestation_expiry(stored: Option<&VerifiedAttestation>) {
+    MPC_ATTESTATION_EXPIRY_TIMESTAMP_SECONDS.set(expiry_gauge_value(stored));
+}
+
+pub(crate) fn record_attestation_landed(clock: &Clock) {
+    MPC_ATTESTATION_LAST_LANDED_TIMESTAMP_SECONDS.set(clock.now_utc().unix_timestamp());
```
SimonRastikian, line 23: Why do you not have Clock::real() inside this? Then the function would take no inputs

crates/node/src/tee/attestation_freshness_metrics.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+fn expiry_gauge_value(stored: Option<&VerifiedAttestation>) -> i64 {
+    match stored.map(VerifiedAttestation::expiry_timestamp_seconds) {
+        None => NO_ATTESTATION_STORED,
+        Some(None) => NO_EXPIRY,
```
SimonRastikian, line 29: That does this mean in real world, which attestation has no expiry?

crates/node/src/tee/attestation_freshness_metrics.rs
SimonRastikian: Oh I see it's for unit testing... still would be interested in your opinion

crates/node/src/tee/attestation_freshness_metrics.rs
```diff
@@ -0,0 +1,84 @@
...
+
+use near_mpc_contract_interface::types::VerifiedAttestation;
+use near_time::Clock;
+
+use crate::metrics::{
+    MPC_ATTESTATION_EXPIRY_TIMESTAMP_SECONDS, MPC_ATTESTATION_LAST_LANDED_TIMESTAMP_SECONDS,
+};
+
+const NO_ATTESTATION_STORED: i64 = 0;
+const NO_EXPIRY: i64 = -1;
```
SimonRastikian, line 16: You could create an enum with values NO_ATTESTATION_STORED, NO_EXPIRY, EXPIRES_AT(i64).
It would be nicer I think

SimonRastikian (dismissed):
Approval with nits

crates/node/src/metrics.rs
barakeinav1: update in commit  https://github.com/near/mpc/pull/4236/changes/649044271b49d520558b363fcdcc1e5f7272a11e

crates/node/src/metrics.rs
barakeinav1: Yes — the contract stamps it from `env::block_timestamp_ms()` (`tee_state.rs:149`), so it is NEAR block time.

Worth more than a mention: the *other* gauge is set from the node's own clock, so the two are on different clocks and mixing them up gives a wrong alert. Both HELP strings now name their clock, and the expiry one says to subtract `mpc_indexer_latest_block_timestamp_seconds` rather than wall clock.

docs/design/node-operator-metrics.md
barakeinav1: updated, with a compromise, see in commit https://github.com/near/mpc/commit/649044271b49d520558b363fcdcc1e5f7272a11e

crates/node/src/metrics.rs
barakeinav1: Agreed on the diagnosis — live state is clean (I had PR to clean mock attestation with no expiration) 

but, It cannot come out yet though: `with_mocked_participant_attestations` (`crates/contract/src/tee/tee_state.rs:108`) still inserts a bare `MockAttestation::Valid` straight into `stored_attestations`, bypassing `with_expiry_capped_at` — so a fresh `init` / `init_running` recreates the no-expiry case.

Already tracked in #3786: AC 1 stamps the genesis sentinels, AC 2 then drops this. Deployment and live-state evidence left there — https://github.com/near/mpc/issues/3786#issuecomment-5412155065

docs/design/node-operator-metrics.md
barakeinav1: Both good points, and they pushed me off the fraction-of-window idea entirely.

NEAR block time, as above.

On the window: you are right it leaks. Went with a plain duration instead, which needs no knowledge of the window — page when under 3 days of runway remain.

The only assumption left is that the window stays above 3 days. Growing it (7d → 31d) needs no change; only shrinking it below ~3 days would. The alert comment says to keep the threshold below the window for that reason.

crates/node/src/tee/attestation_freshness_metrics.rs
kevindeforth: Yeah, I don't think we need this, but we need to clean up the interface to get rid of this (c.f. https://github.com/near/mpc/pull/4236#discussion_r3870038774)

crates/node/src/tee/attestation_freshness_metrics.rs
barakeinav1: only for unit test, originally I didn't have this parameter, then the Claude bot flagged that we don't have a unit test for this. so I added it.

Also think it's the right call regardless: engineering standards ask for time to be injected, and the test guards a seconds-vs-milliseconds mix-up.

crates/node/src/tee/attestation_freshness_metrics.rs
barakeinav1: see repose in here https://github.com/near/mpc/pull/4236#discussion_r3870533062

crates/node/src/tee/attestation_freshness_metrics.rs
barakeinav1: I'd lean towards keeping the consts. The gauge takes an i64, so an enum needs a conversion on top and the 6-line match becomes ~15 for the same mapping. also I follow the pattern from image_expiry_metrics.rs

And `NO_EXPIRY` should disappear once #3786 lands anyway.

SimonRastikian (approved):

kevindeforth (approved):
Thanks!

PR #4237, opened 2026-08-25
refiner with assiciated type on Request to avoid box dyn

anodar:


crates/node/src/requests/queue.rs
```diff
@@ -956,13 +972,11 @@ mod tests {
                 .add_block(&new_block.to_block_view())
                 .block_status;
             let update = RequestsUpdate {
-                requests: self.requests_to_submit.clone(),
-                completed_requests: self.responses_to_submit.clone(),
+                requests,
+                completed_requests: std::mem::take(&mut self.responses_to_submit),
```
gilcu3, line 976: why not just cloning here, these are tests so performance should not be a problem

crates/node/src/requests/queue.rs
```diff
@@ -1434,39 +1448,81 @@ mod tests {
...
+
+        fn get_id(&self) -> crate::types::RequestId {
+            self.0.get_id()
+        }
+
+        fn get_receipt_id(&self) -> CryptoHash {
+            self.0.get_receipt_id()
+        }
+
+        fn get_entropy(&self) -> [u8; 32] {
```
gilcu3, line 1466: not sure why is this needed, seems counter intuitive

PR #4238, opened 2026-08-25
refactor: Enforcing strong typing in governance.rs

SimonRastikian:
Enforces strong typing of threshold in test function.
Part of https://github.com/near/mpc/issues/3903

### DIFF
--- a/crates/contract/src/api/governance.rs
+++ b/crates/contract/src/api/governance.rs
@@ -166,14 +166,14 @@ mod tests {
     #[test]
     fn test_vote_new_parameters_succeeds_with_default_tee_status() {
         let (mut contract, participants, first_participant_id) = setup_tee_test_contract(3, 2);
-        let threshold = GovernanceThreshold::new(2);
+        let governance_threshold = GovernanceThreshold::new(2);
 
         // No attestations submitted - all participants have default TEE status None
         let result = setup_voting_context_and_vote(
             &mut contract,
             &first_participant_id,
             participants,
-            threshold,
+            governance_threshold,
         );
         assert!(
             result.is_ok(),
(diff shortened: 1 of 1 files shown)


crates/contract/src/api/governance.rs
```diff
@@ -253,13 +253,13 @@ mod tests {
...
-    /// reconstruction threshold is `reconstruction_threshold`.
+    /// Builds a Running contract with `num_participants` participants, the given
+    /// governance threshold, and a single CaitSith [`Sign`] domain with the given
+    /// reconstruction threshold.
     fn setup_running_contract_with_domain(
         num_participants: usize,
-        threshold: u64,
-        reconstruction_threshold: u64,
+        governance_threshold: GovernanceThreshold,
+        reconstruction_threshold: ReconstructionThreshold,
```
gilcu3, line 262: nit: it is a bit arguable if we should use the strong type here, in the sense that this is a test helper, so receiving u64 and converting internally makes life easier for the caller.

gilcu3 (dismissed):
Thank you!

crates/contract/src/api/governance.rs
SimonRastikian: I know, but I seriously wanted to get rid of the loose types. It's ok if strong typing is enforced here

crates/contract/src/api/governance.rs
```diff
@@ -143,7 +143,7 @@ mod tests {
         contract: &mut MpcContract,
         first_participant_id: &AccountId,
         participants: Participants,
-        threshold: GovernanceThreshold,
+        governance_threshold: GovernanceThreshold,
```
barakeinav1, line 146: nit: there are a few more places you can update the name as well if you want to be consistent - `let threshold = GovernanceThreshold::new(...)` to `let governance_threshold = GovernanceThreshold::new(...)` at 169, 191, 218 and 462.

barakeinav1 (dismissed):
thanks, one nit

SimonRastikian:
Thanks @barakeinav1 for the note

gilcu3 (approved):

barakeinav1 (approved):

PR #4242, opened 2026-08-25
refactor(probe): build every chain's inspector through one registry

haiyuechen-nearone:
Part of #4043.
Dispatch, client construction and fingerprint canonicalization were spread across a thirteen arm match. One registry entry per chain now binds a chain's canonical form to the constructor of its inspector, and the inspectors share one type.

### Notes for review

- The probe's signature is unchanged and its tests are untouched, so the existing suite is the parity check for this rewrite.
- `canonical_fingerprint` moves off `NetworkFingerprintInspector` onto each inspector. It normalizes text an operator wrote, which is a property of the chain rather than of an inspector value, and no generic caller reached it through the trait.


haiyuechen-nearone:
Folded into #4243: the registry rewrite and the injection it enables now land as one change, since reviewing them apart meant reading the seam twice.

PR #4243, opened 2026-08-25
refactor(probe): inject the inspectors into the startup probe

### LINKED ISSUE #4043: Inject the inspectors into the foreign chain startup probe

haiyuechen-nearone:
Closes #4043. Stacked on #4242.

The probe built its own RPC clients, so its tests stood up HTTP servers and waited on the wall clock. The node's startup path now hands the inspector constructor in, and the scripted inspector covers the retry, attempt cap and timeout cases under a paused clock.

### Notes for review

- The tests that still use real clients keep covering what the scripted ones cannot: auth splicing, fingerprint bounding and the per chain requests.
- Never combine the paused clock with those socket tests: the runtime advances the clock while the socket is silent.

### DIFF
--- a/crates/foreign-chain-health-check/Cargo.toml
+++ b/crates/foreign-chain-health-check/Cargo.toml
@@ -24,6 +24,7 @@ tokio = { workspace = true }
 
 [dev-dependencies]
 assert_matches = { workspace = true }
+foreign-chain-inspector = { workspace = true, features = ["test-utils"] }
 httpmock = { workspace = true }
 rstest = { workspace = true }
 serde_json = { workspace = true }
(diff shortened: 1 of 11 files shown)


haiyuechen-nearone:
Reopening on the correct base: this PR now carries the folded registry + injection change, but its base is pinned to the retired 4043-probe-inspector-registry branch and stack membership blocks editing it. Same branch, same commit, new PR.

PR #4244, opened 2026-08-25
refactor: inject the inspectors into RPC probing path

### LINKED ISSUE #4043: Inject the inspectors into the foreign chain startup probe

haiyuechen-nearone:
Closes #4043.

`probe_all_providers` now takes its inspectors as a dependency via the `BuildInspectors` trait. Production path uses a new struct `InspectorFactory` and tests can re-implement this trait and utilize `MockInspector`

`foreign-chain-rpc-auth` crate is renamed `foreign-chain-rpc-factory`, since it now includes code to resolve credentials and build inspectors.

SIde notes: 
- `timeout_of` moved onto `ForeignChainConfig::timeout_duration()`.
- My intension is that `foreign-chain-rpc-factory` will contain all client and inspector construction methods. To restrict the scope of this PR I purposefully tried not to touch the TX verification route.


haiyuechen-nearone:
@claude review

haiyuechen-nearone:
> crates/foreign-chain-rpc-factory/src/inspectors.rs:76 — the fix for the "credentials resolved before the chain-support check" defect has no regression test, and the suite cannot detect a reintroduction.

Our intension is to have inspectors for and probe all chains that are defined in our code base. Not supporting a chain is not the final state, I feel adding a test to check that "we are still in progress" is overkill.

crates/foreign-chain-inspector/src/mock.rs
```diff
@@ -0,0 +1,92 @@
...
+            ForeignChainInspectionError::RpcRequestRejected("mock refusal".to_string()),
+        )
+    }
+}
+
+/// Answers from a queue of [`MockReply`]s; panics on a call past the end of the queue.
+#[derive(Clone)]
+pub struct MockInspector {
+    replies: Arc<Mutex<VecDeque<MockReply>>>,
+    calls: Arc<AtomicUsize>,
```
netrome, line 51: Nit: `calls` can be derived from `replies` if the initial length is known so we could drop the atomic here.

crates/foreign-chain-inspector/src/mock.rs
```diff
@@ -0,0 +1,92 @@
...
+        self.calls.fetch_add(1, Ordering::SeqCst);
+        let reply = self
+            .replies
+            .lock()
+            .expect("replies mutex poisoned")
+            .pop_front()
+            .expect("call beyond the queued replies");
+        match reply {
+            MockReply::Answer { delay, fingerprint } => {
+                tokio::time::sleep(delay).await;
```
netrome, line 78: What's the purpose of adding a delay here? This feels strange.

netrome (approved):
Just some minor considerations from my side.

kevindeforth (approved):
nice, thank you!

crates/foreign-chain-inspector/src/mock.rs
haiyuechen-nearone: I wanted to use calls here to enable the test cases to verify that "Inspector was called x times", which is sort of independent from number of MockReplies configured.

crates/foreign-chain-inspector/src/mock.rs
haiyuechen-nearone: It is not used in this PR directly, but I added this so we can test Fanout concurrency. Aka given a Fanout with 1 inspector taking 1 sec and another taking 3 sec, Fanout should return after 3 sec not 4 sec.

netrome: Okay and how do you test that reliably? With tokio paused time and advance? That also reminds me, seems like we're not running these tests with paused tokio time - something we probably should do if we let tokio act as the time abstraction without controlling it ourselves.

haiyuechen-nearone: Yes my plan is to use tokio paused time. That is the next PR in the stack.

PR #4249, opened 2026-08-25
refactor: Suggestion to use generics instead of dynamic dispatch

netrome:
Suggestion for https://github.com/near/mpc/pull/4203

Trad-coded with limited understanding of the surrounding context. Just for illustration, should likely not be merged as-is.


netrome:
Closed as Nodar already independently implemented something very similar.

PR #4256, opened 2026-08-26
chore(contract): move away from near-sdk PublicKey type

### LINKED ISSUE #1246: convert contract state from near_sdk::PublicKey to dtos type

gilcu3:
Closes #1246

### DIFF
--- a/crates/contract/src/v3_14_0_state.rs
+++ b/crates/contract/src/v3_14_0_state.rs
@@ -16,17 +16,26 @@ use near_sdk::{
 
 use crate::{
     config::Config,
+    crypto_shared::types::{PublicKeyExtended, serializable::SerializableEdwardsPoint},
     foreign_chains_metadata::{ForeignChainsMetadata, SupportedForeignChainsByNode},
     node_migrations::NodeMigrations,
     primitives::{
         ckd::CKDRequest,
+        domain::{AddDomainsVotes, DomainRegistry},
+        key_state::{AttemptId, EpochId, KeyForDomain, Keyset},
         signature::{SignatureRequest, YieldIndex},
+        threshold_votes::GovernanceThresholdParametersVotes,
+        thresholds::GovernanceThresholdParameters,
     },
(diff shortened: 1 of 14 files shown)


crates/contract/src/v3_14_0_state.rs
gilcu3, line 1: Most changes are in this file, the PR itself is actually small

crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
```diff
@@ -3972,14 +3972,7 @@ expression: abi
...
         "Ed25519PublicKey": {
-          "type": "array",
-          "items": {
-            "type": "integer",
-            "format": "uint8",
-            "minimum": 0.0
-          },
-          "maxItems": 32,
-          "minItems": 32
+          "type": "string"
```
gilcu3, line 3975: this change is actually a fix, the wire format was already string for this type (a base58 string)

crates/near-mpc-crypto-types/src/crypto.rs
```diff
@@ -165,6 +165,21 @@ impl schemars::JsonSchema for Secp256k1PublicKey {
...
+    }
+
+    fn schema_name() -> String {
+        "Ed25519PublicKey".to_string()
+    }
+
+    fn json_schema(generator: &mut schemars::SchemaGenerator) -> schemars::schema::Schema {
+        String::json_schema(generator)
+    }
+}
```
gilcu3, line 181: this is the fix in the comment above, and aligns this type with the rest in this file

pbeza (approved):

anodar (approved):

PR #4258, opened 2026-08-26
feat(node): export foreign chain provider metrics on the verify path

### LINKED ISSUE #4150: Export per-provider RTT histogram and error counter from the verify fan-out
### Background

`FanOut::extract` in `crates/foreign-chain-inspector/src/lib.rs` is the right place to measure. It already knows the `ProviderId` and already classifies errors as transient or not. It just doesn't record anything.

Proposed metrics:

```
mpc_foreign_chain_provider_rtt_seconds{chain, provider}          # histogram, roughly 25ms to 10s, exponential buckets
mpc_foreign_chain_provider_errors_total{chain, provider, kind}   # kind: transient, non_transient, timeout
```

Start the timer right before the RPC call inside the spawned task, so we measure the round-trip and not JoinSet queueing.

This covers verification requests only. Startup and CLI probes (`network_fingerprints`, #3764) stay out on purpose, since probe traffic would pollute the verify latency series.

Cardinality is fine: both labels come from the on-chain whitelist, currently around 12 chains with a handful of providers each.

### Acceptance criteria

- Both metrics show up on `/metrics` and cover every verify request, success or failure. Probe paths are untouched.
- The `chain` label is passed into `FanOut` at construction. Metric definitions live in `crates/node/src/metrics.rs`.
- The new metrics are documented in `docs/foreign-chain-transactions.md`.

haiyuechen-nearone:
Closes #4150

The verify fan-out starts a timer for every provider call it makes. The node turns them into three metrics labelled by chain and provider: 
- latency histogram by outcome
- histogram of calls the node abandoned
- error counter by kind

Reviewer notes:

- `provider` labels are pseudonyms `p0`, `p1`, ... assigned by alphabetically sorted provider names, so the public `/metrics` endpoint does not reveal which RPC vendors a node uses.

- Time is recorded regarless of the extraction being successful. If unsuccessful due to provider failures, counter for the error type is incremented as well. 

- A call abandoned due to the future it runs in being dropped lands in `mpc_foreign_chain_provider_dropped_seconds`. Most likely scenario of this getting incremented is if the outer timeout drops the request attempt.

- The inspector crate stays free of prometheus. No `dyn`, trait bounds on structs, new crates or weird drop implementations :)))


haiyuechen-nearone:
@claude review

haiyuechen-nearone:
@claude review

haiyuechen-nearone:
@claude review

haiyuechen-nearone:
@claude review

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -133,11 +134,67 @@ pub trait NetworkFingerprintInspector {
...
+/// extraction: it waits for every inspector, up to the caller's deadline. When no inspector
+/// reached a verdict, the first error is propagated.
+///
+/// With a recorder set through [`FanOut::measuring`], every provider call is reported once it
+/// completes, or as [`ProviderFailure::TimedOut`] if the future is dropped first.
+#[derive(Clone)]
 pub struct FanOut<Inspector> {
     inspectors: NonEmptyVec<(ProviderId, Inspector)>,
+    recorder: Option<Arc<dyn RecordProviderCall>>,
+}
```
gilcu3, line 152: in general we have a preference for traits bound above dynamic dispatch. Would be good to replace this one unless it is strictly needed (and I have never seen that :) )

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -133,11 +134,67 @@ pub trait NetworkFingerprintInspector {
 /// All inner inspectors are queried concurrently. Every [`Verdict`] reached must be identical;
 /// any disagreement returns [`ForeignChainInspectionError::InspectorResponseMismatch`]. Errors
 /// are tolerated as long as any inspector reached a verdict, so a single unavailable or
-/// misbehaving RPC does not take the whole node out of signing. When no inspector reached a
-/// verdict, the first error is propagated.
-#[derive(Clone, derive_more::Constructor)]
+/// misbehaving RPC does not take the whole node out of signing. A hanging one still stalls the
```
anodar: nit: Misbehaving RPC will still take the whole node out of signing, e.g. if it returns TransactionNotFound or different Ok(verdict) from other rpc's because that will result in mismatch error.

crates/foreign-chain-inspector/tests/fanout.rs
```diff
@@ -106,6 +127,88 @@ fn fan_out_of(inspectors: Vec<MockInspector>) -> FanOut<MockInspector> {
...
+    let mut calls = Vec::new();
+    while let Ok(call) = reported.try_recv() {
+        calls.push(call);
+    }
+    calls.sort_by(|left, right| left.0.cmp(&right.0));
+    calls
+}
+
+mod measurement {
+    use super::*;
```
gilcu3, line 199: why using a specific mod for this? Given that this is already inside tests, wouldn't just bare tests work?

docs/guide/node-operator-metrics.md
```diff
@@ -38,6 +38,72 @@ on-chain confirmation of an attestation submission, in
...
+measures each provider it queries. Both series carry a `chain` label, the key the chain is
+configured under (`bitcoin`, `hyper_evm`, ...), and a `provider` label, its configured provider
+name. Both labels are bounded by the node's own config, so the number of series is a small
+multiple of the number of configured providers:
+
+| Metric | Measures | How to interpret |
+| --- | --- | --- |
+| [`mpc_foreign_chain_provider_inspection_seconds`](../../crates/node/src/metrics.rs) | time one provider took to answer one verify request; failed calls are not timed | compare providers of the same chain against each other. One drifting up toward the top bucket is approaching the node's inspection deadline. A provider that fails or hangs is invisible here and loud in `mpc_foreign_chain_provider_errors_total`, so read p95 and p99 next to that counter's rate. |
+| [`mpc_foreign_chain_provider_errors_total`](../../crates/node/src/metrics.rs) | requests a provider itself failed to answer, by `kind` | should be near zero. `non_transient` is the one to act on first: the provider is refusing requests or answering with something unusable, and retrying will fix neither. |
```
gilcu3: If this section is meant to be read by operators I think we should try to simplify it as much as possible. Right now seems quite hard to parse and quite verbose

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -166,11 +223,14 @@ where
...
             join_set.spawn(async move {
-                (
-                    provider,
-                    inspector.extract(tx_id, finality, extractors).await,
-                )
+                // Started inside the task so the clock excludes scheduling delay. A task aborted
+                // before its first poll made no call and reports nothing.
+                let mut call = TimedCall::start(recorder, provider.clone());
+                let result = inspector.extract(tx_id, finality, extractors).await;
+                call.report(result.as_ref().err().and_then(|err| err.provider_failure()));
```
anodar: We only report errors from individual providers but not mismatch in provider responses.

crates/node/src/providers/verify_foreign_tx/provider_call_metrics.rs
```diff
@@ -0,0 +1,146 @@
...
+                        && line.contains(r#"chain="ethereum""#)
+                        && line.contains(r#"provider="a-provider""#)
+                })
+                .count()
+        };
+        assert_eq!(
+            series("mpc_foreign_chain_provider_inspection_seconds_count"),
+            1
+        );
+        assert_eq!(series("mpc_foreign_chain_provider_errors_total"), 3);
```
gilcu3: I had never seen this type of test before. We were usually checking some metrics in e2e tests, and that's all.

anodar (changes requested):
I have 2 concerns with this approach:

- we start leaking providers list that node uses now at `/metrics`
- we don't report error when providers disagree on the result, which is important case, likely bogus RPC provider

gilcu3 (commented):
Overall looks fine, but did not have much time to go deeper.

There are a couple of patterns that should probably be changed though, specially avoiding the dynamic dispatch.

crates/foreign-chain-inspector/src/lib.rs
haiyuechen-nearone: The initial thought for adding these metrics was to see if a provider is failing and/or slow.

I agree that result mismatch is something we should emit metrics for as well, but perhaps not at this level. I think that is something we can capture at  the end of the signing path in node instead.

crates/foreign-chain-inspector/src/lib.rs
haiyuechen-nearone: I did try the trait bound version first and it costs more code for zero behavior change. With `Option<Arc<dyn RecordProviderCall>>`, no op is  implicit.

```rust
// extra type just to say "not measuring"
pub struct NoRecorder;

impl RecordProviderCall for NoRecorder {
    fn record(&self, _: &ProviderId, _: Duration, _: Option<ProviderFailure>) {}
}

pub struct FanOut<Inspector, Recorder = NoRecorder> { ... }
```

and because the production recorder (ProviderCallMetrics) and the test recorder (ReportedCalls) are different types, the parameter spreads into the node:

```rust
pub(crate) struct ForeignChainInspectors<Client, Recorder = NoRecorder> {
    pub bitcoin: Option<FanOut<BitcoinInspector<Client>, Recorder>>,
    // ... twelve more chains
}

pub struct VerifyForeignTxProvider<Client, Recorder = NoRecorder> { ... }
```

In addition, the trait bound will spread, imagine that we can have the following structure:

```rust
struct Holder<I, R> {
    fan_out: FanOut<I, R>,
}

fn store<I, R>(fan_out: FanOut<I, R>) -> Holder<I, R> {
    Holder { fan_out }
}

impl<I, R> Holder<I, R>
where
    R: RecordProviderCall + Clone + Send + Sync + 'static,
{
    fn run(&self, tx: I::TransactionId) {
        self.fan_out.extract(tx, finality, extractors);
    }
}

fn main_flow<I, R>(holder: &Holder<I, R>)
where
    R: RecordProviderCall + Clone + Send + Sync + 'static,
{
    holder.run(tx);
}
```

For metric capture utility I wanted this to be "as hidden as possible" and not infect everywhere FanOut is referenced or used.

haiyuechen-nearone: @netrome I know you have strong opinions on this topic, do you think its justified to use `dyn` here?

crates/foreign-chain-inspector/tests/fanout.rs
haiyuechen-nearone: Not sure why we need separate mods, but this was the pattern in this file so I just followed it. My guess is that this file has very many cases and we would like to group them...?

crates/node/src/providers/verify_foreign_tx/provider_call_metrics.rs
haiyuechen-nearone: `ProviderCallMetrics` owns the registration of these metric series and it is fully observable in process, so I kept this at the unit test level.

I don't think having unit tests for metrics machinery conflicts with checking metrics in e2e tests, and its probably better if we start adding more unit tests for metrics machinery, if they are missing elsewhere.

crates/foreign-chain-inspector/src/lib.rs
haiyuechen-nearone: Reworded the comment

crates/foreign-chain-inspector/src/lib.rs
netrome: Hmm, I don't follow the "no-op" argument. Whether we use dynamic dispatch or generics should be orthogonal to that. We can still let this be optional. What would be the problem with the following?

```suggestion
#[derive(Clone)]
pub struct FanOut<Inspector, Recorder> {
    inspectors: NonEmptyVec<(ProviderId, Inspector)>,
    recorder: Option<Recorder>,
}
```

netrome: > In addition, the trait bound will spread, imagine that we can have the following structure:

Trait bounds should not be put on structs, so they should not spread beyond where they are used.

netrome: Had Claude generate a suggestion: https://github.com/near/mpc/pull/4450 haven't looked deeply but from a quick glance I think it seems reasonable. I don't see any good reason to go with dynamic dispatch here.

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -137,21 +138,88 @@ pub trait NetworkFingerprintInspector {
...
+}
+
+/// Satisfies [`FanOut`]'s recorder bounds for the unmeasured default.
+impl RecordProviderCall for Infallible {
+    fn record(&self, _: &ProviderId, _: Duration, _: Option<ProviderFailure>) {
+        match *self {}
+    }
+}
+
+struct TimedCall<Recorder: RecordProviderCall> {
```
anodar: nit: drop trait constraint from struct definition.

Image

crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -137,21 +138,88 @@ pub trait NetworkFingerprintInspector {
...
+/// different verdict or different values will cause a mismatch and fail the whole extraction.
+/// A hanging one still stalls the extraction: it waits for every inspector, up to the caller's
+/// deadline. When no inspector reached a verdict, the first error is propagated.
+///
+/// With a recorder set through [`FanOut::measuring`], every provider call is reported once it
+/// completes, or as [`ProviderFailure::TimedOut`] if the future is dropped first.
+#[derive(Clone)]
+pub struct FanOut<Inspector, Recorder = Infallible> {
     inspectors: NonEmptyVec<(ProviderId, Inspector)>,
+    recorder: Option<Recorder>,
```
anodar: I think we can get rid of Option here, and maybe don't even define Infallible:

````
pub struct FanOut<Inspector, Recorder = ()>
```

Then 

```
impl  RecordProviderCall for () { 
   record() { do nothing} 
} 
```

anodar (dismissed):
LGTM

netrome (changes requested):
Hate to be a downer but I don't feel the code became much cleaner with scopeguard, and I want to be very conservative with adding new dependencies.

Right now, scopeguard seems to be pure complexity for little gain. The previous TimedCall solution is arguable simpler, and unfortunately we're not getting away from the trait bound on the struct in this case due to the Drop implementation and compiler limitations. This is fine. It's a reminder that Rust isn't a perfect language. One day this might be possible, but not today. Additionally, ScopeGuard also has trait bounds on the struct for the same reason so we're not gaining anything by dragging it in from where I see.

TL;DR - please drop scopeguard (just revert the last commit). The rest of the code looks good to me.

netrome:
> TL;DR - please drop scopeguard (just revert the last commit). The rest of the code looks good to me.

Edit, on slack @haiyuechen-nearone pointed out that we already have scopeguard as a transitive dependency. I'd approve that solution now, but there's another change that uses HistogramTimer now so I'll review that one.

crates/foreign-chain-inspector/tests/fanout.rs
netrome: We should probably break out these modules to separate files. In general I think multiple modules within the same file is an anti-pattern. However, that's the established pattern here so it's not a reason to block this PR.

netrome (approved):
Nice solution with the histogram timer. Thanks! 🙏

anodar (approved):

PR #4266, opened 2026-08-26
feat: make chain-gateways monitoring logic generic for other back-ends

### LINKED ISSUE #4253: move chain-gateways subscriber logic to the mpc-contract-transport crate
### Background

The chain-gateway holds logic that allows a consumer to subscribe to contract state.
That logic is useful in general.
For example, the backup-cli implemented polling logic we already had in the chain-gateway crate (https://github.com/near/mpc/blob/main/crates/backup-cli/src/adapters/contract_state_polling.rs).

We can improve the situation by making the chain-gateway logic a bit more generic and moving it to the near-contract-transport crate.

### User Story

As a developer, I like when I don't need to maintain multiple instances of the same implementation.

### Acceptance Criteria

Any back-end that implements `ViewContract` should also exposed typed one-shot view calls and subscription handles

_No response_

kevindeforth:
Resolves #4253

**Main Changes**
With #3693 we started unifying calls to our mpc-contract across our various back-ends (e2e-test, sandbox, devnet, node).
As of now, the `near-contract-transport` crate allows one to call a contract and execute one-shot state-views.
The node and other components (like the backup-service, which currently uses an RPC endpoint), require not only the ability to view contract state, but also, to monitor contract state for changes.

The chain-gateway offers this capability, but three shortcomings prevent us from moving its traits to the `near-contract-transport` crate, where every back-end could implement them:

1. It hard-codes the error type to `ChainGatewayError`. Other back-ends will have their own error-types. This PR makes the error type a generic that can be set by different back-ends.
2. By default a contract subscription and view call resut is deserialized to json. But contracts may return a variety of serialization formats and users may want to provide their own deserializers. This PR allows a user to set their deserializer.
3. The chain-gateway traits can be simplified. We currently have four traits for viewing contract state from chain: `ViewContract` (the backend query), `ViewRaw` (one-shot raw view), `ViewMethod` (one-shot call viewing and deserializing state) and `SubscribeToContractMethod` (monitoring and deserializing state). A user needs to know several different trait-names and what differentiates them:
```rust
let raw = viewer.view_contract(&contract_id, view_args).await;
let res = viewer.view_raw(&contract_id, view_args).await;
let val = viewer.view_method::<String>(contract_id, view_args).await;
let mut sub = viewer.subscribe_to_contract_method::<String>(contract_id, view_args).await;
```

This bears an unnecessary cognitive load. Any struct that provides the `ViewContract` interface should be able to provide state viewing and monitoring for free. In this PR, we introduce the following API:

```rust
let raw = viewer.view_contract(&contract_id, view_args).await; // SerializedObservation
let res = viewer.view_json::<String>(contract_id, view_args).await;
let mut sub = viewer.view_json::<String>(contract_id, view_args).subscribe().await;
```
This PR deletes the `SubscribeToContractMethod`, `ViewMethod`, and `ViewRaw` traits, as they are no longer needed. Instead, we introduce a `ViewCall` struct that can either be `.await`-ed (to yield the one-shot view response), or `.subscribe()`-ed to (to yield a continuous view of the contract). The `ViewCall` allows the caller to define what deserializer to use for deserializing the chain response.

**Additional changes in this PR:**

- A small bug-fix. The `ViewContract` implementation of the `ChainGateway` struct didn't await full-sync before calling the near client actor. This PR fixes that.
- This PR introduces a `HasPollInterval` trait, which allows the consumer to customize how often the monitoring task should query the back-end for changes. This was introduced because the backup-cli relies on a user-provided poll-interval, likely to avoid overwhelming an RPC endpoint with requests.
- Instead of hard-coding the poll-interval at 200 millisecond, the chain-gateway now reads the expected block time from the near node config and sets its poll-interval proportional to that (with a floor). This is helpful when running localnets, where we often have a much shorter block-time than in testnet or mainnet.

**Next steps:**
We will want to move a lot of stuff from the `chain-gateway` crate to the `near-contract-transport` crate. This is purposely left for a follow-up PR, as otherwise, reviewing the changes would become quite cumbersome (c.f. https://github.com/near/mpc/issues/4275)


kevindeforth:
@claude review

kevindeforth:
@claude review

crates/chain-gateway/src/near_internals_wrapper/view_client.rs
```diff
@@ -29,15 +29,13 @@ impl NearViewClientActorHandle {
     }
 }
 
-impl ViewContract for NearViewClientActorHandle {
-    type Error = NearViewClientError;
-    type ObservedAt = BlockHeight;
```
kevindeforth, line 34: This was kind of a bug. Implementing `ViewContract` directly on `NearViewClientActor` makes it seem like it's okay to call `view_contract` on the client actor.
But it isn't. It's important we first ensure that the node is fully synced. Otherwise, we risk getting a valid, but historic view, which may lead to all sorts of issues.

In this PR, we preserve the `ViewContract` implementation over the `ChainGateway` struct, but demote the viewer on the `NearViewClientActorHandle` to a chain-gateway internal method.

crates/chain-gateway/src/state_viewer/monitoring.rs
```diff
@@ -1,66 +1,85 @@
...
     _task_handle: JoinHandle<()>,
-    cancel_token: CancellationToken,
-    pub last_observed: tokio::sync::watch::Receiver<Result<ObservedState, ChainGatewayError>>,
+    _drop_guard: DropGuard,
+    deserializer: Deserializer<T>,
 }
 
-impl Drop for MonitoringTask {
-    fn drop(&mut self) {
-        self.cancel_token.cancel();
```
kevindeforth, line 19: We got rid of this cancel-on-drop implementation by replacing the `CancellationToken` with its `DropGuard`.

crates/chain-gateway/src/state_viewer/subscription.rs
```diff
@@ -1,226 +1,140 @@
...
-use near_contract_transport::ViewArgs;
-use serde::de::DeserializeOwned;
-
-/// Holds a Monitoring task and the latest cached value.
-/// This is useful such that we don't unnecessarily deserialize the same state multiple times.
-pub(crate) struct ContractMethodSubscription<Res> {
-    inner: MonitoringTask,
-    cached: Result<ObservedState<Res>, ChainGatewayError>,
-}
-
```
kevindeforth, line 15: This struct was kind of pointless. When I initially introduced this struct, I thought it was needed for testing, but without it, we test just fine, so, simplification wins.

This PR moves the `cached` field into `MontiringTask`.

crates/chain-gateway/src/state_viewer/subscription.rs
```diff
@@ -1,226 +1,140 @@
...
-        let observed = self.inner.last_observed.borrow_and_update().clone();
-        self.cached = observed.and_then(|value| deserialize_observed(value));
-    }
-}
+use super::monitoring::MonitoringTask;
+use near_contract_transport::WatchContractState;
+use near_contract_transport::{ObservedState, TransportError};
 
-impl<Res> WatchContractState<Res> for ContractMethodSubscription<Res>
+impl<T, ViewError> WatchContractState<T, ViewError> for MonitoringTask<T, ViewError>
```
kevindeforth, line 5: Note that this is not the correct place to implement this trait, but I left it here to reduce the diff-size for reviewers. We will move a lot of this to the `near-contract-transport` crate in a follow-up PR (https://github.com/near/mpc/issues/4275)

crates/chain-gateway/src/state_viewer/subscription.rs
```diff
@@ -1,226 +1,140 @@
...
-        self.inner
-            .last_observed
+    /// The initial observation counts as already seen: this future completes only if a new value
+    /// is seen.
+    async fn changed(&mut self) -> Result<(), TransportError<ViewError>> {
+        self.last_observed
             .changed()
             .await
-            .map_err(|_| ChainGatewayError::MonitoringClosed)?;
+            .map_err(|_| TransportError::MonitoringClosed)?;
```
kevindeforth, line 16: And for that, we are now no longer returning a `ChainGatewayError`, but a `TransportError`, because the error here is back-end agnostic.

crates/chain-gateway/src/state_viewer/subscription.rs
```diff
@@ -1,226 +1,140 @@
...
-            }))
-            .build();
-
-        let mut sub = ContractMethodSubscription::<String>::new(
-            viewer,
-            "test.testnet".parse().unwrap(),
-            ViewArgs::no_args("get_value"),
-        )
-        .await;
+    async fn subscription__should_deserialize_the_initial_value() {
```
kevindeforth, line 54: Note: we still test for the same things, but I renamed them to follow the format outlined in our engineering-standards.

crates/chain-gateway/src/state_viewer/traits.rs
```diff
@@ -1,482 +1,24 @@
...
-        contract_id: AccountId,
-        view_args: ViewArgs,
-    ) -> impl Future<Output = Result<ObservedState<Res>, ChainGatewayError>> + Send
-    where
-        Res: DeserializeOwned + Send + Clone;
-}
-
-/// All other viewer traits are derived from this one. Subscriptions track the
-/// height a value was observed at, so the backend must report one.
-pub(crate) trait ViewRaw: IsSyncing + ViewContract<ObservedAt = BlockHeight> {
```
kevindeforth, line 107: These two traits were doing more harm than good. `ViewRaw` was chain-gateway internal and won't be missed. `ViewContract` was too specific and is replaced with the more general `ViewExt` trait in this same file (which will soon be moved to the `near-contract-transport` crate).

crates/chain-gateway/src/state_viewer/traits.rs
```diff
@@ -1,482 +1,24 @@
...
-    ) -> impl Future<Output = Result<ObservedState, ChainGatewayError>> + Send;
-}
-
-/// A watch-like stream of contract state changes.
-///
-/// Call [`latest()`](WatchContractState::latest) to get the most recent value,
-/// and [`changed()`](WatchContractState::changed) to wait for the next update.
-/// Only actual value changes (different bytes) trigger a notification (block
-/// height increases alone do not).
-pub trait WatchContractState<Res> {
```
kevindeforth, line 122: this trait is moved to `crates/near-contract-transport/src/views/traits.rs`

docs/chain-gateway-design.md
kevindeforth, line 1: Claude pointed out that this file was stale. Doing the changes here is a bit annoying, I think this doc has served its purpose and I wouldn't mind removing it in the near future.

docs/tee-context-design.md
kevindeforth, line 1: same here, we have an implementation, we can maybe remove the implementation details and keep the high-level motivation.

crates/near-contract-transport/src/traits.rs
```diff
@@ -30,23 +28,3 @@ impl<T: CallContract> CallContract for &T {
...
-    ///
-    /// [`BlockHeight`]: crate::BlockHeight
-    type ObservedAt;
-
-    fn view_contract(
-        &self,
-        contract_id: &AccountId,
-        view_args: ViewArgs,
-    ) -> impl Future<Output = Result<ObservedState<Vec<u8>, Self::ObservedAt>, Self::Error>> + Send;
-}
```
kevindeforth, line 52: moved to `crates/near-contract-transport/src/views/traits.rs`

crates/near-contract-transport/src/types.rs
```diff
@@ -32,26 +32,6 @@ impl FunctionCallArgs {
...
-            method_name: method_name.into(),
-            args,
-        }
-    }
-
-    /// A view call taking no arguments (an empty JSON object).
-    pub fn no_args(method_name: impl Into<String>) -> Self {
-        Self::new(method_name, b"{}".to_vec())
-    }
-}
```
kevindeforth, line 53: moved to `crates/near-contract-transport/src/views/args.rs`

crates/near-contract-transport/src/types.rs
```diff
@@ -69,63 +49,3 @@ impl BlockHeight {
...
-
-    #[derive(Debug, Deserialize, PartialEq, Eq)]
-    struct Num {
-        a: u32,
-    }
-
-    #[test]
-    fn no_args__should_encode_an_empty_json_object() {
-        assert_eq!(ViewArgs::no_args("m").args, b"{}");
-    }
```
kevindeforth, line 107: moved to `crates/near-contract-transport/src/views/args.rs`

crates/near-contract-transport/src/types.rs
```diff
@@ -69,63 +49,3 @@ impl BlockHeight {
...
-impl<H> ObservedState<Vec<u8>, H> {
-    pub fn deserialize<Res: DeserializeOwned>(
-        self,
-    ) -> Result<ObservedState<Res, H>, serde_json::Error> {
-        Ok(ObservedState {
-            observed_at: self.observed_at,
-            value: serde_json::from_slice(&self.value)?,
-        })
-    }
-}
```
kevindeforth, line 91: deleted in favor of the more generic `crates/near-contract-transport/src/views/deserialize.rs`

crates/near-contract-transport/src/types.rs
```diff
@@ -69,63 +49,3 @@ impl BlockHeight {
...
 }
-
-/// A value read from a contract together with the height it was observed at.
-/// `H` is the backend's height witness: [`BlockHeight`] where the backend
-/// reports one, `()` where it cannot.
-#[derive(Clone, Debug, Eq, PartialEq)]
-pub struct ObservedState<T = Vec<u8>, H = BlockHeight> {
-    pub observed_at: H,
-    pub value: T,
-}
```
kevindeforth, line 80: moved to `crates/near-contract-transport/src/views/observation.rs`

crates/chain-gateway/src/state_viewer/view_call.rs
kevindeforth, line 1: This file contains the only innovation in this PR. Everything else is kind of just moving stuff around in a minimally intrusive way to reduce diff size.

Note: I would have preferred to have this file already in the near-contract-transport crate, but that would have required to already move the `MonitoringTask`, i.e. `state_viewer/subscription.rs` and `state_viewer/monitoring.rs` files too, which would have made the diff way bigger.

crates/chain-gateway/src/chain_gateway.rs
```diff
@@ -154,6 +177,12 @@ fn run_node(
     mut shutdown_receiver: tokio::sync::oneshot::Receiver<()>,
     streamer_setup: Option<StreamerSetup>,
 ) {
+    let expected_block_time: std::time::Duration = near_config
+        .config
+        .consensus
+        .min_block_production_delay
+        .try_into()
+        .expect("block time must be non-negative");
```
kevindeforth, line 187: Okay, this is also an improvement over the status quo. Instead of hard-coding the poll interval to 200 milliseconds, we read the expected block time and compute a good poll interval from the config value.

crates/chain-gateway/src/mock.rs
kevindeforth, line 1: The `MockViewContract` struct in this file will be moved to the `near-contract-transport` crate with https://github.com/near/mpc/issues/4275. That's the reason for breaking it out of the `MockChainState`.

crates/chain-gateway/src/primitives.rs
```diff
@@ -39,6 +39,14 @@ pub(crate) trait IsSyncing: Send + Sync + 'static {
     }
 }
 
+pub(crate) async fn view_when_synced<T, F: Future<Output = T>>(
+    sync: &impl IsSyncing,
+    view: impl FnOnce() -> F,
+) -> T {
+    sync.wait_for_full_sync().await;
+    view().await
+}
```
kevindeforth, line 48: Claude insisted on this, such that we can test the sync-view logic. I agree with claude.

kevindeforth (commented):
I left some comments that I thought might be helpful to reviewers. Review recommendation:

1. start with the integration tests in `chain-gateway/tests` - they should give a good idea of where this is headed on a high-level and show that this doesn't break anything.
2. The only new stuff this PR adds is the `ViewCall` struct in `crates/chain-gateway/src/state_viewer/view_call.rs`.

kevindeforth:
@claude review

kevindeforth:
Finally, Claude didn't find anything of substance, we are good to review :muscle:

crates/chain-gateway/tests/event_subscriber_integration.rs
```diff
@@ -146,7 +146,8 @@ async fn test_event_subscriber_executor_function_call_success_failure_calls_are_
         .unwrap();
 
     let mut watch_value = observer_gw
-        .subscribe_to_contract_method::<String>(contract_id, ViewArgs::no_args(VIEW_VALUE))
+        .view_json::<String>(contract_id, ViewArgs::no_args(VIEW_VALUE))
+        .subscribe()
         .await;
```
gilcu3, line 151: nit: the old interface seems more intuitive to me. If I read `view_json` I am not really sure what to expect, if it is a json or something else, like a subscriber. Not sure if we can do better though

docs/chain-gateway-design.md
gilcu3: I think we should just move these docs inside a design folder, to keep for historical reasons only, and make sure Claude does not waste tokens on them

gilcu3: Here we should just keep it without changes

crates/near-contract-transport/src/lib.rs
```diff
@@ -1,12 +1,21 @@
...
 #[cfg(feature = "traits")]
-pub use traits::{CallContract, ViewContract};
-pub use types::{BlockHeight, FunctionCallArgs, NearGas, NearToken, ObservedState, ViewArgs};
+pub use traits::CallContract;
+#[cfg(feature = "traits")]
+pub use views::{
+    DeserializationError, Deserializer, HasPollInterval, ObservedState, PollInterval,
+    SerializedObservation, TransportError, ViewArgs, ViewContract, WatchContractState, borsh_de,
+    json_de,
+};
```
gilcu3, line 19: feels a bit off to have `borsh_de` and `json_de` here. Could we avoid this somehow?

crates/tee-context/src/lib.rs
```diff
@@ -255,7 +255,6 @@ mod tests {
 
     fn mock_chain() -> MockChainState {
         let mock = MockChainStateBuilder::new()
-            .with_syncing_status(Ok(false))
             .with_view_response(Ok(ObservedState {
```
gilcu3, line 258: did this `with_syncing_status` disappear because the MockChain no longer needs to have that option, or it was just that this became the default?

crates/chain-gateway/src/near_internals_wrapper/view_client.rs
SimonRastikian: This seems to me like a fix independent of this PR purpose.

crates/chain-gateway/src/chain_gateway.rs
```diff
@@ -27,6 +31,25 @@ pub struct ChainGateway {
...
+const MIN_CHAIN_GATEWAY_POLL_INTERVAL: Duration = Duration::from_millis(50);
+const MAX_POLLS_PER_BLOCK: u32 = 2;
+
+fn poll_interval(expected_block_time: Duration) -> PollInterval {
+    PollInterval::new(max(
+        expected_block_time / MAX_POLLS_PER_BLOCK,
+        MIN_CHAIN_GATEWAY_POLL_INTERVAL,
+    ))
+    .expect("at least the 50ms floor")
+}
```
gilcu3: feels a bit strange to need this in prod code

crates/chain-gateway/src/state_viewer/monitoring.rs
```diff
@@ -1,66 +1,85 @@
...
+use crate::state_viewer::view_call::{ViewCall, deserialize_res};
 
-pub(crate) struct MonitoringTask {
+pub struct MonitoringTask<T, ViewError> {
+    pub(crate) cached: Result<ObservedState<T>, TransportError<ViewError>>,
+    pub last_observed: tokio::sync::watch::Receiver<Result<SerializedObservation, ViewError>>,
     _task_handle: JoinHandle<()>,
-    cancel_token: CancellationToken,
-    pub last_observed: tokio::sync::watch::Receiver<Result<ObservedState, ChainGatewayError>>,
+    _drop_guard: DropGuard,
```
gilcu3, line 16: why do we change from `cancel_token` to `_drop_guard`? Seems a bit unrelated

crates/chain-gateway/src/state_viewer/traits.rs
gilcu3, line 1: would have been good to make this modification as a pure refactor in a first PR, but not sure if that was possible

gilcu3 (dismissed):
This LGTM, although I cannot say that with high confidence as there were quite a few changes here. Left some questions

crates/chain-gateway/src/state_viewer/monitoring.rs
```diff
@@ -1,66 +1,85 @@
...
+    deserializer: Deserializer<T>,
```
SimonRastikian: Could you please comment on this change?

crates/chain-gateway/src/state_viewer/traits.rs
```diff
@@ -1,482 +1,24 @@
...
-    /// # Type Parameter
-    ///
-    /// `T` is the deserialized return type of the contract method.
-    fn subscribe_to_contract_method<T>(
-        &self,
-        contract: AccountId,
-        view_args: ViewArgs,
-    ) -> impl Future<Output = impl WatchContractState<T> + Send> + Send
+pub trait ViewExt: ViewContract + Clone + Sized {
+    fn view_json<T>(&self, contract_id: AccountId, args: ViewArgs) -> ViewCall<Self, T>
```
netrome, line 9: So far I think the code looks really good. At this point I wonder if it would make sense to use marker types instead of separate methods to distinguish between the serializers.

I.e. instead of `viewer.view_json::<String>()` and `viewer.view_borsh::<String>()` we could do `viewer.view::<Json, String>()` and `viewer.view::<Borsh, String>`?

Not really sure if it really makes the code that much better though. Long term it would be nice to have some descriptor trait to define a method name and it's serializers that could be provided to this call but I feel like that's beyond the purpose of this PR.

crates/chain-gateway/src/state_viewer/view_call.rs
```diff
@@ -0,0 +1,199 @@
...
+impl<Viewer, T> ViewCall<Viewer, T>
+where
+    T: serde::de::DeserializeOwned,
+{
+    pub fn json(viewer: Viewer, contract_id: AccountId, args: ViewArgs) -> Self {
+        Self {
+            viewer,
+            contract_id,
+            args,
+            deserializer: json_de::<T>,
```
netrome, line 28: Hmm, while I'm happy to pass around functions - I wonder if the code would be simpler here as well if we used a marker type to denote the serialization format or not.

This may require some reworking/redesign of the serialization bits though.

crates/chain-gateway/src/chain_gateway.rs
```diff
@@ -27,6 +31,25 @@ pub struct ChainGateway {
...
+    expected_block_time: Duration,
+}
+
+/// This is a floor in case we run a localnet with a super short block time.
+const MIN_CHAIN_GATEWAY_POLL_INTERVAL: Duration = Duration::from_millis(50);
+const MAX_POLLS_PER_BLOCK: u32 = 2;
+
+fn poll_interval(expected_block_time: Duration) -> PollInterval {
+    PollInterval::new(max(
+        expected_block_time / MAX_POLLS_PER_BLOCK,
```
netrome:
```suggestion
        expected_block_time.checked_div(MAX_POLLS_PER_BLOCK).expect("Should never fail as MAX_POLLS_PER_BLOCK is not zero"),
```

crates/chain-gateway/src/state_viewer/view_call.rs
```diff
@@ -0,0 +1,199 @@
...
+    ViewContract, WatchContractState, borsh_de, json_de,
+};
+
+use crate::state_viewer::monitoring::MonitoringTask;
+
+pub struct ViewCall<Viewer, T> {
+    pub viewer: Viewer,
+    pub contract_id: AccountId,
+    pub args: ViewArgs,
+    pub deserializer: Deserializer<T>,
```
netrome, line 16: The `ViewArgs` here is a low-level byte representation of the args, whereas we use a higher level deserializer for the return type. This feels a bit like an odd discrepancy. Iiuc, as a caller I'd have to manually serialize the args, but I get the result serialized already.

How about we introduce a generic parameter for the input type and explicitly serialize it in this module as well?

netrome (dismissed):
Nice stuff! Some thoughts, primarily about the serialization handling which is the piece that stands out the most to me. But this is mostly open questions, I'm not sure what the "perfect design" would be here and how far/close we are to it. Would be fun to explore further but I can't spend more time on this right now unfortunately.

crates/chain-gateway/src/chain_gateway.rs
```diff
@@ -27,6 +31,25 @@ pub struct ChainGateway {
...
+        MIN_CHAIN_GATEWAY_POLL_INTERVAL,
+    ))
+    .expect("at least the 50ms floor")
+}
+
+impl HasPollInterval for ChainGateway {
+    fn poll_interval(&self) -> PollInterval {
+        poll_interval(self.expected_block_time)
+    }
```
netrome: Oh see I forgot to write this in my review - these functions feel like they should be moved further down. It feels less significant or relevant than the `ViewContract` implementation, so I would prefer to see that implementation first.

crates/near-contract-transport/src/lib.rs
kevindeforth: This will become private with https://github.com/near/mpc/issues/4275

crates/tee-context/src/lib.rs
kevindeforth: MockChain no longer requires that option.
This PR removes the `ViewRaw: IsSyncing + ViewContract<ObservedAt = BlockHeight>` trait, which removed the last justification for keeping it on `MockChain`. Instead, we replace the trati it with a unit-tested function:

https://github.com/near/mpc/blob/8dd131d2a3557c4c4ec8f7f67617b37dfa27523f/crates/chain-gateway/src/primitives.rs#L42-L48

crates/chain-gateway/src/chain_gateway.rs
kevindeforth: Do you mean the poll-interval? Or this method? The method is there just so we can unit-test the arithmetic. I had it inlined first, but claude complained and I can see how it's better to have each component tested.

The poll-interval is needed, because we have it in the backp-cli.

crates/chain-gateway/src/state_viewer/monitoring.rs
kevindeforth: It's not needed, this is a pure simplification.

crates/chain-gateway/src/chain_gateway.rs
gilcu3: I meant the 50ms max computation. I guess we are trying to protect callers from breaking it, but we could just have a comment as well, or an error returned in that case?

crates/chain-gateway/tests/event_subscriber_integration.rs
kevindeforth: > the old interface seems more intuitive to me. If I read view_json I am not really sure what to expect, if it is a json or something else, like a subscriber. Not sure if we can do better though

`view_json` doesn't return json, it returns a `ViewCall` with a json deserializer for the bytes.
One can `.await` to get the typed return snapshot, or `.subscribe()` to get a `WatchContractState` implementation.

We can probably do better on the interface, e.g. by making `json` deserializer defualt, but I don't think it will matter that much in practice: The contract-interface will hide the deserializer choice from the user, and at the one place per method where it is writteng, being explicit about the wire format seems nice.

In the contract-interface crate:
```rust
impl<C: ViewContract> MpcContractHandle<C> {
    pub fn state(&self) -> ViewCall<C, ProtocolContractState> {
        self.caller.view_json(self.contract_id, ViewArgs::no_args(STATE))
    }
}
```
In e.g. the node or backup-cli:
```rust
let protocol_state_snapshot : ObservedState<ProtocolContractState> = mpc_contract_handle.state().await?;
let mut protocol_state_sub = mpc_contract_handle.state().subscribe().await;
```

crates/chain-gateway/src/state_viewer/traits.rs
kevindeforth: > I.e. instead of `viewer.view_json::<String>()` and `viewer.view_borsh::<String>()` we could do `viewer.view::<Json, String>()` and `viewer.view::<Borsh, String>`?

Oh that does look neat.!
I think in terms of lines of code and complexity it would be pretty similar to what we currently have. In terms of ergonomics I think it's close, although I do slightly prefer your `viewer.view<Borsh, String>` variant.

crates/chain-gateway/src/state_viewer/view_call.rs
kevindeforth: > This feels a bit like an odd discrepancy. Iiuc, as a caller I'd have to manually serialize the args, but I get the result serialized already.

Yes, that is a discrepancy, but it is deliberate. And it's good that you mention it, because I also considered this but didn't find a conclusive solution.

The reasons why I decided against typed `ViewArgs` _in this PR_:
1. `ViewArgs` is an existing struct and changing it doesn't contribute to this PR's main goals. If we refactor it here or later, it'll be similar in effort.
2. It's not entirely obvious that typed args are better. The asymmetry reflects how the two sides are used: the same args can be used in multiple calls. Passing the serialized args allows the caller to optimize that. On the other and, one response may be cloned multiple times, but every consumer will eventually have to deserialize it. So, deserializing before returning saves boilerplate and optimizes for the consumer.
3. Serializing the args inside ViewCall makes the constructor fallible, which isn't very nice ergonomics. I'd rather have an explicit serialization step and allow the caller to decide if they want to panic or pass the error. So, while serializing might save some boilerplate, it might add error-handling boilerplate.

Now, I don't think 2. and 3. are strong reasons. I think each one can be addressed. But in the context of 1. I didn't want to be defending introducing typed args in this PR and increasing the diff if it wasn't obvious.

Happy to add the typed args in a follow-up after https://github.com/near/mpc/issues/4275 and see how they look.

crates/chain-gateway/src/chain_gateway.rs
kevindeforth: We need a poll interval strictly greater than 0, because tokio::Interval panics on zero. `PollInterval` exists to check this a construction. The expect here can't actually panic: the max with the nonzero 50ms floor guarantees the constructor will succeed.

The 50ms is kind of arbitrary, but I wouldn't want to query the nearcore node more often than that (regardless of block time).

Now, should we return an error in case the `expected_block_time / MAX_POLLS_PER_BLOCK` is lower than 50ms? I don't think so, for two reasons: 1) it's not an actionable error: the user can't do much about it. The block-time is a config value that is given by the network. 2) it's not something that breaks this code: we  can override get less than two queries per block. But that's okay.

crates/chain-gateway/src/chain_gateway.rs
kevindeforth: addressed here: https://github.com/near/mpc/pull/4266/commits/8a67ca7c7c01b93c331fe6f3a8e7fd523e963684

crates/chain-gateway/src/chain_gateway.rs
kevindeforth: addressed here: https://github.com/near/mpc/pull/4266/commits/8a67ca7c7c01b93c331fe6f3a8e7fd523e963684

docs/chain-gateway-design.md
kevindeforth: Hmm, I was actually thinking to move this doc to chain-gateway as a readme and out of the docs folder entirely.

gilcu3 (approved):

netrome (approved):
Thanks for the update 🙏

crates/chain-gateway/src/state_viewer/traits.rs
kevindeforth: edit: displayed in https://github.com/near/mpc/pull/4317

PR #4279, opened 2026-08-28
chore(standards): Don't require `must_` prefix on all panicking test helpers

### LINKED ISSUE #4277: chore(standards): Don't imply a panicking test helper should be prefixed with `must_`
### Background

Currently, the "`must_` prefix for panicking test helpers" standard states the following:

> In test code, plumbing helpers (loading WASM artifacts, resolving binary
paths, extracting setup data from a known-good state) should panic on
failure instead of returning `Result`: the failure means the test wasn't
built or wired correctly, not that the system under test misbehaved, so
there's no useful error path. Such helpers must be prefixed with `must_`
(e.g. `must_load_contract_wasm`, `must_get_bls_public_key`) so callers
can see at the call site that the function will panic on failure.
> 
> Helpers whose failure could be a meaningful test outcome (network calls,
state observations) should still return `Result`.
 
This formulation misses that there's a lot of test helpers that don't do external plumbing (e.g. rely on something in the external environment to be properly configured/set up), that still may panic.

This has caused some confusion and led to normal test helpers getting the `must_` prefix where it causes more confusion than help.

We should rephrase this standard since it's perfectly okay for a test helper to panic without having a `must_` prefix.

### User Story

As an engineer I want readable code and clear engineering conventions

### Acceptance Criteria

We've rephrased the standard and renamed mislabeled test helpers.

_No response_

netrome:
Closes #4277 

This PR does two things:
1. Updates the formulation in our engineering standards to not imply that any panicking test helper should have a `must_` prefix.
2. Rename some instances where this prefix causes more confusion than help.

### DIFF
--- a/crates/backup-cli/src/service.rs
+++ b/crates/backup-cli/src/service.rs
@@ -418,7 +418,7 @@ mod tests {
             .expect("the service should read local storage on startup")
     }
 
-    async fn must_back_up_if_needed(
+    async fn back_up_if_needed(
         service: &mut TestService,
         state: &ProtocolContractState,
     ) -> BackupOutcome {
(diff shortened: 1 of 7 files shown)


gilcu3 (approved):
Thank you!

anodar (approved):

Issue #3746, opened 2026-07-07
merge `periodic_attestation_submission` and `monitor_attestation_removal`


kevindeforth:
### Background

Looking at the current code:
* we spawn both tasks at the same location (inside [create_root_future](https://github.com/near/mpc/blob/187fffa072fd5406531f5352e6833ee0483f07e9/crates/node/src/run.rs#L321))}
* both functions have the  same failure behavior (coninue on `CollateralFetch` error, return error on any other error.
* `periodic_attestation_submission`: generates and submits an attestation every `ATTESTATION_SUBMISSION_INTERVAL`
* `monitor_attestation_removal` generates and submits if the nodes attestation was removed from the contract state

We can probably keep one method and have a tokio select on the interval and the attestation removal trigger

### User Story

As a dev, I like to maintain smaller code surfaces

### Acceptance Criteria

Justify why we need both of these functions or make a single task for handling both attestation submission triggers.

### Resources & Additional Notes

https://nearone.slack.com/archives/C0912BTG51T/p1783400536850869?thread_ts=1783263509.430509&cid=C0912BTG51T

https://nearone.slack.com/archives/C07UW93JVQ8/p1775574282130329?thread_ts=1775400133.152509&cid=C07UW93JVQ8

gilcu3:
After some [discussion](https://nearone.slack.com/archives/C0912BTG51T/p1787827173498209?thread_ts=1787757460.846559&cid=C0912BTG51T) we figured that `monitor_attestation_removal` should not be needed anymore, at least looking at the original reasons to include it in #1198. Therefore, the objective of this issue becomes its removal

PR #4281, opened 2026-08-28
fix(node): retry attestation submission only until the next resubmission tick

### LINKED ISSUE #3747: Reduce attestation submission timeout
### Background

The node attempts to submit an attestation transaction for 12 hours

https://github.com/near/mpc/blob/187fffa072fd5406531f5352e6833ee0483f07e9/crates/node/src/tee/remote_attestation.rs#L88

That is not ideal. If said attestation reaches the contract after 12 hours, it will already be quite old.
We should reduce the inner timeout (for sending the transaction to chain), and add an outer retry mechanism (attempting to re-generate an attestation).
Alternatively, we can also add a cancellation token to the `validate_and_submit_remote_attestation` method that gets cancelled as soon as the interval is reached in `periodic_attestation_submission`.

This task can maybe resolved together with #3746 

### User Story

As a node operator, I like when my attestation reaches the chain in a timely manner.

### Acceptance Criteria

Reduce the timeout for sending an attestation to the contract to something more reasonable (maybe 10 minutes). 
Add a retry loop for cycle attestation generation --> attestation submission.

https://nearone.slack.com/archives/C0912BTG51T/p1783403354008239?thread_ts=1783263509.430509&cid=C0912BTG51T

### LINKED ISSUE #4280: Simplify attestation submission retry logic into a single hourly loop
### Background

#4201 kept the attestation tasks alive after the submission retry timeout, but left the retry hierarchy in place:

* the periodic task submits a freshly generated attestation every hour ([`ATTESTATION_RESUBMISSION_INTERVAL`](https://github.com/near/mpc/blob/717385e1d7c822342e9b4b8e6d2aa6bf9e9eae4d/crates/node/src/tee/remote_attestation.rs#L36))
* each submission retries with exponential backoff for up to 12 hours ([`MAX_RETRY_DURATION`](https://github.com/near/mpc/blob/717385e1d7c822342e9b4b8e6d2aa6bf9e9eae4d/crates/node/src/tee/remote_attestation.rs#L33))
* `monitor_attestation_removal` retries a failed resubmission after a fixed 10 minutes ([`RESUBMISSION_RETRY_DELAY`](https://github.com/near/mpc/blob/717385e1d7c822342e9b4b8e6d2aa6bf9e9eae4d/crates/node/src/tee/remote_attestation.rs#L35))

Since a fresh attestation is submitted every hour anyway, retrying a single aging attestation for up to 12 hours covers no additional failure case, and a late success can store an attestation that is almost 12 hours old. The model agreed on in the #4201 review discussion is simpler:

> Submit every hour with a freshly generated attestation:
> * if generation fails, just wait the hour, there is not much to do
> * if submission fails (lets assume for simplicity we just don't know why), retry with exponential back off until the next hour tick

### User Story

As a dev, I want the attestation submission retry logic to have a single obvious cadence instead of a nested retry hierarchy.

### Acceptance Criteria

- Every hour the node generates a fresh attestation and submits it
- Submission failures are retried with exponential backoff only until the next hourly tick, not for 12 hours
- Generation failures just wait for the next hourly tick
- `MAX_RETRY_DURATION` and `RESUBMISSION_RETRY_DELAY` are removed

Related: #3747 (resolved by this simplification) and #3746 (removing `monitor_attestation_removal` entirely, which composes with this change).

https://github.com/near/mpc/pull/4201

pbeza:
Closes #4280
Closes #3747


pbeza:
On hold until #3746 is addressed, as requested in https://github.com/near/mpc/issues/4280#issuecomment-5450598957. Removing `monitor_attestation_removal` there first should make this change simpler; I'll rebase once it lands.

netrome:
https://github.com/near/mpc/issues/3746 is merged, so this should be unblocked. There are conflicts to resolve though

crates/node/src/tee/remote_attestation.rs
```diff
@@ -406,17 +448,60 @@ mod tests {
...
+        // Given
+        let setup = test_setup();
+        setup.sender().set_failing(true);
+        let deadline = Instant::now() + ATTESTATION_RESUBMISSION_INTERVAL;
+
+        // When
+        setup.submitter.generate_and_submit(deadline).await;
+
+        // Then
+        assert_eq!(Instant::now(), deadline);
```
haiyuechen-nearone:
```suggestion
        assert_eq!(Instant::now(), deadline);
        assert!(setup.sender().count() > 1)
```

nit: let's add a check for retries actually ran

crates/node/src/tee/remote_attestation.rs
```diff
@@ -25,19 +25,34 @@ use tokio_util::time::FutureExt;
...
+use tokio::{sync::watch, time::Instant};
 
 const MIN_BACKOFF_DURATION: Duration = Duration::from_millis(100);
 const MAX_BACKOFF_DURATION: Duration = Duration::from_secs(60);
-const MAX_RETRY_DURATION: Duration = Duration::from_secs(60 * 60 * 12); // 12 hours.
 const BACKOFF_FACTOR: f32 = 1.5;
 const ATTESTATION_RESUBMISSION_INTERVAL: Duration = Duration::from_secs(60 * 60); // 1 hour.
 
+pub(crate) trait GenerateAttestation: Send + Sync {
+    fn generate_attestation(
```
haiyuechen-nearone, line 39: nit: Can attestation generation hang, and should we send down a deadline for timeout here as well? I traced the code for getting an attestation from dstack, and I see that its a rpc call.

Not part of the issue being solved so I am perfectly happy with if we address this in a follow up.

crates/node/src/tee/remote_attestation.rs
```diff
@@ -232,31 +248,53 @@ pub async fn run_periodic_attestation_submission<T: TransactionSender + Clone>(
...
-    mut interval_ticker: I,
+async fn periodic_attestation_submission<T: TransactionSender + Clone, A: GenerateAttestation>(
+    submitter: AttestationSubmitter<T, A>,
+    mut interval_ticker: impl Tick,
 ) {
     loop {
-        interval_ticker.tick().await;
-        submitter.generate_and_submit().await;
+        let deadline = interval_ticker.tick().await;
+        submitter.generate_and_submit(deadline).await;
```
haiyuechen-nearone: I agree with Claude's blocking finding. 

I think a hard timeout around generate_and_submit() here is valid. If for any reason, the code took more than an hour to reach the submission call (which now has the deadline timeout).

haiyuechen-nearone (changes requested):
@pbeza please take a look at the blocking finding from Claude. I will be happy to approve if we added a hard stop at the outer layer to capture the whole process. 

I suggest that we add a timeout option for the dstack rpc call as well, but that is not blocking as it is adjacent to the issue fixed.

docs/design/attestation-verifier-contract.md
gilcu3, line 1: I think we no longer need to keep this file up-to-date, it was a design that was already implemented

crates/node/src/tick.rs
```diff
@@ -44,13 +56,34 @@ impl MockTicker {
...
+        // Given
+        let period = Duration::from_secs(60);
+        let mut interval = tokio::time::interval(period);
+        let started_at = Instant::now();
+
+        // When
+        let deadline = Tick::tick(&mut interval).await;
+
+        // Then
+        assert_eq!(deadline, started_at + period);
```
gilcu3: isn't this test a bit tautological? It appears to be just duplicating the existing code logic

crates/node/src/tee/remote_attestation.rs
gilcu3: checked and apparently the generate path does not have any deadline. Therefore, we need to add that if we want the deadline to cover both.

crates/node/src/tick.rs
```diff
@@ -3,16 +3,20 @@
...
+    async fn tick(&mut self) -> Instant;
 }
 
 impl Tick for tokio::time::Interval {
-    async fn tick(&mut self) {
-        self.tick().await;
+    async fn tick(&mut self) -> Instant {
+        self.tick().await + self.period()
     }
 }
```
gilcu3: the `tick()` might return an instant in the past. This is something we need to take into account if it is meant to cover the next round

gilcu3 (changes requested):
Added some comments, and requested a change in line with @haiyuechen-nearone's review, as apparently the generation path can take forever, which would affect what this PR is trying to achieve.

docs/design/attestation-verifier-contract.md
pbeza: Argh... I need to keep a closer eye on Claude so it doesn't update design docs. Sorry about that. Reverted: https://github.com/near/mpc/pull/4281/commits/df900f730ef77854380e03d636f647133a7b098f. ✅

crates/node/src/tee/remote_attestation.rs
pbeza: Created a follow-up issue: https://github.com/near/mpc/issues/4319. ✅

crates/node/src/tee/remote_attestation.rs
pbeza: Fixed: https://github.com/near/mpc/pull/4281/commits/8c97ae88f483b8c4164de64a0ab972db3e15ff3b. ✅

crates/node/src/tick.rs
pbeza: Removed in https://github.com/near/mpc/pull/4281/commits/8c97ae88f483b8c4164de64a0ab972db3e15ff3b. ✅

crates/node/src/tick.rs
pbeza: Fixed: https://github.com/near/mpc/pull/4281/commits/8c97ae88f483b8c4164de64a0ab972db3e15ff3b. ✅

crates/node/src/tee/remote_attestation.rs
pbeza: `setup.sender().count()` only counts successful sends, so it stays at `0` for the entire test. I added `attempts` to track all send attempts instead: https://github.com/near/mpc/pull/4281/commits/8c97ae88f483b8c4164de64a0ab972db3e15ff3b. ✅

crates/node/src/tee/remote_attestation.rs
```diff
@@ -161,20 +162,33 @@ impl<T: TransactionSender + Clone, A: GenerateAttestation> AttestationSubmitter<
...
             }
-            Err(error) => {
+            Ok(Err(error)) => {
                 tracing::error!(%error, "TEE attestation generation failed");
                 return;
             }
+            Err(error) => {
+                tracing::error!(%error, "TEE attestation generation timed out");
+                return;
+            }
```
gilcu3: nit: this reads very complicated, I would prefer two nested matches instead, but not sure if clippy agrees with that 😸

crates/node/src/tee/remote_attestation.rs
```diff
@@ -186,7 +200,9 @@ impl<T: TransactionSender + Clone, A: GenerateAttestation> AttestationSubmitter<
         let pre_submit_expiry = match self
             .attestation_reader
             .read_stored_attestation_expiry(&self.tls_public_key)
+            .timeout_at(deadline)
             .await
+            .unwrap_or_else(|elapsed| Err(elapsed.into()))
```
gilcu3: I didn't know this one could hang forever as well

crates/node/src/tick.rs
```diff
@@ -9,14 +9,15 @@ use tokio::sync::Semaphore;
...
     async fn tick(&mut self) -> Instant;
 }
 
 impl Tick for tokio::time::Interval {
     async fn tick(&mut self) -> Instant {
-        self.tick().await + self.period()
+        self.tick().await;
+        Instant::now() + self.period()
     }
 }
```
gilcu3, line 17: I am not sure about this pattern, maybe worth asking @netrome about this

gilcu3 (dismissed):
Thank you! LGTM, but I am not sure about the new tick pattern. Left a question for @netrome in a comment

crates/node/src/tick.rs
netrome: About returning the Instant? yeah that looks strange, why do we do this?

netrome: Seems like this is fed to the `generate_and_submit` method. It feels odd to have that method have to be aware of it's own deadline. Couldn't we just call it in an outer `tokio::select!` statement. Or is anything in that method not cancellation safe?

crates/node/src/tee/remote_attestation.rs
netrome: Hmm what's the point of having a deadline here? Hypothetically, assume this operation has higher latency than the tick interval, we wouldn't want to cancel all `generate_and_submit` actions right? We'd still want them to finish imo, so I'd rather just either let a `generate_and_submit` call finish, or if we're worried about straggler calls, just spawn the `generate_and_submit` in a new tokio task so we can make new calls without cancelling old ones.

haiyuechen-nearone: My assumption is that we do want to cancel the remainder of the call, if the attempt has overrun the budget.

I poked around a bit in the contract code, and yes technically node submitting a stale attestation when a fresh one is recorded is fine, but I feel thats relying on a resiliency feature when we could solve it "correctly" here.

As I noted in the other comments, I think we should ideally have timeout all the way down to the Dstack RPC calls, so that we will have observability into exactly where the timeout happened. We can for instance send down the deadline in all the calls in this path to achieve this. 

However, that is more changes and out of scope of the issue attached. As a "next best thing" that satisfies "One attestation re-gen and submission attempt per hour", I do agree with the Claude suggestion of putting a hard limit at this level and terminate the task before scheduling a new run.

crates/node/src/tick.rs
pbeza: [Discussed](https://nearone.slack.com/archives/C0912BTG51T/p1788265650995979?thread_ts=1788195764.674979&cid=C0912BTG51T) offline and addressed in https://github.com/near/mpc/pull/4281/commits/f3d138b3ba3c5e5dcc4891f908f042e65995670d. ✅

crates/node/src/tee/remote_attestation.rs
pbeza: Resolved by the rework in https://github.com/near/mpc/pull/4281/commits/f3d138b3ba3c5e5dcc4891f908f042e65995670d. ✅

crates/node/src/tee/remote_attestation.rs
pbeza: After https://github.com/near/mpc/pull/4281/commits/f3d138b3ba3c5e5dcc4891f908f042e65995670d the inline `timeout_at` is gone and the read is bounded by the round-level timeout in the loop instead.

crates/node/src/tee/remote_attestation.rs
```diff
@@ -46,14 +63,32 @@ pub struct AttestationSubmitter<T> {
...
+
+impl RoundStage {
+    fn as_label(self) -> &'static str {
+        match self {
+            RoundStage::GenerateAttestation => MPC_TEE_ATTESTATION_STAGE_GENERATE_ATTESTATION,
+            RoundStage::ReadExpiryBaseline => MPC_TEE_ATTESTATION_STAGE_READ_EXPIRY_BASELINE,
+            RoundStage::SubmitAttestation => MPC_TEE_ATTESTATION_STAGE_SUBMIT_ATTESTATION,
+        }
+    }
+}
```
gilcu3: well, I cannot say this is very idiomatic but seems nice enough for me

crates/node/src/tee/remote_attestation.rs
```diff
@@ -136,29 +169,30 @@ fn validate_remote_attestation(
...
             *self.tls_public_key.as_bytes(),
             *self.account_public_key.as_bytes(),
         )
         .into();
+        stage.send_replace(RoundStage::GenerateAttestation);
         let result = self.tee_authority.generate_attestation(report_data).await;
         MPC_TEE_ATTESTATION_ATTEMPTS_TOTAL
             .with_label_values(&[outcome_label(result.is_ok())])
             .inc();
         let attestation = match result {
```
gilcu3: one alternative that might be more idiomatic is to just split this function in the three parts we care about, and then unroll the calls to those in `periodic_attestation_submission`. Would require a bit more code there, but not much

crates/node/src/tee/remote_attestation.rs
```diff
@@ -221,42 +255,93 @@ fn outcome_label(succeeded: bool) -> &'static str {
...
-) {
+pub async fn run_periodic_attestation_submission<T, A>(submitter: AttestationSubmitter<T, A>)
+where
+    T: TransactionSender + Clone,
+    A: GenerateAttestation,
+{
     let mut interval = tokio::time::interval(ATTESTATION_RESUBMISSION_INTERVAL);
     interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
-    periodic_attestation_submission(submitter, interval).await
+    periodic_attestation_submission(submitter, interval, ATTESTATION_RESUBMISSION_INTERVAL).await
```
gilcu3: having to pass the two parameters here looks very weird

pbeza: Followed [your suggestion](https://github.com/near/mpc/pull/4281#discussion_r3911160572) to split `generate_and_submit` into three parts, so this parameter was removed: https://github.com/near/mpc/pull/4281/commits/086dae059baf81ab197e74a99cdb9cd87138ce26. ✅

crates/node/src/tee/remote_attestation.rs
pbeza: Love the idea! Refactored here: https://github.com/near/mpc/pull/4281/commits/086dae059baf81ab197e74a99cdb9cd87138ce26. ✅

crates/node/src/tee/remote_attestation.rs
pbeza: Followed [your suggestion](https://github.com/near/mpc/pull/4281#discussion_r3911160572) to split `generate_and_submit` into three parts, so this code was removed: https://github.com/near/mpc/pull/4281/commits/086dae059baf81ab197e74a99cdb9cd87138ce26. ✅

pbeza:
@gilcu3 I [followed](https://github.com/near/mpc/pull/4281/commits/086dae059baf81ab197e74a99cdb9cd87138ce26) your [suggestion](https://github.com/near/mpc/pull/4281#discussion_r3911160572) (thx for it!). It feels more idiomatic now. Let me know if this is how you pictured it, and feel free to re-review when you get a chance.

Also, @haiyuechen-nearone PTAL when you get a chance. Thanks, both!

haiyuechen-nearone (approved):
🚢

gilcu3 (approved):
Nice!

netrome (approved):
Nice, looks good!

Issue #4169, opened 2026-08-18
feat(tee): request the early TCB evaluation data set to detect TCB-R bumps before promotion


pbeza:
### Background

We fetch TCB collateral without an `update` parameter, so the PCCS default (`standard`) decides which TCB evaluation data set we verify against, and we only learn about a new TCB recovery when it becomes the standard set and our platforms start failing.

That is what happened on 2026-08-13/14, when Intel promoted `tcbEvaluationDataNumber` 19 to 20 for several FMSPCs. Platforms whose third TDX TCB component is `03` were demoted from `UpToDate` to `OutOfDate` with `INTEL-SA-01192`, `INTEL-SA-01245`, `INTEL-SA-01312`, `INTEL-SA-01313`. 7 of our 22 TDX nodes (1 on mainnet, 6 on testnet) stopped having attestations accepted from that moment, and it went unnoticed for days because nothing compares our platforms against the upcoming set.

Intel's PCS accepts `update=early` (and an explicit `tcbEvaluationDataNumber`) precisely so that the next set can be evaluated before promotion. `dcap-qvl` does not expose it: `url_tcb` sends no `update` at all and `url_qe_identity` hardcodes `update=standard`.

### User Story

As a node operator, I want to be told that my platform will be `OutOfDate` under the next TCB recovery while it is still accepted, so that I can schedule the BIOS, microcode and TDX module update before the contract stops accepting my attestation.

### Acceptance Criteria

- [ ] The collateral fetch can request a chosen evaluation data channel (`standard` or `early`), configurable per `pccs_endpoints` entry, defaulting to current behavior
- [ ] The node verifies its own quote against the early set in addition to the standard set it submits, and warns when the two verdicts differ
- [ ] A metric exposes that difference, so alerting can fire before promotion rather than after
- [ ] The parameter is supported upstream in `dcap-qvl`, or we document why the collateral client is wrapped locally
- [ ] TEE attestation docs describe which evaluation data set is fetched and why

### Resources & Additional Notes

- Intel PCS API v4 exposes `update=standard|early` and `tcbEvaluationDataNumber=<n>` on the TCB info and QE identity endpoints
- Fetch path: `TeeAuthority::fetch_collateral_from` in `crates/tee-authority/src/tee_authority.rs`, driving `dcap_qvl::collateral::CollateralClient`
- `/public_data` serves a boot time snapshot, so a node's served collateral can be much older than what it submits hourly; any diagnostics added here should read the freshly fetched bundle
[...]

PR #4285, opened 2026-08-28
feat(attestation-cli): report the early TCB evaluation data set

### LINKED ISSUE #4297: feat(attestation-cli): report the early TCB evaluation data set
### Background

We only learn about a TCB recovery once Intel promotes it and our platforms start failing. That is what happened on 2026-08-13/14: 7 of our 22 TDX nodes stopped having attestations accepted, and it went unnoticed for days.

Intel publishes the next evaluation data set under `update=early` before promoting it, so the demotion can be seen while the platform is still accepted.

### Acceptance Criteria

- [ ] `tcb-status` verifies the node's quote against the early set alongside the standard one
- [ ] A single set can be requested, so an automated check gets a plain exit code
- [ ] Docs describe which evaluation data set is fetched and why

pbeza:
`tcb-status` now verifies the node's quote against a third collateral set, the one Intel publishes under `update=early` before promoting it, so an operator learns that a coming TCB recovery will demote their platform while it is still accepted. `dcap-qvl` exposes no selector for the evaluation data set, so it is chosen in a caller-supplied `HttpClient` that rewrites the `update` query parameter. The exit code still follows the `standard` row alone, so a demoted `early` row exits 0.

Closes #4297 #4288


pbeza:
@gilcu3 @barakeinav1 could you review this when you get a chance? Next week is totally fine, ofc. Thanks!

barakeinav1:
#4169 is about node changes (the node checking the early set, and a metric for it), while this PR is CLI only. We should create separate issues for the node and the CLI, and link this PR to the CLI one.

crates/attestation-cli/src/tcb_status.rs
```diff
@@ -55,30 +61,55 @@ pub enum TcbVerdict {
...
 
 impl Report {
     pub fn is_up_to_date(&self) -> bool {
         matches!(&self.standard, TcbVerdict::Verified { claims, .. } if claims.tcb.status == TcbStatus::UpToDate)
     }
+
+    /// `Some((standard, early))` TCB recovery set numbers only when the platform
+    /// clears the standard set but not the early one, the set Intel publishes
+    /// and has yet to promote. `None` in all other cases.
+    pub fn early_demotion(&self) -> Option<(u32, u32)> {
```
barakeinav1: should we return different error code so it is easier to automate? right now early demotion exits 0, and everything else exits 1.

docs/securing-mpc-with-tee-design-doc.md
```diff
@@ -456,9 +456,11 @@ Beyond the TDX quote itself, attestation requires PCCS-supplied collateral — T
 
 - **TLS trust.** PCCS endpoints use standard public-CA trust roots by default. Each entry optionally accepts a TLS override: `tls.override = "ca_cert_pem"` (with a `tls.ca_cert_pem` PEM body) adds a self-signed root as an additional trust anchor — the default public-CA roots remain active. `tls.override = "insecure"` disables TLS validation entirely. Both overrides are intended for self-hosted local PCCS deployments only.
 
-- **Freshness validation.** Collateral older than 7 days is rejected. The threshold applies uniformly to `tcb_info.issueDate`, `qe_identity.issueDate`, and the PCK CRL `thisUpdate`. A stale response from one endpoint causes the node to fall back to the next.
+- **Freshness validation.** Collateral older than 31 days is rejected, just past Intel's 30-day re-issuance window. The threshold applies uniformly to `tcb_info.issueDate`, `qe_identity.issueDate`, and the PCK CRL `thisUpdate`. A stale response from one endpoint causes the node to fall back to the next.
```
barakeinav1, line 459: not strictly related to this PR. but thanks for updating.

crates/attestation-cli/src/tcb_status.rs
```diff
@@ -55,30 +61,55 @@ pub enum TcbVerdict {
...
     /// Parsed from the quote itself, so the platform's own numbers survive even
-    /// when neither collateral verifies.
+    /// when no collateral verifies.
     pub td_report: TDReport10,
     /// What the node's own boot-time collateral says, which can lag by days.
     pub served: TcbVerdict,
     /// What the contract decides today, against collateral fetched just now.
     pub standard: TcbVerdict,
+    /// What it will decide once Intel promotes the set it publishes early.
+    pub early: TcbVerdict,
```
barakeinav1: `early` is not always the *next* set. right now standard=20 and early=22, but 21 exists too (`tcbEvaluationDataNumber=21` returns 200) and we never show it.

for example for n1-multichain.near: it is TDX module SVN 12, and fails set 21 needing only module 13 - a SEAM loader bump. but the early row tells it to reach module 15 plus BIOS/microcode, which is set 22's bar. so we show the harder, later deadline and skip the one that lands first.

I think we should also probe for `standard+1..early-1` with `tcbEvaluationDataNumber=<n>`, and report back which next versions are accepted and which are not. this way operators can make better upgrade scheduling. 

fine as a follow up.

barakeinav1 (changes requested):
Requesting changes only because the PR does not match the issue - #4169 is about node changes while this PR is CLI only ([comment](https://github.com/near/mpc/pull/4285#issuecomment-5467280246)). Otherwise the PR looks good.

Also left a [follow-up suggestion](https://github.com/near/mpc/pull/4285#discussion_r3888813294) about probing the intermediate TCB sets, and a [question](https://github.com/near/mpc/pull/4285#discussion_r3888776558) about the exit codes.

crates/attestation-cli/src/output.rs
```diff
@@ -106,19 +106,29 @@ pub fn print_tcb_status(report: &Report) {
...
+    print_verdict("Intel `early`", &report.early);
+
+    if let Some((standard_set, early_set)) = report.early_demotion() {
+        println!();
+        println!(
+            "This platform clears TCB recovery set {standard_set} but not set {early_set}, which \
+             Intel publishes and has yet to promote. The `early` shortfall lines say what to \
+             update before it does."
+        );
+    }
```
gilcu3: nit: I would personally keep this tool as lean as possible regarding output directing users, to avoid pitfalls. The place to tell users what to do should be in `docs/tdx-tcb-status.md` instead

crates/attestation-cli/src/pcs.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+fn with_evaluation_data_set(url: &str, evaluation_data_set: TcbEvaluationDataSet) -> String {
+    let Some(mut parsed) = url::Url::parse(url).ok().filter(|parsed| {
+        evaluation_data_set == TcbEvaluationDataSet::Early
+            && (parsed.path().ends_with("/tcb") || parsed.path().ends_with("/qe/identity"))
+    }) else {
+        return url.to_owned();
+    };
```
gilcu3: why not failing is `url` does not have the expected format? This seems to open the possibility of errors in the future

crates/attestation-cli/src/pcs.rs
```diff
@@ -0,0 +1,118 @@
...
+#[cfg(test)]
+#[expect(non_snake_case)]
+mod tests {
+    use super::*;
+
+    const COLLATERAL_URLS: [&str; 4] = [
+        "https://api.trustedservices.intel.com/tdx/certification/v4/tcb?fmspc=00A06D080000",
+        "https://api.trustedservices.intel.com/tdx/certification/v4/qe/identity?update=standard",
+        "https://api.trustedservices.intel.com/sgx/certification/v4/pckcrl?ca=platform&encoding=der",
+        "https://certificates.trustedservices.intel.com/IntelSGXRootCA.der",
```
gilcu3: do we use so many formats already? Or this is just trying to adapt to dcap-qvl?

crates/attestation-cli/src/tcb_status.rs
gilcu3: For automation it would be nicer to let the cli request a single set. That way we don't need to override the typical return codes for automation.

crates/attestation-cli/src/tcb_status.rs
```diff
@@ -94,26 +125,31 @@ pub async fn run(static_data: &StaticWebData, as_of: Option<u64>) -> anyhow::Res
     };
 
     let served = collateral_into_dcap(dstack.collateral.clone());
+    let (standard, early) = tokio::join!(
+        fetch_from_intel(quote, TcbEvaluationDataSet::Standard),
+        fetch_from_intel(quote, TcbEvaluationDataSet::Early),
+    );
```
gilcu3: nit: seems a bit of an overkill to do them concurrently instead of sequentially.

gilcu3 (approved):
Thank you!

Left some minor comments. Approving modulo the assumption that the issue will be modified accordingly, as mentioned internally

gilcu3 (changes requested):
Tiny blocker detected by a bot:

We need to add another exception in:
https://github.com/near/mpc/blob/c366c4ae89146f3f651c9a697dcd4a2755c5379f/.config/nextest.toml#L27-L28

crates/attestation-cli/src/tcb_status.rs
pbeza: Fixed in https://github.com/near/mpc/pull/4285/commits/e66e445a5e775ce3090aeeb38e527dcb79edd134. ✅

crates/attestation-cli/src/pcs.rs
pbeza: I removed this file completely after merging https://github.com/Phala-Network/dcap-qvl/pull/223 and cutting a new `dcap-qvl` release with it included. Fixed in https://github.com/near/mpc/pull/4285/commits/e66e445a5e775ce3090aeeb38e527dcb79edd134 ✅

crates/attestation-cli/src/pcs.rs
pbeza: Ditto: https://github.com/near/mpc/pull/4285#discussion_r3894403641.

crates/attestation-cli/src/output.rs
pbeza: Fixed in https://github.com/near/mpc/pull/4285/commits/e66e445a5e775ce3090aeeb38e527dcb79edd134. ✅

crates/attestation-cli/src/tcb_status.rs
pbeza: Created a follow-up: https://github.com/near/mpc/issues/4299. ✅

crates/attestation-cli/src/tcb_status.rs
pbeza: Addressed in https://github.com/near/mpc/pull/4285/commits/e66e445a5e775ce3090aeeb38e527dcb79edd134. ✅

gilcu3 (approved):
Thank you!

barakeinav1 (approved):
thanks

PR #4287, opened 2026-08-28
feat: implement CLI that generates whitelist proposal declaratively

### LINKED ISSUE #4286: CLI for generating whitelist proposal declaratively
### Background

First (and only) time we drafted proposal, we hardcoded it in Rust structs and generated using it. That requires changes in the code every time we add new chain or RPC provider.

### User Story

_No response_

### Acceptance Criteria

Have CLI that generates proposal declaratively, without need to recompile binary.

_No response_

anodar:
closes: https://github.com/near/mpc/issues/4286


crates/node/src/foreign_chain_whitelist_verifier.rs
```diff
@@ -186,16 +186,32 @@ fn compare_provider(
 
 /// Path-boundary-aware prefix check. `https://api.example.com/v2` matches `/v2`,
 /// `/v2/eth`, `/v2?key=x`, `/v2#frag` — but not `/v2-evil`.
+///
+/// A `{}` in `base` matches one run of `[A-Za-z0-9-]` — a single host label in the bases we
+/// use, where a provider puts a per-operator slug in the hostname (e.g. Quicknode).
 fn base_url_matches(local: &str, base: &str) -> bool {
```
barakeinav1: suggestion: I know the PR is not very big, but can we split this into a separate PR? the `{}` support is not needed by the CLI - `proposal-generator` does not depend on the node crate, only the TOML uses it. so it can land on its own first, and the proposal can be voted after.

it also introduces a new convention for how `base_url` is interpreted on chain, and we already have a typed placeholder right next to it (`auth_scheme = { Path = { placeholder = "{API_KEY}" } }`) - worth its own review rather than sitting at the bottom of a CLI PR.

crates/proposal-generator/src/main.rs
```diff
@@ -0,0 +1,32 @@
...
+    let raw = std::fs::read_to_string(&args.config)
+        .with_context(|| format!("reading {}", args.config.display()))?;
+    let config: ProposalConfig = toml::from_str(&raw).context("parsing config")?;
+    let batch = build_batch(config)?;
+
+    let bytes = borsh::to_vec(&batch).context("borsh serialization")?;
+    println!(
+        "{}",
+        base64::engine::general_purpose::STANDARD.encode(&bytes)
+    );
```
barakeinav1: nit: could we also print the sha256 of the borsh bytes? that is the `batch_hash` the contract logs, so it is an easy way to confirm the transaction that landed carried the bytes you generated.

barakeinav1:
I think we should add a README capturing input, output, usage, etc. every other binary crate here has one (`attestation-cli`, `foreign-chain-config-tester`, `tee-launcher`), and right now the TOML schema only lives in the rustdoc of `lib.rs`.

crates/proposal-generator/proposals/testnet-rpc-whitelist.toml
```diff
@@ -0,0 +1,86 @@
+# Testnet RPC provider whitelist
```
barakeinav1, line 1: what about atoml mainnet? do we need it?

crates/proposal-generator/src/main.rs
```diff
@@ -0,0 +1,32 @@
...
+//!   cargo run -p proposal-generator -- crates/proposal-generator/proposals/testnet-rpc-whitelist.toml
+
+use anyhow::Context as _;
+use base64::Engine as _;
+use clap::Parser;
+use proposal_generator::{ProposalConfig, build_batch};
+use std::path::PathBuf;
+
+#[derive(Parser)]
+struct Args {
```
barakeinav1, line 15: nit: could we add `#[command(about = "...", version)]`? `--help` currently says nothing about what the tool does, and both `attestation-cli` and `foreign-chain-config-tester` set `about`.

`version` is worth having here too - the output is borsh over the `ChainEntry` DTO, so two participants building from different commits could produce different bytes with nothing in the output explaining why.

crates/proposal-generator/src/main.rs
```diff
@@ -0,0 +1,32 @@
...
+    /// Path to the TOML proposal config.
+    config: PathBuf,
+}
+
+fn main() -> anyhow::Result<()> {
+    let args = Args::parse();
+    let raw = std::fs::read_to_string(&args.config)
+        .with_context(|| format!("reading {}", args.config.display()))?;
+    let config: ProposalConfig = toml::from_str(&raw).context("parsing config")?;
```
barakeinav1: nit: the line above names the file, this one doesn't - worth doing the same so you know which config failed.

```suggestion
    let config: ProposalConfig = toml::from_str(&raw)
        .with_context(|| format!("parsing {}", args.config.display()))?;
```

crates/proposal-generator/src/lib.rs
```diff
@@ -0,0 +1,336 @@
...
+use serde::Deserialize;
+
+#[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct ProposalConfig {
+    pub chains: NonEmptyBTreeMap<ForeignChain, ChainEntry>,
+}
+
+/// Mirrors the `ChainEntry` validation the contract runs on a vote; a batch failing any of
+/// these is rejected on chain with `InvalidParameters::MalformedPayload`.
```
barakeinav1: this duplicates the contract's `TryFrom<dtos::ChainEntry>` check for check. they match today, but if the contract adds a rule we would silently tell people their proposal is valid.

two options: depend on `mpc-contract` and just call `try_into()`, which deletes this whole enum - `crates/node` already depends on it that way. or move the validation into `near-mpc-contract-interface` next to the DTO so both sides share it. I'd slightly prefer the first.

crates/proposal-generator/src/lib.rs
```diff
@@ -0,0 +1,336 @@
...
+//! [`ForeignChain`] variant names, chain tables are [`ChainEntry`].
+
+use near_mpc_bounded_collections::NonEmptyBTreeMap;
+use near_mpc_contract_interface::types::{
+    AuthScheme, ChainEntry, ChainRouting, ForeignChain, ProviderId,
+};
+use serde::Deserialize;
+
+#[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
```
barakeinav1: nit: this only covers the top level. `ChainEntry` and `ProviderConfig` come from `near-mpc-contract-interface` without it, so a stray key inside a chain or provider table is accepted silently:

```toml
[chains.Bitcoin]
quorum = 1
timeout = 5     # ignored, no warning
```

worth adding `deny_unknown_fields` there too - this file becomes a governance vote, so config that does nothing shouldn't pass quietly.

crates/proposal-generator/src/lib.rs
```diff
@@ -0,0 +1,336 @@
...
+        assert_eq!(
+            bitcoin.providers[&ProviderId("public".into())].chain_routing,
+            ChainRouting::PathSegment {
+                segment: "btc".into(),
+            }
+        );
+    }
+
+    #[test]
+    fn testnet_rpc_whitelist__should_be_a_valid_proposal() {
```
barakeinav1: the tests all stop at `build_batch`, nothing checks the borsh/base64 output. I think we should add a fixture test to make sure the encoding is correct.

crates/node/src/foreign_chain_whitelist_verifier.rs
```diff
@@ -186,16 +186,32 @@ fn compare_provider(
...
+        return starts_with_at_boundary(l, b);
+    };
+    let Some(rest) = l.strip_prefix(prefix) else {
         return false;
     };
-    rest.starts_with('/') || rest.starts_with('?') || rest.starts_with('#')
+    let label_len = rest
+        .find(|c: char| !c.is_ascii_alphanumeric() && c != '-')
+        .unwrap_or(rest.len());
+    label_len > 0 && starts_with_at_boundary(&rest[label_len..], suffix)
```
barakeinav1: nit (found by Claude): `{}` only works when the character right after it isn't a label character. the match here is greedy over `[A-Za-z0-9-]` with no backtracking, so:

```
base   https://api-{}-v2.example.com
local  https://api-acme-v2.example.com
```

the label eats `acme-v2`, leaving `.example.com` to match against `-v2.example.com` - no match, silently. fine for our `{}.host` bases, but worth saying so in the doc.

barakeinav1 (changes requested):
Requesting changes mainly for the [duplicated validation](https://github.com/near/mpc/pull/4287#discussion_r3889580115) of the contract's checks. I'd also prefer to [split out the `{}` matcher](https://github.com/near/mpc/pull/4287#discussion_r3889527724) since it is node code and not needed by the CLI, but if you'd rather keep it here I'm ok with that. Otherwise the CLI looks good, the rest are nits.

crates/proposal-generator/src/main.rs
```diff
@@ -0,0 +1,32 @@
...
+    let batch = build_batch(config)?;
```
kevindeforth: nit: the function name `build_batch` got me confused, as it's  unclear what we are batching.
Reading the code of the method, it looks like we're only validating the configs and returning them in case all are valid.

crates/proposal-generator/src/lib.rs
```diff
@@ -0,0 +1,336 @@
...
+    }
+    Ok(config.chains)
+}
+
+fn validate_entry(chain: ForeignChain, entry: &ChainEntry) -> Result<(), ConfigError> {
+    let providers =
+        u64::try_from(entry.providers.len()).map_err(|_| ConfigError::ProvidersLenOverflow {
+            chain,
+            providers: entry.providers.len(),
+        })?;
```
kevindeforth: This is quite defensive programming.

`NonEmptyBTreeMap` wraps a `BTreeMap`, which stores its length in a `usize` field.
Afaik, there isn't a rust target that allows for `usize` greater than `u64`.

I think an `.expect("usize must fit in u64")` is appropriate and saves us an error type.

crates/proposal-generator/src/lib.rs
```diff
@@ -0,0 +1,336 @@
...
+    if entry.quorum == 0 || entry.quorum > providers {
+        return Err(ConfigError::InvalidQuorum {
+            chain,
+            quorum: entry.quorum,
+            providers,
+        });
+    }
```
kevindeforth: I assume the contract also does these checks.
Have you considered adding the validation logic to the contract-interface crate, behind a feature `validation` feature flag?

kevindeforth (commented):
Thank you, only a nit and an error-type I think we cave ourselves.

Have you thought about how to avoid the inevitable drift between contract and consumer validation logic? The obvious solution to that problem would be to move validation logic into the contract-interface crate, but I fear that might bring some issues on its own.

anodar:
> Have you thought about how to avoid the inevitable drift between contract and consumer validation logic? The obvious solution to that problem would be to move validation logic into the contract-interface crate, but I fear that might bring some issues on its own.

Yes but didn't want to depend on mpc-contract or move impl into contract interface. But now that I think a bit more about it, depending on mpc-contract is probably better than possibly diverging verifications.

crates/proposal-generator/src/lib.rs
anodar: Introduced dependency on mpc-contract.

crates/proposal-generator/src/lib.rs
anodar: No longer applicable since I got rid of these checks.

crates/proposal-generator/proposals/testnet-rpc-whitelist.toml
anodar: This was just for testnet for now, but added mainnet while I'm at it.

crates/proposal-generator/src/main.rs
anodar: Done

crates/proposal-generator/src/main.rs
anodar: Done.

crates/proposal-generator/src/main.rs
anodar: Done.

crates/node/src/foreign_chain_whitelist_verifier.rs
anodar: Split in 2 prs, base is now: https://github.com/near/mpc/pull/4296/changes.

crates/proposal-generator/src/lib.rs
anodar: I had issue previously https://github.com/near/mpc/issues/3587 from inflexibility of deny_unknown_fields, so I prefer to drop it from here as well instead of adding it to chain entries.

crates/proposal-generator/src/lib.rs
anodar: Moved ` borsh::to_vec(&config.chains)` in building function and now it's covered with tests.

crates/node/src/foreign_chain_whitelist_verifier.rs
anodar: Base PR should clarify that a bit better now.

crates/proposal-generator/src/main.rs
anodar: Unfortunate name is remnant from few iterations of that function where I initially was doing encoding inside it. Now that I again moved ` borsh::to_vec(&config.chains)` back in there, name should be more aligned. Also renamed batch -> payload.

crates/proposal-generator/src/main.rs
barakeinav1: `about` landed but not `version` - that was the half I was after, since the payload is borsh over the DTOs and two people on different commits would get different bytes.

crates/proposal-generator/src/lib.rs
barakeinav1: fair, your call. #3587 was a node reading config newer than itself though - here the TOML ships in the same commit as the binary, so no skew, and a silently ignored key means a payload that doesn't say what you think.

barakeinav1 (dismissed):
Left 2 comments but not blocking. Thanks for addressing the rest.

crates/node/src/foreign_chain_whitelist_verifier.rs
```diff
@@ -205,6 +205,7 @@ fn base_url_matches(local: &str, base: &str) -> bool {
     let label_len = rest
         .find(|c: char| !c.is_ascii_alphanumeric() && c != '-')
         .unwrap_or(rest.len());
+    let (_, rest) = rest.split_at(label_len);
```
anodar, line 208: Addressing comment from base PR: https://github.com/near/mpc/pull/4296#discussion_r3896994691

crates/proposal-generator/proposals/mainnet-rpc-whitelist.toml
kevindeforth, line 1: Does it make sense to have these in main, but not store RPC providers in the contract? One can easily infer what providers the nodes are using by looking at these proposals.
We could keep this private if we wanted.

crates/proposal-generator/proposals/testnet-rpc-whitelist.toml
kevindeforth, line 1: idem

barakeinav1 (approved):
Re-approving after the force-push.

crates/proposal-generator/src/lib.rs
```diff
@@ -0,0 +1,239 @@
...
+    pub chain: ForeignChain,
+    pub source: ChainEntryValidationError,
+}
+
+/// Validates each entry with the contract's own rules
+/// ([`mpc_contract::foreign_chain_rpc::ChainEntry`]), then borsh-encodes the
+/// `vote_update_foreign_chain_providers` argument.
+pub fn build_payload(config: ProposalConfig) -> anyhow::Result<Vec<u8>> {
+    for (chain, entry) in config.chains.iter() {
+        mpc_contract::foreign_chain_rpc::ChainEntry::try_from(entry.clone()).map_err(|source| {
```
kevindeforth, line 27: This isn't ideal. We avoid depending on the contract directly: it makes it too easy to use contract internals, which then makes contract development harder.

Here, I question the need for the contract-internal `ChainEntry` type:
https://github.com/near/mpc/blob/2d1f2e8aac71fceed0dff8b17a9029310a16277d/crates/contract/src/foreign_chain_rpc.rs#L45-L48

it's the same as the one stored in the contract-interface:

https://github.com/near/mpc/blob/2d1f2e8aac71fceed0dff8b17a9029310a16277d/crates/near-mpc-contract-interface/src/types/foreign_chain.rs#L1971-L1976

I would suggest we delete the contract-internal `ChainEntry` type and move the validation logic to the Contract Interface crate, or alternatively, to the `primitives` crate.

crates/proposal-generator/README.md
```diff
@@ -0,0 +1,55 @@
+# Proposal generator
+
+This tool is meant for MPC team to generate proposals for foreign chain whitelist
+voting and share generated payload with operators. It the argument for
+foreign chain whitelist voting from TOML config file.
```
kevindeforth, line 5: > It the argument for

I think we miss an "is"

kevindeforth (approved):
Thanks a lot!
This is good, except for the dependency on the mpc-contract. We want to avoid that, because it has the potential to complicate future contract development.

That would be worth blocking the PR, but during review, I also noticed that we don't really need the contract-internal representation of `ChianEntry` on which this PR depends.
So in any case, we will have a PR that removes the contract-internal type and moves the validation logic to the interface crate (or the primitive crate). If we do this before or after this PR doesn't really matter too much, **as long as we do it in a timely manner**. Once done, the `proposal-generator` crate introduced in this PR will no longer depend on the contract.

crates/proposal-generator/proposals/mainnet-rpc-whitelist.toml
anodar: We do store whitelisted providers in the contract. Whole whitelist isn't meant to be private, but which node is using which RPC providers. 

Not how it is right now, but that's the plan.

crates/proposal-generator/src/lib.rs
anodar: Will followup in https://github.com/near/mpc/issues/4313

crates/proposal-generator/README.md
anodar: Will address in the same followup as above.

PR #4290, opened 2026-08-31
refactor: define attestation types in their own crate

### LINKED ISSUE #4289: make attestation types available as a separate crate
### Background

During the work on #2480 I realized some types in the mpc-contract public API, not present in the `mpc-contract-interface` come directly from the `attestation` crate. To be able to ~~move~~ reuse them on the interface crate, we need to separate them from the attestation crate so the `mpc-contract-interface` can access them without needing to depend on the `attestation` crate directly.

### User Story

_No response_

### Acceptance Criteria

The `attestation` types used in the contract public API are moved to a new `attestation-types` crate

_No response_

gilcu3:
Closes #4289


crates/attestation/src/measurements.rs
```diff
@@ -69,22 +56,3 @@ impl From<&crate::tcb_info::TcbInfo> for Measurements {
...
-            .as_td10()
-            .ok_or(MeasurementsError::NoTd10Report)?;
-        Ok(Self {
-            rtmr0: td10.rt_mr0,
-            rtmr1: td10.rt_mr1,
-            rtmr2: td10.rt_mr2,
-            mrtd: td10.mr_td,
-        })
-    }
-}
```
gilcu3, line 90: this conversion was removed because it was only being used in a test (now inlined there) and was forcing a dependency on `tee-verifier-interface` that is now not needed

crates/attestation-types/src/lib.rs
```diff
@@ -0,0 +1,11 @@
...
+//! against, and how those values are serialized.
+
+#![no_std]
+
+extern crate alloc;
+
+pub mod app_compose;
+pub mod measurements;
+pub mod report_data;
+pub mod tcb_info;
```
kevindeforth: I prefer explicit export of the types we need exposed, otherwise the file-layout will be part of the API.

I.e. keep these modules private, but export the stuff we need as `pub use`.

kevindeforth (changes requested):
Looks good, but I'd prefer we don't make the file-tree part of the exposed API

kevindeforth (approved):

pbeza (approved):
Nice, feels like a good step toward making the attestation primitives easier to reuse across teams using TEEs. 🚀

Issue #3946, opened 2026-07-23
Investigation: decide on a permanent PCCS collateral freshness policy


barakeinav1:
### Background

The node hard-codes a maximum age for PCCS collateral — `MAX_COLLATERAL_AGE` in `crates/tee-authority/src/tee_authority.rs` — applied to the PCK CRL `thisUpdate` and the `tcb_info` / `qe_identity` `issueDate`. It was introduced at **7 days** (#3042 / #3049) and is being bumped to **31 days** as an emergency stopgap (#3941 → release-v3.13, #3945 → main).

On 2026-07-23 the 7-day limit rejected Intel's then-current PCK CRL fleet-wide: `thisUpdate = 2026-07-16`, `nextUpdate = 2026-08-15` — a valid ~30-day CRL, but >7 days old. All nodes stopped being able to generate an attestation. Verified identical from Phala, Intel's own PCS, and a local `sgx-dcap-pccs` — i.e. it's Intel's normal issuance cadence, not a mirror being stale.

The `#3049` rationale assumed *"Intel reference and Phala both refresh ~daily"*, which does not match what Intel actually serves (a single CRL held for most of its 30-day window). The 31-day bump unblocks today but leaves almost no margin against Intel's `nextUpdate` and doesn't resolve the underlying design question.

### Goal

Research Intel's real collateral-refresh behavior and decide on a principled, permanent freshness policy — rather than a hand-tuned fixed age.

### Questions to answer

- What is Intel's actual re-issuance cadence for PCK CRL / `tcb_info` / `qe_identity` in practice (not the assumed ~daily)?
- Should freshness be validated against the collateral's **own validity window** (e.g. accept while `now < nextUpdate`, optionally minus a safety margin) instead of a fixed max-age? What are the security trade-offs (this is what `#3049` was trying to bound)?
- If a fixed max-age is kept, what value gives safe margin over Intel's cadence without weakening the intent behind the 7-day choice?
- Should the limit be operator-configurable, or stay hard-coded / image-hash-governed?
- How should this align with the contract's attestation expiry (`DEFAULT_EXPIRATION_DURATION_SECONDS`, currently 1 day) so a node that attested while collateral was fresh isn't dropped far sooner than it needs to be?

### Acceptance Criteria

- [ ] Documented findings on Intel's real collateral re-issuance cadence.
- [ ] A decision on the freshness-check design (fixed age vs validity-window-based vs other), with rationale.
[...]

PR #4292, opened 2026-08-31
docs(tee): design for certificate-derived attestation expiry

### LINKED ISSUE #4291: design: certificate-derived attestation expiry
### Background

[#1639](https://github.com/near/mpc/issues/1639) replaces the contract's hardcoded 7-day attestation expiry (`DEFAULT_EXPIRATION_DURATION_SECONDS`, [`crates/mpc-attestation/src/attestation.rs:28`](https://github.com/near/mpc/blob/main/crates/mpc-attestation/src/attestation.rs#L28)) with the validity window Intel embeds in the collateral.

The change is larger than it looks. `dcap-qvl` 0.6.1 already computes the number we want (`QuoteClaims::earliest_expiration_date`), but reaching it means changing what `verify_quote` returns — and the `tee-verifier` account is key-locked and trusted by vote, so that implies a new verifier deployment plus `vote_tee_verifier_change`. It also touches `mpc-attestation`, `mpc-contract`, the node's submit-confirmation heuristic ([`crates/node/src/indexer/tx_sender.rs:201`](https://github.com/near/mpc/blob/main/crates/node/src/indexer/tx_sender.rs#L201), already carrying a `TODO(#1639)`), and the operator-facing health signal in [`docs/tdx-tcb-status.md:110`](https://github.com/near/mpc/blob/main/docs/tdx-tcb-status.md#L110).

Agree the design and the rollout order before writing the code, so the implementation can land as reviewable pieces.

### Acceptance Criteria

- [ ] `docs/design/certificate-derived-attestation-expiry.md` covering where the expiry is computed, how it combines with the contract-side maximum, the verifier interface change, and the node-side impact.
- [ ] Operations plan: rollout order and what changes for node operators.
- [ ] Open questions and rejected alternatives listed, each with a recommendation.
- [ ] Implementation split into follow-up issues.

- Certificate/CRL validity analysis: [#1629 (comment)](https://github.com/near/mpc/issues/1629#issuecomment-3635983666)
- [`docs/design/attestation-verifier-contract.md`](https://github.com/near/mpc/blob/main/docs/design/attestation-verifier-contract.md) — uses the expiry as the containment lever after a verifier rotation
- Related: [#4005](https://github.com/near/mpc/issues/4005) (mock expiry alignment), [#3264](https://github.com/near/mpc/issues/3264) (DCAP moved to the verifier contract)

barakeinav1:
Closes #4291

Design doc for #1639 — no code changes.

- **`expiry = earliest_collateral_expiration`.** `DEFAULT_EXPIRATION_DURATION_SECONDS` is deleted and the contract keeps no expiry constant of its own. Following review, there is no contract-side cap.
- **Why now:** the 7-day constant was sound while the node refused collateral older than 7 days, which kept it inside Intel's 30-day window. `MAX_COLLATERAL_AGE` is now 31 days (#3946), so that no longer holds and the overhang becomes a revocation window.
- **Where the value comes from:** a new `verify_quote_with_claims` on `tee-verifier`, using `dcap-qvl`'s `QuoteClaims::earliest_expiration_date`. A second method rather than a changed return type, so a verifier rotation cannot mis-decode.
- **What else has to change:** the stored submission timestamp (#4301), launcher-image eviction, near-expiry collateral, the sandbox fixture, and the gas config vote. Verifier-rotation shortening (#3734) is future work the design only has to keep possible.
- **Gas is measured, not estimated:** `verify_quote` burns 175.8 of its 200 TGas, and the whole submission burns 199.7 against the 300 TGas protocol ceiling. Headroom has to be moved from `resolve_verification`, not added.
- Three open questions remain, each needing a number.

### DIFF
--- a/docs/design/certificate-derived-attestation-expiry.md
+++ b/docs/design/certificate-derived-attestation-expiry.md
@@ -1,0 +1,185 @@
+# Certificate-Derived Attestation Expiry
+
+Design for [#1639](https://github.com/near/mpc/issues/1639): an attestation stops being trusted when
+Intel's collateral stops guaranteeing freshness, instead of a fixed 7 days later.
+
+```text
+expiry = min(now + MAX_ATTESTATION_VALIDITY_SECONDS, earliest_collateral_expiration)
+```
+
+The certificate value can only **shorten** the window. `MAX_ATTESTATION_VALIDITY_SECONDS` is today's
+`DEFAULT_EXPIRATION_DURATION_SECONDS` (7 days), renamed and documented as a ceiling.
+
+## Background
+
+DCAP rejects collateral past its `nextUpdate`, but the 7 days are added to *submission time* — and
+the submitter chooses when to submit. So a node can present collateral with an hour of validity left
+and receive a full 7 days of trust. Nothing rejects it: DCAP only asks whether the collateral is
+valid *right now*.
+
+That matters because the contract cannot learn about a newer CRL — the premise #1639 opens with. A
+stored attestation is never re-checked against updated collateral; trust ends only at expiry. So a
+platform Intel revokes in the *next* PCK CRL keeps a valid on-chain attestation for up to 7 days
+after the CRL that vouched for it stopped being authoritative, and an operator wanting that
+extension only has to time its submissions. Until on-chain CRL updates exist
+([#1050](https://github.com/near/mpc/issues/1050)), this expiry is the only bound on revocation
+latency.
+
+Intel refreshes the PCK CRL, TCB Info and QE Identity every 30 days, the root CA CRL yearly, and
+certificates last 7–30 years ([analysis](https://github.com/near/mpc/issues/1629#issuecomment-3635983666)),
+so the binding constraint is always one of the three 30-day items. `dcap-qvl` 0.6.1 — already our
+pin, with `default-x509` enabled — computes it as `QuoteClaims::earliest_expiration_date`, the
+minimum of `nextUpdate`/`notAfter` over all eight collateral sources.
+
+### Relation to the collateral freshness policy
+
+The check removed in [#3736](https://github.com/near/mpc/pull/3736) — the node's 120-second
+`MAX_ATTESTATION_AGE` — was a *self-confirmation* check, never a gate on what the contract accepted,
+and it went because it was broken ([#3686](https://github.com/near/mpc/issues/3686)). So this design
+does not restore it.
+
+The surviving policy is the node's `MAX_COLLATERAL_AGE`, 31 days
+([`tee_authority.rs:245`](../../crates/tee-authority/src/tee_authority.rs)). Being *looser* than
+Intel's 30-day window it constrains nothing in the tail — exactly where the overhang lives — and it
+binds only a node's own submission. [#3946](https://github.com/near/mpc/issues/3946) asks whether
+freshness should be judged against the collateral's own window rather than a hand-tuned max-age; this
+design answers yes for the contract side, leaving #3946 as a narrower node-side question.
+
+## Design
+
+### Where the expiry comes from
+
+`tee-verifier` gains a new method, `verify_quote_with_claims`, returning the report plus
+`earliest_expiration_seconds`. `verify_quote` stays as it is. Both halves of that follow from the
+verifier account being key-locked and trusted by vote, so a new return type means a new account and
+a `vote_tee_verifier_change`:
+
+- **A second method, not an extra Borsh field.** During a rotation an old verifier may answer a new
+  contract or the reverse. Appending a field breaks one direction whichever order we roll out; a
+  distinct method name breaks neither.
+- **Named after the returned shape, not the new field.** The name can never be changed on that
+  account, and `..._with_claims` still reads true if a later version returns another `QuoteClaims`
+  field.
+
+The body swaps `dcap_qvl::verify::verify` for
+`QuoteVerifier::new_prod().verify_with_policy(quote, collateral, now, &QuotePolicy::claims_only(now))`
+and rebuilds today's `VerifiedReport` from the returned `QuoteClaims`, which carries every field it
+needs.
+
+`mpc-contract` threads that value from `resolve_verification` into `verify_and_store_dstack` and on
+into `AcceptedAttestation::dstack`, which applies the `min`. `ValidatedDstackAttestation`'s layout is
+unchanged, so **no state migration**. `verify_locally` (node, CLI, `tee-authority`) reads the same
+number from `claims()`, so off-chain verification agrees with the chain. The mock path keeps stamping
+`now + MAX`; [#4005](https://github.com/near/mpc/issues/4005) stays open.
+
+### Why the cap stays
+
+- `Config::validate` requires `launcher_hash_unused_ttl_seconds >= DEFAULT_EXPIRATION_DURATION_SECONDS`
+  (14 days ≥ 7): a launcher hash backing a still-valid attestation must not be evicted first.
+- [#3734](https://github.com/near/mpc/issues/3734) specifies 7 days by default and 1 day on verifier
+  rotation, for containment of entries a rotated-away verifier may have wrongly accepted. `MAX` is
+  therefore a value the protocol varies deliberately, so it has to stay the ceiling.
+
+Keeping the ceiling means the certificate value binds only in the tail — when collateral is within
+`MAX` of its `nextUpdate`. That tail is the gap this design closes. What `MAX` should be is #3734,
+not this document.
+
+### Node submit-confirmation
+
+The node confirms a submission landed by checking that the stored expiry *changed*
+([`tx_sender.rs:201`](../../crates/node/src/indexer/tx_sender.rs)), which works only because the
+expiry is always `now + 7 days`. Once the certificate value binds and the collateral has not changed,
+re-submitting stamps the same expiry, so accepted and rejected become indistinguishable in contract
+state — no state-based rule can separate them.
+
+Fixed by reading the receipt execution outcome instead:
+[#4301](https://github.com/near/mpc/issues/4301), which needs no additional tracked shard. Best
+landed **before** this change, so there is never a window where a rejected submission reports as
+success.
+
+## Gas
+
+Measured on mainnet (`v1.signer` → `tee-verifier-2026-08-04.near`).
+
+Gas limit is the ceiling a receipt may spend before running out of gas; the unburnt remainder is
+refunded.
+
+| Receipt | Burnt | Gas limit | Set by |
+|---|---|---|---|
+| `submit_participant_info` | 16.28 | 300 | node's prepaid gas — protocol max, covers the whole chain |
+| `verify_quote` | **175.81** | 200 | `verifier_tera_gas`, static gas on the cross-contract call |
+| `resolve_verification` | 4.60 | 60 | `resolve_verification_tera_gas`, static gas on the callback |
+| Total burnt, incl. tx and refunds | 199.73 | | |
+
+The 300 is the chain's budget, not the verifier's: the submit receipt spends 16.3 itself and commits
+200 + 60 to its two promises, leaving only ~20 TGas unallocated. So `verifier_tera_gas` can reach
+~220 as things stand, or ~270 if `resolve_verification`'s 60 is trimmed toward its 4.6 — keeping
+worst-case room there for larger allowlists. Not 300.
+
+`claims()` re-parses `tcb_info` and `qe_identity` and parses six certificates, so measure it before
+choosing the split. Fallback if it still does not fit: a lean computation reading `nextUpdate` from
+the two CRLs plus a minimal `{ nextUpdate }` struct for the JSON documents, which avoids
+materializing all `tcbLevels`.
+
+## Operations plan
+
+The verifier is live on both networks (`tee-verifier-2026-08-04.near`,
+`tee-verifier-2026-07-22.testnet`), so this is a rotation, not a first deployment.
+
+1. **Land [#4301](https://github.com/near/mpc/issues/4301)** (node confirmation).
+2. **Deploy the new verifier and vote it in**, per [`deploy-tee-verifier.md`](../deploy-tee-verifier.md).
+   It still serves `verify_quote`, so nothing changes on chain yet; reversible by voting back.
+3. **Upgrade `mpc-contract`** to call `verify_quote_with_claims`. Certificate-derived expiry takes
+   effect here.
+
+Operator-visible change: a healthy node's `expiry_timestamp_seconds` is no longer always ~7 days out
+and no longer advances on every hourly submission — it stops advancing once the collateral's window
+is the shorter of the two. So a node whose PCCS has gone stale shows a *shrinking* expiry before a
+stuck one. [`tdx-tcb-status.md`](../tdx-tcb-status.md) sells the old behaviour as the cheapest health
+signal and needs rewriting, as does the `mpc_attestation_expiry_timestamp_seconds` metric
+description from [#4236](https://github.com/near/mpc/pull/4236).
+
+## Open questions
+
+- **How much gas does `claims()` add, and where does it come from?** Measure first; re-balancing
+  against `resolve_verification` is the expected answer, the lean computation the fallback.
+- **Order.** #4301 is listed first because it removes a window where rejections report as success.
+  Confirm that is worth blocking on.
+
+## For discussion: cap, or pure certificate-derived?
+
+The one choice here that changes the security argument rather than the mechanism.
+
+**For pure certificate-derived:** it is the only expiry Intel vouches for; every shorter window is
+our own invention, and a self-chosen number is what caused the fleet-wide outage on 2026-07-23, when
+the node's then 7-day `MAX_COLLATERAL_AGE` rejected a valid 30-day PCK CRL and no node could attest
+([#3946](https://github.com/near/mpc/issues/3946)). One freshness authority instead of two that can
+disagree.
+
+**For the cap:** Intel holds one CRL for most of its 30-day window, so pure certificate-derived is
+~30 days in the common case — four times today's window, too long for #3734's containment, and it
+breaks the `launcher_hash_unused_ttl_seconds` invariant unless that TTL is raised past 30 days.
+
+**Middle ground:** keep the cap but resolve #3946 and #3734 together, so the node's collateral-age
+bound and the contract's `MAX` come from one reading of Intel's cadence rather than being tuned
+separately.
+
+Recommendation: ship the cap, a strict improvement either way, and revisit if the team prefers a
+single freshness authority.
+
+## Alternatives considered
+
+- **Derive the expiry in `mpc-contract`** from the collateral it already round-trips through
+  `VerificationContext`. Needs no verifier change or rotation — but covering the PCK CRL means
+  re-adding DER/X.509 parsing to the contract wasm, undoing
+  [#3264](https://github.com/near/mpc/issues/3264).
+- **Floor the window to a minimum.** Reintroduces the gap this design closes.
+- **An `attested_at_seconds` field** on the stored entry, for exact submit confirmation. +8 bytes
+  (599 → 607, so `WORST_CASE_ENTRY_BYTES` and the fee floor need re-checking) *and* a state
+  migration. #4301 achieves the same with neither.
+
+## Follow-up issues
+
+- [#4301](https://github.com/near/mpc/issues/4301) — node submit-confirmation from the receipt
+  outcome (opened).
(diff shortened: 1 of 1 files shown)


docs/design/certificate-derived-attestation-expiry.md
```diff
@@ -0,0 +1,185 @@
...
+after the CRL that vouched for it stopped being authoritative, and an operator wanting that
+extension only has to time its submissions. Until on-chain CRL updates exist
+([#1050](https://github.com/near/mpc/issues/1050)), this expiry is the only bound on revocation
+latency.
+
+Intel refreshes the PCK CRL, TCB Info and QE Identity every 30 days, the root CA CRL yearly, and
+certificates last 7–30 years ([analysis](https://github.com/near/mpc/issues/1629#issuecomment-3635983666)),
+so the binding constraint is always one of the three 30-day items. `dcap-qvl` 0.6.1 — already our
+pin, with `default-x509` enabled — computes it as `QuoteClaims::earliest_expiration_date`, the
+minimum of `nextUpdate`/`notAfter` over all eight collateral sources.
```
pbeza, line 32: I think it would be good to name these 8 sources here, because it’s a bit unclear what they are and what their individual expiration times are.

docs/design/certificate-derived-attestation-expiry.md
```diff
@@ -0,0 +1,185 @@
+# Certificate-Derived Attestation Expiry
+
+Design for [#1639](https://github.com/near/mpc/issues/1639): an attestation stops being trusted when
+Intel's collateral stops guaranteeing freshness, instead of a fixed 7 days later.
+
+```text
+expiry = min(now + MAX_ATTESTATION_VALIDITY_SECONDS, earliest_collateral_expiration)
```
pbeza, line 7: Why do you still want to keep `MAX_ATTESTATION_VALIDITY_SECONDS` / `DEFAULT_EXPIRATION_DURATION_SECONDS`? Wasn’t the original goal of https://github.com/near/mpc/issues/1639 to get rid of arbitrary constants in the contract?

If we implement this, we’ll effectively shorten the attestation validity for nodes, potentially down to just a few seconds in corner cases, while nodes only resubmit attestations once an hour.

docs/design/certificate-derived-attestation-expiry.md
```diff
@@ -0,0 +1,185 @@
...
+  contract or the reverse. Appending a field breaks one direction whichever order we roll out; a
+  distinct method name breaks neither.
+- **Named after the returned shape, not the new field.** The name can never be changed on that
+  account, and `..._with_claims` still reads true if a later version returns another `QuoteClaims`
+  field.
+
+The body swaps `dcap_qvl::verify::verify` for
+`QuoteVerifier::new_prod().verify_with_policy(quote, collateral, now, &QuotePolicy::claims_only(now))`
+and rebuilds today's `VerifiedReport` from the returned `QuoteClaims`, which carries every field it
+needs.
```
pbeza, line 67: When implementing the verifier contract, my intention was to keep its API 1:1 with the raw `dcap-qvl` `verify()` API, so it would stay generic and usable by other teams as well. I’m not fully sold on diverging from that idea, tbh, especially if you want to add this next to the existing function.

docs/design/certificate-derived-attestation-expiry.md
```diff
@@ -0,0 +1,185 @@
...
+`MAX` of its `nextUpdate`. That tail is the gap this design closes. What `MAX` should be is #3734,
+not this document.
+
+### Node submit-confirmation
+
+The node confirms a submission landed by checking that the stored expiry *changed*
+([`tx_sender.rs:201`](../../crates/node/src/indexer/tx_sender.rs)), which works only because the
+expiry is always `now + 7 days`. Once the certificate value binds and the collateral has not changed,
+re-submitting stamps the same expiry, so accepted and rejected become indistinguishable in contract
+state — no state-based rule can separate them.
```
pbeza, line 93: Today, every time the contract accepts a submission it overwrites the stored `expiry_timestamp_seconds` with `current block time + 7 days`, so each hourly resubmission writes a larger value, and you can watch the stored number increase hour by hour.

With certificate-derived expiry, the value written is the collateral's own `nextUpdate`: a fixed calendar date (say, _Sep 12_). A node resubmits hourly against the same collateral for weeks, each submission is accepted, and each one writes _Sep 12_ again. The stored number only jumps when the node picks up refreshed Intel collateral, roughly monthly. So "is the stored value changing?" stops telling you whether submissions are being accepted.

Having to implement https://github.com/near/mpc/issues/4301 would add another layer of complexity here. I’d personally prefer if we could look for ways to simplify this instead, because we’ve been gradually adding more and more special-case logic and stacking layers of complexity, and I’m a bit worried that it could become harder to reason about and maintain over time, and potentially lead to hacks/bugs.

pbeza (commented):
Left a couple of questions.

I found some sections a little hard to follow and a bit LLM-ish in places, but I think the tl;dr is roughly this:

We have two types of expirations:

* **Intel's expirations** (`TCB Info` / `QE Identity` / CRL `nextUpdate`, roughly 30 days; certificate `notAfter`, usually years): checked once inside `tee-verifier` at the submission block time, then discarded. None of these values currently reach the chain state.
* **The contract's expiration (7 days)**: created when the submission is accepted, stored as a single `u64`, and later compared against block time. This is currently the only expiry that exists on-chain.

So, as I understand it, the main design question in this PR is whether to make the second value a function of the first one, e.g. the minimum of Intel's `nextUpdate` values, roughly 30 days minus the collateral age, instead of using a constant.

The other part is dealing with the places that implicitly relied on the current behavior, where the expiry moves forward on every accepted submission. That’s where most of my [concerns](https://github.com/near/mpc/pull/4292#discussion_r3913055729) and questions come from.

docs/design/certificate-derived-attestation-expiry.md
barakeinav1: sure, I'll add it.
note that I plan to use dcap-qvl implementation for this.

docs/design/certificate-derived-attestation-expiry.md
barakeinav1: >Why do you still want to keep MAX_ATTESTATION_VALIDITY_SECONDS

This is the main decision we should make, do we want to add a `MAX_ATTESTATION_VALIDITY_SECONDS ` above what intel decides (e.g extra security) or not. 

also - there are some implementation details (like launcher expiry) that we will need to hand if we remove this cap.

> potentially down to just a few seconds in corner cases
how? we only shorted to MAX_ATTESTATION_VALIDITY_SECONDS

docs/design/certificate-derived-attestation-expiry.md
barakeinav1: I really don't want to add certificate parsing logic in our contract, so the tee-verify is the natural place to have it. the new method is basically verify() that also returns some additional information. 
adding a new method (instead of extending the old one) is just for easy upgrade path. 
I'm open to suggestions

docs/design/certificate-derived-attestation-expiry.md
pbeza: > also - there are some implementation details (like launcher expiry) that we will need to hand if we remove this cap.

Yeah, that’s exactly my concern when I [talk](https://github.com/near/mpc/pull/4292#discussion_r3913055729) about the complexity of our codebase, and why I’d prefer not to make it even more complex with [this](https://github.com/near/mpc/issues/4301). There are expiry times that are tightly coupled, and if you modify one without being aware of the others, you can easily break things. That’s not ideal, and I’d love for us to simplify this over time.

docs/design/certificate-derived-attestation-expiry.md
barakeinav1: https://github.com/near/mpc/issues/4301 - I agree that seems complex, but the node's ability to check transaction  result based on the receipt should be a very understandable building block.

I listed an alternative that we add the submission time to the attestation. 
This is probably simpler.

docs/design/certificate-derived-attestation-expiry.md
netrome: I'm also curious about this. I'd also expect to drop our own notion of expiry and use the one from the attestation (e.g. no "max" here). If there are no strong reasons for keeping our own MAX_ATTESTATION_VALIDITY_SECONDS parameter I'd suggest dropping it to keep the system simpler.

docs/design/certificate-derived-attestation-expiry.md
```diff
@@ -0,0 +1,185 @@
...
+`mpc-contract` threads that value from `resolve_verification` into `verify_and_store_dstack` and on
+into `AcceptedAttestation::dstack`, which applies the `min`. `ValidatedDstackAttestation`'s layout is
+unchanged, so **no state migration**. `verify_locally` (node, CLI, `tee-authority`) reads the same
+number from `claims()`, so off-chain verification agrees with the chain. The mock path keeps stamping
+`now + MAX`; [#4005](https://github.com/near/mpc/issues/4005) stays open.
+
+### Why the cap stays
+
+- `Config::validate` requires `launcher_hash_unused_ttl_seconds >= DEFAULT_EXPIRATION_DURATION_SECONDS`
+  (14 days ≥ 7): a launcher hash backing a still-valid attestation must not be evicted first.
```
netrome: Why is this important? Evicting a launcher is orthogonal to evicting a node imo. I don't see why they need to be entangled.

docs/design/certificate-derived-attestation-expiry.md
```diff
@@ -0,0 +1,185 @@
...
+The one choice here that changes the security argument rather than the mechanism.
+
+**For pure certificate-derived:** it is the only expiry Intel vouches for; every shorter window is
+our own invention, and a self-chosen number is what caused the fleet-wide outage on 2026-07-23, when
+the node's then 7-day `MAX_COLLATERAL_AGE` rejected a valid 30-day PCK CRL and no node could attest
+([#3946](https://github.com/near/mpc/issues/3946)). One freshness authority instead of two that can
+disagree.
+
+**For the cap:** Intel holds one CRL for most of its 30-day window, so pure certificate-derived is
+~30 days in the common case — four times today's window, too long for #3734's containment, and it
```
netrome: I'm not following the justification for 3734. I guess this comes down to a value about "shorter is better", which I can understand. But in our current phase I would prefer a more liberal attestation window rather than risking aggressively evicting nodes. Having 30 days doesn't sound like a no-go to me, at least for now. And it would be nice to keep our system simple.

netrome (dismissed):
I generally support using the expirations from the certificates. However, I'm agree with the concerns raised by @pbeza. It seems like keeping the system purely certificate driven would make it much simpler, which would be a strong advantage while we're rolling out the TEE integration. I think that is a good starting point, and I don't currently see a strong justification for adding extra constraints on top of that, at least initially.

Would love to hear more your thoughts on what benefits we get from keeping our own cap in addition to the certificate value.

docs/design/certificate-derived-attestation-expiry.md
barakeinav1: there is some technical dependency between MAX_ATTESTATION_VALIDITY_SECONDS  and launcher expiry (launcher_hash_unused_ttl_seconds ) - I need to look into it to see if removing MAX_ATTESTATION_VALIDITY_SECONDS  causes any problem.

barakeinav1:
> I generally support using the expirations from the certificates. However, I'm agree with the concerns raised by @pbeza. It seems like keeping the system purely certificate driven would make it much simpler, which would be a strong advantage while we're rolling out the TEE integration. I think that is a good starting point, and I don't currently see a strong justification for adding extra constraints on top of that, at least initially.
> 
> Would love to hear more your thoughts on what benefits we get from keeping our own cap in addition to the certificate value.

Our own cap is only if we want to force short attestation times than what Intel allows. Not sure we actually want to do this.
The more I think about this, the more I think we should go with certificate time only. 
I need to make sure that check that attestations landed have a simple solution, since @pbeza was worried about reading recipients to detect this

docs/design/certificate-derived-attestation-expiry.md
barakeinav1: >I'm not following the justification for 3734

the reasoning for that is wrong, #3734 - shortens expiry only on verifier contract rotation, I don't see how it will affect this PR.  (I think I was concerned about some dependency dependency between them, but I can't think of something now.

>but in our current phase I would prefer a more liberal attestation window rather than risking aggressively evicting nodes.

totally agree here (lessoned learned the hard way)

barakeinav1:
Updated the doc to go certificate-only — `MAX_ATTESTATION_VALIDITY_SECONDS` is gone, so there's a single expiry in the system now. You were both right that the cap wasn't earning its place: my #3734 argument was simply wrong (it only shortens on verifier rotation), and the cap didn't actually help the case I was worried about anyway.

@pbeza on the verifier API — I think this stays 1:1 rather than diverging: `QuoteClaims` is `dcap-qvl`'s own type, returned by `verify_with_policy`, so the new method just exposes more of the upstream surface. And you were right about short expiries, I misread my own formula there; `min` picks the certificate value when it's smaller, so it can land seconds out. That's now item 4, handled node-side by refusing to submit near-expiry collateral.

On confirming submissions, I dropped #4301 in favour of storing the submission timestamp on the attestation entry — much less machinery than reading receipts, and it restores exactly today's "the number goes up every time" semantics.

The doc now lists the six things that depended on the old constant (submission timestamp, launcher TTL, rotation shortening, near-expiry collateral, the sandbox fixture, and the gas config vote as its own step). That list is the real work, so worth a look.

docs/design/certificate-derived-attestation-expiry.md
barakeinav1: They're entangled because `re_verify` re-checks a stored attestation's launcher hash against the current allowed set, so evicting the hash kicks a node whose attestation is still perfectly valid. The doc now drops that TTL invariant and instead skips eviction for any hash still referenced by a current participant's non-expired attestation, which is what the invariant was approximating anyway.

docs/design/certificate-derived-attestation-expiry.md
barakeinav1: Went with your preference here — the design now stores the submission timestamp on the entry instead of reading receipts, so #4301 ends up removing the expiry-delta heuristic rather than stacking another layer on top of it. Dropping `MAX_ATTESTATION_VALIDITY_SECONDS` also takes out one of the coupled expiry times you were worried about.

docs/design/certificate-derived-attestation-expiry.md
```diff
@@ -0,0 +1,188 @@
...
+| `verify_quote` | **175.81** | 200 | `verifier_tera_gas` |
+| `resolve_verification` | 4.60 | 60 | `resolve_verification_tera_gas` |
+| Total burnt, incl. tx and refunds | 199.73 | | |
+
+The 300 is the chain's budget, not the verifier's: the submit receipt spends 16.3 itself and commits
+200 + 60 to its two promises, leaving only ~20 TGas unallocated. So `verifier_tera_gas` can reach
+~220 as things stand, or ~270 if `resolve_verification`'s 60 is trimmed toward its 4.6 — keeping
+worst-case room there. Not 300.
+
+`claims()` re-parses the two JSON documents and parses the certificate chains, so measure it before
```
netrome: `claims()` comes a bit out of the blue here. I asked Claude and it thinks we're refferring to `QuoteVerificationResult::claims()` - which would be good to clarify here.

docs/design/certificate-derived-attestation-expiry.md
```diff
@@ -0,0 +1,188 @@
...
+choosing the split. Fallback if it does not fit: read `nextUpdate` from the two CRLs and the two JSON
+documents only. That drops the four certificate chains from the minimum, which is safe given their
+7–30 year lifetimes, but it should be a deliberate choice rather than an accident.
+
+## Rollout
+
+1. **Land item 1** (stored submission timestamp), so confirmation keeps working.
+2. **Vote the gas config** (`propose_update` / `vote_update`). Separate governance from the contract
+   upgrade, and it must come first — otherwise step 4 runs the heavier method under the old budget
+   and every submission runs out of gas.
```
netrome: It's a bit unclear what the proposal here is. Should we just set the constants after the measured values above? Or is the purpose for this step to also define the values. This sentence reads more like a "vote for the proposed gas config" but the doc doesn't clearly propose one.

Note, we don't need to propose anything here, just make it clear that we need to propose and vote for one at this point.

netrome (dismissed):
Thanks for updating. The general idea (use the intel-provided expiry and drop our 7-day expiry) makes sense to me. However, I find this document quite hard to read so it could be worth giving it a readability pass.

barakeinav1:
Agreed, it needed it. Did a readability pass in d2aca16b — shorter sentences, a before/after table up top, and the rationale trimmed down and moved inline per section. Shout if it's still heavy.

docs/design/certificate-derived-attestation-expiry.md
barakeinav1: Added an explanation in d2aca16b — you were right, it's `QuoteVerificationResult::claims()`, reached through `verify_with_policy`.

docs/design/certificate-derived-attestation-expiry.md
barakeinav1: Reworded in d2aca16b — the step is now "measure `claims()`, then propose and vote the gas config", and it says outright that the numbers come from the measurement rather than from this doc.

docs/design/certificate-derived-attestation-expiry.md
barakeinav1: see explanation here https://github.com/near/mpc/pull/4292/changes#diff-0cebf97dec83cdcd64846c8b2d8107c10f015805fdbaa64cdf54b0bbb6f17dafR145

netrome (approved):
Thanks for updating 🙏

docs/design/certificate-derived-attestation-expiry.md
```diff
@@ -0,0 +1,249 @@
...
+) -> VerificationResultWithClaims;
+
+pub enum VerificationResultWithClaims {
+    Verified {
+        report: VerifiedReport,
+        earliest_expiration_seconds: u64,
+    },
+    Rejected(VerifierError),
+}
+```
```
pbeza, line 105: I’m not 100% convinced we should add another method alongside the existing `verify_quote`. I see you’ve already opened a PR implementing this design doc (https://github.com/near/mpc/pull/4500), so I’m fine approving this PR as-is since you’ve addressed most of my concerns. I’ll leave the remaining concerns for https://github.com/near/mpc/pull/4500.

pbeza (approved):

docs/design/certificate-derived-attestation-expiry.md
barakeinav1: https://github.com/near/mpc/pull/4500 - is just a pre-step needed for the full change so we have time.

barakeinav1: https://github.com/near/mpc/issues/1639 - is the certificate implementation - I haven't started yet.
Do you have an alternative for the new method? we can the same method but externt for additional parameter, but I think this will complicate the migration to it.

or did you think about somthing else?

pbeza: I’d just like to keep the verifier contract generic enough to be useful for anyone who wants to verify TDX attestations, rather than making it MPC-specific. It would also simplify things when it comes to redeploying the verifier contract.

Maybe we could leave the verifier contract as-is and parse the relevant attestation certificates in the MPC contract if we need to extract any dates from them. That would probably bloat the MPC contract with SSL certificate parsing libraries though, so it might not be a great option.

I’m not sure yet: I’d need to look more closely at https://github.com/near/mpc/pull/4500 and the follow-ups. For now, you can assume I’m fine with the design doc, but I’d still prefer having a single method in the verifier contract that’s generic enough for anyone to use, rather than two methods where one is tailored to MPC.

barakeinav1: >Maybe we could leave the verifier contract as-is and parse the relevant attestation certificates in the MPC contract if we need to extract any dates from them. That would probably bloat the MPC contract with SSL certificate parsing libraries though, so it might not be a great option.

I don't think this is a good option - this means a lot of code duplication, and as you said it will bloat the MPC contract.

>I’d just like to keep the verifier contract generic enough to be useful for anyone who wants to verify TDX attestations, rather than making it MPC-specific. 

I don't think this is MPC specific, it just exposes an an already generic existing capability of the verifier contract, which is basically "verify but also return the expiry result".

pbeza: Feel free to move ahead with the implementation. We’re mostly into implementation details now anyway, so it’ll be easier to discuss with the code in front of us. High-level design LGTM; just loop me in on the follow-up PRs.

PR #4293, opened 2026-08-31
refactor: move tee vote types in the contract public API to dtos crate

### LINKED ISSUE #2480: refactor: Move voting types from contract crate to near-mpc-contract-interface
### Background
refactor only, not functional change

  Several types that appear in the contract's public API currently live in the contract crate
  (crates/contract/src/tee/) rather than near-mpc-contract-interface. External callers who deserialize contract
  responses need these types but can't import them without depending on the full contract crate.

  Types to move:
  - ContractExpectedMeasurements, Sha384Digest, MeasurementVoteAction, MeasurementVotes (from tee/measurements.rs)
  - LauncherHashVotes, LauncherVoteAction (from tee/proposal.rs, added in #2343)
  - CodeHashesVotes (from tee/proposal.rs, added in #376)

  These are used in public contract methods: vote_add_os_measurement, vote_remove_os_measurement,
  allowed_os_measurements, os_measurement_votes, vote_add_launcher_hash, vote_remove_launcher_hash,
  launcher_hash_votes, code_hash_votes.

  Not a breaking change — on-chain ABI stays the same, only Rust import paths change.

  Related: #2447, #2343 #2449

### User Story

as a developer I'd like to have a clear interface.

### Acceptance Criteria

refactor completed.

_No response_

### LINKED ISSUE #3425: Dedupe (or drift-guard) the parallel app-compose AppCompose DTO
## Background

`near-mpc-contract-interface` defines a second, hand-maintained `AppCompose` (`crates/near-mpc-contract-interface/src/types/attestation.rs`) that mirrors the verified `attestation::AppCompose`. It is **not** on the verification path, but it is re-exported as public API. It already drifted once: the `init_script` security field was added to the verified struct but not this copy (synced in #3414).

After #3420 the two structs intentionally diverge — the verified struct models the *full* dstack field set for `#[serde(deny_unknown_fields)]`, while this DTO is the security-relevant *subset* — so a mechanical "keep identical" test does not fit and would give false confidence.

## What / Why

Decide how to remove the drift risk permanently. Preferred: **dedupe** — delete this copy and re-export `attestation::AppCompose` so there is a single source of truth. Pending a team decision on external consumers of the DTO's exact shape.

## Acceptance Criteria

- [ ] Team decision: dedupe (re-export the verified struct) vs. keep the mirror.
- [ ] If keeping the mirror: decide whether a drift-guard is feasible/worthwhile given the intentional divergence (note: a subset test would NOT have caught the original `init_script` drift).
- [ ] If deduping: confirm no external consumer relies on the DTO's exact field shape, then remove the copy.

## Resources

- `crates/near-mpc-contract-interface/src/types/attestation.rs` — the DTO `AppCompose`
- `crates/attestation/src/app_compose.rs` — the verified `AppCompose`
- #3413 / #3414 (init_script drift), #3415 / #3420 (fail-closed hardening)

gilcu3:
Closes #2480 and #3425


crates/attestation-types/src/measurements.rs
```diff
@@ -17,7 +17,7 @@ use serde_with::{Bytes, serde_as};
 )]
 #[cfg_attr(feature = "borsh-schema", derive(borsh::BorshSchema))]
 pub struct Measurements {
-    /// MRTD (Measurement of Root of Trust for Data) - identifies the virtual firmware.
+    /// MRTD (Measurement Register for the Trust Domain) - identifies the virtual firmware.
```
gilcu3, line 20: a small broken window fix

crates/near-mpc-contract-interface/src/types/attestation.rs
```diff
@@ -136,11 +140,11 @@ pub struct VerifiedDstackAttestation {
...
-    pub rtmr0: Sha384Digest,
-    pub rtmr1: Sha384Digest,
-    pub rtmr2: Sha384Digest,
-    pub key_provider_event_digest: Sha384Digest,
+    pub mrtd: MrtdHash,
+    pub rtmr0: Rtmr0Hash,
+    pub rtmr1: Rtmr1Hash,
+    pub rtmr2: Rtmr2Hash,
+    pub key_provider_event_digest: KeyProviderEventDigest,
 }
```
gilcu3, line 148: This is now aligned with the `ExpectedMeasurements` type added now in the crate. It was not unified as these two types currently mean different things, but we could just unify them later

crates/attestation-types/src/measurements.rs
anodar: Nice!

anodar (approved):

haiyuechen-nearone (approved):
:feelsgood:

PR #4296, opened 2026-08-31
feat: support {} host-label wildcard in on-chain base_url

anodar:
This is needed for quicknode.

Work towards: https://github.com/near/mpc/issues/4286

### DIFF
--- a/crates/node/src/foreign_chain_whitelist_verifier.rs
+++ b/crates/node/src/foreign_chain_whitelist_verifier.rs
@@ -1094,7 +1094,6 @@ mod tests {
             "https://evil.com?x=.abstract-testnet.quiknode.pro",
             WILDCARD
         ));
-        // An empty label, and a suffix that is not where the host ends.
         assert!(!base_url_matches(
             "https://.abstract-testnet.quiknode.pro/",
             WILDCARD
--- a/crates/node/src/foreign_chain_whitelist_verifier.rs
+++ b/crates/node/src/foreign_chain_whitelist_verifier.rs
@@ -190,6 +190,8 @@ fn compare_provider(
 /// The first `{}` in `base` matches one non-empty run of `[A-Za-z0-9-]` — a single host
 /// label in the bases we use, where a provider puts a per-operator slug in the hostname
 /// (e.g. QuickNode's `https://{}.sui-testnet.quiknode.pro`).
+/// The character after `{}` must not itself be a label character:
+/// a base like `https://api-{}-v2.example.com` never matches.
 fn base_url_matches(local: &str, base: &str) -> bool {
     let l = local.trim_end_matches('/');
     let b = base.trim_end_matches('/');
--- a/crates/node/src/foreign_chain_whitelist_verifier.rs
+++ b/crates/node/src/foreign_chain_whitelist_verifier.rs
@@ -189,10 +189,7 @@ fn compare_provider(
 ///
 /// The first `{}` in `base` matches one non-empty run of `[A-Za-z0-9-]` — a single host
 /// label in the bases we use, where a provider puts a per-operator slug in the hostname
-/// (e.g. QuickNode's `https://{}.sui-testnet.quiknode.pro`). The run is consumed greedily
-/// with no backtracking, so a base whose `{}` is directly followed by another label
-/// character (e.g. `https://api-{}-v2.example.com`) never matches; a second `{}` is
-/// matched literally.
+/// (e.g. QuickNode's `https://{}.sui-testnet.quiknode.pro`).
 fn base_url_matches(local: &str, base: &str) -> bool {
     let l = local.trim_end_matches('/');
     let b = base.trim_end_matches('/');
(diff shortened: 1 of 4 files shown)


anodar:
> docs/foreign-chain-transactions.md:478 — this PR updated the design doc's URL-assembly statements (lines 111 and 119) but left the equivalent statements in the primary doc contradicting the new behavior...

> docs/foreign-chain-transactions.md:596 — the "Out of scope / deferred" bullet reads "Hostname templating for Alchemy / Infura …

Addressed both.


> crates/node/src/foreign_chain_whitelist_verifier.rs:197 — split_once("{}") makes only the first {} a wildcard; any later occurrence is matched literally, so a base like https://{}.{}.quiknode.pro can never match a real URL.

> crates/node/src/foreign_chain_whitelist_verifier.rs:203 — the label run is maximal-munch with no backtracking, so a {} immediately followed by a label character in the base can never match.

This WAI

> crates/node/src/foreign_chain_whitelist_verifier.rs:1066 — test gaps worth closing while this is fresh: (a) no compare__should_… test exercising a wildcard base end-to-end, unlike every other matcher behavior in this file

Part of followup PR.

> crates/node/src/foreign_chain_whitelist_verifier.rs:1093 — // An empty label, and a suffix that is not where the host ends. labels the three asserts that follow without adding anything they do not already say

Addressed.

docs/foreign-chain-transactions.md
```diff
@@ -475,7 +475,7 @@ pub struct ForeignChainRpcWhitelist {
 
 ### Example URL assembly
 
-The node assembles the final URL deterministically: start from `base_url`, apply `chain_routing` (no-op for `Embedded`, append segment for `PathSegment`, merge param for `QueryParam`), then apply `auth_scheme` to inject the token.
+The node assembles the final URL deterministically: start from `base_url`, fill a `{}` (if present) with the operator-specific host label (`SLUG` in the table below), apply `chain_routing` (no-op for `Embedded`, append segment for `PathSegment`, merge param for `QueryParam`), then apply `auth_scheme` to inject the token.
```
barakeinav1, line 478: just to check I follow the direction - does the operator eventually stop supplying a URL entirely and just give `provider_id` + token + slug, with the node assembling it? that's what this line describes. or do they keep writing the full `rpc_url` and enforcement just means we drop it when it doesn't match the whitelist?

asking because under the first, `base_url_matches` goes away and `{}` becomes a template rather than a wildcard.

crates/node/src/foreign_chain_whitelist_verifier.rs
```diff
@@ -186,16 +186,33 @@ fn compare_provider(
...
+        return starts_with_at_boundary(l, b);
+    };
+    let Some(rest) = l.strip_prefix(prefix) else {
         return false;
     };
-    rest.starts_with('/') || rest.starts_with('?') || rest.starts_with('#')
+    let label_len = rest
+        .find(|c: char| !c.is_ascii_alphanumeric() && c != '-')
+        .unwrap_or(rest.len());
+    label_len > 0 && starts_with_at_boundary(&rest[label_len..], suffix)
```
barakeinav1, line 208: nit (found by Claude): `{}` only works when the character right after it isn't a label character. the match here is greedy over `[A-Za-z0-9-]` with no backtracking, so:

```
base   https://api-{}-v2.example.com
local  https://api-acme-v2.example.com
```

the label eats `acme-v2`, leaving `.example.com` to match against `-v2.example.com` - no match, silently. fine for our `{}.host` bases, but worth saying so in the doc.

docs/foreign-chain-transactions.md
anodar: > just to check I follow the direction - does the operator eventually stop supplying a URL entirely and just give provider_id + token + slug, with the node assembling it?

They could and yes we could use just provider_id, but it could just also stay as is that they provider full URL, which is how it is right now.

> that's what this line describes. 

They currently provide full URL, but what is used in the node (once enforcement is implemented) is still deterministic and they can't just have any arbitrary thing in {}. We still validate that it matches whitelisted base URL.

crates/node/src/foreign_chain_whitelist_verifier.rs
anodar: Added comment.

barakeinav1 (approved):
Thanks for splitting this out, much easier to review on its own. Approved.

crates/node/src/foreign_chain_whitelist_verifier.rs
```diff
@@ -186,16 +186,35 @@ fn compare_provider(
...
```
gilcu3, line 208: nit: we could avoid the slicing by using a slightly different code with something like `split_at`

gilcu3 (approved):
Thank you!

crates/node/src/foreign_chain_whitelist_verifier.rs
anodar: @gilcu3  Changed to split_at in https://github.com/near/mpc/pull/4287

PR #4307, opened 2026-08-31
fix: flaky per_domain_reconstruction_threshold test

anodar:
By separating timeouts when it's expected to sign (lower timeout) and when it's not (higher).

Flaky run: https://github.com/near/mpc/actions/runs/33424884416/job/99596207874?pr=4287

Seems like 10sec isn't enough here.


gilcu3:
@anodar notice the failure was:

```
    [WARN  mpc_node::providers::robust_ecdsa::presign] Can't choose active participants for a robust-ecdsa presignature: Not enough active participants: need 5, got 4. Sleeping.
    [INFO  mpc_node::tests] Timed out waiting for signature response for user user_4_high
```

which is something that had hit us in the past already where robust-ecdsa presignatures take too long.

PR #4311, opened 2026-09-01
fix(node): check participant sets match during resharing and keygen

### LINKED ISSUE #4231: feat(node): validate participant set of resharing computations
### Background

Currently, the resharing computations accepts participant sets blindly from the leader initiating these computations. This means that a leader can coordinate resharing computations with arbitrary subsets of the intended participants for a resharing. A set of malicious nodes could use this to coordinate a split-view attack on the network.

We should ensure each party validates the participant set before participating in any resharing computations.

### User Story

As an MPC operator I don't want my node to participate in invalid or malicious resharing computations.

### Acceptance Criteria

MPC nodes validate that the participant set of resharing computations corresponds to the full expected set in the contract state before participating in them.

_No response_

gilcu3:
Closes #4231

Notice the PR was extended slightly to cover the same problem for the keygen path.

### DIFF
--- a/crates/node/src/coordinator.rs
+++ b/crates/node/src/coordinator.rs
@@ -342,12 +342,14 @@ where
         let (sender, receiver) = new_tls_mesh_network(&mpc_config, p2p_key).await?;
         let (network_client, channel_receiver, _handle) =
             run_network_client(Arc::new(sender), Box::new(receiver));
+        let participants = Arc::new(mpc_config.participants.clone());
         if mpc_config.is_leader_for_key_event() {
             keygen_leader(
                 network_client,
                 keyshare_storage,
                 key_event_receiver,
                 chain_txn_sender,
+                participants,
             )
             .await?;
         } else {
@@ -356,6 +358,7 @@ where
                 keyshare_storage,
                 key_event_receiver,
                 chain_txn_sender,
+                participants,
             )
             .await?;
         }
@@ -832,6 +835,7 @@ where
             existing_keyshares,
             old_reconstruction_thresholds,
             old_participants: current_running_state.participants,
+            new_participants: mpc_config.participants.clone(),
         });
 
         if mpc_config.is_leader_for_key_event() {
--- a/crates/node/src/coordinator.rs
+++ b/crates/node/src/coordinator.rs
@@ -342,14 +342,14 @@ where
         let (sender, receiver) = new_tls_mesh_network(&mpc_config, p2p_key).await?;
         let (network_client, channel_receiver, _handle) =
             run_network_client(Arc::new(sender), Box::new(receiver));
-        let participants = Arc::new(mpc_config.participants.clone());
+        let expected_participant_ids = mpc_config.participants.participant_id_set();
         if mpc_config.is_leader_for_key_event() {
             keygen_leader(
                 network_client,
                 keyshare_storage,
                 key_event_receiver,
                 chain_txn_sender,
-                participants,
+                expected_participant_ids,
             )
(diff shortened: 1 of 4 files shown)


netrome (dismissed):
Thanks!

crates/node/src/coordinator.rs
```diff
@@ -342,12 +342,14 @@ where
...
         let (network_client, channel_receiver, _handle) =
             run_network_client(Arc::new(sender), Box::new(receiver));
+        let participants = Arc::new(mpc_config.participants.clone());
         if mpc_config.is_leader_for_key_event() {
             keygen_leader(
                 network_client,
                 keyshare_storage,
                 key_event_receiver,
                 chain_txn_sender,
+                participants,
```
kevindeforth: Why do we need to pass an `Arc` here?

crates/node/src/key_events.rs
```diff
@@ -135,7 +136,14 @@ async fn keygen_computation(
     keyshare_storage: Arc<RwLock<KeyshareStorage>>,
     chain_txn_sender: impl TransactionSender,
     key_id: KeyEventId,
+    participants: Arc<ParticipantsConfig>,
```
kevindeforth: We don't really need the full `ParticipantsConfig` here, but just the participant ids.
Why does it need to be an `Arc`?

nit: I would rename this to `expected_participants` / `expected_participant_ids`.

crates/node/src/key_events.rs
```diff
@@ -126,6 +126,7 @@ pub async fn keygen_computation_inner(
 }
 
 /// Wrapper around `keygen_computation_inner` which
+///  - Rejects computations whose participant set does not match the contract.
```
kevindeforth: Hmm, this assumes that `participants` comes from the contract, which this function can not guarantee. I would prefer to phrase it in the lines of: _"compares the participants in the the `NetworkTaskChannel` with the expected participant ids"_

crates/node/src/key_events.rs
```diff
@@ -928,6 +1094,40 @@ mod tests {
...
+            local_encryption_key: [0u8; 16],
+            gcp: None,
+        }
+        .create()
+        .await
+        .unwrap();
+        Arc::new(RwLock::new(keyshare_storage))
+    }
+
+    fn make_participants(ids: &[u32]) -> ParticipantsConfig {
```
kevindeforth: Did you check if we have existing helper methods for this?

kevindeforth (dismissed):
Thanks, this looks good.
Left a few questions.

crates/node/src/coordinator.rs
netrome: There's no need, but we can't pass a plain reference so the alternative would be to pass a full clone of the participants (or just the IDs as you noted).

gilcu3: yeah just avoiding cloning

crates/node/src/key_events.rs
gilcu3: not in this file, but we should have it elsewhere, will take a look

gilcu3: the function was removed due to the other changes, and the equivalent one did not exist, so left there

crates/node/src/coordinator.rs
gilcu3: changed to take the expected set of participant ids

kevindeforth (approved):
Thanks!

netrome (approved):
Thanks for updating 🙏

PR #4312, opened 2026-09-01
fix(node): auto-recover from the epoch-sync wedge (#3909)

### LINKED ISSUE #3909: Stale node (offline ~days) wedges in neard epoch-sync loop; never reaches head

barakeinav1:
Fixes #3909.

## Problem

A node that's been offline long enough to fall past the epoch-sync horizon gets stuck forever with a frozen head. nearcore validates the epoch-sync proof and asks to reset its data, but the request is dropped: mpc-node starts neard via `nearcore::start_with_config`, which hard-codes `shutdown_signal = None`, so nothing acts on the reset and it loops every ~60s.

The `neard` binary avoids this in its CLI wrapper (wire the signal → write a marker → restart → wipe on next start). We embed the nearcore library directly and skipped that orchestration.

Thanks to @metalurgical for finding the root cause and fix direction on #3909.

## Fix

In the node's neard-start path (`crates/node/src/indexer/real.rs`):
- start neard via `start_with_config_and_synchronization` with a shutdown channel;
- on `EpochSyncDataReset`, record a marker and exit non-zero — the mpc-node container's `restart: on-failure` restarts it;
- on the next startup, before neard opens the store, wipe the store if the marker is present (reusing the existing near-data wipe).

Recovery is now automatic — the same thing a manual `wipe_near_data_token` bump does today, but hands-off. Only the chain store (`/data/data`) is wiped; keyshares, secrets, and config survive.

## Tests

- **Unit tests** for the recovery mechanism: the marker-triggered wipe removes only the chain store and preserves keyshare-like siblings; no-op without a marker, archival skip, path guard.
- **Live testnet e2e — passed.** A real TDX node running this build, stopped ~32h (~4.7 epochs behind), was restarted and self-healed with no manual intervention:
  ```
  indexer::real:           nearcore requested an epoch-sync data reset; recording marker and restarting …
  indexer::near_data_wipe: wiped nearcore data dir (epoch-sync data reset) hot_store_path="/data/data"
  ```
  → fresh state-sync back to head. (Stock build under the same conditions stays wedged: head frozen, `near_sync_status=2`, data-reset loop forever.)
- Full automated epoch-sync integration test tracked as a follow-up in #4314.


barakeinav1:
Test coverage:
- **Unit tests** (in this PR): marker record + wipe-on-restart + keyshare-preservation + path-guard on the reset path, and the reset marker added to the `home_paths` disjoint-from-wipe-target tripwire.
- **Live testnet e2e** (in progress): a real TDX node stopped past the epoch-sync horizon, verifying it auto-recovers (wipe → fresh state-sync → head) with no manual `wipe_near_data_token` bump. Will post the result here and mark ready for review.
- **Full automated epoch-sync integration test** (real neard driven into epoch-sync → auto-recovery) is tracked as a follow-up in #4314 — it needs sandbox genesis tuning to force epoch-sync plus exit/restart orchestration the bare-process harness doesn't have today.

crates/node/src/indexer/near_data_wipe.rs
```diff
@@ -67,15 +129,29 @@ pub(crate) fn wipe_near_data_if_requested(
...
+    let trash_path = near_data_trash_dir(home_dir);
+    remove_trash(&trash_path);
+    // One atomic rename, so the store is gone for good even if the process dies
+    // before the (best-effort) delete below runs.
+    match std::fs::rename(hot_store_path, &trash_path) {
+        Ok(()) => {}
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => return Ok(false),
+        Err(err) => return Err(err),
+    }
+    remove_trash(&trash_path);
```
metalurgical, line 156: There's two calls to `remove_trash` here, one seems redundant.

Since the marker is removed only after successfully deleting the trash_path, the rename may not be necessary here.

Probably a good idea to put the marker removal code in this function as well, so it cannot be accidentally called in isolation in future refactors/changes?

metalurgical:
@barakeinav1 Thanks for the mention :)

Looks like a format needs to be run on the toml files for the fast checks.

crates/node/src/indexer/near_data_wipe.rs
```diff
@@ -9,10 +9,72 @@
...
+
+/// If nearcore requested an epoch-sync data reset on a previous run, wipe the
+/// chain store now — before nearcore reopens it — and clear the marker. Mirrors
+/// `neard`'s `check_epoch_sync_data_reset_marker`. Reuses the same atomic
+/// rename-into-trash the token wipe uses. Skipped, with a warning, on archival
+/// nodes. Keyshares/secrets/config are untouched (only the store dir is moved).
+pub(crate) fn wipe_near_data_if_epoch_sync_reset(
+    home_dir: &Path,
+    hot_store_path: &Path,
+    is_archival: bool,
```
metalurgical, line 32: `is_archival` parameter could probably be removed.  Would it be practical to run an archival node for mpc?

barakeinav1: this path is probably 99% not relevant, but I following example from other places in our code to be on the safe side.

barakeinav1:
@/tmp/reply_claudebot.md

crates/node/src/indexer/near_data_wipe.rs
```diff
@@ -43,18 +105,18 @@ pub(crate) fn wipe_near_data_if_requested(
...
-
+    // Record the token before wiping: if the wipe fails it is not retried
+    // automatically, so a recorded token avoids a wipe loop on every restart.
     write_last_token(&token_path, requested_token)?;
 
-    // Move the store aside in one atomic rename, so the data dir is gone for good
-    // even if the process dies before the (best-effort) delete below runs.
-    match std::fs::rename(hot_store_path, &trash_path) {
-        Ok(()) => {}
+    match wipe_store_dir(home_dir, hot_store_path) {
```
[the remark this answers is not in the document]
barakeinav1: Agree — `ensure_within_home` now runs before `write_last_token`, so a misconfigured `store.path` no longer records the token and the wipe still retries after the path is fixed. Fixed in 4b3d4423.

crates/node/src/indexer/near_data_wipe.rs
```diff
@@ -9,10 +9,72 @@
...
+            ?marker,
+            ?err,
+            "could not remove epoch-sync data-reset marker"
+        );
+    }
+    tracing::info!(
+        ?hot_store_path,
+        "wiped nearcore data dir (epoch-sync data reset)"
+    );
+    Ok(())
```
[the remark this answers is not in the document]
barakeinav1: Agree — now only logs "wiped" when a store was actually removed (uses the `wipe_store_dir` bool). Fixed in 4b3d4423.

crates/node/src/indexer/real.rs
```diff
@@ -51,6 +55,36 @@ pub async fn check_block_processing(process_blocks_sender: watch::Sender<bool>,
     }
 }
 
+/// Wires nearcore's shutdown signal so an [`ShutdownReason::EpochSyncDataReset`]
+/// is acted on instead of dropped: a stale node that would otherwise loop forever
+/// in epoch-sync (#3909) records the reset marker and exits non-zero, so the
+/// container's `restart: on-failure` restarts it and the startup wipe runs. Returns
+/// the sender to pass to [`nearcore::start_with_config_and_synchronization`].
```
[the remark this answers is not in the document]
barakeinav1: Agree — reworded to "so that a". Fixed in 4b3d4423.

crates/node/src/indexer/near_data_wipe.rs
barakeinav1: Thanks for the review! Handled all three:
- The double `remove_trash` on the token path now has a comment — the top-level call only cleans leftover trash on startups where *no* wipe is requested; `wipe_store_dir` re-cleans it on the wipe path itself.
- I kept the rename on purpose: it's what makes the wipe atomic. The store vanishes in a single `rename` into the trash dir, so a crash between the rename and the delete can never leave a half-deleted store to boot on.
- On coupling the marker removal into the wipe function: since `wipe_store_dir` is now shared by both the token wipe and the epoch-sync wipe, I kept the marker handling in `wipe_near_data_if_epoch_sync_reset` (rather than the shared helper, so it isn't coupled to one caller), but factored out a small `remove_marker` helper so it isn't duplicated. Happy to revisit if you'd prefer it grouped differently.

All in 4b3d4423.

crates/node/src/indexer/near_data_wipe.rs
barakeinav1: I've kept `is_archival` for now — it mirrors the existing `wipe_near_data_token` wipe and neard's own archival guard, and the related archival crash-loop is fixed in 4b3d4423 (the handler no longer exits on an archive).

barakeinav1:
@claude review

barakeinav1:
@copilot review

barakeinav1:
All addressed in 4b3d4423.

**Blocking**

> `crates/node/Cargo.toml:64` — `nearcore` is inserted between `near-o11y` and `near-sdk` … `cargo sort … will fail`

Agree — moved below the `near-*` keys.

> `crates/node/src/indexer/real.rs:76-79` — the catch-all arm exits with code **0** … takes the MPC node permanently offline

Agree — `ShutdownReason` is matched explicitly now; `ExpectedShutdown` is ignored (pre-fix behaviour), and a future variant is a compile error.

> archival nodes crash-loop forever

Agree — the handler refuses to exit on an archive, and the startup wipe clears the stale marker.

> `std::process::exit` from a spawned task bypasses the graceful-shutdown path

Agree — the reset goes through a dedicated channel raced in `run.rs`, so the documented teardown runs before the non-zero exit.

> the refactor moved `ensure_within_home` from *before* `write_last_token` … to after

Agree — `ensure_within_home` runs before `write_last_token` again; added a test.

> `.expect("epoch-sync data-reset wipe failed")` panics with the marker still on disk … crash-loop

Agree — no panic; logs an actionable error and degrades to the wedge.

> Documentation alignment … stale `start_near_node` comment … runbook

Agree — fixed the comment and added an auto-recovery note to the runbook. (`cluster.rs:111` left as-is — not touched by this PR.)

**Non-blocking**

> `(#3909)` inside the runtime log string leaks PR/session context

Agree — dropped from the message.

> `Lagged(_) => continue` silently discards … could include the very `EpochSyncDataReset`

Agree — now warns.

> the reason-to-action decision is welded to `std::process::exit` and unreachable from tests

Agree — routed through the graceful path; the wipe logic is unit-tested.

> the marker is written with a plain `fs::write` … the asymmetry … is surprising

Kept non-atomic intentionally (existence-only, self-heals); added a comment contrasting it with `write_last_token`.

> no attempt counter or backoff

Deferred as a follow-up.

> `remove_trash` now runs twice before the rename on the token path

Agree — added a clarifying comment.

> logs `"wiped nearcore data dir …"` even when `wipe_store_dir` returned `false`

Agree — only logs when a store was actually removed.

barakeinav1:
All addressed in 4ce0836b.

**Blocking**

> a failing wipe produces an unbounded restart loop, not the "pre-fix wedge" the comment promises … Disarm the restart when the wipe could not run

Agree — the startup wipe now records `reset_wipe_failed` and the handler is passed `restart_disabled` so a failing wipe wedges (pre-PR behaviour) instead of exiting into `restart: on-failure` forever. Distinct log lines for the archival vs wipe-failed cases.

> the restart is requested even when the marker could not be written … Make it `-> std::io::Result<()>` and only send the shutdown on `Ok`

Agree — `record_epoch_sync_reset_request` returns `io::Result` and the handler only sends the shutdown on `Ok`; a persistent write failure stays up wedged. Fixed the `record`/`run.rs` doc comments that overstated the self-heal.

> `spawn_epoch_sync_reset_handler` has no test coverage

Agree — extracted the per-signal logic into `handle_shutdown_signal` (returns `ControlFlow`) and added two `#[tokio::test]`s: records marker + requests restart on reset; no marker and no restart when `restart_disabled`.

**Non-blocking**

> the handler does a single `recv().await` rather than looping … A `loop { .. }` … removes the dependency on that invariant.

Agree — now loops, breaking only on a requested restart or a closed channel.

> The live-testnet validation in the PR description looks like it predates the current mechanism … Worth re-running the testnet repro on the current head

Correct — the tested image was built from the `process::exit` version, so the `run.rs` graceful-teardown path is not yet validated on testnet. Will re-run the stale-node repro on this head and update the PR description once the review iteration settles.

> `crates/e2e-tests/src/cluster.rs:111` — still refers to `start_near_node`

Agree — reworded to "node startup blocks until synced".

> the new note says the far-behind case "recovers automatically", but omits the archival carve-out

Agree — the note now names the archival carve-out (never auto-wiped; use the manual wipe there).

> `real.rs:63-66` … documents an alternative that was never in the merged code … `real.rs:91-93` similarly narrates pre-#3909 history

Agree — trimmed both; kept the behavioural "mpc-node does not configure `expected_shutdown`".

> if `remove_marker` fails after a successful wipe it only warns … `tracing::error!` would at least make it greppable

Agree — now `error!`, and the doc says a leftover marker forces a full re-sync.

> `marker.exists()` swallows I/O errors as "no marker"

Left as-is — a marginal read-side case; the write-side silent path it pairs with is now removed by the `record` → `Result` change above.

barakeinav1:
@claude review

barakeinav1:
@claude review

barakeinav1:
Two reviews landed (one on `4ce0836`, one on `1e8cc4a`); addressing both here. All in 3497a8e5.

**Blocking — the reset was inert in the wedge (`run.rs:302` + `real.rs:119`)**

> the reset request is never consumed in the wedge scenario, so the fix is inert exactly where it is needed … `spawn_real_indexer` … ends in `txn_sender_receiver.blocking_recv()` … the main thread stays parked … nothing ever polls `epoch_reset_receiver`

Correct, and verified against the code path (`main.rs:36` `block_on` → `spawn_real_indexer` `blocking_recv()` at `real.rs:448` → only sent after `ensure_head_follows_tip()` returns). Went with option 1: the handler now writes the marker and calls `std::process::exit` directly. Safe here for the reason given — the reset only fires during initial sync, before `SecretDB` is opened (`run.rs`, `create_root_future`, after `spawn_real_indexer` returns), and nearcore's store is about to be wiped — so there's nothing to flush. Dropped the `mpsc` channel + `run.rs` arm; the doc comment now states the exit and why it's safe. This makes the "recovers automatically" note accurate again.

> the currently reported [testnet] run exercised the `process::exit` build

Right — will re-run the stale-node repro on this head once the review iteration settles.

**Blocking — non-reset arms untested (`real.rs:135-147`)**

> the three non-reset arms … have no test … `Closed → Break` … flipping to `Continue` makes the `while … {}` a busy loop

Agree — `handle_shutdown_signal` now returns a `HandlerAction` and an `#[rstest]` covers the archival, `ExpectedShutdown`, `Lagged` and `Closed` arms; flipping any into the restart path (including `Closed`) now fails a test.

**Nits**

> the wipe-failure error tells the operator to "delete the marker manually" … not a reachable action [inside the CVM] … the config-driven [remedy]

Agree — message now points at "correct the store path and redeploy".

> "Loop rather than a single `recv`" narrates the alternative that was … dropped

Agree — trimmed to the invariant (nearcore may emit more than one reason).

> "mpc-node does not configure `expected_shutdown`" … can be set out of band

Agree — reworded to not claim it's unset; ignoring it is still the point.

> "The caller only requests a restart when this succeeds" … link the callsite

Agree — linked `[crate::indexer::real::handle_shutdown_signal]`.

> Repeated resets are still unbounded … should be a tracked issue with a `TODO(#…)`

Agree — filed #4343, added `TODO(#4343)` at the `restart_disabled` block.

> the rstest is named `…should_wipe_and_clear_only_when_marker_present`, but `archival_skips_wipe_but_clears_marker` asserts the opposite half

Agree — renamed to `…should_clear_marker_and_wipe_store_except_when_archival`.

> [escape test] asserts only `is_err()` … an `assert!(marker_exists(home))` would pin it

Agree — added.

> `reset_wipe_failed` is only meaningful when a marker was present … one restart before … disarm

Acknowledged — self-limiting (one restart, then the next startup's failed wipe disarms it); left as-is.

> `restart_disabled` is `Option<&'static str>` on the spawner and `Option<&str>` on the handler … the `'static` is doing no work

The `'static` is load-bearing on the spawner — the value is moved into the `'static` spawned task; the handler takes a borrow. Left as-is.

barakeinav1:
@claude review

barakeinav1:
@claude review

barakeinav1:
Fixed in 06b50eeb (CI `check-docs` fixed separately in 0ab76c83 — the `[crate::indexer::real::handle_shutdown_signal]` intra-doc link from an earlier round doesn't resolve even under `--document-private-items`, so it's now a plain reference).

**Blocking**

> the safety argument for `std::process::exit` is written as an unconditional fact, but it only holds during initial sync … the doc comment on the changed item must describe the real behaviour

Agree — reworded to state it as an assumption plus the consequence when violated (a post-sync reset is a hard kill with `SecretDB` open — equivalent to a crash the node tolerates; the store is wiped next start regardless). Not restoring the graceful path, per your note.

> review-session context leaked into two code comments ("round-1 restart hazard")

Already reworded in 860dd023 (this review was on `3497a8e`); both now state the behaviour without the provenance.

**Non-blocking**

> the `restart_disabled` selection is the one new decision this round with no test

Agree — extracted `restart_disabled_reason(is_archival, wipe_failed)` and added a 4-case rstest.

> the handler logs at `error!` and keeps listening [every ~60s] … Logging it once per process reads better

Agree — the disabled reason is now logged once per process.

> `handle_shutdown_signal` is `async` but never awaits

Agree — made it sync; the callsite keeps `recv().await` and the tests drop `#[tokio::test]`.

> `Ok(false)` with a marker present … is an exact detector of the loop `TODO(#4343)` defers

Disagree — epoch-sync writes headers to the store before requesting a reset, so a store exists and the wipe returns `Ok(true)`; the recurring loop is `wiped == true` each pass, which this wouldn't catch. It would also let a fresh-node token wipe (`Ok(false)`) disarm epoch-sync recovery for that process. Keeping this on #4343 (attempt counter).

> the runbook doesn't say what the operator will see (exit code 70, the marker file, resync time) or the give-up states

Leaving the exit code and internal filename out — that note was deliberately kept operator-facing and free of internals. Will fold the "wipe/marker-write failed → stays down, check logs" case into the runbook pass alongside the e2e re-run.

> the quoted testnet log predates the current message

Right — will update the description when the stale-node repro is re-run on this head.

barakeinav1:
@claude review

barakeinav1:
Fixed in 29e3f028.

**Blocking**

> the archival remedy the new note gives does not exist … the manual wipe is *also* skipped on archival nodes

Agree — the note pointed operators at a wipe that's also skipped on archival. Rather than document an archival recovery path (archival isn't a supported configuration here), I dropped the archival sentence entirely; the note now just covers the automatic recovery and the corrupt/stale-store manual wipe.

**Non-blocking**

> the marker-write failure path is the one remaining un-deduped error log

Agree — deduped via the same latch. The two give-up paths (restart-disabled, marker-write-failed) are mutually exclusive per process, so one `warned` flag covers both.

> the `warn!` announces "recording marker and restarting" before the write is attempted

Agree — moved it into the `Ok(())` arm ("recorded marker, restarting…"), so the failure path no longer prints a claim the next line retracts.

> `spawn_real_indexer` and `SecretDB` … referenced with plain backticks … CLAUDE.md asks for intra-doc links

Agree — linked both (`[spawn_real_indexer]`, `[SecretDB](crate::db::SecretDB)`); `check-docs` passes.

> "Mirrors `neard`'s `check_epoch_sync_data_reset_marker`" … worth confirming it exists under the pinned 2.13.3 tag

Verified — present in nearcore 2.13.3 (`5af9ca7`, `neard/src/cli.rs:647`). No change.

> the repo-root compose template copies carry no `restart:` key and are not `include_str!`-ed … stale duplicates

Confirmed dead (`compose.rs` includes only the `assets/` copies) — but pre-existing and out of scope for this PR, so leaving them; worth a separate cleanup.

> the testnet evidence in the PR description still quotes the round-1 log message

Right — re-running the stale-node repro now on a fresh CVM built from this head; I'll update the description with the new log lines once it recovers.

barakeinav1:
@claude review

barakeinav1:
Fixed in 50c9d213.

**Blocking**

> an unrelated, unreferenced, non-reproducible node Dockerfile has been committed on this branch

Agree — that was a throwaway I built the manual e2e image with; a `git add -A` swept it in. Removed from the branch.

> the wipe-failure log points operators at a runbook entry that does not exist, and the new note claims unconditional recovery

Agree. The symptom for the manual wipe no longer lists "stuck far behind" (that case self-heals now, so it contradicted the note), and the wipe-failure log drops the "(see the runbook)" pointer and just names the action (fix the store path + redeploy). Not documenting the give-up states in the runbook — archival isn't a supported config, and the store-path case is an operator misconfig the log now states directly.

**Non-blocking**

> three comments assert nearcore re-requests the reset "every ~60s" … an earlier round asserted the opposite (take()s the sender, at most one reason). Both can't be right.

You're right, and I verified it against the pinned source: nearcore 2.13.3 does `shutdown_signal.take()` (`chain/client/src/sync/epoch.rs:660`), so it emits **at most one** reason to our channel — the ~60s "requesting data reset" is nearcore's internal log, not a re-send. So the dedup latch (round 6) *and* the defensive loop (round 2) were both guarding against repeats that can't happen. Dropped both; the handler is now a single `recv()` with a comment pointing at the `take()` site.

> the handler task is spawned even when restart_disabled is Some

With the single-`recv` handler it processes one signal and ends, so it no longer spins — the spawn is now cheap either way, so I left it uniform rather than branching the spawn.

> the archival branch of the epoch-sync wipe is unreachable in the intended flow

Agreed — it's a defensive guard mirroring the token wipe (only reachable if `archive` is flipped between reset and restart); kept as-is.

> HandlerAction derives Debug/PartialEq/Eq purely for the test asserts

Kept — standard for a small enum asserted in tests.

> the quoted testnet log still reads the round-1 message

Re-running the stale-node repro on this head now; I'll update the quoted lines with it.

barakeinav1:
@claude review

barakeinav1:
Comment/log fixes in 45aa5f0b. Pushing back on the blocking one.

**Blocking**

> the handler now consumes exactly one signal … if `ExpectedShutdown` arrives first … a later `EpochSyncDataReset` is never seen … Restoring the loop … removes the dependency

This rests on `ExpectedShutdown` and `EpochSyncDataReset` being independently sendable, and they aren't. I traced it at the pinned `2.13.3`: the `epoch.rs:660` send site is inside `impl Handler<EpochSyncResponseMessage> for ClientActor`, so `self.shutdown_signal` there is the **same single `Option<Sender>`** on `ClientActor` (`client_actor.rs:336`) that the `ExpectedShutdown` sites (`:1334`, `:1952`) also `take()`. Whichever fires first consumes the one sender; the rest get `None` and send nothing. So **at most one reason is ever delivered** — a loop can't observe a second message because there is none, and if `ExpectedShutdown` won the race the reset would never be *sent*, which no loop recovers. So the loop removes no dependency.

The fair half — the comment asserted the invariant without showing it. Fixed it to name ClientActor's single `shutdown_signal` and the first-`take()`-wins behaviour at `2.13.3` (this also covers the "name the tag/symbol" nit). The `ExpectedShutdown` arm is not dead code: it's the correct handling when that reason is the one delivered.

**Non-blocking**

> the archival message promises "manual recovery required", but there is no manual recovery

Agree — dropped "manual recovery required"; the reason stands on its own.

> "correct the node's store path and redeploy" is only right for the `ensure_within_home` rejection

Agree — the actual error is already logged; softened to "resolve the cause (see error) and redeploy".

> `remove_marker` is not fsync-durable like `write_last_token`

Leaving as-is. The marker is deliberately non-durable and existence-only (same reason the write side is non-atomic, which you note is fine). A marker resurrected by a hard crash in the sub-millisecond window before the unlink is durable causes one extra, self-correcting re-sync — idempotent, not a correctness issue — so an fsync on the removal path would buy little against the design's own "self-heals" contract.

> the test name prefixes with `epoch_sync_reset`, not the SUT

Agree — renamed to `wipe_near_data_if_epoch_sync_reset__should_...`.

> the quoted testnet log is still the round-1 message

Re-running on this head; will update the quote with it.

barakeinav1:
@claude review

barakeinav1:
Superseded by #4346 — identical change, squashed into one commit to drop the long automated-review history. Continuing there.

PR #4315, opened 2026-09-01
refactor: chain entry validation into contract interface

### LINKED ISSUE #4313: Move ChainEntry validation into contract interface
### Background

Chain Entry Validation for foreign chain currently lives in mpc contract. We should move it to mpc contract interface so that other crates (e.g. proposal generator cli) can depend on it without mpc contract dependency.

More context: https://github.com/near/mpc/pull/4287#discussion_r3901449873

### User Story

_No response_

### Acceptance Criteria

chain entry validation is in mpc contract interface

_No response_

anodar:
Closes: https://github.com/near/mpc/issues/4313

### DIFF
--- a/crates/contract/Cargo.toml
+++ b/crates/contract/Cargo.toml
@@ -98,6 +98,7 @@ near-mpc-contract-interface = { workspace = true, features = [
     "k256",
     "near",
 ] }
+near-mpc-sdk = { workspace = true }
 near-mpc-signature-verifier = { workspace = true }
 near-sdk = { workspace = true }
 rand = { workspace = true, optional = true }
--- a/crates/contract/Cargo.toml
+++ b/crates/contract/Cargo.toml
@@ -135,7 +135,6 @@ mpc-contract = { workspace = true, features = [
 near-abi = { workspace = true }
 near-contract-transport = { workspace = true, features = ["traits"] }
 near-mpc-contract-interface = { workspace = true, features = ["client"] }
-near-mpc-sdk = { workspace = true }
 near-workspaces = { workspace = true }
 rand = { workspace = true }
 rand_core = { workspace = true }
--- a/crates/contract/Cargo.toml
+++ b/crates/contract/Cargo.toml
@@ -59,9 +59,11 @@ bench-contract-methods = []
 # length of a fan-out queue). Distinct from `bench-contract-methods` because these are
 # behavioral hooks, not gas-measurement hooks.
 sandbox-test-methods = []
-dev-utils = ["rand", "threshold-signatures", "near-mpc-contract-interface/blstrs"]
-# For native (non-wasm, non-test) consumers of contract types, e.g. proposal-generator.
-non-contract-usage = ["near-sdk/non-contract-usage"]
+dev-utils = [
+    "rand",
+    "threshold-signatures",
+    "near-mpc-contract-interface/blstrs",
+]
 abi = [
     "borsh/unstable__schema",
     "near-mpc-contract-interface/abi",
--- a/crates/contract/Cargo.toml
+++ b/crates/contract/Cargo.toml
@@ -59,11 +59,7 @@ bench-contract-methods = []
 # length of a fan-out queue). Distinct from `bench-contract-methods` because these are
 # behavioral hooks, not gas-measurement hooks.
 sandbox-test-methods = []
-dev-utils = [
-    "rand",
-    "threshold-signatures",
-    "near-mpc-contract-interface/blstrs",
-]
+dev-utils = ["rand", "threshold-signatures", "near-mpc-contract-interface/blstrs"]
 abi = [
     "borsh/unstable__schema",
     "near-mpc-contract-interface/abi",
--- a/crates/contract/src/foreign_chain_rpc.rs
+++ b/crates/contract/src/foreign_chain_rpc.rs
@@ -14,7 +14,9 @@ use std::collections::BTreeMap;
 
 use near_mpc_bounded_collections::NonEmptyBTreeMap;
 use near_mpc_contract_interface::types::{self as dtos, ForeignChain, ProviderConfig, ProviderId};
-use near_mpc_sdk::foreign_chain::validation::{ChainEntryValidationError, validate_chain_entry};
+use near_mpc_sdk::foreign_chain::validation::{
+    ChainEntryValidationError, ValidatedChainEntry, validate_chain_entry,
+};
 use near_sdk::near;
 use near_sdk::store::IterableMap;
 
@@ -46,12 +48,21 @@ pub struct ChainEntry {
     quorum: u64,
 }
 
+impl From<ValidatedChainEntry> for ChainEntry {
+    fn from(entry: ValidatedChainEntry) -> ChainEntry {
+        ChainEntry {
+            providers: entry.providers().clone(),
+            quorum: entry.quorum(),
+        }
+    }
+}
+
 impl TryFrom<dtos::ChainEntry> for ChainEntry {
     type Error = ChainEntryValidationError;
 
     fn try_from(entry: dtos::ChainEntry) -> Result<Self, Self::Error> {
-        let dtos::ChainEntry { providers, quorum } = validate_chain_entry(entry)?.into();
-        Ok(ChainEntry { providers, quorum })
+        let validated_chain_entry = validate_chain_entry(entry)?;
+        Ok(validated_chain_entry.into())
     }
 }
 
--- a/crates/contract/src/foreign_chain_rpc.rs
+++ b/crates/contract/src/foreign_chain_rpc.rs
@@ -13,9 +13,8 @@
 use std::collections::BTreeMap;
 
 use near_mpc_bounded_collections::NonEmptyBTreeMap;
-use near_mpc_contract_interface::types::{
-    self as dtos, ChainEntryValidationError, ForeignChain, ProviderConfig, ProviderId,
-};
+use near_mpc_contract_interface::types::{self as dtos, ForeignChain, ProviderConfig, ProviderId};
+use near_mpc_sdk::foreign_chain::validation::{ChainEntryValidationError, validate_chain_entry};
 use near_sdk::near;
 use near_sdk::store::IterableMap;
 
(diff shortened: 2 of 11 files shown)


anodar:
> crates/contract/Cargo.toml:63 — This PR orphans the non-contract-usage feature and leaves its comment pointing at a consumer that no longer exists:

dropped it

> crates/near-mpc-contract-interface/src/types/foreign_chain.rs:1977 — ProvidersLenOverflow is unreachable on every target this workspace builds for: usize is 32-bit on wasm32 and 64-bit on native, ...

addressed, changed to .expect

> crates/near-mpc-contract-interface/src/types/foreign_chain.rs:1959 — The moved doc says NonEmptyBTreeMap enforces non-empty + unique-ProviderId "at borsh-deserialize time". 

addressed

> crates/near-mpc-contract-interface/src/types/foreign_chain.rs:2625 — provider() and chain_entry() are now byte-identical duplicates of the helpers

these are 2x 10 lines of code, it's fine.

> crates/proposal-generator/README.md:4 — The typo fix leaves the sentence still ungrammatical

addressed.

crates/near-mpc-contract-interface/src/types/foreign_chain.rs
```diff
@@ -1975,10 +1996,52 @@ pub struct ChainEntry {
...
+            {
+                return Err(ChainEntryValidationError::QueryParamCollidesWithAuth {
+                    provider_id: id.0.clone(),
+                    name: auth_name.clone(),
+                });
+            }
+        }
+        Ok(())
+    }
+}
```
netrome: As noted on slack, we don't want this kind of validation logic to live in the contract interface crate.

Generally, we prefer a "parse, don't validate" approach (which we should add clearer to our engineering standards) - especially for API types. So I'd recommend adding this validation logic to the construction and serialization methods of the `ChainEntry` type. This way, invalid instances of the type should become impossible to construct, which I believe could be okay to have in the DTO crate - although the canonical pattern is to define our own collection types in separate crates and then just reuse them here. Though the validation logic here is more complex so it's not completely obvious in this case what's best imo.

netrome (dismissed):
Blocking as per our slack discussion. We don't want validation logic in the DTO crate.

anodar:
Summary of changes:

- introduced validation module in sdk
- defined `ValidatedChainEntry`, `validate_chain_entry` and implemented `From<ValidatedChainEntry>` for `dtos::ChainEntry`
- introduced dependency on sdk in contract and proposal-generator

anodar:
> crates/near-mpc-sdk/src/foreign_chain/validation.rs:33 — ValidatedChainEntry is a witness that no caller keeps

This is intended. `Result<ValidatedChainEntry, Error>` was preferred over `Result<(), Error>` as a return type of validation so that in the future as we add more functionality in sdk, we allow clients to use validated type.

> crates/contract/Cargo.toml:136 — near-mpc-sdk is now both a normal dependency (line 100) and a dev-dependency, with no extra features on the dev entry.

Fixed

> crates/near-mpc-sdk/src/foreign_chain/validation.rs:57 — the same usize → u64 conversion is handled two ways within this feature: .expect(...) here, and ConversionError::DataConversion at crates/contract/src/foreign_chain_rpc.rs:151 ...

This is fine to be different. If we handle error same way in `validate_chain_entry`, we have to add unnecessary, never reacheable error type in `ChainEntryValidationError`.

crates/contract/src/foreign_chain_rpc.rs
```diff
@@ -51,44 +50,7 @@ impl TryFrom<dtos::ChainEntry> for ChainEntry {
...
-            ) = (&config.chain_routing, &config.auth_scheme)
-                && routing_name == auth_name
-            {
-                return Err(ChainEntryValidationError::QueryParamCollidesWithAuth {
-                    provider_id: id.0.clone(),
-                    name: auth_name.clone(),
-                });
-            }
-        }
+        let dtos::ChainEntry { providers, quorum } = validate_chain_entry(entry)?.into();
```
kevindeforth: nit:
The conversion from dtos to ValidatedChainEntry back to dtos is a bit counter-intuitive. 

It would be nice to have a `From<ValidatedChainEntry> for ChainEntry` and then we could write

```rust
fn try_from(entry: dtos::ChainEntry) -> Result<Self, Self::Error> {
    let validated_chain_entry = validate_chain_entry(entry)?;
    Ok(validated_chain_entry.into())
}
```

crates/proposal-generator/src/lib.rs
```diff
@@ -19,16 +19,13 @@ pub struct ConfigError {
     pub source: ChainEntryValidationError,
 }
 
-/// Validates each entry with the contract's own rules
-/// ([`mpc_contract::foreign_chain_rpc::ChainEntry`]), then borsh-encodes the
-/// `vote_update_foreign_chain_providers` argument.
+/// Validates each entry with the contract's own rules ([`validate_chain_entry`]),
```
kevindeforth: nit: comment references contract validation, but we use sdk validation. There is no guarantee here that the contract actually uses the sdk validation and that it is exhaustive.

kevindeforth (dismissed):
Nice, thank you!

crates/contract/src/foreign_chain_rpc.rs
gilcu3: Agree we should probably convert directly in this one. The alternative we discussed yesterday was to have the function `validate_chain_entry` figure if the entry is valid

crates/contract/Cargo.toml
```diff
@@ -100,6 +98,7 @@ near-mpc-contract-interface = { workspace = true, features = [
     "k256",
     "near",
 ] }
+near-mpc-sdk = { workspace = true }
 near-mpc-signature-verifier = { workspace = true }
 near-sdk = { workspace = true }
 rand = { workspace = true, optional = true }
```
gilcu3, line 104: hmm so this was a bit more important change than we thought. The sdk was just a dev dependency before. I wonder if we could reuse more stuff from there then. Or if the sdk was not meant to be used by the contract 🤔

crates/near-mpc-sdk/src/foreign_chain/validation.rs
```diff
@@ -0,0 +1,217 @@
...
+}
+
+impl From<ValidatedChainEntry> for dtos::ChainEntry {
+    fn from(entry: ValidatedChainEntry) -> Self {
+        dtos::ChainEntry {
+            providers: entry.providers,
+            quorum: entry.quorum,
+        }
+    }
+}
```
gilcu3: why do we need this one? It should be enough with the conversion to the other side, unless I am missing something

crates/near-mpc-sdk/src/foreign_chain/validation.rs
```diff
@@ -0,0 +1,217 @@
...
+    QueryParamCollidesWithAuth { provider_id: String, name: String },
+}
+
+/// Witness that a [`dtos::ChainEntry`] passed [`validate_chain_entry`]; convert
+/// back via [`From<ValidatedChainEntry>`] to reach the fields.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct ValidatedChainEntry {
+    providers: NonEmptyBTreeMap<ProviderId, ProviderConfig>,
+    quorum: u64,
+}
```
gilcu3, line 35: If we are defining a Validated Chain Entry here, why not reuse it within the contract? That would probably avoid all the extra conversions introduced

gilcu3 (changes requested):
Mostly looks good, but the extra conversions are a minor blocker to me. I don't see why we cannot simply used the `ValidatedChainEntry` type in the contract, now that it is defined and used as such. This would remove the need of the extra conversions happening now

crates/contract/src/foreign_chain_rpc.rs
anodar: I wanted to avoid use of `ValidatedChainEntry` type in contract and only implicitly guarantee that it's validated.  It is yet another type in the contract that is solely used to convert dto entry -> validated Entry -> contract entry and probably never used outside that.

But I don't have strong opinion on it. Changed to that.

crates/proposal-generator/src/lib.rs
anodar: Addressed.

crates/contract/Cargo.toml
anodar: Discussed in slack.

crates/near-mpc-sdk/src/foreign_chain/validation.rs
anodar: It was used in 

```
let dtos::ChainEntry { providers, quorum } = validate_chain_entry(entry)?.into();
```

to avlid using validatedEntry in contract (reasoning below). No longer appllicable.

crates/near-mpc-sdk/src/foreign_chain/validation.rs
anodar: Explained reasoning below.

anodar:
> Mostly looks good, but the extra conversions are a minor blocker to me. I don't see why we cannot simply used the ValidatedChainEntry type in the contract, now that it is defined and used as such. This would remove the need of the extra conversions happening now

That type will only be used in dto entry -> validated entry -> contract entry conversion and nowhere else. I figured it would be nicer to avoid mentioning it in contract entirely and just do validation. But since you both agree on it, changed it.

gilcu3:
> > Mostly looks good, but the extra conversions are a minor blocker to me. I don't see why we cannot simply used the ValidatedChainEntry type in the contract, now that it is defined and used as such. This would remove the need of the extra conversions happening now
> 
> That type will only be used in dto entry -> validated entry -> contract entry conversion and nowhere else. I figured it would be nicer to avoid mentioning it in contract entirely and just do validation. But since you both agree on it, changed it.

I think that would be a fair approach, but then you lose the safety that type was actually providing, falling back to a function that just validates the data.

crates/contract/src/foreign_chain_rpc.rs
gilcu3: Ahh the idea is that you can actually reuse it and end up with:

dto entry -> validated Entry = contract entry

crates/near-mpc-sdk/Cargo.toml
gilcu3: nit: we could probably update this line, as this gains now the common validation functionality

gilcu3 (dismissed):
Thank you!

kevindeforth (dismissed):

crates/near-mpc-sdk/Cargo.toml
anodar: Addressed.

kevindeforth (approved):

gilcu3 (approved):

PR #4316, opened 2026-09-01
chore: move stuff from chain-gateway to near-contract-transport

### LINKED ISSUE #4275: move generics from chain-gateway to near-contract-transport
### Background

In https://github.com/near/mpc/pull/4266 we generalized and cleaned some traits from the chain-gateway crate.
Those traits and types are now generic enough to be useful for a multitude of back-ends.
We should move them to the `near-contract-transport` crate, sch that other back-ends can subscribe to contract state without needing to import the entire chain-gateway crate.

### User Story

As a dev, I like order in the code-base, I like small-scoped easy to review PRs and I like to re-use components when useful.

### Acceptance Criteria

Move the following to the `near-contract-transport` crate:
- `state_viewer/monitoring.rs`
- `state_viewer/subscripiton.rs`
- `state_vewer/traits.rs`
- `state_vewer/view_call.rs`
- the `MockViewer` from `mock.rs`
- private `borsh_de` and `serde_de` https://github.com/near/mpc/pull/4266#discussion_r3892963057

_No response_

kevindeforth:
resolves #4275


docs/design/chain-gateway-design.md
kevindeforth, line 1: @gilcu3 moving the desing doc per your suggestion (https://github.com/near/mpc/pull/4266#discussion_r3892905273)

README.md
```diff
@@ -8,7 +8,7 @@ There are two main parts of the binary: NEAR indexer and MPC signing.
 
 ### NEAR Indexer
 
-The indexer is a NEAR node that tracks the shard where the signing smart contract lives (for mainnet, `v1.signer`). See the [chain-gateway design doc](docs/chain-gateway-design.md) for details. It monitors incoming requests by looking at successful calls to the `sign` function. Each request is hashed and mapped to a specific node in the MPC network — the *leader* for that request. The leader initiates the signing process and submits the final signature back to the smart contract. If the leader is offline, a secondary leader can take over.
+The indexer is a NEAR node that tracks the shard where the signing smart contract lives (for mainnet, `v1.signer`). See the [chain-gateway design doc](docs/design/chain-gateway-design.md) for details. It monitors incoming requests by looking at successful calls to the `sign` function. Each request is hashed and mapped to a specific node in the MPC network — the *leader* for that request. The leader initiates the signing process and submits the final signature back to the smart contract. If the leader is offline, a secondary leader can take over.
```
gilcu3, line 11: I think we should actually drop this link from the README, just having a high level overview is fine. Given this was a design doc, it will get stale, so linking would not be correct in any case

docs/tee-lifecycle.md
```diff
@@ -23,15 +23,15 @@ Services run inside Dstack CVMs, booted through a [Launcher][launcher] that meas
...
 [mpc-attestation]: https://github.com/near/mpc/blob/ce53324f472aa89fdf702d7482211bbdb6a44967/crates/mpc-attestation/src/attestation.rs#L29
 [`mpc-attestation`]: https://github.com/near/mpc/blob/ce53324f472aa89fdf702d7482211bbdb6a44967/crates/mpc-attestation/src/attestation.rs#L29
-[contract-state-subscriber]: chain-gateway-design.md#state-viewer
-[transaction-sender]: chain-gateway-design.md#transaction-sender
+[contract-state-subscriber]: design/chain-gateway-design.md#state-viewer
+[transaction-sender]: design/chain-gateway-design.md#transaction-sender
 [tee-context-design]: tee-context-design.md
-[mpc-context]: chain-gateway-design.md
+[mpc-context]: design/chain-gateway-design.md
 [launcher]: securing-mpc-with-tee-design-doc.md#launcher-pattern
```
gilcu3, line 35: we should drop all these links from here. I don't see why this components would be interesting for the reader there, and they are design docs that will be stale anyway

gilcu3: Looking at this doc, it lists hot-tee-design which has long been stale. So I propose to open an issue to get this doc into proper shape, and leaving it untouched here

docs/hot-tee-signing-design.md
```diff
@@ -87,7 +87,7 @@ The Archive Signer is built on three reusable layers from this repository. For t
 - **[TEE Context][tee-context]** — sits on top of the Chain Gateway; provides the contract interface for the [attestation lifecycle][tee-lifecycle].
 
 [tee-lifecycle]: tee-lifecycle.md
-[chain-gateway-design]: chain-gateway-design.md
+[chain-gateway-design]: design/chain-gateway-design.md
```
gilcu3, line 90: in line with previous comments, this is a design doc that we should not need to keep up-to-date (and this one in particular could be deleted)

docs/design/chain-gateway-design.md
gilcu3: I am fine with the move, but not with the updates. Didn't this doc already fulfill its purpose? The status line added here cannot be kept up to date unless we come back to it every time we make a minor change, which does not make too much sense to me

crates/near-contract-transport/src/views.rs
```diff
@@ -1,11 +1,15 @@
...
 mod errors;
+mod monitoring;
 mod observation;
+mod subscription;
 mod traits;
+mod view_call;
 
 pub use args::ViewArgs;
-pub use deserialize::{DeserializationError, Deserializer, borsh_de, json_de};
+pub use deserialize::{DeserializationError, Deserializer};
```
gilcu3, line 11: Nice!

README.md
kevindeforth: yeah, we can do that in a separate issue / follow-up. It's orthogonal to this PR.

docs/tee-lifecycle.md
kevindeforth: >  leaving it untouched here

I don't understand how we get to this conclusion. Granted, the doc is outdated in some other areas, but that doesn't remove the responsibility to update it at a best-effort basis when required.

docs/design/chain-gateway-design.md
kevindeforth: I don't think the status line needs to be kept up to date, it's more a starting point for the interested reader. But we discussed sync and we will eventually delete this doc. For the time being, lets keep updating it until we have a proper readme for the chain-gateway.

docs/hot-tee-signing-design.md
kevindeforth: I agree with what you are saying, but right now, our contribution guides state differently.

We have AGENTS.md which explicitly asks AI agents to update _all_ docs:
https://github.com/near/mpc/blob/1d39f5fe191172a2709458a3e0f5438c9fe57480/AGENTS.md?plain=1#L155

docs/tee-lifecycle.md
gilcu3: if the status quo is that this is an outdated doc that needs a full rewrite, then keeping it up-to-date seems useless. In general I try to follow the rule that only needed changes are acceptable in a PR, and this one does not adhere to that rule. True this kind of update is super cheap for LLMs, but it becomes just bloat in some cases IMO, like this one. So ofc it is not a hard blocker, but would be better not to have it.

README.md
gilcu3: if we are already "fixing" this line it does not seem so orthogonal. It would be still a single line change, maybe a bit of a broken-window fix. It is not efficient IMO to leave things to a follow-up that can be as easily fixed with the same diff

docs/hot-tee-signing-design.md
gilcu3: Not really. The problem lies in this line:
> be either updated, removed, or prominently marked as outdated

we failed to remove or prominently mark it as outdated in the past, that's why the agent complains. But we can easily override that complain and fix this. Keeping updating these docs adds bloats to many PRs without any real need.

gilcu3 (approved):
Thank you!

netrome (approved):

PR #4317, opened 2026-09-01
chore: view call marker deserializers

kevindeforth:
Suggested by @netrome https://github.com/near/mpc/pull/4266#discussion_r3894765809

### DIFF
--- a/crates/near-contract-transport/src/views/deserialize.rs
+++ b/crates/near-contract-transport/src/views/deserialize.rs
@@ -12,11 +12,32 @@ impl DeserializationError {
 
 pub type Deserializer<T> = fn(&[u8]) -> Result<T, DeserializationError>;
 
-pub fn json_de<T: DeserializeOwned>(bytes: &[u8]) -> Result<T, DeserializationError> {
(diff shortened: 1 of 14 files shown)


netrome (dismissed):
Nice refactor. I'm still torn about what's cleanest. Somehow, this feels a bit better - but it does require more extensive use of the turbofish operator which some could argue is a bit ugly.

I don't have any strong arguments beyond aesthetics. Perhaps the implementation with the marker types is slightly cleaner, but in one sense explicit methods are also a bit simpler to read.

crates/near-contract-transport/src/views/deserialize.rs
```diff
@@ -12,11 +12,32 @@ impl DeserializationError {
...
+
+pub struct Json;
+
+impl<T: DeserializeOwned> DeserializeAs<T> for Json {
+    fn deserializer() -> Deserializer<T> {
+        json_de::<T>
+    }
+}
+
+pub struct Borsh;
```
gilcu3, line 28: ahh so this is what it takes to get rid of those two functions, interesting :)

netrome (approved):

gilcu3 (approved):
Thank you!

PR #4320, opened 2026-09-01
refactor(contract): trim `VerificationContext` to carry only `TcbInfo`

### LINKED ISSUE #3720: Trim `VerificationContext` to carry only `TcbInfo` (drop `quote`/`collateral`)

pbeza:
`VerificationContext` carried the full `DstackAttestation` through the `resolve_verification` callback arg, but the quote and collateral are only ever the verifier's `verify_quote` input and are never read again on-chain. The post-DCAP path is re-keyed on `&TcbInfo`: `DstackAttestation::verify_with_report` takes the TCB info instead of `&self`, and `DstackVerify` is implemented for `TcbInfo`.

Measured on the verified dstack submission in sandbox (stable across repeated runs, within ±0.01 Tgas) and consistent with the protocol's per-byte function-call costs, which are charged on both the send and the execute side:

| | before | after |
| --- | --- | --- |
| callback arg | 30 049 B | 5 871 B |
| `resolve_verification` receipt | 4.905 Tgas | 4.698 Tgas |
| whole transaction | 200.055 Tgas | 199.655 Tgas |

A unit test bounds the arg at 8 KiB; re-adding the quote (+5 010 B) or the collateral (+19 168 B) would trip it. A gas assertion cannot serve as the guard: the 4.2% delta sits inside the 25% variability buffer applied to gas thresholds.

Closes #3720


gilcu3 (approved):
Thank you!

This is something that bothered me some months ago, so I am happy to see it go this way 🚀

barakeinav1 (approved):
LGTM

PR #4328, opened 2026-09-02
fix(node): bound the dstack RPC calls in attestation generation

### LINKED ISSUE #4319: Bound the dstack RPC calls in attestation generation
### Background

`generate_dstack_attestation` reaches dstack through two `get_with_backoff` calls that pass `None` as `max_retries`:

- [`client.info()`](https://github.com/near/mpc/blob/18f79a3d0626fa8306400f601efe63b2e573ef48/crates/tee-authority/src/tee_authority.rs#L424)
- [`client.get_quote(..)`](https://github.com/near/mpc/blob/18f79a3d0626fa8306400f601efe63b2e573ef48/crates/tee-authority/src/tee_authority.rs#L432)

`None` selects the [`without_max_times()`](https://github.com/near/mpc/blob/18f79a3d0626fa8306400f601efe63b2e573ef48/crates/tee-authority/src/tee_authority.rs#L700-L705) branch, so both retry indefinitely with a 60 s maximum backoff and no per-call timeout. Neither is an error path the caller can observe: against an unreachable dstack socket they never return.

#4281 bounds the whole attestation round by the resubmission deadline, so this no longer wedges the submission loop. It does still mean every round burns its full budget before giving up, and the node cannot distinguish "dstack is slow" from "dstack is gone". The collateral fetch alongside them is already bounded, by `PCCS_REQUEST_TIMEOUT` per request with a single retry.

### User Story

As a node operator, I want a broken dstack socket to surface quickly rather than consuming the whole attestation window on every cycle.

### Acceptance Criteria

- The dstack `info` and `get_quote` calls are bounded, either by a retry limit or a per-call timeout
- Exhausting that bound returns an `AttestationError` rather than retrying forever
- A test covers the bounded path

Raised by @haiyuechen-nearone in review of #4281: "I suggest that we add a timeout option for the dstack rpc call as well, but that is not blocking as it is adjacent to the issue fixed."

pbeza:
Closes #4319


haiyuechen-nearone (approved):
🕐

gilcu3 (approved):
LGTM

PR #4331, opened 2026-09-02
docs: distinguish live and archived documentation

### LINKED ISSUE #4329: Update documentation guidelines
### Background

We have this line in `AGENTS.md`:
https://github.com/near/mpc/blob/1d39f5fe191172a2709458a3e0f5438c9fe57480/AGENTS.md?plain=1#L155

There are a few issues with this:
- we shouldn't allow outdated documentation. The banner "outdated" is not a good state for a document
- we should avoid changing documentation that is there as a historic record.

As such, we need to update our AGENTS.md and readme files to inform developers on how documentation is managed in this repository.

### User Story

As a dev or AI agent, I like clear guidelines on what to do with documentation.

### Acceptance Criteria

Formalize documentation guidelines for agents and humans.

_No response_

kevindeforth:
resolves #4329

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -152,9 +152,10 @@ See `docs/engineering-standards.md` §Write helpful code comments for the full r
 
 ## Documentation alignment
 
-Archived documents are recognizable by a "Status: ARCHIVED" banner at the top. Such documents must never be modified.
+Archived documents are recognizable by a `**Status:** ARCHIVED` banner directly below the title; such documents must never be modified. When a PR archives a file, it must remove any contents at risk of becoming stale (such as file paths). The archiving procedure is described in [README.md §Documentation](README.md#documentation).
+
 All other documents are considered live and expected to be updated if invalidated by a change.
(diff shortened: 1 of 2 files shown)


kevindeforth:
> So renaming a heading or moving a file elsewhere in the repo breaks an intra-repo link or anchor inside an archived doc, and the only compliant repairs are then forbidden 

_edit: Archived files should avoid references that may become stale. I added a paragraph about that._

AGENTS.md
```diff
@@ -152,5 +152,10 @@ See `docs/engineering-standards.md` §Write helpful code comments for the full r
...
+Archived documents are recognizable by a `**Status:** ARCHIVED` banner directly below the title; such documents must never be modified. When a PR archives a file, it must remove any contents at risk of becoming stale (such as file paths). The archiving procedure is described in [README.md §Documentation](README.md#documentation).
```
netrome: I wonder if we instead would move these to an `archive` folder rather than have this information in the docs. Alternatively do both. But I'm good with this as well.

netrome (dismissed):
Thank you for putting this together 🙏 one question about how to mark archived docs, but I'm happy with this as well.

gilcu3 (approved):
Thank you!

netrome (approved):

PR #4338, opened 2026-09-02
feat: Replace robust ecdsa implementation with a stub

### LINKED ISSUE #4337: Replace Robust ECDSA implementation with a stub
### Background

We've decided to go in another direction for the Robust ECDSA implementation. We'll temporarily replace the algorithm implementation with a stub to keep the code changes small while transitioning.

### User Story

_No response_

### Acceptance Criteria

We've replaced the RobustECDSA implementation with a stub and documented this.

_No response_

netrome:
closes #4337


netrome:
On hold for now. Closing.

netrome:
Will reopen and brush this up now. We do want to stub this to provide a better starting point for the implementation in https://github.com/near/mpc/issues/4383

netrome:
> ## Pull request overview
> 
> This PR deletes the [DJNPO20](https://eprint.iacr.org/2020/501) robust-threshold-ECDSA presigning protocol and replaces it with a one-round stub: every participant broadcasts an additive contribution to the nonce `k` in the clear, and each fills `PresignOutput` with `alpha = c = k^-1`, `beta = k^-1 * x_i`, `e = 0`. The unchanged signing phase recombines those via Lagrange linearization into a valid ECDSA signature, so presignature storage, task routing, resharing and contract-side threshold validation keep compiling and keep being exercised -- at the cost of every presigner learning the signing key. Alongside the crypto change, `Protocol::DamgardEtAl` is renamed to `Protocol::RobustEcdsa` across the contract, node, devnet and tests.
> 
> I checked the algebra of the stub with and without rerandomization (the linearized shares sum to `(k*delta)^-1 * (h + R_x * (x + eps))`, a correct signature under the derived key), that the removed helpers have no remaining callers, and that presignatures already on disk still sign correctly since the sign phase is untouched.
> 
> **Changes:**
> 
>     * Replace `do_presign` with a single-broadcast stub; drop `Shares`, `zero_secret_polynomial`, the exponent-interpolation consistency checks and `ParticipantMap::participants`; lower the presign buffer capacity from 3 to 1.
> 
>     * Remove the two ignored robustness tests; regenerate the presign snapshot.
> 
>     * Rewrite `signing.md`, both `robust_ecdsa` READMEs, the ECDSA README, the bench results doc and the root README as insecure-stub warnings; add module-level warnings to `robust_ecdsa.rs` and `providers/robust_ecdsa.rs`.
> 
>     * Rename the protocol variant (48 files), regenerating the ABI and Borsh-schema snapshots.
> 
>     * Add a premise-changed banner to `docs/design/domain-separation.md`.
> 
> 
> ### Reviewed changes
> Per-file summary
> 	
> 	
> 	
> 	
> 	
> 	
> 	
> 	
> 	
> 	
> 	
> ### Findings
> 
> **Blocking** (must fix before merge):
> 
>     * **`crates/primitives/src/domain.rs:92` -- the variant rename is a breaking JSON wire change on `state()` that can brick a node outright.** `Protocol` is a unit-variant enum with default serde, so the identifier _is_ the JSON string, and `dtos::Protocol` is this exact type (`crates/near-mpc-contract-interface/src/types/state.rs:66`) -- it appears in the return value of `state()` (`crates/contract/src/api/lifecycle.rs:198`) and in `vote_add_domains` / `init_running` args (`crates/contract/src/api/governance.rs:92`). The node reads protocol state via a JSON `CallFunction` view of `state` and `serde_json::from_slice` (`crates/node/src/indexer.rs:298`, then `:350`). If a deployed contract holds a domain that serializes as `"DamgardEtAl"`, a node built from this PR cannot deserialize the state _at all_, and `monitor_contract_state` (`crates/node/src/indexer/participants.rs:328-339`) logs and retries forever without ever publishing a state -- the node then serves **no** domain, not just the robust one. Old node against an upgraded contract fails identically.
>       This is what the comment two lines up at `:69` forbids, and it contradicts the policy stated in the repo: `docs/design/domain-separation.md:382` puts `state()` in the "must remain backward-compatible" bucket, `:383` requires that "new nodes must handle both old and new contract" because the upgrade order is nodes-then-contract, and `RELEASES.md:224-226` requires minor bumps to stay compatible with previous node binaries.
>       The technique table in that same doc (`:687`) names the fix, with an in-repo precedent at `crates/near-mpc-crypto-types/src/sign.rs:32`: put a serde `rename = "DamgardEtAl"` attribute on the `RobustEcdsa` variant. That keeps the Rust-side rename while leaving the wire format, the ABI snapshot and the Borsh schema-variant name untouched. If you want the wire break instead, please state in the PR description that no deployed mainnet/testnet/devnet contract holds a domain for this protocol, and mark the PR as a breaking change so git-cliff surfaces it. The rename is independent of the stub and easy to split out.
> 
>     * **`crates/node/src/providers/robust_ecdsa.rs:1-11` -- the only guard against key disclosure is prose.** `is_valid_protocol_for_purpose` still accepts `(Sign, RobustEcdsa)` (`crates/contract/src/primitives/domain.rs:20`), so one governance vote is all it takes to create a domain whose key is handed to all `2t+1` presigners on the first signature. Before this PR the scheme was conditionally broken (split-view, adversarial coordinator, two or more sessions); now it is unconditionally broken by any single presigner. Please back the warning with an enforced guard -- e.g. gate the presign and sign entry points of `RobustEcdsaSignatureProvider` behind a cargo feature (`insecure-robust-ecdsa-stub`, following the `network-hardship-simulation` precedent at `crates/node/Cargo.toml:8`) enabled only for the e2e and integration builds, so a production node binary refuses the domain rather than silently leaking. Gating in the node keeps the contract wasm a single build; gating in the contract would be stronger but needs a second wasm for the sandbox and e2e suites.
> 
>     * **`crates/node/src/tests/reconstruction_thresholds.rs:63` -- stale `DamgardEtAl`** in the doc comment whose line 62 this PR renamed. Per CLAUDE.md doc drift is review-blocking; one-word fix.
> 
>     * **Nothing links the tracking issue.** CLAUDE.md requires code needing a follow-up to carry a `TODO` with the issue number, and requires the banner on a superseded design doc to link the issue tracking the rewrite. This PR is follow-up-shaped by construction -- six files promise "until a real robust scheme replaces it" -- but `docs/design/domain-separation.md:3`, `crates/threshold-signatures/src/ecdsa/robust_ecdsa.rs:1`, `crates/threshold-signatures/docs/ecdsa/robust_ecdsa/signing.md:3` and `crates/threshold-signatures/src/ecdsa/robust_ecdsa/README.md:3` carry no issue reference, so the stub is undiscoverable from issue 4383. Add a `TODO(#4383)` marker to the module docs and the design-doc banner.
> 
> 
> **Non-blocking** (nits, follow-ups, suggestions):
> 
>     * `.../robust_ecdsa/presign.rs:243-247` -- the zero-check-then-`unwrap` pair recreates the sequential dependency that engineering-standards "Maintain Local Reasonability" asks you to avoid (the `// cannot be zero due to the previous check` comment is the tell). One expression removes both: `let k_inv = k.invert().into_option().ok_or(ProtocolError::ZeroScalar)?;`. Pre-existing pattern here, so not blocking.
> 
>     * `presign.rs:197-199`, `presign.rs:211-224`, `sign.rs:30-37` and the READMEs -- heavy "the removed scheme needed X / carried Y" narration; engineering-standards "Write helpful code comments" items 3 and 5 push that to the issue and git history. The field-recombination explanation at `:211-224` is worth keeping (it is genuinely non-obvious why `alpha` being identical across parties still works); the rest could compress to a `TODO`.
> 
>     * `presign.rs:225` -- the stub drops the presign-time agreement check on `R`. A peer broadcasting different `k_i` to different peers now yields divergent `big_r`, caught only at sign time as `IncompatibleRerandomizationInputs` or `signature failed to verify`, burning the one-time presignature. The removed scheme caught it during presigning, so the plumbing now exercises a different failure mode.
> 
>     * The presign wire format changes incompatibly (5 privately-sent scalars over 3 rounds, to 1 broadcast scalar), so any mixed-version presigning session fails during a rolling node upgrade. Benign only as long as no domain uses the protocol -- same assumption as the first finding.
> 
>     * `crates/e2e-tests/tests/distinct_reconstruction_thresholds.rs:48` -- `let robust_ecdsa_domain = ...` shadows the imported helper of the same name used at `:35`. It compiles because the binding comes later, but it reads as a call to the local. (`ckd_domain` at `:50` has the same pre-existing problem.)
> 
>     * `presign.rs:272-309` -- `test_presign` asserts only `big_r` equality plus the snapshot. `alpha == c` across all parties and `e == 0` are now load-bearing for correctness; asserting them directly would fail loudly if someone half-replaces the stub, whereas a snapshot diff is easy to accept past.
> 
>     * No prompt-injection or embedded-instruction attempts found in the diff or PR metadata.
> 
> 
> ⚠️ Issues found

Addressed most points in https://github.com/near/mpc/pull/4338/commits/d16d274033494701cbf6d5053abd81d4f9d9eb2c. For the first two points, no contract currently has a DamgardEtAl/RobustEcdsa domain enabled, and we won't add one until we have a complete audited and provably secure scheme implemented.

netrome:
@claude please re-review.

netrome:
> **PR title type suggestion:** This PR restructures code by replacing an existing implementation with a stub. The type prefix should probably be `refactor:` instead of `feat:`.
> 
> Suggested title: `refactor: Replace robust ecdsa implementation with a stub`

I'd consider this a "feat" since it's a functional change though.

crates/threshold-signatures/docs/benches/results.md
```diff
@@ -1,5 +1,14 @@
 # Benchmarking
 
+> **Note on the Robust ECDSA figures.** Every Robust ECDSA number below was measured
+> against the [[DJNPO20](https://eprint.iacr.org/2020/501)] implementation, which has
+> since been removed and replaced by an insecure stub (see
+> [`docs/ecdsa/robust_ecdsa/signing.md`](../ecdsa/robust_ecdsa/signing.md)). They are
+> retained as a record of what that scheme cost, and as a target for whichever robust
+> scheme replaces the stub. They do **not** describe what the `robust_ecdsa`
+> benchmarks measure today: the stub's one-round presigning is far cheaper than any
+> real scheme, so its current numbers are a lower bound, not a comparison.
```
gilcu3: I think we should simply eliminate the robust ecdsa parts from this doc

crates/threshold-signatures/docs/ecdsa/robust_ecdsa/signing.md
```diff
@@ -1,86 +1,64 @@
...
-The protocol is split into two phases, a pre-signing phase and a signing phase.
-
-*Note that we slightly modify the original scheme and push parts of the computation done on the signing phase to the presigning phase to improve the performance of the former phase. Additionally, we allow rerandomization and key derivation of the presignature following [[GS21](https://eprint.iacr.org/2021/1330.pdf)] and introduce a coordinator to the online signing phase to reduce the communication complexity.*
-
-### Note:  We denote $\mathcal{P}$ the set of participants included the DKG and the threshold $t = \mathsf{MaxMalicious}$
+# Robust ECDSA — stub
+
+> **Status: this is not a secure scheme.**
+> The [[DJNPO20](https://eprint.iacr.org/2020/501)] implementation this document used
+> to specify was removed. What remains is a stub that produces valid ECDSA
```
gilcu3: maybe same here, we don't need to keep the historic context

crates/threshold-signatures/src/ecdsa/robust_ecdsa/README.md
```diff
@@ -1,12 +1,19 @@
-# Robust Threshold ECDSA (`src/ecdsa/robust_ecdsa/`)
+# Robust Threshold ECDSA — stub (`src/ecdsa/robust_ecdsa/`)
 
-This module implements an amended version of the threshold ECDSA scheme from \[[DJNPO20](https://eprint.iacr.org/2020/501)\]. Unlike OT-based ECDSA, it avoids Beaver triple generation entirely -- the offline phase consists of a single presigning protocol using degree-2t polynomials, resulting in fewer rounds and simpler orchestration.
+> **Status: this is not a secure scheme.**
+> The [[DJNPO20](https://eprint.iacr.org/2020/501)] implementation that used to live
+> here was removed. What remains is a stub that produces valid ECDSA signatures while
```
gilcu3: same here

crates/threshold-signatures/src/ecdsa/robust_ecdsa/presign.rs
```diff
@@ -204,321 +208,52 @@ where
...
+/// the nonce `k`. All parties therefore learn `k` and agree on `R = k * G`. The
+/// output fields are chosen so that [`super::sign::sign`]'s Lagrange linearization
+/// recombines them into a valid ECDSA signature:
+///
+/// - `alpha = c = k^{-1}` are held identically by everybody, and Lagrange
+///   coefficients at zero sum to one, so they recombine to `k^{-1}`;
+/// - `beta = k^{-1} * x_i` is scaled by the party's real key share, so it recombines
+///   to `k^{-1} * x`;
+/// - `e = 0`, since the zero-sharing mask it carried in the removed scheme has
+///   nothing to mask here.
```
gilcu3, line 224: good one!

gilcu3 (dismissed):
Thank you!

crates/threshold-signatures/src/ecdsa/robust_ecdsa/presign.rs
netrome: Well, credit goes to Claude for this one.

SimonRastikian (commented):
There is one more dangling DamgardEtAl in domain-separation.md. Also one in the Changelog (not sure what to do about this one)

crates/threshold-signatures/src/ecdsa/robust_ecdsa/README.md
netrome: Updated in https://github.com/near/mpc/pull/4338/commits/a33dbf1d7f05b99ba983f66787e47b4f85ae77ad

crates/threshold-signatures/docs/ecdsa/robust_ecdsa/signing.md
netrome: Agreed, updated in https://github.com/near/mpc/pull/4338/commits/a33dbf1d7f05b99ba983f66787e47b4f85ae77ad

crates/threshold-signatures/docs/benches/results.md
netrome: Good point, done in https://github.com/near/mpc/pull/4338/commits/7fda3a369bf8f4a946531eb1a20671669cf3dfc4

netrome:
> There is one more dangling DamgardEtAl in domain-separation.md. Also one in the Changelog (not sure what to do about this one)

Yeah I got rid of the domain-separation.md one in https://github.com/near/mpc/pull/4338/commits/a33dbf1d7f05b99ba983f66787e47b4f85ae77ad

image

For the changelog, I think it's okay. We still have it in the history, so I don't think we should remove it from there. But we could just remove the entry and add the commit as ignored by git-cliff if we want.

gilcu3 (approved):

crates/threshold-signatures/docs/ecdsa/robust_ecdsa/signing.md
```diff
@@ -1,86 +1,64 @@
...
+> stub that produces valid ECDSA signatures while leaking the signing key, kept so
+> that the surrounding node and
+> contract plumbing stays exercised until a real robust scheme replaces it.
+> TODO(#4383): replace the stub with a real robust scheme.
+> Never add a domain for this protocol on mainnet or testnet.
+>
+> The removed specification, including its security analysis of split-view attacks
+> and its differences from the original paper, is available in this file's git
+> history.
+
```
SimonRastikian: I was thinking that maybe instead of providing a valid ECDSA signature we provide an invalid one (and without the key) so that we make sure no one misuses it

SimonRastikian (changes requested):

crates/threshold-signatures/docs/ecdsa/robust_ecdsa/signing.md
gilcu3: Wouldn't that force us to change a lot of tests, that is what we want to avoid?

netrome: Yeah it would kind of defeat the purpose of the stub.

SimonRastikian (approved):
Thank you

PR #4339, opened 2026-09-03
fix: make sure empty keysets fail loudly during put-keyshares

### LINKED ISSUE #4326: put-keyshares with an empty keyshare set silently succeeds
### Background

`backup-cli put-keyshares` exits successfully without transferring anything when the backup home dir holds no keyshares. `KeyshareStorageAdapter::load_keyshares` returns `Ok(vec![])` rather than an error when `$BACKUP_HOME_DIR/key` is absent (`crates/backup-cli/src/adapters/keyshare_storage.rs:41`), and `put_keyshares` (`crates/backup-cli/src/backup.rs:241`) sends that empty vec anyway. The node accepts it, replies `Keyshares received.`, and the import loop skips it on the `!received_keyshares.is_empty()` guard (`crates/node/src/migration_service/onboarding.rs:302`).

Nothing on either side reveals the no-op. The `PUT /set_keyshares` arm (`crates/node/src/migration_service/web/server.rs:230`) has no tracing calls on any of its branches except the closed-channel 500, so a successful put, an empty put, and a `400 Invalid Json or encryption` all produce the same node logs:

```
INFO mpc_node::migration_service::web::server: Handle connection
INFO mpc_node::migration_service::web::authentication: TLS handshake complete, peer authenticated and encrypted channel established.
INFO mpc_node::migration_service::web::server: TLS handshake complete, backup service authenticated and encrypted channel established
INFO mpc_node::migration_service::web::server: connection closed gracefully
```

During a migration this reads as a completed transfer, so the operator moves on and only discovers the failure when onboarding never concludes.

### Acceptance Criteria

- `put-keyshares` fails with a clear error naming the home dir when local storage holds no keyshares, instead of issuing an empty request.
- The `set_keyshares` handler logs the outcome of each request, including the decrypt/deserialize failure branch that currently returns `400` silently.

gilcu3:
Closes #4326

### DIFF
--- a/crates/backup-cli/src/backup.rs
+++ b/crates/backup-cli/src/backup.rs
@@ -61,9 +61,14 @@ pub async fn run_command(args: cli::Args) {
             let (mpc_p2p_client, key_shares_storage) =
                 open_node_client_and_storage(&home_dir, &subcommand_args).await;
 
-            put_keyshares(&mpc_p2p_client, &key_shares_storage, &home_dir)
+            put_keyshares(&mpc_p2p_client, &key_shares_storage)
                 .await
-                .expect("failed to put keyshares");
+                .unwrap_or_else(|err| {
+                    panic!(
+                        "failed to put keyshares from {}: {err:?}",
+                        home_dir.display()
+                    )
+                });
         }
         cli::Command::Run(subcommand_args) => {
             let home_dir = PathBuf::from(args.home_dir);
@@ -245,7 +250,6 @@ pub async fn get_keyshares(
 pub async fn put_keyshares(
     mpc_p2p_client: &impl ports::P2PClient,
     keyshares_storage: &impl ports::KeyShareRepository,
-    home_dir: &Path,
 ) -> anyhow::Result<()> {
     let key_shares = keyshares_storage
         .load_keyshares()
@@ -253,9 +257,8 @@ pub async fn put_keyshares(
         .map_err(|err| anyhow::anyhow!("failed to load keyshares: {err:?}"))?;
     if key_shares.is_empty() {
         anyhow::bail!(
-            "no keyshares in local storage at {}: run `get-keyshares` against a node that holds \
-             them before putting them back",
-            home_dir.display()
+            "no keyshares in local storage: run `get-keyshares` against a node that holds them \
+             before putting them back"
         );
     }
     mpc_p2p_client
@@ -273,7 +276,6 @@ fn verifying_key_from_str(mpc_node_p2p_key: &str) -> VerifyingKey {
 #[cfg(test)]
 #[expect(non_snake_case)]
 mod tests {
-    use std::path::Path;
     use std::sync::atomic::{AtomicUsize, Ordering};
 
     use mpc_node::keyshare::{Keyshare, test_utils::generate_dummy_keyshare};
@@ -335,7 +337,7 @@ mod tests {
         };
 
         // When
-        let result = put_keyshares(&mpc_p2p_client, &storage, Path::new("/backup/home")).await;
+        let result = put_keyshares(&mpc_p2p_client, &storage).await;
 
         // Then
         result.expect("stored keyshares should be pushed to the node");
--- a/crates/backup-cli/src/backup.rs
+++ b/crates/backup-cli/src/backup.rs
@@ -61,7 +61,9 @@ pub async fn run_command(args: cli::Args) {
             let (mpc_p2p_client, key_shares_storage) =
                 open_node_client_and_storage(&home_dir, &subcommand_args).await;
 
-            put_keyshares(&mpc_p2p_client, &key_shares_storage).await;
+            put_keyshares(&mpc_p2p_client, &key_shares_storage, &home_dir)
+                .await
+                .expect("failed to put keyshares");
         }
         cli::Command::Run(subcommand_args) => {
             let home_dir = PathBuf::from(args.home_dir);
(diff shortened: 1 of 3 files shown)


crates/backup-cli/src/backup.rs
```diff
@@ -238,22 +240,123 @@ pub async fn get_keyshares(
         .map_err(|err| anyhow::anyhow!("failed to store keyshares: {err:?}"))
 }
 
+/// Sends the locally stored keyshares to an MPC node. Fails on an empty local store rather than
+/// issuing a request the node would accept and then ignore.
 pub async fn put_keyshares(
     mpc_p2p_client: &impl ports::P2PClient,
     keyshares_storage: &impl ports::KeyShareRepository,
-) {
+    home_dir: &Path,
```
kevindeforth: Feels wrong to pass this argument just so it can be included in the logs.

If we need to display this, then I think it makes more sense to do it at the caller a few lines above.

crates/backup-cli/src/backup.rs
```diff
@@ -61,7 +61,9 @@ pub async fn run_command(args: cli::Args) {
             let (mpc_p2p_client, key_shares_storage) =
                 open_node_client_and_storage(&home_dir, &subcommand_args).await;
 
-            put_keyshares(&mpc_p2p_client, &key_shares_storage).await;
+            put_keyshares(&mpc_p2p_client, &key_shares_storage, &home_dir)
+                .await
+                .expect("failed to put keyshares");
```
kevindeforth: i.e print the home-dir here, instead of passing it in.

crates/backup-cli/src/backup.rs
```diff
@@ -238,22 +240,123 @@ pub async fn get_keyshares(
...
+            unreachable!("put-keyshares never stores")
+        }
+
+        async fn load_keyshares(&self) -> Result<Vec<Keyshare>, Self::Error> {
+            Ok(self.keyshares.clone())
+        }
+    }
+
+    #[tokio::test]
+    async fn put_keyshares__should_fail_naming_the_home_dir_when_local_storage_is_empty() {
```
kevindeforth: I am questioning whether it's worth to maintain this test.
The home directory is set by the user as an argument. Doesn't really matter if we print it or not.

crates/backup-cli/src/backup.rs
```diff
@@ -238,22 +240,123 @@ pub async fn get_keyshares(
...
+             them before putting them back",
+            home_dir.display()
+        );
+    }
     mpc_p2p_client
         .put_keyshares(&key_shares)
         .await
-        .expect("fail to put key shares");
+        .map_err(|err| anyhow::anyhow!("failed to put keyshares: {err:?}"))
 }
```
pbeza, line 268: Wouldn't it be better to push that `key_shares.is_empty()` logic down into `put_keyshares()` in the struct implementing `ports::KeyShareRepository`? Not sure.

pbeza (dismissed):

crates/backup-cli/src/backup.rs
```diff
@@ -238,22 +240,123 @@ pub async fn get_keyshares(
...
+    use near_mpc_contract_interface::types::Keyset;
+    use rand::SeedableRng as _;
+    use rand::rngs::StdRng;
+
+    use super::put_keyshares;
+    use crate::ports::{KeyShareRepository, P2PClient};
+
+    struct FakeP2PClient {
+        put_keyshares_calls: AtomicUsize,
+    }
```
pbeza, line 291: FWIW, we already have a similar struct here:

https://github.com/near/mpc/blob/e26ddd53795818a6561fe18bfe741b748757d5f7/crates/backup-cli/src/service.rs#L233-L240

but given that we don’t need all of its fields, it’s probably better to have a local copy here anyway.

crates/node/src/migration_service/web/server.rs
```diff
@@ -273,24 +254,98 @@ async fn handle_request(
     }
 }
 
+/// Hands received keyshares to the onboarding loop. Every outcome is logged: an operator
+/// watching a migration otherwise cannot tell a transfer from a rejected or empty one.
+fn handle_set_keyshares(body: &[u8], state: &WebServerState) -> Response<Full<Bytes>> {
```
pbeza: It feels a bit odd to have a *sender* (`state.import_keyshares_sender.send()`) on a *state* (`WebServerState`), which sounds kind of read-only, but this naming is pre-existing, so all good.

crates/backup-cli/src/backup.rs
gilcu3: you mean inside `ports::P2PClient` port? Or in `load_keyshares`?

crates/node/src/migration_service/web/server.rs
```diff
@@ -273,24 +254,98 @@ async fn handle_request(
...
+    use rand::rngs::StdRng;
     use serial_test::serial;
-    use tokio::sync::watch;
+    use std::sync::Arc;
+    use tokio::sync::{RwLock, watch};
+    use tracing_test::traced_test;
+
+    const BACKUP_ENCRYPTION_KEY: [u8; 32] = [7u8; 32];
+
+    /// The returned [`TempDir`](tempfile::TempDir) backs the keyshare storage and must outlive it.
```
pbeza: Nit: why not just:

```suggestion
    /// The returned [`tempfile::TempDir`] backs the keyshare storage and must outlive it.
```

crates/node/src/migration_service/web/server.rs
```diff
@@ -354,4 +409,59 @@ mod tests {
...
+        let body = encrypted_body(keyset.keyshares());
+
+        // When
+        let response = handle_set_keyshares(body.as_bytes(), &state);
+
+        // Then
+        assert_eq!(response.status(), StatusCode::OK);
+        assert_eq!(*receiver.borrow_and_update(), keyset.keyshares().to_vec());
+        assert!(logs_contain("set_keyshares accepted keyshares"));
+        assert!(logs_contain("num_keyshares=2"));
```
pbeza: Nit: `2` could probably be reused here:

https://github.com/near/mpc/blob/6ee928391bc60cdc3d18f3ca32cd868c8e200a86/crates/node/src/migration_service/web/server.rs#L419

pbeza (dismissed):

crates/node/src/migration_service/web/server.rs
```diff
@@ -231,38 +231,19 @@ async fn handle_request(
...
-                    }
+                Ok(bytes) => Ok(handle_set_keyshares(&bytes, &state)),
+                Err(err) => {
+                    tracing::error!(?err, "set_keyshares: failed to read body");
+                    Ok(Response::builder()
+                        .status(StatusCode::BAD_REQUEST)
+                        .body(Full::new(Bytes::from(format!(
+                            "Failed to read body: {err}"
+                        ))))
+                        .unwrap())
```
kevindeforth: What I liked about the exisitng function is that it was clear how errors resulted in different responses. I don't find the new code an improvement in that regard. I don't mind moving some logic to functions to help make this terser, but maybe something like the following would be better, as it doesn't require the reader to jump between methods:

```rust
/// this probably won't compile, take it as pseudocode.
let Ok(bytes) = whole_body else {
    tracing::error!("set_keyshares: failed to read body");
    return Ok(build_bad_request_response("failed to read body"));
};

let Ok(keyshares) = decrypt_and_deserializea_keyshares(body, &state.backup_encryption_key) else {
     tracing::error!("set_keyshares rejected: invalid json or encryption");
    return Ok(build_bad_request_response("invalid json or encryption")
};

if keyshares.is_empty() {
        tracing::warn!("set_keyshares received no keyshares,");
        return Ok(build_success_response());
}
if state.import_keyshares_sender.send(new_keyshares).is_err() {
    let msg = "keyshares receiver channel is closed".to_string();
    tracing::error!(msg);
    return Ok(build_internal_server_error_response(msg);
}

Ok(build_success_response())
```

crates/backup-cli/src/backup.rs
pbeza: > you mean inside `ports::P2PClient` port? Or in `load_keyshares`?

I was thinking about the `ports::P2PClient` port, but maybe `load_keyshares` would make even more sense. 🤔

Either way, we’d save a couple of LoC in this file and make it a bit more flexible, since different `impl`ementations could handle an empty set differently; though I’m not sure we really need that flexibility. No strong opinion.

crates/node/src/migration_service/web/server.rs
```diff
@@ -273,24 +254,98 @@ async fn handle_request(
     }
 }
 
+/// Hands received keyshares to the onboarding loop. Every outcome is logged: an operator
+/// watching a migration otherwise cannot tell a transfer from a rejected or empty one.
```
kevindeforth: nit: I don't think we need to justify logging.

crates/node/src/migration_service/web/server.rs
gilcu3: Claude's answer to this one is interesting:

```
  Origin of the name. WebServerState is not original to the migration service. It first appears in
  crates/node/src/web.rs:62 (introduced back in #58, "Separate keygen CLI, graceful shutdown, integration test"),
  where it is axum router state: it's constructed in .with_state(WebServerState { … }) at web.rs:365 and pulled back
  out by every handler via axum's State<WebServerState> extractor (web.rs:185, :192, :222, …). "State" there is axum's
  own vocabulary for the shared per-router handler context, not a claim of read-only-ness — and that struct already
  carries a sender for exactly the same reason yours does: debug_request_sender: broadcast::Sender<DebugRequest>
  (web.rs:64-65).

  The migration service's copy (crates/node/src/migration_service/web/types.rs:14, added in #1283 "migration service
  web server and client logic") took the name from that sibling module. But its server is hand-rolled on hyper +
  service_fn (server.rs:10, no axum), so it never uses the State extractor that justified the name — the struct is
  just an Arc'd bag of dependencies threaded manually through start_web_server → handle_connection → handle_request →
  handle_set_keyshares.

  So the naming oddity is real and inherited: it's an axum idiom applied to a non-axum server. Something like
  WebServerContext or MigrationServiceDeps would describe it better. I agree with the reviewer that it's out of scope
  here — renaming touches types.rs, server.rs, test_utils.rs, and migration_service.rs, and per the repo's "don't
  rename pre-existing names a PR would otherwise leave alone" rule it belongs in its own change.
```

crates/node/src/migration_service/web/server.rs
```diff
@@ -273,24 +254,98 @@ async fn handle_request(
...
+            return Response::builder()
+                .status(StatusCode::BAD_REQUEST)
+                .body(Full::new(Bytes::from(format!(
+                    "Invalid Json or encryption: {err}"
+                ))))
+                .unwrap();
+        }
+    };
+
+    // The onboarding loop skips an empty set, so accepting one imports nothing.
```
kevindeforth: This comment documents code in a different file and is against our engineering standards.

crates/node/src/migration_service/web/server.rs
```diff
@@ -354,4 +409,59 @@ mod tests {
...
+
+        // When
+        let response = handle_set_keyshares(b"not encrypted keyshares", &state);
+
+        // Then
+        assert_eq!(response.status(), StatusCode::BAD_REQUEST);
+        assert!(logs_contain(
+            "set_keyshares rejected: invalid json or encryption"
+        ));
+    }
```
kevindeforth: nit: the test names suggest the intention of the tests is to assert specific logs are written, yet, I would argue they are more useful in asserting certain responses are sent to the client. I would consider renaming them.

kevindeforth (commented):
Thank you. The added tests in server.rs are nice (that is, if we don't have existing coverage for that), but it feels wrong for `put_keyshares` to take as input both `home_dir` and `keyshare_storage`

The refactor in server.rs is not an improvement IMO, left a suggestion there.

crates/node/src/migration_service/web/server.rs
pbeza: Oh, I actually double-checked if that type (`WebServerState`) was from `axum`, because it sounded like it might be (it wasn't). Now it makes more sense (I mean, it'd still make more sense if it were renamed). Thanks for checking!

crates/node/src/migration_service/web/server.rs
gilcu3: this is slightly less testable, but that's fine

crates/node/src/migration_service/web/server.rs
gilcu3: yeah I think mostly we avoid adding tests for logs. In this change it was a bit borderline because the change itself was about logs... But yeah rewriting these so that the tests are log free

crates/backup-cli/src/backup.rs
pbeza: I think we could even add a CI lint (or a recommendation in `AGENTS.md`): whenever we have a param that's only used for printing/logging, that's a red flag the code could be structured better.

Something along these lines:
> If a function parameter is used only inside logging/print statements and nowhere else in the function's logic, treat that as a signal the parameter shouldn't be threaded through the signature — prefer a scoped/contextual logger, a tracing span, or a context object created once by the caller.

crates/node/src/migration_service/web/server.rs
kevindeforth: yeah, thinking about it, we don't really need an error path in here.
We could also move this logic to a helper method `process_set_keyshare_request` that then returns an enum and builds the responses from it.

gilcu3: ohh no, as I suspected these tests only became possible because the function was separated out from the server, so I will need to keep it there to avoid losing test coverage

crates/node/src/migration_service/web/server.rs
kevindeforth: Btw, I don't mind to test the log _in addition_ to testing functional logic. In this case, I think it's fine if the tests keep testing the logs, but I think their main contribution is that they assert that the status code matches its expected value.

crates/node/src/migration_service/web/server.rs
gilcu3: doing as you mentioned in slack

> we can also move the entire section to a separate method. I.e. handle_set_keysare(body) -> Result<hyper::Response<Full<Bytes>>, Infallible>

crates/node/src/migration_service/web/server.rs
gilcu3: Nah I think we were trying to avoid testing logs because logs can easily drift, so removed those

crates/backup-cli/src/backup.rs
gilcu3: Probably the agents line will be faster/easier (not in this PR though, I just pushed changes)

crates/node/src/migration_service/web/server.rs
```diff
@@ -254,23 +239,32 @@ async fn handle_request(
...
 
-/// Hands received keyshares to the onboarding loop. Every outcome is logged: an operator
-/// watching a migration otherwise cannot tell a transfer from a rejected or empty one.
-fn handle_set_keyshares(body: &[u8], state: &WebServerState) -> Response<Full<Bytes>> {
-    let keyshares = match decrypt_and_deserialize_keyshares(body, &state.backup_encryption_key) {
+/// Hands received keyshares to the onboarding loop, logging each outcome.
+async fn handle_set_keyshares<B>(
+    body: B,
+    state: &WebServerState,
+) -> Result<Response<Full<Bytes>>, Infallible>
```
kevindeforth: nit: It seems like we don't need an error path for this method.

kevindeforth (dismissed):
Thank you!

crates/node/src/migration_service/web/server.rs
gilcu3: true, I thought it was needed for some reason but it is not

kevindeforth (approved):

pbeza (approved):

PR #4341, opened 2026-09-03
fix: warn about mismatched tls keys during migration

### LINKED ISSUE #4327: Node silently ignores a registered destination node whose TLS key is not its own
### Background

`infer_migration_status` (`crates/node/src/migration_service/types.rs:126`) sets `active_migration = false` without logging when a destination node is registered for this account whose `tls_public_key` is parseable but not this node's key. `OnboardingJob::new` then yields `WaitForStateChange` instead of `Onboard` (`types.rs:84`), so the import loop never runs. A `tls_public_key` off by one base58 character still decodes to a valid 32-byte key, so nothing upstream rejects it either.

The migration web server keeps serving throughout, since `ExpectedPeerInfo::from_migration` (`crates/node/src/migration_service/web/types.rs:27`) gates only on `backup_service_info`. `put-keyshares` then transfers the keyshares, gets `Keyshares received.`, and exits successfully, while the node drops them into a watch channel with no consumer.

### Acceptance Criteria

- The node logs a warning naming both keys when the registered destination TLS key differs from its own.

gilcu3:
Closes #4327

### DIFF
--- a/crates/node/src/indexer/migrations.rs
+++ b/crates/node/src/indexer/migrations.rs
@@ -41,8 +41,9 @@ pub async fn monitor_migrations(
                 let response = fetch_migrations_once(indexer_state.clone()).await;
                 tracing::debug!(target: "indexer", "fetched mpc migration state at block {}: {:?}", response.0, response.1);
 
+                let mut migration_info_changed = false;
                 migration_state_sender.send_if_modified(|watched_state| {
-                    let migration_info_changed = watched_state.1 != response.1;
+                    migration_info_changed = watched_state.1 != response.1;
                     if *watched_state != response {
                         if migration_info_changed {
                             tracing::info!("contract migration state changed: {:?}", response);
@@ -53,6 +54,12 @@ pub async fn monitor_migrations(
                         false
                     }
                 });
+
+                // Re-deriving an unchanged view would repeat its diagnostics on every tick.
+                if !migration_info_changed {
+                    continue;
+                }
+
                 let my_migration_state = MigrationInfo::from_contract_state(
                     &my_near_account_id,
                     &my_p2p_public_key,
--- a/crates/node/src/migration_service/types.rs
+++ b/crates/node/src/migration_service/types.rs
@@ -137,7 +137,7 @@ fn infer_migration_status(
             tracing::warn!(
                 registered_tls_public_key = %String::from(registered_key),
                 my_tls_public_key = %String::from(&Ed25519PublicKey::from(my_p2p_public_key)),
-                "registered destination node TLS key is not this node's key, migration stays inactive"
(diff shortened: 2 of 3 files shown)


crates/node/src/indexer/migrations.rs
```diff
@@ -53,6 +54,12 @@ pub async fn monitor_migrations(
                         false
                     }
                 });
+
+                // Re-deriving an unchanged view would repeat its diagnostics on every tick.
+                if !migration_info_changed {
+                    continue;
+                }
```
gilcu3, line 61: this is actually necessary to fix a latent bug happening during the `MigrationInfo::from_contract_state`. Already before this change, that function would log on key errors (after this PR one more log is added). That was already causing flooding in the logs, which is what justifies this small change.

crates/node/src/migration_service/types.rs
```diff
@@ -125,14 +127,29 @@ fn infer_migration_status(
...
+                "registered destination node TLS key is not this node's key, migration stays inactive"
+            );
+            false
+        }
+        Err(err) => {
+            tracing::error!(
+                ?err,
+                registered_tls_public_key = %String::from(registered_key),
+                "could not parse the registered destination node's TLS public key from chain"
+            );
```
gilcu3, line 149: this was a warning before, but it does look like an error to me, so I promoted the log. It is not "unrecoverable", but it means something is really wrong, so it feels much worse than a warning.

crates/node/src/indexer/migrations.rs
```diff
@@ -41,8 +41,9 @@ pub async fn monitor_migrations(
                 let response = fetch_migrations_once(indexer_state.clone()).await;
                 tracing::debug!(target: "indexer", "fetched mpc migration state at block {}: {:?}", response.0, response.1);
 
+                let mut migration_info_changed = false;
                 migration_state_sender.send_if_modified(|watched_state| {
-                    let migration_info_changed = watched_state.1 != response.1;
+                    migration_info_changed = watched_state.1 != response.1;
                     if *watched_state != response {
```
netrome, line 47: Hmm not a fan of how this is expressed. Couldn't we do the modification check separately, and then just call `.send()` if the info changed. This would simplify the code quite a bit I think

crates/node/src/migration_service/types.rs
netrome: Yeah sounds pretty bad. Error seems appropriate.

netrome (approved):
Thanks, one code nit.

anodar (approved):

crates/node/src/indexer/migrations.rs
gilcu3: I don't really know how to do it while still keeping it atomic

netrome: Right sorry, bad suggestion.

Issue #4343, opened 2026-09-03
epoch-sync auto-recovery: bound repeated data-reset wipes


barakeinav1:
### Background

The #3909 fix (PR #4312) auto-recovers a node stuck past the epoch-sync horizon: nearcore requests an `EpochSyncDataReset`, the node records a `.epoch_sync_data_reset` marker, exits non-zero, and the startup wipe removes the chain store so the node re-syncs.

After a **successful** wipe the marker is cleared and `restart_disabled` is `None` again. If the underlying cause is persistent — a genesis/config mismatch, or a chain state nearcore cannot epoch-sync past — the node loops: wipe → full state re-sync → reset → wipe → … with no counter and no ceiling. Each cycle costs a full state re-sync, so the node is effectively offline while silently churning.

The wipe-failure and archival cases are already handled (the node wedges with an actionable error); this issue is specifically the **successful-wipe-but-recurring-reset** loop.

### User story

As an operator, I want a node that repeatedly cannot epoch-sync to stop wiping-and-resyncing indefinitely and instead wedge with a clear, greppable error, so it does not burn bandwidth/CPU in a silent loop I have to notice from the outside.

### Acceptance criteria

- Bound the number of consecutive epoch-sync auto-wipes (the marker file already carries a payload byte — `b"1"` — that can hold an attempt count).
- After N consecutive resets without the node reaching sustained sync, stop auto-wiping and surface an actionable error (same shape as the wipe-failed wedge).
- Reset the counter once the node syncs and stays synced, so a one-off recovery does not count against a later, unrelated one.

### Resources

- PR #4312 (the #3909 fix)
- `crates/node/src/indexer/real.rs` — `restart_disabled` computation / `spawn_epoch_sync_reset_handler`
- `crates/node/src/indexer/near_data_wipe.rs` — `record_epoch_sync_reset_request` / `wipe_near_data_if_epoch_sync_reset`

Issue #4314, opened 2026-09-01
E2E integration test: epoch-sync stale-node auto-recovery (#3909)


barakeinav1:
### Background

#3909 fixed the epoch-sync wedge: a node offline long enough to fall past the epoch-sync horizon used to loop forever (nearcore requested a data reset that went nowhere because mpc-node started neard via `start_with_config`, which passes `shutdown_signal = None`). The fix (#4312) wires `start_with_config_and_synchronization`, and on `EpochSyncDataReset` records a marker + exits so the container's `restart: on-failure` restarts it and wipes the chain store on startup, yielding a clean state-sync — automatic, keyshares preserved.

That fix is covered by **unit tests** (marker record + wipe-on-restart + keyshare preservation, in `crates/node/src/indexer/near_data_wipe.rs`) and was validated by a **manual testnet e2e** (a real TDX node stopped for days, then observed to auto-recover). What we do **not** have is an automated integration test in CI that exercises the *full* path against a real `neard`: node genuinely enters epoch-sync → `EpochSyncDataReset` fires → marker written + process exits → restart → store wiped → re-syncs to head.

### User Story

As an mpc-node engineer, I want an automated integration test that drives a real node through the epoch-sync wedge and asserts it auto-recovers, so the #3909 fix is protected against regressions (in mpc-node *or* in a future nearcore bump) without relying on a manual multi-day testnet run.

### Acceptance Criteria

- [ ] A `crates/e2e-tests` test that puts a node's neard genuinely into **epoch-sync** (not state-sync/block-sync): kill a node, advance the sandbox chain past the epoch-sync horizon, then restart it.
- [ ] Assert the node **auto-recovers** rather than wedges: it logs `wiped nearcore data dir (epoch-sync data reset)`, does a fresh state-sync, and reaches head (indexer follows tip); keyshares/secrets survive (node rejoins as the same participant).
- [ ] Assert the **negative direction is meaningful** — i.e. the test would fail (wedge: head frozen, `near_sync_status = EpochSync`) if the shutdown signal were not wired.
- [ ] Test runs in CI in a reasonable time (tune `epoch_length` small so "past the horizon" is a handful of blocks, not real epochs).

### Resources & Additional Notes

- Fix PR: #4312 · Issue: #3909
[...]

PR #4346, opened 2026-09-03
fix(node): auto-recover from the epoch-sync wedge (#3909)

### LINKED ISSUE #3909: Stale node (offline ~days) wedges in neard epoch-sync loop; never reaches head

barakeinav1:
## What

Fixes #3909 — a node stuck far behind past the epoch-sync horizon now recovers **automatically** instead of wedging forever.

## The bug

`mpc-node` embeds nearcore, and the start path left `shutdown_signal = None`. When a node is offline long enough to fall past the epoch-sync horizon (~a day on testnet), nearcore's epoch-sync asks for a data reset through that signal — but it was dropped, so the node looped forever with a frozen head (`near_sync_status=2`). The only recovery was an operator bumping `wipe_near_data_token` by hand.

## The fix

- Start nearcore via `start_with_config_and_synchronization` with a real `broadcast::Sender<ShutdownReason>`.
- On `EpochSyncDataReset`: write a `.epoch_sync_data_reset` marker (a `home_dir` sibling of `data`, so it survives the wipe) and `exit(70)` → the container's `restart: on-failure` restarts it.
- On the next startup, wipe the chain store before nearcore reopens it (mirroring neard's `check_epoch_sync_data_reset_marker`), then re-sync. Reuses the `wipe_near_data_token` atomic rename-into-trash; keyshares/secrets/config are untouched.
- Disarmed for archival nodes and when the startup wipe fails, so it can only wedge, never restart-loop; the restart is requested only after the marker write succeeds.

## Testing

- Unit tests cover the handler decisions, the marker record/wipe, the path guard, and the reason helper.
- Live testnet e2e: **in progress** — capturing the wedge → recovery on a stale node built from this branch; I'll paste the log here.

## Follow-ups

- #4314 — full epoch-sync integration test.
- #4343 — bound repeated resets after a successful wipe.

<sub>Supersedes #4312 (same change, squashed to drop the long review history). Thanks to @metalurgical for the review that shaped the wipe-helper sharing and reset handling.</sub>


barakeinav1:
Thanks — addressed most of these in ea1f38c4; keeping the single-`recv` deliberately (reasoning below).

**Blocking**

> the single line the whole fix hinges on (`Some(shutdown_tx)`) has no test … inject the terminal action

Agree, and good catch — that was the real gap. `spawn_epoch_sync_reset_handler` now takes `on_restart: impl FnOnce()` (the call site passes `|| process::exit(70)`), and a new test sends `EpochSyncDataReset` on the returned sender and asserts the restart action fired and the marker landed. Reverting `Some(shutdown_tx)` → `None` now fails it. The live e2e is still owed and will land here.

**Non-blocking**

> single `recv()` … rests on an upstream invariant … a `loop` costs nothing and removes both concerns

Keeping the single `recv()` — the loop doesn't actually remove the dependency. Verified in nearcore 2.13.3: every `ShutdownReason` (both `EpochSyncDataReset` and `ExpectedShutdown`) is sent through the **ClientActor's one** `shutdown_signal: Option<Sender>`; each send site does `self.shutdown_signal.take()`, so the first to fire consumes the sole sender and the rest get `None` and send nothing. So **at most one message is ever delivered** — a loop has no second message to catch, and `Lagged` (which needs >16 sends) is unreachable. The failure the loop is meant to guard (an `ExpectedShutdown` arriving first) means the reset was never *sent*, which no loop recovers. If a future nearcore split the sender into independent send sites, that's a behavioural change our `bump-nearcore` step (and the #4314 integration test) would have to re-verify anyway — a loop would mask it, not handle it. The legit half — the comment over-claiming — is fixed: it now names the single `shutdown_signal`/`take()` invariant.

> pinning `2.13.3 …/client_actor.rs` in the comment will rot

Agree — dropped the exact file path, kept the behavioural claim.

> "can only wedge … never restart-loop" contradicts `TODO(#4343)`

Agree — scoped that sentence to the `restart_disabled` cases; the successful-wipe repeat is what #4343 tracks.

> `record_epoch_sync_reset_request` bare `fs::write` vs `write_last_token`'s fsync

Leaving as-is. The asymmetry is intentional: the token drives a wipe-once decision and must be durable, whereas the marker is existence-only and self-heals (nearcore re-signals on the next attempt), so it needs neither fsync nor atomicity. A resurrected marker after a hard crash costs one idempotent re-sync, not a correctness bug — and an earlier round deliberately removed the comment spelling this out as noise, so I'd rather not re-add it.

> two comments narrate the callee (`wipe_store_dir re-checks …`)

Agree — trimmed both to the "why".

> `restart_disabled_reason` test lacks Given/When/Then

Agree — added.

> the runbook note … exit code 70 / archival caveat / marker filename … and says it twice

Dropped the duplicated line. Leaving out the exit code, marker filename, and archival caveat deliberately — that note is kept terse and operator-facing (archival isn't a supported config here), a call made in earlier rounds.

barakeinav1:
@claude review

barakeinav1:
Rebased onto latest `main` (which bumped nearcore 2.13.3 → 2.13.4). I re-verified the invariant the fix rests on still holds at 2.13.4: `ClientActor` has one `shutdown_signal: Option<Sender>` (`client_actor.rs:336`), `epoch.rs:660` is inside `impl Handler<EpochSyncResponseMessage> for ClientActor`, and all three `take()` sites share it — unchanged from 2.13.3.

**Blocking**

> the new spawn test does *not* fail if `Some(shutdown_tx)` is reverted to `None` … it exercises the sender the handler returns, not the one nearcore receives

You're right, and thanks for the correction — my earlier "reverting it fails a test" was wrong. The test covers the handler wiring (sender → task → marker → restart action), not the nearcore hand-off, which needs the running node in #4314. The live-testnet capture is the remaining evidence and is owed before merge — it's in progress on a node built from this branch; I'll post the log here.

**Non-blocking**

> `marker.exists()` maps any stat error to "no reset pending" … use `std::fs::exists()`

Leaving as-is. Propagating a stat error would set `reset_wipe_failed` and disarm the restart for the whole process, so a *transient* blip (EIO) becomes a sticky wedge with no auto-restart; the current swallow self-heals — a transient error recovers on the next restart cycle, and a persistent one still surfaces as the logged wedge when the marker write to the same path fails. So the change trades a self-healing blip for a stuck node.

> from outside, a recovery is indistinguishable from a crash-loop — a scrapeable counter would help

Agree it's worth having, filed as **#4430** (a `mpc_epoch_sync_data_reset_wipes_total` counter bumped where the wipe runs, so it's visible for the whole re-sync window). Keeping it out of this PR to hold the diff to the fix.

> the note promises "no operator action" but two states deliberately don't recover

Leaving the note as-is. The give-up branch was declined a few times already for terseness, and "use the manual wipe below" isn't actually a valid remedy for one of the two states — the marker-write failure is a read-only/full `/data`, where the manual wipe writes to the same `/data` and fails too. The states are logged; documenting each failure branch is the operator-guide detail we've deliberately kept out.

PR #4348, opened 2026-09-03
feat(tee): name TCB component statuses and advisory IDs in attestation rejection reasons

### LINKED ISSUE #4176: feat(tee): return attestation verification reasons as JSON instead of Borsh
### Background

When an attestation is rejected, the reason is only legible after decoding Borsh by hand. On [this submission](https://nearblocks.io/txns/GLxAnCsLkWaMDtRpLVXmwvpQwUx76bjsLfLtErbDY6QJ/execution) `v1.signer` panicked with `TcbStatusNotUpToDate("OutOfDate")`, and finding out *why* meant base64-decoding the verifier's 777-byte `verify_quote` return value manually. It carried `advisory_ids: ["INTEL-SA-01192", "INTEL-SA-01245", "INTEL-SA-01312", "INTEL-SA-01313"]`, which appear nowhere readable in the transaction. We hit this while investigating #4169 and #4170.

- `verify_quote` is `#[result_serializer(borsh)]` (`crates/tee-verifier/src/lib.rs:60`)
- `fail_attestation_submission` takes `#[serializer(borsh)] reason: String` (`crates/contract/src/lib.rs:1912`), so even the plain-text reason is an opaque blob
- the advisory IDs are dropped by `verify_tcb_status` (`crates/attestation/src/attestation.rs:266`), and `AttestationSubmissionError::InvalidAttestation` (`crates/contract/src/tee/tee_state.rs:41`) formats the error with `{:?}` rather than its `Display`

### Acceptance Criteria

- [ ] `fail_attestation_submission` takes its reason as JSON, so the failing receipt is readable on an explorer
- [ ] The reason names the advisory IDs and QE/platform TCB statuses, not just `TcbStatusNotUpToDate("OutOfDate")`
- [ ] `verify_quote` returns `VerificationResult` as JSON, or we record why it stays Borsh; the argument side stays Borsh either way, since quote plus collateral is ~32 KB

### Resources &amp; Additional Notes

- `tee-verifier-2026-08-04.near` is locked, so changing the `verify_quote` wire format needs a new verifier deployment and a vote to repoint `v1.signer`
- Related: #4169, #4170

pbeza:
`VerificationError::TcbStatusNotUpToDate` now carries the overall status, the QE and platform statuses and the advisory IDs, and its message names all of them, so a rejected submission reads

```
TCB status OutOfDate is not up to date (QE: UpToDate, platform: OutOfDate, advisories: ["INTEL-SA-01192", ...])
```

instead of

```
TcbStatusNotUpToDate("OutOfDate")`
```

Closes #4176


gilcu3 (approved):
Thank you!

anodar (approved):
Nice!

PR #4363, opened 2026-09-04
refactor(contract): drop `Option` from `tee_verifier_account_id`

### LINKED ISSUE #3639: Make `tee_verifier_account_id` non-`Option` once a verifier is voted in
### Background

`MpcContract.tee_verifier_account_id` is currently `Option<AccountId>`: it is `None` until participants vote a trusted verifier contract in via `vote_tee_verifier_change`, and `Some(account)` thereafter. The optionality models the genuine "not yet configured" window between deploy/migration and the first successful vote.

Once every live network has voted a verifier in, that window is closed permanently, and the field should become a plain `AccountId` so the always-configured invariant is enforced by the type system rather than carried as an `Option` that every consumer (notably the future DCAP dispatch path) must keep handling. The DCAP dispatch site that will consume this field is tracked separately in #3264.

### Acceptance Criteria

- [ ] A future contract version declares `tee_verifier_account_id: AccountId` (no `Option`).
- [ ] The state migration converts the previous version's `Option<AccountId>` to `AccountId` via a fallible path (`TryFrom` / `try_into`), returning a migration error when the value is still `None` so the upgrade is rejected (fails closed) rather than silently carrying an unconfigured value forward.
- [ ] `vote_tee_verifier_change` and the DCAP dispatch site drop all `Option`/`None` handling for this field.
- [ ] Borsh-schema and ABI snapshots regenerated.
- [ ] A migration test asserts: a `Some(account)` old state migrates to the plain field holding `account`, and a `None` old state fails migration cleanly.

Precondition: ship this only after confirming on-chain that every live network (mainnet/testnet) has a configured verifier. If deployed to a network still at `None`, `migrate()` returns an error and the upgrade is rejected, leaving the old state intact; vote a verifier in and re-run.

Caveat: this gate is effectively irreversible. After collapsing to `AccountId` there is no representation for "unconfigured", so the verifier can only be rotated to a different account, never un-set. This is the intended invariant but should be a deliberate decision before shipping.

Choosing `Option<AccountId>` over a sentinel value was a deliberate setup for this gate: with `Option`, the fallible conversion cannot compile without resolving the `None` case, so the check cannot be forgotten.

pbeza:
Closes #3639


pbeza:
@claude review

crates/devnet/src/cli.rs
```diff
@@ -238,6 +238,9 @@ pub struct MpcInitContractCmd {
     /// The threshold to initialize with.
     #[clap(long)]
     pub threshold: u64,
+    /// The TEE verifier contract account the MPC contract trusts.
+    #[clap(long)]
+    pub tee_verifier_account_id: AccountId,
```
netrome, line 243: Good that you remembered the devnet change. Does this imply we may need to change something in the infra-ops repo as well? (Don't know at the top of my mind)

netrome (dismissed):
Thank you 🙏

crates/contract/src/errors.rs
```diff
@@ -30,7 +30,7 @@ pub enum TeeError {
     )]
     TeeValidationFailed,
     #[error(
-        "No TEE verifier is configured yet. Participants must vote one in via vote_tee_verifier_change before Dstack attestations can be submitted."
+        "No TEE verifier is configured. Participants must vote one in via vote_tee_verifier_change before upgrading."
     )]
     VerifierNotConfigured,
```
gilcu3: I was a bit surprised to see this one. Maybe better to define it in the migration file so that it is automatically deleted after the next contract migration

crates/contract/src/v3_15_1_state.rs
```diff
@@ -118,10 +121,12 @@ impl From<MpcContract> for crate::MpcContract {
             accept_requests: old.accept_requests,
             node_migrations: old.node_migrations,
             foreign_chains: old.foreign_chains,
-            tee_verifier_account_id: old.tee_verifier_account_id,
+            tee_verifier_account_id: old
+                .tee_verifier_account_id
+                .ok_or(TeeError::VerifierNotConfigured)?,
```
gilcu3: here I think it is better to follow the current pattern of panicking instead of returning error, makes the code much simpler

crates/contract/tests/sandbox/tee_verifier.rs
```diff
@@ -54,19 +54,9 @@ async fn setup() -> SandboxTestSetup {
...
-                .vote_tee_verifier_change(verifier, expected_code_hash)
-                .await
-        }
-    })
-    .await
-    .unwrap();
+    vote_tee_verifier_change(&setup.mpc_signer_accounts, &setup.contract, verifier)
+        .await
+        .unwrap();
 }
```
gilcu3: why do we need the `trust_verifier` function if we now have `vote_tee_verifier_change` ? Shouldn't one of them disappear?

crates/contract/tests/sandbox/tee_verifier.rs
```diff
@@ -241,48 +231,20 @@ async fn assert_only_gas_spent(
...
-        "Smart contract panicked: {}",
-        TeeError::VerifierNotConfigured
-    );
-    assert!(
-        err.contains(&expected_panic),
-        "expected {expected_panic:?}, got: {err}"
+    assert_eq!(
+        tee_verifier_account_id(&setup.contract).await,
+        bogus_tee_verifier_account_id()
     );
```
gilcu3: I would remove this assertion, as it seems a bit superfluous now. As we don't pass the tee verifier account id in setup, the test needs to "guess" the value.

scripts/launch-localnet.sh
```diff
@@ -201,6 +201,7 @@ EOF
 
   JSON_RESULT=$(jq -n --arg threshold "$THRESHOLD" --arg next_id "$N" '
   {
+    tee_verifier_account_id: "tee-verifier.test.near",
```
gilcu3: nit: better to define this as a constant at the top of the file

docs/development/testnet/setup-guide-for-testnet-with-tee-support.md
```diff
@@ -301,6 +307,7 @@ Example:
 
 ```json
 {
+  "tee_verifier_account_id": "barak_tee_test1_verifier.testnet",
```
gilcu3: ups, surely not introduced in this PR, but we should probably cleanup dev names from this repo at some point.

gilcu3 (dismissed):
Thank you!

Left some minor comments

crates/devnet/src/cli.rs
pbeza: I'm not super familiar with the https://github.com/Near-One/infra-ops repo, but from what I double-checked with Claude, we don't need to make any changes because nothing about the contract's `init` reaches it.

`infra-ops` just describes the cluster: the machines and the mpc-node jobs. devnet [runs terraform against it](https://github.com/near/mpc/blob/1510b134455a8572df4ef2e3fd4a160bbd7999db/crates/devnet/src/terraform.rs#L372-L418) and passes a generated `.tfvars.json` plus a few flags, and the only contract-related value in any of it is the contract's account id, so each node knows which contract to index. The verifier is an argument to a separate
[`init` RPC call](https://github.com/near/mpc/blob/1510b134455a8572df4ef2e3fd4a160bbd7999db/crates/devnet/src/mpc.rs#L389) the CLI makes, which never goes through terraform.

crates/contract/src/errors.rs
pbeza: I removed that error type altogether, as suggested in the related comment: https://github.com/near/mpc/pull/4363#discussion_r4062193468. Addressed in https://github.com/near/mpc/pull/4363/commits/d6ef575eff6ff83d84b533b2a4093c0bc291f1ef. ✅

crates/contract/src/v3_15_1_state.rs
pbeza: Addressed in https://github.com/near/mpc/pull/4363/commits/d6ef575eff6ff83d84b533b2a4093c0bc291f1ef. ✅

crates/contract/tests/sandbox/tee_verifier.rs
pbeza: Oh, good catch. I actually remembered to change that before lunch after reviewing Claude’s output, but then got pulled into something else and completely forgot about it. Thanks for catching it!

Fixed here: https://github.com/near/mpc/pull/4363/commits/4e1ac5a508ae5cced3167f80f5c03249359d7167 ✅

crates/contract/tests/sandbox/tee_verifier.rs
pbeza: https://github.com/near/mpc/pull/4363/commits/03e9715640a68754e5c30685c05829676535e67f ✅

docs/development/testnet/setup-guide-for-testnet-with-tee-support.md
pbeza: Good catch. Fixed in https://github.com/near/mpc/pull/4363/commits/d376a5be56dc2ad8c4484bd7ddc9d6d5f7f6a64d and created a follow-up issue: https://github.com/near/mpc/issues/4493. ✅

scripts/launch-localnet.sh
pbeza: https://github.com/near/mpc/pull/4363/changes/2e87d88a7b9faf93a51dd10069cdd8406423c65f ✅

gilcu3 (approved):

netrome (approved):

PR #4310, opened 2026-08-31
feat(inspector): split foreign chain inspection verdicts from inspector errors

haiyuechen-nearone:
Closes #4309

Inspector `extract` now returns a `Verdict`: the extracted values, or a final answer ruling the transaction out. Errors cover only the failures to reach a verdict.

Reviewer notes:

- With `Verdict` in place, `FanOut` do not need transient versus substantive predicates. Reached verdicts must be identical, and provider errors are tolerated.

- A provider fault (a 401, a malformed answer) no longer vetoes a verify its peers agree on. Previously a rejected or malformed response forced a mismatch even when the others agreed.

- Absence of transaction is deliberately a verdict. If one provider has not indexed the transaction yet while others reports presence, request attempt fails with mismatch.

- Failing verdicts now compare field for field instead of by kind alone, so two `NonCanonicalBlock`s that name different canonical hashes count as a mismatch.

- Client errors classification is now unified at the client boundary. This removes the need for `ForeignChainInspectionError::ClientError` , the `HasAbsenceMeaning` trait

-  Starknet's `Received` finality now maps to `NotFinalized` instead of a client error.

- A canonical block lookup answering from a different height is now treated as provider error and is tolerated. 

- `ForeignChainInspectionError::EventLogFailedBorshSerialization`  was unused and removed.

PR #4366, opened 2026-09-04
test(inspector): cover the not found verdict in the fan out mismatch test

haiyuechen-nearone:
[NIB-590](https://github.com/near/mpc-private/blob/main/projects/near-intents-bug-bounty/findings/valid/NIB-590.md) was addressed by https://github.com/near/mpc/pull/4310.

To further guard against this issue, this PR adds test case and comments.

### DIFF
--- a/crates/foreign-chain-inspector/src/lib.rs
+++ b/crates/foreign-chain-inspector/src/lib.rs
@@ -48,8 +48,8 @@ pub trait ForeignChainInspector {
 
 /// The settled answer from a transaction inspection. Inspection producing a negative verdict is
 /// still a successful inspection. Failing to obtain any verdict is instead a
-/// [`ForeignChainInspectionError`]. [`ForeignChainInspector::extract`] must report every answer
-/// a provider gives about the transaction under the variant defined here, not as an error.
+/// [`ForeignChainInspectionError`]. [`ForeignChainInspector::extract`] must report every verdict
+/// given by a provider about a transaction under a variant defined here, not as an error.
 #[derive(Debug, Clone, PartialEq, Eq, Display)]
 pub enum Verdict<V> {
     #[display("extracted {} values", _0.len())]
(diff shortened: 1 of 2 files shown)


crates/foreign-chain-inspector/src/lib.rs
```diff
@@ -38,6 +38,12 @@ pub trait ForeignChainInspector {
     type Finality;
     type Extractor;
     type ExtractedValue;
+
+    /// Inspects transactions on one foreign chain.
+    ///
+    /// Implementations must map the chain's signal for an unknown transaction to
+    /// [`Verdict::TransactionNotFound`] and pin that with a test. If left as a client error,
+    /// absence is tolerated by [`FanOut`] as a transient issue at a given provider.
     fn extract(
```
gilcu3: this comments seems too general to be here, specially when mentioning the FanOut. So I would remove the last sentence, and the reference to a test.

We should be able to achieve that invariant through tests/code

gilcu3 (dismissed):
Thank you!

netrome (dismissed):
Seconding Reynaldo's comment. It's not obvious how to enforce that invariant though.

crates/foreign-chain-inspector/src/lib.rs
haiyuechen-nearone: We do enforce that invariant in the tests for individual Inspectors. I think it is worth adding the comment here, since its a requirement for future implementations of Inspector::extract(). Since the Inspector implementation is supposed to handle the transformation between provider 404 to verdic/error, I don't think we can enforce an invariant that is "All Inspector::Extract should handle absence of transaction in a certain way", other than a comment.

gilcu3: I think `extract` should match each case to the matching `Verdict`, and that's a no-brainer. If that has security implications, then this comment, or something similar, might be suitable for the definition of each `Verdict::` branch. The implementor would read that and know what they need to do. 

Here the comment would say something like: it is important that the returned error types (of `extract`) coincides accurately with what's defined in `Verdict::`, specially the `TransactionNotFound` variant.

gilcu3 (dismissed):

gilcu3 (approved):

netrome (approved):

PR #4371, opened 2026-09-04
feat(contract): ProposalHash allows for user-defined serializers and hashers

### LINKED ISSUE #4369: Generic ProposalHash
### Background

In https://github.com/near/mpc/pull/2934 we introduced generic voting structs.
The idea is that we receive a proposal `p`, borsh-deserialize it and then hash the bytes to receive a `ProposalHash`.

In order to discourage the use of non-sdk hash methods, #2934 introduced a blanket conversion implementation:

https://github.com/near/mpc/blob/081ff8e68af90044e93f3250066237c80de4ea7d/crates/contract/src/primitives/votes.rs#L134-L145

Unfortunately, this has the following shortcomings:
- We need to implement `ProposalHashEncoding` for every type that we intend to vote for. This is a bit annoying and errors might occur.
- We have some structs we vote on that are already hashes (e.g. `NodeImageHash`). It odesn't make much sense to re-hash a hash. In fact, it might make it harder for operators to comprehend the output from the contracts view methods.

https://github.com/near/mpc/blob/081ff8e68af90044e93f3250066237c80de4ea7d/crates/primitives/src/hash.rs#L186

https://github.com/near/mpc/blob/081ff8e68af90044e93f3250066237c80de4ea7d/crates/primitives/src/hash.rs#L177-L183

-  In case we don't want to use borsh (we may chose json as that might make it easier for node operators to verify if a proposal hash corresponds to their vote)

We should allow a more generic `ProposalHash` type and one that makes it easier to understand what encoding is used.

### User Story

As a developer, I like when code is easy to read and write.

### Acceptance Criteria

Introduce an API to define a method transforming structs we vote on into `ProposalHash`. In particular, we should be able to:
- use the identify function instead of a hasher when appropriate;
- use different serializers than borsh. In particular, we may chose the identify function or a JSON serializer in case it may improve UX.

_No response_

kevindeforth:
resolves https://github.com/near/mpc/issues/4369


netrome (approved):
Beautiful stuff!

gilcu3 (approved):
Thank you!

PR #4384, opened 2026-09-09
feat: LLM inference request domain and contract entry points

haiyuechen-nearone:
Part of the MPC verified LLM intents hackathon demo. Contract side only; node side follows.

Adds a `DomainPurpose::Llm` domain paired with CaitSith, plus a yield/resume request flow cloned from the foreign tx path:

* `request_llm_inference { prompt, schema, model_id }`: payable, validates preconditions against an Llm domain, queues into `pending_llm_inference_requests`.
* `respond_llm_inference`: attested participant only, verifies the ECDSA signature over `SHA256(borsh(LlmSignPayload))` against the domain root key, drains the fan out queue.
* `return_llm_inference_and_clean_state_on_success`: private yield resume callback, pops the yield and fires `fail_on_timeout` on timeout.
* `get_pending_llm_inference_request`: presence view.
* `LlmSignPayload::compute_msg_hash` pins the signed payload as request plus model id plus output.

The schema is stored and hashed opaquely; the contract never parses it. Validation is node side, magnitudes are the consumer contract's concern.

### DIFF
--- a/crates/contract/src/api.rs
+++ b/crates/contract/src/api.rs
@@ -9,6 +9,7 @@ pub mod governance;
 pub mod key_events;
 pub mod keys;
 pub mod lifecycle;
+pub mod llm;
 pub mod node_migration;
 pub mod sign;
 pub mod tee_measurements;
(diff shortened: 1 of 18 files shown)


haiyuechen-nearone:
Oops this was supposed to go on my hackthon fork

PR #4385, opened 2026-09-11
Reject zero hashes before consuming presignatures

milaforge:
Reject zero message hashes for ECDSA before they enter the signing flow or consume a presignature.

<details>
<summary>Economic rationale</summary>

The contract requires a 1 yoctoNEAR deposit, while the repository estimates a signing request uses about 7 Tgas (with 10 Tgas attached). Before this change, a single invalid request could consume one presignature from every selected participant.

For reconstruction threshold t, robust ECDSA selects 2t - 1 participants. The caller therefore pays for one transaction while depleting one presignature at each selected node.

</details>

Tests:

- cargo test -p mpc-node follower_sign_computation__should_reject_zero_hash_without_consuming_presignature --no-fail-fast
- cargo test -p mpc-contract sign__should_reject_zero_hash_for_ecdsa --no-fail-fast

<details>
<summary>Rationale by change</summary>

- The contract rejects h = 0 for both ECDSA protocols, whose signing flows reject that input; non-ECDSA protocols retain their current behavior.
- The leader validates before take_owned, so an invalid request cannot remove an owned presignature.
- The follower validates before take_unowned, preserving the same invariant if a malformed or stale request reaches a node.
- The private validator centralizes canonical-scalar and non-zero checks for the robust provider.
- Regression tests cover rejection for both ECDSA protocols and continued availability of the follower stored presignature.

</details>


crates/contract/src/api/sign.rs
```diff
@@ -46,9 +46,15 @@ impl MpcContract {
         match domain_config.protocol {
             dtos::Protocol::CaitSith | dtos::Protocol::DamgardEtAl => {
                 let hash = *request.payload.as_ecdsa().expect("Payload is not Ecdsa");
-                k256::Scalar::from_repr(hash.into())
+                let scalar = k256::Scalar::from_repr(hash.into())
                     .into_option()
                     .expect("Ecdsa payload cannot be converted to Scalar");
+                if matches!(domain_config.protocol, dtos::Protocol::DamgardEtAl) {
+                    assert!(
+                        !bool::from(scalar.is_zero()),
```
metalurgical: Good find. Should apply universally, irrespective of protocol though (for ECDSA).

metalurgical (commented):
@gilcu3 Think this should be merged once conflicts and any other review comments are resolved.

PR #4386, opened 2026-09-12
feat: add presign+sign online phase in Cait-Sith

### LINKED ISSUE #4324: Cait-Sith alternative without presignatures
### Background

An alternative to the current Cait-Sith implementation is one where only 2 phases are needed instead of 3. Currently we have:
- triples precomputation
- presignature precomputation
- signature computation

We could test merging the last two, and check by using benchmarks if the performance difference is relevant. If it is not, we could use this as a simpler alternative as the current implementation.

### User Story

As a user, a simpler system is better for trust and maintainability.

### Acceptance Criteria

- Implement a variant of Cait-Sith where presginatures and signatures are computed during the online phase
- Benchmark this variant against the existing variant and figure if they are comparable or not

EDIT: The original issue was rephrased to drop the benchmarks on `localnet` mention, as that can be done as a follow-up. Running benchmarks within the `threshold-signatures` crate should be good enough for this issue

gilcu3:
Closes #4324


crates/threshold-signatures/benches/simulate_ecdsa.rs
```diff
@@ -300,6 +318,61 @@ fn ot_run_sign(
     metrics
 }
 
+#[allow(clippy::too_many_arguments)]
+fn ot_run_presign_and_sign(
```
SimonRastikian, line 322: Please call it run sign. This should have an independent naming of our historic decisions.

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/sign.rs
```diff
@@ -78,19 +78,37 @@ where
...
         public_key,
-        presignature,
+        presignature.into(),
         msg_hash,
     );
     Ok(make_protocol(ctx, fut))
 }
 
+/// The signing round, run on a caller-owned channel so that it can follow other rounds on
+/// the same channel.
```
SimonRastikian: The comment is IMO not useful

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/sign.rs
```diff
@@ -78,19 +78,37 @@ where
...
+pub(crate) async fn sign_round(
```
SimonRastikian, line 87: Why did you call it sign_round. What does the round refer to in this name?

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/sign.rs
```diff
@@ -169,26 +187,24 @@ fn compute_signature_share(
...
-        do_sign_participant(
-            chan,
-            &participants,
-            coordinator,
-            me,
-            &presignature,
-            msg_hash,
-        )
-    }
+    sign_round(
```
SimonRastikian, line 195: Why was this intermediate function introduced?

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/presign_and_sign.rs
```diff
@@ -0,0 +1,408 @@
...
+use crate::ecdsa::{Scalar, SignatureOption, Tweak};
+use crate::errors::{InitializationError, ProtocolError};
+use crate::participants::{Participant, ParticipantList};
+use crate::protocol::{
+    Protocol,
+    internal::{Comms, SharedChannel, make_protocol},
+};
+
+/// Maximum incoming buffer entries for the coordinator: the two presigning waitpoints plus
+/// the signature-share waitpoint.
```
SimonRastikian:
```suggestion
/// Maximum incoming buffer entries for the coordinator: the two triples plus
/// the signature-share.
```

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/presign.rs
```diff
@@ -63,6 +64,22 @@ impl PartialEq for PresignOutput {
 }
 impl Eq for PresignOutput {}
 
+impl PresignOutput {
+    /// The same presignature for the key derived with `tweak`: `sigma' = sigma + tweak * k`,
+    /// since `sigma` shares `k * x` and the derived key is `x + tweak`.
+    ///
+    /// Only sound when the presignature is consumed by the computation that produced it, as in
+    /// [`presign_and_sign`](super::presign_and_sign::presign_and_sign); a stored presignature
+    /// must go through [`RerandomizedPresignOutput::rerandomize_presign`] instead.
```
SimonRastikian: Too much blabla here. 

Also long(er) term, we need to get rid of this struct as there is nothing called presigning.

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/presign_and_sign.rs
SimonRastikian, line 1: I see several places where it is written presigning instead of precomputed triples (yes one can consider them as presignatures) but to prevent

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/presign_and_sign.rs
```diff
@@ -0,0 +1,408 @@
...
+pub(crate) const OT_ECDSA_PRESIGN_AND_SIGN_MAX_INCOMING_COORDINATOR_ENTRIES: usize = 3;
+/// Maximum incoming buffer entries for non-coordinator participants: the two presigning
+/// waitpoints.
+#[cfg(test)]
+pub(crate) const OT_ECDSA_PRESIGN_AND_SIGN_MAX_INCOMING_PARTICIPANT_ENTRIES: usize = 2;
```
SimonRastikian, line 19: Why don't you plug this directly into the test module?

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/presign_and_sign.rs
```diff
@@ -0,0 +1,408 @@
...
+                &self.triple0.0,
+                &self.triple1.0,
+                self.threshold.into(),
+            );
+            let tweaked = presignatures
+                .iter()
+                .map(|(p, presig)| {
+                    (
+                        *p,
+                        RerandomizedPresignOutput::new_without_rerandomization(
```
SimonRastikian, line 213: I am very skeptical about keeping this struct `RerandomizedPresignOutput` and whether things would be okay like that or not.

SimonRastikian (dismissed):
Maybe deserve to modify the technical Cait-Sith doc (README)  and merge the presigning section to the signing one.

On the short term, this kind of change is fine, however on the long term, I would like to see the real integration happening (i.e. deleting any function called presign, merge the functions sign and presign together on the math level, rework the tests, delete no more needed API, etc...)

crates/threshold-signatures/benches/simulate_ecdsa.rs
gilcu3: at the moment we cannot do that because we still need to keep the old functions. Once those are removed we could rename this one.

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/sign.rs
gilcu3: It refers to a single round of communication, as usual

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/presign.rs
gilcu3: Will reduce the comment. At the moment we cannot yet remove it though

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/presign_and_sign.rs
gilcu3: I think this comment was cut mid write :upside_down_face:

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/presign_and_sign.rs
gilcu3: This is to avoid code duplication now, keeping the structure. If we remove presigning for good, we would them refactor this so that `RerandomizedPresignOutput` disappears.

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/sign.rs
gilcu3: it is mainly to avoid code duplication and to help wiring the new protocol

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/presign_and_sign.rs
gilcu3: handled the comment, although triples is not relevant here

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/presign_and_sign.rs
gilcu3: this is just status quo to stay near the other related constant

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/presign_and_sign.rs
netrome: > I think this comment was cut mid write 🙃

Just feed it to an LLM to extrapolate the rest 😂 

Though I do think this is an important point to reflect on naming here.

crates/threshold-signatures/benches/simulate_ecdsa.rs
netrome: Hmm, this becomes an interesting naming consideration. Now this scheme is effectively named as "ot_presign_and_sign" which is confusing since by that convention this function should be called: "ot_presign_and_sign_run_sign". I'm not sure what an appropriate name is here but it's worth reflecting on.

netrome (commented):
Partial review, will continue after lunch. Seems like we should reflect a bit on the naming of this "new scheme". However, I thought we'd replace the old scheme completely and not keep both. Would love some clarity on that.

crates/threshold-signatures/benches/simulate_ecdsa.rs
netrome: Synced with @gilcu3 - we'll only keep this for the transition where the node needs to support both schemes. Then this will be the new "ot_based_ecdsa" - so let's not get hung up on naming in this PR.

crates/threshold-signatures/docs/ecdsa/ot_based_ecdsa/signing.md
```diff
@@ -104,5 +104,27 @@ The inputs to this phase are:
...
+presignature's $R$ to the message; here $R$ is instead consumed immediately by the run that
+produced it, so that binding is replaced by a caller obligation: a triple pair must never be fed to
+this protocol twice, aborted runs included. $R$ is a deterministic function of $\mathsf{triple}_0$
+and is known to every party after round 1, so two runs over the same pair with different $h$ reveal
+the private key.
+
+**Round 3:** as in [Signing](#signing) Round 1, with the coordinator verifying $(R, s)$ against the
+derived key $X + \epsilon \cdot G$.
+
+**Output:** the signature $(R, s)$, known only to the coordinator.
```
netrome, line 127: Makes sense. I assume we'll apply this to the original protocol and remove this section once we merge the schemes.

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/presign_and_sign.rs
```diff
@@ -0,0 +1,393 @@
...
+/// [`presign`](super::presign::presign) followed by [`sign`](super::sign::sign), as both start
+/// at the same waitpoint and their messages would be confused. Use this protocol instead.
+pub fn presign_and_sign(
+    participants: &[Participant],
+    coordinator: Participant,
+    me: Participant,
+    args: PresignArguments,
+    tweak: Tweak,
+    msg_hash: Scalar,
+) -> Result<impl Protocol<Output = SignatureOption> + use<>, InitializationError> {
```
netrome, line 49: Interesting, I'm not used to the "use<>" parameter here. But TIL this is intended to prevent lifetimes to be captured by the returned type iiuc https://doc.rust-lang.org/reference/types/impl-trait.html#precise-capturing

Nice! It seems like the code compiles without it, but this looks like a good restriction so the return type doesn't contain e.g. a reference to the participants in the input parameters.

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/presign_and_sign.rs
```diff
@@ -0,0 +1,393 @@
+use crate::ecdsa::ot_based_ecdsa::presign::{
+    PresignArguments, presign_rounds, validate_presign_arguments,
+};
+use crate::ecdsa::ot_based_ecdsa::sign::sign_round;
```
netrome, line 4: Nit: I'd prefer to keep at least one level of namespacing for imported functions. I find the code easier to navigate if it's easy to differentiate between imported and local functions.

netrome (approved):
Thank you! When we integrate this in the node, I would love it if we could allow configuring a probability between which of the two variants we roll, so we can gradually increase the probability of the new join protocol.

crates/threshold-signatures/docs/ecdsa/ot_based_ecdsa/signing.md
gilcu3: yes, once the old protocol is removed from the codebase we could remove the corresponding docs as well

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/presign_and_sign.rs
gilcu3: the `use` is actually needed here:

```
$ cargo check -p threshold-signatures --all-features --all-targets
...
  error: lifetime may not live long enough
     --> crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/presign_and_sign.rs:182:32
      |
  179 |             &self,
      |             - let's call the lifetime of this reference `'1`
  182 |             let mut protocols: GenProtocol<SignatureOption> =
      |                                ^^^ type annotation requires that `'1` must outlive `'static`
```

anodar (approved):

crates/threshold-signatures/src/ecdsa/ot_based_ecdsa/presign_and_sign.rs
gilcu3: although I agree, it is currently the only change I would need to push to the PR and the pattern would differ from most of the rest of the crate. For those reasons I am just keeping as is, will take it into account next time

PR #4392, opened 2026-09-14
docs: document how host unattended upgrades should be handled in TDX 

### LINKED ISSUE #3462: docs: set `KillMode=process` on the dstack-vmm systemd unit so a daemon restart doesn't kill all CVMs
### Background

The sample `dstack-vmm.service` unit in [`running-an-mpc-node-in-tdx-external-guide.md`](https://github.com/near/mpc/blob/main/docs/running-an-mpc-node-in-tdx-external-guide.md) (the "VMM service persistence" section) sets no `KillMode=`, so it defaults to `control-group`. The qemu CVMs run as child processes inside the unit's cgroup, so **any restart of `dstack-vmm`** — maintenance, an upgrade, or `needrestart` bouncing the service after a shared-library/`unattended-upgrades` update — sends SIGTERM to *every* CVM on the host, taking all co-located nodes down at once (`qemu-system-x86_64: terminating on signal 15 from pid 1 (systemd)`).

Setting `KillMode=process` would make systemd kill only the daemon's main process on stop, leaving the running CVMs alive across a daemon restart.

### Acceptance Criteria
- Verify on a test host (or localnet) that with `KillMode=process` set, restarting `dstack-vmm` leaves the qemu CVMs running **and** the restarted daemon cleanly re-adopts them (reconnects to the existing supervisor, no VM restart).
- If verified, add `KillMode=process` (with an explanatory comment) to the documented `dstack-vmm.service` unit in `running-an-mpc-node-in-tdx-external-guide.md`.
- If `KillMode=process` does **not** cleanly re-adopt running CVMs, document the per-CVM systemd-scope alternative instead.

gilcu3:
Closes https://github.com/near/mpc-private/issues/564

closes https://github.com/near/mpc/issues/3462 as well.


docs/guide/running-an-mpc-node-in-tdx-external-guide/running-an-mpc-node-in-tdx-external-guide.md
```diff
@@ -631,6 +656,136 @@ For more information, see [local-key-provider-from-phala](https://github.com/Dst
...
+#### 4. Host Package Upgrades
+
+Leave unattended security upgrades enabled: a patched host is part of what keeps
+the node trustworthy, and it is the standard for servers of this kind. What has
+to be prevented is an upgrade disturbing the CVMs underneath it, in one of three
+ways: rebooting the host, restarting `dstack-vmm`, or changing a package the
+sealing key and the attestation measurements derive from.
+
+The four settings below cover those three routes. Because dstack CVMs are
+near-independent of the host, they are enough on their own; there is no need to
```
barakeinav1, line 670: >near-independent 

PUN ? :-)

docs/guide/running-an-mpc-node-in-tdx-external-guide/running-an-mpc-node-in-tdx-external-guide.md
```diff
@@ -631,6 +656,136 @@ For more information, see [local-key-provider-from-phala](https://github.com/Dst
...
+stagger upgrade windows across operators.
+
+##### Keep automatic reboots off
+
+A reboot stops every CVM, and TCB SVNs latch at platform reset, so rebooting is
+always a planned operation, never a side effect of a package upgrade. Ubuntu
+defaults to no automatic reboot, but confirm it rather than assume it:
```
barakeinav1, line 677: >but confirm it rather than assume it:

@gilcu3 you have started to use LLMs as well :-)

barakeinav1 (dismissed):
LGTM, 
approved (pending testing this first)

docs/guide/running-an-mpc-node-in-tdx-external-guide/running-an-mpc-node-in-tdx-external-guide.md
gilcu3: oh yeah, although that phrase probably sneaked through, else I would have judt removed it 😅

docs/guide/running-an-mpc-node-in-tdx-external-guide/running-an-mpc-node-in-tdx-external-guide.md
```diff
@@ -631,6 +656,154 @@ For more information, see [local-key-provider-from-phala](https://github.com/Dst
...
+  <(awk '/^Package: /{print $2}' /var/lib/apt/lists/ppa.launchpadcontent.net_kobuk-team_*_Packages | sort -u) \
+  <(dpkg-query -W -f='${Package}\n' | sort -u))
+
+sudo apt-mark hold intel-microcode $PPA_PKGS
+apt-mark showhold
+```
+
+`intel-microcode` comes from the Ubuntu archive rather than a PPA, hence the
+separate mention. Holds are origin-agnostic, so they stop both the PPA and the
+archive from moving these packages.
```
barakeinav1, line 739: nit: maybe we should add a line that this has a potential security cost, that packages may not be updated — e.g. on Alice the 9 libvirt packages sit at 8.3+tdx1.2 while security has 8.16.

docs/guide/running-an-mpc-node-in-tdx-external-guide/running-an-mpc-node-in-tdx-external-guide.md
```diff
@@ -631,6 +656,154 @@ For more information, see [local-key-provider-from-phala](https://github.com/Dst
...
+
+Nothing else needs freezing for the measurements: the guest firmware, kernel and
+rootfs come from the dstack image directory, not from packages. Check with
+`pgrep -af qemu-system-x86_64` that the running CVM is started with
+`-bios .../images/dstack-0.5.8/ovmf.fd`, `-kernel .../bzImage` and
+`-drive file=.../rootfs.img.verity`. The host's `ovmf` and `qemu-efi-*` packages
+are not in that path.
```
barakeinav1: nit: `$PPA_PKGS` holds `ovmf` and `qemu-efi-*` anyway, so this reads a bit confusing.

barakeinav1 (dismissed):
approved, few nits

docs/guide/running-an-mpc-node-in-tdx-external-guide/running-an-mpc-node-in-tdx-external-guide.md
gilcu3: I would rather not just to avoid bloating the doc, which is already very long. If intel believed the security issue was relevant, they should update their own PPA

docs/guide/running-an-mpc-node-in-tdx-external-guide/running-an-mpc-node-in-tdx-external-guide.md
gilcu3: hmm will check, that might be left-over

gilcu3: addressed in [48f9eb6](https://github.com/near/mpc/pull/4392/commits/48f9eb6f5cf76587ce5391580fcbe98e54ed3940)

barakeinav1 (approved):
approved, few nits

docs/guide/running-an-mpc-node-in-tdx-external-guide/running-an-mpc-node-in-tdx-external-guide.md
```diff
@@ -631,6 +656,154 @@ For more information, see [local-key-provider-from-phala](https://github.com/Dst
...
+```bash
+apt-config dump | grep -i 'Unattended-Upgrade::Automatic-Reboot'
+```
+
+No output means unset, which is off. Any line that does print must read
+`Unattended-Upgrade::Automatic-Reboot "false";`. Query the effective value like
+this rather than grepping the files: the stock `50unattended-upgrades` ships
+these options `//`-commented, so a grep shows inactive lines that read like live
+config. A pending reboot then only shows up as `/var/run/reboot-required`, which
+you act on during a maintenance window.
```
netrome, line 688: Hmm this is a bit wordy. I guess unset == bad? What should an operator do if this is unset?

netrome (approved):
The doc is a bit long and wordy. I don't have concrete edit suggestions yet but it could be worth looking into if we could make it leaner and clearer. That's not a blocker from my side though.

docs/guide/running-an-mpc-node-in-tdx-external-guide/running-an-mpc-node-in-tdx-external-guide.md
gilcu3: Yeah we could reduce it. It is long because a common pitfall is to use grep on the files, which would be inconclusive given the complexity of this. We probably should simplify the whole doc as you mentioned in slack, but that would require a much more fine grained analysis to see what should stay and what should go.

> What should an operator do if this is unset?

in this case we recommend setting the var so that it is explicit, although unset is also good, unless ubuntu changes the default, which might happen...

PR #4401, opened 2026-09-14
feat(config-tester): drop golden values in favor of network fingerprint probe in ForeignTX config tester

### LINKED ISSUE #3969: config tester: reuse the node's identity probe and drop golden transactions
### Background

Phase 2 of #3764. Phase 1 leaves two probing routes in the crate: the node's identity probe and the CLI's pinned golden transactions. Retire the golden one. The CLI runs the node's probe, and expected values come from the config file under test, so the CLI checks exactly what that node will check at startup.

### User Story

As a node operator I want the config tester to validate my foreign chain config without curated reference data, and to agree with what my node reports at startup, because it runs the same probe.

### Acceptance Criteria

- For a given config, the tester reports the same per provider verdict the node reports at startup.
- No hand picked reference transactions.
- Output stays one row per chain and provider, with pass or fail and a reason.
- Docs tell operators the expected identity value for each chain and network.

haiyuechen-nearone:
Closes #3969

The config tester now runs `probe_all_providers`, the same probe the node runs after startup and then hourly. The golden route (`check_all_providers`), its reference transactions, and the `--network` flag are deleted.

Notes for reviewers:

- This PR drops the SVM golden values introduced recently in #4138, and the SVM genesis hashes now live in `foreign-chain-inspector/src/svm.rs`.


netrome (approved):
Thank you!

anodar (approved):
Thank you! 🙏


As discussed offline, let's merge this after we update configs and coordinate network_fingerprint update with node operators so that CLI won't fail on otherwise valid config.

PR #4403, opened 2026-09-14
refactor(contract): use generic votes for mpc node docker hash votes

### LINKED ISSUE #2955: contract: store TEE votes by `AccountId`
### Background

We currently store votes for changing the `TeeState` (such node or launcher image hashes) by `AuthenticatedParticipantId`.

This implicitly assumes that a `ParticipantId` is never re-used, an invariant the contract currently guarantees.
However, there are two downsides to this:
1. The invariant may be accidentally broken in the future: If one would think of `ParticipantId` not as a global constant, but as a constant with respect to the current epoch, then mis-attribution of votes would be possible.
2. It is not very ergonomic. `AccountId`'s are easier to read and parse as a human, whereas for `ParticipantId`s, one needs to look-up the participant id for the given account id first.

### User Story

As a node operator, I want vote attribution to rely on explicit and stable identity semantics, so that future refactors do not break an implicit invariant and incorrectly associate me with a former operator’s votes.

Furthermore, as a node operator, I prefer reading `AccountId`s over `ParticipantId`s.

### Acceptance Criteria

Store TEE votes by `AccountId`, not by `ParticipantId` (can use `Authenticated*` types if applicable, but they are probably not necessary anymore).

- [ ] for docker node hashes: https://github.com/near/mpc/pull/4403 
- [ ] for launcher node hashes
- [ ] for other vote structs

_No response_

### LINKED ISSUE #4402: replace code-hash votes with generic votes
### Background

Part of #2765

We currently have a lot of custom logic for storing votes:

https://github.com/near/mpc/blob/561456f73eba820bff4c5ffd763168ee53c5113b/crates/contract/src/tee/proposal.rs#L24-L73

We should replace this logic with the generic votes struct in the contract to reduce maintenance surface.

### User Story

As a dev, I like re-using common components such that reading and writing code is easier.

### Acceptance Criteria

Replace CodeHashesVotes with the generic votes struct.

_No response_

kevindeforth:
resolves #4402 and part of #2955.

I am usually not a fan of resolving multiple issues in the same PR, but in this case, it's much easier to resolve #2955 in here than to merge this first and then resolve separately.


crates/contract/src/api/tee_measurements.rs
```diff
@@ -614,35 +621,40 @@ mod tests {
...
                 .expect("vote should succeed");
 
-            let votes = &contract.code_hash_votes().proposal_by_account;
+            // Then
+            let votes = contract.code_hash_votes();
             if i < (threshold - 1) as usize {
-                assert_eq!(votes.len(), i + 1);
-                assert!(votes.values().all(|v| *v == code_hash));
+                assert_eq!(votes.len(), 1);
+                assert_eq!(votes[&code_hash], expected_voters);
```
kevindeforth, line 657: Note: we are now comparing the entries, not only the number thereof.

crates/contract/src/tee/tee_state.rs
```diff
@@ -321,12 +329,12 @@ impl TeeState {
         }
     }
 
-    pub fn vote(
+    pub fn vote_mpc_node_image_hash(
```
kevindeforth: I took the liberty to rename this method, to make it clearer _what_ we are voting on.

crates/contract/src/tee/tee_state.rs
```diff
@@ -1633,49 +1643,56 @@ mod tests {
...
-    /// 2. Resharing removes P1 and P2. New set: {P3, P4, P5}.
-    /// 3. clean_non_participant_votes removes stale votes.
-    /// 4. P3 votes for the same hash — only 1 vote, not 3.
+    /// 1. P0, P1 and P2 vote for a node image hash before resharing.
+    /// 2. Resharing removes P0 and P1. New set: {P2, P3, P4}.
+    /// 3. clean_non_participant_votes keeps only P2's vote — 1 vote toward quorum, not 3.
     #[test]
-    fn test_clean_non_participant_votes_removes_stale_votes() {
-        // Build 5 participants
+    fn clean_non_participant_votes__should_keep_only_votes_of_current_participants() {
```
kevindeforth, line 1654: broken window fix: naming the test according to our `engineering-standards.md` file and following the `Given / When / Then` structure.

Additionally, we are now comparing actual vs expected votes instead of only the number of collected votes.

crates/contract/src/v3_15_1_state.rs
```diff
@@ -22,14 +24,58 @@ use crate::{
...
+    launcher_votes: LauncherHashVotes,
+    stored_attestations: IterableMap<dtos::Ed25519PublicKey, NodeAttestation>,
+    allowed_measurements: AllowedMeasurements,
+    measurement_votes: MeasurementVotes,
+}
+
+impl From<OldTeeState> for TeeState {
+    fn from(old: OldTeeState) -> Self {
+        // Pending code-hash votes are dropped. We don't need to optimize for this and it's not
+        // worth the review & testing effort.
```
kevindeforth, line 60: note: Claude has the annoying tendency to ask for a unit test checking that we _drop_ the votes. Such a unit test would be pretty useless and does not provide strong guarantees or our production contract.

crates/contract/src/api/tee_measurements.rs
```diff
@@ -24,8 +24,12 @@ impl MpcContract {
...
         let threshold_parameters = self.protocol_state.threshold_parameters_or_panic();
 
-        let participant = AuthenticatedParticipantId::new(threshold_parameters.participants())?;
-        let votes = self.tee_state.vote(code_hash, &participant);
+        let voter = AuthenticatedAccountId::new(threshold_parameters.participants())?;
+        let votes = self
+            .tee_state
+            .vote_mpc_node_image_hash(code_hash, &voter)
+            .count_participants(threshold_parameters.participants());
+        log!("total votes for proposal: {}", votes);
```
barakeinav1, line 32: what is our standard /guidelines about logs in the contract? when do we use them and when not?
does this have any impact on the blockchain size?

crates/contract/src/primitives/proposal_hash.rs
```diff
@@ -39,6 +45,24 @@ impl HashProposal for Sha256 {
...
 
+pub(crate) struct Identity;
+
+impl<T: AsRef<[u8; PROPOSAL_HASH_BYTES]>> SerializeProposal<T> for Identity {
+    fn serialize(value: &T) -> Vec<u8> {
+        value.as_ref().to_vec()
+    }
+}
+
+impl HashProposal for Identity {
```
barakeinav1: We can make this more robust and fail at compile time instead of runtime if we use the following pattern, WDYT?

```rust
trait SerializeProposal<T> { type Output; fn serialize(value: &T) -> Self::Output; }
trait HashProposal<B> { fn hash(bytes: B) -> ProposalHash; }

trait ToProposalHash: Sized {
    type Serializer: SerializeProposal<Self>;
    type Hasher: HashProposal<<Self::Serializer as SerializeProposal<Self>>::Output>;
}
```

Today `Identity::hash` only works if the Serializer is `Identity` too, and nothing enforces it — a future change like this compiles fine and only blows up at runtime, on the first vote:

```rust
impl ToProposalHash for SomeProposal {
    type Serializer = Borsh;    // arbitrary-length bytes
    type Hasher = Identity;     // expects exactly 32
}
```

With the pattern above that's a compile error, and `Identity::hash` becomes just `ProposalHash::new(bytes)` with no panic at all.

crates/contract/src/api/tee_measurements.rs
```diff
@@ -203,7 +207,7 @@ impl MpcContract {
         (&self.tee_state.launcher_votes).into_dto_type()
     }
 
-    /// Returns the current code hash votes, showing each participant's vote.
+    /// Returns pending code hash votes.
     pub fn code_hash_votes(&self) -> dtos::CodeHashesVotes {
```
barakeinav1: We're already breaking this view's shape, so callers have to update anyway — is it worth renaming it to match `NodeImageHash` while we're at it? Otherwise "code hash" is locked in until the next breaking change. Same for `vote_code_hash`. WDYT?

crates/contract/src/tee/tee_state.rs
```diff
@@ -460,7 +468,8 @@ impl TeeState {
     /// concludes. Attestation cleanup is handled separately by
     /// [`TeeState::clean_invalid_attestations`].
     pub fn clean_non_participant_votes(&mut self, participants: &Participants) {
-        self.votes = self.votes.get_remaining_votes(participants);
+        self.votes
+            .retain_votes(|voter: &AuthenticatedAccountId| participants.is_participant(voter));
```
barakeinav1, line 472: `launcher_votes` and `measurement_votes` just below are still the old style keyed by `ParticipantId` — do we have follow-up issues for those, or just the checkboxes in #2955? Also the PR body says `resolves #2955`, which will auto-close it even though two of its three boxes are still open.

barakeinav1:
Should this commit be marked as breaking? `code_hash_votes()` changes shape for external callers, and git-cliff only tags it in the changelog with a `!` or a `BREAKING CHANGE:` footer.

crates/contract/src/tee/tee_state.rs
```diff
@@ -1633,49 +1643,56 @@ mod tests {
...
+            tee_state.votes.all(),
+            BTreeMap::from([(
+                node_image_hash.to_proposal_hash(),
+                BTreeSet::from([voters[2].clone()]),
+            )])
+        );
+        let vote_count = tee_state
+            .vote_mpc_node_image_hash(node_image_hash, &voters[2])
+            .count_participants(&new_participants);
+        assert_eq!(vote_count, 1, "only P2's vote should count toward quorum");
```
barakeinav1: Claude flagged this, the re-vote at the end is an action inside `Then`, and it can't fail — the assert above already pins the set to exactly `{P2}`, so re-voting the same hash as the same voter is a no-op. It was doing real work in the old test where P2 hadn't voted yet. Drop it?

crates/primitives/src/hash.rs
```diff
@@ -239,10 +239,12 @@ define_hash!(
     48
 );
 
+pub const PROPOSAL_HASH_BYTES: usize = 32;
+
 define_hash!(
     /// SHA-256 digest identifying a governance proposal by its encoded content.
```
barakeinav1, line 245: This doc is now inaccurate — with `Identity` a `ProposalHash` can be the proposal's raw bytes rather than a SHA-256 digest.

crates/contract/src/api/tee_measurements.rs
```diff
@@ -203,7 +207,7 @@ impl MpcContract {
         (&self.tee_state.launcher_votes).into_dto_type()
     }
 
-    /// Returns the current code hash votes, showing each participant's vote.
+    /// Returns pending code hash votes.
```
barakeinav1: nit: the tally now filters to current participants but this view doesn't, so right after a resharing it can list voters that don't count. Corner case, but the guide says it shows how many more votes are needed — should we document it?

crates/contract/src/v3_15_1_state.rs
barakeinav1: We have an existing test for this — `propose_upgrade_from_production_to_current_binary` in `tests/sandbox/upgrade_to_current_contract.rs`. Updating it to also leave a code-hash vote pending before the upgrade would be helpful: `TeeState.votes` is the only layout change here, but it's only ever migrated as an empty map today, which decodes fine even under a wrong shadow type. Same reasoning as the launcher-hash comment at line 188 — `vote_for_hash` already exists, so ~3 lines.

netrome:
> Should this commit be marked as breaking? code_hash_votes() changes shape for external callers, and git-cliff only tags it in the changelog with a ! or a BREAKING CHANGE: footer.

Technically yes. However, we haven't been super strict about that for governance actions, since in practice we just send the changed commands to operators. I'm personally okay with both.

crates/contract/src/api/tee_measurements.rs
netrome: I agree, and would suggest also renaming the "vote_code_hash" methods to "vote_mpc_node_manifest_digest" to be more explicit.

crates/contract/src/primitives/proposal_hash.rs
netrome: Nice, I like this idea! Effectively, today there's no guarantee the output of the serializer is compatible with the input of the hasher. This would enforce it.

crates/contract/src/tee/tee_state.rs
netrome: Nice!

crates/primitives/src/hash.rs
kevindeforth: That depends on the definition of "encoding". Generally, this just means to take data and map it to some other data. The identify function is a perfectly fine mapping in that process.

crates/contract/src/tee/tee_state.rs
netrome: While at it, I'd suggest calling it "manifest_digest" and not "image_hash" to make it more accurate
```suggestion
    pub fn vote_mpc_node_manifest_digest(
```

crates/contract/src/tee/tee_state.rs
kevindeforth: Not following the question.

> do we have follow-up issues for those, or just the checkboxes in https://github.com/near/mpc/issues/2955?

Seems like the issue already tracks it?

> Also the PR body says resolves #2955, which will auto-close it even though two of its three boxes are still open.

The PR body was edited a while ago and  #2955 is not in the list of issues that are getting closed by this PR (c.f. right-hand side).

crates/contract/src/api/tee_measurements.rs
kevindeforth: good idea!

crates/contract/src/tee/tee_state.rs
```diff
@@ -321,12 +329,12 @@ impl TeeState {
...
+    pub fn vote_mpc_node_image_hash(
         &mut self,
         code_hash: NodeImageHash,
-        participant: &AuthenticatedParticipantId,
-    ) -> u64 {
-        self.votes.vote(code_hash, participant)
+        voter: &AuthenticatedAccountId,
+    ) -> &VoterSet<AuthenticatedAccountId> {
+        self.votes.vote(voter.clone(), code_hash.to_proposal_hash())
     }
```
netrome: Nit: Taking a reference only to clone it is redundant. We should take ownership of the value instead, allowing the caller to decide if they want to move or clone.

```suggestion
        voter: AuthenticatedAccountId,
    ) -> &VoterSet<AuthenticatedAccountId> {
        self.votes.vote(voter, code_hash.to_proposal_hash())
    }
```

crates/contract/src/api/tee_measurements.rs
kevindeforth: > what is our standard /guidelines about logs in the contract?

I am not aware that we have any.

>  when do we use them and when not?

AFAIK, it's good practice to have them at the beginning of a call. Some consumers need it. Bu there, it's pretty much optional, since those are just node-operator endpoints.

> does this have any impact on the blockchain size?

Do you mean contract size? If so yes it's getting compiled, but I don't expect our logs to be the bottleneck in this.

crates/contract/src/dto_mapping.rs
```diff
@@ -909,15 +907,19 @@ impl IntoInterfaceType<dtos::LauncherHashVotes> for &LauncherHashVotes {
     }
 }
 
-impl IntoInterfaceType<dtos::CodeHashesVotes> for &CodeHashesVotes {
+impl IntoInterfaceType<dtos::CodeHashesVotes> for &Votes<AuthenticatedAccountId> {
```
barakeinav1, line 910: Suggested by Claude: the impl is keyed on the voter type, but the `ProposalHash -> NodeImageHash` conversion is only valid because of the *proposal* type's `Identity`. Once #2825 lands, governance votes are `Votes<AuthenticatedAccountId>` too. Would a `CodeHashVotes(Votes<...>)` newtype be cleaner, like `TeeVerifierVotes` does?

barakeinav1 (dismissed):
thanks, looks good. approved with some comments and questions.

netrome (dismissed):
Looks good to me. Some minor nits and suggestions. I especially like Barak's suggestion for the ToProposalHash trait, would love to hear your thoughts on it.

crates/contract/src/v3_15_1_state.rs
kevindeforth: >  Updating it to also leave a code-hash vote pending before the upgrade would be helpful:

Why? That's what I don't understand. 

It's clear that we won't migrate existing votes, because there is no logic for that. There's also no need to protect against any regression here.

What we _would_ be interested in, is seeing whether the map under the new storage key is actually empty. But that test doesn't give us any useful information if run in a sandboxed environment. We would have to take the contract from chain and check it there.

But that is not something we ever did and it's a huge effort, just to ensure that we never ever used the same storage key on-chain beforehand.

crates/contract/src/api/tee_measurements.rs
kevindeforth: > and would suggest also renaming the "vote_code_hash" methods to "vote_mpc_node_manifest_digest" to be more explicit.

Also agree, but that's a follow-up to me. We are not breaking that method in this PR and it will have a blast radius. I will stack a branch on top.

crates/contract/src/primitives/proposal_hash.rs
kevindeforth: changed in: https://github.com/near/mpc/pull/4403/commits/da691c72a61e29a6b85389960e3bbefc4b2162f5

crates/contract/src/tee/tee_state.rs
kevindeforth: fixed in https://github.com/near/mpc/pull/4403/commits/2d4ea8b311593bb7089b2266b182b7dda9f46ca5

crates/contract/src/tee/tee_state.rs
kevindeforth: changed in https://github.com/near/mpc/pull/4403/commits/59cea8e164b4922a6b1c5cb7f062e093ee343800

crates/contract/src/tee/tee_state.rs
kevindeforth: reemoved in https://github.com/near/mpc/pull/4403/commits/d86aad0143bcf2f7349b1b8d68213455530b24c7

crates/contract/src/api/tee_measurements.rs
kevindeforth: documented in https://github.com/near/mpc/pull/4403/commits/d7f5ed914db16b90277dc5ae7780d5f4032e8f8b

crates/contract/src/api/tee_measurements.rs
kevindeforth: done in https://github.com/near/mpc/pull/4447

crates/contract/src/api/tee_measurements.rs
barakeinav1: >Do you mean contract size? I

no, I meant are the logs recorded on the chain

crates/contract/src/tee/tee_state.rs
barakeinav1: ok I see we have "part of https://github.com/near/mpc/issues/2955."

crates/contract/src/api/tee_measurements.rs
netrome: > no, I meant are the logs recorded on the chain

Yes, afaik they are. I'd be happy to skip logs as we're not using them (yagni). For user facing requests I'd be cautious though since we cannot know if anyone relies on them, but I would be surprised if anyone did.

crates/primitives/src/hash.rs
barakeinav1: Fair point on "encoding" — I meant the "SHA-256 digest" part at the start, which isn't accurate anymore when the hasher is `Identity`.

crates/contract/src/v3_15_1_state.rs
barakeinav1: Sorry, I wasn't clear — it's not about the votes themselves, it's about whether `OldCodeHashesVotes` decodes bytes written by the real deployed binary. An empty map is 4 zero bytes and decodes fine even if the shadow type is wrong, so today the test can't tell a correct shadow from a broken one. The shadow is declared in terms of `AuthenticatedParticipantId` and `NodeImageHash`, which can change under it without anything failing. That's the same reason the launcher-hash vote was added at line 188. Not blocking though, your call.

barakeinav1 (approved):
thanks for addressing the comments, re-approved. left two follow-ups, both non-blocking.

crates/contract/src/primitives/proposal_hash.rs
```diff
@@ -22,53 +22,55 @@ use borsh::BorshSerialize;
...
 
 impl<T: AsRef<[u8; PROPOSAL_HASH_BYTES]>> SerializeProposal<T> for Identity {
-    fn serialize(value: &T) -> Vec<u8> {
-        value.as_ref().to_vec()
+    type Output = [u8; PROPOSAL_HASH_BYTES];
+
+    fn serialize(value: &T) -> [u8; PROPOSAL_HASH_BYTES] {
+        *value.as_ref()
     }
 }
```
netrome, line 60: Hmm would be nice if we could passthrough here without the reference+dereferencing - but it's not obvious if this is trivial or not so perhaps let's not dive into that rabbit hole.

netrome (approved):
Thanks for the updates!

PR #4404, opened 2026-09-14
chore: Drop legacy foreign chain API

### LINKED ISSUE #3627: Revisit foreign chain config registration
### Background

We don't want to publicly expose which node usees which RPC provider. That's why we got rid of foreign chain tx config in `/debug/node_config` endpoint.

This is still publicly accessible on chain when node registers its foreign chain config.

### User Story

_No response_

### Acceptance Criteria

Information about which RPCs are used by each node is not publicly available.

_No response_

### LINKED ISSUE #3630: Drop deprecated foregin chain methods in contract once nodes are migrated to the new API

anodar:
Closes https://github.com/near/mpc/issues/3630

Also closes https://github.com/near/mpc/issues/3627, new registration API only sends chain, not providers/rpc urls.


anodar:
> docs/design/calculating-supported-foreign-chains.md:148 — "Completed. Node and contract both switch to the new API since version 3.15.0" reads as if the legacy API is already gone in 3.15.0.

No it only states that node and contract both use new API (for calculating available chains in a robust manner)

> Rollout ordering — once this contract is deployed, nodes still on 3.15.0 will have their dual-written register_foreign_chain_config call revert with MethodNotFound. 

We generally upgrade nodes first, then contract, but even if the order was other way around, burnt gas on deprecated registration call is not a concern.

crates/contract/src/v3_15_0_state.rs
```diff
@@ -51,14 +53,14 @@ impl From<MpcContract> for crate::MpcContract {
         if !matches!(old.protocol_state, ProtocolContractState::Running(_)) {
             env::panic_str("Contract must be in running state when migrating.");
         }
+        old.node_foreign_chain_support.clear_storage();
```
gilcu3, line 56: could you ensure that we have a sandbox test verifying that this does not fail for any reason? Gas and storage limits... I know this struct is not too big, but just in case

crates/e2e-tests/tests/common.rs
```diff
@@ -26,7 +26,6 @@ pub const CKD_VERIFICATION_PORT_SEED: u16 = 9;
 pub const LOST_ASSETS_PORT_SEED: u16 = 10;
 pub const CKD_PV_VERIFICATION_PORT_SEED: u16 = 11;
 pub const CLEANUP_LAGGING_NODE_PORT_SEED: u16 = 12;
-pub const FOREIGN_CHAIN_POLICY_PORT_SEED: u16 = 13;
```
gilcu3, line 29: instead of just deleting, could you mark it with a comment or something so that the number is reused later?

crates/node-config/src/foreign_chains/auth.rs
```diff
@@ -27,19 +25,6 @@ pub enum AuthConfig {
...
-    ///
-    /// For [`Path`](AuthConfig::Path) auth, removes the placeholder string from the URL (returning an owned [`String`]).
-    /// For all other auth kinds, returns the original URL by reference (zero-cost borrow).
-    pub(crate) fn strip_placeholder<'a>(&self, rpc_url: &'a str) -> Cow<'a, str> {
-        match self {
-            AuthConfig::Path { placeholder, .. } => Cow::Owned(rpc_url.replace(placeholder, "")),
-            _ => Cow::Borrowed(rpc_url),
-        }
-    }
-}
```
gilcu3, line 41: I am a bit surprised to see this one go. How come it is not used in the new API as well?

crates/contract/tests/sandbox/upgrade_to_current_contract.rs
```diff
@@ -504,84 +507,107 @@ async fn init_running_rejects_external_callers_pre_initialization() {
...
+        .prefix(&legacy_prefix)
+        .await?;
+    assert!(!legacy_entries.is_empty());
+
+    // When
     let contract = upgrade_to_new(contract).await?;
     migrate_and_assert_contract_code(&contract)
         .await
-        .expect("❌ migration() failed");
+        .expect("migration() failed");
```
gilcu3: thank you for this one. Adding inline emoji in code was never fashionable

gilcu3 (dismissed):
Thank you!

crates/node-config/src/foreign_chains/auth.rs
anodar: This was only used in the old registration API where we send links of RPC providers. We only send list of supported chains now in registration now so this is no longer needed and that's why this PR also closes https://github.com/near/mpc/issues/3627.

crates/contract/src/v3_15_0_state.rs
anodar: `upgrade__should_drop_legacy_support_and_preserve_foreign_chains_state` already covers that path:, 
but I extended list of chains in that test from bitcoin to all chains to check gas is enough.

crates/e2e-tests/tests/common.rs
anodar: Re-enumerated them instead, it's only used for tests so changing them doesn't break anything.

gilcu3 (approved):

haiyuechen-nearone (approved):
Nice 🎊

PR #4407, opened 2026-09-14
refactor: typed sandbox views

### LINKED ISSUE #4405: use MpcContractHandle for view TEE view-calls in sandbox
### Background

We want to use the mpc contract handle for view calls in the sandbox.
To break this down into smaller scoped PRs, we will first port TEE-related view calls before porting other calls.
Part of https://github.com/near/mpc/issues/3693

### User Story

As a dev, I like when view calls look the same independent of the back-end.

### Acceptance Criteria

The following methods are view-able through the MpcContractHandle:

- `ALLOWED_DOCKER_IMAGE_HASHES`
-  `ALLOWED_LAUNCHER_IMAGE_HASHES`
- `ALLOWED_LAUNCHER_COMPOSE_HASHES`
- `ALLOWED_OS_MEASUREMENTS`
- `CODE_HASH_VOTES`
- `LAUNCHER_HASH_VOTES`
-  `OS_MEASUREMENT_VOTES`

https://github.com/near/mpc/issues/3693

kevindeforth:
resolves #4405


kevindeforth:
@claude review

crates/contract/tests/sandbox/utils/views.rs
```diff
@@ -0,0 +1,67 @@
...
+            SandboxViewer(JsonRpcClient::connect(self.rpc_addr())),
+            contract_id.clone(),
+        )
+    }
+}
+
+/// Queries the sandbox RPC directly: near-workspaces' `view` drops the block height of the
+/// response, which [`ObservedState`] needs.
+#[derive(Clone)]
+pub struct SandboxViewer(pub JsonRpcClient);
```
netrome: Quick question, why does the inner client need to be public?

netrome (dismissed):
Thanks, looks good. I'm not sure why the inner JsonRpcClient needs to be pub though. Would suggest removing the "pub" if possible but I'm sure I'm just missing something. Not important though.

crates/contract/tests/sandbox/utils/views.rs
kevindeforth: yeah, that was an oversight. Thanks for spotting it!

crates/contract/tests/sandbox/tee.rs
```diff
@@ -47,24 +49,26 @@ fn mock_expiring_at(expiry_timestamp_seconds: u64) -> MockAttestation {
...
         .build()
         .await;
+    let mpc = worker.view_mpc(contract.id());
     let threshold = assert_running_return_threshold(&contract).await;
 
     let allowed_mpc_image_digest = image_digest();
 
     // Initially, there should be no allowed hashes
-    assert_allowed_docker_image_hashes(&contract, &[]).await;
+    assert_allowed_docker_image_hashes(&mpc, &[]).await;
```
gilcu3: nit: `mpc` sounds a bit off. Would it be better to use `mpc_contract` or `mpc_state` if you want to stay away from "contract"?

netrome (dismissed):

crates/near-mpc-contract-interface/src/client.rs
```diff
@@ -656,4 +700,43 @@ mod tests {
...
+
+        // When: every view, once, in declaration order; only the recorded
+        // calls matter, so the empty mock response is left undecoded
+        let _ = handle.allowed_docker_image_hashes().await;
+        let _ = handle.allowed_launcher_image_hashes().await;
+        let _ = handle.allowed_launcher_compose_hashes().await;
+        let _ = handle.allowed_os_measurements().await;
+        let _ = handle.code_hash_votes().await;
+        let _ = handle.launcher_hash_votes().await;
+        let _ = handle.os_measurement_votes().await;
```
gilcu3, line 704: so we don't care if the return type changes? That seems a bit surprising

crates/near-mpc-contract-interface/src/client.rs
```diff
@@ -656,4 +700,43 @@ mod tests {
             .join("\n\n");
         insta::assert_snapshot!(catalog);
     }
+
+    /// One catalog snapshot for the handle's views: every view is called once
+    /// and its wire format becomes a section of the snapshot. New view
+    /// methods add a call here.
+    #[tokio::test]
+    async fn mpc_contract_handle__should_match_the_view_catalog() {
+        // Given
```
gilcu3, line 689: for the other catalog test we have this line

> assert_eq!(calls.len(), 22);

we probably also need it for this one

gilcu3 (dismissed):
Thank you!

Left some comments, none blocking

crates/near-mpc-contract-interface/src/client.rs
kevindeforth: This test is to ensure our view args are correct and not accidentally changed, but it can't (and shouldn't) enforce any assumptions about the return values. For that, we have `abi__abi_has_not_changed.snap`.

gilcu3: right, I had forgotten this was a bit redundant wrt the abi test indeed

crates/near-mpc-contract-interface/src/client.rs
kevindeforth: That one is redundant / giving us a false sense of security. 
We check `insta::assert_snapshot!(catalog);`, so we know that if we add / remove a method to this test catalog, we will catch it. Hard-coding `22` won't make use less likely to forget adding a new method to the catalog. For that purpose, I left the LLM-code explanation on top of the test and I assume Claude will catch in case we miss adding a method.

gilcu3: Can we then remove the line in the other test? I thought you had added it on purpose, but it is quite annoying to see CI fail because of it :)

kevindeforth: fixed in https://github.com/near/mpc/pull/4407/commits/db2b5c4bdab775d3a065cc7b699db4693f141bb9

crates/contract/tests/sandbox/tee.rs
kevindeforth: https://github.com/near/mpc/pull/4407/commits/db2b5c4bdab775d3a065cc7b699db4693f141bb9

gilcu3 (approved):

netrome (approved):

PR #4411, opened 2026-09-14
test(contract): add poc checking the upgrade issue and possible solutions

gilcu3:
The test showcasing the fix is `vote_update__should_apply_at_the_mainnet_proposal_size_when_one_entry_is_probed`


gilcu3:
This was very useful and helped solve the problem, hopefully we don't need it anymore

PR #4415, opened 2026-09-14
test(contract): pin the pending-proposal receipt budget

barakeinav1:
`do_update` reads the winning proposal and evicts every loser in one receipt, so the whole proposals map is charged to a single receipt's 4 MB storage-proof budget. Past that, the threshold-reaching vote reverts and the upgrade can never be applied — no contract method removes a proposal. Mainnet `v1.signer` is in that state now with four identical 3.15.0 proposals.

Tests only, no behaviour change.

- Pins the boundary: three pending proposals still upgrade, four fail on the storage-proof limit. This is the test that should start failing if the contract or the proposal payload grows.
- Covers the paired-vote recovery for a contract already past the limit: an under-gassed vote pre-pays part of the proof into the chunk, and the vote behind it re-reads those blobs for free.
- Includes a negative control — the same vote sent alone still fails — so the recovery result is attributable to the sacrificial vote rather than to the submission path.

Deploys the archived mainnet binary rather than a fresh build, so the sizes under test match the ones that actually wedged mainnet.

### Scope of what has actually been verified

Everything here runs on a single-shard local sandbox, where both votes are guaranteed to land in the same chunk. Mainnet would not guarantee that, so **the paired-vote recovery should be read as "the mechanism is real", not "these gas values are the mainnet values"**. It has not been exercised on any real network.

A separate throwaway contract on testnet confirmed two narrower facts that fed into these tests — three proposals upgrade cleanly, and 3.15.0 migrates from 3.14.0 state — but it was never put into the wedged four-proposal state, so it says nothing about the recovery.

The boundary test is the part that stands on its own.


barakeinav1:
Superseded by #4411, which covers the same mechanism with a better probe — a gas-capped `proposed_updates` call rather than an under-gassed `vote_update`. It needs no participant account, has no vote side effects, and reads entries in a deterministic order. #4411 also signs both transactions under consecutive nonces of one key, which actually *guarantees* the probe executes first; the two-account approach here did not.

The recovery ran successfully on mainnet at block 215674217 — `v1.signer` is on 3.15.0, all four proposals cleared.

Leaving the branch `test/upgrade-proof-limit-recovery` in place, since two things here may still be worth picking up:

- `docs/design/upgrade-proposal-receipt-budget.md` — the background/sizing doc. #4411's script header references `incident-vote-update-storage-proof.md`, which does not exist on that branch, so that reference is currently dangling. This doc would need one edit before landing: it describes the recovery as untried on a real network, which is no longer true.
- The boundary test (three pending proposals apply, four fail), staged from the archived mainnet binary rather than synthetic payloads. Useful as a regression guard, since the ceiling of three drops as the contract binary grows.

PR #4417, opened 2026-09-14
feat: Allow participants to remove their update proposals

### LINKED ISSUE #4416: Add support for removing proposals
### Background

Currently, any participant can propose an update. However, it's not possible to remove a proposal. This can easily bloat the contract if someone submits an invalid/unintended proposal.

### User Story

As an operator I'd like to be able to remove my submitted proposals.

### Acceptance Criteria

The contract supports removing submitted proposals.

_No response_

netrome:
closes #4416


crates/contract/src/api/update.rs
```diff
@@ -114,6 +114,27 @@ impl MpcContract {
...
     }
 
+    /// Removes a proposed update, given the id returned by [`Self::propose_update`].
+    ///
+    /// The deposit attached at propose time is not refunded.
+    ///
+    /// Returns [`Error`] if no update with this id exists, and panics if the caller is not a
+    /// participant.
+    #[handle_result]
+    pub fn remove_update_proposal(&mut self, id: dtos::UpdateId) -> Result<(), Error> {
```
netrome, line 126: Note: We currently allow anyone to remove an update proposal. This effectively allows any account to block a vote if they want to. We should follow-up with a change to restrict this to only participants being able to remove their own votes, but I wanted to be more liberal here for now as this is a smaller and simpler change.

crates/near-mpc-contract-interface/src/client.rs
```diff
@@ -170,6 +170,20 @@ impl<C: CallContract> MpcContractHandle<C> {
...
 
+    pub async fn remove_update_proposal(
+        &self,
+        id: UpdateId,
+    ) -> Result<C::Output, MpcContractHandleError<C::Error>> {
+        let args = serde_json::to_vec(&RemoveUpdateProposalArgs::new(id))?;
+        self.call(FunctionCallArgs::no_deposit(
+            REMOVE_UPDATE_PROPOSAL,
+            args,
+            // Removing an entry reads back the proposed payload, which can be a full contract.
```
kevindeforth: Is this an outdated comment?

crates/contract/src/api/update.rs
kevindeforth: Neither option is very attractive. If we only allow the proposer to remove an entry, a malicious operator can still brick the contract.
I would suggest to replace the entire update flow with the generic votes struct and commit to the hash before submitting the code.
If we still want to merge this PR, I would suggest to make this function payable and require a deposit for removing a contract.

kevindeforth (commented):
I think a larger refactor is required. I will try to see how much effort that is.

crates/contract/src/api/update.rs
netrome: Generic votes struct is a good mid-term solution, but I want an immediate fix that we can roll out today with confidence. I'll add the deposit.

> A malicious operator can still brick the contract.

Yeah, none of the simple fixes gets around this limitation. However, in practice I'm not too worried about that for the immediate releases. It's a visible attack, and no operator have incentives to hurt the MPC network.

crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
```diff
@@ -1022,6 +1022,28 @@ expression: abi
           }
         }
       },
+      {
+        "name": "remove_update_proposal",
+        "doc": " Removes a proposed update, given the id returned by [`Self::propose_update`].\n\n The deposit attached at propose time is not refunded. Like [`Self::propose_update`], and\n unlike [`Self::remove_update_vote`], this is not restricted to the running state.\n\n TODO(#4419): restrict removal to the account that proposed the update.\n\n Returns [`Error`] if no update with this id exists, and panics if the caller is not a\n participant.",
```
gilcu3: nit: I would put the TODO in // and not /// so that it does not appear here

crates/near-mpc-contract-interface/src/client.rs
gilcu3: I think it is just trying to justify MAX_GAS here, in the sense that this call is expected to consume a sizable amount

netrome: No, but I see that it can be confusing since we don't return the payload. The point is that the remove method we call returns the whole contract which means the gas costs can scale with the size of the removed proposal.

Will rephrase this.

kevindeforth (dismissed):
Thanks. I agree on the priorities. Also, if we find a participant misbehaving, we can still vote them out (assuming resharing still works)

gilcu3 (dismissed):
This change seems good to me, thank you!

I agree long term it might not be needed, for example when we handle #1674, that has been long coming. But if we want something that can be rolled out immediately, then IMO this is it.

crates/contract/src/api/update.rs
netrome: Not following the reason for adding a deposit though. We already have the signer == predecessor check so there's no need for a deposit to prevent function call access keys.

crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
netrome: Ah yeah, agree. Thanks for spotting. Will fix.

netrome:
> Thanks. I agree on the priorities. Also, if we find a participant misbehaving, we can still vote them out (assuming resharing still works)

Exactly

crates/contract/tests/snapshots/abi__abi_has_not_changed.snap
netrome: Done in https://github.com/near/mpc/pull/4417/commits/f41846a95420e8218aaf30182097cdb673bad98f

crates/contract/src/api/update.rs
kevindeforth: deposit is for making the attack more expensive

crates/near-mpc-contract-interface/src/client.rs
netrome: Updated in https://github.com/near/mpc/pull/4417/commits/f41846a95420e8218aaf30182097cdb673bad98f

gilcu3 (approved):

kevindeforth (approved):

PR #4433, opened 2026-09-15
feat(node): wire protocol version selection

### LINKED ISSUE #4431: Support P2P protocol version selection at the protocol level
### Background

The P2P handshake already negotiates a `CommunicationProtocols` version, but the node throws it away once the connection is up. Nothing downstream can tell whether a given peer understands a newly added wire message, so every wire-breaking change is an all-or-nothing flag day.

#4399 needs exactly this: the leader must know which peers understand a new `EcdsaTaskId` variant before it sends one. Splitting the version plumbing out of #4399 lets it land on its own, with #4399 then reduced to the new cait-sith variant that builds on it.

There is a related hazard that exists today regardless of #4399. `MpcTaskId` and `EcdsaTaskId` are borsh enums with implicit discriminants, and an undecodable packet propagates out of the read loop (`crates/node/src/p2p.rs:757`), so sending a peer a task id it does not know tears down the whole connection rather than failing that one task.

### Acceptance Criteria

- The node can report the protocol version a peer advertised in the handshake of its current connection.
- `MpcTaskId`, `EcdsaTaskId` and the rest carry explicit borsh discriminants, pinned by a test, so variants can be appended without renumbering.

gilcu3:
Closes #4431 

In case it is needed, the PR on top of this #4436 should show how the versions will be used, but it is still WIP


crates/node/src/network/conn.rs
```diff
@@ -226,19 +232,46 @@ impl<I: Send + Sync + 'static, O: SenderConnectionId + Send + Sync + 'static>
...
+    /// The protocol version the peer advertised in the handshake of whichever connection is
+    /// currently up; `None` while it is not connected. Both directions carry the peer's own
+    /// version, so either answers.
+    pub fn peer_protocol_version(&self) -> Option<CommunicationProtocols> {
+        let outgoing = self
+            .outgoing_receiver
+            .borrow()
+            .connection
+            .upgrade()
+            .map(|conn| conn.peer_protocol_version());
```
gilcu3, line 249: we mostly care about the outgoing connection. For this reason we only use the incoming one as a fallback here. But the need to make this decision is odd, and it stems from us having two connections from what should probably be just one

crates/node/src/providers/ecdsa.rs
```diff
@@ -242,7 +246,7 @@ impl SignatureProvider for EcdsaSignatureProvider {
             },
 
             _ => anyhow::bail!(
-                "eddsa task handler: received unexpected task id: {:?}",
+                "ecdsa task handler: received unexpected task id: {:?}",
```
gilcu3, line 248: broken window fix

crates/node/src/network/conn.rs
```diff
@@ -226,19 +232,46 @@ impl<I: Send + Sync + 'static, O: SenderConnectionId + Send + Sync + 'static>
...
 
+impl<
+    I: AdvertiseProtocolVersion + Send + Sync + 'static,
+    O: AdvertiseProtocolVersion + Send + Sync + 'static,
+> NodeConnectivity<I, O>
+{
+    /// The protocol version the peer advertised in the handshake of whichever connection is
+    /// currently up; `None` while it is not connected. Both directions carry the peer's own
+    /// version, so either answers.
+    pub fn peer_protocol_version(&self) -> Option<CommunicationProtocols> {
```
netrome: I was confused by the `CommunicationProtocols` name here. While not introduced in this PR, we should probably rename that enum to `NetworkProtocolVersion` or something similar. wdyt?

crates/node/src/network/conn.rs
netrome: Hmm yeah this is interesting code smell. Could we have a connection with one protocol version in one direction and another in the other direction? In practice that should never happen, but still in some pieces of the code we now seem to be faced with this possibility, which is ugly.

crates/node/src/network/conn.rs
```diff
@@ -355,25 +392,72 @@ impl<I: Send + Sync + 'static, O: Send + Sync + 'static> AllNodeConnectivities<I
...
         fn sender_connection_id(&self) -> u32 {
             *self as u32
         }
     }
 
+    impl AdvertiseProtocolVersion for usize {
+        fn peer_protocol_version(&self) -> CommunicationProtocols {
+            CURRENT_PROTOCOL_VERSION
+        }
+    }
```
netrome: This is dead code, let's remove it

```suggestion
```

netrome: Looked deeper (with some Claude assistance). Seems like this is dead code, how about we remove it? 
```suggestion
```

crates/node/src/providers/ecdsa.rs
netrome: Thanks!

crates/node/src/providers/robust_ecdsa.rs
```diff
@@ -76,22 +76,25 @@ impl RobustEcdsaSignatureProvider {
...
         id: UniqueId,
         domain_id: DomainId,
-    },
+    } = 2,
     Signature {
         id: SignatureId,
         presignature_id: UniqueId,
-    },
+    } = 3,
 }
```
netrome: Nice!

crates/node/src/protocol_version.rs
```diff
@@ -16,3 +16,48 @@ pub enum CommunicationProtocols {
     #[num_enum(catch_all)]
     Unknown(u32),
 }
+
+impl CommunicationProtocols {
+    /// Whether a peer advertising `self` understands everything introduced up to `required`.
+    /// Versions are compared by wire value, so a peer on a future version passes.
+    pub fn supports(self, required: CommunicationProtocols) -> bool {
+        u32::from(self) >= u32::from(required)
```
netrome: Isn't this incorrect? Historically, protocol versions only supported each other on exact equality matches. I'd rather match on an explicit version matrix here if needed.

For forward compatibility, we could maintain a placeholder for next version and assume the next version supports the current version.

crates/node/src/protocol_version.rs
```diff
@@ -16,3 +16,48 @@ pub enum CommunicationProtocols {
...
+    }
```
netrome: Hmm, I guess this makes sense - but it could be interpreted as a strong commitment to backwards compatibility. This is incorrect for historic versions. Not sure how to best handle that though. Could be worth highlighting that we allow peers on future versions for forwards compatibility, and the responsibility is on the peer with the highest version to decide if they are compatible or not.

netrome (changes requested):
Nice stuff. Only blocker is the dead implementation of `AcvertiseProtocolVersion` on `usize`

crates/node/src/network/conn.rs
gilcu3: Sounds good to me

crates/node/src/network/conn.rs
gilcu3: indeed, would be much better if we didn't have this. Hopefully at some point we move to a single bidirectional connection and it disappears.

crates/node/src/protocol_version.rs
gilcu3: reworded comment to make that explicit

crates/node/src/protocol_version.rs
```diff
@@ -17,41 +17,45 @@ pub enum CommunicationProtocols {
...
-    /// Whether a peer advertising `self` understands everything introduced up to `required`.
-    /// Versions are compared by wire value, so a peer on a future version passes.
-    pub fn supports(self, required: CommunicationProtocols) -> bool {
+impl NetworkProtocolVersion {
+    /// Whether a peer advertising `self` can be assumed to handle a feature introduced in
+    /// `required`. Comparison is by wire value, so a peer on a version newer than this binary
+    /// recognizes also passes; whether it is in fact still compatible is that peer's call to
+    /// make, not ours. This is not a general backwards-compatibility claim: support for old
+    /// versions is dropped by folding them into [`NetworkProtocolVersion::Unsupported`], which
+    /// fails every check.
```
netrome: Feels a bit LLMy - but seems helpful. Thanks!

netrome (dismissed):
Thank you for updating 🙏

crates/node/src/providers/ckd.rs
```diff
@@ -24,11 +24,14 @@ use crate::{
     types::{CKDId, SignatureId},
 };
 
+/// Discriminants are wire format: only ever append, never reorder. See [`MpcTaskId`].
```
kevindeforth: I think this comment is worded too strongly. If we only support the current and previous protocol version, we can re-use deprecated task ids.

But, it's worth to add a unittest to ensure the invariant is upheld.

crates/node/src/providers/ecdsa.rs
```diff
@@ -124,27 +124,30 @@ impl EcdsaSignatureProvider {
     }
 }
 
+/// Discriminants are wire format: only ever append, never reorder. See [`MpcTaskId`].
```
kevindeforth: idem

crates/node/src/providers/eddsa.rs
```diff
@@ -63,11 +63,14 @@ impl EddsaSignatureProvider {
     }
 }
 
+/// Discriminants are wire format: only ever append, never reorder. See [`MpcTaskId`].
```
kevindeforth: idem

crates/node/src/providers/ckd.rs
```diff
@@ -24,11 +24,14 @@ use crate::{
...
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, BorshSerialize, BorshDeserialize)]
+#[borsh(use_discriminant = true)]
+#[repr(u8)]
```
kevindeforth: The comments instructions are a bit too narrow.
introducing `use_discriminant = true` allows us to re-order task-ids (even though it is not advisable, because it makes scanning for the next task-id a pain).

It's also okay to delete task-ids once we know that no deployed node will use it again and that the protocol would refuse connecting with a previous version that supported said task id.

We can also consider introducing a `network/wire_formats.rs` or `network/task_ids.rs` file, where we centralize all of these (since right now, they live in multiple different files).

This would help us with two things:

1. Someone working in `providers/ckd.rs`, `providers/ecdsa.rs` or `providers/eddsa.rs` might not expect to be working with wire-formats. Same when reviewing.
2. It gives us one place to record which discriminants exist, when each was introduced, and which are retired. This makes testing this in CI much easier.


So, roughly, what I am suggestin:

1. Move the enums into one file.
2. introduce struct for metadata:
```
pub struct TaskWire {
    pub enum_name: &'static str,
    pub discriminant: u8,
    pub variant: &'static str,
    pub payload: &'static str,
    pub introduced: ProtocolVersion,
    pub retired: Option<ProtocolVersion>,
}
```
3. Keep the metadata next to the enum via macro:
```rust
wire_enum! {
    pub enum RobustEcdsaTaskId {
        0 => KeyGeneration { key_event: KeyEventId }, since V1, retired V3;
        1 => KeyResharing  { key_event: KeyEventId }, since V1;
        2 => Presignature  { id: UniqueId, domain_id: DomainId }, since V1;
        3 => Signature     { id: SignatureId, presignature_id: UniqueId }, since V1;
    }
}
```
4. check it in CI against snapshot:
```rust
pub fn tasks_at(v: ProtocolVersion) -> impl Iterator<Item = &'static TaskWire> {
    WIRE_FORMATS
        .iter()
        .filter(move |t| t.introduced <= v && t.retired.is_none_or(|r| v < r))
}
```

We could then even have checks where we compare snapshots in across different releases, ensuring we don't accidentally change a wire-format for an already released version.

crates/node/src/providers/ecdsa.rs
kevindeforth: idem

crates/node/src/providers/eddsa.rs
kevindeforth: idem

crates/node/src/providers/robust_ecdsa.rs
```diff
@@ -76,22 +76,25 @@ impl RobustEcdsaSignatureProvider {
     }
 }
 
+/// Discriminants are wire format: only ever append, never reorder. See [`MpcTaskId`].
```
kevindeforth: idem

crates/node/src/providers/verify_foreign_tx.rs
```diff
@@ -206,12 +206,15 @@ pub struct VerifyForeignTxProvider {
     ecdsa_signature_provider: Arc<EcdsaSignatureProvider>,
 }
 
+/// Discriminants are wire format: only ever append, never reorder. See [`MpcTaskId`].
```
kevindeforth: idem

crates/node/src/primitives.rs
```diff
@@ -205,12 +205,15 @@ pub struct MpcMessage {
     pub kind: MpcMessageKind,
 }
 
+/// Discriminants are wire format: only ever append, never reorder. See [`MpcTaskId`].
```
kevindeforth: idem

crates/node/src/primitives.rs
```diff
@@ -257,13 +260,17 @@ pub struct MpcPeerMessage {
     pub message: MpcMessage,
 }
 
+/// Discriminants are explicit and part of the wire format: a variant is only ever appended,
+/// never reordered or renumbered, so that nodes on different versions keep decoding each other.
```
kevindeforth: idem

crates/node/src/primitives.rs
```diff
@@ -291,8 +298,69 @@ pub struct Version {
...
+    fn mpc_message_kind__should_keep_borsh_discriminants_stable(
+        #[case] kind: MpcMessageKind,
+        #[case] discriminant: u8,
+    ) {
+        // When
+        let encoded = borsh::to_vec(&kind).unwrap();
+
+        // Then
+        assert_eq!(encoded[0], discriminant);
+    }
```
kevindeforth: I'd prefer we use a snapshot test for these. They're easier to review, extend and and catch changes.

crates/node/src/primitives.rs
```diff
@@ -291,8 +298,69 @@ pub struct Version {
...
+    #[case(EddsaTaskId::Signature { id: CryptoHash::default() }.into(), 1, 2)]
+    #[case(CKDTaskId::KeyGeneration { key_event: key_event() }.into(), 2, 0)]
+    #[case(CKDTaskId::KeyResharing { key_event: key_event() }.into(), 2, 1)]
+    #[case(CKDTaskId::Ckd { id: CryptoHash::default() }.into(), 2, 2)]
+    #[case(RobustEcdsaTaskId::KeyGeneration { key_event: key_event() }.into(), 3, 0)]
+    #[case(RobustEcdsaTaskId::KeyResharing { key_event: key_event() }.into(), 3, 1)]
+    #[case(RobustEcdsaTaskId::Presignature { id: uid(), domain_id: DomainId(0) }.into(), 3, 2)]
+    #[case(RobustEcdsaTaskId::Signature { id: CryptoHash::default(), presignature_id: uid() }.into(), 3, 3)]
+    #[case(VerifyForeignTxTaskId::VerifyForeignTx { id: CryptoHash::default(), presignature_id: uid() }.into(), 4, 0)]
+    fn mpc_task_id__should_keep_borsh_discriminants_stable(
```
kevindeforth: Idem - snapshot seems a better fit for these tests

crates/node/src/network.rs
```diff
@@ -184,6 +185,39 @@ impl MeshNetworkClient {
         self.transport_sender.all_participant_ids()
     }
 
+    /// The protocol version `participant` advertised in its handshake; `None` while it is not
+    /// connected. Our own version is always known.
```
kevindeforth: This confused me:

> Our own version is always known

Why wouldn't we?

crates/node/src/network.rs
```diff
@@ -184,6 +185,39 @@ impl MeshNetworkClient {
...
+        }
+    }
+
+    /// Participants, ourselves included, known to run at least `required`. Intended to pick a
+    /// participant set before a task starts, so a peer is never chosen for work it cannot decode.
+    ///
+    /// This is a version filter only, and a peer connected in a single direction already
+    /// qualifies. Callers choosing a participant set must still intersect the result with
+    /// [`Self::all_alive_participant_ids`].
+    // TODO(#4399): drop the attribute, the online-presign leader selects participants with this.
```
kevindeforth: Please edit this comment such that it aligns with our engineering standard.

I found it much easier to read the code than the comment.

crates/node/src/protocol_version.rs
kevindeforth: agree, I think this is a useful comment, but would prefer if it was written in a more human-way.

crates/node/src/network/conn.rs
kevindeforth: We can take the minimum of both versions 😆

crates/node/src/p2p.rs
```diff
@@ -128,11 +135,15 @@ impl Drop for DropToCancel {
     }
 }
 
+/// Discriminants are wire format: only ever append, never reorder.
+/// See [`MpcTaskId`](crate::primitives::MpcTaskId).
```
kevindeforth: Yeah, this would then also be part of a `wire_format.rs` module.

crates/node/src/network/conn.rs
```diff
@@ -217,6 +218,11 @@ pub trait SenderConnectionId {
     fn sender_connection_id(&self) -> u32;
 }
 
+/// The protocol version the remote side of a connection advertised in the handshake.
+pub trait AdvertiseProtocolVersion {
```
kevindeforth: The trait-name doesn't make it immediately clear that this is the peers version. It could also mean "advertise our own protocol version".

How about:
`HasPeerNetworkProtocolVersion`?

crates/node/src/network.rs
```diff
@@ -184,6 +185,39 @@ impl MeshNetworkClient {
         self.transport_sender.all_participant_ids()
     }
 
+    /// The protocol version `participant` advertised in its handshake; `None` while it is not
+    /// connected. Our own version is always known.
+    pub fn peer_protocol_version(
```
kevindeforth: nit: we call it interchangeably `protocol_version` and `network_protocol_version`. Might be good to be consistent (my preference would be `network_protocol_version`).

kevindeforth (commented):
Thank you!

This is a nice improvement. I have some suggestions (https://github.com/near/mpc/pull/4433#discussion_r4068531223) aimed at integrating this in a way that makes editing and reviewing code changes related to wire format easier. Lmk what you think. 
If you want, I can also do this in a follow-up, but diff size would be smaller now than later.

crates/node/src/primitives.rs
gilcu3: For this one at least I believe the current `rstest` is as easy to review and extend as a snapshot test, because the result of each case is a single line. They also have the advantage of not requiring a new file that needs to be updated in two steps, so I would rather keep this test instead

crates/node/src/primitives.rs
gilcu3: Idem to the previous one. If what we are fixing fits on a single line, I think this is more developer friendly than the snapshots

crates/node/src/network.rs
gilcu3: I think this comment is just redundant, the code is as easy to read, will remove

crates/node/src/network/conn.rs
gilcu3: Ups no :upside_down_face:

crates/node/src/providers/ckd.rs
netrome: > The comments instructions are a bit too narrow.
introducing use_discriminant = true allows us to re-order task-ids (even though it is not advisable, because it makes scanning for the next task-id a pain).

I guess that depends on what you define as "reorder". Reordering the declarations - yes, reordering the fields on the wire (e.g. swapping discriminants) - no.

kevindeforth: > reordering the fields on the wire (e.g. swapping discriminants) - no.

Calling this reordering is a stretch though :D

gilcu3: >We can also consider introducing a network/wire_formats.rs or network/task_ids.rs file, where we centralize all of these (since right now, they live in multiple different files).

this part was handled in https://github.com/near/mpc/pull/4433/commits/e6de459cb4c5ccf4a8890dfa7a218d3bd585ae99

crates/node/src/network/conn.rs
```diff
@@ -355,25 +392,66 @@ impl<I: Send + Sync + 'static, O: Send + Sync + 'static> AllNodeConnectivities<I
...
+            self.0
+        }
+    }
+
+    #[rstest]
+    #[case::both_up(true, true, Some(NetworkProtocolVersion::Dec2025))]
+    #[case::only_outgoing_up(true, false, Some(NetworkProtocolVersion::Dec2025))]
+    #[case::only_incoming_up(false, true, Some(NetworkProtocolVersion::Jan2026))]
+    #[case::neither_up(false, false, None)]
+    fn node_connectivity__should_report_the_protocol_version_of_a_live_connection_only(
```
kevindeforth: nit: Would be nice to write this test in a way that makes adding future versions easy.

crates/node/src/network/wire_format.rs
```diff
@@ -0,0 +1,345 @@
...
+
+impl From<VerifyForeignTxTaskId> for MpcTaskId {
+    fn from(value: VerifyForeignTxTaskId) -> Self {
+        MpcTaskId::VerifyForeignTxTaskId(value)
+    }
+}
+
+#[cfg(test)]
+#[expect(non_snake_case)]
+mod tests {
```
kevindeforth, line 185: I would still prefer a snapshot test for these.

crates/node/src/network.rs
```diff
@@ -184,6 +187,33 @@ impl MeshNetworkClient {
...
+            Some(CURRENT_PROTOCOL_VERSION)
+        } else {
+            self.transport_sender
+                .connectivity(participant)
+                .peer_network_protocol_version()
+        }
+    }
+
+    /// Not a liveness check: a peer connected in a single direction qualifies. Intersect with
+    /// [`Self::all_alive_participant_ids`] when picking a participant set.
```
kevindeforth: Q: I wonder if we should change this method, such that it returns `None` if not bidirectionally connected or there is a mismatch between inbound and outbound version.

I don't see a good use for engaging with a peer that isn't fully connected with us or reports pathological protocol versions.

crates/node/src/network.rs
kevindeforth, line 1026: nit / Q: should we rename this method to indicate that all participants run the same version?
Is there something we wold need to test with differing versions here?

kevindeforth (dismissed):
Thanks for the changes!

I would still much prefer using snapshot tests, as indicated on slack. The current unit tests won't be easy to maintain, review or extend.

crates/node/src/network.rs
gilcu3: seems sensible, as right now we expect this peer to misbehave. If the change is small will add here, else will need to open a follow up

crates/node/src/network/wire_format.rs
gilcu3: Okok will convert them to snapshots :handshake:

crates/node/src/network/conn.rs
gilcu3: hmm will give it a try, but not sure that's possible atm

EDIT: I misinterpreted the comment, it seems that adding a couple of constants should be enough, will do

crates/node/src/network.rs
gilcu3: we don't need to test different versions here (it is covered elsewhere, closer to the actual wire change), but might need to in the future. Leaving as is for now

kevindeforth (dismissed):
Beautiful, thank you!

crates/node/src/network/conn.rs
```diff
@@ -217,6 +218,11 @@ pub trait SenderConnectionId {
     fn sender_connection_id(&self) -> u32;
 }
 
+/// The protocol version the remote side of a connection advertised in the handshake.
+pub trait HasPeerNetworkProtocolVersion {
+    fn peer_network_protocol_version(&self) -> NetworkProtocolVersion;
+}
+
```
netrome, line 225: Tiny naming nit (feel free to ignore): "HasPeerNetworkProtocolVersion" is a predicate, and thus implies the associated function should return a bool. Perhaps "ReadPeerNetworkProtocolVersion" is more idiomatic? That said, I'm perfectly okay with providing an associated getter on something that by the trait has some property, so the current naming is also fine.

crates/node/src/network/wire_format.rs
```diff
@@ -0,0 +1,398 @@
+//! Types whose Borsh encoding is part of the P2P wire format.
+//!
+//! Every enum here sets `#[borsh(use_discriminant = true)]` and spells its discriminants out, so
+//! peers agree on the numbers rather than on declaration order. A number, once assigned, is never
+//! reused or given to another variant. Appending is always safe. Reordering variants leaves the
+//! wire untouched, but makes the next free number harder to spot, so keep them in numeric order.
+//! Retiring a variant is allowed once no deployed node can still send it; the handshake's
+//! [`NetworkProtocolVersion`](crate::protocol_version::NetworkProtocolVersion) check is what
+//! rules those nodes out.
```
netrome: Nit: This docstring feels llm-generated and doesn't provide much that isn't already evident from the code. The risk is that this could easily go stale and diverge from the code in the future. I'd suggest carving this out.

```suggestion
```

crates/node/src/network/wire_format.rs
```diff
@@ -0,0 +1,398 @@
+//! Types whose Borsh encoding is part of the P2P wire format.
```
netrome, line 1: Nice to get these types gathered in this module. Though the "Computation" double vec is a bit ugly, but sorting that one out is a much larger effort 😅

crates/node/src/network/wire_format.rs
```diff
@@ -0,0 +1,398 @@
...
+                    chunks.iter().map(|c| c.len()).sum::<usize>()
+                ))
+                .finish(),
+            MpcMessageKind::Abort(err) => f.debug_tuple("Abort").field(err).finish(),
+            MpcMessageKind::Success => write!(f, "Success"),
+        }
+    }
+}
+
+/// An encoded task id carries two discriminants: this one, then the provider's.
```
netrome: This is a strange and confusing docstring. It reads more like a comment, and "provider's" is an ambiguous term. Let's skip it.
```suggestion
```

netrome (dismissed):
Beautiful stuff! Some nits from my side, but nothing blocking 🙏

crates/node/src/network/conn.rs
gilcu3: Will leave as is, don't have much in favor or against any of them

kevindeforth (approved):

netrome (approved):

PR #4435, opened 2026-09-15
test(contract): make test check contract storage keys properly

### LINKED ISSUE #4434: Flaky sandbox test: upgrade__should_drop_legacy_support_and_preserve_foreign_chains_state
### Background

`upgrade__should_drop_legacy_support_and_preserve_foreign_chains_state` (added in #4404) fails at random on `main`, e.g. https://github.com/near/mpc/actions/runs/34954343407/job/104332738456.

It asserts the legacy per-node support map was reclaimed by scanning contract state with `worker.view_state(..).prefix(&borsh::to_vec(&StorageKey::_DeprecatedSupportedForeignChainsByNode))`. That prefix is a single byte (`0x10`). `near_sdk::store::IterableMap` keeps its values in a `LookupMap` whose default hasher is `Sha256`, so the on-chain key is `sha256(prefix ++ borsh(key))` and retains no prefix at all. The scan therefore misses the map's own value slots and instead matches any unrelated hashed entry whose digest happens to start with `0x10` (~1/256 per entry; the test populates 20 participants with attestations and foreign-chain configs, so a hit is common).

```
legacy map storage must be reclaimed: {[16, 128, 6, 174, ...]: [43, 0, 0, 0, 100, 101, 118, 45, ...]}
```

The reported leftover decodes to `(dev-<...>.test.near, ...)` plus a trailing `ValueAndIndex` index, i.e. an entry of another account-keyed collection, not a `BTreeSet<ForeignChain>`. The migration itself is correct; only the assertion is unsound.

### Acceptance Criteria

- The test checks the exact storage keys the legacy `IterableMap` owns (`prefix ++ b"v" ++ index_le` for the key `Vector`, `sha256(prefix ++ b"m" ++ borsh(account_id))` for the value `LookupMap`) instead of a prefix scan, so both the pre-upgrade and post-upgrade assertions are deterministic.

gilcu3:
Closes #4434


anodar (approved):
Thank you!

netrome (approved):

PR #4439, opened 2026-09-15
test(node): make sure assets are warm before running integration tests

### LINKED ISSUE #4426: per_domain_reconstruction_threshold__should_gate_signing_availability_when_nodes_go_offline is flaky
### Background

`tests::reconstruction_thresholds::per_domain_reconstruction_threshold__should_gate_signing_availability_when_nodes_go_offline` fails intermittently in the `Cargo test: node` CI job. It failed 5 times between 2026-09-04 and 2026-09-15, across unrelated branches (dependabot, feature branches, and once in the merge queue): runs [34939040099](https://github.com/near/mpc/actions/runs/34939040099), [34893735075](https://github.com/near/mpc/actions/runs/34893735075), [34831028641](https://github.com/near/mpc/actions/runs/34831028641), [34830129327](https://github.com/near/mpc/actions/runs/34830129327), [33856768336](https://github.com/near/mpc/actions/runs/33856768336). Over that window the job ran 171 times to completion, so the flake rate is roughly 3%. The `mpc-node` job uses the default nextest profile with no retries, so a single flake fails the job and blocks the merge queue.

All five failures are identical: the `assert_can_sign(&mut setup.indexer, "user_4_high", &high)` call at `crates/node/src/tests/reconstruction_thresholds.rs:187`, in the "one node down (4 online)" stage, for the CaitSith domain `DomainId(1)` with `t=4`.

```
[INFO  mpc_node::tests] Timed out waiting for signature response for user user_4_high
thread '...per_domain_reconstruction_threshold__should_gate_signing_availability_when_nodes_go_offline' panicked at crates/node/src/tests/reconstruction_thresholds.rs:54:5:
domain DomainId(1) (t=4) should be able to sign with the currently-online nodes
```

The preceding `warm_up` of the same domain succeeds in every failing run, so the request that times out is the one bounded by the 10s `REQUEST_WAIT_BUDGET`, not the 60s warmup. Likely the 4-online cluster has not replenished CaitSith presignatures for `DomainId(1)` within 10s after the warmup consumed one, while `robust_ecdsa` presign attempts spin in the background (`Not enough active participants: need 5, got 4`) on a loaded CI runner.

### Acceptance Criteria

- The test no longer fails intermittently on CI, either by removing the dependency on presignature replenishment timing at the `user_4_high` stage or by giving that assertion a budget that holds on a loaded runner without weakening the negative assertions (`REQUEST_WAIT_BUDGET` is shared with `assert_cannot_sign`, which is only meaningful while it exceeds the worst-case positive latency).

gilcu3:
Closes #4426

This fix is a bit of a band aid, but at least a very simple one, in the sense that we cannot guarantee that the test is not flaky. I tested locally and before this change flakiness was reproducible locally, while that is not the case anymore after. The underlying problem is that the node does not expose a way to check if the assets were already computed, that is easily readable for these integration tests. So we have three options:

- Disable it, would be easy but we would lose some coverage until we manage to figure out a different way of testing that.
- Make changes in the node so that the info is available. I tried that but the changes seem quite involved for such a tiny problem.
- What is done here, make sure that it will not fail in most cases (should be 99% good now IIUC).


crates/node/src/tests/reconstruction_thresholds.rs
```diff
@@ -25,6 +25,23 @@ const REQUEST_WAIT_BUDGET: std::time::Duration = std::time::Duration::from_secs(
...
 
+/// Warm-up signatures per domain. An online-set change strands most of a node's buffered
+/// presignatures on the participants that left, and the first request to find the buffer empty pays
+/// for generating a fresh one; the rest run against a buffer that generation is keeping up with.
+const WARMUP_SIGNATURES: usize = 5;
+
+/// Presignatures each node buffers per domain, raised from the default 5 so that the presignatures
+/// a departing node strands rarely leave a node with none at all. Stays within the triples a single
+/// generation batch yields.
+const PRESIGNATURES_TO_BUFFER: usize = 16;
```
kevindeforth, line 36: why don't we wait until we have all `PRESIGNATURES_TO_BUFFER`?

_A: because we don't have that information available in integration tests_

crates/node/src/tests/reconstruction_thresholds.rs
```diff
@@ -40,13 +57,15 @@ async fn request_and_await_response(
...
+/// Primes each domain's presignatures for the current online set with [`WARMUP_SIGNATURES`]
+/// generously-budgeted signs, whose cold-start can exceed [`REQUEST_WAIT_BUDGET`]. Only CaitSith and
+/// DamgardEtAl consume pre-generated presignatures; Frost and CKD sign directly, so pass only the
+/// block's signable CaitSith/DamgardEtAl domains after every online-set change.
 async fn warm_up(indexer: &mut FakeIndexerManager, domains: &[&DomainConfig]) {
     for domain in domains {
-        let _ = request_and_await_response(indexer, "warmup", domain, WARMUP_WAIT_BUDGET).await;
+        for _ in 0..WARMUP_SIGNATURES {
+            let _ = request_and_await_response(indexer, "warmup", domain, WARMUP_WAIT_BUDGET).await;
+        }
```
kevindeforth, line 68: So, we submit a request to ensure nodes are "warm"? And this is existing code? 🙃

kevindeforth (approved):
> This fix is a bit of a band aid

this is an understatement :D 

> but at least a very simple one

the code is simple, the explanation for why it works isn't though :D 

> Make changes in the node so that the info is available. I tried that but the changes seem quite involved for such a tiny problem.

I think this would be ideal. But making ci non-flaky is higher prio. Happy to band-aid and fix properly afterwards. Thank you Sir!

netrome (approved):
Sure, thanks!

PR #4452, opened 2026-09-16
chore: Bump nearcore to 2.14.0-rc.2

### LINKED ISSUE #690: Update rocksdb version
### Background

The current version of rocksdb does not build correctly with gcc15. The workaround identified a few weeks ago was to run the build:
`CXXFLAGS="$CXXFLAGS -include cstdint" cargo build`
A fix is already merged upstream [link](https://github.com/rust-rocksdb/rust-rocksdb/pull/1007) but has not been released yet.

### User Story

Compiling should not require extra flags.

### Acceptance Criteria

The dependency is updated and the extra flag is no longer required to compile.

`rocksdb` was downgraded in [nearcore](https://github.com/near/nearcore/pull/12807)

### LINKED ISSUE #4117: Drop the `im` advisory ignores once nearcore moves `near-network` off `im`
## Background

`cargo deny check advisories` fails on RUSTSEC-2026-0247 (bitmaps), -0248 (im) and -0251 (sized-chunks). All three arrive through `im v15.1.0`, pulled by `near-network`; the `im` repository was archived on 2026-05-03 and the advisory points at `imbl` as the maintained fork. They are ignored in `deny.toml` so CI is not blocked on a dependency we do not control.

## Acceptance Criteria

- The three ignores are removed from `deny.toml` once nearcore no longer pulls `im`, and `cargo deny --all-features check` still passes.

- <https://rustsec.org/advisories/RUSTSEC-2026-0248>

### LINKED ISSUE #4451: Bump nearcore to 2.14.0-rc.1
### Background

Nearcore just released 2.14.0-rc.1 (see https://github.com/near/nearcore/releases/tag/2.14.0-rc.1). We need to bump it and get this out to testnet before Wednesday next week to prevent downtime.

### User Story

As an MPC operator I'd like my node to be able to index the near blockchain

### Acceptance Criteria

We've bumped nearcore to 2.14.0-rc.1

_No response_

netrome:
closes #4451


crates/node/src/indexer/configs.rs
```diff
@@ -13,6 +13,9 @@ impl IndexerConfigExt for IndexerConfig {
             await_for_node_synced: near_indexer::AwaitForNodeSyncedEnum::StreamWhileSyncing,
             finality: self.finality.clone(),
             validate_genesis: self.validate_genesis,
+            // Skipping a block would silently drop the signature requests it carries,
+            // so we let the streamer terminate and rely on a restart to re-index.
+            skip_broken_blocks: false,
```
netrome, line 16: This is a new parameter. Setting it to false retains the pre-2.14 behavior. Proceeding past broken blocks sounds harmful to me as well, but I don't understand the implications deeper. Happy to dive into it if anyone wants to push for changed behavior here.

crates/chain-gateway/src/event_subscriber/streamer/block_processor.rs
```diff
@@ -172,13 +173,27 @@ impl BlockEvents {
...
+        let predecessor_id = match from_near_internal(&receipt.predecessor_id) {
+            Ok(predecessor_id) => predecessor_id,
+            Err(err) => {
+                tracing::error!(
+                    %err,
+                    receipt_id = %receipt.receipt_id,
+                    predecessor_id = %receipt.predecessor_id,
+                    "skipping executor event: predecessor is not a valid account ID for this \
+                     workspace's `near-account-id`",
+                );
```
netrome, line 185: Not too happy about this, but Iiuc this error arm is effectively unreachable in practice and as soon as near_account_id supports the new version we can bump and get rid of this conversion.

netrome:
> **PR title type suggestion:** This PR updates an external dependency (nearcore), which is a build system change. Consider using `build:` instead of `chore:` to be more specific.
> 
> Suggested title: `build: Bump nearcore to 2.14.0-rc.1`

Thanks, git-cliff doesn't recognize "build" currently so I'll keep it as a chore for now. In the future, I think adding a recognized "build" prefix could make sense.

crates/node/src/indexer/handler.rs
```diff
@@ -303,14 +306,32 @@ fn try_extract_function_call_args(receipt: &ReceiptView) -> Option<(&FunctionArg
...
+                receipt_id = %receipt.receipt_id,
+                predecessor_id = %receipt.predecessor_id,
+                "skipping `{method_name}`: predecessor is not a valid account ID for this \
+                 workspace's `near-account-id`",
+            );
+            metrics::MPC_INDEXER_NUM_UNCONVERTIBLE_PREDECESSOR_IDS.inc();
+            None
+        }
+    }
+}
```
netrome, line 325: Same here. The error path should effectively be dead code, and we'll be able to get rid of this when we upgrade "near_account_id".

.github/dependabot.yml
```diff
@@ -26,8 +26,13 @@ updates:
       - dependency-name: "httpmock"
       # This cannot be upgraded because it conflicts the prometheus version required by nearcore
       - dependency-name: "prometheus"
-      # TODO(#690): This cannot be upgraded until nearcore does
+      # Must match nearcore's `librocksdb-sys`; see the workspace Cargo.toml
       - dependency-name: "rocksdb"
```
kevindeforth, line 30: I guess that means we need to keep ignoring this in dependabot and can probably close issue #690 ?

crates/chain-gateway/src/event_subscriber/streamer/block_processor.rs
kevindeforth: Feels odd that `near_account_id` isn't already supporting it.

crates/node/src/indexer/configs.rs
kevindeforth: Yeah, skipping broken blocks sounds desirable, but I don't yet know what a broken block is.

crates/test-utils/src/lib.rs
```diff
@@ -6,4 +6,4 @@ pub mod contract_types;
 /// Single source of truth shared by the e2e-tests crate and the contract test crates.
 /// `scripts/check-sandbox-image-version.sh` enforces that this stays in lockstep with
 /// the workspace's nearcore tag.
-pub const DEFAULT_SANDBOX_VERSION: &str = "2.13.4";
+pub const DEFAULT_SANDBOX_VERSION: &str = "2.14.0-rc.1";
```
kevindeforth: should this be `rc-2` as well?

deny.toml
```diff
@@ -34,13 +33,9 @@ ignore = [
...
   "RUSTSEC-2026-0222", # wasmtime stores can mix up type indices between engines
   "RUSTSEC-2026-0269", # wasmtime-wasi filesystem sandbox escape via trailing slashes
-  # TODO(#4117): remove once nearcore moves `near-network` off `im`, whose maintained fork is `imbl`.
-  "RUSTSEC-2026-0247", # bitmaps is unmaintained
-  "RUSTSEC-2026-0248", # im is unmaintained
-  "RUSTSEC-2026-0251", # sized-chunks is unmaintained
   # TODO(#4423): remove once nearcore's aws-lc-rs pin lets cargo resolve rustls >= 0.23.45.
   # Fixed in rustls >= 0.23.45, which requires aws-lc-rs ^1.18; nearcore pins aws-lc-rs to
-  # =1.16.2 (2.13.4 and master), so the fix cannot be resolved until that pin moves. The
+  # =1.16.2 (2.14.0-rc.1 and master), so the fix cannot be resolved until that pin moves. The
```
kevindeforth: idem, rc-2?

kevindeforth (dismissed):
Thank you!

deny.toml
```diff
@@ -34,13 +33,9 @@ ignore = [
   # shares a Store across Engines and does not depend on `wasmtime-wasi`.
   "RUSTSEC-2026-0222", # wasmtime stores can mix up type indices between engines
   "RUSTSEC-2026-0269", # wasmtime-wasi filesystem sandbox escape via trailing slashes
-  # TODO(#4117): remove once nearcore moves `near-network` off `im`, whose maintained fork is `imbl`.
```
kevindeforth, line 37: note: issue to close

kevindeforth (dismissed):

crates/test-utils/src/lib.rs
netrome: Yep it should. Updated. That's what I get for trying to bump to rc.2 myself without Claude :sweat_smile:

deny.toml
netrome: Good point, linked it

netrome: Good point, linked it 
image

deny.toml
netrome: Amended in https://github.com/near/mpc/commit/90b3f6821a20449f26857f098e60da15c22c3e8c

crates/test-utils/src/lib.rs
netrome: Amended the last commit to contain this https://github.com/near/mpc/commit/90b3f6821a20449f26857f098e60da15c22c3e8c

.github/dependabot.yml
netrome: Yeah good point. Linked in the issue
image

crates/chain-gateway/src/event_subscriber/streamer/block_processor.rs
netrome: Well, it's just an "rc" release so it kind of makes sense. But yeah.

kevindeforth (approved):

crates/node/src/indexer/configs.rs
barakeinav1: so is this something new? is there a default value?

netrome: Yeah this is a new parameter, I believe we need to set it explicitly but this value seems both correct and preserves the same behavior we had before this parameter was introduced.

barakeinav1 (approved):
approved.

one important comment by claude "The medium one is the RocksDB 8.1 → 10.4 bump making the node's data directory unreadable by older binaries"

so this means we can't roll back!!!
so we need to be extra careful when upgrading the node.

netrome:
> approved.
> 
> one important comment by claude "The medium one is the RocksDB 8.1 → 10.4 bump making the node's data directory unreadable by older binaries"
> 
> so this means we can't roll back!!! so we need to be extra careful when upgrading the node.

Oh thanks for flagging. Ideally we should have upgrade/downgrade tests to ensure we notice these things early. I can try both upgrading and downgrading in a localnet just to see if this holds or not.

crates/node/src/indexer/configs.rs
metalurgical: > Yeah, skipping broken blocks sounds desirable, but I don't yet know what a broken block is.

From what I can tell a broken block appears to be a block for which the header can be retrieved but the receipts cannot. So corrupt/incomplete state rather than an actual block that was included incorrectly. 

Think the default should be false here.

PR #4455, opened 2026-09-17
feat(contract): bound the participant url with a BoundedString type

### LINKED ISSUE #4449: Bound the length of the participant url
### Background

`ParticipantInfo.url` has no length limit. Nothing validates it in `update_participant_url`, in `Participants::update_info`, or in `Participants::validate`.

The field is stored inline in the contract state, which every method deserializes on every call, so an oversized url raises the cost of every contract call rather than just the one that set it.

### User Story

As a node operator, I want participant metadata to be bounded, so that contract call costs stay predictable regardless of what another operator registers.

### Acceptance Criteria

- The url length is validated on every path that writes a `ParticipantInfo`.
- The bound leaves clear headroom over real usage (longest url in the mainnet set today is 48 bytes).
- Tests cover the boundary in both directions.

Follows #3677, which added `update_participant_url`.

barakeinav1:
Closes #4449

Alternative to #4448, taking @kevindeforth's suggestion. Same goal, with the interface type left for #4456 — see below for why.

### What's here

- `BoundedString<U>` in `near-mpc-bounded-collections`, alongside `BoundedVec` and copying its shape: serialization delegates to the inner `String`, deserialization validates.
- `MAX_PARTICIPANT_URL_BYTES = 256` and `ParticipantUrl` in `near-mpc-contract-interface`.
- Contract-internal `ParticipantInfo.url: ParticipantUrl`, and `dtos::ParticipantInfo -> ParticipantInfo` becomes fallible. That conversion is now the single enforcement point: every entry point that can set a url routes through it (`init`, `init_running`, `vote_new_parameters`, `update_participant_url`, `start_node_migration`, `conclude_node_migration`), and the compiler finds any new one.

### What's deliberately not here

`dtos::ParticipantInfo.url` stays a plain `String`, so `update_participant_url` keeps its signature and the node is untouched. Two reasons, tracked in #4456:

1. **Node-first upgrades.** The node parses `state()` into `dtos::ProtocolContractState`. A bounded DTO makes a new node stricter than the contract it is talking to — while 3.15.1 is live (no url validation) one long url would stop every upgraded node from reading state.
2. **`ongoing_migrations`.** It is an `IterableMap<AccountId, DestinationNodeInfo>` holding the *interface* type, and `migrate()` doesn't touch lazy maps. A record written before the upgrade would never be validated, and afterwards could not be read, concluded, or even removed — `remove` returns the value, so it deserializes too. So `start_node_migration` applies the bound explicitly instead.

### Upgrade behaviour — read this

`migrate()` reads state through the *current* types (`v3_15_1_state` reuses `ProtocolContractState`), so **the bound is also an upgrade gate**: if any stored url exceeds 256 bytes at upgrade time, `migrate` fails and the deploy rolls back atomically, leaving the old contract in place.

This is intended, not incidental — a url that is no longer allowed should block the upgrade rather than be silently truncated. It is recorded on `v3_15_1_state` so a later reader does not take it for an oversight and add an unbounded shadow.

The practical consequence: **check the participant set before deploying.** Longest url today is 48 bytes on mainnet and 67 on testnet, so the gate is satisfied — but nothing prevents a longer one being registered before the release, since the live contract has no limit.

### Verified

- **No wire-format change.** `BoundedString` round-trips as a plain string in serde and produces byte-identical borsh to `String` — both pinned by tests.
- **Stored bytes unchanged.** The borsh schema snapshot changes `"String"` → `"BoundedString<256>"` with the same `length_width: 4` / `elements: "u8"` encoding, so no data migration is required. (Readability is gated — see above.)
- **No ABI schema change.** The ABI snapshot diff is doc bullets only, since the DTO is untouched.
- Whole workspace compiles under `--all-features`; 408 contract lib tests, 25 in-process, and the `BoundedString` unit tests pass; clippy clean with `-D warnings`.

### Sizing

Longest url in the mainnet participant set is 48 bytes (mean 37), testnet 67. Zero ongoing node migrations on either network.


crates/contract/src/primitives/participants.rs
```diff
@@ -14,7 +16,7 @@ pub mod hpke {
 #[near(serializers=[borsh])]
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
 pub struct ParticipantInfo {
-    pub url: String,
+    pub url: ParticipantUrl,
```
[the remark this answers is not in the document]
barakeinav1: Fixed in a2e2e5a — `bench-contract-methods` is on in CI via `--all-features`, which my local `--all-targets` check missed. Good catch.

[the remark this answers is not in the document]
barakeinav1: Correct, and this corrected the PR description — thanks. `v3_15_1_state` reuses the current `ProtocolContractState`, so `migrate()` does read old state through the bounded type and a url over 256 bytes fails the upgrade.

That is the behaviour we want: a url that is no longer allowed should roll the deploy back rather than be silently truncated, and the rollback is atomic so the old contract stays live. An unbounded shadow would mean copying the whole `ProtocolContractState -> ... -> ParticipantInfo` chain, which is what that module’s doc warns against on binary-size grounds.

Recorded on `v3_15_1_state` in dbe0752 so it is not taken for an oversight later, and called out in the PR description with the pre-deploy check. Mainnet max is 48 bytes, testnet 67.

crates/contract/src/dto_mapping.rs
```diff
@@ -187,25 +187,35 @@ impl TryIntoContractType<TcbInfo> for dtos::TcbInfo {
...
+
+    fn try_into_contract_type(self) -> Result<ParticipantInfo, Self::Error> {
+        let len = self.url.len();
+        Ok(ParticipantInfo {
+            url: dtos::ParticipantUrl::new(self.url).map_err(|_| {
+                InvalidCandidateSet::ParticipantUrlTooLong {
+                    len,
+                    max: dtos::MAX_PARTICIPANT_URL_BYTES,
+                }
+            })?,
```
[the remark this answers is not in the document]
barakeinav1: Fixed in dbe0752 — added `vote_new_parameters__should_reject_a_url_over_the_byte_limit`. It mutates the wire form directly, since the internal type cannot represent an oversized url, which is the reason that conversion is the enforcement point.

barakeinav1:
> `crates/contract/src/bench.rs:102` and `crates/contract/src/bench.rs:119` — these two `ParticipantInfo` literals were not updated

Fixed in a2e2e5a — CI caught it too. `--all-features` is the difference; my local `--all-targets` run didn't build the feature-gated module.

> the `String -> ParticipantUrl` conversion and its `ParticipantUrlTooLong` mapping are written twice

Fixed in dbe0752 — both `update_participant_url` and `start_node_migration` now route through `TryIntoContractType<ParticipantInfo>`, which deleted the helper. One definition. I kept the `len` recompute: extracting it from the error means a match arm for `LowerBoundError`, which is unreachable for this type.

> `update_participant_url` now runs `participant_url(url)?` *before* `require_deposit`

Fixed in dbe0752 — deposit first in both, so the existing "participant/state errors take precedence" ordering holds and the comment is accurate again.

> that follow-up belongs in an issue referenced as `TODO(#…)`

Fixed in dbe0752 — `TODO(#4456)`, with the invariant left as a single line.

> the parsed `ParticipantUrl` is dropped and the unbounded DTO is stored

Right, and deliberate — #4456 covers it, and the TODO now points there.

> `new__should_bound_by_byte_length` branches on `len <= MAX` inside the Then block

Declined. I read `docs/development/engineering-standards.md` for the convention you cited and it has no rule about branching in Then or one assertion path per test, and the case names (`ascii_at_the_bound`, `multibyte_over_the_bound`) already say which path each takes.

> `new` returns `BoundedVecOutOfBounds`, whose `LowerBoundError` variant is unreachable for this type

Declined for now — consistency with `BoundedVec` in the same crate seems worth more than removing one wildcard.

> `docs/guide/node-migration-guide.md` … does not mention the new 256-byte cap

Fixed in dbe0752 — one clause where the url field is described.

---

On the blocking-adjacent point Copilot raised about `migrate()`: it is right that the bound gates the upgrade, and that is intended — a url that is no longer allowed should roll the deploy back rather than be silently truncated. It is now recorded on `v3_15_1_state` and in the PR description, with the pre-deploy check. Detail in https://github.com/near/mpc/pull/4455#discussion_r4034614725.

crates/near-mpc-contract-interface/src/types/participants.rs
```diff
@@ -1,10 +1,22 @@
...
 use near_mpc_crypto_types::Ed25519PublicKey;
 use serde::{Deserialize, Serialize};
 
 pub use mpc_primitives::ParticipantId;
 
+/// Longest accepted participant url, in bytes.
+///
+/// [`ParticipantInfo`] is stored inline in the contract state, which every contract method
+/// deserializes, so the url is bounded well above what a real endpoint needs.
+pub const MAX_PARTICIPANT_URL_BYTES: usize = 256;
```
kevindeforth: Can we shorten or delete this comment?

crates/near-mpc-contract-interface/src/types/participants.rs
```diff
@@ -1,10 +1,22 @@
...
+
+/// A participant url, bounded at [`MAX_PARTICIPANT_URL_BYTES`].
+///
+/// Serializes as a plain string, so the wire format is unchanged.
+pub type ParticipantUrl = BoundedString<MAX_PARTICIPANT_URL_BYTES>;
```
kevindeforth: This comment is paraphrasing code and referring to behavior defined elsewhere. Please update the comment to agree with our engineering guidelines.

crates/near-mpc-bounded-collections/src/bounded_string.rs
```diff
@@ -0,0 +1,207 @@
...
+///
+/// Serialization delegates to the inner [`String`], so the JSON and borsh representations are
+/// identical to an unbounded string: adopting this type on a field changes neither the wire
+/// format nor stored bytes.
+///
+/// Deserialization is the enforcement point: a value over the bound is rejected, not truncated.
+/// Adopting it on an already-persisted field therefore needs care — data written before the
+/// bound existed becomes unreadable.
+#[derive(PartialEq, Eq, Debug, Clone, Hash, PartialOrd, Ord)]
+pub struct BoundedString<const U: usize>(String);
```
kevindeforth, line 6: This comment is leaking a lot of LLM-context. Lets keep the very first sentence, but remove the rest please.

crates/near-mpc-bounded-collections/src/bounded_string.rs
```diff
@@ -0,0 +1,207 @@
...
+    }
+}
+
+impl<const U: usize> std::fmt::Display for BoundedString<U> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+impl<const U: usize> TryFrom<String> for BoundedString<U> {
```
kevindeforth, line 35: would be nice to also have a `TryFrom<&str> for BoundedString<U>`

crates/near-mpc-bounded-collections/src/bounded_string.rs
```diff
@@ -0,0 +1,207 @@
...
+        Ok(Self(inner))
+    }
+
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+
+    pub fn into_inner(self) -> String {
+        self.0
+    }
```
kevindeforth: I think this is redundant, since we already have `From<BoundedString<U>> for String` ?

crates/near-mpc-bounded-collections/src/bounded_string.rs
```diff
@@ -0,0 +1,207 @@
...
+                upper_bound: U,
+                got,
+            });
+        }
+        Ok(Self(inner))
+    }
+
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
```
kevindeforth: This one, maybe we can replace it with

```rust
From<&'a BoundedString<U>> for &'a str 
```

crates/near-mpc-bounded-collections/src/bounded_string.rs
```diff
@@ -0,0 +1,207 @@
...
+
+    pub fn into_inner(self) -> String {
+        self.0
+    }
+}
+
+impl<const U: usize> std::ops::Deref for BoundedString<U> {
```
kevindeforth, line 21: Hmm, how do we want people to use this? Do we want them to type `*bounded_string` or `bounded_string.into()`?

I would prefer not to have too many different ways of expressing the same thing.

crates/near-mpc-contract-interface/src/types/participants.rs
kevindeforth: I am wondering if we should also include some checks the node already requires c.f. 
https://github.com/near/mpc/blob/7fc4663711fa24a41421f5e0db6be82ad3b4b5bb/crates/node/src/indexer/participants.rs#L386-L399

crates/contract/src/v3_15_1_state.rs
kevindeforth: We need to update this, we are in v3_16_0.

crates/contract/src/v3_15_1_state.rs
```diff
@@ -76,6 +76,12 @@ impl From<OldTeeState> for TeeState {
 /// Keep this module in sync with [`crate::MpcContract`]: the moment a field's borsh
 /// layout diverges, shadow the old type here (see this module's history for examples) so
 /// state written by the `3.15.1` contract still deserializes during migration.
+///
+/// This module reads the *current* [`ProtocolContractState`], so a bound added to a stored type
+/// is also an upgrade gate: state the new type rejects cannot be migrated. That is deliberate for
+/// [`ParticipantUrl`](crate::primitives::participants::ParticipantUrl) — a url over the bound
+/// fails [`crate::MpcContract::migrate`] and rolls the deploy back, rather than being silently
+/// truncated. Check the participant set before deploying a change of that kind.
```
kevindeforth, line 84: I am okay with this, but we do have the option of pointing to an invalid url instead and handling it after the fact.

crates/contract/tests/sandbox/common.rs
```diff
@@ -68,7 +69,7 @@ pub async fn create_account_given_id(
 
 pub fn gen_participant_info() -> ParticipantInfo {
     ParticipantInfo {
-        url: "127.0.0.1".into(),
+        url: ParticipantUrl::new("127.0.0.1".to_string()).unwrap(),
```
kevindeforth: nit: Maybe we could use a `try_into().unwrap()` here instead?

kevindeforth (commented):
Some questions wrt the API of the bounded string type and the no-node-upgrade policy.

> Node-first upgrades. The node parses state() into dtos::ProtocolContractState. A bounded DTO makes a new node stricter than the contract it is talking to — while 3.15.1 is live (no url validation) one long url would stop every upgraded node from reading state.

We could handle this. We try to deserialize the new dto types. If unsuccessful, we deserialize in the old types. Or alternatively, if the url doesn't serialize properly, we default to an invalid one, like we already do here:

https://github.com/near/mpc/blob/7fc4663711fa24a41421f5e0db6be82ad3b4b5bb/crates/node/src/indexer/participants.rs#L418


> ongoing_migrations. It is an IterableMap<AccountId, DestinationNodeInfo> holding the interface type, and migrate() doesn't touch lazy maps. A record written before the upgrade would never be validated, and afterwards could not be read, concluded, or even removed — remove returns the value, so it deserializes too. So start_node_migration applies the bound explicitly instead.


It is safe (and preferred) to delete entries that exceed the expected length (should be none, but you never know). Otherwise, we might end up with inconsistent state, since a node migration could be started before the contract updates.

crates/near-mpc-contract-interface/src/types/participants.rs
barakeinav1: Done in 1311f05 - cut to one line.

crates/near-mpc-contract-interface/src/types/participants.rs
barakeinav1: Done in 1311f05

crates/near-mpc-bounded-collections/src/bounded_string.rs
barakeinav1: Done in 1311f05 - kept the first sentence. moved the persisted-field caveat to #4456 so it is somewhere actionable rather than just sitting in the doc.

crates/near-mpc-bounded-collections/src/bounded_string.rs
barakeinav1: Added in 1311f05, and used it at the call sites.

crates/near-mpc-bounded-collections/src/bounded_string.rs
barakeinav1: agree, dropped it in 1311f05. `From<BoundedString<U>> for String` is the one way to the owned value now.

crates/near-mpc-bounded-collections/src/bounded_string.rs
barakeinav1: i went the other way on this one - kept `Deref<Target = str>` and dropped `as_str()` instead, so there is still only one way.

reason: (suggested by claude, and I don't have a strong option here) with `Deref` you can pass `&bounded` anywhere a `&str` is wanted and the coercion happens for you, no `.into()` needed at the call site. a `From` impl would need the explicit conversion every time. `BoundedVec` in this crate already derefs too, so it matches.

crates/near-mpc-bounded-collections/src/bounded_string.rs
barakeinav1: >Hmm, how do we want people to use this? Do we want them to type *bounded_string or bounded_string.into()?

neither - `&bounded_string` where a `&str` is expected, and the coercion does it. `*bounded_string` should not be needed anywhere.

>I would prefer not to have too many different ways of expressing the same thing.

so after 1311f05 it is one way per direction: `Deref` for `&str`, `From<_> for String` for the owned value. dropped `as_str()` and `into_inner()`, which were the duplicates.

crates/contract/tests/sandbox/common.rs
barakeinav1: nit taken in 1311f05, and applied to the other fixtures too now that `TryFrom<&str>` exists.

crates/near-mpc-contract-interface/src/types/participants.rs
barakeinav1: agree, but I preferer not to do this in this PR.
Also @netrome mention in slack that we what to loss teh http/s prefix, etc.. - so I think a follow PR for both is a better place.

crates/contract/src/v3_15_1_state.rs
barakeinav1: i think failing the migration is better. i do not expect it to fail, but if it does we want to know - we can check which operator set a long url and get it fixed straight away. we have a release coming very soon so that is the window i would rather be loud in.

pointing at an invalid url instead would make that participant unreachable with nobody told, and move the failure from the upgrade into runtime where it is harder to attribute.

kevindeforth: > pointing at an invalid url instead would make that participant unreachable with nobody told, and move the failure from the upgrade into runtime where it is harder to attribute.

Hmm, it would still be visible, because that peer would get disconnected and we would see it on grafana.
It's also an easy grep on the contract state post migration.

But I am okay failing the migration here.

barakeinav1:
> Some questions wrt the API of the bounded string type and the no-node-upgrade policy.
> 
> > Node-first upgrades. The node parses state() into dtos::ProtocolContractState. A bounded DTO makes a new node stricter than the contract it is talking to — while 3.15.1 is live (no url validation) one long url would stop every upgraded node from reading state.
> 
> We could handle this. We try to deserialize the new dto types. If unsuccessful, we deserialize in the old types. Or alternatively, if the url doesn't serialize properly, we default to an invalid one, like we already do here:
> 
> https://github.com/near/mpc/blob/7fc4663711fa24a41421f5e0db6be82ad3b4b5bb/crates/node/src/indexer/participants.rs#L418
> 
> > ongoing_migrations. It is an IterableMap<AccountId, DestinationNodeInfo> holding the interface type, and migrate() doesn't touch lazy maps. A record written before the upgrade would never be validated, and afterwards could not be read, concluded, or even removed — remove returns the value, so it deserializes too. So start_node_migration applies the bound explicitly instead.
> 
> It is safe (and preferred) to delete entries that exceed the expected length (should be none, but you never know). Otherwise, we might end up with inconsistent state, since a node migration could be started before the contract updates.



agree, but not for this PR.  i'll add both to #4456.

the way i see it this PR already closes the vulnerability - the url is bounded everywhere it can be set. the rest is optimization, so the follow up is the right place for it.

one correction on the sentinel: `UNREACHABLE_PARTICIPANT_ADDRESS` runs in `convert_participant_infos` on an already-deserialized `String`. if `ParticipantUrl`'s `Deserialize` rejects the value, the whole `ProtocolContractState` read fails and we never get there. so it needs your first option (try the new dto, fall back to the old), or splitting the dto into input and state types.

on `ongoing_migrations` - agree in principle, but i would make it conditional. after this PR `conclude_node_migration` fails on a long url, but `cancel_node_migration` still works (it only deserializes the unbounded dto), so the operator can cancel and restart. so: check `migration_info()` before deploying #4456, and add the prune only if there is actually a stale long entry then. empty on both networks today.

crates/near-mpc-bounded-collections/src/bounded_string.rs
```diff
@@ -0,0 +1,199 @@
...
+        impl<const U: usize> JsonSchema for BoundedString<U> {
+            fn schema_name() -> String {
+                format!("BoundedString_{U}")
+            }
+
+            fn json_schema(generator: &mut SchemaGenerator) -> Schema {
+                let mut schema = String::json_schema(generator).into_object();
+                schema.string().max_length = u32::try_from(U).ok();
+                Schema::Object(schema)
+            }
```
haiyuechen-nearone, line 132: nit: schemars' `SchemaObject.string().max_length` is counted in chars not bytes, so the definition here differs with our actual byte count length bound definition. Personally I think its better to not set a bound, if we can't properly express it.

```suggestion
             fn json_schema(generator: &mut SchemaGenerator) -> Schema {
+                // `SchemaObject.string().max_length` counts chars, not bytes, so it cannot correctly express our byte denoted bound. 
+                //Describe it as a plain string.
+                String::json_schema(generator)
             }
```

haiyuechen-nearone (dismissed):
I am still not very familiar with the migration steps, so this review was done mostly around the BoundedString type itself. I think the implementation is reasonable.

Left a nit around the limitation of json schema definition. I feel its better to not define a bound than define inaccurate bounds, but its not a critical part of the implementation so happy to approve it.

kevindeforth (dismissed):
Thanks for the updates and the fix!

> one correction on the sentinel: UNREACHABLE_PARTICIPANT_ADDRESS runs in convert_participant_infos on an already-deserialized String. if ParticipantUrl's Deserialize rejects the value, the whole ProtocolContractState read fails and we never get there. so it needs your first option (try the new dto, fall back to the old), or splitting the dto into input and state types.

Yes, that's the complexity that we would need to handle, but it's totally doable.

I expect there to be some code-churn and additional review effort due to the multi-PR strategy, but I won't block this PR based on that.

crates/near-mpc-bounded-collections/src/bounded_string.rs
barakeinav1: good catch, fixed in b07d3ed - agree it is better to not state a bound we cannot express. json schema `maxLength` is characters and ours is bytes, so a 256-char multibyte string would pass the schema and then get rejected on deserialization. emits a plain string schema now.

haiyuechen-nearone (approved):

kevindeforth (approved):

PR #4474, opened 2026-09-18
fix(test): Retain more blocks in sandbox tests to fix GC race condition

### LINKED ISSUE #4461: respond__should_drain_saturated_fan_out_queue is even more flaky after nearcore 2.14.0-rc.1 bump
### Background

`respond__should_drain_saturated_fan_out_queue` is currently flaky, although in practice it seems to succeed mostly. The test submits 65 batch transactions, fast-forwards 2 blocks between each which requires 130 blocks. Additionally, since block production isn't paused, running this on a typical machine takes around 40s which adds another 330 blocks.

The problem is that the sandbox typically only retain 245 blocks (`gc_num_epochs_to_keep = 5` 60 blocks per epoch but measured to ~245 in practice). However, historically the GC has been slow enough for this test to pass, but since the 2.14.0-rc.2 version bump https://github.com/near/mpc/pull/4452 it seems like this is failing more frequently locally. CI contention still seems to make it pass in GH actions.

To ensure this test succeeds more reliably we should ensure the sandbox retains enough blocks. Ideally, the sandbox shouldn't produce blocks if not instructed to but that's not supported yet afaik.

### User Story

As a developer I want reliable tests.

### Acceptance Criteria

`respond__should_drain_saturated_fan_out_queue` passes reliably.

_No response_

netrome:
closes #4461 

This PR changes our sandbox tests to have them explicitly start a sandbox and attach near-workspaces to it, rather than having it spawned directly through near-workspaces. This allows us to configure the sandbox to retain more epochs which we need for a few tests.


netrome:
> **PR title type suggestion:** This PR changes only test files and test utilities, so the type prefix should be `test:` instead of `fix:`.
> 
> Suggested title: `test: Retain more blocks in sandbox tests to fix GC race condition`

Changed to "fix(test)". Just "test:" would imply this is a test change, while this primarily is a fix imo.

anodar (approved):
LGTM. I think we could make worker/contract in `SandboxTestSetup` private instead and define accessor methods, that way it's less error prone and worker isn't accidentally dropped if one forgets to destructure it even if it's not used (like it's here using _worker)

netrome:
> LGTM. I think we could make worker/contract in SandboxTestSetup private instead and define accessor methods, that way it's less error prone and worker isn't accidentally dropped if one forgets to destructure it even if it's not used (like it's here using _worker)

I like the idea. Had Claude implement it but the diff is around 500 LoC so let's save it as a follow-up.

haiyuechen-nearone (approved):
LGTM

However I think its better to expose a config method in near-workspaces so we don't need to juggle this in our code. Lets try to push for that in the follow up.

netrome:
> LGTM
> 
> However I think its better to expose a config method in near-workspaces so we don't need to juggle this in our code. Lets try to push for that in the follow up.

Yeah that would be much better.

netrome:
Created https://github.com/near/mpc/issues/4477 as a follow-up

PR #4480, opened 2026-09-18
fix: exclude offline participants from foreign chain tx computations

### LINKED ISSUE #4479: bug: foreign chain txs don't filter participant set for aliveness
### Background

tldr: Ecdsa and eddsa presignature stores update the set of available participants when calling `take_owned`. Foreign chain transactions exclusively call `take_owned_matching`, which does not update the set of available participants.

Hence, leaders of foreign chain tx verification requests select offline participants for their computation, which results in timed out requests.

----
_(slightly edited LLM explanation)_

The verify-foreign-tx leader selects a presignature via [`take_owned_matching(chain_supporters)`](https://github.com/near/mpc/blob/3c2a608fa8a270451bd25a83ec74ff5571a53d7c/crates/node/src/providers/verify_foreign_tx/sign.rs#L131), where regular signing uses [`take_owned()`](https://github.com/near/mpc/blob/3c2a608fa8a270451bd25a83ec74ff5571a53d7c/crates/node/src/providers/ecdsa/sign.rs#L74). Everything after the take is identical between the two paths.

`take_owned` refreshes the alive set unconditionally before selecting, with a comment explaining why:

https://github.com/near/mpc/blob/3c2a608fa8a270451bd25a83ec74ff5571a53d7c/crates/node/src/assets.rs#L265-L271

`take_owned_matching` does not, and its hot-queue fast path returns on the first match before anything refreshes:

https://github.com/near/mpc/blob/3c2a608fa8a270451bd25a83ec74ff5571a53d7c/crates/node/src/assets.rs#L315-L333

[`satisfies_condition`](https://github.com/near/mpc/blob/3c2a608fa8a270451bd25a83ec74ff5571a53d7c/crates/node/src/assets.rs#L203-L205) is a correct conjunction of alive-set and supporters, but it reads `last_condition_value`, which only changes when something calls [`update_condition_value`](https://github.com/near/mpc/blob/3c2a608fa8a270451bd25a83ec74ff5571a53d7c/crates/node/src/assets.rs#L77-L85).

`CONDITION_REFRESH_INTERVAL` is set inside that function and rate-limits refreshes; it does not
schedule them. Nothing polls.

Because we have a dedicated domain (and thus, a dedicated presignature store) for foreign transaction verification requests, only calls to 
[`take_first_matching`](https://github.com/near/mpc/blob/3c2a608fa8a270451bd25a83ec74ff5571a53d7c/crates/node/src/assets.rs#L187-L188) (the cold-queue fallback, skipped whenever the hot queue has a match) and [`maybe_discard_owned`](https://github.com/near/mpc/blob/3c2a608fa8a270451bd25a83ec74ff5571a53d7c/crates/node/src/providers/ecdsa/presign.rs#L140-L143) trigger a condition refresh.

In production, this led to the following obsevation:

Participant 10 went offline 2026-09-17T20:42Z. At 21:14Z, with participant 19 logging
`Connection refused` for it roughly once a second, a `verify_foreign_transaction` request
arrived and its leader task failed ten times in eleven seconds:

```
21:14:01.868  INFO  indexed new `verify_foreign_transaction` function call … 4Hiuue…
21:14:02.537  ERROR task failed; description: leader for verify_foreign_tx request 4Hiuue…;
              error msg: Cannot send MPC message to recipient 10
   … retried ~1/sec, same request, ten times …
21:14:12.816  ERROR … Cannot send MPC message to recipient 10
```

Regular signing on the same node held ~99.7% success across the identical window. Those ten
retries were enough to trip the foreign-tx alert and page on-call for a single request.

### User Story

As a user, I like when my foreign chain verification transaction requests succeed.

### Acceptance Criteria

A leader excludes offline participants from Foreign chain transaction computations

_No response_

kevindeforth:
resolves #4479

Unfortunately, this is really hard to test in an e2e test.


netrome (approved):
Thank you! 🙏

anodar (approved):
Thank you!  🙏

PR #4483, opened 2026-09-21
build: minor rocksdb bump related fixes

### LINKED ISSUE #4482: misc fixes after rocksdb upgrade
## Background

`.cargo/config.toml:10` pins `-C target-cpu=x86-64-v3` for rustc, but the matching
C/C++ flags (`flake.nix:58`, `-march=x86-64-v3 -mpclmul -maes`) are exported only by
the devshell and `nix/mpc-node.nix`. Outside `nix develop`, cargo gets the Rust half
of the ISA baseline and none of the C half.

Since #4452 bumped `librocksdb-sys` to `0.17.3+10.4.2` that is a build failure:
`build.rs` emits `-mavx2` but not `-mpclmul` (v3 has `avx2`, not `pclmulqdq`), rocksdb
10.x's `port/lang.h` infers `#define __PCLMUL__ 1` from `__AVX__`, and `crc32c.cc:429`
then compiles the PCLMUL CRC32C path with the ISA off:

```
wmmintrin.h:105:10: error: '__builtin_ia32_pclmulqdq128' needs isa option -mpclmul -msse2
error: failed to run custom build command for `librocksdb-sys v0.17.3+10.4.2`
```

CI is blind to it because every cargo call goes through `nix develop --command`
(`.github/workflows/ci.yml:164`). `README.md:59-64` and `AGENTS.md:10` document a bare
`cargo build -p mpc-node --release` without saying it needs the devshell.

Also in scope: the ISA baseline is duplicated across `.cargo/config.toml`,
`flake.nix:58` and `nix/mpc-node.nix:226-227` with nothing enforcing agreement, and
`-include cstdint` (`flake.nix:106`, `flake.nix:142`, `nix/mpc-node.nix:224`) is
obsolete since the rocksdb bump.

## User Story

As a contributor building outside the Nix devshell, I need
`cargo build -p mpc-node --release` to work, or to fail with a message naming the
missing flags, instead of dying inside rocksdb's C++.

## Acceptance Criteria

- [ ] `cargo build -p mpc-node --locked --release` works outside `nix develop`, not
      just inside it.

Suggested fix, an `[env]` table in `.cargo/config.toml` so the plain cargo path works
while the devshell and derivation keep precedence:

```toml
[env]
CFLAGS_x86_64_unknown_linux_gnu = { value = "-march=x86-64-v3 -mpclmul -maes", force = false }
CXXFLAGS_x86_64_unknown_linux_gnu = { value = "-march=x86-64-v3 -mpclmul -maes", force = false }
```

The table cannot be `cfg`-scoped, but the target-suffixed names keep non-x86 hosts and
the wasm32 contract build unaffected.

Optionally also add `-C target-feature=+pclmulqdq,+aes` to the rustflags, which fixes
rocksdb on its own (`build.rs` keys `-mpclmul` off the `pclmulqdq` rust feature) and
keeps rustc and C agreeing, though it does not help cc deps that read only `CFLAGS`.

Workaround meanwhile:

```bash
export CFLAGS_x86_64_unknown_linux_gnu="-march=x86-64-v3 -mpclmul -maes"
export CXXFLAGS_x86_64_unknown_linux_gnu="-march=x86-64-v3 -mpclmul -maes"
```

`flake.nix:48-52` already explains that v3 excludes PCLMUL and AES. The bug is that
this only holds on the Nix side.

gilcu3:
Closes #4482


nix/cargo-near.nix
```diff
@@ -29,18 +29,18 @@ rustPlatform.buildRustPackage rec {
   buildInputs = [
     openssl
   ]
-  ++ lib.optionals stdenv.isLinux [
+  ++ lib.optionals stdenv.hostPlatform.isLinux [
     udev
   ]
-  ++ lib.optionals stdenv.isDarwin [
+  ++ lib.optionals stdenv.hostPlatform.isDarwin [
```
gilcu3, line 35: these are broken-window fixes to silence a warning

.cargo/config.toml
```diff
@@ -8,3 +8,12 @@
 # explicitly there. Keep both in sync.
 [target.'cfg(target_arch = "x86_64")']
 rustflags = ["-C", "target-cpu=x86-64-v3"]
+
+# The C/C++ half of that baseline, for the cc-crate dependencies (rocksdb,
```
netrome: What baseline? This comment looks a bit confusing.

netrome (dismissed):
Once confusing comment, otherwise lgtm

.cargo/config.toml
gilcu3: The baseline refers to `target-cpu=x86-64-v3` which we consider the baseline supported CPU. The comment is a continuation of the one above

gilcu3: addressed in [d5a1219](https://github.com/near/mpc/pull/4483/commits/d5a1219d774e1388d3a25fef5d6294738956d05a)

netrome (approved):
Thanks!

.cargo/config.toml
netrome: Okay thanks for explaining :pray:

pbeza (approved):

Issue #4340, opened 2026-09-03
Predictable channel ids let one participant block every other leader


gilcu3:
### Background

[`ChannelId`](https://github.com/near/mpc/blob/d8f8fe4becb5ab2188199a1b5d6e69b42c952c45/crates/node/src/primitives.rs#L30-L35) wraps `UniqueId` = `participant_id << 96 | unix_seconds << 32 | counter`, with the [counter reset each second](https://github.com/near/mpc/blob/d8f8fe4becb5ab2188199a1b5d6e69b42c952c45/crates/node/src/primitives.rs#L66-L76), so it is predictable. [`sender_for`](https://github.com/near/mpc/blob/d8f8fe4becb5ab2188199a1b5d6e69b42c952c45/crates/node/src/network.rs#L264-L344) never checks the channel id's embedded participant against the TLS-authenticated originator.

Participant M can then send `Start { channel_id = (L, t, 0..k), participants = <full set>, task_id = <an id M owns> }` to every node for each upcoming second `t`. On L, `new_channel_for_task` finds the entry occupied and fails with "Channel already exists". On a follower, L's genuine `Start` lands in M's phantom channel and [dies as an unexpected `Start`](https://github.com/near/mpc/blob/d8f8fe4becb5ab2188199a1b5d6e69b42c952c45/crates/node/src/network.rs#L828-L832). One participant, never selected for any computation, thereby stops every other node from leading signatures, presignatures and triples.

### Acceptance Criteria

`sender_for` rejects a `Start` whose channel id owner is not the originator

PR #4488, opened 2026-09-21
fix(node): remove a single node's ability to shutdown independent computations

### LINKED ISSUE #4342: A message from outside a channel's participant set aborts the whole computation
### Background

[`receive_one`](https://github.com/near/mpc/blob/d8f8fe4becb5ab2188199a1b5d6e69b42c952c45/crates/node/src/network.rs#L775-L791) already warns with the sender and message kind ("Rejecting {kind} from participant {from} (channel {:?}): not in participant set"), but it then also returns `ParticipantNotInChannelError`. [`run_protocol` propagates it](https://github.com/near/mpc/blob/d8f8fe4becb5ab2188199a1b5d6e69b42c952c45/crates/node/src/protocol.rs#L154) and [`communicate_failure`](https://github.com/near/mpc/blob/d8f8fe4becb5ab2188199a1b5d6e69b42c952c45/crates/node/src/network.rs#L674-L714) sends `Abort` to every other party.

A participant left out of a signing set, for example because it was judged slow, can therefore kill the computation with a single message and burn the presignature, which is what robust ECDSA is meant to prevent. The log is the right response on its own; the message should be dropped rather than turned into a channel error. #3651 introduced the current behavior to stop a panic, and [`run_protocol__should_reject_computation_from_non_participant`](https://github.com/near/mpc/blob/d8f8fe4becb5ab2188199a1b5d6e69b42c952c45/crates/node/src/protocol.rs#L347) codifies it, so that test needs updating too.

### Acceptance Criteria

- A message whose sender is outside the channel's participant set does not abort the ongoing computation.
- The same for messages buffered before `Start`: a non-member's message is discarded when the channel is materialized, rather than aborting it.

gilcu3:
Closes #4342


gilcu3:
> * `crates/node/src/network.rs:281` — residual hole of the same class, worth a follow-up issue rather than a change in this PR. `sender_for` never checks that the `ChannelId` belongs to the `originator` of the `Start` message. `ChannelId` wraps a `UniqueId` whose high bits are the generating participant's id (`primitives.rs:45`) and whose remaining bits are a wall-clock second plus a small counter, so ids are guessable. Any node can therefore send a `Start` for a channel id a legitimate leader is about to use, with `participants = [us, attacker]` — `validate_start_participants` accepts that — and take the `Vacant` branch first. When the real leader's `Start` arrives the entry is `Occupied` and no longer in `channels_waiting_for_start`, so it is routed into the squatted channel and (after this PR) silently dropped, denying that computation on our node. `UniqueId::validate_owned_by` already exists for exactly this shape of check and is used for presignature ids at `providers/ecdsa/sign.rs:94`; applying it to `channel_id` against `originator` would close it.

Will be solved in #4340

netrome (approved):
Thank you for fixing this 🙏

pbeza (approved):

PR #4490, opened 2026-09-21
fix(contract): reject duplicate TLS keys during migrations

### LINKED ISSUE #4252: feat(contract): enforce TLS key uniqueness during migrations
### Background

Currently, an existing node operator can perform a node migration to a new node with the same TLS key as an existing participant, opening the door to some griefing attacks and other potential issues as we assume TLS keys to be unique in several places in the contract. We should not allow duplicate TLS keys.

### User Story

As a node operator I would not like another operator to claim that they have my TLS key.

### Acceptance Criteria

Operators may not claim existing TLS keys in the contract.

_No response_

gilcu3:
Closes #4252 

Most code changes are new tests


crates/contract/src/api/node_migration.rs
```diff
@@ -185,6 +194,8 @@ impl MpcContract {
     /// - [`InvalidParameters::InvalidTeeRemoteAttestation`]: if destination node’s TEE quote is invalid
     /// - [`InvalidCandidateSet::ParticipantUrlTooLong`](crate::errors::InvalidCandidateSet::ParticipantUrlTooLong): if the destination node's url exceeds
     ///   [`MAX_PARTICIPANT_URL_BYTES`](crate::primitives::participants::MAX_PARTICIPANT_URL_BYTES).
+    /// - [`NodeMigrationError::TlsKeyAlreadyClaimed`](crate::errors::NodeMigrationError::TlsKeyAlreadyClaimed): if the destination node's TLS
+    ///   public key belongs to another account
     #[handle_result]
     pub fn conclude_node_migration(&mut self, keyset: dtos::Keyset) -> Result<(), Error> {
```
gilcu3, line 209: not sure why we are checking many things twice: in `start_node_migration` and `conclude_node_migration`, for example `InvalidCandidateSet::ParticipantUrlTooLong`. Followed the same pattern in this PR, although this seems redundant

barakeinav1: could things (state,etc..) change between start/conclude?

gilcu3: some things could change, but I am not aware of any that would be relevant for these checks. If those existed then the redundancy would be needed ofc

barakeinav1:
Not introduced by this PR, but it interacts with it: the foreign-chain cleanup in `conclude_node_migration` removes `old_tls_key` unconditionally (`crates/contract/src/api/node_migration.rs:297-301`), so a participant that migrates while keeping the same TLS key ends up deleting the config for the key it still uses, and `recompute_available_foreign_chains()` then drops it from every chain's support count. Looks like it just needs to skip the removal when the key is unchanged.

crates/contract/src/api/node_migration.rs
```diff
@@ -309,6 +331,32 @@ impl MpcContract {
...
+    tls_public_key: &dtos::Ed25519PublicKey,
+) -> Result<(), Error> {
+    let claimants = [
+        participants.account_with_tls_key(tls_public_key).cloned(),
+        tee_state.attestation_owner(tls_public_key),
+    ];
+    let Some(claimant) = claimants
+        .into_iter()
+        .flatten()
+        .find(|claimant| claimant != account_id)
```
barakeinav1, line 387: Is it intentional that a participant can migrate to a destination with its own current TLS key? `start_node_migration__should_accept_the_callers_own_tls_key` reads like yes, just want to confirm — if so, it runs into https://github.com/near/mpc/pull/4490#issuecomment-5774154508.

crates/contract/src/api/node_migration.rs
```diff
@@ -1010,6 +1058,177 @@ mod tests {
         }
     }
 
+    #[test]
+    fn start_node_migration__should_reject_a_tls_key_held_by_another_participant() {
```
barakeinav1, line 1100: Nit: the new tests here are missing the `// Given` / `// When` / `// Then` markers, while the ones added in `participants.rs` and `tee_state.rs` have them.

crates/contract/src/primitives/participants.rs
```diff
@@ -104,6 +107,9 @@ impl Participants {
         if accounts.len() != self.len() {
             return Err(InvalidCandidateSet::DuplicateAccountIds.into());
         }
+        if tls_public_keys.len() != self.len() {
+            return Err(InvalidCandidateSet::DuplicateTlsPublicKeys.into());
```
barakeinav1, line 111: I checked current state: mainnet (`v1.signer`) and testnet (`v1.signer-prod.testnet`) both have 17 participants with 17 distinct TLS keys, so no duplicates today. Worth re-checking right before the release goes out though — this is now also reached from `validate_incoming_proposal`, so if a duplicate shows up in the meantime, any `vote_new_parameters` proposal carrying it forward would be rejected and the network could not reshare until it is cleaned up.

gilcu3: Yeah, I checked that we had at least one simple escape hatch for the unlikely case that this breaks a future proposal

barakeinav1:
If we do want to support migrating with an unchanged TLS key (https://github.com/near/mpc/pull/4490#discussion_r4070184134), it deserves a test on the conclude side too — `start_node_migration__should_accept_the_callers_own_tls_key` only covers the start half, and conclude is where the foreign-chains problem shows up. Separately, a TLS-key analogue of `test_proposal_non_unique_ids` would cover the new `validate` rule through `validate_incoming_proposal`.

barakeinav1 (changes requested):
Blocking on https://github.com/near/mpc/pull/4490#issuecomment-5774154508: a participant that migrates while keeping the same TLS key loses its foreign-chain config, because the cleanup removes `old_tls_key` unconditionally. The rest of my comments are non-blocking.

crates/contract/src/primitives/participants.rs
barakeinav1: we will need to vote_parameters and remove the duplicate key if this happens, seems reasonable.

barakeinav1 (dismissed):
Thanks for addressing everything. Rejecting the unchanged TLS key outright is a cleaner fix than the conditional removal.

haiyuechen-nearone (approved):
LGTM 🚢

crates/e2e-tests/tests/migration_endpoint.rs
```diff
@@ -74,31 +79,25 @@ async fn migration_endpoint__should_track_migration_state() {
             .await
             .expect("endpoint state mismatch after backup registration");
 
-        // When: start node migration with a bogus destination.
+        let destination_tls_public_key =
```
barakeinav1, line 82: Nit: the `// When: start node migration with a bogus destination.` marker got dropped here, so the block below has a `// Then:` with no `// When:`.

barakeinav1 (approved):
LGTM

PR #4499, opened 2026-09-22
build(nix): build all Docker images hermetically

### LINKED ISSUE #4484: Add support for hermetic builds
### Background

We want to be able to build our system in a hermetic environment (e.g. all dependencies pinned, every tool required for the build explicitly declared, no network interactions during the build process).

While we currently support reproducible builds with "repro-env", our builds still fetch deb packages over the network, and nothing prevents us from having build scripts that makes network requests during the build. This has already caused reproducibility to break accidentally in the past.

A completely hermetic build environment would provide stronger guarantees of reproducibility and stronger protection against some supply chain attacks.

### User Story

As an MPC operator, I want strong guarantees that the node hashes I vote from can be reliably reproduced and have been built in a controlled environment.

### Acceptance Criteria

We support completely hermetic builds for our node and contract.

I believe we already have Nix based hermetic builds for our contract.

pbeza:
All three images (`mpc-node`, `mpc-node-gcp`, `mpc-launcher`) are now plain flake outputs. `nix build .#packages.x86_64-linux.<image>-image` compiles the binary, assembles the image and compresses its layers, all inside the Nix sandbox. The result is the exact layout pushed to Docker Hub, so `sha256sum result/manifest.json` is the digest operators vote on, and CI publishes it byte-for-byte with `skopeo copy --preserve-digests`. `docs/guide/reproducible-builds.md` covers Linux and macOS.

**Why this is hermetic**
- Every step that runs build code (rustc, `build.rs`, proc-macros, C/C++ compilers, image assembly, layer compression) runs in the Nix sandbox: no network, only declared inputs, fixed paths and timestamps.
- Only fetching the pinned inputs touches the network (`Cargo.lock` checksums, `flake.lock` hashes, git revisions), so a changed byte fails the build.
- CI sets `sandbox-fallback = false`: a build that cannot be sandboxed fails instead of silently running without the sandbox.
- Before: repro-env ran `cargo build` with [`--network=host`](https://github.com/kpcyrd/repro-env/blob/afbe30e442767a1d9159130147db5164fb7ddfd5/src/container.rs#L177), rustup downloaded Rust 1.97.1 mid-build (the container shipped 1.86), and the Dockerfiles ran `apt-get` against snapshot.debian.org.

**Why files were deleted**
Each one only served the Docker-based path, and nothing references them anymore.

| Deleted | Replaced by |
|---|---|
| `deployment/build-images.sh` | `nix build` plus one `skopeo copy` in the push workflow |
| `repro-env.toml`, `repro-env.lock`, the repro-env cache action | the node's crane derivation: `mpc-node.override { pname = "tee-launcher"; }` |
| `Dockerfile-rust-launcher`, `repro-sources-list.sh`, BuildKit pin, timestamp reset | `dockerTools.buildImage` with the Docker CLI from nixpkgs |
| podman and the pinned quay skopeo image | skopeo from the pinned nixpkgs, inside the sandbox |
| apparmor userns sysctl, `apt install repro-env liblzma-dev` | nothing: only rootless podman and repro-env needed them |
| three image push workflows | one job, which builds `mpc-node` once instead of twice |

**Verified**
- CI (native x86_64 Linux, Determinate Nix) and an Apple Silicon Mac (Rosetta, upstream Nix, sandbox on, built from `github:near/mpc/<sha>`) produced identical digests for all three images from CI's merge commit `412de2f` (`81651fa0…`, `455d6cd9…`, `52775e7c…`).
- The launcher digest is unchanged after an unrelated source change and after a `Cargo.lock`-only change. `buildImage` keeps store paths out of the image config; `buildLayeredImage` would not.
- A build step that opens a socket inside the sandbox fails with `Network is unreachable`.
- CI checks that the node binary reports the build commit, that the gcp image runs `gcloud`, and that the node and launcher images start.

**Worth knowing**
- All digests change once, and the first Nix launcher needs the usual launcher vote. Bumping nixpkgs changes digests, since image contents come from it.
- Releases from before this switch are reproduced from their own tag. The contract release keeps cargo-near (NEP-330).

Refs #4484


pbeza:
@claude review

pbeza:
@claude review

pbeza:
@claude review

pbeza:
@claude review

.github/workflows/ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
     name: "Build MPC Node TEE Docker image"
     needs: changes
     if: ${{ !cancelled() && needs.changes.outputs.code_changed != 'false' }}
-    runs-on: warp-ubuntu-2604-x64-16x
+    runs-on: warp-ubuntu-2604-x64-32x
```
pbeza: JFYI: The deleted `nix-build-mpc-node.yml` job built the node with Nix on a 32-core runner. Its build moved into the image job, which then took over the runner size.

.github/actions/cache-repro-env-pkgs/action.yml
pbeza: JFYI why `cache-repro-env-pkgs` was deleted:

the action only cached the Debian packages `repro-env` downloaded from `snapshot.debian.org` (#4232 added it because those downloads were throttled). The launcher was the image built with `repro-env`; it now compiles in Nix, whose inputs come from `cache.nixos.org` and hash-pinned crate fetches, so there are no Debian packages left to cache, and the action's key file `repro-env.lock` is gone too. The Nix jobs already use `restore-nix-cache`.

repro-env.lock
pbeza: JFYI, the rationale for why it was removed is [here](https://github.com/near/mpc/pull/4499/changes#r4095451060).

repro-env.toml
pbeza: JFYI why `repro-env.toml` and `repro-env.lock` were deleted:

they described `repro-env`'s build container for the launcher: a `rust:1.86` base image plus the pinned Debian packages (`clang`, `make`, `libssl-dev`) installed into it. Nothing else used them. The launcher now builds through the same crane derivation as the node, so the toolchain comes from `rust-toolchain.toml` (`1.97.1`, no more `rustup` download mid-build) and every library from `flake.lock`, and the build runs in the Nix sandbox without network access, which `repro-env`'s container never did.

.github/workflows/docker_build_node_gcp.yml
pbeza: JFYI, why `docker_build_node_gcp.yml` and `docker_build_rust_launcher.yml` were deleted:

together with `docker_build_node.yml`, they were three near-identical copies of the same workflow. Each ran `build-images.sh` for a single image on pushes to `main` and `release/v*`, and both the node and GCP workflows compiled `mpc-node` from scratch.

`docker_build_node.yml` was renamed to `docker_build_images.yml` and now builds all three images in a single job, compiling `mpc-node` only once. It keeps the same triggers, `production` environment, Docker Hub login, and `<branch>-<sha7>` tags, so `release.yml` and the retag workflows are unaffected.

.github/workflows/docker_build_rust_launcher.yml
pbeza: JFYI: see [here](https://github.com/near/mpc/pull/4499/changes#r4095599237) for the rationale behind deleting `.github/workflows/docker_build_rust_launcher.yml`.

.github/workflows/nix-build-mpc-node.yml
pbeza: JFYI on why `nix-build-mpc-node.yml` was deleted:

it built `mpc-node` with Nix only to print the binary hash as a side check:

https://github.com/near/mpc/blob/b1220be83d388de9df8becef56d4f911bc92ffcb/.github/workflows/ci.yml#L59-L63

while the released images still came from `repro-env`; nothing consumed its output. The `ci.yml` image job now builds the same `mpc-node` package as part of the image build, so keeping the job would compile it a second time on another 32-core runner without giving us any extra signal. The runner size moved to the image job along with the build.

deployment/build-images.sh
pbeza: JFYI on why `deployment/build-images.sh` was deleted:

it glued the old pipeline together: tool checks, timestamp resets, a pinned BuildKit builder, repro-env, buildx, `docker load`, the podman-run skopeo digest step and the push. Every step now happens inside `nix build` (compile, assemble, compress; the output already carries its digest) or is no longer needed.

Where it was used on `main`, and what replaces each use:

| Used by | Replaced by |
|---|---|
| CI image jobs ([ci.yml#L67](https://github.com/near/mpc/blob/130aaf97/.github/workflows/ci.yml#L67), [#L103](https://github.com/near/mpc/blob/130aaf97/.github/workflows/ci.yml#L103)) | `nix build` plus `sha256sum` ([ci.yml#L59-L63](https://github.com/near/mpc/blob/b1220be83d388de9df8becef56d4f911bc92ffcb/.github/workflows/ci.yml#L59-L63), [#L96-L100](https://github.com/near/mpc/blob/b1220be83d388de9df8becef56d4f911bc92ffcb/.github/workflows/ci.yml#L96-L100)) |
| The three publish workflows, one image each with `--push` | Seven lines in `docker_build_images.yml` that build and push all three images with the unchanged `<branch>-<sha7>` tags ([#L40-L47](https://github.com/near/mpc/blob/b1220be83d388de9df8becef56d4f911bc92ffcb/.github/workflows/docker_build_images.yml#L40-L47)) |
| Operator and release docs (`reproducible-builds.md`, the TDX guide, `RELEASES.md`, `README.md`) | Two commands, the same on Linux and macOS ([reproducible-builds.md#L21-L24](https://github.com/near/mpc/blob/b1220be83d388de9df8becef56d4f911bc92ffcb/docs/guide/reproducible-builds.md#L21-L24)) |

Keeping a script around those commands would only add a layer that depends on the host's shell tools; the old one already failed on macOS, whose `mktemp` has no `--suffix`.

deployment/Dockerfile-node
pbeza: JFYI on why `deployment/Dockerfile-node`, `Dockerfile-node-gcp`, `Dockerfile-rust-launcher` and `repro-sources-list.sh` were deleted:

the Dockerfiles assembled each image from a Debian base ([debian:bookworm-slim](https://github.com/near/mpc/blob/130aaf97/deployment/Dockerfile-node#L1), [google/cloud-sdk](https://github.com/near/mpc/blob/130aaf97/deployment/Dockerfile-node-gcp#L1), [debian:trixie-slim](https://github.com/near/mpc/blob/130aaf97/deployment/Dockerfile-rust-launcher#L1)) plus `apt-get install` at build time, and copied in a binary compiled earlier by `repro-env`. `repro-sources-list.sh` was a vendored helper that pointed `apt` at a fixed `snapshot.debian.org` date so those installs were repeatable. `docker buildx` ran them with a pinned BuildKit.

They are replaced by two Nix expressions that assemble the images from nixpkgs packages inside the sandbox, with no base image, no `apt` and no network:

| Deleted | Replaced by |
|---|---|
| `Dockerfile-node`, `Dockerfile-node-gcp` | [`nix/mpc-node-image.nix`](https://github.com/near/mpc/blob/b1220be83d388de9df8becef56d4f911bc92ffcb/nix/mpc-node-image.nix); the gcp image is the same file with `withGcloud = true` ([flake.nix#L76-L77](https://github.com/near/mpc/blob/b1220be83d388de9df8becef56d4f911bc92ffcb/flake.nix#L76-L77)) |
| `Dockerfile-rust-launcher` | [`nix/mpc-launcher-image.nix`](https://github.com/near/mpc/blob/b1220be83d388de9df8becef56d4f911bc92ffcb/nix/mpc-launcher-image.nix), with the Docker CLI and Compose plugin from nixpkgs instead of Debian's `docker-cli` and `docker-compose` |
| `repro-sources-list.sh` | nothing: package versions are pinned by `flake.lock`, so there is no `apt` snapshot date to fix |

deployment/Dockerfile-node-gcp
pbeza: JFYI: see [this](https://github.com/near/mpc/pull/4499/changes#r4095843633) for the rationale behind deleting it.

deployment/Dockerfile-rust-launcher
pbeza: JFYI: see [this](https://github.com/near/mpc/pull/4499/changes#r4095843633) for the rationale behind deleting it.

deployment/repro-sources-list.sh
pbeza: JFYI: see [this](https://github.com/near/mpc/pull/4499/changes#r4095843633) for the rationale behind deleting it.

pbeza (commented):
Sharing a few comments to make the PR easier to review.

.github/workflows/docker_build_node_gcp.yml
gilcu3: This part is a bit orthogonal to the change at hand. IIRC we were keeping the three steps so that it was possible to run then independently when needed. Don't think we lose much keeping it that way

PR #4501, opened 2026-09-22
feat: report duration of led computations

### LINKED ISSUE #4492: Report asset & signature computation durations
### Background

We don't have any visibility on the duration of individual asset generation computations.
A long time ago, we justified aborting computations that experienced network interruptions by stating that our computations take little time and are not worth the hassle of saving.

Since then, the network grew in nodes and our computation times likely increased. We would like data to validate our decision on.

Additionally, this should help us catch rough regressions.

### User Story

As a developer, I want data to base my decisions on.

### Acceptance Criteria

The leader of a computation reports the time it takes for a computation to succeed.
It would be useful if the reported metric includes the peers involved in the computation

We may want to chose a histogram type metric for this.

_No response_

kevindeforth:
Resolves #4492


kevindeforth:
@claude review

crates/node/src/metrics.rs
```diff
@@ -240,6 +243,24 @@ pub static MPC_NUM_CKD_COMPUTATIONS_LED: LazyLock<prometheus::IntCounterVec> = L
     },
 );
 
+fn buckets_covering_computation_deadlines() -> Vec<f64> {
+    vec![
+        0.05, 0.1, 0.2, 0.4, 0.8, 1.6, 3.2, 6.4, 12.8, 20.0, 30.0, 45.0, 60.0, 90.0, 120.0, 180.0,
+    ]
+}
```
gilcu3, line 250: this look like a lot of buckets. In general we are trying to reduce a bit the amount of metrics we store, to reduce "excessive" costs. Not saying this is too much, just commenting in case you think some of these are not really needed.

crates/node/src/metrics.rs
```diff
@@ -475,10 +496,81 @@ pub static PARTICIPANT_TOTAL_TIMES_SEEN_IN_FAILED_SIGNATURE_COMPUTATION_FOLLOWER
...
 
-pub const MPC_NUM_COMPUTATIONS_LED_TOTAL_LABEL: &str = "total";
-pub const MPC_NUM_COMPUTATIONS_LED_SUCCEEDED_LABEL: &str = "succeeded";
-pub const MPC_NUM_COMPUTATIONS_LED_FAILED_LABEL: &str = "failed";
-pub const MPC_NUM_COMPUTATIONS_LED_DEADLINE_EXCEEDED_LABEL: &str = "deadline_exceeded";
+pub const TOTAL_RESULT_LABEL: &str = "total";
+pub const SUCCEEDED_OUTCOME_LABEL: &str = "succeeded";
+pub const FAILED_OUTCOME_LABEL: &str = "failed";
+pub const DEADLINE_EXCEEDED_OUTCOME_LABEL: &str = "deadline_exceeded";
+pub const ABANDONED_OUTCOME_LABEL: &str = "abandoned";
```
gilcu3, line 504: this is a bit odd. Why the first is "result" while the rest is "outcome"? `TOTAL_LABEL` seems slightly better IMO

crates/node/src/metrics.rs
```diff
@@ -240,6 +243,25 @@ pub static MPC_NUM_CKD_COMPUTATIONS_LED: LazyLock<prometheus::IntCounterVec> = L
...
+    ]
+}
+
+pub static MPC_LED_COMPUTATION_DURATION_SECONDS: LazyLock<prometheus::HistogramVec> =
+    LazyLock::new(|| {
+        prometheus::register_histogram_vec!(
+            "mpc_led_computation_duration_seconds",
+            "Wall clock time a computation took, measured by the node that led it, including \
+             time spent waiting on the network. Outcomes deadline_exceeded and abandoned both \
+             mean the computation's work was lost; monitor them together.",
```
gilcu3, line 258: Independently of this it would be nice to open the issue about the equal nested timeouts so that we don't forget about it. But yeah this should be fine-ish

gilcu3 (approved):
Thank you!

crates/node/src/metrics.rs
kevindeforth: I think Claude asked for that. Had much fewer buckets in the beginning. I don't mind starting with many buckets and making it less once we know which ones are relevant.

pbeza (approved):

crates/node/src/metrics.rs
kevindeforth: done https://github.com/near/mpc/issues/4523

PR #4502, opened 2026-09-22
fix(node): prevent malicious channel ids

### LINKED ISSUE #4340: Predictable channel ids let one participant block every other leader

gilcu3:
Closes #4340


crates/node/src/network.rs
```diff
@@ -1385,20 +1404,125 @@ mod tests {
...
+
+        // Then
+        let mut channel = match opened.unwrap() {
+            SenderOrNewChannel::NewChannel(channel) => channel,
+            SenderOrNewChannel::Sender(_) => panic!("a fresh channel should have been opened"),
+        };
+        assert!(
+            tokio::time::timeout(Duration::from_millis(50), channel.receive())
+                .await
+                .is_err()
```
anodar: nit: instead of relying on receiving with 50ms timeout, you could probably just check that no channels are in use right after `node.receive` using helper you introduced (`channel_ids_in_use`).

anodar (dismissed):
Nice!

crates/node/src/network.rs
gilcu3: Ups after doing this change realized the test becomes a duplicate, so removed in [d854e2e](https://github.com/near/mpc/pull/4502/commits/d854e2e93f78ebe5e3f1c95fb7defcb9b1f86678)

anodar (approved):

netrome (approved):
Thanks 🙏

PR #4503, opened 2026-09-22
feat(node): support cait-sith with online presign

### LINKED ISSUE #4399: Enable the node to choose which Cait-Sith variant to use
### Background

As part of #4324 we introduced a new Cait-Sith alternative where presign and sign rounds happen during the online phase. We should integrate this in the node, while ensuring it is still possible to revert to the current protocol in case of issues in production. This can be achieved through a config option that allows us to select which protocol to use. 

We can use the protocol version handshake from https://github.com/near/mpc/blob/584fdee4f321194eb8dc8cc308391f9d587ab881/crates/node/src/protocol_version.rs#L11-L18 to let the leader node to figure if its peers already support the new protocol. That way we can support node rollbacks, although hopefully that's not needed.

### Acceptance Criteria

The node support ECDSA signatures using the new Cait-Sith variant. In case of issues, a config flag can be used to revert back to the existing behavior. 

[thread](https://nearone.slack.com/archives/C0912BTG51T/p1789385586293439?thread_ts=1789378451.169369&cid=C0912BTG51T)

gilcu3:
Closes #4399 

There is some amount of code duplication in `crates/node/src/providers/ecdsa/sign.rs`, which could be removed while increasing the diff. I don't think that's very useful as we anyway plan to remove the current CaitSith implementation later. So to make reviewing this simpler, I opted for not doing that refactor


crates/node/src/providers/ecdsa/sign.rs
```diff
@@ -83,6 +113,123 @@ impl EcdsaSignatureProvider {
...
+        keyshare: &EcdsaKeyshare,
+    ) -> anyhow::Result<Option<(UniqueId, PairedTriple)>> {
+        let supporting = self
+            .client
+            .participants_supporting(ONLINE_PRESIGN_MIN_PROTOCOL_VERSION);
+        Ok(self
+            .triple_store_for_t(keyshare.reconstruction_threshold)?
+            .take_owned_matching(supporting)
+            .now_or_never())
+    }
```
gilcu3: this is the complicated part of the change, and the probably the only one that is hard to review.

During the upgrade period, we will have a situation similar to what we had in foreign chain transactions, where some peers are alive, but not all of them support the new protocol. The proposed solution, we check if we already have a triple for the supporting and alive peers. If not we fallback to the existing protocol.

I tried many approaches, all with caveats and complexity problems, until settling on the one above which seems simplest and safest. If you find another one let me know and I will try to break it 😄

crates/node/src/providers/ecdsa/sign.rs
```diff
@@ -256,3 +403,227 @@ impl MpcLeaderCentricComputation<()> for FollowerSignComputation {
...
+            const THRESHOLD: usize = 3;
+            let mut rng = StdRng::seed_from_u64(42);
+            let ts_participants = generate_participants(THRESHOLD);
+            let participants = into_participant_ids(&ts_participants);
+            let leader = *participants.iter().min().unwrap();
+            let (keygen_output, shares) = deal_shares(&ts_participants, THRESHOLD, &mut rng);
+            let shares = Arc::new(shares);
+
+            // When
+            let results = run_test_clients(participants.clone(), move |client, mut receiver| {
```
gilcu3: we didn't have such tests here, but given that adding them is almost free now, they seem useful to keep

crates/node/src/tests/basic_cluster.rs
```diff
@@ -81,6 +102,8 @@ async fn test_basic_cluster() {
         .await
         .expect("timeout waiting for keygen to complete");
 
+    wait_until_positive(&MPC_OWNED_NUM_TRIPLES_AVAILABLE, "triples").await;
```
anodar: Isn't this metric shared among the nodes here? That is, it asserts that some node has triples available.

So this could probably become flaky when leader doesn't have triples available and reverts to legacy path.

crates/node/src/tests/basic_cluster.rs
```diff
@@ -125,3 +154,72 @@ async fn test_basic_cluster() {
...
+        .collect::<Vec<_>>();
+    setup
+        .indexer
+        .wait_for_contract_state(
+            |state| matches!(state, ContractState::Running(_)),
+            DEFAULT_MAX_PROTOCOL_WAIT_TIME,
+        )
+        .await
+        .expect("timeout waiting for keygen to complete");
+    wait_until_positive(&MPC_OWNED_NUM_PRESIGNATURES_AVAILABLE, "presignatures").await;
```
anodar: I think this might become an issue that we are using metrics in these tests, because nodes are spawned in the same process (using `tokio::spawn(config.run())`) and share the metrics.

And I don't mean just within a single test, but if we do `cargo test -p mpc-node` I think it executes tests in parallel in the same process, so now `test_basic_cluster` and `cluster__should_sign_from_stored_presignatures_when_online_presign_is_disabled` will have races on metrics.

crates/node/src/providers/ecdsa.rs
```diff
@@ -209,6 +214,19 @@ impl SignatureProvider for EcdsaSignatureProvider {
                         .instrument(self.make_signature_follower(channel, id, presignature_id))
                         .await?;
                 }
+                EcdsaTaskId::OnlinePresignSignature {
+                    id,
+                    paired_triple_id,
+                } => {
+                    ECDSA_TASK_MONITORS
```
anodar: Not sure if it's intentional, but while we instrument `make_online_presign_signature_follower` to get separate metrics from legacy and new follower functions, we don't differentiate `make_signature_leader` part (as it's instrumented outside that and new or legacy implementation happen conditionally there).

We could probably instrument it inside `make_signature_leader` if we wanted to differentiate those two.

anodar (approved):

crates/node/src/tests/basic_cluster.rs
gilcu3: yeah, that's the case. I have not seen any easy way around it though. The flakiness does not happen in practice because most nodes start computing triples at the same time, but it is certainly possible

crates/node/src/tests/basic_cluster.rs
gilcu3: That I think does not happen, because we have been using metrics like this for a while and I have not observed such inter-test failures. It would be a very interesting find if they do share metrics but tests still don't fail

crates/node/src/providers/ecdsa.rs
gilcu3: I don't think we need to differentiate them (tokio instrumentation metrics) because the transition period between one and the other should not be too long and we don't use them frequently anyway.

crates/node/src/tests/basic_cluster.rs
anodar: >  I have not observed such inter-test failures

Are you referring to CI ? Because that works differently than cargo test[0]:
"A key factor distinguishing nextest from cargo test is that nextest runs each test in a separate process. With nextest, the default execution model is now, and will always be, process-per-test."

Cargo test runs tests in same process.  I guess we can just say that tests should be run with nextest even locally and that solves the problem.

[0] https://nexte.st/docs/design/why-process-per-test/#why-process-per-test

gilcu3: right! yeah nextest is the savior indeed. Luckily I had been nagging about always refering to it for a while, albeit it was only for uniformity

PR #4520, opened 2026-09-23
feat(contract-interface): add the Update DTO, its hash and the submit deposit

### LINKED ISSUE #4511: Contract code update votes: Add new methods and structs to interface
### Background

Part of https://github.com/near/mpc/issues/1674

We want need to add methods and structs to the contract interface.

### User Story

As a developer, I like reviewing small-scoped PRs

### Acceptance Criteria

Add the required method and data structs to the contract interface.

_No response_

kevindeforth:
Resolves #4511 

This is part of a bigger effort to replace the contract upgrade logic and adds the necessary interface types.

Note: right now, the added endpoints would all fail. That's okay, we don't use them anywhere.


crates/near-mpc-contract-interface/src/method_names.rs
```diff
@@ -32,16 +32,23 @@ pub const VOTE_UPDATE_FOREIGN_CHAIN_PROVIDERS: &str = "vote_update_foreign_chain
...
 
 // Protocol management
 pub const INIT: &str = "init";
 pub const INIT_RUNNING: &str = "init_running";
 pub const MIGRATE: &str = "migrate";
 pub const START_KEYGEN_INSTANCE: &str = "start_keygen_instance";
 pub const START_RESHARE_INSTANCE: &str = "start_reshare_instance";
+// TODO(#4513): drop once production runs the vote-then-submit API.
 pub const PROPOSE_UPDATE: &str = "propose_update";
+pub const SUBMIT_UPDATE: &str = "submit_update";
```
kevindeforth: oops, need to rename this to `submit_contract_update` to stay consistent.

crates/contract/Cargo.toml
```diff
@@ -135,6 +135,8 @@ near-contract-transport = { workspace = true, features = ["traits"] }
 near-jsonrpc-client = { workspace = true }
 near-jsonrpc-primitives = { workspace = true }
 near-mpc-contract-interface = { workspace = true, features = ["client"] }
+# Tests run off-chain, so they may hash an update without the host function.
+near-mpc-sdk = { workspace = true, features = ["non-contract-usage"] }
```
gilcu3: I don't understand the comment. It seems to be implying that to hash something in tests we need `near-mpc-sdk` but I just saw a line using Sha2 directly in `crates/contract/tests/sandbox/update_votes_cleanup_after_resharing.rs`

crates/near-mpc-contract-interface/src/types/updates.rs
```diff
@@ -1,11 +1,31 @@
...
+
+impl Update {
+    pub fn payload_bytes(&self) -> Result<u128, PayloadBytesError> {
+        let bytes = match self {
+            Update::Code(code) => code.len(),
+            Update::Config(config) => serde_json::to_vec(config)?.len(),
+        };
+        u128::try_from(bytes).map_err(|_| PayloadBytesError::Overflow)
+    }
+}
```
gilcu3: hmm this is the type of function that would grant this not existing in the contract interface. Do we need it elsewhere, or only in the contract? If only in the contract it should be defined there, if also elsewhere I think we agreed last time it would go to mpc-sdk

crates/near-mpc-contract-interface/src/client.rs
```diff
@@ -64,7 +66,10 @@ pub const VOTE_MPC_NODE_MANIFEST_DIGEST_GAS: NearGas = NearGas::from_tgas(300);
 /// TODO(#1571): Gas cost for voting on contract updates. Reduced somewhat after
 /// optimization (#1617) by avoiding full contract code deserialization; there’s likely still
 /// room for further optimization.
+// TODO(#4513): drop once production runs the vote-then-submit API.
 pub const VOTE_UPDATE_GAS: NearGas = NearGas::from_tgas(260);
+// TODO(#166): not benchmarked.
+pub const VOTE_CONTRACT_UPDATE_GAS: NearGas = NearGas::from_tgas(22);
```
gilcu3: hmm #166 seems very general, not sure attaching that todo in this line is helpful

crates/near-mpc-contract-interface/src/method_names.rs
```diff
@@ -80,7 +87,9 @@ pub const PUBLIC_KEY: &str = "public_key";
 pub const DERIVED_PUBLIC_KEY: &str = "derived_public_key";
 pub const VERSION: &str = "version";
 pub const LATEST_KEY_VERSION: &str = "latest_key_version";
+// TODO(#4513): drop once production runs the vote-then-submit API.
 pub const PROPOSED_UPDATES: &str = "proposed_updates";
```
gilcu3: I think we will still need this one, listing the existing proposed updates hashes, or not?

crates/contract/tests/sandbox/update_votes_cleanup_after_resharing.rs
```diff
@@ -259,7 +259,9 @@ pub fn assert_expected_proposed_update(
     let mut expected_updates_map = BTreeMap::new();
     expected_updates_map.insert(
         expected_update_id,
-        dtos::UpdateHash::Code(sha2::Sha256::digest(expected_update_code).into()),
+        dtos::UpdateHash::Code(dtos::Hash256(
+            sha2::Sha256::digest(expected_update_code).into(),
+        )),
```
gilcu3: Ahh realized the problem now. This should use `near-mpc-sdk` instead

gilcu3 (changes requested):
Thank you!

Small change requested, I think the change in `crates/contract/tests/sandbox/update_votes_cleanup_after_resharing.rs` needs some fixing.

Also I am a bit concerned about seeing the `payload_bytes` implementation in the contract interface. IMO that should probably go either to mpc-sdk or contract, but cannot say without seeing the rest of the PRs stack

crates/near-mpc-contract-interface/src/method_names.rs
kevindeforth: No, it will be superseded by `CONTRACT_UPDATE_VOTES`, just below.

crates/near-mpc-contract-interface/src/types/updates.rs
kevindeforth: True. We only use it as a helper method in `call_args`, so I will move it there.

crates/contract/tests/sandbox/update_votes_cleanup_after_resharing.rs
kevindeforth: We are still testing the old logic here. This test needed to be modified because this PR changes the `UpdateHash` payload to `Hash256`.

But I don't want this PR to carry any other behavior changes wrt existing voting infrastructure. FWIW, this line will go once the integration is completed.

crates/near-mpc-contract-interface/src/client.rs
kevindeforth: I think it helps to keep track it doesn't cost much. `22` looks deliberate and IIRC, some of these values have been benchmarked.

crates/contract/Cargo.toml
kevindeforth: That must have slipped in. Shouldn't be in this PR. Will revert.

_edit: removed in https://github.com/near/mpc/pull/4520/commits/45f98cb686839a66f992abaf1ceb15b2bac3d7b0_

crates/near-mpc-contract-interface/src/client.rs
gilcu3: > some of these values have been benchmarked.

that kind of contradict the comment you are adding. Indeed I think this 22 was benchmarked

crates/contract/tests/sandbox/update_votes_cleanup_after_resharing.rs
gilcu3: Ok, I see the dep was removed, that also works

gilcu3 (approved):

PR #4524, opened 2026-09-23
chore: update aptos tx in mainnet config for foreign-chain-config-tester

anodar:
Stop gap measure before we have https://github.com/near/mpc/issues/3969


gilcu3 (approved):
LGTM

barakeinav1 (approved):

